PR #1841, opened 2018-09-04
peer: Synchronous registration of gossip queries

cfromknecht:
This commit moves the gossip sync dispatch
such that it is more tightly coupled to the
life cycle of the peer. In testing, I noticed
that the gossip syncer needs to be dispatched
before the first gossip messages come across
the wire.

The prior spawn location in the server happens
after starting all of the peer's goroutines,
which could permit an ordering where the
gossip syncer has not yet been registered.
The new location registers the gossip syncer
within the read handler such that the call
blocks before any messages are read.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -35,6 +35,11 @@ var (
 	// ErrGossiperShuttingDown is an error that is returned if the gossiper
 	// is in the process of being shut down.
 	ErrGossiperShuttingDown = errors.New("gossiper is shutting down")
+
+	// ErrGossipSyncerNotFound signals that we were unable to find an active
+	// gossip syncer corresponding to a gossip query message received from
+	// the remote peer.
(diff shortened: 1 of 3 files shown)


Roasbeef:
The gossiper already handles this case here: https://github.com/lightningnetwork/lnd/blob/master/discovery/gossiper.go#L903

When it gets a new gossip sync message, it tries to find the existing syncer. If it's able to find it, then it'll use that directly, otherwise, it'll make a new one.

cfromknecht:
Hmm, then why have `InitSyncState` at all?

cfromknecht:
The fallback in `fetchGossipSyncer` doesn't properly honor `cfg.NoChanUpdates`.

cfromknecht:
@Roasbeef added a commit to remove the fallback since it is no longer required, PTAL

PR #1843, opened 2018-09-05
Start deprecating noencryptwallet

cfromknecht:
This PR begins the process of deprecating the confusing `noencryptwallet` flag, which has initially designed for testing purposes. As people have started to use this flag in the wild, it cannot be fully removed in one fell swoop due to backwards-compatibility requirements, though this will start the process of deprecating its use.

The primary change in this PR is to rename the flag to `noseedbackup`, as this the more critical information the user should be aware of when using this setting. The help description printed presents a more alarming tone than the prior message, and includes the caveat that no seed will be provided.

References to noencryptwallet have been migrated to noseedbackup in comments throughout the codebase. The section on disabling wallet encryption has also been removed from the install docs and `sample-lnd.conf`. Lastly, the docker start script has been changed to pass the `noseedbackup` flag instead.

NOTE: Any users that are actively using `noencryptwallet` will have to switch any scripts/confs to use`noseedbackup` as a result of this PR, though no further modification should be required.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -1332,7 +1332,7 @@ var unlockCommand = cli.Command{
 	The unlock command is used to decrypt lnd's wallet state in order to
 	start up. This command MUST be run after booting up lnd before it's
 	able to carry out its duties. An exception is if a user is running with
-	--noencryptwallet, then a default passphrase will be used.
+	--noseedbackup, then a default passphrase will be used.
 	`,
 	Flags: []cli.Flag{
 		cli.IntFlag{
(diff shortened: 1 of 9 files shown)


NicolasDorier:
You should add a proper error message in case a user try to use `noencryptwallet` because of outdated documentation.

Roasbeef:
It'll fail to start as is.

On Sep 13, 2018 1:40 PM, "Nicolas Dorier" <notifications@github.com> wrote:

You should add a proper error message in case a user try to use
noencryptwallet because of outdated documentation.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/lightningnetwork/lnd/pull/1843#issuecomment-421145569>,
or mute the thread

dannypaz:
@Roasbeef fails to start with something like `lnd_btc_1  | /home/lnd/lnd.conf:18: unknown option: noencryptwallet` which helps, but I think @NicolasDorier is asking for a deprecation message instead of a generic one.

PR #1844, opened 2018-09-05
peer: prune persistent peer connection on zero on-disk channels

wpaulino:
In this commit, we fix a small bug with regards to the persistent peer
connection pruning logic. Before this commit, it'd be the case that we'd
prune a persistent connection to a peer if all links happen to be
inactive. This isn't ideal, as the channels are still open, so we should
always be atttempting to connect to them. We fix this by looking at the
set of channels on-disk instead and prune the persistent connection if
there aren't any.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -299,11 +299,16 @@ func (p *peer) Start() error {
 		return err
 	}
 
+	if len(activeChans) == 0 {
+		p.server.prunePersistentPeerConnection(p.pubKeyBytes)
+	}
+
 	// Next, load all the active channels we have with this peer,
 	// registering them with the switch and launching the necessary
 	// goroutines required to operate them.
 	peerLog.Debugf("Loaded %v active channels from database with "+
 		"NodeKey(%x)", len(activeChans), p.PubKey())
+
 	if err := p.loadActiveChannels(activeChans); err != nil {
 		return fmt.Errorf("unable to load channels: %v", err)
 	}
@@ -1993,6 +1998,7 @@ func (p *peer) finalizeChanClosure(chanCloser *channelCloser) {
 
 	go waitForChanToClose(chanCloser.negotiationHeight, notifier, errChan,
 		chanPoint, &closingTxid, closingTx.TxOut[0].PkScript, func() {
+
 			// Respond to the local subsystem which requested the
 			// channel closure.
 			if closeReq != nil {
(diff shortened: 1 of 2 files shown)


peer.go
```diff
@@ -1961,12 +1968,20 @@ func (p *peer) finalizeChanClosure(chanCloser *channelCloser) {
...
 			// Remove the persistent connection to this peer if we
-			// no longer have open channels with them.
-			p.activeChanMtx.Lock()
-			numActiveChans := len(p.activeChannels)
-			p.activeChanMtx.Unlock()
+			// no longer have open channels with them. We'll make
+			// sure to look at the on-disk state to ensure we don't
+			// prune the connection just becasue there aren't any
+			// _active_ links at the moment.
+			openChannels, err := p.server.chanDB.FetchOpenChannels(
```
cfromknecht: on second thought, do we wanna pull this into `WipeChannel`?

Roasbeef: Yeah this should be pulled into wipe channel.

Roasbeef: Other relevant area: https://github.com/lightningnetwork/lnd/blob/b31884a0e21860b471ae69953dcd1af3dd7e8deb/peer.go#L2039

wpaulino: Since we're now looking at the on-disk state, we should ensure that we attempt to fetch the set of open channels only _after_ the channel has been marked as closed by the `ChannelArbitrator` within the database. This happens on the dispatch of a spend notification. Pulling this into `WipeChannel` would not work as we can't guarantee this has already happened by then.

One approach is to thread down a callback from the server to the `ChannelArbitrator` to ensure this is done once the channel has been marked as closed within the database.

wpaulino: Not sure how the peer exiting is relevant to this? Pruning the persistent connection just removes the persistent connection logic, it doesn't disconnect the peer.

I'm specifically thinking of the following scenario: we receive a link error from the switch, so we'll call `WipeChannel` and then attempt to force close the channel. If this is the only channel with this peer, then `WipeChannel` will retrieve it when fetching the set of open channels, which will end up not pruning the persistent connection.

If the server restarts, sure, we won't connect to them anymore, but this handles the case where we don't restart at all.

cfromknecht: in that case we could reorder wipe channel to be after the force close? force closing should remove the link from the switch either way, so would seem safe

cfromknecht: > Not sure how the peer exiting is relevant to this?

The quit chan at top breaks out if the peer has already exited. Can this lead to a situation where the callback isn't executed if the peer quits?

wpaulino: Well we still need to wait for the confirmed spend notification in order to mark it as closed, so I don't think that'd work either as the process is asynchronous. It looks like the force close call will just return once the closing transaction has been crafted/broadcasted.

cfromknecht: What i'm saying is, there's no explicit requirement that it has to be perfectly synchronized with the channel closure. We just need to ensure that if a close happens, `prunePersistentPeers` is reliably called _at some point_ afterwards. For example, we could have something that prunes persistent connection on peer start, if it detects that there are no active channels. The channel could have closed w/in scope of the last peer, but we don't really care as long as we eventually mark this is a non-perm peer.

PR #1592, opened 2018-07-20
travis - avoid deep compare error

offerm:
This problem will show up when running TestLightningWireProtocol when the random number of numChanIDs is zero. This problem is depends on the random function so we will see it only from time to time,

to simulate the problem just add

    numChanIDs = 0

    after line 701 at lnwire_test.go (so you force the random number to bu zero) to make it:

    			numChanIDs := rand.Int31n(5000)
    			numChanIDs = 0

    			req.ShortChanIDs = make([]ShortChannelID, numChanIDs)
    and run the test you will see the problem.

when running the tests you will get:

	lnwire_test.go:231: messages don't match after re-encoding: (*lnwire.ReplyChannelRange)(0xc420226230)({
		 QueryChannelRange: (lnwire.QueryChannelRange) {
		  ChainHash: (chainhash.Hash) (len=32 cap=32) b4d468adaa51fd220dad7373cb0442ca6ac4978671d01e63adececfaff69624d,
		  FirstBlockHeight: (uint32) 647747300,
		  NumBlocks: (uint32) 438264374
		 },
		 Complete: (uint8) 0,
		 EncodingType: (lnwire.ShortChanIDEncoding) 1,
		 ShortChanIDs: ([]lnwire.ShortChannelID) {
		 }
		})
		 vs (*lnwire.ReplyChannelRange)(0xc420226000)({
		 QueryChannelRange: (lnwire.QueryChannelRange) {
		  ChainHash: (chainhash.Hash) (len=32 cap=32) b4d468adaa51fd220dad7373cb0442ca6ac4978671d01e63adececfaff69624d,
		  FirstBlockHeight: (uint32) 647747300,
		  NumBlocks: (uint32) 438264374
		 },
		 Complete: (uint8) 0,
		 EncodingType: (lnwire.ShortChanIDEncoding) 1,
		 ShortChanIDs: ([]lnwire.ShortChannelID) <nil>
		})

Process finished with exit code 1

PR #1853, opened 2018-09-06
lnwire/lnwire_test: assert nil 0-length sid reply msg

cfromknecht:
Thank you to @offerm for pointing this out, our tests would incorrectly allocate a zero-length short chan id array when checking via quick tests.

Replaces #1592


offerm:
@cfromknecht much better fix. Don't know how I missed the difference between `MsgQueryShortChanIDs` and `MsgReplyChannelRange`. Clearly, this is the way to go!

PR #1855, opened 2018-09-06
Add amt_paid_msat to invoice

halseth:
Adds back a field for `msat` amount paid, which was removed in #1846.

Note that the original tag number for the `msat` field is restored, for backwards compatability.


cfromknecht:
Should be noted that any LND users that do mission-critical computation w/ amtpaid field will now be off by a factor of 1000. Makes me wonder if we should just add `amt_paid_sat` and `amt_paid_msat`, keeping `amd_paid` in msat for backwards compatibility.

halseth:
We could do that, and deprecate the old field.

cfromknecht:
@Roasbeef yes i'm saying we should revert the changes from two days ago, and add new fields with specified units. The old one can be deprecated going forward.

> I suppose it's up to them to detect a change like this

I don't think we can expect this of our users, nor should they have to monitor for this. Users exporting this field into an external db will have a column with inconsistent units. The change is not documented in release notes--are we expecting them to keep up actively with commits and change downstream logic on the fly?

lnrpc/rpc.proto
```diff
@@ -1669,15 +1669,21 @@ message Invoice {
...
+    int64 amt_paid = 18 [json_name = "amt_paid", deprecated = true];
+
     /**
-    The amount that was accepted for this invoice. This will ONLY be set if
-    this invoice has been settled. We provide this field as if the invoice was
-    created with a zero value, then we need to record what amount was
-    ultimately accepted. Additionally, it's possible that the sender paid MORE
-    that was specified in the original invoice. So we'll record that here as
-    well.
+    The amount that was accepted for this invoice, in satoshis. This will ONLY
```
cfromknecht: Great comment :) can we add the same thing w/ different units on `amt_paid_msat`?

halseth: Done!

PR #1856, opened 2018-09-06
multi: fix various typos in comments

maurycy:


maurycy:
The checks here do not make any sense: no functional changes.

PR #1295, opened 2018-05-29
rpc: add support for a `forgetchannel` RPC only enabled w/ the debug flag

joostjager:
Implements https://github.com/lightningnetwork/lnd/issues/1225

PR #1857, opened 2018-09-06
channeldb: clean out borked channels [WIP]

joostjager:
This PR builds on #1295. #1295 adds functionality to abandon borked channels when `lnd` is running.

The intention of this PR is to add functionality to clean borked channels very early in the startup process. 

The current implementation is to perform the cleaning as a database migration. On a second thought, this may not be the best idea. The code depends quite heavily on serialize/deserialize methods and future changes to those will break the migration.

A further step is to trigger the clean up code not from a migration, but to put it behind a command line flag that is only available in debug builds. That way it will always deal with the latest version of the database and we won't risk breaking the logic in the future.


joostjager:
Within this PR, related state in the contractcourt, htlcswitch, and utxonursery can be cleaned up too, see @Roasbeef  comment in #1295.

Advantage is that is doesn't need to be done "hot".

joostjager:
Closing for now. Issue needs more consideration as to what the best way is to deal with this

Issue #1854, opened 2018-09-06
Handle starting LND with unsynced chain backend.


alevchuk:
### Background

LND fails to start on Litecoind testnet

### Your environment

* lnd version 0.5.0-beta commit=baee07ef480c15a731e0c87dc98d27269c5a80fa
* Linux 3.16.0-6-amd64 #1 SMP Debian 3.16.56-1+deb8u1 (2018-05-08) x86_64 GNU/Linux
* Litecoin Core Daemon version v0.16.2  eba0e1ca3fb4088998461b02ec6831858cc24144
```
~/.litecoin/litecoin.conf
testnet=1
txindex=1
server=1
; daemon=1
debug=1
discardfee=0.00000001
mintxfee=0.00000001
minrelaytxfee=0.00000001
rpcuser=jason
rpcpassword=litecoin
zmqpubrawblock=tcp://127.0.0.1:28332
zmqpubrawtx=tcp://127.0.0.1:28333
```

```
tail -n1000 ~/.litecoin/testnet4/debug.log |grep UpdateTip   | awk '{print $1, $2, $6, $10, $11, $12}'
2018-09-06 03:38:31 height=724100 date='2018-09-06 03:38:17' progress=1.000000
2018-09-06 03:40:34 height=724101 date='2018-09-06 03:40:21' progress=1.000000
2018-09-06 03:41:07 height=724102 date='2018-09-06 03:40:34' progress=1.000000
2018-09-06 03:50:05 height=724103 date='2018-09-06 03:49:48' progress=1.000000
2018-09-06 03:54:18 height=724104 date='2018-09-06 03:54:05' progress=1.000000
```

### Steps to reproduce

```
lnd --externalip=$(dig +short myip.opendns.com @resolver1.opendns.com)
```

### Expected behaviour

Start normal

### Actual behaviour

```
2018-09-05 20:51:18.488 [INF] LTND: Version 0.5.0-beta commit=baee07ef480c15a731e0c87dc98d27269c5a80fa
2018-09-05 20:51:18.489 [INF] LTND: Active chain: Litecoin (network=testnet)
2018-09-05 20:51:18.489 [INF] CHDB: Checking for schema update: latest_version=6, db_version=6
2018-09-05 20:51:18.506 [INF] RPCS: password gRPC proxy started at 127.0.0.1:8080
2018-09-05 20:51:18.507 [INF] RPCS: password RPC server listening on 127.0.0.1:10009
[...]

PR #1858, opened 2018-09-06
Add debug log to Auto Pilot on why no more channels are needed

alevchuk:
This is to address confusion I ran into in issue #1854

A test was to run LND with debug log enabled and 0 funds in the chain wallet:
```
2018-09-06 10:50:49.117 [INF] ATPL: Autopilot Agent starting
2018-09-06 10:50:49.119 [DBG] ATPL: Applying external balance state update
2018-09-06 10:50:49.123 [DBG] ATPL: Pending channels: (map[autopilot.NodeID]autopilot.Channel) {
}

2018-09-06 10:50:49.124 [DBG] ATPL: No more channels needed. Funds allocation reached. Total channel allocation: 0 BTC. Total funds: 0 BTC
```

### DIFF
--- a/autopilot/agent_constraints.go
+++ b/autopilot/agent_constraints.go
@@ -88,6 +88,9 @@ func (h *agentConstraints) ChannelBudget(channels []Channel,
 	// If we're already over our maximum allowed number of channels, then
 	// we'll instruct the controller not to create any more channels.
 	if len(channels) >= int(h.chanLimit) {
+                log.Debugf("No more channels needed because channel limit was reached. " +
+                        "Limit: %v. Current number of channels: %v",
+                        h.chanLimit, len(channels))
 		return 0, 0
 	}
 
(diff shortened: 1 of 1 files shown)


alevchuk:
Take a look when you get a chance. It’s a 2-line change, easy to verify.

halseth:
Needs to be tested to figure out if logs get too spammy.

alevchuk:
I ran this in September and the debug log was printed once every few minutes. Would you like me to re-test rebased on latest?

Roasbeef:
Yeah an update sample run would be useful.

alevchuk:
Rebased and tested. Here is the sample run: https://gist.github.com/alevchuk/6a94593ebec4db8800f96f9cd9c1becc#file-gistfile1-txt-L12-L20

^ shows that in some cases it prints multiple times a second. So no go. What do you want to do with this?

halseth:
Multiple times a second is def too spammy. I would either make trace or add some sort of check to ensure we only print after a certain time has passed.

PR #1859, opened 2018-09-06
add C# gRPC doc

BrettCawley:


Roasbeef:
cc @joostjager

docs/grpc/c#.md
```diff
@@ -0,0 +1,219 @@
...
+
+IEnumerable<SendRequest> SendPayment()
+{
+    while (true)
+    {
+        SendRequest request = new SendRequest() {
+            DestString = "<DEST_PUB_KEY>",
+            Amt = 100,
+            PaymentHashString = "<R_HASH>",
+            FinalCltvDelta = "<CLTV_DELTA>"
```
joostjager: Remove quotes around `FinalCltvDelta` value.

docs/grpc/c#.md
```diff
@@ -0,0 +1,219 @@
...
+        };
+        yield return request;
+        System.Threading.Thread.Sleep(2000);
+    }
+}
+```
+This example will send a payment of 100 satoshis every 2 seconds.
```
joostjager: The idea is clear. However the example will not work completely as expected because `lnd` is currently checking for invoices that are already paid. So only the first payment succeeds.

docs/grpc/c#.md
```diff
@@ -0,0 +1,219 @@
...
+
+// combine the cert credentials and the macaroon auth credentials using interceptors
+// so every call is properly encrypted and authenticated
+Task AddMacaroon(AuthInterceptorContext context, Metadata metadata)
+{
+    metadata.Add(new Metadata.Entry("macaroon", macaroon));
+    return Task.CompletedTask;
+}
+var macaroonInterceptor = new AsyncAuthInterceptor(AddMacaroon);
+var combinedCreds = ChannelCredentials.Create(sslCreds, CallCredentials.FromInterceptor(macaroonInterceptor));
```
joostjager: Works :+1:

docs/grpc/c#.md
```diff
@@ -0,0 +1,219 @@
...
+
+* Fetch the `nuget` executable (Windows is used in this example), then install the `Grpc.Tools` package:
+```bash    
+curl -o nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
+./nuget install Grpc.Tools
+```
+
+* `cd` to the location of the `protoc.exe` tool:  (note the version name in the folder)
+```bash
+cd Grpc.Tools.X.XX.X/tools/windows_x86
```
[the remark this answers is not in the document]
BrettCawley: It seems to be necessary to copy the files to the directory, imports wont work as files are expected to be in a certain directory. Will update the command to create output files at the base of the folder.

docs/grpc/c#.md
```diff
@@ -0,0 +1,219 @@
...
+Console.WriteLine(response);
+```
+
+#### Response-streaming RPC
+
+```c#
+var request = new InvoiceSubscription();
+using (var call = client.SubscribeInvoices(request))
+{
+    while (await call.ResponseStream.MoveNext())
```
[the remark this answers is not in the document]
BrettCawley: Can definitely do that, but I thought it better to always get the latest versions as this project progresses so quickly! Better to have docs that are 95% correct than ones that are quickly out of date, in my opinion. Your thoughts?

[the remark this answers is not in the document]
joostjager: I don't think the `protoc` version is moving that quickly. I think for example code, it doesn't hurt to add enough details so that a user can run it. Maybe accompanied by a note that says that actual lnd may already be on a higher version?

docs/grpc/c#.md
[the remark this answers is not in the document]
joostjager: What I was thinking about, is that the `packages` dir is managed by nuget. It isn't really meant to write anything into manually. Just wondering if you can read `rpc.proto` from the source dir and write the output to the source dir as well.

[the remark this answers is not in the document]
BrettCawley: I understand what you mean - Changed to add files to the root of the project

docs/grpc/c#.md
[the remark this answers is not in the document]
BrettCawley: can't reproduce on windows using .net core. `MoveNext()` has a signature with and without a cancellation token. Maybe the async context caused the issue

docs/grpc/c#.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Setup and Installation
+
+`lnd` uses the `gRPC` protocol for communication with clients like `lncli`.
+
+`gRPC` is based on protocol buffers, and as such, you will need to compile the `lnd` proto file in C# before you can use it to communicate with `lnd`.
+
+This assumes you are using a Windows machine, but it applies equally to Mac and Linux.
+
+* Open Visual Studio, and create a new `.net core` console application called `lnrpc` at the root directory (Windows : `C:/`)
```
joostjager: Now that you switched to `.net core`, you could just as well list the `dotnet` cli commands here. That way visual studio is not a requirement anymore, allowing the steps to be followed without improvisation on mac/linux. 
```
mkdir lndclient
cd lndclient
dotnet new console
dotnet add package Grpc.Tools --version 1.17.0
```

docs/grpc/c#.md
```diff
@@ -0,0 +1,217 @@
...
+
+* Copy Google's [descriptor.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto) to the correct folder:
+```bash
+mkdir Grpc/google/protobuf
+curl -o Grpc/google/protobuf/descriptor.proto -s https://raw.githubusercontent.com/protocolbuffers/protobuf/master/src/google/protobuf/descriptor.proto
+```
+
+* Compile the proto file using `protoc.exe` from nuget package `Grpc.Tools` (remember to replace "YOUR_USER", and possibly version "1.17.0" in both paths):
+```bash
+# linux + mac nuget package location: ~/.nuget/packages
```
joostjager: and replace os reference in command line

I needed to do: (last tilde wasn't expanded)
`~/.nuget/packages/grpc.tools/1.17.0/tools/linux_x64/protoc --csharp_out . --grpc_out . rpc.proto --plugin=protoc-gen-grpc=/home/joost/.nuget/packages/grpc.tools/1.17.0/tools/linux_x64/grpc_csharp_plugin`

docs/grpc/c#.md
```diff
@@ -0,0 +1,217 @@
...
+
+After installing these, use the code below to set up a channel and client to connect to your `lnd` node:
+
+```c#
+
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Lnrpc;
```
joostjager: Also needed to do:
`dotnet add package Google.Protobuf --version 3.6.1`

joostjager: And `dotnet add package Grpc.Core --version 1.17.0`

joostjager: And `dotnet add package Google.Api.CommonProtos --version 1.4.0`

Don't know if all these separately downloade annotation protos are also somewhere in nuget?

docs/grpc/c#.md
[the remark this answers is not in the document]
joostjager: Can't reproduce it anymore myself. Probably we had different version previously and not anymore now.

Roasbeef:
So it does indeed compile? 🤔

joostjager:
I rebased this onto master and went through the steps and examples again, on Linux.

Because there is no full example code, some modifications are needed. But this is limited to adding `using` statements and changing paths. With that, all examples work.

@Roasbeef I'd recommend to merge this as it is now and possibly amend a bit if things prove to be unclear when people start using this.

PR #1861, opened 2018-09-06
server: ensure persistent connections to tor addrs are made if active

wpaulino:
In this commit, we restrict the persistent connection logic on startup
to only attempt to establish connections to Tor addresses if Tor
outbound support is enabled. Otherwise, we'll continually attempt to
reach the address even though we never will.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1555,14 +1555,13 @@ type nodeAddresses struct {
 }
 
 // establishPersistentConnections attempts to establish persistent connections
-// to all our direct channel collaborators.  In order to promote liveness of
-// our active channels, we instruct the connection manager to attempt to
-// establish and maintain persistent connections to all our direct channel
-// counterparties.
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -1614,7 +1620,12 @@ func (s *server) establishPersistentConnections() error {
 		if ok {
 			for _, lnAddress := range linkNodeAddrs.addresses {
 				switch lnAddress.(type) {
-				case *net.TCPAddr, *tor.OnionAddr:
+				case *net.TCPAddr:
+					addrSet[lnAddress.String()] = lnAddress
+
+				// We'll only attempt to connect to Tor
+				// addresses if Tor outbound support is enabled.
+				case *tor.OnionAddr:
```
cfromknecht: should this also check `cfg.Tor.Active`?

PR #1763, opened 2018-08-22
channeldb+rpc: add pagination of listinvoices

valentinewallace:
In this PR, we introduce support for pagination of invoices at the RPC level. We do this by using a cursor on the invoice bucket, seeking until we have enough results as specified by `--max_invoices` or we have no more invoices to iterate through. 

Fixes #1514.
Replaces #1614 and #1720.

PR #1863, opened 2018-09-07
channeldb+cmd/lncli: enable backwards pagination of listinvoices

### LINKED ISSUE #1852: Q: listinvoices after 0.5 update returns old not newly created invoices
### Background

lncli --network=testnet listinvoices

returns old invoices, does not show a newly created invoice 

Looks like something was changed and i need to remove / update something.

### Your environment

 version 0.5 commit=a1f549754b61c1ba84466f0226997687127f7fc0
 4.15.0-24-generic #26~16.04.1-Ubuntu
Bitcoin Core RPC client version v0.16.2.0-g2848aa808f

### Steps to reproduce

create a new invoice
lncli --network=testnet addinvoice --amt 40000 --memo testinv

list invoices
 lncli --network=testnet listinvoices

### Expected behaviour

Expect to see new invoice at the end of listinvoices

### Actual behaviour

only old invoices shown

edit: added version and environment info

wpaulino:
In this commit, we change the default behaviour when querying the
database for invoices and not providing an index offset. Before, if an
index offset wasn't provided, we'd assume that the user would like to
start their query from the first invoice ever created. This isn't as
useful, as users will usually be interested in their most recent
invoices due to older ones expiring. Because of this, we'll now return
the most recent NumMaxInvoices invoices when an index offset isn't
provided. If a user wishes to specify more granularity within their
query, they can still do so by specifying the existing parameters.

Fixes #1852.


halseth:
I think it would be more intuitive to just add a `--reversed` flag, that would flip the order, keeping the pagination and everything else as before?

Roasbeef:
We can add that, but IMO it should be the default, as the goal here is to preserve the _prior_ behavior before pagination. As is, it's a bit unexpected that if a user creates a new invoice, and then does `listinvoices`, the new invoice won't be returned in the response.

Roasbeef:
With this new interface, how's a caller able to obtain the very _first_ invoice? If they specify an index offset of 1, then they'll get the invoice with an add_index of 2.

lnrpc/rpc.proto
```diff
@@ -1734,6 +1734,12 @@ message ListInvoiceRequest {
 
     /// The max number of invoices to return in the response to this query.
     uint32 num_max_invoices = 5 [json_name = "num_max_invoices"];
+
+    /**
+    If set, the invoices returned will result from seeking backwards from the
+    specified index offset.
+    */
+    bool reversed = 6 [json_name = "reversed"];
```
wpaulino: Not sure if we should leave this as `6`. There were some other fields that were removed when the `listinvoices` PR was merged and the ones above were not modified to reflect that.

lnrpc/rpc.proto
```diff
@@ -1743,10 +1749,16 @@ message ListInvoiceResponse {
...
+    The index of the last item in the set of returned invoices. This can be used
+    to seek further, pagination style.
     */
     uint32 last_index_offset = 2 [json_name = "last_index_offset"];
+
+    /**
+    The index of the last item in the set of returned invoices. This can be used
+    to seek backwards, pagination style.
+    */
+    uint32 first_index_offset = 3 [json_name = "first_index_offset"];
```
wpaulino: Should we modify these to be `uint64` as that's what's used within the invoice add index?

channeldb/invoices.go
```diff
@@ -494,19 +512,47 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
...
+				continue
+			default:
+				return err
+			}
+		}
+
+		// If we iterated through the add index in reverse order, then
+		// we'll need to reverse the slice of invoices to return them in
+		// forward order.
+		if q.Reversed {
```
halseth: I imagine when the `reversed` flag is provided, one assumes the results will be returned in reverse order also?

channeldb/invoices.go
```diff
@@ -439,51 +450,58 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
...
+			}
+		} else {
+			// Otherwise, we'll seek to the given offset and begin
+			// iterating either forwards or backwards depending on
+			// the Reversed flag.
+			var startIndex [8]byte
+			if q.Reversed {
+				byteOrder.PutUint64(startIndex[:], q.IndexOffset-1)
+			} else {
+				byteOrder.PutUint64(startIndex[:], q.IndexOffset+1)
```
halseth: I would expect this to just be `q.IndexOffset`? (no `+1`)

channeldb/invoices.go
```diff
@@ -439,51 +450,58 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
...
-		// or reach our max number of events.
-		cursor := invoiceAddedIndex.Cursor()
-		_, invoiceKey := cursor.Seek(startIndex[:])
-		for ; invoiceKey != nil; _, invoiceKey = cursor.Next() {
+		// forEachInvoice is a helper closure that will determine
+		// whether we should include an invoice within the response to
+		// the query. errDone is returned if we've reached the maximum
+		// number of invoices in the response.
+		errDone := errors.New("done")
+		forEachInvoice := func(invoiceKey []byte) error {
```
halseth: Would it be possible to (for readability and a smaller diff) to define 
```go
nextKey := func(c Cursor) {
    if q.Reverserd {
       return c.Prev()
    }
    return c.Next()
}

for ; invoiceKey != nil; _, invoiceKey = nextKey(cursor) {
```

channeldb/invoices.go
```diff
@@ -439,51 +450,58 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
...
+		// our cursor depending on the parameters set within the query.
+		cursor := invoiceAddIndex.Cursor()
+		var invoiceKey []byte
+
+		// It's possible that the index offset is set to 0 due to it
+		// either being explicitly set or not by the user. In order to
+		// handle this edge case, we'll look at the Reversed flag to
+		// determine whether we should show the first or last
+		// NumMaxInvoices invoices.
+		if q.IndexOffset == 0 {
```
halseth: This could be made more readable:

```go
start = q.IndexOffset
if q.Reversed {
     start = cursor.Last() - q.IndexOffset
}

invoiceKey = cursor.Seek(start)


```

lnrpc/rpc.proto
halseth: Can you point to the commit that removed those fields? Generally one should not reuse field tags, as long as they are added we should be okay :)

channeldb/invoices.go
wpaulino: I think it still makes sense to provide them in order. Perhaps it should be renamed to `backwards`?

channeldb/invoices.go
wpaulino: No, as the index they should be providing should be the index of either the first or last invoice in the response. In that case, we don't want to show that invoice again, so we add/subtract one respectively.

channeldb/invoices.go
wpaulino: Good idea! Fixed.

channeldb/invoices.go
wpaulino: Took a somewhat similar approach. Unfortunately we can't subtract the offset from the cursor as `Last()` just returns the `(key, value)` pair.

lnrpc/rpc.proto
wpaulino: The removed fields won't show as they weren't committed yet, but this is the commit where they were added. 

https://github.com/lightningnetwork/lnd/pull/1763/commits/1c7f10a467952e0d0977847db80577f3315c6ac8#diff-f6ec8c0f52caccfaec5656028d316c7fR1720

lnrpc/rpc.proto
cfromknecht: According to the "Updating A Message Type" section in https://developers.google.com/protocol-buffers/docs/proto, it should be safe to do this. It's one of 15 things you "just have to remember" lol.

channeldb/invoices.go
halseth: Callers can easily reverse themselves, so guess it's just a subjective preference.

cmd/lncli/commands.go
```diff
@@ -2310,6 +2310,15 @@ var listInvoicesCommand = cli.Command{
...
 	Usage:    "List all invoices currently stored.",
+	Description: `
+	This command enables the retrieval of all invoices currently stored
+	within the database. It has full support for pagination-style responses,
+	allowing users to query for specific invoices through their add_index.
+	This can be done by using either the first_index_offset or
+	last_index_offset fields included in the response. The reversed flag is
+	set by default in order to paginate backwards. If you wish to paginate
+	forwards, you must explicitly set the flag to false. If none of the
+	parameters are specified, then the last 100 invoices will be returned.`,
```
halseth: Usage description of "index_offset" need an update?

channeldb/invoices.go
```diff
@@ -439,51 +449,57 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
...
+				return c.Prev()
+			}
+			return c.Next()
+		}
+
+		// We'll be using a cursor to seek into the database and return
+		// a slice of invoices. We'll need to determine where to start
+		// our cursor depending on the parameters set within the query.
+		c := invoiceAddIndex.Cursor()
+		invoiceKey := keyForIndex(c, q.IndexOffset+1)
```
halseth: Does this mean the invoiceKeys are 1-indexed? Don't see how you can fetch invoice `0` now.

Roasbeef: The add_index value starts at 1.

cmd/lncli/commands.go
Roasbeef: Will update it in an upcoming commit.

channeldb/invoices.go
wpaulino: Yep, the invoice add index starts at 1, so it's safe to do this.

PR #1865, opened 2018-09-07
config: Experimental support for Litecoin/simnet

### LINKED ISSUE #1830: [Feature request] LND support for Litecoin on Simnet
Tried to run LND  with litecoin on simnet:

```
loadConfig: simnet mode for litecoin not currently supported
```

Is there a good reason for this not to be supported?
May I suggest a PR?

offerm:
This PR adds the configuration needed to run LND with Litecoin on simnet. The change is minimal and has no impact for users that don't enable this mode. When using this configuration, the user is being warned that this mode is not officially supported.

Fixes #1830.

### DIFF
--- a/lnd.go
+++ b/lnd.go
@@ -123,7 +123,7 @@ func lndMain() error {
 	case cfg.Bitcoin.MainNet || cfg.Litecoin.MainNet:
 		network = "mainnet"
 
-	case cfg.Bitcoin.SimNet:
+	case cfg.Bitcoin.SimNet || cfg.Litecoin.SimNet:
 		network = "simnet"
 
 	case cfg.Bitcoin.RegTest:
--- a/chainparams.go
+++ b/chainparams.go
@@ -54,6 +54,14 @@ var bitcoinSimNetParams = bitcoinNetParams{
 	CoinType: keychain.CoinTypeTestnet,
 }
 
+// litecoinSimNetParams contains parameters specific to the simulation test
+// network.
+var litecoinSimNetParams = litecoinNetParams{
+	Params:   &litecoinCfg.SimNetParams,
+	rpcPort:  "18556",
+	CoinType: keychain.CoinTypeTestnet,
+}
+
 // litecoinTestNetParams contains parameters specific to the 4th version of the
 // test network.
 var litecoinTestNetParams = litecoinNetParams{
--- a/config.go
+++ b/config.go
@@ -541,10 +541,6 @@ func loadConfig() (*config, error) {
 			"litecoin.active must be set to 1 (true)", funcName)
 
 	case cfg.Litecoin.Active:
-		if cfg.Litecoin.SimNet {
-			str := "%s: simnet mode for litecoin not currently supported"
-			return nil, fmt.Errorf(str, funcName)
-		}
 		if cfg.Litecoin.RegTest {
 			str := "%s: regnet mode for litecoin not currently supported"
 			return nil, fmt.Errorf(str, funcName)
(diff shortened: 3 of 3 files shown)


lnd.go
```diff
@@ -134,6 +134,10 @@ func lndMain() error {
 		strings.Title(registeredChains.PrimaryChain().String()),
 		network,
 	)
+	if cfg.Litecoin.SimNet {
```
halseth: Not needed, no risk running on simnet anyway.

offerm: I added that here just to inform the user that he can't expect community and LND's team help if encountered problems related to Litecoin and SimNet.

Should I take it out?

Let me know @halseth  please.

offerm:
@halseth took out per your suggestion.

Anything else?

cfromknecht:
The simnet params for ltcd should be updated as well: https://github.com/ltcsuite/ltcd/blob/master/chaincfg/params.go#L492

Currently they are just copied from bitcoin simnet. Some things that would need to change:
 * `SubsidyReductionInterval`,`TargetTimespan`, `TargetTimePerBlock`, `MinDiffReductionTime` should be made equal to the mainnet values
 * `MinDiffReductionTime` should be set to the testnet value
 * `HDCoinType` should be 1

offerm:
> The simnet params for ltcd should be updated as well: https://github.com/ltcsuite/ltcd/blob/master/chaincfg/params.go#L492
> 
> Currently they are just copied from bitcoin simnet. Some things that would need to change:
> 
> * `SubsidyReductionInterval`,`TargetTimespan`, `TargetTimePerBlock`, `MinDiffReductionTime` should be made equal to the mainnet values
> * `MinDiffReductionTime` should be set to the testnet value
> * `HDCoinType` should be 1

@cfromknecht do you suggest to open a PR against ltcsuite/ltcd and wait for a change? or maybe fix these values on LND side once `SimNetParams` for litecoin selected? 

Also, since this is a private network why these parameters are so important?  
Any idea why these did change on ltcsuite/ltcd ?

please guide me.

cfromknecht:
@offerm I would definitely welcome a PR to ltcd making those param changes, this is the primary reason ltc simnet was never exposed in lnd.

> or maybe fix these values on LND side once SimNetParams for litecoin selected?

I'd prefer not to hack this in, the changes to ltcd are pretty minimal and worth doing properly IMO. FWIW, the ltcd backend would then have different parameters from LND.

Roasbeef:
FWIW, you're aware that `litecoind` supports regtest right?

offerm:
@Roasbeef @cfromknecht LTCD is supporting simnet already (even if using wrong parameters).
The only thing which is missing is this small configuration to LND.

@cfromknecht two questions related to these parameters which should change on LTCD:
a. can you please let me know how do you know that these parameters should change and what their value should be?  I would like to get educated before opening such PR. 
b. Why to mix the current PR to LND with an unrelated one for  LTCD? 

 

> > The simnet params for ltcd should be updated as well: https://github.com/ltcsuite/ltcd/blob/master/chaincfg/params.go#L492
> > Currently they are just copied from bitcoin simnet. Some things that would need to change:
> > 
> > * `SubsidyReductionInterval`,`TargetTimespan`, `TargetTimePerBlock`, `MinDiffReductionTime` should be made equal to the mainnet values
> > * `MinDiffReductionTime` should be set to the testnet value
> > * `HDCoinType` should be 1
> 
> @cfromknecht do you suggest to open a PR against ltcsuite/ltcd and wait for a change? or maybe fix these values on LND side once `SimNetParams` for litecoin selected?
> 
> Also, since this is a private network why these parameters are so important?
> Any idea why these did change on ltcsuite/ltcd ?
> 
> please guide me.

dannypaz:
Hoping that this answer helps you

@offerm A) litecoin/bitcoin are very similar in ways, but also have their own protocol level differences (e.g. transaction times or supply). You'll have to look at differences between the two, to know which settings need to be changed. I've found that simply searching for `bitcoin chain params` or `litecoin chain params` results provide good information on the what/why/how of some of these settings.

B) Its probably meant as a quality issue. You will want the LTCD changes merged before making the change to LND or we wont be running a 'correct' simnet.

offerm:
@cfromknecht @halseth per your request I created 
https://github.com/ltcsuite/ltcd/pull/12

Issue #1227, opened 2018-05-10
rpc+contractcourt: merge the contractResolver state into the pendingchannels RPC response 


Roasbeef:
Currently, the `pendingchannels` RPC will only display information drawn from the `utxoNursery` for pending force close channels. However, for a channel that was forced closed with active HTLCs, the HTLC outputs won't be sent to the nursery until they're fully mature. As a result, atm, there's a gap where all information isn't displayed on the RPC layer. This should be reconciled by having the `pendingchannels` RPC query the `chainArbitrator` to obtain information regarding all active `contractResolvers` for a particular channel point.

There are currently four types of [`contractResolvers`](https://github.com/lightningnetwork/lnd/blob/master/contractcourt/contract_resolvers.go#L23): timeout, success, timeout contest, and success contest. The first two are terminal states and will result in outputs being swept on chain by the [ `utxoNursery`](https://github.com/lightningnetwork/lnd/blob/master/utxonursery.go#L214) or the resolver directly. The latter two, are _indeterminate_ states. In this state we're either waiting for an HTLC to timeout, a pre-image to be discovered off-chain, or a pre-image to be discovered on-chain. During this period of time (before expiry or sweeping), there's no information available on the RPC layer concerning these HTLCs. In order to reconcile this, we'll need to extend the `pendingchannels` RPC call to also pull in information concerning these HTLCs. 

One possible path would be to extend the [`HTLC`](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/rpc.proto#L757) with a new enum showing the current status of the HTLC itself.

orbitalturtle:
I'll work on this.

Roasbeef:
Forgot to comment, but this had been assigned, sorry!

On May 20, 2018 10:38 AM, "orbitalturtle" <notifications@github.com> wrote:

I'll work on this.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/lightningnetwork/lnd/issues/1227#issuecomment-390498494>,
or mute the thread

Roasbeef:
Resolved w/ #1875.

PR #1875, opened 2018-09-10
rpc+contractcourt: merge the contractResolver state into the pendingchannels RPC response [DO NOT REVIEW]

joostjager:
#1227


Roasbeef:
I'd like to pick this back up, with the exclusion of the commit to detect spends in the nursery. Since then, we've created the Sweeper which itself will handle this notification case. It still stands that as is, the API is blind to these HTLCs that haven't yet timed out causing much user confusion and an inability to properly track all funds throughout the timeout cycle. Even after the new sweeper changes have been merged, we'll still need this for the lifetime that the current nursery exists (and the resolvers will still report directly back to the API).

joostjager:
Some observations on the current reporting situation:

* When the channel is fully closed, the close summary doesn't show any information on the recovered balance. Just the settled balance at the time of closure.
* During pending close, only recovered balance from locked outputs is reported.
* During pending close, recovered balance stays in the report even when contract resolvers have already finished. The data is kept in the `utxonursery` graduation table.
* Some nursery report fields are unused. I removed them in commit `utxonursery: remove unused report fields`.

Is this the desired behaviour? Ideally I think you'd want to see:
* In addition to the settled balance, report in the channel close summary: recovered balance, lost (to remote) balance and chain fees. These should add up to the settled balance.
* During pending close, show limbo balance, recovered balance, lost (to remote) balance and chain fees. All reported per output. This should again add up to the settled balance. Contrary to current behaviour, also add commitment no delay outputs as recovered balance.
*  Open question is how to attribute the on-chain sweeper fee to the various inputs. One option is to make some kind of proportional calculation it the sweeper and report the per input fee back as part of the sweeper result struct.

Implementation implications:
* Currently the channel close summary is already written before the contract resolvers finish their work. This order would need to change.
* The point that we are working towards is to get rid of `utxonursery` completely. That means that all reporting should originate from the contract resolvers. Currently contract resolvers disappear after the have finished. At that point, the information on recovered/lost/fee balances is lost. To be able to keep reporting this, it should be persisted somewhere else (`briefcase`?). Or alternatively, contract resolvers should be kept alive until all of them have finished.
* How to make a small first step? An option is to remove `UtxoNursery.NurseryReport()` and pull all data from the contract resolvers. If we don't want the changeset to get big, this would mean that reporting on recovered balance during pending close would disappear. I doubt this is a problem, as recovered balance is currently already incomplete. In exchange for this, we'd get accurate limbo balance reporting, because the source is now the contract resolver that has full information on what is going on.

Roasbeef:
> When the channel is fully closed, the close summary doesn't show any information on the recovered balance. Just the settled balance at the time of closure.

Yes we atm we only record the fully settled amount, as depending on a number of parameters (if it's batched, the fee rate, etc) we may actually sweep less than the total amount settled at closing time. 

> During pending close, only recovered balance from locked outputs is reported.

We report the CSV outputs, and HTLC outputs as they transition from stage 1 to stage 2. We also report the total limbo balance, and the amount of coins recovered at that given instance. 

> During pending close, recovered balance stays in the report even when contract resolvers have already finished. The data is kept in the utxonursery graduation table.

Yep, it's part of the base report. Only once _all_ the outputs that were created by a channel have been resolved do we no longer report on it. 

> In addition to the settled balance, report in the channel close summary: recovered balance, lost (to remote) balance and chain fees. These should add up to the settled balance.

Agree this would be a good addition in order to ensure users have access to this data without doing any manual block chain crawling. I think it's out of the scope for this PR though, which should be focused on fixing the egregious blind spots. 

> During pending close, show limbo balance, recovered balance, lost (to remote) balance and chain fees. All reported per output. This should again add up to the settled balance. Contrary to current behavior, also add commitment no delay outputs as recovered balance.

All this but chain fees and the non delay output are already reported. Agree it would be nice to also show the confirmation status of the non-delay output as well. I don't think this would expand the scope of this PR too significantly, as arguably it's a blind spot in the current report and confirmations can take some time on mainnet. 

> Open question is how to attribute the on-chain sweeper fee to the various inputs. 

Yeah we also have this issue where we want to account for the full amount recovered when we need to claim HTLCs on chain. I think this can be deferred from this PR and either picked up in #2075, or a sort of "fee report" message that can be requested from the sweeper (new PR all together). 


> Currently the channel close summary is already written before the contract resolvers finish their work. This order would need to change.

We can still write the close summary as soon as we detect a channel close, but then update the fields as we go. 

> At that point, the information on recovered/lost/fee balances is lost. To be able to keep reporting this, it should be persisted somewhere else (briefcase?).

Yeah so we'd basically progressively move over this state from the nursery over to the main channel arb.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -166,6 +167,26 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
+		finalAmt = btcutil.Amount(
+			h.htlcResolution.SignedSuccessTx.TxOut[0].Value,
+		)
+	}
+
+	return &ContractReport{
+		Outpoint:       h.htlcResolution.ClaimOutpoint,
+		Incoming:       true,
+		Amount:         finalAmt,
+		MaturityHeight: h.htlcExpiry,
```
Roasbeef: Perhaps we also want to show the number of blocks until it's mature? Can likely compute that at the RPC level before passing it onto the caller.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -166,6 +167,26 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 	}
 }
 
+func (h *htlcIncomingContestResolver) report() *ContractReport {
+	// No locking needed as these values are read-only.
```
Roasbeef: Can likely elevate this comment to a "NOTE: ", above for the `godoc` comment for the method itself.

contractcourt/channel_arbitrator.go
```diff
@@ -717,7 +717,19 @@ func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
 		log.Infof("ChannelArbitrator(%v): still awaiting contract "+
 			"resolution", c.cfg.ChanPoint)
 
-		nextState = StateWaitingFullResolution
+		numUnresolved, err := c.log.FetchUnresolvedContracts()
```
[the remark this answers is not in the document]
joostjager: This not fixing a bug, but merely a refactoring to make the flow more understandable.

Previously, the channel arbitrator state machine could finish in `StateWaitingFullResolution`, because it was short cut when the last contract resolved. In that case, also the persistent state wasn't advanced.

Also, `c.cfg.MarkChannelResolved()` was called in two places.

contractcourt/channel_arbitrator.go
```diff
@@ -358,6 +344,23 @@ func (c *ChannelArbitrator) Start() error {
 	return nil
 }
 
+func (c *ChannelArbitrator) relaunchResolvers() error {
```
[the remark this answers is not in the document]
joostjager: Yes

contractcourt/channel_arbitrator.go
```diff
@@ -1421,6 +1490,20 @@ func (c *ChannelArbitrator) resolveContract(currentContract ContractResolver) {
...
+				c.activeResolversLock.Lock()
+				for i, r := range c.activeResolvers {
+					if r == currentContract {
+						c.activeResolvers[i] =
+							nextContract
+					}
+				}
+				c.activeResolversLock.Unlock()
+
+				// TODO: Clear active resolver.
```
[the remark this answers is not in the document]
joostjager: I do it already by swapping above. Outdated todo, removed.

contractcourt/htlc_incoming_contest_resolver.go
joostjager: I like to keep it inside the function, as it is an implementation detail.

contractcourt/htlc_incoming_contest_resolver.go
joostjager: It is currently calculated inside `rpcserver`.

rpcserver.go
```diff
@@ -2175,6 +2132,63 @@ func (r *rpcServer) PendingChannels(ctx context.Context,
 	return resp, nil
 }
 
+func (r *rpcServer) setNurseryForceCloseInfo(chanPoint *wire.OutPoint,
```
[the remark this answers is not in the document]
joostjager: Made a change inspired by your suggestion.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: Rebased onto master

contractcourt/channel_arbitrator.go
```diff
@@ -132,6 +134,36 @@ type ChannelArbitratorConfig struct {
 	ChainArbitratorConfig
 }
 
+// ContractReport provides a summary of a commitment tx output.
+type ContractReport struct {
+	// Outpoint is the final output that will be swept back to the wallet.
+	Outpoint wire.OutPoint
+
+	// Incoming indicates whether the htlc was incoming to this channel.
+	Incoming bool
```
halseth: Maybe we should make this a enum instead of bool? That way we could expose more about what type of output.

contractcourt/channel_arbitrator.go
```diff
@@ -132,6 +134,36 @@ type ChannelArbitratorConfig struct {
 	ChainArbitratorConfig
 }
 
+// ContractReport provides a summary of a commitment tx output.
```
halseth: Is this only for HTLC outputs or also other types?

contractcourt/channel_arbitrator.go
```diff
@@ -1722,29 +1734,18 @@ func (c *ChannelArbitrator) channelAttendant(bestHeight int32) {
...
-
-			// Otherwise, our job is finished here, the contract is
-			// now fully resolved! We'll mark it as such, then exit
-			// ourselves.
-			if err := c.cfg.MarkChannelResolved(); err != nil {
-				log.Errorf("unable to mark contract "+
-					"resolved: %v", err)
+			// If we don't have anything further to do after
+			// advancing our state, then we'll exit.
+			if nextState == StateFullyResolved {
```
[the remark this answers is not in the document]
joostjager: Readded

contractcourt/channel_arbitrator.go
joostjager: It is currently just used for two types of (htlc) outputs, but the intention is to eventually use this for all commitment tx outputs.

contractcourt/channel_arbitrator.go
joostjager: This is also how it is reported over rpc. I'd like to postpone this until we have a good reason to change it.

rpcserver.go
```diff
@@ -2091,13 +2091,25 @@ func (r *rpcServer) PendingChannels(ctx context.Context,
...
+			// and channel arbitrator will be the single source for
+			// these kind of reports.
 			err := r.nurseryPopulateForceCloseResp(
 				&chanPoint, currentHeight, forceClose,
 			)
 			if err != nil {
 				return nil, err
 			}
 
+			err = r.arbitratorPopulateForceCloseResp(
```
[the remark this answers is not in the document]
joostjager: Comment above applies to this statement as well

rpcserver.go
```diff
@@ -2143,6 +2155,42 @@ func (r *rpcServer) PendingChannels(ctx context.Context,
...
 }
 
+// arbitratorPopulateForceCloseResp populates the pending channels response
+// message with channel resolution information from the contract resolvers.
+func (r *rpcServer) arbitratorPopulateForceCloseResp(chanPoint *wire.OutPoint,
+	currentHeight int32,
+	forceClose *lnrpc.PendingChannelsResponse_ForceClosedChannel) error {
+
+	// Query for contract resolvers state.
+	arbitrator, err := r.server.chainArb.GetChannelArbitrator(*chanPoint)
```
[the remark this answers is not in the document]
joostjager: See earlier comment

joostjager:
ptal @halseth

contractcourt/htlc_incoming_contest_resolver.go
Roasbeef: Yep, missed that!

contractcourt/channel_arbitrator.go
```diff
@@ -1430,6 +1498,18 @@ func (c *ChannelArbitrator) resolveContract(currentContract ContractResolver) {
 						"contract: %v", err)
 				}
 
+				// Swap contract in resolvers list. This is to
+				// make sure that reports are queried from the
+				// new resolver.
+				c.activeResolversLock.Lock()
+				for i, r := range c.activeResolvers {
+					if r == currentContract {
```
Roasbeef: Comparing pointers makes me nervous at times, instead this can use the `ResolverKey()` method directly.

joostjager: Yes valid point. Fixed and replace operation extract to method.

PR #1878, opened 2018-09-10
release: Add ARMv6 platform to release.sh

davidknezic:
This PR adds a build for the `ARMv6` platform to `release.sh`. That's the platform that the *Raspberry Pi Zero* and the *Raspberry Pi Zero W* are built on. It does that by [providing `GOARM=6`](https://github.com/golang/go/wiki/GoArm#supported-architectures) to the `go build`.

This generates an entirely new package `lnd-linux-armv6-XYZ.tar.gz`.
For consistency, it also changes the release name of the previous `linux-arm` build to `lnd-linux-armv7-XYZ.tar.gz`.

It would be fantastic to have `Linux ARMv6` included, so that Pi Zero hackers (like me) don't have to build the released binaries ourselves or don't have to rely on unofficial releases.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -21,7 +21,7 @@ MAINDIR=$PACKAGE-$TAG
 mkdir -p $MAINDIR
 cd $MAINDIR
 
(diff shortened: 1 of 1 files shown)


davidknezic:
If you're looking at the diff and wondering why I've removed `linux-arm64` from the `SYS` variable: it's because it's a duplicate entry.

PR #1886, opened 2018-09-11
channeldb: relax bucket assumptions for latest db migration

Roasbeef:
In this commit, we no longer assume that the bucket hierarchy has been
created properly when applying the latest DB migration. On older nodes
that never obtained a channel graph, or updated _before_ the query sync
stuff was added, then they're missing buckets that the migration expects
them to have.

We fix this by simply creating the buckets as we go, if needed.

### DIFF
--- a/channeldb/migrations.go
+++ b/channeldb/migrations.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/binary"
-	"errors"
 	"fmt"
 
 	"github.com/coreos/bbolt"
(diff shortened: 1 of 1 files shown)


channeldb/migrations.go
```diff
@@ -474,20 +473,24 @@ func migratePruneEdgeUpdateIndex(tx *bolt.Tx) error {
...
 	}
-	edgeUpdateIndex := edges.Bucket(edgeUpdateIndexBucket)
-	if edgeUpdateIndex == nil {
-		return nil
+	edgeUpdateIndex, err := edges.CreateBucketIfNotExists(
+		edgeUpdateIndexBucket,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to create/fetch edge update " +
+			"index bucket")
```
wpaulino: Not sure how helpful the errors returned by `CreateBucketIfNotExists` are, but we could include them here.

Roasbeef: IIRC, it doesn't give any more context than this.

PR #1890, opened 2018-09-12
utxonursery: handle remote spends [DO NOT REVIEW]

joostjager:
**PR will probably be abandoned, as we are on track to eventually remove nursery (#2000)**

In this PR, utxonursery is made aware of the possibility that the remote party can spent outputs with the payment hash pre-image. 

Currently, outgoing htlcs are sometimes handed off from resolvers to nursery when they are still contested. This can happen when:
- In the broadcast delta period. This instantiates a timeout resolver that is not watching for remote spends.
- Between expiry of the CTLV lock (that is when hand off to nursery takes place) and confirmation of the timeout tx (local commitment published) or sweep tx (remote commitment published).

Possible consequences of a remote spend after hand-off are:

1. Incorrect htlc status reported in the `pendingchannels` rpc call.
2. Trying to include a spent output in the sweep tx, making the sweep fail and losing the other mature outputs on that height.
3. Not registering the pre-image, causing potential loss of funds by not being able to claim upstream.

This PR addresses issue 1 and part of issue 2. When a remote spend is detected, nursery moves the output into a seperate `spnd` bucket and removes it from the height index. It will not be included anymore in the subsequent sweep tx, unless the sweep tx was already constructed (*). When outputs are moved to the spnd bucket, they will be reported as "stage 0" in the htlc maturity report and also not count towards limbo balance anymore.

(*) This problem is to be solved in a follow-up PR. Possible fix could be to reconstruct the sweep tx when the height index is changed.

For issue 3, the ground work is laid by detecting the remote spend. One possible way to build this out is to extract the pre-image and signal it to the pre-image cache. 

Now that nursery has a broader view on the status of the outputs that it is managing, the road is opened to broadcasting relevant output events (in particular the final states graduated and remote spend) to the contract resolvers. This will remove the duplication of tracking code that is present in the resolvers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,8 +175,8 @@ flakehunter: build
 
 flake-unit:
 	@$(call print, "Flake hunting unit tests.")
-	$(UNIT) -count=1
-	while [ $$? -eq 0 ]; do /bin/sh -c "$(UNIT) -count=1"; done
+	GOTRACEBACK=all $(UNIT) -count=1
+	while [ $$? -eq 0 ]; do /bin/sh -c "GOTRACEBACK=all $(UNIT) -count=1"; done
 
 # ======
 # TRAVIS
--- a/nursery_store.go
+++ b/nursery_store.go
@@ -109,6 +109,10 @@ type NurseryStore interface {
 	// kidOutput's CSV delay.
 	CribToKinder(*babyOutput) error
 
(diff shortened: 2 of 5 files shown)


joostjager:
All code in this pr is either merged or superseded by the sweeper development. Closing.

PR #1892, opened 2018-09-12
travis cache

halseth:
## UPDATED
This PR restructures the Travis build pipeline to speed up the existing build, and prepare for multibackend integration test builds.

* The builds using `go 1.10` are removed. We only run on `go 1.11.x` now.
* The integration test and unit-coverage build is separated into two builds. 
* The build directories are cached, which makes building as well as the unit-race runs significantly faster.

Overall the longest build is now about 16 minutes if a cache exists. This is down from about 28 minutes.

### Original
This PR adds caching for a few directories used by Travis, to speed up the bulid process. 

|            | go 1.9.4 | go 1.10 | go 1.9.4 (cache) | go 1.10 (cache) |
|------------|----------|---------|------------------|-----------------|
| race=false | 28m 31s  | 23m 46s | 27m 36s          | 21m 08s         |
| race=true  | 31m 39s  | 27m 24s | 31m 28s          | 12m 01s         |

(https://travis-ci.org/lightningnetwork/lnd/builds/428112578 and https://travis-ci.org/lightningnetwork/lnd/builds/428125178)

Most notably is the caching of the `$GOCACHE` dir, reduces the `unit-race` test build for `go 1.10` from 27m to 12m! This is a result of test results now being cached. Also the other runs are sped up a bit, but not by as much.

The next step would be to break up the `race=false` test into separate tests for `itest` and `unit-cover`, as they both are long, and independent. This together with `go 1.10+` should make the longest running build about 12-14m.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: go
-matrix:
-  include:
-  - go: "1.11.x"
-    env: RACE=true LINT=true
-  - go: "1.11.x"
-    env: ITEST=true COVER=true LINT=true
+go:
+  - "1.11.x"
+
+env:
+  matrix:
+    - RACE=true LINT=true
+    - ITEST=true
+    - COVER=true
 
 sudo: required
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,15 +9,17 @@ matrix:
     env: RACE=TRUE USE_LINT=FALSE
   - go: "1.10.x"
     env: RACE=FALSE USE_LINT=FALSE
+
 sudo: required
 install:
   - sudo add-apt-repository -y ppa:bitcoin/bitcoin -y
   - sudo apt-get update -q
   - sudo apt-get install bitcoind -y
   - export PATH=$PATH:$PWD/linux-amd64/
+
 script:
-  - export PATH=$PATH:$GOPATH/bin
   - make travis
+
 after_script:
   - echo "Uploading to termbin.com..." && find *.log | xargs -I{} sh -c "cat {} | nc termbin.com 9999 | xargs -r0 printf '{} uploaded to %s'"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,17 +2,27 @@ language: go
 go:
   - "1.9.4"
   - "1.10"
+cache:
+  directories:
+    - vendor/
```
cfromknecht: might we run into issues with this due to the inconsistencies we've been experiencing with `dep`? If the build breaks, is there any way for us to remove `vendor` manually?

Another option may be to cache `$GOPATH`, that way all of the packages are already installed (or some recent version), but the `vendor` directory is still fresh. This should cause it to check out and copy packages that are unchanged, and only fetch diffs for dependencies that have updated.

halseth: You can always manually delete the cache from travis-ci.org.

I added caching of the $GOCACHE dir (only for go 1.10+) which speeds up building, and also caches tests results! This means that unit tests run much faster now.

Regarding caching the $GOPATH, you cannot do that entirely, as the lnd directory will conflict. I did add caching of a few of the deps , however, as it sped up some of the steps (make lint, btcd, dep). Caching more dirs can be considered, but it doesn't seem like there's much more to gain there (I modified travis.yml to do each step at a time, such that we can see how long it takes).

halseth:
Blocking this by https://github.com/lightningnetwork/lnd/pull/1786, as upgrading past `go 1.9` is really required for this PR to make much difference.

.travis.yml
Roasbeef: We'll be switching over to go modules soon after this is merged, which'll eliminate all the issues we're seeing with dep.

Roasbeef:
The README states that the race condition tests when down from 27 mins to 12 mins, but the latest run still shows it at 27 mins?

Roasbeef:
Could be a cold cache issue I guess?

PR #1894, opened 2018-09-12
build: update to new btcwallet to set birthday for bitcoind backend

wpaulino:
In this PR, we update our `btcwallet` dependency in order to correctly set the wallet's backing chain client's birthday correctly when running with a bitcoind backend. This problem arose due to https://github.com/lightningnetwork/lnd/commit/1ac1092aaab3b8712e9612df283300689a79ec3e. When running `lnd` with `--noencryptwallet`, this would pass the default birthday of `time.Now()` into the `BitcoindClient` used in the wallet, causing us to ignore blocks that were mined while we were offline.

Depends on https://github.com/btcsuite/btcwallet/pull/542.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -106,7 +106,7 @@
   revision = "ab6388e0c60ae4834a1f57511e20c17b5f78be4b"
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


Roasbeef:
Replaced by #1896

PR #1895, opened 2018-09-13
fuzz: new fuzz package and lnwire parsing harnesses

Crypt-iQ:
This PR updates the fuzzing harness and converts it into one harness per message, updates the README for updated building instructions, and updates the corpus to include support for the newer messages.  A new **fuzz** package was introduced which contains all of the message harnesses.

TODO:
- [x] Update the README with newer build instructions.

- [x] Build script

- [x] Update the go-fuzz fuzzing corpus to include support for the newer lnwire messages.

- [x] Add a fuzzing harness for each lnwire message so that things like correct encodings / signatures can be added and so the fuzzer doesn't get "confused".

Future Work:
-  Flame graph tooling.

-  Fuzz the `wtwire` messages.

-  Add better corpus messages for increased code coverage.

-  Add correct signatures / encodings in the harnesses and other things the fuzzer won't guess.

-  Run for a long time. See: https://github.com/bitcoin/bitcoin/issues/11045#issuecomment-323203253


Roasbeef:
Ping on this PR, is it good to go/review as is, or do you plan to address the lingering TODO's (or update the list)?

Crypt-iQ:
Experiencing this problem: https://github.com/dvyukov/go-fuzz/issues/195
because of go modules.  Was able to get around it though, but not ideal.

Crypt-iQ:
I was talking to @tuxx42 and we can get very good coverage information similar to AFL by using [gccgo](https://blog.golang.org/cover) - so if the fuzzer is trying to guess a magic value, we will know and can add this input to the corpus.

Crypt-iQ:
Removed message prefixes from the seed corpus, so everything is good to go now.  Going to do some coverage comparison tests on a compute engine instance (new approach vs. old, general harness).

Crypt-iQ:
Old, slightly patched harness fuzzes some messages more than others (24 hour test).  Specifically `openchannel`, `acceptchannel`, `nodeannouncement`, `queryshortchanids`, & `replychannelrange` (bad!!).  These are pretty shallow fuzzing targets (no state machines or anything fancy, just deserialization + serialization), so individually fuzzing messages might not give us much more coverage.  The `decodeShortChanIDs` function doesn't hit the `EncodingSortedZlib` block, so that might be worth fuzzing individually... :neckbeard:

Crypt-iQ:
Added zlib harnesses to hit the `EncodingSortedZlib` block

Crypt-iQ:
Updated the PR with just 2 commits and included better building instructions since `go-fuzz-build` master branch fails to build the targets.  Also included a one-liner in `fuzz.md` that builds all fuzzing binaries for a given package.  The only thing left to do would be to generate a decent corpus for each message and zip it up in each harness directory.  However, I don't have the resources to do that right now.

Crypt-iQ:
Removed the corpus from the PR as it's overly burdensome to review with hundreds of files in the way.  Updated coverage is [here](https://crypt-iq.github.io/lnwire-coverage-with-zlib.html#file9).

fuzz/lnwire/channelannouncement/channel_announcement.go
```diff
@@ -0,0 +1,21 @@
...
+
+package channelannouncement
+
+import (
+	fuzz "github.com/lightningnetwork/lnd/fuzz/lnwire"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// Fuzz is used by go-fuzz.
+func Fuzz(data []byte) int {
```
Roasbeef: High level question: why do certain message targets like `AcceptChannel` use a custom scenario rather than the `fuzz.Harness` scenario as many of these below do?

fuzz/lnwire/fundinglocked/funding_locked.go
```diff
@@ -0,0 +1,21 @@
+// +build gofuzz
+
+package fundinglocked
```
Roasbeef: What's the rationale for each of them having a new package rather than just a series of files in a blanket package?

fuzz/lnwire/fuzz_utils.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"reflect"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// PrefixWithMsgType takes []byte and adds a wire protocol prefix
+// to make the []byte into an actual message to be used in fuzzing.
+func PrefixWithMsgType(data []byte, prefix lnwire.MessageType) []byte {
+	prefixBytes := make([]byte, 2)
```
Roasbeef: Alternatively we can just declare it as: 
```go
var prefixBytes [2]byte
```

fuzz/lnwire/replychannelrangezlib/reply_channel_range_zlib.go
```diff
@@ -0,0 +1,60 @@
...
+	fuzz "github.com/lightningnetwork/lnd/fuzz/lnwire"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// Fuzz is used by go-fuzz.
+func Fuzz(data []byte) int {
+
+	var buf bytes.Buffer
+	zlibWriter := zlib.NewWriter(&buf)
+	_, err := zlibWriter.Write(data)
```
Roasbeef: So we're inputting random data rather than well formed (and sorted) short channel IDs? Won't this always trigger our checks to ensure that the input data is properly sorted? Or will the fuzzer eventually incorporate these constraints into the input data it generates?

fuzz/lnwire/channelannouncement/channel_announcement.go
Crypt-iQ: Yeah so this could have been more explicit. it's because of parsing UpfrontShutdownScript - before the test starts, if it's empty (nil slice), it will be converted into a non-nil slice at the very end of the test (the decoding+encoding changes it). since there's no way to only reflect.DeepEqual subsets of a struct (or like switch on error / determine which field the error occured), had to do it this way

fuzz/lnwire/fundinglocked/funding_locked.go
Crypt-iQ: yeah so at the time of writing the tests you couldn't have multiple Fuzz funcs in the same package, but that's changed now so can change it to this - would make the directory look cleaner too..

fuzz/lnwire/replychannelrangezlib/reply_channel_range_zlib.go
Crypt-iQ: So it *should* be the case that the fuzzer learns... checking [here](https://crypt-iq.github.io/lnwire-coverage-with-zlib.html#file25) (messed up the coverage format a bit but look for decodeShortChanIDs) you can see by hovering the mouse that `if cid.ToUint64() <= lastChanID.ToUint64() {` has 15557 hits while the error message `return 0, nil, ErrUnsortedSIDs{lastChanID, cid}` has 3 hits so it is passing the checks

Crypt-iQ: Fuzzers mainly get confused with more complicated structuring or with any sort of auth check.

Crypt-iQ:
Will get rid of all the packages. NOTE: Some of the harnesses are a bit different (`acceptchannel`, `nodeannouncement`, `openchannel`) since the channel ones have a quirk during serialization +deserialization with the UpfrontShutdownScript. `nodeannouncement` contains `[]net.Addr` which don't serialize + deserialize back to the byte-by-byte same address (some compression / better encoding can be taken advantage of sometimes)

Crypt-iQ:
Flattened all packages to single `lnwirefuzz` package. Had to modify the `decodeShortChanIds` function so that invalid `EncodingType` returns an error if there are no short channel ids rather than no error.

Crypt-iQ:
@Roasbeef rebased+squashed, not sure why Travis is failing.  I would also like to point to your attention 0b78f59 which errors if the decoded `EncodingType` is unknown.  I think the fix more accurately follows the ideal behavior of decoding+mimics the encoding, plus it also helps in fuzz testing since the general harness will [panic](https://github.com/lightningnetwork/lnd/pull/1895/commits/cdd33f8da9401754023028957ef57a7f138858ff#diff-172a0738c715b4304ad9bbc83ed046c4R53) if `WriteMessage` fails.

lnwire/query_short_chan_ids.go
```diff
@@ -152,6 +152,10 @@ func decodeShortChanIDs(r io.Reader) (ShortChanIDEncoding, []ShortChannelID, err
 	// type was specified, meaning that there're no further bytes to be
 	// parsed.
 	if len(queryBody) == 0 {
+		// Check whether the encodingType is valid or not.
+		if encodingType > EncodingSortedZlib {
+			return 0, nil, ErrUnknownShortChanIDEncoding(encodingType)
+		}
```
cfromknecht: is it possible just to remove this whole check? to me it looks like both of the existing sections below handle the zero-length case gracefully, and then error will instead be produced by the `default` case

Crypt-iQ: Yeah I can just remove this check

Crypt-iQ: Removed the check, but had to add a special check [here](https://github.com/lightningnetwork/lnd/pull/1895/files#diff-172a0738c715b4304ad9bbc83ed046c4R90). The first `ReadMessage` call would succeed and return an empty list of `ShortChannelIDs` (it would succeed because there would be bytes, but not enough to be a zlib-encoded `*ShortChannelID`); the subsequent call would error on `ReadMessage` when creating the zlib reader because there were no more bytes left to be read.

docs/gofuzz/fuzz.md
```diff
@@ -0,0 +1,58 @@
...
+```
+If:
+
+- `-1` is returned, the fuzzing input is ignored
+- `0` is returned, `go-fuzz` will add the input to the corpus and deprioritize it in future mutations.
+- `1` is returned, `go-fuzz` will add the input to the corpus and prioritize it in future mutations.
+
+### Conclusion ###
+Citizens,
+do your part and `go-fuzz` `lnd` today!
```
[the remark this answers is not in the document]
Crypt-iQ: Done

lnwire/query_short_chan_ids.go
Crypt-iQ: On second thought, removing this check makes `TestReplyChannelRangeEmpty` fail. Now any `QueryShortChanIDs` or `ReplyChannelRange` message with zlib encoding and no `ShortChannelIDs` will fail with `unable to create zlib reader: unexpected EOF`.  I don't think the peer should fail on a legitimate message like this.

lnwire/query_short_chan_ids.go
```diff
@@ -148,13 +148,6 @@ func decodeShortChanIDs(r io.Reader) (ShortChanIDEncoding, []ShortChannelID, err
 	// as that was just the encoding type.
 	queryBody = queryBody[1:]
 
-	// At this point, if there's no body remaining, then only the encoding
-	// type was specified, meaning that there're no further bytes to be
-	// parsed.
-	if len(queryBody) == 0 {
```
cfromknecht: i think we just need to move the check to the beginning of the zlib case and the tests should pass again?

Roasbeef:
New test failure? 
```
--- FAIL: TestReplyChannelRangeEmpty (0.00s)
    --- FAIL: TestReplyChannelRangeEmpty/empty_zlib_encoding (0.00s)
        reply_channel_range_test.go:86: unable to decode req: unable to create zlib reader: unexpected EOF
```

Roasbeef:
New test failure? 
```
--- FAIL: TestReplyChannelRangeEmpty (0.00s)
    --- FAIL: TestReplyChannelRangeEmpty/empty_zlib_encoding (0.00s)
        reply_channel_range_test.go:86: unable to decode req: unable to create zlib reader: unexpected EOF
```

Crypt-iQ:
> New test failure?
> 
> ```
> --- FAIL: TestReplyChannelRangeEmpty (0.00s)
>     --- FAIL: TestReplyChannelRangeEmpty/empty_zlib_encoding (0.00s)
>         reply_channel_range_test.go:86: unable to decode req: unable to create zlib reader: unexpected EOF
> ```

Have to do this: https://github.com/lightningnetwork/lnd/pull/1895#discussion_r384063219

Crypt-iQ:
Ran all the fuzzers one last time and no more bugs with the fuzzing harness from what i can tell. Should be g2g now

PR #1896, opened 2018-09-13
lnwallet+rpc: allow unconfirmed transactions to be returned over RPC, fix SubscribeTransaction unconf for neutrino

### LINKED ISSUE #1422: Cannot get list of unconfirmed transactions via RPC
### Background

Accessing the LND API via GRPC, you can use GetTransactions to list all on-chain transactions but if there are any pending unconfirmed transactions, that are related to my wallet, they will be invisible until they are confirmed.

I see that in the go source code here there is a function that can do this if we specify 0 confirmations, but it is not exposed to the RPC.  

https://github.com/lightningnetwork/lnd/blob/cd8a21c4b829def9bf317a1a9b42f07e36b5891c/lnwallet/btcwallet/btcwallet.go#L221

### Your environment

* version of `lnd` 0.42
* which operating system (`uname -a` on *Nix) Linux
* version of `btcd`, `bitcoind`, or other backend btcd 0.12

Roasbeef:
In this PR, we add a new test to the set of lnwallet integration
tests. In this new test, we aim to ensure that all backends are able to
display unconfirmed transactions in ListChainTransactions. As of this
commit, this test fails as no backends will return unconfirmed
transactions properly.

In this PR, we fix a bug in the arguments to GetTransactions for the
btcwallet implementation of the WalletController interface. Before this
commit, we wouldn't properly return unconfirmed transactions. The issue
was that we didn't specify the special mempool height of "-1", as the
ending height. The mempool height is actually internally converted to
the highest possible height that can fit into a int32.

In this commit, we set the start to zero, and end to -1 (actually
2^32-1) to properly scan for unconfirmed transactions.

In this PR, we also add a new test to ensure that all backends will
properly send out notifications when an unconfirmed transcation that we
send is inserted into the tx store. Before we updated the btcwallet
build commit in dep, this would fail for neutrino but now passes.

Fixes #1422.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -106,7 +106,7 @@
   revision = "ab6388e0c60ae4834a1f57511e20c17b5f78be4b"
 
 [[projects]]
-  digest = "1:012039cbb6694894aa9e9890645e6ade8a5aff5a02c279714779eff3eae53bc2"
+  digest = "1:e6575f3ee773e72a38319dbd8f4cb7f1b574eccfed7cb103c3a25c254d5c250f"
   name = "github.com/btcsuite/btcwallet"
   packages = [
     "chain",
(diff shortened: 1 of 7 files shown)


neogeno:
Yay

PR #1902, opened 2018-09-13
Move chan series to discovery

cfromknecht:
This PR moves the contents of `chan_series.go` to be within the `discovery` package. This code move is in line with future goals of packaging up the daemon to be more modular, and generally move things out of the `main` package.


discovery/chan_series.go
```diff
@@ -89,23 +146,7 @@ func (c *chanSeries) UpdatesInHorizon(chain chainhash.Hash,
 		return nil, err
 	}
 	for _, nodeAnn := range nodeAnnsInHorizon {
-		// Ensure we only forward nodes that are publicly advertised to
```
Roasbeef: New diff removes the protection against leaking out nodes with purely private channels.

cfromknecht: good catch! filtering has been restored

PR #1505, opened 2018-07-06
fundingmanager: send messages directly to peers

wpaulino:
In this PR, we introduce a similar optimization to the one done in #1345. Currently, each call to `server.SendToPeer` acquires the server's mutex to obtain the active peer. We work around this by passing the `peer` struct itself within the internal `fundingManager` messages.

Fixes #1495.

PR #1910, opened 2018-09-14
Fundingmanager state machine cleanup

halseth:
This PR contains a set of commit to clean up the `fundingmanager` state machine. We now share a common codepath between all funding flows, both as initiator and responder. We also reuse this codepath at restart, greatly reducing the number of cases we handle.

### DIFF
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -525,6 +525,8 @@ func (f *fundingManager) start() error {
 		go func(ch *channeldb.OpenChannel) {
 			go f.waitForFundingWithTimeout(ch, confChan, timeoutChan)
 
+			var shortChanID *lnwire.ShortChannelID
+			var ok bool
 			select {
 			case <-timeoutChan:
 				// Timeout channel will be triggered if the number of blocks
@@ -547,43 +549,30 @@ func (f *fundingManager) start() error {
 				if err := ch.CloseChannel(closeInfo); err != nil {
 					fndgLog.Errorf("Failed closing channel "+
 						"%v: %v", ch.FundingOutpoint, err)
+					return
 				}
+				return
 
 			case <-f.quit:
 				// The fundingManager is shutting down, and will
 				// resume wait on startup.
-			case shortChanID, ok := <-confChan:
+				return
+
+			case shortChanID, ok = <-confChan:
 				if !ok {
 					fndgLog.Errorf("Waiting for funding" +
 						"confirmation failed")
 					return
 				}
-
-				// The funding transaction has confirmed, so
-				// we'll attempt to retrieve the remote peer
-				// to complete the rest of the funding flow.
-				peerChan := make(chan lnpeer.Peer, 1)
-
-				var peerKey [33]byte
-				copy(peerKey[:], ch.IdentityPub.SerializeCompressed())
-
-				f.cfg.NotifyWhenOnline(peerKey, peerChan)
-
-				var peer lnpeer.Peer
-				select {
-				case peer = <-peerChan:
-				case <-f.quit:
-					return
-				}
-				err := f.handleFundingConfirmation(
-					peer, ch, shortChanID,
-				)
-				if err != nil {
-					fndgLog.Errorf("Failed to handle "+
-						"funding confirmation: %v", err)
-					return
-				}
+				// Fallthrough.
 			}
+
+			// Success, funding transaction was confirmed.
+			fndgLog.Debugf("ChannelID(%v) is now fully confirmed! "+
+				"(shortChanID=%v)", chanID, shortChanID)
+
+			f.wg.Add(1)
+			go f.advanceFundingState(ch, nil)
 		}(channel)
 	}
 
@@ -1604,14 +1593,12 @@ func (f *fundingManager) handleFundingCreated(fmsg *fundingCreatedMsg) {
 		}
 
 		// Success, funding transaction was confirmed.
-		err := f.handleFundingConfirmation(
-			fmsg.peer, completeChan, shortChanID,
-		)
-		if err != nil {
-			fndgLog.Errorf("failed to handle funding"+
-				"confirmation: %v", err)
-			return
-		}
+		fndgLog.Debugf("ChannelID(%v) is now fully confirmed! "+
+			"(shortChanID=%v)", channelID, shortChanID)
+
+		f.wg.Add(1)
+		go f.advanceFundingState(completeChan, nil)
+
 	}()
 }
 
@@ -2018,43 +2005,6 @@ func (f *fundingManager) waitForFundingConfirmation(
 		close(discoverySignal)
 	}
 	f.localDiscoveryMtx.Unlock()
-}
-
-// handleFundingConfirmation is a wrapper method for creating a new
-// lnwallet.LightningChannel object, calling sendFundingLocked,
-// addToRouterGraph, and annAfterSixConfs. This is called after the funding
-// transaction is confirmed.
-func (f *fundingManager) handleFundingConfirmation(peer lnpeer.Peer,
-	completeChan *channeldb.OpenChannel,
-	shortChanID *lnwire.ShortChannelID) error {
-
-	// We create the state-machine object which wraps the database state.
-	lnChannel, err := lnwallet.NewLightningChannel(
-		nil, completeChan, nil,
-	)
-	if err != nil {
-		return err
-	}
-
-	chanID := lnwire.NewChanIDFromOutPoint(&completeChan.FundingOutpoint)
-
-	fndgLog.Debugf("ChannelID(%v) is now fully confirmed!", chanID)
-
-	err = f.sendFundingLocked(peer, completeChan, lnChannel, shortChanID)
-	if err != nil {
-		return fmt.Errorf("failed sending fundingLocked: %v", err)
-	}
-	err = f.addToRouterGraph(completeChan, shortChanID)
-	if err != nil {
-		return fmt.Errorf("failed adding to router graph: %v", err)
-	}
-	err = f.annAfterSixConfs(completeChan, shortChanID)
-	if err != nil {
-		return fmt.Errorf("failed sending channel announcement: %v",
-			err)
-	}
-
-	return nil
 }
 
 // sendFundingLocked creates and sends the fundingLocked message.
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -1792,6 +1792,14 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 			}
 		}
 
+		err = f.handleFundingConfirmation(completeChan, *shortChanID)
+		if err != nil {
+			fndgLog.Errorf("unable to handle funding confirmation "+
+				"for ChannelPoint(%v): %v",
+				completeChan.FundingOutpoint, err)
+			return
+		}
+
 		fndgLog.Debugf("Channel with ShortChanID %v now confirmed",
 			shortChanID.ToUint64())
 
@@ -1870,6 +1878,16 @@ func (f *fundingManager) waitForFundingWithTimeout(completeChan *channeldb.OpenC
 				// Failed waiting for confirmation, close
 				// confChan to indicate failure.
 				close(confChan)
+				return
+			}
+
+			err = f.handleFundingConfirmation(
+				completeChan, *shortChanID,
+			)
+			if err != nil {
+				fndgLog.Errorf("unable to handle funding "+
+					"confirmation for ChannelPoint(%v): %v",
+					completeChan.FundingOutpoint, err)
 				return
 			}
 
@@ -1988,12 +2006,6 @@ func (f *fundingManager) waitForFundingConfirmation(
 		BlockHeight: confDetails.BlockHeight,
 		TxIndex:     confDetails.TxIndex,
 		TxPosition:  uint16(fundingPoint.Index),
-	}
-
-	err = f.handleFundingConfirmation(completeChan, shortChanID)
-	if err != nil {
-		fndgLog.Errorf("Unable to handle funding confirmation: %v", err)
-		return
 	}
 
 	select {
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -851,23 +851,7 @@ func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
 	// The funding transaction was confirmed, but we did not successfully
 	// send the fundingLocked message to the peer, so let's do that now.
 	case markedOpen:
-		peerChan := make(chan lnpeer.Peer, 1)
-
-		var peerKey [33]byte
-		copy(peerKey[:], channel.IdentityPub.SerializeCompressed())
-
-		f.cfg.NotifyWhenOnline(peerKey, peerChan)
-
-		var peer lnpeer.Peer
-		select {
-		case peer = <-peerChan:
-		case <-f.quit:
-			return ErrFundingManagerShuttingDown
-		}
-
-		err := f.sendFundingLocked(
-			peer, channel, lnChannel, shortChanID,
-		)
+		err := f.sendFundingLocked(channel, lnChannel, shortChanID)
 		if err != nil {
 			return fmt.Errorf("failed sending fundingLocked: %v",
 				err)
@@ -2057,7 +2041,7 @@ func (f *fundingManager) handleFundingConfirmation(
 // sendFundingLocked creates and sends the fundingLocked message.
 // This should be called after the funding transaction has been confirmed,
 // and the channelState is 'markedOpen'.
-func (f *fundingManager) sendFundingLocked(peer lnpeer.Peer,
+func (f *fundingManager) sendFundingLocked(
 	completeChan *channeldb.OpenChannel, channel *lnwallet.LightningChannel,
 	shortChanID *lnwire.ShortChannelID) error {
 
@@ -2088,8 +2072,18 @@ func (f *fundingManager) sendFundingLocked(peer lnpeer.Peer,
 	// send fundingLocked until we succeed, or the fundingManager is shut
 	// down.
 	for {
-		fndgLog.Debugf("Sending FundingLocked for ChannelID(%v) to "+
-			"peer %x", chanID, peerKey)
+		connected := make(chan lnpeer.Peer, 1)
+		f.cfg.NotifyWhenOnline(peerKey, connected)
+
+		var peer lnpeer.Peer
+		select {
+		case peer = <-connected:
+		case <-f.quit:
+			return ErrFundingManagerShuttingDown
+		}
+
+		fndgLog.Infof("Peer(%x) is online, sending FundingLocked "+
+			"for ChannelID(%v)", peerKey, chanID)
 
 		if err := peer.SendMessage(false, fundingLockedMsg); err == nil {
 			// Sending succeeded, we can break out and continue the
@@ -2099,20 +2093,6 @@ func (f *fundingManager) sendFundingLocked(peer lnpeer.Peer,
 
 		fndgLog.Warnf("Unable to send fundingLocked to peer %x: %v. "+
 			"Will retry when online", peerKey, err)
-
-		connected := make(chan lnpeer.Peer, 1)
-		f.cfg.NotifyWhenOnline(peerKey, connected)
-
-		select {
-		case <-connected:
-			fndgLog.Infof("Peer(%x) came back online, will retry "+
-				"sending FundingLocked for ChannelID(%v)",
-				peerKey, chanID)
-
-			// Retry sending.
-		case <-f.quit:
-			return ErrFundingManagerShuttingDown
-		}
 	}
 
 	return nil
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -598,10 +598,8 @@ func (f *fundingManager) start() error {
 	}
 
 	for _, channel := range openChannels {
-		err = f.advanceFundingState(channel)
-		if err != nil {
-			return err
-		}
+		f.wg.Add(1)
+		go f.advanceFundingState(channel, nil)
 	}
 
 	f.wg.Add(1) // TODO(roasbeef): tune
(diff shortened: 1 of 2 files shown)


halseth:
Not ready for review yet.

halseth:
Abstracted even more common code, and added a fix for a previous non-atomic DB transaction that could cause problems (last commit).

Ready for review.

fundingmanager.go
```diff
@@ -1898,31 +1832,102 @@ func (f *fundingManager) waitForFundingConfirmation(completeChan *channeldb.Open
...
-	//
-	// TODO(halseth): make the two db transactions (MarkChannelAsOpen and
-	// saveChannelOpeningState) atomic by doing them in the same transaction.
-	// Needed to be properly fault-tolerant.
-	err = f.saveChannelOpeningState(&completeChan.FundingOutpoint, markedOpen,
-		&shortChanID)
+	// useful to resume the opening process in case of restarts. We set the
+	// opening state before we mark the channel opened in the database,
+	// such that we can receover from one of the db writes failing.
+	err := f.saveChannelOpeningState(&fundingPoint, markedOpen, &shortChanID)
```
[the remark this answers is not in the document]
halseth: Fixed by https://github.com/lightningnetwork/lnd/pull/1910/commits/26edf3996d18c613878b6c607666bc21df971eff ?

fundingmanager.go
```diff
@@ -801,17 +755,89 @@ func (f *fundingManager) reservationCoordinator() {
...
+				peerChan := make(chan lnpeer.Peer, 1)
+				f.cfg.NotifyWhenOnline(ch.IdentityPub, peerChan)
+
+				var peer lnpeer.Peer
+				select {
+				case peer = <-peerChan:
+				case <-f.quit:
+					return
+				}
+				f.failFundingFlow(peer, pendingChanID, err)
```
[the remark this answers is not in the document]
halseth: That's correct. However, seems like a discussion for a follow-up PR. Made a TODO.

fundingmanager.go
```diff
@@ -909,6 +909,31 @@ func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
...
 			"added to router graph.", chanID, shortChanID)
 
+		// Give the caller a final update notifying them that
+		// the channel is now open.
+		// TODO(roasbeef): only notify after recv of funding locked?
+		fundingPoint := channel.FundingOutpoint
+		upd := &lnrpc.OpenStatusUpdate{
+			Update: &lnrpc.OpenStatusUpdate_ChanOpen{
+				ChanOpen: &lnrpc.ChannelOpenUpdate{
+					ChannelPoint: &lnrpc.ChannelPoint{
```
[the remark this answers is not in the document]
halseth: Fixed!

halseth:
All pending channels will be republished then `waitForFundingWithTimeout->handleFundingConfirmation` which will again call `MarkAsOpen`. I think that should cover it?

wpaulino:
Needs a rebase!

halseth:
Rebased.

fundingmanager.go
```diff
@@ -824,17 +779,94 @@ func (f *fundingManager) reservationCoordinator() {
...
+				}
+				// TODO(halseth): should this send be made
+				// reliable?
+				f.failFundingFlow(peer, pendingChanID, err)
+			}()
+
+			return
+
+		} else if err != nil {
+			fndgLog.Errorf(err.Error())
```
wpaulino: Nit: log should include more context.

fundingmanager.go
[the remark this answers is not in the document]
wpaulino: Seems like we'd also want to call `MarkAsOpen` within the `markedOpen` case of  `stateStep`.

fundingmanager.go
```diff
@@ -931,6 +837,114 @@ func (f *fundingManager) reservationCoordinator() {
...
+				return
+			}
+		}(channel)
+
+	default:
+		return fmt.Errorf("undefined channelState: %v",
+			channelState)
+	}
+
+	return nil
```
Roasbeef: This will never be hit.

fundingmanager.go
```diff
@@ -931,6 +837,114 @@ func (f *fundingManager) reservationCoordinator() {
...
+func (f *fundingManager) advanceFundingState(
+	channel *channeldb.OpenChannel) error {
+
+	channelState, shortChanID, err := f.getChannelOpeningState(
+		&channel.FundingOutpoint)
+	if err == ErrChannelNotFound {
+		// Channel not in fundingManager's opening database,
+		// meaning it was successfully announced to the
+		// network.
+		return nil
```
Roasbeef: Do we also want to add that consistency check here? Basically, if the channel is open, then also check to ensure that it's been added to to the graph. Alternatively, it can just add it again, since it's a idempotent operation.

fundingmanager.go
```diff
@@ -608,10 +608,8 @@ func (f *fundingManager) start() error {
 	}
 
 	for _, channel := range openChannels {
-		err = f.advanceFundingState(channel)
```
Roasbeef: In the commit message: 
> qyeru the datanase 

should be: 
> query the database

fundingmanager.go
```diff
@@ -840,109 +838,125 @@ func (f *fundingManager) reservationCoordinator() {
...
+				"channel opening state(%v): %v",
+				channel.FundingOutpoint, err)
+			return
+		}
+
+		// If we did find the channel in the opening state database, we
+		// have seen the funding transaction being confirmed, but there
+		// are still steps left of the setup procedure. We continue the
+		// procedure where we left off.
+		err = f.stateStep(channel, lnChannel, shortChanID,
```
Roasbeef: Folding should be: 
```go
err = f.stateStep(
    channel, lnChannel, shortChanID, channelState, 
    updateChan, 
)
```

fundingmanager.go
```diff
@@ -840,109 +838,125 @@ func (f *fundingManager) reservationCoordinator() {
...
+		nil, channel, nil,
+	)
+	if err != nil {
+		fndgLog.Errorf("Unable to create LightningChannel(%v): %v",
+			channel.FundingOutpoint, err)
+		return
+	}
+
+	for {
+		channelState, shortChanID, err := f.getChannelOpeningState(
```
Roasbeef: Do we need to read it from disk each time? We already know what the new state is, and only this loop can modify the the state.

fundingmanager.go
```diff
@@ -840,109 +838,125 @@ func (f *fundingManager) reservationCoordinator() {
...
-			// different states, to avoid this code
-			// duplication.
-			err = f.annAfterSixConfs(dbChan, shortChanID)
-			if err != nil {
-				fndgLog.Errorf("error sending channel "+
-					"announcements: %v", err)
-				return
-			}
-		}(channel)
+		return nil
```
Roasbeef: Why was the prior call to `annAfterSixConf` removed here?

fundingmanager.go
```diff
@@ -1748,62 +1774,8 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
...
-		}
-
-		err = f.annAfterSixConfs(completeChan, shortChanID)
-		if err != nil {
-			fndgLog.Errorf("failed sending channel announcement: %v",
-				err)
-			return
-		}
+		f.wg.Add(1)
+		go f.advanceFundingState(completeChan, resCtx.updates)
```
Roasbeef: 👍

fundingmanager.go
```diff
@@ -840,109 +838,125 @@ func (f *fundingManager) reservationCoordinator() {
...
+}
+
+// stateStep advances the confirmed channel one step in the funding state
+// machine. This method is synchronous and the new channel opening state will
+// have been written to the database when it successfully returns. The
+// updateChan can be set non-nil to get OpenStatusUpdates.
+func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
+	lnChannel *lnwallet.LightningChannel,
+	shortChanID *lnwire.ShortChannelID, channelState channelOpeningState,
+	updateChan chan<- *lnrpc.OpenStatusUpdate) error {
```
Roasbeef: Where do we actually send to the `updateChan`?

fundingmanager.go
```diff
@@ -912,8 +912,20 @@ func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
 				err)
 		}
 
+		// As the fundingLocked message is now sent to the peer, the
+		// channel is moved to the next state of the state machine. It
+		// will be moved to the last state (actually deleted from the
+		// database) after the channel is finally announced.
+		err = f.saveChannelOpeningState(&channel.FundingOutpoint,
```
Roasbeef: Folding should be: 
```go
f.saveChannelOpeningState(
     &channel.FundingOutpoint, fundingLockedSent, shortChanID,
)
```

Roasbeef: Alternatively, we can have this logic be present in the caller function, then based on the final state returned,  either persist or delete the funding state.

fundingmanager.go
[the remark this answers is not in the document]
Roasbeef: Similar line folding comment here.

fundingmanager.go
```diff
@@ -1990,12 +1990,35 @@ func (f *fundingManager) waitForFundingConfirmation(
...
+	case <-f.quit:
+		return
+	}
+}
+
+// handleFundingConfirmation marks a channel as open in the database, and set
+// the channelOpeningState markedOpen. In addition it will report the now
+// decided short channel ID to the switch, and close the local discovery signal
+// for this channel.
+func (f *fundingManager) handleFundingConfirmation(
```
Roasbeef: Why not also fold this into the `stateStep` method? This method has been deleted, then re-added (in a slightly different format), in this diff so far.

fundingmanager.go
```diff
@@ -536,54 +536,7 @@ func (f *fundingManager) start() error {
 		}
 
 		f.wg.Add(1)
-		go func(ch *channeldb.OpenChannel) {
```
Roasbeef: In the commit message: 
> pening before 

should be: 
> pending before

fundingmanager.go
```diff
@@ -824,17 +779,94 @@ func (f *fundingManager) reservationCoordinator() {
...
 // NOTE: This MUST be run as a goroutine.
 func (f *fundingManager) advanceFundingState(channel *channeldb.OpenChannel,
-	updateChan chan<- *lnrpc.OpenStatusUpdate) {
+	pendingChanID [32]byte, updateChan chan<- *lnrpc.OpenStatusUpdate) {
 	defer f.wg.Done()
 
+	// If the channel is still pending we must wait for the funding
+	// transaction to confirm.
+	if channel.IsPending {
+		shortChanID, err := f.waitForFundingWithTimeout(channel)
```
Roasbeef: To reduce the vertical distance a bit in this method, this should be tucked away into another method IMO. It would then make it easier to test isolation as well.

fundingmanager.go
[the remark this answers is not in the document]
Roasbeef: How do we handle this on the receiving side? This can be interpreted as a signal to double-spend our funding output.

fundingmanager.go
```diff
@@ -499,59 +499,58 @@ func (f *fundingManager) start() error {
...
 
-	// For any channels that were in a pending state when the daemon was
-	// last connected, the Funding Manager will re-initialize the channel
-	// barriers and will also launch waitForFundingConfirmation to wait for
-	// the channel's funding transaction to be confirmed on the blockchain.
-	for _, channel := range pendingChannels {
-		f.barrierMtx.Lock()
-		fndgLog.Tracef("Loading pending ChannelPoint(%v), creating chan "+
-			"barrier", channel.FundingOutpoint)
+	for _, channel := range allChannels {
```
Roasbeef: 👍

fundingmanager.go
```diff
@@ -1992,34 +1992,32 @@ func (f *fundingManager) handleFundingConfirmation(
 	fundingPoint := completeChan.FundingOutpoint
 	chanID := lnwire.NewChanIDFromOutPoint(&fundingPoint)
 
-	// Now that the channel has been fully confirmed, we'll mark it as open
```
Roasbeef: In the commit message:  regitster -> register

fundingmanager.go
halseth: It will in all but the `default` case.

fundingmanager.go
halseth: Yeah, that could be a way to handle the missing graph edges indeed.

fundingmanager.go
halseth: Fixed.

fundingmanager.go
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/1910/commits/36ea8b14297150c4d35eeb60709e0e9c71b4c4a5

halseth: It is to simplify the state machine, so it can easily be resumed from any state. Returning the new state from `stateStep` to only read it once would be an optimization, but now sure how critical.

fundingmanager.go
```diff
@@ -840,109 +838,125 @@ func (f *fundingManager) reservationCoordinator() {
...
+
+	for {
+		channelState, shortChanID, err := f.getChannelOpeningState(
+			&channel.FundingOutpoint)
+		if err == ErrChannelNotFound {
+			// Channel not in fundingManager's opening database,
+			// meaning it was successfully announced to the
+			// network.
+			return
+		} else if err != nil {
```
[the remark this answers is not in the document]
halseth: Find it cleaner with else just to make it clear :)

fundingmanager.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/1910/commits/36ea8b14297150c4d35eeb60709e0e9c71b4c4a5

fundingmanager.go
halseth: Done in the next commit.

fundingmanager.go
```diff
@@ -840,109 +838,125 @@ func (f *fundingManager) reservationCoordinator() {
...
 	chanID := lnwire.NewChanIDFromOutPoint(&channel.FundingOutpoint)
-	fndgLog.Debugf("channel (%v) with opening state %v found",
-		chanID, channelState)
-
-	// If we did find the channel in the opening state database, we have
-	// seen the funding transaction being confirmed, but we did not finish
-	// the rest of the setup procedure before we shut down. We handle the
-	// remaining steps of this setup by continuing the procedure where we
-	// left off.
+	fndgLog.Debugf("Channel(%v) with ShortChanID %v has opening state %v.",
```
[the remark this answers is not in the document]
halseth: Done.

fundingmanager.go
halseth: The state machine will advance to the next state (`addedToRouterGraph`) where it will be called.,

fundingmanager.go
halseth: Fixed folding: addedToRouterGraph

fundingmanager.go
halseth: It is a new method really, name colliding with the removed one is a unfortunate coincidence...

It cannot easily be folded into `stateStep`, since the state machine really starts from the `markedOpen` state, which is set here. Or maybe I misunderstood what you meant?

fundingmanager.go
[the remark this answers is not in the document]
halseth: > Seems like we'd also want to call `MarkAsOpen` within the `markedOpen` case of `stateStep`.

`markedOpen` step assumes it is already marked open in the db.

> Similar line folding comment here.

Fixed

fundingmanager.go
```diff
@@ -1792,6 +1792,12 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 			}
 		}
 
+		err = f.handleFundingConfirmation(completeChan, *shortChanID)
+		if err != nil {
+			fndgLog.Errorf("Unable to handle funding confirmation: %v", err)
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/1910/commits/079ce9afbe62de26c7aa8f4f6b138a741fd100f7

fundingmanager.go
```diff
@@ -529,20 +535,18 @@ func (f *fundingManager) start() error {
...
-			var ok bool
-			select {
-			case <-timeoutChan:
-				// Timeout channel will be triggered if the number of blocks
-				// mined since the channel was initiated reaches
-				// maxWaitNumBlocksFundingConf and we are not the channel
-				// initiator.
+			shortChanID, err := f.waitForFundingWithTimeout(ch)
+			if err == ErrConfirmationTimeout {
+				fndgLog.Warnf(err.Error())
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/1910/commits/592921650b591c5a4bec9624b11293e10e7ed92d

fundingmanager_test.go
```diff
@@ -245,7 +245,7 @@ func createTestFundingManager(t *testing.T, privKey *btcec.PrivateKey,
 	chainNotifier := &mockNotifier{
 		oneConfChannel: make(chan *chainntnfs.TxConfirmation, 1),
 		sixConfChannel: make(chan *chainntnfs.TxConfirmation, 1),
-		epochChan:      make(chan *chainntnfs.BlockEpoch, 1),
+		epochChan:      make(chan *chainntnfs.BlockEpoch, 2),
```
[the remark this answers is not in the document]
halseth: I went for this approach, as not dispatching the block would hide the fact if the initiator is actually listening waiting for blocks.

fundingmanager.go
halseth: Fixed.

fundingmanager.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/3e9e509128dcd9b919f24f363d58f85661e8de30

fundingmanager.go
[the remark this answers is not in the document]
halseth: We will get a generic funding error on the receiving side, since it is not a whitelisted error. We could change this to some error code we understand, to act accordingly.

fundingmanager.go
halseth: Fixed.

fundingmanager.go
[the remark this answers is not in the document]
wpaulino: >> Seems like we'd also want to call MarkAsOpen within the markedOpen case of stateStep.

> markedOpen step assumes it is already marked open in the db.

Ah I see where the confusion arose. Pending channels are now handled through the same flow so it'll go through `handleFundingConfirmation` again if `MarkAsOpen` didn't succeed.

fundingmanager.go
```diff
@@ -1022,6 +984,85 @@ func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
...
+			Capacity:                ch.Capacity,
+			SettledBalance:          localBalance,
+			CloseType:               channeldb.FundingCanceled,
+			RemoteCurrentRevocation: ch.RemoteCurrentRevocation,
+			RemoteNextRevocation:    ch.RemoteNextRevocation,
+			LocalChanConfig:         ch.LocalChanCfg,
+		}
+
+		if err := ch.CloseChannel(closeInfo); err != nil {
+			return fmt.Errorf("Failed closing channel "+
```
wpaulino: Nit: errors returned should start lowercase.

fundingmanager.go
```diff
@@ -499,59 +499,58 @@ func (f *fundingManager) start() error {
...
+				channel.FundingOutpoint)
+
+			f.newChanBarriers[chanID] = make(chan struct{})
+			f.barrierMtx.Unlock()
+
+			f.localDiscoverySignals[chanID] = make(chan struct{})
+
+			// Rebroadcast the funding transaction for any pending
+			// channel that we initiated. If this operation fails
+			// due to a reported double spend, we treat this as an
```
wpaulino: Comment is out of date as `PublishTransaction` no longer returns `ErrDoubleSpend` for transactions previously broadcast.

fundingmanager.go
halseth: Would also need to return the new state from the `stateStep` method, and define a new state or error for when to exit the loop.

fundingmanager.go
```diff
@@ -942,6 +778,292 @@ func (f *fundingManager) reservationCoordinator() {
...
+
+	// The funding transaction was confirmed, but we did not successfully
+	// send the fundingLocked message to the peer, so let's do that now.
+	case markedOpen:
+		peerChan := make(chan lnpeer.Peer, 1)
+
+		var peerKey [33]byte
+		copy(peerKey[:], channel.IdentityPub.SerializeCompressed())
+
+		f.cfg.NotifyWhenOnline(peerKey, peerChan)
```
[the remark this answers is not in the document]
halseth: It is done to get access to the `lnpeer.Peer` struct really

fundingmanager.go
halseth: fixed: https://github.com/lightningnetwork/lnd/pull/1910/commits/a7d880e6588fbd67e04d28749f4c22b0fb5cf771

fundingmanager.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/1910/commits/ff6d6ae8a22df54a3a31ad4b60a95f637e0cc9d4

fundingmanager.go
```diff
@@ -942,6 +778,292 @@ func (f *fundingManager) reservationCoordinator() {
...
+	for {
+		channelState, shortChanID, err := f.getChannelOpeningState(
+			&channel.FundingOutpoint,
+		)
+		if err == ErrChannelNotFound {
+			// Channel not in fundingManager's opening database,
+			// meaning it was successfully announced to the
+			// network.
+			// TODO(halseth): could do graph consistency check
+			// here, and re-add the edge if missing.
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/d95d7495a616c2bc8082cfd3f21cb555feacfa4f

fundingmanager.go
wpaulino: :+1: can be done as a follow-up though.

fundingmanager.go
wpaulino: Fine with me either way, doesn't seem very critical.

fundingmanager.go
```diff
@@ -840,109 +838,128 @@ func (f *fundingManager) reservationCoordinator() {
...
-		return nil
-	} else if err != nil {
-		return err
+// announcing the channel. The updateChan can be set non-nil to get
+// OpenStatusUpdates.
+//
+// NOTE: This MUST be run as a goroutine.
+func (f *fundingManager) advanceFundingState(channel *channeldb.OpenChannel,
+	updateChan chan<- *lnrpc.OpenStatusUpdate) {
+	defer f.wg.Done()
```
wpaulino: Nit: add newline before.

fundingmanager.go
```diff
@@ -943,6 +943,32 @@ func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
...
+		// the channel is now open.
+		// TODO(roasbeef): only notify after recv of funding locked?
+		fundingPoint := channel.FundingOutpoint
+		cp := &lnrpc.ChannelPoint{
+			FundingTxid: &lnrpc.ChannelPoint_FundingTxidBytes{
+				FundingTxidBytes: fundingPoint.Hash[:],
+			},
+			OutputIndex: fundingPoint.Index,
+		}
+		upd := &lnrpc.OpenStatusUpdate{
```
wpaulino: Nit: can limit the scope of this to be within the if statement below.

fundingmanager.go
```diff
@@ -557,43 +559,30 @@ func (f *fundingManager) start() error {
...
-					peer, ch, shortChanID,
-				)
-				if err != nil {
-					fndgLog.Errorf("Failed to handle "+
-						"funding confirmation: %v", err)
-					return
-				}
-			}
+			f.wg.Add(1)
+			go f.advanceFundingState(ch, nil)
```
wpaulino: Nit: looks like there's no need to spawn this in another goroutine since we're already in one?

fundingmanager.go
```diff
@@ -1612,14 +1601,12 @@ func (f *fundingManager) handleFundingCreated(fmsg *fundingCreatedMsg) {
...
-		if err != nil {
-			fndgLog.Errorf("failed to handle funding"+
-				"confirmation: %v", err)
-			return
-		}
+		fndgLog.Debugf("ChannelID(%v) is now fully confirmed! "+
+			"(shortChanID=%v)", channelID, shortChanID)
+
+		f.wg.Add(1)
+		go f.advanceFundingState(completeChan, nil)
```
wpaulino: Same comment here regarding the goroutine.

fundingmanager.go
```diff
@@ -942,6 +775,296 @@ func (f *fundingManager) reservationCoordinator() {
...
+		// If we did find the channel in the opening state database, we
+		// have seen the funding transaction being confirmed, but there
+		// are still steps left of the setup procedure. We continue the
+		// procedure where we left off.
+		err = f.stateStep(
+			channel, lnChannel, shortChanID, channelState,
+			updateChan,
+		)
+		if err != nil {
+			fndgLog.Errorf("unable to advance state: %v", err)
```
wpaulino: Nit: capitalize error log and include channel context.

fundingmanager.go
```diff
@@ -1935,9 +1929,12 @@ func makeFundingScript(channel *channeldb.OpenChannel) ([]byte, error) {
 // when a channel has become active for lightning transactions.
 // The wait can be canceled by closing the cancelChan. In case of success,
 // a *lnwire.ShortChannelID will be passed to confChan.
+//
+// NOTE: This MUST be run as a goroutine.
 func (f *fundingManager) waitForFundingConfirmation(
 	completeChan *channeldb.OpenChannel, cancelChan <-chan struct{},
 	confChan chan<- *lnwire.ShortChannelID) {
+	defer f.wg.Done()
```
wpaulino: Nit: add newline before.

fundingmanager.go
```diff
@@ -1829,82 +1800,50 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
...
+		return nil, ErrFundingManagerShuttingDown
 
-			select {
-			case confChan <- shortChanID:
-			case <-f.quit:
-				return
-			}
+	case shortChanID, ok = <-confChan:
+		if !ok {
+			return nil, fmt.Errorf("Waiting for funding" +
```
wpaulino: Nit: error not logged should start lowercase.

fundingmanager.go
```diff
@@ -1022,6 +984,85 @@ func (f *fundingManager) stateStep(channel *channeldb.OpenChannel,
...
+
+	shortChanID, err := f.waitForFundingWithTimeout(channel)
+	if err == ErrConfirmationTimeout {
+		// We'll get a timeout if the number of blocks mined
+		// since the channel was initiated reaches
+		// maxWaitNumBlocksFundingConf and we are not the
+		// channel initiator.
+		ch := channel
+		localBalance := ch.LocalCommitment.LocalBalance.ToSatoshis()
+		closeInfo := &channeldb.ChannelCloseSummary{
```
wpaulino: Nit: could extract all this code into a method.

fundingmanager.go
```diff
@@ -499,59 +499,58 @@ func (f *fundingManager) start() error {
...
-	}
-
-	for _, channel := range openChannels {
-		chanID := lnwire.NewChanIDFromOutPoint(&channel.FundingOutpoint)
-
+		// We will restart the funding state machine for all channels,
+		// which will wait for the channel's funding transaction to be
+		// confirmed on the blockchain, and retransmit the messages
+		// necessary for the channel to be operational.
+		// TODO(halseth): retransmission of messages to make the
```
wpaulino: Which messages?

fundingmanager.go
halseth: It more or less is?

fundingmanager.go
halseth: funding locked and announcements.

fundingmanager.go
halseth: It is just more natural, since the method decrements a waitgroup,

fundingmanager.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/f89e1c4268e16a581616eb408f40fc256a9a948b

fundingmanager.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/bd228a403b7817ed6aac55e6cd27744f801917d2

fundingmanager.go
halseth: Yep 👍 https://github.com/lightningnetwork/lnd/pull/1910/commits/479f80dc7040a92c2a27baa177fdac89e8ca675a

fundingmanager.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/53987d9017c7731b8f298b933b7658a4ff1fee52

fundingmanager.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/e148c0a1af38e6b4565f666261cec638c27ddebe

joostjager:
```
htlcswitch/switch.go:470:41: Using a reference for the variable on range scope `targetLink` (scopelint)
			cid := lnwire.NewChanIDFromOutPoint(&targetLink)
			                                     ^
lnd.go:99:1: cyclomatic complexity 60 of func `Main` is high (> 30) (gocyclo)
func Main() error {
^
htlcswitch/circuit_test.go:33:2: `testExtracter` is a global variable (gochecknoglobals)
	testExtracter *htlcswitch.SphinxErrorEncrypter
	^
fundingmanager.go:1981:30: ST1005: error strings should not be capitalized (stylecheck)
				timeoutChan <- fmt.Errorf("Epoch client " +
				                         ^
htlcswitch/iterator.go:9: File is not `goimports`-ed (goimports)
	"github.com/lightningnetwork/lightning-onion"
htlcswitch/circuit_test.go:12: File is not `goimports`-ed (goimports)
	"github.com/lightningnetwork/lightning-onion"
```

fundingmanager.go
wpaulino: I was referring to the timeout logic, but not blocking.

fundingmanager.go
wpaulino: Funding locked seems to already be sent reliably? As for the announcements, signatures and channel updates are also sent reliably.

fundingmanager.go
```diff
@@ -484,230 +490,53 @@ func (f *fundingManager) start() error {
...
+	for _, channel := range allChannels {
 		chanID := lnwire.NewChanIDFromOutPoint(&channel.FundingOutpoint)
-		fndgLog.Debugf("channel (%v) with opening state %v found",
-			chanID, channelState)
 
+		// For any channels that were in a pending state when the
+		// daemon was last connected, the Funding Manager will
+		// re-initialize the channel barriers, and republish the
+		// funding transaction if we're the initiator.
 		if channel.IsPending {
```
wpaulino: Any reason to not include the pending logic in the `advanceFundingState` closure as well?

fundingmanager.go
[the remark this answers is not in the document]
wpaulino: Perhaps remove the notification request from `sendFundingLocked` then?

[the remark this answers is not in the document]
halseth: Should be okay, as we only request it in case sending fails in `sendFundingLocked`?

fundingmanager.go
halseth: Not really I think, could be done.

halseth: Not really I think, could be done.

fundingmanager.go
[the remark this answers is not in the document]
wpaulino: Yeah, it's just about making the code cleaner.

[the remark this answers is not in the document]
halseth: I think this can just stay as-is for now.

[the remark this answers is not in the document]
wpaulino: Should be pretty straight forward to address, just remove the `lnpeer.Peer` parameter from `sendFundingLocked`.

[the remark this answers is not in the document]
halseth: I see what you mean now, and you're right that is pretty nice :) Had to tweak the log messages a bit, but I think it works out: https://github.com/lightningnetwork/lnd/pull/1910/commits/ccf95de8a3ff7b816c0d833c254af56c4162cae6 👍

fundingmanager.go
```diff
@@ -484,230 +490,53 @@ func (f *fundingManager) start() error {
...
-				channelState)
+			}
 		}
+
+		// We will restart the funding state machine for all channels,
+		// which will wait for the channel's funding transaction to be
+		// confirmed on the blockchain, and transmit the messages
+		// necessary for the channel to be operational.
+		f.wg.Add(1)
+		go f.advanceFundingState(channel, chanID, nil)
```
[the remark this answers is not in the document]
halseth: It's only done at startup, shouldn't that be fine?

fundingmanager.go
```diff
@@ -933,6 +762,283 @@ func (f *fundingManager) reservationCoordinator() {
...
+			return fmt.Errorf("error sending channel "+
+				"announcement: %v", err)
+		}
+
+		// We delete the channel opening state from our internal
+		// database as the opening process has succeeded. We can do
+		// this because we assume the AuthenticatedGossiper queues the
+		// announcement messages, and persists them in case of a daemon
+		// shutdown.
+		err = f.deleteChannelOpeningState(&channel.FundingOutpoint)
```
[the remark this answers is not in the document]
halseth: Assumption is that `annAfterSixConfs` is idempotent (as is every other step in the state machine) and it will just be redone at restart.

fundingmanager.go
```diff
@@ -1971,37 +1916,108 @@ func (f *fundingManager) waitForFundingConfirmation(
...
 
+	// Now that the channel has been fully confirmed and we successfully
+	// saved the opening state, we'll mark it as open within the database.
+	if err := completeChan.MarkAsOpen(shortChanID); err != nil {
+		return fmt.Errorf("error setting channel pending flag to false: "+
+			"%v", err)
+	}
+
+	// Inform the ChannelNotifier that the channel has transitioned from
+	// pending open to open.
```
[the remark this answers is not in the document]
halseth: it is "pending open" to "open"

fundingmanager.go
```diff
@@ -1971,37 +1916,108 @@ func (f *fundingManager) waitForFundingConfirmation(
...
+		case epoch, ok := <-epochClient.Epochs:
+			if !ok {
+				timeoutChan <- fmt.Errorf("epoch client " +
+					"shutting down")
+				return
+			}
+
+			// Close the timeout channel and exit if the block is
+			// aboce the max height.
+			if uint32(epoch.Height) >= maxHeight {
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightningnetwork/lnd/pull/1910/commits/16ad4992708fc52236261c14f864d861ceedd069

fundingmanager.go
```diff
@@ -2098,8 +2074,18 @@ func (f *fundingManager) sendFundingLocked(peer lnpeer.Peer,
 	// send fundingLocked until we succeed, or the fundingManager is shut
 	// down.
 	for {
-		fndgLog.Debugf("Sending FundingLocked for ChannelID(%v) to "+
-			"peer %x", chanID, peerKey)
+		connected := make(chan lnpeer.Peer, 1)
+		f.cfg.NotifyWhenOnline(peerKey, connected)
```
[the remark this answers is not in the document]
wpaulino: Maybe not that crucial since we can't guarantee that they processed the message?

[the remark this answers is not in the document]
halseth: It definitely looks like it should be sync, otherwise the error from that call will be more or less meaningless. The same is true other places we send messages in this file.

It looks like it changed from bing sync to async here: https://github.com/lightningnetwork/lnd/pull/1505

Do you remember why it was changed, or if that was unintentional?

fundingmanager.go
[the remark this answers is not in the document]
cfromknecht: ah okay, i misread that 😅

fundingmanager.go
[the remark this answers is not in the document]
cfromknecht: gotcha, that should be fine then

fundingmanager.go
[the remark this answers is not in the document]
cfromknecht: you're right, i think that's fine as well

fundingmanager.go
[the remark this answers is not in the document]
cfromknecht: yeah the async variant can never return an error, which is fine. most likely the only real errors that would be returned here are shutting down errors or some error that indicates the write failed, e.g. the remote peer disconnects.

it should be fine though to leave it async, as @wpaulino points out, since we can never be sure if they process the message even if we want until it's written

halseth:
Squashed

fundingmanager.go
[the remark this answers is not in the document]
wpaulino: The change seems unintentional. Agreed offline that we'll address it as a separate PR.

PR #1911, opened 2018-09-14
server: start tor controller if we should listen for inbound connections

wpaulino:
over Tor

In this commit, we fix a small bug where we would attempt to start the
Tor controller even if we were not requested to automatically create and
onion service in order to listen for inbound connections over Tor.

### DIFF
--- a/server.go
+++ b/server.go
@@ -432,7 +432,7 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
 	// If we were requested to route connections through Tor and to
 	// automatically create an onion service, we'll initiate our Tor
 	// controller and establish a connection to the Tor server.
-	if cfg.Tor.Active {
+	if cfg.Tor.Active && (cfg.Tor.V2 || cfg.Tor.V3) {
 		s.torController = tor.NewController(cfg.Tor.Control)
 	}
 


qubenix:
This fixes my unexpected shutdown on Whonix, where the control commands are transparently proxied to Tor on the remote gateway and restricted [by design](https://www.whonix.org/wiki/Dev/Control_Port_Filter_Proxy).

server.go
```diff
@@ -432,7 +432,7 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
 	// If we were requested to route connections through Tor and to
 	// automatically create an onion service, we'll initiate our Tor
 	// controller and establish a connection to the Tor server.
-	if cfg.Tor.Active {
+	if cfg.Tor.Active && (cfg.Tor.V2 || cfg.Tor.V3) {
```
halseth: what if `disableListen` is set?

halseth: It will also be set `true` here: https://github.com/lightningnetwork/lnd/blob/master/config.go#L495

Desuuuu: My point is, when `disableListen` is set, you can be sure both `tor.v2` and `tor.v3` are off.

In the code you linked, `tor.v2` and `tor.v3` being off are a condition to set `disableListen`.

halseth: Ok, so we could do
```go
if cfg.Tor.Active && !cfg.DisableListen {
```
? 😄

Desuuuu: AFAICT, the result would be exactly the same!

Since this condition determines whether or not we should start the Tor controller (to automatically create a v2/v3 hidden service), I feel like the other is clearer 😸

PR #1926, opened 2018-09-17
Move daemon from main to lnd package

halseth:
This PR renames the `main` package to `lnd`, and moves the `lnd` entry point into `cmd/lnd`.

This intends to make `lnd` suitable for use as a library, for instance within in other apps. This is also a prerequisite for using `lnd` with `gomobile`.

### DIFF
--- a/lnd.go
+++ b/lnd.go
@@ -18,13 +18,15 @@ import (
 	"math/big"
 	"net"
 	"net/http"
-	_ "net/http/pprof"
 	"os"
 	"path/filepath"
 	"runtime/pprof"
 	"strings"
 	"sync"
 	"time"
+
+	// Blank import to set up profiling HTTP handlers.
+	_ "net/http/pprof"
 
 	"gopkg.in/macaroon-bakery.v2/bakery"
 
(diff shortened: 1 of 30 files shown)


lnd.go
```diff
@@ -92,7 +91,7 @@ var (
 // lndMain is the true entry point for lnd. This function is required since
 // defers created in the top-level scope of a main method aren't executed if
 // os.Exit() is called.
-func lndMain() error {
+func LndMain() error {
```
guggero: The comment should be upper case too. And the linter suggests to name this func `Main` because it is now in the package `lnd`.

yaslama:
What do you think about adding parameters to Main() in order to be able to run lnd as a library without lnd.conf ? In this case ,the args needs to be passed to loadConfig() as an array of strings and loadConfig needs to use flags.ParseArgs() instead of flags.Parse()

halseth:
@yaslama that's a good idea! Probably warrants its own PR though, but will definitely consider it after this change goes in 👍

Roasbeef:
Needs rebase!

halseth:
Rebased!

lnd.go
```diff
@@ -87,10 +88,11 @@ var (
...
 
-// lndMain is the true entry point for lnd. This function is required since
-// defers created in the top-level scope of a main method aren't executed if
-// os.Exit() is called.
-func lndMain() error {
+// Main is the true entry point for lnd. This function is required since defers
+// created in the top-level scope of a main method aren't executed if os.Exit()
+// is called.
+func Main() error {
+
```
wpaulino: Nit: extra newline.

halseth: Fixed.

PR #1932, opened 2018-09-18
tor/controller: fix response parsing for values containing the '=' character

### LINKED ISSUE #1931: Tor hidden service private key not saved
### Background

When listening through a Tor hidden service, the private key is not saved (most of the time).

### Your environment

* lnd v0.5-beta

### Steps to reproduce

* Start an instance of lnd with `--tor.active --tor.v2` or `--tor.active --tor.v3`

### Expected behaviour

The private key file corresponding to the Tor hidden service should be saved at the `tor.privatekeypath` path.

### Actual behaviour

The private key file is not saved, **unless** the private key does not contain the character `=` (which seems to happen on occasion when using a v2 hidden service).

### Details

The problem lies in the way the Tor controller is parsing replies:

https://github.com/lightningnetwork/lnd/blob/v0.5-beta/tor/controller.go#L148-L159

This part is actually discarding valid `key=value` pairs when the value contains an `=` character. As a result this condition evaluates to `false`:

https://github.com/lightningnetwork/lnd/blob/v0.5-beta/tor/controller.go#L506

Desuuuu:
Allow parsing `key=value` pairs from Tor control's responses when `value` contains one or more `=` characters.

Fixes #1931

### DIFF
--- a/tor/controller.go
+++ b/tor/controller.go
@@ -149,7 +149,7 @@ func parseTorReply(reply string) map[string]string {
 		// "KEY=VALUE". If the parameter doesn't contain "=", then we
 		// can assume it does not provide any other relevant information
 		// already known.
-		keyValue := strings.Split(content, "=")
+		keyValue := strings.SplitN(content, "=", 2)
 		if len(keyValue) != 2 {
 			continue
 		}


Roasbeef:
Excellent first time contribution! Thanks 🤗

PR #1936, opened 2018-09-19
conctractcourt: check sweep sanity + reliable publication

cfromknecht:
This PR modifies the signing and sweeping of commit to-remote sweeps and htlc success sweeps, such that we validate the generated transaction's sanity before publication. It also ensures that the resolvers checkpoint the transaction before publishing. This ensures that we will watch for the same txid after a restart in case a crash happens between publication and checkpointing. If such a crash happens, both will attempt to publish the sweep transaction again in order to ensure the transaction propagates.

In addition, we return an error if checkpointing fails from a number of places within the contract resovlers, and also ignore errors that can arise from duplicate publications of the same htlc success transaction.

Note: would like to add some unit tests exercising this behavior, though this brings the resolvers in line with the intended behavior of other subsystems performing similar functions.

### DIFF
--- a/contractcourt/contract_resolvers.go
+++ b/contractcourt/contract_resolvers.go
@@ -171,6 +171,7 @@ func (h *htlcTimeoutResolver) Resolve() (ContractResolver, error) {
 
 		if err := h.Checkpoint(h); err != nil {
 			log.Errorf("unable to Checkpoint: %v", err)
+			return nil, err
 		}
 	}
 
@@ -485,6 +486,7 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 			// Checkpoint our state.
 			if err := h.Checkpoint(h); err != nil {
 				log.Errorf("unable to Checkpoint: %v", err)
+				return nil, err
 			}
 		}
 
@@ -559,6 +561,7 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 
 		if err := h.Checkpoint(h); err != nil {
 			log.Errorf("unable to Checkpoint: %v", err)
+			return nil, err
 		}
 	}
 
@@ -1323,6 +1326,7 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 
 			if err := c.Checkpoint(c); err != nil {
 				log.Errorf("unable to Checkpoint: %v", err)
+				return nil, err
 			}
 		case <-c.Quit:
 			return nil, fmt.Errorf("quitting")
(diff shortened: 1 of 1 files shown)


joostjager:
Looking at this PR again, I realize it partially overlaps with https://github.com/lightningnetwork/lnd/pull/1978, at least the sanity checking.

Roasbeef:
Needs rebase!

contractcourt/contract_resolvers.go
```diff
@@ -1315,20 +1315,32 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
+		if err := c.Checkpoint(c); err != nil {
+			log.Errorf("unable to Checkpoint: %v", err)
+			return nil, err
+		}
+
+		// With the sweep transaction checkpointed, we'll fallthrough to
+		// actually publish the transaction. Upon restart, the resolver
+		// will immediately take the case below since the sweep tx is
+		// checkpointed.
+		fallthrough
```
Roasbeef: IMO the control flow is easier to follow if we retain the publishing of the tx in this case, and also the ones that need a tx published.

contractcourt/contract_resolvers.go
```diff
@@ -513,21 +513,21 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
...
-			// Finally, we'll broadcast the sweep transaction to
-			// the network.
-			//
-			// TODO(roasbeef): validate first?
-			if err := h.PublishTx(h.sweepTx); err != nil {
-				log.Infof("%T(%x): unable to publish tx: %v",
-					h, h.payHash[:], err)
-				return nil, err
-			}
+		// Regardless of whether an existing transaction was found or newly
```
Roasbeef: 👍

contractcourt/contract_resolvers.go
```diff
@@ -1258,18 +1258,33 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 		log.Infof("%T(%v): sweeping commit output with tx=%v", c,
 			c.chanPoint, spew.Sdump(c.sweepTx))
 
-		// Finally, we'll broadcast the sweep transaction to the
-		// network.
-		if err := c.PublishTx(c.sweepTx); err != nil {
+		// With the sweep transaction constructed, we'll now Checkpoint
+		// our state.
+		if err := c.Checkpoint(c); err != nil {
+			log.Errorf("unable to Checkpoint: %v", err)
```
halseth: Print resolvertype+channelpoint?

Roasbeef: We'll print the chan point in the char arb.

PR #1937, opened 2018-09-19
Data loss protect resending

halseth:
This PR makes the latest channel reestablishment message being stored as part of the channel close summary, such that it can be sent even when a channel ahs been closed.

This makes it possible for nodes to recover from the case where they have been offline, lost state, and comes back online, attempting to reseync the channel.

Note: This PR contains a DB migration. To ensure deserialization code is not changed in the future, and breaking old migrations, a versioned copy of the current deserialization logic is added to the file `legacy_serialization.go`. Backwards compatible addition of the new field turned out to be complex because of the presence of optional fields (see the now removed commit https://github.com/lightningnetwork/lnd/commit/14648b236fa5f203a76e876d75026735f2356d1f).

TODO:
- [x] Unit test for `FetchClosedChannelForID`

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -2059,7 +2059,12 @@ func serializeChannelCloseSummary(w io.Writer, cs *ChannelCloseSummary) error {
 	// If this is a close channel summary created before the addition of
 	// the new fields, then we can exit here.
 	if cs.RemoteCurrentRevocation == nil {
-		return nil
+		return WriteElements(w, false)
+	}
+
+	// If fields are present, write boolean to indicate this, and continue.
+	if err := WriteElements(w, true); err != nil {
+		return err
 	}
 
 	if err := WriteElements(w, cs.RemoteCurrentRevocation); err != nil {
@@ -2070,14 +2075,22 @@ func serializeChannelCloseSummary(w io.Writer, cs *ChannelCloseSummary) error {
 		return err
 	}
 
-	// We'll write this field last, as it's possible for a channel to be
-	// closed before we learn of the next unrevoked revocation point for
-	// the remote party.
-	if cs.RemoteNextRevocation == nil {
-		return nil
+	// The RemoteNextRevocation field is optional, as it's possible for a
+	// channel to be closed before we learn of the next unrevoked
+	// revocation point for the remote party. Write a boolen indicating
+	// whether this field is present or not.
+	if err := WriteElements(w, cs.RemoteNextRevocation != nil); err != nil {
+		return err
 	}
 
-	return WriteElements(w, cs.RemoteNextRevocation)
+	// Write the field, if present.
+	if cs.RemoteNextRevocation != nil {
+		if err = WriteElements(w, cs.RemoteNextRevocation); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func fetchChannelCloseSummary(tx *bolt.Tx,
@@ -2111,15 +2124,19 @@ func deserializeCloseChannelSummary(r io.Reader) (*ChannelCloseSummary, error) {
 
 	// We'll now check to see if the channel close summary was encoded with
 	// any of the additional optional fields.
-	err = ReadElements(r, &c.RemoteCurrentRevocation)
-	switch {
-	case err == io.EOF:
-		return c, nil
+	var hasNewFields bool
+	err = ReadElements(r, &hasNewFields)
+	if err != nil {
+		return nil, err
+	}
 
-	// If we got a non-eof error, then we know there's an actually issue.
-	// Otherwise, it may have been the case that this summary didn't have
-	// the set of optional fields.
-	case err != nil:
+	// If fields are not present, we can return.
+	if !hasNewFields {
+		return c, nil
+	}
+
+	// Otherwise read the new fields.
+	if err := ReadElements(r, &c.RemoteCurrentRevocation); err != nil {
 		return nil, err
 	}
 
@@ -2129,15 +2146,20 @@ func deserializeCloseChannelSummary(r io.Reader) (*ChannelCloseSummary, error) {
 
 	// Finally, we'll attempt to read the next unrevoked commitment point
 	// for the remote party. If we closed the channel before receiving a
-	// funding locked message, then this can be nil. As a result, we'll use
-	// the same technique to read the field, only if there's still data
-	// left in the buffer.
-	err = ReadElements(r, &c.RemoteNextRevocation)
-	if err != nil && err != io.EOF {
-		// If we got a non-eof error, then we know there's an actually
-		// issue. Otherwise, it may have been the case that this
-		// summary didn't have the set of optional fields.
+	// funding locked message then this might not be present. A boolean
+	// indicating whether the field is present will come first.
+	var hasRemoteNextRevocation bool
+	err = ReadElements(r, &hasRemoteNextRevocation)
+	if err != nil {
 		return nil, err
+	}
+
+	// If this field was written, read it.
+	if hasRemoteNextRevocation {
+		err = ReadElements(r, &c.RemoteNextRevocation)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return c, nil
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -3,6 +3,7 @@ package channeldb
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -1832,6 +1833,10 @@ type ChannelCloseSummary struct {
 
 	// LocalChanCfg is the channel configuration for the local node.
 	LocalChanConfig ChannelConfig
+
+	// LastChanSyncMsg is the ChannelReestablish message for this channel
+	// for the state at the point where it was closed.
+	LastChanSyncMsg *lnwire.ChannelReestablish
 }
 
 // CloseChannel closes a previously active Lightning channel. Closing a channel
@@ -2090,6 +2095,18 @@ func serializeChannelCloseSummary(w io.Writer, cs *ChannelCloseSummary) error {
 		}
 	}
 
+	// Write whether the channel sync message is present.
+	if err := WriteElements(w, cs.LastChanSyncMsg != nil); err != nil {
+		return err
+	}
+
+	// Write the channel sync message, if present.
+	if cs.LastChanSyncMsg != nil {
+		if err := WriteElements(w, cs.LastChanSyncMsg); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
@@ -2160,6 +2177,32 @@ func deserializeCloseChannelSummary(r io.Reader) (*ChannelCloseSummary, error) {
 		if err != nil {
 			return nil, err
 		}
+	}
+
+	// Check if we have a channel sync message to read.
+	var hasChanSyncMsg bool
+	err = ReadElements(r, &hasChanSyncMsg)
+	if err == io.EOF {
+		return c, nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	// If a chan sync message is present, read it.
+	if hasChanSyncMsg {
+		// We must pass in reference to a lnwire.Message for the codec
+		// to support it.
+		var msg lnwire.Message
+		if err := ReadElements(r, &msg); err != nil {
+			return nil, err
+		}
+
+		chanSync, ok := msg.(*lnwire.ChannelReestablish)
+		if !ok {
+			return nil, errors.New("unable cast db Message to " +
+				"ChannelReestablish")
+		}
+		c.LastChanSyncMsg = chanSync
 	}
 
 	return c, nil
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -12,6 +12,7 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/coreos/bbolt"
 	"github.com/go-errors/errors"
+	"github.com/lightningnetwork/lnd/lnwire"
 )
 
 const (
@@ -609,6 +610,54 @@ func (d *DB) FetchClosedChannel(chanID *wire.OutPoint) (*ChannelCloseSummary, er
 		chanSummary, err = deserializeCloseChannelSummary(summaryReader)
 
 		return err
+	}); err != nil {
+		return nil, err
+	}
+
+	return chanSummary, nil
+}
+
+// FetchClosedChannelForID queries for a channel close summary using the
+// channel ID of the channel in question.
+func (d *DB) FetchClosedChannelForID(cid lnwire.ChannelID) (
+	*ChannelCloseSummary, error) {
+
+	var chanSummary *ChannelCloseSummary
+	if err := d.View(func(tx *bolt.Tx) error {
+		closeBucket := tx.Bucket(closedChannelBucket)
+		if closeBucket == nil {
+			return ErrClosedChannelNotFound
+		}
+
+		// The first 30 bytes of the channel ID and outpoint will be
+		// equal.
+		cursor := closeBucket.Cursor()
+		op, c := cursor.Seek(cid[:30])
+
+		// We scan over all possible candidates for this channel ID.
+		for ; op != nil && bytes.Compare(cid[:30], op[:30]) <= 0; op, c = cursor.Next() {
+			var outPoint wire.OutPoint
+			err := readOutpoint(bytes.NewReader(op), &outPoint)
+			if err != nil {
+				return err
+			}
+
+			// If the found outpoint does not correspond to this
+			// channel ID, we continue.
+			if !cid.IsChanPoint(&outPoint) {
+				continue
+			}
+
+			// Deserialize the close summary and return.
+			r := bytes.NewReader(c)
+			chanSummary, err = deserializeCloseChannelSummary(r)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		}
+		return ErrClosedChannelNotFound
 	}); err != nil {
 		return nil, err
 	}
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -80,6 +80,13 @@ var (
 			number:    6,
 			migration: migratePruneEdgeUpdateIndex,
 		},
+		{
+			// The DB version that migrates the ChannelCloseSummary
+			// to a format where optional fields are indicated with
+			// boolean flags.
+			number:    7,
+			migration: migrateOptionalChannelCloseSummaryFields,
+		},
 	}
 
 	// Big endian is the preferred byte order, due to cursor scans over
--- a/channeldb/db_test.go
+++ b/channeldb/db_test.go
@@ -5,6 +5,9 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lnwire"
 )
 
 func TestOpenWithCreate(t *testing.T) {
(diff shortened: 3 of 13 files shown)


channeldb/channel.go
```diff
@@ -2203,6 +2224,32 @@ func deserializeCloseChannelSummary(r io.Reader) (*ChannelCloseSummary, error) {
 		}
 	}
 
+	// Check if we have a channel sync message to read.
+	var hasChanSyncMsg bool
+	err = ReadElements(r, &hasChanSyncMsg)
+	if err == io.EOF {
```
halseth: Note: we can add write the `false` marker here as part of the migration to avoid this `EOF` check.

halseth: Currently the addition of the `LastChanSyncMsg` field is kept in its own commit to show how new fields can easily be added after the presence flags are introduced in the serialization.

channeldb/channel.go
```diff
@@ -2203,6 +2224,32 @@ func deserializeCloseChannelSummary(r io.Reader) (*ChannelCloseSummary, error) {
...
+		return c, nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	// If a chan sync message is present, read it.
+	if hasChanSyncMsg {
+		// We must pass in reference to a lnwire.Message for the codec
+		// to support it.
+		var msg lnwire.Message
```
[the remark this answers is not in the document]
cfromknecht: i'm guessing that `**lnwire.ChannelReestablish` and `*lnwire.Message` are distinct concrete types, and the type switch doesn't try to determine if the suffix `*lnwire.ChannelReestablish` implements the other suffix `lnwire.Message`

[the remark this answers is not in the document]
halseth: Yes, I think that's right. I think the type switch is checking if the passed `interface{}` is of concrete type `*lnwire.Message`, not if it is satisfying the interface. TIL you cannot do that: https://play.golang.org/p/SmL37Hobvvs

channeldb/db.go
```diff
@@ -616,6 +617,55 @@ func (d *DB) FetchClosedChannel(chanID *wire.OutPoint) (*ChannelCloseSummary, er
 	return chanSummary, nil
 }
 
+// FetchClosedChannelForID queries for a channel close summary using the
+// channel ID of the channel in question.
+func (d *DB) FetchClosedChannelForID(cid lnwire.ChannelID) (
```
Roasbeef: This as is scales with the number of closed channels, which will increase for all time. Instead, we can use a different algorithm, which will be constant-time ish, as in we know the hard upper bound

  1. Fetch the entire set of channel point candidates for a channel ID via `GenPossibleOutPoints`
  2. Run through each of these, and attempt to read the key from the `closedChannelBucket`

This should terminate in 1, and sometimes two iterations. Only once people start doing this like splicing or funding coin joins will this start to swell. Even then, it should be well below 1000 or so iterations.

channeldb/channel.go
```diff
@@ -3,6 +3,7 @@ package channeldb
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
```
Roasbeef: Lingering TODO in the commit.

channeldb/channel.go
Roasbeef: Why are we mixing the two approaches? We should either use EOF everywhere (so no migration) or the bool based approach.

Roasbeef: Ah scratch that, I see the usage now

contractcourt/chain_watcher.go
```diff
@@ -488,6 +489,19 @@ func (c *chainWatcher) toSelfAmount(tx *wire.MsgTx) btcutil.Amount {
...
 
+// chanSyncMsg fetches the latest channel sync message for the channel.
+func (c *chainWatcher) chanSyncMsg() (*lnwire.ChannelReestablish, error) {
+	lnChannel, err := lnwallet.NewLightningChannel(
+		nil, nil, c.cfg.chanState,
+	)
+	if err != nil {
+		return nil, err
+	}
+	defer lnChannel.Stop()
```
Roasbeef: Bit of a code smell that we need to do this (make a set of goroutines) just to grab this field. Instead we can [efactor this method slightly](https://github.com/lightningnetwork/lnd/blob/5803da382a4d16c155cd5396d585c0340736ecad/lnwallet/channel.go#L3449:29) to make a (possible private version) that accepts the channel state and commitments heights and use that in place. Recall the issues we had in the funding manager that caused us to just pass around the channel state everywhere vs the full state machine.

contractcourt/chain_watcher.go
```diff
@@ -590,6 +613,15 @@ func (c *chainWatcher) dispatchLocalForceClose(
 		closeSummary.TimeLockedBalance += htlcValue
 	}
 
+	// Attempt to add a channel sync message to the close summary.
```
Roasbeef: What about remote force close?

contractcourt/chain_watcher.go
```diff
@@ -749,6 +781,15 @@ func (c *chainWatcher) dispatchContractBreach(spendEvent *chainntnfs.SpendDetail
...
+	chanSync, err := c.chanSyncMsg()
+	if err != nil {
+		log.Errorf("ChannelPoint(%v): unable to create channel sync "+
+			"message: %v", c.cfg.chanState.FundingOutpoint, err)
+	} else {
+		closeSummary.LastChanSyncMsg = chanSync
+	}
+
```
Roasbeef: Unrelated to this PR, but shouldn't the `BRAR` be the one that's making this DB state here? Re the whole reliable hand off thing.

lnd_test.go
```diff
@@ -276,6 +276,58 @@ func closeChannelAndAssert(ctx context.Context, t *harnessTest,
 	return closingTxid
 }
 
+// waitForChannelPendingForceClose waits for the node to report that the
```
Roasbeef: This commit can be deleted when rebasing to master as it was merged on its own.

peer.go
```diff
@@ -868,7 +868,33 @@ func newChanMsgStream(p *peer, cid lnwire.ChannelID) *msgStream {
...
+				switch {
+
+				// If we failed to find the link in question,
+				// and the message received was a channel sync
+				// message, then this might be a peer trying to
+				// resync closed channel. In this case we'll
+				// try to resend our last channel sync message,
+				// such that the peer can recover funds from
+				// the closed channel.
+				case err != nil && isChanSyncMsg:
```

contractcourt/chain_watcher.go
```diff
@@ -403,16 +404,41 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
-				log.Errorf("Unable to retrieve commitment "+
-					"point for channel(%v) with lost "+
-					"state: %v",
-					c.cfg.chanState.FundingOutpoint, err)
-				return
+			// can sweep our funds. If we cannot find the commit
+			// point, there's not much we can do other than wait
+			// for us to retrieve it. We will attempt to retrieve
+			// it from the peer each time we connect to it.
+			// TODO(halseth): actively initiate re-connection to
```
Roasbeef: This can eventually be hooked up into the channel life cycle management stuff. So to request a notification once a new node connects. Actually we already have this for use in the funding manager, so perhaps it can be used here?

lnd_test.go
```diff
@@ -7198,6 +7198,85 @@ func testDataLossProtection(net *lntest.NetworkHarness, t *harnessTest) {
 
 	assertNodeNumChannels(t, ctxb, dave, 0)
 	assertNodeNumChannels(t, ctxb, carol, 0)
+
```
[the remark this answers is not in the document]
Roasbeef: Least blocking route would be a comment that delineates this new testing scenario.

contractcourt/chain_watcher.go
cfromknecht: the brar writes the retributions to disk, then acks back. after that the chain watcher knows it's safe to mark the channel as pending closed

contractcourt/chain_watcher.go
cfromknecht: yeah i think we can use `NotifyWhenOnline` to just try if they ever reconnect

channeldb/db.go
Roasbeef: Actually scratch that, I properly see the seek!

channeldb/channel.go
```diff
@@ -2094,7 +2094,12 @@ func serializeChannelCloseSummary(w io.Writer, cs *ChannelCloseSummary) error {
 	// If this is a close channel summary created before the addition of
 	// the new fields, then we can exit here.
 	if cs.RemoteCurrentRevocation == nil {
-		return nil
+		return WriteElements(w, false)
```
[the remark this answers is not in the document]
halseth: The problem with optional fields is that it only works for the last field. If there are more than one optional field, then you won't know what the data present represents. This way we attempt to make it future proof.

By adding booleans to indicate presence, we can make sure old code will read the new fields (they won't attempt to read any extra data they don't understand) and that new code will read the old format (booleans will indicate whether fields are present, so it can know which fields are not present from the old format).

channeldb/channel.go
```diff
@@ -2105,14 +2110,28 @@ func serializeChannelCloseSummary(w io.Writer, cs *ChannelCloseSummary) error {
...
-	// We'll write this field last, as it's possible for a channel to be
-	// closed before we learn of the next unrevoked revocation point for
-	// the remote party.
+	// The RemoteNextRevocation field is optional, as it's possible for a
+	// channel to be closed before we learn of the next unrevoked
+	// revocation point for the remote party. Write a boolen indicating
+	// whether this field is present or not.
 	if cs.RemoteNextRevocation == nil {
-		return nil
+		if err := WriteElements(w, false); err != nil {
```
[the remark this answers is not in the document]
halseth: Won't work for the case where `cs.RemoteNextRevocation == nil` and `chanSyncMsg != nil`, as the deserialization code will attempt to read the `chanSyncMsg` as `RemoteNextRevocation`.

channeldb/migrations.go
```diff
@@ -573,3 +573,40 @@ func migratePruneEdgeUpdateIndex(tx *bolt.Tx) error {
 
 	return nil
 }
+
+// migrateOptionalChannelCloseSummaryFields migrates the serialized format of
```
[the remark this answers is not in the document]
halseth: I first attempted to do this without a migration, but it turned out to be complex: https://github.com/lightningnetwork/lnd/commit/14648b236fa5f203a76e876d75026735f2356d1f

channeldb/channel.go
halseth: This is how you would add new fields in the future without a migration.

channeldb/channel.go
halseth: I think it makes sense to NOT do it in the migration (the only difference would be that we could remove the EOF check below), so will remove the TODO from the commit message if no protests.

contractcourt/chain_watcher.go
halseth: We could let the `BRAR` have that responsibility, as we could get rid of ACK logic.

contractcourt/chain_watcher.go
halseth: Fixed!

contractcourt/chain_watcher.go
halseth: Good catch! Fixed.

lnd_test.go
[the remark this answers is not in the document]
halseth: Improved the comments.

Agreed that some of the tests would benefit from being split up. I think this might cause our integration tests to run even longer though, bc of the added overhead. Might be worth looking into in combination with running integration tests in parallel (perhaps using containers?).

lnd_test.go
halseth: Deleted.

lnd_test.go
```diff
@@ -6839,183 +6891,229 @@ func testDataLossProtection(net *lntest.NetworkHarness, t *harnessTest) {
...
-		ctxt, t, net, carol, dave,
-		lntest.OpenChannelParams{
-			Amt: chanAmt,
-		},
-	)
+	// We create a helper method that will create a slice of payment
+	// requests for us. The incrementing invoiceIndex will ensure they are
+	// unique.
+	invoiceIndex := 0
+	createPayReqs := func(node *lntest.HarnessNode,
```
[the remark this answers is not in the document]
halseth: Done!

contractcourt/chain_watcher.go
halseth: I actually like the current polling more, since we might need some time to process the `ChannelReestablish` after they reconnect. In that case the first lookup after they connect might fail, so we have to wait a little bit anyway. 

If we don't wait before retrying at that point, we risk the peer coming online send us the `ChannelReestablish` and disconnects, and we will stall here waiting for them to connect again.

halseth:
Comments address, PTAL 🐶

contractcourt/chain_watcher.go
cfromknecht: the cleaner solution IMO is just to write resolutions in `CloseChannel`. then there is no need for a reliable handoff. the resolutions can be passed in memory to kick off the process, but if we crash the brar would look for all breached channels in the db and load resolutions from there

channeldb/channel.go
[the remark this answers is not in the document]
Roasbeef: Yep, realized this later that day. I think in the future, we should do a clean sweep to just move to a TLV format everywhere. You read it all, then at the end check the array/map for the keys you actually know of then parse that. With that, we gain a way to handle an arbitrary number of fields in the future in a uniform manner throughout the codebase.

channeldb/migrations.go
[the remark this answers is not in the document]
Roasbeef: Ah ok, yeah that's pretty convoluted.

lnd_test.go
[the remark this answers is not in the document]
Roasbeef: 👍 

Yeah we'll need to white board out a better architecture as we add more backends, and also eventually starting other platforms other than mac OS.

channeldb/channel.go
Roasbeef: SGTM

channeldb/db_test.go
```diff
@@ -71,3 +74,76 @@ func TestWipe(t *testing.T) {
 			ErrNoClosedChannels, err)
 	}
 }
+
+// TestFetchClosedChannelForID tests that we are able to properly retrieve a
```
Roasbeef: 👍

contractcourt/chain_watcher.go
```diff
@@ -402,16 +403,39 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
-			if err != nil {
+			// can sweep our funds. If we cannot find the commit
+			// point, there's not much we can do other than wait
+			// for us to retrieve it. We will attempt to retrieve
+			// it from the peer each time we connect to it.
+			// TODO(halseth): actively initiate re-connection to
+			// the peer?
+			var commitPoint *btcec.PublicKey
+			backoff := time.Second
+			maxBackoff := time.Hour
```
Roasbeef: This seems a bit long of a max back off, perhaps something in the 10 of minutes instead?

contractcourt/chain_watcher.go
Roasbeef: Don't see it fixed in the latest diff: https://github.com/lightningnetwork/lnd/blob/4124c6db8159bf79327181029f613fc9d404f4a1/contractcourt/chain_watcher.go#L654

lnwallet/channel.go
```diff
@@ -1023,11 +1023,10 @@ type commitmentChain struct {
...
 }
 
-// newCommitmentChain creates a new commitment chain from an initial height.
-func newCommitmentChain(initialHeight uint64) *commitmentChain {
+// newCommitmentChain creates a new commitment chain.
+func newCommitmentChain() *commitmentChain {
 	return &commitmentChain{
-		commitments:    list.New(),
-		startingHeight: initialHeight,
+		commitments: list.New(),
```
Roasbeef: Where do we set the new starting height of a commitment chain? Don't see why this change is necessary atm. There're other spots where we depend on the height being set properly.

lnwallet/channel.go
```diff
@@ -5113,6 +5112,15 @@ func NewUnilateralCloseSummary(chanState *channeldb.OpenChannel, signer Signer,
 		LocalChanConfig:         chanState.LocalChanCfg,
 	}
 
+	// Attempt to add a channel sync message to the close summary.
+	chanSync, err := ChanSyncMsg(chanState)
```
Roasbeef: Ah ok I see this now re remote close. We could also shuffle the other spots we populate this information into this package. Don't consider it a blocker though.

Roasbeef:
Final lingering comments are re if we should have a lower backoff, and also if a minor refactor is necessary or not given.

lnwallet/channel.go
```diff
@@ -3446,27 +3445,27 @@ func (lc *LightningChannel) ProcessChanSyncMsg(
 //      it.
 //   3. We didn't get the last RevokeAndAck message they sent, so they'll
 //      re-send it.
-func (lc *LightningChannel) ChanSyncMsg() (*lnwire.ChannelReestablish, error) {
+func ChanSyncMsg(c *channeldb.OpenChannel) (*lnwire.ChannelReestablish, error) {
```
cfromknecht: should this have a mutex around it? esp now that this is called in more places than just the switch

also seems like something that should be implemented in channeldb as a method of `OpenChannel`, since it is otherwise unrelated to anything in `lnwallet`. would be nice to keep the `.ChanSyncMsg()` notation, and not use a package level function

channeldb/migrations.go
```diff
@@ -573,3 +573,40 @@ func migratePruneEdgeUpdateIndex(tx *bolt.Tx) error {
...
+// ChannelCloseSummary to a format where optional fields' presence is indicated
+// with boolean markers.
+func migrateOptionalChannelCloseSummaryFields(tx *bolt.Tx) error {
```
cfromknecht: would be nice to have a unit test for this migration, asserting it behaves properly in all four cases

channeldb/channel.go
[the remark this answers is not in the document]
halseth: ooooh, yeah that would be nice!

lnwallet/channel.go
halseth: Good idea, added mutex.

The reason I didn't make this a method on `OpenChannel` is that it calls `ComputeCommitmentPoint`, which is a method from `lnwallet`, leading to an import cycle. We could however move this method to a new package (`lnutil`? 😛) if we want this, but I felt it was not worth it.

Also moving it to `channeldb` felt wrong 🤔

contractcourt/chain_watcher.go
```diff
@@ -403,16 +404,41 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
+			for {
```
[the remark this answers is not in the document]
halseth: kewl, sticking with the fixup for now

lnwallet/channel.go
halseth: We never read it, we just look at the height of the commitment in the chain.

Agreed that the change is not directly relevant to this PR, I just noticed when transforming the `ChanSync` method to not being dependent on the chain heights.

Removed and made a separate PR: https://github.com/lightningnetwork/lnd/pull/2206

channeldb/legacy_serialization.go
```diff
@@ -0,0 +1,53 @@
+package channeldb
+
+import "io"
+
+// deserializeCloseChannelSummaryV6 reads the v6 database format for
+// ChannelCloseSummary.
+//
+// NOTE: deprecated, only for migration.
+func deserializeCloseChannelSummaryV6(r io.Reader) (*ChannelCloseSummary, error) {
```
halseth: This was after an idea by @joostjager. Lmk what you think :)

channeldb/migrations.go
halseth: Migration added! (only three cases though?)

halseth:
Comments addressed and commits squashed. PTAL.

Issue #1513, opened 2018-07-07
routing+rpc+lncli: add new absolute time lock limit within path finding 


Roasbeef:
Currently when sending payments, we allow the user to specify a max fee value, either as a percentage of an absolute amount. This is very useful for services that implement any sort of withdrawal as they can be sure the user isn't attempting to force them to transport the HTLC over an edge that has abnormally high fees. 

A similar constraint should be added, but this time for the _total absolute_ time lock value for a payment. This allows users to express a ceiling on the time value of their Bitcoin in the worst case. For example, the user can ensure that their funds will be suspended for _at most_ 1 day in the case of a total breakdown along each hop of the path.

# Steps To Completion 

- [ ] Add a new `max_cltv_delay` field to the `SendPayment` RPC, and expose it on the command line. 

- [ ] Add a similar field to the `QueryRoutes` RPC, as this can let clients further constrain the set of paths that should be returned. 

- [ ] Modify `rpcserver.go` to pass through the new field into the `ChannelRouter` for both `QueryRoutes` and all variants of `SendPayment`. 

- [ ] During path finding, maintain a `cltv_budget`, and ignore an edge if traversing it would exceed this budget. 

- [ ] As always, as set of tests should be added to exercise all of the changes above. 

The implementation of this should likely wait until #1321 has been merged.

Bluetegu:
I'd like to work on this

ZapUser77:
@Bluetegu Are you still working on this one?

Bluetegu:
Sounds good

johng:
@ZapUser77 are you still working on this?

wpaulino:
Done in #2640.

PR #1941, opened 2018-09-19
Absolute Time Lock Limit

johng:
As per #1513 I've added a field to `queryroutes` and `sendpayment` which limits the maximum total time funds can be locked in htlcs along a given route. 

The main changes involve replacing the maximum fee parameter with a struct which contains both restrictions.


routing/pathfind_test.go
```diff
@@ -668,20 +692,23 @@ func testBasicGraphPathFindingCase(t *testing.T, graph *channeldb.ChannelGraph,
...
-		if err == nil {
-			t.Fatal("expected no path to be found")
-		}
-		return
-	}
+
 	if err != nil {
-		t.Fatalf("unable to find path: %v", err)
+
+		if IsError(err, test.expectedFailure) {
```
joostjager: If think the default expectedFailure (if not specified in the testcase) is ErrNoPathFound. So what happens here is that if an unexpected error occurs and it happens to be ErrNoPathFound, the test will pass?

routing/pathfind_test.go
```diff
@@ -1124,9 +1156,19 @@ func TestNewRoute(t *testing.T) {
 				createHop(0, 1000, 1000000, 144),
 				createHop(0, 1000, 1000000, 144),
 			},
-			feeLimit:          0,
+			pathRestrictions:  PathRestrictions{FeeLimit: 0, MaxCltvDelay: noCltvLimit},
 			expectError:       true,
 			expectedErrorCode: ErrFeeLimitExceeded,
+		}, {
+			name:          "two hop failure with fee limit (zero)",
```
joostjager: Check name

routing/router.go
```diff
@@ -1523,6 +1523,20 @@ func generateSphinxPacket(route *Route, paymentHash []byte) ([]byte,
...
+type PathRestrictions struct {
+
+	// FeeLimit is the maximum fee in millisatoshis that the payment should
+	// accept when sending it through the network. The payment will fail
+	// if there isn't a route with lower fees than this limit.
+	FeeLimit lnwire.MilliSatoshi
+
+	// MaxCltvDelay is the maximum absolute time lock value accross the
+	// hops in the payment
+	MaxCltvDelay uint32
```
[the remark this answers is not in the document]
joostjager: Also, we might consider moving the ignore edges and vertices in here too. Those are a restriction as well, although not provided by an external source but by our own mission control. Probably good to reduce the number of args to findPaths a bit.

cmd/lncli/commands.go
```diff
@@ -2520,11 +2528,14 @@ func getNodeInfo(ctx *cli.Context) error {
...
-	Category:    "Payments",
-	Usage:       "Query a route to a destination.",
-	Description: "Queries the channel router for a potential path to the destination that has sufficient flow for the amount including fees",
-	ArgsUsage:   "dest amt",
+	Name:     "queryroutes",
+	Category: "Payments",
+	Usage:    "Query a route to a destination.",
+	Description: `
+			Queries the channel router for a potential path to the destination that has sufficient flow for 
+			the amount including fees. By default the fee limit capped at the send amount.
```
joostjager: Capped at send amount, is this implemented? I think the default is no cap?

johng: It defaults to the amt sent in the rpc server. https://github.com/lightningnetwork/lnd/blob/4c0ca3717423367be59973928b5ff908ff260259/rpcserver.go#L1969

routing/pathfind_test.go
```diff
@@ -47,6 +47,9 @@ const (
 	// can use this value to signal there is no fee limit since payments
 	// should never be larger than this.
 	noFeeLimit = lnwire.MilliSatoshi(math.MaxUint32)
+
+	// noFeeLimit
+	noCltvLimit = math.MaxUint32
```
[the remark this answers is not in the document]
johng: We don't do any casting from uints to ints. If we do in the future that will cause issues if the limit is sufficiently large but should be picked up by the test cases

joostjager:
I haven't looked at the code changes themselves yet, but the commit structure is not yet what it should be. You should use `git rebase` to get it right. Commits that fix problems that were introduced in earlier commits that are also part of this PR should be squashed together. 

The end result should be a set of commits that build on top of each other and it should look like it was done first time right.

joostjager:
The branch needs to be rebased onto master as well to fix conflicts.

joostjager:
Breaking it up logically makes reviewing quicker, because it is easier to assess unrelated changes separately. Reviewing before clean up means that a second review (albeit shorted) is needed after clean up.

Also, when people look at a particular change in the future (after merge), they are directed to a logical commit of which that change was part. This helps to understand the intention of the change.

In this PR, there are so many commits, that is probably easier to start with all changes in the working directory and then stage and commit bit by bit (see structure that I suggested earlier). So rebuild the commit structure from scratch rather than trying to transform the current commits into the intended structure.

joostjager:
There are some commits from others in the diff now that don't belong in this PR.

Also, not all of the changes that you first had are in?

W/r to commit structure, the idea is to group the changes functionally. Like in the example breakdown I gave above. So "giving test cases a name" is a commit that can exist in isolation. Also it is preferred to have commits building successfully.

routing/heap.go
```diff
@@ -21,6 +21,9 @@ type nodeWithDist struct {
 	// amount that includes also the fees for subsequent hops.
 	amountToReceive lnwire.MilliSatoshi
 
+	// Total cltv amount from this node to the destination
```
halseth: nit: missing period.

Also, should this be `destination` or `source`? I see `source` is used elsewhere in this struct, but I guess we changed to destination at one point.

routing/pathfind.go
```diff
@@ -279,8 +279,14 @@ func newRoute(amtToSend, feeLimit lnwire.MilliSatoshi, sourceVertex Vertex,
 		// backwards below, this next hop gets closer and closer to the
 		// sender of the payment.
 		nextIncomingAmount lnwire.MilliSatoshi
+
+		// Total intermediate cltv
```
halseth: missing period. Here and other places.

routing/pathfind.go
```diff
@@ -311,6 +317,11 @@ func newRoute(amtToSend, feeLimit lnwire.MilliSatoshi, sourceVertex Vertex,
 			fee = computeFee(amtToForward, pathEdges[i+1])
 		}
 
+		// Keep track of the total intermediate cltv in the path
+		if i != len(pathEdges)-1 && i != 0 {
```
halseth: The boolean condition should have an explanation.

routing/pathfind.go
```diff
@@ -361,9 +372,13 @@ func newRoute(amtToSend, feeLimit lnwire.MilliSatoshi, sourceVertex Vertex,
...
 	// Invalidate this route if its total fees exceed our fee limit.
 	if newRoute.TotalFees > feeLimit {
-		err := fmt.Sprintf("total route fees exceeded fee "+
+		return nil, newErrf(ErrFeeLimitExceeded, "total route fees exceeded fee "+
 			"limit of %v", feeLimit)
-		return nil, newErrf(ErrFeeLimitExceeded, err)
+	}
+
+	if maxCltvDelay > 0 && totalIntermediateCltv > maxCltvDelay {
+		return nil, newErrf(ErrMaxCltvDelayExceeded, "total intermidate cltv delay exceeded "+
```
halseth: long line

routing/pathfind.go
```diff
@@ -361,9 +372,13 @@ func newRoute(amtToSend, feeLimit lnwire.MilliSatoshi, sourceVertex Vertex,
 
 	// Invalidate this route if its total fees exceed our fee limit.
 	if newRoute.TotalFees > feeLimit {
-		err := fmt.Sprintf("total route fees exceeded fee "+
+		return nil, newErrf(ErrFeeLimitExceeded, "total route fees exceeded fee "+
```
halseth: long line

routing/pathfind.go
```diff
@@ -361,9 +372,13 @@ func newRoute(amtToSend, feeLimit lnwire.MilliSatoshi, sourceVertex Vertex,
...
 			"limit of %v", feeLimit)
-		return nil, newErrf(ErrFeeLimitExceeded, err)
+	}
+
+	if maxCltvDelay > 0 && totalIntermediateCltv > maxCltvDelay {
```
halseth: this needs a comment as well.

routing/pathfind.go
```diff
@@ -461,10 +476,16 @@ func findPath(tx *bolt.Tx, graph *channeldb.ChannelGraph,
...
-	amt lnwire.MilliSatoshi, feeLimit lnwire.MilliSatoshi,
+	amt lnwire.MilliSatoshi, pathRestrictions PathRestrictions,
 	bandwidthHints map[uint64]lnwire.MilliSatoshi) ([]*channeldb.ChannelEdgePolicy, error) {
 
-	var err error
+	var (
+		err error
+	)
+
+	feeLimit := pathRestrictions.FeeLimit
```
halseth: move these inside the `var ( .. )` as well?

routing/pathfind.go
```diff
@@ -599,6 +621,19 @@ func findPath(tx *bolt.Tx, graph *channeldb.ChannelGraph,
 			timeLockDelta = edge.TimeLockDelta
 		}
 
+		// We only calculate the timelockDelta for intermediate hops
```
halseth: It is not obvious to me why this is the case, so add to the comment why.

routing/router.go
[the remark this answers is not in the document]
halseth: Note that this might have to change with the merge of https://github.com/lightningnetwork/lnd/pull/2115

cmd/lncli/commands.go
```diff
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/lightningnetwork/lnd/routing"
```
halseth: import not ordered.

cmd/lncli/commands.go
```diff
@@ -2728,6 +2750,7 @@ func queryRoutes(ctx *cli.Context) error {
 		Amt:            amt,
 		FeeLimit:       feeLimit,
 		NumRoutes:      int32(ctx.Int("num_max_routes")),
+		MaxCltvDelay:   int32(ctx.Int("max_cltv_delay")),
```
halseth: Should be set also for `sendPayment`?

cmd/lncli/commands.go
```diff
@@ -2644,11 +2652,14 @@ func getNodeInfo(ctx *cli.Context) error {
...
-	Category:    "Payments",
-	Usage:       "Query a route to a destination.",
-	Description: "Queries the channel router for a potential path to the destination that has sufficient flow for the amount including fees",
-	ArgsUsage:   "dest amt",
+	Name:     "queryroutes",
+	Category: "Payments",
+	Usage:    "Query a route to a destination.",
+	Description: `
+			Queries the channel router for a potential path to the destination that has sufficient flow for 
+			the amount including fees. By default the fee limit capped at the send amount.
```
halseth: I think the fee comment belongs with the actual fee arguments?

routing/missioncontrol.go
```diff
@@ -375,8 +375,7 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
 	path, err := findPath(
 		nil, p.mc.graph, p.additionalEdges, p.mc.selfNode,
 		payment.Target, pruneView.vertexes, pruneView.edges,
-		payment.Amount, payment.FeeLimit, p.bandwidthHints,
-	)
+		payment.Amount, payment.Restrictions, p.bandwidthHints)
```
halseth: style nit: should keep `)` on the next line.

routing/pathfind_test.go
```diff
@@ -620,56 +628,86 @@ type expectedHop struct {
...
+		pathRestrictions:      PathRestrictions{FeeLimit: noFeeLimit, MaxCltvDelay: noCltvLimit},
+		expectedTotalTimeLock: 245, expectedTotalAmt: 100110,
 		expectedHops: []expectedHop{
 			{alias: "songoku", fwdAmount: 100000, fee: 110, timeLock: 101},
 			{alias: "sophon", fwdAmount: 100000, fee: 0, timeLock: 101},
 		}},
 
 	// Basic direct (one hop) route.
-	{target: "luoji", paymentAmt: 100, feeLimit: noFeeLimit,
+	{name: "BasicRouteDirect", target: "luoji", paymentAmt: 100,
```
halseth: It's a bit hard to see the changing parameters now. I think it would be easier to parse of we did it like 
```go
{
    name: "BasicRouteDirect", 
    target: "luoji", 
    paymentAmt: 100,
...
}.
```

routing/pathfind_test.go
```diff
@@ -620,56 +628,86 @@ type expectedHop struct {
 }
 
 type basicGraphPathFindingTestCase struct {
+	name                  string
```
halseth: 👍

routing/pathfind_test.go
```diff
@@ -718,20 +756,25 @@ func testBasicGraphPathFindingCase(t *testing.T, graphInstance *testGraphInstanc
...
-		if err == nil {
-			t.Fatal("expected no path to be found")
-		}
-		return
-	}
+
 	if err != nil {
-		t.Fatalf("unable to find path: %v", err)
+
+		if !test.expectError {
```
halseth: Can just use `test.expectedFailure = nil` for cases where we don't expect an error? Then we won't need the `expectError` field.

routing/pathfind_test.go
```diff
@@ -820,7 +863,7 @@ func testBasicGraphPathFindingCase(t *testing.T, graphInstance *testGraphInstanc
 	}
 
 	if route.TotalTimeLock != test.expectedTotalTimeLock {
-		t.Fatalf("expected time lock of %v, instead have %v", 2,
+		t.Fatalf("expected total-time lock of %v, instead have %v", test.expectedTotalTimeLock,
```
halseth: long line. Here and elsewhere.

routing/pathfind_test.go
```diff
@@ -1048,7 +1094,7 @@ func TestNewRoute(t *testing.T) {
 			expectedTimeLocks:     []uint32{1},
 			expectedTotalAmount:   100000,
 			expectedTotalTimeLock: 1,
-			feeLimit:              noFeeLimit,
+			pathRestrictions:      PathRestrictions{FeeLimit: noFeeLimit, MaxCltvDelay: noCltvLimit},
```
halseth: just define this restriction as a global test constant

routing/pathfind_test.go
```diff
@@ -1035,7 +1081,7 @@ func TestNewRoute(t *testing.T) {
 		// expectError is true.
 		expectedErrorCode errorCode
 
-		feeLimit lnwire.MilliSatoshi
+		pathRestrictions PathRestrictions
```
halseth: needs a comment

routing/pathfind_test.go
```diff
@@ -620,56 +628,86 @@ type expectedHop struct {
...
+		expectError:      true, expectedFailure: ErrNoPathFound,
+	},
+
+	// Basic route with max cltv delay limit.
+	{name: "MaxCltvLimit", target: "elst", paymentAmt: 100,
+		pathRestrictions: PathRestrictions{FeeLimit: noFeeLimit, MaxCltvDelay: 9},
+		expectError:      true, expectedFailure: ErrNoPathFound,
+	},
+
+	// Test routing a value for which there is no capacity
```
halseth: It is a bit hard to see what are the actual test additions here. Ideally the commits would be structured like
1. Modify the test data, any necessary refactoring and existing tests to work with this data.
2. Add the new tests.

routing/pathfind_test.go
```diff
@@ -620,56 +628,86 @@ type expectedHop struct {
...
-		expectFailureNoPath: true,
+	{name: "MaxFeeLimit", target: "sophon", paymentAmt: 100,
+		pathRestrictions: PathRestrictions{FeeLimit: 50, MaxCltvDelay: noCltvLimit},
+		expectError:      true, expectedFailure: ErrNoPathFound,
+	},
+
+	// Basic route with max cltv delay limit.
+	{name: "MaxCltvLimit", target: "elst", paymentAmt: 100,
+		pathRestrictions: PathRestrictions{FeeLimit: noFeeLimit, MaxCltvDelay: 9},
+		expectError:      true, expectedFailure: ErrNoPathFound,
```
halseth: Are we testing the case where there exists two paths to the target, where the cheapest one is above the cltv limit?

cmd/lncli/commands.go
```diff
@@ -2685,17 +2685,17 @@ var queryRoutesCommand = cli.Command{
...
-			Value: 1440,
-		},
 		cli.Int64Flag{
 			Name: "final_cltv_delta",
 			Usage: "number of blocks the last hop has to reveal " +
 				"the preimage",
 			Value: routing.DefaultFinalCLTVDelta,
 		},
+		cli.Int64Flag{
+			Name:  "max_cltv_delay",
```
halseth: this commit should be squashed with the one where this was originally introduced, or be a `fixup` commit.

routing/pathfind_test.go
johng: Would require creating a nill errortype. Should it be possible to throw a nill error?

routing/pathfind.go
johng: Any reason? They're initialised there.

routing/heap.go
johng: You're right, here source node is the actual destination since we do the path finding in reverse

routing/pathfind_test.go
halseth: All methods return a `nil` error in the success case?

johng:
But what do we put in `test.expectedFailure` if we remove `test.expectError`?
Currently we don't have a "nil" error.

routing/pathfind_test.go
johng: But what do we put in `test.expectedFailure` if we remove `test.expectError`?
Currently we don't have a "nil" error.

halseth: Can't you set `test.expectedFailure = nil` ?

johng: No, the expectedFailure is an int

halseth: Can you make it `error` instead?

johng: If we did we’d still need to pass in the error code somewhere and get back to where we started

joostjager:
No recent activity on this PR. I reworked the changes on top of the current master branch, but unfortunately not much of the code remained.

https://github.com/lightningnetwork/lnd/pull/2640

wpaulino:
Thanks for getting this started @johng! Superseded by #2640.

PR #1944, opened 2018-09-20
lnrpc: allow pubkey in SendToRoute

### LINKED ISSUE #2139: Allow private channels to be used in sendtoroute
### Background

Currently sendtoroute takes as an argument a set of hops that assume the channels specified reference channels known to the channel graph.

Completely specifying all the route details including public keys of node hops is not possible, which precludes the ability to use sendtoroute for things like sending to private recipients.

### Expected behaviour

Ideally sendtoroute would accept public keys in its arguments and follow crafted routes as specified without needing to consult the channel graph for edge and node lookups.

joostjager:
The information in the `Route` RPC message is not enough to directly send out the payment. First the local channel graph needs to be queried to complete the data. A lookup is executed to retrieve the public key of a node based on channel id. This does not work with private channels.

In this PR, we allow the hop pubkeys to be passed in so that execution of the payment becomes independent of the channel graph.

It is also a step towards the unbundling of lnd. (maybe, in the future, we get a node that doesn't track the channel graph at all anymore and is just a payment executor. mobile maybe?)

Fixes #2139.


joostjager:
`godoc` added

rpcserver.go
```diff
@@ -3509,57 +3509,109 @@ func (r *rpcServer) marshallRoute(route *routing.Route) *lnrpc.Route {
...
-		if err != nil {
-			return nil, fmt.Errorf("unable to fetch channel edges by "+
-				"channel ID for hop (%d): %v", i, err)
+		var routeHop *routing.Hop
+
+		if hop.PubKey == "" {
+			// If no pub key is given of the hop, the local channel
+			// graph needs to be queried to complete the information
+			// necessary for routing.
+			routeHop, err = unmarshallHopByChannelLookup(graph,
```
halseth: Regarding commit structure: it looks like this change (extracting logic into `unmarshallHopByChannelLookup`) could be done in an individual commit.

rpcserver.go
```diff
@@ -3509,57 +3509,109 @@ func (r *rpcServer) marshallRoute(route *routing.Route) *lnrpc.Route {
...
 	hops := make([]*routing.Hop, len(rpcroute.Hops))
 	for i, hop := range rpcroute.Hops {
-		// Discard edge policies, because they may be nil.
-		edgeInfo, _, _, err := graph.FetchChannelEdgesByID(hop.ChanId)
-		if err != nil {
-			return nil, fmt.Errorf("unable to fetch channel edges by "+
-				"channel ID for hop (%d): %v", i, err)
+		var routeHop *routing.Hop
+
+		if hop.PubKey == "" {
```
halseth: I think I would prefer this check be done inside the `unmarshal method`. Shouldn't be a problem passing in the graph even though it might not be needed.

rpcserver.go
```diff
@@ -3509,57 +3509,109 @@ func (r *rpcServer) marshallRoute(route *routing.Route) *lnrpc.Route {
 			Fee:              int64(fee.ToSatoshis()),
 			FeeMsat:          int64(fee),
 			Expiry:           uint32(hop.OutgoingTimeLock),
+
+			// Include public key of hop in response, so that
```
halseth: I think this comment belongs by the definition of the `PubKey` field.

rpcserver.go
```diff
@@ -3509,6 +3509,8 @@ func (r *rpcServer) marshallRoute(route *routing.Route) *lnrpc.Route {
 			Fee:              int64(fee.ToSatoshis()),
 			FeeMsat:          int64(fee),
 			Expiry:           uint32(hop.OutgoingTimeLock),
+			PubKey: hex.EncodeToString(
```
halseth: hmm, strange wrapping?

joostjager: I thought so too, but this is what gofmt does.

rpcserver.go
```diff
@@ -3548,6 +3550,41 @@ func unmarshallHopByChannelLookup(graph *channeldb.ChannelGraph, hop *lnrpc.Hop,
...
+		AmtToForward:     lnwire.MilliSatoshi(hop.AmtToForwardMsat),
+		PubKeyBytes:      pubKeyBytes,
+		ChannelID:        hop.ChanId,
+	}, nil
+}
+
+// unmarshallHop unmarshalls an rpc hop that may or may not contain a node
+// pubkey.
+func unmarshallHop(graph *channeldb.ChannelGraph, hop *lnrpc.Hop,
+	prevNodePubKey []byte) (*routing.Hop, error) {
```
halseth: `prevNodePubKey` should be made `[33]byte`

rpcserver.go
joostjager: No

rpcserver.go
joostjager: Why? This way there is less byte copying.

halseth: It makes the length clear, and makes sure it is non-nil. I don't think any extra byte copying makes a difference in this case.

joostjager: Changed. Not sure it was worth it. It is a minor thing which to me is also a matter of preference.

PR #1954, opened 2018-09-21
server: ensure each time we update a node ann we also update the on-disk version 

Roasbeef:
In this commit, we remedy the `genNodeAnnouncement` method to ensure that it also updates the on-disk node announcement. Before this commit, in both the Tor code, and the dynamic IP update code (for NAT traversal), we would fail to update the on-disk version of the node announcement. As a result, we would never propagate a new authenticated announcement to the network. 

Along the way, we clean up some duplicated node with the help of a new `LightningNode.NodeAnnouncement` method.

### DIFF
--- a/chan_series.go
+++ b/chan_series.go
@@ -89,7 +89,7 @@ func (c *chanSeries) UpdatesInHorizon(chain chainhash.Hash,
 		return nil, err
 	}
 	for _, nodeAnn := range nodeAnnsInHorizon {
-		nodeUpdate, err := makeNodeAnn(&nodeAnn)
+		nodeUpdate, err := nodeAnn.NodeAnnouncement(true)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


channeldb/graph.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"github.com/coreos/bbolt"
+	"github.com/go-errors/errors"
```
wpaulino: Heh I guess this snuck in back when I wrote this 😛

chan_series.go
```diff
@@ -89,7 +89,7 @@ func (c *chanSeries) UpdatesInHorizon(chain chainhash.Hash,
 		return nil, err
 	}
 	for _, nodeAnn := range nodeAnnsInHorizon {
-		nodeUpdate, err := makeNodeAnn(&nodeAnn)
+		nodeUpdate, err := nodeAnn.NodeAnnouncement(true)
```
wpaulino: 👍

server.go
```diff
@@ -1518,32 +1518,59 @@ func (s *server) genNodeAnnouncement(refresh bool,
...
+	// We'll now update the timestamp, ensuring that with each update, the
+	// timestamp monotonically increases.
 	newStamp := uint32(time.Now().Unix())
 	if newStamp <= s.currentNodeAnn.Timestamp {
 		newStamp = s.currentNodeAnn.Timestamp + 1
 	}
-
 	s.currentNodeAnn.Timestamp = newStamp
+
+	// Now that the announce tn is fully updated, we'll generate a new
```
wpaulino: announcement*

discovery/gossiper.go
```diff
@@ -270,27 +270,6 @@ func (d *AuthenticatedGossiper) SynchronizeNode(syncPeer lnpeer.Peer) error {
 	// containing all the messages to be sent to the target peer.
 	var announceMessages []lnwire.Message
 
-	makeNodeAnn := func(n *channeldb.LightningNode) (
```
halseth: Nice cleanup! 🌮

server.go
```diff
@@ -1518,32 +1518,59 @@ func (s *server) genNodeAnnouncement(refresh bool,
...
 	if err != nil {
 		return lnwire.NodeAnnouncement{}, err
 	}
-
 	s.currentNodeAnn.Signature, err = lnwire.NewSigFromSignature(sig)
 	if err != nil {
 		return lnwire.NodeAnnouncement{}, err
 	}
 
+	// Finally, we'll update the on-disk version of our announcement so it
```
halseth: This commit should be reverted/squashed with the one that moves it to the tor controller.

Roasbeef:
@halseth it depends on the situation. Other than the tor innit stuff (where we just need to set the source node as upon connect we'll send it to out first node), we only need to send it to the gossiper.

Roasbeef:
The OG version of this had a bug where since the funding manager would trigger a new node ann to be generated, and we wrote that to disk, then the gossiper would reject the ann as it has the same timestamp. In this current version, we'll send directly to the gossiper instead.

PR #1957, opened 2018-09-21
Update install.md: replace obsolete "gocode" path with "go"

bithobbes:


cfromknecht:
ACK, though can these be squashed into one commit? Also the commit message should be formatted according to the [code contribution guidelines](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#44-model-git-commit-messages).

This changes is fairly small, so an essay is certinaly not required. You can get an idea of the appropriate level of detail by examining past commits updating the install docs:
https://github.com/lightningnetwork/lnd/commits/master/docs/INSTALL.md

bithobbes:
Thanks, will try again.

PR #1958, opened 2018-09-21
Update install.md: dep installation via apt

bithobbes:
Installation via go did not work as described, with apt it's easier.


Roasbeef:
The command as is, is actually OS agnostic. What issues did you have with the prior instructions?

bithobbes:
Figured it out I think. Because of earlier GOPATH issues I had installed dep via apt and also ran it. Later even with the apt version removed the go version was not found as bash had cached ('hashed') dep's location. ´hash -r´ fixed things.

PR #1961, opened 2018-09-21
Add SubscribePayments RPC endpoint

roeierez:
This PR adds a missing piece for an LND client to be able to effectively sync with incoming/outgoing payments history.
For the incoming payments, represented by settled invoices, there is an API endpoint "SubscribeInvoices" that can take an "addIndex" and "settledIndex" which both serve as "checkpoints" and can be used by the client to get only changes occurred since it was last connected.
The current API doesn't have the same mechanism for outgoing payments and today clients need to query all payments (ListPayments) and don't have a way to be notified when a new payment is done.
This PR solves that by adding the SubscribePayments RPC API for outgoing payments.


Roasbeef:
> don't have a way to be notified when a new payment is done.
The streaming RPC for _sending_ payments already notifies this. As each action is user initiated (unlike receiving a payment), why does this RPC need to exist? Upon restart, simply fetch the set of existing payments, display that, and set up event handlers on payment success.

PR #1962, opened 2018-09-21
Add SubscribePayments RPC endpoint

roeierez:
** Sorry for duplicate this PR, I mistakenly closed the previous one and couldn't re-open it.

This PR adds a missing piece for an LND client to be able to effectively sync with incoming/outgoing payments history.
For the incoming payments, represented by settled invoices, there is an API endpoint "SubscribeInvoices" that can take an "addIndex" and "settledIndex" which both serve as "checkpoints" and can be used by the client to get only changes occurred since it was last connected.
The current API doesn't have the same mechanism for outgoing payments and today clients need to query all payments (ListPayments) and don't have a way to be notified when a new payment is done.
This PR solves that by adding the SubscribePayments RPC API for outgoing payments.


Roasbeef:
> don't have a way to be notified when a new payment is done.

The streaming RPC for _sending_ payments already notifies this. As each action is user initiated (unlike receiving a payment), why does this RPC need to exist? Upon restart, simply fetch the set of existing payments, display that, and set up event handlers on payment success.

roeierez:
@Roasbeef following your comment on https://github.com/lightningnetwork/lnd/pull/1961
Despite the fact this is a user initiated action sometimes the process that is responsible for the "paying" action is different than the process that is responsible for synchronization with the payments.
Also on recovery/startup it in a light client (like mobile) it is problematic to query the whole payments database in order to sync.

Roasbeef:
>  the process that is responsible for the "paying" action is different than the process that is responsible for synchronization with the payments.

So then pass this information between the layers? 

> Also on recovery/startup it in a light client (like mobile) it is problematic to query the whole payments database in order to sync.

Have you done any measurements/benchmarks of the load and latency for this query with various (1k, 10k, etc) amount of payments in the database? Assuming the client the client keeps no state, then wouldn't it need to do this query on startup in any case in order to populate the initial UI or state?

roeierez:
> So then pass this information between the layers?

Agree that can be done, the same way it can be done for tracking added invoices. But the fact that we can subscribe to added invoices today, despite the fact it is user initiated, is very convenient.

> Have you done any measurements/benchmarks of the load and latency for this query with various (1k, 10k, etc) amount of payments in the database? Assuming the client the client keeps no state, then wouldn't it need to do this query on startup in any case in order to populate the initial UI or state?

for the initial UI in light client like mobile, we would only need to load the first page (10,20 items)
These clients also needs to expose more advanced query options on top of payments (last N, filtering, etc...) and for this job they will need their own payments store, this is why a sync mechanism is needed. For this purpose the subscription pattern I believe best fit.

roeierez:
> We provide this interface for subscribing invoices as this is an action that can be initiated by _any node on the network_. _Sending_ a payment on the other hand, can only be initiated by the end user.

This is true for settled invoices not for added invoices that is initiated by the node only (we can still get notified on those). I hope I understood well and not missing anything.

> Sure, if they're stateful, then they may not even need the full query, and can instead use pagination to only extract a portion

I believe any reasonable app that shows the payments must be stateful, means it will have its own store of payments. Paging is only one problem, I don't think LND should take the responsibility for providing API of different queries, filtering, searching, etc.. that decent UI needs these days. Being able to sync effectively bypass all these problems, remove the need to expose such APIs and maintain freedom for clients to implement their own queries.

> If we added an _entirely_ new RPC each time a user of `lnd`, hits a problem that can be solved with some straight forward programming, we'd have hundreds of them

I understand and agree with this argument. In my opinion ListXXX and SubscribeXXX for the dominant persistent objects like transactions, payments, invoices, etc... is a good convention that solves a lot of hidden use cases.

roeierez:
@Roasbeef thanks for the feedback.
Is paging API for payments is aligned with your future plans and something that you will consider to merge in a higher priority? If so, I will go for it.

Roasbeef:
Pagination sounds good to me 👍

wpaulino:
Closing since we'll want to proceed with pagination rather than a new RPC.

PR #1970, opened 2018-09-24
rpc: populate additional information in AbandonChannel close summary

Roasbeef:
In this commit, we extend the AbandonChannel RPC method to store as much
information as possible within the CloseSummary. Most importantly, we
store the current and next (if available) revocation of the remote party
along with the channel config for the local party. With this
information, we have enough data to be able to generate a custom signer
for a new commitment state, or to prove that the channel was indeed
ours if needed.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1190,12 +1190,16 @@ out:
 func (r *rpcServer) AbandonChannel(ctx context.Context,
 	in *lnrpc.AbandonChannelRequest) (*lnrpc.AbandonChannelResponse, error) {
 
+	// If this isn't the debug build, then we won't allow the RPC to be
+	// executed, as it's an advanced feature and won't be activated in
+	// regular production/release builds.
 	if !DebugBuild {
 		return nil, fmt.Errorf("AbandonChannel RPC call only " +
 			"available in debug builds")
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -1224,23 +1228,46 @@ func (r *rpcServer) AbandonChannel(ctx context.Context,
...
+		SettledBalance:          dbChan.LocalCommitment.LocalBalance.ToSatoshis(),
+		ShortChanID:             dbChan.ShortChanID(),
+		RemoteCurrentRevocation: dbChan.RemoteCurrentRevocation,
+		LocalChanConfig:         dbChan.LocalChanCfg,
+	}
+
+	// Only if we ever received a FundingLocked message for this channel
+	// will be RemoteCurrentRevocation field be populated.
+	if dbChan.RemoteNextRevocation != nil {
+		summary.RemoteNextRevocation = dbChan.RemoteNextRevocation
```
cfromknecht: This seems equivalent to just setting
```golang
     RemoveNextRevocation: dbChan.RemoteNextRevocation
```

Roasbeef: Fixed!

PR #1972, opened 2018-09-24
Fix problems finding bitcoind auth cookie on Windows

dvide:
Fix for some problems I encountered when trying to get lnd to find the bitcoind auth cookie file on a Windows machine:

- An improper use of `path.Dir` rather than `filepath.Dir`, which meant that it didn't work with backslashes and so it was just returning `'.'`. The `filepath` package seems to be properly used everywhere else, just not in this one small instance.
- When reading bitcoind's `datadir` from `bitcoin.conf`, the regex didn't allow for spaces in the path name.
- When `bitcoind.dir` isn't present in `lnd.conf`, it uses a default dir of `AppData\Local\Bitcoin` on Windows, but this should be `AppData\Roaming\Bitcoin`. The fix also corrects this for `litecoind`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1282,7 +1282,7 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 	// Next, we'll try to find an auth cookie. We need to detect the chain
 	// by seeing if one is specified in the configuration file.
 	dataDir := filepath.Dir(bitcoindConfigPath)
-	dataDirRE, err := regexp.Compile(`(?m)^\s*datadir\s*=\s*([^\s]+)`)
+	dataDirRE, err := regexp.Compile(`(?m)^\s*datadir\s*=\s*([\S ]+?)\s*$`)
 	if err != nil {
 		return "", "", "", "", err
 	}
--- a/config.go
+++ b/config.go
@@ -83,8 +83,8 @@ var (
 	defaultLtcdDir         = btcutil.AppDataDir("ltcd", false)
 	defaultLtcdRPCCertFile = filepath.Join(defaultLtcdDir, "rpc.cert")
 
-	defaultBitcoindDir  = btcutil.AppDataDir("bitcoin", false)
-	defaultLitecoindDir = btcutil.AppDataDir("litecoin", false)
+	defaultBitcoindDir  = btcutil.AppDataDir("bitcoin", true)
+	defaultLitecoindDir = btcutil.AppDataDir("litecoin", true)
 
 	defaultTorSOCKS   = net.JoinHostPort("localhost", strconv.Itoa(defaultTorSOCKSPort))
 	defaultTorDNS     = net.JoinHostPort(defaultTorDNSHost, strconv.Itoa(defaultTorDNSPort))
--- a/config.go
+++ b/config.go
@@ -11,7 +11,6 @@ import (
 	"net"
 	"os"
 	"os/user"
-	"path"
 	"path/filepath"
 	"regexp"
 	"sort"
@@ -1282,7 +1281,7 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 
 	// Next, we'll try to find an auth cookie. We need to detect the chain
 	// by seeing if one is specified in the configuration file.
-	dataDir := path.Dir(bitcoindConfigPath)
+	dataDir := filepath.Dir(bitcoindConfigPath)
 	dataDirRE, err := regexp.Compile(`(?m)^\s*datadir\s*=\s*([^\s]+)`)
 	if err != nil {
 		return "", "", "", "", err
--- a/config.go
+++ b/config.go
@@ -1279,8 +1279,8 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 		return "", "", "", "", err
 	}
 
-	// Next, we'll try to find an auth cookie. We need to detect the chain
-	// by seeing if one is specified in the configuration file.
+	// Next, we'll try to find an auth cookie. We first follow the datadir
+	// if one is specified in the configuration file.
 	dataDir := filepath.Dir(bitcoindConfigPath)
 	dataDirRE, err := regexp.Compile(`(?m)^\s*datadir\s*=\s*([\S ]+?)\s*$`)
 	if err != nil {
@@ -1291,17 +1291,19 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 		dataDir = string(dataDirSubmatches[1])
 	}
 
-	chainDir := "/"
+	// Next we must append the appropriate chain dir
 	switch activeNetParams.Params.Name {
 	case "testnet3":
-		chainDir = "/testnet3/"
+		dataDir = filepath.Join(dataDir, "testnet3")
 	case "testnet4":
-		chainDir = "/testnet4/"
+		dataDir = filepath.Join(dataDir, "testnet4")
 	case "regtest":
-		chainDir = "/regtest/"
+		dataDir = filepath.Join(dataDir, "regtest")
 	}
 
-	cookie, err := ioutil.ReadFile(dataDir + chainDir + ".cookie")
+	// We can now try to read the authentication details from the
+	// .cookie file in this directory, if one exists.
+	cookie, err := ioutil.ReadFile(filepath.Join(dataDir, ".cookie"))
 	if err == nil {
 		splitCookie := strings.Split(string(cookie), ":")
 		if len(splitCookie) == 2 {


dvide:
Updated the regex so it doesn't include any trailing whitespace in the match.

molxyz:
@dvide Hi, 

The default directory for lnd bitcoin has always been `AppData\Local\Bitcoin` and I prefer very much it's this way. To me if you run your bitcoin-qt with real money/coins in the normal bitcoin default directory (`AppData\Roaming\Bitcoin`) on the same machine you run LND then it's better that your money stays in a separate directory, and run an empty wallet bitcoin-qt for LND in its own directory `AppData\Local\Bitcoin`.

Another reason to think about: Eclair Desktop also uses bitcoin-qt as their backend plus wallet, and their default directory is in `AppData\Roaming\Bitcoin` with a different bitcoin.conf setup. This can cause some confusion for users who would like to run both LND and Eclair if both are in the same directory. However, do you know you don't really have to run LND in any default directory?  It's very easy to configure your bitcoin.conf and lnd.conf to get LND connect to your Bitcoin in any directory. But if you want to make this easy for newbies, please leave the default directory alone for LND as it has always been like this since the beginning of LND. 

Also, could you elaborate the issue of LND not finding cookie auth and how how it has to do with the name of the datadir?  I use `rpcauth` to create credentials for `rpcuser` and `rpcpassword` in bitcoin.conf, because I've noticed, for windows, LND can't access bitcoin if it doesn't have rpcuser and rpcpassword in bitcoin.conf, not because of any space in the name of the file because I have the datadir file as "Bitcoin", not "Bitcoin Core". I'm not sure if there's an issue with LND not being able to read cookie auth but I'm wondering if it's a separate issue because of Windows.  But even if cookie auth doesn't work, we can use rpcauth and it's better that it can be used for multi users from what I heard. Another subject that might be worthwhile to look into as now bitcoin core has multiple wallet capability. 

How I get LND to work with Bitcoin on windows:  I have multiple testnet datadirs in C:\ directory, and I name them with whatever name you can think of. Then I create `rpcauth` and put them in their bitcoin.confs, plus the zmq args.  And I have the same credential in lnd.conf.  So it doesn't matter if my datadir isn't named "Bitcoin" or if it has space, LND still can read it.  I'm also a windows user, has been running/testing LND on windows for over a year, and this is my two satoshis.

dvide:
@molxyz Hi, thanks for responding.

I'm not really sure about what you're saying regarding the default bitcoind datadir. I don't understand why you would want lnd to look in `AppData\Local\Bitcoin` by default, since that's not the default datadir of bitcoind. For the vast majority of users, it's not going to find anything there? And even for those who have changed their datadir, they probably wouldn't haven choosen `AppData\Local\Bitcoin`? To me, it looks like it was a simple mistake that `Local` instead of `Roaming` was used originally, rather than it having been a deliberate choice. Of course, I don't want to step on anybody's toes here, if that's what changing this would do.

I get that you can change bitcoind's datadir to `AppData\Local\Bitcoin` if you want to, or to any other arbitrary directory. But if a user has done that, then it's fair to expect them to point `lnd` to that different directory, by specifying it in `lnd.conf` (`bitcoind.dir`). I think they will understand that they will need to do that, no? (That being said, `bitcoind.dir` doesn't seem to be documented anywhere, and it's not in the `sample_lnd.conf` either, but hopefully you get my point.)

I mean, the whole point of `extractBitcoindRPCParams` is to try to work with bitcoind "out of the box", right? So it seems strange to me to assume Windows users have a more complicated setup, and that they don't wish for their default bitcoind instance to be used. Why assume by default, but only for Windows users, that they've probably changed their bitcoind datadir, and specifically to `AppData\Local\Bitcoin`; or that they have multiple bitcoind instances running, and the one they want lnd to connect to happens to have its RPC details in `AppData\Local\Bitcoin`?

On a Linux machine, for example, it would simply look in bitcoind's default datadir (i.e. in `~/.bitcoin`), instead of assuming a more complicated setup exists where the user would rather that bitcoind instance not be touched, and instead look somewhere else. So why not have it be that simple on Windows too? It's not as though it's documented anywhere, in a guide or whatever, that this is the preferred setup for using bitcoind on Windows with lnd. Does that make sense, or am I missing something?

However, that being said of course I'll happily remove that commit from the PR if you guy's prefer it that way.

---

With regard to your second question, this is just if `bitcoin.conf` in the `bitcoind.dir` directory has a `datadir=foo` line in it. `extractBitcoindRPCParams` attempts to follow the datadir specified in `bitcoin.conf` to the 'real' datadir.

So for example, with bitcoind, you could have a folder structure like this:

```
In %APPDATA%\Bitcoin\
    bitcoin.conf:
        datadir=D:\Bitcoin\
        txindex=1
        dbcache=4000

In D:\Bitcoin\:
    blocks\
    chainstate\
    .cookie
    mempool.dat
    wallet.dat
    etc...
```

With this setup, you can run `bitcoind.exe` without `-datadir=D:\Bitcoin` on its command-line, and `bitcoin-cli.exe` the same way.

So `extractBitcoindRPCParams` is attempting to do this, in order to look for a `.cookie`, but I noticed that it wouldn't properly follow to a datadir that contained spaces. For example, if in the `bitcoin.conf` file you have `datadir=F:\foo bar`, it would instead look at `F:\foo\.cookie`, because the regex didn't properly match to a value with spaces. I changed it so that it would.

Hope that makes sense.

molxyz:
@dvide Hi,

First off, in your setup for bitcoin.conf, you donot have parameters for ZMQ, without zmq LND can't create connection to bitcoind.  Pllease see the install guide:  `https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md`

Secondly, you have a peculiar setup for your bitcoin node, so in order to find out if there's a problem, I tried to reproduce this setup on Testnet.  Now I have a bitcoin testnet node in `D:\Bitcoin\`.  Then I start LND and it works.  Then I shut off both nodes, change the name of "Bitcoin"  in `D:\Bitcoin` to `D:\Bitcoin Core`, restart Bitcoin, then restart LND.  Success. 

I am wondering it's because of incorrect setup for LND that you have some issues.  This is an example of my bitcoin.conf: 
`testnet=1
server=1
zmqpubrawblock=tcp://127.0.0.1:28332
zmqpubrawtx=tcp://127.0.0.1:28333
rpcuser=myusername
rpcpassword=mylongpassword`

And this my lnd.conf:
`[Application Options]
debuglevel=info
listen=:9735
maxpendingchannels=10
minchansize=50000

[Bitcoin]
bitcoin.active=1
bitcoin.testnet=1
bitcoin.node=bitcoind

[Bitcoind]
bitcoind.rpchost=127.0.0.1
bitcoind.zmqpubrawblock=tcp://127.0.0.1:28332
bitcoind.zmqpubrawtx=tcp://127.0.0.1:28333
bitcoind.rpcuser=myusername
bitcoind.rpcpass=mylongpassword`

Note that rpcuser & rpcpass/word have to be the same in both files.  Also, note I don't need to use bitcoind.dir. 

>And even for those who have changed their datadir, they probably wouldn't haven choosen AppData\Local\Bitcoin? To me, it looks like it was a simple mistake that Local instead of Roaming was used originally, rather than it having been a deliberate choice. 

Afaik, no windows users have ever complained about this directory, whether it's a mistake or deliberate, it's all moot because it's not a big deal to complain about. As long as it's something good for the devs to work on to benefit the community, I don't see a reason to change this. If someone doesn't know how to cut and paste a folder, then I'm sorry, maybe they need to take a class on basic computing.  But for those windows users that I know, and I even asked around, one of them said this default directory is not a problem.  Bitcoin is in the same directory with Lnd, Btcd, Btcctl, and that is where it should remain. Who knows if the long time "Roaming" default for Bitcoin might be changed in the future because there's an on going project from the Core devs to separate the wallet from the daemon, which remains to be seen.

>I mean, the whole point of extractBitcoindRPCParams is to try to work with bitcoind "out of the box", right? So it seems strange to me to assume Windows users have a more complicated setup, and that they don't wish for their default bitcoind instance to be used. Why assume by default, but only for Windows users, that they've probably changed their bitcoind datadir, and specifically to AppData\Local\Bitcoin; or that they have multiple bitcoind instances running, and the one they want lnd to connect to happens to have its RPC details in AppData\Local\Bitcoin?

Currently LND doesn't work out of the box for any operating system that I know of.  The things that will work out of the box will be apps such as the Lightning App with Neutrino that the LND devs are trying to work on.  I'm not sure what makes you think windows user have a more complicated setup when you ignored the basic setup, if Linux users do this it will not for them either.  The difficulty for windows users doesn't even have anything to do with LND, currently windows users can't even compile LND because Windows doesn't have `make` and makefile or all the good tools like Linux. So unless they know where and how to get the tools, their choice is limited, and again, not by LND, but by Windows itself. So please think of it in a broader mindset.  

To sum it up, you can keep your bitcoin node in your favorite windows default directory and run LND just fine if you set up your bitcoin.conf and lnd.conf like I suggested above.  And if you have further questions I would be happy to help you out if I can.  But I'm sorry, I don't see this change can benefit anyone else because it only has to do with your personal setup of your nodes.  So I would have to object this. 

NACK

dvide:
I legitimately don't understand where this is coming from. ☹

> First off, in your setup for bitcoin.conf, you donot have parameters for ZMQ, without zmq LND can't create connection to bitcoind. Pllease see the install guide: `https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md`

It's not "my setup"; it was just an example to further explain what you asked me to elaborate on (i.e. following the `datadir` from `bitcoin.conf`). ZMQ has nothing to do with that. Don't read too much into it.

> Secondly, you have a peculiar setup for your bitcoin node, so in order to find out if there's a problem, I tried to reproduce this setup on Testnet. Now I have a bitcoin testnet node in `D:\Bitcoin\`. Then I start LND and it works. Then I shut off both nodes, change the name of "Bitcoin" in `D:\Bitcoin` to `D:\Bitcoin Core`, restart Bitcoin, then restart LND. Success.

You seem to think that I need help setting up LND or something, because I cannot get it to work? You're reading waaay too much into everything. I'm not asking for all that.

I'm not saying it "won't work on Windows" at all. I'm just trying to submit a few bugfixes for `extractBitcoindRPCParams`. The setup you're talking about wouldn't even call `extractBitcoindRPCParams`, because you're providing the RPC authentication details directly. I already know that it works if you're doing that.

My original post says "when trying to get lnd to find the bitcoind auth cookie file on a Windows machine", not "when trying to get LND to work on a Windows machine".

> I am wondering it's because of incorrect setup for LND that you have some issues. This is an example of my bitcoin.conf:
> 
> testnet=1
> server=1
> zmqpubrawblock=tcp://127.0.0.1:28332
> zmqpubrawtx=tcp://127.0.0.1:28333
> rpcuser=myusername
> rpcpassword=mylongpassword
> 
> And this my lnd.conf:
> 
> [Application Options]
> debuglevel=info
> listen=:9735
> maxpendingchannels=10
> minchansize=50000
> [Bitcoin]
> bitcoin.active=1
> bitcoin.testnet=1
> bitcoin.node=bitcoind
> [Bitcoind]
> bitcoind.rpchost=127.0.0.1
> bitcoind.zmqpubrawblock=tcp://127.0.0.1:28332
> bitcoind.zmqpubrawtx=tcp://127.0.0.1:28333
> bitcoind.rpcuser=myusername
> bitcoind.rpcpass=mylongpassword
> 
> Note that rpcuser & rpcpass/word have to be the same in both files. Also, note I don't need to use bitcoind.dir.

Like I said, you're setting `bitcoind.rpcuser` and `bitcoind.rpcpass`, so it's not even going to attempt to find the auth cookie file because you've already supplied the authentication details. It doesn't need to use `bitcoind.dir` then either. It's completely irrelevant to my patch.

Look, I didn't write `extractBitcoindRPCParams` to look for the auth cookie. It's not up to me to decide whether or not lnd should support using the auth cookie file. (It's as though you have a problem with anyone that would want this.) But it does appear to support it. And yet that support doesn't work on Windows for the reasons I've explained. I was just trying to fix some small bugs with its attempt to locate the file on Windows.

Have you looked at the patch I submitted? Do you not agree that these are issues with what the code was clearly trying to do? It's quite a small patch and I've explained the changes in detail I think. Probably in more detail than was necessary.

> Afaik, no windows users have ever complained about this directory, whether it's a mistake or deliberate, it's all moot because it's not a big deal to complain about. If someone doesn't know how to cut and paste a folder, then I'm sorry, maybe they need to take a class on basic computing.

I'm not "complaining" about it. And I don't understand where this weird hostility is coming from?

>But for those windows users that I know, and I even asked around, one of them said this default directory is not a problem. 

I get that looking in the wrong directory by default is 'not a problem', because you can always change the default, or just provide the RPC params directly like you are doing (and like most people probably do). But that's not a very compelling reason to have the default value be some random arbitrary directory when there's an obvious default value, which is bitcoind's default datadir. Like I said, surely this choice originally was a mistake rather than some intentional decision to try to avoid connecting to the user's default bitcoind instance?

>Bitcoin is in the same directory with Lnd, Btcd, Btcctl, and that is where it should remain.

What do you mean 'is in'? We're not talking about where 'Bitcoin' 'is', we're talking about where lnd looks by default when it decides to call `extractBitcoindRPCParams`. And I still don't understand why one would want it to look by default in `AppData\Local`, when the default bitcoind datadir is in `AppData\Roaming`? That just makes no sense to me. And to only look in a completely different directory by default on Windows, but not on any other OS? Why?

Unless, like I say, I'm missing something? I dunno. You seem to feel very strongly about this. I'll definitely remove that commit if a more influential reviewer expresses the same sentiment.

> Who knows if the long time "Roaming" default for Bitcoin might be changed in the future because there's an on going project from the Core devs to separate the wallet from the daemon, which remains to be seen.

So just change it when and if that happens? Kind of a weird argument to justify it looking in the wrong directory by default right now.

> I'm not sure what makes you think windows users have a more complicated setup when you ignored the basic setup, if Linux users do this it will not for them either.

This is getting so overly confused for such a simple patch. I was saying that _you_ seem to be saying that lnd should assume a more complicated setup exists by default for Windows users, but only for Windows users. That was _your_ justification for leaving the `bitcoind.dir` default as `AppData\Local`, because you said that users should be running two separate instances of bitcoind anyway (one for their 'real money' and one with an empty wallet). So therefore, lnd shouldn't look in the default bitcoind datadir by default, because Windows users wouldn't want lnd to connect to that particular instance of bitcoind anyway. I was trying to deal with what I thought was your reasoning, though I admitted I didn't really understand what you were saying (because it seems to be not a very good argument, so I'm trying to give you the benefit of the doubt by granting to you that I may have misinterpreted it). Obviously I don't wish to break anything, if that's what this change would do in some way that I'm not seeing.

>The difficulty for windows users doesn't even have anything to do with LND, currently windows users can't even compile LND because Windows doesn't have `make` and makefile or all the good tools like Linux. So unless they know where and how to get the tools, their choice is limited, and again, not by LND, but by Windows itself. So please think of it in a broader mindset.

I still don't understand. Broader mindset? Compiling has nothing to do with anything here. I don't understand how that would broaden my mindset on this issue.

> But I'm sorry, I don't see this change can benefit anyone else because it only has to do with your personal setup of your nodes. So I would have to object this.

This has nothing to do with my personal setup.

molxyz:
@dvide Hi, 

I apologize if I sounded hostile. I didnot mean to come across like this. 
Your diagram didn't show the common setup for the .conf files (with ZMQ and such) so I thought you were having issues because of not having a proper setup.  
Also, I forgot to mention that LND could read the .cookie file in bitcoind before, but since I upgraded to the latest release I notice it can't read the .cookie file for some reason. However, I've been using bitcoin mostly not in the default directory so I often not to care about the .cookie file and from the experience I had with it before, .cookie file in bitcoin was flaky, not consistent for a long time. 
But you may be right that we need this change. Give me some hours, let me compile and test your PR and will get back with you.  
Thanks.

dvide:
@molxyz No worries. I apologize too. Thanks for testing the PR. 👍

>I forgot to mention that LND could read the .cookie file in bitcoind before, but since I upgraded to the latest release I notice it can't read the .cookie file for some reason.

On Windows though yeah? I think if your bitcoind's `bitcoin.conf` has a `datadir=` line in it, and the dir doesn't contain spaces, it would have always worked. But if it doesn't have a `datadir` line, from what I can gather, I don't think it ever would have worked properly on Windows due to the `path.Dir` issue. Using git blame, I see that the use of `path.Dir` was always there [from the beginning](https://github.com/lightningnetwork/lnd/commit/9a02884d0bd16bc36ad79b0e576a3a4604ebfc79#diff-b4bda758a2aef091432646c354b4dc59R727) of when this stuff was added.

And even if you specifically set your `bitcoind.dir` in `lnd.conf` using forward slashes, like `C:/foo/bar/bitcoin`, in an attempt to work around this, it gets turned back into `C:\foo\bar\bitcoin` with backslashes anyway, so there's no way to get it to work. By the time the path gets to the `path.Dir` call, it always just returns `'.'` due to the backslashes, and so it ends up trying to open `./.cookie`. 

The only time it doesn't use the result of the call to `path.Dir` is if your `bitcoin.conf` has a `datadir` line in it (if the regex finds a match). So that's one way I can see that it may have worked before for you, and then stopped working at a future date if you at some point removed that line.

molxyz:
@dvide Hi, 

Ok I compiled and ran your code, LND works great exactly with the location %APPDATA%\Bitcoin for Bitcoin node as the default, and also with the datadir forwarding if Bitcoin is in another location like in your diagram, it can find the .cookie file nicely and bypass the space in the name of the bitcoin folder like you said.  

However, when I try with Bitcoin in other locations than this default location, LND can't read rpcauth (rpcuser and rpcpassword).  I have the credentials in both .conf files as shown in my last post, plus I also put `bitcoind.dir=<bitcoin-location>` in lnd.conf, but LND still can't read the credentials, and keeps giving me this error:  `lnd.conf:24: unknown option: zmqpubrawblock`. 

Is there something I'm missing?

dvide:
> However, when I try with Bitcoin in other locations than this default location, LND can't read rpcauth (rpcuser and rpcpassword). I have the credentials in both .conf files as shown in my last post, plus I also put `bitcoind.dir=<bitcoin-location>` in lnd.conf, but LND still can't read the credentials, and keeps giving me this error: `lnd.conf:24: unknown option: zmqpubrawblock`.

Did you put `bitcoind.zmqpubrawblock` or just `zmqpubrawblock`? That would be the simplest explanation. This error would come from parsing the `lnd.conf` file, which my changes shouldn't have affected. I briefly tried to cause this error in the way you're describing, but I couldn't replicate it, and I can't imagine a way in which my patch would cause this.

If you're still getting it, could you post a full `lnd.conf` & `bitcoin.conf` which replicates the error on this PR, but not on master, so I can check that out. Thanks!

> EDIT: I forgot to mention, and because Bitcoin is not in the default location, LND won't be able to read the .cookie file so we need to use rpcauth. In my opinion, it would be best to be able to do this as we've always done it, and not stuck with just a default location.

What do you mean it won't be able to? It should be able to (if you set `bitcoind.dir` properly). There's nothing preventing it from reading a .cookie file in a different location.

And if you want to use rpcauth you can still do that like before. This stuff shouldn't even come into relevance if you set `bitcoind.rpcuser`, `bitcoind.rpcpass`, `bitcoind.zmqpubrawblock`, and `bitcoind.zmqpubrawtx` in the `lnd.conf`. Even if you set `bitcoind.dir` alongside them, it wouldn't do anything with that as it doesn't need to try to extract those details then, and that's the only thing it uses `bitcoind.dir` for. That's the same on master as in my PR.

So I'm not quite sure what you mean? You shouldn't be stuck with a default location.

molxyz:
@dvide  Yup that was the error I missed from deleting and pasting my lnd.conf file.  Then I got another error, had to get that figured out. But I got it to work now.  This is nice that it can work with bitcoin from any directory with or without .cookie file.  👍 

Thanks and again, my apology.  :-)

Roasbeef:
Still relevant?

wpaulino:
I believe so. @dvide could you rebase this on the current master branch?

PR #1975, opened 2018-09-25
[tests] Force tick on FwdEventTicker

halseth:
Ensures events are properly flushed on `FwdEventTick`. Should also help with the flaky coverage reports.

### DIFF
--- a/htlcswitch/switch_test.go
+++ b/htlcswitch/switch_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/ticker"
 )
 
 func genPreimage() ([32]byte, error) {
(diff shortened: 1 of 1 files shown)


Roasbeef:
Is all of this really necessary _just_ to appease a code coverage tool which itself is known to be overly sensitive?

htlcswitch/switch_test.go
```diff
@@ -1835,7 +1836,58 @@ func TestMultiHopPaymentForwardingEvents(t *testing.T) {
...
+	}
+
+	bobLog, ok := n.bobServer.htlcSwitch.cfg.FwdingLog.(*mockForwardingLog)
+	if !ok {
+		t.Fatalf("mockForwardingLog assertion failed")
+	}
+
+	// After sending 5 of the payments, trigger the forwarding ticker, to
+	// make sure the events are properly flushed.
+	bobTicker := n.bobServer.htlcSwitch.cfg.FwdEventTicker.(*ticker.Mock)
```
wpaulino: Nit: also make type check like we do above?

htlcswitch/switch_test.go
```diff
@@ -1835,7 +1836,58 @@ func TestMultiHopPaymentForwardingEvents(t *testing.T) {
...
+
+	// We'll trigger the ticker, and wait for the events to appear in Bob's
+	// forwarding log.
+	timeout := time.After(15 * time.Second)
+	for {
+
```
wpaulino: Nit: extra newline.

htlcswitch/switch_test.go
```diff
@@ -1835,7 +1836,58 @@ func TestMultiHopPaymentForwardingEvents(t *testing.T) {
...
+		bobLog.Unlock()
+
+		// Otherwise wait a little bit before checking again.
+		select {
+		case <-time.After(50 * time.Millisecond):
+		case <-timeout:
+			bobLog.Lock()
+			t.Fatalf("expected 5 events in event log, instead "+
+				"found: %v", spew.Sdump(bobLog.events))
+			bobLog.Unlock()
```
wpaulino: Perhaps this should be deferred? The `t.Fatal` call _shouldn't_ allow it to get to this point.

halseth: True, good catch :)

PR #1595, opened 2018-07-20
discovery/gossiper: reliably send channel update msg to remote peer

wpaulino:
In this commit, we address an issue where it's possible that we failed
to send the channel update message for a channel to the remote peer due
to them being offline. This would cause issues with private channels, as
a channel update message is needed for the remote peer to be aware of
the channel's forwarding policy. Due to the latest changes to
NotifyWhenOnline, we can now attempt to send the message every time we
detect the remote peer is online until success.

Depends on #1505.
Fixes #1347.

PR #1981, opened 2018-09-26
rpc+server: ensure we don't leak unadvertised nodes within invoice routing hints

wpaulino:
In this commit, we ensure that we don't include routing hints for
unadvertised nodes at the time of invoice creation. Otherwise, this
would lead us to leak these unadvertised nodes to anyone who can get
their hands on the invoice being created. To prevent this, we'll now
look at the network graph and ensure that the node in unadvertised if
all of their edges are unadvertised and only extend to us.

### DIFF
--- a/chan_series.go
+++ b/chan_series.go
@@ -89,6 +89,22 @@ func (c *chanSeries) UpdatesInHorizon(chain chainhash.Hash,
 		return nil, err
 	}
 	for _, nodeAnn := range nodeAnnsInHorizon {
+		// Ensure we only forward nodes that are publicly advertised to
+		// prevent leaking information about nodes.
+		isNodePublic, err := c.graph.IsPublicNode(nodeAnn.PubKeyBytes)
+		if err != nil {
+			srvrLog.Errorf("Unable to determine if node %x is "+
+				"advertised: %v", nodeAnn.PubKeyBytes, err)
+			continue
+		}
+
+		if !isNodePublic {
+			srvrLog.Tracef("Skipping forwarding announcement for "+
+				"node %x due to being unadvertised",
+				nodeAnn.PubKeyBytes)
+			continue
+		}
+
 		nodeUpdate, err := nodeAnn.NodeAnnouncement(true)
 		if err != nil {
 			return nil, err
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"image/color"
 	"io"
@@ -1804,6 +1805,47 @@ func (l *LightningNode) NodeAnnouncement(signed bool) (*lnwire.NodeAnnouncement,
 	return nodeAnn, nil
 }
 
+// isPublic determines whether the node is seen as public within the graph from
+// the source node's point of view. An existing database transaction can also be
+// specified.
+func (l *LightningNode) isPublic(tx *bolt.Tx, sourcePubKey []byte) (bool, error) {
+	// In order to determine whether this node is publicly advertised within
+	// the graph, we'll need to look at all of its edges and check whether
+	// they extend to any other node than the source node. errDone will be
+	// used to terminate the check early.
+	nodeIsPublic := false
+	errDone := errors.New("done")
+	err := l.ForEachChannel(tx, func(_ *bolt.Tx, info *ChannelEdgeInfo,
+		_, _ *ChannelEdgePolicy) error {
+
+		// If this edge doesn't extend to the source node, we'll
+		// terminate our search as we can now conclude that the node is
+		// publicly advertised within the graph due to the local node
+		// knowing of the current edge.
+		if !bytes.Equal(info.NodeKey1Bytes[:], sourcePubKey) &&
+			!bytes.Equal(info.NodeKey2Bytes[:], sourcePubKey) {
+
+			nodeIsPublic = true
+			return errDone
+		}
+
+		// Since the edge _does_ extend to the source node, we'll also
+		// need to ensure that this is a public edge.
+		if info.AuthProof != nil {
+			nodeIsPublic = true
+			return errDone
+		}
+
+		// Otherwise, we'll continue our search.
+		return nil
+	})
+	if err != nil && err != errDone {
+		return false, err
+	}
+
+	return nodeIsPublic, nil
+}
+
 // FetchLightningNode attempts to look up a target node by its identity public
 // key. If the node isn't found in the database, then ErrGraphNodeNotFound is
 // returned.
@@ -2564,6 +2606,35 @@ func (c *ChannelGraph) FetchChannelEdgesByID(chanID uint64) (*ChannelEdgeInfo, *
 	}
 
 	return edgeInfo, policy1, policy2, nil
+}
+
+// IsPublicNode is a helper method that determines whether the node with the
+// given public key is seen as a public node in the graph from the graph's
+// source node's point of view.
+func (c *ChannelGraph) IsPublicNode(pubKey [33]byte) (bool, error) {
+	var nodeIsPublic bool
+	err := c.db.View(func(tx *bolt.Tx) error {
+		nodes := tx.Bucket(nodeBucket)
+		if nodes == nil {
+			return ErrGraphNodesNotFound
+		}
+		ourPubKey := nodes.Get(sourceKey)
+		if ourPubKey == nil {
+			return ErrSourceNodeNotSet
+		}
+		node, err := fetchLightningNode(nodes, pubKey[:])
+		if err != nil {
+			return err
+		}
+
+		nodeIsPublic, err = node.isPublic(tx, ourPubKey)
+		return err
+	})
+	if err != nil {
+		return false, err
+	}
+
+	return nodeIsPublic, nil
 }
 
 // genMultiSigP2WSH generates the p2wsh'd multisig script for 2 of 2 pubkeys.
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -1613,13 +1613,26 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 			return nil
 		}
 
-		// Node announcement was successfully proceeded and know it
-		// might be broadcast to other connected nodes.
-		announcements = append(announcements, networkMsg{
-			peer:   nMsg.peer,
-			source: nMsg.source,
-			msg:    msg,
-		})
+		// In order to ensure we don't leak unadvertised nodes, we'll
+		// make a quick check to ensure this node intends to publicly
+		// advertise itself to the network.
+		isPublic, err := d.cfg.Router.IsPublicNode(node.PubKeyBytes)
+		if err != nil {
+			log.Errorf("Unable to determine if node %x is "+
+				"advertised: %v", node.PubKeyBytes, err)
+			nMsg.err <- err
+			return nil
+		}
+
+		// If it does, we'll add their announcement to our batch so that
+		// it can be broadcast to the rest of our peers.
+		if isPublic {
+			announcements = append(announcements, networkMsg{
+				peer:   nMsg.peer,
+				source: nMsg.source,
+				msg:    msg,
+			})
+		}
 
 		nMsg.err <- nil
 		// TODO(roasbeef): get rid of the above
--- a/discovery/gossiper_test.go
+++ b/discovery/gossiper_test.go
@@ -616,40 +616,10 @@ func TestProcessAnnouncement(t *testing.T) {
 		}
 	}
 
-	// Create node valid, signed announcement, process it with
-	// gossiper service, check that valid announcement have been
-	// propagated farther into the lightning network, and check that we
-	// added new node into router.
-	na, err := createNodeAnnouncement(nodeKeyPriv1, timestamp)
-	if err != nil {
-		t.Fatalf("can't create node announcement: %v", err)
-	}
-
 	nodePeer := &mockPeer{nodeKeyPriv1.PubKey(), nil, nil}
 
-	select {
-	case err = <-ctx.gossiper.ProcessRemoteAnnouncement(na, nodePeer):
-	case <-time.After(2 * time.Second):
-		t.Fatal("remote announcement not processed")
-	}
-	if err != nil {
-		t.Fatalf("can't process remote announcement: %v", err)
-	}
-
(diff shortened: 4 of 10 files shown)


server.go
```diff
@@ -3126,3 +3126,50 @@ func (s *server) watchChannelStatus() {
 		}
 	}
 }
+
+// isNodePublic attempts to determine if a node with the given public key is
+// publicly advertised within the network.
+func (s *server) isNodePublic(pubKey *btcec.PublicKey) (bool, error) {
+	// First, we'll attempt to retrieve the node from the graph.
+	remoteNode, err := s.chanDB.ChannelGraph().FetchLightningNode(pubKey)
```
Roasbeef: Isn't this enough? In that if the node isn't found in our graph, then we can assume that it likely has purely non-advertised channels. For example, here's me trying to lookup laptop node (running the app on testnet) from the faucet: 
```
kek@segspam:~/gocode/src/github.com/lightningnetwork/lnd# tlncli getnodeinfo 028916c993156097a1fdb5b4bbe4116f63bf9afbf4c7210d1c638dd2127abe7ed4
tlncli: command not found
```

Roasbeef: Ah actually this doesn't work if a direct connection exists between the two endpoints.

server.go
```diff
@@ -3126,3 +3126,50 @@ func (s *server) watchChannelStatus() {
...
+		return false, err
+	}
+
+	// In order to determine whether the node is publicly advertised within
+	// the graph, we'll need to look at all of its edges and check whether
+	// they extend to any other node than us. errDone will be used to
+	// terminate the check early.
+	nodeIsPublic := false
+	ourPub := s.identityPriv.PubKey().SerializeCompressed()
+	errDone := errors.New("done")
```
Roasbeef: Nice touch with `errDone` here.

server.go
```diff
@@ -3126,3 +3126,50 @@ func (s *server) watchChannelStatus() {
 		}
 	}
 }
+
+// isNodePublic attempts to determine if a node with the given public key is
+// publicly advertised within the network.
+func (s *server) isNodePublic(pubKey *btcec.PublicKey) (bool, error) {
```
Roasbeef: Seems like we can pull this into a method directly on the node itself in the database. This would allow us to unit test this section with a small mock graph, possibly writing some helper methods to make it easier to make table-driven tests for this use case.

Roasbeef:
Will start to test this on the faucet (using laptop app).

server.go
```diff
@@ -3126,3 +3126,50 @@ func (s *server) watchChannelStatus() {
...
+
+	// In order to determine whether the node is publicly advertised within
+	// the graph, we'll need to look at all of its edges and check whether
+	// they extend to any other node than us. errDone will be used to
+	// terminate the check early.
+	nodeIsPublic := false
+	ourPub := s.identityPriv.PubKey().SerializeCompressed()
+	errDone := errors.New("done")
+
+	err = remoteNode.ForEachChannel(nil, func(_ *bolt.Tx,
```
halseth: Seems like this can be simplified down to checking whether we have a node announcement for the node?

Roasbeef: That's insufficient as we'll still have a node announcement for nodes that we have _private_ channels with.

halseth: Interesting, looking at the gossiper it seems like we are not checking this before forwarding incoming node announcements, meaning we potentially are leaking their presence there as well.

halseth: I'm also wondering if it makes sense for a node to send a node announcement if it plans to stay off the radar.

halseth: Agreed. We should send node announcements also for private channels, but only to the channel counterparty (non-broadcast).

wpaulino: Fixed!

server.go

channeldb/graph.go
```diff
@@ -2529,6 +2571,35 @@ func (c *ChannelGraph) FetchChannelEdgesByID(chanID uint64) (*ChannelEdgeInfo, *
...
+		}
+		node, err := fetchLightningNode(nodes, pubKey[:])
+		if err != nil {
+			return err
+		}
+
+		nodeIsPublic, err = node.isPublic(tx, ourPubKey)
+		return err
+	})
+	if err != nil && err != ErrGraphNodeNotFound {
```
halseth: Why check `ErrGraphNodeNotFound` here?

discovery/gossiper.go
```diff
@@ -1634,8 +1634,18 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
-		// Node announcement was successfully proceeded and know it
-		// might be broadcast to other connected nodes.
+		// In order to ensure we don't leak unadvertised nodes, we'll
+		// make a quick check to ensure this node intends to publicly
+		// advertise itself to the network.
+		isPublic, err := d.cfg.Router.IsPublicNode(node.PubKeyBytes)
+
+		// If it doesn't, we'll skip broadcasting this announcement to
+		// the rest of our peers.
+		if !isPublic || err != nil {
```
halseth: just treat `err` first, as a regular error?

discovery/gossiper.go
```diff
@@ -1634,8 +1634,18 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
+			nMsg.err <- nil
+			return announcements
```
halseth: I was thinking of the case where we receive a node announcement, and later we learn about a public channel for the node. E.g. we open a private channel with the node, receive a node announcement, then they open a public channel which we learn about. I _thiiink_ it's okay now, as they probably should send out a `NodeAnnouncement` after they open the public channel.

discovery/gossiper_test.go
```diff
@@ -1966,6 +1990,68 @@ func TestDeDuplicatedAnnouncements(t *testing.T) {
...
+		t.Fatal("remote announcement not processed")
+	}
+
+	// The gossiper should also not broadcast the node announcement due to
+	// it not being part of any advertised channels.
+	select {
+	case <-ctx.broadcastedMessage:
+		t.Fatal("gossiper should not have broadcast node announcement")
+	case <-time.After(2 * trickleDelay):
+	}
```
halseth: extend with a check adding a public channel + node announcement?

fundingmanager.go
```diff
@@ -2149,16 +2149,36 @@ func (f *fundingManager) addToRouterGraph(completeChan *channeldb.OpenChannel,
...
+		case <-f.quit:
+			return ErrFundingManagerShuttingDown
+		}
+
+		nodeAnn, err := f.cfg.CurrentNodeAnnouncement()
+		if err != nil {
+			return fmt.Errorf("unable to retrieve current node "+
+				"announcement: %v", err)
+		}
+		if err := peer.SendMessage(true, &nodeAnn); err != nil {
```
halseth: Add a log before sending?

chan_series.go
```diff
@@ -89,6 +89,13 @@ func (c *chanSeries) UpdatesInHorizon(chain chainhash.Hash,
 		return nil, err
 	}
 	for _, nodeAnn := range nodeAnnsInHorizon {
+		// Ensure we only forward nodes that are publicly advertised to
+		// prevent leaking information about nodes.
+		isNodePublic, err := c.graph.IsPublicNode(nodeAnn.PubKeyBytes)
+		if !isNodePublic || err != nil {
```
halseth: same, error should be treated.

rpcserver.go
```diff
@@ -2713,16 +2713,26 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 
 			// Make sure the channel is active.
 			link, err := r.server.htlcSwitch.GetLink(chanPoint)
-			if err != nil {
-				rpcsLog.Errorf("Unable to get link for "+
-					"channel %v: %v", chanPoint, err)
+			if err != nil || !link.EligibleToForward() {
```
halseth: I think we should keep the existing way of logging the error.

rpcserver.go
```diff
@@ -2713,16 +2713,26 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-					"being eligible to forward payments",
-					chanPoint)
+			// To ensure we don't leak unadvertised nodes, we'll
+			// make sure our counterparty is publicly advertised
+			// within the network. Otherwise, we'll end up leaking
+			// information about nodes that intend to stay
+			// unadvertised, like in the case of a node only having
+			// private channels.
+			var remotePub [33]byte
+			copy(remotePub[:], channel.IdentityPub.SerializeCompressed())
```
halseth: why copy?

rpcserver.go
```diff
@@ -2713,16 +2713,26 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+			isRemoteNodePublic, err := graph.IsPublicNode(remotePub)
+			if err != nil || !isRemoteNodePublic {
```
halseth: same, treat error

channeldb/graph.go
wpaulino: Not needed indeed - fixed.

discovery/gossiper.go
wpaulino: Fixed.

discovery/gossiper_test.go

fundingmanager.go
wpaulino: This should already be covered by the `PEER` log, no?

chan_series.go
wpaulino: Fixed.

rpcserver.go

rpcserver.go
wpaulino: `SerializeCompressed` returns a byte slice, while `IsPublicNode` requires a `[33]byte`.

rpcserver.go

fundingmanager.go
```diff
@@ -2149,16 +2149,36 @@ func (f *fundingManager) addToRouterGraph(completeChan *channeldb.OpenChannel,
...
+	// If this channel is not meant to be announced to the greater network,
+	// we'll only send our NodeAnnouncement to our counterparty to ensure we
+	// don't leak any of our information.
 	announceChan := completeChan.ChannelFlags&lnwire.FFAnnounceChannel != 0
 	if !announceChan {
 		fndgLog.Debugf("Will not announce private channel %v.",
 			shortChanID.ToUint64())
+
+		peerChan := make(chan lnpeer.Peer, 1)
+		f.cfg.NotifyWhenOnline(completeChan.IdentityPub, peerChan)
```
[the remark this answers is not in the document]
wpaulino: For now yeah. It can be extended once https://github.com/lightningnetwork/lnd/pull/1595 lands.

discovery/gossiper.go
```diff
@@ -1634,8 +1634,24 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
+		if err != nil {
+			log.Errorf("Unable to determine if node %x is "+
+				"advertised: %v", node.PubKeyBytes, err)
+			nMsg.err <- err
+			return nil
+		}
+
+		// If it doesn't, we'll skip broadcasting this announcement to
+		// the rest of our peers.
+		if !isPublic {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

fundingmanager.go
```diff
@@ -2149,16 +2148,36 @@ func (f *fundingManager) addToRouterGraph(completeChan *channeldb.OpenChannel,
...
+
+		var peer lnpeer.Peer
+		select {
+		case peer = <-peerChan:
```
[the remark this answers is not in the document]
wpaulino: It seems like all the uses of `annAfterSixConfs` are already async, so I don't think we'd gain anything from not blocking here.

Roasbeef:
Needs rebase!

fundingmanager.go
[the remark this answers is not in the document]
halseth: Yeah, I think that's fine, as we can retry at next startup and the channel is most likely not operational anyways before this succeeds.

fundingmanager.go
halseth: yes, but if we are grepping on funding logs it would be nice to easily find it.

chan_series.go
```diff
@@ -89,6 +89,19 @@ func (c *chanSeries) UpdatesInHorizon(chain chainhash.Hash,
...
+		// Ensure we only forward nodes that are publicly advertised to
+		// prevent leaking information about nodes.
+		isNodePublic, err := c.graph.IsPublicNode(nodeAnn.PubKeyBytes)
+		if err != nil {
+			srvrLog.Errorf("Unable to determine if node %x is "+
+				"advertised: %v", nodeAnn.PubKeyBytes, err)
+			continue
+		}
+
+		if !isNodePublic {
```
halseth: trace log "skipping non-public node"?

fundingmanager.go
wpaulino: Fixed.

chan_series.go
wpaulino: Fixed.

fundingmanager.go
```diff
@@ -2149,16 +2148,39 @@ func (f *fundingManager) addToRouterGraph(completeChan *channeldb.OpenChannel,
...
+		}
+
+		nodeAnn, err := f.cfg.CurrentNodeAnnouncement()
+		if err != nil {
+			return fmt.Errorf("unable to retrieve current node "+
+				"announcement: %v", err)
+		}
+
+		pubKey := peer.PubKey()
+		fndgLog.Debugf("Sending our NodeAnnouncement to %x", pubKey)
```
halseth: log channelpoint as well

wpaulino: Done

fundingmanager.go
```diff
@@ -2149,16 +2148,44 @@ func (f *fundingManager) addToRouterGraph(completeChan *channeldb.OpenChannel,
...
+	// If this channel is not meant to be announced to the greater network,
+	// we'll only send our NodeAnnouncement to our counterparty to ensure we
+	// don't leak any of our information.
 	announceChan := completeChan.ChannelFlags&lnwire.FFAnnounceChannel != 0
 	if !announceChan {
 		fndgLog.Debugf("Will not announce private channel %v.",
 			shortChanID.ToUint64())
+
+		peerChan := make(chan lnpeer.Peer, 1)
+		f.cfg.NotifyWhenOnline(completeChan.IdentityPub, peerChan)
```
Roasbeef: We shouldn't send this _before_ we actually send over the chan update and chan announce. Otherwise, if this is their first private channel for example, they won't have any visible channels in the graph, so we won't accept the node announcement. The timing is a bit tricky.

PR #1983, opened 2018-09-27
docs/INSTALL: small improvements

bithobbes:


docs/INSTALL.md
```diff
@@ -34,15 +34,16 @@
     Alternatively, one can download the pre-compiled binaries hosted on the
     [golang download page](https://golang.org/dl/). If one seeks to install
     from source, then more detailed installation instructions can be found
-    [here](http://golang.org/doc/install).
-
+    [here](http://golang.org/doc/install).
```
cfromknecht: can we make this https?

docs/INSTALL.md
```diff
@@ -234,7 +235,8 @@ the following:
 - Since `lnd` uses
   [ZeroMQ](https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md) to
   interface with `bitcoind`, *your `bitcoind` installation must be compiled with
-  ZMQ*. Note that if you installed `bitcoind` from source and ZMQ was not present, 
+  ZMQ*. The official Bitcoin Core binaries do support ZMQ.
```
cfromknecht: nice!

docs/INSTALL.md
```diff
@@ -345,4 +347,6 @@ Bitcoin chain. `lnd` also supports Litecoin testnet4 (but not both BTC and LTC
 at the same time), so when working with Litecoin be sure to set to parameters
 for Litecoin accordingly. For node configuration, the sections are called
 `[Btcd]`, `[Bitcoind]`, `[Neutrino]`, `[Ltcd]`, and `[Litecoind]` depending on
-which chain and node type you're using.
+which chain and node type you're using.  
+Make sure to place options in the proper section.
```
cfromknecht: might be worth elaborating that "proper section" means that it must be under the section with the option's prefix, e.g. `bitcoind.xxxx` should go in `[Bitcoind]`

bithobbes: Hmm that is not always the case, other options don't have the prefix so it might get confusing?

bithobbes: What I think would be nice would be another paragraph "getting started" or "first steps" that explains how to activate autopilot and/or connect to a node to be able to transfer in both directions. This is something for another PR, though.

docs/INSTALL.md
```diff
@@ -34,15 +34,16 @@
...
     from source, then more detailed installation instructions can be found
-    [here](http://golang.org/doc/install).
-
+    [here](https://golang.org/doc/install).  
+  
+    Set $GOPATH:  
     At this point, you should set your `$GOPATH` environment variable, which
     represents the path to your workspace. By default, `$GOPATH` is set to
-    `~/go`. You will also need to add `$GOPATH/bin` to your `PATH`. This ensures
+    `$HOME/go`. You will also need to add `$GOPATH/bin` to your `PATH`. This ensures
```
halseth: Is `$HOME/` more accurate than using `~/`

bithobbes:
@cfromknecht Regarding your earlier concerns about multiple small commits being untidy (which I totally agree with): maybe this helps keeping the commit history clean while still allowing to separately review the commits: https://blog.github.com/2016-04-01-squash-your-commits/

PR #1984, opened 2018-09-27
Add listunspent RPC call

AdamISZ:
Returns a brief json summary of each utxo found by calling
ListUnspentWitness in the wallet. ~~The only argument is the
number of confirmations (0=include unconfirmed), acting as
a filter.~~
Arguments: minconfs maxconfs (so filter functions as in Core).

moli on IRC suggested this might be a desirable thing, which is certainly understandable, for debugging/checking. It currently shows (address, scriptpubkey, txid:n outpoint and amount in satoshis) in the json output. 

Adding the number of confirmations seems a little non-trivial as that's available in the call to the underlying `wallet.ListUnspent` `ListUnspentResponse` object but is not returned in the `WalletController` interface (which exposes `ListUnspentWitness` which returns objects of type `Utxo`, which don't contain the number of confirmations). That could be changed but it didn't seem enough of a big deal to address here.

I'll probably add a brief test function.


molxyz:
tACK!

This is awesome! I just compiled on a testnet node and here's the result:  http://paste.ubuntu.com/p/HGPs3bTKHG/ 

Would love to have this for LND.  Thanks, Adam.  :)

AdamISZ:
Squashed and rebased.

lnrpc/rpc.proto
```diff
@@ -647,6 +657,24 @@ service Lightning {
...
+
+	/// the value of the unspent coin in satoshis
+	int64 amount = 3 [json_name = "amount"];
+
+	/// the scriptpubkey in hex
+	string scriptpubkey = 4 [json_name = "scriptpubkey"];
+
+	/// The outpoint in format txid:n
+	string outpoint = 5 [json_name = "outpoint"];
+}
```
joostjager: Consider reusing the structured types `enum AddressType` and `message ChannelPoint`, to prevent string parsing client-side.

AdamISZ: Thanks, that's helpful. I'll take a look.

cmd/lncli/commands.go
```diff
@@ -242,6 +242,66 @@ func sendCoins(ctx *cli.Context) error {
 	return nil
 }
 
+var listUnspentCommand = cli.Command{
+	Name:      "listunspent",
+	Category:  "On-chain",
+	Usage:     "List all utxos available for spending",
+	ArgsUsage: "minconfs maxconfs",
```
halseth: These should be listed under `Flags` as well. Take  a look at other commands where this is done :)

cmd/lncli/commands.go
```diff
@@ -242,6 +242,66 @@ func sendCoins(ctx *cli.Context) error {
...
+
+func listUnspent(ctx *cli.Context) error {
+	var (
+		MinConfirms int64
+		MaxConfirms int64
+		err         error
+	)
+	args := ctx.Args()
+
+	if !(ctx.NArg() != 2 || ctx.NArg() != 1) || ctx.NumFlags() != 0 {
```
halseth: this can also be simplified with the change to `Flags`.

lnrpc/rpc.proto
```diff
@@ -647,6 +657,24 @@ service Lightning {
...
+message Utxo {
+
+	/// The type of address
+	string address_type = 1 [json_name = "address_type"];
+
+	/// The address
+	string address = 2 [json_name = "address"];
+
+	/// the value of the unspent coin in satoshis
+	int64 amount = 3 [json_name = "amount"];
```
halseth: should be named `amount_sat`

lnrpc/rpc.proto
```diff
@@ -647,6 +657,24 @@ service Lightning {
...
 
+message Utxo {
+
+	/// The type of address
+	string address_type = 1 [json_name = "address_type"];
+
+	/// The address
+	string address = 2 [json_name = "address"];
+
+	/// the value of the unspent coin in satoshis
```
halseth: nit: Start comments with capital letter.

rpcserver.go
```diff
@@ -456,6 +460,83 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
 	}
 }
 
+func (r *rpcServer) ListUnspent(ctx context.Context,
```
halseth: Missing godoc.

rpcserver.go
```diff
@@ -456,6 +460,83 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+	}
+
+	utxos := make([]*lnwallet.Utxo, 0, len(utxos1))
+
+	if usingMax {
+		utxos2, err := r.server.cc.wallet.ListUnspentWitness(maxConfs)
+		if err != nil {
+			return nil, err
+		}
+		// We want the symmetric difference between the first set (utxos1)
```
halseth: A better solution would be to modify the result struct to inlcude the number of confirmations: https://github.com/lightningnetwork/lnd/blob/b00e43eec1f19e55f910f7eb527bf7437b4069b4/lnwallet/interface.go#L52

This will also be useful other places, and makes sure we list unspents in one atomic db transaction! 😀

rpcserver.go
```diff
@@ -456,6 +460,83 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+		} else if utxo.AddressType == 1 {
+			utxoresp.AddressType = "np2wkh"
+		} else {
+			utxoresp.AddressType = "unknown"
+		}
+		utxoresplist = append(utxoresplist, &utxoresp)
+	}
+
+	// This logs the txid:n string only, server side.
+	rpcsLog.Infof("[listunspent] generated utxos: %v", utxos)
```
halseth: This might be a bit spammy? Could make `debug`.

rpcserver.go
```diff
@@ -456,6 +460,83 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+		for _, u := range utxos1 {
+			if _, ok := utxo2map[u.OutPoint.String()]; !ok {
+				utxos = append(utxos, u)
+			}
+		}
+	} else {
+		utxos = utxos1
+	}
+
+	utxoresplist := []*lnrpc.Utxo{}
```
halseth: nit: just create the `ListUnspentResponse` here, and append directly to it.

rpcserver.go
```diff
@@ -456,6 +460,83 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+			utxo.PkScript, activeNetParams.Params)
+		if err != nil {
+			return nil, err
+		}
+		if len(outAddresses) != 1 {
+			return nil, fmt.Errorf("An output was unexpectedly multisig")
+		}
+		utxoresp.Address = outAddresses[0].String()
+		if utxo.AddressType == 0 {
+			utxoresp.AddressType = "p2wkh"
```
halseth: as @joostjager mentioned, here we should be using the already defined enums.

rpcserver.go
AdamISZ: Thanks, so, just to check I understood: I mentioned in the first comment in the thread that I wanted to, but wasn't able, to include the number of confs due to that `Utxo` struct not containing it; so you're suggesting here that I do change that result struct. I was a bit wary of messing with more "internal" code for this PR, but I guess there's no possibility of that causing an issue elsewhere.

halseth: Oh sorry, didn't see that comment 😛 But yeah, I think it should be fairly trivial and non-disruptive, as you have the number of confirmations available as part of the returned result here: https://github.com/lightningnetwork/lnd/blob/b00e43eec1f19e55f910f7eb527bf7437b4069b4/lnwallet/btcwallet/btcwallet.go#L305

halseth: The only think remaining is to add the same information to the `Utxo` struct, correct 👍

lnrpc/rpc.proto
AdamISZ: Fixed as per 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4 although I have minor questions about this, see comment below.

rpcserver.go
AdamISZ: Fixed in 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4

rpcserver.go
AdamISZ: Fixed in 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4

rpcserver.go
AdamISZ: Switched it to debug, no prob removing it if preferred.

rpcserver.go
AdamISZ: Done 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4

rpcserver.go
AdamISZ: Done, but have Qs about this, see comment below

cmd/lncli/commands.go
AdamISZ: Done in 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4, it should now be in line with other commands

lnrpc/rpc.proto
AdamISZ: Done in 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4

lnrpc/rpc.proto
AdamISZ: Done in 60ae2afc236c6eb3d1095cca9dfed5d9760a23b4

joostjager:
>The message type ChannelPoint in the rpc is now reused for the Outpoint field in the lnrpc.Utxo message type. However this is a bit silly naming-wise since of course our spendable utxos are not channel points. As previously, just trying to avoid spreading code changes all over the place if it's not needed.

Yes, I agree that it should probably be renamed to OutPoint. Also, one could argue if the `oneof` is justifiable. It makes the interface wider while decoding hex bytes could easily by done by the caller too.

wpaulino:
Needs a proto rebase (this happens quite often)!

> (1) The message type ChannelPoint in the rpc is now reused for the Outpoint field in the lnrpc.Utxo message type. However this is a bit silly naming-wise since of course our spendable utxos are not channel points. As previously, just trying to avoid spreading code changes all over the place if it's not needed.

It doesn't seem like type aliases are supported, so I think it's probably better to just rename it or define a new OutPoint message to prevent the rename change.

> (2) For the AddressType I factored out the enum so it was reusable (i.e. not restricted to the NewAddressRequest message), which required a couple of very minor edits. I presume that's OK/desirable

Indeed 👍 

> (3) There is an issue with casting the number of confirmations from int64 to int32 here, but as per comment it seemed it was OK ... note the origin of the mismatch is that what is returned from ListUnspent in the btcd codebase here is an int64, so that's what I've stored in the Utxo struct that ListUnspentWitness returns.

Yeah, I'd say just keep things as they are now. Modifying this to be `int64` would probably be a  somewhat large diff.

rpcserver.go
```diff
@@ -456,6 +460,88 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+			"confirmations")
+	}
+	usingMax := maxConfs != math.MaxInt32
+	if usingMax {
+		rpcsLog.Infof("[listunspent] min=%v max=%v", minConfs, maxConfs)
+	} else {
+		rpcsLog.Infof("[listunspent] min=%v", minConfs)
+	}
+
+	utxos1, err := r.server.cc.wallet.ListUnspentWitness(minConfs)
```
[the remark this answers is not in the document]
AdamISZ: Right, extending the interface makes more sense, but: If I change the existing one (ListUnspentWitness) I'd have to modify all calls to it, wouldn't I? It seems a bit invasive but happy to do it if that's what's desired.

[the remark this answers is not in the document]
wpaulino: Indeed. Definitely warrants its own PR.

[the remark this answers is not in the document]
AdamISZ: OK so I can PR that and make this a dependency then, sound good? (I do agree it's more logical!)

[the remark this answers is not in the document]
wpaulino: Yup!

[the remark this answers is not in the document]
AdamISZ: See #2112 ; I'll wait on that before updating here, in case for some reason people decide against it.

rpcserver.go
```diff
@@ -456,6 +460,88 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+			Outpoint:      outpoint,
+			Confirmations: utxo.Confirmations,
+		}
+		_, outAddresses, _, err := txscript.ExtractPkScriptAddrs(
+			utxo.PkScript, activeNetParams.Params)
+		if err != nil {
+			return nil, err
+		}
+		if len(outAddresses) != 1 {
+			return nil, fmt.Errorf("An output was unexpectedly multisig")
```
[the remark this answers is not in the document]
AdamISZ: Done

rpcserver.go
```diff
@@ -456,6 +460,88 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+	var respList = lnrpc.ListUnspentResponse{Utxos: []*lnrpc.Utxo{}}
+	for _, utxo := range utxos2 {
+		// Translate lnwallet address type to gRPC proto address type:
+		var addrType lnrpc.AddressType
+		switch utxo.AddressType {
+		case lnwallet.WitnessPubKey:
+			addrType = lnrpc.AddressType_WITNESS_PUBKEY_HASH
+		case lnwallet.NestedWitnessPubKey:
+			addrType = lnrpc.AddressType_NESTED_PUBKEY_HASH
+		}
```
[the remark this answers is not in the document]
AdamISZ: Done (it is logged on server but not returned; don't see how it can happen for now anyway)

rpcserver.go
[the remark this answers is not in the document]
AdamISZ: Done (#2112 merged)

rpcserver.go
```diff
@@ -456,6 +460,88 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+// user, with 0 meaning unconfirmed.
+func (r *rpcServer) ListUnspent(ctx context.Context,
+	in *lnrpc.ListUnspentRequest) (*lnrpc.ListUnspentResponse, error) {
+	minConfs := in.Minconfs
+	maxConfs := in.Maxconfs
+	if minConfs > maxConfs {
+		return nil, fmt.Errorf("Max confirmations must be >= min " +
+			"confirmations")
+	}
+	usingMax := maxConfs != math.MaxInt32
```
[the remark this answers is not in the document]
AdamISZ: Done

rpcserver.go
```diff
@@ -456,6 +460,88 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+// the information returned is: outpoint, amount in satoshis, address,
+// address type, scriptPubKey in hex and number of confirmations.
+// The result is filtered to contain outputs whose number of confirmations
+// is between a minimum and maximum number of confirmations specified by the
+// user, with 0 meaning unconfirmed.
+func (r *rpcServer) ListUnspent(ctx context.Context,
+	in *lnrpc.ListUnspentRequest) (*lnrpc.ListUnspentResponse, error) {
+	minConfs := in.Minconfs
+	maxConfs := in.Maxconfs
+	if minConfs > maxConfs {
```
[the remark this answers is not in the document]
AdamISZ: Done

lnrpc/rpc.proto
```diff
@@ -231,6 +231,16 @@ service Lightning {
         };
     }
 
+    /** lncli: `listunspent`
+    ListUnspent returns a list of all utxos spendable by the wallet with a
+	number of confirmations between the specified minimum and maximum.
+    */
+    rpc ListUnspent (ListUnspentRequest) returns (ListUnspentResponse) {
+        option (google.api.http) = {
+            get: "/v1/txouts"
```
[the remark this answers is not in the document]
AdamISZ: Sure; done

cmd/lncli/commands.go
```diff
@@ -242,6 +242,101 @@ func sendCoins(ctx *cli.Context) error {
...
+			Usage: "the maximum number of confirmations for a utxo",
+		},
+	},
+	Action: actionDecorator(listUnspent),
+}
+
+func listUnspent(ctx *cli.Context) error {
+	var (
+		MinConfirms int64
+		MaxConfirms int64
```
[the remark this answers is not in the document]
AdamISZ: Done

cmd/lncli/commands.go
```diff
@@ -242,6 +242,101 @@ func sendCoins(ctx *cli.Context) error {
 	return nil
 }
 
+var listUnspentCommand = cli.Command{
+	Name:      "listunspent",
+	Category:  "On-chain",
+	Usage:     "List utxos available for spending",
```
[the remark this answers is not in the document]
AdamISZ: Done

lnrpc/rpc.proto
```diff
@@ -647,6 +657,29 @@ service Lightning {
...
+	AddressType type = 1 [json_name = "address_type"];
+
+	/// The address
+	string address = 2 [json_name = "address"];
+
+	/// The value of the unspent coin in satoshis
+	int64 amount_sat = 3 [json_name = "amount_sat"];
+
+	/// The scriptpubkey in hex
+	string scriptpubkey = 4 [json_name = "scriptpubkey"];
```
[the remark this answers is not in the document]
AdamISZ: Done

lnrpc/rpc.proto
```diff
@@ -797,18 +830,31 @@ message SendCoinsResponse {
     string txid = 1 [json_name = "txid"];
 }
 
+message ListUnspentRequest {
+	/// The minimum number of confirmations to be included.
+	int32 minconfs = 1;
```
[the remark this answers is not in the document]
AdamISZ: Done

lnrpc/rpc.proto
```diff
@@ -797,18 +830,31 @@ message SendCoinsResponse {
...
+
+	/// The maximum number of confirmations to be included.
+	int32 maxconfs = 2;
```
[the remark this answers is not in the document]
AdamISZ: Done

lnrpc/rpc.proto
```diff
@@ -797,18 +829,31 @@ message SendCoinsResponse {
...
+
 /** 
 `AddressType` has to be one of:
 
 - `p2wkh`: Pay to witness key hash (`WITNESS_PUBKEY_HASH` = 0)
 - `np2wkh`: Pay to nested witness key hash (`NESTED_PUBKEY_HASH` = 1)
 */
-message NewAddressRequest {
-    enum AddressType {
+enum AddressType {
```
halseth: Will this be a breaking RPC change?

AdamISZ: I don't think so. It tests fine on the lncli command line; I also just tested it over grpc using a python client and it seemed to work fine.

Roasbeef:
After one final rebase (hopefully!?) that also squashes the commits, I think this is ready to land!

AdamISZ:
Rebased again. I removed the created change to `go.sum`; I actually don't know what it is, so let me know if that was the wrong thing to do.

cfromknecht:
tACK, just tested on my node; thanks a bunch @AdamISZ!

I have one minor cosmetic comment, which I think has an easy fix. Since we use the `ChannelPoint` proto object to serialize the outpoints, the json fields show `"funding_txid_str"`:
```
            "outpoint": {
                "funding_txid_str": "b616b49afa2b724a71469e798c9dbff24f11f3bfddd38a3f3ef356e6a388926a",
                "output_index": 0
            },
```

I think it might be worth adding an `OutPoint` type to lnrpc and using it here. (IMO arguably `ChanPoint` should just use this as well, but that's out of scope). Idt this is a blocker on the PR though, as it would be an easy fix in a follow up and since we're in between releases it should be fine if the final API changes slightly.

Appreciate all the hard work on this feature. I know a lot people have requested it, so thanks again :)

lnrpc/rpc.proto
halseth: It will break usages of `NewAddressRequests` for projects importing `lnrpc`, but I think that's fine. It is only a simple name change.

rpcserver.go
```diff
@@ -676,6 +680,85 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+
+	if minConfs < 0 {
+		return nil, fmt.Errorf("min confirmations must be >= 0")
+	}
+	if minConfs > maxConfs {
+		return nil, fmt.Errorf("max confirmations must be >= min " +
+			"confirmations")
+	}
+
+	usingMax := maxConfs != math.MaxInt32
```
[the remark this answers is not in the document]
AdamISZ: Done

rpcserver.go
```diff
@@ -676,6 +680,85 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+	if usingMax {
+		rpcsLog.Infof("[listunspent] min=%v max=%v", minConfs, maxConfs)
```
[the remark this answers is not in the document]
AdamISZ: Done

rpcserver.go
```diff
@@ -676,6 +680,85 @@ func determineFeePerKw(feeEstimator lnwallet.FeeEstimator, targetConf int32,
...
+			OutputIndex: utxo.OutPoint.Index,
+		}
+		utxoresp := lnrpc.Utxo{
+			Type:          addrType,
+			AmountSat:     int64(utxo.Value),
+			ScriptPubkey:  hex.EncodeToString(utxo.PkScript),
+			Outpoint:      outpoint,
+			Confirmations: utxo.Confirmations,
+		}
+		_, outAddresses, _, err := txscript.ExtractPkScriptAddrs(
```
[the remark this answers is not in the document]
AdamISZ: Done

AdamISZ:
Squashed + addressed nits, cheers.

alexbosworth:
May be an issue with this method when listing unspents after a seed restore

PR #1987, opened 2018-09-27
rpc: remove NewWitnessAddress RPC

### LINKED ISSUE #1986: REST API newaddress not taking addresstype param
### Background

REST API newaddress not taking addresstype param

### Your environment
lnd v0.5.0
Windows 10
bitcoind 0.16.2
WebUI: Angular 6, Node 8.12.0, TypeScript 2.7.2

### Steps to reproduce

NA

### Expected behaviour

REST API should be able to accept 'addressType' as request param.

### Actual behaviour

No request param.
newaddress

wpaulino:
In this commit, we remove the unnecessary NewWitnessAddress RPC. We do
this as, at the moment, we only expose the wallet to generate witness
addresses. Now that the RPC has been removed, its REST endpoint has been
assigned to the NewAddress RPC, which should be the de facto way of
having the wallet generate any type of supported address.

Fixes #1986.


lnrpc/rpc.proto
```diff
@@ -249,12 +249,7 @@ service Lightning {
     /** lncli: `newaddress`
     NewAddress creates a new address under control of the local wallet.
     */
-    rpc NewAddress (NewAddressRequest) returns (NewAddressResponse);
-
-    /**
-    NewWitnessAddress creates a new witness address under control of the local wallet.
-    */
-    rpc NewWitnessAddress (NewWitnessAddressRequest) returns (NewAddressResponse) {
+    rpc NewAddress (NewAddressRequest) returns (NewAddressResponse) {
```
Roasbeef: The below needs to capture a value such that the user can specify which address type on the REST interface.

wpaulino: Tested this as is and the parameter can still be specified, but it's optional. In the case that it isn't specified, the default will be used (P2WPKH).

rodasmith:
I'm pretty late to this party, but I can finally Ack this, having verified that these commands work as expected:
* `$ curl -X GET --cacert "$LND_DIR/tls.cert" --header "$MACAROON_HEADER" https://localhost:8080/v1/newaddress?type=NESTED_PUBKEY_HASH`
* `$ curl -X GET --cacert "$LND_DIR/tls.cert" --header "$MACAROON_HEADER" https://localhost:8080/v1/newaddress?type=WITNESS_PUBKEY_HASH`

PR #1988, opened 2018-09-28
rpc: Add SubscribeChannels RPC.

### LINKED ISSUE #1235: rpc+lnrpc: add new SubcribeChannel RPC for async notifications of channel state changes
We currently have streaming RPC's to be notified of changes to our network view, on-chain balances, off-chain balances, and a few others. However, we don't yet have a streaming RPC that's dedicated to notifying callers of changes w.r.t the state of our set of channels. Potential state transitions that we'd want to notify on include: new open channels, new closed channels, channels becoming active/inactive, channel force close events, partial contract sweep events. Implementing a set of rich responses for the `SubscribeChannel` streaming RPC would allow GUIs to be more reactive as they'll no longer have to poll to obtain the latest channel state. 

Steps To Completion:

- [ ] Identify all areas in the codebase that can trigger changes to the set of open/active channels. Starting locations include: the `fundingManager`, the `contractcourt` package, the `chanCloser` (within the peer), and the `channelLink`. 

- [ ] Once these areas are identified, either introduce a new set of message between these sub-systems, or intercept the messages that are usually delivered via callbacks/closures, and then pipe these through to the new sub-system to dispatch notifications to subscribed callers.

valentinewallace:
Closes #1235. 

In this PR, we add a new streaming RPC call that allows clients to be notified whenever a channel becomes inactive, active, or closed. This is useful for any type of application that would otherwise need to poll the channel state to keep up to date on what channels are active, inactive, or closed. 

In service of this goal, we add an additional subsystem called `ChannelNotifier` that notifies its clients upon any of the aforementioned channel state changes. It uses two hooks within the peer to know when channels become active/inactive. 

For closed channel events, we add a proxy in the `ChainArbitrator` that proxies between all of its `chainWatcher`s' closed channel notifications and pipes them into the `ChannelNotifier`.


contractcourt/chain_arbitrator.go
```diff
@@ -155,6 +159,11 @@ type ChainArbitrator struct {
 	// that are still considered open.
 	activeWatchers map[wire.OutPoint]*chainWatcher
 
+	// newChainWatcher is used by the closedChanNotifier goroutine to find
+	// out about new chain watchers and subscribe to their channel event
+	// notifications, which will be passed on to the chanNotifier.
+	newChainWatcher chan *chainWatcher
```
halseth: I think a better approach is to utilize the `MarkChannelClosed` closure to get the close events.

rpcserver.go
```diff
@@ -1804,69 +1798,94 @@ func (r *rpcServer) ListChannels(ctx context.Context,
...
+	nodePub := dbChannel.IdentityPub
+	nodeID := hex.EncodeToString(nodePub.SerializeCompressed())
+	chanPoint := dbChannel.FundingOutpoint
+
+	// With the channel point known, retrieve the network channel
+	// ID from the database.
+	var chanID uint64
+	chanID, _ = graph.ChannelID(&chanPoint)
+
+	// Next, we'll determine whether we should add this channel to
```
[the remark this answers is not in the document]
valentinewallace: fixed

channotifier.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/lightningnetwork/lnd/chainntnfs"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/contractcourt"
+	"github.com/lightningnetwork/lnd/lnwallet"
+)
+
+// chanNotifier is a subsystem which all active, inactive, and closed channel
+// events pipe through. It takes subscriptions for its events, and whenever
+// it receives a new event it notifies its subscribers over the proper channel.
+type chanNotifier struct {
```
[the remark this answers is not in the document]
valentinewallace: haha true, fixed

channotifier.go
```diff
@@ -0,0 +1,281 @@
+package main
```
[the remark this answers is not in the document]
valentinewallace: fixed

peer.go
```diff
@@ -584,6 +584,28 @@ func (p *peer) addLink(chanPoint *wire.OutPoint,
 	// this channel can be used to dispatch local payments and also
 	// passively forward payments.
 	return p.server.htlcSwitch.AddLink(link)
+// fetchDBChannel fetches a specific open channel from the database from its
```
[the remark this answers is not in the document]
valentinewallace: fixed since all changes in `peer.go` have been removed

peer.go
```diff
@@ -583,7 +583,25 @@ func (p *peer) addLink(chanPoint *wire.OutPoint,
 	// With the channel link created, we'll now notify the htlc switch so
 	// this channel can be used to dispatch local payments and also
 	// passively forward payments.
-	return p.server.htlcSwitch.AddLink(link)
+	err := p.server.htlcSwitch.AddLink(link)
+	if err != nil {
+		return err
+	}
+
+	// Since the link addition has succeeded, this chanel is now officially
```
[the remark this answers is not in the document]
valentinewallace: fixed

peer.go
```diff
@@ -1622,10 +1640,18 @@ out:
 				TimeLockDelta: defaultPolicy.TimeLockDelta,
 			}
 
+			// We'll fetch the channel from the database so we can
+			// tell the chanNotifier about it.
+			dbChan, err := p.fetchDBChannel(newChan.ShortChanID())
```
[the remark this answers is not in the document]
valentinewallace: fixed

peer.go
```diff
@@ -1748,9 +1748,24 @@ out:
...
 			for _, channel := range p.activeChannels {
 				channel.ResetState()
+
+				dbChan, err := p.fetchDBChannel(channel.ShortChanID())
+				if err != nil {
+					peerLog.Errorf("unable to fetch channel from db: %v", err)
+					continue
+				}
+
+				event := &activeChanEvent{
```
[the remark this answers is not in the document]
valentinewallace: good one, fixed

contractcourt/chain_arbitrator.go
valentinewallace: agree, implemented!

server.go
```diff
@@ -345,6 +345,18 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
 		Notifier:               s.cc.chainNotifier,
+		NotifyActiveChannelEvent: func(chanID lnwire.ShortChannelID) {
+			select {
+			case s.channelNotifier.NewActiveChannels.ChanIn() <- chanID:
+			case <-s.quit:
+			}
+		},
+		NotifyInactiveChannelEvent: func(chanID lnwire.ShortChannelID) {
+			select {
+			case s.channelNotifier.NewInactiveChannels.ChanIn() <- chanID:
```
halseth: To make sure is always non-blocking, I'm wondering if we should select on `default` here? Or somehow determine that the queue is active somehow.

Need to think about this abit.

lnrpc/rpc.proto
```diff
@@ -1301,6 +1309,18 @@ message PendingChannelsResponse {
...
+message ChannelSubscription {
+}
+
+message ChannelUpdate {
+    oneof channel {
+        Channel open_channel = 1 [ json_name = "channel" ];
+        ChannelCloseSummary closed_channel = 2 [ json_name = "closed_channel" ];
+    }
+
+    string type = 3 [ json_name = "type" ];
```
halseth: should this be an enum instead?

rpcserver.go
```diff
@@ -1829,70 +1801,129 @@ func (r *rpcServer) ListChannels(ctx context.Context,
...
 		}
+	}
 
-		resp.Channels = append(resp.Channels, channel)
+	return channel
+}
+
+// createRPCClosedChannel creates an *lnrpc.ChannelChannelSummary from a
+// *channeldb.ChannelCloseSummary.
+func createRPCClosedChannel(dbChannel *channeldb.ChannelCloseSummary) *lnrpc.ChannelCloseSummary {
```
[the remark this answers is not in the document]
valentinewallace: fixed

rpcserver.go
```diff
@@ -1829,70 +1801,129 @@ func (r *rpcServer) ListChannels(ctx context.Context,
...
+			HashLock:         rHash[:],
+			ExpirationHeight: htlc.RefundTimeout,
 		}
+	}
 
-		resp.Channels = append(resp.Channels, channel)
+	return channel
+}
+
+// createRPCClosedChannel creates an *lnrpc.ChannelChannelSummary from a
```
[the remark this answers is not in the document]
valentinewallace: :woman_facepalming: fixed lol

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// ChannelNotifier is a subsystem which all active, inactive, and closed channel
+// events pipe through. It takes subscriptions for its events, and whenever
+// it receives a new event it notifies its subscribers over the proper channel.
+type ChannelNotifier struct {
+	sync.RWMutex
+
+	// These queues hold the events that the notifier is receiving,
```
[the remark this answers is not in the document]
valentinewallace: ty fixed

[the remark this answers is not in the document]
valentinewallace: ty fixed!

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,185 @@
...
+
+	started uint32
+	stopped uint32
+	wg      sync.WaitGroup
+	quit    chan struct{}
+}
+
+// ActiveChannelEvent represents a new event where a channel becomes active.
+type ActiveChannelEvent struct {
+	Channel *channeldb.OpenChannel
```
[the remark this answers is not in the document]
valentinewallace: fixed

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,185 @@
...
+
+// ClosedChannelEvent represents a new event where a channel becomes closed.
+type ClosedChannelEvent struct {
+	CloseSummary *channeldb.ChannelCloseSummary
+}
+
+// NewChannelNotifier creates a new channel notifier. The ChannelNotifier gets channel
+// events from peers and from the chain arbitrator, and dispatches them to
+// its clients.
+func NewChannelNotifier(chanDB *channeldb.DB) *ChannelNotifier {
```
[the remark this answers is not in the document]
valentinewallace: fixed

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,185 @@
...
+		ntfnServer:          subscribe.NewServer(),
+		db:                  chanDB,
+		quit:                make(chan struct{}),
+	}
+}
+
+// Start starts the ChannelNotifier and all goroutines it needs to carry out its task.
+func (c *ChannelNotifier) Start() error {
+	if !atomic.CompareAndSwapUint32(&c.started, 0, 1) {
+		return errors.New("ChannelNotifier already started")
```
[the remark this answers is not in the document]
valentinewallace: :ok_hand: fixed

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,185 @@
...
+
+		case <-c.quit:
+			return
+		}
+	}
+}
+
+// fetchDBChannel fetches a specific open channel from the database from its
+// short channel ID.
+func (c *ChannelNotifier) fetchDBChannel(id lnwire.ShortChannelID) (*channeldb.OpenChannel, error) {
```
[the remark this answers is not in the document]
valentinewallace: n/a :)

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,185 @@
...
+	var dbChan *channeldb.OpenChannel
+	activeChans, err := c.db.FetchAllOpenChannels()
```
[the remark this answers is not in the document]
valentinewallace: good idea, implemented :)

contractcourt/chain_arbitrator.go
```diff
@@ -251,10 +255,13 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 			return chanMachine.ForceClose()
 		},
 		MarkCommitmentBroadcasted: channel.MarkCommitmentBroadcasted,
-		MarkChannelClosed:         channel.CloseChannel,
-		IsPendingClose:            false,
-		ChainArbitratorConfig:     c.cfg,
-		ChainEvents:               chanEvents,
+		MarkChannelClosed: func(summary *channeldb.ChannelCloseSummary) error {
+			c.cfg.NotifyClosedChannel(summary)
+			return channel.CloseChannel(summary)
```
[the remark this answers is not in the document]
valentinewallace: good one fixed

server.go
valentinewallace: let me know if you like my solution to this!

lnrpc/rpc.proto
valentinewallace: true fixed

channeldb/db.go
```diff
@@ -288,6 +290,32 @@ func fileExists(path string) bool {
...
 }
 
+// FetchOpenChannel starts a new database transaction and returns the stored
+// active/open channel associated with the target nodeID and channel ID. In the
+// case that there is no channel matching this criteria, an error is returned.
+// In the case that we fail to fetch any open channels, an error is returned.
+func (d *DB) FetchOpenChannel(
+	nodeID *btcec.PublicKey, chanID lnwire.ShortChannelID) (*OpenChannel, error) {
+
+	channels, err := d.FetchOpenChannels(nodeID)
```
halseth: Instead this should be done by fetching the channel directly from the bucket, instead of iterating over all the node's channels. Not sure if this is compatible with the `shortChanID`, so might have to pass in the `chanPoint` instead,

valentinewallace: (probably) fixed. now we iterate over a node's `chainhash` buckets, but I think that is acceptable?

valentinewallace: fixed, we still iterate over the `chainhash` buckets for a node but I think that's acceptable?

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,192 @@
...
+		log.Warnf("unable to notify on active channel event: active " +
+			"channels queue unexpectedly closed before notifier quit")
+	}
+}
+
+// NotifyInactiveChannelEvent notifies the channelEventNotifier goroutine that a
+// channel has become inactive.
+func (c *ChannelNotifier) NotifyInactiveChannelEvent(channel *channeldb.OpenChannel) {
+	select {
+	case c.NewInactiveChannels.ChanIn() <- channel:
```
halseth: Using these concurrent queues we risk that events don't reach clients in the order they are notified. Instead we should share the queue for all events, or just send the event directly to the `channelEventNotifier` here.

htlcswitch/switch.go
```diff
@@ -1932,6 +1940,14 @@ func (s *Switch) AddLink(link ChannelLink) error {
 		)
 	}
 
+	// Since the link addition has succeeded, this channel is now officially
+	// active and we can tell the ChannelNotifier about it.
+	channel, err := s.cfg.DB.FetchOpenChannel(
+		link.Peer().PubKey(), link.ChannelPoint())
+	if err == nil {
```
halseth: print error

htlcswitch/switch.go
```diff
@@ -2024,6 +2040,15 @@ func (s *Switch) removeLink(chanID lnwire.ChannelID) ChannelLink {
 		return nil
 	}
 
+	// We can now tell the ChannelNotifier that the channel corresponding to this
+	// link is no longer active.
+	channel, err := s.cfg.DB.FetchOpenChannel(
+		link.Peer().PubKey(), link.ChannelPoint())
+
```
halseth: nit: extra newline

htlcswitch/switch.go
```diff
@@ -2024,6 +2040,15 @@ func (s *Switch) removeLink(chanID lnwire.ChannelID) ChannelLink {
...
+	if err == nil {
```
halseth: print error

contractcourt/chain_arbitrator.go
```diff
@@ -710,13 +720,7 @@ func (c *ChainArbitrator) WatchNewChannel(newChan *channeldb.OpenChannel) error
 // SubscribeChannelEvents returns a new active subscription for the set of
 // possible on-chain events for a particular channel. The struct can be used by
 // callers to be notified whenever an event that changes the state of the
-// channel on-chain occurs. If syncDispatch is true, then the sender of the
```
halseth: 👍

lnd_test.go
```diff
@@ -5052,6 +5052,218 @@ func testMultiHopOverPrivateChannels(net *lntest.NetworkHarness, t *harnessTest)
 	closeChannelAndAssert(ctxt, t, net, carol, chanPointCarol, false)
 }
 
+func testChannelSubscription(net *lntest.NetworkHarness, t *harnessTest) {
```
halseth: two suggestions:
1. Make the integration test change its own commit.
2. Can the same be tested by adding the channel subscription to one or more existing tests? The integration tests are long-running at this point, and if we can avoid the overhead of extra tests that would be a win I think :)

rpcserver.go
```diff
@@ -1865,8 +1870,7 @@ func createRPCChannel(r *rpcServer, graph *channeldb.ChannelGraph,
 	var chanID uint64
 	chanID, _ = graph.ChannelID(&chanPoint)
 
-	// Next, we'll determine whether we should add this channel to
-	// our list depending on the type of channels requested to us.
+	// Next, we'll determine whether the channel is public or not.
```
halseth: minor nit: make part of the original commit that changed this.

rpcserver.go
```diff
@@ -1966,7 +1970,61 @@ func createRPCClosedChannel(
...
+				updateStream.Send(update)
+			case channelnotifier.ClosedChannelEvent:
+				closedChannel := createRPCClosedChannel(event.CloseSummary)
+				update := &lnrpc.ChannelUpdate{
+					Type: lnrpc.ChannelUpdate_CLOSED_CHANNEL,
+					Channel: &lnrpc.ChannelUpdate_ClosedChannel{
+						ClosedChannel: closedChannel,
+					},
+				}
+				updateStream.Send(update)
```
halseth: inspect the errors of the `Send`s.

rpcserver.go
```diff
@@ -1966,7 +1970,61 @@ func createRPCClosedChannel(
 	}
 }
 
-	return resp, nil
+// SubscribeChannels returns a uni-directional stream (server -> client) for
+// notifying the client of newly active, inactive or closed channels.
+func (r *rpcServer) SubscribeChannels(req *lnrpc.ChannelSubscription,
```
halseth: 👍

htlcswitch/link.go
```diff
@@ -1736,6 +1737,11 @@ func (l *channelLink) Peer() lnpeer.Peer {
 	return l.cfg.Peer
 }
 
+// ChannelPoint returns the channel outpoint for the channel link.
```
halseth: add ` // NOTE: Part of the ChannelLink interface.`

channeldb/db.go
```diff
@@ -288,6 +288,79 @@ func fileExists(path string) bool {
 	return true
 }
 
+// FetchOpenChannel starts a new database transaction and returns the stored
+// active/open channel associated with the target nodeID and channel ID. In the
+// case that there is no channel matching this criteria, an error is returned.
+// In the case that we fail to fetch any open channels, an error is returned.
+func (d *DB) FetchOpenChannel(
+	nodeID [33]byte, chanPoint *wire.OutPoint) (*OpenChannel, error) {
```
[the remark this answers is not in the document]
valentinewallace: fixed

channeldb/db.go
```diff
@@ -288,6 +288,79 @@ func fileExists(path string) bool {
...
+			if chanBucket == nil {
+				return nil
+			}
+
+			// Finally, with desired channel bucket retrieved, fetch
+			// the *channeldb.OpenChannel that corresponds to it.
+			channel, err = fetchOpenChannel(chanBucket, chanPoint)
+			return err
+		})
+	})
```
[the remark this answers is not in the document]
valentinewallace: fixed

channeldb/db.go
```diff
@@ -288,6 +288,79 @@ func fileExists(path string) bool {
...
+
+	if channel == nil {
+		err = fmt.Errorf("unable to find channel with outpoint=%v", chanPoint)
```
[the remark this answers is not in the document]
valentinewallace: fixed

channeldb/db.go
```diff
@@ -288,6 +288,79 @@ func fileExists(path string) bool {
 	return true
 }
 
+// FetchOpenChannel starts a new database transaction and returns the stored
```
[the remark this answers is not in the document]
valentinewallace: done :)

channelnotifier/channelnotifier.go
valentinewallace: good point, fixed

htlcswitch/switch.go
valentinewallace: fixed

htlcswitch/switch.go
valentinewallace: fixed

htlcswitch/switch.go
valentinewallace: fixed

lnd_test.go
valentinewallace: fixed, added to `testBasicChannelCreation` and renamed the test

rpcserver.go
valentinewallace: fixed

rpcserver.go
valentinewallace: fixed

htlcswitch/link.go
valentinewallace: fixed

rpcserver.go
```diff
@@ -1868,70 +1840,130 @@ func (r *rpcServer) ListChannels(ctx context.Context,
...
-			var rHash [32]byte
-			copy(rHash[:], htlc.RHash[:])
-			channel.PendingHtlcs[i] = &lnrpc.HTLC{
-				Incoming:         htlc.Incoming,
-				Amount:           int64(htlc.Amt.ToSatoshis()),
-				HashLock:         rHash[:],
-				ExpirationHeight: htlc.RefundTimeout,
-			}
+// createRPCChannel creates an *lnrpc.Channel from the *channeldb.Channel.
+func createRPCChannel(r *rpcServer, graph *channeldb.ChannelGraph,
```
cfromknecht: what do you think about naming this `createRPCOpenChannel` to mimic `createRPCClosedChannel`?

rpcserver.go
```diff
@@ -1868,70 +1840,130 @@ func (r *rpcServer) ListChannels(ctx context.Context,
...
+	dbChannel *channeldb.OpenChannel, isActive bool) *lnrpc.Channel {
+
+	nodePub := dbChannel.IdentityPub
+	nodeID := hex.EncodeToString(nodePub.SerializeCompressed())
+	chanPoint := dbChannel.FundingOutpoint
+
+	// With the channel point known, retrieve the network channel
+	// ID from the database.
+	var chanID uint64
+	chanID, _ = graph.ChannelID(&chanPoint)
```
cfromknecht: should we handle the error here?

rpcserver.go
```diff
@@ -1859,7 +1831,7 @@ func (r *rpcServer) ListChannels(ctx context.Context,
 		// Next, we'll determine whether we should add this channel to
 		// our list depending on the type of channels requested to us.
 		isActive := peerOnline && linkActive
-		isPublic := dbChannel.ChannelFlags&lnwire.FFAnnounceChannel != 0
+		channel := createRPCChannel(r, graph, dbChannel, isActive)
```
cfromknecht: great refactor 👍

subscribe/subscribe.go
```diff
@@ -0,0 +1,216 @@
...
+// Quit is a channel that will be closed in case the server decides to no
+// longer deliver updates to this client.
+func (c *Client) Quit() <-chan struct{} {
+	return c.quit
+}
+
+// Server is a struct that manages a set of subscriptions and their
+// corresponding clients. Any update will be delivered to all active clients.
+type Server struct {
+	clientCounter uint64 // To be used atomically.
```
cfromknecht: 64 b4 32, nice @halseth :)

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// ChannelNotifier is a subsystem which all active, inactive, and closed channel
+// events pipe through. It takes subscriptions for its events, and whenever
+// it receives a new event it notifies its subscribers over the proper channel.
+type ChannelNotifier struct {
+	ntfnServer *subscribe.Server
+
+	started uint32
```
cfromknecht: atomic vars need to be first on structs, in this case to to ensure they're 32-bit aligned on all architectures

server.go
```diff
@@ -954,6 +959,9 @@ func (s *server) Start() error {
 	if err := s.utxoNursery.Start(); err != nil {
 		return err
 	}
+	if err := s.channelNotifier.Start(); err != nil {
```
cfromknecht: we may want to move this service up in the start order, to ensure that we don't miss any notifications. perhaps after the chain notifier?

htlcswitch/switch.go
```diff
@@ -1932,6 +1940,18 @@ func (s *Switch) AddLink(link ChannelLink) error {
...
+	// active and we can tell the ChannelNotifier about it.
+	peerPub := link.Peer().PubKey()
+	channel, err := s.cfg.DB.FetchOpenChannel(peerPub[:], link.ChannelPoint())
+	if err != nil {
+		log.Warnf("errored fetching channel with channelpoint=%v: %v",
+			link.ChannelPoint(), err)
+		return nil
+	}
+
+	s.cfg.NotifyActiveChannelEvent(channel)
```
cfromknecht: one design question here, is do we want to notify once the channel becomes active for forwarding? or just notify on whether or not it's in the switch. right now, this will notify on pending and live links

the distinction is that a pending link is still waiting for confirmation, but indicates that funding has completed successfully. it will not be usable until the transaction confirms sufficiently, and the peers have exchanged funding lock msgs. perhaps we want to make distinction in the notification api between a pending and a live link? we could then later send a live ntfn when `UpdateShortChanID` is called on the pending link

htlcswitch/switch.go
```diff
@@ -2024,14 +2044,25 @@ func (s *Switch) removeLink(chanID lnwire.ChannelID) ChannelLink {
...
 
+	// We can now tell the ChannelNotifier that the channel corresponding to this
+	// link is no longer active.
+	peerPub := link.Peer().PubKey()
+	channel, err := s.cfg.DB.FetchOpenChannel(peerPub[:], link.ChannelPoint())
+	switch err {
+	case nil:
+		s.cfg.NotifyInactiveChannelEvent(channel)
+	default:
+		log.Warnf("errored fetching channel with channelpoint=%v: %v",
```
cfromknecht: nit: capitalize log messages

htlcswitch/switch.go
```diff
@@ -1932,6 +1940,18 @@ func (s *Switch) AddLink(link ChannelLink) error {
 		)
 	}
 
+	// Since the link addition has succeeded, this channel is now officially
+	// active and we can tell the ChannelNotifier about it.
+	peerPub := link.Peer().PubKey()
+	channel, err := s.cfg.DB.FetchOpenChannel(peerPub[:], link.ChannelPoint())
+	if err != nil {
+		log.Warnf("errored fetching channel with channelpoint=%v: %v",
```
cfromknecht: Just for consistency, we typically use present tense in most log msgs, e.g. "Unable to fetch"

channeldb/channel_test.go
```diff
@@ -651,6 +651,82 @@ func TestChannelStateTransition(t *testing.T) {
...
+			RefundTimeout: 1,
+			OnionBlob:     []byte("onionblob"),
+		},
+	}
+	if err := state.FullSync(); err != nil {
+		t.Fatalf("unable to save and serialize channel state: %v", err)
+	}
+
+	pubKey := state.IdentityPub.SerializeCompressed()
+	openChannel, err := cdb.FetchOpenChannel(pubKey, testOutpoint)
```
cfromknecht: error should be handled here, and if non-nil should be `Fatalf` for the test

channeldb/channel_test.go
```diff
@@ -651,6 +651,82 @@ func TestChannelStateTransition(t *testing.T) {
 	}
 }
 
+func TestFetchOpenChannel(t *testing.T) {
```
cfromknecht: solid test!

lnd_test.go
```diff
@@ -5288,14 +5288,27 @@ func testInvoiceSubscriptions(net *lntest.NetworkHarness, t *harnessTest) {
 	closeChannelAndAssert(ctxt, t, net, net.Alice, chanPoint, false)
 }
 
-// testBasicChannelCreation test multiple channel opening and closing.
-func testBasicChannelCreation(net *lntest.NetworkHarness, t *harnessTest) {
+// testBasicChannelCreationAndUpdates tests multiple channel opening and closing,
+// and ensures that if a node is subscribed to channel updates they will be
+// received correctly for both cooperative and force closed channels.
+func testBasicChannelCreationAndUpdates(net *lntest.NetworkHarness, t *harnessTest) {
```
cfromknecht: excellent integration test!

channelnotifier/channelnotifier.go
valentinewallace: hm TIL, fixed

channeldb/channel_test.go
valentinewallace: oo fixed

rpcserver.go
valentinewallace: agree, fixed!

rpcserver.go
valentinewallace: looked into this. If we error here, `testOpenChannelAfterReorg` fails because when it calls `closeChannelAndAssert`, `graph.ChannelID()`will be called here and fail because it's unable to retrieve a channel from the edge bucket (due to the channel being reorged out).

Abandoning the channel instead of closing it isn't a possible solution because the tx remains in the mempool, causing later tests to fail.

Mining a few blocks also doesn't solve it because lnd thinks the channel is already open, so it won't add it to the edge bucket again.

Not sure if there's a solution that would fit with this PR here -- am I missing something?

server.go
valentinewallace: good point, fixed

rpcserver.go
halseth: Since this was not done even before this refactor, I think handling the error is out of scope for this PR. The question why the error is not checked here remains though, and should maybe be looked into in a separate issue/PR.

channeldb/channel_test.go
```diff
@@ -651,6 +651,85 @@ func TestChannelStateTransition(t *testing.T) {
...
+		CloseType:         CooperativeClose,
+	}
+	if err := state.CloseChannel(closeSummary); err != nil {
+		t.Fatalf("unable to close channel: %v", err)
+	}
+
+	// As the channel is now closed, attempting to fetch it from the db
+	// should error.
+	openChannel, err = cdb.FetchOpenChannel(pubKey, testOutpoint)
+	if err == nil {
```
halseth: nit: Can be check against `ErrChanNotFound` or similar?

channeldb/channel_test.go
```diff
@@ -651,6 +651,85 @@ func TestChannelStateTransition(t *testing.T) {
...
+		t.Fatalf("fetching nonexistent open channel did not error")
+	}
+	if openChannel != nil {
```
halseth: nit: If `err` is non-nil, usually we don't have to car checking the other return values.

lnd_test.go
```diff
@@ -5307,13 +5320,158 @@ func testBasicChannelCreation(net *lntest.NetworkHarness, t *harnessTest) {
...
+	// has been properly closed on-chain and ensuring that the correct
+	// notifications are received.
+	for i, chanPoint := range chanPoints {
 		ctx, _ := context.WithTimeout(context.Background(), timeout)
-		closeChannelAndAssert(ctx, t, net, net.Alice, chanPoint, false)
+
+		// Force close half of the channels.
+		var force bool
+		switch {
+		case i%1 == 0:
```
halseth: should be `%2`?

halseth: also, could be written
```go
force := i%2 == 0
```

lnd_test.go
```diff
@@ -5307,13 +5320,158 @@ func testBasicChannelCreation(net *lntest.NetworkHarness, t *harnessTest) {
...
+			force = true
+		default:
+			force = false
+		}
+		closeChannelAndAssert(ctx, t, net, net.Alice, chanPoint, force)
+
+		// Since the channel just became inactive, we should receive an
+		// inactive notification from each subscription.
+		inactiveChannelUpdate, err := bobChannelSubscription.Recv()
```
halseth: Do we timeout here if we don't receive a notification? If not, these calls should be extracted into helper methods that let us set a timeout (I think we do something similar for graph updates).

htlcswitch/switch.go
```diff
@@ -1952,6 +1960,16 @@ func (s *Switch) addLiveLink(link ChannelLink) {
 		s.interfaceIndex[peerPub] = make(map[lnwire.ChannelID]ChannelLink)
 	}
 	s.interfaceIndex[peerPub][link.ChanID()] = link
+
+	// Since the link addition has succeeded, this channel is now officially
+	// active and we can tell the ChannelNotifier about it.
+	channel, err := s.cfg.DB.FetchOpenChannel(peerPub[:], link.ChannelPoint())
```
halseth: Not sure if there are performance considerations we should take doing DB queries when adding and removing links. Will let @cfromknecht comment on this . :)

channeldb/channel_test.go
valentinewallace: fixed

channeldb/channel_test.go
valentinewallace: thx, fixed

lnd_test.go
valentinewallace: oops, fixed

channeldb/channel_test.go
```diff
@@ -651,6 +651,83 @@ func TestChannelStateTransition(t *testing.T) {
 	}
 }
 
+func TestFetchOpenChannel(t *testing.T) {
```
halseth: should have a short description

lnd_test.go
```diff
@@ -5386,13 +5387,151 @@ func testInvoiceSubscriptions(net *lntest.NetworkHarness, t *harnessTest) {
...
+	}
+
+	// We'll launch a goroutine that will be responsible for proxying all
+	// notifications recv'd from the client into the channel below.
+	errChan := make(chan error, 1)
+	quit := make(chan struct{})
+	chanUpdates := make(chan *lnrpc.ChannelUpdate, 20)
+	go func() {
+		for {
+			defer cancelFunc()
```
halseth: I don't think the `cancelFunc` should be in the loop?

lnd_test.go
```diff
@@ -5386,13 +5387,151 @@ func testInvoiceSubscriptions(net *lntest.NetworkHarness, t *harnessTest) {
...
+// channel subscriptions.
+type channelSubscription struct {
+	updateChan chan *lnrpc.ChannelUpdate
+	errChan    chan error
+	quit       chan struct{}
+}
+
+// subscribeChannelNotifications subscribes to channel updates and launches a
+// goroutine that forwards these to the returned channel.
+func subscribeChannelNotifications(ctxb context.Context, t *harnessTest,
```
halseth: `ctxb` not used

lnd_test.go
```diff
@@ -5407,11 +5546,118 @@ func testBasicChannelCreation(net *lntest.NetworkHarness, t *harnessTest) {
...
+		forceType lnrpc.ChannelCloseSummary_ClosureType) {
+
+		defer wg.Done()
+
+		// Ensure two notifications are received for each closed channel.
+		numChannelUpds := 0
+		for numChannelUpds < 2*numChannels {
+			// Every other channel should be force closed.
+			force := (numChannelUpds/2)%2 == 0
+			// Inactive ntfns should be received before close ntfns
```
halseth: nit: add newline above

lnd_test.go
halseth: Yes, very thorough!  😀

lnd_test.go
```diff
@@ -5407,11 +5546,118 @@ func testBasicChannelCreation(net *lntest.NetworkHarness, t *harnessTest) {
...
+					"chanupds", numChannelUpds, 2*numChannels)
+				return
+			}
+		}
+	}
+
+	// Verify Bob receives all closed channel notifications. He should
+	// receive a remote force close notification for force closed channels.
+	wg.Add(1)
+	go verifyCloseUpdatesReceived(bobChanSub,
```
halseth: What is the reason you are not checking this sequentially? If for runtime reasons, how much longer does the sequential test run?

channeldb/channel_test.go
valentinewallace: fixed

lnd_test.go
valentinewallace: fixed -- also an issue [here](https://github.com/lightningnetwork/lnd/blob/master/lnd_test.go#L7696). Can open a pr!

lnd_test.go
valentinewallace: fixed, now used

lnd_test.go
valentinewallace: It only made a difference of 0.03 seconds actually, so I got rid of it.

lnd_test.go
valentinewallace: Fixed

lnrpc/rpc.proto
```diff
@@ -1306,6 +1314,24 @@ message PendingChannelsResponse {
     repeated WaitingCloseChannel waiting_close_channels = 5 [ json_name = "waiting_close_channels" ];
 }
 
+message ChannelSubscription {
+}
+
+message ChannelUpdate {
```
halseth: To prevent confusion with the gossip messages, I think we should rename this. What about `ChannelEvent`?

lnrpc/rpc.proto
```diff
@@ -336,6 +336,14 @@ service Lightning {
         };
     }
 
+    /** lncli: `subscribechannels`
+    SubscribeChannels creates a uni-directional stream from the server to the
+    client in which any updates relevant to the state of the channels are sent
+    over. Events include new active channels, inactive channels, and closed
+    channels.
+    */
+    rpc SubscribeChannels (ChannelSubscription) returns (stream ChannelUpdate);
```
halseth: name suggestion: `SubscribeChannelEvents` and `ChannelEventSubscription`.

rpcserver.go
```diff
@@ -2192,7 +2197,65 @@ func createRPCClosedChannel(
...
+					},
+				}
+			case channelnotifier.ClosedChannelEvent:
+				closedChannel := createRPCClosedChannel(event.CloseSummary)
+				update = &lnrpc.ChannelUpdate{
+					Type: lnrpc.ChannelUpdate_CLOSED_CHANNEL,
+					Channel: &lnrpc.ChannelUpdate_ClosedChannel{
+						ClosedChannel: closedChannel,
+					},
+				}
```
halseth: should have a default case returning an error

lnd_test.go
```diff
@@ -5351,14 +5351,151 @@ func testInvoiceSubscriptions(net *lntest.NetworkHarness, t *harnessTest) {
...
+	return nil
+}
+
+// testBasicChannelCreationAndUpdates tests multiple channel opening and closing,
+// and ensures that if a node is subscribed to channel updates they will be
+// received correctly for both cooperative and force closed channels.
+func testBasicChannelCreationAndUpdates(net *lntest.NetworkHarness, t *harnessTest) {
 	const (
 		numChannels = 2
+		timeout     = time.Duration(time.Second * 5)
```
halseth: check out the new `defaultTimeout` :)

lnd_test.go
```diff
@@ -5351,14 +5351,151 @@ func testInvoiceSubscriptions(net *lntest.NetworkHarness, t *harnessTest) {
...
 		amount      = maxBtcFundingAmount
 	)
+	ctxb := context.Background()
```
halseth: style nit: define ctxb first line in test.

lnrpc/rpc.proto
valentinewallace: agree, fixed

lnrpc/rpc.proto
valentinewallace: fixed

lnd_test.go
valentinewallace: cool :) fixed

lnd_test.go
valentinewallace: fixed

rpcserver.go
Roasbeef: We don't check the error here since if the channel hasn't been announced yet, or marked fully open, then it won't have a channel ID set. In this case, it'll display as a chan ID of zero (since it doesn't have one yet).

lnrpc/rpc.proto
```diff
@@ -1362,6 +1370,24 @@ message PendingChannelsResponse {
...
+message ChannelEventSubscription {
+}
+
+message ChannelEventUpdate {
+    oneof channel {
+        Channel open_channel = 1 [ json_name = "channel" ];
+        ChannelCloseSummary closed_channel = 2 [ json_name = "closed_channel" ];
+    }
+
+    enum UpdateType {
```
Roasbeef: As is, this PR is missing pending channels, and also newly opened channels. A channel can be open without actually being active. Please refer back to the original issue.

htlcswitch/switch.go
```diff
@@ -1952,6 +1960,10 @@ func (s *Switch) addLiveLink(link ChannelLink) {
 		s.interfaceIndex[peerPub] = make(map[lnwire.ChannelID]ChannelLink)
 	}
 	s.interfaceIndex[peerPub][link.ChanID()] = link
+
+	// Since the link addition has succeeded, this channel is now officially
+	// active and we can tell the ChannelNotifier about it.
+	s.cfg.NotifyActiveChannelEvent(link.ChannelState())
```
Roasbeef: A channel only becomes active once this method returns true: https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L459

htlcswitch/interfaces.go
```diff
@@ -110,6 +110,10 @@ type ChannelLink interface {
 	// will use this function in forwarding decisions accordingly.
 	EligibleToForward() bool
 
+	// ChannelState returns a pointer to a channeldb.OpenChannel representing
+	// the state of the channel.
+	ChannelState() *channeldb.OpenChannel
```
Roasbeef: I don't think we need to expose underlying details like this, it effectively weakens the abstraction. Instead, the switch only needs to tell the channel notifier of a channel point, and the notifier can gain a new specific interface which allows it to fetch a channel by its outpoint.

lnrpc/rpc.proto
Roasbeef: Events for pending channels should be emitted from the funding manager. Events for opened channels should be emitted either at the funding manager or peer's `channelManager` goroutine.

subscribe/subscribe_test.go
```diff
@@ -0,0 +1,110 @@
...
+		// were cancelled.
+		case i < numClients/3:
+			select {
+			case <-c.Quit():
+				continue
+			case <-time.After(1 * time.Second):
+				t.Fatalf("cancelled client %v did not quit", i)
+			}
+
+		// The next third should receive all updats.
```
Roasbeef: updats -> updates

channelnotifier/channelnotifier.go
```diff
@@ -0,0 +1,105 @@
...
+	// Send the active event to all channel event subscribers.
+	event := ActiveChannelEvent{Channel: channel}
+	if err := c.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send active channel update: %v", err)
+	}
+}
+
+// NotifyInactiveChannelEvent notifies the channelEventNotifier goroutine that a
+// channel has become inactive.
+func (c *ChannelNotifier) NotifyInactiveChannelEvent(channel *channeldb.OpenChannel) {
```
Roasbeef: See comment re taking a chan point and having the notifier look it up itself. We also avoid passing around a context to live channel state more than we actually need to.

lnrpc/rpc.proto
Roasbeef: Also the following state transitions are missing:
  * open -> pending close -> closed

channelnotifier/channelnotifier.go
AndrewSamokhvalov: Somewhere in ideal world, I think API developers might expect updates on following events 😄:

- opening
  - on enter of state
  - on confirmation change
- opened
  - on enter of state
  - on commit fee change (not for every, but in 5-10 sec window)
  - on local / remote balance change (not for every, but in 5-10 sec window)
  - on pending htlc appear / disappear
- waiting close
  - on enter of state
- closing
  - on enter of state
  - on pending htlc swipe
- closed
  - on enter of state
  - on time locked balance change

htlcswitch/interfaces.go
valentinewallace: fixed -- now we expose `ChannelPoint()` however

subscribe/subscribe_test.go
valentinewallace: fixed

channelnotifier/channelnotifier.go
valentinewallace: fixed :+1:

lnrpc/rpc.proto
valentinewallace: discussed, pending ntfns will be in separate PR

htlcswitch/switch.go
valentinewallace: fixed if i interpreted this correctly..

peer.go
```diff
@@ -1693,6 +1693,11 @@ out:
 				continue
 			}
 
+			// Inform the ChannelNotifier that this channel has transitioned
+			// from pending open to open.
+			p.server.channelNotifier.NotifyOpenChannelEvent(newChan.IdentityPub,
```
halseth: Would a better place for this be in the `fundingmanager`? If we want this to notify about the channel being "marked open" in the database, then it could be done here: https://github.com/lightningnetwork/lnd/blob/441307e8f7ede6f15ea7c1897de732ade6544ab9/fundingmanager.go#L1935

Roasbeef: Agreed would be nice to move to the `fundingManager`. Would also get rid of another instance of usage of the server's "God" pointer.

valentinewallace: fixed :)

server.go
Jircs1: 👍

PR #1991, opened 2018-09-28
routing/pathfind: adding flag to check insufficient funds for path

### LINKED ISSUE #1680: Incorrect payment error message
### Background

I have an open channel with "local_balance": "0", and I try to pay an Invoice to the node I'm direcly connected by the channel

### Your environment

* lnd version 0.4.2-beta commit=f0f5e11b826e020c11c37343bcbaf9725627378b
* Linux wepa 4.15.0-29-generic #31-Ubuntu SMP Tue Jul 17 15:39:52 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* btcd version 0.12.0-beta

### Steps to reproduce
```
$ lncli payinvoice lnsb10u....cqde6eyq
```
### Expected behaviour

I'm a sender with not enough funds for this invoice, I should get an error message that indicates to me that I don't have enough funds

```
{
	"payment_error": "not enough funds",
	"payment_preimage": "",
	"payment_route": null
}
```

### Actual behaviour

I receive this error message

```
{
	"payment_error": "unable to find a path to destination",
	"payment_preimage": "",
	"payment_route": null
}
```

ccdle12:
This commit adds a flag 'bandwithFlag' in findPath(). If finding a
complete path to the source node has failed, we can check the
bandwithFlag to see if the reason was due to insufficient bandwith.
If bandwithFlag is activated we return an error 'unable to find path
with sufficient funds to the destination', instead of simply 'unable
to find path to destination'.

Fixes #1680

### DIFF
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -554,6 +554,10 @@ func findPath(tx *bolt.Tx, graph *channeldb.ChannelGraph,
 	// mapped to within `next`.
 	next := make(map[Vertex]*ChannelHop)
 
+	// We'll activate this flag if the bandwidth of the channel edge
+	// does not have enough funds to route the transaction.
+	var bandwithFlag bool
+
 	// processEdge is a helper closure that will be used to make sure edges
 	// satisfy our specific requirements.
 	processEdge := func(fromNode *channeldb.LightningNode,
@@ -561,6 +565,9 @@ func findPath(tx *bolt.Tx, graph *channeldb.ChannelGraph,
 		bandwidth lnwire.MilliSatoshi, toNode Vertex) {
 
 		fromVertex := Vertex(fromNode.PubKeyBytes)
+
+		// Reset the bandWithFlag each time we process an edge node.
+		bandwithFlag = false
 
 		// If the edge is currently disabled, then we'll stop here, as
 		// we shouldn't attempt to route through it.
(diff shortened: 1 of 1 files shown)


joostjager:
In the path finding loop, multiple routes are explored until one is found that satisfies all conditions. If none is found, there may be multiple reasons why. It could be that there is no connection at all, channels can be disabled, min htlc not satisfied, fee limit exceeded or not enough capacity in channels, and together all these conditions on all possible routes lead to the routing failure. 

So I'd say it is difficult to attribute a routing failure to a single cause. If I see it right, in your code you base the failure reason only on the result of the last edge explored, which seems random knowing that there can also be other reasons. The order of exploration becomes a factor in determining the failure cause.

Maybe a clearer statement can be obtained by just checking whether none of the sender node's enabled direct outbound channels have enough balance. If not, it is probably fair to conclude that there are insufficient funds.

ccdle12:
Thanks for the review @joostjager.

I made the assumption that the last edge processed would be the direct hop (the node with the open channel) but after reviewing `findPath()`, as you said this could be any node explored, not necessarily the direct hop.

I'll follow up on the suggestion of calling the DB to fetch all the open channels and making the check most likely in the rpcserver.

I'll close this PR and start working on that.

joostjager:
One complication that I think of now, is that you don't yet know how much balance you need if you'd include fees to be paid along the way.

ccdle12:
That's true, 

The check could be a very simple check.

1. Get all the Open Channels
2. If all the channels cannot satisfy `is payAmt < channel.LocalCommitment.LocalBalance `
3. Then there are no direct open channels that can handle the routing

You're correct with the issue of fees for each hop. You could have have a pay amount just under the available local balance in a channel but with each hop, it goes beyond what's available in the local balance. In that case the error that will be returned, will be the one from `findPath()` which would be `unable to find path to destination`

PR #2006, opened 2018-10-04
Scoring based autopilot attachment heuristic

halseth:
This PR makes the current autopilot attachment heuristic _scoring based_, meaning that instead of returning a set of candidate nodes to open channels to, it takes a list of nodes (currently all nodes in the graph), and scores them according to the existing model.

This is done as a step in making it possible for the autopilot to select nodes from a diverse set of models. Making them scoring based makes it easier to combine the result from multiple heuristics when considering candidates for channels.

The new API `NodeScores` takes the current channel graph, the existing local channels and available funds, and a set of nodes. It returns a map of scored nodes, where a higher score indicates it is more preferred as a channel counterparty. Note that we currently only require the returned scores to be non-negative, higher meaning more preferred. As we want to pick a node at random according to this score distribution, we therefore must normalize the scores before picking a channel candidate.

### TODO:
- [x] Update existing unit tests to new API
- [ ] Add more unit test for new `NodeScores` implementation
- [ ] Add more unit tests for the agent's use of `NodeScores`
- [ ] Add unit tests for `weightedChoice` and `chooseN`
- [ ] Increase logging around channel opening loop


autopilot/agent.go
```diff
@@ -473,17 +475,33 @@ func (a *Agent) controller() {
...
+		// channel attachment heuristics to determine if we should open
 		// up any additional channels or modify existing channels.
-		availableFunds, numChans, needMore := a.cfg.Heuristic.NeedMoreChans(
-			totalChans, a.totalBalance,
-		)
-		if !needMore {
+		var heuristic AttachmentHeuristic
+		var availableFunds btcutil.Amount
+		var numChans uint32
+		for _, h := range a.cfg.Heuristics {
```
joostjager: I would introduce a struct called `CompositeAttachmentHeuristic` that executes this first come first serve logic. The list of child-heuristics is maintained in that struct. Then at least this code doesn't need to be aware of multiple heuristics being active. 

It is then also an option to move the `moreChans` / `distribute` calls in that struct.

halseth:
Blocked by https://github.com/lightningnetwork/lnd/pull/2039

Roasbeef:
> So every time autopilot wakes up, it will iterate over all nodes and for every candidate node make a call into a heuristic. 

The idea is to remove the idea of candidate nodes (which are generated via a single heuristic), and instead move to a model where we have a set of scoring heuristics, we obtain the top N from each source, then normalize the scores into a composite heuristic. 

I think the idea of a pre/post candiate attachment evaluation is worthwhile! We can eventually later experiment with actions like: "if I connect to this node, whole will that affect a network/graph wide metric?". From my PoV, they're a distinct, but complementary. This PR is attempting to abstract out the set of heuristics so we can combine data sources, while I think what you're suggesting is more of an incremental optimization engine. Although, thinking about it a bit more, I think this can be layered one, we take the candidates with the best normalized score, then query the global heuristic to determine how attaching to this node will affect the global node score, and open a channel to them based on that information. 

One component that's still generally unaddressed is: _how large of a channel to open?_ Our current thinking is that this will be based on node acceptance policies once they start to more closely monitor their inflows and outflow to achieve a nice equilibrium. The current code as is, will simply attempt to open the largest possible channel, and if it has enough funds, open as many of them as possible. On testnet this is fine, but obviously on mainnet, we don't want to put all our satoshis in one basket.

Roasbeef:
>  So every time autopilot wakes up, it will iterate over all nodes and for every candidate node make a call into a heuristic. It will supply the heuristic with candidate node and info on current channels, graph and available on-chain balance

Where would the set of initial candidates be drawn from? Let's assume the current pref attachment heuristic goes away all together for a sec

autopilot/heuristic_constraints.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// Now that we know we need more funds, we'll compute the amount of
+	// additional funds we should allocate towards channels.
+	targetAllocation := btcutil.Amount(float64(totalFunds) * h.Allocation)
+	fundsAvailable := targetAllocation - totalChanAllocation
+	return fundsAvailable, numAdditionalChans, true
+}
+
+// distribute will take the available funds and passed directives, and return a
```
Roasbeef: Perhaps the heuristics should instead suggest the size of the channel to open? What's implement here was meant to be nothing more than a stop gap.

autopilot/heuristic_constraints.go
```diff
@@ -0,0 +1,121 @@
...
+	// Allocation the percentage of total funds that should be committed to
+	// automatic channel establishment.
+	Allocation float64
+}
+
+// moreChans returns whether we can open more channels and still be within the
+// set constraints. In the case that is possible, the funds available for new
+// channels, and the number of additional channels that can opened is also
+// returned.
+func (h *HeuristicConstraints) moreChans(channels []Channel,
```
Roasbeef: This was initially written with the assumption that there'd only ever be a _single_ heuristic that drove the entire agent. With this multi-heuristics scoring based model, I think we may need to re-think this approach all together.

autopilot/prefattach.go
```diff
@@ -61,45 +52,8 @@ var _ AttachmentHeuristic = (*ConstrainedPrefAttachment)(nil)
...
-	if !needMore {
-		return 0, 0, false
-	}
-
-	// Now that we know we need more funds, we'll compute the amount of
-	// additional funds we should allocate towards channels.
-	targetAllocation := btcutil.Amount(float64(totalFunds) * p.threshold)
-	fundsAvailable := targetAllocation - totalChanAllocation
-	return fundsAvailable, numAdditionalChans, true
+	// We'll try to open more channels as long as the constraints allow it.
```
Roasbeef: So with this change to the scoring based heuristics, I envisioned removing this method `NeedMoreChans` all together. Instead the agent determines this.

Roasbeef: So with this change, we'd have an interface something like: 
```go
type AttachmentHeuritic interface {
    SelectTopN(self *btcec.PublicKey, graph ChannelGraph, skipNodes map[NodeID]struct{}) ([]AttachmentScore, error) 
}
```

Eventually once we move to the retry/queue based model, we remove the `skipNodes` all together.

Roasbeef: In this case `AttachmentScore` is more or less a two-tuple, of the node info (pubkey, etc) and the `float64` score (or an int, whichever).

autopilot/graph.go
```diff
@@ -136,6 +136,42 @@ func (d *databaseChannelGraph) ForEachNode(cb func(Node) error) error {
 	})
 }
 
+// FetchNode returns the Node from the graph for the given NodeID, or
+// ErrNodeNotFound if the node is not found. ErrNoAddresses is returned if the
+// node is found, but has no addresses to connected to.
+//
+// NOTE: Part of the autopilot.ChannelGraph interface.
+func (d *databaseChannelGraph) FetchNode(pubkeyBytes NodeID) (Node, error) {
```
Roasbeef: We may eventually also want to expose the feature bit information as well. As eventually the pilot may be aware of things like watch towers. As an example, if we don't have any channels yet, and want to open a channel, then it may make sense to open a channel up to a tower so we can pay it for tokens to watch over other channels. ofc, we'd need to ensure that we don't have it watch channels that we have directly with the node.

autopilot/scoreattach.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"bytes"
+	"sort"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcutil"
+)
+
+// ScoreAttachment is an implementation of the AttachmentHeuristic
```
Roasbeef: Referring back to my earlier comment, I'd intended this to instead be a modification of the existing interface. So they'd _all_ be scoring based.

Roasbeef: Although for what we have in mind, could pin this as an `ExternalScoringHeuristic`.

autopilot/scoreattach.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Sort them, such that the nodes with the highest score will come
+	// first in the slice.
+	sort.Slice(prefNodes, func(i, j int) bool {
+		return s.nodeScores[prefNodes[i]] > s.nodeScores[prefNodes[j]]
+	})
+
+	// We'll continue our attachment loop until we've exhausted the current
+	// amount of available funds.
+	for _, node := range prefNodes {
```
Roasbeef: For the external scores, we may want to insert a bit of jitter in here. So for example adding a supplementary score based on the xor distance (or just regular distance) of a node's pubkey to our own.

autopilot/agent.go
```diff
@@ -473,17 +475,33 @@ func (a *Agent) controller() {
...
 
 		// Now that we've updated our internal state, we'll consult our
-		// channel attachment heuristic to determine if we should open
+		// channel attachment heuristics to determine if we should open
 		// up any additional channels or modify existing channels.
-		availableFunds, numChans, needMore := a.cfg.Heuristic.NeedMoreChans(
-			totalChans, a.totalBalance,
-		)
-		if !needMore {
+		var heuristic AttachmentHeuristic
```
Roasbeef: IMO the constraints should be lifted to the agent level. So instead, this loop would query all heuristics for their top candidates, and then (to start something simple) sum the scores from each heuristic, and use that as a basis (along with jitter) to select a possible candidate.

lnrpc/autopilot/autopilot.proto
```diff
@@ -42,3 +53,10 @@ message EnableRequest{
 }
 
 message EnableResponse {}
+
+message SetNodeScoresRequest{
```
Roasbeef: 👍

joostjager:
> Where would the set of initial candidates be drawn from? Let's assume the current pref attachment heuristic goes away all together for a sec

The candidate set are in this case all nodes known. So for every node known, heuristics will be queried and they'll return the 'gain' of connecting to this node.

halseth:
Thanks for the review and suggestions! 

The scope of this PR has been reduced to only making the _existing_ attachment heuristic scoring based. It is intended to keep the current autopilot behavior, but making the API scoring based to later making it easy to add heuristics like the discussed "External scoring based" one.

autopilot/interface.go
```diff
@@ -129,6 +145,16 @@ type AttachmentHeuristic interface {
...
 		skipNodes map[NodeID]struct{}) ([]AttachmentDirective, error)
+
+	// NodeScores is a method that given the current channel graph, current
+	// set of local channels and funds available, scores the given nodes
+	// according the the preference of opening a channel with them. The
+	// returned channel candidates maps the NodeID to a score and a channel
+	// size. The scores will be in the range [0, M], where 0 indicates no
+	// improvement in connectivity if a channel is opened to this node,
+	// while M is the maximum possible improvement in connectivity.
+	NodeScores(g ChannelGraph, chans []Channel, fundsAvailable btcutil.Amount,
```
Roasbeef: Depending on how the future heuristics specify this score, we may want to add a degree of normalization across all the scores. Otherwise, we may be comparing something like the in-degree of nodes to the aggregate amount that we've routed through them in the past (just an example).

autopilot/agent.go
```diff
@@ -529,57 +530,135 @@ func (a *Agent) controller() {
...
-		pendingMtx.Lock()
-		if uint16(len(pendingOpens)) >= a.cfg.Constraints.MaxPendingOpens {
-			pendingMtx.Unlock()
-			log.Debugf("Reached cap of %v pending "+
-				"channel opens, will retry "+
-				"after success/failure",
-				a.cfg.Constraints.MaxPendingOpens)
-			continue
-		}
+			s := scores[candidate]
```
Roasbeef: Check the second value here just to guard against bugs in `weightedChoice`?

autopilot/agent.go
```diff
@@ -342,6 +346,38 @@ func mergeChanState(pendingChans map[NodeID]Channel,
...
+
+	// Create a map of normalized scores such, that they sum to 1.0.
+	norm := make(map[NodeID]float64)
+	for k, v := range s {
+		norm[k] = v.Score / sum
+	}
+
+	// Pick a random number in the range [0.0, 1.0), and iterate the map
+	// until the number goes below 0. This means that each index is picked
+	// with a probablity eqaual to their normalized score.
```
Roasbeef: eqaual -> equal

autopilot/agent.go
```diff
@@ -529,57 +530,135 @@ func (a *Agent) controller() {
...
+			log.Debugf("%d chans remaining, %v funds remaining",
+				remChans, remFunds)
+
+			// To avoid looping forever in case of an error, we
+			// decrease the number of remaining channels now.
+			remChans--
+
+			// Now use the score to make a weighted choice which
+			// node to open a channel to.
+			candidate, err := weightedChoice(scores)
```
Roasbeef: The proposed design was to select the node (aggregating all the heuristics) that results in the highest "gain". Atm, this will randomly select a node from the single heuristic. is the idea to leave the generic selection algorithm to a follow up PR?

autopilot/agent.go
```diff
@@ -342,6 +346,38 @@ func mergeChanState(pendingChans map[NodeID]Channel,
...
+	norm := make(map[NodeID]float64)
+	for k, v := range s {
+		norm[k] = v.Score / sum
+	}
+
+	// Pick a random number in the range [0.0, 1.0), and iterate the map
+	// until the number goes below 0. This means that each index is picked
+	// with a probablity eqaual to their normalized score.
+	r := rand.Float64()
+	for k, v := range norm {
```
Roasbeef: I find this selection a bit hard to analyze. First, the map iteration of `norm` is already randomized. So the probability that a node is selected, is dependent on both the order of map iteration _and_ the random value selected. As an example, if the iteration _wasn't_ randomized, and the value .50 was selected each time, and both nodes had the _same_ score (let's say .50 for the sake of the example), then the first node would _always_ be selected. 

One thing we could do, in order to assert guarantees w.r.t the selection here, is to write a set of `testing/quick` tests that target selection probabilities based on scores+ranges.

autopilot/graph.go
```diff
@@ -82,7 +82,7 @@ func (d dbNode) Addrs() []net.Addr {
 //
 // NOTE: Part of the autopilot.Node interface.
 func (d dbNode) ForEachChannel(cb func(ChannelEdge) error) error {
-	return d.node.ForEachChannel(d.tx, func(tx *bolt.Tx,
+	return d.node.ForEachChannel(nil, func(tx *bolt.Tx,
```
Roasbeef: If it's closed or not? The idea is to not require a new db transaction for each node we explore, and instead do it under a single database transaction.

autopilot/agent.go
halseth: Idea is to either here combine scores from different heuristics before making the choice, or create a combined heuristic that abstract this away.

autopilot/agent.go
halseth: Added some explanation. Will add some unit/quick tests when the design is settled.

autopilot/graph.go
halseth: Was able to avoid this change by no longer keeping the `Node` objects around.

halseth:
Significant rewrite pushed, ready for another round of review :)

Tests should also be passing now :D

autopilot/agent.go
```diff
@@ -485,211 +486,221 @@ func (a *Agent) controller() {
...
+			log.Errorf("Unable to open channels: %v", err)
 		}
+	}
+}
 
-		if len(chanCandidates) == 0 {
-			log.Infof("No eligible candidates to connect to")
-			continue
-		}
+// openChans queries the agent's heuristic for a set of channel candidates, and
```
Roasbeef: 👍 

I think some of the comments and code fragments below can be opened up a bit to take advantage of the new column space.

autopilot/agent.go
```diff
@@ -485,211 +486,221 @@ func (a *Agent) controller() {
...
+	for _, chanCandidate := range chanCandidates {
+		// Skip candidates which we are already trying
+		// to establish a connection with.
+		nodeID := chanCandidate.NodeID
+		if _, ok := a.pendingConns[nodeID]; ok {
 			continue
 		}
+		a.pendingConns[nodeID] = struct{}{}
+
+		go func(directive AttachmentDirective) {
```
Roasbeef: Now that the maps etc are a part of the agent, we can also break this goroutine out into its own function. This change would reduce the amount of vertical space utilized by this method in general.

autopilot/prefattach.go
```diff
@@ -257,3 +257,120 @@ func (p *ConstrainedPrefAttachment) Select(self *btcec.PublicKey, g ChannelGraph
...
+		// score of 0, as we don't need another channel.
+		case ok:
+			score = 0
+
+		// If there are no channels in the graph we cannot determine
+		// any preferences, so we set all scores to zero.
+		case graphChans == 0:
+			score = 0
+
+		// Set score 0 if channel is too small.
```
Roasbeef: Seems we can just check this above (outside of the select loop), and continue altogether, not even including this node in the final scoring map.

autopilot/agent.go
Roasbeef: Typo still present.

autopilot/agent.go
Roasbeef: Much clearer now with the comment, thanks!

autopilot/agent.go
```diff
@@ -362,6 +365,73 @@ func mergeChanState(pendingChans map[NodeID]Channel,
...
+	for k, v := range norm {
+		r -= v
+		if r <= 0 {
+			return k, nil
+		}
+	}
+	return NodeID{}, fmt.Errorf("no choice made")
+}
+
+// chooseN picks at random min[n, len(s)] nodes if from the
```
Roasbeef: 👍

autopilot/agent.go
```diff
@@ -578,18 +579,58 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
 
-	// If we reach this point, then according to our heuristic we
-	// should modify our channel state to tend towards what it
-	// determines to the optimal state. So we'll call Select to get
-	// a fresh batch of attachment directives, passing in the
-	// amount of funds available for us to use.
-	chanCandidates, err := a.cfg.Heuristic.Select(
-		a.cfg.Self, a.cfg.Graph, availableFunds,
-		numChans, nodesToSkip,
+	// Gather the set of all nodes in the graph, except those we
```
Roasbeef: Typo in commit message: copmatible -> compatible

autopilot/agent.go
```diff
@@ -578,18 +579,58 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
+			return nil
+		}
+
+		nodes[nID] = struct{}{}
+		return nil
+	}); err != nil {
+		return fmt.Errorf("unable to get graph nodes: %v", err)
+	}
+
+	// Use the heuristic to calculate a score for each node in the
```
Roasbeef: I dig the control flow of this new loop, haven't got to the rest of the commits yet, but as joost suggested in a prior commit, we can make a composite heuristic that itself combines other heuristics and allows us to hide the details w.r.t the eventual normalization scoring we add in.

autopilot/agent.go
```diff
@@ -458,17 +460,17 @@ func (a *Agent) controller() {
 			return
 		}
 
-		pendingMtx.Lock()
-		log.Debugf("Pending channels: %v", spew.Sdump(pendingOpens))
-		pendingMtx.Unlock()
+		a.pendingMtx.Lock()
+		log.Debugf("Pending channels: %v", spew.Sdump(a.pendingOpens))
+		a.pendingMtx.Unlock()
```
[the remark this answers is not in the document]
halseth: Fixed in following commit to keep this one purely a move.

autopilot/agent.go
halseth: Done!

autopilot/agent.go
halseth: Fixed.

autopilot/agent.go
halseth: Done!

autopilot/prefattach.go
halseth: Done.

autopilot/prefattach.go
```diff
@@ -257,3 +257,120 @@ func (p *ConstrainedPrefAttachment) Select(self *btcec.PublicKey, g ChannelGraph
...
+// NOTE: This is a part of the AttachmentHeuristic interface.
+func (p *ConstrainedPrefAttachment) NodeScores(g ChannelGraph, chans []Channel,
+	fundsAvailable btcutil.Amount, nodes map[NodeID]struct{}) (
+	map[NodeID]*AttachmentDirective, error) {
+
+	// Fill the map of candidates to return.
+	// NOTE: we'll set the addrs, score and chansize below.
+	candidates := make(map[NodeID]*AttachmentDirective)
+	for nID := range nodes {
+		pub, err := btcec.ParsePubKey(nID[:], btcec.S256())
```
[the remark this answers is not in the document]
halseth: Fixed by removing the `NodeKey` from the `AttachmentDirective` completely, and parsing only when calling `ConnectToPeer`.

autopilot/prefattach.go
```diff
@@ -257,3 +257,120 @@ func (p *ConstrainedPrefAttachment) Select(self *btcec.PublicKey, g ChannelGraph
...
+
+		_, ok := existingPeers[nID]
+
+		var score float64
+		switch {
+
+		// If the node is among or existing channel peers we give it a
+		// score of 0, as we don't need another channel.
+		case ok:
+			score = 0
```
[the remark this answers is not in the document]
halseth: Done!

PR #2007, opened 2018-10-04
Server: make maximum backoff configurable

icota:
In case of spotty wifi coverage a mobile device will rapidly switch from wifi to 4G and vice versa. In these cases the persistent peer will disconnect and the backoff period will increase exponentially. After a couple reconnects we are talking about minutes which makes for bad user experience.

### DIFF
--- a/config.go
+++ b/config.go
@@ -53,6 +53,7 @@ const (
 	defaultInactiveChanTimeout = 20 * time.Minute
 	defaultMaxLogFiles         = 3
 	defaultMaxLogFileSize      = 10
+	defaultMaxBackoff          = time.Hour
 
 	defaultTorSOCKSPort            = 9050
 	defaultTorDNSHost              = "soa.nodes.lightning.directory"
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -51,6 +51,7 @@ const (
 	defaultInactiveChanTimeout = 20 * time.Minute
```
ccdle12: The travis tests are failing due to `config.go` not being correctly formatted. 

I think you'll have to run `go fmt`

Roasbeef:
Needs a rebase!

icota:
Rebased.

server.go
```diff
@@ -2853,6 +2849,7 @@ func parseHexColor(colorStr string) (color.RGBA, error) {
 func computeNextBackoff(currBackoff time.Duration) time.Duration {
 	// Double the current backoff, truncating if it exceeds our maximum.
 	nextBackoff := 2 * currBackoff
+	maximumBackoff := cfg.MaxBackoff
```
wpaulino: Nit: no need to redefine this here, can just check `nextBackoff > cfg.MaxBackoff` below.

icota: @wpaulino You're right, fixed.

wpaulino:
Could you squash the commits into one as well?

PR #2012, opened 2018-10-06
rpc+cmd/lncli: add LookupPaymentStatus RPC

wpaulino:
In this PR, we add a new RPC, LookupPaymentStatus, that'll allow us to look up the status of an outgoing payment.

As is, when using the `lookuppaymentstatus` command, the status will be returned as one of the following:
* `GROUNDED`
* `IN_FLIGHT`
* `COMPLETED`

Perhaps these should be parsed to provide a more user-friendly description? I can foresee some users wondering what `GROUNDED` means.

Regarding integration tests - we usually add one to test some of the RPC logic, although I don't think we'll gain anything in this case due to the payment status logic already being tested in other parts of the codebase.


ccdle12:
Hi @wpaulino,

I cloned the branch and ran the commits in simnet.

I was receiving this error: 
`[lncli] rpc error: code = Unknown desc = /lnrpc.Lightning/LookupPaymentStatus: unknown permissions required for method`

I added the below code to the permissions maps RPC calls in `rpcserver.go` and it now works for me, not sure if this was a future commit you planning to implement :)

```
"/lnrpc.Lightning/LookupPaymentStatus": {{
    Entity: "info",
    Action: "read",
}},
```

In terms of `GROUNDED`, it does seem a bit confusing at first. It makes sense in the context of comparing it to `IN_FLIGHT` and `COMPLETED`. I think a parsed response with an explanation would generally be more helpful to the user.

Not sure the cleanest way to implement the parsed status description. I did experiment with a `message ParsedPaymentStatus { string payment_status }`, and then return that in `LookupPaymentStatus()`,  check the Enum type and assign a string message to the `lnrpc.ParsedPaymentStatus`.

rpcserver.go
```diff
@@ -3709,6 +3709,56 @@ func marshallTopologyChange(topChange *routing.TopologyChange) *lnrpc.GraphTopol
...
+		}
+	} else {
+		rHash = req.RHash
+	}
+
+	// Ensure that the payment hash is *exactly* 32-bytes.
+	if len(rHash) != 32 {
+		return nil, fmt.Errorf("payment hash must be exactly 32 "+
+			"bytes, is instead %v", len(rHash))
+	}
```
joostjager: Code duplication from LookupInvoice

rpcserver.go
```diff
@@ -3709,6 +3709,56 @@ func marshallTopologyChange(topChange *routing.TopologyChange) *lnrpc.GraphTopol
...
+	var status lnrpc.PaymentStatus_Status
+	switch paymentStatus {
+	case channeldb.StatusGrounded:
+		status = lnrpc.PaymentStatus_GROUNDED
+	case channeldb.StatusInFlight:
+		status = lnrpc.PaymentStatus_IN_FLIGHT
+	case channeldb.StatusCompleted:
+		status = lnrpc.PaymentStatus_COMPLETED
+	default:
+		return nil, fmt.Errorf("unknown payment status %v",
```
joostjager: Unknown payment (hash) instead of unknown payment status?

joostjager:
One thing to add to this: `SendPayment` and `SendPaymentSync` are also in a sense returning streaming info on a payment/payments. In the case of `SendPaymentSync` it is just a single update that is pushed through the channel, but it definitely isn't polling. 

It may be worth considering to align those streams with the stream returned from `SubscribePaymentStatus`. Keep the same format / updates.

joostjager:
WIth regards to the terms, I prefer FAILED over GROUNDED, although I do like the airplane analogy.

Roasbeef:
Perhaps we also need to insert an "on going" status? Which would indicate that the router is still actively trying to dispatch the payment?

joostjager:
> > The same functionality can be had by calling SubscribePaymentStatus, receiving the first update (actual status) and closing the stream.
> 
> In this case, do you keep reading the messages until success? I guess how does one know when to stop reading the responses in the case that the payment itself is actually still in flight?

The implementation of `SubscribePaymentStatus` is such that it immediately pushes the current status into the stream. So the reader can immediately receive this and close the stream if it isn't interested in further updates and needs just a snapshot.

joostjager:
>Perhaps we also need to insert an "on going" status? Which would indicate that the router is still actively trying to dispatch the payment?

I was thinking to repurpose the ON_FLIGHT state on the _payment_  level (opposed to _payment attempt_ level).

treygriffith:
As a potential user of an rpc like this, I'm interested in @joostjager's idea of a payment status that has the same signature as `SendPayment` as it allows me to keep mostly the same handling on my end and allows me to stream *or* query the state.

It would also allow me to use the same endpoint as a way to get the preimage for a single payment without querying for all payments with `ListPayments`.

The state of the payment could be implicit based on the behavior of the stream (e.g. failure with `NOT_FOUND` is a non-existent payment, open stream with no messages is an in-flight payment, `payment_error` is a failed payment, `payment_preimage` is a successful payment) or the stream could add an explicit status message that is sent on stream open and on updates. In the latter, explicit case, it would be nice if the same status was applied in `SendPayment`.

Is there interest in approaching payment status this way? It might be something I can pick up.

Roasbeef:
@treygriffith does #2183 address your needs? Mainly if you do `SendPayment` again with a payment hash after we've already settled it, you'll get the pre-image again. If you do it while in progress, it'll block until the payment is completed. If you do it after a payment fails, you'll get the error.

treygriffith:
@Roasbeef It's really close, but there are times when I would want to get that same feedback but without sending a payment if I haven't yet. But it seems like if that issue is resolved it would greatly simplify adding another rpc that does the same thing but without dispatching a new payment.

Is #2183 a near-term priority?

halseth:
@treygriffith #2183 should be very near-term 👍

PR #2014, opened 2018-10-07
Re-create forwardings after externalIP change

frennkie:
I noticed that the `nat=true` feature works really well to detect when the external IP address of a NAT device has changed (in my case using UPnP). But this requires that `lnd` also takes over the creation of the port forwarding (e.g. `9735`). In my case after every IP change the port forwardings were gone. I assume my NAT device removes them when the line is disconnected and re-established. 

In `lnd` I found no code that would re-create the port forwardings in this case. Port forwardings are only setup in `newServer` at server start (`configurePortForwarding` in line 405).

This PR make sure that the forwardings are re-created after a change in the external IP is detected.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1145,6 +1145,21 @@ out:
 				continue
 			}
 
+			// Periodically renew the NAT port forwarding.
+			for _, port := range forwardedPorts {
+				err := s.natTraversal.AddPortMapping(port)
+				if err != nil {
+					srvrLog.Warnf("Unable to automatically "+
+						"re-create port forwarding using %s: %v",
+						s.natTraversal.Name(), err)
+				} else {
+					srvrLog.Debugf("Automatically re-created "+
+						"forwarding for port %d using %s to "+
+						"advertise external IP",
+						port, s.natTraversal.Name())
+				}
+			}
+
 			if ip.Equal(s.lastDetectedIP) {
 				continue
 			}
--- a/nat/pmp.go
+++ b/nat/pmp.go
@@ -65,10 +65,6 @@ func (p *PMP) AddPortMapping(port uint16) error {
 	p.forwardedPortsMtx.Lock()
 	defer p.forwardedPortsMtx.Unlock()
 
-	if _, exists := p.forwardedPorts[port]; exists {
-		return nil
-	}
-
 	_, err := p.client.AddPortMapping("tcp", int(port), int(port), 0)
 	if err != nil {
 		return err
--- a/nat/upnp.go
+++ b/nat/upnp.go
@@ -62,10 +62,6 @@ func (u *UPnP) AddPortMapping(port uint16) error {
 	u.forwardedPortsMtx.Lock()
 	defer u.forwardedPortsMtx.Unlock()
 
-	if _, exists := u.forwardedPorts[port]; exists {
-		return nil
-	}
-
 	if err := u.device.Forward(port, ""); err != nil {
 		return err
 	}


server.go
```diff
@@ -1133,6 +1133,29 @@ out:
...
 			srvrLog.Infof("Detected new external IP address %s", ip)
 
+			// After detecting the new external IP try to re-create the
+			// port forwardings for the used ports.
+			for _, port := range forwardedPorts {
+				// first remove a possibly still existing forward
+				if err := s.natTraversal.DeletePortMapping(port); err != nil {
+					srvrLog.Debugf("Unable to remove forward %d: %v", port, err)
+					// assume the port is no longer forwarded
+					s.natTraversal.DeleteForwardedPort(port)
```
cfromknecht: this seems like it should be done in the success case, no? how do we know the port mapping was actually removed if an error was returned?

frennkie:
@wpaulino I fully agree with you. I will try to find some time to work on adding the ` --nat-watch-only ` (or something like this) flag in a separate PR.

frennkie:
Regarding the port forwarding logic I think the code really doesn't need to be very complicated. I checked the RfCs and both `UPnP` and `NAT-PMP` do not mind if a port forwarding that is already present is requested again. 

Actually on the contrary: NAT-PMP requires port forwardings to have a lifetime set. This value can be 0 but the recommendation is 7200 seconds (2 hours) and the client is expected to renew it in time. https://tools.ietf.org/rfc/rfc6886.txt

frennkie:
I (significantly) simplified this PR. Now all this does is to re-create the port forwardings after a change of the external IP has been detected.

The `u.forwardedPorts[port]; exists` check assumes that it has knowledge about the actual status of the port forwarding on the NAT device. But (at least for me) this is not the case as all UPnP forwardings are removed when the external IP changes. Therefore I removed the check. Should the fowarding still be in place then re-createing it doesn't actually do any harm. This is even the recommended behavior described in the PCP RfC.

PR has been tested and is working on my node.

server.go
```diff
@@ -1133,6 +1133,19 @@ out:
...
 
+			// After detecting the new external IP re-create the port forwardings.
+			for _, port := range forwardedPorts {
+				if err := s.natTraversal.AddPortMapping(port); err != nil {
+					srvrLog.Errorf("Unable to automatically re-create port "+
+						"forwarding using %s: %v",
+						s.natTraversal.Name(), err)
+				} else {
+					srvrLog.Infof("Automatically re-created port forwarding "+
+						"using %s to advertise external IP",
```
wpaulino: We should log new external IP as well.

frennkie: That's already done in line 1134, no?

frennkie:
@wpaulino Thanks for your support. I now now squashed this PR as requested.

frennkie:
Damn... I applied this PR, compiled it and deployed it to my Lightning node on Friday. On Saturday morning it was ok, this morning it was not ok (no UPNP port forwardings on my router). I re-connected my internet line and `lnd` successfully re-created the forwards.

I checked the logs and there seems to be a *race condition*.

The normal flow is

1. ISP disconnects line
2. Router reconnects
3. Router cleans port forwardings
4. `lnd` notices the new IP and re-creates the port forwardings (this runs every 15 minutes)

According to the logs what happened this morning was:

1. ISP disconnects line
2. Router reconnects
3. `lnd` notices the new IP and re-creates the port forwardings (this runs every 15 minutes)
4. Router cleans port forwardings

My guess is that the Router either also has a scheduled job or that it waited with cleaning the port forwardings because there were still connections in a local NAT table.

I see two solutions:

* A. Always schedule a 2nd re-create task (may be 5-10 minutes) later when a new IP is detected.
* B. Always re-create the forwardings (even if the IP hasn't changed.

My suggestions would be **option B**. Because:

* Simple
* There is not much load/traffic (~40 packets with 4Kbytes in total)
* This also catches other situations when the NAT devices loses the port forwardings (but may still have the same external IP)
* This would comply with the NAT-PMP RFC6886

frennkie:
Is there anything I can or have to do to get this merged?

halseth:
@frennkie A bit short on review cycles atm!

frennkie:
@halseth @Roasbeef Anything I can do to help get this into master? I don't really enjoy building my own patched `lnd` binary everytime you do a new (beta) release.

server.go
```diff
@@ -1145,6 +1145,19 @@ out:
 				continue
 			}
 
+			// Periodically renew the NAT port forwarding
```
halseth: nit: missing period

server.go
```diff
@@ -1145,6 +1145,19 @@ out:
...
+			for _, port := range forwardedPorts {
+				if err := s.natTraversal.AddPortMapping(port); err != nil {
```
halseth: lines look a bit long. Can do 
```go
err := s.nat...
if err != nil {
```

server.go
```diff
@@ -1145,6 +1145,19 @@ out:
...
+					srvrLog.Errorf("Unable to automatically re-create port "+
+						"forwarding using %s: %v",
+						s.natTraversal.Name(), err)
+				} else {
+					srvrLog.Debugf("Automatically re-created port forwarding "+
```
halseth: Should log the port in the log

server.go
```diff
@@ -1145,6 +1145,19 @@ out:
 				continue
 			}
 
+			// Periodically renew the NAT port forwarding
+			for _, port := range forwardedPorts {
+				if err := s.natTraversal.AddPortMapping(port); err != nil {
+					srvrLog.Errorf("Unable to automatically re-create port "+
```
halseth: make `warn`

frennkie:
@halseth I made the requested changes. Let me know when/if I should do another squash.

PR #2017, opened 2018-10-08
peer: only send send enable update for disabled chans

halseth:
To avoid spamming on each connection.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -491,6 +491,12 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
 
 		// To ensure we can route through this channel now that the peer
 		// is back online, we'll attempt to send an update to enable it.
+		// If the channel is already enabled in the database, we won't
+		// send an update, to avoid spamming unnecessarily.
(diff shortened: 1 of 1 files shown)


peer.go
```diff
@@ -491,6 +491,12 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
 
 		// To ensure we can route through this channel now that the peer
 		// is back online, we'll attempt to send an update to enable it.
+		// If the channel is already enabled in the database, we won't
```
Roasbeef: Won't this already be done by the logic further into this execution pipeline? In that it will only send out a new update if this is a shift from the state of the past 20 mins?

halseth: True, if we have already sent a disable+enable for this channel, the re-enabling will be ignored.

That means that this will only avoid the case where we first connect to a peer after startup, and send out a channel update for all our N channels (if all our peers are online). Maybe still good for the health of the graph not to send out N new updates after each restart?

halseth:
Might not be needed after strict channel enabling and related PRs. Making P4 for now.

halseth:
Replaced by #2080

PR #2018, opened 2018-10-08
Mark channels borked in case of remote error

halseth:
According to BOLT#1, a when receiving an error from the remote, we MUST fail the targeted channel: https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#the-error-message

This PR implements this behavior by marking the channel `Borked` in cases like this. We don't automatically force close, to avoid the remote tricking us into force closing.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -1900,9 +1900,14 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 		// Error received from remote, MUST fail channel, but should
 		// only print the contents of the error message if all
 		// characters are printable ASCII.
-		l.fail(LinkFailureError{code: ErrRemoteError},
+		l.fail(
+			LinkFailureError{
+				code:             ErrRemoteError,
+				PermanentFailure: true,
+			},
 			"ChannelPoint(%v): received error from peer: %v",
-			l.channel.ChannelPoint(), msg.Error())
+			l.channel.ChannelPoint(), msg.Error(),
+		)
 	default:
 		l.log.Warnf("received unknown message of type %T", msg)
 	}
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -5129,6 +5129,11 @@ func TestChannelLinkFail(t *testing.T) {
 		// force close the channel in response to the actions performed
 		// during the linkTest.
 		shouldForceClose bool
+
+		// permanentFailure indicates whether we expect the link to
+		// consider the failure permanent in response to the actions
+		// performed during the linkTest.
+		permanentFailure bool
 	}{
 		{
 			// Test that we don't force close if syncing states
@@ -5143,6 +5148,7 @@ func TestChannelLinkFail(t *testing.T) {
 			func(t *testing.T, c *channelLink, _ *lnwallet.LightningChannel) {
 				// Should fail at startup.
 			},
+			false,
 			false,
 		},
 		{
@@ -5160,6 +5166,7 @@ func TestChannelLinkFail(t *testing.T) {
 				// Should fail at startup.
 			},
 			false,
+			false,
 		},
 		{
 			// Test that we force close the channel if we receive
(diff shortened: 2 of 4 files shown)


peer.go
```diff
@@ -1920,6 +1920,14 @@ type linkFailureReport struct {
 // force closing the channel depending on severity, and sending the error
 // message back to the remote party.
 func (p *peer) handleLinkFailure(failure linkFailureReport) {
+
```
wpaulino: Nit: extra newline.

peer.go
```diff
@@ -1952,6 +1960,13 @@ func (p *peer) handleLinkFailure(failure linkFailureReport) {
 		}
 	}
 
+	// If this is a permanent failure, we will mark the channel borked.
```
wpaulino: Should we add a sanity check to make sure both `ForceClose` and `PermanentFailure` are not set?

peer.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2018/commits/3dad2d4413702011bf34e01aa6ffa113d1675e32

peer.go
halseth: Not sure if they will always be mutually exclusive? 

Added a logging statement: https://github.com/lightningnetwork/lnd/pull/2018/commits/03f38071e6a68dd98def37746983173f0cbb7260

halseth:
PTAL @wpaulino

peer.go
cfromknecht: in the process of force closing a channel, we will already mark it as borked. should we skip this if `ForceClose` is true as well?

peer.go
```diff
@@ -212,14 +212,6 @@ type peer struct {
 	// peer during the connection handshake.
 	remoteGlobalFeatures *lnwire.FeatureVector
 
-	// failedChannels is a set that tracks channels we consider `failed`.
-	// This is a temporary measure until we have implemented real failure
-	// handling at the link level, to handle the case where we reconnect to
-	// a peer and try to re-sync a failed channel, triggering a disconnect
-	// loop.
-	// TODO(halseth): remove when link failure is properly handled.
-	failedChannels map[lnwire.ChannelID]struct{}
```
cfromknecht: yay!! been wanting to delete this for a while :)

peer.go
```diff
@@ -2232,6 +2240,13 @@ func (p *peer) handleLinkFailure(failure linkFailureReport) {
 		}
 	}
 
+	// If this is a permanent failure, we will mark the channel borked.
+	if failure.linkErr.PermanentFailure && lnChan != nil {
```
carlaKC: When would `lnChan` be nil?

peer.go
```diff
@@ -2232,6 +2240,13 @@ func (p *peer) handleLinkFailure(failure linkFailureReport) {
...
+		if err := lnChan.State().MarkBorked(); err != nil {
+			peerLog.Errorf("Unable to mark channel borked: %v", err)
```
carlaKC: nit-est nit, but maybe log the channel ID as well?

Roasbeef:
Let's revive this for 0.11?

halseth:
@Roasbeef could def do 0.11, pretty small touch ups needed. One thing we should make sure is that we don't send error messages in cases we shouldn't, since after this PR there is no way returning to an operational channel.

cfromknecht:
@Roasbeef @halseth 0.12?

halseth:
> @Roasbeef @halseth 0.12?

I would be okay with that

halseth:
Could be useful for connection stability in case of force close between the nodes: https://github.com/lightningnetwork/lnd/issues/4613

Crypt-iQ:
Needs rebase

halseth:
Rebased!

peer.go
halseth: Just sanity check in case it is not in the map.

peer.go
halseth: Done!

carlaKC (approved):
LGTM!

cfromknecht (approved):
LGTM 🔥

PR #2022, opened 2018-10-09
htlcswitch: hodl invoice

joostjager:
In this PR, a new feature called "hold invoice" is added to `lnd`. A hold invoice is new type of invoice that triggers a different flow at the receiver end. 

Instead of immediately locking in and settling the htlc when the payment arrives, the htlc for a hold invoice is only locked in and not yet settled. At that point, it is not possible anymore for the sender to revoke the payment, but the receiver still can choose whether to settle or cancel the htlc and invoice.

From the sender perspective, a hold invoice payment request looks identical to a regular payment request. There is no way for the sender to know when a hold invoice is paid to.

Hold invoices enable several new use cases:

* Receiver gains the ability to check a condition before settling. If the condition is not met, the invoice can still be canceled and a potential refund does not need to be given. 

  _Example 1: web shop performing an inventory check. If the item is no longer in stock at the time the invoice is paid, the payment can be canceled._

  _Example 2: exchange accepting spontaneous LN deposits (building on top of #2455). The account to credit is specified as user data in the EOB and the exchange first checks whether this account is valid before settling the incoming htlc._

* A service can use a hold invoice as an insurance against bad behaviour by its users, similar to a fidelity bond. The user gains access to the service by paying the hold invoice. When the session is over, the service cancels the invoice and the user regains its funds. The only cost for the user has been the time value of those funds. When the user misbehaves during the session, the service can settle the hold invoice as a penalty. 

  _Example 3: prevention of dos attacks_

* A hold invoice may also be created if the preimage of the payment hash is not yet known to the receiver. Receiver can accept the payment and perform a follow up action to obtain the preimage. Only once the preimage is obtained, the receiver can settle the payment.

  _Example 4: pizza delivery service wanting to offer "no pizza no pay". A customer generates a random preimage and adds the hash of it to the pizza order. The delivery service creates a hold invoice tied to this hash and returns the invoice to the customer. Customer pays the invoice. At this point, the delivery service can accept the payment, but not settle it because it doesn't know the preimage yet. Then the driver goes out to deliver the pizza. At the door he asks the customer for the preimage, verifies that it is indeed the preimage of the hold invoice for this order and hands over the pizza. Verification can even be done offline. The delivery service can then settle the invoice when the driver returns with the preimage. Atomic pizza swap completed._

  _Example 5: parcel delivery service. Similar to pizza, but here the payer and recipient are different people. Payer pays the hold invoice of the delivery service and gives the preimage to the recipient of the parcel._ 

  _Example 6: taking this one step further in a three party example: a customer, a web shop and a delivery service. Customer generates preimage, customer sends hash of preimage to web shop, web shop creates hold invoice, customer pays hold invoice, web shop sends hash to delivery service, delivery service creates hold invoice for delivery cost, web shop pays delivery hold invoice, delivery service delivers to customer in exchange for preimage, delivery service settles delivery invoice with preimage, web shop settles customer invoice with preimage._

  [In this example there is a risk associated with reuse of the payment hash. Intermediate nodes on both paths may shortcut the preimage and disallow the webshop from receiving any money.]

  _Example 7: using hold invoice to execute a trustless atomic off-chain to on-chain swap. User A creates a secret preimage and passes the hash of it to user B. User B creates a hold invoice tied to this hash. User A pays the invoice. User B publishes an on-chain htlc tied to the same hash. When user A sweeps the on-chain htlc with the preimage it knows, user B learns it too and can pull the off-chain invoice._

### Usage

To enable hold invoice functionality, `lnd` needs to be built with the invoices sub server enabled:

`make tags="invoicesrpc"`

#### Creation of a hold invoice

To add a holdinvoice, invoke:

` lncli addholdinvoice <hash> --amt <amt>`

#### Waiting for acceptance

Typically an application wants to receive a notification when the htlc is accepted. This event isn't notified through the general `SubscribeInvoices` rpc. The reason for this is not to break backwards compatibility and prevent difficulties in the interaction with the add and settle indices of that call.Instead, the new `SubscribeSingleInvoice` rpc can be used. 

`SubscribeSingleInvoice` is not available through `lncli`. For testing, the invoice state can be observed through `lncli lookupinvoice <hash>`.

#### Settling or canceling

Once the htlc has been accepted in `lnd`, it is held there until either a cancel or settle rpc is received. Both can be issued through `lncli`:

`lncli cancelinvoice <hash>`

or

`lncli settleinvoice <preimage>`

### Todo
This PR is still work in progress:

* Automatic invoice cancelation when the incoming htlc is about to time out. Canceling the invoice fails back the htlc. If the invoice wouldn't be canceled, the htlc remains pending and the channel will be force closed on chain. Typical use of hodl invoices involves an application that is monitoring this timeout as well. That application would have either canceled or settled the invoice already before automatic cancelation is triggered. Automatic cancelation can be seen as a safety net to prevent channel closure.

* Automatic invoice cancelation when the invoice expires (invoice `expiry` property). This does not only apply to hodl invoices, but to regular invoices as well. Currently `expiry` is only checked by the sender before sending the payment and can be worked around. Automatic cancelation would prevent any payment of the invoice from being accepted by the receiver, thereby preventing potential refund flows.

* Handling of multiple htlcs paying to the same accepted invoice. Currently `amtPaid` does not reflect the total amount accepted. This is true for regular invoices as well. Calls for a higher level plan of when and when not to accept/settle htlcs.

* Support the accepted invoice state in contract resolvers. Related to #2075.

* Make the settle/fail resolution path through `htlcswitch` safe. There currently exists a gap where htlc and invoice administration may get out of sync.


htlcswitch/iterator.go
```diff
@@ -40,11 +40,14 @@ func (c NetworkHop) String() string {
...
-	exitHop lnwire.ShortChannelID
+	//
+	// We must make sure that these special channel ids are unique, because
+	// they are mainly compared by value.
+	exitHop = lnwire.ShortChannelID{TxPosition: 0}
 
 	// sourceHop is a sentinel value denoting that an incoming HTLC is
 	// initiated by our own switch.
-	sourceHop lnwire.ShortChannelID
+	sourceHop = lnwire.ShortChannelID{TxPosition: 1}
```
cfromknecht: this value cannot change, it would break just about everything in the switch since this is compared to persisted entries in the forwarding packages.

htlcswitch/link.go
```diff
@@ -2437,7 +2485,46 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
...
 			}
 
+			// We don't have the pre-image yet, but will hold onto
+			// the HTLC as forward it to the switch as if there're
+			// additional hops. We use a special delayed resolution
+			// channelID, and then use the HTLC ID of the add index
+			// for the original invoice.
 			preimage := invoice.Terms.PaymentPreimage
+			if !bytes.Equal(preimage[:], unknownPreimage[:]) {
+				err = l.cfg.Registry.AcceptInvoice(
```
joostjager: @cfromknecht, I worked your PoC into this PR. You wrote:

>one thing we may consider is logically grouping`switchSettledPackets` from switchPackets. both can be forwarded in a single call to forwardBatch, but we should post-process accepting the switchSettledPackets after they have been forwarded. rationale here is that AcceptInvoice will fire a notification, which if instantly acted upon, could result in the circuit not being opened when the switch tries to lookup the circuit
delaying until after the circuits have been atomically opened and committed will ensure this case doesn't arise

Is this still how we want to do this? Or is it better to make the `AcceptInvoice` call inside the switch?

joostjager: And do we need a `if fwdPkg.State == channeldb.FwdStateLockedIn` check here?

joostjager:
@cfromknecht I park the link/switch comments for now. First define the rpc interface for this in #2263, implement that (new invoice states and events) and then continue with the switch/link modifications.

invoiceregistry.go
```diff
@@ -397,6 +462,11 @@ type invoiceSubscription struct {
 	// StartingInvoiceIndex field.
 	NewInvoices chan *channeldb.Invoice
 
+	// AcceptedInvoices is a channel that we'll use to send all newly created
+	// invoices with an invoice index greater than the specified
+	// StartingInvoiceIndex field.
+	AcceptedInvoices chan *channeldb.Invoice
```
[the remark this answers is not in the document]
joostjager: No need anymore for distinct channels for new states with introduction of `SubscribleSingeInvoice`

rpcserver.go
```diff
@@ -2807,13 +2807,17 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+	// If neither a preimage or hash was specified, then we'll generate a
+	// new preimage from fresh cryptographic randomness.
+	case len(invoice.RPreimage) == 0 && len(invoice.RHash) == 0:
 		if _, err := rand.Read(paymentPreimage[:]); err != nil {
 			return nil, err
 		}
 
+	// If only a hash is specified, preimage can remain empty and will be
+	// supplied from an external source later.
+	case len(invoice.RPreimage) == 0 && len(invoice.RHash) > 0:
```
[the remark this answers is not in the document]
joostjager: This is now solved with the introduction of `lnhash.NewHash` (similar to `chainhash.Hash`, but without the reversed `String()`)

rpcserver.go
```diff
@@ -2826,6 +2830,16 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
 	}
 
+	// Next, generate the payment hash itself from the preimage if
+	// necessary. This will be used by clients to query for the state of a
+	// particular invoice.
+	var rHash [32]byte
+	if len(invoice.RHash) == 0 {
+		rHash = sha256.Sum256(paymentPreimage[:])
+	} else {
+		copy(rHash[:], invoice.RHash)
```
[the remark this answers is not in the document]
joostjager: Current state of the pr is that `AddInvoice` only looks at preimage and `AddHoldInvoice` only at the hash field.

channeldb/invoices.go
```diff
@@ -123,6 +147,9 @@ type Invoice struct {
 	// SettleDate is the exact time the invoice was settled.
 	SettleDate time.Time
 
+	// SettleDate is the exact time the invoice was settled.
```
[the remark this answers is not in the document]
joostjager: `AcceptDate` is history, as we want to avoid db migration for now.

channeldb/invoices.go
```diff
@@ -1101,3 +1141,34 @@ func acceptInvoice(invoices, acceptedIndex *bbolt.Bucket, invoiceNum []byte,
...
+	}
+
+	invoice.Terms.State = ContractRemoved
+
+	var buf bytes.Buffer
+	if err := serializeInvoice(&buf, &invoice); err != nil {
+		return err
+	}
+
+	if err := invoices.Put(invoiceNum[:], buf.Bytes()); err != nil {
```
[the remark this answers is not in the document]
joostjager: Combining with the `return` statement would still return the `nextAddSeqNo` in the error case, even though it isn't committed to the db. I think it is better to keep it deterministic by returning 0.

channeldb/invoices.go
```diff
@@ -95,6 +95,9 @@ const (
 
 	// ContractAccepted means the HTLC has been accepted but not settled yet.
 	ContractAccepted = 2
+
+	// ContractRemoved means the invoice has been removed.
+	ContractRemoved = 3
```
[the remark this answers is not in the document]
joostjager: Last comment is still applicable to the pr as it is atm. Invoices are never removed, only marked `Canceled`.

channeldb/invoices.go
```diff
@@ -1101,3 +1141,34 @@ func acceptInvoice(invoices, acceptedIndex *bbolt.Bucket, invoiceNum []byte,
...
+	// Add idempotency to duplicate removed state, return here to avoid
+	// overwriting the previous info.
+	switch invoice.Terms.State {
+	case ContractRemoved:
+		return nil
+	case ContractSettled:
+		return fmt.Errorf("invoice already settled")
+	}
+
+	invoice.Terms.State = ContractRemoved
```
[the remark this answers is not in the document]
joostjager: Should be implemented in the current as it is right now and (some of) these cases are covered in the hold invoice test in `invoiceregistry_test.go` too.

htlcswitch/iterator.go
joostjager: This has long been reverted.

ZapUser77:
Will this lock up funds, prevent them from being transferred in either direction, of each intermediary hop for each hold?  This could cause serious problems, if hold transfers were used as "bonds" and frozen for days.  Is there a method for hubs to prevent their node from relaying "hold invoices"?

joostjager:
Yes, it will lock up funds. There is no way to recognize a "bond" htlc for the intermediate nodes, but what they could do is lower the max timelock that they will except. #2051 introduces this limit at 5000 blocks. 

It is an interesting question how LN will develop if hold invoices become common.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -0,0 +1,297 @@
...
+		},
+		cli.StringFlag{
+			Name: "fallback_addr",
+			Usage: "fallback on-chain address that can be used in " +
+				"case the lightning payment fails",
+		},
+		cli.Int64Flag{
+			Name: "expiry",
+			Usage: "the invoice's expiry time in seconds. If not " +
+				"specified an expiry of 3600 seconds (1 hour) " +
```
ZapUser77: Needs a comma
"specified, an expiry of 3600 seconds (1 hour) " +

ZapUser77:
Currently, "Max timelock" is a constant, that isn't adjustable, correct?
With 5,000, funds could be locked for over a month, unable to be spent, unable to be used for routing any other payment, correct? Or are they still usable until the invoice is accepted, and just breaks if the path found no longer has liquidity to service the payment?
Either way, couldn't they create this invoice... get the goods/service, then one of the nodes along the path go down, making acceptance impossible, and when the HTLC runs out get their money back?  [Sounds like CC charge backs to me.]
Have you considered all the different ways a 'hold invoice' could be used as an attack vector?  Either drying liquidity of multiple nodes (asymmetrically 1 bitcoin locked by attacker locks up to 20 nodes @ 1 BTC each), or preventing the HTLC from executing after getting goods and services?

I may misunderstand how this works, but it's food for thought.  ^_^

joostjager:
>Currently, "Max timelock" is a constant, that isn't adjustable, correct?

Yes the constant isn't adjustable at the moment. It would require surfacing it as a config settings. But decreasing it to a low number will also prevent you from forwarding payments at the start of a long chain. Those htlcs will have long cltv locks too.

>With 5,000, funds could be locked for over a month

Correct and not usable for anything else during that time.

>Either way, couldn't they create this invoice... get the goods/service, then one of the nodes along the path go down, making acceptance impossible, and when the HTLC runs out get their money back?

No, once the final node accepts the htlc (and this happens before delivering the goods/service), there is no way an intermediate node can prevent the final node from settling.

>Have you considered all the different ways a 'hold invoice' could be used as an attack vector?

Any such attack vector would exist already now. Hold invoice doesn't change the protocol and the sender and intermediates nodes aren't changed either.

A limitation of LN as it is today, is that the forwarding fee is not dependent on the requested timelock and also not on the actual timelock. Making it dependent on the requested time lock could be implemented as an extension to the fee function.

joostjager:
Experimented with refactoring hold invoice into the 'exit link' concept.

https://github.com/lightningnetwork/lnd/compare/master...joostjager:exithop

Doing this, I realized that the idea of an 'exit link' that settles invoices may not be so good. initially i thought, it's nice, because there we can also do things like aggregating multiple incoming htlcs for amp. But the fundamental difference with the exit link vs a normal outgoing link, is that in case of the exit link it is tightly coupled to the incoming link. only if the incoming links settles the htlc, the invoice is settled. 

We can't just settle the invoice in the exit link and send an htlc_fulfill back to the incoming link. because if something goes wrong there, the invoice shouldn't have been marked as settled. this is different from normal links, because they are in the end fully independent.

So what you'd get is:
- incoming links forwards htlc to exit link
- exit link decides the htlc is right for the invoice
- exit link signals back fulfill message to incoming link
- incoming link settles htlc
but then it is not done yet:
- incoming link somehow signals exit link again that htlc is settled
- exit link marks invoice as settled

Too much ping pong

Roasbeef:
With cancel in, this can now be rebased! Still under discussion if we'll go with this route, or the exit link route. By rebasing it'll at least be easier to examine the implementation details in this PR.

joostjager:
Zoomed in further on the exit link design: by detaching invoice settling from the invoice link, a risk is created that the invoice is settled, but the incoming htlc not. This risk didn't exist when all has handled in the link, because the link would just retry on startup. 

Isolating invoice settling to the exit link, requires the exit link to resend a resolution message to the incoming link if necessary. It would need a SettleAck filter similar to regular outgoing links. This filter would be set when the incoming link has settled the htlc. As long as the filter isn't set, the exit link would resend the resolution message.

Also, a variation on `TrimOpenCircuits` would be needed that looks at the invoice database to decide when to trim the circuit and trigger reprocessing.

joostjager:
Push redesigned non-persistent holdinvoice implementation.

@Roasbeef @cfromknecht ptal

htlcswitch/link_test.go
```diff
@@ -1066,22 +1066,15 @@ func TestChannelLinkMultiHopUnknownPaymentHash(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Generate payment: invoice and htlc.
-	invoice, htlc, err := generatePayment(amount, htlcAmt, totalTimelock,
+	// Generate payment invoice and htlc, but don't add this invoice to the
+	// receiver registry. This should trigger an unknown payment hash
+	// failure.
+	_, htlc, err := generatePayment(amount, htlcAmt, totalTimelock,
```
halseth: why is this change needed really?

htlcswitch/mock.go
```diff
@@ -699,82 +701,83 @@ func (f *mockChannelLink) UpdateShortChanID() (lnwire.ShortChannelID, error) {
...
 }
 
-func (i *mockInvoiceRegistry) SettleInvoice(rhash lntypes.Hash,
-	amt lnwire.MilliSatoshi) error {
+type mockInvoiceRegistry struct {
+	settleChan chan lntypes.Hash
 
-	i.Lock()
-	defer i.Unlock()
+	registry *invoices.InvoiceRegistry
```
halseth: it the mock just wraps a real registry, why not just use the registry directly?

htlcswitch/link.go
```diff
@@ -1064,6 +1071,51 @@ out:
 	}
 }
 
+// processHodlEvent applies a received hodl event to the provided htlc. When
+// this function returns without an error, the commit tx should be updated.
+func (l *channelLink) processHodlEvent(htlcs []hodlHtlc,
+	hodlEvent invoices.HodlEvent) error {
+
+	hash := hodlEvent.Hash
+	if hodlEvent.Preimage == nil {
```
halseth: would it be more clear to actually define a event type field?

htlcswitch/mock.go
joostjager: The only thing it really adds is the settle channel, to be able to intercept this event.

htlcswitch/link_test.go
joostjager: Because the invoice isn't modified anymore in the way it was before. See also commit message.

htlcswitch/link.go
```diff
@@ -1064,6 +1071,51 @@ out:
 	}
 }
 
+// processHodlEvent applies a received hodl event to the provided htlc. When
+// this function returns without an error, the commit tx should be updated.
+func (l *channelLink) processHodlEvent(htlcs []hodlHtlc,
```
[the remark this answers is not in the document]
joostjager: good suggestion, fixed.

htlcswitch/link.go
joostjager: I think I am fine with this. `Hash` is already redundant if `Preimage` is set, and this would add another redundant field.

invoices/invoiceregistry.go
```diff
@@ -438,45 +446,54 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
-	// Implement idempotency by returning success if the invoice was already
-	// settled.
-	if err == channeldb.ErrInvoiceAlreadySettled {
-		log.Debugf("Invoice %v already settled", rHash)
-		return nil
-	}
+	// If invoice is already settled, settle htlc. This means we accept more
+	// payments to the same invoice hash.
+	case channeldb.ErrInvoiceAlreadySettled:
+		return createEvent(&invoice.Terms.PaymentPreimage), nil
```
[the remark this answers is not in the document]
joostjager: I think this will be fixed by `settle_index`? That is incremented in the database and clients will request from the point they know of.

joostjager:
@halseth @cfromknecht comments processed. ptal

invoices/invoiceregistry.go
```diff
@@ -62,6 +62,14 @@ type InvoiceRegistry struct {
 	// value from the payment request.
 	decodeFinalCltvExpiry func(invoice string) (uint32, error)
 
+	// subscriptions is a map from a payment hash to a list of subscribers.
+	// It is used for efficient notification of links.
+	hodlSubscriptions map[lntypes.Hash]map[chan<- interface{}]struct{}
+
+	// reverseSubscriptions tracks hashes subscribed to per subscriber. This
+	// is used to unsubscribe from all hashes efficiently.
+	hodlReverseSubscriptions map[chan<- interface{}]map[lntypes.Hash]struct{}
```
Roasbeef: Using a map as a key here makes be uneasy. Above as well, can we instead use a monotonic client counter?

Roasbeef: Also why `interface{}` if we already know the type? We may get one from the queue itself, but doesn't mean we need to sacrifice type safety everywhere.

invoices/invoiceregistry.go
```diff
@@ -449,8 +461,17 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+// function moves the invoice to the accepted state. When SettleHoldInvoice is
+// called later, a resolution message will be send back to the caller via the
+// provided hodlChan. Invoice registry sends on this channel what action needs
+// to be taken on the htlc (settle or cancel). The caller needs to ensure that
+// the channel is either buffered or received on from another goroutine to
+// prevent deadlock.
 func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
-	amtPaid lnwire.MilliSatoshi) (*HodlEvent, error) {
+	amtPaid lnwire.MilliSatoshi, hodlChan chan<- interface{}) (
+	*HodlEvent, error) {
```
Roasbeef: The notification channel can be returned _within_ the `HodlEvent` response. The registry can handle buffering it properly, and the client then checks directly or indirectly (via the method) if there's a delayed response/future.

invoices/invoiceregistry.go
```diff
@@ -770,3 +833,58 @@ func (i *InvoiceRegistry) SubscribeSingleInvoice(
 
 	return client
 }
+
+func (i *InvoiceRegistry) notifyHodlSubscribers(hodlEvent HodlEvent) {
```
Roasbeef: Missing `godoc` comment.

htlcswitch/link_test.go
```diff
@@ -5586,3 +5586,190 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
 		t.Fatalf("expected unknown payment hash, but got %v", err)
 	}
 }
+
+func TestChannelLinkHoldInvoiceSettle(t *testing.T) {
```
Roasbeef: Missing `godoc` comment.

htlcswitch/test_utils.go
```diff
@@ -755,6 +767,51 @@ func preparePayment(sendingPeer, receivingPeer lnpeer.Peer,
...
+	// Generate payment: invoice and htlc.
+	invoice, htlc, err := generatePaymentWithPreimage(invoiceAmt, htlcAmt, timelock, blob,
+		channeldb.UnknownPreimage, rhash)
+	if err != nil {
+		paymentErr <- err
+		return paymentErr
+	}
+
+	// Check who is last in the route and add invoice to server registry.
+	if err := receiver.registry.AddInvoice(*invoice, rhash); err != nil {
```
Roasbeef: Won't need to handle interactions like this, if we mock out the registry fully.

htlcswitch/link_test.go
```diff
@@ -5586,3 +5586,190 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
...
+
+	channels, cleanUp, _, err := createClusterChannels(
+		btcutil.SatoshiPerBitcoin*3,
+		btcutil.SatoshiPerBitcoin*5)
+	if err != nil {
+		t.Fatalf("unable to create channel: %v", err)
+	}
+	defer cleanUp()
+
+	n := newThreeHopNetwork(t, channels.aliceToBob, channels.bobToAlice,
```
Roasbeef: Folding style violation.

htlcswitch/link_test.go
```diff
@@ -5586,3 +5586,190 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
 		t.Fatalf("expected unknown payment hash, but got %v", err)
 	}
 }
+
+func TestChannelLinkHoldInvoiceSettle(t *testing.T) {
+	t.Parallel()
+
+	channels, cleanUp, _, err := createClusterChannels(
```
Roasbeef: Folding style violation.

htlcswitch/link_test.go
Roasbeef: Few others in this test as well, won't point out the rest to reduce spam.

htlcswitch/link_test.go
Roasbeef: Also on the test below.

htlcswitch/link_test.go
Roasbeef: There's also a large portion of overlap between the tests that can likely be consolidated.

htlcswitch/link.go
```diff
@@ -2565,16 +2569,6 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		return true, nil
 	}
 
-	// Reject htlcs for canceled invoices.
```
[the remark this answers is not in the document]
joostjager: More details added to commit message

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -23,10 +23,66 @@ service Invoices {
...
+    rpc AddHoldInvoice(AddHoldInvoiceRequest) returns (AddHoldInvoiceResp);
 }
 
 message CancelInvoiceMsg {
     /// Hash corresponding to the invoice to cancel.
     bytes payment_hash = 1;
 } 
 message CancelInvoiceResp {}
+
+message AddHoldInvoiceRequest {
```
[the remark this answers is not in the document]
joostjager: I specifically created a new message because exposing unused fields in an interface leads to confusion. I also got this feedback from a user.

htlcswitch/mock.go
```diff
@@ -699,82 +701,83 @@ func (f *mockChannelLink) UpdateShortChanID() (lnwire.ShortChannelID, error) {
...
-	if invoice.Terms.State == channeldb.ContractSettled {
-		return nil
+	decodeExpiry := func(invoice string) (uint32, error) {
+		return 3, nil
 	}
 
-	invoice.Terms.State = channeldb.ContractSettled
-	invoice.AmtPaid = amt
-	i.invoices[rhash] = invoice
+	registry := invoices.NewRegistry(cdb, decodeExpiry)
```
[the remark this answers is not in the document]
joostjager: The reason that a real instance is used here, is to be able to test the behavior of the link in combination with the actual registry implementation. This could be mocked out, but the mock would need to be implemented and controlled carefully to make sure it does the same thing. I think using the actual implementation is a safer solution. If performance is a concern, I'd prefer mocking the db layer underneath the registry.

contractcourt/htlc_success_resolver.go
```diff
@@ -179,8 +179,10 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 		}
 
 		// With the HTLC claimed, we can attempt to settle its
-		// corresponding invoice if we were the original destination.
-		err = h.Registry.SettleInvoice(h.payHash, h.htlcAmt)
+		// corresponding invoice if we were the original destination. As
+		// the htlc is already settled at this point, we don't need to
+		// process the result.
+		_, err = h.Registry.NotifyExitHopHtlc(h.payHash, h.htlcAmt)
```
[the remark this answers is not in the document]
joostjager: Do you mean the naming? The point of renaming the method is to set the stage for removing invoice awareness from the links.

htlcswitch/link.go
```diff
@@ -2674,30 +2726,22 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
...
-		l.sendHTLCError(pd.HtlcIndex, failure, obfuscator, pd.SourceRef)
-
-		return true, nil
-	}
-
-	// Handle unexpected errors.
+	// Notify the invoiceRegistry of the exit hop htlc. If we crash right
+	// after this, this code will be re-executed after restart. We will
+	// receive back a resolution event.
+	event, err := l.cfg.Registry.NotifyExitHopHtlc(
```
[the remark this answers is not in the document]
joostjager: In this commit, hodl invoices don't exist yet and afaik there is no other outcome possible. Either the invoice exists and we settle the htlc or it doesn't exist and the htlc should be failed.

channeldb/invoices.go
```diff
@@ -611,11 +624,14 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
 	return resp, nil
 }
 
-// SettleInvoice attempts to mark an invoice corresponding to the passed
-// payment hash as fully settled. If an invoice matching the passed payment
-// hash doesn't existing within the database, then the action will fail with a
-// "not found" error.
-func (d *DB) SettleInvoice(paymentHash [32]byte,
+// AcceptOrSettleInvoice attempts to mark an invoice corresponding to the passed
```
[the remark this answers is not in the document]
joostjager: Where would it be useful now? I removed it because it would otherwise be dead code.

channeldb/invoices.go
```diff
@@ -932,40 +988,99 @@ func deserializeInvoice(r io.Reader) (Invoice, error) {
...
+	}
+
+	if err := invoices.Put(invoiceNum[:], buf.Bytes()); err != nil {
+		return nil, err
+	}
+
+	return &invoice, nil
+}
+
+func setSettleFields(settleIndex *bbolt.Bucket, invoiceNum []byte,
```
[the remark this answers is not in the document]
joostjager: `update` suggests that this method can be called a second time to update the fields to other values, but it can't. It is intended to set the fields once per invoice.

channeldb/invoices.go
```diff
@@ -932,40 +988,99 @@ func deserializeInvoice(r io.Reader) (Invoice, error) {
...
 	}
 
 	var seqNoBytes [8]byte
 	byteOrder.PutUint64(seqNoBytes[:], nextSettleSeqNo)
 	if err := settleIndex.Put(seqNoBytes[:], invoiceNum); err != nil {
-		return nil, err
+		return err
 	}
 
-	invoice.AmtPaid = amtPaid
```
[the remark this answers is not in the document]
joostjager: It is set in `acceptOrSettleInvoice`. So for hodl invoice, we set `AmtPaid` on acceptance.

channeldb/invoices.go
```diff
@@ -932,40 +988,99 @@ func deserializeInvoice(r io.Reader) (Invoice, error) {
...
+		return &invoice, ErrInvoiceAlreadyCanceled
+	case ContractSettled:
+		return &invoice, ErrInvoiceAlreadySettled
+	}
+
+	// If preimage is already set, return.
+	if invoice.Terms.PaymentPreimage != UnknownPreimage &&
+		invoice.Terms.PaymentPreimage != preimage {
+
+		return &invoice, errors.New("invalid preimage")
```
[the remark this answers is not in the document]
joostjager: When we get here, the invoice must be in the `Accepted` state. Otherwise the switch above would have returned an error. An invoice in the current implementation cannot be `Accepted` while having the preimage known. So this if-condition should never be true. 

Removed the check.

htlcswitch/link.go
```diff
@@ -1065,12 +1082,73 @@ out:
...
+		// resolution for a previously accepted htlc.
+		case hodlItem := <-l.hodlQueue.ChanOut():
+			hodlEvent := hodlItem.(invoices.HodlEvent)
+			err := l.processHodlQueue(hodlEvent)
+			if err != nil {
+				l.fail(LinkFailureError{code: ErrInternalError},
+					fmt.Sprintf("process hodl queue: %v",
+						err.Error()),
+				)
+				break out
```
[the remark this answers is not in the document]
joostjager: It depends on how strict we want to be about errors. As this code is only executed for hodl invoices and hodl invoices are newly introduced, I think we should be strict about it. Otherwise we risk both not noticing the error when it happens as well as having to deal with consequential damage from for example not being able to update the commitment tx inside `processHodlQueue`.

invoices/invoiceregistry.go
joostjager: I don't use a map as a key. I use a chan as a key.

Assuming you meant to say using a chan as a key makes you uneasy: why is that? I'd agree if the key was a struct. Those can be copied accidentally and thereby introduce bugs. A channel cannot be copied (?), so I think this is safe to do.

joostjager: W/r `interface{}`, how would that work if I'd put `chan <-HodlEvent` here? I cannot just connect channels of different types without another goroutine just for the conversion.

invoices/invoiceregistry.go
joostjager: The `hodlChan` argument is there, because I want all HodlEvents to be sent on the same channel. So that the link only needs to listen on that single channel in its main event loop. If every call to `NotifyExitHopHtlc` would yield a new channel, how to listen to them all in a clean way?

htlcswitch/link_test.go
joostjager: Folding styles fixed.
Overlap consolidated.
Changed use of three hop network to two hop network, as the third hop wasn't used in the tests.

htlcswitch/test_utils.go
joostjager: See comment above about testing link+registry as a single unit.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -112,33 +123,62 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 		blockEpochs.Cancel()
 	}()
 
+	// Create a buffered hodl chan to prevent deadlock.
+	hodlChan := make(chan interface{}, 1)
```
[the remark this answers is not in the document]
joostjager: See comment below about connecting to queue.

htlcswitch/interfaces.go
```diff
@@ -21,13 +21,20 @@ type InvoiceDatabase interface {
...
 
 	// CancelInvoice attempts to cancel the invoice corresponding to the
 	// passed payment hash.
 	CancelInvoice(payHash lntypes.Hash) error
+
+	// SettleHodlInvoice settles a hold invoice.
+	SettleHodlInvoice(preimage lntypes.Preimage) error
+
+	// HodlUnsubscribeAll unsubscribes from all hodl events.
+	HodlUnsubscribeAll(subscriber chan<- interface{})
```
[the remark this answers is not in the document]
joostjager: I am not completely sure if I get what you mean here. A link can have multiple outstanding hodl htlcs. Do you want to unsubscribe from them individually? Part of this is also a consequence of wanting to have a single channel per link for hodl events.

htlcswitch/link.go
```diff
@@ -2569,6 +2569,19 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		return true, nil
 	}
 
+	// Reject invoices with unknown preimages.
```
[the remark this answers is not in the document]
joostjager: Yes, that issue is fixed now. Verified by running both the old and the current hold invoice branch and checking the fwding logs. It was a side effect of settling invoices through the switch. Not having to fix that issue is another advantage of the current approach.

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -0,0 +1,364 @@
...
+
+// AddInvoiceConfig contains dependencies for invoice creation.
+type AddInvoiceConfig struct {
+	// InvoiceRegistry is a central registry of all the outstanding invoices
+	// created by the daemon.
+	InvoiceRegistry *invoices.InvoiceRegistry
+
+	// Switch is used by the invoices subserver to control acceptance and
+	// cancelation of invoices.
+	Switch *htlcswitch.Switch
```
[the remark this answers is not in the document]
joostjager: Changed to closures for `Switch` and `InvoiceRegistry`. `ChanDB` has two methods used, left that as is.

I think the better time to fix this is when a unit test is added for this code.

lnrpc/invoicesrpc/config_active.go
```diff
@@ -26,7 +30,26 @@ type Config struct {
 	// created by the daemon.
 	InvoiceRegistry *invoices.InvoiceRegistry
 
+	// Switch is used by the invoices subserver to control acceptance and
```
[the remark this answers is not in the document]
joostjager: Changed to closure

invoices/invoiceregistry.go
```diff
@@ -52,7 +50,9 @@ type InvoiceRegistry struct {
 	// that *all* nodes are able to fully settle.
 	debugInvoices map[lntypes.Hash]*channeldb.Invoice
 
-	activeNetParams *chaincfg.Params
+	// decodeFinalCltvExpiry is a function used to decode the final expiry
```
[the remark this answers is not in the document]
joostjager: See comment about wanting to test link+registry as a unit.

contractcourt/chain_arbitrator.go
```diff
@@ -137,10 +137,9 @@ type ChainArbitratorConfig struct {
 	// Sweeper allows resolvers to sweep their final outputs.
 	Sweeper *sweep.UtxoSweeper
 
-	// SettleInvoice attempts to settle an existing invoice on-chain with
-	// the given payment hash. ErrInvoiceNotFound is returned if an invoice
-	// is not found.
-	SettleInvoice func(lntypes.Hash, lnwire.MilliSatoshi) error
+	// Registry is the invoice database that is used by resolvers to lookup
+	// preimages and settle invoices.
+	Registry *invoices.InvoiceRegistry
```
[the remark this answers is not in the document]
joostjager: This interface already exists as `htlcswitch.InvoiceDatabase`. In a previous iteration, I used that in `chain_arbitrator.go`. But in order to do so, the interface needed to be moved outside of `htlcswitch` to prevent a circular dependency. There was some objection to that, so I took it out again.

Added the commit back in now (`invoices+htlcswitch: move registry interface`). I am in favor of creating the decoupling in that way. Let me know if you see it differently. If ok, I can squash the commit.

joostjager:
@Roasbeef changes made, comments replied to, ptal

contractcourt/chain_arbitrator.go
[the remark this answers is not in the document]
joostjager: Btw, we can't just have the `SettleInvoice` closure, because we use two methods from `InvoiceRegistry`.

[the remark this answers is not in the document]
Roasbeef: Will check out the commit, but an advantage of introducing a slimmer/similar interface is that the package expresses to outside packages its requirements, and doesn't need to import another package to express its interface requirements. 

Yeah, it would be more than just `SettleInvoice`: it would also be the additional methods it needs as well.

htlcswitch/mock.go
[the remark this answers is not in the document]
Roasbeef: Understood, thanks for the explanation.

contractcourt/htlc_success_resolver.go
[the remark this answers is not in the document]
Roasbeef: Commented obsoleted by later commits which added a new return value.

htlcswitch/interfaces.go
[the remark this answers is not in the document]
Roasbeef: I mean that it seems awkward that the "key" to a subscription be a channel rather than some internal integer.

invoices/invoiceregistry.go
Roasbeef: Ahh I see what you're getting at here: a single channel with multiple events sent over it, rather than needing to have a new channel for even hodl event.

htlcswitch/interfaces.go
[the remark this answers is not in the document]
joostjager: Yes, I agree. But decided for myself that awkward without a solid underlying reason probably isn't awkward then...

invoices/invoiceregistry.go
joostjager: Yes that indeed. Otherwise we'd get that reflection-based select over a list.

contractcourt/chain_arbitrator.go
```diff
@@ -139,7 +139,7 @@ type ChainArbitratorConfig struct {
 
 	// Registry is the invoice database that is used by resolvers to lookup
 	// preimages and settle invoices.
-	Registry *invoices.InvoiceRegistry
+	Registry invoices.InvoiceDatabase
```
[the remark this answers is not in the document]
joostjager: What it adds is that it removes the dependency on the concrete type, allowing a mock to be put in place. But removed the commit for now, can also do this when needed. Interested to see your example PR.

joostjager:
Rebased

joostjager:
@cfromknecht @halseth ptal

rpcserver.go
```diff
@@ -3277,309 +3276,29 @@ func (r *rpcServer) sendPaymentSync(ctx context.Context,
...
-					chanPoint, remotePub)
-				continue
-			}
-
-			// Fetch the policies for each end of the channel.
-			chanID := channel.ShortChanID().ToUint64()
-			info, p1, p2, err := graph.FetchChannelEdgesByID(chanID)
+	addInvoiceCfg := &invoicesrpc.AddInvoiceConfig{
+		AddInvoice: r.server.invoices.AddInvoice,
+		IsChannelActive: func(chanID lnwire.ChannelID) (bool, error) {
```
cfromknecht: is it needed that this return an error? there is an existing method that just returns a boolean value, could be
```golang
    IsChannelActive: r.server.htlcSwitch.HasActiveLink,
```

subrpcserver_config.go
```diff
@@ -155,9 +162,37 @@ func (s *subRPCServerConfigs) PopulateDependencies(cc *chainControl,
 			subCfgValue.FieldByName("InvoiceRegistry").Set(
 				reflect.ValueOf(invoiceRegistry),
 			)
+			subCfgValue.FieldByName("IsChannelActive").Set(
+				reflect.ValueOf(func(chanID lnwire.ChannelID) (
```
cfromknecht: would imagine this also becomes simpler, e.g.
```golang
reflect.ValueOf(htlcSwitch.HasActiveLink)
```

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -112,33 +123,62 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
+	// on-chain, so that we will get a hodl event when a corresponding hodl
+	// invoice is settled.
+	event, err := h.Registry.NotifyExitHopHtlc(h.payHash, h.htlcAmt, hodlChan)
+	if err != nil && err != channeldb.ErrInvoiceNotFound {
+		return nil, err
+	}
+	defer h.Registry.HodlUnsubscribeAll(hodlChan)
+
+	// If the htlc can be settled directly, we can progress to the inner
+	// resolved immediately.
```
cfromknecht: s/resolved/resolver

channeldb/invoices.go
```diff
@@ -650,6 +670,46 @@ func (d *DB) SettleInvoice(paymentHash [32]byte,
 	return settledInvoice, err
 }
 
+// SettleHoldInvoice sets the preimage of a hodl invoice and marks the invoice
+// as settled.
+func (d *DB) SettleHoldInvoice(preimage lntypes.Preimage) (*Invoice, error) {
+	var updatedInvoice *Invoice
+	hash := preimage.Hash()
+	err := d.Batch(func(tx *bbolt.Tx) error {
```
cfromknecht: qq: did this use `Batch` previously? i think we want to set `updatedInvoice` to nil during each invocation to avoid side-effects reexecution of batch updates

rpcserver.go
joostjager: Ok, nice cleanup :+1:

channeldb/invoices.go
joostjager: It was `Update` before, but iirc it was your own suggestion to use `Batch`. But now I actually reverted it back to `Update`. I don't think there is a practical advantage of `Batch` here, only more things that can go wrong.

joostjager:
@cfromknecht ptal

channeldb/invoices.go
cfromknecht: was thinking the same thing, can always revisit if there is evidence that this is a bottleneck

mandelmonkey:
You mention that
```
SubscribeSingleInvoice is not available through lncli. For testing, the invoice state can be observed through lncli lookupinvoice <hash>.
```

however in the invoices.proto there is no rhash returned in the addHoldInvoiceResponse
https://github.com/lightningnetwork/lnd/blob/master/lnrpc/invoicesrpc/invoices.proto#L87

should we be expected to decode the payment_request to get this?

joostjager:
For AddHoldInvoice, you pass in the hash yourself. Returning it in the response wouldn't be new information.

mandelmonkey:
oh I see so the rhash is the hash you pass, makes sense thanks

mandelmonkey:
im trying this out with gomobile which uses some native bindings for grpc, so I had to create the bindings from the invoice proto.

I can see the invoice server is running in the logs 
```
Invoices RPC server successfully registered with root
```

and I created an InvoicesClient that is listening on the rpcServer connection

when I call addHoldInvoice I get the error

```
rpc error: code = Unimplemented desc = unknown service invoicesrpc.Invoices
```
at
https://github.com/lightningnetwork/lnd/blob/master/lnrpc/invoicesrpc/invoices.pb.go#L502

any ideas as to what im doing wrong?

mandelmonkey:
what is the reason this requites a special make tag? will this be removed in the future so it is automatically part of the default LND?

joostjager:
Yes, we will remove the build tag in the future and make it part of default `lnd`. Eventually we want to also start removing redundant calls from the main rpc server.

mandelmonkey:
Awesome thanks

AndySchroder:
Reading through this pull request and all the comments, it's hard for me to figure out if the open issue related to automatic cancelling of the accepted but unsettled HOLD invoice has been addressed. Has it? If so, how does it actually work? Is there a new htlc that is created that just reverts back to the original channel state and everyone in the route would rather do that than deal with a channel closure? At what time does the automatic cancellation occur? It seems like the default `cltv_expiry` is `40`, yet `(expiry_height)-(accept_height)` seems to be `43`. Is there any additional margin applied for the automatic cancellation?

Also, has the other open issue related to the automatic cancellation of all invoices based on the `expiry` been handled? It seems like this cancellation is related to a node policy and nothing to do with htlc's?

alexbosworth:
auto canceling the held htlc has been implemented yes

PR #2033, opened 2018-10-11
lnrpc: Add payment hash to SendResponse

gitlikeagirl:
When SendPayment grpc call fails, there is currently no way of associating the failure with a payment. 
Payment hash is already extracted by extractPaymentIntent, so it can just be added to SendResponse.
Did not add hash to successful sends, since they have the preimage anyway.


gitlikeagirl:
ah, thanks @ccdle12 🙈 
any idea what I can do about the timeout in `TestLightningWallet/btcwallet/neutrino:reorg_wallet_balance` (14.68s)` ? Can't see my change affecting the test.

wpaulino:
@gitlikeagirl that is a known issue with our integration tests, I wouldn't worry about it for now.

joostjager:
There is one more stream.send. The success case. It does contain the preimage, so the hash can be computed. For consistency I'd think it is better to always supply the hash, as it is the main key to identify a payment.

wpaulino:
Needs a rebase!

halseth:
LGTM after rebase!

Roasbeef:
Ready to land after a rebase, will have to synchronize with any other PRs that also modify the protos.

lnrpc/rpc.proto
```diff
@@ -730,6 +730,7 @@ message SendResponse {
     string payment_error = 1 [json_name = "payment_error"];
     bytes payment_preimage = 2 [json_name = "payment_preimage"];
     Route payment_route = 3 [json_name = "payment_route"];
+    bytes payment_hash = 4 [json_name ="payment_hash"];
```
halseth: missing space after `=`

gitlikeagirl:
Space added and rebase done. Just not sure how to address the merge conflict in rpc.pb.go, is it ok to leave as is?

halseth:
To resolve the merge conflict, checkout the master version: `git checkout master lnrcp/rpc.pb.go` and regen the protos: `make rpc`.

PR #2039, opened 2018-10-12
Add Autopilot sub-RPC server

halseth:
This PR adds new subserver that manages `lnd`s `autopilot`, enabled by providing the `autopilotrpc` flag at build time.

With this enabled, a new RPC for getting the status if the autopilot agent, and enabling/disabling the agent at runtime is set up on the `gRPC` server.

For builds without the `autopilotrpc` flag, no behavior is changed, and the `autopilot.enable` flag in the config will still determine whether the agent is enabled or not.


halseth:
Blocked by https://github.com/lightningnetwork/lnd/pull/2081

halseth:
Rebased on top of the `subserver` PR.

halseth:
Rebased on subserver-enabled master, ready for review and testing! 💃

AndrewSamokhvalov:
Would you consider adding here something similar to sending payment RPC?
1. `router` => `agent`
2. `queryroutes` => `querynodesscores`
3. `sendtoroute` => `openchannelswith`

The end goal is to be able to use the wisdom of lnd itself, but adding some knowledge from outside as well during process of agent openning channels with nodes in the network.

As far as I understand there are two ways how third-party could influence the process of opening channel today is either (1) work directly with `openchannel/closechannel` rpc and avoid agent,  or (2) rewrite heuristic interface to use remote metrics fetching.

In second case third-party lose information from `ConstrainedPrefAttachment` heuristic.

As an example of additional metric might be “percentage of time node is being online”, or “percentage of success payments routed through node”… which is simple by itself and probably shouldn’t require sending PRs in lnd. Based on this metrics third=party could filter nodes and than send the result back to agent.

Roasbeef:
>  rewrite heuristic interface to use remote metrics fetching, but in this case lose information from ConstrainedPrefAttachment heuristic.

No, they can simply be composed. 

> As an example of additional metric might be “percentage of time node is being online”, or “percentage of success payments routed through node”. 

Check out the PR, all of these can be added as new score based heuristics.

autopilot/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	// Manager.
+	PilotCfg *Config
+
+	// ChannelState is a function closure that returns the current set of
+	// channels managed by this node.
+	ChannelState func() ([]Channel, error)
+
+	// SubscribeTransactions is used to get a subscription for transactions
+	// relevant to this node's wallet.
+	SubscribeTransactions func() (lnwallet.TransactionSubscription, error)
```
Roasbeef: Also needs the ability to cancel the subscription? Example scenario being when we turn off autopilot programmatically.

autopilot/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	close(r.quit)
+	r.wg.Wait()
+
+	return nil
+}
+
+// GetStatus returns the current active status of the autopilot agent.
+func (r *Manager) GetStatus() bool {
```
Roasbeef: Returns a `bool` atm, instead should possibly be named: `IsActive`?

autopilot/manager.go
```diff
@@ -0,0 +1,277 @@
...
+		quit: make(chan struct{}),
+	}, nil
+}
+
+// Start starts the Manager.
+func (r *Manager) Start() error {
+	if !atomic.CompareAndSwapUint32(&r.started, 0, 1) {
+		return nil
+	}
+
```
Roasbeef: Should this start the agent?

autopilot/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	// topology updates.
+	txnSubscription, err := r.cfg.SubscribeTransactions()
+	if err != nil {
+		defer r.pilot.Stop()
+		return err
+	}
+	graphSubscription, err := r.cfg.SubscribeTopology()
+	if err != nil {
+		defer r.pilot.Stop()
+		defer txnSubscription.Cancel()
```
Roasbeef: Why not just call these methods (and the ones above) directly as we're shutting down for certain?

autopilot/manager.go
```diff
@@ -0,0 +1,277 @@
...
+		return err
+	}
+
+	// We'll launch a goroutine to provide the agent with notifications
+	// whenever the balance of the wallet changes.
+	// TODO(halseth): can lead to panic if in process of shutting down.
```
Roasbeef: What leads to a panic? The `wg`?

lnrpc/autopilotrpc/autopilot.proto
```diff
@@ -0,0 +1,33 @@
...
+service Autopilot {
+    /**
+    GetStatus returns the status of the daemon's autopilot agent.
+    */
+    rpc GetStatus(GetStatusRequest) returns (GetStatusResponse);
+
+    /**
+    Enable is used to enable or disable the autopilot agent.
+    */
+    rpc Enable(EnableRequest) returns (EnableResponse);
```
Roasbeef: Possible can name it something like `ModifyStatus`? As it's a bit awkward that atm you need to call `Enable` with a false bool rather than something like `Disable`.

lnrpc/autopilotrpc/autopilot_server.go
```diff
@@ -0,0 +1,204 @@
...
+	// SubServerConfigDispatcher instance recognize as the name of our
+	subServerName = "AutopilotRPC"
+)
+
+var (
+	// macaroonOps are the set of capabilities that our minted macaroon (if
+	// it doesn't already exist) will have.
+	macaroonOps = []bakery.Op{
+		{
+			Entity: "autopilot",
```
Roasbeef: So one question is if we want to require an _entirely_ new macaroon for interaction with the pilot sub-sever, or if we want to use the existing entity mapping. FWIW, the latter route would allow the `admin.macaroon` to use the service as is without having to generate an additional set of macaroons. However, if we make these a series of "sub-commands" for `lncli`, then these sub-commands can themselves know where to fetch the proper macaroon.

.gitignore
```diff
@@ -24,10 +24,10 @@ _testmain.go
 *.test
 *.prof
 
-lnd
-lnd-debug
-lncli
-lncli-debug
+/lnd
```
Roasbeef: Why do we _ever_ want to include binaries?

cmd/lncli/main.go
```diff
@@ -292,6 +292,9 @@ func main() {
 		forwardingHistoryCommand,
 	}
 
+	// Add any extra commands determined by build flags.
+	app.Commands = append(app.Commands, extraCommands()...)
```
Roasbeef: What do you think of making all of these sub-commands? So it would be something along the lines of: 
```
lncli autopilot enable
lncli autopilot disable
```

halseth:
@AndrewSamokhvalov Check out the PR #2212. It is still WIP, but I think it covers what you want :) It will let you provide your own scores for node, and choose the weight given to the active heuristics.

.gitignore
halseth: Problem is that this will match on folders as well. In the case of this PR, the files added to `cmd/lncli` wouldn't show up, as they were ignored.

lnrpc/autopilotrpc/autopilot.proto
halseth: What do you think about `SetEnabled` or `SetActive`?

lnrpc/autopilotrpc/autopilot_server.go
halseth: I'm fine with reusing existing macaroons, as you don't really gain more access than what you already have with the regular macaroon.

However, it would be nice if subservers had a standardized macaroon setup.

cmd/lncli/main.go
halseth: TIL! Nice 🤠

autopilot/manager.go
halseth: It is being done by calling `Cancel` on the returned subscription.

autopilot/manager.go
halseth: Done!

autopilot/manager.go
halseth: I decided not to, to keep the current behavior where autopilot is only started if the `--autopilot.enabled` config flag is set.

autopilot/manager.go
halseth: Fixed.

autopilot/manager.go
halseth: Yes, as @cfromknecht has mentioned earlier, this could lead to panic in certain very rare cases.

lnrpc/autopilotrpc/autopilot.proto
Roasbeef: Those could work as well, personally I lean towards `ModifyStatus` as in the future we can do things like modify the status of things like it's "operating type", so: end user, routing node, etc. Of course, this can also be a follow up later.

lnrpc/autopilotrpc/autopilot_server.go
Roasbeef: Yeah re macaroon set up, once the primary sub-servers are in, following your suggestion, we can modify them to simply present which macaroons they need, and then the (to be fleshed out) bakery would handle baking them.

.gitignore
Roasbeef: Hmm, ok never had that happen to me, but can't hurt I guess.

lnrpc/autopilotrpc/autopilot_server.go
Roasbeef: Had some thoughts about this earlier w.r.t a "catch all" capability in the admin macaroon. This is a follow up though, and would require one last (hopefully) change to the admin macaroon to basically give it a "God" capability. All other method calls would by pass their regular checks if a macaron with this capability was presented.

Roasbeef:
Ready to test this out after we rebase and squash the fixup commits!

lnrpc/autopilotrpc/autopilot.proto
halseth: Renamed RPCs to `Status` and `ModifyStatus`.

lnrpc/autopilotrpc/autopilot_server.go
halseth: Ended up removing custom autopilot macaroons completely for now, and instead require the `info/read` permission for getting status, and `offchain/write + onchain/write` for modifying the status.

halseth:
Rebased , squashed and last comments addressed.

PR #2042, opened 2018-10-12
Peer events subscription RPC

halseth:
This PR adds a new RPC for subscribing to peer events. This is useful to track connection status of peers.

1. The `ConcurrentQueue` is moved to its own package `queue`, since it is useful for reusing across packages.
2. A new package `subscribe` is added, which contains a generic client-server design for handling event subscription. This is intended for use other places in the code base (e.g. graph topology updates).
3. The `subscribe` package is used in the `server` to notify any interested grpc client about connections status changes.


Roasbeef:
Replaced by #1988.

PR #1787, opened 2018-08-25
chainntnfs: Isolate conf notifications during historical scans

cfromknecht:
In this PR, we modify the `TxConfNotifier` to isolate confirmation notifications while a historical dispatch is being processed. The purpose of this is to segment which notifications should have their height hints updated in the cache. During a historical rescan, we don't yet know if the txn is in the chain, so we shouldn't continue to update its height hint until after finishing the rescan. Once the rescan finishes, the notification is grouped along with all other notifications at tip so that their height hints can progress as blocks are added.

This PR also introduces:
 - a reorg safety delta to height hints, currently 144 blocks
   - with the recent LRU cache added to neutrino, this shouldn't result too much unnecessary network I/O
   - the height hint is still clamped to whatever the client initial prescribes, so initial registration doesn't take a hit either
 - moves querying of confirmation height hints into the `TxConfNotifier`
 - reorders the persistence/spend-notification dispatch to improve external consistency

In a follow PR, we will do the same for spend notifications.

This PR depends on #1951, which contains the first four commits

PR #1621, opened 2018-07-25
[Integration tests] Split chain backend and miner

halseth:
This PR alters the integration test setup to use a chain backend separate from the miner. This is done to make it possible to switch the chain backend with a different implementation (currently we are using `btcd`).

The separation of chain backend and miner increases the delay of transaction and block propagation, since there is one extra node every event must be relayed through. This caused some test flakes to manifest themselves, and they are fixed one by one in the commits before splitting backend and miner. 

A follow-up PR will enable `bitcoind` and `neutrino` as chain backends.

### Blocked by:
- [x] `btcd` Node RPC: https://github.com/btcsuite/btcd/pull/1217
- [x] `btcd` configurable `trickleTimeout`: https://github.com/btcsuite/btcd/pull/1222

### Builds on:
- [x] https://github.com/lightningnetwork/lnd/pull/1608
- [x] https://github.com/lightningnetwork/lnd/pull/1613
- [x] https://github.com/lightningnetwork/lnd/pull/2116
- [x] https://github.com/lightningnetwork/lnd/pull/2117

PR #2044, opened 2018-10-12
chainntnfs: re-enable height hint cache

wpaulino:
In this PR, we attempt to re-enable the height hint cache for the different `ChainNotifier` implementations. This will lead to performance improvements upon startup when running a full node backend without `txindex` or when running as a light client, like Neutrino.

~~Depends on #1787 and #2004.~~

### DIFF
--- a/chainntnfs/bitcoindnotify/bitcoind.go
+++ b/chainntnfs/bitcoindnotify/bitcoind.go
@@ -568,14 +568,19 @@ func (b *BitcoindNotifier) confDetailsManually(txid *chainhash.Hash,
 // transactions included this block will processed to either send notifications
 // now or after numConfirmations confs.
 func (b *BitcoindNotifier) handleBlockConnected(block chainntnfs.BlockEpoch) error {
+	// First, we'll fetch the raw block as we'll need to gather all the
+	// transactions to determine whether any are relevant to our registered
+	// clients.
 	rawBlock, err := b.chainConn.GetBlock(block.Hash)
 	if err != nil {
 		return fmt.Errorf("unable to get block: %v", err)
 	}
-
 	txns := btcutil.NewBlock(rawBlock).Transactions()
-	err = b.txNotifier.ConnectTip(
-		block.Hash, uint32(block.Height), txns)
+
+	// We'll then extend the txNotifier's height with the information of
+	// this new block, which will handle all of the notification logic for
+	// us.
+	err = b.txNotifier.ConnectTip(block.Hash, uint32(block.Height), txns)
 	if err != nil {
 		return fmt.Errorf("unable to connect tip: %v", err)
 	}
@@ -583,15 +588,15 @@ func (b *BitcoindNotifier) handleBlockConnected(block chainntnfs.BlockEpoch) err
 	chainntnfs.Log.Infof("New block: height=%v, sha=%v", block.Height,
 		block.Hash)
 
-	// We want to set the best block before dispatching notifications so
-	// if any subscribers make queries based on their received block epoch,
-	// our state is fully updated in time.
+	// Now that we've guaranteed the new block extends the txNotifier's
+	// current tip, we'll proceed to dispatch notifications to all of our
+	// registered clients whom have had notifications fulfilled. Before
+	// doing so, we'll make sure update our in memory state in order to
+	// satisfy any client requests based upon the new block.
 	b.bestBlock = block
 
-	// Lastly we'll notify any subscribed clients of the block.
 	b.notifyBlockEpochs(block.Height, block.Hash)
-
-	return nil
+	return b.txNotifier.NotifyHeight(uint32(block.Height))
 }
 
 // notifyBlockEpochs notifies all registered block epoch clients of the newly
--- a/chainntnfs/txnotifier.go
+++ b/chainntnfs/txnotifier.go
@@ -887,9 +887,13 @@ func (n *TxNotifier) dispatchSpendDetails(ntfn *SpendNtfn, details *SpendDetail)
 //   confirmation registration for.
 //
 // In the event that the transaction is relevant, a confirmation/spend
-// notification will be dispatched to the relevant clients. Confirmation
-// notifications will only be dispatched for transactions that have met the
-// required number of confirmations required by the client.
+// notification will be queued for dispatch to the relevant clients.
+// Confirmation notifications will only be dispatched for transactions that have
+// met the required number of confirmations required by the client.
+//
+// NOTE: In order to actually dispatch the relevant transaction notifications to
+// clients, NotifyHeight must be called with the same block height in order to
+// maintain correctness.
 func (n *TxNotifier) ConnectTip(blockHash *chainhash.Hash, blockHeight uint32,
 	txns []*btcutil.Tx) error {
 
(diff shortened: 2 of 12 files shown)


wpaulino:
Ready for review! There are some additional commits (might seem unrelated) that solve some integration test flakes due to timing issues upon re-enabling the height hint cache.

One thing that's left to address is whether we want to clear the existing state of users' caches. I believe this won't be needed in the case that users have successfully started up after the cache was disabled, as a rescan will occur from the broadcast height and catch everything since then. If this hasn't happened however, it's possible that they begin their rescan after the fact due to previous incorrect behavior of the cache.

chainntnfs/txnotifier.go
```diff
@@ -891,7 +891,7 @@ func (n *TxNotifier) dispatchSpendDetails(ntfn *SpendNtfn, details *SpendDetail)
 // notifications will only be dispatched for transactions that have met the
 // required number of confirmations required by the client.
 func (n *TxNotifier) ConnectTip(blockHash *chainhash.Hash, blockHeight uint32,
-	txns []*btcutil.Tx) error {
+	txns []*btcutil.Tx, notify bool) error {
```
halseth: godoc should be updated.

chainntnfs/bitcoindnotify/bitcoind.go
```diff
@@ -568,30 +568,37 @@ func (b *BitcoindNotifier) confDetailsManually(txid *chainhash.Hash,
...
 	}
-
 	txns := btcutil.NewBlock(rawBlock).Transactions()
+
+	// We'll then extend the txNotifier's height with the information of
+	// this new block, which will handle all of the notification logic for
+	// us.
 	err = b.txNotifier.ConnectTip(
-		block.Hash, uint32(block.Height), txns)
+		block.Hash, uint32(block.Height), txns, false,
```
halseth: If this will always be false now, I would remove the `notifyHeight` call from `ConenctTip` instead.

chainntnfs/bitcoindnotify/bitcoind.go
```diff
@@ -568,30 +568,37 @@ func (b *BitcoindNotifier) confDetailsManually(txid *chainhash.Hash,
...
+	// registered clients whom have had notifications fulfilled. Before
+	// doing so, we'll make sure update our in memory state in order to
+	// satisfy any client requests based upon the new block.
 	b.bestBlock = block
 
-	// Lastly we'll notify any subscribed clients of the block.
 	b.notifyBlockEpochs(block.Height, block.Hash)
-
-	return nil
+	return b.txNotifier.NotifyHeight(uint32(block.Height))
```
halseth: can `notifyBlockEpochs` be moved into `NotifyHeight` now?

discovery/gossiper.go
```diff
@@ -2344,13 +2344,63 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 				source: nMsg.source,
 				msg:    e1Ann,
 			})
+
+			// In addition to sending this edge's policy, we'll also
+			// attempt to send the backing node's announcement if we
+			// happen to know about it.
+			pubKey, _ := chanInfo.NodeKey1()
```
halseth: How is this change related to the rest of the PR? Could it be made its own PR?

chainntnfs/txnotifier.go

chainntnfs/bitcoindnotify/bitcoind.go
wpaulino: Agreed, fixed.

chainntnfs/bitcoindnotify/bitcoind.go
wpaulino: Nope, as the `TxNotifier` is not aware of the block epoch clients. Would also prefer to keep the distinction between the block epoch clients and transaction-related clients.

discovery/gossiper.go
wpaulino: It's related in that it solves some integration test issues that became more apparent with the cache enabled. Not opposed to making it its own PR - let me know what you think.

chainntnfs/txnotifier.go
```diff
@@ -887,9 +887,12 @@ func (n *TxNotifier) dispatchSpendDetails(ntfn *SpendNtfn, details *SpendDetail)
...
 // In the event that the transaction is relevant, a confirmation/spend
-// notification will be dispatched to the relevant clients. Confirmation
-// notifications will only be dispatched for transactions that have met the
-// required number of confirmations required by the client.
+// notification will be queued for dispatch to the relevant clients.
+// Confirmation notifications will only be dispatched for transactions that have
+// met the required number of confirmations required by the client.
+//
+// NOTE: In order to actually dispatch the relevant transaction notifications to
+// clients, NotifyHeight must be called.
```
cfromknecht: should probably note that this method MUST be called if `ConnectTip` succeeds in order to maintain correctness of the notifications. the current godoc makes it sounds semi-optional imo

chainntnfs/txnotifier.go
```diff
@@ -1017,22 +1020,32 @@ func (n *TxNotifier) ConnectTip(blockHash *chainhash.Hash, blockHeight uint32,
...
 
-	// Next, we'll dispatch an update to all of the notification clients for
-	// our watched transactions with the number of confirmations left at
+	return nil
+}
+
+// NotifyHeight dispatches confirmation and spend notifications to the clients
+// who registered for a notification which has been fulfilled at the passed
+// height.
+func (n *TxNotifier) NotifyHeight(height uint32) error {
```
cfromknecht: great addition to the interface, external consistency ftw :D

discovery/gossiper.go
halseth: Yeah, let's make it a own PR, should be fairly easy to review and get in.

chainntnfs/bitcoindnotify/bitcoind.go
halseth: Agreed for now. I think in the future it would make sense to unify them somehow to avoid the duplication across backends.

discovery/gossiper.go
halseth: It should also have an accompanying test to demonstrate the problem.

wpaulino: Turns out these changes are actually no longer needed due to #1981!

chainntnfs/txnotifier.go

wpaulino:
There shouldn't be any issues with the integration tests other than:
```
    --- FAIL: TestLightningNetworkDaemon/revoked_uncooperative_close_retribution_zero_value_remote_output (35.39s)
        lnd_test.go:74: Failed: (revoked uncooperative close retribution zero value remote output): exited with error:
            *errors.errorString unable to find Carol's force close tx in mempool: wanted 1, found 2 txs in mempool: [98075d6b639105d4ed259c77ed171c7e14f3d9daeb22fce5b31a5287668d52f2 b8b042732be88531c640f3f4179cfde1a5baf8a3cd5423c3df0723598c39c0e2]
            /Users/wilmer/go/src/github.com/lightningnetwork/lnd/lnd_test.go:6417 (0x17ade6a)
            	testRevokedCloseRetributionZeroValueRemoteOutput: t.Fatalf("unable to find Carol's force close tx in mempool: %v",
            /Users/wilmer/go/src/github.com/lightningnetwork/lnd/lnd_test.go:99 (0x1788ae6)
            	(*harnessTest).RunTestCase: testCase.test(net, h)
            /Users/wilmer/go/src/github.com/lightningnetwork/lnd/lnd_test.go:12591 (0x17e1c44)
            	TestLightningNetworkDaemon.func3: ht.RunTestCase(testCase, lndHarness)
            /usr/local/Cellar/go/1.11.1/libexec/src/testing/testing.go:827 (0x10f590f)
            	tRunner: fn(t)
            /usr/local/Cellar/go/1.11.1/libexec/src/runtime/asm_amd64.s:1333 (0x105dcb1)
            	goexit: BYTE	$0x90	// NOP
```

I rarely hit this on Travis though, can only reproduce it reliably locally. This will be resolved by https://github.com/lightningnetwork/lnd/pull/1621/commits/dbc11c7ed3c4608dffe7aa6fd9d1f67512028311.

halseth:
Also note that I split the integration test fix into https://github.com/lightningnetwork/lnd/pull/2116, so we can get them in before doing the bigger multi-backend change.

Roasbeef:
I don't think we need to worry about clearing any existing caches, as it was never included in a major release. Additionally, any entries that users had in their caches are now long expired since many weeks have passed since the initial implementation, and the commit that disabled all the caches. The users that had prior height hints also in a way will _benefit_ from them as they won't need to start from the initial broadcast (or confirm or w/e height) and can start a bit earlier in the chain.

Issue #2034, opened 2018-10-11
[tor] if run with "tor.v3" option, then "lnd must *only* be listening on localhost when running with Tor inbound support enabled"


BobleChinois:
### Background

I got the following error when starting LND with tor service activated for the first time : 
```
Attempting automatic RPC configuration to bitcoind
Automatically obtained bitcoind's RPC credentials
2018-10-11 15:55:13.562 [INF] LTND: Shutdown complete
lnd must *only* be listening on localhost when running with Tor inbound support enabled
```

I have already been running bitcoind with tor for some times without issues.

I think I'm getting something wrong with either LND or tor configuration, but can't tell what.

### Your environment

* lnd version 0.5.0-beta commit=25145acc46cc5d18e4e348eed097300b1391d2a7
* Linux Celeri 4.15.0-36-generic #39-Ubuntu SMP Mon Sep 24 16:19:09 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* Bitcoin Core Daemon version v0.16.3
* Tor version : 0.3.2.10 (git-0edaa32732ec8930).
* `lnd.conf` content : 
```
debuglevel=trace
bitcoin.active=1
bitcoin.mainnet=1
bitcoin.node=bitcoind
alias=88325dc3ad81fd9caf2249ea74dcff62ad8542b0ab35d1a1e25d5d2f7ac29f34

tor.active=1
tor.v3=1
```

### Steps to reproduce

I started a new LND node from scratch, created a wallet, and let it synchronize. Tor was NOT activated then.

When synch was over, I turned it off, and created the lnd.conf file with content as seen upside.

When running lnd again, I get the error message already provided.

I turned off `tor.v3=1` inside the conf file, and then lnd can start normally.

Here is the relevant piece of code I could find.

https://github.com/lightningnetwork/lnd/blob/8dfcbc160d056cba039c7e1d7931d4867bc385de/config.go#L937-L947

### Expected behaviour

LND should start up with listening on tor v3 enabled.

### Actual behaviour

LND would only start if `listening on the P2P interface is disabled`

wpaulino:
This is intended to prevent unintentionally leaking any identifying information. You must run `lnd` with `listen=localhost` if you intend to use Tor.

BobleChinois:
[...]

PR #2053, opened 2018-10-15
update 'configuring_tor' file

BobleChinois:
I recently ran lnd with tor for the first time. While following instructions on [configuring_tor file](https://github.com/lightningnetwork/lnd/blob/4dd4f7cf452709d90fa1d16062b13874623c929a/docs/configuring_tor.md), I wasn't able to run it at first because I didn't add `listen=localhost` in my config file (#2034).

I think we should add it to the instructions.

I also add another minor issue because my tor version was not 0.3.3.6, I think it should help to add this requirement in the same file.


Roasbeef:
In this case, shouldn't we automatically listen on `localhost`?

PR #2054, opened 2018-10-15
htlcswitch: settle invoices behind switch [WIP]

joostjager:
This PR is an exploration of what would happen when we move invoice settling behind the switch instead of inside the channel link.

Goals:
- Leverage CircuitMap to store the link between the final htlc and an invoice (proposal: outgoingchannelid=exithop, outgoinghtlcidx=invoice addindex). This link can be used when hold invoices are settled/failed over rpc.
- Prepare for amp. With amp, invoices are settled via multiple channel links. Therefore the invoice settling needs to be lifted out of the link level.

Implementation:
- Remove the complete 'if exithop' case from the link
- Have the switch match exithop with the outgoing channel id and forward to a new struct InvoiceSettler.
- InvoiceSettler is managing the interaction with the invoice database.

### DIFF
--- a/htlcswitch/invoice_settler.go
+++ b/htlcswitch/invoice_settler.go
@@ -1,0 +1,243 @@
+package htlcswitch
+
+import (
+	"github.com/lightningnetwork/lnd/contractcourt"
+)
+
+// InvoiceSettler handles settling and failing of invoices.
+type InvoiceSettler struct {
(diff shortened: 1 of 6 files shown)


joostjager:
Main problem atm is when and how to get the invoice addindex as outgoing htlc index in the circuitmap.

joostjager:
This PR is partially woven into #2022, closing.

PR #2060, opened 2018-10-17
Ignore redundant connection request to peer

roeierez:
This PR solves the case where trying to connect to a peer that already have a pending connection request.
In that case the system would try to fulfill these two requests in parallel instead of just ignoring the later.
This fix the issue by returning early from the function, which I believe was the original intention of the author.

### DIFF
--- a/server.go
+++ b/server.go
@@ -2687,19 +2687,18 @@ func (s *server) ConnectToPeer(addr *lnwire.NetAddress, perm bool) error {
 
 	// Peer was not found, continue to pursue connection with peer.
 
-	// If there's already a pending connection request for this pubkey,
-	// then we ignore this request to ensure we don't create a redundant
-	// connection.
-	if reqs, ok := s.persistentConnReqs[targetPub]; ok {
-		srvrLog.Warnf("Already have %d persistent connection "+
-			"requests for %v, connecting anyway.", len(reqs), addr)
-	}
-
-	// If there's not already a pending or active connection to this node,
-	// then instruct the connection manager to attempt to establish a
-	// persistent connection to the peer.
 	srvrLog.Debugf("Connecting to %v", addr)
 	if perm {
+
+		// If there's already a pending connection request for this pubkey,
+		// then we cancel pending request to ensure we don't create a redundant
+		// connection.
+		if reqs, ok := s.persistentConnReqs[targetPub]; ok {
+			srvrLog.Warnf("Already have %d persistent connection "+
+				"requests for %v, connecting anyway.", len(reqs), addr)
+			s.cancelConnReqs(pubStr, nil)
+		}
+
 		connReq := &connmgr.ConnReq{
 			Addr:      addr,
 			Permanent: true,


roeierez:
Although the comments indicate this was the intention I am not sure now about this fix.
Before the fix if there was a pending connection (might even be stuck on the backoff timeout) the "ConnectToPeer" gRPC request would have been gone through and attempt to connect in a forceful way.
After the fix it would be ignored, leaving the pending connection hang on its current backoff.
As it is useful to be able to force connect especially in mobile devices when the network status is varying, thinking maybe just to document this better?

roeierez:
I have updated the PR to maintain the old behavior of connecting anyway but to also cancel old pending requests.
The guidelines for this change are:
* One of the connection attempts should be canceled, it is not good to have two connection attempts for the same peer in parallel.
* It is better to let the user override pending requests, for example when some pending request is stuck on a long backoff retry.

server.go
```diff
@@ -2604,11 +2604,12 @@ func (s *server) ConnectToPeer(addr *lnwire.NetAddress, perm bool) error {
 	// Peer was not found, continue to pursue connection with peer.
 
 	// If there's already a pending connection request for this pubkey,
-	// then we ignore this request to ensure we don't create a redundant
+	// then we cancel older requests to ensure we don't create a redundant
 	// connection.
 	if reqs, ok := s.persistentConnReqs[targetPub]; ok {
 		srvrLog.Warnf("Already have %d persistent connection "+
 			"requests for %v, connecting anyway.", len(reqs), addr)
+		s.cancelConnReqs(targetPub, nil)
```
halseth: Log above should be updated?

roeierez: done!

wpaulino: As noted above, we should only cancel them if the connection attempt was successful or if `perm` was set.

wpaulino:
I believe we would also need to mark peers as persistent within `peer.Start` if there are any active channels.

server.go
roeierez: > we should only cancel them if the connection attempt was successful

This may cause more than one connection in progress which I think will cause the remote node to drop the duplicate connection. Isn't it cleaner to make sure we only have one pending connection?



> or if `perm` was set.

Do you think there is a problem with not differentiating the "perm" case here? I think it would be nice to have the same behavior for both cases.

Thanks.

wpaulino: > This may cause more than one connection in progress which I think will cause the remote node to drop the duplicate connection. Isn't it cleaner to make sure we only have one pending connection?

If we assume that the peer will drop the duplicate connection, then there shouldn't be an issue. Another alternative is to cancel the persistent request and proceed to connect. If we fail to connect, we can restore the persistent request.

> Do you think there is a problem with not differentiating the "perm" case here? I think it would be nice to have the same behavior for both cases.

Not sure what you mean by differentiating the `perm` case. If it is set we'll end up with duplicate persistent connection requests, which we want to avoid.

roeierez: > Another alternative is to cancel the persistent request and proceed to connect. If we fail to connect, we can restore the persistent request.

I think we don't need to write any code to restore the persistent request because if the new connection request fails it will try again using backoff if this peer is persistent. My intention was just to replace a pending request with another.



> Not sure what you mean by differentiating the `perm` case

Sorry I wasn't clear about it. I was saying why limiting the cancelation only for the "perm" case and what problem do you see in canceling pending requests also for non permanent connection if the user just called ConnectToPeer again?

wpaulino: Let's say there was a persistent connection request to a peer because we have a channel open with them. The request is still active since the peer is offline. If the user tries to manually connect to this peer and fails, if we cancel the request without restoring it, then we wouldn't be able to reconnect to the peer once they come back online.

roeierez: Got you! , makes sense. Now I understand also why you suggested to mark persistent connections at peers.Start. This will make sure the persistent pees will survive cancellations and stay persistent.
So in summary, the problematic case is when the user tries to connect manually (perm=false) to a persistent peer with a pending connection. At this case we want to make sure the connection stays persistent regardless of the new connection attempt results.
So I think to limit the cancelation only for the cases where perm=true (as you suggested) will solve this case right?  I will update the PR.

roeierez: @wpaulino I have moved the cancelation code into the block that is executed only in case of perm connection. This intends to cover the first comment of the review (only cancel in case of perm connection).
As for the second comment (cancel pending requests if connection is successful) I saw this is already handled here: https://github.com/lightningnetwork/lnd#L2214

server.go
```diff
@@ -2687,19 +2687,18 @@ func (s *server) ConnectToPeer(addr *lnwire.NetAddress, perm bool) error {
 
 	// Peer was not found, continue to pursue connection with peer.
 
-	// If there's already a pending connection request for this pubkey,
```
cfromknecht: good catch! this comment was outdated for sure. i think we used to return an error here, and then replaced it with a log but looks like the comment never got updated.

server.go
```diff
@@ -2687,19 +2687,18 @@ func (s *server) ConnectToPeer(addr *lnwire.NetAddress, perm bool) error {
...
-	// If there's not already a pending or active connection to this node,
-	// then instruct the connection manager to attempt to establish a
-	// persistent connection to the peer.
 	srvrLog.Debugf("Connecting to %v", addr)
 	if perm {
+
+		// If there's already a pending connection request for this pubkey,
+		// then we cancel pending request to ensure we don't create a redundant
+		// connection.
+		if reqs, ok := s.persistentConnReqs[targetPub]; ok {
```
cfromknecht: indeed i think this is the right place to move this check!

server.go
```diff
@@ -2687,19 +2687,18 @@ func (s *server) ConnectToPeer(addr *lnwire.NetAddress, perm bool) error {
...
+			srvrLog.Warnf("Already have %d persistent connection "+
+				"requests for %v, connecting anyway.", len(reqs), addr)
+			s.cancelConnReqs(pubStr, nil)
```
cfromknecht: imo we don't need to do this (sorry if the old comment was confusing), it should be fine to have multiple pending connection requests to the same peer. note that these may actually be at different addresses, even though they have the same pubkey, and we will use whichever succeeds first

roeierez: @cfromknecht thanks for your review. I agree it creates a problem.
Originally I wanted to solve the case where two requests simultaneously tries  to connect but somehow they prevent from each other to connect.
The behavior I experienced is a constantly connect/disconnect effect even though the code aims to handle this case.
Could be that the behavior I experienced is related to a problem that might be solved by this PR: https://github.com/btcsuite/btcd/pull/1330
without the above PR connections are not cancelled if they are waiting for retry. so it could be that the cancelation of other requests when one succeed didn't work properly.

Anyway, we both agree the solution is not to cancel at this point and I will need to repro that behavior so I can suggest a better solution.

halseth:
@roeierez what's the status of this PR?

roeierez:
> @roeierez what's the status of this PR?

@halseth the problem was solved by merging https://github.com/btcsuite/btcd/pull/1330.
We can close this one for now.

halseth:
@roeierez Thanks!

PR #2063, opened 2018-10-17
multi: modify sweeper.CreateSweepTx to accept conf target, style changes

Roasbeef:
In this commit, we modify the newly introduced UtxoSweeper.CreateSweepTx
to accept the confirmation target as a param of the method rather than a
struct level variable. We do this as this allows each caller to decide
at sweep time, what the fee rate should be, rather than using a global
value that is meant to work in all scenarios. For example, anytime
we're sweeping an output with a CLTV lock that's has a dependant
transaction we need to sweep/cancel, we may require a higher fee rate
than a regular force close with a CSV output.

### DIFF
--- a/contractcourt/contract_resolvers.go
+++ b/contractcourt/contract_resolvers.go
@@ -5,9 +5,10 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
-	"github.com/lightningnetwork/lnd/sweep"
 	"io"
 	"io/ioutil"
+
+	"github.com/lightningnetwork/lnd/sweep"
 
 	"github.com/btcsuite/btcd/wire"
 	"github.com/davecgh/go-spew/spew"
@@ -18,6 +19,12 @@ import (
 
 var (
 	endian = binary.BigEndian
+)
+
+const (
+	// sweepConfTarget is the default number of blocks that we'll use as a
+	// confirmation target when sweeping.
+	sweepConfTarget = 6
 )
 
 // ContractResolver is an interface which packages a state machine which is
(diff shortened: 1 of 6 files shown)


contractcourt/contract_resolvers.go
```diff
@@ -1216,16 +1225,26 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
+		// With out input constructed, we'll now request that the
+		// sweeper construct a valid sweeping transaction for this
+		// input.
+		//
 		// TODO: Set tx lock time to current block height instead of
 		// zero. Will be taken care of once sweeper implementation is
 		// complete.
 		c.sweepTx, err = c.Sweeper.CreateSweepTx(
-			[]sweep.Input{&input}, 0)
+			[]sweep.Input{&input}, 6, 0,
```
cfromknecht: make 6 a const?

sweep/sweeper.go
```diff
@@ -166,9 +161,7 @@ func (s *UtxoSweeper) getWeightEstimate(inputs []Input) (int64, int, int) {
 	// For each output, use its witness type to determine the estimate
 	// weight of its witness, and add it to the proper set of spendable
 	// outputs.
-	csvCount := 0
-	cltvCount := 0
-
+	var csvCount, cltvCount int
```
cfromknecht: 👍

utxonursery.go
```diff
@@ -871,15 +872,14 @@ func (u *utxoNursery) graduateClass(classHeight uint32) error {
...
-				len(kgtnOutputs))
+			sweepInputs := make([]sweep.Input, len(kgtnOutputs))
 			for i := range kgtnOutputs {
 				sweepInputs[i] = &kgtnOutputs[i]
 			}
 
 			finalTx, err = u.cfg.Sweeper.CreateSweepTx(
-				sweepInputs, classHeight)
-
+				sweepInputs, sweepTxConfTarget, classHeight,
```
cfromknecht: thoughts on passing this in as a nursery config option, instead of accessing a global variable?

Roasbeef: SGTM, added in the latest diff.

contractcourt/contract_resolvers.go
Roasbeef: Fixed.

Issue #2045, opened 2018-10-12
HTLC not getting failed upstream after they timeout downstream


pm47:
### Background

I see quite a lot of htlcs timing out instead of failing on mainnet, which is quite annoying because that leads to channel closing. This could be normal if the counterparty was down, but in that particular case, I was pretty sure the peer was up and operational, so I tried to find out what happened.

### Your environment

* version of `lnd`: v0.5.0-beta

### Steps to reproduce

Scenario is a payment being relayed through `eclair` and then through `lnd`:

... -----> `eclair` ----- htlcA ----->	`lnd`	----- htlcB -----> ...

htlcA  lnd_channel_id=`544814:1271:1`  expiry=545073  htlc_id=7
htlcB  lnd_channel_id=`540823:2707:1`  expiry=545069  htlc_id=15

From the point of view of `eclair`, nothing happens until block #545073 gets mined, at which point htlcA has timed out, causing `eclair` to unilaterally close the channel.

I was able to get `lnd`'s logs ([log.zip](https://github.com/lightningnetwork/lnd/files/2474665/log.zip), thanks @AndrewSamokhvalov), here is what I think are the relevant parts:

```
"October 9th 2018, 16:39:57.044",DBG,PEER,"Received UpdateAddHTLC(chan_id=b0d69f3c998bfabfc32702e55fbe0ff3e2551d91ca234c692e0fc1b0a1a8745e, id=7, amt=135635112 mSAT, expiry=545073, hash=7202da369b8dae66cb9ac6b0dec53bfe7e254efe53ebfb0d3ef6b6f1ba90d4bc) from 34.239.230.56:9735"
...
"October 9th 2018, 16:39:58.660",DBG,PEER,"Sending UpdateAddHTLC(chan_id=8c8481ae7e2a76af1691599057f2e4bafa7cad382f68eb73a9a1650ee410344d, id=15, amt=135634976 mSAT, expiry=545069, hash=7202da369b8dae66cb9ac6b0dec53bfe7e254efe53ebfb0d3ef6b6f1ba90d4bc) to 47.184.129.94:43178"
...
"October 9th 2018, 16:39:59.664",TRC,HSWC,"ChannelLink(540823:2707:1) revocation window exhausted, unable to send: 1, dangling_opens=([]channeldb.CircuitKey) (len=1 cap=1) {
 (channeldb.CircuitKey) (Chan ID=544814:1271:1, HTLC ID=7)
}
[...]

PR #2069, opened 2018-10-19
discovery: replace buffered channel with ConcurrentQueue in syncer

Roasbeef:
In this commit, we fix an issue due to the rate limiting that could
cause us to stop reading ALL messages from a peer all together, rather
than just rate limiting the number of gossip query messages that we'll
process for that peer. In order to fix this, we now introduce a
potentially unbounded queue (that'll use a buffer amount then switch to
the greater queue) for each gossip syncer. With this change, we ensure
that we won't block the entire read handler just to throttle the
messages of a particular peer.

A longer term fix for this is a proper DoS engine.

Possibly related to #2045.

### DIFF
--- a/discovery/syncer.go
+++ b/discovery/syncer.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/queue"
 	"golang.org/x/time/rate"
 )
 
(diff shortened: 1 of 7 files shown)


discovery/syncer.go
```diff
@@ -426,19 +442,32 @@ func (g *gossipSyncer) channelGraphSyncer() {
 			// an ending reply, or just another query from the
 			// remote peer.
 			select {
-			case msg := <-g.gossipMsgs:
+			case msg := <-g.gossipMsgs.ChanOut():
+				// Ensure that this is at least an
+				// lnwire.Message, otherwise, we can't properly
+				// manipulate it.
+				netMsg, ok := msg.(lnwire.Message)
```
halseth: I think we should really type assert without checking `ok` here, as anything not `lnwire.Message` shouldn't be on this channel.

Roasbeef:
Closing this as the real issue was found and fixed.

PR #2073, opened 2018-10-19
Enable Tor control via (hashed)ControlPassword

### LINKED ISSUE #2070: tor integration missing HashedControlPassword
When using tor it seems to be best to have tor running on a different machine. lnd seems to support this, as tor related configuration options allow specification of an IP address.

Yet " // Its [lnds] only supported method of authentication is the SAFECOOKIE method."

the safecookie is regenerated with each restart of tor [0].

So when trying to run lnd via tor in such an isolated environment, with each restart of tor one has to manually sync up the cookie or sacrifice isolation. 

[0] https://www.torproject.org/docs/tor-manual.html.en

svewa:
fixes  #2070 by providing support for Tor HASHEDPASSWORD authentication


config.go
```diff
@@ -154,6 +155,7 @@ type torConfig struct {
 	DNS             string `long:"dns" description:"The DNS server as host:port that Tor will use for SRV queries - NOTE must have TCP resolution enabled"`
 	StreamIsolation bool   `long:"streamisolation" description:"Enable Tor stream isolation by randomizing user credentials for each connection."`
 	Control         string `long:"control" description:"The host:port that Tor is listening on for Tor control connections"`
+	ControlPassword string `long:"controlpassword" default-mask:"-" description:"Password for tor HashedControlPassword authentication"`
```
wpaulino: Nit: s/tor/Tor/

config.go
```diff
@@ -936,15 +939,15 @@ func loadConfig() (*config, error) {
 
 	// Finally, ensure that we are only listening on localhost if Tor
 	// inbound support is enabled.
-	if cfg.Tor.V2 || cfg.Tor.V3 {
+	if (cfg.Tor.V2 || cfg.Tor.V3) && !(len(cfg.Tor.ControlPassword) > 0) {
```
wpaulino: What's the rationale for this change? We can't assume that if a ControlPassword is provided then we're dealing with a remote Tor server.

tor/controller.go
```diff
@@ -176,6 +181,15 @@ func (c *Controller) authenticate() error {
 			"%v", err)
 	}
 
+	// If a controlPassword is provided, we try to authenticate using this.
```
wpaulino: Since we're introducing a new authentication method, I think it'd be better to refactor this into the following:

```
func (c *Controller) authenticate() error {
    // If a control password was provided, we'll assume the caller intends to use HashedControlPassword authentication.
    if len(c.controlPass) > 0 {
        return c.authenticateWithHashedPass()
    }
    return c.authenticateWithSafeCookie()
}
```

config.go
svewa: Right, ControlPassword is a bad indicator. Having a remote Tor server configured is a way better indicator.

svewa:
I fixed what you suggested and refactored, makes sense to me.
A recent merge caused #2114 for me, which I also reverted.

svewa:
Created #2119 and #2120 for the 2 distinct functionalities as suggested.

Roasbeef:
Closing this since it was superseded by PRs (one of which is open) further down the line.

Issue #2032, opened 2018-10-10
Resolved hashes in payments do not reflect back their settled state


alexbosworth:
### Background

When sending a payment that did not resolve immediately, the eventual preimage revelation via an on-chain HTLC success transaction did not reflect as a success in `listpayments`.

### Your environment

* `0.5.0-beta commit=b341dea3733cb721d6aecefa54123b5b1ac7b126`
* MacOS Mojave
* Bitcoin Core 0.17.0
* Sending to lnd 0.5.0

### Steps to reproduce

1. Sent to a destination that held on to the HTLC and did not resolve it
2. Restarted my lnd
3. Waited for eventual resolution of the HTLC which happened on chain
4. Settlement message appeared in logs, but no record was added to `listpayments` of the payment

### Expected behaviour

Payments should be listed in `listpayments` when funds have been successfully transfered

### Actual behaviour

A message indicating completion of the HTLC comes in, but there is an error regarding a payment lookup.

```
2018-10-09 22:45:43.943 [DBG] HSWC: ChannelLink(544281:1936:0): settle-fail-filter &{1 [0]}
2018-10-09 22:45:43.943 [DBG] HSWC: Closed completed SETTLE circuit for b5d5bf498e1a8ba0a3e8d05d5d666e00f2943c9f498dfb6a92af8795d98ceaa2: (0:0:0, 19003) <-> (544281:1936:0, 27)
2018-10-09 22:45:43.955 [DBG] HSWC: Tearing down open circuit with SETTLE pkt, removing circuit=(Chan ID=0:0:0, HTLC ID=19003) with keystone=(Chan ID=544281:1936:0, HTLC ID=27)
2018-10-09 22:45:43.967 [DBG] HSWC: Closed completed SETTLE circuit for b5d5bf498e1a8ba0a3e8d05d5d666e00f2943c9f498dfb6a92af8795d98ceaa2: (0:0:0, 19003) <-> (544281:1936:0, 27)
2018-10-09 22:45:43.967 [ERR] HSWC: Cannot find pending payment with ID 19003
```

Payment `b5d5bf498e1a8ba0a3e8d05d5d666e00f2943c9f498dfb6a92af8795d98ceaa2` is absent from listpayments

treygriffith:
@halseth is this issue now resolved with #2761 and #2762 ? Or is there something else that still needs to be fixed?

alexbosworth:
[...]

PR #2075, opened 2018-10-19
contractcourt: settle invoice when claiming HTLC on-chain

wpaulino:
In this PR, we extend the `htlcSuccessResolver` to settle the invoice,
if any, of the corresponding on-chain HTLC sweep. This ensures that the
invoice state is consistent as when claiming the HTLC "off-chain".

Partly addresses #2032 (the receiver side).

### DIFF
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -1084,33 +1084,11 @@ func (c *ChannelArbitrator) prepContractResolutions(htlcActions ChainActionMap,
 	inResolutionMap := make(map[wire.OutPoint]lnwallet.IncomingHtlcResolution)
 	for i := 0; i < len(incomingResolutions); i++ {
 		inRes := incomingResolutions[i]
-
-		// If we have a success transaction, then the htlc's outpoint
-		// is the transaction's only input. Otherwise, it's the claim
-		// point.
-		var htlcPoint wire.OutPoint
-		if inRes.SignedSuccessTx != nil {
-			htlcPoint = inRes.SignedSuccessTx.TxIn[0].PreviousOutPoint
-		} else {
-			htlcPoint = inRes.ClaimOutpoint
-		}
-
-		inResolutionMap[htlcPoint] = inRes
+		inResolutionMap[inRes.HtlcPoint()] = inRes
 	}
 	for i := 0; i < len(outgoingResolutions); i++ {
 		outRes := outgoingResolutions[i]
-
-		// If we have a timeout transaction, then the htlc's outpoint
-		// is the transaction's only input. Otherwise, it's the claim
-		// point.
-		var htlcPoint wire.OutPoint
-		if outRes.SignedTimeoutTx != nil {
-			htlcPoint = outRes.SignedTimeoutTx.TxIn[0].PreviousOutPoint
-		} else {
-			htlcPoint = outRes.ClaimOutpoint
-		}
-
-		outResolutionMap[htlcPoint] = outRes
+		outResolutionMap[outRes.HtlcPoint()] = outRes
 	}
 
 	// We'll create the resolver kit that we'll be cloning for each
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -358,11 +358,106 @@ func (c *ChannelArbitrator) relaunchResolvers() error {
 		return err
 	}
 
+	// Retrieve the commitment tx hash from the log.
+	contractResolutions, err := c.log.FetchContractResolutions()
+	if err != nil {
+		log.Errorf("unable to fetch contract resolutions: %v",
+			err)
+		return err
+	}
+	commitHash := contractResolutions.CommitHash
+
+	// Reconstruct the htlc outpoints and data from the chain action log.
+	// The purpose of the constructed htlc map is to supplement to resolvers
+	// restored from database with extra data. Ideally this data is stored
+	// as part of the resolver in the log. This is a workaround to prevent a
+	// db migration.
+	htlcMap := make(map[wire.OutPoint]*channeldb.HTLC)
+	chainActions, err := c.log.FetchChainActions()
+	if err != nil {
+		log.Errorf("unable to fetch chain actions: %v", err)
+		return err
+	}
+	for _, htlcs := range chainActions {
+		for _, htlc := range htlcs {
+			outpoint := wire.OutPoint{
+				Hash:  commitHash,
+				Index: uint32(htlc.OutputIndex),
+			}
+			htlcMap[outpoint] = &htlc
+		}
+	}
+
 	log.Infof("ChannelArbitrator(%v): relaunching %v contract "+
 		"resolvers", c.cfg.ChanPoint, len(unresolvedContracts))
 
+	for _, resolver := range unresolvedContracts {
+		supplementResolver(resolver, htlcMap)
+	}
+
 	c.launchResolvers(unresolvedContracts)
 
+	return nil
+}
+
+// supplementResolver takes a resolver as it is restored from the log and fills
+// in missing data from the htlcMap.
+func supplementResolver(resolver ContractResolver,
+	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
+
+	switch r := resolver.(type) {
+
+	case *htlcSuccessResolver:
+		return supplementSuccessResolver(r, htlcMap)
+
+	case *htlcIncomingContestResolver:
+		return supplementSuccessResolver(
+			&r.htlcSuccessResolver, htlcMap,
+		)
+
+	case *htlcTimeoutResolver:
+		return supplementTimeoutResolver(r, htlcMap)
+
+	case *htlcOutgoingContestResolver:
+		return supplementTimeoutResolver(
+			&r.htlcTimeoutResolver, htlcMap,
+		)
+	}
+
+	return nil
+}
+
+// supplementSuccessResolver takes a htlcSuccessResolver as it is restored from
+// the log and fills in missing data from the htlcMap.
+func supplementSuccessResolver(r *htlcSuccessResolver,
+	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
+
+	res := r.htlcResolution
+	htlcPoint := res.HtlcPoint()
+	htlc, ok := htlcMap[htlcPoint]
+	if !ok {
+		return errors.New(
+			"htlc for success resolver unavailable",
+		)
+	}
+	r.htlcAmt = htlc.Amt
+	return nil
+}
+
+// supplementTimeoutResolver takes a htlcSuccessResolver as it is restored from
+// the log and fills in missing data from the htlcMap.
+func supplementTimeoutResolver(r *htlcTimeoutResolver,
+	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
+
+	res := r.htlcResolution
+	htlcPoint := res.HtlcPoint()
+	htlc, ok := htlcMap[htlcPoint]
+	if !ok {
+		return errors.New(
+			"htlc for timeout resolver unavailable",
+		)
+	}
+	r.htlcAmt = htlc.Amt
 	return nil
 }
 
@@ -1146,6 +1241,7 @@ func (c *ChannelArbitrator) prepContractResolutions(htlcActions ChainActionMap,
 					htlcResolution:  resolution,
 					broadcastHeight: height,
 					payHash:         htlc.RHash,
+					htlcAmt:         htlc.Amt,
 					ResolverKit:     resKit,
 				}
 				htlcResolvers = append(htlcResolvers, resolver)
@@ -1173,6 +1269,7 @@ func (c *ChannelArbitrator) prepContractResolutions(htlcActions ChainActionMap,
 					htlcResolution:  resolution,
 					broadcastHeight: height,
 					htlcIndex:       htlc.HtlcIndex,
+					htlcAmt:         htlc.Amt,
 					ResolverKit:     resKit,
 				}
 				htlcResolvers = append(htlcResolvers, resolver)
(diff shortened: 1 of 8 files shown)


witness_beacon.go
```diff
@@ -129,5 +130,16 @@ func (p *preimageBeacon) AddPreimage(pre []byte) error {
...
 
+// SettleInvoice attempts to settle an existing invoice with the given payment
+// hash. ErrInvoiceNotFound is returned if an invoice is not found.
+func (p *preimageBeacon) SettleInvoice(paymentHash [32]byte,
+	amt lnwire.MilliSatoshi) error {
+
+	p.Lock()
+	defer p.Unlock()
+
+	return p.invoices.SettleInvoice(paymentHash, amt)
```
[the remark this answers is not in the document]
joostjager: It is currently just a pass-through to InvoiceRegistry, so maybe it is better to make InvoiceRegistry a dependency of ChainArbitratorConfig so that the resolver can access it.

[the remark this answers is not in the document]
wpaulino: Fixed - this made the changes even simpler.

contractcourt/contract_resolvers.go
```diff
@@ -522,8 +523,24 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 			return nil, fmt.Errorf("quitting")
 		}
 
+		// Finally, with the HTLC claimed, we'll attempt to settle its
+		// corresponding invoice if we were the original destination.
+		// We'll retrieve the preimage from the incoming HTLC resolution
+		// and settle the invoice with the output's amount, taking into
+		// account on-chain fees (if any).
+		paymentHash := sha256.Sum256(h.htlcResolution.Preimage[:])
```
joostjager: I meant even simpler, `paymentHash := h.payHash`

contractcourt/contract_resolvers.go
```diff
@@ -585,6 +602,21 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 		return nil, fmt.Errorf("quitting")
 	}
 
+	// Finally, with the HTLC claimed, we'll attempt to settle its
+	// corresponding invoice if we were the original destination. We'll
+	// retrieve the preimage from the success transaction's witness and
+	// settle the invoice with the output's amount, taking into account
+	// on-chain fees (if any).
+	paymentHash := sha256.Sum256(h.htlcResolution.Preimage[:])
+	amt := btcutil.Amount(h.htlcResolution.SignedSuccessTx.TxOut[0].Value)
```
joostjager: Here we take as amount the output value of the success tx, which is the paid amount minus the success tx fee. 

This amount will be reduced further by the nursery sweep which needs fees too, before the funds are back in the wallet. Nursery may sweep multiple inputs in a single tx, so difficult to attribute (part) of the fee to this invoice payment.

However, in the remote commitment case above, as invoice amount the paid value minus the sweep fee is recorded. Looks inconsistent? Also it will be a problem when that sweep is (soon) also combined with other inputs.

From the perspective of the sender of the payment, subtracting any on chain fees may not be justifiable. It may look like not the full amount has been paid, while it isn't his fault that the second-last hop caused channel closure.

lnd_test.go
```diff
@@ -9092,6 +9089,20 @@ func testMultiHopReceiverChainClaim(net *lntest.NetworkHarness, t *harnessTest)
...
+	if err != nil {
+		t.Fatalf("unable to retrieve invoices: %v", err)
+	}
+	if len(invoicesResp.Invoices) != 1 {
+		t.Fatalf("expected 1 invoice, got %d", len(invoicesResp.Invoices))
+	}
+	if !invoicesResp.Invoices[0].Settled {
+		t.Fatalf("expected invoice to be settled")
+	}
+
```
joostjager: You could consider adding test coverage on the unit level. In general, I myself prefer to test things only in an integration test if there is really no other option. 

I have an (unmerged) contract resolver test for the outgoingcontestresolver, https://github.com/lightningnetwork/lnd/pull/1875/commits/33c348b369143d07132d3b015478f0811d24deed. 

But I can also imagine that you consider this too much effort in comparison to the size of the change.

contractcourt/contract_resolvers.go
wpaulino: Fixed.

contractcourt/contract_resolvers.go
wpaulino: Agreed - fixed. Now we'll use just the value of the output we're sweeping directly. One thing I can see come up in the future is node operators wondering where the remaining amount went to when sweeping the HTLC on-chain. Currently (I think), there isn't a way to distinguish an invoice being settled off-chain vs on-chain.

lnd_test.go
wpaulino: I opted for the integration test change as we're interested in making sure that the invoice state is reflected correctly within the database. I guess we could have this done at the unit test level though. Let me know if you'd like to proceed with that instead.

contractcourt/contract_resolvers.go
joostjager: I think it is safer to settle the invoice with the amount on the commitment tx, excluding both stage 1 and stage 2 fees. To indeed prevent invoices from marked as settled in the db with a value less than the original invoice amt. But we should do that in both the local and remote close case.

lnd_test.go
```diff
@@ -9118,6 +9115,20 @@ func testMultiHopReceiverChainClaim(net *lntest.NetworkHarness, t *harnessTest)
...
+	// on-chain.
+	invoicesReq := &lnrpc.ListInvoiceRequest{}
+	invoicesResp, err := carol.ListInvoices(ctxb, invoicesReq)
+	if err != nil {
+		t.Fatalf("unable to retrieve invoices: %v", err)
+	}
+	if len(invoicesResp.Invoices) != 1 {
+		t.Fatalf("expected 1 invoice, got %d", len(invoicesResp.Invoices))
+	}
+	if !invoicesResp.Invoices[0].Settled {
```
Roasbeef: For future proofing, it should start to use the new invoice state field recently added.

lnd_test.go
```diff
@@ -9118,6 +9115,20 @@ func testMultiHopReceiverChainClaim(net *lntest.NetworkHarness, t *harnessTest)
 		t.Fatalf(predErr.Error())
 	}
 
+	// The invoice should show as settled for Carol, even if it was swept
```
Roasbeef: The remote chain claim test should also be updated with this assertion.

contractcourt/contract_resolvers.go
Roasbeef: IMO it's correct to use the output amount directly, and not the original value of the output within the commitment. If we use the original HTLC value, then from an external accounting point of view, we create an inconsistent state. We'd show the HTLC as being settled with a value of 1k satoshis, but then by the time we actually sweep it, it may be 500 or so satoshis (just an example). We already have the `AmtPaid` field in the database and can use that along with a new settle state `ContractSettledChain` to indicate that we had to go on chain to sweep it. Once the resolvers are fully connected to the sweeper, then the amount swept is even hazier as part of the HTLC amount will go to fees for a batched sweep.

wpaulino: > Once the resolvers are fully connected to the sweeper, then the amount swept is even hazier as part of the HTLC amount will go to fees for a batched sweep.

In this case, how would we go about determining the settled amount (HTLC amount - chain fees)? We could weigh all the inputs and determine the settled amount with a ratio, but not sure if that's worth the hassle.

joostjager: Would the `AmtPaid` field contain the original HTLC value then? I am worried that if we don't do that, it could be that the sender paid his invoice in full, but the final amount was lower because the receiver needed to go on chain. This is to be blamed on an intermediate node, but it may look like the sender didn't fulfill his obligation.

wpaulino: That's why we would add a new settle state `ContractSettledChain` to signal this to the receiver. We can also make it clear in the docs that when an invoice has been settled with this state, the amount received will be less due to chain fees.

Roasbeef: Yeah to circle back on this we'd add a new settle state, as generally accurate fee accounting in the face of batching is a whole project in itself.

joostjager: I now realize that once we do hold invoice, we'll also need to implement the accepted state in the contract resolvers. That'll be another settled state then, `ContractAcceptedChain`.

joostjager: Another way to look at it, could be similar to traditional banking: I pay various service fees to my bank and I book them as a separate cost. I am not trying to attribute those to the individual invoices I sent out. Could we do it similarly for lnd? Keep track of a global costs administration that includes chain fees and book invoices with the original htlc value?

Roasbeef: Circled back on this and we decided to just use the full HTLC value for now. We'll then defer some sort of fee audit functionality for a channel that can tell you the amount paid in on-chain fees from open to close of the channel to a follow up PR.

lnd_test.go
wpaulino: Fixed.

contractcourt/contract_resolvers.go
wpaulino: Fixed. Realized we can't even use the HTLC value as we don't persist it on disk and there doesn't seem to be a way to retrieve it from what's currently stored (something we might want to store in the future perhaps?). As an alternative, I went with just settling the invoice with whatever it originally requested. I believe this is still fine to use as the value is inaccurate anyway.

wpaulino:
Comments addressed, PTAL @Roasbeef @joostjager

contractcourt/contract_resolvers.go
```diff
@@ -530,8 +530,17 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 			return nil, fmt.Errorf("quitting")
 		}
 
+		// Finally, with the HTLC claimed, we'll attempt to settle its
+		// corresponding invoice if we were the original destination.
+		err = h.SettleInvoiceOnChain(h.payHash)
```
joostjager: Isn' t the original value in `h.htlcResolution.SweepSignDesc.Output.Value`  ?

joostjager: I looked into it, it for the local force case, it is the success tx amount, which is less than the htlc amount. For the remote case below, the number is the original htlc value.

But also for the local force case, the info doesn't need to be stored in the database. If I understand it correctly, `ChannelArbitrator.activeHTLCs` contains them. The original HTLC object could possibly be passed into this `Resolve` function.

I think it is important to not take the shortcut here by using the invoice amt.

joostjager: This data is also needed in https://github.com/lightningnetwork/lnd/pull/1875.

See commit `cnct: make original htlc amt available for contract resolvers`

Roasbeef: Why is the invoice amount a short cut? Isn't that what we want, the full value that was requested? The HTLC outputs for second level transactions are smaller than the _actual_ HTLC value since we pay fees from the HTLC value for that transaction. If the value after we subtract fees is below dust, then we consider that a dust output all together.

channeldb/invoices.go
```diff
@@ -81,18 +81,29 @@ const (
...
 
-	// ContractSettled means the htlc is settled and the invoice has been
-	// paid.
+	// ContractSettled represents a HTLC that has been claimed off-chain to
+	// settle an invoice.
 	ContractSettled ContractState = 1
+
+	// ContractSettledOnChain represents a HTLC that has been claimed
+	// on-chain to settle an invoice.
+	ContractSettledOnChain = 2
```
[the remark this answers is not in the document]
Roasbeef: Yeah during our discussion we more or less settled on the lack of a new state, in favor of keeping things simple in the short term, and working out a more comprehensive fee report for a channel.

contractcourt/contract_resolvers.go
joostjager: > I went with just settling the invoice with whatever it originally requested

This is not the amount paid. It could be that more was paid.

Roasbeef: Ah yeah very true, over looked that!

Roasbeef: Final conclusion is that we can get this data from the logged `ChainActions`. They contain the original set of `channeldb.HTLC`s.

joostjager: For implementation, see commit above. Now using `ChainActions`.

channeldb/invoices.go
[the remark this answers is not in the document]
wpaulino: State removed.

contractcourt/contract_resolvers.go
wpaulino: Sweet, thanks @joostjager! I believe this should be good to go now.

contractcourt/htlc_success_resolver.go
```diff
@@ -174,6 +176,14 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 			return nil, fmt.Errorf("quitting")
 		}
 
+		// With the HTLC claimed, we can attempt to settle its
+		// corresponding invoice if we were the original destination.
```
joostjager: We are the original destination, aren't we?

joostjager:
I still think a unit test would've been more appropriate for this PR (see earlier comment above), but if a second reviewer approves, I am good.

Other than that, LGTM

contractcourt/channel_arbitrator.go
```diff
@@ -353,14 +353,105 @@ func (c *ChannelArbitrator) relaunchResolvers() error {
...
+	case *htlcOutgoingContestResolver:
+		return supplementTimeoutResolver(
+			&r.htlcTimeoutResolver, htlcMap,
+		)
+	}
+
+	return nil
+}
+
+func supplementSuccessResolver(r *htlcSuccessResolver,
```
Roasbeef: Missing `godoc` comment here and below.

contractcourt/htlc_success_resolver.go
Roasbeef: Either we're the final destination, or we're just an intermediate node that happend to already go on-chain for the incoming HTLC (the outgoing was pulled off-chain).

contractcourt/channel_arbitrator.go
wpaulino: Fixed.

contractcourt/htlc_success_resolver.go
joostjager: Ah yes, right.

PR #2080, opened 2018-10-22
peer: send reenable updates after 20m

cfromknecht:
This commit adds a short delay before sending
out channel reenable messages. The aim is to
prevent flooding the network if we have
flapping peers. The approach ensures that
reenabling will only occur within a peer's
lifecycle, which can prevent lingering
goroutines from interfering with
enable/disable attempts across different
connections with the same peer.

### DIFF
--- a/config.go
+++ b/config.go
@@ -31,28 +31,28 @@ import (
 )
 
 const (
-	defaultConfigFilename      = "lnd.conf"
-	defaultDataDirname         = "data"
-	defaultChainSubDirname     = "chain"
-	defaultGraphSubDirname     = "graph"
-	defaultTLSCertFilename     = "tls.cert"
-	defaultTLSKeyFilename      = "tls.key"
-	defaultAdminMacFilename    = "admin.macaroon"
-	defaultReadMacFilename     = "readonly.macaroon"
-	defaultInvoiceMacFilename  = "invoice.macaroon"
-	defaultLogLevel            = "info"
-	defaultLogDirname          = "logs"
-	defaultLogFilename         = "lnd.log"
-	defaultRPCPort             = 10009
-	defaultRESTPort            = 8080
-	defaultPeerPort            = 9735
-	defaultRPCHost             = "localhost"
-	defaultMaxPendingChannels  = 1
-	defaultNoSeedBackup        = false
-	defaultTrickleDelay        = 30 * 1000
-	defaultInactiveChanTimeout = 20 * time.Minute
-	defaultMaxLogFiles         = 3
-	defaultMaxLogFileSize      = 10
+	defaultConfigFilename     = "lnd.conf"
+	defaultDataDirname        = "data"
+	defaultChainSubDirname    = "chain"
+	defaultGraphSubDirname    = "graph"
+	defaultTLSCertFilename    = "tls.cert"
+	defaultTLSKeyFilename     = "tls.key"
+	defaultAdminMacFilename   = "admin.macaroon"
+	defaultReadMacFilename    = "readonly.macaroon"
+	defaultInvoiceMacFilename = "invoice.macaroon"
+	defaultLogLevel           = "info"
+	defaultLogDirname         = "logs"
+	defaultLogFilename        = "lnd.log"
+	defaultRPCPort            = 10009
+	defaultRESTPort           = 8080
+	defaultPeerPort           = 9735
+	defaultRPCHost            = "localhost"
+	defaultMaxPendingChannels = 1
+	defaultNoSeedBackup       = false
+	defaultTrickleDelay       = 30 * 1000
+	defaultChanStatusInterval = 20 * time.Minute
+	defaultMaxLogFiles        = 3
+	defaultMaxLogFileSize     = 10
 
 	defaultTorSOCKSPort            = 9050
 	defaultTorDNSHost              = "soa.nodes.lightning.directory"
@@ -229,8 +229,8 @@ type config struct {
 
 	NoSeedBackup bool `long:"noseedbackup" description:"If true, NO SEED WILL BE EXPOSED AND THE WALLET WILL BE ENCRYPTED USING THE DEFAULT PASSPHRASE -- EVER. THIS FLAG IS ONLY FOR TESTING AND IS BEING DEPRECATED."`
 
-	TrickleDelay        int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
-	InactiveChanTimeout time.Duration `long:"inactivechantimeout" description:"If a channel has been inactive for the set time, send a ChannelUpdate disabling it."`
+	TrickleDelay       int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
+	ChanStatusInterval time.Duration `long:"chanstatusinterval" description:"The duration that should elapse between detecting that a channel has been enabled or disabled and announcing the updated status to the network."`
 
 	Alias       string `long:"alias" description:"The node alias. Used as a moniker by peers and intelligence services"`
 	Color       string `long:"color" description:"The color of the node in hex format (i.e. '#3399FF'). Used to customize node appearance in intelligence services"`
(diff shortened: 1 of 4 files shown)


Roasbeef:
`announceChanStatus` will already delay this, and only enable if the state hasn't changed in the last 20 minutes. In addition, this is an incomplete solution as we need to decouple enable/disable in our local graph, from the announcement we send out to the network. We should immediately enable within our graph, but delay the announcement until we reach a stable point. cc @halseth

cfromknecht:
@Roasbeef correct, this does not address the decoupling of local graph from network gossip, but neither is that the intention of this PR.

The purpose of this PR is
  1. not enable immediately to avoid flooding in case the peer is unstable.
  2. to remove the stray goroutine by ensuring reenabling happens within the peer lifecycle, as this can properly "cancel" the reenable if the peer disconnects before stability is reached.

> announceChanStatus will already delay this, and only enable if the state hasn't changed in the last 20 minutes

where does this happen? to me, it looks like `announceChanStatus` signs and sends a new update to the gossiper immediately, which is then only delayed by 30s?

> this is an incomplete solution as we need to decouple enable/disable in our local graph, from the announcement we send out to the network

again, that is mostly orthogonal to this PR. decoupling them allows us greater flexibility in picking timeouts for each. After decoupling, we are free to increase this timeout before sending out a new announcement.

Additionally, just letting `channelStatusWatcher` reenable has the downside that it may take up to almost twice the intended timeout before the channel is actually reenabled, where as this will reliably reenable after the chosen timeout.

halseth:
Yep, https://github.com/lightningnetwork/lnd/pull/2115 needs to go in first though.

halseth:
With #2115 merged I think this PR can replace #2091 if we set the `channelReannounceDelay` to `InactiveChanTimeout` (or create a new config option).

cfromknecht:
Created a new config option called `ChanStatusInterval`, PTAL

peer.go
```diff
@@ -1723,6 +1702,16 @@ out:
...
 			p.finalizeChanClosure(chanCloser)
+
+		// The channel reannounce delay has elapsed, broadcast the
+		// reenabled channel updates to the network. This should only
+		// fire once, so we set the dispatchStatuses channel to nil to
+		// mark it for garbage collection. If the peer is torn down
+		// before firing, reenabling will not be attempted.
+		case <-reenableTimeout:
+			p.reenableActiveChannels()
+			reenableTimeout = nil
```
halseth: is nilling necessary?

config.go
```diff
@@ -227,18 +226,15 @@ type config struct {
...
+	ChanStatusInterval time.Duration `long:"chanstatusinterval" description:"The duration that should elapse between detecting that a channel has been enabled or disabled and announcing the updated status to the network."`
 
 	Alias       string `long:"alias" description:"The node alias. Used as a moniker by peers and intelligence services"`
 	Color       string `long:"color" description:"The color of the node in hex format (i.e. '#3399FF'). Used to customize node appearance in intelligence services"`
 	MinChanSize int64  `long:"minchansize" description:"The smallest channel size (in satoshis) that we should accept. Incoming channels smaller than this will be rejected"`
 
 	NoChanUpdates bool `long:"nochanupdates" description:"If specified, lnd will not request real-time channel updates from connected peers. This option should be used by routing nodes to save bandwidth."`
 
-	RejectPush bool `long:"rejectpush" description:"If true, lnd will not accept channel opening requests with non-zero push amounts. This should prevent accidental pushes to merchant nodes."`
```
halseth: needs a rebase

server.go
```diff
@@ -3078,7 +3078,7 @@ func (s *server) watchChannelStatus() {
 	status := make(map[wire.OutPoint]activeStatus)
 
 	// We'll check in on the channel statuses every 1/4 of the timeout.
-	unchangedTimeout := cfg.InactiveChanTimeout
+	unchangedTimeout := cfg.ChanStatusInterval
```
halseth: nit: just do the variable name change and this as the first commit?

cfromknecht: was trying to do it in a way such that the intermediate commits would build :P

peer.go
cfromknecht: shaves some CPU cycles from select, figured why not?

server.go
halseth: Haha, yeah I think if you do the `config+server` change it will build 😀

halseth:
Needs a rebase now that #2115 is in!

peer.go
halseth: adds some CPU cycles in my head at least

halseth: Why btw will it save CPU cycles? I think maybe that is only for tickers.

cfromknecht: because the runtime doesn't have to acquire a lock to check if a value is being sent on a nil channel, it just skips it. it saves CPU cycles on every subsequent select

cfromknecht:
```
func main() {
        const n = 1000000
        const numChans = 10

        chans := make([]chan int, numChans)
        for i := range chans {
                chans[i] = make(chan int)
        }   

        start := time.Now()
        for i := 0; i < n; i++ {
                select {
                case <-chans[0]:
                case <-chans[1]:
                case <-chans[2]:
                case <-chans[3]:
                case <-chans[4]:
                case <-chans[5]:
                case <-chans[6]:
                case <-chans[6]:
                case <-chans[8]:
                case <-chans[9]:
                default:
                }   
        }   
        fmt.Printf("non-nil: %v\n", time.Since(start))

        for i := range chans {
                chans[i] = nil 
        }   

        start = time.Now()
        for i := 0; i < n; i++ {
                select {
                case <-chans[0]:
                case <-chans[1]:
                case <-chans[2]:
                case <-chans[3]:
                case <-chans[4]:
                case <-chans[5]:
                case <-chans[6]:
                case <-chans[6]:
                case <-chans[8]:
                case <-chans[9]:
                default:
                }   
        }   
        fmt.Printf("nil: %v\n", time.Since(start))
}
```
running this gives, OMM:
```
non-nil: 521.206479ms
nil: 155.758026ms
```

config.go
cfromknecht: fixed

peer.go
```diff
@@ -1753,6 +1732,16 @@ out:
 			// relevant sub-systems and launching a goroutine to
 			// wait for close tx conf.
 			p.finalizeChanClosure(chanCloser)
+
+		// The channel reannounce delay has elapsed, broadcast the
+		// reenabled channel updates to the network. This should only
+		// fire once, so we set the dispatchStatuses channel to nil to
```
wpaulino: s/dispatchStatuses/reenableTimeout

peer.go
cfromknecht: In case anyone was interested, I modified the demo to show how the latency is linearly dependent on the number of nil channels:
```
func main() {
        const n = 1000000
        const numChans = 10

        chans := make([]chan int, numChans)
        for i := range chans {
                chans[i] = make(chan int)
        }   

        testNNil := func(ii int) {
                start := time.Now()
                for i := range chans[:ii] {
                        chans[i] = nil 
                }   

                start = time.Now()
                for i := 0; i < n; i++ {
                        select {
                        case <-chans[0]:
                        case <-chans[1]:
                        case <-chans[2]:
                        case <-chans[3]:
                        case <-chans[4]:
                        case <-chans[5]:
                        case <-chans[6]:
                        case <-chans[6]:
                        case <-chans[8]:
                        case <-chans[9]:
                        default:
                        }   
                }   
                fmt.Printf("%d-nil: %v\n", ii, time.Since(start))
        }   

        for i := 0; i < 11; i++ {
                testNNil(i)
        }   
}
```
outputs:
```
0-nil: 519.117008ms
1-nil: 505.169358ms
2-nil: 485.513103ms
3-nil: 456.82318ms
4-nil: 430.449548ms
5-nil: 387.429949ms
6-nil: 340.134363ms
7-nil: 236.290499ms
8-nil: 238.719919ms
9-nil: 191.990285ms
10-nil: 157.386867ms
```

halseth: Wow, nice! TIL :)

halseth:
Another rebase needed 🤣

peer.go
```diff
@@ -1548,6 +1522,11 @@ func (p *peer) genDeliveryScript() ([]byte, error) {
 func (p *peer) channelManager() {
 	defer p.wg.Done()
 
+	// reenableTimeout will fire once after the configured channel status
+	// interval  has elapsed. This will trigger us to sign new channel
+	// updates and broadcast them with the "disabled" flag unset.
+	reenableTimeout := time.After(cfg.ChanStatusInterval)
```
Roasbeef: Shouldn't it be possible for the existing `watchChannelStatus` method to also take on this responsibility? As is right now, it's possible for both of these goroutines to conflict, and cause a double-ish re-enable for channels. 

Generally, I find the existing logic in `watchChannelStatus` hard to follow w.r.t the precise delay between enable and disable. I wouldn't be against abstracting it further to isolate to its own package so we can properly unit test it (with necessary interfaces added).

cfromknecht: we decided on not using long-polling for reenable as it can lead to false positives. agree with the comments on the existing logic, i will extend this PR to refactor that to only be responsible for 1) dampening reenables and 2) disable still-enabled channels after the configured interval via long polling

halseth: I think it would be cleanest to make the long polling only responsible for disabling channels.

Since we poll to disable, and require 20 minutes (configurable) steady connection to enable, I don't think they should conflict?

cfromknecht:
Closing since this was replaced by the #2411

PR #2081, opened 2018-10-23
lnrpc+lnd+rpc: add new Signer RPC service, and sub RPC server infrastructure 

Roasbeef:
In this commit, we add a new sub RPC server, the SignerService, along with infrastructure to allow developers to easily create these new sub-servers in the future. With a goal of making thighs fully generalized, and new RPC service (after this RPC), only needs to do the following: 
  1. Add a new package, and within the `init()` method call `lnrpc.RegisterSubServer``
  2. Add a new case statement within the new sub-server config parsing to properly set the arguments of the config of the sub-server. 

In order to achieve the above, we utilize the reflect package and build flags heavily throughout the scaffolding. 

## New `signrpc` Package

In this PR, we introduce a new sub-package within the greater RPC package.  This new sub-package will house a new set of sub-RPC servers to expose experimental features behind build flags for upstream consumers. In this commit, we add the first config for the service, which will simply expose the lnwallet.Signer interface over RPC.

In the default file, we have what the config will be if the build tag (signerrpc) is off. In this case, the config parser won't detect any times, and if specified will error out. In the active file, we have the true config that the server will use. With this new set up, we'll exploit these build flags heavily in order to create a generalized framework for adding additional sub RPC servers.

The package also defines a new tool for generating protos, and also the new protos for the SIgner service itself. 

## Sub RPC Server Registration System 

In this PR, we add the scaffolding for the future sub-server RPC system. The idea is that each sub server will implement this particular interface. From there on, a "root" RPC server is able to query this registry, and dynamically create each  sub-sever instance without knowing the details of each sub-server.

In the init() method of the package of a sub-server, the sub-server is to call: RegisterSubServer to claim its namespace. Afterwards, the root RPC server can use the RegisteredSubServers() method to obtain a slice of ALL registered sub-servers. Once this list is obtained, it can use the New() method of the SubServerDriver struct to create a new sub-server instance.

Each sub-server needs to be able to locate it's primary config using the SubServerConfigDispatcher interface. This can be a map of maps, or a regular config struct. The main requirement is that the sub-server be able to find a config under the same name that it registered with. This string of abstractions will allow the main RPC server to find, create, and run each sub-server without knowing the details of its configuration or its role.

## New SignerServer gRPC Service Implementation 

In this PR, we add a full implementation of the new SignerServer sub RPC service within the main root RPC service. This service is able to fully manage its macaroons, and service any connected clients. Atm, this service only has a single method: SignOutputRaw which mimics the existing `lnwallet.Signer` interface within lnd itself. As the API's are so similar, it will be possible for a client to directly use the `lnwallet.Signer` interface, and have a proxy that sends the request over RPC, and translates the proto layer on both sides. To the client, it doesn't know that it's using a remote, or local RPC.

The server uses the new driver registration system to register itself as a relevant sub RPC server at the `lnrpc` package level. We also do a bit of validation on the config passed in, in order to ensure it has all the items we need to function. 

## Dynamic Sub Server Config Parsing 

In this commit, we add the glue infrastructure to make the sub RPC server system work properly. Our high level goal is the following: using only the lnrpc package (with no visibility into the sub RPC servers), the RPC server is able to find, create, run, and manage the entire set of present and future sub RPC servers. In order to achieve this, we use the reflect package and build tags heavily to permit a loosely coupled configuration parsing system for the sub RPC servers.

We start with a new `subRpcServerConfigs` struct which is _always_ present.  This struct has its own group, and will house a series of sub-configs, one for each sub RPC server. Each sub-config is actually gated behind a build flag, and can be used to allow users on the command line or in the config to specify arguments related to the sub-server. If the config isn't present, then we don't attempt to parse it at all, if it is, then that means the RPC server has been registered, and we should parse the contents of its config.

The `subRpcServerConfigs` struct has two main methods: `PopulateDependancies` and `FetchConfig`. The `PopulateDependancies` is used to dynamically locate and set the config fields for each new sub-server. As the config may not actually have any fields (if the build flag is off), we use the reflect pacakge to determine if things are compiled in or not, then if so, we dynamically set each of the config parameters. The `PopulateDependancies` method implements the `lnrpc.SubServerConfigDispatcher` interface. Our goal is to allow sub servers to look up their actual config in this main config struct. We achieve this by using reflect to look up the target field _as if it were a key in a map_. If the field is found, then we check if it has any actual attributes (it won't if the build flag is off), if it is, then we return it as we expect it to be populated already.

## The Current RPC Server as the Root gRPC Server 

In this PR, we modify the existing rpcServer to fully manage the macaroons, gRPC server, and also seek out and create all sub-servers. With this change, the RPC server gains more responsibility, as it becomes the "root" server in the hierarchy of gRPC sub-servers.

In addition to creating each sub-server, it will also merge the set of macaroon permissions for each sub-server, with the permissions of the rest of the RPC infra. As a result, each sub-server is able to
independently specify what it needs w.r.t macaroon permissions and have that taken care of by the RPC server. In order to achieve this, we need to unify the creation of the RPC interceptors, and also fully manage the gRPC server ourselves.

As a bonus, on shutdown, the RPC server will now call `rpcServer.GracefulStop()`, which will stop it from accepting new connections, and wait for existing RPC's to execute before shutting down the entire server.

Some examples with various build configs:
```
⛰i  make build
 Building debug lnd and lncli.
go build -v -tags="dev" -o lnd-debug -ldflags "-X github.com/lightningnetwork/lnd/build.Commit=v0.5-beta-143-gb2069914c4b76109b7c59320dc48f8a5f30deb75-dirty" github.com/lightningnetwork/lnd
go build -v -tags="dev" -o lncli-debug -ldflags "-X github.com/lightningnetwork/lnd/build.Commit=v0.5-beta-143-gb2069914c4b76109b7c59320dc48f8a5f30deb75-dirty" github.com/lightningnetwork/lnd/cmd/lncli

⛰i  ./lnd-debug --debuglevel=debug --signrpc.signermacaroonpath=~/sign.macaroon
unknown flag `signrpc.signermacaroonpath'
unknown flag `signrpc.signermacaroonpath'

⛰i  make build tags=signerrpc
 Building debug lnd and lncli.
go build -v -tags="dev signerrpc" -o lnd-debug -ldflags "-X github.com/lightningnetwork/lnd/build.Commit=v0.5-beta-143-gb2069914c4b76109b7c59320dc48f8a5f30deb75-dirty" github.com/lightningnetwork/lnd
go build -v -tags="dev signerrpc" -o lncli-debug -ldflags "-X github.com/lightningnetwork/lnd/build.Commit=v0.5-beta-143-gb2069914c4b76109b7c59320dc48f8a5f30deb75-dirty" github.com/lightningnetwork/lnd/cmd/lncli

⛰i  ./lnd-debug --debuglevel=debug --signrpc.signermacaroonpath=~/sign.macaroon
2018-10-22 17:31:01.132 [INF] LTND: Version: 0.5.0-beta commit=v0.5-beta-143-gb2069914c4b76109b7c59320dc48f8a5f30deb75-dirty, build=development, logging=default
2018-10-22 17:31:01.133 [INF] LTND: Active chain: Bitcoin (network=simnet)
2018-10-22 17:31:01.140 [INF] CHDB: Checking for schema update: latest_version=6, db_version=6
2018-10-22 17:31:01.236 [INF] LTND: Primary chain is set to: bitcoin
2018-10-22 17:31:02.391 [INF] LNWL: Opened wallet
2018-10-22 17:31:03.315 [INF] LNWL: The wallet has been unlocked without a time limit
2018-10-22 17:31:03.315 [INF] LTND: LightningWallet opened
2018-10-22 17:31:03.319 [INF] LNWL: Catching up block hashes to height 3060, this will take a while...
2018-10-22 17:31:03.320 [INF] HSWC: Restoring in-memory circuit state from disk
2018-10-22 17:31:03.320 [INF] LNWL: Done catching up block hashes
2018-10-22 17:31:03.320 [INF] HSWC: Payment circuits loaded: num_pending=0, num_open=0
2018-10-22 17:31:03.322 [DBG] LTND: Populating dependencies for sub RPC server: Signrpc
```

As for the config, an example is:
```
[signrpc]
signrpc.signermacaroonpath=~/signer.macaroon
```


subrpcserver_config.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		if !fieldElem.CanSet() {
+			continue
+		}
+
+		switch cfg := field.Interface().(type) {
+		case *signrpc.Config:
+			subCfgValue := extractReflectValue(cfg)
+
+			subCfgValue.FieldByName("MacService").Set(
```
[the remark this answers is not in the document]
Roasbeef: Nope, they won't always exist if the build tag is off. This _always_ compiles regardless of if the build tags are on or not.

subrpcserver_config.go
```diff
@@ -0,0 +1,128 @@
+package main
```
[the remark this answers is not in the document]
Roasbeef: At a future point possibly? Things aren't yet decoupled enough to make the move.

subrpcserver_config.go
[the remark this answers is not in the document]
cfromknecht: Gotcha, makes sense

subrpcserver_config.go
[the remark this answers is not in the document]
cfromknecht: the imports in this file are minimal and usage seems to be done exclusively through public methods, which part isn't decoupled?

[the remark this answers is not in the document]
cfromknecht: chain control?

rpcserver.go
```diff
@@ -339,19 +345,137 @@ type rpcServer struct {
...
+	lnrpc.RegisterLightningServer(grpcServer, rootRpcServer)
+
+	// Now the main RPC server has been registered, we'll iterate through
+	// all the sub-RPC servers and register them to ensure that requests
+	// are properly routed towards them.
+	for _, subServer := range subServers {
+		err := subServer.RegisterWithRootServer(grpcServer)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register "+
+				"sub-server %v with root: %v", err)
```
halseth: missing first param

rpcserver.go
```diff
@@ -339,19 +345,137 @@ type rpcServer struct {
...
+
+	var (
+		subServers     []lnrpc.SubServer
+		subServerPerms []lnrpc.MacaroonPerms
+	)
+
+	// Before we create any of the sub-servers, we need to ensure that all
+	// the dependencies they need are properly populated within each sub
+	// server configuration struct.
+	err := subServerCgs.PopulateDependancies(
```
halseth: Why is this level of indirection needed? Seems like we could just pass these parameters to `subserver.New()` or have a common `subserverConfig` passed to `New()` instead of using reflection.

subrpcserver_config.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/lightningnetwork/lnd/macaroons"
+)
+
+// subRpcServerConfigs is special sub-config in the main configuration that
+// houses the configuration for the optional sub-servers. These sub-RPC servers
+// are meant to house experimental new features that may eventually make it
+// into the main RPC server that lnd exposes. Special methods are present on
+// this struct to allow the main RPC server to create and manipulate the
+// sub-RPC servers in a generalized manner.
+type subRpcServerConfigs struct {
```
joostjager: I think using a separate config file per sub server in the `.lnd` directory is simpler. No reflection needed and less references from lnd to the sub services. Are there serious downsides? I can even imagine a separate config to be more in line with the plan to unbundle lnd separate processes.

lnrpc/signrpc/driver.go
```diff
@@ -0,0 +1,66 @@
...
+			"subserver %s, expected %T got %T", subServerName,
+			&Config{}, signServerConf)
+	}
+
+	// Before we try to make the new signer service instance, we'll perform
+	// some sanity checks on the arguments to ensure taht they're useable.
+	switch {
+	case config.MacService == nil:
+		return nil, nil, fmt.Errorf("MacService must be set to create " +
+			"Signrpc")
```
joostjager: This doesn't allow running with `NoMacaroons`

lnd.go
```diff
@@ -602,9 +556,9 @@ func genCertPair(certFile, keyFile string) error {
 	return nil
 }
 
-// genMacaroons generates three macaroon files; one admin-level, one
-// for invoice access and one read-only. These can also be used
-// to generate more granular macaroons.
+// genMacaroons generates three macaroon files; one admin-level, one for
+// invoice access and one read-only. These can also be used to generate more
+// granular macaroons.
 func genMacaroons(ctx context.Context, svc *macaroons.Service,
```
joostjager: How to generate an admin macaroon that can access all sub services? In signer, a new entity and action is introduced. I think admin should be able to have permissions to access signer?

lnrpc/signrpc/driver.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	return New(config)
+}
+
+func init() {
+	subServer := &lnrpc.SubServerDriver{
+		SubServerName: subServerName,
+		New: func(c lnrpc.SubServerConfigDispatcher) (lnrpc.SubServer, lnrpc.MacaroonPerms, error) {
```
cfromknecht: nit: long line

lnrpc/signrpc/gen_protos.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+protoc -I/usr/local/include -I. \
```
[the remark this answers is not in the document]
halseth: Example on the recursive version: https://github.com/halseth/lnd/commit/f3f81294462494ed81b6010631342622ca9c44a9

lnrpc/signrpc/config_active.go
```diff
@@ -0,0 +1,33 @@
+// +build signerrpc
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

lnrpc/signrpc/log.go
```diff
@@ -0,0 +1,42 @@
...
+
+// The default amount of logging is none.
+func init() {
+	DisableLog()
+}
+
+// DisableLog disables all library log output.  Logging output is disabled
+// by default until UseLogger is called.
+func DisableLog() {
+	log = btclog.Disabled
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

rpcserver.go
```diff
@@ -339,19 +345,137 @@ type rpcServer struct {
...
+	// listenerCleanUp are a set of closures functions that will allow this
+	// main RPC server to clean up all the listening socket created for the
+	// server.
+	listenerCleanUp []func()
+
+	// restServerOpts are a set of gRPC dial options that the REST server
+	// proxy will use to connect to the main gRPC server.
+	restServerOpts []grpc.DialOption
+
+	// tlsCfg is the TLS configu that allows the REST server proxy to
```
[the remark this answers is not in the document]
Roasbeef: dunzo

subrpcserver_config.go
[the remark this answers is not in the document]
Roasbeef: As in, there're gonna be future sub-servers added in the next week or so. At that point, we'll have a good idea w.r.t all their dependancies and can make one global move to integrate them all. 

I lean against putting it in the `lnrpc` package though, as this also contains items which will be parsed from either the command line, or the the config files.

lnrpc/signrpc/config_default.go
```diff
@@ -0,0 +1,10 @@
+// +build !signerrpc
+
+package signrpc
+
+// Config is the primary configuration struct for the signer RPC server. It
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

lnrpc/signrpc/signer.proto
```diff
@@ -0,0 +1,116 @@
...
+    /// The family of key being identified.
+    int32 key_family = 1;
+
+    /// The precise index of the key being identified.
+    int32 key_index = 2;
+}
+
+message KeyDescriptor {
+     /**
+     The raw bytes of the key being identified. Either this or the KeyLocator
```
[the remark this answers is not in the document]
Roasbeef: Both _can_ be specified FWIW. `oneof`s are a bit awkward within Go IMO.

lnrpc/signrpc/gen_protos.sh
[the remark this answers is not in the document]
Roasbeef: Made recursive version (leaves out REST etc for now)

lnrpc/sub_server.go
```diff
@@ -0,0 +1,131 @@
+package lnrpc
+
+import (
+	fmt "fmt"
```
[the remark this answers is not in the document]
Roasbeef: Fixed, dunno what added that...

lnrpc/sub_server.go
```diff
@@ -0,0 +1,131 @@
...
+	"sync"
+
+	"google.golang.org/grpc"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+)
+
+// MacaroonPerms is a map from the FullMethod of an invoked gRPC command.  to
```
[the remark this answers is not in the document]
Roasbeef: Fixed! Sentence somehow was chopped off, filled that in properly

lnrpc/sub_server.go
```diff
@@ -0,0 +1,131 @@
...
+// the set of operations that the macaroon presented with the command MUST
+// satisfy. With this map, all sub-servers are able to communicate to the
+// primary macaroon service what type of macaroon must be passed with each
+// method present on the service of the sub-server.
+type MacaroonPerms map[string][]bakery.Op
+
+// SubServer is a child server of the main lnrpc gRPC server. Sub-servers allow
+// lnd to expose discrete services that can be use with or independent of the
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

lnrpc/signrpc/driver.go
joostjager: I don't think so. The idea of `NoMacaroons` is that you can use lnd without macaroons? It probably comes back to why we have the `NoMacaroons` flag in the first place. I use it during dev to make it easier to switch between alice, bob and carol on regtest and testnet.

lnd.go
joostjager: Is the solution then now to use two macaroons from the client (admin and signer)? I think this also means that two tcp connections need to be created?

lnrpc/signrpc/signer.proto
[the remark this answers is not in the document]
joostjager: This is also happening in for example in `rpc.proto` in the `ChannelPoint` message. Do we really need to offer this service to accept data in two formats? To me it bloats the interface while the client itself could also do the conversion to the single accepted format.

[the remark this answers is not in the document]
halseth: Agree that its better to either accept only one format, or use a `oneof` to make the API clearer.

lnrpc/signrpc/signer_server.go
```diff
@@ -0,0 +1,309 @@
...
+	if cfg.SignerMacPath == "" {
+		cfg.SignerMacPath = filepath.Join(
+			cfg.NetworkDir, DefaultSignerMacFilename,
+		)
+	}
+
+	// Now that we know the full path of the signer macaroon, we can check
+	// to see if we need to create it or not.
+	macFilePath := cfg.SignerMacPath
+	if !fileExists(macFilePath) && cfg.MacService != nil {
```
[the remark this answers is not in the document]
halseth: I initially thought that all services would require the same macaroon creation, but if they might require multiple in the future probably better to leave this as is for now. Can extract into common logic when entering the multiservice era.

subrpcserver_config.go
```diff
@@ -0,0 +1,128 @@
...
+	// Signrpc is a sub-RPC server that exposes signing of arbitrary inputs
+	// as a gRPC service.
+	Signrpc *signrpc.Config `group:"signrpc" namespace:"signrpc"`
+}
+
+// PopulateDependancies attempts to iterate through all the sub-server configs
```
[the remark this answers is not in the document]
halseth: s/PopulateDependancies/PopulateDependencies

rpcserver.go
halseth: This will still have the problem of needing to pass in more parameters in the future, if new subservers require them in their configs.

halseth: One possibility is to use a map from `subserver name -> subserverconfig`, see https://github.com/lightningnetwork/lnd/commit/b9c74ff61ca4c3200c552f3259adb7c9bbdd7365

rpcserver.go
```diff
@@ -369,8 +560,32 @@ func (r *rpcServer) Stop() error {
...
 	close(r.quit)
 
+	// After we've signalled all of our active goroutines to exit, we'll
+	// then do the same to signal a graceful shutdown of all the sub
+	// servers.
+	for _, subServer := range r.subServers {
+		rpcsLog.Infof("Stopping %v Sub-RPC Server",
+			subServer.Name())
+
+		if err := subServer.Stop(); err != nil {
```
halseth: print error?

subrpcserver_config.go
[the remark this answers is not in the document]
halseth: A way to make this possible, is to move the actual population to the build tag protected config files, see https://github.com/lightningnetwork/lnd/commit/b9c74ff61ca4c3200c552f3259adb7c9bbdd7365. 

This lets us remove this file entirely, and the use of reflection with it.

rpcserver.go
Roasbeef: That ignores that the sub-servers also may have command line arguments that we may need to parse out which are in the config.

rpcserver.go
Roasbeef: Now printed

lnrpc/signrpc/signer_server.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"google.golang.org/grpc"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+)
+
+const (
+	// subServerName is the name of the sub rpc server. We'll use this name
+	// to register ourselves, and we also require that the main
+	// SubServerConfigDispatcher instance recognize as the name of our
```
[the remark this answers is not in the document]
Roasbeef: Fixed

lnrpc/signrpc/driver.go
Roasbeef: Fixed

lnrpc/signrpc/driver.go
Roasbeef: Latest version is now usable with no macaroons!

subrpcserver_config.go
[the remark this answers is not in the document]
Roasbeef: actually, actually fixed this time :)

lnrpc/signrpc/driver.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// createNewSubServer is a helper method that will create the new signer sub
+// server given the main config dispatcher method. If we're unable to find the
+// config that is meant for us in the config dispatcher, then we'll exit with
+// an error.
+func createNewSubServer(configRegistry lnrpc.SubServerConfigDispatcher) (lnrpc.SubServer, lnrpc.MacaroonPerms, error) {
```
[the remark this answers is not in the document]
Roasbeef: Adopted

Roasbeef:
Pushed out a new version with the following major changes:
  * now passes linter so should be buildable 
  * recursive proto generation script
  * proper line wrapping as necessary

lnrpc/signrpc/signer_server.go
```diff
@@ -0,0 +1,332 @@
...
+
+// Stop signals any active goroutines for a graceful closure.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (s *Server) Stop() error {
+	if atomic.AddInt32(&s.shutdown, 1) != 1 {
+		return nil
+	}
+
+	close(s.quit)
```
cfromknecht: we don't use `quit` anywhere besides here atm. should be fine if both `Start` and `Stop` return nil? also means we can remove `started` and `stopped`

lnd.go
joostjager: Is this something that can be controlled client side, or is a change on the server needed?

lnrpc/signrpc/config_active.go
```diff
@@ -1,4 +1,4 @@
-// +build signerrpc
+// +build signrpc
```
halseth: This commit should be squashed with the first.

lnrpc/signrpc/driver.go
```diff
@@ -34,10 +34,13 @@ func createNewSubServer(configRegistry lnrpc.SubServerConfigDispatcher) (
 			&Config{}, signServerConf)
 	}
 
+	// Before we try to make the new signer service instance, we'll perform
```
halseth: this should be squashed with the commit where `driver` was introduced.

lnrpc/signrpc/signer_server.go
```diff
@@ -1,4 +1,4 @@
-// +build signerrpc
+// +build signrpc
```
halseth: should be fixup commit also

rpcserver.go
halseth: See https://github.com/lightningnetwork/lnd/commit/b9c74ff61ca4c3200c552f3259adb7c9bbdd7365#r31173493

lnrpc/signrpc/gen_protos.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
halseth: Commit can be deleted.

config.go
```diff
@@ -221,7 +221,7 @@ type config struct {
 
 	Tor *torConfig `group:"Tor" namespace:"tor"`
 
-	SubRpcServers *subRpcServerConfigs `group:"subrpc"`
+	SubRPCServers *subRPCServerConfigs `group:"subrpc"`
```
halseth: squash changes with relevant commits.

lnrpc/signrpc/signer.proto
```diff
@@ -0,0 +1,118 @@
...
+    int32 key_index = 2;
+}
+
+message KeyDescriptor {
+     oneof key {
+         /**
+         The raw bytes of the key being identified. Either this or the KeyLocator
+         must be specified.
+         */
+         bytes raw_key_bytes = 1;
```
joostjager: What is the use case for passing in `raw_key_bytes`?

rpcserver.go
```diff
@@ -572,6 +572,8 @@ func (r *rpcServer) Stop() error {
 			subServer.Name())
 
 		if err := subServer.Stop(); err != nil {
+			rpcsLog.Errorf("unable to stop sub-server %v: %v",
```
cfromknecht: nit: capitalise log message

lnd.go
joostjager: I did a quick try passing multiple macaroon credentials to the dial call, but that is not going to work. `PerRPCCredentials`, didn't try that, but it also means we need to pass around those macaroons. Is it worth the effort at this point? Otherwise I'd say, just make signing possible with the admin macaroon for now.

Roasbeef: Why do you need to pass multiple macaroon credentials? You should only need to pass the `signer.macaroon` if you're interacting with the signer sub-server. The systems that interact w/ the sub-servers will need to get used to managing their various macaroons, as we can easily implemented domain isolation using them. For example, merchants are already used to using the `invoice.macaroon` everywhere. 

It's possible to create custom macarons that can access _only_ what's needed (so signer and new address for example), as there's an open PR for this, but it isn't yet merged.

lnrpc/signrpc/signer.proto
Roasbeef: When you know the pubkey, but not the key locator information.

lnd.go
joostjager: So I need to open multiple tcp connections then? Swaplet needs signer, walletkit and the main rpc.

lnrpc/signrpc/signer.proto
```diff
@@ -0,0 +1,148 @@
...
+message KeyLocator {
+    /// The family of key being identified.
+    int32 key_family = 1;
+
+    /// The precise index of the key being identified.
+    int32 key_index = 2;
+}
+
+message KeyDescriptor {
+     oneof key {
```
joostjager: This oneof construct is not working with walletkit.DeriveNextKey. It returns both the locator and the bytes.

lnd.go
Roasbeef: Possibly, will look into if we can modify the way we manually set the credentials and get back to you!

PR #1662, opened 2018-07-31
lnrpc: better payment failure reporting

joostjager:
### Background
Many ideas float around to build functions/algorithms on top of `lnd`. For example: rebalancing, probing, routing service, etc. The current way for that kind of scripts to operate is to make calls to `SendToRoute`. 
Unfortunately limited error information is returned to the caller. `router` will keep trying the supplied routes and if none of them works, return a `routes exhausted` error without the real cause. Even if only one route is supplied. This makes it harder for scripts to guess what happened.

### Implementation details
- In order to keep it simple, the option to pass in multiple routes is removed. Only a single route per payment is allowed. If multiple routes need to be tried (with or without waiting for the result), they can still be passed in one by one. Otherwise, to get a proper error result for every route in a list of routes, the result object should also need to become a list. In the streaming version of the `SendToRoute` call, it would become a list (of routes) in a list (of payments).
- Structured error information is represented as a (flat) list of properties. The alternative would be to use protobuf `oneof`. This would for example result in the following json:
```
{
    "payment_error": "UnknownPaymentHash",
    "StructuredResult": {
        "FailUnknownPaymentHash": {}
    }
}
```
Maybe not the most intuitive representation in json and I am not sure how this is supported in the proto generators for various languages. It also seems that the oneof key cannot be assigned a `json_name`. Feedback please.

PR #2083, opened 2018-10-23
rpc+lnd: add new sub RPC server, Router service

Roasbeef:
In this PR, we implement the full RouterServer as specified by the
newly added sub-service as defined in router.proto. This new sub-server
has its own macaroon state (but overlapping permissions which can be
combined with the current admin.macaroon), and gives users a simplified
interface for a gRPC service that is able to simply send payment. Much
of the error reporting atm, is a place holder, and a follow up commit
will put up the infrastructure for a proper set of errors.

**NOTE:** The only new functionality in this PR are the last 3 commits, which show how one can add a new sub-server given the existing scaffolding. 

Depends on #2081.


joostjager:
Is this also the venue where we are going to stage #1662?

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,221 @@
...
+
+// Stop signals any active goroutines for a graceful closure.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (s *Server) Stop() error {
+	if atomic.AddInt32(&s.shutdown, 1) != 1 {
+		return nil
+	}
+
+	close(s.quit)
```
cfromknecht: similar comment here wrt removing `quit`, `started`, and `stopped`

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,221 @@
...
+	if cfg.RouterMacPath == "" {
+		cfg.RouterMacPath = filepath.Join(
+			cfg.NetworkDir, DefaultRouterMacFilename,
+		)
+	}
+
+	// Now that we know the full path of the router macaroon, we can check
+	// to see if we need to create it or not.
+	macFilePath := cfg.RouterMacPath
+	if !fileExists(macFilePath) && cfg.MacService != nil {
```
cfromknecht: would be great if this logic were abstracted in `lnrpc`, so we don't have to reimplement it with each new subserver. i imagine all subservers we add will have to use macaroons anyway

Roasbeef:
Abstract this conditional or making the macaroons? Making the macaroons is
something each sub server will need to do. However, the set of perms they
add will likely be diff and also they may eventually create multiple
macaroons.

Roasbeef:
Pushed up a new rebased version, also added a new call to obtain a routing fee quote for a `(dest, amt)` pair.

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,90 @@
...
+    */
+    int64 amt_sat = 2;
+}
+message RouteFeeResponse {
+    /**
+    A lower bound of the estimated fee to the target destination within the
+    network.
+    */
+    int64 routing_fee_sat = 1;
+}
```
joostjager: should this return the time lock too? I'd like to have that too, so the user can consent to it (times a factor) and I can pass it as a limit to sendpayment (once the timelock limit pr is in)

so it isn't a lower bound of the fee, but a lower bound of the weight function including the tradeoff fee/timelock.

lnrpc/routerpc/log.go
```diff
@@ -0,0 +1,45 @@
...
+// UseLogger uses a specified Logger to output package logging info.
+// This should be used in preference to SetLogWriter if the caller is also
+// using btclog.
+func UseLogger(logger btclog.Logger) {
+	log = logger
+}
+
+// logClosure is used to provide a closure over expensive logging operations so
+// don't have to be performed when the logging level doesn't warrant it.
+type logClosure func() string
```
joostjager: Does this `logClosure` needs to be duplicated in every package? Or can it reference/alias a common 'implementation'? (comment not related to router)

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,90 @@
...
+
+service Router {
```
joostjager: The naming `Router` is quite low level. That routing is happening is from some perspective an implementation detail. Shouldn't this service be just named `Payment`, in line with what it does?

lnrpc/routerpc/log.go
Roasbeef: We can likely extract this out to the `build` package perhaps? Agree that this type is unnecessarily being duplicated across _all_ the current packages.

lnrpc/routerpc/router.proto
Roasbeef: Done!

lnrpc/routerpc/router.proto
Roasbeef: As is, it also does more than payments in that it can also return estimates for a route/payment.

Roasbeef:
> General question: where is this requirement coming from?

Part of the sub-server architecture that enables things like managing a cluster of `lnd` nodes with one primarily sending payments (due to better graph placement and liquidity) and for example a set of other nodes (even may not be fully fledged `lnd` nodes) which handle on-chain actions. By abstracting behavior at the service level, we gain a greater degree of flexibility and are able to implement things like rolling restarts and also ensure we have service level redundancy.

lnrpc/routerpc/router.proto
joostjager: Also for a fee estimate, it isn't necessary to refer to the 'implementation detail' of routing either. I think I prefer a `Payment` service, but it's only a suggestion.

joostjager:
> By abstracting behavior at the service level, we gain a greater degree of flexibility and are able to implement things like rolling restarts and also ensure we have service level redundancy.

Yes, this part I understand. My question was about why it needs to be a 'simplified' interface? Or do you mean with simplified the isolation of several calls in the sub server rather than simplify the calls themselves?

Roasbeef:
Simplified w.r.t isolation! A side effect of the sub-server system it that it forces us to not just pile new things (like extensive business logic outside of marshaling) into the `rpcserver.go` file.

Roasbeef:
Rebased on top of the latest master.

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,97 @@
+syntax = "proto3";
+
+package routerpc;
+
+message PaymentRequest {
+    /**
+    A serialized BOLT-11 payment request that contains all information
+    required to dispatch the payment. If the pay req is invalid, or expired,
+    an error will be returned.
+
```
halseth: extra newline.

lnrpc/routerpc/driver.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// createNewSubServer is a helper method that will create the new router sub
+// server given the main config dispatcher method. If we're unable to find the
+// config that is meant for us in the config dispatcher, then we'll exit with
+// an error.
+func createNewSubServer(configRegistry lnrpc.SubServerConfigDispatcher) (lnrpc.SubServer, lnrpc.MacaroonPerms, error) {
```
halseth: line should be wrapped

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,259 @@
+// +build routerpc
+
+package routerpc
+
+import (
+	"context"
+	fmt "fmt"
```
halseth: unnecessary naming

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/lightningnetwork/lnd/zpay32"
+	"google.golang.org/grpc"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+)
+
+const (
+	// subServerName is the name of the sub rpc server. We'll use this name
+	// to register ourselves, and we also require that the main
+	// SubServerConfigDispatcher instance recognize as the name of our
+	subServerName = "RouteRPC"
```
halseth: s/RouteRPC/Route**r**RPC

subrpcserver_config.go
```diff
@@ -41,7 +48,7 @@ type subRPCServerConfigs struct {
 // FetchConfig method.
 func (s *subRPCServerConfigs) PopulateDependencies(cc *chainControl,
 	networkDir string, macService *macaroons.Service,
-	atpl *autopilot.Manager) error {
+	atpl *autopilot.Manager, server *server) error {
```
halseth: since the server is passed in, we   can remove the separate `chainControl`? Or we could pass in the `chanRouter` directly.

Roasbeef:
Comments addressed, PTAL @halseth

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,96 @@
...
+    /**
+    The destination once wishes to obtain a routing fee quote to.
+    */
+    bytes dest = 1;
+
+    /**
+    The amount one wishes to send to the target destination.
+    */
+    int64 amt_sat = 2;
+}
```
halseth: nit: add newline

lnrpc/routerpc/config_active.go
```diff
@@ -0,0 +1,43 @@
...
+package routerpc
+
+import (
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/lightningnetwork/lnd/routing"
+)
+
+// Config is the main configuration file for the router RPC server. It contains
+// all the items required for the router RPC server to carry out it duties.
```
halseth: nit: s/it/its

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,259 @@
+// +build routerpc
+
+package routerpc
```
halseth: I just noticed these are all named "route-rpc", instead of "router-rpc" as I expected. Same for build flags. Is this intentional?

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, err
+	}
+
+	// Now that all the information we need has been parsed, we'll map this
+	// proto request into a proper request that our backing router can
+	// understand.
+	finalDelta := uint16(payReq.MinFinalCLTVExpiry())
+	payment := routing.LightningPayment{
+		Target:            payReq.Destination,
+		Amount:            *payReq.MilliSat,
```
halseth: This will crash on zero-value invoices I believe

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,96 @@
...
+message RouteFeeResponse {
+    /**
+    A lower bound of the estimated fee to the target destination within the
+    network.
+    */
+    int64 routing_fee_sat = 1;
```
halseth: Doesn't it make sense to expose msat fees instead, this way we might miscalculate bc of rounding

halseth: Well, I guess since it's a lower bound then maybe it's okay?

lnrpc/routerpc/router_server.go
Roasbeef: It's a personal aesthetic thing for me: `routerpc` looks better than `routerrpc`. The consecutive "r" there makes it harder to parse.

lnrpc/routerpc/router_server.go
Roasbeef: Fixed!

Roasbeef:
Latest instance returns fee quotes in msat.

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,97 @@
...
+    /**
+    The amount one wishes to send to the target destination.
+    */
+    int64 amt_sat = 2;
+}
+
+message RouteFeeResponse {
+    /**
+    A lower bound of the estimated fee to the target destination within the
+    network, expressed in satoshis.
```
halseth: nit: comment should be updated to msat

lnrpc/routerpc/router.proto
```diff
@@ -0,0 +1,97 @@
...
+    */
+    string pay_req = 1;
+
+    /**
+    An absolute limit on the highest fee we should pay when looking for a
+    route to the destination. Routes with fees higher than this will be
+    ignored, if there are no routes with a fee below this amount, an error
+    will be returned.
+    */
+    int64 fee_limit = 2;
```
cfromknecht: needs a unit suffix

lnrpc/routerpc/router_server.go
cfromknecht: bump

subrpcserver_config.go
```diff
@@ -31,6 +33,12 @@ type subRPCServerConfigs struct {
 	// AutopilotRPC is a sub-RPC server that exposes methods on the running
 	// autopilot as a gRPC service.
 	AutopilotRPC *autopilotrpc.Config `group:"autopilotrpc" namespace:"autopilotrpc"`
+
+	// RouterRPC is a sub-RPC server the exposes functionality that allows
+	// clients to send payments on the network, and perform Lightning
+	// payment related queries such as requests for estimates of off-chain
+	// fees.
+	RouterRPC *routerpc.Config `group:"routerpc" namespace:"routerpc"`
```
cfromknecht: group and namespace should be `routerrpc`, missing an r

lnrpc/routerpc/config_default.go
```diff
@@ -0,0 +1,7 @@
+// +build !routerpc
+
+package routerpc
```
joostjager: Agree with @cfromknecht, this doesn' t like good imo and needs to be `routerrpc`.

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,265 @@
...
+		},
+	}
+
+	// macPermissions maps RPC calls to the permissions they require.
+	macPermissions = map[string][]bakery.Op{
+		"/routerpc.Router/SendPayment": {{
+			Entity: "offchain",
+			Action: "write",
+		}},
+	}
```
joostjager: Add `EstimateRouteFee`

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// Now that all the information we need has been parsed, we'll map this
+	// proto request into a proper request that our backing router can
+	// understand.
+	finalDelta := uint16(payReq.MinFinalCLTVExpiry())
+	payment := routing.LightningPayment{
+		Target:            payReq.Destination,
+		Amount:            *payReq.MilliSat,
+		FeeLimit:          lnwire.MilliSatoshi(req.FeeLimit),
```
joostjager: I the fee limit is in msat, I think it should also be passed into the rpc as msat? Or alternatively change the field in  `LightningPayment` and further to `btcutil.Amount`?

lnrpc/routerpc/router_server.go
```diff
@@ -0,0 +1,265 @@
...
+		return nil, err
+	}
+
+	// Next, we'll convert the amount in satoshis to mSAT, which are the
+	// native unit of LN.
+	amtMsat := lnwire.NewMSatFromSatoshis(btcutil.Amount(req.AmtSat))
+
+	// Finally, we'll query for a route to the destination that can carry
+	// that target amount, we'll only request a single route.
+	routes, err := s.cfg.Router.FindRoutes(
```
joostjager: Imo we should not increase the dependency on the k-shortest algorithm as we might remove it in the future. Because it is a single route, we can use `findPath` here directly?

joostjager: It would indeed require adding a `FindRoute` function that generates the bandwidth hints before calling `findPath`. But I don't see much more is needed other than that.

lnrpc/routerpc/log.go
```diff
@@ -0,0 +1,45 @@
+package routerpc
```
halseth: wrong package and folder name

Roasbeef: I think you commented on a dated diff.

Roasbeef: Ah fixed, the comment.

lnrpc/routerpc/router_server.go
Roasbeef: Indeed, this PR is focused on just adding the sub-server. Adding a singular `FindRoute` (along with adding tests, etc) will need to be when/if we decide to deprecate `FindRoutes` all together. I imagine that eventually we'll not even attempt a regular query, and instead may want more of a fee range to be returned to give the caller an idea w.r.t both the min and max.

joostjager:
I think it is better to take out EstimateFee completely. We currently have no way to calculate a realistic fee estimate. Just returning the minimum fee as we do now is of limited practical use. The rpc promises functionality that we currently cannot deliver. Besides, it is hardly more than a wrapper around QueryRoutes.

halseth:
Needs rebase.

Issue #274, opened 2017-08-27
Don't unconditionally create a new NodeAnnouncement upon startup


Roasbeef:
# Issue

[Atm, upon start up we'll unconditnioally create a new node announcement, write it to the database, sign the announcement ~~and broadcast it to the network~~](https://github.com/lightningnetwork/lnd/blob/master/server.go#L212). This is unnecessary, as we should only do this if configuration options between this running instance and the last have modified what our `NodeAnnouncement` _should_ be within the network. 

# Steps to Completion 

- [ ] Rather than doing the above unconditionally, we should instead craft a new node announcement based on the relevant configuration options, and compare that with the one on disk. If they differ, then we should update the node announcement in the database.

blico:
I would like to give this one a shot! (seems easy enough)

halseth:
As noted in https://github.com/lightningnetwork/lnd/pull/2084#discussion_r321007023, we still unconditionally update our node announcement on startup (but we don't broadcast it). I updated the issue description to reflect this.

halseth:
Discussion is whether updating and sending on startup is actually okay. Currently we don't send the update one it seems like.

Roasbeef:
These days we have node announcement propagation issues, so this would be a stop backwards.

PR #2084, opened 2018-10-23
Retransmit NodeAnnouncement on regular intervals

### LINKED ISSUE #2005: Send out updated announcements periodically
Some nodes will attempt to prune "zombie" nodes and channels from the channel graph. This can happen if a new ~~`ChannelUpdate` or~~ `NodeAnnouncement`  with an updated timestamp is sent after a time on the order of a few weeks.

To avoid getting pruned, we should periodically check the timestamp of our last sent `NodeAnnouncement` ~~and `ChannelUpdate`~~, and broadcast updated ones if the timestamp is a few weeks in the past.

halseth:
This PR makes the gossiper retransmit our `NodeAnnouncement` every 24 hrs, to avoid other nodes considering it being a zombie. This is similar to what is done for channels.

Fixes #2005.


discovery/gossiper.go
```diff
@@ -1563,6 +1563,34 @@ func (d *AuthenticatedGossiper) processRejectedEdge(
 	return announcements, nil
 }
 
+// addNode processes the given node announcement, and adds it to our channel
+// graph.
+func (d *AuthenticatedGossiper) addNode(msg *lnwire.NodeAnnouncement) error {
+	if err := routing.ValidateNodeAnn(msg); err != nil {
+		err := fmt.Errorf("unable to validate node announcement: %v",
+			err)
+		return err
```
wpaulino: Nit: just craft the error here?

discovery/gossiper_test.go
```diff
@@ -170,8 +166,25 @@ func (r *mockGraphSource) ForEachNode(func(node *channeldb.LightningNode) error)
...
+		e, ok := r.edges[info.ChannelID]
+		if !ok {
+			return fmt.Errorf("policy for chanID %v not found",
+				info.ChannelID)
+		}
+		if len(e) != 2 {
+			return fmt.Errorf("had %d policies for chanID %v",
+				len(e), info.ChannelID)
+		}
+		if err := cb(info, e[0]); err != nil {
```
wpaulino: Doesn't seem like we can guarantee that the first edge will always be the outgoing one?

discovery/gossiper_test.go
```diff
@@ -63,6 +63,8 @@ var (
 	retransmitDelay     = time.Hour * 1
 	rebroadcastInterval = time.Second * 100
 	proofMatureDelta    uint32
+
+	testTimestamp = uint32(time.Now().Unix())
```
wpaulino: Any reason this always needs to be the current timestamp? IIRC, @joostjager has had some remarks on using dynamic values within tests.

discovery/gossiper_test.go
```diff
@@ -2308,6 +2310,156 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
...
+	case <-time.After(2 * time.Second):
+		t.Fatal("did not process local announcement")
+	}
+	if err != nil {
+		t.Fatalf("unable to process :%v", err)
+	}
+
+	select {
+	case <-ctx.broadcastedMessage:
+	// TODO: enable after wilmer's private node announcement thing is merged.
```
wpaulino: Looks like this TODO can be removed. One thing I forgot to add to that PR (which has already been merged), was to address this:
https://github.com/lightningnetwork/lnd/blob/3b9aa2e7b5f84b8332846c561774720d76d80698/discovery/gossiper_test.go#L209

We should return true instead to mimic what would actually happen.

discovery/gossiper_test.go
```diff
@@ -2308,6 +2310,156 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
...
+			t.Fatalf("announcement %v wasn't broadcast", i)
+		}
+	}
+
+	// Now let the retransmit ticker tick, which should trigger updates to
+	// be rebroadcast.
+	now := time.Unix(int64(testTimestamp), 0)
+	future := now.Add(rebroadcastInterval + 10*time.Second)
+	select {
+	case ctx.gossiper.cfg.RetransmitTicker.(*ticker.Mock).Force <- future:
```
wpaulino: 👍

discovery/gossiper.go
```diff
@@ -1385,13 +1387,46 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels(now time.Time) error {
...
+				"announment: %v", err)
+		}
+
+		signedUpdates = append(signedUpdates, &newNodeAnn)
+		nodeAnnStr = " and our refreshed node announcement"
+
+		// Before broadcasting the refreshed node announcement, add it
+		// to our own graph.
+		if err := d.addNode(&newNodeAnn); err != nil {
+			log.Error("unable to add refreshed node announcement "+
```
wpaulino: Nit: s/Error/Errorf/

Errors logged should also be capitalized.

server.go
```diff
@@ -565,11 +565,12 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
 		NotifyWhenOnline: s.NotifyWhenOnline,
-		ProofMatureDelta: 0,
-		TrickleDelay:     time.Millisecond * time.Duration(cfg.TrickleDelay),
-		RetransmitDelay:  time.Minute * 30,
-		DB:               chanDB,
-		AnnSigner:        s.nodeSigner,
+		ProofMatureDelta:    0,
+		TrickleDelay:        time.Millisecond * time.Duration(cfg.TrickleDelay),
+		RetransmitDelay:     time.Minute * 30,
+		RebroadcastInterval: time.Hour * 24,
```
wpaulino: Any reason this is the default? It could be made less spammy IMO.

discovery/gossiper_test.go
joostjager: I indeed use static time values in unit tests to keep them deterministic.

halseth: Nice, will see if I can do the same! 😀

server.go
halseth: Could reasonably be made longer, but don't think one update every 24 hrs is the biggest spam problem atm.

discovery/gossiper_test.go
halseth: Made a new PR for this change: https://github.com/lightningnetwork/lnd/pull/2123

discovery/gossiper_test.go
halseth: Done!

discovery/gossiper_test.go
halseth: Fixed!

wpaulino:
Needs a rebase!

halseth:
Rebased.

discovery/gossiper.go
```diff
@@ -1231,13 +1233,46 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels(now time.Time) error {
...
+		return fmt.Errorf("unable to get current node announment: %v",
+			err)
+	}
+
+	timestamp := time.Unix(int64(currentNodeAnn.Timestamp), 0)
+	timeElapsed := now.Sub(timestamp)
+
+	// If it's been a full day since we've re-broadcasted the
+	// node announcement, refresh it and resend it.
+	// TODO(halseth): only send if we have public channels.
```
cfromknecht: should this be a part of this pr? I think we can use the `IsPublicNode` method used below?

discovery/gossiper_test.go
```diff
@@ -2910,8 +2920,9 @@ func TestRetransmit(t *testing.T) {
 		t.Fatalf("unable to force tick")
 	}
 
-	// The channel announcement + channel update should be re-broadcast.
-	for i := 0; i < 2; i++ {
+	// The channel announcement + channel update + node announcement should
+	// be re-broadcast.
+	for i := 0; i < 3; i++ {
```
cfromknecht: also assert proper types are being retransmitted?

discovery/gossiper_test.go
```diff
@@ -2768,6 +2781,156 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
...
+		t.Fatalf("unable to parse pubkey: %v", err)
+	}
+	remoteKey, err := btcec.ParsePubKey(batch.nodeAnn2.NodeID[:], btcec.S256())
+	if err != nil {
+		t.Fatalf("unable to parse pubkey: %v", err)
+	}
+	remotePeer := &mockPeer{remoteKey, nil, nil}
+
+	// Process a local channel annoucement, channel update and node
+	// announcement. Node should be broadcased yet, since no proof has been
```
cfromknecht: should not?

discovery/gossiper.go
```diff
@@ -1178,14 +1185,12 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels() error {
...
 
-		const broadcastInterval = time.Hour * 24
-
 		timeElapsed := time.Since(edge.LastUpdate)
 
 		// If it's been a full day since we've re-broadcasted the
 		// channel, add the channel to the set of edges we need to
 		// update.
-		if timeElapsed >= broadcastInterval {
+		if timeElapsed >= d.cfg.RebroadcastInterval {
```
wpaulino: Nit: update comment since rebroadcast interval is now configurable.

discovery/gossiper_test.go
```diff
@@ -1498,7 +1501,8 @@ func TestSignatureAnnouncementRetryAtStartup(t *testing.T) {
 		NotifyWhenOffline:    ctx.gossiper.reliableSender.cfg.NotifyWhenOffline,
 		Router:               ctx.gossiper.cfg.Router,
 		TrickleDelay:         trickleDelay,
-		RetransmitDelay:      retransmitDelay,
+		RetransmitTicker:     ticker.NewForce(retransmitDelay),
```
wpaulino: Nit: can be squashed with the previous commit to allow the tests to build.

discovery/gossiper_test.go
```diff
@@ -2771,6 +2771,155 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
...
+	case err = <-ctx.gossiper.ProcessLocalAnnouncement(batch.localChanAnn,
+		localKey):
+	case <-time.After(2 * time.Second):
+		t.Fatal("did not process local announcement")
+	}
+	if err != nil {
+		t.Fatalf("unable to process :%v", err)
+	}
+	select {
+	case <-ctx.broadcastedMessage:
```
wpaulino: Unsure if something like this already exists, but given that we repeat this pattern so much, we should extract it into a helper function, e.g. `assertBroadcastedMessage`.

discovery/gossiper.go
```diff
@@ -1231,13 +1233,46 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels(now time.Time) error {
...
+
+	// If it's been a full day since we've re-broadcasted the
+	// node announcement, refresh it and resend it.
+	// TODO(halseth): only send if we have public channels.
+	nodeAnnStr := ""
+	if timeElapsed >= d.cfg.RebroadcastInterval {
+		newNodeAnn, err := d.cfg.SelfNodeAnnouncement(true)
+		if err != nil {
+			return fmt.Errorf("unable to get refreshed node "+
+				"announment: %v", err)
```
wpaulino: Nit: s/announment/announcement/

discovery/gossiper.go
```diff
@@ -1231,13 +1233,46 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels(now time.Time) error {
...
 
-	// If we don't have any channels to re-broadcast, then we'll exit
+	// We'll also check that our NodeAnnouncement is not too old.
+	currentNodeAnn, err := d.cfg.SelfNodeAnnouncement(false)
+	if err != nil {
+		return fmt.Errorf("unable to get current node announment: %v",
+			err)
+	}
+
+	timestamp := time.Unix(int64(currentNodeAnn.Timestamp), 0)
```
wpaulino: Just something to note, the timestamp of our node announcement is updated to `time.Now` every time the server is started.

discovery/gossiper_test.go
```diff
@@ -1493,10 +1498,15 @@ func TestSignatureAnnouncementRetryAtStartup(t *testing.T) {
...
 	gossiper := New(Config{
-		Notifier:             ctx.gossiper.cfg.Notifier,
-		Broadcast:            ctx.gossiper.cfg.Broadcast,
-		NotifyWhenOnline:     ctx.gossiper.reliableSender.cfg.NotifyWhenOnline,
-		NotifyWhenOffline:    ctx.gossiper.reliableSender.cfg.NotifyWhenOffline,
+		Notifier:          ctx.gossiper.cfg.Notifier,
+		Broadcast:         ctx.gossiper.cfg.Broadcast,
+		NotifyWhenOnline:  ctx.gossiper.reliableSender.cfg.NotifyWhenOnline,
+		NotifyWhenOffline: ctx.gossiper.reliableSender.cfg.NotifyWhenOffline,
+		SelfNodeAnnouncement: func(bool) (lnwire.NodeAnnouncement, error) {
```
wpaulino: Nit: can use `ctx.gossiper.cfg.SelfNodeAnnouncement`?

discovery/gossiper.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2084/commits/50980662ef58fdab9bd5e1215ccb97bf52d9fd00

discovery/gossiper_test.go
halseth: Done!

discovery/gossiper_test.go
```diff
@@ -2454,6 +2456,155 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
 	}
 }
 
+// TestRetransmit checks that the expected announcements are retransmittet when
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2084/commits/68a0ee61985a089d8af7c66ca570842896d35436

discovery/gossiper_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2084/commits/68a0ee61985a089d8af7c66ca570842896d35436

discovery/gossiper_test.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/2084/commits/77370540255c8edd10ce212c3ecdb52a0a22fa32

discovery/gossiper.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2084/commits/e7424d7fa91a7b72d9b3a03643f08367248d5e82

discovery/gossiper.go
halseth: Figured it was even simpler to just add a bool: https://github.com/lightningnetwork/lnd/pull/2084/commits/512efc855aae68a0b28c5544e34e4848b2642cc5

discovery/gossiper.go
```diff
@@ -1157,16 +1172,21 @@ func (d *AuthenticatedGossiper) isRecentlyRejectedMsg(msg lnwire.Message) bool {
 
 // retransmitStaleChannels examines all outgoing channels that the source node
 // is known to maintain to check to see if any of them are "stale". A channel
-// is stale iff, the last timestamp of its rebroadcast is older then
-// broadcastInterval.
-func (d *AuthenticatedGossiper) retransmitStaleChannels() error {
+// is stale iff, the last timestamp of its rebroadcast is older than the
+// RebroadcastInterval. We also check if a refreshed node announcement should be
+// resent.
+func (d *AuthenticatedGossiper) retransmitStaleChannels(now time.Time) error {
```
wpaulino: Rename the method to `retransmitStaleAnns` since we also do our node announcement now?

discovery/gossiper.go
wpaulino: Because of this, if a node restarts before it reaches a full day of uptime, it'll never rebroadcast its node announcement.

discovery/gossiper_test.go
```diff
@@ -2768,6 +2777,185 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to parse pubkey: %v", err)
+	}
+	remotePeer := &mockPeer{remoteKey, nil, nil}
+
+	// Process a local channel annoucement, channel update and node
+	// announcement. No messages should be broadcasted yet, since no proof
+	// has been exchanged.
+	select {
+	case err = <-ctx.gossiper.ProcessLocalAnnouncement(
```
wpaulino: Would be nice to have a helper for these too, not a blocker though.

server.go
wpaulino: Sure but given that we only consider things stale after 2 weeks it could be set a bit higher.

discovery/gossiper.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/2084/commits/ec91c340243da0e61b58a01aee5e5cebd411f5e1

discovery/gossiper.go
wpaulino: I don't think that's the case. We only seem to send when opening a channel or when running with `--nat`.

discovery/gossiper_test.go
halseth: Done.

halseth:
> LGTM, though I still think the node announcement timestamp issue should be addressed in order to be consistent, likely as a separate PR.

I updated and reopened https://github.com/lightningnetwork/lnd/issues/274. Let me know if this is what you had in mind.

wpaulino:
> I updated and reopened #274. Let me know if this is what you had in mind.

SGTM

PR #2086, opened 2018-10-23
test: rename files to exclude from non-test build

joostjager:
This PR prevents `test_utils.go` in various packages from being included in non-test builds by giving them the `_test` suffix.


joostjager:
Not a good idea, because now these files cannot be used from another package test.

PR #2088, opened 2018-10-24
chainntnfs/interface_test: lower mempool spend check timeout

wpaulino:
In this commit, we lower the mempool spend check timeout to be twice as
long as the trickle interval of the miner node, which will greatly
improve the execution time of this specific test. We're able to do this
now since we can specify custom trickle intervals for our test
harnesses.

### DIFF
--- a/chainntnfs/interface_test.go
+++ b/chainntnfs/interface_test.go
@@ -320,11 +320,7 @@ func testSpendNotification(miner *rpctest.Harness,
 	// Make sure notifications are not yet sent. We launch a go routine for
 	// all the spend clients, such that we can wait for them all in
 	// parallel.
-	//
-	// Since bitcoind is at times very slow at notifying about txs in the
-	// mempool, we use a quite large timeout of 10 seconds.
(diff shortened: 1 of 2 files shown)


Roasbeef:
Still relevant?

wpaulino:
Yes, though IIRC @halseth had some concerns about this. I tested that the timeout actually works by allowing bitcoind to dispatch on mempool spends and that the notification is received within it.

halseth:
Earlier `bitcoind` could use several seconds before notifying about mempool spends, which is why the long timeout was originally 10 seconds. Did anything change such that this is no longer true?

wpaulino:
Not sure, but that doesn't seem to be the case anymore. Perhaps it was just due to the trickle interval not being custom like it is now?

wpaulino:
> how many runs has this had on travis? do we see in regressions in stability?

Have given it a couple runs multiple times and haven't seen an issue.

halseth:
Since the timeout is used to trigger continuation of the success path, lowering it should only make the test more likely to pass 😛 (which is why a long timeout was used in the first place)

PR #2091, opened 2018-10-24
Strict channel enabling

halseth:
This PR enables strict channel enabling in `lnd`. This means that similarly to how we disable channels, we require a disabled link to stay active for 20 minutes (default) before re-enabling it.

This is done in two steps:
1. The path finding algorithm is modified to take into account an extra set of active local channels, that it will use when considering whether to ignore local channels. This improves the current path finding logic when using `mission control`, as we will no longer attempt to route over local channels that are inactive.
2. The channel re-enabling logic carried out when a peer connects is removed, as we will now instead let the server's re-enabling logic take care of enabling the disabled channels when they have been stable for 20 minutes.

This essentially decouples the graph's disabled status from the local path finding, as the local channel status will be more up-to-date than what is found in the graph.


htlcswitch/switch.go
```diff
@@ -1990,6 +1990,19 @@ func (s *Switch) getLinkByShortID(chanID lnwire.ShortChannelID) (ChannelLink, er
...
 
+// ActiveLinks returns a list of ChannelLinks currently active within the
+// switch.
+func (s *Switch) ActiveLinks() []ChannelLink {
+	s.indexMtx.RLock()
+	defer s.indexMtx.RUnlock()
+
+	var links []ChannelLink
+	for _, l := range s.linkIndex {
+		links = append(links, l)
```
cfromknecht: should maybe filter by `EligibleToForward()`?

routing/pathfind.go
```diff
@@ -457,16 +499,12 @@ func edgeWeight(lockedAmt lnwire.MilliSatoshi, fee lnwire.MilliSatoshi,
 // destination node back to source. This is to properly accumulate fees
 // that need to be paid along the path and accurately check the amount
 // to forward at every node against the available bandwidth.
-func findPath(tx *bolt.Tx, graph *channeldb.ChannelGraph,
-	additionalEdges map[Vertex][]*channeldb.ChannelEdgePolicy,
-	sourceNode *channeldb.LightningNode, target *btcec.PublicKey,
-	ignoredNodes map[Vertex]struct{}, ignoredEdges map[uint64]struct{},
-	amt lnwire.MilliSatoshi, feeLimit lnwire.MilliSatoshi,
-	bandwidthHints map[uint64]lnwire.MilliSatoshi) ([]*channeldb.ChannelEdgePolicy, error) {
+func findPath(p *findPathParams) ([]*channeldb.ChannelEdgePolicy, error) {
```
cfromknecht: 👍

config.go
```diff
@@ -227,7 +227,7 @@ type config struct {
 	NoSeedBackup bool `long:"noseedbackup" description:"If true, NO SEED WILL BE EXPOSED AND THE WALLET WILL BE ENCRYPTED USING THE DEFAULT PASSPHRASE -- EVER. THIS FLAG IS ONLY FOR TESTING AND IS BEING DEPRECATED."`
 
 	TrickleDelay        int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
-	InactiveChanTimeout time.Duration `long:"inactivechantimeout" description:"If a channel has been inactive for the set time, send a ChannelUpdate disabling it."`
+	InactiveChanTimeout time.Duration `long:"inactivechantimeout" description:"If a channel has been inactive for the set time, send a ChannelUpdate disabling it. If the channel becomes active again, send an enabling ChannelUpdate after the same timeout."`
```
cfromknecht: I think it'd be useful to have these be distinctly configurable. The naming is somewhat confusing since the config option is overloaded

peer.go
```diff
@@ -488,33 +487,8 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
...
 
-	// As a final measure we launch a goroutine that will ensure the newly
-	// loaded public channels are not currently disabled, as that will make
-	// us skip it during path finding.
-	go func() {
-		for _, chanPoint := range activePublicChans {
-			// Set the channel disabled=false by sending out a new
-			// ChannelUpdate. If this channel is already active,
-			// the update won't be sent.
-			err := p.server.announceChanStatus(chanPoint, false)
```
cfromknecht: I still think we want to have this logic w/in the peer. Depending on the timing of the `channelStatusWatcher`, it could result in the reenable happening at 2x the specified timeout. I think this is okay for disables, but not something we want for enabling a channel

cfromknecht: #2080 still seems like it should go in, and then thread the reenable timeout to override the default

config.go
halseth: would you say it is sufficient to rename it? Making them distinct will complicate the logic a bit, and also seems they should be symmetric.

htlcswitch/switch.go
halseth: Ah, good idea! Should the same filtering be done for `HasActiveLink` below?

peer.go
halseth: I think the logic within the `channelStatusWatcher` should ensure that an active channel will stay `disabled` for at most `InactiveChanInterval * 5 / 4` ?

The reason we want this for enabling as well is to attempt making the network converge to a state where nodes going offline often will have their channels disabled. Eventually we might even consider enabling them with an exponential backoff.

cfromknecht: > I think the logic within the channelStatusWatcher should ensure that an active channel will stay disabled for at most InactiveChanInterval * 5 / 4 ?

Indeed, it is 1.25x the specified timeout

From my understanding, the `channelStatusWatcher` is effectively long-polling, and is susceptible to false positives where a peer is online for say 5 mins at a time, but disconnects and reconnects quickly so each time we think it's enabled. 

IMO we should only reenable if the connection is actually stable for 20 (or w/e) minutes, which can't really be deduced via long polling. #2080 fixes this by ensuring we always do so only after the prescribed duration since the connection was initialized. We avoid this currently by immediately sending a reenable, so the watcher is never actually detecting the disable -> enable transition.

FWIW, I think long polling is sufficient for detecting disables, as being down four consecutive times should be an indicator that the channel isn't reliable. For enabling however, passing four times isn't really sufficient for determining if the connection is actually stable.

cfromknecht: > Eventually we might even consider enabling them with an exponential backoff.

Yeah was pondering this as well, I think it'd be a good idea

config.go
cfromknecht: hmm my thinking was that they would be asymmetric, or at least that the ability to configure them as such would be useful even if the defaults were the same

peer.go
halseth: > IMO we should only reenable if the connection is actually stable for 20 (or w/e) minutes, which can't really be deduced via long polling.

Totally agree, I like this idea! I would say if we do this we should also make this true for disables; if you cannot maintain a stable connection for 20 minutes, we'll disable you. I think the code would be easier to maintain and extend if they share the logic here.

To allow nodes to do infrequent restarts, we could say something like they must keep a rolling 18.5 out of 20 min uptime, with each disconnect counting as minimum 1 min downtime say ( I haven't given this too much thought yet). I imagine @valentinewallace's `channel event service` could come in handy here :)

halseth:
The disable bit decoupling is now split into a separate PR, to reduce the size of this one, and since they are really do different discussions: #2115

config.go
cfromknecht: this is also somewhat of a breaking change from the user's perspective. prior to this change, their channels will be reenabled immediately. after applying this change, their channels won't be routable for 20 mins. this behavior happens w/o making any changes to the config because we're overloading the param

halseth: It should help with the health of the network, and in the future nodes might get punished (by getting used less for routing) for not disabling unreliable channels.

cfromknecht: > It should help with the health of the network

I agree, which to me indicates that the defaults should be symmetric. though i still don't see a reason for not giving users the option to choose asymmetric timeouts if they want, esp if the code paths for enable/disable will be distinct anyway.

misnaming/overloading parameters has led to a great deal of confusion previously, imo we should try keep argument names as self-descriptive as possible, since most users will never read `lnd -h`

halseth: Agree on changing the parameter name to something more descriptive. 

But I'm reluctant to making it possible to set them asymmetrically, as we want to discourage that (I think?), and also clutters up the config.

PR #2092, opened 2018-10-25
make: ensure make fmt is run with -s flag

halseth:
Linting will fail if code is not formatted with the -s flag, so make
sure we always use it when formatting.

Inspired by https://github.com/elastic/beats/pull/780

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,14 +36,13 @@ GOBUILD := go build -v
 GOINSTALL := go install -v
 GOTEST := go test -v
 
-GOLIST := go list $(PKG)/... | grep -v '/vendor/'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -211,7 +209,7 @@ endif
 
 fmt:
 	@$(call print, "Formatting source.")
-	$(GOLIST) | $(XARGS) go fmt -x
+	gofmt -s -l -w .
```
cfromknecht: won't this format the vendor dir?

halseth: yeah, but does it matter? 😛  I guess it could increase the duration...

halseth: Updated to exclude the vendor dir!

PR #2093, opened 2018-10-25
lnrpc: add new WalletKit sub-RPC server

Roasbeef:
In this PR, we add a new sub-RPC server to the existing set of gRPC
servers. This new sub-RPC server is the WalletKit. It's a utility
toolkit that contains method which allow clients to perform common
interactions with a wallet such as getting a new address, or sending a
transaction. It also includes some supplementary actions such as fee
estimation.

One thing to note in the RPC file is that we _import_ the existing
signer.proto file in order to get at some existing proto definitions
which are useful in our use case.

Depends on #2081.


lnrpc/walletrpc/walletkit_server.go
```diff
@@ -0,0 +1,366 @@
...
+		}},
+		"/walletrpc.WalletKit/HasSpendingTx": {{
+			Entity: "onchain",
+			Action: "read",
+		}},
+		"/walletrpc.WalletKit/EstimateFee": {{
+			Entity: "onchain",
+			Action: "read",
+		}},
+	}
```
joostjager: SendOutputs missing

lnrpc/walletrpc/walletkit.proto
```diff
@@ -0,0 +1,157 @@
...
+    /**
+    A slice of the outputs that should be created in the transaction produced.
+    */
+    repeated signrpc.TxOut outputs = 2;
+}
+message SendOutputsResponse {
+    /**
+    The txid of the transaction sent out on the network.
+    */
+    bytes txid = 1;
```
joostjager: Can this return the mapping of outputs to tx out indices?

joostjager: Proposal:
https://github.com/btcsuite/btcwallet/pull/568/files

Just returning the change index won't work because of the output randomization that is happening.

lnrpc/walletrpc/walletkit.proto
```diff
@@ -0,0 +1,157 @@
...
+    */
+    int32 index = 2;
+}
+
+message SendOutputsRequest {
+    /**
+    The number of satoshis per kilo weight that should be used when crafting
+    this transaction.
+    */
+    int64 sat_per_kw = 1;
```
joostjager: In addition to this, we need a maximum total fee as well. We don't know how many inputs are needed, so the fee can potentially become (too) large.

lnrpc/walletrpc/walletkit.proto
```diff
@@ -0,0 +1,157 @@
...
+    */
+    rpc NextAddr(AddrRequest) returns (AddrResponse);
+
+    /**
+    PublishTransaction attempts to publish the passed transaction to the
+    network. Once this returns without an error, the wallet will continually
+    attempt to re-broadcast the transaction on start up, until it enters the
+    chain.
+    */
+    rpc PublishTransaction(Transaction) returns (PublishResponse);
```
joostjager: How is this going to work with the rebroadcaster? It needs sign descriptors too.

joostjager: Suggestion: instead of `PublishTransaction`, this could be `SendInputs` which takes sign descriptors.

joostjager:
Need a way to persist the `SendOutputs` tx before it is published, to make sure we never accidentally pay to the same output again.

lnrpc/walletrpc/walletkit.proto
Roasbeef: No longer needed now that we return the entire transaction.

lnrpc/walletrpc/walletkit.proto
Roasbeef: So a max for the request, or a max in the response? Once you know the fee rate, you can use the weight estimator to compute what the final fee will be.

lnrpc/walletrpc/walletkit.proto
Roasbeef: So just to check in again, I think we're find with `PublishTransaction` and may follow up with the two-stage craft+sign process?

lnrpc/walletrpc/walletkit_server.go
Roasbeef: Fixed!

lnrpc/walletrpc/walletkit.proto
joostjager: I am suggesting 3. Set a hard cap on the actually paid tx fee if I instruct the wallet to create a certain nof outputs.

lnrpc/walletrpc/driver.go
```diff
@@ -0,0 +1,77 @@
...
+			&Config{}, walletKitServerConf)
+	}
+
+	// Before we try to make the new WalletKit service instance, we'll
+	// perform some sanity checks on the arguments to ensure that they're
+	// useable.
+	switch {
+	case config.MacService == nil:
+		return nil, nil, fmt.Errorf("MacService must be set to " +
+			"create WalletKit RPC server")
```
joostjager: It seems walletkit cannot be used with nomacaroons

lnrpc/walletrpc/walletkit.proto
Roasbeef: Gotcha, will make an issue to track this as this can be added in a new PR.

lnrpc/walletrpc/driver.go
Roasbeef: Fixed!

Roasbeef:
Addressed all lingering comments, PTAL!

subrpcserver_config.go
```diff
@@ -97,6 +122,8 @@ func (s *subRPCServerConfigs) FetchConfig(subServerName string) (interface{}, bo
 		return nil, false
 	}
 
+	configValElem := configVal.Elem()
```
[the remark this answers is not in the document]
Roasbeef: It's actually a bug fix left over from the prior instance, it catches a panic.

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -0,0 +1,362 @@
...
+// sub RPC server to register itself with the main gRPC root server. Until this
+// is called, each sub-server won't be able to have requests routed towards it.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (w *WalletKit) RegisterWithRootServer(grpcServer *grpc.Server) error {
+	// We make sure that we register it with the main gRPC server to ensure
+	// all our methods are routed properly.
+	RegisterWalletKitServer(grpcServer, w)
+
+	log.Debugf("WalletKit RPC server successfully register with " +
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

lnrpc/walletrpc/log.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// log is a logger that is initialized with no output filters.  This means the
+// package will not perform any logging by default until the caller requests
+// it.
+var log btclog.Logger
+
+// The default amount of logging is none.
+func init() {
+	UseLogger(build.NewSubLogger("CRTR", nil))
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -0,0 +1,360 @@
...
+	// macaroon that we expect to find via a file handle within the main
+	// configuration file in this package.
+	DefaultWalletKitMacFilename = "walletkit.macaroon"
+)
+
+// WalletKit is a sub-RPC server that exposes a tool kit which allows clients
+// to execute common wallet operations. This includes requesting new addresses,
+// keys (for contracts!), and publishing transactions.
+type WalletKit struct {
+	started  int32 // To be used atomically.
```
cfromknecht: `started`, `stopped`, and `quit` don't do anything, they can be removed

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return walletKit, macPermissions, nil
+}
+
+// Start launches any helper goroutines required for the sub-server to function.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (w *WalletKit) Start() error {
+	if atomic.AddInt32(&w.started, 1) != 1 {
+		return nil
```
cfromknecht: `Start()` can just return nil

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return nil
+}
+
+// Stop signals any active goroutines for a graceful closure.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (w *WalletKit) Stop() error {
+	if atomic.AddInt32(&w.shutdown, 1) != 1 {
+		return nil
```
cfromknecht: `Stop` can just return nil

joostjager:
Didn't you want to add signer permission to the admin macaroon in this pr too?

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -0,0 +1,360 @@
...
+		cfg.WalletKitMacPath = filepath.Join(
+			cfg.NetworkDir, DefaultWalletKitMacFilename,
+		)
+	}
+
+	// Now that we know the full path of the wallet kit macaroon, we can
+	// check to see if we need to create it or not.
+	macFilePath := cfg.WalletKitMacPath
+	if !fileExists(macFilePath) && cfg.MacService != nil {
+		log.Infof("Making macaroons for WaleltKit RPC Server at: %v",
```
wpaulino: suggest
```suggestion
		log.Infof("Baking macaroons for WalletKit RPC Server at: %v",
```
😛

PR #2094, opened 2018-10-25
config: validate node color after reading config

### LINKED ISSUE #2082: Validate color flag immediately
### Background

Very minor annoyance, but passing an invalid `--color` (e.g. forgetting the #) doesn't get caught until you unlock the wallet. Color should be validated immediately.

### Your environment

* version of `lnd` - 0.5.0-beta
* which operating system (`uname -a` on *Nix) - macOS
* version of `btcd`, `bitcoind`, or other backend - btcd

### Steps to reproduce

Run with `--color=somecolororwhatever`

### Expected behaviour

It should die immediately

### Actual behaviour

It runs until you unlock wallet

cfromknecht:
Adds simple check to validate color after parsing config file.

Closes #2082

### DIFF
--- a/config.go
+++ b/config.go
@@ -934,8 +934,8 @@ func loadConfig() (*config, error) {
 		}
 	}
 
-	// Finally, ensure that we are only listening on localhost if Tor
-	// inbound support is enabled.
+	// Ensure that we are only listening on localhost if Tor inbound support
+	// is enabled.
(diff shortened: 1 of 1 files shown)


Empact:
utACK b0318e2 a test would be nice

ErikEk:
Added a PR #2143 to improve the color validation in parseHexColor.

ErikEk:
Now when #2143 is merged this looks good to me.

PR #2097, opened 2018-10-25
sample-lnd.conf dir update

reez:
**Documentation fix:**

On macOS the default directory seems to be `~/Library/Application Support/Lnd` instead of `~/Library/Application Support/Lnd/data`

Requested change implemented are just updating documentation.


wpaulino:
The documentation is correct as is. `~/Library/Application Support/Lnd` is the default directory for all of `lnd`'s files (data, logs, etc.), but `~/Library/Application Support/Lnd/data` is the default data directory.

PR #2106, opened 2018-10-26
sweep: Preauthorized Inputs

cfromknecht:
This PR adds a new `PreAuthorized` witness type and modifies the `sweep.Input` interfaces to support externally provided witness types. The previous `BuildWitness()` function has been split into two stages. 

`BuildWitness` now computes the witness and caches the witness stack inside the `sweep.Input`. A subsequent call to `Witness()` returns the witness if one has been computed or provided externally via `SetWitnessStack()`

These changes are needed as a preliminary step for watchtowers, since the tower doesn't have the signing keys to compute the witness. Instead, the witness is constructed after decrypting the encrypted blob, and set via `SetWitnessStack`. A later PR will be made the uses this functionality.

### DIFF
--- a/breacharbiter.go
+++ b/breacharbiter.go
@@ -761,6 +761,8 @@ type breachedOutput struct {
 	secondLevelWitnessScript []byte
 
 	witnessFunc lnwallet.WitnessGenerator
+
+	witnessStack [][]byte
 }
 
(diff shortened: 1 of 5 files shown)


cfromknecht:
Closing in favor an alternative approach

PR #2107, opened 2018-10-26
gRPC ruby doc updates

vwoo:
Fixed a typo in the documentation.

Also the Interceptor class example only worked with simple request/response calls like GetInfo.

I've updated it to work with server streaming calls like SubscribeChannelGraph.

The Interceptor class can also be updated to work with client streaming and bidirectional streaming calls but AFAIK LND doesn't have such calls.


Roasbeef:
> The Interceptor class can also be updated to work with client streaming and bidirectional streaming calls but AFAIK LND doesn't have such calls.

The existing bi-directional streaming call are `SendPayments` and `SendToRoute`.

PR #2109, opened 2018-10-27
config+rpcserver: adding a Private field to enable private nodes

### LINKED ISSUE #1298: Feature request: ability to disable forwarding for node
### Background

Some nodes don't want to necessarily forward payments, it would be nice to have a way of disabling forwarding all together.

ccdle12:
In this PR, we add a Private field in `config.go` to allow a node to 
create private channels by default and deploy a private node.
 
This also adds a check in `rpcserver.go` when `OpenChannel()` 
is called, to see if the Private field was set in the configuration.
If the field was set, then the channel will be created as private.

This will also prevent the private node from forwarding HTLCs 
unless the HTLC is meant for a node on the private channel.

Fixes #1298


Roasbeef:
Users can already make each channel un advertised via the command line or the direct RPC, do we really need this extra functionality? I think that issue wants to be able to have public channels, yet instruct the switch to cancel back all HTLCs, and also ensure that the side of the channel belonging to the node is disabled (channel update) at all times.

wpaulino:
Closing in favor of #2203.

PR #2111, opened 2018-10-28
Bump Go versions and use '.x' to always get latest patch versions

HaraldNordgren:


halseth:
Note that this change is already a part of https://github.com/lightningnetwork/lnd/pull/1892. That being said, it does make sense to make this change in a separate PR 😊

HaraldNordgren:
@halseth Thanks for letting me know! I'm all for "modularity". So several smaller changes to reach a goal seem like a reasonable approach to me.

cfromknecht:
LGTM 🎉re: coverage change, that's a known issue with the way coveralls measures go coverage. mostly it stems from probabilistic executions of code paths due to asynchrony. safe to ignore in this case

PR #2112, opened 2018-10-28
Add a maxconfirms argument to ListUnspentWitness

AdamISZ:
This change was inspired by #1984 - the underlying call to
ListUnspent supports a (min, max) range so it makes sense that
the WalletController interface can also support this; a
default no-maximum can be expressed using a MaxInt32 value.


AdamISZ:
@Roasbeef yes it is enough to do it without this change. "It" being #1984 for me.

That's how I did it, but @wpaulino suggested this alternative, so I guess it's between you two to figure out which is preferable :)

wpaulino:
My reason for preferring this is to prevent another linear scan just to satisfy `maxConfs`. I don’t think this interface change is intrusive at all. If the implementations of the interface can already filter out by `minConfs`, why not allow it to do `maxConfs` as well?

PR #2115, opened 2018-10-29
path finding: Disable bit decoupling

halseth:
This PR consist of the first part of #2091 (Strict channel enabling):

1. The path finding algorithm is modified to ignore the disable bit for local channels, allowing us to send payments on these channels even though they are publicly disabled. The assumption is that bandwidth hints of local channels will be set to a value reflecting their active status.

This essentially decouples the graph's disabled status from the local path finding, as the local channel status will be more up-to-date than what is found in the graph.

### DIFF
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -406,9 +406,17 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
 	// to our destination, respecting the recommendations from
 	// missionControl.
 	path, err := findPath(
-		nil, p.mc.graph, p.additionalEdges, p.mc.selfNode,
-		payment.Target, pruneView.vertexes, pruneView.edges,
-		payment.Amount, payment.FeeLimit, p.bandwidthHints,
+		&graphParams{
+			graph:           p.mc.graph,
+			additionalEdges: p.additionalEdges,
+			bandwidthHints:  p.bandwidthHints,
+		},
+		&restrictParams{
+			ignoredNodes: pruneView.vertexes,
+			ignoredEdges: pruneView.edges,
+			feeLimit:     payment.FeeLimit,
+		},
+		p.mc.selfNode, payment.Target, payment.Amount,
 	)
 	if err != nil {
 		return nil, err
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -446,6 +446,42 @@ func edgeWeight(lockedAmt lnwire.MilliSatoshi, fee lnwire.MilliSatoshi,
 	return int64(fee) + timeLockPenalty
 }
 
+// graphParams wraps the set of graph parameters passed to findPath.
+type graphParams struct {
+	// tx can be set to an existing db transaction. If not set, a new
+	// transaction will be started.
+	tx *bbolt.Tx
+
+	// graph is the ChannelGraph to be used during path finding.
+	graph *channeldb.ChannelGraph
+
+	// additionalEdges is an optional set of edges that should be
+	// considered during path finding, that is not already found in the
+	// channel graph.
+	additionalEdges map[Vertex][]*channeldb.ChannelEdgePolicy
+
+	// bandwidthHints is an optional map from channels to bandwidths that
+	// can be populated if the caller has a better estimate of the current
+	// channel bandwidth than what is found in the graph.
+	bandwidthHints map[uint64]lnwire.MilliSatoshi
+}
+
+// restrictParams wraps the set of restrictions passed to findPath that the
+// found path must adhere to.
+type restrictParams struct {
+	// ignoredNodes is an optional set of nodes that should be ignored if
+	// encountered during path finding.
+	ignoredNodes map[Vertex]struct{}
+
+	// ignoredEdges is an optional set of edges that should be ignored if
+	// encountered during path finding.
+	ignoredEdges map[uint64]struct{}
+
+	// feeLimit is a maximum fee amount allowed to be used on the path from
+	// the source to the target.
+	feeLimit lnwire.MilliSatoshi
+}
+
 // findPath attempts to find a path from the source node within the
 // ChannelGraph to the target node that's capable of supporting a payment of
 // `amt` value. The current approach implemented is modified version of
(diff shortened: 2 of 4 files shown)


routing/missioncontrol.go
```diff
@@ -373,9 +373,17 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
 	// to our destination, respecting the recommendations from
 	// missionControl.
 	path, err := findPath(
-		nil, p.mc.graph, p.additionalEdges, p.mc.selfNode,
-		payment.Target, pruneView.vertexes, pruneView.edges,
-		payment.Amount, payment.FeeLimit, p.bandwidthHints,
+		&findPathParams{
```
cfromknecht: nit: struct misspelled in commit msg

routing/pathfind_test.go
```diff
@@ -1501,6 +1501,112 @@ func TestRouteFailDisabledEdge(t *testing.T) {
 	}
 }
 
+// TestPathSourceEdgesBandwidth tests that explicitly passing in a set of
+// bandwidth hints is used by the path finding algorithm to consider whether to
+// use a local channel.
+func TestPathSourceEdgesBandwidth(t *testing.T) {
```
cfromknecht: nice test!

routing/pathfind.go
```diff
@@ -484,7 +484,11 @@ type findPathParams struct {
 
 	// bandwidthHints is an optional map from channels to bandwidths that
 	// can be populated if the caller has a better estimate of the current
-	// channel bandwidth than what is found in the graph.
+	// channel bandwidth than what is found in the graph. If set, it will
```
cfromknecht: nit: also misspelled in commit msg

routing/pathfind.go
```diff
@@ -446,6 +446,48 @@ func edgeWeight(lockedAmt lnwire.MilliSatoshi, fee lnwire.MilliSatoshi,
...
+	// feeLimit is a maximum fee amount allowed to be used on the path from
+	// the source to the target.
+	feeLimit lnwire.MilliSatoshi
+
+	// bandwidthHints is an optional map from channels to bandwidths that
+	// can be populated if the caller has a better estimate of the current
+	// channel bandwidth than what is found in the graph.
+	bandwidthHints map[uint64]lnwire.MilliSatoshi
+}
+
```
joostjager: This structs is quite a mix of parameters. How about creating two separate structs?

- graphData (`tx`, `graph`, `additionalEdges`, `bandwidthHints`)

- restrictions (`ignoredNodes`, `ignoredEdges`, `feeLimit`) (#1941 is adding a time lock limit to this)

And keep passing in `sourceNode`, `targetNode` and `amt` as direct function arguments. So 5 arguments in total.

routing/pathfind.go
```diff
@@ -484,7 +484,11 @@ type findPathParams struct {
 
 	// bandwidthHints is an optional map from channels to bandwidths that
 	// can be populated if the caller has a better estimate of the current
-	// channel bandwidth than what is found in the graph.
+	// channel bandwidth than what is found in the graph. If set, it will
+	// override the capacities and disabled flags found in the graph when
+	// doing path finding. In particular, it should be set to the current
```
joostjager: >If set, it will override the disabled flags found in the graph.

It isn't `bandwidthHints` that overrides the disabled flag, is it? Only the fact that a channel is local, overrides the disabled flag?

halseth: Yes, you're right. I forgot to update the comment with the latest revision. I was contemplating whether it would also make sense to let the bandwidth hint override disabled flags for non-local channels, but figured for now that is not needed (as we don't use it).

routing/pathfind.go
halseth: Done!

routing/pathfind.go
halseth: Updated!

routing/pathfind_test.go
```diff
@@ -1551,6 +1551,132 @@ func TestRouteFailDisabledEdge(t *testing.T) {
...
+	t.Parallel()
+
+	graph, err := parseTestGraph(basicGraphFilePath)
+	defer graph.cleanUp()
```
cfromknecht: i think this will panic if a non-nil error is returned, since we always return a nil `graph` in such cases

routing/pathfind_test.go
```diff
@@ -1551,6 +1551,132 @@ func TestRouteFailDisabledEdge(t *testing.T) {
...
+	bandwidths[roasToSongoku] = 2 * payAmt
+	_, e1, e2, err := graph.graph.FetchChannelEdgesByID(roasToSongoku)
+	if err != nil {
+		t.Fatalf("unable to fetch edge: %v", err)
+	}
+	e1.Flags |= lnwire.ChanUpdateDisabled
+	if err := graph.graph.UpdateEdgePolicy(e1); err != nil {
+		t.Fatalf("unable to update edge: %v", err)
+	}
+	e2.Flags |= lnwire.ChanUpdateDisabled
```
cfromknecht: why is it that we set the disabled bit in both directions, instead of just in the outbound direction?

routing/pathfind_test.go
halseth: Hm, yeah good catch. Added a commit to fix this all places in the file.

routing/pathfind_test.go
halseth: I did this to avoid having to loop up which direction is the outbound one, and I figured that the ability to find a path when it's disabled in both directions is a stronger test than only one direction.

cfromknecht: Fair enough, agreed!

joostjager:
I think as is it looks good, but I have two concerns about bandwidth hints:
1. `bandwidthhints` and `ignorededges` overlap. an offline edge could also be passed in as an ignored edges, which may be clearer than using `bandwidthhint` (as the offline edge does have a bandwidth other than 0)
2. `bandwidthhints` are bidirectional, which is not what it is in reality. maybe it could be redefined as `bandwidthhint` from smaller to larger node pubkey. the other way around, it is `capacity - hint`. This is also related to #2243 where directionality for ignored edges is added. 

(but these are preexisting issues)

halseth:
> bandwidthhints and ignorededges overlap. an offline edge could also be passed in as an ignored edges, which may be clearer than using bandwidthhint (as the offline edge does have a bandwidth other than 0)

Currently `bandwidthHints` are only used for local edges, while `ignoredEdges` are used for edges that fail during a `paymentSession` (most likely non-local edges?). I agree that this can be cleaned up. What about only passing in bandwidth hints, and set them to 0 for edges we want to ignore? We could also get rid of the ignored nodes by setting all their edges to bandwidth=0.

> bandwidthhints are bidirectional, which is not what it is in reality. maybe it could be redefined as bandwidthhint from smaller to larger node pubkey. the other way around, it is capacity - hint. This is also related to #2243 where directionality for ignored edges is added.

Good observation! We should definitely make them bidirectional, as future more advanced routing might make use of that.

Sounds like these two improvements can go together in a follow-up PR or two, as it shouldn't be changing the behaviour achieved here, and as you mentioned are preexisting issues.

joostjager:
Yes, agree that this can be follow up PRs.

For those PRs: We can choose to use `bandwidthHints` to ignore edges, but then we need to have separate hints for both directions of a channel. That way, you can have a channel that has zero bandwidth in both directions (meaning both directions are disabled). 

Alternatively, we could supply a bandwidth hint only for the direction smaller to greater pubkey and assume the opposite direction bandwidth as the capacity minus that bandwidth. In that case, it isn't possible anymore to disable both directions using the bandwidth hint.

halseth:
I think explicitly having one hint for each direction is cleaner. That let us treat the directions induvidually. Also seems like the implicit method could lead to problems with the channel reserve (since bandwidth then not necessarily will be capacity-opposite bandwidht).

PR #2120, opened 2018-10-29
Remote Tor server support

svewa:
Enables configurations where Tor is running on a remote server by 

a) allowing non-localhost listeners if Tor is not on localhost
b) including the IP of the listener in onion-setup

 As requested in #2073 a partial pr for this functionality only

### DIFF
--- a/config.go
+++ b/config.go
@@ -935,8 +935,8 @@ func loadConfig() (*config, error) {
 	}
 
 	// Finally, ensure that we are only listening on localhost if Tor
-	// inbound support is enabled.
-	if cfg.Tor.V2 || cfg.Tor.V3 {
+	// inbound support via Tor on localhost is enabled.
+	if (cfg.Tor.V2 || cfg.Tor.V3) && lncfg.IsLoopback(cfg.Tor.SOCKS) {
 		for _, addr := range cfg.Listeners {
 			if lncfg.IsLoopback(addr.String()) {
 				continue
@@ -944,7 +944,7 @@ func loadConfig() (*config, error) {
 
 			return nil, errors.New("lnd must *only* be listening " +
 				"on localhost when running with Tor inbound " +
-				"support enabled")
+				"support enabled, and Tor on localhost")
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


Roasbeef:
So this would allow user to connect to a remote Tor server over a _plaintext_ connection?

svewa:
This is useful for isolated networks, where lnd is operated on a box behind a Tor gateway, similar to the way Whonix operates.

Currently lnd supports remote Tor servers. The thing that is missing (and fixed by this pr) is the proper creation of onion services, sorry for the bad title. So even without this patch it is possible to participate in lightning via remote Tor, just with limited connectivity.

While it sounds really bad the way you put it, it is a very important feature imho. If you take a look on how Tor users got de-anonymized in recent operations, it is very common to target the protected application. If the lnd host gets compromised in a single host deployment it's anonymity is likely to be lost. Hence the way to go is isolation between the to-be-anonymized application and the anonymizing gateway (like Whonix)

I'm not too happy with the way the (hashed)passwortauthentication works, and I believe we need improvements beyond that on the Tor side as well (challenge/response auth, limited commands/macaroons). But it is still a huge improvement on the isolation of a hidden service.

qubenix:
Relevant: https://github.com/lightningnetwork/lnd/issues/2176

qubenix:
>a) allowing non-localhost listeners if Tor is not on localhost

This PR does not allow me to listen on `0.0.0.0` with `listen=0.0.0.0`, `tor.active=1`, and `tor.socks=remoteip:port`.

svewa:
> > a) allowing non-localhost listeners if Tor is not on localhost
> 
> This PR does not allow me to listen on `0.0.0.0` with `listen=0.0.0.0`, `tor.active=1`, and `tor.socks=remoteip:port`.

This patch should not be needed, unless you use --tor.v2 / --tor.v3 for automatic setup of the .onion. 
`lnd  --tor.socks=10.0.0.1 --listen=0.0.0.0 --bitcoin.active`
shold work. But this would not allow for inbound connections, and your address is not announced. So listening is useless anyway. Adding the .onion as an extra --listen parameter results in
`[ERR] SRVR: unable to create server: address [snip].onion: no suitable address found`

qubenix:
>This patch should not be needed, unless you use --tor.v2 / --tor.v3 for automatic setup of the .onion.

Why is that? If I set up my own onion instead of using the control port I still need to listen on an interface other than localhost if my tor is on a remote gateway.

>Adding the .onion as an extra --listen parameter results in

Onion should be added with `externalip=` not `listen=`.

Why aren't you using `tor.active=1`?

wpaulino:
Superseded by #2490.

PR #2122, opened 2018-10-29
[watchtower/blob] variable length sweep addr

cfromknecht:
This PR fixes an oversight in the previous
design of the watchtower blob, by introducing
a length byte for sweep addresses. The previous
format supposed that addresses would be padded
to 42 bytes, but had no indication of the
address's actual length.

To remedy this, we introduce a single byte
indicating the actual size of the address,
such that the padding can be removed upon
decoding.

In addition, the watchtower/blob unit tests are
modified to use ` t.Run` so that the unit test
names are properly displayed in logs.

### DIFF
--- a/watchtower/blob/justice_kit.go
+++ b/watchtower/blob/justice_kit.go
@@ -33,14 +33,19 @@ const (
 	CiphertextExpansion = 16
 
 	// V0PlaintextSize is the plaintext size of a version 0 encoded blob.
-	//    sweep address:                  42 bytes
+	//    sweep address length:            1 byte
+	//    padded sweep address:           42 bytes
 	//    revocation pubkey:              33 bytes
 	//    local delay pubkey:             33 bytes
 	//    csv delay:                       4 bytes
 	//    commit to-local revocation sig: 64 bytes
 	//    commit to-remote pubkey:        33 bytes, maybe blank
 	//    commit to-remote sig:           64 bytes, maybe blank
-	V0PlaintextSize = 273
+	V0PlaintextSize = 274
+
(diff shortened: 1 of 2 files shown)


watchtower/blob/justice_kit.go
```diff
@@ -33,14 +33,19 @@ const (
...
 	//    csv delay:                       4 bytes
 	//    commit to-local revocation sig: 64 bytes
 	//    commit to-remote pubkey:        33 bytes, maybe blank
 	//    commit to-remote sig:           64 bytes, maybe blank
-	V0PlaintextSize = 273
+	V0PlaintextSize = 274
+
+	// MaxSweepAddrSize defines the maximum sweep address size that can be
+	// encoded in a blob.
+	MaxSweepAddrSize = 42
```
Roasbeef: 👍 

Seems more future looking to just encode the entire serialized script (as this allows).

watchtower/blob/justice_kit_test.go
```diff
@@ -27,10 +27,20 @@ func makeSig(i int) lnwire.Sig {
 	return sig
 }
 
+func makeAddr(size int) []byte {
+	addr := make([]byte, size)
+	if _, err := io.ReadFull(rand.Reader, addr); err != nil {
```
Roasbeef: We may want to consider using the `math/rand` reader here. Not a blocker atm, but if we switch to this source, then the unit tests become I/O free as they no longer have to obtain any file descriptors etc. For the sake of moving forward, I think a TODO here is enough, possibly a blanket issue within `lnd` to do a sweep and remove instances of `crypto/rand` that aren't actually required.

PR #2123, opened 2018-10-30
[Tests] No forwarding of stale node announcements

halseth:
This commit adds a new test TestNodeAnnouncementNoChannels that ensures
a node announcement for an unknown node is not forwarded by the gossiper.

Also updates the router mock to mimic the Router's `IsStaleNode` behaviour.

### DIFF
--- a/discovery/gossiper_test.go
+++ b/discovery/gossiper_test.go
@@ -206,7 +206,18 @@ func (r *mockGraphSource) IsStaleNode(nodePub routing.Vertex, timestamp time.Tim
 		}
 	}
 
-	return false
+	// If we did not find the node among our existing graph nodes, we
+	// require the node to already have a channel in the graph to not be
+	// considered stale.
+	for _, info := range r.infos {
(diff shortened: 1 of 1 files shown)


discovery/gossiper_test.go
```diff
@@ -2418,6 +2429,109 @@ func TestExtraDataNodeAnnouncementValidation(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to process announcement: %v", err)
+	}
+
+	// This time the node announcement should be forwarded. The same should
+	// the channel announcement and update be.
+	for i := 0; i < 3; i++ {
+		select {
+		case <-ctx.broadcastedMessage:
+		case <-time.After(time.Second):
```
cfromknecht: should be `2*trickleDelay` like elsewhere?

halseth: In the cases we expect success we usually use a longer timeout I think.

PR #2128, opened 2018-10-30
queryroutes: set queryroutes numRoutes param default to 10

mrfelton:
## Problem

When calling `queryroutes` from gRPC, the default value for the `num_routes` parameter  is set to zero. Having a default value of zero is confusing as there would be little point of making the call if you wanted to fetch zero results. Additionally, the documentation states that the default should be 10.

This is inconsistent with the CLI, where the default of 10 is correctly applied.

Also, the documentation for the cli `queryroutes` command has bad output which this PR addresses as below:

**Output from  `lncli queryroutes --help` (before changes):**

```
--num_max_routes value     the max number of routes to be returned (default: 10) (default: 10)
```

## Solution
Set the default value as `10` when no value is provided by the caller.

**Updated output from  `lncli queryroutes --help` (after changes):**
```
--num_max_routes value     the max number of routes to be returned (default: 10)
```

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -2671,7 +2671,7 @@ var queryRoutesCommand = cli.Command{
 		},
 		cli.Int64Flag{
 			Name:  "num_max_routes",
-			Usage: "the max number of routes to be returned (default: 10)",
+			Usage: "the max number of routes to be returned",
 			Value: 10,
 		},
 		cli.Int64Flag{
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3443,6 +3443,12 @@ func (r *rpcServer) QueryRoutes(ctx context.Context,
 
 	feeLimit := calculateFeeLimit(in.FeeLimit, amtMSat)
 
+	// numRoutes will default to 10 if not specified explicitly.
+	numRoutesIn := uint32(in.NumRoutes)
+	if numRoutesIn == 0 {
+		numRoutesIn = 10
+	}
+
 	// Query the channel router for a possible path to the destination that
 	// can carry `in.Amt` satoshis _including_ the total fee required on
 	// the route.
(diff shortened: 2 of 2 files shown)


cmd/lncli/commands.go
```diff
@@ -2671,7 +2671,7 @@ var queryRoutesCommand = cli.Command{
 		},
 		cli.Int64Flag{
 			Name:  "num_max_routes",
-			Usage: "the max number of routes to be returned (default: 10)",
```
wpaulino: What's the rationale for removing this note?

lnrpc/rpc.pb.go
```diff
@@ -2956,7 +2956,7 @@ func (m *QueryRoutesRequest) GetNumRoutes() int32 {
 	if m != nil {
 		return m.NumRoutes
 	}
-	return 0
+	return 10
```
wpaulino: This unfortunately isn't the place to make this change as this file and others are generated automatically by the proto compiler. Instead, the check should be done at the [QueryRoutes RPC level](https://github.com/lightningnetwork/lnd/blob/832bd8101c889efb9ace39bd1f4364d34900f351/rpcserver.go#L3407).

cmd/lncli/commands.go
mrfelton: It doesn't need to be documented there because it is defined just below in the `value` param, which automatically gets appended to the cli help text output. See my original comments on the issue that shows the output before and after this change.

```
		cli.Int64Flag{
			Name:  "num_max_routes",
			Usage: "the max number of routes to be returned",
			Value: 10, <----- this
		},
```

lnrpc/rpc.pb.go
mrfelton: Thanks for the pointer @vapopov  - I've updated the PR.

Roasbeef:
IMO, it should just error out (at the RPC level) if a value isn't specified. In many places we keep cli level defaults distinct from actual RPC level defaults.

mrfelton:
@Roasbeef What would be a good reason to leave the default empty for RPC whilst setting it to 10 for CLI users? Seems like a better DX and UX to have both aligned, and to have a sensible default supplied (as is the case for the CLI right now).

I think diverging the interfaces in this way introduces unnecessary confusion.

wpaulino:
👍 for having a sensible default for the RPC itself, rather than returning an error.

Roasbeef:
Fails travis atm with: 
```
rpcserver.go:1::warning: file is not gofmted with -s (gofmt)
```

Should be fixed by running `make lint` or `make fmt` locally.

PR #2131, opened 2018-10-31
contractcourt/chain_arbitrator: prevent force closing same channel twice

wpaulino:
In this commit, we prevent the ChainArbitrator from sending a force
close request for a channel if it has previously already sent one. We do
this to prevent blocking the caller of ForceCloseContract.


contractcourt/chain_arbitrator.go
```diff
@@ -599,7 +599,12 @@ func (c *ChainArbitrator) ForceCloseContract(chanPoint wire.OutPoint) (*wire.Msg
 	arbitrator, ok := c.activeChannels[chanPoint]
 	c.Unlock()
 	if !ok {
-		return nil, fmt.Errorf("unable to find arbitrator")
+		return nil, errors.New("unable to find arbitrator")
+	}
+
+	if arbitrator.state != StateDefault {
```
halseth: This check is already done in the `channel arbitrator`. We should instead return the error on the err chan from there I think.

wpaulino: Fixed and test added.

PR #2132, opened 2018-10-31
htlcswitch: Eliminate misleading log statement

alexbosworth:
Eliminate incorrect log statement that announces sending and receiving of satoshis with inaccurate totals.


halseth:
Or fix the calculation?

PR #2135, opened 2018-11-02
[discovery] Isolate gossip rate limiting

cfromknecht:


discovery/gossiper.go
```diff
@@ -456,12 +456,67 @@ func (d *AuthenticatedGossiper) Stop() {
 func (d *AuthenticatedGossiper) ProcessRemoteAnnouncement(msg lnwire.Message,
 	peer lnpeer.Peer) chan error {
 
+	errChan := make(chan error, 1)
+
+	// For messages in the known set of channel series queries, we'll
+	// dispatch the message directly to the gossipSyncer, and skip the main
+	// processing loop.
+	switch m := msg.(type) {
```
halseth: I would be more confident in this change if we instead created a separate event loop for gossip query messages that we handed them off to.

cfromknecht: If they all go through one event loop then we’re back to square one. Are you suggesting one event loop per gossip syncer?

halseth: I was thinking one event loop per peer, but I see now that that might get very complex 🤔

PR #2137, opened 2018-11-02
rpcserver: fix error handling in bidirectional sendPayment

halseth:
This PR fixes various cases of missing payments and potential deadlocks within the bidirectional `sendPayment` in the `rpcserver`. See commit messages for detailed info.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3696,9 +3696,7 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 			select {
 			case <-reqQuit:
 				return
-			case <-r.quit:
-				errChan <- nil
-				return
+
 			default:
 				// Receive the next pending payment within the
 				// stream sent by the client. If we read the
@@ -3755,6 +3753,9 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 		select {
 		case err := <-errChan:
 			return err
+
+		case <-r.quit:
+			return errors.New("rpc server shutting down")
 
 		case payIntent := <-payChan:
 			// We launch a new goroutine to execute the current
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3680,18 +3680,23 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 		htlcSema <- struct{}{}
 	}
 
+	// We keep track of the running goroutines and set up a quit signal we
+	// can use to request them to exit if the method returns because of an
+	// encountered error.
+	var wg sync.WaitGroup
+	reqQuit := make(chan struct{})
+	defer close(reqQuit)
+
 	// Launch a new goroutine to handle reading new payment requests from
 	// the client. This way we can handle errors independently of blocking
 	// and waiting for the next payment request to come through.
-	reqQuit := make(chan struct{})
-	defer func() {
-		close(reqQuit)
-	}()
-
 	// TODO(joostjager): Callers expect result to come in in the same order
 	// as the request were sent, but this is far from guarantueed in the
 	// code below.
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
+
 		for {
 			select {
 			case <-reqQuit:
@@ -3778,11 +3783,18 @@ sendLoop:
 			// We launch a new goroutine to execute the current
 			// payment so we can continue to serve requests while
 			// this payment is being dispatched.
+			wg.Add(1)
 			go func() {
+				defer wg.Done()
+
 				// Attempt to grab a free semaphore slot, using
 				// a defer to eventually release the slot
 				// regardless of payment success.
-				<-htlcSema
+				select {
+				case <-htlcSema:
+				case <-reqQuit:
+					return
+				}
 				defer func() {
 					htlcSema <- struct{}{}
 				}()
@@ -3853,6 +3865,10 @@ sendLoop:
 			}()
 		}
 	}
+
+	// Wait for all goroutines to finish before closing the stream.
+	wg.Wait()
+	return nil
 }
 
 // SendPaymentSync is the synchronous non-streaming version of SendPayment.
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -2394,8 +2394,7 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 				} else if err != nil {
 					select {
 					case errChan <- err:
-					case <-reqQuit:
-						return
+					default:
```
cfromknecht: should we log the error instead of just dropping it in the default case? agree that the first error is all that matters in returning to caller, but would be nice to see if other things are failing as well

rpcserver.go
```diff
@@ -2441,7 +2442,14 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 		case <-r.quit:
 			return errors.New("rpc server shutting down")
 
-		case payIntent := <-payChan:
+		case payIntent, ok := <-payChan:
+			// If the receive loop is done, we break the send loop
+			// and wait for the ongoing payments to finish before
+			// exiting.
+			if !ok {
+				break sendLoop
```
cfromknecht: is returning equivalent?

rpcserver.go
```diff
@@ -2365,18 +2365,23 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 		htlcSema <- struct{}{}
 	}
 
+	// We keep track of the running goroutines and set up a quit signal we
+	// can use to request them to exit if the method returns because of an
+	// encountered error.
+	var wg sync.WaitGroup
+	reqQuit := make(chan struct{})
+	defer close(reqQuit)
```
cfromknecht: 👍

rpcserver.go
halseth: Good idea! Decided to log in both cases, as I imagine it is useful to see these error on the grpc server as well.

rpcserver.go
halseth: With the following commit it will no longer be, as we wait for goroutines to exit.

rpcserver.go
```diff
@@ -2482,12 +2525,22 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
...
+					default:
+					}
 					return
 				}
 			}()
 		}
 	}
+
+	// Wait for all goroutines to finish before closing the stream.
+	wg.Wait()
```
Crypt-iQ: Not meaning to gravedig or anything, but I was looking at this code today.  Won't this fail to be called in some circumstances and therefore not wait for all of the goroutines to finish?  Like if one of the payments failed to send and sent an error on `errChan`, then this send loop would call `return err` and therefore never call this.

halseth:
Yes, these issues are still present. Rebased.

rpcserver.go
```diff
@@ -3237,20 +3246,41 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
...
 			go func() {
+				defer wg.Done()
+
 				// Attempt to grab a free semaphore slot, using
 				// a defer to eventually release the slot
 				// regardless of payment success.
-				<-htlcSema
+				select {
+				case <-htlcSema:
+				case <-reqQuit:
```
cfromknecht: doesn't seem like this case can ever be executed, since `close(reqQuit)` is deferred but now we wait for the goroutines to exit. i think this is fixed by deferring `wg.Wait()` before deferring the close

rpcserver.go
```diff
@@ -3168,39 +3168,44 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 		htlcSema <- struct{}{}
 	}
 
+	// We keep track of the running goroutines and set up a quit signal we
+	// can use to request them to exit if the method returns because of an
+	// encountered error.
+	var wg sync.WaitGroup
+	reqQuit := make(chan struct{})
```
bhandras: This is a bit tricky and may not be obvious from the comment (reqQuit signalling quit on this execution line upon returning). I had to read thorugh the long function to find the one place where this might happen. Maybe add consider adding a comment there too.

rpcserver.go
```diff
@@ -3216,18 +3221,22 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
...
+							"sending on "+
+							"stream: %v", err)
+
 						select {
 						case errChan <- err:
-						case <-reqQuit:
-							return
+						default:
 						}
+						return
```
bhandras: IMHO it's a bit cleaner if the return is in the select scope. (Disregard my comment if this is common stlye in this code base.)

rpcserver.go
bhandras: If I understand this correctly, there's a return on the main execution line outside of the goroutines (line 3256 in the current PR state). That return indicates an error and signals all gorotutines to quit as soon as possible.

rpcserver.go
bhandras: Sorry, got confused here, this is OK, please discard my previous comment.

bhandras:
Only nits, LGTM.

rpcserver.go
halseth: Yeah, the waitgroup is only being waited on in the non-error case, where there are no payments left to send. This handle the error cases where we return early.

rpcserver.go
```diff
@@ -3237,6 +3243,9 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 
 	for {
 		select {
+
+		// If we encounter and error either during sending or
+		// receiving, we return directly, clsoing the stream.
```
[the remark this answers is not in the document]
halseth: Fixed

halseth:
back from the dead!

rpcserver.go
cfromknecht: got it thanks for the explanation, makes sense

PR #2143, opened 2018-11-04
Improving color validation

ErikEk:
Improved the color hex format validation code using regexp.

### DIFF
--- a/server.go
+++ b/server.go
@@ -10,6 +10,7 @@ import (
 	"math/big"
 	"net"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -70,6 +71,10 @@ var (
 	// ErrServerShuttingDown indicates that the server is in the process of
 	// gracefully exiting.
 	ErrServerShuttingDown = errors.New("server is shutting down")
+
+	// validColorRegexp is a regexp that lets you check if a particular
+	// color string matches the standard hex color format #RRGGBB.
+	validColorRegexp = regexp.MustCompile("^#[A-Fa-f0-9]{6}$")
 )
 
 // server is the main server of the Lightning Network Daemon. The server houses
@@ -2821,8 +2826,10 @@ func (s *server) Peers() []*peer {
 // form "#RRGGBB", parses the hex color values, and returns a color.RGBA
 // struct of the same color.
 func parseHexColor(colorStr string) (color.RGBA, error) {
-	if len(colorStr) != 7 || colorStr[0] != '#' {
-		return color.RGBA{}, errors.New("Color must be in format #RRGGBB")
+	// Check if the hex color string is a valid color representation.
+	if !validColorRegexp.MatchString(colorStr) {
+		return color.RGBA{}, errors.New("Color must be specified " +
+			"using a hexadecimal value in the form #RRGGBB")
 	}
 
 	// Decode the hex color string to bytes.
--- a/server_test.go
+++ b/server_test.go
@@ -5,32 +5,39 @@ package main
 import "testing"
 
 func TestParseHexColor(t *testing.T) {
-	empty := ""
-	color, err := parseHexColor(empty)
-	if err == nil {
-		t.Fatalf("Empty color string should return error, but did not")
+	var colorTestCases = []struct {
+		test  string
+		valid bool // If valid format
+		R     byte
+		G     byte
+		B     byte
+	}{
+		{"#123", false, 0, 0, 0},
+		{"#1234567", false, 0, 0, 0},
+		{"$123456", false, 0, 0, 0},
+		{"#12345+", false, 0, 0, 0},
+		{"#fFGG00", false, 0, 0, 0},
+		{"", false, 0, 0, 0},
+		{"#123456", true, 0x12, 0x34, 0x56},
+		{"#C0FfeE", true, 0xc0, 0xff, 0xee},
 	}
 
-	tooLong := "#1234567"
-	color, err = parseHexColor(tooLong)
-	if err == nil {
-		t.Fatalf("Invalid color string %s should return error, but did not",
-			tooLong)
-	}
+	// Perform the table driven tests.
+	for _, ct := range colorTestCases {
 
-	invalidFormat := "$123456"
-	color, err = parseHexColor(invalidFormat)
-	if err == nil {
-		t.Fatalf("Invalid color string %s should return error, but did not",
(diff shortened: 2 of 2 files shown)


Roasbeef:
Thanks for the PR! What do you think of fixing #2082 as well?

ErikEk:
It's already beeing addressed in #2094 by connor.

cfromknecht:
Thanks @ErikEk! Indeed i think this is complementary to #2094! Since we're modifying `parseHexColor`, would also be great to add some negative tests to https://github.com/lightningnetwork/lnd/blob/master/server_test.go#L7. Bonus points for making a table driven test :)

server.go
```diff
@@ -2821,7 +2822,8 @@ func (s *server) Peers() []*peer {
 // form "#RRGGBB", parses the hex color values, and returns a color.RGBA
 // struct of the same color.
 func parseHexColor(colorStr string) (color.RGBA, error) {
-	if len(colorStr) != 7 || colorStr[0] != '#' {
+	// Check if the hex color string is a valid color representation
+	if res, _ := regexp.MatchString(`^#[A-Fa-f0-9]{6}$`, colorStr); res == false {
```
cfromknecht: what's the rationale for not handling the error value?

server.go
```diff
@@ -2821,7 +2822,8 @@ func (s *server) Peers() []*peer {
 // form "#RRGGBB", parses the hex color values, and returns a color.RGBA
 // struct of the same color.
 func parseHexColor(colorStr string) (color.RGBA, error) {
-	if len(colorStr) != 7 || colorStr[0] != '#' {
+	// Check if the hex color string is a valid color representation
```
cfromknecht: style nit: missing period

cfromknecht:
Looks like the test already includes some negative tests, though your proposal will also catch values that are the correct length but not within the accepted character set. would be a great addition to the test case

server.go
cfromknecht: I see, error value is returned if the regex doesn't compile. to get around this, we can consider using `MustCompile`https://golang.org/pkg/regexp/#MustCompile to safely initialize the pattern

ErikEk:
Don't review the latest commit. I sent the wrong code.

server_test.go
```diff
@@ -4,33 +4,44 @@ package main
...
+	{"#C0FfeE", true, 0xc0, 0xff, 0xee},
+}
+
 func TestParseHexColor(t *testing.T) {
-	empty := ""
-	color, err := parseHexColor(empty)
-	if err == nil {
-		t.Fatalf("Empty color string should return error, but did not")
-	}
+	// Perform the table driven tests
```
halseth: nit: period at end of sentence.

server_test.go
```diff
@@ -4,33 +4,44 @@ package main
 
 import "testing"
 
+var colortests = []struct {
+	in    string
```
halseth: style nit: usually name this field `test` :)

server_test.go
```diff
@@ -4,33 +4,44 @@ package main
 
 import "testing"
 
+var colortests = []struct {
```
halseth: naming suggestion nit: `colorTestCases`

server_test.go
```diff
@@ -4,33 +4,44 @@ package main
...
-	tooLong := "#1234567"
-	color, err = parseHexColor(tooLong)
-	if err == nil {
-		t.Fatalf("Invalid color string %s should return error, but did not",
-			tooLong)
-	}
+		color, err := parseHexColor(ct.in)
+		if !ct.valid && err == nil {
+			if ct.in != "" {
+				t.Fatalf("Invalid color string %s should return "+
```
halseth: Just do `t.Fatalf("Invalid color string <%s> should return "...` and you don't have to special case the empty string case.

ErikEk: @halseth Thanks for the review. I've included your suggestions.

server.go
```diff
@@ -2821,8 +2822,14 @@ func (s *server) Peers() []*peer {
 // form "#RRGGBB", parses the hex color values, and returns a color.RGBA
 // struct of the same color.
 func parseHexColor(colorStr string) (color.RGBA, error) {
-	if len(colorStr) != 7 || colorStr[0] != '#' {
-		return color.RGBA{}, errors.New("Color must be in format #RRGGBB")
+	// Parses a regular expression and returns a Regexp object.
+	// Panics if the expression cannot be parsed.
+	validColorRegexp := regexp.MustCompile("^#[A-Fa-f0-9]{6}$")
```
cfromknecht: would recommend moving this to a global variable, instead of calling `MustCompile` in this method. rationale is that we'd rather `MustCompile` panic when initializing the runtime, as opposed to when the function is invoked

server_test.go
```diff
@@ -4,33 +4,40 @@ package main
 
 import "testing"
 
+var colorTestCases = []struct {
```
cfromknecht: great work, nice set of test!

server.go
```diff
@@ -70,6 +71,12 @@ var (
 	// ErrServerShuttingDown indicates that the server is in the process of
 	// gracefully exiting.
 	ErrServerShuttingDown = errors.New("server is shutting down")
+
+	// Parses a regular expression. The functions in this module let you
+	// check if a particular color string matches the standard hex color
+	// format #RRGGBB.
+	// Panics if the expression cannot be parsed.
```
halseth: this sentence is unnecessary.

server.go
```diff
@@ -70,6 +71,12 @@ var (
 	// ErrServerShuttingDown indicates that the server is in the process of
 	// gracefully exiting.
 	ErrServerShuttingDown = errors.New("server is shutting down")
+
+	// Parses a regular expression. The functions in this module let you
```
halseth: godocs should start with `// validColorRegexp is a regexp that...`

ErikEk: @halseth Done

server_test.go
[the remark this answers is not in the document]
ErikEk: @wpaulino Not really. If you want me to move the struct into the function I'll do that.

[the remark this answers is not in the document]
wpaulino: Sure, would say makes sense to not pollute the global namespace.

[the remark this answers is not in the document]
ErikEk: @wpaulino There you go. Thanks for the review.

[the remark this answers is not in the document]
halseth: Yep, agreed! 👍

Issue #2144, opened 2018-11-04
Handling of CLTV expiry too soon


jimpo:
I think the link handling of cases where an HTLC's timeout is too soon has some issues.

When a node is the final hop in a payment circuit, it fails the payment if the CLTV is too soon, by checking against the constant `expiryGraceDelta` [here](https://github.com/lightningnetwork/lnd/blob/b6009850635cf966eb3bd637de5da89cb9262589/htlcswitch/link.go#L2292). But `expiryGraceDelta` is only 2 and it seems like it should certainly be greater than BroadcastDelta in ChainArbitrator, for which the default value is 10. Actually, it should be larger than `BroadcastDelta * redeemMultiplier` [which is used when redeeming](https://github.com/lightningnetwork/lnd/blob/master/contractcourt/channel_arbitrator.go#L905).

Secondly, when forwarding a payment, it also checks if the CLTV is too soon, but in this check it uses the outbound channel TimelockDelta (see [here](https://github.com/lightningnetwork/lnd/blob/b6009850635cf966eb3bd637de5da89cb9262589/htlcswitch/link.go#L1946)) and it should probably use the same `expiryGraceDelta` as in the final hop case once that is raised to a safe value.

jimpo:
Hmm, there's actually a more difficult issue, which is that the default final_cltv_expiry per BOLT 11 is 9 blocks, so when constructing routes without a specified final_cltv_expiry, the final CLTV will be current height (at time of route construction) + 9. But 9 is less than the default broadcast delta of 10, so an lnd node should either reject the HTLC for having a CLTV too soon or reveal the preimage and immediately go to chain to unilaterally close b/c it's within the broadcast delta.

I think this means that either the default broadcast delta should be decreased to < 9 or lnd should only generate invoices with a `c` tag that is greater than the default expiry grace delta.

PR #2145, opened 2018-11-04
Improve enforcement of CLTV too soon rejects

jimpo:
As explained in #2144, the CLTV too soon failure checks need to enforce longer periods between HTLCs accepted/offered and the current block height. Currently, the grace period for forwarded HTLCs is 0 blocks and only 2 blocks when we are the final hop. However, the default broadcast delta is 10 blocks. This means that we might forward an HTLC and add it on an outbound channel, then immediately go to chain to close unilaterally because the CLTV is within the broadcast delta from the current block height.

This PR sets ExpiryGraceDelta as a configuration on links, and the server sets it so that it is greater than the broadcast delta.

### DIFF
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -13,15 +13,6 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
-const (
-	// broadcastRedeemMultiplier is the additional factor that we'll scale
-	// the normal broadcastDelta by when deciding whether or not to
-	// broadcast a commitment to claim an HTLC on-chain. We use a scaled
-	// value, as when redeeming we want to ensure that we have enough time
-	// to redeem the HTLC, well before it times out.
-	broadcastRedeemMultiplier = 2
-)
-
 var (
 	// errAlreadyForceClosed is an error returned when we attempt to force
 	// close a channel that's already in the process of doing so.
@@ -910,7 +901,6 @@ func (c *ChannelArbitrator) checkChainActions(height uint32,
 		"height=%v", c.cfg.ChanPoint, height)
 
 	actionMap := make(ChainActionMap)
-	redeemCutoff := c.cfg.BroadcastDelta * broadcastRedeemMultiplier
 
 	// First, we'll make an initial pass over the set of incoming and
 	// outgoing HTLC's to decide if we need to go on chain at all.
@@ -943,7 +933,7 @@ func (c *ChannelArbitrator) checkChainActions(height uint32,
 			continue
 		}
 		haveChainActions = haveChainActions || c.shouldGoOnChain(
-			htlc.RefundTimeout, redeemCutoff, height,
+			htlc.RefundTimeout, c.cfg.BroadcastDelta, height,
 		)
 	}
 
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -26,16 +26,6 @@ func init() {
 }
 
 const (
-	// expiryGraceDelta is a grace period that the timeout of incoming
-	// HTLC's that pay directly to us (i.e we're the "exit node") must up
-	// hold. We'll reject any HTLC's who's timeout minus this value is less
-	// that or equal to the current block height. We require this in order
-	// to ensure that if the extending party goes to the chain, then we'll
-	// be able to claim the HTLC still.
-	//
-	// TODO(roasbeef): must be < default delta
-	expiryGraceDelta = 2
-
 	// maxCltvExpiry is the maximum outgoing time lock that the node accepts
 	// for forwarded payments. The value is relative to the current block
 	// height. The reason to have a maximum is to prevent funds getting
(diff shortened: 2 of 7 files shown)


Roasbeef:
New failure in the integration tests due to this PR: 
```
    --- FAIL: TestLightningNetworkDaemon/multi-hop_htlc_error_propagation (6.63s)
        lnd_test.go:74: Failed: (multi-hop htlc error propagation): exited with error: 
            *errors.errorString payment should have failed due to unknown payment hash, instead failed due to: unable to route payment to destination: ExpiryTooSoon(update=(lnwire.ChannelUpdate) {
```

contractcourt/channel_arbitrator.go
```diff
@@ -12,15 +12,6 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
-const (
-	// broadcastRedeemMultiplier is the additional factor that we'll scale
```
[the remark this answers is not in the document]
jimpo: Hmm, OK, I buy the idea that broadcasting for outgoing HTLCs is less urgent than for incoming since you have the CLTV delta on the other side to make sure you don't lose money. Can I flip it so the broadcast delta becomes 20 and it gets halved in the timeout case instead of getting doubled in the incoming case? So that BroadcastDelta can be taken as a safe upper bound on the broadcast delta in all cases.

htlcswitch/link.go
```diff
@@ -1942,11 +1944,10 @@ func (l *channelLink) HtlcSatifiesPolicy(payHash [32]byte,
 	// We want to avoid accepting an HTLC which will expire in the near
 	// future, so we'll reject an HTLC if its expiration time is too close
 	// to the current height.
```
joostjager: Is this comment accurate still? The check seems to make sure that we are not creating an outgoing htlc that is close to the broadcast point. This is a stronger check than checking the incoming htlc, because the incoming expiration is always after the outgoing.

peer.go
```diff
@@ -573,6 +580,7 @@ func (p *peer) addLink(chanPoint *wire.OutPoint,
 		UnsafeReplay:        cfg.UnsafeReplay,
 		MinFeeUpdateTimeout: htlcswitch.DefaultMinLinkFeeUpdateTimeout,
 		MaxFeeUpdateTimeout: htlcswitch.DefaultMaxLinkFeeUpdateTimeout,
+		ExpiryGraceDelta:    defaultBroadcastDelta + extraExpiryGraceDelta,
```
joostjager: How do we know that `defaultBroadcastDelta` is indeed the value that is used in `ChainArbitrator`? It currently is all the same value, but if the idea is that all configuration is done in `server.go`, shouldn't this value then also be passed in from `server.go`?

htlcswitch/link.go
```diff
@@ -237,6 +227,18 @@ type ChannelLinkConfig struct {
 	// fee rate. A random timeout will be selected between these values.
 	MinFeeUpdateTimeout time.Duration
 	MaxFeeUpdateTimeout time.Duration
+
+	// ExpiryGraceDelta is the minimum difference between the current block
+	// height and the CLTV of an incoming HTLC we require to forward it as an
+	// intermediary hop or reveal the preimage as the final hop. We'll reject
```
joostjager: It is also the minimum diff for our own outgoing htlcs that we create?

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: Isn't this race condition why we need to keep watching for remote spends even after we publish our sweep transaction? (which is implemented in the new sweeper)

htlcswitch/link.go
jimpo: Good catch.

peer.go
jimpo: My assumption was that since peer and server are both in the main package, server refers directly to defaults, and there's no existing peer config, that it'd be OK. But I agree that's it's probably better to pass it into `newPeer`. I'll make that change.

Roasbeef:
Seem to be getting this test failure now semi reliably: 
```
    --- FAIL: TestLightningNetworkDaemon/test_multi-hop_htlc_receiver_chain_claim (27.19s)
        lnd_test.go:77: Failed: (test multi-hop htlc receiver chain claim): exited with error: 
            *errors.errorString expected transaction not found in mempool: wanted 1, found 0 txs in mempool: []
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd_test.go:9015 (0xbc5df2)
```

May be able to dig into it later sometime this week.

htlcswitch/link.go
```diff
@@ -1939,14 +1941,13 @@ func (l *channelLink) HtlcSatifiesPolicy(payHash [32]byte,
...
 
-	// We want to avoid accepting an HTLC which will expire in the near
-	// future, so we'll reject an HTLC if its expiration time is too close
-	// to the current height.
-	timeDelta := policy.TimeLockDelta
-	if incomingTimeout-timeDelta <= heightNow {
+	// We want to avoid offering an HTLC which will expire in the near
+	// future, so we'll reject an HTLC if the outgoing expiration time is too
+	// close to the current height.
+	if outgoingTimeout-l.cfg.ExpiryGraceDelta <= heightNow {
```
cfromknecht: Should we make this a strict inequality? Otherwise, this is equivalent to:
```golang
outgoingTimeout < heightNow + l.cfg.ExpiryGraceDelta + 1
```

The previous constant in the switch was 2, and via the non-strict inequality was actually enforcing 3. Now that the constant in main is 3, this is actually now `4+broadcastDelta`. I assume the intent is to keep this as 3?

Alternatively, the constant in main could be set to 2 again and maintain the current grace delta w/o changing the code here.

jimpo: Oh, I think I just changed it from 2 to 3 because 3 felt like a safer buffer but didn't give much thought to it. I'm fine changing it back to 2. Slight preference for non-strict inequality.

joostjager:
It may not be significant, but there is also a small window between accepting an incoming htlc on the channel and checking the expiry of the htlc and failing it in case it violates the rules. If a new block arrives in this window, the channel arbitrator may go to chain if the expiry is withing the broadcast delta.

joostjager:
Rebased and restructured commits in https://github.com/lightningnetwork/lnd/pull/2839

Roasbeef:
Closing in favor of #2839.

Issue #2134, opened 2018-11-01
Make `RPCServer` gRPC methods context aware


halseth:
The `RPCServer` within `lnd` acts as a `gRPC` server that serves clients such as `lncli`. Currently some of the long running methods in the `RPCServer` will keep executing until they finish or encounter an error, even though the client has canceled the API call.

We should make the `RPCServer` context aware, ceasing execution when the `gRPC` context is canceled.

## Steps to completion
1. Make the server properly listen for stream closes in case of streaming APIs, or context cancelation in case of non-streaming. 
2. Stop any long running task in case of cancelation, and clean up (more information: https://grpc.io/docs/tutorials/basic/go.html).

Companion issue: https://github.com/lightningnetwork/lnd/issues/2133

mrwhythat:
@halseth should listening to stream closes/context cancellation be done manually on a per-handler basis (i.e. just add `case <-ctx.Done` for all `select` statements in RPC handlers)? Another option is to implement a generic task killing mechanism but since we can't normally kill goroutines in Golang, this will be too hacky and dangerous. Maybe there are other options I'm not aware of?

halseth:
I was thinking more like the first option. For select statements within the `rpcServer` we should listen on `ctx.Done()`, while if we call into other (long-running) parts of LND, we should supply a `quit` channel.

Issue #2133, opened 2018-11-01
[Feature request] Cancel gRPC contex on lncli ctrl-C


halseth:
When executing a long-running `lncli` command, it can be canceled by a regular `ctrl-C`. However, we currently don't cancel the `gRPC` context in this case, so whatever API that was triggered within `lnd` will keep executing the command.

We should properly capture the `interrupt` and cancel the `gRPC` context when `ctrl-C` is pressed.

## Steps to completion

1. Change all gRPC calls made by `lncli` to use `context.WithCancel` instead of `context.Background()`.
2. Listen for OS interrupts each time a `lncli` command is launched. Similar to what is already done in `lnd`: https://github.com/lightningnetwork/lnd/blob/master/signal/signal.go
3. When the shutdown is requested, immediately cancel the `gRPC` call's context and close the `gRPC` connection before exiting.

Companion issue: https://github.com/lightningnetwork/lnd/issues/2134

PR #2147, opened 2018-11-04
lncli: task cancellation

rodentrabies:
This is an attempt at implementing #2133 and #2134.

Progress:
- [x] `lncli` interrupt listener;
- [x] wrap context with interrupt listener for all `lncli` RPC calls;
- [x] `lnd` cancelled context / closed stream handling.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -40,13 +40,13 @@ import (
 const defaultRecoveryWindow int32 = 2500
 
 func getContext() context.Context {
-	ctxb, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(context.Background())
 	go func() {
 		<-signal.ShutdownChannel()
 		cancel()
 
 	}()
-	return ctxb
+	return ctx
 }
 
 func printJSON(resp interface{}) {
@@ -128,12 +128,12 @@ var newAddressCommand = cli.Command{
 	Action: actionDecorator(newAddress),
 }
 
-func newAddress(ctx *cli.Context) error {
-	ctxb := getContext()
-	client, cleanUp := getClient(ctx)
+func newAddress(cliCtx *cli.Context) error {
+	ctx := getContext()
+	client, cleanUp := getClient(cliCtx)
 	defer cleanUp()
 
-	stringAddrType := ctx.Args().First()
+	stringAddrType := cliCtx.Args().First()
 
 	// Map the string encoded address type, to the concrete typed address
 	// type enum. An unrecognized address type will result in an error.
@@ -148,7 +148,7 @@ func newAddress(ctx *cli.Context) error {
 			"are: p2wkh and np2wkh", stringAddrType)
 	}
 
-	addr, err := client.NewAddress(ctxb, &lnrpc.NewAddressRequest{
+	addr, err := client.NewAddress(ctx, &lnrpc.NewAddressRequest{
 		Type: addrType,
 	})
 	if err != nil {
@@ -181,21 +181,21 @@ var estimateFeeCommand = cli.Command{
 	Action: actionDecorator(estimateFees),
 }
 
-func estimateFees(ctx *cli.Context) error {
+func estimateFees(cliCtx *cli.Context) error {
 	var amountToAddr map[string]int64
 
-	jsonMap := ctx.Args().First()
+	jsonMap := cliCtx.Args().First()
 	if err := json.Unmarshal([]byte(jsonMap), &amountToAddr); err != nil {
 		return err
 	}
 
-	ctxb := getContext()
-	client, cleanUp := getClient(ctx)
+	ctx := getContext()
+	client, cleanUp := getClient(cliCtx)
 	defer cleanUp()
 
-	resp, err := client.EstimateFee(ctxb, &lnrpc.EstimateFeeRequest{
+	resp, err := client.EstimateFee(ctx, &lnrpc.EstimateFeeRequest{
 		AddrToAmount: amountToAddr,
-		TargetConf:   int32(ctx.Int64("conf_target")),
+		TargetConf:   int32(cliCtx.Int64("conf_target")),
 	})
 	if err != nil {
 		return err
@@ -257,27 +257,27 @@ var sendCoinsCommand = cli.Command{
 	Action: actionDecorator(sendCoins),
 }
 
-func sendCoins(ctx *cli.Context) error {
+func sendCoins(cliCtx *cli.Context) error {
 	var (
 		addr string
 		amt  int64
 		err  error
 	)
-	args := ctx.Args()
+	args := cliCtx.Args()
 
-	if ctx.NArg() == 0 && ctx.NumFlags() == 0 {
-		cli.ShowCommandHelp(ctx, "sendcoins")
+	if cliCtx.NArg() == 0 && cliCtx.NumFlags() == 0 {
+		cli.ShowCommandHelp(cliCtx, "sendcoins")
 		return nil
 	}
 
-	if ctx.IsSet("conf_target") && ctx.IsSet("sat_per_byte") {
+	if cliCtx.IsSet("conf_target") && cliCtx.IsSet("sat_per_byte") {
 		return fmt.Errorf("either conf_target or sat_per_byte should be " +
 			"set, but not both")
 	}
 
 	switch {
-	case ctx.IsSet("addr"):
-		addr = ctx.String("addr")
+	case cliCtx.IsSet("addr"):
+		addr = cliCtx.String("addr")
 	case args.Present():
 		addr = args.First()
 		args = args.Tail()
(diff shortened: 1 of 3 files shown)


cmd/lncli/commands.go
```diff
@@ -1536,7 +1565,9 @@ func walletBalance(ctx *cli.Context) error {
 	defer cleanUp()
 
 	req := &lnrpc.WalletBalanceRequest{}
-	resp, err := client.WalletBalance(ctxb, req)
+	resp, err := handleSignals(ctxb, func(ctx context.Context) (proto.Message, error) {
```
halseth: I think a better way of handling this would be to define a new method `getContext()` that returns a `context` that will be canceled in case of an interrupt. Passing `ctx` to `client.WalletBalance(ctx, req)` should then stop correctly.

cmd/lncli/commands.go
```diff
@@ -46,6 +47,15 @@ func printJSON(resp interface{}) {
 	out.WriteTo(os.Stdout)
 }
 
+func getContext() context.Context {
```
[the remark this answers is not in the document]
halseth: Yep, exactly! 👍

halseth:
@mrwhythat I think that will entail way to much work, see my suggestion in the issue :)

cmd/lncli/commands.go
```diff
@@ -1177,7 +1179,7 @@ func monowidthColumns(words []string, ncols int) []string {
 }
 
 func create(ctx *cli.Context) error {
-	ctxb := context.Background()
+	ctxc := getContext()
```
halseth: nit: convention is naming it `ctx`

cmd/lncli/commands.go
```diff
@@ -2855,11 +2841,10 @@ var listChainTxnsCommand = cli.Command{
 }
 
 func listChainTxns(ctx *cli.Context) error {
-	ctxb := context.Background()
 	client, cleanUp := getClient(ctx)
 	defer cleanUp()
 
-	resp, err := client.GetTransactions(ctxb, &lnrpc.GetTransactionsRequest{})
+	resp, err := client.GetTransactions(getContext(), &lnrpc.GetTransactionsRequest{})
```
halseth: I also think I would prefer a `ctx := getContext()` at the start of the method, then use that directly here.

cmd/lncli/commands.go
halseth: Ah, true! :) 

Maybe use `ctxt`?

cmd/lncli/commands.go
mrwhythat: Done.

cmd/lncli/commands.go
mrwhythat: While I'm at it, I refactored it to use conventional naming in 6b5d165 and 019dec2 and it turns out to be not that big a diff, so if it's ok, this can be merged too.

mrwhythat: Variable renaming is pretty easy when you have a static type system:)

cmd/lncli/commands.go
```diff
@@ -122,12 +122,12 @@ var newAddressCommand = cli.Command{
 	Action: actionDecorator(newAddress),
 }
 
-func newAddress(ctx *cli.Context) error {
+func newAddress(clictx *cli.Context) error {
```
halseth: I would name this `cliCtx`, and also do this commit first. Then we could make the other context named `ctx` without first naming it `ctxt`.

cmd/lncli/commands.go
```diff
@@ -123,7 +123,7 @@ var newAddressCommand = cli.Command{
 }
 
 func newAddress(clictx *cli.Context) error {
-	ctxt := getContext()
+	ctx := getContext()
```
halseth: this should be squashed with the first commit.

mrwhythat: done

cmd/lncli/commands.go
mrwhythat: done

halseth:
This would be good to finally get in. We should start with a rebase, then check whether the places we wait for the context done signal are safe to return from.

mrwhythat:
I completely forgot about this PR. Will work on that this weekend.

mrwhythat:
Rebase is done, still have to recheck all the `rpcserver.go` changes.

cmd/lncli/commands.go
```diff
@@ -266,8 +266,7 @@ func sendCoins(cliCtx *cli.Context) error {
 	args := cliCtx.Args()
 
 	if cliCtx.NArg() == 0 && cliCtx.NumFlags() == 0 {
-		cli.ShowCommandHelp(cliCtx, "sendcoins")
-		return nil
+		return cli.ShowCommandHelp(cliCtx, "sendcoins")
```
halseth, line 269: do these ever return an error?

rpcserver.go
```diff
@@ -4263,6 +4274,13 @@ func (r *rpcServer) sendPayment(stream *paymentStream) error {
 			case <-reqQuit:
 				return
 
+			case <-ctx.Done():
```
halseth, line 4277: This kinda _looks_ correct, but it is also a bit scary change to make.

I would suggest breaking this PR up a bit:
1. The lncli changes are large enough in themselves, and we can add the context awareness for a few "safe" RPCs (for instance the Subscribe calls), that don't write state.
2. A follow-up PR that adds context awareness to more RPCs can be made, when we are confident that the first context cancellation works as intended.

cmd/lncli/commands.go
```diff
@@ -38,6 +39,16 @@ import (
 
 const defaultRecoveryWindow int32 = 2500
 
+func getContext() context.Context {
+	ctxb, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-signal.ShutdownChannel()
+		cancel()
+
```
guggero, line 47: nit: extra newline.

cmd/lncli/commands.go
```diff
@@ -118,6 +129,7 @@ var newAddressCommand = cli.Command{
 }
 
 func newAddress(ctx *cli.Context) error {
+	ctxb := getContext()
```
guggero: The commands have been split up into multiple files since this PR was originally created. Would be great to also apply these changes to the commands in the other files (just one example: `cmd/lncli/cmd_pay.go`).

lntest/itest/log_error_whitelist.txt
```diff
@@ -153,14 +153,24 @@
...
 <time> [ERR] RPCS: [/lnrpc.Lightning/SendToRoute]: rpc server shutting down
 <time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelEvents]: rpc error: code = Canceled desc = context canceled
-<time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelGraph]: ChannelRouter shutting down
-<time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelGraph]: router not started
+<time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelEvents]: rpc error: code = DeadlineExceeded desc = context deadline exceeded
+<time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelEvents]: context canceled
+<time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelEvents]: context deadline exceeded
 <time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelGraph]: rpc error: code = Canceled desc = context canceled
 <time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelGraph]: rpc error: code = DeadlineExceeded desc = context deadline exceeded
+<time> [ERR] RPCS: [/lnrpc.Lightning/SubscribeChannelGraph]: context canceled
```
guggero, line 165: This makes me think that perhaps we should add a new regular expression in `log_substitutions.txt` that replaces the `/lnrpc.Lightning/xxx` with just `<URL>` then we could collapse all of these context errors into one line in the whitelist.

rpcserver.go
```diff
@@ -1923,6 +1923,8 @@ out:
 
 				break out
 			}
+		case <-updateStream.Context().Done():
```
guggero, line 1926: Would be great to also apply these changes to the subservers in `lnrpc/...`, for example `lnrpc/routerrpc/router_server.go`.

cmd/lncli/commands.go
```diff
@@ -128,12 +128,12 @@ var newAddressCommand = cli.Command{
...
-func newAddress(ctx *cli.Context) error {
-	ctxb := getContext()
-	client, cleanUp := getClient(ctx)
+func newAddress(cliCtx *cli.Context) error {
+	ctx := getContext()
+	client, cleanUp := getClient(cliCtx)
 	defer cleanUp()
 
-	stringAddrType := ctx.Args().First()
+	stringAddrType := cliCtx.Args().First()
```
guggero, line 136: I'd prefer to keep the CLI context named as `ctx` and instead rename all `ctxb` to `ctxc` to make it more clear that it's backed by a cancel.

ellemouton:
@rodentrabies, are you planning on working on this further? If not, im keen to rebase your commits and continue with this.

ellemouton:
ok im gonna yolo-open the rebase PR but happy to close it if @rodentrabies re-appears :)

rodentrabies:
@ellemouton sorry I don't have enough time to wrap this up atm, so thanks a lot for picking it up!

Issue #2096, opened 2018-10-25
SignMessage & VerifyMessage missing from REST API


wbobeirne:
### Background

The SignMessage and VerifyMessage commands available to gRPC are not available via the REST API. Not sure if there's a particular reason for this, but ideally the APIs should be at parity with each other.

### Your environment

* version of `lnd` - 0.5.0-beta
* which operating system (`uname -a` on *Nix) - macOS
* version of `btcd`, `bitcoind`, or other backend - btcd

### Steps to reproduce

N/A

### Expected behaviour

N/A

### Actual behaviour

N/A

wpaulino:
This is because they're missing a REST endpoint like the other RPCs do.
https://github.com/lightningnetwork/lnd/blob/5c6c9668916461536e15b5be11e06adee0cfa89d/lnrpc/rpc.proto#L263

PR #2148, opened 2018-11-04
Add missing REST endpoints for SignMessage and VerifyMessage

xsb:
Fixes https://github.com/lightningnetwork/lnd/issues/2096


lnrpc/rpc.proto
```diff
@@ -260,15 +260,25 @@ service Lightning {
...
     /** lncli: `verifymessage`
     VerifyMessage verifies a signature over a msg. The signature must be
     zbase32 encoded and signed by an active node in the resident node's
     channel database. In addition to returning the validity of the signature,
     VerifyMessage also returns the recovered pubkey from the signature.
     */
-    rpc VerifyMessage (VerifyMessageRequest) returns (VerifyMessageResponse);
+    rpc VerifyMessage (VerifyMessageRequest) returns (VerifyMessageResponse) {
+        option (google.api.http) = {
+            post: "/v1/verifymessage"
```
Roasbeef: Perhaps this should be a `GET` instead as it doesn't mutate any state, doesn't produce anything new, and is also a stateless operation?

xsb: Both SignMessage and VerifyMessage could in theory be implemented as GET. Problem is, as GET requests are not supposed to have a body this implies the message needs to go in the URL. Which produces potentially very long URLs and in certain scenarios the user could find limits on the size of the message that can be passed to these functions. What do you think?

xsb: Thinking more about it, URL size limits are defined in clients not servers, and afaik only old browsers are problematic in that sense. So I guess this is not a real issue? I'd prefer to use the correct http verb here but I am just not 100% sure it's not an issue to pass messages in URLs.

xsb: @Roasbeef I can make it a GET but I have a question. The field `msg` is of type `bytes`. Should I change it to `string` instead as it needs to be passed through a URL? Looks like that would break other stuff.

wpaulino: `bytes` fields are encoded as base64 strings, but they're not URL safe, so perhaps it's better to keep it as a `POST`? Either that or we can note that the base64 string should be padded so that it's URL safe (https://stackoverflow.com/questions/1374753/passing-base64-encoded-strings-in-url), but this isn't very user friendly IMO.

wpaulino:
Needs a rebase!

xsb:
@wpaulino rebased

PR #2149, opened 2018-11-05
 discovery/gossiper: send node anns when constructing full chan proof

wpaulino:
In this PR, we allow the gossiper to also broadcast the corresponding node announcements, if we know of them, of a channel when constructing its full proof. We do this to ensure peers (other than our
remote peer) receive all the relevant announcements for a channel.

This also addresses an existing flake in the integration test `testNodeAnnouncement`. This problem arose due to the node announcement being sent before the connection between Dave (node announcement sender) and Alice (node announcement receiver) was initiated and the full channel proof was constructed.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -1642,6 +1642,9 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 				source: nMsg.source,
 				msg:    msg,
 			})
+		} else {
+			log.Tracef("Skipping broadcasting node announcement "+
+				"for %x due to being unadvertised", msg.NodeID)
 		}
 
 		nMsg.err <- nil
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -2366,6 +2366,31 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 			})
 		}
 
+		// We'll also send along the node announcements for each channel
+		// participant if we know of them.
+		node1Ann, err := d.fetchNodeAnn(chanInfo.NodeKey1Bytes)
+		if err != nil {
+			log.Debugf("Unable to fetch node announcement for "+
+				"%x: %v", chanInfo.NodeKey1Bytes, err)
+		} else {
+			announcements = append(announcements, networkMsg{
+				peer:   nMsg.peer,
+				source: nMsg.source,
+				msg:    node1Ann,
+			})
+		}
+		node2Ann, err := d.fetchNodeAnn(chanInfo.NodeKey2Bytes)
+		if err != nil {
+			log.Debugf("Unable to fetch node announcement for "+
+				"%x: %v", chanInfo.NodeKey2Bytes, err)
+		} else {
+			announcements = append(announcements, networkMsg{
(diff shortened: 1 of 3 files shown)


discovery/gossiper.go
```diff
@@ -2357,13 +2357,61 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
 				msg:    e1Ann,
 			})
+
+			// Since the edge exists, then we can also send the node
+			// announcement of the edge's backing node.
+			node, err := d.cfg.Router.FetchLightningNode(
+				chanInfo.NodeKey1Bytes,
+			)
+			if err != nil {
+				err := fmt.Errorf("unable to retrieve node "+
```
halseth: maybe just log error and continue execution?

discovery/gossiper.go
```diff
@@ -2357,13 +2357,61 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
 		if e2Ann != nil {
 			announcements = append(announcements, networkMsg{
 				peer:   nMsg.peer,
 				source: nMsg.source,
 				msg:    e2Ann,
 			})
+
+			// Since the edge exists, then we can also send the node
+			// announcement of the edge's backing node.
+			node, err := d.cfg.Router.FetchLightningNode(
```
halseth: extract into a common local method `fetchNodeAnn`?

discovery/gossiper.go
```diff
@@ -2357,13 +2357,61 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 				source: nMsg.source,
 				msg:    e1Ann,
 			})
+
```
halseth: I don't think this needs to be done within the `if e1ann != nil`? Doesn't seem like a channel update us needed for the `node_anncounement` to be valid.

discovery/gossiper_test.go
```diff
@@ -2200,6 +2392,16 @@ func TestReceiveRemoteChannelUpdateFirst(t *testing.T) {
 	case <-time.After(2 * trickleDelay):
 	}
 
+	err = <-ctx.gossiper.ProcessLocalAnnouncement(batch.nodeAnn1, localKey)
+	if err != nil {
+		t.Fatalf("unable to process :%v", err)
+	}
+	select {
+	case <-ctx.broadcastedMessage:
+		t.Fatal("channel update announcement was broadcast")
```
halseth: nit: this error should say node ann

discovery/gossiper_test.go
```diff
@@ -2152,8 +2333,19 @@ func TestReceiveRemoteChannelUpdateFirst(t *testing.T) {
...
+	case <-time.After(2 * trickleDelay):
+	}
 
+	err = <-ctx.gossiper.ProcessRemoteAnnouncement(batch.nodeAnn2, remotePeer)
+	if err != nil {
+		t.Fatalf("unable to process node ann: %v", err)
+	}
```
halseth: nit: below err should say node ann

discovery/gossiper_test.go

discovery/gossiper_test.go

PR #2150, opened 2018-11-05
lnrpc: modify AbandonChannel REST endpoint

wpaulino:
In this commit, we modify the AbandonChannel REST endpoint to avoid
conflicting with the CloseChannel's. Otherwise, if a debug build of lnd
is being used, there's no way of closing channels through the REST API
as it's been overwritten by AbandonChannel.


halseth:
Needs a rebase!

wpaulino:
Rebased!

Issue #2066, opened 2018-10-18
channel update from onion error for private channel is discarded


SomberNight:
I am paying lnd1 -> lnd2 -> electrum, where the last channel is private.
The invoice contains an `r` field for this private channel, but the content of this `r` field is not up-to-date (i.e. `fee_base_msat`, `fee_proportional_millionths`, `cltv_expiry_delta` are incorrect).

lnd1 attempts the payment, gets back the htlc failure/onion error, but then does NOT retry with the new channel update from the onion error payload.

`0.5.0-beta commit=v0.5-beta-133-gd52e691d5f051153b4fd07db40309a54ea56ef9a-dirty`

```
$ lncli -n regtest --lnddir=~/.lnd2 --rpcserver localhost:10011 sendpayment --pay_req=lnbcrt5u1pdu324dpp5qjyfmf8f59euhhyn6zctctn737ehu5ppxkz0chfnr9e33srfc33qdqqcqzysrzjqtsxecaeca06399tv9953ae503pufh48j7jzfrmau4xmr2n4sef3cqq89vqqqqsqqqqqqqqqqqqqqqqqqyxhkn45fshhjlwkgqkxjm7huq8kzl6934m4xvjxs2nkwwxkzq2wypxsej4wf30ug8lp4cltg6aqzhq8qzzw0qjlxs5dn8dh3t26mgj3qqwdupkq
Description: 
Amount (in satoshis): 500
Destination: 022cdc2c2c0409cf30c2a524024ae7d6555794612c1722f0e1970bca3337de3c03
Confirm payment (yes/no): yes
{
[...]

PR #2151, opened 2018-11-05
Update routing hints upon insufficient fees failure.

Bluetegu:
This pull request address issue #2066 

Reconstruct the problem:
Setup composed of 3 nodes, A -> B -> C, where B -> C is private.
Create an invoice on C   (C:addinvoice with private flag)
Increase fee on B->C   (B:updatechanpolicy to a higher fee)
Attempt to pay invoice from A  (A:sendpayment --pay_req=xxx)
Receive payment error ("unable to route payment to destination: FeeInsufficient ... ")

After the fix the payment goes through.

A unit test on the 'basic graph' was added. The test adds a private channel between son goku and elst, with lower fees compared with the alternative routes from roasbeef to elst. A fee error is returned on first attempt to route through the private channel, with a higher fee, but still better compared with the alternative routes. The test checks that the after the update the payment is still routed through the private channel but with the updated higher fees.

A bug in the existing test TestSendPaymentErrorRepeatedFeeInsufficient was fixed too. I did a minimal fix. I believe this test can be further improved. The major issue was that the simulated error returned was rejected due to signature failure, and hence didn't simulate correctly the insufficient fees error as intended. In order to correctly sign the errors, an optional private key attribute was added to the basic graph and its parser.

### DIFF
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -1,6 +1,8 @@
 package routing
 
 import (
+	"bytes"
+	"fmt"
 	"sync"
 	"time"
 
@@ -188,6 +190,11 @@ func (m *missionControl) NewPaymentSession(routeHints [][]zpay32.HopHint,
 			// Finally, create the channel edge from the hop hint
 			// and add it to list of edges corresponding to the node
 			// at the start of the channel.
+			v := NewVertex(hopHint.NodeID)
+			var flags lnwire.ChanUpdateFlag
+			if bytes.Compare(v[:], endNode.PubKeyBytes[:]) == 1 {
+				flags |= lnwire.ChanUpdateDirection
+			}
 			edge := &channeldb.ChannelEdgePolicy{
 				Node:      endNode,
 				ChannelID: hopHint.ChannelID,
@@ -198,9 +205,9 @@ func (m *missionControl) NewPaymentSession(routeHints [][]zpay32.HopHint,
 					hopHint.FeeProportionalMillionths,
 				),
 				TimeLockDelta: hopHint.CLTVExpiryDelta,
+				Flags:         flags,
 			}
 
-			v := NewVertex(hopHint.NodeID)
 			edges[v] = append(edges[v], edge)
 		}
 	}
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -2,7 +2,6 @@ package routing
 
 import (
 	"bytes"
-	"fmt"
 	"sync"
 	"time"
 
@@ -191,7 +190,7 @@ func (m *missionControl) NewPaymentSession(routeHints [][]zpay32.HopHint,
 			// and add it to list of edges corresponding to the node
 			// at the start of the channel.
 			v := NewVertex(hopHint.NodeID)
-			var flags lnwire.ChanUpdateFlag
+			var flags lnwire.ChanUpdateChanFlags
 			if bytes.Compare(v[:], endNode.PubKeyBytes[:]) == 1 {
 				flags |= lnwire.ChanUpdateDirection
 			}
@@ -205,7 +204,7 @@ func (m *missionControl) NewPaymentSession(routeHints [][]zpay32.HopHint,
 					hopHint.FeeProportionalMillionths,
 				),
 				TimeLockDelta: hopHint.CLTVExpiryDelta,
-				Flags:         flags,
+				ChannelFlags:  flags,
 			}
 
 			edges[v] = append(edges[v], edge)
--- a/routing/pathfind_test.go
+++ b/routing/pathfind_test.go
@@ -82,11 +82,14 @@ type testGraph struct {
 
 // testNode represents a node within the test graph above. We skip certain
 // information such as the node's IP address as that information isn't needed
-// for our tests.
+// for our tests. Private keys are optional. If set, they should be consistent
+// with the public key. The private key is used to sign error messages
+// sent from the node.
 type testNode struct {
-	Source bool   `json:"source"`
-	PubKey string `json:"pubkey"`
-	Alias  string `json:"alias"`
+	Source  bool   `json:"source"`
+	PubKey  string `json:"pubkey"`
+	PrivKey string `json:"privkey"`
+	Alias   string `json:"alias"`
 }
 
 // testChan represents the JSON version of a payment channel. This struct
@@ -164,6 +167,7 @@ func parseTestGraph(path string) (*testGraphInstance, error) {
 	}
 
 	aliasMap := make(map[string]Vertex)
+	privKeyMap := make(map[string]*btcec.PrivateKey)
 	var source *channeldb.LightningNode
 
 	// First we insert all the nodes within the graph as vertexes.
@@ -194,6 +198,25 @@ func parseTestGraph(path string) (*testGraphInstance, error) {
 		// alias map for easy lookup.
 		aliasMap[node.Alias] = dbNode.PubKeyBytes
 
+		// private keys are needed for signing error messages. If set
+		// check the consistency with the public key.
+		privBytes, err := hex.DecodeString(node.PrivKey)
+		if err != nil {
+			return nil, err
+		}
+		if len(privBytes) > 0 {
+			key, derivedPub := btcec.PrivKeyFromBytes(btcec.S256(),
+				privBytes)
+
+			if !pub.IsEqual(derivedPub) {
+				return nil, fmt.Errorf("%s public key and "+
+					"private key are inconsistent",
+					node.Alias)
+			}
+
+			privKeyMap[node.Alias] = key
+		}
+
 		// If the node is tagged as the source, then we create a
 		// pointer to is so we can mark the source in the graph
 		// properly.
@@ -204,7 +227,8 @@ func parseTestGraph(path string) (*testGraphInstance, error) {
 			// node can be the source in the graph.
 			if source != nil {
 				return nil, errors.New("JSON is invalid " +
-					"multiple nodes are tagged as the source")
+					"multiple nodes are tagged as the " +
+					"source")
 			}
 
 			source = dbNode
@@ -291,9 +315,10 @@ func parseTestGraph(path string) (*testGraphInstance, error) {
 	}
 
 	return &testGraphInstance{
-		graph:    graph,
-		cleanUp:  cleanUp,
-		aliasMap: aliasMap,
+		graph:      graph,
+		cleanUp:    cleanUp,
+		aliasMap:   aliasMap,
+		privKeyMap: privKeyMap,
 	}, nil
 }
 
--- a/routing/pathfind_test.go
+++ b/routing/pathfind_test.go
@@ -208,7 +208,9 @@ func parseTestGraph(path string) (*testGraphInstance, error) {
 			key, derivedPub := btcec.PrivKeyFromBytes(btcec.S256(),
 				privBytes)
 
-			if !pub.IsEqual(derivedPub) {
+			if !bytes.Equal(dbNode.PubKeyBytes[:],
+				derivedPub.SerializeCompressed()) {
+
 				return nil, fmt.Errorf("%s public key and "+
 					"private key are inconsistent",
 					node.Alias)
--- a/routing/payment_session.go
+++ b/routing/payment_session.go
@@ -37,6 +37,51 @@ type paymentSession struct {
 	pathFinder pathFinder
 }
 
+// edgeLocatorOfEdgePolicy returns the edge locator corresponding to
+// the channel edge policy
+func edgeLocatorOfEdgePolicy(ep *channeldb.ChannelEdgePolicy) *edgeLocator {
+	return &edgeLocator{
+		channelID: ep.ChannelID,
+		direction: edgePolicyDirection(ep),
+	}
+}
+
+// edgePolicyDirection returns the direction of the edge policy in the format
+// used  by edge locators (i.e. 1 if lnwire.ChanUpdateDirection is set)
+func edgePolicyDirection(ep *channeldb.ChannelEdgePolicy) uint8 {
+	if ep.ChannelFlags&lnwire.ChanUpdateDirection == 0 {
+		return 0
+	}
+	return 1
+}
+
+// updateEdgePolicy updates the channel edge policy parameters
+func updateEdgePolicy(up *lnwire.ChannelUpdate,
+	ep *channeldb.ChannelEdgePolicy) {
+
+	ep.FeeBaseMSat = lnwire.MilliSatoshi(up.BaseFee)
+	ep.FeeProportionalMillionths = lnwire.MilliSatoshi(up.FeeRate)
+	ep.TimeLockDelta = up.TimeLockDelta
+	ep.MinHTLC = lnwire.MilliSatoshi(up.HtlcMinimumMsat)
+}
+
+// UpdateEdgePolicy updates edge policy of the routing hints kept in the
+// payment session.
+func (p *paymentSession) UpdateEdgePolicy(el *edgeLocator,
+	update *lnwire.ChannelUpdate) {
+
+	log.Debugf("Updating edge %v policy in Mission Control", el)
+
+	// Run over hints and update policy if edge id and direction match.
+	for _, edges := range p.additionalEdges {
+		for _, edge := range edges {
+			if *el == *edgeLocatorOfEdgePolicy(edge) {
+				updateEdgePolicy(update, edge)
+			}
+		}
+	}
+}
+
 // ReportVertexFailure adds a vertex to the graph prune view after a client
 // reports a routing failure localized to the vertex. The time the vertex was
 // added is noted, as it'll be pruned from the shared view after a period of
--- a/routing/payment_session.go
+++ b/routing/payment_session.go
@@ -39,10 +39,10 @@ type paymentSession struct {
 
 // edgeLocatorOfEdgePolicy returns the edge locator corresponding to
 // the channel edge policy
-func edgeLocatorOfEdgePolicy(ep *channeldb.ChannelEdgePolicy) *edgeLocator {
-	return &edgeLocator{
-		channelID: ep.ChannelID,
-		direction: edgePolicyDirection(ep),
+func edgeLocatorOfEdgePolicy(ep *channeldb.ChannelEdgePolicy) *EdgeLocator {
+	return &EdgeLocator{
+		ChannelID: ep.ChannelID,
+		Direction: edgePolicyDirection(ep),
 	}
 }
 
@@ -67,7 +67,7 @@ func updateEdgePolicy(up *lnwire.ChannelUpdate,
 
 // UpdateEdgePolicy updates edge policy of the routing hints kept in the
 // payment session.
-func (p *paymentSession) UpdateEdgePolicy(el *edgeLocator,
+func (p *paymentSession) UpdateEdgePolicy(el *EdgeLocator,
 	update *lnwire.ChannelUpdate) {
 
 	log.Debugf("Updating edge %v policy in Mission Control", el)
--- a/routing/router.go
+++ b/routing/router.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"fmt"
+	"math"
 	"runtime"
 	"sort"
 	"sync"
@@ -1208,15 +1209,19 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 		// Now that we know this isn't a stale update, we'll apply the
 		// new edge policy to the proper directional edge within the
 		// channel graph.
-		if err = r.cfg.Graph.UpdateEdgePolicy(msg); err != nil {
-			err := errors.Errorf("unable to add channel: %v", err)
-			log.Error(err)
-			return err
-		}
+		if exists {
+			err = r.cfg.Graph.UpdateEdgePolicy(msg)
+			if err != nil {
+				err := errors.Errorf("unable to update "+
+					"channel policy: %v", err)
+				log.Error(err)
+				return err
+			}
 
-		invalidateCache = true
-		log.Tracef("New channel update applied: %v",
-			newLogClosure(func() string { return spew.Sdump(msg) }))
+			invalidateCache = true
+			log.Tracef("New channel update applied: %v",
+				newLogClosure(func() string { return spew.Sdump(msg) }))
+		}
 
 	default:
 		return errors.Errorf("wrong routing update message type")
@@ -1837,8 +1842,13 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
 			paySession.ReportEdgeFailure(
 				failedEdge,
 			)
+			return
 		}
 
+		// Update routing hints.
+		paySession.UpdateEdgePolicy(failedEdge, update)
+
+		// Report edge policy failure.
 		paySession.ReportEdgePolicyFailure(
 			NewVertex(errSource), failedEdge,
 		)
@@ -2060,13 +2070,21 @@ func (r *ChannelRouter) applyChannelUpdate(msg *lnwire.ChannelUpdate,
 		return true
 	}
 
-	ch, _, _, err := r.GetChannelByID(msg.ShortChannelID)
-	if err != nil {
-		log.Errorf("Unable to retrieve channel by id: %v", err)
+	// Verify update signature validates.
+	if err := ValidateChannelUpdateAnn(pubKey, math.MaxInt64, msg); err != nil {
+		log.Errorf("Failed to validate channel update: %v", err)
 		return false
 	}
 
-	if err := ValidateChannelUpdateAnn(pubKey, ch.Capacity, msg); err != nil {
+	// If channel is in database, do further checks. Else, the
+	// update is for an additional private edge added by hints.
+	ch, _, _, err := r.GetChannelByID(msg.ShortChannelID)
+	if err != nil {
+		log.Errorf("Unable to retrieve channel by id: %v", err)
+		return true
+	}
+
+	if err := ValidateOptionalFields(ch.Capacity, msg); err != nil {
 		log.Errorf("Unable to validate channel update: %v", err)
 		return false
 	}
--- a/routing/router.go
+++ b/routing/router.go
@@ -1810,7 +1810,9 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
 	// update with id may not be available.
 	failedEdge, err := getFailedEdge(route, errVertex)
 	if err != nil {
-		return true
+		// If channel identification fails, ignore the error
+		// message and try again, until timeout.
+		return false
 	}
 
 	// processChannelUpdateAndRetry is a closure that
@@ -1829,7 +1831,8 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
 		pubKey *btcec.PublicKey) {
 
 		// Try to apply the channel update.
-		updateOk := r.applyChannelUpdate(update, pubKey)
+		updateOk := r.applyChannelUpdate(update,
+			pubKey, failedEdge)
 
 		// If the update could not be applied, prune the
 		// edge. There is no reason to continue trying
@@ -1896,7 +1899,8 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
 	// that sent us this error, as it doesn't now what the
 	// correct block height is.
 	case *lnwire.FailExpiryTooSoon:
-		r.applyChannelUpdate(&onionErr.Update, errSource)
+		r.applyChannelUpdate(&onionErr.Update,
+			errSource, failedEdge)
 		paySession.ReportVertexFailure(errVertex)
 		return false
 
@@ -1941,7 +1945,8 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
 	// forward one is currently disabled, so we'll apply
 	// the update and continue.
 	case *lnwire.FailChannelDisabled:
-		r.applyChannelUpdate(&onionErr.Update, errSource)
+		r.applyChannelUpdate(&onionErr.Update,
+			errSource, failedEdge)
 		paySession.ReportEdgeFailure(failedEdge)
 		return false
 
@@ -1949,7 +1954,8 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
 	// sufficient capacity, so we'll prune this edge for
 	// now, and continue onwards with our path finding.
 	case *lnwire.FailTemporaryChannelFailure:
-		r.applyChannelUpdate(onionErr.Update, errSource)
+		r.applyChannelUpdate(onionErr.Update,
+			errSource, failedEdge)
 		paySession.ReportEdgeFailure(failedEdge)
 		return false
 
@@ -2063,7 +2069,7 @@ func getFailedEdge(route *Route, errSource Vertex) (
 // applyChannelUpdate validates a channel update and if valid, applies it to the
 // database. It returns a bool indicating whether the updates was successful.
 func (r *ChannelRouter) applyChannelUpdate(msg *lnwire.ChannelUpdate,
-	pubKey *btcec.PublicKey) bool {
+	pubKey *btcec.PublicKey, e *EdgeLocator) bool {
 	// If we get passed a nil channel update (as it's optional with some
 	// onion errors), then we'll exit early with a success result.
 	if msg == nil {
@@ -2089,9 +2095,16 @@ func (r *ChannelRouter) applyChannelUpdate(msg *lnwire.ChannelUpdate,
 		return false
 	}
 
+	// Trust the direction of the edge locator, and not the one
+	// in the message.
+	if e.Direction == 1 {
+		msg.ChannelFlags |= lnwire.ChanUpdateDirection
+	} else {
+		msg.ChannelFlags ^= lnwire.ChanUpdateDirection
+	}
 	err = r.UpdateEdge(&channeldb.ChannelEdgePolicy{
 		SigBytes:                  msg.Signature.ToSignatureBytes(),
-		ChannelID:                 msg.ShortChannelID.ToUint64(),
+		ChannelID:                 e.ChannelID,
 		LastUpdate:                time.Unix(int64(msg.Timestamp), 0),
 		MessageFlags:              msg.MessageFlags,
 		ChannelFlags:              msg.ChannelFlags,
--- a/routing/router_test.go
+++ b/routing/router_test.go
@@ -420,10 +420,10 @@ func TestChannelUpdateValidation(t *testing.T) {
 	ctx, cleanUp, err := createTestCtxFromGraphInstance(startingBlockHeight,
 		testGraph)
 
-	defer cleanUp()
 	if err != nil {
 		t.Fatalf("unable to create router: %v", err)
 	}
+	defer cleanUp()
 
 	// Assert that the initially configured fee is retrieved correctly.
 	_, policy, _, err := ctx.router.GetChannelByID(
@@ -583,7 +583,7 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
 	// to sophon. We'll obtain this as we'll need to to generate the
 	// FeeInsufficient error that we'll send back.
 	chanID := uint64(3495345)
-	_, _, edgeUpateToFail, err := ctx.graph.FetchChannelEdgesByID(chanID)
+	_, _, edgeUpdateToFail, err := ctx.graph.FetchChannelEdgesByID(chanID)
 	if err != nil {
 		t.Fatalf("unable to fetch chan id: %v", err)
 	}
@@ -682,10 +682,10 @@ func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
 
 	const startingBlockHeight = 101
 	ctx, cleanUp, err := createTestCtxFromFile(startingBlockHeight, basicGraphFilePath)
-	defer cleanUp()
 	if err != nil {
 		t.Fatalf("unable to create router: %v", err)
 	}
+	defer cleanUp()
 
 	// Craft a LightningPayment struct that'll send a payment from roasbeef
 	// to elst, through a private channel between son goku and elst
@@ -699,14 +699,19 @@ func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
 	feeBaseMSat := uint32(15)
 	feeProportionalMillionths := uint32(10)
 	expiryDelta := uint16(32)
-	hopHint := HopHint{
-		NodeID:                    ctx.aliases["songoku"],
+	sgNode := ctx.aliases["songoku"]
+	sgNodeID, err := btcec.ParsePubKey(sgNode[:], btcec.S256())
+	if err != nil {
+		t.Fatal(err)
+	}
+	hopHint := zpay32.HopHint{
+		NodeID:                    sgNodeID,
 		ChannelID:                 privateChannelID,
 		FeeBaseMSat:               feeBaseMSat,
 		FeeProportionalMillionths: feeProportionalMillionths,
 		CLTVExpiryDelta:           expiryDelta,
 	}
-	routeHints := [][]HopHint{{hopHint}}
+	routeHints := [][]zpay32.HopHint{{hopHint}}
 	payment := LightningPayment{
 		Target:      ctx.aliases["elst"],
 		Amount:      amt,
@@ -749,8 +754,15 @@ func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
 
 		if !errorReturned {
 			errorReturned = true
+			sourceKey, err := btcec.ParsePubKey(
+				sourceNode[:], btcec.S256(),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+
 			return [32]byte{}, &htlcswitch.ForwardingError{
-				ErrorSource: sourceNode,
+				ErrorSource: sourceKey,
 
 				// Within our error, we'll add a channel update
 				// which is meant to reflect the new fee
@@ -797,12 +809,10 @@ func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
 	}
 
 	// The route should have son goku as the first hop.
-	if !bytes.Equal(route.Hops[0].PubKeyBytes[:],
-		ctx.aliases["songoku"].SerializeCompressed()) {
-
+	if route.Hops[0].PubKeyBytes != ctx.aliases["songoku"] {
 		t.Fatalf("route should go through son goku as first hop, "+
 			"instead passes through: %v",
-			getAliasFromPubKey(route.Hops[0].PubKeyBytes[:],
+			getAliasFromPubKey(route.Hops[0].PubKeyBytes,
 				ctx.aliases))
 	}
 
@@ -931,20 +941,20 @@ func TestSendPaymentPrivateEdgeDirection(t *testing.T) {
 		}, c2eChanID)}
 
 	testGraph, err := createTestGraphFromChannels(testChannels)
-	defer testGraph.cleanUp()
 	if err != nil {
 		t.Fatalf("unable to create graph: %v", err)
 	}
+	defer testGraph.cleanUp()
 
 	const startingBlockHeight = 101
 
 	ctx, cleanUp, err := createTestCtxFromGraphInstance(startingBlockHeight,
 		testGraph)
 
-	defer cleanUp()
 	if err != nil {
 		t.Fatalf("unable to create router: %v", err)
 	}
+	defer cleanUp()
 
 	// Create private part of the network.
 	// f is a private node connected to both d and e.
@@ -955,24 +965,34 @@ func TestSendPaymentPrivateEdgeDirection(t *testing.T) {
 		0, 0, 0, 0, 0, 0, 0, 80,
 	}
 	_, pubKey := btcec.PrivKeyFromBytes(btcec.S256(), keyBytes)
-	ctx.aliases["f"] = pubKey
+	ctx.aliases["f"] = NewVertex(pubKey)
 
 	// Prepare hints.
 	expiryDelta := uint16(32)
 	d2eChanID := uint64(100)
 	d2fChanID := uint64(101)
 	e2fChanID := uint64(102)
-	routeHints := [][]HopHint{
+	dNode := ctx.aliases["d"]
+	dNodeID, err := btcec.ParsePubKey(dNode[:], btcec.S256())
+	if err != nil {
+		t.Fatal(err)
+	}
+	eNode := ctx.aliases["e"]
+	eNodeID, err := btcec.ParsePubKey(eNode[:], btcec.S256())
+	if err != nil {
+		t.Fatal(err)
+	}
+	routeHints := [][]zpay32.HopHint{
 		{
 			{
-				NodeID:                    ctx.aliases["d"],
+				NodeID:                    dNodeID,
 				ChannelID:                 d2eChanID,
 				FeeBaseMSat:               20,
 				FeeProportionalMillionths: 0,
 				CLTVExpiryDelta:           expiryDelta,
 			},
 			{
-				NodeID:                    ctx.aliases["e"],
+				NodeID:                    eNodeID,
 				ChannelID:                 e2fChanID,
 				FeeBaseMSat:               30,
 				FeeProportionalMillionths: 0,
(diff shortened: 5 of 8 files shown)


routing/router.go
```diff
@@ -1838,6 +1840,8 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
 			// newly updated fees.
 			case *lnwire.FailFeeInsufficient:
```
joostjager: This is not the only case where we can get back a channel update. For example, `FailIncorrectCltvExpiry`. I think we should try to update `additionalEdges` also in those cases.

joostjager: Can the update maybe be done inside `applyChannelUpdate` ?

routing/router_test.go
```diff
@@ -521,7 +548,7 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
 	}
 
 	// Craft a LightningPayment struct that'll send a payment from roasbeef
-	// to luo ji for 100 satoshis.
+	// to sophon for 1000 satoshis.
```
joostjager: I think the fix to this existing test can be moved to a separate commit, stacking on top of the priv key map commit.

routing/pathfind_test.go
```diff
@@ -74,11 +75,14 @@ type testGraph struct {
...
+// for our tests. Private keys are optional. If set, they should be consistent
+// with the public key. The private key is used to sign error messages
+// sent from the node.
 type testNode struct {
-	Source bool   `json:"source"`
-	PubKey string `json:"pubkey"`
-	Alias  string `json:"alias"`
+	Source  bool   `json:"source"`
+	PubKey  string `json:"pubkey"`
+	PrivKey string `json:"privkey"`
```
joostjager: Addition of a private key to the json can be a separate commit.

routing/router_test.go
```diff
@@ -602,13 +639,157 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+	sourceNode := ctx.aliases["songoku"]
+
+	// We'll now modify the SendToSwitch method to return an error on
+	// the first attempt, and check that the correct route and fees
+	// where used the second attempt.
+	errorReturned := false
+	ctx.router.cfg.SendToSwitch = func(firstHop lnwire.ShortChannelID,
+		_ *lnwire.UpdateAddHTLC, _ *sphinx.Circuit) ([32]byte, error) {
+
+		if !errorReturned {
```
joostjager: Flipping the boolean looks quite implementation-dependent. Is it possible to do it in a more robust way? For example by inspecting the contents of the `UpdateAddHTLC` and matching with the updated policy?

routing/router_test.go
```diff
@@ -602,13 +639,157 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
 
+// TestSendPaymentErrorFeeInsufficientPrivateEdge tests that if we receive
+// a fee related error from a private channel that we're attempting to route
+// through, then we'll update the fees in the route hints and successfuly
+// route through the private channel in the second attempt.
+func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
+	t.Parallel()
+
+	const startingBlockHeight = 101
+	ctx, cleanUp, err := createTestCtxFromFile(startingBlockHeight, basicGraphFilePath)
```
joostjager: There are currently two ways of writing these kind of path finding tests. One is the way you did it, using the json file. The other option is used in for example `TestChannelUpdateValidation`. It allows programmatically building the test graph and generates some values like the keys by itself (as the actual value isn't important for the tests, it just needs a value). You could consider using that instead. I think it is clearer than keep adding new channels to the json file.

Bluetegu:
Thanks for the review!
I rebased and changed the commit message(s). The code in router.go looks much better all around.
I addressed most of the comments. Please see responses to individual comments inline.

routing/missioncontrol.go
```diff
@@ -291,6 +291,27 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
 	return bandwidthHints, nil
 }
 
+// UpdateEdgePolicy updates the edge policy of the routing hints kept in the
+// payment session. The number of hints updated is returned.
+func (p *paymentSession) UpdateEdgePolicy(e *channeldb.ChannelEdgePolicy) uint32 {
```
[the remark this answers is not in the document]
Bluetegu: Done

routing/router.go
Bluetegu: Done.

routing/router_test.go
Bluetegu: I prefer to leave it as is. Its a pain to extract those changes, and I think the fix is aligned with the signature checks that were added to the new test, so its not an unrelated. Anyhow, if you believe it's important let me know and I'll extract to a separate commit.

routing/pathfind_test.go
Bluetegu: Done

routing/router_test.go
Bluetegu: What I wanted there is a simple way to send an error only the first time; that is, if its wasn't Go, I would have a static variable named say 'once' and change it to True inside the function. I think inspecting the content of UpdateAddHTLC would obfuscates this. Maybe the name errorReturned is confusing? Is there a better way in Go to achieve this 'static' variable 'do once' logic?

routing/router_test.go
Bluetegu: I think that you will still have the problem of needing a private key even if you use TestChannelUpdateValidation approach, as the validation of the error message is part of the tested code. Anyhow, I like the approach of using the basic graph in many tests all based on a non trivial topology. It requires the writer of the test to rethink whether the logic indeed covers all scenarios, beyond the simplest topology needed for testing the particular feature or fix she was aiming to do.

Bluetegu:
I rebased again, but for some reason the third CI test (coverage) fails in auto pilot and I don't think this has anything to do with the code modification in this PR. I forced push twice already and got the same error there.

routing/router.go
```diff
@@ -1818,6 +1820,12 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
 				update *lnwire.ChannelUpdate,
 				pubKey *btcec.PublicKey) {
 
+				err := ValidateChannelUpdateAnn(pubKey, update)
+				if err != nil {
+					log.Errorf("Unable to validate channel update: %v", err)
```
joostjager: This line and others too exceed the max line width of 80 chars (project coding standard).

routing/router.go
```diff
@@ -1818,6 +1820,12 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
 				update *lnwire.ChannelUpdate,
 				pubKey *btcec.PublicKey) {
 
+				err := ValidateChannelUpdateAnn(pubKey, update)
```
joostjager: Why is this added here? Validation is also done inside `r.applyChannelUpdate`.

routing/router.go
```diff
@@ -1834,6 +1842,21 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
...
+				failedChanID := failedEdge.channelID
+				paySession.UpdateEdgePolicy(
+					&channeldb.ChannelEdgePolicy{
+						ChannelID:                 failedChanID,
+						MinHTLC:                   uMinHTLC,
+						TimeLockDelta:             uTimeLockDelta,
+						FeeBaseMSat:               uBaseFee,
+						FeeProportionalMillionths: uFeeRate,
+					})
+
```
joostjager: You've added it in here, but for other cases where `r.applyChannelUpdate` is called directly, the payment session isn't updated.

routing/router_test.go
```diff
@@ -110,6 +112,7 @@ func createTestCtxFromGraphInstance(startingHeight uint32, graphInstance *testGr
 		router:    router,
 		graph:     graphInstance.graph,
 		aliases:   graphInstance.aliasMap,
+		privKeys:  graphInstance.privKeyMap,
```
joostjager: This looks like it belongs more to the previous commit where priv keys are introduced.

routing/router_test.go
joostjager: If you remove the mission control policy update code and `TestSendPaymentErrorFeeInsufficientPrivateEdge` (and leaving `signErrChanUpdate` in), I think a clean commit remains that fixes the existing test. And then commit on top of that the new functionality and its tests.

In know spending time on clean commit history can feel like a pain. But a large part of it is getting used to it, using the right tools and understanding git rebase commands. I use vscode for this sometimes. With all changes in the working directory, it allows you to select chunks of files and stage only those. There are other editors that offer similar functionality.

The basic assumption here is that it indeed takes the author of the PR extra time (but with some experience this can be marginal) than just creating a single big commit, but more time is saved by the (at least two) people who review the code and all future developers looking up this commit to understand the greater picture of a particular change.

routing/router_test.go
joostjager: That's okay, just wanted to mention the option.

routing/router_test.go
joostjager: Oh no, I don't mean that there is anything wrong with this code on the language level. 

This SendToSwitch is supposed to simulate the real behaviour to cover the private channel policy update case. If you flip the boolean, you rely on the fact the two htlcs will be sent out in a particular order. That's what I meant by '(router) implementation specific'. I was just thinking, if you just check the total amount in `UpdateAddHtlc` and decide based on that to reject or not, it is a little more robust.

routing/router.go
Bluetegu: My idea is that the hints are updated only when a retry is aimed at. It now handles all those cases (not only insufficient fees). Otherwise the vertex or channel is removed completely, and there is no need for the update. I thought that is the whole purpose of defining processChannelUpdateAndRetry (which includes the paySession in its closure). 
Said in another way, it doesn't seem right to me to plug it into applyChannelUpdate, exactly because this function is called for changing database edges/vertexes (and don't have paySession in closure).

routing/router.go
Bluetegu: I agree this is ugly. My excuse it that I'm trying to minimize the changes I do in code to those mandatory for this PR. Therefore, if validation fails, processChannelUpdateAndRetry should immediately return, and should not update paySession at all (neither update hints or report edge policy failure), so the check must be there. However we can not rely on the error returned by applyChannelUpdate, as it may well error in other scenarios (i.e. not validation errors) which are perfectly ok, and do require retries. Initially I thought of doing the validation much earlier, i.e. before the switch on the error types, but this looks like a change that is too major for this PR; hence my feeling was that adding validation here is the right way to go, with only bad consequence being that its checked twice.

routing/router.go
joostjager: Yes, I think you're right. All other calls to `applyChannelUpdate` prune immediately, so those edges will never show up in subsequent payment attempts anymore.

routing/missioncontrol.go
```diff
@@ -299,6 +299,22 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
 }
 
+// UpdateEdgePolicy updates the edge policy of the routing hints kept in the
+// payment session. The number of hints updated is returned.
+func (p *paymentSession) UpdateEdgePolicy(e *channeldb.ChannelEdgePolicy) {
+
+	// run over hints and update policy if edge id matches
+	for _, edges := range p.additionalEdges {
+		for _, edge := range edges {
+			if edge.ChannelID == e.ChannelID {
```
joostjager: Are you updating the channel in both directions? I think you may also need to match the pubkey in additional edges to update only a single direction.

Bluetegu: The channel is updated only in one direction. That is, the node generates an error on the outgoing channel policy, and you update the hints accordingly. Am I missing something here?

routing/router.go
Bluetegu: The log line is of 77 width. I believe I kept the <80 width everywhere.

routing/missioncontrol.go
joostjager: I mean, suppose you have two additional edges relating to the same channel id. They will both be updated?

routing/router.go
joostjager: I have it at 96 chars. Do you have tab size 8?

routing/router.go
joostjager: I think it is ok to not update the payment session if the graph update fails. `ReportEdgeFailure` is important to call in case of a validation error. If we get an incorrect update we don't want to try that channel anymore.

routing/router_test.go
Bluetegu: Ok. Done

routing/router.go
joostjager: The standard is 80 chars max @ tab size 8.

routing/router_test.go
Bluetegu: Fixed

routing/router.go
Bluetegu: Ok. I removed the extra validation and returned if applyChannelUpdate fails after the edge is pruned.

I'm uncomfortable with the handling of error messages that fail in signature validation, as they do prune edges. Doesn't that allow an attacker to send error messages and cause the routing to fail?

routing/missioncontrol.go
Bluetegu: Yes. I changed it such that it takes direction into account too, such that policy will be updated only if channel ID and direction both match.
I think this also fixes a general hidden bug in routing through hint hops, not related to this PR, as additionalEdges policies didn't take into account direction. That is, the Flags were not initialized in NewPaymentSession, and routing assumes the direction is initialized in all edge policies.

routing/router_test.go
Bluetegu: I don't agree here.
This is a unit test that orchestrate a scenario in which the first attempt to pay returns an error while the second one does not.
It's explicitly designed to fail only once, and not say, to fail twice and then succeed. 
Once the second attempt succeeds, the resulting route is checked, including a check that the same route is taken, and that the fee was indeed updated according to the error simulated.
I do 'rely on the fact the two htlcs will be sent out in a particular order'. Anything else will be detected as error. 
If instead I base the check on values in UpdateAddHtlc, I obfuscate the scenario I wanted to simulate (error only once), I have to avoid loops if the expected values do not appear, have to count the number of attempts to send to switch and compare later to 2, etc.

routing/router_test.go
Bluetegu: Done

Bluetegu:
Ready for inspection :-)
By some miracle the cover test passed too.
Please see inline my response to the review issues.
Thanks!

routing/router_test.go
joostjager: It indeed depends on what exactly you want to assert. Ok, no change needed then.

routing/missioncontrol.go
```diff
@@ -299,6 +306,25 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+// payment session.
+func (p *paymentSession) UpdateEdgePolicy(e *channeldb.ChannelEdgePolicy) {
+
+	// run over hints and update policy if edge id and direction matches
+	for _, edges := range p.additionalEdges {
+		for _, edge := range edges {
+			if edge.ChannelID == e.ChannelID &&
+				edge.Flags&lnwire.ChanUpdateDirection ==
+					e.Flags&lnwire.ChanUpdateDirection {
+				edge.FeeBaseMSat = e.FeeBaseMSat
```
joostjager: New line after wrapped line

routing/router.go
```diff
@@ -1832,7 +1834,30 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
...
 						failedEdge,
 					)
+					return
+				}
+
+				// Update routing hints.
+				var direction lnwire.ChanUpdateFlag
+				if failedEdge.direction == 1 {
+					direction = lnwire.ChanUpdateDirection
+				}
```
joostjager: `failedEdge.direction` can be used directly.

routing/missioncontrol.go
```diff
@@ -299,6 +306,25 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
 	return bandwidthHints, nil
 }
 
+// UpdateEdgePolicy updates the edge policy of the routing hints kept in the
+// payment session.
+func (p *paymentSession) UpdateEdgePolicy(e *channeldb.ChannelEdgePolicy) {
+
```
joostjager: remove newline

routing/missioncontrol.go
```diff
@@ -299,6 +306,25 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	// run over hints and update policy if edge id and direction matches
```
joostjager: Capital at beginning of sentence. Period at end.

routing/missioncontrol.go
```diff
@@ -209,6 +210,12 @@ func (m *missionControl) NewPaymentSession(routeHints [][]HopHint,
 			// Finally, create the channel edge from the hop hint
 			// and add it to list of edges corresponding to the node
 			// at the start of the channel.
+			v := NewVertex(hopHint.NodeID)
+			var direction lnwire.ChanUpdateFlag
+			if bytes.Compare(v[:],
+				endNode.PubKeyBytes[:]) == 1 {
```
joostjager: Can fit on a single line

routing/router.go
```diff
@@ -1203,14 +1203,16 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
...
 		// new edge policy to the proper directional edge within the
 		// channel graph.
-		if err = r.cfg.Graph.UpdateEdgePolicy(msg); err != nil {
-			err := errors.Errorf("unable to add channel: %v", err)
-			log.Error(err)
-			return err
-		}
+		if exists {
+			if err = r.cfg.Graph.UpdateEdgePolicy(msg); err != nil {
+				err := errors.Errorf("unable to update channel policy: %v", err)
```
joostjager: Exceeds line len

routing/router.go
```diff
@@ -1203,14 +1203,16 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 		// Now that we know this isn't a stale update, we'll apply the
 		// new edge policy to the proper directional edge within the
 		// channel graph.
-		if err = r.cfg.Graph.UpdateEdgePolicy(msg); err != nil {
-			err := errors.Errorf("unable to add channel: %v", err)
-			log.Error(err)
-			return err
-		}
+		if exists {
```
joostjager: Is this a fix in itself? Ideally this could be yet another separate commit, but if you find it too small for that, it's fine with me.

routing/router.go
```diff
@@ -1832,7 +1834,30 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
...
+				var direction lnwire.ChanUpdateFlag
+				if failedEdge.direction == 1 {
+					direction = lnwire.ChanUpdateDirection
+				}
+
+				fChanID := failedEdge.channelID
+				uBaseFee := lnwire.MilliSatoshi(update.BaseFee)
+				uFeeRate := lnwire.MilliSatoshi(update.FeeRate)
+				uMinHTLC := lnwire.MilliSatoshi(
+					update.HtlcMinimumMsat)
```
joostjager: This vars are introduced because of line length I assume. But maybe that indicates that now is the time to extract this closure to a separate function.

routing/router_test.go
```diff
@@ -647,6 +647,150 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+
+		t.Fatalf("route should go through son goku as first hop, "+
+			"instead passes through: %v",
+			getAliasFromPubKey(route.Hops[0].PubKeyBytes[:],
+				ctx.aliases))
+	}
+
+	// The route should pass via the private channel
+	if !route.containsChannel(privateChannelID) {
+		t.Fatalf("route did not pass through private channel between pham nuwen " +
```
joostjager: Line length.

routing/router_test.go
joostjager: Thanks. For the non-author it really makes it much easier to understand what parts of the full set of changes relate to each other.

routing/missioncontrol.go
```diff
@@ -219,9 +226,9 @@ func (m *missionControl) NewPaymentSession(routeHints [][]HopHint,
 					hopHint.FeeProportionalMillionths,
 				),
 				TimeLockDelta: hopHint.CLTVExpiryDelta,
+				Flags:         direction,
```
joostjager: This was indeed not set, but I think it didn't have impact before, because that flag was never read right?

routing/router_test.go
```diff
@@ -162,6 +165,30 @@ func createTestCtxFromFile(startingHeight uint32, testGraph string) (*testCtx, f
 	return createTestCtxFromGraphInstance(startingHeight, graphInstance)
 }
 
+// Add valid signature to channel update simulated as error
+// received from the network.
+func signErrChanUpdate(key *btcec.PrivateKey,
```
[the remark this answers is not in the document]
Bluetegu: Done.

routing/missioncontrol.go
Bluetegu: Done

routing/router.go
Bluetegu: Not sure. failedEdge direction is unit8 while Flags are unit16.

routing/missioncontrol.go
Bluetegu: Done

routing/missioncontrol.go
Bluetegu: Fixed

routing/missioncontrol.go
Bluetegu: Fixed

routing/router.go
Bluetegu: Fixed

routing/router_test.go
Bluetegu: Fixed

routing/missioncontrol.go
Bluetegu: I think it did have impact, i.e. it was an open bug, but the situations where it matters were rare enough such that it wasn't noticed. I haven't fully thought this out. What happened was that when I tried to fix UpdateEdgePolicy to consider direction as well, following your comment, the (new) test failed, and adding direction here fixed it.

Bluetegu:
Changes made.
The CI cover test failed, but I believe its a bug in the CI test itself (there is an open PR that fixes the coverage test).
Please see inline my response to the review issues.
Thanks!

routing/missioncontrol.go
joostjager: Yes, it was an omission that surfaced when you fixed `UpdateEdgePolicy`. But I think that before that, the situations where it mattered where non-existent. Just interested to know whether that was true. In any case, it is fixed now.

routing/router.go
joostjager: Ok, you may need to do a type cast, but the if statement is not necessary.

routing/router.go
joostjager: There are plenty of internal functions that are called only once, it does help to make the code more understandable. Inline functions can be useful especially if a lot of local vars are bound.

How about just creating a function taking `failedEdge` and `update` and returning the `ChannelEdgePolicy`? Or better, updating the signature of `UpdateEdgePolicy` to take  `failedEdge` and `update`. Also consistent with other pay session funcs that take an `edgeLocator` too.

routing/router.go
Bluetegu: Done

Bluetegu:
Getting closer...
Open issues addressed. Please see inline.
Thanks!

routing/missioncontrol.go
```diff
@@ -299,6 +305,33 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	// Prepare updated values.
+	uMinHTLC := lnwire.MilliSatoshi(update.HtlcMinimumMsat)
+	uFeeRate := lnwire.MilliSatoshi(update.FeeRate)
+	uFeeBase := lnwire.MilliSatoshi(update.BaseFee)
+	uDelta := update.TimeLockDelta
+
+	// Run over hints and update policy if edge id and direction match.
+	for _, edges := range p.additionalEdges {
+		for _, edge := range edges {
+			if edge.ChannelID == e.channelID &&
```
joostjager: I think it could be slightly better to extract the if+update block into a separate function. Saves some indents and temp. vars.

routing/missioncontrol.go
```diff
@@ -299,6 +305,33 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+				uint8(edge.Flags&lnwire.ChanUpdateDirection) ==
+					e.direction {
```
joostjager: Newline after wrapped line.

routing/missioncontrol.go
```diff
@@ -299,6 +305,33 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
 	return bandwidthHints, nil
 }
 
+// UpdateEdgePolicy updates edge policy of the routing hints kept in the
+// payment session.
+func (p *paymentSession) UpdateEdgePolicy(e *edgeLocator,
+	update *lnwire.ChannelUpdate) {
+	log.Debugf("Updating edge %v policy in Mission Control", e)
```
joostjager: New line after wrapped line.

routing/router_test.go
```diff
@@ -647,6 +647,150 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+	if !bytes.Equal(route.Hops[0].PubKeyBytes[:],
+		ctx.aliases["songoku"].SerializeCompressed()) {
+
+		t.Fatalf("route should go through son goku as first hop, "+
+			"instead passes through: %v",
+			getAliasFromPubKey(route.Hops[0].PubKeyBytes[:],
+				ctx.aliases))
+	}
+
+	// The route should pass via the private channel
```
joostjager: Dot at end of sentence.

routing/router_test.go
```diff
@@ -647,6 +647,150 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+	if !route.containsChannel(privateChannelID) {
+		t.Fatalf("route did not pass through private channel " +
+			"between pham nuwen and elst")
+	}
+
+	// The route should have the updated fee
```
joostjager: Dot at end of sentence.

routing/router_test.go
```diff
@@ -647,6 +647,150 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+	expectedFee := updatedFeeBaseMSat +
+		(updatedFeeProportionalMillionths*uint32(amt))/1000000
+	if route.HopFee(0) != lnwire.MilliSatoshi(expectedFee) {
+		t.Fatalf("expected %v MSat fee to forward to the private "+
+			"channel, instead got %v MSat", expectedFee,
+			route.HopFee(0))
+	}
+
```
joostjager: Remove newline

joostjager:
Yes, almost there. Hope the quality level we try to maintain didn't discourage you too much!

Bluetegu:
Not at all. I have full appreciation for the need to maintain and improve quality. My experience till now was that the review process led to better end results, not only in term of code quality and conformance, but also in term of what the PR eventually achieves. For example this PR was written to address a bug of handling of insufficient fees error, and after review addresses all similar errors (all errors which lead to a second reroute attempt), as well as probably fixing another unrelated bug.

Changes in code made per review comments.
Thanks!

routing/router_test.go
Bluetegu: Removed

routing/router_test.go
Bluetegu: Y

routing/router_test.go
Bluetegu: Y

routing/missioncontrol.go
Bluetegu: Y

routing/missioncontrol.go
Bluetegu: Y

routing/missioncontrol.go
Bluetegu: Done

routing/missioncontrol.go
```diff
@@ -299,6 +305,36 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
 	return bandwidthHints, nil
 }
 
+// updateEdgePolicy updates the channel edge policy parameters if both channel
+// ID and direction match those of the edge locator.
+func updateEdgePolicy(el *edgeLocator, up *lnwire.ChannelUpdate,
+	ep *channeldb.ChannelEdgePolicy) {
+
+	if ep.ChannelID == el.channelID &&
+		uint8(ep.Flags&lnwire.ChanUpdateDirection) == el.direction {
```
halseth: This is a bit brittle as it assumes that the direction flag is the last bit (this is also true elsewhere in the codebase, but better not to keep assuming it) :)

Instead you can do `(ep.Flags&lnwire.ChanUpdateDirection != 0) == (el.direction != 0)` or similar.

routing/missioncontrol.go
```diff
@@ -219,9 +225,9 @@ func (m *missionControl) NewPaymentSession(routeHints [][]HopHint,
 					hopHint.FeeProportionalMillionths,
 				),
 				TimeLockDelta: hopHint.CLTVExpiryDelta,
+				Flags:         direction,
```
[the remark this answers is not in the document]
halseth: Ideally this would be fixed in its own commit to make it clear what the consequence of this change is.

routing/router.go
halseth: I _think_ maybe the proper fix here is something like this: https://github.com/lightningnetwork/lnd/pull/2279/commits/ad92e5c9638c95b53f8756d462424d6317de76bf (feel free to cherry-pick it into this PR if needed).

I believe then you wouldn't have to flip the `AssumeChannelValid` in the config during tests?

routing/router_test.go
```diff
@@ -647,6 +647,149 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+				ctx.aliases))
+	}
+
+	// The route should pass via the private channel.
+	if !route.containsChannel(privateChannelID) {
+		t.Fatalf("route did not pass through private channel " +
+			"between pham nuwen and elst")
+	}
+
+	// The route should have the updated fee.
```
halseth: could also check the value of `errorReturned` here.

routing/missioncontrol.go
```diff
@@ -299,6 +305,36 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+// payment session.
+func (p *paymentSession) UpdateEdgePolicy(e *edgeLocator,
+	update *lnwire.ChannelUpdate) {
+
+	log.Debugf("Updating edge %v policy in Mission Control", e)
+
+	// Run over hints and update policy if edge id and direction match.
+	for _, edges := range p.additionalEdges {
+		for _, edge := range edges {
+			updateEdgePolicy(e, update, edge)
```
halseth: This is a bit hard to follow IMO. A better approach could be 
```go
for _, edges := range p.additionalEdges {
	for _, edge := range edges {
                if edge.ChannelID != e.ChannelID {
                      continue
                }

                if ..!checkDirection... {
                      continue
                }
 
                 ... update policy ....
        }
}
```
Also I don't think you need to pass the `edgeLocator` into this method? Might have everything you need in the `ChannelUpdate`. If not just pass the `pubKey` as in `applyChannelUpdate`.

routing/router_test.go
```diff
@@ -647,6 +647,149 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
 	}
 }
 
+// TestSendPaymentErrorFeeInsufficientPrivateEdge tests that if we receive
+// a fee related error from a private channel that we're attempting to route
+// through, then we'll update the fees in the route hints and successfuly
+// route through the private channel in the second attempt.
+func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
```
halseth: Good test! 👍

routing/missioncontrol.go
Bluetegu: Done

routing/missioncontrol.go
Bluetegu: Its strange that go doesn't have casting to bool in a simple way. Anyhow, I extracted it into the most straightforward help function, as to me this casting to bool is not self explanatory.

routing/missioncontrol.go
```diff
@@ -209,6 +210,11 @@ func (m *missionControl) NewPaymentSession(routeHints [][]HopHint,
 			// Finally, create the channel edge from the hop hint
 			// and add it to list of edges corresponding to the node
 			// at the start of the channel.
+			v := NewVertex(hopHint.NodeID)
+			var direction lnwire.ChanUpdateFlag
```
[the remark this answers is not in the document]
Bluetegu: Done

routing/router_test.go
```diff
@@ -566,7 +595,7 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
 	ctx.router.cfg.SendToSwitch = func(firstHop lnwire.ShortChannelID,
 		_ *lnwire.UpdateAddHTLC, _ *sphinx.Circuit) ([32]byte, error) {
 
-		roasbeefSongoku := lnwire.NewShortChanIDFromInt(chanID)
+		roasbeefSongoku := lnwire.NewShortChanIDFromInt(12345)
```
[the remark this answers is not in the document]
Bluetegu: This is another (copy and paste) error in the test. Note that this is supposed to be roasbeefSongoku channel, not chanID that is channel edge from son goku to sophon.

routing/router_test.go
```diff
@@ -583,7 +612,12 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
 		return preImage, nil
 	}
 
-	// Send off the payment request to the router, route through satoshi
+	// Avoid dropping the channel information upon check if the
```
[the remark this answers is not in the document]
Bluetegu: I rephrased it. Hope this is clearer now.

routing/pathfind_test.go
```diff
@@ -190,6 +194,23 @@ func parseTestGraph(path string) (*testGraphInstance, error) {
...
 		aliasMap[node.Alias] = pub
 
+		// private keys are needed for signing error messages. If set
+		// check the consistency with the public key.
+		privBytes, err := hex.DecodeString(node.PrivKey)
+		if err != nil {
+			return nil, err
+		}
+		if len(privBytes) > 0 {
+			key, derivedPub := btcec.PrivKeyFromBytes(btcec.S256(), privBytes)
```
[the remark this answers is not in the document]
Bluetegu: Fixed

routing/router_test.go
Bluetegu: I didn't understand this one.

routing/router_test.go
Bluetegu: Thanks

routing/missioncontrol.go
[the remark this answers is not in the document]
Bluetegu: The new test fails if the direction is not initialized here. I did try to add a unit test that directly checks this problem, but it didn't work as I expected; and I got into an impasse. 

I wanted the test to include two hints, one using a private channel in one direction and a second hint using the other direction of the same private channel. I wanted to have error on one of the directions, and check that the payment is routered through the other hint, through the private channel in a direction that didn't error. 

The setup was the following:  a graph including three nodes, a, b and c, connected by public edges in the following way.

               a  ------ b
               |
               |
               c

The payment should be routed to a node d. Node d is connected to b and c in private channels, and also c and b are connected in a private channel.

I included two hints   
h1: c -> b -> d
h2: b -> c -> d

I expected the route to be either 
r1 a -> c -> b -> d
r2 a -> b -> c -> d

But it seems that I'm getting either:
a1 a -> b -> d
a2 a -> c -> d

That is, I can't force routing to go through b -> c and c -> b, although that is what I explicitly indicate in the hints.
I was a bit surprised by that. Is routing expected to behave this way?

Initially I thought that I'd add negative fee on the b -- c channel to force routing to go through them, but unfortunately it doesn't seem that LND support negative fees at this point, as the fees are uint. 

I didn't have time to go deeper and make sure that I understand what is happening there correctly. I'll give it another try later.

Do you see a better test scenario?

Bluetegu:
Thanks for the review!

Most comments addressed. I did try to create a new unit test to directly test the subtle bug found in the paySession initialization, but got into an impasse.

Please see inline.

routing/missioncontrol.go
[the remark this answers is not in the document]
Bluetegu: I documented my finding regarding the above attempted test here #2346. 

My conclusion is that up till now in all legit scenarios there was no need for adding direction to additionalEdges during initialization in NewPaymentSession. This was only really required when direction was considered in UpdateEdgePolicy (both in missioncontrol). 
I don't recommend removing the direction initialization and checking, it is the right thing to do, but it doesn't solve any real life scenario, i.e. all real life scenarios were ok if both were omitted. 

IMO this means that there was no 'subtle bug' found (or its so subtle its invisible) and
that initializing the direction in NewPaymentSession doesn't deserve its own commit and its own unit test. It is part of the other logic that needs to address the original bug this PR is supposed to fix. The unit test that was added test the directionality already (it fails without initialization of the direction in NewPaymentSession) covers it.

I can add a test that attempt to send a malformed onion error that reports the wrong direction of the reported edge and check that it doesn't update the additionalEdges, but it seems too contrived. 

Let me know what you think.
Thanks!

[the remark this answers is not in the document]
halseth: Thanks for the detailed explanation! 

So if I understand it correctly, the reason this hasn't been a problem is because we don't this flag when doing path finding? We infer the direction of the edge by looking at the to/from nodes I think.

One way to possibly test this is to look at the `ignoredEdges` map. This one uses `edgeLocators` which also takes into account directionality. So if you could add the first direction to this map, then include a routing hint for the second direction, you might be able to make it fail finding a path.

routing/router_test.go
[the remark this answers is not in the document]
halseth: I see. The reason this test was passing before was that a signature error was returned instead?

routing/router_test.go
halseth: You can check that `errorReturned == true`.

routing/missioncontrol.go
```diff
@@ -299,6 +305,46 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+	log.Debugf("Updating edge %v policy in Mission Control", el)
+
+	// Run over hints and update policy if edge id and direction match.
+	for _, edges := range p.additionalEdges {
+		for _, edge := range edges {
+			if edge.ChannelID != el.channelID {
+				continue
+			}
+			if edgePolicyDirection(edge) != el.direction {
```
halseth: We could also add a method for creating `edgeLocator` from a `ChannelUpdate`.

routing/missioncontrol.go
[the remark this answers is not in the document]
Bluetegu: > So if I understand it correctly, the reason this hasn't been a problem is because we don't this flag when doing path finding? We infer the direction of the edge by looking at the to/from nodes I think.

Not exactly. What I'm saying is that if you didn't consider direction when pruning it'll be the same in real life scenarios.

Going back to the example detailed in the issue, suppose that we asked the router for 4 routes. 

We would get:
r1 A -> B -> D -> F
r2 A ->C -> E -> F
r3 A ->B -> D -> E -> F
r4 A ->C ->E -> D ->F

So it seems that we have both D->E and E->D on our list of candidates, and therefore we should make sure in code that if we receive an error on the D->E edge, we must prune only that direction D->E, but not E->D, hence the need for a direction check.

We should get an error on either directions of the E-D edge only once we attempt to route through it. We don't go through E-D in r1 and r2, so those need to fail first. 

So assume for our example that r1 failed due to D->F, so you prune it. and assume that r2 failed due to A->C, so you prune it too.

Now r3 is legit, so we try it. We pass through the D->E and we get an error.

With direction checking we prune D->E, but leave E->D. 

Without direction checking (as was before), we prune both E->D and D->E.

But r4 is useless anyhow (A->C already pruned), so it doesn't matter!

I'm quite confident that this can be formally proved for any graph whose cost at each hop is non-negative.

So bottom line, I don't see any real situation that I can simulate in a test that will require to check direction too, in addition to edge id, prior to pruning. I can create a contrived scenario, but as I said, the new test already test the direction checking anyhow.

routing/missioncontrol.go
Bluetegu: Added

routing/router_test.go
Bluetegu: Ah. Added

routing/router_test.go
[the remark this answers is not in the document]
Bluetegu: Yes.

halseth:
Blocked by #2279

routing/missioncontrol.go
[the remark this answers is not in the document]
halseth: OK, I _think_ I see the problem you are facing testing this now. Here's a possible solution (using your original A-B-C-D graph):
* Include routing hints B-C-D and C-B-D
* Let the router try route A-B-C-D by making sure it is the cheapest path.
* Make the hop B-C fail, so it should be pruned.
* Now the router will pick either A-B-D or A-C-D, since it is impossible for A-C-B-D to be cheaper than both. Let's say it picks A-C-D. Make C-D fail.
* Now it picks A-B-D. Make A-B fail.
* The only possible route left is A-C-B-D. Before your bugfix this route wouldn't be found.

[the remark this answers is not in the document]
Bluetegu: > Let the router try route A-B-C-D by making sure it is the cheapest path.
That was my original plan.
But, I can't. The router will select A->C->D
That is why I opened the issue #2346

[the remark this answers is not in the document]
halseth: Well, what if you make the fees on channels A-C and B-D super high? If it is still choosing the routes A-C-D or A-B-D then, you found a bug.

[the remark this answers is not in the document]
Bluetegu: Yes. I somehow convinced myself that its not possible. Anyhow, I separated this to a separate commit and added a dedicated unit test for that. Thanks!

Bluetegu:
@halseth  Hi Johan, Is the block tag necessary? Let's try to push this to completion. Thanks!

btcdanny:
> @halseth Hi Johan, Is the block tag necessary? Let's try to push this to completion. Thanks!

This would be great if possible, as we at Fastbitcoins are encountering the issue this PR addresses.

Bluetegu:
Ok, great. Let me know if anything is needed on my side.

halseth:
@Bluetegu #2549 is now merged, we can move this forward :)

Bluetegu:
@halseth Thanks!

routing/router_test.go
```diff
@@ -587,7 +616,13 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
 		return preImage, nil
 	}
 
-	// Send off the payment request to the router, route through satoshi
+	// processUpdate ensures that the funding transaction is not spent,
+	// before further processing an edge update. Toggel AssumeChannelValid
+	// to avoid dropping the channel update, as the utxo set query is
+	// not mocked in this test.
+	ctx.router.cfg.AssumeChannelValid = true
```
halseth: Is this still needed?

routing/router_test.go
```diff
@@ -652,6 +653,373 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+// r3: A->C->E->F     total fees: 90
+// r4: A->B->D->E->F  total fees: 150
+//
+// with the following errors:
+// r1: E->D is pruned
+// r2: D->F is pruned
+// r3: C->E is pruned
+// leaving r4 as viable 4th route
+//
+// This test was added to test direction initialization in NewPaymentSession;
```
halseth: Excellent test description! 👍

routing/router_test.go
```diff
@@ -652,6 +653,373 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to create graph: %v", err)
+	}
+
+	const startingBlockHeight = 101
+
+	ctx, cleanUp, err := createTestCtxFromGraphInstance(startingBlockHeight,
+		testGraph)
+
+	defer cleanUp()
```
halseth: move `defer cleanUp()` after error check

routing/router_test.go
```diff
@@ -652,6 +653,373 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+		errorSeq = errorSeq[1:]
+
+		return [32]byte{}, &htlcError
+	}
+
+	// processUpdate ensures that the funding transaction is not spent,
+	// before further processing an edge update. Toggel AssumeChannelValid
+	// to avoid dropping the channel update, as the utxo set query is
+	// not mocked in this test.
+	ctx.router.cfg.AssumeChannelValid = true
```
halseth: not needed anymore I think

routing/router_test.go
```diff
@@ -652,6 +653,373 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+
+	// Send payment
```
halseth: nit: missing period.

routing/router_test.go
```diff
@@ -652,6 +653,373 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+	}
+
+	// The route should have the updated fee.
+	if route.TotalFees != lnwire.MilliSatoshi(150) {
+		t.Fatalf("expected %v MSat total fee, "+
+			"instead got %v MSat", 150, route.TotalFees)
+	}
+
+	// Now switch directions and test again.
+	ctx.router.missionControl.ResetHistory()
```
halseth: Instead of manually driving the test twice, it should be made table driven, such that the two scenarios can be run by the same logic.

That being said, this is an excellent test! 😀

routing/missioncontrol.go
```diff
@@ -184,7 +183,7 @@ func (m *missionControl) NewPaymentSession(routeHints [][]HopHint,
 			// and add it to list of edges corresponding to the node
 			// at the start of the channel.
 			v := NewVertex(hopHint.NodeID)
-			var flags lnwire.ChanUpdateFlag
+			var flags lnwire.ChanUpdateChanFlags
```
halseth: Needs a rebase I guess?

routing/payment_session.go
```diff
@@ -35,6 +35,51 @@ type paymentSession struct {
 	preBuiltRoutes []*Route
 }
 
+// edgeLocatorOfEdgePolicy returns the edge locator corresponding to
+// the channel edge policy
+func edgeLocatorOfEdgePolicy(ep *channeldb.ChannelEdgePolicy) *edgeLocator {
```
halseth: just return non-pointer value.

routing/payment_session.go
```diff
@@ -35,6 +35,51 @@ type paymentSession struct {
...
+
+	ep.FeeBaseMSat = lnwire.MilliSatoshi(up.BaseFee)
+	ep.FeeProportionalMillionths = lnwire.MilliSatoshi(up.FeeRate)
+	ep.TimeLockDelta = up.TimeLockDelta
+	ep.MinHTLC = lnwire.MilliSatoshi(up.HtlcMinimumMsat)
+}
+
+// UpdateEdgePolicy updates edge policy of the routing hints kept in the
+// payment session.
+func (p *paymentSession) UpdateEdgePolicy(el *edgeLocator,
```
halseth: also better to take non-pointer for the `edgeLocator` here.

routing/router.go
halseth: This can be removed now that #2549 is merged.

routing/router_test.go
```diff
@@ -653,6 +653,155 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+// TestSendPaymentErrorFeeInsufficientPrivateEdge tests that if we receive
+// a fee related error from a private channel that we're attempting to route
+// through, then we'll update the fees in the route hints and successfuly
+// route through the private channel in the second attempt.
+func TestSendPaymentErrorFeeInsufficientPrivateEdge(t *testing.T) {
+	t.Parallel()
+
+	const startingBlockHeight = 101
+	ctx, cleanUp, err := createTestCtxFromFile(startingBlockHeight, basicGraphFilePath)
+	defer cleanUp()
```
halseth: same comment wrt `cleanup`

routing/router_test.go
```diff
@@ -653,6 +653,155 @@ func TestSendPaymentErrorRepeatedFeeInsufficient(t *testing.T) {
...
+		}
+
+		return preImage, nil
+	}
+
+	// processUpdate ensures that the funding transaction is not spent,
+	// before further processing an edge update. Toggel AssumeChannelValid
+	// to avoid dropping the channel update, as the utxo set query is
+	// not mocked in this test.
+	ctx.router.cfg.AssumeChannelValid = true
```
halseth: can be removed

Bluetegu:
I did the rebase, but I cannot figure out why the new direction unitest TestSendPaymentPrivateEdgeDirection fails. It didn't fail in the version before the rebase.
I started a new position and unfortunately I no longer have time to work on LND. I tried to fix this issue the last two weekends without success.
I uploaded my latest attempt. The last commit include all the unfinished code.
I hope someone can continue and complete this work (as well as the other of my open PRs), as it contains fixes for two bugs.
I'll do my best to answer questions etc., and can do some work in weekend, but my bandwidth is very limited now.

halseth:
@Bluetegu No worries! Thanks for your excellent contributions 👍

halseth:
Somebody should pick this up and rebase the still relevant parts.

Roasbeef:
Replaced by https://github.com/lightningnetwork/lnd/pull/5332

PR #2152, opened 2018-11-05
added num_routes param to rest api

evgeniy-scherbina:
It seems that lnd's QueryRoutes rest call does not work because num_routes always zero by default.


wpaulino:
It can be set as an optional parameter by appending `?num_routes=N` to the end of the URL. The real fix here is https://github.com/lightningnetwork/lnd/pull/2128.

Roasbeef:
Closing in favor of #2128.

PR #2155, opened 2018-11-06
lnwallet: update to new SendOutputs signature

joostjager:
Depends on https://github.com/btcsuite/btcwallet/pull/568/files (merged)

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -408,7 +408,13 @@ func (r *rpcServer) sendCoinsOnChain(paymentMap map[string]int64,
 		return nil, err
 	}
 
-	return r.server.cc.wallet.SendOutputs(outputs, feeRate)
+	tx, err := r.server.cc.wallet.SendOutputs(outputs, feeRate)
+	if err != nil {
+		return nil, err
+	}
+
+	txHash := tx.TxHash()
+	return &txHash, err
 }
 
 // determineFeePerKw will determine the fee in sat/kw that should be paid given
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -106,7 +106,7 @@
   revision = "ab6388e0c60ae4834a1f57511e20c17b5f78be4b"
 
 [[projects]]
-  digest = "1:ac01796bf202b80d6cb1b92d42df0fbbf3aff964451376bfa7c39bc2bf94d108"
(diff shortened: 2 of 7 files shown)


rpcserver.go
```diff
@@ -408,7 +408,9 @@ func (r *rpcServer) sendCoinsOnChain(paymentMap map[string]int64,
 		return nil, err
 	}
 
-	return r.server.cc.wallet.SendOutputs(outputs, feeRate)
+	tx, err := r.server.cc.wallet.SendOutputs(outputs, feeRate)
+	txHash := tx.TxHash()
```
wpaulino: I believe this will panic if the error is non-nil due to a nil `wire.MsgTx` being returned, so we want to check the error before doing this.

joostjager: Fixed

rpcserver.go
```diff
@@ -408,7 +408,14 @@ func (r *rpcServer) sendCoinsOnChain(paymentMap map[string]int64,
 		return nil, err
 	}
 
-	return r.server.cc.wallet.SendOutputs(outputs, feeRate)
+	tx, err := r.server.cc.wallet.SendOutputs(outputs, feeRate)
+
+	if tx == nil {
```
halseth: Error should be checked instead?

joostjager: I think this is closest to what it was. If SendOutputs misbehaves, it could still be that tx and err are both nil.

rpcserver.go
```diff
@@ -408,7 +408,14 @@ func (r *rpcServer) sendCoinsOnChain(paymentMap map[string]int64,
 		return nil, err
 	}
 
-	return r.server.cc.wallet.SendOutputs(outputs, feeRate)
+	tx, err := r.server.cc.wallet.SendOutputs(outputs, feeRate)
+
```
wpaulino: Style nit: remove newline.

rpcserver.go
wpaulino: Since it's not possible for `SendOutputs` to return `nil, nil`, it think it's better to just check the error. Usually, changes that allow returning `nil, nil` from a function/method also add a note to its `godoc`.

joostjager: I don't think it matters, because the diff is only in return value combinations that never happen. But changed anyway to yours liking :smile:

PR #2116, opened 2018-10-29
Integration tests fixes and cleanups

halseth:
This PR is split off from #1621 (Split chain backend and miner).

It is a pure integration test change:
1. Extracting various `timeout` and `CSV` parameters into common constants
2. Fix up some usages of `context.Background` which would stall the tests in certain cases
3. Fixes a few flakes caused by transaction leftover from previous test cases. 
4. Make sure the tests don't fail because a chain backend is still syncing after a restart
5. Various small cleanups.

Builds on:
- [x] https://github.com/lightningnetwork/lnd/pull/2153
- [x] https://github.com/lightningnetwork/lnd/pull/2156
- [x] https://github.com/lightningnetwork/lnd/pull/2157
- [x] https://github.com/lightningnetwork/lnd/pull/2158
- [x] https://github.com/lightningnetwork/lnd/pull/2159
- [x] https://github.com/lightningnetwork/lnd/pull/2160
- [x] #2224 
- [x] #2225 
- [x] https://github.com/lightningnetwork/lnd/pull/2229
- [x] https://github.com/lightningnetwork/lnd/pull/2230

PR #2156, opened 2018-11-06
[Integration tests]: make sure miner have seen tx before mining

halseth:
(Split off from #2116)

This commit adds a call to waitForTxInMempool before mining blocks we
assumed previously would always confirm a tx. This ensures that the tx
in question has actually propagated to the miner's mempool before mining
the block.

Requires: 
- [x] https://github.com/lightningnetwork/lnd/pull/2160
- [x] #2153 

Follow-up: Replace calls to `Generate` expecting txs in mempool to avoid more races.


halseth:
Maybe we can just let `mineBlocks` take an extra `int` argument?

halseth:
Done!

PR #2162, opened 2018-11-06
[Makefile] Define `make imports`

halseth:
This PR defines `make imports` for running `goimports` on the source. It also enables it for the linter and formats the affected files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ GOINSTALL := GO111MODULE=on go install -v
 GOTEST := GO111MODULE=on go test 
 
 GOVERSION := $(shell go version | awk '{print $$3}')
-GOFILES_NOVENDOR = $(shell find . -type f -name '*.go' -not -path "./vendor/*")
+GOFILES_NOVENDOR = $(shell find . -type f -name '*.go' -not -path "./vendor/*" -not -name "*pb.go" -not -name "*pb.gw.go")
 
 RM := rm -f
 CP := cp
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -209,7 +211,11 @@ endif
 # UTILITIES
 # =========
 
-fmt:
+imports: 
+	@$(call print, "Fixing imports.")
+	goimports -w $(GOFILES_NOVENDOR)
+
+fmt: imports
```
wpaulino: `goimports` will also run `gofmt`, so should we just replace it?

halseth: Good observation, will look into it.

halseth: Looks like `goimports` doesn't support code simplification, like we add here: https://github.com/lightningnetwork/lnd/pull/2118

cfromknecht: Should we just run both in `make fmt`?

halseth: As in don't define `make imports`?

wpaulino: Yep, agree it's better to keep all formatting tools under one target (`make fmt`).

halseth: Makes sense, done!

halseth:
Also included the commit from #2118 to run formatting with simplification.

halseth:
This has some performance problems with the usage of go modules. Will defer for now.

Relevant: https://github.com/golang/go/issues/27287

guggero:
Looks like this is checked by `golangci-lint` now. Can be closed?

halseth:
> Looks like this is checked by `golangci-lint` now. Can be closed?

Looks like it is not totally checked by the linter. I rebased it and re-ran. One import change only though :)

wpaulino (approved):

guggero (approved):
Nice, LGTM :taco:

PR #1782, opened 2018-08-24
rpcserver: show inactive channels in GetInfo

xsb:
Adds the `num_inactive_channels` field in the response of the `GetInfo` gRPC call. Inactive channels is calculated substracting the number of active channels to the sum of all currently open channels.

Fixes #1779

PR #2164, opened 2018-11-07
[trivial] lnrpc: reorder num_inactive_channels

halseth:
Small fixup from https://github.com/lightningnetwork/lnd/pull/1782

Move it close to num_active_channels for an output from lncli getinfo
that is easier to parse.


lnrpc/rpc.proto
```diff
@@ -1063,6 +1063,9 @@ message GetInfoResponse {
     /// Number of active channels
     uint32 num_active_channels = 4 [json_name = "num_active_channels"];
 
+    /// Number of inactive channels
+    uint32 num_inactive_channels = 15 [json_name = "num_inactive_channels"];
```
cfromknecht: IMO it's more important to keep the proto definitions in increasing order, so it stays clear what the last allocate number is. This also helps avoid gaps (where did 7 go?) and other inconsistencies.

In the past, we've reordered fields by defining a display struct at the RPC level with the desired order and mapping the fields into that before printing. There are a number of fields in getinfo that I think would benefit from a reordering, and this provides a better long-term solution by decoupling the display from the proto definitions

halseth: Agreed! 👍

halseth:
Closing per https://github.com/lightningnetwork/lnd/pull/2164#discussion_r232560960

Roasbeef:
Alternatively, we can just re-order it on the command line, similar to how we handle doing custom responses in order to print hex rather than base58: https://github.com/lightningnetwork/lnd/blob/master/cmd/lncli/commands.go#L1985-L1993

PR #2173, opened 2018-11-09
peer: Fix handshake timeout error message

chokoboko:


peer.go
```diff
@@ -269,10 +272,10 @@ func (p *peer) Start() error {
 
 	select {
 	// In order to avoid blocking indefinitely, we'll give the other peer
-	// an upper timeout of 15 seconds to respond before we bail out early.
-	case <-time.After(time.Second * 15):
-		return fmt.Errorf("peer did not complete handshake within 5 " +
-			"seconds")
+	// an upper timeout to respond before we bail out early.
+	case <-time.After(handshakeTimeout):
+		return fmt.Errorf("peer did not complete handshake within %d "+
```
halseth: I think using %d will actually print in nanoseconds. Instead you should use %v which will automatically print 15s

cfromknecht: Good catch @halseth

halseth:
@chokoboko The two commits should be squashed into one! :)

PR #2188, opened 2018-11-15
Added simnet support for litecoind

durkmurder:
This PR adds support of simnet of litecoind. 
There is PR in [litecoind](https://github.com/litecoin-project/litecoin/pull/559) that adds support of simnet based on ltcd implementation.


Roasbeef:
Added in #1865.

PR #2191, opened 2018-11-16
release: update release script to allow user to specify systems

xsb:
Whenever there is a new tagged release I find it useful to use `release.sh` targeting my own architecture. This PR introduces the option to specify a subset of systems/architectures or to try experimental ones without editing it.

Example:

```sh
LNDBUILDSYS="linux-arm64 darwin-amd64" ./release.sh
```

Notice that this is totally optional. When `LNDBUILDSYS` is unset the script works as always did.


halseth:
Concept ACK

PR #2203, opened 2018-11-20
htlcswitch+config+server: adding RejectHTLC flag

### LINKED ISSUE #1298: Feature request: ability to disable forwarding for node

ccdle12:
In this PR, we add a flag to the command line config `--rejecthtlc` and a 
field in the HTLC Switch config `RejectHTLC bool`.

The user should be able to start lnd with the `--rejecthtlc` flag which will
mean that lnd will reject any onward HTLCs. The switch will still allow receiving
HTLCs meant for the node and can still send HTLCs.

fixes #1298 and #2109

### DIFF
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -8842,6 +8842,169 @@ out:
 	cleanupForceClose(t, net, net.Bob, chanPointBob)
 }
 
+// testRejectHTLC tests that a node can be created with the flag --rejecthtlc.
+// This means that the node will reject all forwarded HTLCs but can still
+// accept direct HTLCs as well as send HTLCs.
+func testRejectHTLC(net *lntest.NetworkHarness, t *harnessTest) {
+	//             RejectHTLC
+	// Alice ------> Carol ------> Bob
+	//
+	const chanAmt = btcutil.Amount(1000000)
+	ctxb := context.Background()
+	timeout := time.Duration(time.Second * 5)
+
+	// Create Carol with reject htlc flag.
+	carol, err := net.NewNode("Carol", []string{"--rejecthtlc"})
+	if err != nil {
+		t.Fatalf("unable to create new node: %v", err)
+	}
+	defer shutdownAndAssert(net, t, carol)
+
+	// Connect Alice to Carol.
+	if err := net.ConnectNodes(ctxb, net.Alice, carol); err != nil {
+		t.Fatalf("unable to connect alice to carol: %v", err)
+	}
+
+	// Connect Carol to Bob.
+	if err := net.ConnectNodes(ctxb, carol, net.Bob); err != nil {
+		t.Fatalf("unable to conenct carol to net.Bob: %v", err)
+	}
+
+	// Send coins to Carol.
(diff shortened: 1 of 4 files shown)


halseth:
Needs rebase

ccdle12:
@halseth thanks, 

There's a bit for me to do in with refactoring the itest both in fixing the failing test and some general formatting.

server.go
```diff
@@ -443,6 +443,7 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 			htlcswitch.DefaultLogInterval),
 		NotifyActiveChannel:   s.channelNotifier.NotifyActiveChannelEvent,
 		NotifyInactiveChannel: s.channelNotifier.NotifyInactiveChannelEvent,
+		RejectHTLC:            false,
```
cfromknecht: probably don't need this commit since it's the default

test_utils.go
```diff
@@ -365,6 +365,7 @@ func createTestPeer(notifier chainntnfs.ChainNotifier,
 			htlcswitch.DefaultFwdEventInterval),
 		LogEventTicker: ticker.New(
 			htlcswitch.DefaultLogInterval),
+		RejectHTLC: false,
```
cfromknecht: similar comment here re: not being needed

lntest/itest/lnd_test.go
```diff
@@ -8843,6 +8843,159 @@ out:
...
+	// Bob adds the invoice to his database.
+	resp, err = net.Bob.AddInvoice(ctxb, bobInvoice)
+	if err != nil {
+		t.Fatalf("unable to add invoice: %v", err)
+	}
+
+	// Alice attempts to pay Bobs invoice.
+	ctxt, _ = context.WithTimeout(ctxb, timeout)
+	err = completePaymentRequests(ctxt, net.Alice, []string{resp.PaymentRequest}, true)
+	if err == nil {
```
cfromknecht: we may also want to test which error we're getting back here

test_utils.go
ccdle12: fixed!

server.go
ccdle12: fixed!

ccdle12: @cfromknecht just realised think that was an old commit. 

Server needs to pull the bool flag from configuration 
```
RejectHTLC: cfg.RejectHTLC
```

This will then enable the `htlcswitch/switch.go - handlePacketForward` to reject onward htlcs according to the configuration flag when creating the `htlcswitch.Config{...}` in `server.go`.

lntest/itest/lnd_test.go
ccdle12: fixed!

lntest/itest/lnd_test.go
```diff
@@ -8848,6 +8848,159 @@ out:
...
+		t.Fatalf("should have been rejected since carol will not accept onward htlcs")
```
halseth: nit: I haven't heard "onward htlcs" being used before. Use "forwarded" instead?

ccdle12: thanks fixed!

PR #2204, opened 2018-11-20
rpcserver: make methods context aware

### LINKED ISSUE #2134: Make `RPCServer` gRPC methods context aware

chokoboko:
In this PR, we ensure that rpcServer methods will stop execution after gRPC context is canceled.

fixes #2134

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -898,6 +898,11 @@ out:
 
 				break out
 			}
+
+		// The client canceled stream context, so we exit.
+		case <-updateStream.Context().Done():
+			return updateStream.Context().Err()
+
 		case <-r.quit:
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


halseth:
Does this implement anything that is not handled by https://github.com/lightningnetwork/lnd/pull/2147 ?

chokoboko:
Hey thanks for your comment! It's almost the same. The only difference is that in this PR, client ctx is referenced as a filed in `paymentStream{}`, instead of passing it as an argument to `sendPayment()` and we also handle nested selects in the goroutines started by `sendPayment()`.

PR #2205, opened 2018-11-21
lnwallet/btcwallet: return best header timestamp while wallet is resc…

valentinewallace:
…anning.

One way applications built on top of lnd can estimate sync percentage is
through comparing the current time to the best known timestamp of the
lnd wallet's sync state. Therefore, we should always return this
information even if the the wallet is not synced.


Roasbeef:
Curious, this didn't seem to be an issue in the past, what changed?

wpaulino:
Now, upon initial sync, the app won't be able to estimate how long it'll take until it finishes syncing as it depends on the best header timestamp. Without this change, the best header timestamp won't be returned until it's done syncing, which prevents the estimate calculation.

PR #2212, opened 2018-11-23
Bos score enabled Autopilot

halseth:
This PR adds an API for providing Autopilot with a set of bos scores that will be queried when opening channels.

Builds on
- [x] #2081 
- [x] #2039 
- [x] #2006 
- [x] #2341
- [x] #2405 
- [x] #2350


Roasbeef:
Needs rebase!

halseth:
Rebased! PTAL @Roasbeef

lnrpc/autopilotrpc/autopilot_server.go
```diff
@@ -41,6 +41,13 @@ var (
 			Entity: "info",
 			Action: "read",
 		}},
+		"/autopilotrpc.Autopilot/SetScores": {{
+			Entity: "onchain",
```
Roasbeef: Only actually needs on chain perms?

lnrpc/autopilotrpc/autopilot_server.go
```diff
@@ -217,3 +224,12 @@ func (s *Server) QueryScores(ctx context.Context, in *QueryScoresRequest) (
...
 	return resp, nil
 }
+
+// SetScores sets the scores of the external score heuristic, if active.
+//
+// NOTE: Part of the AutopilotServer interface.
+func (s *Server) SetScores(ctx context.Context,
+	in *SetScoresRequest) (*SetScoresResponse, error) {
+
+	return &SetScoresResponse{}, nil
```
Roasbeef: Purposeful lack of implementation here?

autopilot/combinedattach.go
```diff
@@ -126,3 +127,23 @@ func (c *WeightedCombAttachment) NodeScores(g ChannelGraph, chans []Channel,
 
 	return scores, nil
 }
+
+// SetNodeScores is used to set the internal map from NodeIDs to scores if a
+// score settable sub-heuristic is active.
+//
+// NOTE: This is a part of the ScoreSettable interface.
+func (c *WeightedCombAttachment) SetNodeScores(
+	newScores map[NodeID]float64) error {
```
Roasbeef: Seems like we should instead make it a bit more precise to _only_ target the specified heuristic. Otherwise, in the future we may accidentally override a set of heuristics inadvertently.

lnrpc/autopilotrpc/autopilot_server.go
```diff
@@ -231,5 +231,24 @@ func (s *Server) QueryScores(ctx context.Context, in *QueryScoresRequest) (
 func (s *Server) SetScores(ctx context.Context,
 	in *SetScoresRequest) (*SetScoresResponse, error) {
 
+	scores := make(map[autopilot.NodeID]float64)
```
Roasbeef: 👍 disregard prior comment

autopilot/externalscoreattach_test.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := h.SetNodeScores(scores); err != nil {
+		t.Fatal(err)
+	}
+
+	// Query all scores, half should be set, half should be zero.
+	q := make(map[autopilot.NodeID]struct{})
+	for _, nID := range pubkeys {
+		q[nID] = struct{}{}
+	}
+	resp, err := h.NodeScores(nil, nil,
```
Roasbeef: Funky line break here, would expect something more like: 
```
resp, err := h.NodeScores(
    nil, nil, btcutil.Amount(btcutil.SatoshiPerBitcoin), q,
)
```

lnrpc/autopilotrpc/autopilot_server.go
halseth: Rationale was that you can impact what channel gets opened by hitting this API, not sure if that is considered off-chain... Similar rationale for the `ModifyStatus` RPC above.

lnrpc/autopilotrpc/autopilot_server.go
halseth: Yes, just to make the commit compile.

autopilot/combinedattach.go
halseth: True. I prefer deferring this to when we actually have more heuristics to target though, as I'm not sure what is the proper way of targetting this as long as we have only one.

autopilot/externalscoreattach_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2212/commits/0e8c185f32bd2ba9aa26b9db936b3b268bedbc1b

halseth:
PTAL @Roasbeef

lnrpc/autopilotrpc/autopilot.proto
```diff
@@ -54,3 +60,9 @@ message QueryScoresResponse {
 
     repeated HeuristicResult results = 1 [json_name = "results"];
 }
+
+message SetScoresRequest{
+    map<string, int64> scores= 1 [json_name = "scores"];
```
wpaulino: Nit: add space between `scores=`.

autopilot/combinedattach.go
wpaulino: Could use an enum for each heuristic, but agree that can be deferred until needed.

lnrpc/autopilotrpc/autopilot_server.go
```diff
@@ -207,3 +214,31 @@ func (s *Server) QueryScores(ctx context.Context, in *QueryScoresRequest) (
...
+		if err != nil {
+			return nil, err
+		}
+		pubKey, err := btcec.ParsePubKey(pubHex, btcec.S256())
+		if err != nil {
+			return nil, err
+		}
+		nID := autopilot.NewNodeID(pubKey)
+
+		scores[nID] = float64(score)
```
cfromknecht: score is an int, so this only supports scores of 0 and 1?

autopilot/externalscoreattach.go
```diff
@@ -0,0 +1,124 @@
...
+// A compile time assertion to ensure ExternalScoreAttachment meets the
+// AttachmentHeuristic and ScoreSettable interfaces.
+var _ AttachmentHeuristic = (*ExternalScoreAttachment)(nil)
+var _ ScoreSettable = (*ExternalScoreAttachment)(nil)
+
+// Name returns the name of this heuristic.
+//
+// NOTE: This is a part of the AttachmentHeuristic interface.
+func (s *ExternalScoreAttachment) Name() string {
+	return "externalscore"
```
cfromknecht: Not a blocker, but would be nice if you could name external scores for debugging purposes, e.g. `”externalscore-<name>”`

lnrpc/autopilotrpc/autopilot.proto
cfromknecht: Needs comment, specifically to explain what the keys and values should be

Why use `int64` and not `double`?

autopilot/externalscoreattach.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/btcsuite/btcutil"
+)
+
+// ScoreSettable is an interface that indicates that the scores returned by the
+// heuristic can be set by an external caller. The ExternalScoreAttachment
+// imlements this interface, and so should any heuristic that is using the
+// ExternalScoreAttachment as a sub-heuristic, to allow access to setting the
+// internal scores.
+type ScoreSettable interface {
```
cfromknecht: non-blocking naming suggestion: `MutableHeuristic`?

TBH, my preferred alternative is actually to roll this into the `AttachmentHeuristic` interface, and non-mutable heuristics just return nil. This avoids the type casting and the targeting issues discussed elsewhere

autopilot/combinedattach.go
```diff
@@ -126,3 +127,23 @@ func (c *WeightedCombAttachment) NodeScores(g ChannelGraph, chans []Channel,
...
+
+	// Check if the external score attachement is active.
+	for _, h := range c.heuristics {
+		s, ok := h.AttachmentHeuristic.(ScoreSettable)
+		if !ok {
+			continue
+		}
+
+		return s.SetNodeScores(newScores)
```
cfromknecht: This only sets the first score found, though I’d expect this method forward the call to all relevant sub-heuristics. Note that the existing behavior is non-deterministic due to map range order if multiple mutable heuristics are present

autopilot/externalscoreattach.go
cfromknecht: If we keep the additional interface, it belongs in interface.go

autopilot/externalscoreattach.go
cfromknecht: This can also be used to target a specific external score if multiple are used in the weighted combination

lnrpc/autopilotrpc/autopilot.proto
halseth: Good catch, this should be `double`. It initially was using integers, but was changed to floats. The RPC I forgot...

autopilot/externalscoreattach.go
Roasbeef: Moving it to the `AttachmentHeuristic` interface would cause all implementation to have to implement this new method even if they don't actually utilize it. Splitting up the interface creates a specific interface which can be targeted precisely via type assertions. I don't think it's the case that type assertions are inherently "bad", they're simply a feature of the language to upgrade/downgrade interfaces.

cfromknecht: > I don't think it's the case that type assertions are inherently "bad", they're simply a feature of the language to upgrade/downgrade interfaces.

Certainly, type assertions have their place. IME however an assertion from interface->interface typically indicates that the abstractions are insufficient or surrounding logic is overloaded, even if using a type assertion readily enables compilation. 

Side note: not that performance is a huge concern in this particular instance, but interface to interface conversion is also the most expensive variant by a significant margin.

> Moving it to the `AttachmentHeuristic` interface would cause all implementation to have to implement this new method even if they don't actually utilize it.

How is that different that forcing `WeightedCombAttachment` to implement the `ScoreSettable` interface, even though it is it not the source of the scores it outputs? Its output is purely a dervied value. Unless calling `SetNodeScores` caused it to override the weighted combination (which is what is required as per the godoc comment on `SetNodeScores`), idt it makes sense to force it into the `ScoreSettable` interface. 

This also breaks the abstraction at the interface level, since you can't determine the behavior of calling `SetNodeScores` on a `ScoreSettable` without also knowing or asserting which `AttachmentHeuristic` you're calling it on: i.e. will it return exactly the scores I tell it, or will it return a dervied/weighted score for which these are only a component?

Perhaps a solution is to differentiate between a source (either static or mutable, conceptually a map) and a heuristic (derived from one or more sources). For instance:
```golang
type MutableHeuristicSource interface {
        AttachmentHeuristic

        SetNodeScores(map[NodeID]float64) error
}
...
var _ MutableHeuristicSource = (*ExternalScoreAttachment)(nil)
```

Since `ExternalScoreAttachment` then implements `AttachmentHeuristic` as a subinterface it can be used directly to instantiate `NewWeightedCombAttachment` as is the case right now. However, the combined interface lets us retain the same pointer with access to mutate the underlying source.

In addition to accepting the active `Heuristic` within the manager's config, we also accept an optional `MutableSources []MutableHeuristicSource` parameter. Upon initialization, the manager constructs an index of `map[source.Name()]  = source`. The method for updating scores then looks like:
```golang
func (m *Manager) SetSourceScores(name string, scores map[NodeID]float64) error {
	source, ok := m.mutableSourceIndex[name]
	if !ok {
		return ErrUnknownSource
	}

 	return source.SetNodeScores(scores)
}
```
where `name` can be additional parameter passed via rpc to target a specific source. Since the same object is referenced inside the `WeightedCombAttachment` we can safely update its scores without any additional coordination besides the mutex internal to `ExternalScoreAttachment`.

The major benefits here are that derived heuristics won't need to implement `SetNodeScores` unless they are 1) pure source or 2) we actually want to be able to override them. In addition, `WeightedCombAttachment` does not need to be modified at all and better isolates the heuristic derivation pipeline from the actual values flowing through it. Lastly, it provides a way of target specific mutable heuristics within the pipline without having to squeeze that logic through higher-level heuristics via `SetNodeScore`, which otherwise doesn't take any sort of identifier.

autopilot/combinedattach.go
halseth: This was a simplification I made since we are only expecting at most one such heuristic at this point. I'm warming up to making it possible to target sub-heuristics with a name parameter....

autopilot/externalscoreattach.go
halseth: Good suggestions :) I do see the benefit of being able to target a specific sub-heuristic to set the scores. 

Note that the API currently assumes (at least the few docs that are there) that the scores you set are also the scores the heuristic returns. Good point that in the case of the `Weighted` heuristic this won't be the exact case.

By specifying a name, it could be more clear that you are actually setting the scores on the `externalScore` sub-heuristic, and trying to target the `WeightedComb` directly would fail.

In regards to how to structure the interfaces etc, I don't have any strong opinions about it, I just used what was easiest to get this added. Since this is an experimental sub-server RPC, I think we can just start with something that works for our use case, and then modify it recklessly if we figure out there are better ways to organize it once we have more `ScoreSettable` sub-heuristics (if that even makes sense).

lnrpc/autopilotrpc/autopilot.proto
halseth: Fixed.

lnrpc/autopilotrpc/autopilot_server.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2212/commits/2b70f6d1245ed3175d29eba248797def8cdb1bf7

autopilot/externalscoreattach.go
halseth: Defined a new `SuperHeuristic` that can be used to get all sub-heuristics. From there we now target the name of the sub-heuristic as suggested.

autopilot/combinedattach.go
halseth: Fixed.

halseth:
Comments addressed. The agent now holds a `SuperHeuristic`, which is just a regular `AttachmentHeuristic` that can return its list of sub-heuristics. The RPC now lets the caller target a named heuristic, which makes it possible to add more heuristics later, that also accepts external scores.

PTAL @Roasbeef @cfromknecht @wpaulino

autopilot/manager.go
```diff
@@ -287,3 +287,35 @@ func (m *Manager) QueryHeuristics(nodes []NodeID) (HeuristicScores, error) {
 	log.Debugf("Querying heuristics for %d nodes", len(n))
 	return m.pilot.queryHeuristics(n)
 }
+
+// SetNodeScores is used to set the scores of the given heuristic, if it is
+// active, and ScoreSettable.
+func (m *Manager) SetNodeScores(name string, scores map[NodeID]float64) error {
+	// Check if the targeted heuristc is among the active sub-heuristics.
+	found := false
+	for _, h := range m.cfg.PilotCfg.Heuristic.SubHeuristics() {
```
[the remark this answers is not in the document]
halseth: Intentionally left this as an array to support the case with duplicate sub-heuristics with the same name.

autopilot/agent.go
```diff
@@ -24,14 +24,15 @@ type Config struct {
...
-	Heuristic AttachmentHeuristic
+	// Heuristic is a super-heuristic that will combine scores from its
+	// active sub-heuristics to govern to whom we open channels to, and
+	// also what those channels look like in terms of desired capacity. The
+	// Heuristic will take into account the current state of the graph, our
+	// set of open channels, and the amount of available funds when
+	// determining how channels are to be opened.  Additionally, a
+	// heuristic make also factor in extra-graph information in order to
+	// make more pertinent recommendations.
+	Heuristic SuperHeuristic
```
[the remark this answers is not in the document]
halseth: Yes, it was nicer that the interface supported both a regular heuristic and a super heuristic like in the previous iterations.

What about this proposal: 
1. let the `ScoreSettable`'s method `SetNodeScores` take the targetted heuristic name.
2. In `Manager.SetNodeScores`, we assert whether the active heuristic is `ScoreSettable`, if so set the scores, passing along the targeted name of the heuristic.
3. In the `SetNodeScores` method of the heuristic, we choose if we should set the scores (for regular heuristics, and if the name matches) or recursively to sub-heuristics. So it will be up to the regular "leaf" heuristics to decide whether they should apply the scores based on the name passed.
4. They can return a `bool` or specific error that indicates whether a matching leaf-heuristic was found.

cc @cfromknecht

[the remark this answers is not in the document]
cfromknecht: SGTM, this sounds like a good compromise :)

halseth:
PTAL @cfromknecht @Roasbeef @wpaulino

autopilot/externalscoreattach.go
```diff
@@ -0,0 +1,122 @@
...
+
+	s.Lock()
+	defer s.Unlock()
+
+	// Fill the map of candidates to return.
+	candidates := make(map[NodeID]*NodeScore)
+	for nID := range nodes {
+		var score float64
+		if nodeScore, ok := s.nodeScores[nID]; ok {
+			score = float64(nodeScore)
```
cfromknecht: nit: unnecessary conversion, `nodeScore` is already a `float64`

autopilot/manager.go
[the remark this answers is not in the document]
cfromknecht: not a blocker, but having trouble determining when this would be useful. if multiple subheurisitcs share the same name and are the same object, we only need to invoke `SetNodeScores` once. If they are distinct objects, it's ambiguous which one is being targeted (I would think this should produce an error when creating the manager) and may result in undefined behavior. In the case of `WeightedCombAttach`, if two distinct heuristics both share the name `externalattach` with weights 0.25 and 0.5, they could equally be represented as a single heuristic w/ weight 0.75.

Issue #1792, opened 2018-08-26
Docker stuck on simnet


yancyribbens:
I think there's a problem with the docker implementation.  Following the instructions, `export NETWORK="testnet"` for example, followed by `sudo docker-compose up "btcd"` shows `Command: btcd --simnet --debuglevel=info --rpcuser=devuser --rpcpass=devpass --datadir=/data --logdir=/data --rpccert=/rpc/rpc.cert --rpckey=/rpc/rpc.key --rpclisten=0.0.0.0 --txindex`.  As you can see, it's starting simnet instead of testnest.

I can get it to run testnet by explicitly changing https://github.com/lightningnetwork/lnd/blob/master/docker/btcd/start-btcd.sh#L45

Another way to pass the environment would be to use a .env file

yancyribbens:
@Roasbeef i'd be happy to pick this up also.

Roasbeef:
Excellent! So issue seems that it no longer properly reads out the environment variables?

yancyribbens:
yeah.  It doesn't scope up the env vars from the host OS.

Should be able to just use the `environment:` key in `docker-compose.yml`, although maybe being on `Version 2` is the issue.

```
    environment:
      RPCUSER: $RPCUSER
      RPCPASS: $RPCPASS
      DEBUG: $DEBUG
      NETWORK: $NETWORK
```

https://docs.docker.com/compose/environment-variables/#set-environment-variables-in-containers
https://docs.docker.com/compose/environment-variables/#the-env-file

yancyribbens:
this seems to work now.  Therefore closing issue.
`Command: btcd --testnet --debuglevel=info --rpcuser=devuser --rpcpass=devpass --datadir=/data --logdir=/data --rpccert=/rpc/rpc.cert --rpckey=/rpc/rpc.key --rpclisten=0.0.0.0 --txindex`

yancyribbens:
PR to fix issue: https://github.com/lightningnetwork/lnd/pull/2220

yancyribbens:
just an update to this issue.  it looks like the problem with not switching networks has to do with my environment and how I set the env vars.  For example, it works ok to use `sudo NETWORK=some_net ..` however, `export NETOWRK=some_net` will not work for me.  At any rate, the PR I submitted still is needed to fix the mainnet implementation from what I can tell.

PR #2220, opened 2018-11-26
fix bug which prevents btcd from starting different networks using en…

yancyribbens:
PR for issue: https://github.com/lightningnetwork/lnd/issues/1792#issuecomment-416408260

to run btcd using mainnet, there should be no network defined.  I added an additional check for if the network `mainnet` is defined also.  In addition, I removed the default value setters in the `.sh` scripts instead setting the default values using `.env`.  I think the docker setup could be simplified further by just removing `start-btcctl.sh` and `start-btcd.sh` and just set the entrypoint to `btcctl` and `btcd` respectively.

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -1,0 +1,6 @@
+# modify these values locally to change the behavior of btcd
+
+RPCUSER=devuser
+RPCPASS=devpass
+NETWORK=simnet
+DEBUG=info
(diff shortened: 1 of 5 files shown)


halseth:
utACK

yancyribbens:
this is now outdated. I'll create a new PR at some future time if needed.

Issue #2179, opened 2018-11-12
Private invoices are not marked as private on lookup


alexbosworth:
### Background

Looking up invoices that were created as "private" invoices does not reflect their private status as created.

### Your environment

* lnd - commit `d4b042dc1946ece8b60d538ade8e912f035612fe`
* MacOS Mojave
* btcd - commit `67e573d211ace594f1366b4ce9d39726c4b19bd0`

### Steps to reproduce

- GRPC `addInvoice` - `private` set to `true
- GRPC `lookupInvoice` - invoice id

### Expected behaviour

Response should indicate that the invoice is a "private" invoice

### Actual behaviour

Response indicates private is false

Roasbeef:
We use the exact same proto/struct for both the response and request. However, atm, we don't also include this value when printing out the response. This should definitely be made consistent, I've had a few other people bring this up in the past.

ErikEk:
@Roasbeef Do you want me write the private setting into the invoice db bucket?
If so, we might need to add some nasty backward compatibility handling if we don't modify or delete the old invoices.
Alternatively, we could just set private to false if there are no route_hints, but that feels wrong imo.

ErikEk:
Atm we don't store the private setting in the database. Is there a reason for not doing this?
My experimental code adds this setting to all new invoices before they are stored in the database. Because of this I'm forced to do some nasty tests to keep backward comparability when reading old invoices from the db, which would be nice to get rid of. I'm not sure how I should precede with this one.

halseth:
`--private` is only indicating that private __channels__ should be used as hop hints. There's no such thing as a private invoice I think.

With current `lnd` behavior if we want to make the lookup reflect the initial setting, we can just check if there are any routing hints in the invoice, meaning it was created with the `--private` option.

brandoncurtis:
More clarity is needed on how exactly `--private` is supposed to behave.

If I have a node and it has only private channels, my invoice MUST use routing hints and `lnd` DOES append routing hints regardless of whether I use `--private`.

If I have a node and it only has public channels, my invoice doesn't need routing hints and `lnd` DOES NOT append routing hints regardless of whether I use `--private`.

If my node has both public and private channels, what should I expect the behavior of `--private` to be?

Testing a node with one public and one private channel to the same node, and `lnd` DOES append routing hints regardless of whether I use `--private`.

Testing a node with one public and one private channel to different nodes, and `lnd` DOES append routing hints regardless of whether I use `--private`.

[...]

PR #2222, opened 2018-11-26
Invoices with --private marked as such on LookupInvoice

ErikEk:
When looking at invoices that were created with the private option on they do not reflect their private status correctly. This PR suggest a fix for that.
Fix for #2179.


ErikEk:
I though about that too, but since I didn't want to change the structure for saved invoices I guess this is good enough for now.

alexbosworth:
What if private were just renamed to something like `include_hop_hints` and then the response did not mirror the creation arguments?

ErikEk:
Sounds good to me. I'll update the pr if we can reach a consensus.

ErikEk:
I don't think we really need to store or print this value at all. The routing hits themself would be adequate.

PR #2224, opened 2018-11-26
lnd_test: increase test timeouts

halseth:
In preparation for the added propagation delay by separating the miner
and the chain backend, we increase several timeouts throughout the test,
and extract them into constants that can easily be altered.


lnd_test.go
```diff
@@ -5405,8 +5382,7 @@ func testMaxPendingChannels(net *lntest.NetworkHarness, t *harnessTest) {
 	maxPendingChannels := defaultMaxPendingChannels + 1
 	amount := maxBtcFundingAmount
 
-	timeout := time.Duration(time.Second * 10)
-	ctx, _ := context.WithTimeout(context.Background(), timeout)
+	ctx, _ := context.WithTimeout(context.Background(), defaultTimeout)
```
wpaulino: Looks like this isn't needed.

lnd_test.go
```diff
@@ -7840,7 +7807,7 @@ func testGraphTopologyNotifications(net *lntest.NetworkHarness, t *harnessTest)
 
 	// Now we'll test that updates are properly sent after channels are closed
 	// within the network.
-	ctxt, _ = context.WithTimeout(context.Background(), timeout)
+	ctxt, _ = context.WithTimeout(context.Background(), channelCloseTimeout)
```
wpaulino: Nit: use `ctxb`?

lnd_test.go
```diff
@@ -7977,7 +7944,7 @@ out:
 	}
 
 	// Close the channel between Bob and Carol.
-	ctxt, _ = context.WithTimeout(context.Background(), timeout)
+	ctxt, _ = context.WithTimeout(context.Background(), channelCloseTimeout)
```
wpaulino: Same here to use `ctxb`.

lnd_test.go
```diff
@@ -1098,7 +1094,7 @@ func testUpdateChannelPolicy(net *lntest.NetworkHarness, t *harnessTest) {
 	pushAmt := chanAmt / 2
 
 	// Create a channel Alice->Bob.
-	ctxt, _ := context.WithTimeout(ctxb, timeout)
+	ctxt, _ := context.WithTimeout(ctxb, channelOpenTimeout)
```
wpaulino: There are some `subscribeGraphNotifications` calls above that are missing a timeout.

halseth: We cannot give them a timeout, since they will need the context open for the entirety of the test.

lnd_test.go
halseth: Fixed

lnd_test.go
halseth: Done!

lnd_test.go
halseth: Done!

PR #2225, opened 2018-11-26
lnd_test: use context with timeout when instead of context.Background

halseth:
Split off from #2116

To ensure the test properly time out in case of a failed RPC call.


Roasbeef:
Needs rebase due to latest merges.

halseth:
Rebased!

PR #2237, opened 2018-11-28
Remove unneeded backend config

CirroStorm:
Since the chain source is abstracted there seems to be no reason that all wallet drivers shouldn't work with every backend. So there is no need for each wallet driver to specify which backends it supports.


CirroStorm:
This change became part of another branch.

PR #2245, opened 2018-11-30
build: switch to btcd version that uses go modules, switch from bolt to bbolt

Roasbeef:


.travis.yml
```diff
@@ -27,7 +27,7 @@ install:
   - export PATH=$PATH:$PWD/linux-amd64/
 
 script:
-  - env GO111MODULE=on
+  - export GO111MODULE=on
```
halseth: instead this can be set as a global env variable.

Issue #2183, opened 2018-11-14
channeldb+routing+htlcswitch: extract payment tracking from rpcserver, unify in router+switch


Roasbeef:
Tracking of outgoing payments within `lnd` currently has a number of bugs:
  * If a payment is initiated, but `lnd` crashes before it's complete, even once the pre-image comes back, we'll never mark it as fully settled in the database.
  * If a user uses the streaming RPC, and a payment is initiated, but the client is closed before it completes, then we'll again not even create the payment within the database. 
  * It's possible for two concurrent sending attempts from clients to flip-flop the control tower state during retries. 

Both of the bugs above stem from the fact that we _only_ write the payment in the same goroutines that creates the payment. To solve this, we need to move the payment tracking into the two sub-systems that will always be around when the payment is created (`ChannelRouter`) and ultimately settled (`HtlcSwitch`). 

While we're at it, we can also implement a much requested feature: for instances of `SendPayment` to be idempotent. Namely, once a payment has already been completed, a call to `SendPayment` with the same payment hash and amount should return the pre-image and final route from the database just as if it had _just_ completed. Additionally, it should be possible to _cancel_ an in-flight payment, which marks it as dead, so the channel router no longer attempts to satisfy the payment requirements. Finally, if a call to `SendPayment` is issued, and a payment with the same payment hash is in-flight, it should block until completed. 

# Steps To Completion 

- [x] Move _creation_ of the payment state into the `ChannelRouter` after each payment attempt. This way, we can properly track the last route used, as well as the fees, which won't be known to the switch at payment time. If a payment ends in an unsatisfied state (timeout, or no route), then it should be marked as such in the database. 

- [x] Move finalization of a payment's state into the `HtlcSwitch`, once a payment comes back across, we attempt to look up the sender in an in-memory map. In the idea case (before these changes) an entry is within the map, however, if the daemon has re-started, then we won't find anything in this map. Instead, another case needs to be added to dispatch the payment (via a new callback/interface perhaps) back to the primary caller. 

- [x] Within `SendPayment`, check the payment database to see if a payment for this payment hash was already completed. If so, just return the existing payment, possibly with fields indicating that this is the state from a _prior_ attempt. 

- [ ] Add pending payment attempt tracking to the `ChannelRouter` so we can cancel payments with a new RPC call. 

- [ ] With the payment tracking moved into the client, if a timeout hasn't been hit, then we can continue to retry a payment attempt in the background after a restart.

[...]

PR #2246, opened 2018-11-30
[WIP] channeldb+routing+switch: move payment tracking from rpcserver to router+switch

treygriffith:
This change moves responsibility for creating payments in the database
from the rpcserver and the original sender of the payment to the router
and the switch.

Previously, anytime the original sender was no longer available (e.g.
because lnd restarted or the client closed the stream) the payment would
not be saved in the database, and the preimage would not be available.

Now, the payment is originally created in the router before any payment
attempts are created, but is saved without a preimage. When a payment is
attempted, the payment in the database is updated with the route being attempted.
When the payment is finally complete and a preimage is available, the payment
is updated in the database with the preimage by the switch, which is guaranteed
to be alive when the payment completes.

Making this change required creating a new index for the payments by their
hash so that they can be retrieved and updated without knowing the order
in which they were inserted.

Specific changes include:
- Add an index for payments by payment hash
- Require a payment hash when adding a payment
- Add UpdatePaymentRoute function to the DB
- Add UpdatePaymentPreimage function to the DB
- Create payments in the Router, add Preimages in the Switch

This is in response to the first two items listed in https://github.com/lightningnetwork/lnd/issues/2183 and should fix https://github.com/lightningnetwork/lnd/issues/2032 (although I haven't tested it).

This PR is **work in progress** and does not contain any updates to test cases. I'm posting it now for early feedback about the direction this PR is taking prior to going any further.

### DIFF
--- a/channeldb/error.go
+++ b/channeldb/error.go
@@ -39,6 +39,14 @@ var (
 	// created.
 	ErrNoPaymentsCreated = fmt.Errorf("there are no existing payments")
 
+	// ErrDuplicatePayment is returned when a payment with the target payment
+	// hash already exists.
+	ErrDuplicatePayment = fmt.Errorf("payment with payment hash already exists")
+
+	// ErrPaymentNotFound is returned when a payment with the target payment
+	// hash can't be found.
+	ErrPaymentNotFound = fmt.Errorf("unable to locate payment with that hash")
+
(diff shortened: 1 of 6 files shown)


Roasbeef:
Superseded by #2761 and co. Thanks for getting the ball rolling on this change!

Issue #2217, opened 2018-11-25
Remote node returns error about exceeding max pending channels while no pending channels


cryptagoras:
### Background

I am trying to open a channel with the OpenNode lnd node, with which I used to have a channel in the past but after some time it got forced closed for some reason.
I then discovered that there was a ghost inbound channel from OpenNode for which the TX was never broadcasted. I force-closed all of them and waited till the locktime was over.

Now when I try to open a channel with the same node I get an error about max pending channels.
I contacted the João from OpenNode and there are no pending channels with my node (meanwhile he managed to open a channel to my node just fine).

It seems like there's something got out-of-sync around the pending channels.

All this happened before both of us upgrade to 0.5.1+ so the migrations didn't help. I still get the same remote error about max pending channels.

If the issue with ghost channels (i.e. creating channels by spending already-spent outputs) is fixed, the might not happen again, but it would be nice if there was a way to fix past out-of-sync state.

### Your environment

* `lnd` 4da1c867c3209dab4e4a824b73d89fc38b616b37
* OS: Ubuntu 16.04 LTS 4.4.0-119-generic
* `bitcoind` 0.16.3

### Steps to reproduce

Difficult to reproduce. Maybe, with maxpendingchannels=1, have NodeA creating somehow a ghost channel with NodeB. NodeB force-closing that ghost channel, then NodeB trying to open a channel with NodeA. (or NodeB force-closing that ghost channel as well)

```
lncli openchannel 03abf6f44c355dec0d5aa155bdbdd6e0c8fefe318eff402de65c6eb2e1be55dc3e 5217000
[lncli] rpc error: code = Code(101) desc = Number of pending channels exceed maximum
```

lnd logs
```
2018-11-25 15:57:40.837 [INF] HSWC: ChannelPoint(f794f34a4bfbb5ef81259194399a819a71e20523412c863c0ad2b814923e4b38:0): updating commit fee to 9023 sat/kw
2018-11-25 15:57:41.460 [INF] FNDG: Initiating fundingRequest(localAmt=0.05217 BTC, remoteAmt=0 mSAT, capacity=0.05217 BTC, chainhash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f, peer=03abf6f44c355dec0d5aa155bdbdd6e0c8fefe318eff402de65c6eb2e1be55dc3e, dustLimit=0.00000573 BTC, min_confs=1)
[...]

PR #2248, opened 2018-11-30
channeldb/db: prevent filtering channels with unconfirmed close txs

wpaulino:
In this commit, we address an issue with the FetchWaitingCloseChannels
method where it would not properly return channels that are unconfirmed
and also have an unconfirmed closing transaction because they were
filtered out. We fix this by modifying the filtering logic to only apply
when the channel has already confirmed.

Related to #2217.

### DIFF
--- a/channeldb/channel_test.go
+++ b/channeldb/channel_test.go
@@ -205,7 +205,7 @@ func createTestChannelState(cdb *DB) (*OpenChannel, error) {
 	return &OpenChannel{
 		ChanType:          SingleFunder,
 		ChainHash:         key,
-		FundingOutpoint:   *testOutpoint,
+		FundingOutpoint:   wire.OutPoint{Hash: key, Index: rand.Uint32()},
 		ShortChannelID:    chanID,
 		IsInitiator:       true,
 		IsPending:         true,
--- a/channeldb/channel_test.go
+++ b/channeldb/channel_test.go
@@ -852,6 +852,79 @@ func TestFetchClosedChannels(t *testing.T) {
 	}
 }
 
+// TestFetchWaitingCloseChannels ensures that the correct channels that are
+// waiting to be closed are returned.
+func TestFetchWaitingCloseChannels(t *testing.T) {
+	t.Parallel()
+
+	const numChannels = 2
+	const broadcastHeight = 99
+	addr := &net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: 18555}
+
+	// We'll start by creating two channels within our test database. One of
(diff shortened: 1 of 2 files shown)


channeldb/db.go
```diff
@@ -511,7 +511,7 @@ func fetchChannels(d *DB, pending, waitingClose bool) ([]*OpenChannel, error) {
 						"node_key=%x: %v", chainHash[:], k, err)
 				}
 				for _, channel := range nodeChans {
-					if channel.IsPending != pending {
+					if !waitingClose && channel.IsPending != pending {
```
halseth: Hmm.. this code is not straight forward to follow. Would it be better to instead add the channel if either
1) `channel.IsPending == pending`
2) `channelWaitingClose == waitingClose`

Another option is to only consider the `pending` argument if it is true, and use both `pending` and `waitingClose` as a way of filter channels.

cfromknecht: I'm not sure either is equivalent to the new behavior? Switching from 1 && 2 to 1 || 2 doesn't offer enough constraint to target a single quadrant of the state space. The new behavior ignores `pending` if we are looking for `waitingClose=true` channels, which is not the same as only considering `pending` if `pending=true`.

Docs + comments would help out here for sure

halseth: yeah, but with this ther will also be a quadrant impossible to target? (`1 && 2`)

cfromknecht: The simplest change (the one proposed) is to treat channels that are pending and waiting close as waiting close, so it’s collapsing the two quadrants where waiting close is true. 

We could leave the filtering logic as is, and just query for (pending=true, waitingClose=true) and (pending=false, waitingClose=true) and combine the results, which would be equivalent

halseth: Latter suggestion SGTM :)

wpaulino: Fixed.

Roasbeef:
Ping!

wpaulino:
@Roasbeef @halseth test added!

channeldb/channel_test.go
```diff
@@ -852,6 +852,79 @@ func TestFetchClosedChannels(t *testing.T) {
 	}
 }
 
+// TestFetchWaitingCloseChannels ensures that the correct channels that are
+// waiting to be closed are returned.
+func TestFetchWaitingCloseChannels(t *testing.T) {
```
halseth: Generally, the fixing commit should be added before the test case, to ensure the commit  history is always building/passing. Not a blocker for this PR tho.

PR #2252, opened 2018-12-01
queue: Add reliable 100% coverage

cfromknecht:
This PR fixes flakey coverage of the `queue` package, primarily due to inconsistencies on which lines are executed during shutdown. The main issue was due to the lingering goroutine potentially never reading from the quit channel before the test ends.

To remedy this, we know waitgroup the queue's goroutine, and wait for it to exit during `Stop`. The `Start` and `Stop` methods have also been made idempotent, to avoid spawning a goroutine twice and avoid closing the quit channel more than once.

Additional tests have been added to exercise these cases reliably, bringing the packages coverage up to 100%.


queue/queue_test.go
```diff
@@ -6,21 +6,53 @@ import (
...
 		item := <-queue.ChanOut()
 		if i != item.(int) {
-			t.Fatalf("Dequeued wrong value: expected %d, got %d", i, item.(int))
+			t.Fatalf("Dequeued wrong value: expected %d, got %d",
+				i, item.(int))
 		}
 	}
 }
+
+// TestConcurrentQueue tests that the queue properly adds 1000 items, drain all
```
Roasbeef: Missing `t.Parallel()` everywhere.

cfromknecht: Fixed

PR #2253, opened 2018-12-01
Fix the build error.

### LINKED ISSUE #2254: seems like master is on go modules now and no longer uses glide
### Background

Docker build now fails with the following

Step 5/19 : RUN git clone https://github.com/btcsuite/btcd . &&  glide install &&  go install . ./cmd/...
 ---> Running in 477bfdfeb76c
Cloning into '.'...
[ERROR] Failed to find glide.yaml file in directory tree: Cannot resolve parent of /
The command '/bin/sh -c git clone https://github.com/btcsuite/btcd . &&  glide install &&  go install . ./cmd/...' returned a non-zero code: 2

### Steps to reproduce

docker-compose build

### Expected behaviour

successful build

### Actual behaviour

failed to build

cryptcoin-junkey:
It looks the current master branch of `btcd` doesn't use Glide.

Fixes #2254.

### DIFF
--- a/docker/btcd/Dockerfile
+++ b/docker/btcd/Dockerfile
@@ -1,18 +1,15 @@
-FROM golang:1.10-alpine as builder
+FROM golang:1.11-alpine as builder
 
 MAINTAINER Olaoluwa Osuntokun <laolu@lightning.network>
 
 # Install build dependencies such as git and glide.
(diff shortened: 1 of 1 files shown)


yancyribbens:
tested ACK 👍

yancyribbens:
:shipit:

PR #2257, opened 2018-12-02
fix dockerfile dependency bug

yancyribbens:
fix `Dockerfile` build failure due to missing dependencies

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,10 @@ FROM golang:alpine as builder
 ENV GODEBUG netdns=cgo
 
 # Install dependencies and build the binaries.
-RUN apk add --no-cache \
+RUN apk add --no-cache --update alpine-sdk \
     git \
     make \
+    gcc \
 &&  git clone https://github.com/lightningnetwork/lnd /go/src/github.com/lightningnetwork/lnd \
 &&  cd /go/src/github.com/lightningnetwork/lnd \
 &&  make \


Dockerfile
```diff
@@ -5,9 +5,10 @@ FROM golang:alpine as builder
 ENV GODEBUG netdns=cgo
 
 # Install dependencies and build the binaries.
-RUN apk add --no-cache \
+RUN apk add --no-cache --update alpine-sdk \
     git \
     make \
+    gcc \
```
halseth: why is gcc needed?

MDrollette: It does seem to fail building without it.
```bash
$ docker-composer up ...
...
net
# net
exec: "gcc": executable file not found in $PATH
github.com/lightningnetwork/lnd/build
github.com/lightningnetwork/lnd/htlcswitch/hodl
make: *** [Makefile:112: install] Error 2
ERROR: Service 'lnd_btc' failed to build: The command '/bin/sh -c apk add --no-cache     git     make &&  cd /go/src/github.com/lightningnetwork/lnd &&  make &&  make install' returned a non-zero code: 2
```

MDrollette: But it also doesn't work after I added GCC. So maybe the Dockerfile needs more love than just this change.

```bash
$ docker build .
...
# net
/usr/local/go/src/net/cgo_linux.go:10:19: fatal error: netdb.h: No such file or directory
 #include <netdb.h>
                   ^
```

yancyribbens: yeah you also need to add `apk add --no-cache --update alpine-sdk`

yancyribbens: @halseth the makefile requires it.  if you try running `make VERBOSE=1` it will complain in the docker setup without it.  Also, the file you code reviewed https://github.com/lightningnetwork/lnd/pull/2253/files has that change as well.

halseth: Alright, SGTM :)

PR #2265, opened 2018-12-03
htlcswitch: reliable HTLC tracking

halseth:
This PR is a primary step towards fixing https://github.com/lightningnetwork/lnd/issues/2183

We handle a few problems regarding atomicity of payments within the switch. Since the writing of a payment hash to the `ControlTower` and switch `CircuitMap` is not atomic, we risked these getting out of sync, and potentially leaving preimages in limbo. This PR attempts to fix a few of these issues by giving the `htlcswitch` the responsibility for storing the preimage a payment succeeds, and immediately returning them if we notice a replay of a payment. We do this by using the `OutgoingPayment` bucket within the DB also for in-fligh payments (not only for completed payments as before).

We ensure all outgoing payments are given a unique ID, since this is what is used to determine if a payment is already committed to the switch circuit.

To ensure the payment DB and `CircuitMap` stay in sync, we acquire a mutex tied to the particular `paymentID` before using them.

This is also a step towards idempotent payments, since we'll now return a `pendingPayment` as normal even if the HTLC is already sent.

### TODO
* Fix compilation errors with existing unit tests.
* Unit tests for the following scenarios:
    - [x] A a payment in flight, and we replay it on restart. It will be added to the DB, and at the same time it gets settled. Since we haven't yet added it to the pending payment map, no response is sent. We then add it to the pending payment map,  but the event has already happened. We fix this by using a multimutex, this ensures that the payment  won't be attempted settled between we add it to the DB and adding it to the pendingpayment map, and at the same time allows distinct payment IDs to be cleared concurrently.
    - [x] Another race: we add the payment to the DB, add to pending payment map, a settle comes in and is removed from the circuit map, we add it to the circuit map again.


routing/router.go
```diff
@@ -41,6 +41,10 @@ var (
 	// construct a new sphinx packet, but provides an empty set of hops for
 	// each route.
 	ErrNoRouteHopsProvided = fmt.Errorf("empty route hops provided")
+
+	// ErrRouterShuttingDown is returned if the router is in the process of
+	// shutting down.
+	ErrRouterShuttingDown = fmt.Errorf("router shutting down")
```
cfromknecht: 👍

routing/router.go
```diff
@@ -1745,9 +1746,19 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
...
 			firstHop, htlcAdd, circuit,
 		)
+
+		var preImage [32]byte
+		var sendError error
+		select {
+		case preImage = <-preImageChan:
+		case sendError = <-errChan:
+		case <-r.quit:
+			return preImage, nil, fmt.Errorf("router shutting down")
```
cfromknecht: use `ErrRouterShuttingDown`?

htlcswitch/switch.go
```diff
@@ -361,34 +361,48 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID,
...
-		err:          errChan,
-		preimage:     preimageChan,
-		paymentHash:  htlc.PaymentHash,
-		amount:       htlc.Amount,
-		deobfuscator: deobfuscator,
-	}
+	// Before sending, double check that we don't already have 1) an
+	// in-flight payment to this payment hash, or 2) a complete payment for
+	// the same hash.
+	s.pendingMutex.Lock()
```
cfromknecht: do we need to hold the lock around `ClearForTakeoff`? this limits it to single writer, and we lose out on all benefits of using `Batch`

htlcswitch/control_tower.go
```diff
@@ -79,8 +88,12 @@ func NewPaymentControl(strict bool, db *channeldb.DB) ControlTower {
 
 // ClearForTakeoff checks that we don't already have an InFlight or Completed
 // payment identified by the same payment hash.
-func (p *paymentControl) ClearForTakeoff(htlc *lnwire.UpdateAddHTLC) error {
-	var takeoffErr error
```
cfromknecht: 👍

htlcswitch/switch.go
```diff
@@ -350,36 +350,89 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+		// Create payment and add to the map of payment in order later to be
+		// able to retrieve it and return response to the user.
+		payment := &pendingPayment{
+			err:          errChan,
+			preimage:     preimageChan,
+			paymentHash:  htlc.PaymentHash,
+			amount:       htlc.Amount,
+			deobfuscator: deobfuscator,
+		}
+		s.pendingPayments[paymentID] = payment
```
cfromknecht: what if this overwrites an existing entry? seems the prior caller would never receive anything on their channels. do we need to support multiple subscribers?

routing/router.go
halseth: lol, rebase fail

htlcswitch/switch.go
halseth: The reason I did this was to handle the case where `ClearForTakeoff` is called concurrently with a HTLC for the same payment has gets settled. Without the lock we risk getting `ErrPaymentInFlight` returned, but before we add the pending payment to the map the HTLC is settled.

I do see the lost benefit of batch... Time to bring out the `multimutex`? 😁

htlcswitch/switch.go
halseth: The thought here was only to handle the restart case, and otherwise make the caller responsible for not sending the HTLC more than once. But I definitely see the advantages of just handling duplicate entries, will see if can be easily handled!

cfromknecht: > and otherwise make the caller responsible for not sending the HTLC more than once

Isn't that hard to do if they have to call `SendHTLC` to see if they've already sent/re-register for ntfns?

Can you describe the scenario we're trying to solve?

htlcswitch/control_tower.go
```diff
@@ -90,67 +102,79 @@ func (p *paymentControl) ClearForTakeoff(htlc *lnwire.UpdateAddHTLC) error {
...
+			}
+			return ErrPaymentInFlight
 
 		case channeldb.StatusCompleted:
-			// We've already completed a payment to this payment hash,
-			// forbid the switch from sending another.
-			takeoffErr = ErrAlreadyPaid
+			// We've already completed a payment to this payment
+			// hash, forbid the switch from sending another.
+			preimage, err = channeldb.FetchPaymentPreimageTx(
```
cfromknecht: the side-effects of this transaction are not idempotent, which means we may see invalid reads from any of these subcommands. specifically, `pid` and `preimage` should be set to zero values each time the function is invoked

htlcswitch/switch.go
```diff
@@ -350,36 +350,91 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
 	}
 
-	s.pendingMutex.Lock()
-	s.pendingPayments[paymentID] = payment
-	s.pendingMutex.Unlock()
+	// Send the HTLC async, returning the response channels immediately to
+	// the caller.
+	s.wg.Add(1)
+	go s.sendAsync(
+		paymentID, firstHop, htlc, deobfuscator, preimageChan, errChan,
```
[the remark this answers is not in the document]
cfromknecht: Nvm we'd use the new one only if the payment was grounded which I think is okay

htlcswitch/control_tower.go
```diff
@@ -106,51 +101,41 @@ func (p *paymentControl) ClearForTakeoff(htlc *lnwire.UpdateAddHTLC) error {
...
 		case channeldb.StatusCompleted:
-			// We've already completed a payment to this payment hash,
-			// forbid the switch from sending another.
-			takeoffErr = ErrAlreadyPaid
+			// We've already completed a payment to this payment
+			// hash, forbid the switch from sending another.
+			return ErrAlreadyPaid
 
 		default:
-			takeoffErr = ErrUnknownPaymentStatus
```
cfromknecht: This commit should be reverted, and leave the batch error handling as is. If the function returns an error, then each will be run in isolation. This is the intent behind capturing the errors, but allowing the transaction to return `nil`

https://github.com/etcd-io/bbolt/blob/master/db.go#L787

halseth: Ah, I see TIL! Will revert and maybe add a comment explaining this 👍

htlcswitch/control_tower.go
halseth: Shouldn't matter, since `pid` is only used in case of `ErrPaymentInFlight` and `preimage` only in case of `ErrAlreadyPaid`. But probably doesn't hurt to do what you suggest.

htlcswitch/control_tower.go
halseth: I was kinda puzzled by why it was done this way tbh, probably should have asked before changing it :p

halseth:
Discussed a bit offline with @cfromknecht, have some ideas on how to simplify this quite a bit, and will push an update in a few days. Stay tuned.

htlcswitch/switch.go
halseth: Used a multimutex, was the only way to really ensure the `CircuitMap` and `ControlTower` would always stay in sync.

htlcswitch/switch.go
halseth: Went back to return results to all callers, so no overwriting would happen!

halseth:
New version pushed! Still some of the unit tests aren'r added, but the new design for idempotent payments should be in place.

Ended up using a multimutex to ensure synchronizing the `ControlTower` and `CircuitMap`. Another potential way of solving this would be to unify them into one database tx. With the mutex we make sure that a payment with this exact payment ID cannot be settled/failed between querying the control tower and committing the packet to the circuit map.

channeldb/payments.go
```diff
@@ -291,6 +296,41 @@ func FetchPaymentIDTx(tx *bbolt.Tx, paymentHash [32]byte) (uint64, error) {
 	return pid, nil
 }
 
+// UpdatePaymentPreimageTx stores the mapping from the given paymentHash to
```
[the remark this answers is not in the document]
halseth: Agreed. Will see if a little bit of scaffolding will let me reuse the existing structure.

halseth:
Gave this a rewrite. The failure scenarios are a bit tricky, so to ensure they are handled correctly I reduced the scope of this PR to only handle the `htlcswitch`'s responsibility of persistently handle replays of unique paymentIDs.

Unit tests are not currently compiling, but putting this up for some initial feedback.

Basically the tricky part is ensuring the atomicity between creating a payment with a unique payment ID and committing it to the circuit. Same for completing a payment and tearing down the circuit.

Since these are currently done in different DB transactions, we must handle the case where the first one succeeds but the latter fails. This is attempted by making each of the transitions `add in-flight payment -> commit to circuit -> tear down circuit -> complete payment` idempotent, such that we can resume at any point.

halseth:
Actually, don't review yet, might have a better approach.

htlcswitch/switch.go
```diff
@@ -974,6 +946,34 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
...
+	}
+
+	// Next, we'll remove the circuit since we are about to complete an
+	// fulfill/fail of this HTLC. Since we've already removed the
+	// settle/fail fwdpkg reference, the response from the peer cannot be
+	// replayed internally if this step fails. If this happens, this logic
+	// will be executed when a provided resolution message comes through.
+	// This can only happen if the circuit is still open, which is why this
+	// ordering is chosen.
+	if err := s.teardownCircuit(pkt); err != nil {
```
halseth: qq @cfromknecht: should we store the preimage before calling ackSettleFail/teardownCircuit? If we do, and crash before successfully storing the preimage, will we receive it again on restart?

cfromknecht:
@halseth the preimage should already have been added to the preimage cache whenever we receive a settle from the remote peer https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L1408. we should be able to query for it. note that that itself is not 100% reliable since it spawns and forgets the goroutine. we could make the change to not do that call asynchronously for the sake of correctness, and then come back and optimize it later

channeldb/payments.go
```diff
@@ -249,6 +257,40 @@ func FetchPaymentStatusTx(tx *bbolt.Tx, paymentHash [32]byte) (PaymentStatus, er
 	return paymentStatus, nil
 }
 
+// UpdatePaymentIDTx stores the association between the given payment hash and
```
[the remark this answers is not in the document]
cfromknecht: peep the method directly above ;)

halseth:
The new design just pushed is much simpler. It reuses the `PreimageCache` as a proxy for whether a resent HTLC is already settled.

The payment flow is as follows (steps 5-9 is handled in this PR).
1. The router creates gets a `LightningPayment` to pay to a given hash.
2. The router crafts a route, the HTLC and gives this bundle (`payAttempt`) a unique `paymentID`.
3. The router stores this `payAttemt` to the database.
4. The router send this payment attempt to the switch.
**<<<<<<<<<< (this PR)**
5. The switch checks the `PreimageCache` whether this payment ID is already completed.
6. If it is already completed, the preimage is returned directly to the router. If it is not completed, and we continue.
7. The switch attempts to commit the HTLC to the circuit map using the key `(linkID, paymentID)`. It doesn't care whether it was already committed, or if this was the first time it got committed.
8. When the payment result comes back, the preimage gets added to the `PreimageCache`, before the circuit is torn down.
9. The result is handed to the router if still online.
**this PR >>>>>>>>>>>>>>**
10. The router saves the outgoing successful payment in the db and deletes the pay attempt.
On restart the router will resend any `payAttempts` in the DB to the switch.
The follow-up that handles the router parts is here: https://github.com/lightningnetwork/lnd/pull/2475

Note that this assumes that the preimage is always in the `PreimageCache` if the payment was successful before the circuit gets torn down. For this assumption to hold, https://github.com/lightningnetwork/lnd/pull/2501 is required.

This is ready for review now, I'm still working on the tests, but the non-test code is done for now. I tried to make each commit carefully explain the rationale behind each change. 
PTAL @cfromknecht @Roasbeef

htlcswitch/switch.go
```diff
@@ -367,9 +370,20 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID,
...
+
+	// To ensure atomicity between checking the PreimageCache and
+	// forwarding the HTLC (committing the circuit), we aquire a multimutex
+	// for this paymentID.
+	s.paymentIDMtx.Lock(paymentID)
+
 	// If we already know the preimage, we can return it immediately.
 	p, ok := s.cfg.PreimageCache.LookupPreimage(htlc.PaymentHash[:])
 	if ok {
+		s.paymentIDMtx.Unlock(paymentID)
```
joostjager: Is it possible that anyone else locks in between this unlock and the lock above with the same `paymentID`? This `paymentID` is just generated and not known anywhere else at this point?

htlcswitch/switch.go
```diff
@@ -407,13 +416,15 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID,
...
 	if err := s.forward(packet); err != nil && err != ErrDuplicateAdd {
+		s.paymentIDMtx.Unlock(paymentID)
 		s.removePendingPayment(paymentID)
 		if err := s.control.Fail(htlc.PaymentHash); err != nil {
 			return zeroPreimage, err
 		}
 
 		return zeroPreimage, err
 	}
+	s.paymentIDMtx.Unlock(paymentID)
```
joostjager: Alternative is to `Unlock` directly after the call to `s.forward` and only then start checking the `err` value.

htlcswitch/switch.go
```diff
@@ -367,6 +367,14 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID,
 		return zeroPreimage, err
 	}
 
+	// If we already know the preimage, we can return it immediately.
+	p, ok := s.cfg.PreimageCache.LookupPreimage(htlc.PaymentHash[:])
+	if ok {
+		var preImg [32]byte
+		copy(preImg[:], p[:])
+		return preImg, nil
```
joostjager: How is this going to work when we not only need to get the preimage from the previous successful payment attempt, but also the route taken?

joostjager: Okay, you explained it nicely above. It is stored persistently by the router.

htlcswitch/switch.go
joostjager: Okay, I get this one too now. Generate of the `paymentID` is moved to router in the follow-up PR.

joostjager: Hmm.. but even then the paymentID passed into `SendHTLC` will always be unique?

htlcswitch/switch.go
halseth: Done!

htlcswitch/switch.go
halseth: Added the commit that moves paymentID generation to the router to this PR, good idea :)

halseth:
Addressed comments and added unit tests. Also moved the paymentID generation to the router.

PTAL @joostjager @cfromknecht @Roasbeef

htlcswitch/mock.go
```diff
@@ -32,16 +32,22 @@ import (
 type mockPreimageCache struct {
 	sync.Mutex
 	preimageMap map[[32]byte][]byte
+
+	// lookupDelay is used to delay the preimage lookup after the mutex is
+	// released.
+	lookupDelay time.Duration
```
joostjager: Can this delay be removed here and instead write an interceptor for the circuitmap, so that a channel can be exposed that is waited on before attempting to write the circuit?

Or a different approach to get rid of the delay?

htlcswitch/switch_test.go
```diff
@@ -2032,3 +2032,810 @@ func TestMultiHopPaymentForwardingEvents(t *testing.T) {
...
+			t.Fatalf("received incorrect preimage")
+		}
+	case <-time.After(time.Second):
+		t.Fatal("response wasn't received")
+	}
+
+	if s.numPendingPayments() != 0 {
+		t.Fatal("wrong amount of pending payments")
+	}
+}
```
joostjager: Is it possible to create a test context or at least some helper methods to reduce code duplication in this series of tests?

cfromknecht:
> If we already know the preimage, do we really want to pay to it? If we allow that, we risk any intermediate node also knowing it, being able to claim the payment before it reaches the receiver.

This is also true, but the new scenario joost is referring to is that we send the payment, then get used as an intermediary, restart, then report the payment succeeding even though we haven't heard back on the exact status of the payment.

> However, as discussed in #2265 (comment) this meant that an redundant preimage bucket structure was added, and the current version of the PR uses the preimage cache to avoid this.

Would be nice to find a way to avoid the redundant storage, while also retaining the strictness of the original proposal.

server.go
```diff
@@ -504,12 +504,19 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
 	}
 	s.currentNodeAnn = nodeAnn
 
+	// The router will get access to the payment ID sequencer, such that it
```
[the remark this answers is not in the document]
cfromknecht: Even if it's not used by the switch directly, it is exposed so that users of the `htlcswitch` package properly adhere to the semantics required by `SendHTLC`.

halseth:
It seems that what we want is the ability to distinguish between a previous attempt that completed and that we have the preimage in our cache for some other reason. 

This could be solved as in the original proposal with a persistent map `pid->preimage`, but as discussed earlier this necessitates a new bucket structure as far as I can tell. I think however this is okay, as it allows us to nicely separate the knowledge of the different parts of the payment flow between the router (`OutgoingPayment`, invoice, routes, map `pid->paymenthash` etc) and the switch (map `pid->circuit`, map `pid->preimage`).

We could use the reuse the `OutgoingPayment` bucket structure (https://github.com/lightningnetwork/lnd/blob/master/channeldb/payments.go) as much as possible, since these are already indexed by paymentID in the database. Currently this struct contains a payment preimage, but we could move this to its own bucket ("preimage bucket"), indexed by the same `pid`.

To send a payment, we would get a new unique `pid`, and create an outgoing payment in the database. Since no preimage was yet stored for this `pid`, this would indicate it is still in flight.

Sending the corresponding HTLC to the switch, it would commit the circuit, using the unique pid as key. Here we could add a check to the new "`preimage` bucket" in the same DB transaction to ensure we don't yet know the preimage. If we do, we return it.

When the result comes back, we would tear down the circuit and store the preimage (in case of success) in the same db transaction. The outgoing payment would therefore be considered settled, since the preimage now is found in the bucket.

In case the payment fails we could consider deleting the `OutgoingPayment`, but I think maybe it could be useful also keep failed payment attempts around?

This approach is essentially what I did in an earlier branch (https://github.com/lightningnetwork/lnd/compare/master...halseth:reliable-payments-reused-buckets), except for storing the preimages in their own bucket. That approach was awkward since the switch needed to be aware of the whole `OutgoingPayment` and all its fields.

lmk what you think.

htlcswitch/switch.go
```diff
@@ -218,6 +219,7 @@ type Switch struct {
 	// integer ID when it is created.
 	pendingPayments map[uint64]*pendingPayment
 	pendingMutex    sync.RWMutex
+	paymentIDMtx    *multimutex.Mutex
```
joostjager: I think the commits `lookup preimage in cache`, `ignore ErrDuplicateAdd` and `use multimutex` should be squashed in a single commit.

The first two introduce functionality that doesn't really work reliable until the last is in place too. Checking out the intermediate commits doesn't give useful behaviour.

They are all small too and (at least to me) it would make review easier. To see a set of changes that together does something meaningful.

htlcswitch/switch.go
```diff
@@ -366,6 +371,20 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID,
 		return zeroPreimage, err
```
joostjager: Consider moving the control tower up to the `router` level already in this pr. To finalize the required switch changes for reliable payments.

htlcswitch/switch.go
```diff
@@ -364,9 +367,15 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
 	}
 
+	// To ensure atomicity between checking the PreimageCache and
+	// forwarding the HTLC (committing the circuit), we aquire a multimutex
+	// for this paymentID.
+	s.paymentIDMtx.Lock(paymentID)
```
joostjager: Can the multi mutex be avoided by looking up the preimage after commiting the circuit? If it is present then, the circuit could be teared down again. Or left in half open state, where it will be cleaned up later.

joostjager:
Pasting my own notes used during review here for reference:
```
Switch.SendHTLC
        ControlTower.ClearForTakeoff
        multimutex lock
        LookupPreimage (return success if present)
        add pending payment in memory
        forward
                CommitCircuit (fail if duplicate payment id)
                async handoff to link
        multimutex unlock
        wait for payment complete

ChannelLink.handleUpstreamMsg
        AddPreimage
        Switch.handlePacketForward
                CloseCircuit
                Switch.handleLocalResponse
                        multimutex lock
                        Teardown circuit
                        remove pending payment in memory
                        multimutex unlock
                        ControlTower.success
                        signal payment complete
```

halseth:
After discussion with @joostjager we came to a design that should be both simpler and easier to review. This is done by splitting up the `forward` method, such that we can lookup the preimage in the cache between committing the circuit and forwarding the HTLC to the link. This let us avoid the use of a `multimutex`, since we can just bail out after committing the circuit and finding the preimage present.

PTAL @joostjager @cfromknecht @Roasbeef

htlcswitch/switch.go
halseth: Done. Note that there still might be some switch changes left (most notably making `SendHTLC` async), but I figured this PR was big enough as is now.

htlcswitch/switch.go
halseth: Very good idea, don!

htlcswitch/switch.go
```diff
@@ -867,60 +907,28 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
...
-	default:
-		log.Warnf("Received unknown response type: %T", pkt.htlc)
-		return
+		s.pendingPayments[paymentID] = payment
 	}
+	s.pendingMutex.Unlock()
+
+	// Set the result packet and signal that to the callers it is ready.
+	payment.result = pkt
+	close(payment.ready)
```
joostjager: Should this happen before teardown? To prevent this:

```
Teardown circuit

        CommitCircuit (succeeds)

        get or create pending payment in memory (create)

get or create pending payment in memory + signal result

        async handoff to link if commit succeeded (pay twice)

        wait for pending payment result
```

halseth: Yes, indeed! I also believe the pending payment should be created by `SendHTLC` after `CommitCircuit`, such that the two methods perform the operations in the opposite order: https://github.com/lightningnetwork/lnd/pull/2265/commits/6ffe4eb7183d09f96c4a2cec88472ec3dbd29873

halseth:
Blocked by https://github.com/lightningnetwork/lnd/pull/2501

Roasbeef:
Chatted with Johan offline, and I think we went down this rabbit hole due to one of my earlier comments that was a bit vague. We discussed a simpler version that hopefully should be easier to implement and also reason about during review as it involves much less shuffling around of the existing control flow and doesn't introduce any new dependancies.

Roasbeef:
Closing in favor of #2762.

PR #2275, opened 2018-12-05
fundingmanager: rejects funding requests with number of confirmations too large

### LINKED ISSUE #2273: No `FundingLocked` is sent from LND funded private channel
### Background

LND does not send a `FundingLocked` in a private channel where it is a funder, so channel never enters operational state. This does not happen when LND is a fundee.

When looking at message log while having an opening channel I see this (funding tx has 8 confirmations at this point):

```
<-- Init(,82)

--> ChannelReestablish(34d6ca2e25265c70abee4af7ad96f4d99d4804e9138586b33878805a1949cdd8,1,0,Some(0000000000000000000000000000000000000000000000000000000000000000),Some(02055c7176524bffc348d496980306d8687a1f2c3ca8e40498039421aae6c264ef))

<-- ChannelReestablish(34d6ca2e25265c70abee4af7ad96f4d99d4804e9138586b33878805a1949cdd8,1,0,Some(0000000000000000000000000000000000000000000000000000000000000000),Some(02967d09d22fdea3cdc26e1b34bd1ce6887bb91963be08299f522b67e18bf8b921))

--> FundingLocked(34d6ca2e25265c70abee4af7ad96f4d99d4804e9138586b33878805a1949cdd8,02348b6183f96595ab74364622c4553a0d6e30138cad80598e20be68c5c293626b)

<pings, gossip but no FundingLocked from LND>
```

### Your environment

0.5.1-beta commit=v0.5.1-beta-42-g668bd4db33b4c1ceb3b86836061e7faf99b3d080

### Steps to reproduce

Fund a private channel using LND.

### Expected behaviour

`FundingLocked` is sent by LND once funding tx gets enough confirmations.

### Actual behaviour

No `FundingLocked` is sent.

wpaulino:
In this PR, we allow the funding manager to reject any funding requests which impose a number of confirmations that is deemed as too large by the underlying `ChainNotifier`. This will properly prevent the funding state machine from continuing and avoid bugs like #2273, where we're unable to ever see a channel confirm as the notification registration will always fail due to a too large number of confirmations.

Fixes #2273.


fundingmanager.go
```diff
@@ -1228,10 +1227,10 @@ func (f *fundingManager) handleFundingAccept(fmsg *fundingAcceptMsg) {
 	// We'll also specify the responder's preference for the number of
 	// required confirmations, and also the set of channel constraints
 	// they've specified for commitment states we can create.
-	resCtx.reservation.SetNumConfsRequired(uint16(msg.MinAcceptDepth))
 	err = resCtx.reservation.CommitConstraints(
 		msg.CsvDelay, msg.MaxAcceptedHTLCs, msg.MaxValueInFlight,
 		msg.HtlcMinimum, msg.ChannelReserve, msg.DustLimit,
+		uint16(msg.MinAcceptDepth),
```
halseth: About time we wrap these in a struct? 😛

chainntnfs/txnotifier.go
```diff
@@ -10,14 +10,28 @@ import (
...
+	// ReorgSafetyLimit is the chain depth beyond which it is assumed a
+	// block will not be reorganized out of the chain. This is used to
+	// determine when to prune old confirmation requests so that reorgs are
+	// handled correctly. The average number of blocks in a day is a
+	// reasonable value to use.
+	ReorgSafetyLimit = 144
+
+	// MaxNumConfs is the maximum number of confirmations that can be
+	// requested on a transaction.
+	MaxNumConfs = ReorgSafetyLimit
```
halseth: Any reason why we won't set this to a different number? 144 seems a bit high for `MaxConfs` IMO.

wpaulino: Not really, it just needs to be `<= ReorgSafetyLimit` to ensure notifications are properly tracked within the `TxNotifier`.

fundingmanager.go
wpaulino: Fixed!

chainntnfs/txnotifier_test.go
```diff
@@ -99,6 +99,36 @@ func newMockHintCache() *mockHintCache {
 	}
 }
 
+// TestTxNotifierMaxConfs ensures that we are not able to register for more
+// confirmations on a transaction than the maximum supported.
+func TestTxNotifierMaxConfs(t *testing.T) {
```
halseth: 👍

chainntnfs/txnotifier.go
halseth: Should we set it lower?

fundingmanager.go
```diff
@@ -1082,10 +1082,15 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 
 	// We'll also validate and apply all the constraints the initiating
 	// party is attempting to dictate for our commitment transaction.
-	err = reservation.CommitConstraints(
-		msg.CsvDelay, msg.MaxAcceptedHTLCs, msg.MaxValueInFlight,
-		msg.HtlcMinimum, msg.ChannelReserve, msg.DustLimit,
-	)
+	channelConstraints := &channeldb.ChannelConstraints{
```
halseth: ooooh, this is nice

channeldb/channel.go
```diff
@@ -161,6 +161,13 @@ type ChannelConstraints struct {
 	// acted upon in the case of a unilateral channel closure or a contract
 	// breach.
 	MaxAcceptedHtlcs uint16
+
+	// CsvDelay is the relative time lock delay expressed in blocks. Any
+	// settled outputs that pay to the owner of this channel configuration
+	// MUST ensure that the delay branch uses this value as the relative
+	// time lock. Similarly, any HTLC's offered by this node should use
+	// this value as well.
+	CsvDelay uint16
```
halseth: Interesting... Does this work since `CsvDelay` is still serialized in the DB as part of the `ChannelConfig`?

channeldb/channel.go
```diff
@@ -168,6 +168,10 @@ type ChannelConstraints struct {
 	// time lock. Similarly, any HTLC's offered by this node should use
 	// this value as well.
 	CsvDelay uint16
+
+	// NumConfs is the minimum number of confirmations required before the
+	// channel can begin operating between both parties.
+	NumConfs uint32
```
halseth: must also be serialized in the db?

wpaulino: I'd say it doesn't need to be since we don't persist funding requests anyway. The only inconsistency I see is that when reading the `ChannelConfig` from the DB, this will always be 0. We can either leave a `NOTE` detailing so, or enforce the `MaxConfs` check elsewhere.

chainntnfs/txnotifier.go
wpaulino: Sure, but we must also keep in mind that lowering this also impacts the number of confirmations clients can request to the `ChainNotifier`. Do you have any suggestions on a lower number?

halseth: 143? Nah, can probably keep this as is 😛

lnwallet/reservation.go
```diff
@@ -282,72 +282,81 @@ func (r *ChannelReservation) SetNumConfsRequired(numConfs uint16) {
 // of satoshis that can be transferred in a single commitment. This function
 // will also attempt to verify the constraints for sanity, returning an error
 // if the parameters are seemed unsound.
-func (r *ChannelReservation) CommitConstraints(csvDelay, maxHtlcs uint16,
-	maxValueInFlight, minHtlc lnwire.MilliSatoshi,
-	chanReserve, dustLimit btcutil.Amount) error {
-
+func (r *ChannelReservation) CommitConstraints(constraints *channeldb.ChannelConstraints) error {
```
halseth: nit: long line

lnwallet/reservation.go
```diff
@@ -282,72 +282,81 @@ func (r *ChannelReservation) SetNumConfsRequired(numConfs uint16) {
...
-		return ErrCsvDelayTooLarge(csvDelay, maxDelay)
+	if constraints.CsvDelay > maxDelay {
+		return ErrCsvDelayTooLarge(constraints.CsvDelay, maxDelay)
 	}
 
 	// The dust limit should always be greater or equal to the channel
 	// reserve. The reservation request should be denied if otherwise.
-	if dustLimit > chanReserve {
-		return ErrChanReserveTooSmall(chanReserve, dustLimit)
+	if constraints.DustLimit > constraints.ChanReserve {
```
halseth: suggestion: rename `constraints -> c`?

fundingmanager.go
```diff
@@ -1241,6 +1243,7 @@ func (f *fundingManager) handleFundingAccept(fmsg *fundingAcceptMsg) {
 		MinHTLC:          msg.HtlcMinimum,
 		MaxAcceptedHtlcs: msg.MaxAcceptedHTLCs,
 		CsvDelay:         msg.CsvDelay,
+		NumConfs:         msg.MinAcceptDepth,
```
halseth: We can just check the value here instead of making it part of the constraints.

channeldb/channel.go
halseth: I don't think we should make it part of the `ChannelConstraints` if we don't plan to add it to the database. `ChannelConstraints` are meant for asymmetric values that need to be adhered to during channel operation.

lnwallet/reservation.go
wpaulino: Fixed.

fundingmanager.go
wpaulino: Fixed.

lnwallet/reservation.go
wpaulino: Fixed.

channeldb/channel.go
wpaulino: Fixed.

channeldb/channel.go
Roasbeef: Yeah it works since we just [list out all the fields when we're writing the config, and this is embedded](https://github.com/lightningnetwork/lnd/blob/master/channeldb/channel.go#L2211). Probably could have been left out this PR all together, but works as is.

fundingmanager.go
Roasbeef: Ah yeh see the rationale now.

chainntnfs/txnotifier.go
```diff
@@ -17,6 +17,10 @@ const (
 	// handled correctly. The average number of blocks in a day is a
 	// reasonable value to use.
 	ReorgSafetyLimit = 144
+
+	// MaxNumConfs is the maximum number of confirmations that can be
+	// requested on a transaction.
```
halseth: nit: would add the rationale for setting it equal to the comment.

fundingmanager.go
```diff
@@ -1230,6 +1230,18 @@ func (f *fundingManager) handleFundingAccept(fmsg *fundingAcceptMsg) {
 
 	fndgLog.Infof("Recv'd fundingResponse for pendingID(%x)", pendingChanID[:])
 
+	// The required number of confirmations should not be greater than the
+	// maximum number of confirmations required by the ChainNotifier to
+	// properly dispatch confirmations.
+	if msg.MinAcceptDepth > chainntnfs.MaxNumConfs {
+		err := lnwallet.ErrNumConfsTooLarge(
```
halseth: We should probably find a better pattern than having these errors defined in `lnwallet`, but this is okay for now.

PR #2277, opened 2018-12-05
Add --notls to disable TLS for REST endpoint

NicolasDorier:
Make it possible to disable TLS for REST endpoint.

I need it inside BTCPay because in the internal network, LND and BTCPay communicate together.

Internal network
```
BTCPAY <-> LND
```

For internet requests, I use HTTPS dispensed by NGINX.

Internet 
```
LND <-> NGINX <-> Internet request
```

Now, both BTCPay and NGINX complain if TLS is activated at LND level because the certificate is untrusted. Making both of them trust this certificate is significant work, so I am just disabling TLS with this commit on my fork.

### DIFF
--- a/config.go
+++ b/config.go
@@ -190,6 +190,7 @@ type config struct {
 	RawRESTListeners []string `long:"restlisten" description:"Add an interface/port/socket to listen for REST connections"`
 	RawListeners     []string `long:"listen" description:"Add an interface/port to listen for peer connections"`
 	RawExternalIPs   []string `long:"externalip" description:"Add an ip:port to the list of local addresses we claim to listen on to peers. If a port is not specified, the default (9735) will be used regardless of other parameters"`
+	DisableTLS       bool     `long:"notls" description:"Disable TLS for RPC connetions"`
 	RPCListeners     []net.Addr
 	RESTListeners    []net.Addr
 	Listeners        []net.Addr
--- a/lnd.go
+++ b/lnd.go
@@ -713,7 +713,15 @@ func waitForWalletPassword(grpcEndpoints, restEndpoints []net.Addr,
 	srv := &http.Server{Handler: mux}
 
 	for _, restEndpoint := range restEndpoints {
-		lis, err := lncfg.TLSListenOnAddress(restEndpoint, tlsConf)
+		var (
+			lis net.Listener
+			err error
+		)
+		if !cfg.DisableTLS {
+			lis, err = lncfg.TLSListenOnAddress(restEndpoint, tlsConf)
+		} else {
+			lis, err = net.Listen("tcp", restEndpoint.String())
+		}
 		if err != nil {
 			ltndLog.Errorf(
 				"password gRPC proxy unable to listen on %s",


Roasbeef:
>  Making both of them trust this certificate is significant work,
Wouldn't it be as simple as accepting the self signed cert? So just skip validation all together. Is the work on the nginx or C# side?

NicolasDorier:
> Wouldn't it be as simple as accepting the self signed cert?

Not so simple. In C# level I can skip HTTPS validation, but then it means that I can't use the `IHttpClientFactory`, which mean it then go into [this type of issue](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) and [this type of issue](https://contrivedexample.com/2017/07/01/using-httpclient-as-it-was-intended-because-youre-not/). (a very .NET Core specific problem)

Then after, I need to make it pass at NGINX level, which mean I need to do additional research as well as having the nginx container access the LND container just to get the certificate. Headaches which should not even be needed because we are in an internal network.

cfromknecht:
This sounds like more of a complication in architecture and language decisions made by BTCPay, rather than a pain-point caused by LND's enforcement of basic security practices.

> having the nginx container access the LND container just to get the certificate. Headaches which should not even be needed because we are in an internal network.

You want to communicate _between_ docker instances _in plaintext_? Assuming a network is private is a shaky thing to hang your hat on when totally unecessary, defensible, and historically inaccurate.

> which mean I need to do additional research

Is this really a substantive argument?

If BTCPay is to be as successful as you and I both want, I'd prefer we have a solution that doesn't compromise basic security safeguards for a large set of users running LND, just for the sake of convenience. It's hard to believe that this is the only way to solve the problem despite the additional architectural constraints.

IMO this introduces too much of a configuration footgun to be present in the mainline repo; all to save a trivial rebase every few months. I do hope some further research will lead to a workable configuration w/ TLS enabled. I'm happy to help out on this front.

NicolasDorier:
Blindly trusting a certificate does not protect against network level attacks.
If you don't want this commit, feel free to close, but I will keep it in btcpayserver LND fork.

wereHamster:
I would also appreciate an option to disable TLS. I understand the goal to be safe-by-default, but sometimes that makes the tool much harder to use in certain situations.

I agree with @NicolasDorier that blindly trusting (self-signed) certificates does not offer a good protection. Even worse, it forces projects like https://lightningjoule.com/ to tell people to [click away the Chrome self-signed certificate warning](https://github.com/wbobeirne/joule-extension/blob/25ad56541f4710f18f97142bb36822fbfece6cf2/src/app/components/SelectNode/InputAddress.tsx#L59-L60). So much about educating people of good practices.

In my case I want to do TLS termination in nginx, so that I can automatically provision Let's Encrypt certificates. The connection between nginx and lnd is already protected by the underlying network infrastructure. Anyone who has access to that, also has access to the nginx and lnd processes. So TLS on that connection doesn't provide much more security as is.

NicolasDorier:
Same position as @wereHamster. The TLS termination is in nginx, even if I made TLS works on the internal server between btcpay<->lnd and lnd<->nginx, this is just busy work providing no more security and is yet another new thing which can break when a new update for LND come.

ramontayag:
Hi, is this still something that you intend to merge? In my case, I want to set it up in a Kubernetes cluster. If you throw in TLS for individual services within the cluster, it complicates things. I'm open to other methods of handling this in Kubernetes, but I haven't heard anything convincing.

NicolasDorier:
@ramontayag we are using it in our fork https://github.com/btcpayserver/lnd/commit/bd96b95d062a675f0fc70117a1079d09aa0e512f 

I think this can be closed if there is no intent to merge.

ramontayag:
Thanks @NicolasDorier I will use that in the mean time.

halseth:
@NicolasDorier Is this something you still want? If yes we can probably rebase and merge it.

config.go
```diff
@@ -190,6 +190,7 @@ type config struct {
 	RawRESTListeners []string `long:"restlisten" description:"Add an interface/port/socket to listen for REST connections"`
 	RawListeners     []string `long:"listen" description:"Add an interface/port to listen for peer connections"`
 	RawExternalIPs   []string `long:"externalip" description:"Add an ip:port to the list of local addresses we claim to listen on to peers. If a port is not specified, the default (9735) will be used regardless of other parameters"`
+	DisableTLS       bool     `long:"notls" description:"Disable TLS for RPC connetions"`
```
halseth, line 193: type

lnd.go
```diff
@@ -713,7 +713,15 @@ func waitForWalletPassword(grpcEndpoints, restEndpoints []net.Addr,
 	srv := &http.Server{Handler: mux}
 
 	for _, restEndpoint := range restEndpoints {
-		lis, err := lncfg.TLSListenOnAddress(restEndpoint, tlsConf)
+		var (
+			lis net.Listener
+			err error
+		)
+		if !cfg.DisableTLS {
```
halseth, line 720: should do the same for the subservers and the wallet unlocker.

halseth (changes requested):

NicolasDorier:
@halseth  yes it is something I still want.

We are using it in our fork and keep rebasing on our fork. Feel free to takeover this PR.

NicolasDorier:
Here is on our fork: https://github.com/btcpayserver/lnd/commit/98ec34c5231d8262e5732d0993d796ff0f4ce243

halseth:
@NicolasDorier Can you check if https://github.com/lightningnetwork/lnd/pull/2277 does what you need?

NicolasDorier:
yes, it seems to do the job, thanks. Closing this one then.

PR #2280, opened 2018-12-05
Try to parse bitcoind cookie to get RPC credentials before automatic setup

### LINKED ISSUE #2266: Unable to setup Bitcoind cookie authentication in Docker environment
I believe that checks in `config.go` are too restrictive and prevent setting up `lnd` + `bitcoind` with cookie authentication in `docker` environment. Consider the following setup:
1. `bitcoin.conf` is outside of `bitcoind` data directory and configured with the `-conf=` option
1. `lnd.conf` is outside `lnddir` and configured with `--configfile=` option
1. `bitcoind` and `lnd` are in separate containers (network namespace is different)
1. `bitcoind` data directory is mounted to `lnd` container read-only so that `lnd` is able to read `.cookie` file

While setting `bitcoind.dir`, `bitcoind.rpchost`, `bitcoind.zmqpubrawblock` and `bitcoind.zmqpubrawtx` should be enough to make this setup work, in fact it doesn't. Moreover, leaving `bitcoin.conf` in its default location and setting fake `rpcuser` and `rpcpass` doesn't help – `lnd` is unable to authenticate.

I think checks in `config.go` should be relaxed as RPC password authentication in `bitcoind` is deprecated, and the aforementioned setup is quite typical in containerized environment.

leshik:
Closes #2266.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1154,13 +1154,6 @@ func parseRPCParams(cConfig *chainConfig, nodeConfig interface{}, net chainCode,
 			}
 		}
 
-		// If all of RPCUser, RPCPass, ZMQBlockHost, and ZMQTxHost are
-		// set, we assume those parameters are good to use.
-		if conf.RPCUser != "" && conf.RPCPass != "" &&
-			conf.ZMQPubRawBlock != "" && conf.ZMQPubRawTx != "" {
-			return nil
-		}
-
 		// Get the daemon name for displaying proper errors.
 		switch net {
 		case bitcoinChain:
@@ -1171,6 +1164,24 @@ func parseRPCParams(cConfig *chainConfig, nodeConfig interface{}, net chainCode,
 			daemonName = "litecoind"
 			confDir = conf.Dir
 			confFile = "litecoin"
+		}
+
+		// If all of RPCUser, RPCPass, ZMQBlockHost, and ZMQTxHost are
+		// set, we assume those parameters are good to use.
+		if conf.ZMQPubRawBlock != "" && conf.ZMQPubRawTx != "" {
+			if conf.RPCUser != "" && conf.RPCPass != "" {
+				return nil
+			}
+			// Try to get RPCUser and RPCPass from an auth cookie.
+			rpcUser, rpcPass, err :=
+				getBitcoindRPCCredentialsFromCookie(confDir)
+			if err != nil {
+				return fmt.Errorf("unable to retrieve RPC "+
+					"credentials from cookie: %v", err,
+				)
+			}
+			conf.RPCUser, conf.RPCPass = rpcUser, rpcPass
+			return nil
 		}
 
 		// If not all of the parameters are set, we'll assume the user
@@ -1264,7 +1275,7 @@ func extractBtcdRPCParams(btcdConfigPath string) (string, string, error) {
 	}
 	passSubmatches := rpcPassRegexp.FindSubmatch(configContents)
 	if passSubmatches == nil {
-		return "", "", fmt.Errorf("unable to find rpcuser in config")
+		return "", "", fmt.Errorf("unable to find rpcpass in config")
 	}
 
 	return string(userSubmatches[1]), string(passSubmatches[1]), nil
@@ -1275,7 +1286,8 @@ func extractBtcdRPCParams(btcdConfigPath string) (string, string, error) {
 // location of bitcoind's bitcoin.conf on the target system. The routine looks
 // for a cookie first, optionally following the datadir configuration option in
 // the bitcoin.conf. If it doesn't find one, it looks for rpcuser/rpcpassword.
-func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string, string, error) {
+func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string,
+	string, string, error) {
 	// First, we'll open up the bitcoind configuration file found at the
 	// target destination.
 	bitcoindConfigFile, err := os.Open(bitcoindConfigPath)
@@ -1319,8 +1331,7 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 		return "", "", "", "", err
 	}
 
-	// Next, we'll try to find an auth cookie. We need to detect the chain
-	// by seeing if one is specified in the configuration file.
+	// Next, we'll try to find an auth cookie.
 	dataDir := path.Dir(bitcoindConfigPath)
 	dataDirRE, err := regexp.Compile(`(?m)^\s*datadir\s*=\s*([^\s]+)`)
 	if err != nil {
@@ -1331,24 +1342,13 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 		dataDir = string(dataDirSubmatches[1])
 	}
 
-	chainDir := "/"
-	switch activeNetParams.Params.Name {
-	case "testnet3":
-		chainDir = "/testnet3/"
-	case "testnet4":
-		chainDir = "/testnet4/"
-	case "regtest":
-		chainDir = "/regtest/"
-	}
-
-	cookie, err := ioutil.ReadFile(dataDir + chainDir + ".cookie")
+	rpcUser, rpcPass, err := getBitcoindRPCCredentialsFromCookie(dataDir)
 	if err == nil {
-		splitCookie := strings.Split(string(cookie), ":")
-		if len(splitCookie) == 2 {
-			return splitCookie[0], splitCookie[1], zmqBlockHost,
-				zmqTxHost, nil
-		}
+		return rpcUser, rpcPass, zmqBlockHost, zmqTxHost, nil
 	}
+	fmt.Printf("unable to retrieve RPC credentials from cookie: %v, "+
+		"falling back to rpcuser/rpcpassword\n", err.Error(),
+	)
 
 	// We didn't find a cookie, so we attempt to locate the RPC user using
 	// a regular expression. If we  don't have a match for our regular
@@ -1462,4 +1462,34 @@ func normalizeNetwork(network string) string {
 	}
 
 	return network
+}
+
+// getBitcoindRPCCredentialsFromCookie attempts to extract RPC credentials from
+// the bitcoind authentication cookie.
+func getBitcoindRPCCredentialsFromCookie(bitcoindDataDir string) (string,
+	string, error) {
+	fmt.Println("Attempting to extract RPC credentials from auth cookie")
+	// We need to detect the chain by seeing if one is specified
+	// in the configuration file.
+	chainDir := "/"
+	switch activeNetParams.Params.Name {
+	case "testnet3":
+		chainDir = "/testnet3/"
+	case "testnet4":
+		chainDir = "/testnet4/"
+	case "regtest":
+		chainDir = "/regtest/"
+	}
+
+	cookie, err := ioutil.ReadFile(bitcoindDataDir + chainDir + ".cookie")
+	if err != nil {
+		return "", "", err
+	}
+	splitCookie := strings.Split(string(cookie), ":")
+	if len(splitCookie) != 2 {
+		return "", "", fmt.Errorf("invalid data in %s%s.cookie found",
+			bitcoindDataDir, chainDir)
+	}
+
+	return splitCookie[0], splitCookie[1], nil
 }
--- a/config.go
+++ b/config.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/url"
 	"os"
 	"os/user"
 	"path"
@@ -1142,8 +1143,8 @@ func parseRPCParams(cConfig *chainConfig, nodeConfig interface{}, net chainCode,
 		}
 
 	case *bitcoindConfig:
-		// Ensure that if the ZMQ options are set, that they are not
-		// equal.
+		// Ensure that if the ZMQ options are set, that they are valid
+		// and not equal.
 		if conf.ZMQPubRawBlock != "" && conf.ZMQPubRawTx != "" {
 			err := checkZMQOptions(
 				conf.ZMQPubRawBlock, conf.ZMQPubRawTx,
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -1378,3 +1387,27 @@ func normalizeNetwork(network string) string {
 
 	return network
 }
+
+// We need to detect the chain by seeing if one is specified in the configuration file.
```
wpaulino: Nit: function comment should be of the form `getBitcoindRPCCredentialsFromCookie ...` that describes what the function does at a glance, like all the others.

config.go
```diff
@@ -1362,10 +1355,26 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 // ZMQ rawblock and rawtx notifications are different.
 func checkZMQOptions(zmqBlockHost, zmqTxHost string) error {
 	if zmqBlockHost == zmqTxHost {
-		return errors.New("zmqpubrawblock and zmqpubrawtx must be set" +
+		return errors.New("zmqpubrawblock and zmqpubrawtx must be set " +
 			"to different addresses")
 	}
 
+	zmqBlockHostURL, err := url.Parse(zmqBlockHost)
```
wpaulino: Nit: could use `ResolveTCPAddr` instead to make a stricter check as that's what we require.

config.go
```diff
@@ -1362,10 +1355,26 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
+	if err != nil {
+		return errors.New("invalid data found in zmqpubrawblock")
+	}
+
+	zmqTxHostURL, err := url.Parse(zmqTxHost)
+	if err != nil {
+		return errors.New("invalid data found in zmqpubrawtx")
+	}
+
+	if strings.Split(zmqBlockHostURL.Host, ":")[0] == "0.0.0.0" ||
```
wpaulino: Nit: could use `net.SplitHostPort` instead.

config.go
leshik: @wpaulino you mean, to add `net.ResolveTCPAddr` in addition to `url.Parse`? The latter is needed anyway as the option value is in the URL format, e.g. `tcp://0.0.0.0:8334`

leshik: I reworked the whole function, hope I checked all cases.

config.go
leshik: Done.

leshik:
> Not a blocker, but would prefer if the commits were split up to reason about the changes better. For example, the changes in checkZMQOptions could warrant their own commit with a message along the lines of:
> 
> `config: enforce stricter checks when parsing ZMQ options`

@wpaulino I've split it to two commits.

config.go
```diff
@@ -1155,6 +1148,22 @@ func parseRPCParams(cConfig *chainConfig, nodeConfig interface{}, net chainCode,
...
+				return nil
+			}
+			// Try to get RPCUser and RPCPass from an auth cookie
+			rpcUser, rpcPass, err :=
+				getBitcoindRPCCredentialsFromCookie(confDir)
+			if err == nil {
+				conf.RPCUser, conf.RPCPass = rpcUser, rpcPass
+				return nil
+			}
+			fmt.Println(err)
```
halseth: Why print instead of return error here?

config.go
```diff
@@ -1362,13 +1363,65 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
 			"to different addresses")
 	}
 
+	// TCP transport specific checks.
+	if zmqBlockURL.Scheme == "tcp" && zmqTxURL.Scheme == "tcp" {
+		zmqBlockResolved, err := net.ResolveTCPAddr("tcp", zmqBlockURL.Host)
+		if err != nil {
+			return err
+		}
+		zmqBlockHost, zmqBlockPort, err :=
```
halseth: style nit:
```
zmqBlockHost, zmqBlockPort, err := net.SplitHostPort(
           zmqBlockResolved.String(),
)
```

config.go
```diff
@@ -1362,13 +1363,65 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
+		}
+		zmqTxHost, zmqTxPort, err := net.SplitHostPort(zmqTxResolved.String())
+		if err != nil {
+			return err
+		}
+		zmqTxHostIP := net.ParseIP(zmqTxHost)
+
+		if zmqBlockPort == zmqTxPort &&
+			// Protects against one is 127.0.0.1 and the other is ::1
+			((zmqBlockHostIP.IsLoopback() && zmqTxHostIP.IsLoopback()) ||
```
halseth: This is a bit hard to read. Can you split the cases into two, and ideally return a slightly different error to differentiate? :)

config.go
leshik: @halseth because we don't have to stop and return here, the print is purely informational (otherwise one can't know that the cookie e.g. contains garbage, or the file wasn't found).

config.go
leshik: Uh, you really want different errors here? The `IsLoopback` is used here just because `Equal` returns `false` in case of `127.0.0.1` and `::1` (basically, because `127.0.0.1` maps to `0:0:0:0:0:ffff:7f00:1`). So it's kinda workaround here.

config.go
wpaulino: Ah yes, that was my mistake, `net.ResolveTCPAddr` isn't needed here. The new changes should mostly be unnecessary now. `url.Parse` can't parse addresses without a scheme prefix, e.g. `0.0.0.0:8334`, so as long as that isn't `tcp`, we can just return an error like `invalid ZMQ address, must be TCP` or something like that.

config.go
wpaulino: In that case, would suggest adding some additional context to it like so:

`Unable to retrieve RPC credentials from cookie (path=/path/to/cookie), falling back to rpcuser/rpcpassword`

config.go
leshik: @wpaulino
> we can just return an error like `invalid ZMQ address, must be TCP`

Well, but why? Is it because LND currently can't use anything else than TCP?

leshik: @wpaulino I found in tests that at least `ipc:///` is supported, so maybe we shouldn't be so strict about TCP-only sockets...

leshik:
There is another issue I stumbled upon while testing this config with `.cookie` authentication. Each time `bitcoind` starts, it generates a new cookie, which means LND should be restarted too, otherwise RPC authentication breaks. Would be nice if `.cookie` were re-read in this case so one didn't have to relaunch LND, unlock it and then wait till it catches up. Might be a separate PR IMO, but I'm not sure how this should be addressed. Suggestions are welcome. :)

config.go
leshik: Besides that, I believe `net.ResolveTCPAddr` is a good idea anyway, as it's better resolve once and then work with the IP instead of issuing DNS requests on every connection.

config.go
halseth: you could do 
```
switch {
// Comment 1
case zmqBlockHostIP.IsLoopback() && zmqTxHostIP.IsLoopback:
     fallthrough

// Comment 2..
case zmqBlockHostIP.Equal(zmqTxHostIP):
     return errors.New("....")
}
```

config.go
```diff
@@ -1362,13 +1363,65 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
+			return err
+		}
+		zmqBlockHost, zmqBlockPort, err :=
+			net.SplitHostPort(zmqBlockResolved.String())
+		if err != nil {
+			return err
+		}
+		zmqBlockHostIP := net.ParseIP(zmqBlockHost)
+
+		zmqTxResolved, err := net.ResolveTCPAddr("tcp", zmqTxURL.Host)
```
halseth: You should add some comments to what is being checked here.

leshik: Done.

config.go
leshik: Ah, ok. Done!

config.go
leshik: Done.

leshik:
@halseth @wpaulino Is there anything else that prevents this from merging?

config.go
```diff
@@ -1362,13 +1363,75 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
 }
 
 // checkZMQOptions ensures that the provided addresses to use as the hosts for
-// ZMQ rawblock and rawtx notifications are different.
+// ZMQ rawblock and rawtx notifications are valid and different.
 func checkZMQOptions(zmqBlockHost, zmqTxHost string) error {
-	if zmqBlockHost == zmqTxHost {
-		return errors.New("zmqpubrawblock and zmqpubrawtx must be set" +
+	// ZMQ can be used with diferent transports
```
halseth: nit: s/diferent/different

config.go
```diff
@@ -1372,13 +1373,75 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
 
+	// TCP transport specific checks.
+	if zmqBlockURL.Scheme == "tcp" && zmqTxURL.Scheme == "tcp" {
+		// Ensure the host:port is correct and resolve it to ip:port if needed,
+		// then split the result to IP and port
+		zmqBlockResolved, err := net.ResolveTCPAddr("tcp", zmqBlockURL.Host)
+		if err != nil {
+			return err
+		}
+		zmqBlockHost, zmqBlockPort, err := net.SplitHostPort(
```
wpaulino: Don't think there's a need for this to retrieve the IP and port. `ResolveTCPAddr` returns a `TCPAddr`, which already exposes the IP and port as fields.

config.go
```diff
@@ -1323,24 +1334,13 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
+	rpcUser, rpcPass, err := getBitcoindRPCCredentialsFromCookie(dataDir)
 	if err == nil {
-		splitCookie := strings.Split(string(cookie), ":")
-		if len(splitCookie) == 2 {
-			return splitCookie[0], splitCookie[1], zmqBlockHost,
-				zmqTxHost, nil
-		}
+		return rpcUser, rpcPass, zmqBlockHost, zmqTxHost, nil
 	}
+	fmt.Println("unable to retrieve RPC credentials from cookie: " +
```
wpaulino: Nit: could use `fmt.Printf` here.

config.go
```diff
@@ -1454,3 +1454,33 @@ func normalizeNetwork(network string) string {
...
+		chainDir = "/regtest/"
+	}
+
+	cookie, err := ioutil.ReadFile(bitcoindDataDir + chainDir + ".cookie")
+	if err != nil {
+		return "", "", err
+	}
+	splitCookie := strings.Split(string(cookie), ":")
+	if len(splitCookie) != 2 {
+		return "", "", errors.New("invalid data in " + bitcoindDataDir +
```
wpaulino: Nit: could use `fmt.Errorf` here.

config.go
leshik: Oh yeah, you're right! Thanks, fixed.

config.go
leshik: Fixed.

config.go
leshik: Fixed.

config.go
```diff
@@ -1173,6 +1166,24 @@ func parseRPCParams(cConfig *chainConfig, nodeConfig interface{}, net chainCode,
 			confFile = "litecoin"
 		}
 
+		// If all of RPCUser, RPCPass, ZMQBlockHost, and ZMQTxHost are
+		// set, we assume those parameters are good to use.
+		if conf.ZMQPubRawBlock != "" && conf.ZMQPubRawTx != "" {
+			if conf.RPCUser != "" && conf.RPCPass != "" {
+				return nil
+			}
+			// Try to get RPCUser and RPCPass from an auth cookie
```
wpaulino: Nit: add period.

config.go
```diff
@@ -1173,6 +1166,24 @@ func parseRPCParams(cConfig *chainConfig, nodeConfig interface{}, net chainCode,
...
+			rpcUser, rpcPass, err :=
```
wpaulino: This snippet below can be rewritten like so:
```
user, pass, err := getCookieCreds()
if err != nil {
    return fmt.Errorf("unable to retrieve RPC credentials from cookie: %v", err)
}

conf.RPCUser, conf.RPCPass = user, pass
return nil
```

config.go
```diff
@@ -1380,13 +1381,61 @@ func extractBitcoindRPCParams(bitcoindConfigPath string) (string, string, string
...
 // checkZMQOptions ensures that the provided addresses to use as the hosts for
-// ZMQ rawblock and rawtx notifications are different.
+// ZMQ rawblock and rawtx notifications are valid and different.
 func checkZMQOptions(zmqBlockHost, zmqTxHost string) error {
-	if zmqBlockHost == zmqTxHost {
-		return errors.New("zmqpubrawblock and zmqpubrawtx must be set" +
+	// ZMQ can be used with different transports
+	// (http://api.zeromq.org/4-0:zmq-bind). At first, parse the socket URL
+	// and perform generalized comparison.
+	zmqBlockURL, err := url.Parse(zmqBlockHost)
```
wpaulino: Realized that by making this change, users will now be forced to prepend their ZMQ values with `tcp://`, so it'll break some existing configs out in the wild. Because of that, I think we should defer this until we make the breaking changes to fully support bitcoind's cookie authentication.

config.go
leshik: Fixed.

config.go
leshik: Done.

config.go
leshik: Hmm, I didn't think of this before. Added checks to ensure compatibility with old setups, will it go through?

bhandras:
Hi @leshik 
This PR needs a rebase.

leshik:
@bhandras I'm sorry, I'm almost out of time right now, if it can wait I will look into it later, otherwise, anyone else can pick it up.

bhandras:
> @bhandras I'm sorry, I'm almost out of time right now, if it can wait I will look into it later, otherwise, anyone else can pick it up.

Thanks for the update @leshik.

Roasbeef:
Closing due to inactivity.

PR #2290, opened 2018-12-07
Add `debughtlc` option to `sample-lnd.conf`.

BrianSipple:
I saw the `debughtlc` option being [used in the tutorial](https://dev.lightning.community/tutorial/01-lncli/index.html#configuring-lndconf) and noticed it wasn't a part of the sample config. 

Figured it would be useful 😄.

### DIFF
--- a/config.go
+++ b/config.go
@@ -205,7 +205,7 @@ type config struct {
 
 	Profile string `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65535"`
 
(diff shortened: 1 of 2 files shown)


cfromknecht:
`debughtlc` is only used for testing and not available in production builds of LND because it's not safe to use with real funds. We removed it from the sample-conf a few months back. The docs you linked to are quite out of date, maybe it's time we update those as well :)

BrianSipple:
@cfromknecht Gotcha. It looks like https://github.com/lightninglabs/lightning-dev-site/pull/50 did actually take care of that. Perhaps those changes just haven't been deployed yet or somethings being cached?

Roasbeef:
Echoing what was said above, we'd prefer to not promote this flag, as it even requires building `lnd` in a special way to expose it. Instead, we'll be updating the dev site with the latest conf settings, thanks!

PR #2293, opened 2018-12-07
[autopilot] weighted choice unit tests and optimizations

halseth:
This is a follow-up to #2006 that adds unit tests for the `weightedChoice` and `chooseN` algorithms.

In working on the test cases I noticed that the algorithms became slow when handling a large number of choices. This turned out to mostly be caused by map allocations and iterations. A commit to optimize this by instead keeping what we need in slices is added.

Builds on #2304

### DIFF
--- a/autopilot/choice.go
+++ b/autopilot/choice.go
@@ -52,7 +52,7 @@ func weightedChoice(s map[NodeID]*AttachmentDirective) (NodeID, error) {
 
 // chooseN picks at random min[n, len(s)] nodes if from the
 // AttachmentDirectives map, with a probability weighted by their score.
-func chooseN(n int, s map[NodeID]*AttachmentDirective) (
+func chooseN(n uint32, s map[NodeID]*AttachmentDirective) (
 	map[NodeID]*AttachmentDirective, error) {
 
 	// Keep a map of nodes not yet choosen.
@@ -64,7 +64,7 @@ func chooseN(n int, s map[NodeID]*AttachmentDirective) (
 	// Pick a weighted choice from the remaining nodes as long as there are
 	// nodes left, and we haven't already picked n.
 	chosen := make(map[NodeID]*AttachmentDirective)
-	for len(chosen) < n && len(rem) > 0 {
+	for len(chosen) < int(n) && len(rem) > 0 {
 		choice, err := weightedChoice(rem)
 		if err == ErrNoPositive {
 			return chosen, nil
--- a/autopilot/choice.go
+++ b/autopilot/choice.go
@@ -10,28 +10,23 @@ import (
 // weights left to choose from.
 var ErrNoPositive = errors.New("no positive weights left")
 
-// weightedChoice draws a random index from the map of channel candidates, with
-// a probability propotional to their score.
-func weightedChoice(s map[NodeID]*AttachmentDirective) (NodeID, error) {
-	// Calculate the sum of scores found in the map.
+// weightedChoice draws a random index from the slice of weights, with a
+// probability propotional to the weight at the given index.
+func weightedChoice(w []float64) (int, error) {
+	// Calculate the sum of weights.
 	var sum float64
-	for _, v := range s {
-		sum += v.Score
+	for _, v := range w {
+		sum += v
 	}
 
 	if sum <= 0 {
-		return NodeID{}, ErrNoPositive
+		return 0, ErrNoPositive
 	}
 
-	// Create a map of normalized scores such, that they sum to 1.0.
-	norm := make(map[NodeID]float64)
-	for k, v := range s {
-		norm[k] = v.Score / sum
-	}
-
-	// Pick a random number in the range [0.0, 1.0), and iterate the map
-	// until the number goes below 0. This means that each index is picked
-	// with a probablity equal to their normalized score.
+	// Pick a random number in the range [0.0, 1.0) and multiply it with
+	// the sum of weights. Then we'll iterate the weights until the number
+	// goes below 0. This means that each index is picked with a probablity
+	// equal to their normalized score.
 	//
 	// Example:
 	// Items with scores [1, 5, 2, 2]
@@ -40,14 +35,15 @@ func weightedChoice(s map[NodeID]*AttachmentDirective) (NodeID, error) {
 	// in [0, 1.0]:
 	// [|-0.1-||-----0.5-----||--0.2--||--0.2--|]
 	// The following loop is now equivalent to "hitting" the intervals.
-	r := rand.Float64()
-	for k, v := range norm {
-		r -= v
+	r := rand.Float64() * sum
+	for i := range w {
+		r -= w[i]
 		if r <= 0 {
-			return k, nil
+			return i, nil
 		}
 	}
-	return NodeID{}, fmt.Errorf("unable to make choice")
+
+	return 0, fmt.Errorf("unable to make choice")
 }
 
 // chooseN picks at random min[n, len(s)] nodes if from the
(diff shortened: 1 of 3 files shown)


autopilot/choice_test.go
```diff
@@ -13,6 +13,101 @@ var (
...
+	}
+}
+
+// TestWeightedChoiceSingleIndex tests that choosing randomly in a slice with one
+// positive element always returns that one index.
+func TestWeightedChoiceSingleIndex(t *testing.T) {
+	t.Parallel()
+
+	w := []float64{0, 0.5, 0, 0}
+	for i := 0; i < 1000; i++ {
```
Roasbeef: We can modify this slightly to use `testing.Quick` instead. It would look something like: 
```go
property := func(weights [10]float64) bool {
    allButOneZero := func() (bool, float64) {
        var (
           numZero uint32
           nonZero float64
        )
       
        for _, w := range weights {
          if w != 0 {
             numZero++
             nonZero = w
         }
       }
       
      return numZero == 1, nonZero
    }
  
   conditionMet, nonZeroElem := allButOneZero()
   if conditionMet {
    choice, err := weightedChoice(w) 
    if err != nil {
      return false
    }
    return choice == nonZeroEleme
   }
}

if err := quick.Check(property, nil); err != nil {
    t.Fatalf(err)
}
```

Roasbeef: The above example illustrates how to write property based tests in Go using `testing/quick`. Given a set of random (or constrained data via the config), we make a predicate function that targets a particular property. In this case, the property is: if there's a single non-zero element, then we should always select that. I made it a fixed slice to make the code a bit shorter, but we can also accept a `[]float64` and use the config/generator to restrict the form of the input data.

autopilot/choice_test.go
```diff
@@ -13,6 +13,101 @@ var (
...
+	for i := 0; i < iterations; i++ {
+		res, err := weightedChoice(w)
+		if err != nil {
+			t.Fatalf("unable to get choice: %v", err)
+		}
+		chosen[res]++
+	}
+
+	// Since this is random we check that the number of times chosen is
+	// within %20 of the expected value.
```
Roasbeef: %20 -> 20%

autopilot/choice_test.go
```diff
@@ -13,6 +13,101 @@ var (
...
+	// The sum of choices must be exactly iterations of course.
+	if sum != iterations {
+		t.Fatalf("expected a sum of %d, got %d", iterations, sum)
+	}
+
+}
+
+// TestWeightedChoiceDistribution asserts that the weighted choice algorithm
+// chooses among indexes according to their scores.
+func TestWeightedChoiceDistribution(t *testing.T) {
```
Roasbeef: I think a similar technique as outlined above (using `testing/quick`) can be applied here as well. One way to go about it would be to compare two ideas: one that will always pick option A 25% of the time (as an example) and the other which is actually the `weightedChoice` function. We'd that assert the two to be within some statistical difference of each other.

autopilot/choice.go
```diff
@@ -10,28 +10,23 @@ import (
...
-		norm[k] = v.Score / sum
-	}
-
-	// Pick a random number in the range [0.0, 1.0), and iterate the map
-	// until the number goes below 0. This means that each index is picked
-	// with a probablity equal to their normalized score.
+	// Pick a random number in the range [0.0, 1.0) and multiply it with
+	// the sum of weights. Then we'll iterate the weights until the number
+	// goes below 0. This means that each index is picked with a probablity
+	// equal to their normalized score.
```
Roasbeef: Comment is now slightly out of data since the scores are no longer being normalized.

autopilot/choice.go
```diff
@@ -40,40 +35,52 @@ func weightedChoice(s map[NodeID]*AttachmentDirective) (NodeID, error) {
...
 }
 
 // chooseN picks at random min[n, len(s)] nodes if from the
 // AttachmentDirectives map, with a probability weighted by their score.
 func chooseN(n int, s map[NodeID]*AttachmentDirective) (
 	map[NodeID]*AttachmentDirective, error) {
 
-	// Keep a map of nodes not yet choosen.
-	rem := make(map[NodeID]*AttachmentDirective)
+	// Keep track of the number of nodes not yet chosen, in additon to
```
Roasbeef: additon -> addition

autopilot/choice_test.go
```diff
@@ -0,0 +1,154 @@
...
+	nID4 = NodeID([33]byte{4})
+)
+
+// TestChooseNEmptyMap checks that chooseN returns an empty result when no
+// nodes are chosen among.
+func TestChooseNEmptyMap(t *testing.T) {
+	t.Parallel()
+
+	nodes := map[NodeID]*AttachmentDirective{}
+	for i := 0; i < 10; i++ {
```
Roasbeef: Loop can go away here with usage of `testing/quick`.

autopilot/choice_test.go
```diff
@@ -0,0 +1,154 @@
...
+		binary.BigEndian.PutUint32(nID[:], i)
+		nodes[nID] = &AttachmentDirective{
+			Score: s,
+		}
+	}
+
+	// For each value of N we'll check that the nodes are picked the
+	// expected number of times over time.
+	for _, n := range []int{1, 5, 10, 20, 50} {
+		// Since choosing more nodes will be slower we decrease the
```
Roasbeef: slower -> slowly

autopilot/choice_test.go
```diff
@@ -0,0 +1,154 @@
...
+		for i := 0; i < iterations; i++ {
+			res, err := chooseN(n, nodes)
+			if err != nil {
+				t.Fatalf("failed choosing nodes: %v", err)
+			}
+
+			for nID := range res {
+				count[nID]++
+			}
+
```
Roasbeef: Extra new line.

autopilot/choice_test.go
halseth: Done!

autopilot/choice_test.go
halseth: Fixed.

autopilot/choice_test.go
halseth: Done!

autopilot/choice.go
halseth: "each index is picked with a probablity equal to their normalized score " is still true though :p

autopilot/choice.go
halseth: Fixed.

autopilot/choice_test.go
halseth: Done!

autopilot/choice_test.go
halseth: Fixed.

autopilot/choice_test.go
halseth: Fixed.

halseth:
Enhanced with `quick tests`.

autopilot/choice_test.go
```diff
@@ -32,40 +34,79 @@ func TestChooseNEmptyMap(t *testing.T) {
 	}
 }
 
+// candidateMapVarLen is a type we'll use to generate maps of various lengths
+// up to 255 to be used during QuickTests.
+type candidateMapVarLen map[NodeID]*AttachmentDirective
+
+// Generate generates a value of type candidateMapVarLen to be used during
+// QuickTests.
+func (candidateMapVarLen) Generate(rand *rand.Rand, size int) reflect.Value {
```
Roasbeef: Nice, yeah I think the `Generate` router with custom types is a bit easier than specifying the value generating function each time in the main config.

halseth:
Rebased!

PR #2296, opened 2018-12-09
Implement the new max_htlc ChannelUpdate value.

### LINKED ISSUE #2182: lnwire+funding+routing: populate, implement and adhere to the new max_htlc ChannelUpdate value

# Steps To Completion 

valentinewallace:
Partially closes #2182. This PR is the first in a series of PRs to fully implement the max HTLC feature, not all of which will necessarily be merged.

Here we implement the new `max_htlc` value that has recently been added to the ChannelUpdate message via a spec amendment: lightningnetwork/lightning-rfc#465. 

In the short term, this value will allow for such benefits as smarter light client routing and enabling nodes to preserve channel bandwidth if desired. In the long term, this value can be used by AMP implementations to help with payment splitting.

The implementation of this new field consists of a few parts (spread over 6 PRs total): 
1. Setting a default `max_htlc` when sending ChannelUpdates for newly funded channels. The default chosen in this PR is the max pending amount for the channel. (#2436)
2. Allow callers of `UpdateChannelPolicy` RPC and cli commands to specify their desired `max_htlc` or a specific channel or all channels. (#2438)
3. Ensure this specified value is written to disk locally and broadcasted to the network. (this PR)
4. Validate remote channel updates to ensure that the `max_htlc` they specify is sane. (this PR)
5. Ensure that the `max_htlc` that all received channel updates specify is written to disk, broadcasted properly, and shared with remote nodes when synchronizing. (this PR)
6. Tell graph topology subscribers and `DescribeGraph` RPC callers about the `max_htlc` of channel edges and policies. (#2434)
7. Finally, use `max_htlc` in path finding and when deciding whether to forward an HTLC as an intermediate hop. (#2436 and #2437)


discovery/chan_series.go
```diff
@@ -327,7 +327,7 @@ func (c *ChanSeries) FetchChanUpdates(chain chainhash.Hash,
 			ChainHash:       chanInfo.ChainHash,
 			ShortChannelID:  shortChanID,
 			Timestamp:       uint32(e1.LastUpdate.Unix()),
-			Flags:           e1.Flags,
+			ChannelFlags:    e1.Flags,
```
halseth: Should `MsgFlags` also be set?

lnwire/channel_update.go
```diff
@@ -54,7 +54,12 @@ type ChannelUpdate struct {
 	// corresponds to the first node in the previously sent channel
 	// announcement and 1 otherwise. If the second bit is set, then the
 	// channel is set to be disabled.
-	Flags ChanUpdateFlag
+	ChannelFlags ChanUpdateFlag
+
+	// MsgFlags is a bitfield that describes additional meta-data concerning
+	// how the update is to be interpreted. Currently, the least-significant
+	// bit must be set to 1 if the optional field MaxHtlc is present.
+	MsgFlags ChanUpdateFlag
```
halseth: suggestion: name this `MessageFlags` to reflect what's in the spec?

lnwire/lnwire_test.go
```diff
@@ -603,7 +603,7 @@ func TestLightningWireProtocol(t *testing.T) {
 			req := ChannelUpdate{
 				ShortChannelID:  NewShortChanIDFromInt(uint64(r.Int63())),
 				Timestamp:       uint32(r.Int31()),
-				Flags:           ChanUpdateFlag(r.Int31()),
+				ChannelFlags:    ChanUpdateFlag(r.Int31()),
```
halseth: also set `MsgFlags`? Same question goes several places in this commit.

lnwire/channel_update.go
```diff
@@ -143,6 +147,7 @@ func (a *ChannelUpdate) Encode(w io.Writer, pver uint32) error {
 		a.MsgFlags,
 		a.TimeLockDelta,
 		a.HtlcMinimumMsat,
+		a.HtlcMaximumMsat,
```
halseth: don't think this will work, as this field is part of what was previously opaque data.

fundingmanager.go
```diff
@@ -2483,8 +2497,9 @@ func (f *fundingManager) newChanAnnouncement(localPubKey, remotePubKey,
 		ChainHash:       chainHash,
 		Timestamp:       uint32(time.Now().Unix()),
 		ChannelFlags:    chanFlags,
-		TimeLockDelta:  uint16(f.cfg.DefaultRoutingPolicy.TimeLockDelta),
+		MsgFlags:        1,
```
halseth: This should be extracted into a describing constant  :p

peer.go
```diff
@@ -1642,14 +1642,17 @@ out:
 			}
 
 			// We'll query the localChanCfg of the new channel to
-			// determine the minimum HTLC value that can be
-			// forwarded. For fees we'll use the default values, as
-			// they currently are always set to the default values
-			// at initial channel creation.
+			// We'll query the localChanCfg of the new channel to determine the
```
halseth: first line is repeated

lnrpc/rpc.proto
```diff
@@ -1936,6 +1936,9 @@ message PolicyUpdateRequest {
 
     /// The required timelock delta for HTLCs forwarded over the channel.
     uint32 time_lock_delta = 5 [json_name = "time_lock_delta"];
+
+    /// The maximum HTLC size in milli-satoshis.
+    uint64 max_htlc_msat = 6 [json_name = "max_htlc_msat"];
```
halseth: I don't think this field is meant to be changed after channel creation? 🤔

channeldb/graph.go
```diff
@@ -3186,6 +3190,9 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 	if err := binary.Write(&b, byteOrder, uint64(edge.MinHTLC)); err != nil {
 		return err
 	}
+	if err := binary.Write(&b, byteOrder, uint64(edge.MaxHTLC)); err != nil {
```
halseth: This won't work, as it will not be a backwards compatible db change. Instead, care must be taken to write it last, as part of what was previously opaque data :)

lnrpc/rpc.proto
```diff
@@ -1475,9 +1475,10 @@ message NodeAddress {
 message RoutingPolicy {
     uint32 time_lock_delta = 1 [json_name = "time_lock_delta"];
     int64 min_htlc = 2 [json_name = "min_htlc"];
-    int64 fee_base_msat = 3 [json_name = "fee_base_msat"];
-    int64 fee_rate_milli_msat = 4 [json_name = "fee_rate_milli_msat"];
-    bool disabled = 5 [json_name = "disabled"];
+    uint64 max_htlc = 3 [json_name = "max_htlc"];
```
halseth: To reduce the size of this PR, I suggest this change can be done in a follow-up! 😄 

(proto fields must be given new numbers btw).

htlcswitch/link.go
```diff
@@ -1912,6 +1912,25 @@ func (l *channelLink) HtlcSatifiesPolicy(payHash [32]byte,
 		return failure
 	}
 
+	// Next, ensure that the passed HTLC isn't too large. If so, we'll cancel
+	// the HTLC directly.
+	if policy.MaxHTLC != 0 && amtToForward > policy.MaxHTLC {
```
halseth: This is also a restriction that is questionable... Currently we  do allow forwarding on disabled channels for instance. Whether we should allow it or not is a discussion in itself, and I think it warrants a separate PR :)

routing/pathfind.go
```diff
@@ -589,6 +589,13 @@ func findPath(g *graphParams, r *restrictParams,
 			return
 		}
 
+		// If this edge was constructed from a hop hint, we won't have access to
+		// its max HTLC. Therefore, only consider discarding this edge here if
+		// the field is set.
+		if edge.MaxHTLC != 0 && edge.MaxHTLC < amountToSend {
```
halseth: 👍

discovery/chan_series.go
valentinewallace: fixed

lnwire/channel_update.go
valentinewallace: fixed

lnwire/lnwire_test.go
valentinewallace: fixed

lnwire/channel_update.go
valentinewallace: moved to the end. i hope that fixes this feedback, let me know if not

fundingmanager.go
valentinewallace: fixed

lnrpc/rpc.proto
valentinewallace: Fixed in #2434

peer.go
valentinewallace: fixed

htlcswitch/link.go
valentinewallace: Submitted as a separate PR n #2436

lnrpc/rpc.proto
valentinewallace: talked to @Roasbeef and seems to think it should still be included. I submitted this feature as a separate PR: #2438.

channeldb/graph.go
valentinewallace: thanks :) fixed

routing/pathfind.go
valentinewallace: I moved the routing to a separate PR as well: #2437

channeldb/graph.go
```diff
@@ -3386,6 +3393,11 @@ func deserializeChanEdgePolicy(r io.Reader,
 	}
 	edge.FeeProportionalMillionths = lnwire.MilliSatoshi(n)
 
+	if err := binary.Read(r, byteOrder, &n); err != nil {
```
Roasbeef: This breaks any existing deployments as they won't have this new field serialized. Instead we either need a migration, or the any new fields added are to be at the end of the struct.

Roasbeef:
We should also add a provision to this PR to set the max_htlc size for any existing channel updates which don't have this value set. This can be done [at the same time we usually check for updates that we need to retransmit](https://github.com/lightningnetwork/lnd/blob/master/discovery/gossiper.go#L1301:33).

valentinewallace:
@halseth thank you for all the feedback! the dependency graph for the max HTLC PRs is as follows:
https://i.imgur.com/ztiOt3H.png

routing/ann_validation.go
```diff
@@ -142,5 +144,31 @@ func ValidateChannelUpdateAnn(pubKey *btcec.PublicKey,
 			"update %v", spew.Sdump(a))
 	}
 
+	return validateOptionalFields(capacity, a)
```
cfromknecht: would recommend checking the optional fields before checking the signature since the latter is more expensive. checking the optional fields allows us to fail faster in the event that they're incorrect

discovery/gossiper.go
```diff
@@ -2514,15 +2521,22 @@ func (d *AuthenticatedGossiper) updateChannel(info *channeldb.ChannelEdgeInfo,
 	if timestamp <= edge.LastUpdate.Unix() {
 		timestamp = edge.LastUpdate.Unix() + 1
 	}
+
+	// Check whether optional fields are present.
+	msgFlags := 1
+	if edge.MaxHTLC == 0 {
```
cfromknecht: It looks like we will always set `MaxHTLC` before calling this, perhaps we don't need to have this conditional, and we can always set have the field enabled? Perhaps I'm missing something tho

lnwire/channel_update.go
```diff
@@ -145,6 +149,7 @@ func (a *ChannelUpdate) Encode(w io.Writer, pver uint32) error {
 		a.HtlcMinimumMsat,
 		a.BaseFee,
 		a.FeeRate,
+		a.HtlcMaximumMsat,
```
halseth: Similarly, should only be written if the flag is set.

channeldb/graph.go
```diff
@@ -3201,6 +3205,9 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 	if err := binary.Write(&b, byteOrder, uint64(edge.FeeProportionalMillionths)); err != nil {
 		return err
 	}
+	if err := binary.Write(&b, byteOrder, uint64(edge.MaxHTLC)); err != nil {
```
halseth: must be moved right before opaque data field. Also should only be written if flag is set?

channeldb/graph.go
halseth: Since the field data might already be present within the opaque data, we can either keep the current format, where we read and write the field to the var-bytes opaque data, or we can do a migration to transition it to a new format.

It is nice to avoid a migration, but reading and writing to the varbytes blob might be a bit cumbersome. I also think it might fail if the message flags indicate that the `max_htlc` field should be there, but it actually isn't, as we haven't validated the field before adding the policy to the DB.

So maybe a migration is cleanest?

halseth: What was the rationale behind storing this field as `varbytes` btw? If we just always interpreted the last unknown bytes as the extra blob, I think it would be easier to let new fields "eat into" the extra bytes.

discovery/chan_series.go
```diff
@@ -321,20 +321,21 @@ func (c *ChanSeries) FetchChanUpdates(chain chainhash.Hash,
 		return nil, err
 	}
 
-	// msgFlags indicates the presence of optional fields in a channel update.
-	// Zero indicates no optional fields.
-	msgFlags := lnwire.ChanUpdateFlag(0)
-
 	chanUpdates := make([]*lnwire.ChannelUpdate, 0, 2)
 	if e1 != nil {
+		msgFlags := 1
```
halseth: this needs a comment

discovery/gossiper.go
halseth: I think we can keep this and remove the setting before calling `updateChannel`.

discovery/gossiper.go
```diff
@@ -1344,6 +1344,13 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels() error {
 
 	var signedUpdates []lnwire.Message
 	for _, chanToUpdate := range edgesToUpdate {
+		// Ensure that the max HTLC field is set. If it isn't, set it to the
+		// capacity of the channel.
+		if chanToUpdate.edge.MaxHTLC == 0 {
```
halseth: I don't think we have to do this, as we will just send out a `channel_update` with the flag set to 0.

halseth:
Awesome work splitting up the PRs @valentinewallace! 😀

peer.go
```diff
@@ -1237,7 +1237,7 @@ func messageSummary(msg lnwire.Message) string {
 	case *lnwire.ChannelUpdate:
 		return fmt.Sprintf("chain_hash=%v, short_chan_id=%v, flag=%v, "+
 			"update_time=%v", msg.ChainHash,
-			msg.ShortChannelID.ToUint64(), msg.Flags,
+			msg.ShortChannelID.ToUint64(), msg.ChannelFlags,
```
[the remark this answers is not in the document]
valentinewallace: fixed + added String() method in 988f8dd70ea411d0524ff592f46df5f91feb71ed

routing/router.go
```diff
@@ -2067,7 +2067,7 @@ func (r *ChannelRouter) applyChannelUpdate(msg *lnwire.ChannelUpdate,
 		SigBytes:                  msg.Signature.ToSignatureBytes(),
 		ChannelID:                 msg.ShortChannelID.ToUint64(),
 		LastUpdate:                time.Unix(int64(msg.Timestamp), 0),
-		Flags:                     msg.Flags,
+		Flags:                     msg.ChannelFlags,
```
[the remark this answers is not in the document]
valentinewallace: fixed in 7f01706abb02d8eb1010c6cfc290c3cb94bd0bb6

lnwire/lnwire_test.go
```diff
@@ -608,7 +608,8 @@ func TestLightningWireProtocol(t *testing.T) {
...
+				MessageFlags:    0,
```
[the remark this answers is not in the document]
valentinewallace: true, fixed!

discovery/chan_series.go
```diff
@@ -346,7 +351,8 @@ func (c *ChanSeries) FetchChanUpdates(chain chainhash.Hash,
 			ChainHash:       chanInfo.ChainHash,
 			ShortChannelID:  shortChanID,
 			Timestamp:       uint32(e2.LastUpdate.Unix()),
-			Flags:           e2.Flags,
+			ChannelFlags:    e2.Flags,
+			MessageFlags:    msgFlags,
```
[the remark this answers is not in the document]
valentinewallace: fixed

lnwire/lnwire_test.go
```diff
@@ -609,9 +609,10 @@ func TestLightningWireProtocol(t *testing.T) {
 				ShortChannelID:  NewShortChanIDFromInt(uint64(r.Int63())),
 				Timestamp:       uint32(r.Int31()),
 				ChannelFlags:    ChanUpdateFlag(r.Int31()),
-				MessageFlags:    0,
+				MessageFlags:    1,
```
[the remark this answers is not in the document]
valentinewallace: I decided to follow this [comment](https://github.com/lightningnetwork/lnd/pull/2296#discussion_r245914562) and just randomly assign it

routing/ann_validation.go
valentinewallace: nice fixed!

lnwire/channel_update.go
```diff
@@ -109,6 +112,7 @@ func (a *ChannelUpdate) Decode(r io.Reader, pver uint32) error {
 		&a.HtlcMinimumMsat,
 		&a.BaseFee,
 		&a.FeeRate,
+		&a.HtlcMaximumMsat,
```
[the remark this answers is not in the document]
valentinewallace: Ah that's a good point, fixed

lnwire/channel_update.go
valentinewallace: fixed

channeldb/graph.go
valentinewallace: fixed

discovery/chan_series.go
valentinewallace: Removed this since we added `MessageFlags` to `ChannelEdgePolicy`, so we should just be able to use that instead.

discovery/gossiper.go
valentinewallace: Fixed. FWIW it's worth it was requested in this [comment](https://github.com/lightningnetwork/lnd/pull/2296#issuecomment-452174100)

halseth:
Will pick this one up! 😄

halseth:
Replaced by https://github.com/lightningnetwork/lnd/pull/2460

PR #2308, opened 2018-12-10
rpcserver: add channels to GetNodeInfo response

### LINKED ISSUE #2287: Feature request: get channels for node
### Background

When getting node info, the channels of a node are not revealed unless the entire describegraph is shown.

It would be convenient to have a way to only reveal a single node's channels instead of the entire graph.

### Your environment

* lnd 0.5.1
* MacOS Mojave
* Bitcoin Core 0.17.0

### Steps to reproduce

1. Ask for graph node info

### Expected behavior

Would be nice to have related channels. When looking at a channel it's interesting to see the nature of their channels, for example if you aren't interested in the total capacity of the node but rather the median capacity of their channels, you will need to query describegraph for that

### Actual behavior

Channels are not returned as part of a node query

chokoboko:
This PR adds an array of ChannelEdges to GetNodeInfo response, with json_name "channels"

fixes #2287


chokoboko:
NB: current implementation may result unexpected results, as we don't include unannounced channels in the "channels" list, but include all channels in numChannels and totalCapacity aggregators. Is this a problem? I'm trying to make minimal code and behavior changes as possible.

rpcserver.go
```diff
@@ -3592,11 +3592,25 @@ func (r *rpcServer) GetNodeInfo(ctx context.Context,
 		numChannels   uint32
 		totalCapacity btcutil.Amount
 	)
+
+	channels := make([]*lnrpc.ChannelEdge, 0)
```
halseth: style nit: `var channels []*lnrpc.ChannelEdge`

rpcserver.go
```diff
@@ -3592,11 +3592,25 @@ func (r *rpcServer) GetNodeInfo(ctx context.Context,
...
 		numChannels++
 		totalCapacity += edge.Capacity
+
+		// Do not include unannounced channels - private channels or public
+		// channels whose authentication proof were not confirmed yet.
+		if edge.AuthProof == nil {
+			return nil
+		}
+
+		// Convert the database's edge format into the network/RPC edge format
```
halseth: nit: missing period at end.

rpcserver.go
```diff
@@ -3834,11 +3834,25 @@ func (r *rpcServer) GetNodeInfo(ctx context.Context,
 		numChannels   uint32
 		totalCapacity btcutil.Amount
 	)
+
+	var channels []*lnrpc.ChannelEdge
+
```
cfromknecht: nit: extra new line

lnrpc/rpc.proto
```diff
@@ -1744,6 +1744,8 @@ message NodeInfo {
 
     uint32 num_channels = 2 [json_name = "num_channels"];
     int64 total_capacity = 3 [json_name = "total_capacity"];
+
```
cfromknecht: could use a comment here so that it is automatically updated in our documentation.

if you don't mind, could you also add docs for `num_channels` and `total_capacity` while we're here? :)

cfromknecht:
Thanks for the quick turnaround @chokoboko. IMO this is good to go, but it will conflict with #2312. To save everyone from extra rebases, I’ll ping once that’s landed so you can rebase on master and we can get a clean Travis run with the latest build fixes

halseth:
@chokoboko can now be rebased! :)

PR #2313, opened 2018-12-11
multi: implement new safe static channel backup and recovery scheme, RPCs, and cli commands

### LINKED ISSUE #175: channeldb+funding: implement streaming and regular RPCs for static backup format
A request has come in to allow online backup of the databases used by `lnd` so that the process doesn't have to be stopped for backup.

The channel DB can be backed up using https://godoc.org/github.com/boltdb/bolt#Tx.WriteTo

The wallet DB can be backed up using https://github.com/btcsuite/btcwallet/blob/master/walletdb/interface.go#L203 (`DB.Copy()`).

Roasbeef:
# Overview

In this PR, we implement a new safe scheme for static channel backups (SCB's) for  `lnd`. We say _safe_, as care has been taken to ensure that there are no foot guns in this method of backing up channels, vs doing things like `rsync`ing or copying the `channel.db` file periodically. Those methods can be dangerous as one never knows if they have the latest state of a channel or not. Instead, we aim to provide a simple safe instead to allow users to recover the settled funds in their channels in the case of partial or complete data loss. The backups themselves are encrypted using the a key derived from the user's seed, this way we protect privacy of the users channels in the back up state, and ensure that a random node can't attempt to import another user's channels. 

Once this PR is merged, given their **seed and the latest back up file**, the user will be able to recover _both_ their on-chain funds, and also funds that are fully settled within their channels. By "fully settled" we mean funds that are in the base commitment outputs, and not HTLCs. We can only restore these funds as right after the channel is created, we have all the data required to make a backup. In contrast, in order to resolve HTLCs, we would also need to update the backup state with each new channel update, which is tricky to do without additional infrastructure. This infrastructure will be built out in the near future, but until then we have this scheme which will also be a fall back in the scenario that any higher level mechanisms fail. 

At a later point, we also plan to propose this backup scheme as an addition to the spec, as even with the change to make the "to self" outputs static, we still need this SCB information in order to restore user funds. Additionally, the current serialization format is a bit up in the air. Atm, we use the same "codec" as we do within the wire protocol for the BOLT specs. However, we'll likely move to a TLV (type-length-value) format as it's extremely flexible and allows us to add/remove fields in the future once we gain new channel types, or modifications are made in the protocol that warrant a change to the backup format. Most importantly, if `aezeed` and this `chanbackup` scheme are added to the spec, then it will be possible to write a simple program, that given a seed+backup from _any_ of the implementations, will be able to recover all funds (sweep to an address) the shutdown. 

# Recovery Flow 

Skipping the backup flow for a second, given their 24-word `aezeed` seed, and a special `channels.backup` file, the recovery flow would be something like the following

1. The user uses `lncli create` or the gRPC `WalletUnlocker.Init` call to input their seed _and_ fully serialized backups. 

    1. Alternatively, if they already have a new node set up, they can use the cli and RPC commands to import channels one at a time, or the entire file. 

2. `lnd` boots up and the wallet performs a rescan from the wallet's birthday (encoded in their `aezeed`) to restore _all_ on-chain funds. Once this process is complete, the main `lnd` server will start up. 
3. Given the set of channels to recover, the server will then (using the new `chanbackup`) package, will insert a series of "channel shells" into the database. These contains _only_ the information required to initiate the DLP (data loss protection) protocol and nothing more. As a result, they're makred as "recovered" channels in the database, and we'll disallow trying to use then for any other process. 
4. Once the channel is recovered, the `chanbackup` package will attempt to insert a `LinkNode` that contains all prior addresses that we were able to reach the peer at. During the process, we'll also insert the edge for that channel (only out outgoing direction) into the database as well. 
5. `lnd` will then start up, and as usual attempt to establish connections to all peers that we have channels open with. 
6. Once we connect with a peer, we'll then initiate the DLP protocol. The remote peer will discover that we've lost data, and then immediately force close their channel. Before they do though, they'll send over their latest unrevoked commitment point which we need to derive keys (will be fixed in BOLT 1.1 by making the key static) to sweep our funds. 
7. Once the commitment transaction confirms, given information within the `SCB` we'll re-derive all keys we need, and then sweep the funds. 

# Backup + Recovery Methods 

This PR exposes multiple _safe_ ways to backup and recover a channel. We expect only one of them to be used primarily by unsophisticated end users, but have provided other mechanisms for more advanced users and business that already script `lnd` via the gRPC system. 

First, the **easiest method for backup+recovery**. After this PR, `lnd` will maintain a **`channels.backup`** file in the _same_ location that we store all the other files. Users will *at any time* be able to safely copy and backup this file. Each time a channel is opened or closed, `lnd` will update this file with the latest channel state. Users can use scripts to detect changes to the file, and upload them to their backup location. Something like `fsnotify` can notify a script each time the file changes to be backed up once again. The file is encrypted using an AEAD scheme, so it can safely be stored plainly in cloud storage, your SD card, etc. The file uses a special format and can be used to import via any of the recovery methods described below. 

The second mechanism is via the new `SubscribeChanBackups` _steaming gRPC method_. Each time an channel is opened or closed, you'll get a new notification with all the `chanbackup.Single` files (described below), and a single `chanbackup.Multi` that contains all the information for all channels. 

Finally, users are able to request a backup of a single channel, or all the channels via the cli and RPC methods.  Here's an example, of a few ways users can obtain backups, see the PR for full details: 
```
⛰ lncli --network=simnet exportchanbackup --chan_point=29be6d259dc71ebdf0a3a0e83b240eda78f9023d8aeaae13c89250c7e59467d5:0
{
    "chan_point": "29be6d259dc71ebdf0a3a0e83b240eda78f9023d8aeaae13c89250c7e59467d5:0",
    "chan_backup": "02e7b423c8cf11038354732e9696caff9d5ac9720440f70a50ca2b9fcef5d873c8e64d53bdadfe208a86c96c7f31dc4eb370a02631bb02dce6611c435753a0c1f86c9f5b99006457f0dc7ee4a1c19e0d31a1036941d65717a50136c877d66ec80bb8f3e67cee8d9a5cb3f4081c3817cd830a8d0cf851c1f1e03fee35d790e42d98df5b24e07e6d9d9a46a16352e9b44ad412571c903a532017a5bc1ffe1369c123e1e17e1e4d52cc32329aa205d73d57f846389a6e446f612eeb2dcc346e4590f59a4c533f216ee44f09c1d2298b7d6c"
}

⛰ lncli --network=simnet exportchanbackup --all
{
    "chan_points": [
        "29be6d259dc71ebdf0a3a0e83b240eda78f9023d8aeaae13c89250c7e59467d5:0"
    ],
    "multi_chan_backup": "fd73e992e5133aa085c8e45548e0189c411c8cfe42e902b0ee2dec528a18fb472c3375447868ffced0d4812125e4361d667b7e6a18b2357643e09bbe7e9110c6b28d74f4f55e7c29e92419b52509e5c367cf2d977b670a2ff7560f5fe24021d246abe30542e6c6e3aa52f903453c3a2389af918249dbdb5f1199aaecf4931c0366592165b10bdd58eaf706d6df02a39d9323a0c65260ffcc84776f2705e4942d89e4dbefa11c693027002c35582d56e295dcf74d27e90873699657337696b32c05c8014911a7ec8eb03bdbe526fe658be8abdf50ab12c4fec9ddeefc489cf817721c8e541d28fbe71e32137b5ea066a9f4e19814deedeb360def90eff2965570aab5fedd0ebfcd783ce3289360953680ac084b2e988c9cbd0912da400861467d7bb5ad4b42a95c2d541653e805cbfc84da401baf096fba43300358421ae1b43fd25f3289c8c73489977592f75bc9f73781f41718a752ab325b70c8eb2011c5d979f6efc7a76e16492566e43d94dbd42698eb06ff8ad4fd3f2baabafded"
}

⛰ lncli --network=simnet exportchanbackup --all --output_file=channels.backup

⛰ ll channels.backup
-rw-r--r--  1 roasbeef  staff   381B Dec  9 18:16 channels.backup
```

# Static Channel Backup Scheme 

## Crypto 

For encryption, we utilize `chacha20poly1305` with a random 24 byte nonce. We use a larger nonce size as this can be safely generated via a CSPRNG without fear of frequency collisions between nonces generated. To encrypt a blob, we then use this nonce as the AD (associated data) and prepend the nonce to the front of the ciphertext package.

For key generation, in order to ensure the user only needs their passphrase and the backup file, we utilize the existing keychain to derive a private key. In order to ensure that at we don't force any hardware signer to be aware of our crypto operations, we instead opt to utilize a public key that will be hashed to derive our private key. The assumption here is that this key will only be exposed to this software, and never derived as a public facing address.

## `chanbackup.Single`

The SCB contains all information required to initiate the data loss protection protocol once we restore the channel and connect to the remote channel peer.

The primary way outside callers will interact with this package are via the Pack and Unpack methods. Packing means writing a serialized+encrypted version of the SCB to an io.Writer. Unpacking does the opposite.

The encoding format itself uses the same encoding as we do on the wire within Lightning. Each encoded backup begins with a version so we can easily add or modify the serialization format in the future, if new channel types appear, or we need to add/remove fields. The backup contains:
  * The chain a channel belongs to. 
  * The `chanPoint` of the channel. 
  * The `shortChanID` of the channel. 
  * The public key of the remote node. 
  * The series of addresses that we can use to reach the node. 
  * The CSV delay of the channel (required to later reconstruct our output script after BOLT 1.1)
  * A `keychain.KeyLocator` that allows us to re-derive the payment bas epoint we need to sweep our funds .
  * A `keychain.KeyDescriptor` that we need in order to re-derive our `shachain` root to validate the information the remote party gives us during the DLP protocol. (see the next section for the complications that arose here)

## `chanbackup.Multi`

Multi is a series of static channel backups. This type of backup can contains ALL the channel
backup state in a single packed blob. This is suitable for storing on your file system, cloud storage, etc. Systems will be in place within lnd to ensure that one can easily obtain the latest version of the Multi for the node, and also that it will be kept up to date if channel state changes.

# Implementation Complications and Open Questions

The main complication that arose during the implementation was that I realized late in development, that we _also_ need to backup the details w.r.t _how_ we derive out `shachain` root. We got a bit lucky here as we store the private key we use as the root, and _not_ the public key itself. In order to derive the `shachain` roots, we use a special `keychain.KeyFamily`. However, we don't store the `keychain.KeyLocator` information which is a two-tuple that allow us to derive a key w/o knowing the public key or having any state in the wallet. Instead, within the backup, we're forced to store the entire public key and not just the key locator information. [As a result, I needed to modify `keychain.SecretKeyRing.DerivePrivKey` to support a _brute force scan_ to allow us to derive the key](https://github.com/lightningnetwork/lnd/commit/3d58d4fc7608965816fb6e9be4c8a0c039eaae68). In the future, we'll want to do a migration to also store the key locator information so we don't need to _always_ do this brute force. In order to ensure we don't scan to infinity if we don't actually know the public key, I've added a cap on the max number of iterations. 

As a result of the case above, it's now the case that any future hardware signers need to be aware of the `shachain` protocol, in order to generate and validate any points we receive. 

The one other section that we maybe want to modify is the way we derive the key we use for encryption. We made an attempt to ensure that any future hardware signers _don't_ actually need to understand our encryption protocol. So instead what we do is use a _public point_ with the assumption that it will never be used for an address and be unveiled to the outside world. One alternative that I had (but scrapped, idk why TBH) is use a point, but then have the hardware signer provide us with an ECDH of that point and another. This would ensure that the key is derived from secret data, but allow us to not store any private data in the backup. 

# TODO's 

- [ ] write integration tests

- [x] write additional unit tests in `channeldb`

- [ ] real world recovery attempts 

- [ ] update docs on how to use the recovery tools

- [x] after #1988 is in, finish hooking up the `chanbackup.SubSwapper` so we can auto update the backup file on disk 

Fixes #175


chanbackup/backupfile.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// tempFile is an open handle to the temp back up file.
+	tempFile *os.File
+}
+
+// NewMultiFile create a new multi-file instance at the target location on the
+// file system.
+func NewMultiFile(fileName string) *MultiFile {
+
+	// We'll our temporary backup file in the very same directory as the
```
alexbosworth:
```suggestion
	// We'll store our temporary backup file in the very same directory as the
```

chanbackup/backupfile.go
```diff
@@ -0,0 +1,160 @@
...
+		backupFileDir, DefaultTempBackupFileName,
+	)
+
+	return &MultiFile{
+		fileName:     fileName,
+		tempFileName: tempFileName,
+	}
+}
+
+// UpdateAndSwap will attempt write a new temporary backup file to disk with
```
alexbosworth:
```suggestion
// UpdateAndSwap will attempt to write a new temporary backup file to disk with
```

chanbackup/crypto.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// TODO(roasbeef): interface in front of?
+
+// baseEncryptionKeyLoc is the KeyLocator that we'll use to derive the base
+// encryption key used for encrypting all static channel backups. We use this
+// to then derive the actual key that we'll use for encryption via ECDH. We do
+// this rather than using the raw key, as it's possibly for us to delegate the
+// ECDH to a HSM (or the like) then use the raw keys ourselves, rather than
+// requiring the HSM to know our target cipher fore encryption.
```
alexbosworth:
```suggestion
// requiring the HSM to know our target cipher for encryption.
```

chanrestore.go
```diff
@@ -0,0 +1,128 @@
...
+		channeldb.ChanStatusRestored
+
+	chanShell.Chan.ApplyChanStatus(chanStatus)
+
+	return &chanShell, nil
+}
+
+// RestoreChansFromSingles attempts to map the set of single channel backups to
+// channel shells that will be stored persistently. Once these shells have been
+// stored on disk, we'll be able to connect to the channel peer an execute the
```
alexbosworth:
```suggestion
// stored on disk, we'll be able to connect to the channel peer and execute the
```

lsching17:
"First, the easiest method for backup+recovery. After this PR, lnd will maintain a channels.backup file in the same location that we store all the other files. .."

Can a dedicated folder be used? If it is mounted with sshfs or nfs, the channels.backup and channel.db files can be separated into different machine.

Roasbeef:
>Can a dedicated folder be used?  

I don't see why not. We can add a config flag for the backup file location.

Roasbeef:
Alrighty, I've broken this PR up into 5 distinct PR's. Each new PR depends on the prior PR. As a result, they can go in one by one and be reviewed in smaller units, rather than waiting for the final dependents of this larger PR to be finalized. I'll keep this one as is though as it has the full description, and also builds allowing users to experiment with the set of commands. Once the final PR is ready for review (as all the prior PRs have been merged), I'll rebase this on on top of that, so everyone can use this as a central point of end to end testing.

Roasbeef:
Pushed up a rebased version as all the dependent PRs have been merged. Once in #1988 is in, then I'll start the final push to getting this merged!

Roasbeef:
Pushed up a new version that maintains the backup file on disk and modifies it based on new/closed channels. Will push up the integration tests next, and after that it's ready for review.

lnrpc/rpc.proto
```diff
@@ -674,6 +694,48 @@ service Lightning {
...
+    single encrypted multi-chan backup and attempts to recover any funds
+    remaining within the channel. If we are able to unpack the backup, then the
+    new channel will be shown under listchannels, as well as pending channels.
+    */
+    rpc RestoreChannelBackups(RestoreChanBackupRequest) returns (RestoreBackupResponse)  {
+    };
+
+    /**
+    SubscribeChannelBackups allows a client to sub-subscribe to the most up to
+    date information concerning the state of all channel back ups. Each time a
```
alexbosworth:
```suggestion
    date information concerning the state of all channel backups. Each time a
```

lnrpc/rpc.proto
```diff
@@ -2114,3 +2178,64 @@ message ForwardingHistoryResponse {
    /// The index of the last time in the set of returned forwarding events. Can be used to seek further, pagination style.
    uint32 last_offset_index = 2 [json_name = "last_offset_index"];
 }
+
+message ChannelBackup {
+    /**
+    Identifies the channel that this back up belongs to.
```
alexbosworth:
```suggestion
    Identifies the channel that this backup belongs to.
```

lnrpc/rpc.proto
```diff
@@ -2114,3 +2178,64 @@ message ForwardingHistoryResponse {
...
+}
+
+message MultiChanBackup {
+    /**
+    Is the set of all channels that are included in this multi-channel backup.
+    */
+    repeated ChannelPoint chan_points = 1 [ json_name = "chan_points" ];
+
+    /**
+    A single encrypted blob containig all the static channel backups of the
```
alexbosworth:
```suggestion
    A single encrypted blob containing all the static channel backups of the
```

walletunlocker/service.go
```diff
@@ -167,6 +189,43 @@ func (u *UnlockerService) GenSeed(ctx context.Context,
...
+		backups.PackedMultiChanBackup = chanbackup.PackedMulti(
+			multiBackup.MultiChanBackup,
+		)
+	}
+
+	if chanBackups.SingleChanBackups == nil {
+		return &backups
+	}
+
+	// Finally, we can extract all the single chan back ups as well.
```
alexbosworth:
```suggestion
	// Finally, we can extract all the single chan backups as well.
```

walletunlocker/service.go
```diff
@@ -167,6 +189,43 @@ func (u *UnlockerService) GenSeed(ctx context.Context,
...
+// sub-systems.
+func extractChanBackups(chanBackups *lnrpc.ChanBackupSnapshot) *ChannelsToRecover {
+	// If there aren't any populated channel backups, then we can exit
+	// early as there's nothing to extract.
+	if chanBackups == nil || (chanBackups.SingleChanBackups == nil &&
+		chanBackups.MultiChanBackup == nil) {
+		return nil
+	}
+
+	// Now that we know there's at least a single back up populated, we'll
```
alexbosworth:
```suggestion
	// Now that we know there's at least a single backup populated, we'll
```

Roasbeef:
Alright, this PR is now final and ready for testing! I've added a new `verifychanbackup` CLI command to allow users to verify that they have a proper backup, since we use randomized encryption and backups change over time as they're updated. I need to still do a bit more testing on the CLI workflows (integration tests exercise the RPC interface), but other than that it's good to go on my end.

lnd_test.go
```diff
@@ -13011,6 +13013,194 @@ func testSweepAllCoins(net *lntest.NetworkHarness, t *harnessTest) {
...
+			ctxt, t, net, net.Alice, carol,
+			lntest.OpenChannelParams{
+				Amt: chanAmt,
+			},
+		)
+
+		chanPoints = append(chanPoints, chanPoint)
+	}
+
+	// Using this helper function, we'll maintain a pointe rot he latest
```
halseth: nit: typo

lnd_test.go
```diff
@@ -13011,6 +13013,194 @@ func testSweepAllCoins(net *lntest.NetworkHarness, t *harnessTest) {
...
+				return fmt.Errorf("unable to read backup "+
+					"file: %v", err)
+			}
+
+			// As each back up file will be encrypted with a fresh
+			// nonce, we can't compare them directly, so instead
+			// we'll compare the length which is a proxy for the
+			// number of channels that the multi-backup contains.
+			rawBackup := currentBackup.MultiChanBackup.MultiChanBackup
+			if len(rawBackup) != len(packedBackup) {
```
halseth: can we compare the bytes directly?

lnd_test.go
```diff
@@ -13201,6 +13201,159 @@ func testChannelBackupUpdates(net *lntest.NetworkHarness, t *harnessTest) {
...
+				chanBackup.ChanPoint.String())
+		}
+	}
+
+	// Before we proceed, we'll make two utility methods we'll use below
+	// for our primary assertions.
+	assertNumSingleBackups := func(numSingles int) {
+		ctxt, _ := context.WithTimeout(ctxb, defaultTimeout)
+		req := &lnrpc.ChanBackupExportRequest{}
+		chanSnapshot, err := carol.ExportAllChannelBackups(ctxt, req)
```
halseth: Feels like this can be incorporated with the previous test case, to avoid the overhead of running a new test

lnd_test.go
```diff
@@ -13590,6 +13992,9 @@ var testsCases = []*testCase{
 		test: testExportChannelBackup,
 	},
 	{
+		name: "channel backup restore",
+		test: testChannelBackupRestore,
```
halseth: we could make these test part of a new file `backup_test.go`, and assign a separate travis build for it.

lnd_test.go
```diff
@@ -13354,6 +13354,408 @@ func testExportChannelBackup(net *lntest.NetworkHarness, t *harnessTest) {
...
+	if err != nil {
+		t.Fatalf("unable to create new node: %v", err)
+	}
+	carol, err := net.NewNode("carol", nil)
+	if err != nil {
+		t.Fatalf("unable to make new node: %v", err)
+	}
+
+	defer func() {
+		shutdownAndAssert(net, t, dave)
```
halseth: this should be called right after creating `dave` , before creating `carol`.

lnd_test.go
```diff
@@ -13354,6 +13354,408 @@ func testExportChannelBackup(net *lntest.NetworkHarness, t *harnessTest) {
...
+		},
+	}
+
+	// TODO(roasbeef): online vs offline close?
+
+	// TODO(roasbeef): need to re-trigger the on-disk file once the node
+	// ann is updated?
+
+	for _, testCase := range testCases {
+		if testCase.restoreMethod == nil {
```
halseth: unnecessary?

lnwallet/channel.go
```diff
@@ -3496,9 +3496,9 @@ func (lc *LightningChannel) ProcessChanSyncMsg(
 	// is valid.
 	var commitPoint *btcec.PublicKey
 	switch {
-	case msg.NextLocalCommitHeight == remoteTailHeight+2:
-		commitPoint = lc.channelState.RemoteNextRevocation
-
+	// If their height is one beyond what we know their current height to
```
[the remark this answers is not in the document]
Roasbeef: Ended up moving it below, it's in the main diff.

lnrpc/rpc.proto
```diff
@@ -674,6 +694,48 @@ service Lightning {
...
+
+    /** lncli: `exportchanbackup`
+    ExportChannelBackup attempts to return an encrypted static channel backup
+    for the target channel identified by it channel point. The backup is
+    encrypted with a key generated from the aezeed seed of the user. The
+    returned backup can either be restored using the RestoreChannelBackup
+    method once lnd is running, or via the InitWallet and UnlockWallet methods
+    from the WalletUnlocker service.
+    */
+    rpc ExportChannelBackup(ChannelPoint) returns (ChannelBackup) {
```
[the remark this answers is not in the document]
Roasbeef: Good idea, fixed!

rpcserver.go
```diff
@@ -4600,3 +4594,218 @@ func (r *rpcServer) ForwardingHistory(ctx context.Context,
...
+	}
+	err = unpackedMultiBackup.PackToWriter(&b, r.server.cc.keyRing)
+	if err != nil {
+		return nil, fmt.Errorf("unable to multi-pack backups: %v", err)
+	}
+
+	multiBackupResp := &lnrpc.MultiChanBackup{
+		MultiChanBackup: b.Bytes(),
+	}
+	for _, singleBackup := range singleBackupResp.ChanBackups {
```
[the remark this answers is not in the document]
Roasbeef: I don't think it makes a difference either way. I have an integration test for this, and the current iteration ensures that we don't accidentally collate any chan points which didn't actually have e a backup produced for them.

channel_notifier.go
```diff
@@ -0,0 +1,145 @@
...
+	// AddrsForNode returns all known addresses for the target node public
+	// key.
+	AddrsForNode(nodePub *btcec.PublicKey) ([]net.Addr, error)
+}
+
+// channelNotifier is an implementation of the chanbackup.ChannelNotifier
+// interface using the existing channelnotifier.ChannelNotifier struct. This
+// implementation allows us to satisfy all the dependencies of the
+// chanbackup.SubSwapper struct.
+type channelNotifier struct {
```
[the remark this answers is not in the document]
Roasbeef: Typically we place concrete implementation of package interfaces into the main package. This allows us to keep the target that defines the interfaces clean, as it doesn't need to import things like the channel notifier. 

Going for something like: 
> The dependency inversion principle encourages you to push the responsibility for the specifics, as high as possible up the import graph, to your main package or top level handler, leaving the lower level code to deal with abstractions–interfaces.


From: https://dave.cheney.net/2016/08/20/solid-go-design

lnwallet/channel.go
```diff
@@ -3572,6 +3579,13 @@ func ChanSyncMsg(c *channeldb.OpenChannel) (*lnwire.ChannelReestablish, error) {
 		return nil, err
 	}
 
+	// If we've restored this channel, then we'll purposefully give them an
```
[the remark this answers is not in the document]
Roasbeef: We need this if we've lost state before we've actually done any updates. Otherwise, everything checks out to the remote party, and they don't force close at all. In any case the remote party should force close as we sent it funky data. We already do this, and I think it's also in the spec to force close if you can't verify their commitment secret.

contractcourt/chain_watcher.go
```diff
@@ -338,12 +338,14 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
-		// having an input sequence number that's finalized.
-		// This won't happen with regular commitment
-		// transactions due to the state hint encoding scheme.
+		// Next, we'll check to see if this is a cooperative channel
+		// closure or not. This is characterized by having an input
+		// sequence number that's finalized.  This won't happen with
+		// regular commitment transactions due to the state hint
+		// encoding scheme.
 		if commitTxBroadcast.TxIn[0].Sequence == wire.MaxTxInSequenceNum {
+			// TODO(roasbeef): rare but possible, need itest case
```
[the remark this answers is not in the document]
Roasbeef: A TODO with out context I guess, I meant that it's possible that the remote party broadcasts with a _cooperative close_ transaction they stashed somewhere _after_ we've lost state and need to restore.

contractcourt/chain_watcher.go
```diff
@@ -425,8 +424,12 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
 		// commitment (we write them to disk before sending out), then
 		// this means that we've lost data. In this case, we'll enter
-		// the DLP protocol.
-		case broadcastStateNum > remoteStateNum:
+		// the DLP protocol. Otherwise, if we've recovered our channel
+		// state from scratch, then we don't know what the precise
+		// current state is, so we assume either the remote party
+		// forced closed or we've been "breached. In the latter case,
+		// our tower will take care of us.
+		case broadcastStateNum > remoteStateNum || isRecoveredChan:
```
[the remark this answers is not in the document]
Roasbeef: Don't think it makes a big diff either way.

rpcserver.go
```diff
@@ -4651,6 +4659,65 @@ func (r *rpcServer) ExportChannelBackup(ctx context.Context,
...
+	}
+
+	// By default, we'll assume that both backups are valid.
+	resp := lnrpc.VerifyChanBackupResponse{
+		SinglesValid: true,
+		MultiValid:   true,
+	}
+
+	if in.GetSingleChanBackups() != nil {
+		// First, we'll convert the raw byte sliice into a type we can
```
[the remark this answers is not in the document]
Roasbeef: sliice sliice baby

lnd_test.go
Roasbeef: No, since we use a random nonce each time to create a backup. I created a the `VerifyChanBackup` command to help with this, as it allows users to ensure they have a valid backup, since it can change on disk.

lnd_test.go
Roasbeef: There's a single test here? Commented on an outdated diff perhaps?

lnd_test.go
Roasbeef: Yeah when we want to enable the remaining 24 test cases, that sounds like a good idea. But unless they run on some sort of "nightly" build, then they'll still add to the execution time of the integration tests.

Roasbeef:
PTAL @halseth

chanbackup/pubsub.go
```diff
@@ -49,7 +50,7 @@ type ChannelEvent struct {
 type ChannelSubscription struct {
 	// ChanUpdates is a read-only channel that will be sent upon once the
 	// primary channel state is updated.
-	ChanUpdates <-chan ChannelEvent
+	ChanUpdates chan ChannelEvent
```
cfromknecht: why is it now send/receive? comment should be updated if it's no longer read-only

chanrestore.go
```diff
@@ -0,0 +1,187 @@
...
+			RevocationProducer:      shaChainProducer,
+		},
+	}
+
+	// TODO(roasbeef): move this mapping elsewhere?
+
+	// When we make a channel, we mark that the channel has been restored,
+	// this will signal to other sub-systems to not attempt to use the
+	// channel as if it was a regular one.
+	chanStatus := channeldb.ChanStatusDefault |
```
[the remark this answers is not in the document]
Roasbeef: A later commit modified things to handle this within `channeldb` itself, as that method requires the channel state to already have been created.

peer.go
```diff
@@ -438,7 +438,12 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
 
 		// Skip adding any permanently irreconcilable channels to the
 		// htlcswitch.
-		if dbChan.ChanStatus() != channeldb.ChanStatusDefault {
+		switch {
```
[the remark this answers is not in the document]
Roasbeef: Only the first bit is always default or none. For the rest, they're treated as a bit vector, so we still need to do a bitmask to compare. Additionally, it's nice that this is explicit, rather than a blanket check since it'll force us to consider any future combinations.

rpcserver.go
```diff
@@ -353,7 +350,19 @@ var (
 			Entity: "offchain",
 			Action: "write",
 		}},
-		"/lnrpc.Lightning/ForwardingHistory": {{
```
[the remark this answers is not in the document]
Roasbeef: Should be reverted in a later commit, I think you're looking at a dated diff.

lnwire/lnwire.go
```diff
@@ -416,6 +416,14 @@ func WriteElement(w io.Writer, element interface{}) error {
 			return err
 		}
 
+	case bool:
```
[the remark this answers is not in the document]
Roasbeef: To align things with the wire protocol, as I'd like to eventually write this up to become part of the BOLT spec. This way, we don't need to special case things that differ between `channeldb` and `lnwire` like the way we serialize outpoints.

contractcourt/chain_watcher.go
```diff
@@ -320,22 +320,35 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
-		// exit here as we don't have any further processing we
-		// need to do (we can't cheat ourselves :p).
-		commitmentHash := localCommit.CommitTx.TxHash()
-		isOurCommitment := commitSpend.SpenderTxHash.IsEqual(
-			&commitmentHash,
+		// If this channel has been recovered, then we'll modify our
+		// behavior as it isn't possible for us to close out the
+		// channel off-chain ourselves. It can only be the remote party
+		// force closing, or a cooperative closure we signed off on
+		// before losing data getting confirmed in the chain.
```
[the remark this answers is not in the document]
Roasbeef: Hmm, yeah I guess that's a possibility. However the comparison below is based on knowing the txid of the commitment transaction that force closed the commitment. We may know it for the very first state, but wouldn't for any other subsequent states. If they _did_ force close before restoring, then it's also possible that node is still active. Additionally, we won't have any of the pre-signed HTLC transactions, so we won't be able to sweep those. Therefore in order to identify this case generically, we'd likely need to brute force all possible output scripts for our local commitment and look for that.

rpcserver.go
```diff
@@ -138,10 +139,6 @@ var (
 			Entity: "invoices",
 			Action: "write",
 		},
-		{
```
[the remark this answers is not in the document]
Roasbeef: rebase error reverted

chanbackup/pubsub.go
Roasbeef: Made it easier to write some unit tests

Roasbeef:
@halseth @cfromknecht PTAL!

lnwallet/channel.go
```diff
@@ -3501,7 +3501,16 @@ func (lc *LightningChannel) ProcessChanSyncMsg(
 	// as that's what they should send.
 	case msg.NextLocalCommitHeight == remoteTailHeight+1:
 		commitPoint = lc.channelState.RemoteCurrentRevocation
+
+	// Alternatively, if their height is two beyond what we know their best
+	// height to be, then they're holding onto two commitments, and the
+	// highest unrevoked point it their next revocation.
+	//
+	// TODO(roasbeef): verify this in the spec...
+	case msg.NextLocalCommitHeight == remoteTailHeight+2:
```
halseth: this needs a unit test...

contractcourt/chain_watcher.go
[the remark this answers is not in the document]
halseth: True, but it doesn't say anything about this being during restore

rpcserver.go
```diff
@@ -4670,17 +4668,17 @@ func (r *rpcServer) ForwardingHistory(ctx context.Context,
 // once lnd is running, or via the InitWallet and UnlockWallet methods from the
 // WalletUnlocker service.
 func (r *rpcServer) ExportChannelBackup(ctx context.Context,
-	in *lnrpc.ChannelPoint) (*lnrpc.ChannelBackup, error) {
+	in *lnrpc.ExportChannelBackupRequest) (*lnrpc.ChannelBackup, error) {
```
halseth: squash with original commit

server.go
```diff
@@ -1035,206 +1038,233 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
-	// Already running?
-	if !atomic.CompareAndSwapInt32(&s.started, 0, 1) {
-		return nil
-	}
-
-	if s.torController != nil {
-		if err := s.initTorController(); err != nil {
-			return err
+	var startErr error
+	s.start.Do(func() {
```
halseth: Hm, not sure if I prefer this new pattern... (indentation and complicates error handling)

What about 
```go
if !atomic.CompareAndSwapInt32(&s.started, 0, 1) {
     return nil
}
defer atomic.StoreInt32(&s.active, 1)
```

lntest/node.go
```diff
@@ -464,16 +464,14 @@ func (hn *HarnessNode) Init(ctx context.Context,
 // should be called after the restart of a HarnessNode that was created with a
 // seed+password. Once this method returns, the HarnessNode will be ready to
 // accept normal gRPC requests and harness command.
-func (hn *HarnessNode) Unlock(ctx context.Context, password []byte) error {
+func (hn *HarnessNode) Unlock(ctx context.Context,
```
halseth: sqaush with commit that introduces the password. 

This won't compile.

lnd_test.go
halseth: I meant that `testExportChannelBackup` and `testChannelBackupUpdates` could be combined into one. In the test you could just first check that you get the backup updates, then you can manually export the channel backup, and make sure everything adds up.

channel_notifier.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// SubscribeChans requests a new channel subscription relative to the initial
+// set of known channels. We use the knownChans as a synchronization point to
+// ensure that the chanbackup.SubSwapper does not miss any channel open or
+// close events in the period between when it's created, and when it requests
+// the channel subscription.
+//
+// NOTE: This is part of the chanbackup.ChannelNotifier interface.
+func (c *channelNotifier) SubscribeChans(startingChans map[wire.OutPoint]struct{}) (
```
halseth: `startingChans` now used?

chanrestore.go
```diff
@@ -0,0 +1,187 @@
...
+		channeldb.ChanStatusRestored
+
+	chanShell.Chan.ApplyChanStatus(chanStatus)
```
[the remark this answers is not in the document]
Roasbeef: This isn't in the latest diff.

chanrestore.go
```diff
@@ -0,0 +1,187 @@
...
+	for _, addr := range addrs {
+		netAddr := &lnwire.NetAddress{
+			IdentityKey: nodePub,
+			Address:     addr,
+		}
+
+		ltndLog.Infof("Attempting to connect to %v for SCB restore "+
+			"DLP", netAddr)
+
+		if err := s.ConnectToPeer(netAddr, true); err != nil {
```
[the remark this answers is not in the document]
Roasbeef: Nice suggestion! Modified to continue and try the next address if there's an error, ultimately returning and error if nothing connects.

 One question with the logic that sits above this is: if we restore one channel, but then fail to connect to the peer, should that halt restoring all other channels? Atm, the answer is yes, but it wouldn't be too involved to continue to restore the other channels, and then return the set of peers that we're unable to connect to.

channel_notifier.go
```diff
@@ -0,0 +1,145 @@
...
+						NewChans: []chanbackup.ChannelWithAddrs{
+							{
+								OpenChannel: channel,
+								Addrs:       nodeAddrs,
+							},
+						},
+					}
+
+					select {
+					case chanUpdates <- chanEvent:
```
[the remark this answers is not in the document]
Roasbeef: The `chanbackup.SubSwapper` already updates the on-disk backups with the current available channel set on start up.

log.go
```diff
@@ -140,6 +143,7 @@ var subsystemLoggers = map[string]btclog.Logger{
 	"NTFR": ntfnLog,
 	"IRPC": irpcLog,
 	"CHNF": chnfLog,
+	"CHBU": chbuLog,
```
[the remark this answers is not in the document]
Roasbeef: Would rather have a new change that switches everything over (if we really like this new style). IMO it's a bit less clear what the names of all the loggers are at a glance.

lntest/node.go
```diff
@@ -439,13 +440,30 @@ func (hn *HarnessNode) Init(ctx context.Context,
...
-	}, 5*time.Second); err != nil {
 	return hn.initClientWhenReady()
 }
+
+// Unlock attempts to unlock the wallet of the target HarnessNode. This method
+// should be called after the restart of a HarnessNode that was created with a
+// seed+password. Once this method returns, the HarnessNode will be ready to
+// accept normal gRPC requests and harness command.
+func (hn *HarnessNode) Unlock(ctx context.Context, password []byte) error {
+
```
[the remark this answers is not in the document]
Roasbeef: I prefer to add a new line if we need to wrap the function args as this is modified to do in a follow up commit.

lnwallet/channel.go
Roasbeef: It's really implementation specific. For example, `lnd` never holds two commitments. Even within the spec, it's unclear _which_ point to expect/use in this scenario. I'd rather add the test in a follow change (or remove it from this diff) just to keep this thing moving forwards. K

server.go
Roasbeef: We're moving to using `sync.Once` elsewhere, and if the indentation gets too out of hand, we can also lift it into a new method rather than using a function closure.

channel_notifier.go
Roasbeef: Only there to meet the interface, but it isn't being used atm as we send ourselves the fresh set of channels at start up now.

lnd_test.go
Roasbeef: Yeah, but also nice to have some integration tests not be so "mega"...

molxyz:
Tested on a testnet node that has been running with `noseedbackup`, SCB still let me do `exportchanbackup`. Shouldn't this result in an error message instead?
https://hastebin.com/raw/urohupeful

Roasbeef:
@molxyz at runtime, `lnd` doesn't know if you actually got a seed or not.

Roasbeef:
In that case, you wouldn't actually be able to decrypt the SCB unless you read out the private data of the database.

ZapUser77:
Any chance you can include what the commands are to restore (exact syntax), and what the expected outputs would be (just and example)?  Considering how important this is, just guess and 'tying to figure it out' may not be the best idea.

From my understanding, this isn't actually a "back up" of the channels, and is instead a "channel funds recovery mechanism".  Correct?  If you restored using this, you'd have a node with zero channels, and would have to start open channels from scratch. Correct?

Roasbeef:
Check out the PR description, more docs will be provided later.

ZapUser77:
"Check out the PR description"
I did, read the entire thing, multiple times.  I wouldn't have asked before reading.

"more docs will be provided later."
Thanks as always for your diligent hard work.  It really is appreciated.

gijswijs:
Would it be possible to create a "backup" manually, in the scenario where a node is lost and accessing the original channels.backup file isn't accessible anymore?

Put in another way, given that you know the `chan_point`, could you insert the "channel shell" into the database, so that the DLP protocol can be initiated?

rootzoll:
@gijswijs I think thats an LND question and should be best answered at the LND Slack/GitHub

PR #2315, opened 2018-12-11
lnrpc: remove reference to GOPATH

joostjager:
As go modules allow us to build outside of the GOPATH, we shouldn't look for .proto imports there anymore.

Before this change, it wasn't possible to run `gen_protos.sh` without errors.

To reproduce the problem, make sure there is no `lnd` source in `GOPATH`. Otherwise it seems to work.


lnrpc/signrpc/signer.proto
```diff
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package signrpc;
 
+option go_package = "github.com/lightningnetwork/lnd/lnrpc/signrpc";
```
halseth: what's this?

joostjager: To make it work outside the gopath with modules. The dir structure doesn't need to match the package structure anymore there.

lnrpc/gen_protos.sh
```diff
@@ -30,8 +30,7 @@ do
     DIRECTORY=$(dirname ${file})
     echo "Generating protos from ${file}, into ${DIRECTORY}"
 
-    protoc -I/usr/local/include -I. \
-           -I$GOPATH/src \
-           --go_out=plugins=grpc:. \
+    protoc -I/usr/local/include -I.\
+           --go_out=plugins=grpc,paths=source_relative:. \
```
wpaulino: Should also update the other uses of `protoc` with these changes.

joostjager: I am not sure, because it still needs to find that external dep somewhere. `protoc` doesn't care about go modules I think.

joostjager:
I used this code again, but it doesn't seem to work. It generates this file for me `/home/joost/lightninglabs/lnd/lnrpc/github.com/lightningnetwork/lnd/lnrpc/signrpc/signer.pb.go`

lnrpc/gen_protos.sh
joostjager: The other references to `GOPATH` are dependencies of `protoc`, not `lnd`. Those need to be installed somewhere. I think it is okay that those are installed in `GOPATH` as is also instructed in the readme.

joostjager: I think it worked for you because you also had `lnd` still in `GOPATH`? Only if you fully `rm -rf` lnd sources from there, you can see the difference.

joostjager:
Pushed update to proto deps to get `source_relative` support.

PR #2320, opened 2018-12-12
[watchtower] add wtpolicy.Policy and wtserver message logging

cfromknecht:
This PR makes some preliminary changes in advance of the watchtower client and watchtower capstone integrations.

The primary additions are the `wtpolicy.Policy`, which bundles up the negotiable watchtower parameters with the aim of later exposing them via command line. Since the same policy is shared between client and server, the struct is placed in it's own `wtpolicy` package. We also define `blob.Type` which allows for further negotiation at the session level (see #2439 for more info).

In addition, this PR adds log messages for wtserver messages that are sent and received, doing so with a `MessageSummary` method in `wtwire`. This code can then be shared across client and server, to ensure both sides see mirrored log messages. This final top-level PR is predominately this change.

Finally, we fix some broken log messages in the `lookout` package, which would apply hex formatting to the real value twice. We now serialize those values using the txid's string representation (see #2433).

Depends on:
 - [x] #2433 
 - [x] #2439

### DIFF
--- a/watchtower/wtserver/server.go
+++ b/watchtower/wtserver/server.go
@@ -191,6 +191,12 @@ func (s *Server) handleClient(peer Peer) {
 	defer s.removePeer(&id, peer.RemoteAddr())
 
 	msg, err := s.readMessage(peer)
+	if err != nil {
+		log.Errorf("Unable to read message from client %s@%s: %v",
+			id, peer.RemoteAddr(), err)
+		return
+	}
+
 	remoteInit, ok := msg.(*wtwire.Init)
 	if !ok {
 		log.Errorf("Client %s@%s did not send Init msg as first "+
--- a/watchtower/wtserver/server.go
+++ b/watchtower/wtserver/server.go
@@ -188,7 +188,7 @@ func (s *Server) handleClient(peer Peer) {
 		peer.Close()
 		return
 	}
-	defer s.removePeer(&id)
+	defer s.removePeer(&id, peer.RemoteAddr())
 
 	msg, err := s.readMessage(peer)
 	remoteInit, ok := msg.(*wtwire.Init)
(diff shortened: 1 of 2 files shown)


watchtower/wtwire/summary.go
```diff
@@ -0,0 +1,35 @@
+package wtwire
+
+import "fmt"
+
+// MessageSummary creates a human-readable description of a given Message. If
+// the type is unknown, an empty string is returned.
+func MessageSummary(msg Message) string {
```
Roasbeef: Can alternatively use the `MsgType()` method here. Although that would mean you still need a type assertion somewhere below.

watchtower/wtserver/server.go
```diff
@@ -191,6 +191,12 @@ func (s *Server) handleClient(peer Peer) {
 	defer s.removePeer(&id, peer.RemoteAddr())
 
 	msg, err := s.readMessage(peer)
+	if err != nil {
```
Roasbeef: 👍

watchtower/wtwire/summary.go
halseth: Using `MsgType()` would also work with non-pointer messages, while this won't?

watchtower/wtserver/server.go
```diff
@@ -632,6 +625,27 @@ func (s *Server) removePeer(id *wtdb.SessionID) {
...
+		action = "Sending"
+		preposition = "to"
+	}
+
+	summary := wtwire.MessageSummary(msg)
+	if len(summary) > 0 {
+		summary = "(" + summary + ")"
+	}
+
+	log.Debugf("%s %s%v %s %x@%s", action, msg.MsgType(), summary,
```
halseth: elegant

watchtower/wtwire/summary.go
halseth: Why? Non-pointers can also implement interfaces.

cfromknecht: Indeed they can

> Using MsgType() would also work with non-pointer messages, while this won't?

The current version will work with non-pointer messages as well

halseth: Realized that the non-pointer type actually doesn't implement the `Message` interface, so this is good :)

PR #2329, opened 2018-12-15
multi: replace per channel sigPool with global daemon level sigPool

Roasbeef:
In this commit, we remove the per channel `sigPool` within the
`lnwallet.LightningChannel` struct. With this change, we ensure that as
the number of channels grows, the number of gouroutines idling in the
sigPool stays constant. It's the case that currently on the daemon, most
channels are likely inactive, with only a hand full actually
consistently carrying out channel updates. As a result, this change
should reduce the amount of idle CPU usage, as we have less active
goroutines in select loops.

In order to make this change, the `SigPool` itself has been publicly
exported such that outside callers can make a `SigPool` and pass it into
newly created channels. Since the sig pool now lives outside the
channel, we were also able to do away with the Stop() method on the
channel all together.

Finally, the server is the sub-system that is currently responsible for
managing the `SigPool` within lnd.


lnwallet/channel.go
```diff
@@ -1467,18 +1459,6 @@ func (lc *LightningChannel) createStateHintObfuscator() {
 	}
 }
 
-// Stop gracefully shuts down any active goroutines spawned by the
-// LightningChannel during regular duties.
-func (lc *LightningChannel) Stop() {
```
wpaulino: Nit: looks like we can also remove the now unused `shutdown` and `quit` members.

cfromknecht:
breacharbiter_test doesn't compile atm
```
# github.com/lightningnetwork/lnd [github.com/lightningnetwork/lnd.test]
./breacharbiter_test.go:1573:43: undefined: bobPool
```

Roasbeef:
Fixed, PTAL!

PR #2331, opened 2018-12-16
Add sample endpoint to neutrino.connect

keeshux:
Should help having everything in place without looking for further Neutrino documentation.


halseth:
Not sure if this is the best idea, for two reasons:
1) When using `connect=`, neutrino will only stay connected to that node, and not attempt to find more peers.
2) We risk everybody connecting to this one node by default, which is centralizing.

PR #2259, opened 2018-12-02
update lnd dockerfile image

yancyribbens:
running `docker-compose build` currently fails from `/lnd/docker` because `lnd/Dockerfile` is outdated.  This PR updates the dependencies.

PR #2333, opened 2018-12-17
fixed dockerfile lnd inside docker folder

guilhermehas:


halseth:
What does this fix? Should have a description.

Also, might conflict with #2259

wpaulino:
Looks like it can actually replace #2259. This PR will allow us to no longer assume the source code is present within the current directory.

OluAgunloye:
Works, in conjunction w/ #2334 , on Ubuntu 16.04

tuxx42:
i'd suggest the method proposed in my PR #2334 instead.
However, in #2334 one needs to build the containers from the root of `lnd` as such, but its better than cloning the git repo :
`host@~/lnd/$ sudo docker-compose -f docker/docker-compose.yml build lnd`

but i think this is more useful than cloning the repo, because this builds against whatever is currently checked out in the root of the `lnd` directory (e.g. a feature branch)

guilhermehas:
After writing this command in the guide
`docker-compose run -d --name alice lnd_btc`
I got this error:
`ERROR: Service 'lnd_btc' failed to build: The command '/bin/sh -c apk add --no-cache     git     make &&  cd /go/src/github.com/lightningnetwork/lnd &&  make &&  make install' returned a non-zero code: 2 (edited) `

docker/lnd/Dockerfile
```diff
@@ -1,18 +1,15 @@
-FROM golang:1.10-alpine as builder
-
-MAINTAINER Olaoluwa Osuntokun <lightning.engineering>
```
yancyribbens: Is there a reason you removed the Maintainer contact?

docker/lnd/Dockerfile
```diff
@@ -1,18 +1,15 @@
...
-
-# Copy in the local repository to build from.
-COPY . /go/src/github.com/lightningnetwork/lnd
```
yancyribbens: Is there a reason to clone from the repo on L12 (which requires http connection) instead of using COPY?

yancyribbens:
> Looks like it can actually replace #2259. This PR will allow us to no longer assume the source code is present within the current directory.

@wpaulino yes, but now to build you add a dependacy of needing http and more importantly GitHub.  I think this actually makes the build more brittle.

yancyribbens:
@wpaulino yeah Github is used other places to pull files into the container.  Either options work, I just think it adds extra complexity since the state of github may be different than the state locally, and also github could go down which does happen from time to time.  Also, build time is slightly faster if not relying on an external service to pull int files that could otherwise be provided locally.  The net result is the same as a way to copy files into a docker container.

Roasbeef:
Superseded by #2259

PR #2341, opened 2018-12-18
[autopilot] Decouple agent constraints from heuristics

halseth:
This is a preparation for the `external scoring based autopilot` #2212 and `dry run autopilot` #2350. As it turned out to be quite massive, I'm breaking it up into three separate PRs.

This PR cleans up some of the affected `autopilot` interfaces, mainly by moving logic from the `heuristics` to the `agent`. This is done to make it easier to share this code between several heuristics when they are enabled.

~~We also make the `agent` now take a weighted list of heuristics. For now this list just contains a single item with a weight of `1.0`, but this will change in the follow-up PRs.~~ Broken into new PR: #2405


autopilot/agent.go
```diff
@@ -24,14 +24,15 @@ type Config struct {
...
-	Heuristic AttachmentHeuristic
+	// Heuristics is a slice of attachment heuristics which will govern to
+	// whom we open channels to. The Heuristics will take into account the
+	// current state of the graph, our set of open channels, and desired
+	// channel size when determining how channels are to be opened.
+	// Additionally, a heuristic might also factor in extra-graph
+	// information in order to make more pertinent recommendations. The
+	// weight given to each heuristic is a given each heuristic's weight,
+	// which must sum to 1.0 for the slice.
+	Heuristics []*WeightedHeuristic
```
Roasbeef: Alternatively, in order to keep the logic at the agent simple, we can create a new composite node score heuristic. So from the PoV of the agent, it queries a _single_ heuristic which then implements the weighted heuristic logic underneath. In a sense, it's just an additional layer of indirection that keeps the agent simple and the bulk of the logic within the composite node score heuristic.

Roasbeef: By abstracting the selection logic, we can eventually experiment with other things like the original "information gain model", the self-aware model (takes into account how its own score will be modified by accepting/making a channel), and so on.

halseth: I went the composite heuristic way first, but it turned out to be more complex than necessary when I wanted to query each individual sub-heuristic for the dry run.

But I do agree that it would be nice to keep the agent as simple as possible, and it would possibly also make it easier to experiment with more advanced combinations than just a weighted average.

Will see if I can combine the two approaches, still keeping the dryrun easy to implement.

halseth: Done!

autopilot/agent_constraints.go
```diff
@@ -0,0 +1,151 @@
...
+
+// AvailableChans should, given the passed parameters, and its internal state,
+// return whether more channels should be opened within the channel graph. If
+// the constraints allow us to open more channels, then the first return value
+// will represent the amount of additional funds available towards creating
+// channels. The second return value is the exact *number* of additional
+// channels that available.
+//
+// Note: part of the AgentConstraints interface.
+func (h *agentConstraints) AvailableChans(channels []Channel,
```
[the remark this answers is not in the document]
halseth: Went for `ChannelBudget`.

autopilot/agent_constraints.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// MaxChanSize returns largest channel that the autopilot agent should
+	// create.
+	MaxChanSize() btcutil.Amount
+}
+
+// agenConstraints is an implementation of the AgentConstraints interface that
+// indicate the constraints the autopilot agent must adhere to when opening
+// channels.
+type agentConstraints struct {
```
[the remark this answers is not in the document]
halseth: keeping name for now, can change if we add other types of constraints.

autopilot/agent_constraints.go
```diff
@@ -0,0 +1,151 @@
...
+		maxPendingOpens: maxPendingOpens,
+	}
+}
+
+// AvailableChans should, given the passed parameters, and its internal state,
+// return whether more channels should be opened within the channel graph. If
+// the constraints allow us to open more channels, then the first return value
+// will represent the amount of additional funds available towards creating
+// channels. The second return value is the exact *number* of additional
+// channels that available.
```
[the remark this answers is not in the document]
halseth: Fixed.

PR #2352, opened 2018-12-19
lnrpc: set JS_STRING option for chan_id

sangaman:
This sets the `jstype` option to `JS_STRING` for all `chan_id` fields in the proto rpc definition. `chan_id` is a 64 bit integer, which is not natively supported by javascript's floating-point `number` with only 52 bit precision. Nevertheless, by default protobuf will use the `number` type for 64 bit integer fields in javascript, which can cause loss of precision problems with `chan_id`. Explicitly setting the type for javascript as a string will prevent these issues, and should not interfere with its use as an identifier.

I've run into this issue in a javascript project, and making these changes to the proto file manually was the solution. I figure this change would save the headache for others who may run into this as well down the road.


sangaman:
I reopened this branch/PR and rebased onto master (including adding the jstype attribute in a few more places where channel id fields are used). This definitely seems helpful to have in the main repo as it would prevent bugs in javascript code that uses an unmodified version of the proto definitions.

halseth:
cc @tanx any thoughts on this?

erkarl:
Bump.

tanx:
> cc @tanx any thoughts on this?

We don't use `chan_id` in the app. So no strong feelings here. Casting to a string makes sense to prevent errors when dealing with 64 bit integers.

halseth:
Need another rebase 😅

sangaman:
@halseth done!

sangaman:
@halseth Rebased again.

sangaman:
@cfromknecht @halseth All your rebase are belong to us.

sangaman:
Re-rebased.

guggero:
@sangaman could you tell us which JS library you are using?
It looks like this setting doesn't really have any impact with either `grpc` or `@grpc/grpc-js`.
But rather setting `{ longs: String }` in the `@grpc/proto-loader` options seemed to fix this.

sangaman:
I'm using the `grpc` package on npm, but I think the main factor there is what you are using for protobuf.

PR #2355, opened 2018-12-20
invoices: create package

joostjager:
Isolates the invoice registry in a separate package. It is preparation for the creation of an invoices sub server.


Roasbeef:
Fails travis atm with linter checks.

joostjager:
Hmm different error now

joostjager:
Another circular dep just in the test

joostjager:
Good now, PTAL

PR #2356, opened 2018-12-20
invoices: add subscribesingleinvoice [no review]

joostjager:
PR builds on top of #2355.

Work in progress.


lnrpc/invoicesrpc/log.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// log is a logger that is initialized with no output filters.  This means the
+// package will not perform any logging by default until the caller requests
+// it.
+var log btclog.Logger
+
+// The default amount of logging is none.
+func init() {
+	UseLogger(build.NewSubLogger("INVC_RPC", nil))
```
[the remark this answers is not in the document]
joostjager: Changed

invoices/invoiceregistry.go
```diff
@@ -140,73 +140,80 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
...
+		// index it has. We'll use this to ensure we
+		// don't send a notification twice, which can
+		// happen if a new event is added while we're
+		// catching up a new client.
+		switch event.state {
+		case channeldb.ContractSettled:
+			client.settleIndex = invoice.SettleIndex
+		case channeldb.ContractOpen:
+			client.addIndex = invoice.AddIndex
+		}
```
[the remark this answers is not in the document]
joostjager: restored

invoices/invoiceregistry.go
```diff
@@ -129,25 +134,64 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
 			// continue.
 			i.notificationClients[newClient.id] = newClient
 
+		// A new single invoice subscription has arrived. We'll query
+		// for any backlog notifications, then add it to the set of
+		// clients.
+		case newClient := <-i.newSingleSubscriptions:
+			err := i.deliverSingleBacklogEvents(newClient)
+			if err != nil {
```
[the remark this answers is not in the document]
joostjager: yes, that's better. the client could register in anticipation of the invoice.

lnrpc/utils.go
```diff
@@ -0,0 +1,117 @@
...
+	activeNetParams *chaincfg.Params) (*Invoice, error) {
+
+	paymentRequest := string(invoice.PaymentRequest)
+	decoded, err := zpay32.Decode(paymentRequest, activeNetParams)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode payment request: %v",
+			err)
+	}
+
+	descHash := []byte("")
```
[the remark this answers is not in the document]
joostjager: i think so. just copied from `rpcserver.go`. but fixed.

lnrpc/invoicesrpc/invoices_server.go
```diff
@@ -0,0 +1,102 @@
...
+
+	cfg *Config
+}
+
+// A compile time check to ensure that Server fully implements the
+// InvoicesServer gRPC service.
+var _ InvoicesServer = (*Server)(nil)
+
+// fileExists reports whether the named file or directory exists.
+func fileExists(name string) bool {
```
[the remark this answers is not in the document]
joostjager: extracted

lnrpc/invoicesrpc/invoices_server.go
```diff
@@ -47,16 +73,48 @@ func fileExists(name string) bool {
...
-	// existing onchain/offchain permissions.
+	// If the path of the invoices macaroon wasn't specified, then we'll
+	// assume that it's found at the default network directory.
+	macFilePath := filepath.Join(
+		cfg.NetworkDir, DefaultInvoicesMacFilename,
+	)
+
+	// Now that we know the full path of the invoices macaroon, we can
+	// check to see if we need to create it or not.
+	if !fileExists(macFilePath) && cfg.MacService != nil {
```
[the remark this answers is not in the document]
joostjager: Gave it a try, but it gets many parameters. I think it may be better to, in a separate pr, have subservers expose some macaroon info and bake the macaroon somewhere else.

joostjager:
@cfromknecht PTAL

invoices/invoiceregistry.go
```diff
@@ -140,73 +140,81 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
...
-					client.addIndex = invoice.AddIndex
-				default:
-					log.Errorf("unknown invoice "+
-						"state: %v", event.state)
-				}
-			}
+// dispatchToClients passes the supplied event to all notification clients that
+// subscribed to all invoices. Add and settle indices are used to make sure that
+// clients don't receive duplicate or unwanted events.
+func (i *InvoiceRegistry) dispatchToClients(event *invoiceEvent) {
```
cfromknecht: really nice refactor, more breathing room

invoices/invoiceregistry.go
```diff
@@ -129,19 +134,58 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
 			// continue.
 			i.notificationClients[newClient.id] = newClient
 
+		// A new single invoice subscription has arrived. We'll query
+		// for any backlog notifications, then add it to the set of
+		// clients.
+		case newClient := <-i.newSingleSubscriptions:
+			err := i.deliverSingleBacklogEvents(newClient)
+			if err != nil {
+				log.Errorf("unable to deliver backlog invoice "+
```
cfromknecht: nit: log message should be capitalized

lnrpc/invoicesrpc/invoices_server.go
```diff
@@ -0,0 +1,184 @@
...
+	"path/filepath"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+const (
+	// subServerName is the name of the sub rpc server. We'll use this name
+	// to register ourselves, and we also require that the main
+	// SubServerConfigDispatcher instance recognize tt as the name of our
```
cfromknecht: s/recognize tt/recognizes it?

lnrpc/invoicesrpc/invoices_server.go
```diff
@@ -0,0 +1,184 @@
...
+// used to identify the sub-server and also de-duplicate them.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (s *Server) Name() string {
+	return subServerName
+}
+
+// RegisterWithRootServer will be called by the root gRPC server to direct a
+// sub RPC server to register itself with the main gRPC root server. Until this
+// is called, each sub-server won't be able to have
```
cfromknecht: nit: line wrapped too early

lnrpc/invoicesrpc/invoices_server.go
```diff
@@ -0,0 +1,184 @@
...
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (s *Server) Start() error {
+	return nil
+}
+
+// Stop signals any active goroutines for a graceful closure.
+//
+// NOTE: This is part of the lnrpc.SubServer interface.
+func (s *Server) Stop() error {
+	close(s.quit)
```
cfromknecht: i think it'd be worth protecting this with a compare and swap to prevent us from accidentally closing the same channel twice, as it would lead us to panic during shutdown. Start should be fine though since it's a NOP

joostjager: Calling `Stop()` twice is invalid use of the subserver and adding the compare/swap would camouflage this. We would loose the opportunity to discover the bug of the caller. In general I'd like to avoid defensive code, unless we are dealing with legacy that has become too complicate to reason about and we want to take no risk.

joostjager:
@cfromknecht PTAL

invoices/invoiceregistry_test.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"crypto/sha256"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"io/ioutil"
+	"testing"
+	"time"
+
+	"github.com/fortytw2/leaktest"
```
cfromknecht: is this import meant to be permanent? looks like the lib isn't compatible with `t.Parallel()`, so although it may work here it would be hard to use it elsewhere in the codebase

lnrpc/invoicesrpc/invoices_server.go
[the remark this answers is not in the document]
cfromknecht: sounds good, thanks!

invoices/invoiceregistry_test.go
joostjager: I was planning to leave it in. Is it a problem if we just don't use `t.Parallel()`? I try to make all my tests without any `time.Sleep`, so they are fast.

joostjager:
@cfromknecht ptal

joostjager:
Added `lnhash.Hash` structure to get rid of confusion in printing/logging payment hashes and preimages. Also a preparation for AddHoldInvoice.

invoices/invoiceregistry_test.go
cfromknecht: I think the use of parallelism greatly speeds up the unit tests, especially on travis and/or with the race detector running. attempting to not use sleeps is definitely the way to go, though there are certain areas of the codebase that test timing assumptions and it probably can't be avoided. IMO this would be a good tool to have when debugging, but don't know if it needs to be committed in mainline. perhaps @Roasbeef has some thoughts?

joostjager:
rebased

invoices/invoiceregistry_test.go
```diff
@@ -0,0 +1,163 @@
+package invoices
+
+import (
+	"crypto/sha256"
+	"github.com/btcsuite/btcd/chaincfg"
```
halseth: imports not sorted.

invoices/invoiceregistry_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	hash = sha256.Sum256(preimage[:])
+)
+
+// TestSettleInvoice tests settling of an invoice and related notifications.
+func TestSettleInvoice(t *testing.T) {
+	defer leaktest.Check(t)()
+
+	cdb, cleanup, err := newDB()
```
halseth: check error

invoices/invoiceregistry_test.go
```diff
@@ -0,0 +1,163 @@
...
+				"expected state ContractOpen, but got %v",
+				settledInvoice.Terms.State,
+			)
+		}
+	case <-time.After(testTimeout):
+		t.Fatal("no update received")
+	}
+
+	// Try to settle again.
+	err = registry.SettleInvoice(hash, amtPaid)
```
halseth: What happens if a duplicate settle with a different amount is performed?

invoices/invoiceregistry_test.go
halseth: Interesting. Will this detect cases of for instance a client goroutine within the invoiceregistry not exiting before the test is over?

invoices/invoiceregistry.go
```diff
@@ -264,8 +264,8 @@ func (i *InvoiceRegistry) deliverBacklogEvents(client *InvoiceSubscription) erro
 // by the passed preimage. Once this invoice is added, subsystems within the
 // daemon add/forward HTLCs that are able to obtain the proper preimage
 // required for redemption in the case that we're the final destination.
-func (i *InvoiceRegistry) AddDebugInvoice(amt btcutil.Amount, preimage chainhash.Hash) {
-	paymentHash := chainhash.Hash(sha256.Sum256(preimage[:]))
+func (i *InvoiceRegistry) AddDebugInvoice(amt btcutil.Amount, preimage lnhash.Hash) {
```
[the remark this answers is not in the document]
joostjager: That is an idea. Need to discuss with others to see if we want to go that way.

lnrpc/invoicesrpc/config_active.go
```diff
@@ -0,0 +1,16 @@
+// +build invoicesrpc
```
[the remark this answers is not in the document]
joostjager: What do you mean exactly? The conditional tag?

rpcserver.go
```diff
@@ -3558,7 +3458,9 @@ func (r *rpcServer) SubscribeInvoices(req *lnrpc.InvoiceSubscription,
 	for {
 		select {
 		case newInvoice := <-invoiceClient.NewInvoices:
-			rpcInvoice, err := createRPCInvoice(newInvoice)
+			rpcInvoice, err := invoicesrpc.CreateRPCInvoice(
```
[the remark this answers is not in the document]
joostjager: The idea is to group invoice rpc code in a separate package. Putting it in `lnrpc` would defeat that goal. Eventually the invoice code will be removed from `rpcserver` completely and then it is imo completely natural that it is `invoicesrpc`.

invoices/invoiceregistry_test.go
joostjager: yes

invoices/invoiceregistry_test.go
joostjager: Nothing. Test amended.

lnhash/hash.go
```diff
@@ -0,0 +1,53 @@
+package lnhash
```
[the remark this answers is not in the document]
joostjager: Added

lnrpc/invoicesrpc/config_active.go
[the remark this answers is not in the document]
joostjager: Closing because of later comment.

joostjager:
@halseth ptal

invoices/invoiceregistry.go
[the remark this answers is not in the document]
joostjager: Converted to `lntypes.Hash` and `lntypes.Preimage`.

Added `Preimage.Hash()` function as suggested by @halseth. Nice

joostjager:
@cfromknecht import order fixed

joostjager:
Rebased

lntypes/preimage.go
```diff
@@ -0,0 +1,59 @@
+package lntypes
```
Roasbeef: Not the biggest fan of a package like this...but I guess it works

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/btcsuite/goleveldb v1.0.0 // indirect
 	github.com/coreos/bbolt v0.0.0-20180223184059-7ee3ded59d4835e10f3e7d0f7603c42aa5e83820
 	github.com/davecgh/go-spew v1.1.1
+	github.com/fortytw2/leaktest v1.3.0
```
Roasbeef: Why's this needed? Would rather use external tools that add another dep.

lnrpc/gen_protos.sh
```diff
@@ -28,6 +28,7 @@ do
     echo "Generating protos from ${file}, into ${DIRECTORY}"
 
     protoc -I/usr/local/include -I.\
+           -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
```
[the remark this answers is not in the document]
Roasbeef: Why's this needed? All the other sub-severs don't yet expose an HTTP proxy.

invoices/invoiceregistry.go
```diff
@@ -413,13 +494,27 @@ func (i *InvoiceRegistry) notifyClients(invoice *channeldb.Invoice,
 	}
 }
 
+type invoiceSubscriptionKit struct {
```
Roasbeef: Missing `godoc` comment.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -0,0 +1,67 @@
...
+	conn := getClientConn(ctx, false)
+
+	cleanUp := func() {
+		conn.Close()
+	}
+
+	return invoicesrpc.NewInvoicesClient(conn), cleanUp
+}
+
+var subscribeSingleInvoiceCommand = cli.Command{
```
Roasbeef: Do we need this `cli` edition? We don't have cli versions of the existing subscription type RPC's, and it's an odd use case IMO to watch invoices on the command line, as then you'd need some other system (automated or not) to do anything with the response. Also less code if we get rid of it all together.

invoices/invoiceregistry_test.go
Roasbeef: > Is it a problem if we just don't use t.Parallel()? 

That'd make the _entire_ test suite run slower, not to mention the additional impact on tests running w/ the race condition detector. We can use external tooling to catch stray goroutine tools, would rather not further extend the execution time of our integration tests.

lntypes/preimage.go
joostjager: Yes, if it annoys us after some time, we can do something about it. Most important is that we have the strict typing now on hash and preimage.

lnrpc/gen_protos.sh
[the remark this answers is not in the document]
joostjager: Because it is importing `rpc.proto`.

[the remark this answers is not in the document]
joostjager: It is not for rest proxy generation.

cmd/lncli/invoicesrpc_active.go
joostjager: The reason I wrote it, is primarily to test the new rpc. I can imagine a user using this call instead of polling `lookupinvoice` on the command line. This use case wouldn't involve another system, but just giving the user a direct notification when the invoice is paid.

@cfromknecht what is your opinion on this cli command?

invoices/invoiceregistry.go
```diff
@@ -484,9 +593,11 @@ func (i *InvoiceRegistry) SubscribeNotifications(addIndex, settleIndex uint64) *
 		SettledInvoices: make(chan *channeldb.Invoice),
 		addIndex:        addIndex,
 		settleIndex:     settleIndex,
-		inv:             i,
-		ntfnQueue:       queue.NewConcurrentQueue(20),
-		cancelChan:      make(chan struct{}),
+		invoiceSubscriptionKit: invoiceSubscriptionKit{
```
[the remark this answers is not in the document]
joostjager: I think in initializer you have to do it like this, can't use the embedded fields as if they were not embedded. Just in initializers.

rpcserver.go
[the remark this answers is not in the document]
joostjager: Invoice is in `channeldb`. I don't think rpc marshalling code belongs there?

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -1,11 +1,20 @@
 syntax = "proto3";
 
+import "google/api/annotations.proto";
+import "rpc.proto";
```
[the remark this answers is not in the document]
joostjager: It is code duplication, how does it make it better? I don't mind this as a temporary structure. Would you like it more if the dependency was reversed? So move make the main rpc depend on invoices rpc?

subrpcserver_config.go
```diff
@@ -53,7 +54,8 @@ type subRPCServerConfigs struct {
 func (s *subRPCServerConfigs) PopulateDependencies(cc *chainControl,
 	networkDir string, macService *macaroons.Service,
 	atpl *autopilot.Manager,
-	invoiceRegistry *invoices.InvoiceRegistry) error {
+	invoiceRegistry *invoices.InvoiceRegistry,
+	activeNetParams *chaincfg.Params) error {
```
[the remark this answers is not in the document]
joostjager: `lnd` main package is not a library, so cannot access it.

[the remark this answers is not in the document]
cfromknecht: Better on pass it in IMO than rely on the presence of global variables. Makes the code less brittle

lntypes/preimage.go
```diff
@@ -0,0 +1,59 @@
...
+	preimage, err := hex.DecodeString(newPreimage)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewPreimage(preimage)
+}
+
+// Hash returns the sha256 hash of the preimage.
+func (p *Preimage) Hash() Hash {
```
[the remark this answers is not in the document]
joostjager: Yes, good idea. Although not as widely used as `Hash()`.

invoices/invoiceregistry.go
```diff
@@ -378,6 +378,14 @@ func (i *InvoiceRegistry) SettleInvoice(rHash lntypes.Hash,
 	// If this isn't a debug invoice, then we'll attempt to settle an
 	// invoice matching this rHash on disk (if one exists).
 	invoice, err := i.cdb.SettleInvoice(rHash, amtPaid)
+
+	// Implement idempotency by returning success if the invoice was already
+	// settled.
+	if err == channeldb.ErrInvoiceAlreadySettled {
```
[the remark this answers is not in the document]
joostjager: I like it more as two separate if statements. Cannot define a common `err`, because the != condition.

invoices/invoiceregistry.go
```diff
@@ -267,6 +313,37 @@ func (i *InvoiceRegistry) deliverBacklogEvents(client *InvoiceSubscription) erro
...
+// subscribing. Only in case the invoice does not yet exist, nothing is sent
+// yet.
+func (i *InvoiceRegistry) deliverSingleBacklogEvents(
+	client *SingleInvoiceSubscription) error {
+
+	invoice, err := i.cdb.LookupInvoice(client.hash)
+
+	// It is possible that the invoice does not exist yet, but the client is
+	// already watching it in anticipation.
+	if err == channeldb.ErrInvoiceNotFound {
```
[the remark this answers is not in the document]
joostjager: See above

invoices/invoiceregistry.go
```diff
@@ -413,13 +494,29 @@ func (i *InvoiceRegistry) notifyClients(invoice *channeldb.Invoice,
...
 
+// invoiceSubscriptionKit defines that are common to both all invoice
+// subscribers and single invoice subscribers.
+type invoiceSubscriptionKit struct {
+	cancelled uint32 // To be used atomically.
+
+	id uint32
+
+	inv *InvoiceRegistry
+
```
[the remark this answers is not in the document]
joostjager: returned to wire :smile:

go.mod
joostjager: Not entirely convinced, because I don't see the practical downside of having this dependency in the project and only using it in tests that run fully without blocking. For those test parallel wouldn't be a significant speed up. But removed it anyway.

invoices/invoiceregistry_test.go
joostjager: See comment above

invoices/invoiceregistry_test.go
```diff
@@ -0,0 +1,188 @@
...
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+	}
+
+	hash = sha256.Sum256(preimage[:])
+
+	// testPayReq is a dummy payment request that does parse properly. It
+	// has no relation with the real invoice parameters and isn't used in
+	// this test.
+	testPayReq = "lnbc500u1pwywxzwpp5nd2u9xzq02t0tuf2654as7vma42lwkcjptx4yzfq0umq4swpa7cqdqqcqzysmlpc9ewnydr8rr8dnltyxphdyf6mcqrsd6dml8zajtyhwe6a45d807kxtmzayuf0hh2d9tn478ecxkecdg7c5g85pntupug5kakm7xcpn63zqk"
```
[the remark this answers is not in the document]
joostjager: Comment updated.

cmd/lncli/invoicesrpc_active.go
joostjager: Removed

joostjager:
Leaktest removed. CLI command removed.

PTAL @halseth @Roasbeef @cfromknecht

go.sum
```diff
@@ -46,6 +46,8 @@ github.com/coreos/bbolt v0.0.0-20180223184059-7ee3ded59d4835e10f3e7d0f7603c42aa5
 github.com/davecgh/go-spew v0.0.0-20171005155431-ecdeabc65495/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/fortytw2/leaktest v1.3.0 h1:u8491cBMTQ8ft8aeV+adlcytMZylmA5nnwwkRZjI8vw=
```
Roasbeef: Needs a `go mod tidy` run?

joostjager: tidied

joostjager:
Stale review dismissed, ptal

invoices/invoiceregistry_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+var (
+	testTimeout = 5 * time.Second
+
+	preimage = lntypes.Preimage([32]byte{
```
cfromknecht: fwiw, we should be able to instantiate this directly using
```golang
preimage = lntypes.Preimage{
    ...
}
```
not a blocker tho

halseth:
Needs a rebase!

invoices/invoiceregistry_test.go
joostjager: fixed as i needed to rebase anyway

joostjager:
@halseth @cfromknecht @Roasbeef 

Rebased, ptal

lnrpc/gen_protos.sh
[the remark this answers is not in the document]
Roasbeef: It _is_, the annotations are what the REST proxy generation uses to generate the client+server proxy code.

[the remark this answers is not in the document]
joostjager: Yes okay, you are right. But not for REST generation for the subserver. Just not to get an error when importing `rpc.proto`.

PR #2357, opened 2018-12-20
htlcswitch: remove lnrpc dependency

joostjager:
Removes an unwanted dependency that prevents moving parts of rpcserver to a sub server.


htlcswitch/switch.go
```diff
@@ -101,6 +100,18 @@ const (
 	CloseBreach
 )
 
+// PendingUpdate describes the pending state of a closing channel.
+type PendingUpdate struct {
+	Txid        []byte
```
Roasbeef: Missing `godoc` field comments here and below.

lnrpc/rpc.proto
```diff
@@ -1153,13 +1153,6 @@ message GetInfoResponse {
     uint32 num_inactive_channels = 15 [json_name = "num_inactive_channels"];
 }
 
-message ConfirmationUpdate {
```
Roasbeef: The message currently isn't used, but recently we've gained the ability to actually populate it. Would prefer to leave it in tact (but possibly remove the entry in the oneof so the docs are accurate).

htlcswitch/switch.go
```diff
@@ -101,6 +100,18 @@ const (
...
+	OutputIndex uint32
+}
+
+// ChannelCloseUpdate contains the outcome of the close channel operation.
```
Roasbeef: FWIW, we've added two new structs to this package in this PR, yet they aren't referenced at all in the package. Instead, the structs are only created in the peer/rpcserver. This indicates that perhaps these structs shouldn't live in this package at all, or even a more aggressive refactoring is warranted.

htlcswitch/switch.go
joostjager: Moved structs and made internal.

lnrpc/rpc.proto
joostjager: Reverted delete.

htlcswitch/switch.go
joostjager: I moved them into `peer.go` where they are primarily used. More aggressive refactoring than this (for example creating a different pathway than that `interface{}` channel), I'd like to leave for another time. This PR is already a sub sub task.

Roasbeef: SGTM

PR #2360, opened 2018-12-21
channeldb+lnrpc: remove receipt invoice field

joostjager:
These fields are used nowhere in the code.

@roasbeef:
https://github.com/lightningnetwork/lnd/pull/2263#discussion_r240857344


lnrpc/rpc.proto
```diff
@@ -1716,13 +1713,6 @@ message Invoice {
     */
     string payment_request = 9 [json_name = "payment_request"];
 
-    /**
```
Roasbeef: We currently don't use the description hash field, but unlike the receipt field, it's actually part of BOLT 11. It's tagged field `h`. Atm, is specified, it _will_ be stored in the database in the form of the payreq itself. As a result, removing it would break existing behavior that works, unlike removing the receipt which doesn't do anything at all atm.

joostjager: Suppose we _would_ use the description hash field, what would it be used for in `lnd`? Is there more to it than store and return?

Roasbeef: To commit to anything that's greater than the 600 byte or so limit of the normal description field. Or it can be used to store a 256-bit identifier for an invoice. In these cases, it wouldn't be used _directly_ by `lnd`, but then again, many of the invoice fields like the description are like that: `lnd` stores it so users can query against it and implement higher level logic.

joostjager: Yes, that is what I mean. Just trying to properly interpret https://github.com/lightningnetwork/lnd/pull/2263#discussion_r240857344

But it seems that description hash should not be touched at all, correct?

Roasbeef: Yep, when I made that comment I forgot it was a part of the invoice itself.

lnrpc/rpc.proto
```diff
@@ -1685,9 +1685,6 @@ message Invoice {
     */
     string memo = 1 [json_name = "memo"];
 
-    /// An optional cryptographic receipt of payment
-    bytes receipt = 2 [json_name = "receipt"];
```
halseth: When removing proto fields, the field number should be marked `reserved`.

I also think we should just mark it deprecated for now, and remove it entirely in the future, along with other deprecated fields.

Roasbeef:
> Do we want a db migration for receipt, going through all invoices and reserializing without this field? Or can a migration be postponed and grouped with the next bigger invoice migration that comes up?

I'd say let's postpone for the bigger overhaul we've discussed offline. At that point, we'd have an entirely different database structure so we can handle it all in a single migration.

lnrpc/rpc.proto
joostjager: Ok, rewinded the removal and just marked as deprecated.

Roasbeef:
Ready to land after rebase+sign!

Roasbeef:
Another `lnrpc` related PR landed, so this needs to be rebased, but then it'll be the first one in afterwards.

PR #2361, opened 2018-12-21
lnwallet: Extract funding tx creation to a separate method

federicobond:
`handleContributionMsg` does a lot of things and could use some refactoring to improve readability. I have extracted the funding tx creation to a separate method but I'm open to rethinking or expanding the scope of this refactor.

We are testing out some alternative funding arrangements at @muun and this would also make it easier to keep our changes synced with upstream.

### DIFF
--- a/lnwallet/wallet.go
+++ b/lnwallet/wallet.go
@@ -683,34 +683,14 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
 	pendingReservation.Lock()
 	defer pendingReservation.Unlock()
 
-	// Create a blank, fresh transaction. Soon to be a complete funding
-	// transaction which will allow opening a lightning channel.
-	pendingReservation.fundingTx = wire.NewMsgTx(1)
-	fundingTx := pendingReservation.fundingTx
-
 	// Some temporary variables to cut down on the resolution verbosity.
 	pendingReservation.theirContribution = req.contribution
 	theirContribution := req.contribution
 	ourContribution := pendingReservation.ourContribution
-
-	// Add all multi-party inputs and outputs to the transaction.
-	for _, ourInput := range ourContribution.Inputs {
-		fundingTx.AddTxIn(ourInput)
-	}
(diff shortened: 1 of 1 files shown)


lnwallet/wallet.go
```diff
@@ -921,6 +868,72 @@ func (l *LightningWallet) handleSingleContribution(req *addSingleContributionMsg
 	return
 }
 
+// createAndSignFundingTx is called to create a funding transaction for a
+// pending reservation using the given multi-sig output. It will populate
+// the fundingTx and ourFundingInputScripts fields of the reservation.
+func (l *LightningWallet) createAndSignFundingTx(reservation *ChannelReservation, multiSigOut *wire.TxOut) error {
```
wpaulino: Nit: line should be 80 chars max.

lnwallet/wallet.go
```diff
@@ -921,6 +868,74 @@ func (l *LightningWallet) handleSingleContribution(req *addSingleContributionMsg
...
+// createAndSignFundingTx is called to create a funding transaction for a
+// pending reservation using the given multi-sig output. It will populate
+// the fundingTx and ourFundingInputScripts fields of the reservation.
+func (l *LightningWallet) createAndSignFundingTx(
+	reservation *ChannelReservation, multiSigOut *wire.TxOut) error {
+
+	reservation.fundingTx = wire.NewMsgTx(1)
+	fundingTx := reservation.fundingTx
+
+	ourContribution := reservation.ourContribution
```
halseth: Style suggestion: to avoid mutating passed arguments, you can make the method take `our|theirContribution` and return the `fundingTx` + `ourFundingInputScripts`.

federicobond: That's a great idea. I just pushed a new commit addressing your suggestion.

halseth:
Since this PR is not actually changing behavior or adding tests, I'll close awaiting upcoming changes to the wallet which can incorporate a similar refactoring.

PR #2368, opened 2018-12-25
keychain: extend DerivePrivKey to derive based on pubkey+KeyFamily

Roasbeef:
In this commit, we extend the DerivePrivKey method to allow callers that
don't know the full KeyLocator information to attempt to derive a
private key via a brute force mechanism. If we don't now the full
KeyLoactor, then given the KeyFamily, we can walk down the derivation
path and compare keys one by one. In order to ensure we don' t enter an
infinite loop when given an unknown public key, we cap the number of
keys derived at 100k.

An upcoming feature to lnd that adds static channel backups will utilize
this feature, as we need to derive the shachain root given only the
public key and key family, as we don't currently store this KeyLocator
on disk.

Split off of #2313.

### DIFF
--- a/keychain/btcwallet.go
+++ b/keychain/btcwallet.go
@@ -269,24 +269,74 @@ func (b *BtcWalletKeyRing) DerivePrivKey(keyDesc KeyDescriptor) (*btcec.PrivateK
 			return err
 		}
 
-		// Now that we know the account exists, we can safely derive
-		// the full private key from the given path.
-		path := waddrmgr.DerivationPath{
+		// If the public key isn't set or they have a non-zero index,
+		// then we know that the caller instead knows the derivation
+		// path for a key.
+		if keyDesc.PubKey == nil || keyDesc.Index > 0 {
+			// Now that we know the account exists, we can safely
+			// derive the full private key from the given path.
+			path := waddrmgr.DerivationPath{
+				Account: uint32(keyDesc.Family),
+				Branch:  0,
+				Index:   uint32(keyDesc.Index),
+			}
+			addr, err := scope.DeriveFromKeyPath(addrmgrNs, path)
+			if err != nil {
+				return err
+			}
+
+			key, err = addr.(waddrmgr.ManagedPubKeyAddress).PrivKey()
+			if err != nil {
+				return err
+			}
+
+			return nil
+		}
+
+		// If the public key isn't nil, then this indicates that we
+		// need to scan for the private key, assuming that we know the
+		// valid key family.
+		nextPath := waddrmgr.DerivationPath{
 			Account: uint32(keyDesc.Family),
 			Branch:  0,
-			Index:   uint32(keyDesc.Index),
-		}
-		addr, err := scope.DeriveFromKeyPath(addrmgrNs, path)
-		if err != nil {
-			return err
+			Index:   0,
 		}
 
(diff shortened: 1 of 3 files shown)


keychain/interface_test.go
```diff
@@ -356,6 +361,56 @@ func TestSecretKeyRingDerivation(t *testing.T) {
...
+						Family: keyFam,
+					},
+				}
+				privKey, err = secretKeyRing.DerivePrivKey(keyDesc)
+				if err != nil {
+					t.Fatalf("unable to derive priv key "+
+						"via scanning: %v", err)
+				}
+
+				// Having to resort to scanning, we should be able to find
```
cfromknecht: comment looks cut off?

keychain/btcwallet.go
```diff
@@ -269,24 +269,71 @@ func (b *BtcWalletKeyRing) DerivePrivKey(keyDesc KeyDescriptor) (*btcec.PrivateK
...
-			return err
+			Index:   0,
 		}
 
-		key, err = addr.(waddrmgr.ManagedPubKeyAddress).PrivKey()
-		if err != nil {
-			return err
+		// We'll now iterate through our key range in an attempt to
+		// find the target public key.
+		for i := 0; i < MaxKeyRangeScan; i++ {
```
cfromknecht: would be great if this were parallelizable, though it seems the exclusive mutex held during `DeriveFromKeyPath` would prevent doing so directly. Should we add a TODO to move the scanning into btcwallet, and expose as an API call?

keychain/interface_test.go
```diff
@@ -304,6 +305,10 @@ func TestSecretKeyRingDerivation(t *testing.T) {
 		},
 	}
 
+	// We'll clamp the max range scan to constrain the run time of the
+	// test.
+	MaxKeyRangeScan = 3
```
cfromknecht: This seems like it could be racey, since it's modified in a parallel test. might be better to set this in an `init()` method in this file

keychain/derivation.go
```diff
@@ -20,6 +24,18 @@ const (
 	BIP0043Purpose = 1017
 )
 
+var (
+	// MaxKeyRangeScan is the maximum number of keys that we'll attempt to
+	// scan with if a caller knows the public key, but not the KeyLocator
+	// and wishes to derive a private key.
+	MaxKeyRangeScan = 100000
```
cfromknecht: i was curious about how long this would take so i tried it OMM, seems to take about 10s for an unknown key. dunno how long it would take on mobile, but seems like something we can try to optimize in the future to reduce recovery latency. not a blocker on the PR tho

Roasbeef: Yeah ideally this only gets thru a hundred or so keys before it finds the target key. This is meant to be a stop gap, as a future version of `lnd` can come with a database migration to start to store the full key locator for the sha chain root.

keychain/interface_test.go
Roasbeef: Fixed

keychain/btcwallet.go
Roasbeef: TODO added

keychain/interface_test.go
Roasbeef: Fixed!

PR #2370, opened 2018-12-25
chanbackup: add new package implementing static channel backups

Roasbeef:
In this PR, we introduce a new package that implements our current scheme for static channel backups (SCB's). Note that the serialization format will likely change to use a TLV schema rather than a raw packed format. However the serialization is isolated to a single instance and can easily be modified when we decide on the final TLV format. As a result, this PR c an be reviewed as is. 

Depends on #2369. 

# Static Channel Backup Scheme 

## Crypto 

For encryption, we utilize `chacha20poly1305` with a random 24 byte nonce. We use a larger nonce size as this can be safely generated via a CSPRNG without fear of frequency collisions between nonces generated. To encrypt a blob, we then use this nonce as the AD (associated data) and prepend the nonce to the front of the ciphertext package.

For key generation, in order to ensure the user only needs their passphrase and the backup file, we utilize the existing keychain to derive a private key. In order to ensure that at we don't force any hardware signer to be aware of our crypto operations, we instead opt to utilize a public key that will be hashed to derive our private key. The assumption here is that this key will only be exposed to this software, and never derived as a public facing address.

## `chanbackup.Single`

The SCB contains all information required to initiate the data loss protection protocol once we restore the channel and connect to the remote channel peer.

The primary way outside callers will interact with this package are via the Pack and Unpack methods. Packing means writing a serialized+encrypted version of the SCB to an io.Writer. Unpacking does the opposite.

The encoding format itself uses the same encoding as we do on the wire within Lightning. Each encoded backup begins with a version so we can easily add or modify the serialization format in the future, if new channel types appear, or we need to add/remove fields. The backup contains:
  * The chain a channel belongs to. 
  * The `chanPoint` of the channel. 
  * The `shortChanID` of the channel. 
  * The public key of the remote node. 
  * The series of addresses that we can use to reach the node. 
  * The CSV delay of the channel (required to later reconstruct our output script after BOLT 1.1)
  * A `keychain.KeyLocator` that allows us to re-derive the payment bas epoint we need to sweep our funds .
  * A `keychain.KeyDescriptor` that we need in order to re-derive our `shachain` root to validate the information the remote party gives us during the DLP protocol. (see the next section for the complications that arose here)

## `chanbackup.Multi`

Multi is a series of static channel backups. This type of backup can contains ALL the channel
backup state in a single packed blob. This is suitable for storing on your file system, cloud storage, etc. Systems will be in place within lnd to ensure that one can easily obtain the latest version of the Multi for the node, and also that it will be kept up to date if channel state changes.

Broken off of #2313.

### DIFF
--- a/chanbackup/backupfile.go
+++ b/chanbackup/backupfile.go
@@ -1,0 +1,160 @@
+package chanbackup
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+const (
+	// DefaultBackupFileName is the default name of the auto updated static
+	// channel backup fie.
+	DefaultBackupFileName = "channel.backup"
+
+	// DefaultTempBackupFileName is the default name of the temporary SCB
+	// file that we'll use to atomically update the primary back up file
+	// when new channel are detected.
+	DefaultTempBackupFileName = "temp-dont-use.backup"
+)
+
+var (
+	// ErrNoBackupFileExists is returned if caller attempts to call
+	// UpdateAndSwap with the file name not set.
+	ErrNoBackupFileExists = fmt.Errorf("back up file name not set")
+
+	// ErrNoTempBackupFile is returned if caller attempts to call
+	// UpdateAndSwap with the temp back up file name not set.
+	ErrNoTempBackupFile = fmt.Errorf("temp backup file not set")
+)
+
+// MultiFile represents a file on disk that a caller can use to read the packed
+// multi backup into an unpacked one, and also atomically update the contents
+// on disk once new channels have been opened, and old ones closed. This struct
+// relies on an atomic file rename property which most widely use file systems
+// have.
+type MultiFile struct {
+	// fileName is the file name of the main back up file.
+	fileName string
+
+	// mainFile is an open handle to the main back up file.
+	mainFile *os.File
+
+	// tempFileName is the name of the file that we'll use to stage a new
+	// packed multi-chan backup, and the rename to the main back up file.
+	tempFileName string
+
+	// tempFile is an open handle to the temp back up file.
+	tempFile *os.File
+}
+
+// NewMultiFile create a new multi-file instance at the target location on the
+// file system.
+func NewMultiFile(fileName string) *MultiFile {
+
+	// We'll our temporary backup file in the very same directory as the
+	// main backup file.
+	backupFileDir := filepath.Dir(fileName)
+	tempFileName := filepath.Join(
+		backupFileDir, DefaultTempBackupFileName,
+	)
+
+	return &MultiFile{
+		fileName:     fileName,
+		tempFileName: tempFileName,
+	}
+}
+
+// UpdateAndSwap will attempt write a new temporary backup file to disk with
+// the newBackup encoded, then atomically swap (via rename) the old file for
+// the new file by updating the name of the new file to the old.
+func (b *MultiFile) UpdateAndSwap(newBackup PackedMulti) error {
+	// If the main backup file isn't set, then we can't proceed.
+	if b.fileName == "" {
+		return ErrNoBackupFileExists
+	}
+
+	// If the old back up file still exists, then we'll delete it before
+	// proceeding.
+	if _, err := os.Stat(b.tempFileName); err == nil {
+		log.Infof("Found old temp backup @ %v, removing before swap",
+			b.tempFileName)
+
+		err = os.Remove(b.tempFileName)
+		if err != nil {
+			return fmt.Errorf("unable to remove temp "+
+				"backup file: %v", err)
+		}
+	}
+
+	// Now that we know the staging area is clear, we'll create the new
+	// temporary back up file.
+	var err error
+	b.tempFile, err = os.Create(b.tempFileName)
+	if err != nil {
+		return err
+	}
+
+	// With the file created, we'll write the new packed multi backup and
+	// remove the temporary file all together once this method exits.
+	_, err = b.tempFile.Write([]byte(newBackup))
+	if err != nil {
+		return err
+	}
+	if err := b.tempFile.Sync(); err != nil {
+		return err
+	}
+	defer os.Remove(b.tempFileName)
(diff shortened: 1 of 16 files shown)


chanbackup/pubsub.go
```diff
@@ -0,0 +1,247 @@
...
+
+// ChannelNotifier represents a system that allows the chanbackup.SubSwapper to
+// be notified of any changes to the primary channel state.
+type ChannelNotifier interface {
+	// SubcribeChans requests a new channel subscription relative to the
+	// initial set of known channels. We use the knownChans as a
+	// synchronization point to ensure that the chanbackup.SubSwapper does
+	// not miss any channel open or close events in the period between when
+	// it's created, and when it requests the channel subscription.
+	SubcribeChans(map[wire.OutPoint]struct{}) (*ChannelSubscription, error)
```
cfromknecht: s/SubcribeChans/SubscribeChans

cfromknecht:
@Roasbeef needs rebase

chanbackup/crypto.go
```diff
@@ -0,0 +1,140 @@
...
+// genEncryptionKey method.
+func decryptPayloadFromReader(payload io.Reader,
+	keyRing keychain.KeyRing) ([]byte, error) {
+
+	// First, we'll re-generate the encryption key that we use for all the
+	// SCBs.
+	encryptionKey, err := genEncryptionKey(keyRing)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Roasbeef: See above.

Roasbeef:
Comments responded to and rebased! PTAL @cfromknecht

chanbackup/crypto.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// First, we'll derive the key that we'll use to encrypt the payload
+	// for safe storage without giving away the details of any of our
+	// channels.  The final operation is:
+	//
+	//  key = SHA256(baseKey)
+	encryptionKey, err := genEncryptionKey(keyRing)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
cfromknecht: can it hurt? it's at worst a marginal improvement, having one less copy in memory. it looks like `Sum256` will copy a static array with the final result back to the caller, so would say its likely that there's at least one other reference

Roasbeef:
@cfromknecht ping!

chanbackup/crypto_test.go
```diff
@@ -0,0 +1,164 @@
...
+			mutator: func(p *[]byte) {
+				// Modify the cipher text to be zero length.
+				*p = []byte{}
+			},
+			valid: false,
+		},
+		{
+			plaintext: []byte("payload test plain text"),
+			mutator: func(p *[]byte) {
+				// Modify the cipher text to be zero length.
```
cfromknecht: duplicate test case?

chanbackup/single.go
```diff
@@ -0,0 +1,328 @@
...
+	if s.ShaChainRootDesc.PubKey != nil {
+		copy(
+			shaChainPub[:],
+			s.ShaChainRootDesc.PubKey.SerializeCompressed(),
+		)
+	}
+
+	return lnwire.WriteElements(
+		w,
+		byte(s.Version),
```
cfromknecht: do you think we should also write out the serialized size of the `Single` for each version, VLB (version-length-bytes)? if we try to restore a multi backup containing versions we don't understand, we may not be able to parse anything after an unknown version bc we don't know how big it is

chanbackup/backupfile.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+	defer os.Remove(b.tempFileName)
+
+	log.Infof("Swapping old multi backup file from %v to %v",
+		b.tempFileName, b.fileName)
+
+	// Finally, we'll attempt to atomically rename the temporary file to
+	// the main back up file. If this succeeds, then we'll only have a
+	// single file on disk once this method exit.s
```
cfromknecht: s/exit.s/exits.

chanbackup/backupfile_test.go
```diff
@@ -0,0 +1,289 @@
...
+		},
+
+		// Main file not read, file name is there, but file doesn't
+		// exist.
+		{
+			fileName: "kek",
+			pass:     false,
+		},
+
+		// Mail file not read, should be able to read multiple times.
```
cfromknecht: s/Mail/Main

chanbackup/backupfile_test.go
```diff
@@ -0,0 +1,289 @@
...
+			tempFileName: filepath.Join(
+				tempTestDir, DefaultTempBackupFileName,
+			),
+			valid: true,
+		},
+	}
+	for i, testCase := range testCases {
+		// Ensure that all created files are removed at the end of the
+		// test case.
+		defer os.Remove(testCase.fileName)
```
cfromknecht: fwiw, these will return an error in the first case when the filenames are `""`. don't think it's an issue here though

chanbackup/crypto_test.go
Roasbeef: Ah yeh, fixed!

chanbackup/single.go
Roasbeef: Hadn't thought of that, updated to write out the length as a `uint16` before the raw SCB bytes.

chanbackup/backupfile_test.go
Roasbeef: Yep, just a defer in a test.

chanbackup/single.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		return err
+	}
+
+	s.Version = SingleBackupVersion(version)
+
+	switch s.Version {
+	case DefaultSingleVersion:
+	default:
+		return fmt.Errorf("unable to de-serialize w/ unknown "+
```
cfromknecht: just making a note that once we add other versions, this should be deferred until after reading the length. then if we don't understand it, discard `length` bytes and return to the caller so that we advance the reader. preferably this would be a concrete error so that the multi reader can skip/log it during deserialization

PR #2371, opened 2018-12-25
channeldb: prepatory modifications for full SCB implementation 

Roasbeef:
In this PR, we add number of methods that will be used for the final SBC implementation. Along the way we refactor some core functionality within the package to ensure that we're able to properly re-use database transactions at higher levels. See the commits for further details along with the parent PR to this. 

Broken off of #2313.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -102,6 +102,11 @@ var (
 	// ErrNoCommitPoint is returned when no data loss commit point is found
 	// in the database.
 	ErrNoCommitPoint = fmt.Errorf("no commit point found")
+
+	// ErrNoRestoredChannelMutation is returned when a caller attempts to
+	// mutate a channel that's been recovered.
+	ErrNoRestoredChannelMutation = fmt.Errorf("cannot mutate restored " +
+		"channel state")
 )
 
 // ChannelType is an enum-like type that describes one of several possible
@@ -958,6 +963,13 @@ func (c *OpenChannel) UpdateCommitment(newCommitment *ChannelCommitment) error {
 	c.Lock()
 	defer c.Unlock()
 
+	// If this is a restored channel, then we want to avoid mutating the
+	// state as all, as it's impossible to do so in a protocol compliant
+	// manner.
+	if c.hasChanStatus(ChanStatusRestored) {
+		return ErrNoRestoredChannelMutation
+	}
+
 	err := c.Db.Update(func(tx *bbolt.Tx) error {
 		chanBucket, err := fetchChanBucket(
 			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
@@ -1379,6 +1391,13 @@ func (c *OpenChannel) AppendRemoteCommitChain(diff *CommitDiff) error {
 	c.Lock()
 	defer c.Unlock()
 
+	// If this is a restored channel, then we want to avoid mutating the
+	// state as all, as it's impossible to do so in a protocol compliant
+	// manner.
+	if c.hasChanStatus(ChanStatusRestored) {
+		return ErrNoRestoredChannelMutation
+	}
+
 	return c.Db.Update(func(tx *bbolt.Tx) error {
 		// First, we'll grab the writable bucket where this channel's
 		// data resides.
@@ -1502,6 +1521,13 @@ func (c *OpenChannel) InsertNextRevocation(revKey *btcec.PublicKey) error {
 func (c *OpenChannel) AdvanceCommitChainTail(fwdPkg *FwdPkg) error {
 	c.Lock()
 	defer c.Unlock()
+
+	// If this is a restored channel, then we want to avoid mutating the
+	// state as all, as it's impossible to do so in a protocol compliant
+	// manner.
+	if c.hasChanStatus(ChanStatusRestored) {
+		return ErrNoRestoredChannelMutation
+	}
 
 	var newRemoteCommit *ChannelCommitment
 
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -972,7 +972,9 @@ func (c *OpenChannel) UpdateCommitment(newCommitment *ChannelCommitment) error {
 
 		// With the proper bucket fetched, we'll now write toe latest
 		// commitment state to dis for the target party.
-		err = putChanCommitment(chanBucket, newCommitment, true)
+		err = putChanCommitment(
+			chanBucket, newCommitment, true,
+		)
 		if err != nil {
 			return fmt.Errorf("unable to store chan "+
 				"revocations: %v", err)
@@ -1540,7 +1542,9 @@ func (c *OpenChannel) AdvanceCommitChainTail(fwdPkg *FwdPkg) error {
 		if err != nil {
 			return err
 		}
-		err = putChanCommitment(chanBucket, &newCommit.Commitment, false)
+		err = putChanCommitment(
+			chanBucket, &newCommit.Commitment, false,
+		)
 		if err != nil {
 			return err
 		}
@@ -2353,12 +2357,22 @@ func putChanCommitment(chanBucket *bbolt.Bucket, c *ChannelCommitment,
 }
 
 func putChanCommitments(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
-	err := putChanCommitment(chanBucket, &channel.LocalCommitment, true)
+	// If this is a restored channel, then we don't have any commitments to
+	// write.
+	if channel.hasChanStatus(ChanStatusRestored) {
+		return nil
+	}
+
+	err := putChanCommitment(
+		chanBucket, &channel.LocalCommitment, true,
+	)
 	if err != nil {
 		return err
 	}
 
-	return putChanCommitment(chanBucket, &channel.RemoteCommitment, false)
+	return putChanCommitment(
+		chanBucket, &channel.RemoteCommitment, false,
+	)
 }
 
 func putChanRevocationState(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
@@ -2472,6 +2486,12 @@ func fetchChanCommitment(chanBucket *bbolt.Bucket, local bool) (ChannelCommitmen
 
 func fetchChanCommitments(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
 	var err error
+
+	// If this is a restored channel, then we don't have any commitments to
+	// read.
+	if channel.hasChanStatus(ChanStatusRestored) {
+		return nil
+	}
 
 	channel.LocalCommitment, err = fetchChanCommitment(chanBucket, true)
 	if err != nil {
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -320,20 +322,54 @@ var (
 	ChanStatusRestored ChannelStatus = 1 << 3
 )
 
+// chanStatusStrings maps a ChannelStatus to a human friendly string that
+// describes that status.
+var chanStatusStrings = map[ChannelStatus]string{
+	ChanStatusDefault:           "ChanStatusDefault",
+	ChanStatusBorked:            "ChanStatusBorked",
+	ChanStatusCommitBroadcasted: "ChanStatusCommitBroadcasted",
+	ChanStatusLocalDataLoss:     "ChanStatusLocalDataLoss",
+	ChanStatusRestored:          "ChanStatusRestored",
+}
+
+// orderedChanStatusFlags is an in-order list of all that channel status flags.
+var orderedChanStatusFlags = []ChannelStatus{
+	ChanStatusDefault,
+	ChanStatusBorked,
+	ChanStatusCommitBroadcasted,
+	ChanStatusLocalDataLoss,
+	ChanStatusRestored,
+}
+
 // String returns a human-readable representation of the ChannelStatus.
 func (c ChannelStatus) String() string {
-	switch c {
-	case ChanStatusDefault:
-		return "ChanStatusDefault"
-	case ChanStatusBorked:
-		return "ChanStatusBorked"
-	case ChanStatusCommitBroadcasted:
-		return "ChanStatusCommitBroadcasted"
-	case ChanStatusLocalDataLoss:
-		return "ChanStatusLocalDataLoss"
-	default:
-		return fmt.Sprintf("Unknown(%08b)", c)
+	// If no flags are set, then this is the default case.
+	if c == 0 {
+		return chanStatusStrings[ChanStatusDefault]
 	}
+
+	// Add individual bit flags.
+	statusStr := ""
+	for _, flag := range orderedChanStatusFlags {
+		if c&flag == flag {
+			statusStr += chanStatusStrings[flag] + "|"
+			c -= flag
+		}
+	}
+
+	// Remove anything to the right of the final bar, including it as well.
+	statusStr = strings.TrimRight(statusStr, "|")
+
+	// Add any remaining flags which aren't accounted for as hex.
+	if c != 0 {
+		statusStr += "|0x" + strconv.FormatUint(uint64(c), 16)
+	}
+
+	// If this was purely an unknown flag, then remove the extra bar at the
+	// start of the string.
+	statusStr = strings.TrimLeft(statusStr, "|")
+
+	return statusStr
 }
 
 // OpenChannel encapsulates the persistent and dynamic state of an open channel
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -316,8 +316,8 @@ var (
 	// channel.
 	ChanStatusLocalDataLoss ChannelStatus = 1 << 2
 
-	// ChanStatusRestored is a status flag that signals that the chanel has
-	// been restored, and doesn't have all the fields a typical channel
+	// ChanStatusRestored is a status flag that signals that the channel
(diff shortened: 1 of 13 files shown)


channeldb/channel.go
```diff
@@ -531,6 +531,29 @@ func (c *OpenChannel) ChanStatus() ChannelStatus {
 	return c.chanStatus
 }
 
+// ApplyChanStatus allows the caller to modify the internal channel state in a
+// thead-safe manner.
+func (c *OpenChannel) ApplyChanStatus(status ChannelStatus) {
+	c.Lock()
+	defer c.Unlock()
+
+	c.chanStatus |= status
```
halseth: this will make it out of sync with the db version. Should instead use `putChanStatus`?

channeldb/db.go
```diff
@@ -408,6 +408,29 @@ func (d *DB) fetchNodeChannels(chainBucket *bbolt.Bucket) ([]*OpenChannel, error
 	return channels, nil
 }
 
+// FetchChannel attempts to locate a channel specified by the passed channel
+// point. If the channel canno tbe found, then an error will be returned.
```
halseth: typo

channeldb/db.go
```diff
@@ -539,16 +539,12 @@ func fetchChannels(d *DB, pending, waitingClose bool) ([]*OpenChannel, error) {
...
-					// than Default, then it means it is
-					// waiting to be closed.
-					channelWaitingClose :=
-						channel.ChanStatus() != ChanStatusDefault
-
 					// Only include it if we requested
 					// channels with the same waitingClose
 					// status.
-					if channelWaitingClose != waitingClose {
+					if channel.HasChanStatus(
```
halseth: This is not totally equivalent, as it could also be `DataLoss` or `Borked`

channeldb/graph.go
```diff
@@ -483,101 +483,106 @@ func (c *ChannelGraph) deleteLightningNode(nodes *bbolt.Bucket,
 // the channel supports. The chanPoint and chanID are used to uniquely identify
 // the edge globally within the database.
 func (c *ChannelGraph) AddChannelEdge(edge *ChannelEdgeInfo) error {
+
```
halseth: nit: extra newline

channeldb/graph.go
```diff
@@ -1639,28 +1644,25 @@ func delChannelByEdge(edges *bbolt.Bucket, edgeIndex *bbolt.Bucket,
...
+		return updateEdgePolicy(tx, edge)
 	})
 }
 
 // updateEdgePolicy attempts to update an edge's policy within the relevant
 // buckets using an existing database transaction.
-func updateEdgePolicy(edges, edgeIndex, nodes *bbolt.Bucket,
-	edge *ChannelEdgePolicy) error {
+func updateEdgePolicy(tx *bbolt.Tx, edge *ChannelEdgePolicy) error {
+	edges, err := tx.CreateBucketIfNotExists(edgeBucket)
```
halseth: unintentional change from `Bucket` to `CreateBucketIfNotExists` here and below?

lnwallet/channel.go
```diff
@@ -5578,14 +5578,24 @@ func (lc *LightningChannel) ForceClose() (*LocalForceCloseSummary, error) {
 	lc.Lock()
 	defer lc.Unlock()
 
+	// If the channel isn't in the default channel state, then we won't
+	// allow a force close, as it may be the case that we have a dated
+	// version of the commitment, or this is actually a channel shell.
+	if lc.channelState.ChanStatus() != channeldb.ChanStatusDefault {
+		return nil, fmt.Errorf("cannot force close channel it non " +
```
halseth: add state to error message?

halseth:
Travis is failing.

Roasbeef:
Added a new commit to use the new `FetchChannel` method in a few areas instead of scanning. Should fix this:
screen shot 2018-12-31 at 1 09 18 pm

contractcourt/chain_arbitrator.go
```diff
@@ -220,23 +220,10 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 			// With the channels fetched, attempt to locate
 			// the target channel according to its channel
 			// point.
-			dbChannels, err := c.chanSource.FetchAllChannels()
+			channel, err := c.chanSource.FetchChannel(chanPoint)
```
cfromknecht: much better :)

channeldb/nodes.go
```diff
@@ -149,40 +149,44 @@ func (db *DB) deleteLinkNode(tx *bbolt.Tx, identity *btcec.PublicKey) error {
...
+		return nil, ErrNodeNotFound
+	}
+
+	// Finally, decode an allocate a fresh LinkNode object to be returned
+	// to the caller.
+	nodeReader := bytes.NewReader(nodeBytes)
+	return deserializeLinkNode(nodeReader)
 }
 
+// TODO(roasbeef): update link node addrs in server upon connection
```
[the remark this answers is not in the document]
Roasbeef: Was left over from waaay back. It's basically the first step towards proper up time tracking within the daemon.

channeldb/nodes.go
```diff
@@ -62,13 +62,13 @@ type LinkNode struct {
 // NewLinkNode creates a new LinkNode from the provided parameters, which is
 // backed by an instance of channeldb.
 func (db *DB) NewLinkNode(bitNet wire.BitcoinNet, pub *btcec.PublicKey,
-	addr net.Addr) *LinkNode {
+	addrs []net.Addr) *LinkNode {
```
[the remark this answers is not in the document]
Roasbeef: Nope, changed over.

channeldb/db.go
```diff
@@ -520,7 +520,7 @@ func fetchChannels(d *DB, pending, waitingClose bool) ([]*OpenChannel, error) {
 					// than Default, then it means it is
 					// waiting to be closed.
 					channelWaitingClose :=
-						channel.ChanStatus() != Default
+						channel.ChanStatus() != ChanStatusDefault
```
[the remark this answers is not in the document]
Roasbeef: Arguably, it _is_ as we can't do anything with it before the remote party force closes. This section was modified slightly, but I ended up introducing a bug the first time around. I have a plan to make a follow up PR to extend `listchannels` to also display the full string chan status as well. This will help users to identify which channels have been restored and are still waiting for the recovery process to kick ouff.

contractcourt/chain_arbitrator.go
Roasbeef: yeee

Roasbeef:
Comments addressed, PTAL @halseth @cfromknecht!

channeldb/db_test.go
```diff
@@ -271,3 +277,172 @@ func TestFetchChannel(t *testing.T) {
...
+			RemoteCurrentRevocation: pub,
+			IsPending:               false,
+			RevocationStore:         shachain.NewRevocationStore(),
+			RevocationProducer:      shaChainProducer,
+		},
+	}, nil
+}
+
+// TestRestoreChannelShells tests that we're able to insert a partially
+// populated to disk. This is useful for channel recovery purposes. We should
```
halseth: nit: partially populated **edge**?

contractcourt/chain_arbitrator.go
halseth: 🎉

channeldb/db.go
```diff
@@ -408,6 +408,103 @@ func (d *DB) fetchNodeChannels(chainBucket *bbolt.Bucket) ([]*OpenChannel, error
...
+
+				chainBucket := nodeChanBucket.Bucket(chainHash)
+				if chainBucket == nil {
+					return fmt.Errorf("unable to read "+
+						"bucket for chain=%x", chainHash[:])
+				}
+
+				// Finally we reach the leaf bucket that stores
+				// all the chanPoints for this node.
+				return chainBucket.ForEach(func(k, v []byte) error {
```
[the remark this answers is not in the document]
Roasbeef: Ah didn't think of it, good idea! Implemented in the latest diff.

channeldb/db_test.go
Roasbeef: Should be channel, fixed.

channeldb/channel.go
```diff
@@ -1537,7 +1539,9 @@ func (c *OpenChannel) AdvanceCommitChainTail(fwdPkg *FwdPkg) error {
 		if err != nil {
 			return err
 		}
-		err = putChanCommitment(chanBucket, &newCommit.Commitment, false)
+		err = putChanCommitment(
+			chanBucket, &newCommit.Commitment, false, false,
```
cfromknecht: i think it might be better to read the channel status and compute `isRestoredChan` before calling here an above. this way we will fail if we try to put commitments on restored chans

channeldb/channel.go
```diff
@@ -2470,6 +2486,12 @@ func fetchChanCommitment(chanBucket *bbolt.Bucket, local bool) (ChannelCommitmen
 func fetchChanCommitments(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
 	var err error
 
+	// If this is a restored channel, then we don't have any commitments to
+	// read.
+	if channel.hasChanStatus(ChanStatusRestored) {
```
cfromknecht: this also could return an error, and lift the conditional above? that way other places that call `fetchChanCommitments` don't succeed silently when the channels have been restored

Roasbeef: See comment above, the idea is that it does, since we don't write any commitment data for restored channels. Threading through this check allowed me to minimize the number of special cases when reading out channels that are only actually shells.

channeldb/channel.go
Roasbeef: See comment above. In light of that, do you mean to error out all together if any of the mutation calls are executed for a restarted channel?

channeldb/channel.go
```diff
@@ -2332,7 +2336,11 @@ func serializeChanCommit(w io.Writer, c *ChannelCommitment) error {
 }
 
 func putChanCommitment(chanBucket *bbolt.Bucket, c *ChannelCommitment,
-	local bool) error {
+	restoredChan, local bool) error {
+
+	if restoredChan {
+		return nil
```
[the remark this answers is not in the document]
Roasbeef: Admittedly, this is a bit hacky. The idea is that if this is a restored channel, then we don't have any commitments, so they don't need to be written at all. Open to other ways to handle this scenario. An Alternative it to just use a place holder commitment in both instances.

channeldb/db.go
```diff
@@ -869,6 +870,110 @@ func (d *DB) PruneLinkNodes() error {
...
+			// With the edge info shell constructed, we'll now add
+			// it to the graph.
+			err = chanGraph.addChannelEdge(tx, &edgeInfo)
+			if err != nil {
+				return err
+			}
+
+			// Similarly, we'll construct a channel edge shell and
+			// add that itself to the graph.
+			chanEdge := ChannelEdgePolicy{
```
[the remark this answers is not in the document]
Roasbeef: Yes, since we don't know the policy of the channel at the time of closure. We don't really need any of these fields, we only need the struct to be present so we can connect to the remote peer on restart.

channeldb/db.go
```diff
@@ -869,6 +870,110 @@ func (d *DB) PruneLinkNodes() error {
...
+
+			// First, we'll attempt to create a new open channel
+			// and link node for this channel. If the channel
+			// already exists, then in order to ensure this method
+			// is idempotent, we'll continue to the next step.
+			channel.Db = d
+			err := syncNewChannel(
+				tx, channel, channelShell.NodeAddrs,
+			)
+			if err != nil && err != ErrChanAlreadyExists {
```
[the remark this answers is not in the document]
Roasbeef: So if the channel is found, but none of the graph edges or related data are there, we'd still want to create those.

Roasbeef:
Responded, to set of questions, waiting on clarification before any modifications are made. PTAL @cfromknecht

channeldb/db.go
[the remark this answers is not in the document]
cfromknecht: i see, that makes sense. tho it doesn't look like we check for the existence of the edges/policies before adding them. under what event would we have the channel w/o having added the edge? should we alway overwrite it even if its different?

channeldb/channel.go
cfromknecht: yeah i think that'd be a good safety net, that way we catch any misuses of restored channels, see other comment

channeldb/channel.go
```diff
@@ -2279,12 +2358,20 @@ func putChanCommitment(chanBucket *bbolt.Bucket, c *ChannelCommitment,
 }
 
 func putChanCommitments(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
-	err := putChanCommitment(chanBucket, &channel.LocalCommitment, true)
+	// If this is a restored channel, then we don't have any commitments to
+	// write.
+	isRestoredChan := channel.hasChanStatus(ChanStatusRestored)
```
cfromknecht: what do you think about doing this here, this is the simplest change that comes to mind:
```golang
if channel.hasChanStatus(ChanStatusRestored) {
        return nil
}

err := putChanCommitment(
         chanBucket, &channel.LocalCommitment, false, true,
)
...
```
and then other calls to `putChanCommitment` can be invoked with a computed `isRestoredChan` and return a concrete error? It doesn't seem worth it to propagate an `ignoreRestoreError` bool all the way up to `syncChannel`, but in all other instances it looks like we assume `isRestoredChan=false`, in which case we would want that error

Roasbeef: Removed the extra param from `putChanCommitment` all together in the latest fix-up commit.

channeldb/channel.go
Roasbeef: Added a safety net, also extended the tests to exercise this.

channeldb/db.go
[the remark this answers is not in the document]
Roasbeef: Hmm, good question. In the case of a clean restore, that shouldn't be possible. Removed the case, never even had a test case for it, was more of a spidey sense thing.

Roasbeef:
PTAL @halseth @cfromknecht

lnwallet/channel.go
```diff
@@ -5578,14 +5578,24 @@ func (lc *LightningChannel) ForceClose() (*LocalForceCloseSummary, error) {
 	lc.Lock()
 	defer lc.Unlock()
 
+	// If the channel isn't in the default channel state, then we won't
+	// allow a force close, as it may be the case that we have a dated
+	// version of the commitment, or this is actually a channel shell.
+	if lc.channelState.HasChanStatus(channeldb.ChanStatusLocalDataLoss) {
```
halseth: Comment and returned error says non-default state, while this checks for `DataLoss` only. What's correct?

channeldb/codec.go
```diff
@@ -147,6 +147,9 @@ func WriteElement(w io.Writer, element interface{}) error {
 	case *wire.MsgTx:
 		return e.Serialize(w)
 
+	case wire.MsgTx:
```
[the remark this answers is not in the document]
Roasbeef: Nice catch, no longer needed anymore! Had this in place when I was experimenting with another way of inserting channel shells into the database (would write a blank tx instead of having the distinct channel status).

lnwallet/channel.go
Roasbeef: Data loss, comment was left over from a check that was too broad which ended up introducing a test failure.

Roasbeef:
Latest comments addressed, PTAL!

Roasbeef:
Ping @cfromknecht @halseth !

channeldb/channel.go
```diff
@@ -656,17 +656,18 @@ func (c *OpenChannel) fullSync(tx *bbolt.Tx) error {
...
 	if err := writeOutpoint(&chanPointBuf, &c.FundingOutpoint); err != nil {
 		return err
 	}
-	chanBucket, err := chainBucket.CreateBucketIfNotExists(
+	chanBucket, err := chainBucket.CreateBucket(
 		chanPointBuf.Bytes(),
 	)
 	if err != nil {
-		return err
+		// If this channel already exists, then in order to avoid
```
halseth: Should possibly check that the returned error is `ErrBucketExists`.

channeldb/channel.go
```diff
@@ -970,7 +970,7 @@ func (c *OpenChannel) UpdateCommitment(newCommitment *ChannelCommitment) error {
 		// With the proper bucket fetched, we'll now write toe latest
 		// commitment state to dis for the target party.
 		err = putChanCommitment(
-			chanBucket, newCommitment, false, true,
+			chanBucket, newCommitment, true,
```
halseth: the  changes to this file should be squashed with the previous commit.

channeldb/db.go
```diff
@@ -772,6 +870,163 @@ func (d *DB) PruneLinkNodes() error {
...
+			// Similarly, we'll construct a channel edge shell and
+			// add that itself to the graph.
+			chanEdge := ChannelEdgePolicy{
+				ChannelID:  edgeInfo.ChannelID,
+				LastUpdate: time.Now(),
+			}
+			if selfIsSmaller {
+				chanEdge.ChannelFlags = 0
+			} else {
+				chanEdge.ChannelFlags = 1
```
cfromknecht: should use `lnwire.ChanUpdateDirection` instead of magic number :)

channeldb/channel.go
```diff
@@ -1427,6 +1519,13 @@ func (c *OpenChannel) AdvanceCommitChainTail(fwdPkg *FwdPkg) error {
 	c.Lock()
 	defer c.Unlock()
 
+	// If this is a restored channel, then we want to avoid mutating the
+	// state as all, as it's impossible to do so in a protocol compliant
+	// manner.
+	if c.hasChanStatus(ChanStatusRestored) {
```
cfromknecht: much prefer this approach!

Roasbeef:
PTAL @cfromknecht @halseth !

PR #2374, opened 2018-12-25
cmd/lncli: increase lightning gRPC block size to 50MB

Roasbeef:
In this commit, we increase the gRPC block size from 4MB to 50MB.
Recently, the output of `lncli describegraph` has hit the block size cap
due to the expansion of the mainnet graph. Without this attempts to
fetch the graph return an error of:
```
[lncli] rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4246753 vs. 4194304)
```

With this commit, we give ourselves some breathing room. It's important
to note that the max message size limit is a _client side_ setting. As a
result, any developers driving `lnd` with gRPC will also need to raise
their block size limit as well if they wish to fetch the graph over
gRPC.

### DIFF
--- a/cmd/lncli/main.go
+++ b/cmd/lncli/main.go
@@ -37,6 +37,10 @@ const (
 var (
 	defaultLndDir      = btcutil.AppDataDir("lnd", false)
 	defaultTLSCertPath = filepath.Join(defaultLndDir, defaultTLSCertFilename)
+
+	// maxMsgRecvSize is the largest message our client will receive. We
+	// set this to ~50Mb atm.
+	maxMsgRecvSize = grpc.MaxCallRecvMsgSize(1 * 1024 * 1024 * 50)
 )
 
 func fatal(err error) {
@@ -131,11 +135,10 @@ func getClientConn(ctx *cli.Context, skipMacaroons bool) *grpc.ClientConn {
 
 	// We need to use a custom dialer so we can also connect to unix sockets
 	// and not just TCP addresses.
-	opts = append(
-		opts, grpc.WithDialer(
-			lncfg.ClientAddressDialer(defaultRPCPort),
-		),
-	)
+	genericDialer := lncfg.ClientAddressDialer(defaultRPCPort)
+	opts = append(opts, grpc.WithDialer(genericDialer))
+	opts = append(opts, grpc.WithDefaultCallOptions(maxMsgRecvSize))
+
 	conn, err := grpc.Dial(ctx.GlobalString("rpcserver"), opts...)
 	if err != nil {
 		fatal(fmt.Errorf("unable to connect to RPC server: %v", err))


guggero:
Nice! While we're increasing the block size, we should also re-enable some deprecated features and then call it gRPC Cash!
It's Poon/Dryja's original vision :-D

nickfogle:
@Roasbeef this fixes the issue for me. I also had to update my client side `grpc.max_receive_message_length` setting as you mentioned above.

ekerstein:
Following what @nickfogle said, I used the following Python options to fix the gRPC example for describegraph in the API docs: https://api.lightning.community/#describegraph
`
channel = grpc.secure_channel('localhost:10009', ssl_creds, options=[('grpc.max_receive_message_length',1024*1024*50)])
`

hpbock:
This is the option to fix this issue for node.js clients:
`
const lightning = new lnrpc.Lightning(program['lnd.rpcserver'], creds, {'grpc.max_receive_message_length': 50*1024*1024});
`

C-Otto:
The description of ForwardingHistory still references the old limit and needs to be updated

misovan:
I tested the PR and I got it working only after extending both lncli client and lnd server options in `lnd.go`. On server side the required option is `MaxSendMsgSize(m int)`. Thus this commit covers only client side.

rstmsn:
Is there a 'max_receive_message_length' equivalent when using cURL ? 
An example of the full mainnet graph being successfully retrieved from the REST interface using cURL would be most useful.

gijswijs:
@rstmsn 

> Is there a 'max_receive_message_length' equivalent when using cURL ?

Yes, use headers:
```
curl --header "grpc.max_receive_message_length: 52428800"
```

ujwalakoriraj:
Hello. I have been getting this error. Tried everything that has been discussed here, but no luck!

"""
raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "failed to connect to all addresses"
        debug_error_string = "{"created":"@1650280490.570000000","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3129,"referenced_errors":[{"created":"@1650280490.570000000","description":"failed to connect to all addresses","file":"src/core/lib/transport/error_utils.cc","file_line":163,"grpc_status":14}]}"  
"""


This is the program:
"""
from pyhelm.chartbuilder import ChartBuilder
from pyhelm.tiller import Tiller
t = Tiller('127.0.0.1',44134)
chart = ChartBuilder({'name': 'mariadb', 'source': {'type': 'directory', 'location': 'C:\projects\helm_trial\pyhelm\pyhelm\pyhelm-w8r3tnsf\mariadb'}})
t.install_release(chart.get_helm_chart(), dry_run=False, namespace='default') 

"""

Anything I need to change in the program? Have I given the Tiller Host correct?

guggero:
I think you posted in the wrong repo. There's no mariadb and no helm charts (currently) in `lnd`.

PR #2384, opened 2018-12-29
Fix build failures on docker-compose.

cryptcoin-junkey:
This PR is not for any Golang code. So I omited `Pull Request Checklist`

`docker/lnd/Dockerfile` haven't been updated even though lnd was moved to Golang 1.11.
This will fix it.

### DIFF
--- a/docker/lnd/Dockerfile
+++ b/docker/lnd/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.10-alpine as builder
+FROM golang:1.11-alpine as builder
 
 MAINTAINER Olaoluwa Osuntokun <lightning.engineering>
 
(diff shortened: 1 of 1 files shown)


yancyribbens:
you could also use `alpine-sdk` instead of `musl-dev`: https://github.com/lightningnetwork/lnd/pull/2259/files.  Any reason for preferring `musl-dev`?

yancyribbens:
I've confirmed that `musl-dev` also works although `alpine-sdk` reports a _much_ smaller packer size, so all things being equal it seems like `alpine-sdk` is a better choice imo.

https://pkgs.alpinelinux.org/package/v3.3/main/x86/musl-dev
https://pkgs.alpinelinux.org/package/edge/main/x86/alpine-sdk

wpaulino:
Superseded by #2259.

PR #2385, opened 2018-12-29
peer: write buffer pool

cfromknecht:
In this PR, we implement a write buffer pool for LN peers. Currently, each peer object embeds a 65KB byte array, which is used to serialize messages before writing them to the wire. As a result, every new peer causes a large memory allocation, which places unnecessary burden on the garbage collector when faced with short-lived or flapping peers. This has been observed historically on well-peered nodes and more recently reported by @alexbosworth on y'alls node.

The solution proposed here implements a buffer pool, that dynamically grows and shrinks based on the demand for write buffers corresponding to active peers. This greatly helps when there is a high level of churn in peer activity, even if there is only one flapping peer.

To accomplish this, the bulk of this proposal implements a new queue type, called a `GCQueue` residing in the `queue` package. A `queue.GCQueue` is parameterized by a `gcInterval` and an `expiryInterval`. The `expiryInterval` specifies the minimum amount of time an item can remain in the queue before it becomes eligible to be released back to the runtime. The `gcInterval` specifies how often the queue will perform a sweep to perform the eviction.

Due to the above, the lifetime of an object within the queue is in the range `[expiryInterval, expiryInterval+gcInterval]`. As this is meant to be a generic solution, these parameters are configurable, such that specific applications may tune 1) the overhead corrleated to frequency of garbage collection and 2) how long the pool will hold on items before returning them to the runtime.

The `queue.GCQueue` is compatible with any objects implementing the `queue.Recycler` interface. Objects implementing this interface a required to have a `Recycle()` method, which allows them to be zeroed and returned as new objects to requesters.

Finally, the `lnpeer.WriteBufferPool` is implemented using an underlying `queue.GCQueue` and defining a new type `lnpeer.WriteBuffer`, which implements the `queue.Recycler` interface.

NOTE: Read buffers are managed by the brontide package, have a different size (as they include an HMAC), and have a less clearly defined life cycle than that of `peer` objects. The `queue.GCQueue` is designed with these constraints in mind so that the core logic can be shared across both buffer pools, so expect a follow up PR that addresses this separately.

### DIFF
--- a/lnpeer/write_buffer_pool.go
+++ b/lnpeer/write_buffer_pool.go
@@ -1,0 +1,79 @@
+package lnpeer
+
+import (
+	"time"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/queue"
+)
+
+const (
+	// DefaultGCInterval is the default interval that the WriteBufferPool
+	// will perform a sweep to see which expired buffers can be released to
+	// the runtime.
+	DefaultGCInterval = 15 * time.Second
+
+	// DefaultExpiryInterval is the default, minimum interval that must
+	// elapse before a WriteBuffer will be released. The maximum time before
+	// the buffer can be released is equal to the expiry interval plus the
+	// gc interval.
+	DefaultExpiryInterval = 30 * time.Second
+)
+
+// WriteBuffer is static byte array occupying to maximum-allowed
+// plaintext-message size.
+type WriteBuffer [lnwire.MaxMessagePayload]byte
+
+// Recycle zeroes the WriteBuffer, making it fresh for another use.
+// Zeroing the buffer using a logarithmic number of calls to the optimized copy
+// method. Benchmarking shows this to be ~30 times faster than a for loop that
+// sets each index to 0 for this buffer size. Inspired by:
+// https://stackoverflow.com/questions/30614165/is-there-analog-of-memset-in-go
+//
+// This is part of the queue.Recycler interface.
+func (b *WriteBuffer) Recycle() {
+	b[0] = 0
+	for i := 1; i < lnwire.MaxMessagePayload; i *= 2 {
+		copy(b[i:], b[:i])
+	}
+}
+
+// newRecyclableWriteBuffer is a constructor that returns a WriteBuffer typed as
+// a queue.Recycler.
+func newRecyclableWriteBuffer() queue.Recycler {
+	return new(WriteBuffer)
+}
+
+// A compile-time constraint to ensure that *WriteBuffer implements the
+// queue.Recycler interface.
+var _ queue.Recycler = (*WriteBuffer)(nil)
+
+// WriteBufferPool acts a global pool of WriteBuffers, that dynamically
+// allocates and reclaims buffers in response to load.
+type WriteBufferPool struct {
+	pool *queue.GCQueue
+}
+
+// NewWriteBufferPool returns a freshly instantiated WriteBufferPool, using the
+// given gcInterval and expiryIntervals.
+func NewWriteBufferPool(
+	gcInterval, expiryInterval time.Duration) *WriteBufferPool {
+
+	return &WriteBufferPool{
+		pool: queue.NewGCQueue(
+			newRecyclableWriteBuffer, 100,
+			gcInterval, expiryInterval,
+		),
+	}
+}
+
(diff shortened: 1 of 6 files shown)


peer.go
```diff
@@ -602,6 +604,11 @@ func (p *peer) WaitForDisconnect(ready chan struct{}) {
 	}
 
 	p.wg.Wait()
+
+	// Now that we are certain all active goroutines which could have been
+	// modifying the write buffer have exited, return the buffer to the pool
+	// to be reused.
+	p.server.writeBufferPool.Return(p.writeBuf)
```
Roasbeef: Alternatively, the peer that take the buffer _each_ time it goes to write a message. As is, would seem that we still need to commit 65KB to each active peer, with most of them likely being idle other than pings (once the inv system is set up). If they take the buffer each time they go to write, then eventually we'll enter a steady state of the total number of write bufs necessary to handle pings, and the links that are actually active. This'll likely have a _slight_ impact on performance, but when we get there I wager there're other lower hanging fruits to optimize away in the forwarding critical path.

Roasbeef: Re-reading the PR description, as is this would indeed still help with the allocations bursts with flappy peers, what I described above whoever would tend us to a minimal working set w.r.t the number of write buffers. Since the package is also nicely abstracted, we can eventually utilize this both within _lnwire_ and also _channeldb_ for serder purposes.

cfromknecht: Indeed we can go further, though I think that would require a little more research into the access patterns, which would inform us of how frequently we should return buffers to the pool. I'm not sure the performance penalty of doing it for every message is necessarily worth it, as traffic can often be bursty and we might send a few thousand messages at once.

One way I've though of doing this is to have a timer, that returns the buffer if no messages have been written in the last n seconds. Another alternative would be to allow the write handler to inspect the queue size, and return it if it ever goes to 0. I think both would work, but a little more involved. I agree we should first ensure our memory usage is bounded by the number of active peers, and then follow up to make it bounded by number of concurrent messages.

cfromknecht:
@roasbeef PTAL

queue/gc_queue.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+// Return adds the returned item to freelist if the queue's returnBuffer has
+// available capacity. Under load, items may be dropped to ensure this method
+// does not block.
+func (q *GCQueue) Return(item Recycler) {
+	// Recycle the item to ensure that a dirty instance is never offered
+	// from Take. The call is done here so that the CPU cycles spent
+	// clearing the buffer are down by the caller, and not by the queue
```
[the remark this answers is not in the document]
cfromknecht: Yep, with with owned!

queue/gc_queue.go
```diff
@@ -0,0 +1,210 @@
...
+			for e := q.freeList.Front(); e != nil; e = next {
+				// Cache the next element, since it will become
+				// unreachable from the current element if it is
+				// removed.
+				next = e.Next()
+				entry := e.Value.(gcQueueEntry)
+
+				// Use now - insertTime > expiryInterval to
+				// determine if this entry has expired.
+				if time.Since(entry.time) > q.expiryInterval {
```
wpaulino: Nit: can use the opposite assumption to `break` early instead.

lnpeer/write_buffer_pool.go
```diff
@@ -0,0 +1,79 @@
...
+// method. Benchmarking shows this to be ~30 times faster than a for loop that
+// sets each index to 0 for this buffer size. Inspired by:
+// https://stackoverflow.com/questions/30614165/is-there-analog-of-memset-in-go
+//
+// This is part of the queue.Recycler interface.
+func (b *WriteBuffer) Recycle() {
+	b[0] = 0
+	for i := 1; i < lnwire.MaxMessagePayload; i *= 2 {
+		copy(b[i:], b[:i])
+	}
```
wpaulino: 🤓

queue/gc_queue.go
```diff
@@ -0,0 +1,210 @@
...
+	// release elements that have been in the queue longer than
+	// expiryInterval.
+	recycleTicker ticker.Ticker
+
+	// freeList maintains a list of gcQueueEntries, sorted in order of
+	// increasing time of arrival.
+	freeList *list.List
+
+	wg   sync.WaitGroup
+	quit chan struct{}
```
wpaulino: Looks like these unused fields can be removed.

PR #1160, opened 2018-04-30
lncli: add command to create new macaroon

guggero:
As discussed in #1147 there is a need to create macaroons with custom permissions set since the three existing macaroon files `admin.macaroon`, `invoice.macaroon` and `readonly.macaroon` aren't fine-grained enough.

A new gRPC method named `NewMacaroon` is added:
* Introduces a new permission entity named `macaroon` (could be used for macaroon based RPCs mentioned in #291 too).
* `write` access to the entity macaroon is necessary to call the method NewMacaroon.
* The `admin.macaroon` gets `write` access to the entity `macaroon`.
* A new command is added to lncli named `newmacaroon` that calls this gRPC method.
* As parameters to the `NewMacaroon` method a list of entity/action pairs for the allowed operations can be passed.

Example:
```bash
lncli newmacaroon --permission=invoices/write --permission=invoices/read --save_to=~/.lnd/custom-invoice.macaroon --timeout=10
```
Creates a macaroon that is valid for reading and writing invoices during the next 10 seconds.

Closes #283, #1147, #3516.

**NOTE** for release notes: Users will need to delete or move their `admin.macaroon`, `readonly.macaroon` and `invoices.macaroon` before starting `0.9`, otherwise they won't get regenerated macaroons that have the required permission (`macaroon:generate`) to mint custom macaroons.

PR #1152, opened 2018-04-29
itest: add tests for macaroon authentication

guggero:
This PR adds a new integration test for the macaroon authentication of the gRPC interface, as mentioned in #284.

The following six scenarios are tested:
1. Expect an error when using no macaroons.
1. Expect an error when using an invalid macaroon.
1. Expect an error when trying to invoke a `write` with a read-only macaroon.
1. Expect an error when the first-party caveat `TimeoutConstraint` is not satisfied.
1. Expect an error when the first-party caveat `IPLockConstraint` is not satisfied.
1. Expect a successful response when the correct macaroon with valid first-party caveats is used.

Closes #284.

PR #1147, opened 2018-04-28
lncli: add command to delegate macaroons

guggero:
This PR addresses #283 and introduces the new command `delegatemacaroon` to the `lncli` tool.

Currently, this just exposes the functionality that the `lncli` already uses internally to the command line interface:
* Take an existing macaroon (by specifying `--macaroonpath`, default is `~/.lnd/admin.macaroon`)
* [optional] Add a timeout to it (by specifying `--timeout`)
* [optional] Restrict it to a specific IP address (by specifying `--ip_address`)

Depending on the `--format` parameter, the following output is produced (truncated for brevity):

**`lncli delegatemacaroon --format=json --timeout=3600 --ip_address=123.123.123.123`**:
```json
{
        "c": [
                {
                        "i": "time-before 2018-04-28T16:02:23.47112259Z"
                },
                {
                        "i": "ipaddr 123.123.123.123"
                }
        ],
        "l": "lnd",
        "i64": "AwoQ7V1quyefYWYwh9TyFCJCBhIBMBoWCgdhZGRyZXNzEgRy....",
        "s64": "BqjN2qrARiy4dRfX7H__3t0XRWqXtVjNs5gyXL7oU9Q"
}
```

**`lncli delegatemacaroon --format=hex --timeout=3600 --ip_address=123.123.123.123`**:
```json
{
        "macaroonHex": "0201036c6e6402bb01030a10ed5d6abb279f61663..."
}
```

Once new first-party caveats are introduced, they can be implemented here as well.

@aakselrod:
Does this cover your idea as described in #283?
As I see it, there are at least two ways to delegate macaroons in `lnd`:
1. Take one of the three existing macaroon files (`admin.macaroon`, `invoice.macaroon` and `readonly.macaroon`) and add first-party caveats to them (as implemented in this PR).
1. Bake fresh macaroons by using the encrypted root key in `macaroons.db` and define custom operations/permissions. Then add first-party caveats to those newly baked macaroons.
The second option has the advantage that the user can customize the allowed operations/permissions. But the big disadvantage that the `macaroons.db` need to be unlocked (`lnd` needs to be stopped) and decrypted with the wallet password.

Am I correct in the assumption that you meant to implement option 1 in your issue, instead of option 2?

Thanks for your feedback!

Issue #291, opened 2017-08-29
Add support for accounting-based macaroons


aakselrod:
There is a large set of functionality that can be added by extending `lnd`'s macaroon implementation by adding internal state. This partially draws on #286, but also adds root key deletion/expiration with garbage collection. This functionality could enable use cases described in [issue #20](https://github.com/lightningnetwork/lnd/issues/20#issuecomment-270564394) such as:

* Giving a process or user a macaroon with a predetermined, one-time budget
* Giving a process or user a macaroon with a predetermined, periodic budget
* Giving a process or user the ability to manually open channels and send on-chain payments
* Allowing a process or user to use only channels it has opened

This would be a pretty massive functionality addition as well as reimagining of much of the API. To do this, we would need to do the following:

- [ ] Add per-macaroon off-chain and on-chain balance tracking to the macaroon database with expiration and expired macaroon deletion.
- [ ] Add on-chain balance macaroon generation to `NewAddress`, `NewWitnessAddress`, and `CloseChannel` and consumption to `OpenChannel`, `OpenChannelSync`, `SendCoins`, and `SendMany`.
- [ ] Add off-chain balance macaroon generation to `OpenChannel` and `OpenChannelSync` and consumption to `SendPayment`, `SendPaymentSync`, and `CloseChannel`.
- [ ] Add RPCs for `CheckMacaroonBalance`, `SubdivideMacaroonBalance`, and `CombineMacaroonBalance` to manage macaroon balances.

There are other possibilities and potential issues to watch out for listed in [my previous comment on issue #20](https://github.com/lightningnetwork/lnd/issues/20#issuecomment-270564394), so this is only preliminary. This implementation is likely to get fairly complicated.

guggero:
I'll start working on accounting-based macaroons. This issue lists features for several PRs IMHO, so I'll start with a basic off-chain balance macaroon. Then we'll see where we get from there.

Kixunil:
I'm definitely interested in this. It would be useful for mitigating attacks against Joule browser extension.

Roasbeef:
Will be obviated by https://github.com/lightningnetwork/lnd/pull/5101

PR #2390, opened 2018-12-31
macaroons: initial implementation of macaroon based accounting

guggero:
In this PR we add a new accounting system to the macaroons package.
This first implementation only partially satisfies the features requested in #291 but further PRs will follow.

As we introduce it in this first PR, an account is a simple construct that has a balance and an optional expiration date. The balance represents satoshis that can be spent by the "owner" of the account.
Every invoice that is paid from/with an account reduces the balance of that account by the payment amount plus fees.
Once the balance of an account is 0, further payments will be refused from/with that account.

Applying an account to a macaroon is a *restriction*. The bearer of a macaroon that is locked to an account will only be able to spend **at most** as many satoshis as the account's balance.
If a macaroon is not locked to an account, there is no restriction.

Accounts only assert a maximum amount spendable. Having a certain account balance does not guarantee that the node has the channel liquidity to actually spend that amount.

There are three commands in `lncli` that can be used to manage accounts:
* `createaccount balance [expiration_date]` Creates a new account with the given balance and the optional expiration date.
* `listaccounts` Lists all accounts that are currently stored in the account
  database.
* `removeaccount id` Removes the account with the given ID.

Example output of `listaccounts`:
```json
{
    "accounts": [
        {
            "id": "945ab38d3890c267",
            "initial_balance": "100",
            "current_balance": "100",
            "last_update": "1546254113",
            "expiration_date": "0"
        }
    ]
}
```
A new macaroon that is locked to an account can then be created using the `delegatemacaroon` command:  
`lncli --macaroonpath /some/dir/payment.macaroon delegatemacaroon --save_to
/some/dir/account.macaroon --account_id abcdef01abcdef01`
Of course this only makes sense if we can create a macaroon that is only allowed to pay invoices, once #1160 is merged.

With this first version there are several limitations that will be addressed in further PRs:
* Accounts cannot be replenished.
* Only payment methods (`SendPayment`, `SendPaymentSync`, `SendToRoute` and `SendToRouteSync`) are currently checked for the account constraint. It can be discussed if the `ListAccounts` method should only show the current account when the macaroon is locked to an account.
* There is no link between accounts and the payments that have been paid from/with them.
* There are no accounts with on-chain balance.
* There are no accounts with a periodic balance (e.g. for subscription payments).

Depends on #1147 and #1152 (the first three commits of this PR will be removed as soon as these two PRs are merged to master).


guggero:
Realized this probably shouldn't be built into lnd itself but some kind of wrapper library. Will continue working on this in a different context.

PR #2397, opened 2019-01-02
[bugfix] Process fee updates as any other update message

### LINKED ISSUE #1908: LND rejects a valid commitment
### Background

`eclair`-`lnd` channel, an htlc has been failed right after a disconnection, and reconnection `lnd` rejects a sig which seems valid.

Note the `lnd` node definitely seems to be having issues (see huge delays between `sig` and `rev` below), but that doesn't explain the disagreement over the commitment

### Your environment

* version of `lnd`: bitrefill node `024a2e265cd66066b78a788ae615acdc84b5b0dec9efac36d7ac87513015eaf6ed`

### Steps to reproduce

```
A                B
  |<-----add-------| htlc for 9 satoshi
  |<-----sig-------|
  |------rev------>|
  |------sig------>|
  |                | an hour elapses (?)
  |<-----rev-------|
  |------fail----->| we fail immediately
  |------sig------>|
  |                | 10 minutes
  |------fee------>|
  |                | 10 minutes
  |------fee------>|
  |                |
  |     (...)      | 9 update fee in total, one every 10 minutes
  |                |
  |------fee------>|
  |                |
  |<------rev------| eventually we get the revocation two hours later
  |-------sig----->| we sign the fee update
  |<------sig------| we receive a sig with the htlc removed
  |-------rev----->|
  |                |
  |        X       | connection reset
  |                |
  |--reestablish-->|
  |<-reestablish---|
  |-----fee x9---->| we resend all the fees
  |------sig------>| and the sig
  |------rev------>| and the rev
  |<-----err-------| error: rejected commitment
```

### Expected behaviour

`lnd` accepts the sig.

### Actual behaviour

`lnd` rejects the signature, and provides the tx it expected, which does include the htlc that was previously removed (and which removal was acked by `lnd`). OTOH this tx has the correct feerate.

This looks like everything from the `fail` until the disconnection was lost on `lnd`'s side, but shouldn't the state have been persisted to disk?

### LINKED ISSUE #2348: Concurrency issue with `update_fee`
### Background

I don't know the version of `lnd` so this issue may have already been fixed, just thought it would be interesting to keep a reference just in case.

Right after having sent an `update_fee`, `lnd` rejects a `commit_sig` that I sent, and by looking at the attached error message, I can see that `lnd` was *incorrectly* expecting that I use the new unacknowledged feerate.

```
  eclair              lnd
    |                  |
    |-------sig------->|
    |<------fee--------|
    |<------err--------|
    |                  |

```

halseth:
Instead of special casing the UpdateFee messages, we instead add them to
the update logs like any other HTLC update message. This lets us avoid
having to keep an extra set of variables to keep track of the fee
updates, and instead reuse the commit/ack logic used for other updates.

This fixes a bug where we would reset the pendingFeeUpdate variable
after signing our next commitment, which would make us calculate the new
fee incorrectly if the remote sent a commitment concurrently.

A test to exercise the previous failure case is added.

Fixes #2348 
Fixes #1908

### DIFF
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -165,6 +165,10 @@ const (
 	// original add entry from the remote party's log after the next state
 	// transition.
 	Settle
+
+	// FeeUpdate is an update type sent by the channel initiator that
+	// updates the fee rate used when signing the commitment transaction.
+	FeeUpdate
 )
 
 // String returns a human readable string that uniquely identifies the target
@@ -179,6 +183,8 @@ func (u updateType) String() string {
 		return "MalformedFail"
 	case Settle:
 		return "Settle"
+	case FeeUpdate:
+		return "FeeUpdate"
 	default:
 		return "<unknown type>"
 	}
@@ -190,7 +196,7 @@ func (u updateType) String() string {
 // the original added HTLC.
 //
 // TODO(roasbeef): LogEntry interface??
-//  * need to separate attrs for cancel/add/settle
+//  * need to separate attrs for cancel/add/settle/feeupdate
 type PaymentDescriptor struct {
 	// RHash is the payment hash for this HTLC. The HTLC can be settled iff
 	// the preimage to this hash is presented.
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -1211,6 +1211,9 @@ func compactLogs(ourLog, theirLog *updateLog,
 			nextA = e.Next()
 
 			htlc := e.Value.(*PaymentDescriptor)
+
+			// We skip Adds, as they will be removed along with the
+			// fail/settles below.
 			if htlc.EntryType == Add {
 				continue
 			}
@@ -1229,6 +1232,16 @@ func compactLogs(ourLog, theirLog *updateLog,
 			if remoteChainTail >= htlc.removeCommitHeightRemote &&
 				localChainTail >= htlc.removeCommitHeightLocal {
 
+				// Fee updates have no parent htlcs, so we only
+				// remove the update itself.
+				if htlc.EntryType == FeeUpdate {
+					logA.removeUpdate(htlc.LogIndex)
+					continue
+				}
+
+				// The other types (fail/settle) do have a
+				// parent HTLC, so we'll remove that HTLC from
+				// the other log.
 				logA.removeUpdate(htlc.LogIndex)
 				logB.removeHtlc(htlc.ParentIndex)
 			}
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -351,7 +351,7 @@ type PaymentDescriptor struct {
 // NOTE: The provided `logUpdates` MUST corresponding exactly to either the Adds
 // or SettleFails in this channel's forwarding package at `height`.
 func PayDescsFromRemoteLogUpdates(chanID lnwire.ShortChannelID, height uint64,
-	logUpdates []channeldb.LogUpdate) []*PaymentDescriptor {
+	logUpdates []channeldb.LogUpdate) ([]*PaymentDescriptor, error) {
 
 	// Allocate enough space to hold all of the payment descriptors we will
 	// reconstruct, and also the list of pointers that will be returned to
@@ -423,13 +423,18 @@ func PayDescsFromRemoteLogUpdates(chanID lnwire.ShortChannelID, height uint64,
 					Index:  uint16(i),
 				},
 			}
+
+		// NOTE: UpdateFee is not expected since they are not forwarded.
+		case *lnwire.UpdateFee:
+			return nil, fmt.Errorf("unexpected update fee")
+
 		}
 
 		payDescs = append(payDescs, pd)
 		payDescPtrs = append(payDescPtrs, &payDescs[i])
 	}
 
-	return payDescPtrs
+	return payDescPtrs, nil
 }
 
 // commitment represents a commitment to a new state within an active channel.
@@ -1566,6 +1571,23 @@ func (lc *LightningChannel) logUpdateToPayDesc(logUpdate *channeldb.LogUpdate,
 			EntryType:                MalformedFail,
 			FailCode:                 wireMsg.FailureCode,
 			ShaOnionBlob:             wireMsg.ShaOnionBlob,
+			removeCommitHeightRemote: commitHeight,
+		}
+
+	// For fee updates we'll create a FeeUpdate type to add to the log. We
+	// reuse the amount field to hold the fee rate. Since the amount field
+	// is denominated in msat we won't lose precision when storing the
+	// sat/kw denominated feerate. Note that we set both the add and remove
+	// height to the same value, as we consider the fee update locked in by
+	// adding and removing it at the same height.
+	case *lnwire.UpdateFee:
+		pd = &PaymentDescriptor{
+			LogIndex: logUpdate.LogIndex,
+			Amount: lnwire.NewMSatFromSatoshis(
+				btcutil.Amount(wireMsg.FeePerKw),
+			),
+			EntryType:                FeeUpdate,
+			addCommitHeightRemote:    commitHeight,
 			removeCommitHeightRemote: commitHeight,
 		}
 	}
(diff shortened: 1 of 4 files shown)


halseth:
Possibly also fixes #1908

lnwallet/channel.go
```diff
@@ -421,6 +421,11 @@ func PayDescsFromRemoteLogUpdates(chanID lnwire.ShortChannelID, height uint64,
 					Index:  uint16(i),
 				},
 			}
+
+		// NOTE: UpdateFee is not expected since they are not forwarded.
```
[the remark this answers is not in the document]
halseth: Done!

lnwallet/channel.go
```diff
@@ -2432,8 +2443,8 @@ func (lc *LightningChannel) createCommitmentTx(c *commitment,
 // TODO(halseth): return htlcs to mutate instead of mutating inside
 // method.
 func (lc *LightningChannel) evaluateHTLCView(view *htlcView, ourBalance,
-	theirBalance *lnwire.MilliSatoshi, nextHeight uint64,
-	remoteChain, mutateState bool) *htlcView {
+	theirBalance *lnwire.MilliSatoshi, nextHeight uint64, remoteChain,
+	mutateState bool, feePerKw SatPerKWeight) (*htlcView, SatPerKWeight) {
```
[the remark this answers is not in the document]
halseth: Yes, that's much better!

Roasbeef:
New set of tests failures with the latest rebase.

lnwallet/channel.go
```diff
@@ -1566,6 +1571,19 @@ func (lc *LightningChannel) logUpdateToPayDesc(logUpdate *channeldb.LogUpdate,
 			ShaOnionBlob:             wireMsg.ShaOnionBlob,
 			removeCommitHeightRemote: commitHeight,
 		}
+
+	// For fee updates we'll create a FeeUpdate type to add to the log. We
+	// reuse the amount field to hold the fee rate. Since the amount field
+	// is denominated in msat we won't lose precision when storing the
+	// sat/kw denominated feerate.
+	case *lnwire.UpdateFee:
+		pd = &PaymentDescriptor{
```
Roasbeef: `addCommitHeightRemote` isn't set. We need to properly record this in order to handle the retransmission case.

lnwallet/channel.go
```diff
@@ -3626,6 +3672,11 @@ func (lc *LightningChannel) validateCommitmentSanity(theirLogCounter,
 	ourInitialBalance := commitChain.tip().ourBalance
 	theirInitialBalance := commitChain.tip().theirBalance
 
+	// Initiate feePerKw to the last committed fee for this chain as we'll
+	// need this to determine which HTLCs are dust, and also the final fee
+	// rate.
+	view.feePerKw = commitChain.tip().feePerKw
```
Roasbeef: This seems incorrect. We only know the final fee rate for this view after we fully evaluate it (as we do below).

lnwallet/channel.go
```diff
@@ -2435,7 +2443,9 @@ func (lc *LightningChannel) evaluateHTLCView(view *htlcView, ourBalance,
 	theirBalance *lnwire.MilliSatoshi, nextHeight uint64,
 	remoteChain, mutateState bool) *htlcView {
 
-	newView := &htlcView{}
+	newView := &htlcView{
+		feePerKw: view.feePerKw,
```
Roasbeef: How can we know what the fee rate is before we've fully evaluated the view?

lnwallet/channel.go
Roasbeef:
```
--- FAIL: TestChannelRetransmissionFeeUpdate (0.13s)
    channel_test.go:3998: unable to complete bob's state transition: rejected commitment: commit_height=2, invalid_commit_sig=30440220682b6dd35e569da6cbe26568e5388bae4c76a9ed61162087bc77471c4558ae9102207a8dad8ffd4a8297c8f50463c2c8a1177ded151f2ea003ee8bedf83f20bdfc1b, commit_tx=0200000001b794385f2d1ef7ab4d9273d1906381b44f2f6f2588a3efb96a491883319847530000000000d1af1c8003204e0000000000002200204e62432fb93d883e0910926ed09e70556f2ca6d62fa74096d016fc389ec55cdbe016cd1d00000000220020d7ba6761e1ac17cbb0722c5818bb61d312ea660cf3505f0eeef4b39f81a2d4330065cd1d000000001600144e8187664a78afec5b8c6991dbc52d575ef5e27116539e20, sig_hash=9ffec9eac9de4aaeb680da3c95bb650468f48a2b57f7ef7eac81caa5bd6863ee
```

lnwallet/channel.go
```diff
@@ -3504,44 +3545,12 @@ func (lc *LightningChannel) computeView(view *htlcView, remoteChain bool,
 	// We evaluate the view at this stage, meaning settled and failed HTLCs
 	// will remove their corresponding added HTLCs.  The resulting filtered
 	// view will only have Add entries left, making it easy to compare the
-	// channel constraints to the final commitment state.
+	// channel constraints to the final commitment state. If any fee
```
Roasbeef: As the fee rate is included in the returned view now, we no longer need to return the fee rate as a distinct parameter.

lnwallet/channel.go
halseth: You're right. Test failure was due to https://github.com/lightningnetwork/lnd/commit/516a6a794cb8b2048c51a2430b9e8784de2d2d45, which is fixed in https://github.com/lightningnetwork/lnd/pull/2397/commits/821cb23f529651cf3202ff58beaf3577a364cebe

lnwallet/channel.go
halseth: This is the initial fee rate. If `fee_updates` are found among the log updates, it will be updated, otherwise it will stay the same. 

Added a comment.

lnwallet/channel.go
halseth: It is not returned as distinct parameter. I just set it to its own variable to minimize the diff here.

lnwallet/channel.go
halseth: Yeah, reverted this, as it was the culprit of the test failures

lnwallet/channel.go
halseth: Ah, I see what you mean now. `feePerKw` is returned from `computeView`. Fixed!

lnwallet/channel.go
```diff
@@ -1571,7 +1571,8 @@ func (lc *LightningChannel) logUpdateToPayDesc(logUpdate *channeldb.LogUpdate,
 			Amount: lnwire.NewMSatFromSatoshis(
 				btcutil.Amount(wireMsg.FeePerKw),
 			),
-			EntryType: FeeUpdate,
+			EntryType:                FeeUpdate,
+			removeCommitHeightRemote: commitHeight,
```
Roasbeef: Shouldn't this be add?

halseth: It should indeed.... Fixed!

halseth: Thninking a bit more about this, maybe it should really be `remove`.. To make the `feeUpdates` being removed when compacting logs, then the remove height must be set. lmk what you think.

halseth: Since `FeeUpdates` doesn't really have the same property of getting added and removed at different heights, I decided to just set both fields at the same time.

When compacting logs we'll now properly remove it if the remove height has passed: https://github.com/lightningnetwork/lnd/pull/2397/commits/277949cb0e08095091a6a56c44d34d553a4d02d1

Also added a test to check that it gets removed during log compacting: https://github.com/lightningnetwork/lnd/pull/2397/commits/e6ee835bbe3238365a7e5054f4fd65b2dea05a92

Roasbeef: Cool yeah it's slightly different since it doesn't have the same add/remove semantics as the others. However, I think we can emulate that based on the two add heights. Nice was going to recommend we extend the tests to ensure that the log is empty after fully processing a fee update, will check out those extra commits!

cfromknecht: Yeah I thought about this as well, but it makes sense since we want the fee update to be "insta-removed" in a sense, plus there's so secondary message to cleanup the feeupdate add. might it be correct to set both to the same height?

lnwallet/channel.go
```diff
@@ -1229,6 +1232,16 @@ func compactLogs(ourLog, theirLog *updateLog,
 			if remoteChainTail >= htlc.removeCommitHeightRemote &&
 				localChainTail >= htlc.removeCommitHeightLocal {
 
+				// Fee updates have no parent htlcs, so we only
```
cfromknecht: 👍

Roasbeef:
Faucet testing went off without a hitch, ready to land!

PR #2402, opened 2019-01-02
docs/INSTALL.md: update Go installation options for Linux

Roasbeef:
In this commit, we modify the Linux installation instructions to use the
pre-built archives rather than install from the ppa. We do this as
typically the ppa lags behind the latest releases, and also by
downloading it in this new manner, we're able to verify the sha256 sum
of the installed Go release.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -32,12 +32,35 @@
 
 
     On Linux:
+
+    (x86-64)
     ```
-    sudo apt-get install golang-1.11-go
+    wget https://dl.google.com/go/go1.11.4.linux-amd64.tar.gz
+    sha256sum go1.11.4.linux-amd64.tar.gz | awk -F " " '{ print $1 }'
     ```
-    > Note that golang-1.11-go puts binaries in /usr/lib/go-1.11/bin. If you want them on your PATH, you need to make that change yourself. Alternatively, you can run:
+
+    The final output of the command above should be
+    `fb26c30e6a04ad937bbc657a1b5bba92f80096af1e8ee6da6430c045a8db3a5b`. If it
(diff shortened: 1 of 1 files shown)


docs/INSTALL.md
```diff
@@ -65,7 +65,7 @@
 
     On Mac OS X:
     ```
-    brew install go
+    brew install go@1.11
```
halseth: brew will install `v1.11` or newer already?

Roasbeef: I don't think it will if the user hasn't upgraded the "kegs" (or w/e they're called) for some time.

halseth: `brew update && brew install go`? 😛

docs/INSTALL.md
```diff
@@ -32,12 +32,35 @@
...
+    sha256sum go1.11.4.linux-amd64.tar.gz | awk -F " " '{ print $1 }'
+    ```
+
+    The final output of the command above should be
+    `fb26c30e6a04ad937bbc657a1b5bba92f80096af1e8ee6da6430c045a8db3a5b`. If it
+    isn't, then the target REPO HAS BEEN MODIFIED, and you shouldn't install
+    this version of Go. If it matches, then proceed to install Go:
     ```
-    sudo apt-get install golang-1.11-go
+    tar -C /usr/local -xzf go1.11.4.linux-amd64.tar.gz
```
cfromknecht: needs `sudo`? reported by @plantOS

cfromknecht:
verified that `fb26c30e6a04ad937bbc657a1b5bba92f80096af1e8ee6da6430c045a8db3a5b` is the correct hash on x86-64

PR #2405, opened 2019-01-03
[autopilot] Weighted combined attachment heuristic

halseth:
Follow-up to #2341. 

We make a new autopilot heuristic `WeightedCombAttachment` that combines scores from multiple sub-heuristics into one that can be used by the agent. 

To make this possible we make sure the `NodeScore` API always returns scores in the range `0.0-1.0`.

For now we only use the existing heuristic with a weight of 1.0, but this will change in the follow-up PRs. 

Builds on #2341

### DIFF
--- a/autopilot/combinedattach.go
+++ b/autopilot/combinedattach.go
@@ -1,0 +1,121 @@
+package autopilot
+
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcutil"
+)
+
+// WeightedHeuristic is a tuple that associates a weight to an
+// AttachmentHeuristic. This is used to determining a node's final score when
+// querying several heuristics for scores.
+type WeightedHeuristic struct {
+	// Weight is this AttachmentHeuristic's relative weight factor. It
+	// should be between 0.0 and 1.0.
+	Weight float64
+
+	AttachmentHeuristic
+}
+
+// WeightedCombAttachment is an implementation of the AttachmentHeuristic
+// interface that combines the scores given by several sub-heuristics into one.
+type WeightedCombAttachment struct {
+	heuristics []*WeightedHeuristic
+}
+
+// NewWeightedCombAttachment creates a new instance of a WeightedCombAttachment.
+func NewWeightedCombAttachment(h ...*WeightedHeuristic) (
+	AttachmentHeuristic, error) {
+
+	// The sum of weights given to the sub-heuristics must sum to exactly
+	// 1.0.
+	var sum float64
+	for _, w := range h {
+		sum += w.Weight
+	}
+
+	if sum != 1.0 {
+		return nil, fmt.Errorf("weights MUST sum to 1.0 (was %v)", sum)
+	}
+
+	return &WeightedCombAttachment{
+		heuristics: h,
+	}, nil
+}
+
+// A compile time assertion to ensure WeightedCombAttachment meets the
+// AttachmentHeuristic interface.
+var _ AttachmentHeuristic = (*WeightedCombAttachment)(nil)
+
+// NodeScores is a method that given the current channel graph, current set of
+// local channels and funds available, scores the given nodes according to the
+// preference of opening a channel with them. The returned channel candidates
+// maps the NodeID to an attachment directive containing a score and a channel
+// size.
+//
+// The scores is determined by quering the set of sub-heuristics, then
+// combining these scores into a final score according to the active
+// configuration.
+//
+// The returned scores will be in the range [0, 1.0], where 0 indicates no
+// improvement in connectivity if a channel is opened to this node, while 1.0
+// is the maximum possible improvement in connectivity.
+//
+// NOTE: This is a part of the AttachmentHeuristic interface.
+func (c *WeightedCombAttachment) NodeScores(g ChannelGraph, chans []Channel,
+	chanSize btcutil.Amount, nodes map[NodeID]struct{}) (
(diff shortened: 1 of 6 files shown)


Roasbeef:
Needs rebase!

autopilot/prefattach.go
```diff
@@ -127,8 +132,9 @@ func (p *PrefAttachment) NodeScores(g ChannelGraph, chans []Channel,
 		}
 
 		// Otherwise we score the node according to its fraction of
-		// channels in the graph.
-		score := float64(nodeChans) / float64(graphChans)
+		// channels in the graph, scaled such that the highest-degree
+		// node will be given a score of 1.0.
+		score := float64(nodeChans) / float64(maxChans)
```
Roasbeef: This change significantly alters the existing algorithm. I don't see why this change is needed, since without this, the score is _still_ in the range of 0 to 1.

autopilot/combinedattach.go
```diff
@@ -0,0 +1,117 @@
...
+			// how much weight we should give to this particular
+			// score.
+			score.Score += h.Weight * sub.Score
+		}
+
+		switch {
+		// Instead of adding a node with score 0 to the returned set,
+		// we just skip it.
+		case score.Score == 0:
+			continue
```
Roasbeef: Assert the score is in the proper range after the combination?

autopilot/prefattach.go
halseth: The idea here is that it will make it easier to combine multiple heuristics. I ran into issues when attempting to combine the scores returned from the existing algorithm with the bos scores. 

Before this change the `prefattach` scores would look like `[A: 0.002, B: 0.03, C: 0.0024, ...]`, i.e. very small scores for most nodes. The `bos` scores would be binary in the sense that nodes not in the list would implicitly be given a zero-score: `[A: 1.0, C: 1.0, ...]`. 

This made it very hard to find the correct weighting if I wanted to combine them. For instance if I wanted to consider the two heuristics equally, I would give them both a weight of `0.5`. However, for the example above this would result in scores like `[A: 0.5001, B: 0.015, C: 0.5012]`. Since the scales of the scores would be vastly different, to find the right weighting was hard. Also nodes considered good by the `prefattach` algorithm would almost never get picked, since they wasn't found in the score list.

To your first point: this _shouldn't_ change the algorithm, only the scale of the returned scores. Consider the old version for score given to node `i`:
```
score_i := nodeChans_i / graphChans
```
This change scales all these scores by finding the maximum score and multiplying all by `(1.0 / score_max)`:
```
score_max := maxChans / graphChans

score_i = score_i * (1.0 / score_max)
             = (nodeChans_i / graphChans) * (1.0 / (maxChans / graphChans))
             = (nodeChans_i / graphChans) * (graphChans / maxChans)
             = nodeChans_i / maxChans

```

autopilot/combinedattach.go
```diff
@@ -0,0 +1,117 @@
...
+	Weight float64
+
+	AttachmentHeuristic
+}
+
+// WeightedCombAttachment is an implementation of the AttachmentHeuristic
+// interface that combines the scores given by several sub-heuristics into one.
+type WeightedCombAttachment struct {
+	heuristics []*WeightedHeuristic
+	sync.Mutex
```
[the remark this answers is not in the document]
halseth: It is not! Removed 😅

autopilot/combinedattach.go
```diff
@@ -0,0 +1,117 @@
...
+func NewWeightedCombAttachment(h ...*WeightedHeuristic) (AttachmentHeuristic, error) {
+
+	// The sum of weights given to the sub-heuristics must sum to exactly
+	// 1.0.
+	var sum float64
+	for _, w := range h {
+		sum += w.Weight
+	}
+
+	if sum != 1.0 {
```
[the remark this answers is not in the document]
halseth: Currently is is based on user config, so shouldn't be a problem atm.

autopilot/combinedattach.go
halseth: Do you suggest returning an error in that case? It shouldn't happen, so possibly best to surface it.

autopilot/combinedattach.go
```diff
@@ -0,0 +1,117 @@
...
+// WeightedCombAttachment is an implementation of the AttachmentHeuristic
+// interface that combines the scores given by several sub-heuristics into one.
+type WeightedCombAttachment struct {
+	heuristics []*WeightedHeuristic
+	sync.Mutex
+}
+
+// NewWeightedCombAttachment creates a new instance of a WeightedCombAttachment.
+func NewWeightedCombAttachment(h ...*WeightedHeuristic) (AttachmentHeuristic, error) {
+
```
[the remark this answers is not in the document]
halseth: properly wrapped the above line: https://github.com/lightningnetwork/lnd/pull/2405/commits/d87f5b55b8b752beaadeb452c86ad3007664147d

autopilot/prefattach.go
Roasbeef: > Also nodes considered good by the prefattach algorithm would almost never get picked, since they wasn't found in the score list.

That isn't necessarily a bad thing, they could be node that _seem_ well connected, but in reality aren't able to properly route payment.

autopilot/combinedattach.go
Roasbeef: Yep

halseth: Done.

PR #2407, opened 2019-01-03
cnct: use sweeper in commit resolver

joostjager:
Now that the sweeper is available, it isn't necessary anymore for the
commit resolver to craft its own sweep tx. Instead it can offer its
input to the sweeper and wait for the outcome.

When this PR is merged, a follow up is to use sweeper in `htlcSuccessResolver` too.

### DIFF
--- a/contractcourt/commit_sweep_resolver.go
+++ b/contractcourt/commit_sweep_resolver.go
@@ -1,17 +1,14 @@
 package contractcourt
 
 import (
-	"bytes"
 	"encoding/binary"
 	"fmt"
-	"github.com/lightningnetwork/lnd/input"
 	"io"
-	"io/ioutil"
+
+	"github.com/lightningnetwork/lnd/input"
 
 	"github.com/btcsuite/btcd/wire"
-	"github.com/davecgh/go-spew/spew"
 	"github.com/lightningnetwork/lnd/lnwallet"
-	"github.com/lightningnetwork/lnd/sweep"
 )
 
 // commitSweepResolver is a resolver that will attempt to sweep the commitment
@@ -33,11 +30,6 @@ type commitSweepResolver struct {
 
 	// chanPoint is the channel point of the original contract.
 	chanPoint wire.OutPoint
-
-	// sweepTx is the fully signed transaction which when broadcast, will
-	// sweep the commitment output into an output under control by the
-	// source wallet.
-	sweepTx *wire.MsgTx
 
 	ResolverKit
 }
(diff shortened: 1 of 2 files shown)


Roasbeef:
Before we convert over the REST of the HTLC resolvers, we should first address #1875, otherwise we create more RPC blind spots along the way.

contractcourt/contract_resolvers.go
```diff
@@ -1239,124 +1233,101 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
-		// until the commitment output has been spent.
-		spendNtfn, err := c.Notifier.RegisterSpendNtfn(
-			&c.commitResolution.SelfOutPoint,
-			c.commitResolution.SelfOutputSignDesc.Output.PkScript,
-			c.broadcastHeight,
-		)
-		if err != nil {
-			return nil, err
-		}
+	// Otherwise we are dealing with a local commitment transaction and the
```
Roasbeef: Even in this case, can't we bypass the nursery and wait out the time lock ourselves here?

joostjager:
@Roasbeef ptal. I think this is not blocked by #1875, as we don't create new blind spots in this PR.

contractcourt/commit_sweep_resolver.go
```diff
@@ -271,10 +237,6 @@ func (c *commitSweepResolver) Encode(w io.Writer) error {
 		return err
 	}
 
-	if c.sweepTx != nil {
```
cfromknecht: since we are in fact leaving data on-disk that will no longer be read, we should add a comment here explaining that there may be instances where this transaction is still present. this way we don't forget to handle/migrate if we ever need to append data to the on-disk format. would probably be useful for the decode case as well

contractcourt/commit_sweep_resolver.go
```diff
@@ -86,127 +78,101 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
 	log.Infof("%T(%v): waiting for commit sweep txid=%v conf", c, c.chanPoint,
-		c.sweepTx.TxHash())
+		sweepTx.TxHash())
 
 	// Now we'll wait until the sweeping transaction has been fully
 	// confirmed.  Once it's confirmed, we can mark this contract resolved.
-	sweepTXID := c.sweepTx.TxHash()
-	sweepingScript := c.sweepTx.TxOut[0].PkScript
+	sweepTXID := sweepTx.TxHash()
+	sweepingScript := sweepTx.TxOut[0].PkScript
```
cfromknecht: iirc, the spend notification will only trigger after it has one conf, so i'm not sure this second stage does anything? obviously this existed prior and independent of this PR, but perhaps we should consider bumping the number of confirmations

joostjager: I indeed think this second stage does nothing, but not sure if we want to change anything to it in this pr. The greater plan is to remove nursery completely and wait out the csv inside the resolver.

contractcourt/commit_sweep_resolver.go
joostjager: Comments added

joostjager:
@cfromknecht ptal

contractcourt/commit_sweep_resolver.go
Roasbeef: Yeah this was left over from when we still did mempool confirmations. IMO there's no harm in leaving it in as we should move away from single conf dispatch in many areas of the codebase.

cfromknecht:
@joostjager needs rebase now

joostjager:
I prefer to do the reporting in a follow up pr.

Rebased.

ptal @cfromknecht @Roasbeef

PR #2411, opened 2019-01-04
netann: channel status manager

cfromknecht:
In this PR, we introduce a new subsystem called the `netann.ChanStatusManager` responsible for managing the announcement of channel updates which toggle a channel's disabled bit. Most of the logic has been moved into a new `netann` package, which offers greater unit testability of the subsystem's behavior.

The `netann.ChanStatusManager` exposes two methods:
```golang
func RequestEnable(outpoint wire.OutPoint) error {}
func RequestDisable(outpoint wire.OutPoint) error {}
```

Together, these allow other subsystems to request a toggle in a particular direction. The `netann.ChanStatusManager` then handles the task of dropping duplicate requests, and also ensuring that it's network behavior follows a well-defined state machine.

## State Machine Description

At any point, the `netann.ChanStatusManager` categorizes known channels into three distinct `ChanStatus` states:
```golang
type ChanStatus uint8

const (
    // ChanStatusActive indicates that the channel's last announcement has
    // the disabled bit cleared.
    ChanStatusActive ChanStatus = iota

    // ChanStatusPendingInactive indicates that the channel's last
    // announcement has the disabled bit cleared, but that the channel was
    // detected in an inactive state. Channels in this state will have a
    // disabling announcement sent after the ChanInactiveTimeout expires
    // from the time of the first detection--unless the channel is explicitly
    // reenabled before the disabling occurs.
    ChanStatusPendingInactive

    // ChanStatusInactive indicates that the channel's last announcement has
    // the disabled bit set.
    ChanStatusInactive
)
```

Channels will always start in either `ChanStatusActive` or `ChanStatusInactive` on startup or after we detect a new outpoint (perhaps for a newly created channel) by examining the last channel update we have on disk. 

One key distinction from the existing design is that the `netann.ChanStatusManager` only uses long polling to detect inactive channels, meaning that channels can ONLY be reenabled via an explict call to `RequestEnable` for an outpoint. This allows us to use a more accurate metric for enabling channels: connection duration.

Once a channel has been detected as inactive within the switch, presumably because of a disconnection, this kicks off a timed progression from `ChanStatusActive` -> `ChanStatusPendingInactive` -> `ChanStatusInactive`.

If `RequestEnable` is not received before the progression terminates, a new announcement setting the disable bit will be broadcast to the network.

Otherwise, any call to `RequestEnable` before the progression finishes will cancel the disable from being sent, and leaves the channel in the its still-enabled state on the network. This allows the `netann.ChanStatusManager` to tolerate short-lived reconnections, without causing the node to spam the network with channel updates.

Using the exposed configurations, users can tune this to a specific threshold, e.g. require ~95% uptime over a 20 minute interval, in order for the channel to remain enabled. 

Supersedes #2080 

Depends on:
 - [x] #2417 
 - [x] #2416 
 - [x] #2445 
 - [x] #2418 
 - [x] #2452

### DIFF
--- a/config.go
+++ b/config.go
@@ -31,30 +31,32 @@ import (
 )
 
 const (
-	defaultConfigFilename      = "lnd.conf"
-	defaultDataDirname         = "data"
-	defaultChainSubDirname     = "chain"
-	defaultGraphSubDirname     = "graph"
-	defaultTLSCertFilename     = "tls.cert"
-	defaultTLSKeyFilename      = "tls.key"
-	defaultAdminMacFilename    = "admin.macaroon"
-	defaultReadMacFilename     = "readonly.macaroon"
-	defaultInvoiceMacFilename  = "invoice.macaroon"
-	defaultLogLevel            = "info"
-	defaultLogDirname          = "logs"
-	defaultLogFilename         = "lnd.log"
-	defaultRPCPort             = 10009
-	defaultRESTPort            = 8080
-	defaultPeerPort            = 9735
-	defaultRPCHost             = "localhost"
-	defaultMaxPendingChannels  = 1
-	defaultNoSeedBackup        = false
-	defaultTrickleDelay        = 90 * 1000
-	defaultInactiveChanTimeout = 20 * time.Minute
-	defaultMaxLogFiles         = 3
-	defaultMaxLogFileSize      = 10
-	defaultMinBackoff          = time.Second
-	defaultMaxBackoff          = time.Hour
+	defaultConfigFilename           = "lnd.conf"
+	defaultDataDirname              = "data"
+	defaultChainSubDirname          = "chain"
+	defaultGraphSubDirname          = "graph"
+	defaultTLSCertFilename          = "tls.cert"
+	defaultTLSKeyFilename           = "tls.key"
+	defaultAdminMacFilename         = "admin.macaroon"
+	defaultReadMacFilename          = "readonly.macaroon"
+	defaultInvoiceMacFilename       = "invoice.macaroon"
+	defaultLogLevel                 = "info"
+	defaultLogDirname               = "logs"
+	defaultLogFilename              = "lnd.log"
+	defaultRPCPort                  = 10009
+	defaultRESTPort                 = 8080
+	defaultPeerPort                 = 9735
+	defaultRPCHost                  = "localhost"
+	defaultMaxPendingChannels       = 1
+	defaultNoSeedBackup             = false
+	defaultTrickleDelay             = 90 * 1000
+	defaultChanStatusSampleInterval = time.Minute
+	defaultChanEnableTimeout        = 19 * time.Minute
+	defaultChanDisableTimeout       = 20 * time.Minute
+	defaultMaxLogFiles              = 3
+	defaultMaxLogFileSize           = 10
+	defaultMinBackoff               = time.Second
+	defaultMaxBackoff               = time.Hour
 
 	defaultTorSOCKSPort            = 9050
 	defaultTorDNSHost              = "soa.nodes.lightning.directory"
@@ -234,8 +236,10 @@ type config struct {
 
 	NoSeedBackup bool `long:"noseedbackup" description:"If true, NO SEED WILL BE EXPOSED AND THE WALLET WILL BE ENCRYPTED USING THE DEFAULT PASSPHRASE -- EVER. THIS FLAG IS ONLY FOR TESTING AND IS BEING DEPRECATED."`
 
-	TrickleDelay        int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
-	InactiveChanTimeout time.Duration `long:"inactivechantimeout" description:"If a channel has been inactive for the set time, send a ChannelUpdate disabling it."`
+	TrickleDelay             int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
+	ChanEnableTimeout        time.Duration `long:"chan-enable-timeout" description:"The duration that a peer connection must be stable before attempting to send a channel update to reenable or cancel a pending disables of the peer's channels on the network (default: 19m)."`
+	ChanDisableTimeout       time.Duration `long:"chan-disable-timeout" description:"The duration that must elapse after first detecting that an already active channel is actually inactive and sending channel update disabling it to the network. The pending disable can be canceled if the peer reconnects and becomes stable for chan-enable-timeout before the disable update is sent. (default: 20m)"`
+	ChanStatusSampleInterval time.Duration `long:"chan-status-sample-interval" description:"The polling interval between attempts to detect if an active channel has become inactive due to its peer going offline. (default: 1m)"`
 
 	Alias       string `long:"alias" description:"The node alias. Used as a moniker by peers and intelligence services"`
 	Color       string `long:"color" description:"The color of the node in hex format (i.e. '#3399FF'). Used to customize node appearance in intelligence services"`
@@ -317,11 +321,13 @@ func loadConfig() (*config, error) {
 				"preferential": 1.0,
 			},
 		},
-		TrickleDelay:        defaultTrickleDelay,
-		InactiveChanTimeout: defaultInactiveChanTimeout,
-		Alias:               defaultAlias,
-		Color:               defaultColor,
-		MinChanSize:         int64(minChanFundingSize),
+		TrickleDelay:             defaultTrickleDelay,
+		ChanStatusSampleInterval: defaultChanStatusSampleInterval,
+		ChanEnableTimeout:        defaultChanEnableTimeout,
+		ChanDisableTimeout:       defaultChanDisableTimeout,
+		Alias:                    defaultAlias,
+		Color:                    defaultColor,
+		MinChanSize:              int64(minChanFundingSize),
 		Tor: &torConfig{
 			SOCKS:   defaultTorSOCKS,
 			DNS:     defaultTorDNS,
--- a/lnd_test.go
+++ b/lnd_test.go
@@ -12755,7 +12755,7 @@ func testSendUpdateDisableChannel(net *lntest.NetworkHarness, t *harnessTest) {
 	// We create a new node Eve that has an inactive channel timeout of
 	// just 2 seconds (down from the default 20m). It will be used to test
 	// channel updates for channels going inactive.
-	eve, err := net.NewNode("Eve", []string{"--inactivechantimeout=2s"})
+	eve, err := net.NewNode("Eve", []string{"--chan-disable-timeout=2s"})
 	if err != nil {
 		t.Fatalf("unable to create eve's node: %v", err)
 	}
--- a/lnd_test.go
+++ b/lnd_test.go
@@ -1051,6 +1051,32 @@ out:
 	}
 }
 
+// assertNoChannelUpdates ensures that no ChannelUpdates are sent via the
+// graphSubscription. This method will block for the provided duration before
+// returning to the caller if successful.
+func assertNoChannelUpdates(t *harnessTest, subscription graphSubscription,
+	duration time.Duration) {
+
+	timeout := time.After(duration)
+	for {
+		select {
+		case graphUpdate := <-subscription.updateChan:
+			if len(graphUpdate.ChannelUpdates) > 0 {
+				t.Fatalf("received %d channel updates when "+
+					"none were expected",
+					len(graphUpdate.ChannelUpdates))
+			}
+
+		case err := <-subscription.errChan:
+			t.Fatalf("graph subscription failure: %v", err)
+
+		case <-timeout:
+			// No updates received, success.
+			return
+		}
+	}
+}
+
 // assertChannelPolicy asserts that the passed node's known channel policy for
 // the passed chanPoint is consistent with the expected policy values.
 func assertChannelPolicy(t *harnessTest, node *lntest.HarnessNode,
@@ -12734,7 +12760,13 @@ func testSendUpdateDisableChannel(net *lntest.NetworkHarness, t *harnessTest) {
 		},
 	)
 
-	carol, err := net.NewNode("Carol", nil)
+	carol, err := net.NewNode("Carol", []string{
+		"--minbackoff=10s",
+		"--unsafe-disconnect",
+		"--chan-enable-timeout=1.5s",
+		"--chan-disable-timeout=3s",
+		"--chan-status-sample-interval=.5s",
+	})
 	if err != nil {
 		t.Fatalf("unable to create carol's node: %v", err)
 	}
@@ -12755,7 +12787,12 @@ func testSendUpdateDisableChannel(net *lntest.NetworkHarness, t *harnessTest) {
 	// We create a new node Eve that has an inactive channel timeout of
 	// just 2 seconds (down from the default 20m). It will be used to test
 	// channel updates for channels going inactive.
-	eve, err := net.NewNode("Eve", []string{"--chan-disable-timeout=2s"})
+	eve, err := net.NewNode("Eve", []string{
+		"--minbackoff=10s",
+		"--chan-enable-timeout=1.5s",
+		"--chan-disable-timeout=3s",
+		"--chan-status-sample-interval=.5s",
+	})
 	if err != nil {
 		t.Fatalf("unable to create eve's node: %v", err)
 	}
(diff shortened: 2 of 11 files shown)


halseth:
Apart from the addition of tests, I'm struggling to see what this adds that is not already covered by #2080. How is this logic different than taking what we already have in #2080, making sure we only send updates if they alter the `disable` bit already in the DB?

cfromknecht:
> Apart from the addition of tests, I'm struggling to see what this adds that is not already covered by #2080. 

To me this is the most crucial part, as we don't have much testing on the server or insight into the existing behavior apart from the integration tests.

Apart from that, the primary difference is that the state machine only performs the passive enable -> disable transition, while the current one will perform the transition in both directions. My hunch is that this is the primary thing leading to the instability we see on the network, and would probably have strange interactions with #2080 if left unchanged.

Another difference is that all state changes are serialized via the `ChanStatusManager`s main event loop, and are kept in sync with the on-disk state. Right now external commands can modify the persistent view of the graph separately from `watchChannelStatus`'s view of each channel's current state.

With those things in mind, the goals in my mind are to reduce the state space, consolidate the logic into one unit, and provide a way of throughly testing it. It's also a good chance to work on our server bloat :)

cfromknecht:
Testing framework has been significantly reworked, and additional state machine tests for the `ChanStatusManager` have been finished. The rework now brings just over 80% coverage to the `netann` package :D

config.go
```diff
@@ -234,7 +236,10 @@ type config struct {
...
 
 	TrickleDelay        int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
-	InactiveChanTimeout time.Duration `long:"inactivechantimeout" description:"If a channel has been inactive for the set time, send a ChannelUpdate disabling it."`
+	ChanActiveTimeout   time.Duration `long:"chan-active-timeout" description:"The duration that a peer connection must be stable before attempting to send a channel announcement that enables or reenables a given channel on the network."`
+	ChanInactiveTimeout time.Duration `long:"chan-inactive-timeout" description:"The duration that must elapse after first detecting that an already active channel is actually inactive and the channel announcement disabling the channel is announced to the network."`
+
+	ChanStatusSampleInterval time.Duration `long:"chan-status-sample-interval" description:"The polling interval between attempts to detect if an active channel has become inactive with respect to the channel's ability to forward."`
```
halseth: Do we really want to make this configurable? Do you think it will provide value to the user?

config.go
```diff
@@ -313,11 +318,13 @@ func loadConfig() (*config, error) {
...
 		},
-		TrickleDelay:        defaultTrickleDelay,
-		InactiveChanTimeout: defaultInactiveChanTimeout,
-		Alias:               defaultAlias,
-		Color:               defaultColor,
-		MinChanSize:         int64(minChanFundingSize),
+		TrickleDelay:             defaultTrickleDelay,
+		ChanStatusSampleInterval: defaultChanStatusInterval,
+		ChanActiveTimeout:        defaultChanActiveTimeout,
+		ChanInactiveTimeout:      defaultChanInactiveTimeout,
```
halseth: Should also do this rename in the `server`, for this to compile.

peer.go
```diff
@@ -1775,6 +1762,23 @@ out:
 			// relevant sub-systems and launching a goroutine to
 			// wait for close tx conf.
 			p.finalizeChanClosure(chanCloser)
+
+		// The channel reannounce delay has elapsed, broadcast the
+		// reenabled channel updates to the network. This should only
+		// fire once, so we set the reenableTimeout channel to nil to
+		// mark it for garbage collection. If the peer is torn down
+		// before firing, reenabling will not be attempted.
+		case <-reenableTimeout:
```
halseth: I'm wondering if it would be even nicer to move this logic to the `ChanStatusManager`. That way the peer/server would only notify the `ChanStatusManager` about changes to the active status of the link: `becameActive`/`becameInactive`, and then the `manager` would make sure to wait for a full interval duration without any inactive events for the link before re-enabling it.

halseth: To keep the current behaviour, it would send `becameActive` when entering the `channelManager`, and `becameInactive` when exiting.

server.go
```diff
@@ -1038,11 +1059,6 @@ func (s *server) Start() error {
 		srvrLog.Infof("Auto peer bootstrapping is disabled")
 	}
 
-	// Start a goroutine that will periodically send out ChannelUpdates
-	// based on a channel's status.
-	s.wg.Add(1)
-	go s.watchChannelStatus()
```
halseth: Just remove the method as well?

netann/chan_status_manager.go
```diff
@@ -0,0 +1,577 @@
...
+
+	switch curState.Status {
+
+	// Channel is already disabled, nothing to do.
+	case ChanStatusInactive:
+		return nil
+
+	// We'll sign a new update disabling the channel if the current status
+	// is enabled or pending-inactive.
+	case ChanStatusActive, ChanStatusPendingInactive:
```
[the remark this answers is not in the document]
cfromknecht: > Otherwise, we'll go ahead and sign a new update and also reset the SendDisabledTime for this channel. 

Not sure I follow, this is intentional because channels in `ChanStatusPendingInactive` are enabled on the network, so we want to immediately disable them and then forgo the passive disable.

> Ideally the caller doesn't repeatedly call RequestDisable for the same channel, but would be nice to not delegate that responsibility.

If the disable is requested multiple times, the subsequent ones will be ignored as they end up in the `ChanStatusInactive`

netann/chan_status_manager.go
```diff
@@ -0,0 +1,577 @@
...
+		log.Infof("Announcing channel(%v) disabled [requested]",
+			outpoint)
+
+		err := m.signNextUpdate(outpoint, true)
+		if err != nil {
+			return err
+		}
+	}
+
+	m.chanStates.markInactive(outpoint)
```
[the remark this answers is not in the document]
cfromknecht: sure, i will also update the `ChanStatus` types to use enable/disable terminology

netann/chan_status_manager.go
```diff
@@ -0,0 +1,577 @@
...
+		m.chanStates.markPendingInactive(c.FundingOutpoint, disableTime)
+	}
+}
+
+// disableInactiveChannels scans through the set of monitored channels, and
+// broadcast a disable update for any pending inactive channels whose
+// SendDisableTime has been superseded by the current time.
+func (m *ChanStatusManager) disableInactiveChannels() {
+	// Now, disable any channels whose inactive chan timeout has elapsed.
+	now := time.Now()
```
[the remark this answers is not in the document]
cfromknecht: i prefer this method as it's more accurate. the time returned from the channel may be stale if they aren't dequeued promptly, as subsequent ones are dropped

netann/chan_status_manager.go
```diff
@@ -0,0 +1,577 @@
...
+			continue
+		}
+
+		// We'll skip any private channels, as they aren't used for
+		// routing within the network by other nodes.
+		if c.ChannelFlags&lnwire.FFAnnounceChannel == 0 {
+			continue
+		}
+
+		channels = append(channels, c)
```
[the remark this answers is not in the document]
cfromknecht: this can only add channels into indices less than or equal to the index of the item in the original array

netann/interface.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/channeldb"
+)
+
+// DB abstracts the required database functionality needed by the
+// ChanStatusManager.
+type DB interface {
+	// FetchAllOpenChannels returns a slice of all open channels known to
+	// the daemon. This may include private or pending channels.
+	FetchAllOpenChannels() ([]*channeldb.OpenChannel, error)
```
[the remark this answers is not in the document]
cfromknecht: @wpaulino because then we need to wrap the channeldb call in some other struct/closure that implements this. unless you're also suggesting add `FetchPublicOpenChannels` to the db as well

netann/interface.go
```diff
@@ -0,0 +1,23 @@
+package netann
+
+import (
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/channeldb"
+)
+
+// DB abstracts the required database functionality needed by the
+// ChanStatusManager.
+type DB interface {
```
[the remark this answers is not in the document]
cfromknecht: agree that it could be more specific, though given that it's a two line change maybe we can address that when we need to instead of trying to anticipate how the chan status manager's needs are different from some yet-to-be-realized subsystem? :)

peer.go
cfromknecht: I thought about this, and ultimately decided it wasn't worth it at this point in time. if we do so, we then need extra goroutines/timers inside the chan status manager to properly time the reenable, instead of reusing the existing peer's resources/lifecycle to govern this. it would be nice to have this fully self-contained, tho I think it will add a fair amount of additional complexity to the state machine and bookkeeping. Given how much time has already been spent on this refactor, I would vote to defer this

config.go
cfromknecht: I don't think it hurts to expose it since we validate it properly. If the user wants to set super short/long timeouts then the default interval may be inappropriate.

server.go
```diff
@@ -2935,91 +2935,6 @@ func (s *server) fetchNodeAdvertisedAddr(pub *btcec.PublicKey) (net.Addr, error)
 	return node.Addresses[0], nil
 }
 
-// announceChanStatus disables a channel if disabled=true, otherwise activates
-// it. This is done by sending a new channel update across the network with the
-// disabled flag set accordingly. The result of disabling the channel is it not
-// being able to forward payments.
-func (s *server) announceChanStatus(op wire.OutPoint, disabled bool) error {
```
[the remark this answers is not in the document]
cfromknecht: i don't think that's practical in this case given the refactorings and number of files that must change to hook up the new code and have the commit also build

server.go
cfromknecht: squashed the removal into the commit where everything is hooked up

cfromknecht:
@halseth @wpaulino PTAL

netann/common_test.go
```diff
@@ -0,0 +1,198 @@
...
+				lnwire.ChanUpdateDisabled != 0
+			if disabled != tc.disable {
+				t.Fatalf("expected disable:%v, found:%v",
+					tc.disable, disabled)
+			}
+
+			// Finally, validate the signature using the router's
+			// verification logic.
+			err = routing.ValidateChannelUpdateAnn(
+				pubKey, 0, newUpdate,
```
wpaulino: We'll actually need the capacity here in case the channel supports the `MaxHTLC` flag, otherwise this new update will fail validation.

netann/chan_status_manager.go
```diff
@@ -0,0 +1,577 @@
...
+	// peers.
+	ApplyChannelUpdate func(*lnwire.ChannelUpdate) error
+
+	// DB stores the set of channels that are to be monitored.
+	DB DB
+
+	// Graph stores the channel info and policies for channels in DB.
+	Graph ChannelGraph
+
+	// ChanActiveTimeout is the duration a peer's connect must remain stable
```
wpaulino: Nit: also rename these to be enabled/disabled?

netann/chan_status_manager_test.go
```diff
@@ -0,0 +1,648 @@
...
+func (h *testHarness) assertNoUpdates(duration time.Duration) {
+	h.t.Helper()
+
+	h.assertNumUpdates(0, false, duration)
+}
+
+// assertNumUpdates waits for the specified duration, asserting that n updates
+// are received on the network and that all of their disable bits are set to
+// match expEnabled. The expEnabled parameter is ignored if n is 0.
+func (h *testHarness) assertNumUpdates(n int, expEnabled bool, duration time.Duration) {
```
wpaulino: Could rework this so that we also take in the channels which we expect to have new updates? This way we can ensure there's exactly one update per channel.

wpaulino:
@cfromknecht looks like the latest version introduced a test failure:
```
    --- FAIL: TestLightningNetworkDaemon/update_channel_policy (7.19s)
        lnd_test.go:75: Failed: (update channel policy): exited with error: 
            *errors.errorString unable to send payment: received payment error: unable to find a path to destination
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd_test.go:1489 (0xbe2523)
            	testUpdateChannelPolicy: t.Fatalf("unable to send payment: %v", err)
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd_test.go:100 (0xbdac66)
            	(*harnessTest).RunTestCase: testCase.test(net, h)
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd_test.go:13282 (0xc39f74)
            	TestLightningNetworkDaemon.func4: ht.RunTestCase(testCase, lndHarness)
            /home/travis/.gimme/versions/go1.11.5.linux.amd64/src/testing/testing.go:827 (0x4f8caf)
            	tRunner: fn(t)
            /home/travis/.gimme/versions/go1.11.5.linux.amd64/src/runtime/asm_amd64.s:1333 (0x45dd91)
            	goexit: BYTE	$0x90	// NOP
```

cfromknecht:
@wpaulino indeed still looking into

peer.go
halseth: I agree. Should add a todo for this.

netann/common.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Helper function to extract the owner of the given policy.
+	owner := func(edge *channeldb.ChannelEdgePolicy) []byte {
+		var pubKey *btcec.PublicKey
+		if edge.ChannelFlags&lnwire.ChanUpdateDirection == 0 {
+			pubKey, _ = info.NodeKey1()
+		} else {
+			pubKey, _ = info.NodeKey2()
+		}
+
```
cfromknecht: i should note that in the current version, we have a check here that does
```golang
if pubKey == nil {
        return nil
}
```
do we need this check? are there cases where we don't have both? @halseth

halseth: Yes, it was made like this in order to handle the ignored error case, as this helper method is used directly in a byte comparison.

netann/chan_status_manager.go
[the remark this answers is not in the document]
halseth: I don't think this non-obvious pattern is worth it, as it doesn't look very perfromance critical :)

config.go
```diff
@@ -31,30 +31,32 @@ import (
...
+	defaultLogDirname         = "logs"
+	defaultLogFilename        = "lnd.log"
+	defaultRPCPort            = 10009
+	defaultRESTPort           = 8080
+	defaultPeerPort           = 9735
+	defaultRPCHost            = "localhost"
+	defaultMaxPendingChannels = 1
+	defaultNoSeedBackup       = false
+	defaultTrickleDelay       = 90 * 1000
+	defaultChanStatusInterval = time.Minute
```
halseth: `defaultChanStatusSampleInterval` to match the config option?

config.go
halseth: It just bloats the exposed config, and seems like all three could be derived from just setting the longest one, even during integration tests.

test_utils.go
```diff
@@ -387,13 +411,16 @@ func createTestPeer(notifier chainntnfs.ChainNotifier,
...
+		chanActiveTimeout: chanActiveTimeout,
+
 		queueQuit: make(chan struct{}),
 		quit:      make(chan struct{}),
 	}
 
 	chanID := lnwire.NewChanIDFromOutPoint(channelAlice.ChannelPoint())
 	alicePeer.activeChannels[chanID] = channelAlice
 
+	alicePeer.wg.Add(1)
```
halseth: Hmm, is this an unrelated change really? If that's the case (nit): Warrant its own commit, with a description?

cfromknecht: moved to separate commit

netann/common.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// ExtractChannelUpdate attempts to retrieve a lnwire.ChannelUpdate message from
+// an edge's info and a set of routing policies.
+//
+// NOTE: The passed policies can be nil.
+func ExtractChannelUpdate(ownerPubKey []byte,
```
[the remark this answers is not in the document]
cfromknecht: fixed, now under one commit

config.go
cfromknecht: still in favor of giving users full control , this is an important piece of allowing routing node operators to start specifying uptime SLAs for their peers. saying you want 90% uptime over a day has very diffent meanings depending on the sample interval. any deterministic function will produce unsatisfactory values for a subset of users and timescales.

netann/common.go
cfromknecht: done!

cfromknecht:
@wpaulino @halseth @Roasbeef comments addressed, commits restructured, and additional unit tests added. See [this force-push](https://github.com/lightningnetwork/lnd/compare/a3c282070ecb6c2183563c779ab31d24a5f6d57d..888fa8ae58e9655b58b8b6a3e67eb21a69c13919) for the majority of the changes

netann/chan_status_manager.go
```diff
@@ -0,0 +1,595 @@
...
+	m.wg.Wait()
+
+	return nil
+}
+
+// RequestEnable submits a request to immediately enable a channel identified by
+// the provided outpoint. If the channel is already enabled, no action will be
+// taken. If the channel is marked pending-disable the channel will be returned
+// to an active status as the scheduled disable was never sent. Otherwise if the
+// channel is found to be disabled, a new announcment will be signed with the
```
wpaulino: Nit: s/announcment/announcement/. Saw some others like this in this file as well ;)

netann/channel_update.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// ChannelUpdateModifier is a closure that makes in-place modifications to an
```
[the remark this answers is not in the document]
cfromknecht: i have a follow up PR that moves modifiers+signing of all announcement types into this package, it ends up being a very clean separation :)

netann/chan_status_manager.go
cfromknecht: grr, i've been wondering why my autocomplete keeps doing that, thanks

netann/channel_update.go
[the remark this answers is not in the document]
cfromknecht: right now there's a lot of code duplication in the gossiper on this front that gets removed as a result

cfromknecht:
@wpaulino comments addressed, @halseth @Roasbeef ptal

netann/chan_status_manager.go
```diff
@@ -0,0 +1,595 @@
...
+		}
+	}
+
+	// If the disable was requested via the manager's public interface, we
+	// will remove the output from our map of channel states. Typically this
+	// signals that the channel is being closed, so this frees up the space
+	// in the map. If for some reason the channel isn't closed, the state
+	// will be repopulated on subsequent calls to RequestEnable or
+	// RequestDisable via a db lookup, or on startup.
+	delete(m.chanStates, outpoint)
```
Roasbeef: 👍 

I think I ran into this issue in a prior diff of this PR. It appears to be fixed now.

netann/chan_status_manager.go
```diff
@@ -0,0 +1,595 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter out private channels.
+	var channels []*channeldb.OpenChannel
+	for _, c := range allChannels {
+		// We'll skip any private channels, as they aren't used for
+		// routing within the network by other nodes.
+		if c.ChannelFlags&lnwire.FFAnnounceChannel == 0 {
```
Roasbeef: Not a blocker, but this could have been hidden behind this new interface. With that path, then the `ChanStatusManager` doesn't need to know how to interpret the bits of the channel flag, or what an advertised channel even is.

lnd_test.go
```diff
@@ -12755,7 +12787,12 @@ func testSendUpdateDisableChannel(net *lntest.NetworkHarness, t *harnessTest) {
 	// We create a new node Eve that has an inactive channel timeout of
 	// just 2 seconds (down from the default 20m). It will be used to test
 	// channel updates for channels going inactive.
-	eve, err := net.NewNode("Eve", []string{"--chan-disable-timeout=2s"})
+	eve, err := net.NewNode("Eve", []string{
+		"--minbackoff=10s",
```
Roasbeef: I have a feeling that we'll need to adjust these values due to Travis flakiness in the future...

peer.go
```diff
@@ -1782,6 +1780,25 @@ out:
 			// relevant sub-systems and launching a goroutine to
 			// wait for close tx conf.
 			p.finalizeChanClosure(chanCloser)
+
+		// The channel reannounce delay has elapsed, broadcast the
```
Roasbeef: Nice, I think this'll probably make the biggest difference in our current handling, so we don't immediately enable and span the network due to a flappy peer.

PR #2412, opened 2019-01-04
lnwire+peer: fully validate node aliases on the wire, don't d/c due to invalid aliases

Roasbeef:


peer.go
```diff
@@ -996,6 +996,12 @@ out:
 				idleTimer.Reset(idleTimeout)
 				continue
 
+			// If the NodeAnnouncement has an invalid alias, then
+			// we'll log that error above and continue so we can
+			// continue to read messges from the peer.
+			case lnwire.ErrInvalidNodeAlias:
```
halseth: should `idleTimer` be reset, as in the other cases?

Roasbeef: Fixed!

Roasbeef:
Comment addressed, PTAL @halseth

peer.go
```diff
@@ -996,6 +996,13 @@ out:
...
```
halseth: This should have a test if possible. I think maybe this should be a pointer as well?

Roasbeef: There isn't a good way to make an accurate end to end test for this as we don't allow local nodes to set the value, and we don't yet have a p2p testing framework in place. Correct on pointer, just added! We do currently have unit tests that cover validation however.

PR #2416, opened 2019-01-04
netann: move NodeSigner out of main package

cfromknecht:
A preliminary step for #2411, this PR moves the node signer out of the main package and updates the server to use the imported object.


server.go
```diff
@@ -87,7 +88,7 @@ type server struct {
 
 	// nodeSigner is an implementation of the MessageSigner implementation
 	// that's backed by the identity private key of the running lnd node.
-	nodeSigner *nodeSigner
+	nodeSigner *netann.NodeSigner
```
halseth: nit: ideally should be done in same commit as rename to ensure compilation is intact

cfromknecht: split into add, rename, remove to maintain compilation and distinct files for rebasability. Final diff should appear the same as a rename

cfromknecht: yeah seems the only way if we want to try to keep the one commit/file :S

PR #2419, opened 2019-01-05
brontide: read buffer pool

cfromknecht:
This PR is a follow up of #2385, and adds read buffer pool to the `brontide` package. The underlying implementation uses the `queue.GCQueue`, added in #2385. 

Whenever a new peer connects, we currently allocate a 65KB+16 byte array to use as a read buffer. The read buffer stores the ciphertext and MAC read from the wire, and used to decrypt and then decode messages from the peer.

Because the read buffer is implemented at the connection-level, as opposed to the peer-level like write buffers, simply opening a TCP connection will cause this allocation. Therefore peers that send no messages, or do not complete the handshake, will add to this memory overhead even if they are released promptly.

Adding a read buffer pool is thus MORE critical than adding a write buffer pool, as peers that use write buffers are a subset of those that require read buffers. Coupled with the potential for asymmetric resource usage, adding a read buffer pool serves also a modest DOS hardening measure for `brontide` as a standalone package.

Along the way, I made some small modifications in terms of moving the noise prologue and key generation functions to be global variables, as this reduces the total number of allocations when initializing a `brontide.Machine`. The PR also adds error handling when setting connection deadlines, which were previously unhandled.

The commit containing the implementation of the `queue.GCQueue` is included in this so that the tests will compile, but its removal will reduce the final diff by 200 lines.

Depends on:
 - [x] #2385


brontide/noise.go
```diff
@@ -47,6 +47,10 @@ var (
 	// the cipher session exceeds the maximum allowed message payload.
 	ErrMaxMessageLengthExceeded = errors.New("the generated payload exceeds " +
 		"the max allowed message length of (2^16)-1")
+
+	// lightningPrologue identifies is the noise prologue that is used to
```
wpaulino: Nit: `identifies` not needed?

brontide/read_buffer_pool.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"time"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/queue"
+)
+
+const (
+	// DefaultGCInterval is the default interval that the WriteBufferPool
```
wpaulino: Nit: s/WriteBufferPool/ReadBufferPool. There are some other cases below as well.

brontide/read_buffer_pool.go
```diff
@@ -0,0 +1,85 @@
...
+	return new(ReadBuffer)
+}
+
+// A compile-time constraint to ensure that *ReadBuffer implements the
+// queue.Recycler interface.
+var _ queue.Recycler = (*ReadBuffer)(nil)
+
+// ReadBufferPool acts as a global pool of ReadBuffers, that dynamically
+// allocates and reclaims buffers in response to load.
+type ReadBufferPool struct {
```
wpaulino: Perhaps this can be abstracted further to prevent some code duplication between this, the `WritePoolBuffer` and any future uses of a buffer pool?

brontide/noise.go
```diff
@@ -739,6 +740,15 @@ func (b *Machine) ReadMessage(r io.Reader) ([]byte, error) {
 		return nil, err
 	}
 
+	// If this is the first message being read, take a read buffer from the
+	// buffer pool. This is delayed until this point to avoid allocating
+	// read buffers until after the peer has successfully completed the
+	// handshake, and is ready to begin sending lwnire messages.
```
wpaulino: Nit: s/lwnire/lnwire

brontide/noise.go
```diff
@@ -739,6 +740,15 @@ func (b *Machine) ReadMessage(r io.Reader) ([]byte, error) {
...
+	if b.nextCipherText == nil {
+		b.nextCipherText = readBufferPool.Take()
+		runtime.SetFinalizer(b, freeReadBuffer)
```
wpaulino: 👍

brontide/read_buffer_pool.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Return returns the ReadBuffer to the pool, so that it can be cycled or
+// released.
+func (p *ReadBufferPool) Return(buf *ReadBuffer) {
+	p.pool.Return(buf)
+}
+
+// readBufferPool is a singleton instance of the buffer pool, used to conserve
+// memory allocations due to read buffers across the entire brontide package.
+var readBufferPool = NewReadBufferPool(DefaultGCInterval, DefaultExpiryInterval)
```
wpaulino: Nit: redefine in `noise.go` where it's used?

brontide/conn.go
```diff
@@ -61,7 +61,11 @@ func Dial(localPriv *btcec.PrivateKey, netAddr *lnwire.NetAddress,
 	// We'll ensure that we get ActTwo from the remote peer in a timely
 	// manner. If they don't respond within 1s, then we'll kill the
 	// connection.
-	conn.SetReadDeadline(time.Now().Add(handshakeReadTimeout))
+	err = conn.SetReadDeadline(time.Now().Add(handshakeReadTimeout))
```
Roasbeef: 👍

brontide/read_buffer_pool.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// readBufferSize represents the size of the maximum message that can be
+	// read off  the wire by brontide. The buffer is used to hold the
+	// ciphertext while the  brontide state machine decrypts the message.
+	readBufferSize = lnwire.MaxMessagePayload + macSize
+)
+
+// ReadBuffer is a static byte array sized to the maximum-allowed message size,
+// plus 16 bytes for the MAC.
+type ReadBuffer [readBufferSize]byte
```
Roasbeef: Odd to have a public static array that has a private size. Wonder how this shows up in the `godoc`.

brontide/read_buffer_pool.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Recycle zeroes the ReadBuffer, making it fresh for another use.
```
Roasbeef: Possibly make this a function instead so the block can be re-used rather than copied between contexts.

brontide/read_buffer_pool.go
cfromknecht: i'm not sure we can do that without sacrificing the type safety that this imposes on the . underlying queue. at best, we'd just be pushing this step up another layer

brontide/read_buffer_pool.go
```diff
@@ -0,0 +1,85 @@
...
+func (b *ReadBuffer) Recycle() {
+	b[0] = 0
+	for i := 1; i < readBufferSize; i *= 2 {
+		copy(b[i:], b[:i])
+	}
+}
+
+// newRecyclableReadBuffer is a constructor that returns a ReadBuffer as a
+// queue.Recycler.
+func newRecyclableReadBuffer() queue.Recycler {
```
Roasbeef: Is it necessary to return the interface rather than the struct? If you return the struct/type, then it case be passed into all contexts a  `queue.Recycler` is accepted and also where a regular slice is accepted (once you slice it). As is now, callers will have to type assert it back to a plain buffer.

brontide/read_buffer_pool_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/lightningnetwork/lnd/brontide"
+)
+
+// TestReadBufferPool verifies that the buffer pool properly resets used read
+// buffers.
+func TestReadBufferPool(t *testing.T) {
```
Roasbeef: Should be possible to abstract that other test a bit in order to make it a table driven tests of the `queue.Recycler` implementations.

brontide/noise.go
Roasbeef: Have we confirmed that this is indeed being executed in a timely manner, and isn't left dangling due to a pointer in some struct?

brontide/read_buffer_pool.go
cfromknecht: that's a good point, tho my preference is to keep this in the file that has the same name as the variable :)

cfromknecht:
@wpaulino comments addressed and rebased on master, ptal!

brontide/read_buffer_pool.go
cfromknecht: the gcqueue has to accept some generic `New()` method, the only other thing that makes sense is a `func() interface{}`, otherwise we'd have to build a unique gcqueue for each distinct return type.

In this `func() interface{}` case, then the gcqueue has to perform an expensive interface -> interface conversion to make it a `queue.Recycler`. So there is a performance argument to having it return the `queue.Recycler`directly. Type asserting from interface to element is more efficient so I opted for this

The `queue.Recycler` can also be used for things that are not byte arrays, so idt we want to assume they can always be sliced

brontide/read_buffer_pool_test.go
cfromknecht: this can get become involved with the current layout since they live in different packages.

there could be an argument for structuring the packages like:
```
buffer/
  read.go
  write.go
pool/
  read_buffer.go
  write_buffer.go
```
is that something we want to do here or in a followup?

brontide/noise.go
cfromknecht: when running on my node the profiles didn't hint at this happening. it matched what was being done for the write pools w/o the finalizer

cfromknecht:
@Roasbeef @wpaulino ptal

queue/gc_queue.go
```diff
@@ -15,6 +15,22 @@ type Recycler interface {
 	Recycle()
 }
 
+// RecycleSlice zeroes byte slice, making it fresh for another use.
+// Zeroing the buffer using a logarithmic number of calls to the optimized copy
+// method.  Benchmarking shows this to be ~30 times faster than a for loop that
+// sets each index to 0 for this buffer size. Inspired by:
+// https://stackoverflow.com/questions/30614165/is-there-analog-of-memset-in-go
+func RecycleSlice(b []byte) {
```
wpaulino: Don't see how this fits within the `queue` package, but fine with me.

brontide/read_buffer_pool_test.go
wpaulino: IMO can be done in #2474.

queue/gc_queue.go
cfromknecht: yes that bothered me too, i'm thinking it more appropriately belongs in a `buffer` package. i can add this lone method to that package now, which would unblock #2602. does that sound good?

wpaulino: Yep!

cfromknecht: kk will fix

cfromknecht:
Updated the PR to include a `buffer` and `pool` package. This keeps all recycling out of the queue package, which keeps it more generalized as one would expect from a queue package. Let me know what you think

@roasbeef @wpaulino ptal

brontide/read_buffer_pool.go
cfromknecht: @wpaulino ended up moving this to `noise.go` as the `read_buffer_pool.go` file no longer exists in `brontide` :)

pool/recycle.go
```diff
@@ -0,0 +1,52 @@
+package pool
```
Roasbeef: Nice, now I have a spot to move the existing `lnwallet.SigPool` struct to! Probably something like `pool.Sig`.

pool/recycle_test.go
```diff
@@ -0,0 +1,193 @@
...
+	switch pp := p.(type) {
+	case *pool.WriteBuffer:
+		pp.Return(item.(*buffer.Write))
+
+	default:
+		t.Fatalf("unknown pool type: %T", p)
+	}
+}
+
+func dirtyGeneric(t *testing.T, i interface{}) {
```
Roasbeef: `interface{}`, a necessary evil...

queue/gc_queue.go
```diff
@@ -36,15 +21,15 @@ type gcQueueEntry struct {
 type GCQueue struct {
 	// takeBuffer coordinates the delivery of items taken from the queue
 	// such that they are delivered to requesters.
-	takeBuffer chan Recycler
+	takeBuffer chan interface{}
```
Roasbeef: Not a blocker, but what's the rationale in downgrading to an `interface{}` (was the `Recycler` struct before) ? I'm guessing in order to be able to consolidate the tests? Even just having be interface be specified in the file makes things a bit more readable IMO.

pool/recycle_test.go
```diff
@@ -71,6 +75,14 @@ func TestConcreteRecyclePoolTests(t *testing.T) {
 				)
 			},
 		},
+		{
+			name: "read buffer pool",
+			newPool: func() interface{} {
```
Roasbeef: Ah, required to abstract the buffer/pool itself atm.

PR #2247, opened 2018-11-30
rpcserver: forward gRPC proxy requests to localhost when listening on all interfaces

wpaulino:
This prevents certificate issues when accessing the gRPC REST proxy externally.

Fixes #1567.

PR #2428, opened 2019-01-07
lnrpc: let's encrypt certificates

joostjager:
This commit enables lnd to request and renew a Let's Encrypt certificate. This certificate is used both for the grpc as well as the rest listeners. It allows clients to connect without having a copy of the (public) server certificate.

In addition to this, it fixes a bug where wildcard ip instantiation (introduced in https://github.com/lightningnetwork/lnd/pull/2247) wasn't carried out for the wallet unlocker service.

## Configuration

Most important: the `lnd` instance must be reachable via a domain name. Let's Encrypt certificates are tied to domain names.

New flags added to the `lnd` command line and `lnd.conf`:

* `--letsencryptport`
The port on which lnd will listen for Let's Encrypt challenges. Let's Encrypt will always try to contact on port 80. Often non-root processes are not allowed to bind to ports lower than 1024. This configuration option allows a different port to be used, but must be used in combination with port forwarding from port 80.

* `--letsencryptdir`
The directory to store Let's Encrypt certificates within. By default this is `.lnd/letsencrypt`.

* `--letsencryptdomain`
Request a Let's Encrypt certificate for the domain specified using this flag.

The `lnd` instance must be reachable from outside on port 80. Either directly or through port forwarding to `letsencryptport`.

## lncli

When `lncli` cannot find a `tls.cert` file, it will assume the server has a valid (Let's Encrypt) certificate. It is important to pass the domain name as a command line flag to `lncli`:

`lncli --rpcserver my.domain.org:10009`

This is necessary as well when connecting to `localhost`.

### DIFF
--- a/cmd/lncli/profile.go
+++ b/cmd/lncli/profile.go
@@ -36,6 +36,10 @@ type profileEntry struct {
 
 // cert returns the profile's TLS certificate as a x509 certificate pool.
 func (e *profileEntry) cert() (*x509.CertPool, error) {
+	if e.TLSCert == "" {
+		return nil, nil
+	}
+
 	cp := x509.NewCertPool()
 	if !cp.AppendCertsFromPEM([]byte(e.TLSCert)) {
 		return nil, fmt.Errorf("credentials: failed to append " +
@@ -113,11 +117,16 @@ func profileFromContext(ctx *cli.Context, store bool) (*profileEntry, error) {
 		return nil, err
 	}
 
-	// Load the certificate file now. We store it as plain PEM directly.
-	tlsCert, err := ioutil.ReadFile(tlsCertPath)
-	if err != nil {
-		return nil, fmt.Errorf("could not load TLS cert file %s: %v",
-			tlsCertPath, err)
+	// Load the certificate file now, if specified. We store it as plain PEM
+	// directly.
+	var tlsCert []byte
+	if lnrpc.FileExists(tlsCertPath) {
+		var err error
+		tlsCert, err = ioutil.ReadFile(tlsCertPath)
+		if err != nil {
+			return nil, fmt.Errorf("could not load TLS cert file "+
+				"%s: %v", tlsCertPath, err)
+		}
 	}
 
 	// Now load and possibly encrypt the macaroon file.
--- a/config.go
+++ b/config.go
@@ -64,6 +64,8 @@ const (
 	defaultMaxLogFileSize                = 10
 	defaultMinBackoff                    = time.Second
 	defaultMaxBackoff                    = time.Hour
+	defaultLetsEncryptDirname            = "letsencrypt"
+	defaultLetsEncryptPort               = 80
 
 	defaultTorSOCKSPort            = 9050
 	defaultTorDNSHost              = "soa.nodes.lightning.directory"
@@ -127,8 +129,9 @@ var (
 
 	defaultTowerDir = filepath.Join(defaultDataDir, defaultTowerSubDirname)
 
-	defaultTLSCertPath = filepath.Join(DefaultLndDir, defaultTLSCertFilename)
-	defaultTLSKeyPath  = filepath.Join(DefaultLndDir, defaultTLSKeyFilename)
+	defaultTLSCertPath    = filepath.Join(DefaultLndDir, defaultTLSCertFilename)
+	defaultTLSKeyPath     = filepath.Join(DefaultLndDir, defaultTLSKeyFilename)
+	defaultLetsEncryptDir = filepath.Join(DefaultLndDir, defaultLetsEncryptDirname)
 
 	defaultBtcdDir         = btcutil.AppDataDir("btcd", false)
 	defaultBtcdRPCCertFile = filepath.Join(defaultBtcdDir, "rpc.cert")
@@ -178,6 +181,10 @@ type Config struct {
 	MaxLogFiles     int           `long:"maxlogfiles" description:"Maximum logfiles to keep (0 for no rotation)"`
 	MaxLogFileSize  int           `long:"maxlogfilesize" description:"Maximum logfile size in MB"`
 	AcceptorTimeout time.Duration `long:"acceptortimeout" description:"Time after which an RPCAcceptor will time out and return false if it hasn't yet received a response"`
+
+	LetsEncryptDir    string `long:"letsencryptdir" description:"The directory to store Let's Encrypt certificates within"`
+	LetsEncryptPort   int    `long:"letsencryptport" description:"The port on which lnd will listen for Let's Encrypt challenges. Let's Encrypt will always try to contact on port 80. Often non-root processes are not allowed to bind to ports lower than 1024. This configuration option allows a different port to be used, but must be used in combination with port forwarding from port 80."`
+	LetsEncryptDomain string `long:"letsencryptdomain" description:"Request a Let's Encrypt certificate for this domain. Note that the certicate is only requested and stored when the first rpc connection comes in."`
 
 	// We'll parse these 'raw' string arguments into real net.Addrs in the
 	// loadConfig function. We need to expose the 'raw' strings so the
@@ -318,6 +325,8 @@ func DefaultConfig() Config {
 		DebugLevel:      defaultLogLevel,
 		TLSCertPath:     defaultTLSCertPath,
 		TLSKeyPath:      defaultTLSKeyPath,
+		LetsEncryptDir:  defaultLetsEncryptDir,
+		LetsEncryptPort: defaultLetsEncryptPort,
 		LogDir:          defaultLogDir,
 		MaxLogFiles:     defaultMaxLogFiles,
 		MaxLogFileSize:  defaultMaxLogFileSize,
@@ -520,6 +529,9 @@ func ValidateConfig(cfg Config, usageMessage string) (*Config, error) {
 	lndDir := CleanAndExpandPath(cfg.LndDir)
 	if lndDir != DefaultLndDir {
 		cfg.DataDir = filepath.Join(lndDir, defaultDataDirname)
+		cfg.LetsEncryptDir = filepath.Join(
+			lndDir, defaultLetsEncryptDirname,
+		)
 		cfg.TLSCertPath = filepath.Join(lndDir, defaultTLSCertFilename)
 		cfg.TLSKeyPath = filepath.Join(lndDir, defaultTLSKeyFilename)
 		cfg.LogDir = filepath.Join(lndDir, defaultLogDirname)
@@ -558,6 +570,7 @@ func ValidateConfig(cfg Config, usageMessage string) (*Config, error) {
 	cfg.DataDir = CleanAndExpandPath(cfg.DataDir)
 	cfg.TLSCertPath = CleanAndExpandPath(cfg.TLSCertPath)
 	cfg.TLSKeyPath = CleanAndExpandPath(cfg.TLSKeyPath)
+	cfg.LetsEncryptDir = CleanAndExpandPath(cfg.LetsEncryptDir)
 	cfg.AdminMacPath = CleanAndExpandPath(cfg.AdminMacPath)
 	cfg.ReadMacPath = CleanAndExpandPath(cfg.ReadMacPath)
 	cfg.InvoiceMacPath = CleanAndExpandPath(cfg.InvoiceMacPath)
--- a/lnd.go
+++ b/lnd.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	_ "net/http/pprof" // Blank import to set up profiling HTTP handlers.
 	"os"
 	"path/filepath"
 	"runtime/pprof"
@@ -18,19 +19,14 @@ import (
 	"sync"
 	"time"
 
-	// Blank import to set up profiling HTTP handlers.
-	_ "net/http/pprof"
-
-	"gopkg.in/macaroon-bakery.v2/bakery"
-	"gopkg.in/macaroon.v2"
-
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
-
 	"github.com/btcsuite/btcutil"
 	"github.com/btcsuite/btcwallet/wallet"
 	proxy "github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/lightninglabs/neutrino"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+	"gopkg.in/macaroon.v2"
 
 	"github.com/lightningnetwork/lnd/autopilot"
 	"github.com/lightningnetwork/lnd/build"
--- a/lnd.go
+++ b/lnd.go
@@ -23,6 +23,7 @@ import (
 	"github.com/btcsuite/btcwallet/wallet"
 	proxy "github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/lightninglabs/neutrino"
+	"golang.org/x/crypto/acme/autocert"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"gopkg.in/macaroon-bakery.v2/bakery"
@@ -264,12 +265,14 @@ func Main(cfg *Config, lisCfg ListenerCfg, shutdownChan <-chan struct{}) error {
 	defer cleanUp()
 
 	// Only process macaroons if --no-macaroons isn't set.
-	tlsCfg, restCreds, restProxyDest, err := getTLSConfig(cfg)
+	tlsCfg, restCreds, restProxyDest, cleanUp, err := getTLSConfig(cfg)
 	if err != nil {
 		err := fmt.Errorf("unable to load TLS credentials: %v", err)
 		ltndLog.Error(err)
 		return err
 	}
+
+	defer cleanUp()
 
 	serverCreds := credentials.NewTLS(tlsCfg)
 	serverOpts := []grpc.ServerOption{grpc.Creds(serverCreds)}
@@ -748,7 +751,7 @@ func Main(cfg *Config, lisCfg ListenerCfg, shutdownChan <-chan struct{}) error {
 // getTLSConfig returns a TLS configuration for the gRPC server and credentials
 // and a proxy destination for the REST reverse proxy.
 func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
-	string, error) {
+	string, func(), error) {
 
 	// Ensure we create TLS key and certificate if they don't exist.
 	if !fileExists(cfg.TLSCertPath) && !fileExists(cfg.TLSKeyPath) {
@@ -759,7 +762,7 @@ func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
 			cfg.TLSDisableAutofill, cert.DefaultAutogenValidity,
 		)
 		if err != nil {
-			return nil, nil, "", err
+			return nil, nil, "", nil, err
 		}
 		rpcsLog.Infof("Done generating TLS certificates")
 	}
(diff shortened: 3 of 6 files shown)


joostjager:
Difficult implementation detail: the reverse proxy connects to the grpc listener and needs to use the correct domain name. Otherwise there is no matching certificate in the grpc listener.

joostjager:
Rebased after merge of #2247

vegardengen:
I think it might become confusing if you can't any more connect to localhost when operating on the server itself.

The following patch implementes the following changes:

- Create self-signed cert even if we are doing LetsEncrypt
- Adding the self-signed cert as a fallback and return it if there was no hostname match in the Letsencrypt-certificate
- Instantiate the internal truststore in lncli from system CAs anyhow, but add the self-signed tls.cert if it is found. 

This way, you can choose whether you use localhost, IP-adresses or the hostname(s) in the Letsencrypt-generated certificate, LND will return the correct certificate, and lncli will trust the self-signed if it can find the tls.cert in ~/.lnd/tls.cert


[letsencrypt-changes.txt](https://github.com/lightningnetwork/lnd/files/2799587/letsencrypt-changes.txt)

vegardengen:
Regarding my last comment: It also solves the problem that might arise if the IP-adress in tlsextradomain is not on the machine itself, but outside of a NATing firewall.

Should we have a letsencryptdomains separate from the tlsextradomains ? It could be that there are domains in tlsextradomains that can't be managed by letsencrypt because they are not known in public DNS, so you want to add them in the self-signed certificate instead.

joostjager:
>Should we have a letsencryptdomains separate from the tlsextradomains ? It could be that there are domains in tlsextradomains that can't be managed by letsencrypt because they are not known in public DNS, so you want to add them in the self-signed certificate instead.

Yes, I was in doubt about this. But now that we can have both certificates active at the same time, I think it should indeed be a different flag.

joostjager:
@vegardengen I tacked on another commit that implements your suggestion on setting the domain. Removed the `letsencrypt` bool, because it was actually redundant.

joostjager:
If we are happy with what we have here, I will restructure the commits to get this ready for merge.

vegardengen:
We *could*, theoretically, want to add more than one letsencryptdomain...couldn't we?

joostjager:
Yes, we could split the `letsencryptdomain` field by comma for example, but I'd think we capture 95% of the use cases already now. Any patches welcome though ;)

vegardengen:
Take a look at how it is done with tlsextradomain?


Sendt fra min Samsung Galaxy-smarttelefon.
-------- Opprinnelig melding --------Fra: Joost Jager <notifications@github.com> Dato: 28.01.2019  23:21  (GMT+01:00) Til: lightningnetwork/lnd <lnd@noreply.github.com> Ko: vegardengen <vegard@engen.priv.no>, Mention <mention@noreply.github.com> Emne: Re: [lightningnetwork/lnd] lnrpc: let's encrypt certificates (#2428) 
Yes, we could split the letsencryptdomain field by comma for example, but I'd think we capture 95% of the use cases already now. Any patches welcome though ;)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/lightningnetwork/lnd","title":"lightningnetwork/lnd","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/lightningnetwork/lnd"}},"updates":{"snippets":[{"icon":"PERSON","message":"@joostjager in #2428: Yes, we could split the `letsencryptdomain` field by comma for example, but I'd think we capture 95% of the use cases already now. Any patches welcome though ;)"}],"action":{"name":"View Pull Request","url":"https://github.com/lightningnetwork/lnd/pull/2428#issuecomment-458325504"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/lightningnetwork/lnd/pull/2428#issuecomment-458325504",
"url": "https://github.com/lightningnetwork/lnd/pull/2428#issuecomment-458325504",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

joostjager:
I know how to do it, just thought you might like it :wink:

joostjager:
Commit structure updated. Decided on not supporting multiple Let's Encrypt domains in this PR.

lnd.go
```diff
@@ -26,6 +26,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/crypto/acme/autocert"
```
halseth: import can be move together with other `golang.org` imports

lnd.go
```diff
@@ -417,10 +419,61 @@ func getTLSConfig(cfg *config) (*tls.Config, *credentials.TransportCredentials,
...
 	}
 
-	tlsConf = &tls.Config{
-		Certificates: []tls.Certificate{cert},
-		CipherSuites: tlsCipherSuites,
-		MinVersion:   tls.VersionTLS12,
+	// If Let's Encrypt is enabled, instantiate autocert to request/renew
+	// the certificates.
+	if cfg.LetsEncryptDomain != "" {
+		ltndLog.Infof(
```
halseth: nit: wrapping not needed

lnd.go
```diff
@@ -417,10 +419,61 @@ func getTLSConfig(cfg *config) (*tls.Config, *credentials.TransportCredentials,
...
+		)
+		manager := autocert.Manager{
+			Cache:      autocert.DirCache(cfg.LetsEncryptDir),
+			Prompt:     autocert.AcceptTOS,
+			HostPolicy: autocert.HostWhitelist(cfg.LetsEncryptDomain),
+		}
+		go func() {
+			addr := fmt.Sprintf(":%v", cfg.LetsEncryptPort)
+
+			ltndLog.Infof(
```
halseth: nit: wrapping

lnd.go
```diff
@@ -417,10 +419,61 @@ func getTLSConfig(cfg *config) (*tls.Config, *credentials.TransportCredentials,
...
+			err := http.ListenAndServe(
+				addr,
+				manager.HTTPHandler(nil),
+			)
+			if err != nil {
+				ltndLog.Errorf("autocert http: %v", err)
+			}
+		}()
+
+		getCertificate := func(h *tls.ClientHelloInfo) (
```
halseth: can function be declared directly in the returned config? If not too messy.

joostjager: tried, but didn't like it too much

rpcserver.go
```diff
@@ -476,6 +479,7 @@ func newRPCServer(s *server, macService *macaroons.Service,
 		grpcServer:     grpcServer,
 		server:         s,
 		quit:           make(chan struct{}, 1),
+		proxyDest:      proxyDest,
```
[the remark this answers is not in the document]
joostjager: The thing is that a bug was also fixed en passant. For the wallet unlocker service, the translation to a real ip to connect to didn't happen.

If I'd make this function, I would have to call it twice.

joostjager:
Processed comments. Renamed a few variables in the first commit to make their purpose clearer.

@halseth ptal

rpcserver.go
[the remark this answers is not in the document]
halseth: Yes, we would call the method twice, but we would avoid having to add more fields to the rpcServer and method signature. So it is a style preference I guess :)

config.go
```diff
@@ -185,6 +188,10 @@ type config struct {
 	MaxLogFiles    int    `long:"maxlogfiles" description:"Maximum logfiles to keep (0 for no rotation)"`
 	MaxLogFileSize int    `long:"maxlogfilesize" description:"Maximum logfile size in MB"`
 
+	LetsEncryptDir    string `long:"letsencryptdir" description:"The directory to store Let's Encrypt certificates within"`
```
halseth: The config is starting to become quite big, what do you think about grouping auth related settings in their own namespace? :)

lnd.go
halseth: Not fixed?

lnd.go
halseth: Not fixed?

lnd.go
halseth: Not fixed?

lnd.go
```diff
@@ -412,6 +398,45 @@ func lndMain() error {
 	return nil
 }
 
+// getTLSConfig returns a tls configuration struct for the grpc server and
+// credentials and a proxy destination for the rest reverse proxy.
+func getTLSConfig(cfg *config) (*tls.Config, *credentials.TransportCredentials,
+	string, error) {
+	var tlsConf *tls.Config
```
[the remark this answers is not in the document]
halseth: not fixed?

config.go
joostjager: I think it is a good idea, but not sure if it is considered a breaking change? @Roasbeef

joostjager:
I am pretty sure I fixed those 'not fixeds'. Something must have gone wrong somewhere. Did it again.

joostjager:
@halseth ptal

lnd.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	"golang.org/x/net/context"
 
+	"golang.org/x/crypto/acme/autocert"
```
halseth: nit: could be grouped with above import. tbh I don't know if there is any structure behind the imports in this file 😛

joostjager: We definitely need an import robot. Or alternatively don't care about import order in reviews (my pref) and just touch up whenever you feel like it.

Made three groups now in this file: base libs, other libs and lnd libs.

config.go
cfromknecht: @halseth good point, this might be a good time to start taking advantage of the `lncfg` package. add a file like `lncfg/lets_encrypt.go` with `lncfg.LetsEncrypt` config struct. could at least start by having let's encrypt configs not directly on the main config?

cfromknecht: the namespacing works out quite nicely i think, `--letsencrypt.dir=xxx`, etc

joostjager: created the group `letsencrypt`

joostjager:
@cfromknecht ptal

halseth:
Can be closed?

joostjager:
I still want it. It is a small diff to get rid of passing around the `tls.cert` file to every `lnd` client tool/application. The workaround is setting up a proxy which is not as simple as adding the `letsencrypt` flag that this pr introduces. There are 5 thumbs up on the pr...

joostjager:
2020 rebase executed

cmd/lncli/profile.go
```diff
@@ -36,8 +37,16 @@ type profileEntry struct {
 
 // cert returns the profile's TLS certificate as a x509 certificate pool.
 func (e *profileEntry) cert() (*x509.CertPool, error) {
-	cp := x509.NewCertPool()
-	if !cp.AppendCertsFromPEM([]byte(e.TLSCert)) {
+	// Retrieve the system certificate pool. If not available (Windows),
+	// create a new pool.
+	cp, _ := x509.SystemCertPool()
```
guggero, line 42: Apparently there's a workaround for the Windows problem: https://github.com/golang/go/issues/16736#issuecomment-540373689. Do you think it would be worth implementing that?
Also see my comment in #4549 about this.

cmd/lncli/profile.go
```diff
@@ -234,3 +248,13 @@ func (f *profileFile) marshalJSON() ([]byte, error) {
 	out.WriteString("\n")
 	return out.Bytes(), nil
 }
+
+// fileExists reports whether the named file or directory exists.
+func fileExists(name string) bool {
```
guggero, line 253: nit: use `lnrpc.FileExists`.

lnd.go
guggero: See https://github.com/lightninglabs/faraday/pull/74, maybe this is something we'd want to use in `lnd` too?

lnd.go
```diff
@@ -827,6 +828,48 @@ func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
...
+			HostPolicy: autocert.HostWhitelist(cfg.LetsEncryptDomain),
+		}
+		go func() {
+			addr := fmt.Sprintf(":%v", cfg.LetsEncryptPort)
+
+			ltndLog.Infof("Autocert challenge listener started "+
+				"at %v", addr)
+			err := http.ListenAndServe(
+				addr,
+				manager.HTTPHandler(nil),
```
guggero, line 850: nit: can merge two lines.

cmd/lncli/profile.go
guggero: Maybe the question is: Do we need to provide a cert pool to `gRPC` if the cert is valid in the first place? Maybe it will fall back to the system CA store by default and we've been looking at this the wrong way all along?

lnd.go
```diff
@@ -827,6 +828,48 @@ func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
...
+	// If Let's Encrypt is enabled, instantiate autocert to request/renew
+	// the certificates.
+	if cfg.LetsEncryptDomain != "" {
+		ltndLog.Infof("Using Let's Encrypt certificate for domain %v",
+			cfg.LetsEncryptDomain)
+
+		manager := autocert.Manager{
+			Cache:      autocert.DirCache(cfg.LetsEncryptDir),
+			Prompt:     autocert.AcceptTOS,
+			HostPolicy: autocert.HostWhitelist(cfg.LetsEncryptDomain),
```
guggero, line 841: The issue with the `autocert` manager is that the certificate is only really requested/generated once the first outside request for one of the white listed domains comes in. This means the user might expect the certificate to be created on startup as before and nothing will happen. Not sure if we should try to force a request here or if we should at least document the behavior.

sample-lnd.conf
```diff
@@ -50,6 +50,19 @@
 ; or want to expose the node at a domain.
 ; externalhosts=my-node-domain.com
 
+; Sets the directory to store Let's Encrypt certificates within
+; letsencryptdir=
```
guggero, line 54: nit: I think we started to provide valid/sane example values for all options in this file.

guggero (commented):
Great functionality, I'd really like to see this being added to `lnd`!

lnd.go
joostjager: But isn't this completely transparent to the user? Just as if the cert would be expired and a new one requested invisibly?

guggero: Yes, I guess it would be. But I know we ran into the issue with `aperture` where we thought something didn't work because the certificate wasn't created (automatically). So we should maybe at least mention it somewhere?

cmd/lncli/profile.go
joostjager: I tried with a new pool, but it doesn't work:
`[lncli] rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority"`

I don't know if the workaround is worth implementing. I think this PR is also valuable without it, so rather do it in a follow-up.

cmd/lncli/profile.go
joostjager: dibe

lnd.go
joostjager: Possibly, but not in this PR :)

lnd.go
joostjager: Added to description of the `letsencryptdomain` flag.

sample-lnd.conf
joostjager: added

rpcserver.go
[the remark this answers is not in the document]
joostjager: This comment isn't applicable anymore. Much has changed since Feb 2019

lnd.go
guggero: No, of course not in this PR. But if you agree with the proposed idea, you could order the imports that way already, just an idea.

cmd/lncli/profile.go
guggero: Okay, I agree with not implementing the workaround in this PR.

I wasn't really clear about what I meant with "no cert pool". I meant, if instead of creating an empty pool or a pool with the system cert pool, what happens if you don't add the `grpc.WithTransportCredentials(creds)` at all? Does `gRPC` then fall back to the system pool? If so, we don't need `x509.SystemCertPool()`, as we don't really care about its content, just it being used.

sample-lnd.conf
```diff
@@ -50,6 +50,19 @@
...
+; letsencryptdir=~/.lnd/letsencrypt
+
+; Sets the port on which lnd will listen for Let's Encrypt challenges. Let's 
+; Encrypt will always try to contact on port 80. Often non-root processes are
+; not allowed to bind to ports lower than 1024. This configuration option allows
+; a different port to be used, but must be used in combination with port
+; forwarding from port 80.
+; letsencryptport=8080
+
+; Request a Let's Encrypt certificate for this domain
```
guggero, line 63: nit: update this text too.

cmd/lncli/profile.go
joostjager: No, removing `grpc.WithTransportCredentials(creds)` doesn't work:

`[lncli] unable to connect to RPC server: grpc: no transport security set (use grpc.WithInsecure() explicitly or set credentials)`

lnd.go
joostjager: Yes, I agree with the idea. Added commit.

sample-lnd.conf
joostjager: Added

lnd.go
```diff
@@ -11,26 +11,22 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	_ "net/http/pprof" // Blank import to set up profiling HTTP handlers.
```
joostjager, line 14: Can't keep the comment on a separate line, because auto-formatting will add a new line, which makes `impi` disapprove.

cmd/lncli/profile.go
guggero: Apparently using `credentials.NewTLS(&tls.Config{})` instead of loading any certificates or pools causes `gRPC` to fall back to the system cert pool. I tested this with Windows as well and it seems to work.

joostjager: Great find. Updated

guggero (approved):
LGTM, excited to have two TLS related issues fixed with the same PR!

Tested the CLI part, wasn't able to set up the daemon part with Let's Encrypt myself.

lnd.go
```diff
@@ -823,6 +824,47 @@ func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
...
+	if cfg.LetsEncryptDomain != "" {
+		ltndLog.Infof("Using Let's Encrypt certificate for domain %v",
+			cfg.LetsEncryptDomain)
+
+		manager := autocert.Manager{
+			Cache:      autocert.DirCache(cfg.LetsEncryptDir),
+			Prompt:     autocert.AcceptTOS,
+			HostPolicy: autocert.HostWhitelist(cfg.LetsEncryptDomain),
+		}
+		go func() {
```
halseth, line 839: should have a wait group

cmd/lncli/profile.go
```diff
@@ -36,6 +36,10 @@ type profileEntry struct {
 
 // cert returns the profile's TLS certificate as a x509 certificate pool.
 func (e *profileEntry) cert() (*x509.CertPool, error) {
+	if e.TLSCert == "" {
+		return nil, nil
```
halseth, line 40: Define some error `ErrTLSNotSet` instead?

joostjager: I think the `nil` is fine too. There is no pool, so we return `nil`.

halseth (approved):

lnd.go
joostjager: made it wait, but in a slightly different way.

PR #2430, opened 2019-01-08
lnwallet: Remove breach txn from NewBreachRetribution

cfromknecht:
In this PR, we remove the passed `breachTxn` argument from `NewBreachRetribution`. We already store the transaction on disk, which we retrieve via the state number. Currently, we pull values from both the argument and the transaction loaded from disk, which is unnecessary since they should both be the same.

We modify this method to only inspect the on-disk transaction, as this allows `NewBreachRetribution` to be queried for a specific state number w/o needing to have the breach transaction itself.

This is a preliminary step to adding the watchtower client, which may need to be able to retroactively sign justice transactions for any particular revoked state. This also implies that the client/links don't have to do any bookkeeping/recording of breach transactions outside the wallet.

### DIFF
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -1897,10 +1897,7 @@ type BreachRetribution struct {
 // passed channel, at a particular revoked state number, and one which targets
 // the passed commitment transaction.
 func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
-	broadcastCommitment *wire.MsgTx,
 	breachHeight uint32) (*BreachRetribution, error) {
-
-	commitHash := broadcastCommitment.TxHash()
 
 	// Query the on-disk revocation log for the snapshot which was recorded
 	// at this particular state num.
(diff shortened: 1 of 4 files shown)


cfromknecht:
rebased and squashed

lnwallet/channel.go
```diff
@@ -1952,7 +1951,7 @@ func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
 	remoteOutpoint := wire.OutPoint{
 		Hash: commitHash,
 	}
-	for i, txOut := range broadcastCommitment.TxOut {
+	for i, txOut := range revokedSnapshot.CommitTx.TxOut {
```
halseth: It is surprising to me that we store the whole remote commit tx for each state num! Has this always been the case?

Seems a bit wasteful and not necessary?

cfromknecht: We considered removing them at one point and then didn't iirc. Turns out they're useful! 

> Seems a bit wasteful and not necessary?

Never hurts to keep channel data, deleting is forever. if it becomes an issue we should reconsider it, but i don't think that'll be any time soon :)

halseth: True, must leave some low-hanging optimizations around in case we run out of work 🤣

PR #2432, opened 2019-01-08
lnwire: Adjust Stringer to standard chan id format

alexbosworth:
This would adjust logging output to use a more easily selectable short channel id components string

### DIFF
--- a/lnwire/short_channel_id.go
+++ b/lnwire/short_channel_id.go
@@ -44,5 +44,5 @@ func (c ShortChannelID) ToUint64() uint64 {
 
 // String generates a human-readable representation of the channel ID.
 func (c ShortChannelID) String() string {
-	return fmt.Sprintf("%d:%d:%d", c.BlockHeight, c.TxIndex, c.TxPosition)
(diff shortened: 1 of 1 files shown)


Roasbeef:
Would say we should wait on this until we update lncli to accept this format and convert it to the regular format. We can even just keep our RPCs as is for time being, and expect that lncli or the client is able to convert the short channel ID properly.

guggero:
What's the dependency for this? What needs to be changed in `lncli` to get this in?

Roasbeef:
> What's the dependency for this? What needs to be changed in lncli to get this in?

I don't think anything does? Given that `lncli` always accepts the integer version since it's generally easier to copy/paste. We could also log the integer version in parens or something in this new stringer.

cfromknecht:
> What's the dependency for this? What needs to be changed in lncli to get this in?

Ideally all of our commands in `lncli` where a channelid is accepted we also accept the new format. Otherwise users will need to convert back into the integer format, e.g. when trying to use `getchaninfo`.

PR #2436, opened 2019-01-08
Check max HTLC when forwarding payments.

### LINKED ISSUE #2182: lnwire+funding+routing: populate, implement and adhere to the new max_htlc ChannelUpdate value

valentinewallace:
Depends on #2434 (which depends on #2296). Partially closes #2182.

In our previous PR, we implement parsing, storing and re-sending the new max HTLC field that has been added to ChannelUpdates.

In this PR, we verify an HTLC does not exceed the max HTLC for a given link before forwarding it.


Roasbeef:
Needs a rebase.

valentinewallace:
Updated [pending travis]! @Roasbeef

htlcswitch/link_test.go
```diff
@@ -770,6 +770,69 @@ func TestLinkForwardMinHTLCPolicyMismatch(t *testing.T) {
...
 
+// TestLinkForwardMaxHTLCPolicyMismatch tests that if a node is an intermediate
+// node and receives an HTLC which is _above_ its max HTLC policy then the
+// HTLC will be rejected.
+func TestLinkForwardMaxHTLCPolicyMismatch(t *testing.T) {
+	t.Parallel()
+
+	channels, cleanUp, _, err := createClusterChannels(
+		btcutil.SatoshiPerBitcoin*5,
+		btcutil.SatoshiPerBitcoin*5)
```
wpaulino: Nit: Args should be on their own lines if they can't fit, e.g.:
```
createClusterChannels(
    btcutil.SatoshiPerBitcoin*5, btcutil.SatoshiPerBitcoin*5,
)
```
There are a couple others like this throughout the commit.

valentinewallace: Fixed, although the commit has moved to #2434. Closing this PR now :)

PR #2437, opened 2019-01-08
Consider max HTLC during pathfinding.

### LINKED ISSUE #2182: lnwire+funding+routing: populate, implement and adhere to the new max_htlc ChannelUpdate value

valentinewallace:
Depends on #2434 (which depends on #2296). Partially closes #2182.

In our previous PR, we implement parsing, storing and re-sending the new max HTLC field that has been added to ChannelUpdates.

In this PR, we verify an HTLC does not exceed the max HTLC for a given link during pathfinding.


Roasbeef:
Needs a rebase when free.

Roasbeef:
Ping x3.

valentinewallace:
@Roasbeef rebased!

routing/pathfind_test.go
```diff
@@ -1443,6 +1443,87 @@ func TestRouteFailMinHTLC(t *testing.T) {
 	}
 }
 
+// TestRouteFailMaxHTLC tests that if we attempt to route an HTLC which is
+// larger than the advertised max HTLC of an edge, then path finding fails.
+func TestRouteFailMaxHTLC(t *testing.T) {
```
wpaulino: This commit can be squashed with the one that introduces the pathfinding check.

routing/pathfind_test.go
```diff
@@ -1443,6 +1443,87 @@ func TestRouteFailMinHTLC(t *testing.T) {
...
+	_, _, elstEdge, err := graph.graph.FetchChannelEdgesByID(15433)
+	elstEdge.MessageFlags = 1
+	elstEdge.MaxHTLC = 100000
+	if err := graph.graph.UpdateEdgePolicy(elstEdge); err != nil {
+		t.Fatalf("unable to update edge: %v", err)
+	}
+
+	// Now, if we attempt to route through that edge with a payment above 100k
+	// msat, we should get a faiure.
+	payAmt = lnwire.MilliSatoshi(100001)
```
wpaulino: Nit: Rather than redefining these, keep `payAmt` constant and work with that value instead. So in the `findPath` call below, you'd use `payAmt+1`, etc.

routing/pathfind_test.go
```diff
@@ -1443,6 +1443,87 @@ func TestRouteFailMinHTLC(t *testing.T) {
 	}
 }
 
+// TestRouteFailMaxHTLC tests that if we attempt to route an HTLC which is
+// larger than the advertised max HTLC of an edge, then path finding fails.
+func TestRouteFailMaxHTLC(t *testing.T) {
+	t.Parallel()
+
+	graph, err := parseTestGraph(basicGraphFilePath)
```
wpaulino: I believe we now use `symmetricTestChannel` for new pathfinding tests added to set up  a test graph rather than relying on the test graph file. You can take a look at some of the other tests for examples on how to use it.

routing/pathfind_test.go
valentinewallace: fixed

routing/pathfind_test.go
valentinewallace: fixed

routing/pathfind_test.go
valentinewallace: fixed

Roasbeef:
Needs a rebase now with #2434 in!

PR #2438, opened 2019-01-08
Enable users to update the  max HTLC for a given channel(s).

### LINKED ISSUE #2182: lnwire+funding+routing: populate, implement and adhere to the new max_htlc ChannelUpdate value

valentinewallace:
Depends on #2437 (which depends on #2434, which depends on #2296). Partially closes #2182.

In the previous PRs, we implement parsing, storing and re-sending the new max HTLC field that has been added to ChannelUpdates. We also consider the max HTLC during pathfinding.

In this PR, we enable users to update the max HTLC for a given channel or channels.

Edit 08/6/2019: currently it's a mandatory field for `UpdateChannelPolicy`, gonna edit to make it optional.
Edit 2: done

Edit 8/30/19: depends on #3439 to resolve [this comment](https://github.com/lightningnetwork/lnd/pull/2438#discussion_r316519137) and similar.


Roasbeef:
Needs rebase.

rpcserver.go
```diff
@@ -4512,16 +4512,65 @@ func (r *rpcServer) FeeReport(ctx context.Context,
...
 
+// checkMaxHtlc errors if the user tries to update their channel policy with
+// a max HTLC value that is too high or too low for a given list of channels.
+func checkMaxHtlc(maxHtlc uint64, channels []*channeldb.OpenChannel) error {
+	var (
+		minSupportedHtlc lnwire.MilliSatoshi
+		maxSupportedHtlc = lnwire.MilliSatoshi(math.MaxUint64)
+	)
+
+	for _, ch := range channels {
```
wpaulino: I think we should be applying the constraint to each channel individually. If there are two channels with `min_htlc`s of 4 and 6 and we're proposing a `max_htlc` of 5, we should not allow this `max_htlc` for the the channel with the higher `min_htlc`, while the current diff allows it. Doing so will also allow us to include the channel point in the error as well.

routing/router.go
```diff
@@ -214,6 +214,10 @@ type ChannelPolicy struct {
 	// TimeLockDelta is the required HTLC timelock delta to be used
 	// when forwarding payments.
 	TimeLockDelta uint32
+
+	// MaxHTLC is the maximum HTLC size including fees we are allowed to forward
```
wpaulino: Nit: line too long.

discovery/gossiper.go
```diff
@@ -1277,6 +1277,11 @@ func (d *AuthenticatedGossiper) processChanPolicyUpdate(
 		)
 		edge.TimeLockDelta = uint16(policyUpdate.newSchema.TimeLockDelta)
 
+		// MaxHTLC is optional in channel policy update requests.
+		if policyUpdate.newSchema.MaxHTLC != 0 {
+			edge.MaxHTLC = policyUpdate.newSchema.MaxHTLC
```
wpaulino: We should add/extend a test case for this.

htlcswitch/link.go
```diff
@@ -2171,6 +2171,9 @@ func (l *channelLink) UpdateForwardingPolicy(newPolicy ForwardingPolicy) {
 	if newPolicy.MinHTLC != 0 {
 		l.cfg.FwrdingPolicy.MinHTLC = newPolicy.MinHTLC
 	}
+	if newPolicy.MaxHTLC != 0 {
+		l.cfg.FwrdingPolicy.MaxHTLC = newPolicy.MaxHTLC
```
wpaulino: Same here w.r.t. adding a test case.

cmd/lncli/commands.go
```diff
@@ -3478,6 +3485,19 @@ func updateChannelPolicy(ctx *cli.Context) error {
 		return fmt.Errorf("time_lock_delta argument missing")
 	}
 
+	switch {
+	case ctx.IsSet("max_htlc_msat"):
+		maxHtlcMsat = ctx.Uint64("max_htlc_msat")
+	case args.Present():
+		maxHtlcMsat, err = strconv.ParseUint(args.First(), 10, 64)
```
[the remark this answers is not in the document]
valentinewallace: Lmk if this isn't actually resolved, tested it out and seemed to work.

rpcserver.go
valentinewallace: Discussed, hopefully the code is more clear now.

rpcserver.go
```diff
@@ -4521,16 +4521,55 @@ func (r *rpcServer) FeeReport(ctx context.Context,
 // 0.000001, or 0.0001%.
 const minFeeRate = 1e-6
 
+// checkMaxHtlc errors if the user tries to update their channel policy with
+// a max HTLC value that is too high or too low for a given list of channels.
+func checkMaxHtlc(maxHtlc uint64, channels []*channeldb.OpenChannel) error {
+	var maxHtlcRequested = lnwire.MilliSatoshi(maxHtlc)
```
wpaulino: Nit: can use `:=` notation.

rpcserver.go
```diff
@@ -4521,16 +4521,55 @@ func (r *rpcServer) FeeReport(ctx context.Context,
...
+	for _, ch := range channels {
+		if maxHtlcRequested < ch.LocalChanCfg.MinHTLC {
+			return fmt.Errorf("max htlc size of %v msat is below "+
+				"min HTLC policy of %v msat for channel %v",
+				maxHtlc, ch.LocalChanCfg.MinHTLC,
+				ch.ShortChannelID)
```
wpaulino: Perhaps use the channel point instead here and below as that's what is input by the user?

cmd/lncli/commands.go
```diff
@@ -3363,7 +3363,8 @@ var updateChannelPolicyCommand = cli.Command{
 	Category: "Channels",
 	Usage: "Update the channel policy for all channels, or a single " +
 		"channel.",
-	ArgsUsage: "base_fee_msat fee_rate time_lock_delta [channel_point]",
+	ArgsUsage: "base_fee_msat fee_rate time_lock_delta max_htlc_msat " +
```
wpaulino: So now we'll only allow `max_htlc_msat` to be set with the flag explicitly, but this seems to indicate that you can specify it without the flag?

routing/router.go
```diff
@@ -214,6 +214,10 @@ type ChannelPolicy struct {
 	// TimeLockDelta is the required HTLC timelock delta to be used
 	// when forwarding payments.
 	TimeLockDelta uint32
+
+	// MaxHTLC is the maximum HTLC size including fees we are allowed to
+	// forward over this channel.
+	MaxHTLC lnwire.MilliSatoshi
```
wpaulino: Note in the `godoc` that it is optional?

htlcswitch/link_test.go
```diff
@@ -975,6 +975,48 @@ func TestUpdateForwardingPolicy(t *testing.T) {
...
+	).Wait(30 * time.Second)
+	if err != nil {
+		t.Fatalf("unable to send payment: %v", err)
+	}
+
+	// Now we'll update Bob's policy to lower his max HTLC to an extent
+	// that'll cause him to reject the same HTLC that we just sent.
+	newPolicy = n.globalPolicy
+	newPolicy.MaxHTLC = amountNoFee - 1
+	n.secondBobChannelLink.UpdateForwardingPolicy(newPolicy)
```
joostjager: check err?

htlcswitch/link_test.go
```diff
@@ -975,6 +975,48 @@ func TestUpdateForwardingPolicy(t *testing.T) {
...
+		t.Fatalf("payment should've been rejected")
+	}
+
+	ferr, ok = err.(*ForwardingError)
+	if !ok {
+		t.Fatalf("expected a ForwardingError, instead got (%T): %v",
+			err, err)
+	}
+	switch ferr.FailureMessage.(type) {
+	case *lnwire.FailTemporaryChannelFailure:
```
joostjager: Is this the error we return? :thinking: For a policy problem, we'd rather have a different error, so we can retry with the new channel update if that was attached.

lntest/itest/lnd_test.go
```diff
@@ -1659,6 +1679,47 @@ func testUpdateChannelPolicy(net *lntest.NetworkHarness, t *harnessTest) {
...
+		t.Fatalf("expected policy update to fail, instead got %v", err)
+	}
+
+	// Ensure that we can't update a channel's max HTLC with a value below the
+	// min HTLC.
+	req.MaxHtlcMsat = defaultMinHtlc - 1
+	_, err = net.Bob.UpdateChannelPolicy(ctxb, req)
+	if err == nil || !strings.Contains(err.Error(), "is below min HTLC") {
+		t.Fatalf("expected policy update to fail, instead got %v", err)
+	}
```
joostjager: I am not sure if it is justified to cover this in an integration test. Integration tests come at the cost of build/test time on travis and additional maintenance cost. To me this looks like it could just as well be tested in a unit test?

htlcswitch/link_test.go
valentinewallace: Sorry, `UpdateForwardingPolicy` doesn't return an error, am I missing something?

routing/router.go
valentinewallace: Left out the "optional" since this struct is only used in `rpcserver` and `gossiper`, and `MaxHtlc` will always be set by one of them (either the user in `UpdateChannelPolicy` or the `gossiper` if the user leaves `MaxHtlc` unset).

lnrpc/rpc.proto
```diff
@@ -2304,6 +2304,9 @@ message PolicyUpdateRequest {
 
     /// The required timelock delta for HTLCs forwarded over the channel.
     uint32 time_lock_delta = 5 [json_name = "time_lock_delta"];
+
+    /// The maximum HTLC size in milli-satoshis.
```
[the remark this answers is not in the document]
valentinewallace: Do you prefer that happens in this PR, or a separate PR? This PR seems decently sizable as is, haha. But, maybe it's best to do it all in one.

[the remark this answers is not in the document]
joostjager: I have no preference, just wanted to mention that `min_htlc` has been neglected a bit. But maybe it is also the less exciting parameter of the two.

htlcswitch/link_test.go
joostjager: Hmm, don't know where that comment came from.

htlcswitch/link_test.go
valentinewallace: TemporaryChannelFailure was [the only error option I saw](https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#failure-messages) that includes a `ChannelUpdate` and isn't inaccurate to the reason for failure. Lmk if I'm missing a more descriptive failure type option.

It seems like since the spec has an error for `htlc_minimum_msat` it should have an error for `htlc_maximum_msat` too :thinking:

rpcserver.go
```diff
@@ -4521,16 +4521,55 @@ func (r *rpcServer) FeeReport(ctx context.Context,
 // 0.000001, or 0.0001%.
 const minFeeRate = 1e-6
 
+// checkMaxHtlc errors if the user tries to update their channel policy with
+// a max HTLC value that is too high or too low for a given list of channels.
+func checkMaxHtlc(maxHtlc uint64, channels []*channeldb.OpenChannel) error {
```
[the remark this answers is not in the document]
valentinewallace: Sort of, the `rpcserver` checks that a MaxHtlc is below the `MaxPendingAmount` for our channel, but the `gossiper` only checks that it's below the channel capacity, since it doesn't have access to `MaxPendingAmount` info (iinm). 

Still, lmk if you think this check should just be removed -- it creates a bit of inconsistency, even though `MaxPendingAmount` is a better upper limit.

lnrpc/rpc.proto
```diff
@@ -2304,6 +2304,9 @@ message PolicyUpdateRequest {
...
+    uint64 max_htlc_msat = 6 [json_name = "max_htlc_msat"];
```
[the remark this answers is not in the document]
valentinewallace: Ah, I agree. Changed the commit structure.

lntest/itest/lnd_test.go
valentinewallace: Hm, yeah. I removed the parts that test incorrect `MaxHtlc`'s.

lnrpc/rpc.proto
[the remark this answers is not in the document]
valentinewallace: Yeah, it makes sense as a feature.

htlcswitch/link_test.go
joostjager: Without an `htlc_maximum_msat` error, you'd need to make an assumption on the sender's implementation and what triggers them to retry. For `lnd` senders, I'd say it is best to send `htlc_minimum_msat` also for exceeding the max. What we'll do is apply the channel update and retry. But maybe you want to hear other's opinions about this too.

rpcserver.go
```diff
@@ -4524,16 +4524,55 @@ func (r *rpcServer) FeeReport(ctx context.Context,
...
-	// target any channels by their channel point.
+	// If the request is targeting all active channels, then we need to retrieve
+	// them to check that the request's max HTLC value is within range.
 	case *lnrpc.PolicyUpdateRequest_Global:
+		openChannels, err := r.server.chanDB.FetchAllOpenChannels()
+		if err != nil {
+			return nil, err
+		}
+
+		pendingChannels, err := r.server.chanDB.FetchPendingChannels()
```
joostjager: There is a small gap here if a channel moves from pending to open in between these calls.

rpcserver.go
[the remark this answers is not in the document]
joostjager: I think that ideally you want to have the check only in one place and the `ValidateChannelUpdateAnn` is already there. Gossiper is also already querying `Router` for outgoing channels. What would it take to also retrieve the `MaxPendingAmount` there and check it?

discovery/gossiper.go
joostjager: Also here, shouldn't we be careful not to interpret `0` as "leave unchanged"?

cmd/lncli/commands.go
```diff
@@ -3365,7 +3365,8 @@ var updateChannelPolicyCommand = cli.Command{
 	Category: "Channels",
 	Usage: "Update the channel policy for all channels, or a single " +
 		"channel.",
-	ArgsUsage: "base_fee_msat fee_rate time_lock_delta [channel_point]",
+	ArgsUsage: "base_fee_msat fee_rate time_lock_delta " +
+		"[--max_htlc_msat=N] [channel_point]",
```
joostjager: Do we specify those flags like this in `ArgsUsage` for other commands?

htlcswitch/link_test.go
valentinewallace: @wpaulino, thoughts?

cmd/lncli/commands.go
valentinewallace: Got it from [this line](https://github.com/lightningnetwork/lnd/blob/master/cmd/lncli/commands.go#L152)

htlcswitch/link.go
```diff
@@ -2157,6 +2157,9 @@ func (l *channelLink) UpdateForwardingPolicy(newPolicy ForwardingPolicy) {
 	if newPolicy.MinHTLC != 0 {
 		l.cfg.FwrdingPolicy.MinHTLC = newPolicy.MinHTLC
 	}
+	if newPolicy.MaxHTLC != 0 {
```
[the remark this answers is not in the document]
valentinewallace: True, but it seems like this function is intended to be usable with an incomplete policy: https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L2152

discovery/gossiper.go
valentinewallace: If it's `0` here, it'll get set a few lines later in [`updateChannel`](https://github.com/lightningnetwork/lnd/blob/master/discovery/gossiper.go#L2429).

halseth:
Needs a rebase!

rpcserver.go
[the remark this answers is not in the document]
valentinewallace: Some options:
1. In the gossiper's `updateChannel`, use the `ChannelEdgePolicy`'s `DB` pointer to get the `MaxPendingAmount`
2. Add a closure for fetching the `MaxPendingAmount` of a channel to the gossiper's `cfg`
3. Pass [`targetDBChans`](https://github.com/lightningnetwork/lnd/pull/2438/files#diff-01ada5680aaa18fc79cbf78b2d55a37bR4560) to the gossiper's `PropagateChanPolicyUpdate`

rpcserver.go
valentinewallace: Added TODO :grimacing:

lnrpc/rpc.proto
```diff
@@ -2320,6 +2320,9 @@ message PolicyUpdateRequest {
 
     /// The required timelock delta for HTLCs forwarded over the channel.
     uint32 time_lock_delta = 5 [json_name = "time_lock_delta"];
+
+    /// The maximum HTLC size in milli-satoshis.
```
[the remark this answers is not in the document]
valentinewallace: Added to comment.

Yeah, not sure the best way to handle fields with a meaningful zero value. Might like the idea of having `-1` as a nil/unset value.

routing/router.go
joostjager: Optional is still in the comment.

lnrpc/rpc.proto
```diff
@@ -2395,6 +2395,9 @@ message PolicyUpdateRequest {
 
     /// The required timelock delta for HTLCs forwarded over the channel.
     uint32 time_lock_delta = 5 [json_name = "time_lock_delta"];
+
+    /// If set, the maximum HTLC size in milli-satoshis. If unset, the maximum HTLC will be unchanged.
+    uint64 max_htlc_msat = 6 [json_name = "max_htlc_msat"];
```
joostjager: Should it be unset or zero?

rpcserver.go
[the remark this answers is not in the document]
joostjager: I'd think that `Router.ForAllOutgoingChannels` is the place to do it. Return more complete info on the set of outgoing channels that includes `MaxPendingAmount`.

cmd/lncli/commands.go
```diff
@@ -3362,6 +3363,11 @@ var updateChannelPolicyCommand = cli.Command{
 			Usage: "the CLTV delta that will be applied to all " +
 				"forwarded HTLCs",
 		},
+		cli.Uint64Flag{
+			Name: "max_htlc_msat",
+			Usage: "if set, the max HTLC size that will be applied " +
+				"to all forwarded HTLCs",
```
joostjager: Does not setting mean that there is no max or that the max is left unchanged? Comment could be explicit about this.

rpcserver.go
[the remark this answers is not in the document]
joostjager: Discussed with @halseth what a final design could be to move towards in small steps. It seems that having a distinct object `LocalChannelManager` that manages our local channel state including the forwarding policy would be an improvement. It could expose `ForAllOutgoingChannels`, but also something like `ValidatePolicy`. This also allows to, later, split all the persistent data associated with our own channels to a separate db. When an updates happens in `LocalChannelManager`, it is pushed to the gossiper. No further checking is needed in the gossiper anymore.

I think that ideally we'd want to make a step in this PR towards that. Or towards another final design that we think is good. The current change with another loop over channels in `rpcserver` feels to me like creating technical debt. For an urgent bug fix I can understand that, but if we are not in a hurry with this max htlc setting, I'd prefer doing it right.

htlcswitch/link_test.go
wpaulino: > For a policy problem, we'd rather have a different error, so we can retry with the new channel update if that was attached.

I think we can proceed with this as is for now. We already return the error in the master branch, the change being done here is for a test.

joostjager: Ok, good point. We don't make it worse here.

joostjager:
Picking up this PR in #3523

PR #2439, opened 2019-01-08
wtpolicy: migrate to using wtpolicy.Policy + blob bitvector flags

cfromknecht:
This PR refactors the internal handling of session parameters within the existing watchtower packages in preparation for the watchtower client integration.  First, we introduce a `blob.Type` bitvector, composed of `blob.Flag`s. Blob types are meant to replace the `uint16` versioning scheme, and are used to signal what features the client is requesting from the tower.

Two flags are introduced: `FlagReward` and `FlagCommitOutputs`:
 - `FlagReward` is used to indicate if the tower will be getting a portion of the breached balance. Otherwise, the tower will sweep all funds to the client.
 - `FlagCommitOutputs` instructs the tower that the payload is the fixed-size, v0 blob format. This is added as a flag so that it could, in theory, be turned off and replaced with a different encoding.

Finally, we consolidate our negotiated session parameters into a new `wtpolicy.Policy` struct. This struct is then threaded through the existing watchtower wire/server/lookout code, as it is already used natively in the client to follow. The server will also now reject blob types for which it doesn't understand, to ensure we don't accept sessions with clients that require a different set of features.

The final commit is added to improve coverage on the `JusticeKit`'s construction of witness scripts and stacks. These are tested indirectly in the lookout package, but now they are tested explicitly in this package.

### DIFF
--- a/watchtower/blob/justice_kit_test.go
+++ b/watchtower/blob/justice_kit_test.go
@@ -38,8 +38,8 @@ func makeAddr(size int) []byte {
 
 type descriptorTest struct {
 	name                 string
-	encVersion           uint16
-	decVersion           uint16
+	encVersion           blob.Type
+	decVersion           blob.Type
 	sweepAddr            []byte
 	revPubKey            blob.PubKey
 	delayPubKey          blob.PubKey
@@ -52,11 +52,15 @@ type descriptorTest struct {
 	decErr               error
 }
 
+var rewardAndCommitType = blob.TypeFromFlags(
+	blob.FlagReward, blob.FlagCommitOutputs,
+)
+
 var descriptorTests = []descriptorTest{
 	{
 		name:             "to-local only",
-		encVersion:       0,
-		decVersion:       0,
+		encVersion:       blob.TypeDefault,
+		decVersion:       blob.TypeDefault,
 		sweepAddr:        makeAddr(22),
 		revPubKey:        makePubKey(0),
 		delayPubKey:      makePubKey(1),
@@ -65,8 +69,8 @@ var descriptorTests = []descriptorTest{
 	},
 	{
 		name:                 "to-local and p2wkh",
-		encVersion:           0,
-		decVersion:           0,
+		encVersion:           rewardAndCommitType,
+		decVersion:           rewardAndCommitType,
 		sweepAddr:            makeAddr(22),
 		revPubKey:            makePubKey(0),
 		delayPubKey:          makePubKey(1),
@@ -78,30 +82,30 @@ var descriptorTests = []descriptorTest{
 	},
 	{
 		name:             "unknown encrypt version",
-		encVersion:       1,
-		decVersion:       0,
+		encVersion:       0,
+		decVersion:       blob.TypeDefault,
 		sweepAddr:        makeAddr(34),
 		revPubKey:        makePubKey(0),
 		delayPubKey:      makePubKey(1),
 		csvDelay:         144,
 		commitToLocalSig: makeSig(1),
-		encErr:           blob.ErrUnknownBlobVersion,
+		encErr:           blob.ErrUnknownBlobType,
 	},
 	{
 		name:             "unknown decrypt version",
-		encVersion:       0,
-		decVersion:       1,
+		encVersion:       blob.TypeDefault,
+		decVersion:       0,
 		sweepAddr:        makeAddr(34),
 		revPubKey:        makePubKey(0),
 		delayPubKey:      makePubKey(1),
 		csvDelay:         144,
 		commitToLocalSig: makeSig(1),
-		decErr:           blob.ErrUnknownBlobVersion,
+		decErr:           blob.ErrUnknownBlobType,
 	},
 	{
 		name:             "sweep addr length zero",
-		encVersion:       0,
-		decVersion:       0,
+		encVersion:       blob.TypeDefault,
+		decVersion:       blob.TypeDefault,
 		sweepAddr:        makeAddr(0),
 		revPubKey:        makePubKey(0),
 		delayPubKey:      makePubKey(1),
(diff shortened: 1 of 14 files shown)


watchtower/wtpolicy/policy.go
```diff
@@ -0,0 +1,68 @@
...
+	DefaultMaxUpdates = 1024
+
+	// DefaultRewardRate specifies the fraction of the channel that the
+	// tower takes if it successfully sweeps a breach. The value is
+	// expressed in millionths of the channel capacity.
+	DefaultRewardRate = 10000
+
+	// DefaultSweepFeeRate specifies the fee rate used to construct justice
+	// transactions. The value is expressed in satoshis per kilo-weight.
+	DefaultSweepFeeRate = 3000
```
Roasbeef: Should there even be a default? Instead we can pass in a `lnwallet.FeeEstimator` instance and if we're on simnet or w/e, then that can be a static version of it.

watchtower/blob/justice_kit_test.go
```diff
@@ -212,3 +216,195 @@ func testBlobJusticeKitEncryptDecrypt(t *testing.T, test descriptorTest) {
 			"want: %v, got %v", boj, boj2)
 	}
 }
+
+// TestJusticeKitRemoteWitnessConstruction tests that a JusticeKit returns the
+// proper to-remote witnes script and to-remote witness stack. This should be
```
Roasbeef: witnes -> witness

watchtower/blob/justice_kit_test.go
```diff
@@ -212,3 +216,195 @@ func testBlobJusticeKitEncryptDecrypt(t *testing.T, test descriptorTest) {
...
+
+	// Now, compute the to-remote witness script returned by the justice
+	// kit.
+	toRemoteScript, err := justiceKit.CommitToRemoteWitnessScript()
+	if err != nil {
+		t.Fatalf("unable to compute to-remote witness script: %v", err)
+	}
+
+	// Assert this is exactly the to-remote, compressed pubkey.
+	if !bytes.Equal(toRemoteScript, toRemotePubKey[:]) {
```
Roasbeef: Why should the script match the pubkey? The script will have an `OP_0` there along with a data push.

watchtower/blob/justice_kit_test.go
```diff
@@ -212,3 +216,195 @@ func testBlobJusticeKitEncryptDecrypt(t *testing.T, test descriptorTest) {
...
+		t.Fatalf("to-remote witness script should be equal to "+
+			"to-remote pubkey, want: %x, got %x",
+			toRemotePubKey[:], toRemoteScript)
+	}
+
+	// Next, compute the to-remote witness stack, which should be a p2wkh
```
Roasbeef: The stack for `p2wkh` should be: the sig, and then the pubkey.

watchtower/blob/justice_kit_test.go
```diff
@@ -212,3 +216,195 @@ func testBlobJusticeKitEncryptDecrypt(t *testing.T, test descriptorTest) {
...
+
+	// Next, compute the to-local witness stack returned by the justice kit.
+	toLocalWitnessStack, err := justiceKit.CommitToLocalRevokeWitnessStack()
+	if err != nil {
+		t.Fatalf("unable to compute to-local witness stack: %v", err)
+	}
+
+	// A valid witness that spends the revocation path should have exactly
+	// two elements on the stack.
+	if len(toLocalWitnessStack) != 2 {
```
Roasbeef: The revoke witness stack should have 3 items: the sig, the 0x01, and the witness script.

watchtower/wtpolicy/policy.go
cfromknecht: This is the fee rate that is negotiated between client and server. The client itself can have a fee estimator, and use that to prescribe what it attempts to negotiate. This is just a shared struct for holding the negotiated values.

cfromknecht: Sorry, ignore the prior comment, misread where the comment was.

There doesn't need to be a default, yes in practice you would use whatever the client fee estimator proposes. I've been using this mostly to generate valid policies for testing. Will remove it

watchtower/blob/justice_kit_test.go
cfromknecht: this is actually the witness script, i will change the naming

watchtower/blob/justice_kit_test.go
cfromknecht: i've been using witness stack here to describe everything excluding the witness script, which the caller already has via `CommitToLocal/RemoteWitnessScript()`. they are later joined via: https://github.com/lightningnetwork/lnd/blob/master/watchtower/lookout/justice_descriptor.go#L293

watchtower/blob/justice_kit_test.go
cfromknecht: see below comment

watchtower/blob/justice_kit_test.go
Roasbeef: To me, this isn't the witness script. Instead, it's the data push of the witness program. Not a blocker, just a matter of tending to consistent language across the codebase.

watchtower/blob/justice_kit_test.go
Roasbeef: Even with the below, p2wkh doesn't have a witness script as its last item. Instead it always has the sig+pubkey combo. Technically, the witness script is actually the full unrolled p2pkh type script.

watchtower/blob/justice_kit_test.go
Roasbeef: Witness script: the pre-image of the witness program for p2wsh. This is the first thing on the stack. 
Witness program: the pkScript of a segwit output. 
Witness stack: the full witness that allows one to spend a witness program.

Issue #2420, opened 2019-01-05
Feature request: extend ListPayments response by original invoice memo


misovan:
### Background

I would like to propose to extend ListPayment response by memo information from the invoice the payment was based on. This information is necessary for easy identification of the payment by the user. Many standard payment formats implement such behavior.

### Proposed behaviour

```
        {
            "payment_hash": "14e42ffc411edea0e44ba00437b1e7a7e34995e1299d511506190f40341c30ae",
            "value": "2",
            "creation_date": "1546714044",
            "path": [
            ],
            "memo": "Payment for 2 pixels at satoshis.place.",
            "fee": "1",
            "payment_preimage": "ffaad08cf2a3983a46488bb2e44af3136832c8bf7705b6e868ed5e52f5ff638b"
        }
```

### Actual behaviour

ListPayments grpc api currently [returns](https://api.lightning.community/?shell#payment) an array of `Payment` objects. `Payment` does not contain `Memo` field from the invoice it was based on.

ListPayments returns list of payments stored in lnd database. Currently LND stores each payment into db (see AddPayment in [channeldb/payments.go](https://github.com/lightningnetwork/lnd/blob/master/channeldb/payments.go)) as `OutgoingPayment` type. This type definition already contains `Invoice`, but the actual memo data are not passed to it when saving payment to db (see savePayment in rpcserver.go).

### Proposal

1. Extend `rpcPaymentIntent` type by `memo` (in [rpcserver.go](https://github.com/lightningnetwork/lnd/blob/master/rpcserver.go))
2. Extend extractPaymentIntent function so it sets `payIntent.memo` to `payReq.Description` value
3. Pass `memo` in dispatchPaymentIntent to savePayment function (function signature extended)
4. Extend listPayments response in rpcserver.go so it returns `Memo` set to stored `payment.Invoice.Memo` value (with Memo added as well to used `lnrpc.Payment` type)

Proposed change does not require any changes in db layer nor in channeldb/payments.go as relevant `OugoingPayment` type already implements `Invoice`.

Thanks for feedback.
Michal

misovan:
I've [created a PR #2440](https://github.com/lightningnetwork/lnd/pull/2440) implementing the change

alexbosworth:
This should be covered already, the full payment request is in the payment response

misovan:
Yes it has been covered in a way, that full payReq data are available. While more flexible, that approach requires additional call / client impl of decoding it when invoice memo is needed.

[...]

PR #2440, opened 2019-01-08
Extend ListPayments response by original invoice memo (#2420)

misovan:
Pull request extends grpc ListPayments api by `memo` from the invoice the payment is based on. See [2420](https://github.com/lightningnetwork/lnd/issues/2420) for proposal reasoning and details.

### DIFF
--- a/lnrpc/rpc.pb.go
+++ b/lnrpc/rpc.pb.go
@@ -6423,10 +6423,12 @@ type Payment struct {
 	Fee int64 `protobuf:"varint,5,opt,name=fee,proto3" json:"fee,omitempty"`
 	// / The payment preimage
 	PaymentPreimage string `protobuf:"bytes,6,opt,name=payment_preimage,proto3" json:"payment_preimage,omitempty"`
+	// / The memo from the invoice the payment was based on
+	Memo string `protobuf:"bytes,7,opt,name=memo,proto3" json:"memo,omitempty"`
 	// / The value of the payment in satoshis
-	ValueSat int64 `protobuf:"varint,7,opt,name=value_sat,proto3" json:"value_sat,omitempty"`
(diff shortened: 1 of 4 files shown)


michaelWuensch:
Can't wait for this feature to be included. This would improve user experience so much.

halseth:
Blocked by #2761

misovan:
> Blocked by #2761

Do you plan to store invoice memo or eventually full payreq data alongside the outgoing payment within #2761?
I'll close this PR then.

halseth:
not necessarily, it is just that there is a code conflict currently. Feel free to keep this open until #2761 is merged :)

misovan:
OK, I believe that this listPayments rpc change would be better applied on top of your PR, as your one targets different concern and is much more complex. At the same time I found https://github.com/lightningnetwork/lnd/pull/2717 that is more recent and aims to implement similar/same thing - just took the option to expose whole payreq in listPayments.

misovan:
Closing, listPayments rpc has been extended by payreq (which includes memo data) in #2717 that is now merged.

michaelWuensch:
Actually I still would like to have this. The payreq is now included, but I now have to decode that payreq in order to get the memo. 
My project is in Java and I can't find a library to do this.
Therefore if I want to show the memo on my payment history page, I will have to make hundreds of additional rpc calls to decode the payment request. This causes unwanted delay, traffic and complexity.

halseth:
@michaelWuensch there's `lnd` RPC to do that

michaelWuensch:
@halseth I know, that is exactly what I am refering to in the second part of my last comment and I already implemented that.
But as I said, if the memo was included in the listPayments rpc I would need exactly one rpc call to display 100 Payments with memos in the history.
When it is not included, I need one call to get the 100 payments without the memo. Then I have to loop through them an make one more decodePayReq request for each of the payments.
This results in a total of 101 rpc calls which cause unwanted traffic and delay.

PR #2445, opened 2019-01-09
netann: Add NANN subsystem logger

cfromknecht:
In preparation for #2411, this PR introduces the NANN subsystem logger and tidies the imports in `log.go`

### DIFF
--- a/log.go
+++ b/log.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


cfromknecht:
Allows me to break up the body of #2411 if this is in place, gets rid of 2 of the 20 commits, and this import constantly breaks my rebasing on the watchtower series

PR #2446, opened 2019-01-10
lncfg: NetAddress parsing

cfromknecht:
In this PR, we extend the `lncfg` package to support parsing of `lnwire.NetAddress` strings. As we will be rolling out watchtowers initially via the use of private towers, users will have to specify a `pubkey@addr` string in their config files. Here, we add logic to support parsing and validating a raw string value that can be given to the watchtower client.

We introduce a new `ParseLNAddressString` method, which is meant to mirror the `ParseAddressString` method introduced previously. Internally, `ParseLNAddressString` calls `ParseAddressString`, retaining the capability of using default ports and resolving TCP addrs via a custom resolver.

The unit tests are extended to assert `ParseLNAddressString` can support the variety of valid address types accessible to users today. We also make some slight refactorings to take advantage of the overlap between the two test cases.

### DIFF
--- a/lncfg/address.go
+++ b/lncfg/address.go
@@ -2,12 +2,15 @@ package lncfg
 
 import (
 	"crypto/tls"
+	"encoding/hex"
 	"fmt"
 	"net"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/tor"
 )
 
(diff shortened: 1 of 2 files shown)


lncfg/address.go
```diff
@@ -162,6 +165,59 @@ func ParseAddressString(strAddress string, defaultPort string,
 	}
 }
 
+// ParseLNAddressString converts a string of the form <pubkey>@<addr> into an
+// lnwire.NetAddress. The <pubkey> must be presented in hex, and result in a
+// 33-byte, compressed public key that lies on the secp256k1 curve. The <addr>
+// may be any address supported by ParseAddressString. If no port is specified,
+// the defaultPort will be used. Any tcp addresses that need resolving will be
+// resolved using the custom tcpResolver.
+func ParseLNAddressString(strAddress string, defaultPort string,
```
Roasbeef: Nice! We can eventually also use this method in the `ConnectPeer` RPC method, and also possibly in a few areas in `lncli`.

lncfg/address_test.go
```diff
@@ -78,55 +82,186 @@ var (
...
-		}
-		isAddrUnix := IsUnix(normalized[0])
-		if testVector.isUnix != isAddrUnix {
-			t.Fatalf("#%v: mismatched unix detection: expected "+
-				"%v, got %v for addr %s",
-				i, testVector.isUnix, isAddrUnix,
-				testVector.address,
-			)
-		}
+	for _, test := range addressTestVectors {
```
Roasbeef: Solid test refactoring in this file 👍

PR #2460, opened 2019-01-11
Support the max_htlc field from ChannelUpdates

halseth:
Continuation of #2296.

Replaces #2296 and #2435


lnwire/channel_update.go
```diff
@@ -8,16 +8,20 @@ import (
...
 // how the ChannelUpdate message is to be interpreted.
-type ChanUpdateFlag uint16
+type ChanUpdateChanFlags uint8
 
 const (
 	// ChanUpdateDirection indicates the direction of a channel update. If
 	// this bit is set to 0 if Node1 (the node with the "smaller" Node ID)
 	// is updating the channel, and to 1 otherwise.
-	ChanUpdateDirection ChanUpdateFlag = 1 << iota
+	ChanUpdateDirection ChanUpdateChanFlags = 1 << iota
```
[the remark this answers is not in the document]
Roasbeef: In part, they also have distinct purposes so follows that they also have distinct types.

lnwire/channel_update.go
```diff
@@ -12,6 +13,11 @@ import (
 // present in the ChannelUpdate.
 type ChanUpdateMsgFlags uint8
 
+// String returns the bitfield flags as a string.
+func (c ChanUpdateMsgFlags) String() string {
+	return fmt.Sprintf("%08b", c)
```
Roasbeef: Nice didn't even know this directive was a thing!

Roasbeef: Alternatively, [we can use something like this here](https://github.com/lightningnetwork/lnd/pull/2371/commits/d6ce3bf4531fd01063ef4ebad65aa772fcda6e27). This route makes the logs even more readable IMO. Not a blocker though, we can make an issue to make the logs prettier after this lands.

routing/ann_validation.go
```diff
@@ -144,3 +150,25 @@ func ValidateChannelUpdateAnn(pubKey *btcec.PublicKey,
 
 	return nil
 }
+
+// validateOptionalFields validates a channel update's message flags and
+// corresponding update fields.
+func validateOptionalFields(capacity btcutil.Amount,
+	msg *lnwire.ChannelUpdate) error {
+
+	if msg.MessageFlags&lnwire.ChanUpdateOptionMaxHtlc != 0 {
```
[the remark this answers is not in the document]
Roasbeef: It won't ever be zero, we're checking here to see if the max HTLC bit is set in the message flags variable. If it isn't set (all zeroes), then this won't execute. If it is, then the only thing left on the LHS there will be the max HTLC bit.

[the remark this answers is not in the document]
Roasbeef: Agree we should reject that. However, this should be in the `lnwire` layer: if as we're reading, we see the max HTLC bit set, but then there's no opaque data, or not enough, we should reject the message.

channeldb/graph.go
```diff
@@ -3351,11 +3355,26 @@ func serializeChanEdgePolicy(w io.Writer, edge *ChannelEdgePolicy,
...
+		err := binary.Write(&opaqueBuf, byteOrder, uint64(edge.MaxHTLC))
+		if err != nil {
+			return err
+		}
+	}
+
 	if len(edge.ExtraOpaqueData) > MaxAllowedExtraOpaqueBytes {
 		return ErrTooManyExtraOpaqueBytes(len(edge.ExtraOpaqueData))
 	}
+	if _, err := opaqueBuf.Write(edge.ExtraOpaqueData); err != nil {
```
Roasbeef: This write will cause the buf to expand and allocate more data. Why not just write the buf, then the opaque bytes which have already been allocated?

routing/ann_validation.go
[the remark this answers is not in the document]
Roasbeef: In this case, I personally prefer to write this instead: 
```go
if msg.MessageFlags&lnwire.ChanUpdateOptionMaxHtlc == lnwire.ChanUpdateOptionMaxHtlc
```

Usually with a helper:
```go
func hasFlag(flags byte, flag byte) bool {
    return flags&flag == flag 
}

File: channeldb/graph.go [hunk 5]
```diff
@@ -3207,11 +3207,13 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
 		// (now outdated by the new one), and delete its corresponding
-		// entry within the update index.
+		// entry within the update index. We'll ignore any
+		// ErrEdgePolicyOptionalFieldNotFound error, as we only need
+		// the channel ID and update time to delete the entry.
 		oldEdgePolicy, err := deserializeChanEdgePolicy(
 			bytes.NewReader(edgeBytes), nodes,
 		)
-		if err != nil {
+		if err != nil && err != ErrEdgePolicyOptionalFieldNotFound {
```
Roasbeef: We may want to eventually even just do a migration (or call it a database sanity check) that finds and deletes any channel updates that are malformed.

discovery/gossiper.go
```diff
@@ -1321,6 +1321,17 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels() error {
 			return nil
 		}
 
+		// If this edge has a ChannelUpdate that was created before the
```
Roasbeef: 👍

server.go
```diff
@@ -3065,6 +3065,7 @@ func createChannelUpdate(info *channeldb.ChannelEdgeInfo,
 		ChannelFlags:    policy.ChannelFlags,
 		TimeLockDelta:   policy.TimeLockDelta,
 		HtlcMinimumMsat: policy.MinHTLC,
+		HtlcMaximumMsat: policy.MaxHTLC,
```
Roasbeef: Commit message nor body doesn't match this diff. For ref it's: 
> server: set max htlc in ChannelUpdate sent on HtlcSatifiesPolicy failure

and this diff touches only the server.

fundingmanager.go
```diff
@@ -2257,13 +2264,20 @@ func (f *fundingManager) annAfterSixConfs(completeChan *channeldb.OpenChannel,
 		// HTLC it deems economically relevant.
 		fwdMinHTLC := completeChan.LocalChanCfg.MinHTLC
 
+		// We'll obtain the max HTLC value we can forward in our direction, as
+		// we'll use this value within our ChannelUpdate. This value must be <=
+		// channel capacity and <= the maximum in-flight msats set by the peer,
+		// so we default to max in-flight msats as this value will always be <=
+		// channel capacity.
+		fwdMaxHTLC := completeChan.LocalChanCfg.MaxPendingAmount
```
Roasbeef: FWIW, with the gossiper changes, on the first retransmission tick, we'll override this with the current capacity of the channel.

lnwire/channel_update.go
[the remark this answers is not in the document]
halseth: Yeah, type safety makes it easier to not assign one to the other, checking the bits on the correct flag etc :)

lnwire/lnwire_test.go
```diff
@@ -178,6 +178,50 @@ func randAddrs(r *rand.Rand) ([]net.Addr, error) {
 	return []net.Addr{tcp4Addr, tcp6Addr, v2OnionAddr, v3OnionAddr}, nil
 }
 
+// TestChanUpdateChanFlags ensures that converting the ChanUpdateChanFlags and
+// ChanUpdateMsgFlags bitfields to a string behaves as expected.
+func TestChanUpdateChanFlags(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
```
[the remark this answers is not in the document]
halseth: \o/

lnwire/channel_update.go
halseth: Hm yeah, could be a nice follow-up :)

routing/ann_validation.go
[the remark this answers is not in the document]
halseth: I personally prefer the current `flags&flag != 0` style because it is less verbose (C style), but agree that a helper method makes it more readable!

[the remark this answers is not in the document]
halseth: If the message flag is set, but there's not enough data in the byte stream for the field to be there, we *do reject* it in the `lnwire` layer.

If the message flag is not set, we just interpret the data as opaque if it's there.

channeldb/graph.go
halseth: Not sure I understand what you mean?

The rationale behind doing it this way is that we need to write the combined bytes of`max_htlc` + `opaque_bytes` as `VarBytes` to be compatible with the old format. Are you saying we should do something like 
```go
scratch [8]byte = bytes(max_htlc)
wire.WriteVarBytes(w, 0, append(scratch[:], edge.ExtraOpaqueData...))
```
?

channeldb/graph.go
halseth: Added a TODO.

server.go
halseth: I think maybe it's because `HtlcSatifiesPolicy` eventually calls this method when fetching the updated `ChannelUpdate`? @valentinewallace

fundingmanager.go
halseth: We'll do that only for existing edges not having the field already set.

routing/ann_validation.go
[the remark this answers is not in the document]
halseth: Added a helper here: https://github.com/lightningnetwork/lnd/pull/2460/commits/1ffebc8d16e892d0ec37da0ff9c409e16b1ab070

lmk if that's what you envisioned!

channeldb/graph.go
```diff
@@ -3207,11 +3207,13 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 		// *prior* update time in order to delete it. To do this, we'll
 		// need to deserialize the existing policy within the database
 		// (now outdated by the new one), and delete its corresponding
-		// entry within the update index.
+		// entry within the update index. We'll ignore any
```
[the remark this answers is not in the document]
halseth: Fixed.

channeldb/graph_test.go
```diff
@@ -2601,6 +2611,173 @@ func TestNodeIsPublic(t *testing.T) {
 	)
 }
 
+// TestEdgePolicyMissingMaxHtcl tests that if we find a ChannelEdgePolicy in
+// the DB that indicates that it should support the htlc_maximum_value_msat
+// field, but it is not part of the opaque data, then we'll handle it as it is
+// unknown. It also checks that we are correclty able to overwrite it when we
```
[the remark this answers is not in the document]
halseth: Fixed.

halseth:
PTAL

server.go
valentinewallace: @halseth that's what I was originally thinking, but checking again now it looks like I may be mistaken. This function gets called down the stack by `announceChanStatus`, which is used in 4+ places for different reasons.

Roasbeef:
Fails lint: 
```
lnwallet/interface_test.go:440:46:error: not enough arguments in call to aliceChanReservation.CommitConstraints (vet)
lnwallet/interface_test.go:872:46:error: not enough arguments in call to aliceChanReservation.CommitConstraints (vet)
```

halseth:
Lint fixed in https://github.com/lightningnetwork/lnd/pull/2460/commits/50893043e94c3dbf9069c7ec902c255cebf5cad4

channeldb/graph.go
```diff
@@ -3169,49 +3177,13 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
-
-	if err := binary.Write(&b, byteOrder, edge.ChannelID); err != nil {
+	if err := serializeChanEdgePolicy(&b, edge, to); err != nil {
 		return err
 	}
 
 	var scratch [8]byte
 	updateUnix := uint64(edge.LastUpdate.Unix())
 	byteOrder.PutUint64(scratch[:], updateUnix)
-	if _, err := b.Write(scratch[:]); err != nil {
```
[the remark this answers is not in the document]
halseth: Done!

halseth:
Rebased.

PR #2463, opened 2019-01-12
server: implement notifications for disconnected peers

wpaulino:
In this commit, we introduce a new method that allows the different subsystems of `lnd` to request a notification for when a particular peer has disconnected.

This will serve as a basis for #1595.


server.go
```diff
@@ -2527,8 +2558,16 @@ func (s *server) peerTerminationWatcher(p *peer, ready chan struct{}) {
 	// in question.
 	s.removePeer(p)
 
-	// Next, check to see if this is a persistent peer or not.
+	// If there were any notification requests for when this peer
```
cfromknecht: would say this should be moved up to just after acquiring the lock, otherwise we won't deliver notifications for teardowns that take the first conditional

PR #2464, opened 2019-01-12
channeldb+discovery: add gossip message store

wpaulino:
In this PR, we add a new store within the database that'll be responsible for storing gossip messages which we need to reliably send to peers. This aims to replace the current messageStore that exists within the gossiper, so much of this logic is borrowed from there. One of the main differences between the two is that we now index messages with a new key format in which we take into account the message's type. This allows us to store different messages for a specific channel with a peer. The new key format will require a migration, which has been included in this PR as well.

This will serve as a basis for #1595.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -2,7 +2,6 @@ package discovery
 
 import (
 	"bytes"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"runtime"
@@ -13,7 +12,6 @@ import (
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/coreos/bbolt"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/channeldb"
@@ -25,13 +23,6 @@ import (
 )
 
 var (
-	// messageStoreKey is a key used to create a top level bucket in
-	// the gossiper database, used for storing messages that are to
-	// be sent to peers. Currently this is used for reliably sending
-	// AnnounceSignatures messages, by persisting them until a send
-	// operation has succeeded.
-	messageStoreKey = []byte("message-store")
-
 	// ErrGossiperShuttingDown is an error that is returned if the gossiper
 	// is in the process of being shut down.
 	ErrGossiperShuttingDown = errors.New("gossiper is shutting down")
@@ -136,6 +127,10 @@ type Config struct {
 	// DB is a global boltdb instance which is needed to pass it in waiting
 	// proof storage to make waiting proofs persistent.
 	DB *channeldb.DB
+
+	// MessageStore is a persistent storage of gossip messages which we will
+	// use to determine which messages need to be resent for a given peer.
+	MessageStore GossipMessageStore
 
 	// AnnSigner is an instance of the MessageSigner interface which will
 	// be used to manually sign any outgoing channel updates. The signer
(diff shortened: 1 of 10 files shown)


channeldb/messagestore.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	// Serialize the message with its wire encoding.
+	var b bytes.Buffer
+	if _, err := lnwire.WriteMessage(&b, msg, 0); err != nil {
+		return err
+	}
+
+	return s.db.Update(func(tx *bbolt.Tx) error {
+		messageStore, err := tx.CreateBucketIfNotExists(messageStoreBucket)
```
cfromknecht: would prefer this to be created on initialization of the store, then from then on we can assume it exists

channeldb/messagestore.go
```diff
@@ -0,0 +1,164 @@
...
+	peerPubKey [33]byte) ([]lnwire.Message, error) {
+
+	var msgs []lnwire.Message
+	err := s.db.View(func(tx *bbolt.Tx) error {
+		messageStore := tx.Bucket(messageStoreBucket)
+		if messageStore == nil {
+			return nil
+		}
+
+		return messageStore.ForEach(func(k, v []byte) error {
```
cfromknecht: since all messages for the same peer should be adjacent, we should be able to use `Seek` instead of iterating over all stored messages

discovery/gossiper.go
```diff
@@ -814,126 +809,69 @@ func (d *deDupedAnnouncements) Emit() []msgWithSenders {
...
-			"messageStore", t.msg.ChannelID)
-		if err := d.cfg.DB.Update(func(tx *bbolt.Tx) error {
-			bucket := tx.Bucket(messageStoreKey)
-			if bucket == nil {
-				return fmt.Errorf("bucket " +
-					"unexpectedly did not exist")
+				// If the channel cannot be found, it is most likely a
+				// leftover message for a channel that was closed.  In
+				// this case we delete it from the message store.
+				log.Warnf("unable to fetch channel info for "+
```
cfromknecht: nit: log message should be capitalized

discovery/gossiper.go
```diff
@@ -814,126 +809,69 @@ func (d *deDupedAnnouncements) Emit() []msgWithSenders {
...
+					"chanID=%v from graph: %v. Will delete local"+
```
cfromknecht: style nit: Will delete -> Deleting

channeldb/messagestore.go
wpaulino: Fixed.

channeldb/messagestore.go
wpaulino: Fixed.

discovery/gossiper.go
wpaulino: This is more of a code move and will be replaced in #1595, so don't think it's worth addressing.

discovery/gossiper.go
wpaulino: Same comment as above.

wpaulino:
Pushed a new version that avoid the database migration.

channeldb/messagestore.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		// If we're attempting to delete an AnnounceSignatures message
+		// from the store, we'll first check if it's using the legacy
+		// key format. We'll only proceed to delete it if it exists,
+		// otherwise, we'll go ahead with the new key format.
+		if msg, ok := msg.(*lnwire.AnnounceSignatures); ok {
+			oldKey := oldMessageStoreKey(msg, peerPubKey)
+			if messageStore.Get(oldKey) != nil {
+				return messageStore.Delete(oldKey)
```
halseth: Could we risk both keys being in the store? Seems like we might as well just delete both keys always.

channeldb/messagestore.go
```diff
@@ -0,0 +1,204 @@
...
+	var msgs []lnwire.Message
+	err := s.db.View(func(tx *bbolt.Tx) error {
+		messageStore := tx.Bucket(messageStoreBucket)
+		if messageStore == nil {
+			return ErrCorruptedMessageStore
+		}
+
+		c := messageStore.Cursor()
+		k, v := c.Seek(peerPubKey[:])
+		for ; bytes.HasPrefix(k, peerPubKey[:]); k, v = c.Next() {
```
halseth: 👍

channeldb/messagestore_test.go
```diff
@@ -0,0 +1,242 @@
...
+	// ErrUnsupportedMessage.
+	err := msgStore.AddMessage(unsupportedMsg, peer)
+	if err != ErrUnsupportedMessage {
+		t.Fatalf("expected ErrUnsupportedMessage, got %v", err)
+	}
+}
+
+// TestLegacyGossipMessageStore ensures a message with a legacy key format
+// within the store can be properly removed from it.
+func TestLegacyGossipMessageStore(t *testing.T) {
```
halseth: 👍

channeldb/messagestore_test.go
```diff
@@ -0,0 +1,242 @@
...
+	if err := msgStore.DeleteMessage(msg, peerPubKey); err != nil {
+		t.Fatalf("unable to delete legacy message: %v", err)
+	}
+	msgs, err = msgStore.MessagesForPeer(peerPubKey)
+	if err != nil {
+		t.Fatalf("unable to retrieve messages for peer: %v", err)
+	}
+	if len(msgs) != 0 {
+		t.Fatalf("expected 0 messages for peer, got %d", len(msgs))
+	}
```
halseth: Maybe add with a new key, ensure it overwrites the legacy one?

channeldb/messagestore.go
```diff
@@ -0,0 +1,204 @@
+package channeldb
```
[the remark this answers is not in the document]
wpaulino: Not really, moved.

channeldb/messagestore_test.go
wpaulino: The current implementation doesn't check for a legacy entry when adding the same message again. Are you saying that should be done and the test should also be expanded to cover this? In any case, don't think this is really necessary as we'll end up removing the message (both new and legacy entries) once it's stale anyway.

discovery/messagestore.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	// Serialize the message with its wire encoding.
+	var b bytes.Buffer
+	if _, err := lnwire.WriteMessage(&b, msg, 0); err != nil {
+		return err
+	}
+
+	return s.db.Update(func(tx *bbolt.Tx) error {
```
Roasbeef: Can be made it use `Batch`.

discovery/messagestore.go
```diff
@@ -0,0 +1,204 @@
...
+	peerPubKey [33]byte) error {
+
+	// Construct the key for which we'll find this message with in the
+	// store.
+	msgKey, err := messageStoreKey(msg, peerPubKey)
+	if err != nil {
+		return err
+	}
+
+	return s.db.Update(func(tx *bbolt.Tx) error {
```
Roasbeef: Similar comment re `Batch`.

discovery/gossiper.go
```diff
@@ -814,126 +809,69 @@ func (d *deDupedAnnouncements) Emit() []msgWithSenders {
...
-		if err := bucket.ForEach(func(k, v []byte) error {
-			// The database value represents the encoded
-			// AnnounceSignatures message.
-			r := bytes.NewReader(v)
-			msg := &lnwire.AnnounceSignatures{}
-			if err := msg.Decode(r, 0); err != nil {
-				return err
-			}
+		for _, msg := range msgsToResend {
+			msg := msg.(*lnwire.AnnounceSignatures)
```
Roasbeef: No type assertion to ensure it's actually an `AnnounceSignatures`. Should also be a type switch since this can now have other message types.

discovery/gossiper.go
```diff
@@ -124,9 +124,12 @@ type Config struct {
...
 
-	// DB is a global boltdb instance which is needed to pass it in waiting
-	// proof storage to make waiting proofs persistent.
-	DB *channeldb.DB
+	// WaitingProofStore is a persistent storage of partial channel proof
+	// announcement messages. We use it to buffer half of the material
+	// needed to reconstruct a full authenticated channel announcement.
+	// Once we receive the other half the channel proof, we'll be able to
+	// properly validate it and re-broadcast it out to the network.
+	WaitingProofStore *channeldb.WaitingProofStore
```
Roasbeef: Should add a TOOD here to make this an interface in order to not depend directly on `channeldb` for this functionality.

wpaulino: Done.

discovery/gossiper.go
wpaulino: At this point, it's not possible for other messages to exist within it. In the follow-up PR (#1595) most of this diff will be removed/modified.

discovery/messagestore.go
wpaulino: Fixed.

PR #2474, opened 2019-01-15
lnpeer+brontide: reduce memory footprint using read/write pools for message encode/decode

cfromknecht:
This PR is a follow up to #2419 and #2385, introducing an `lnpeer.WritePool` and `brontide.ReadPool`.  Requests to read/write messages are submitted to the respective pool, which maintains a set of workers that handle the encoding/decoding. This builds on the prior PRs, since each pool is backed by a `lnpeer.WriteBufferPool` and `brontide.ReadBufferPool`, ensuring that buffer used in each are efficiently reused before being returned to the runtime. 

With #2419 and #2385, the total memory that will be allocated is linear in the number of peers/connections maintained by the daemon. With the read and write pools, we modify that to be at most linear in the _number of specified workers_. More importantly, these changes completely decouple read and write buffer allocations from the peer/connection lifecycle, allowing LND to tolerate flapping peers with minimal overhead.

Nodes that have a large number of peers will see the most drastic benefit. In testing, I was able to connect (w/o gossip queries) to nearly 900 unique nodes, all while keeping LNDs total memory under 15 MB. This configuration could have easily connected to more nodes, though that was all that reachable via the bootstrapper.

This same test used between 90-100MB on master, and continues to grow as more connections are established. In contrast, the memory used with read/write pools remains constant even as more peers are added.

Note that of the 15MB, only about 1MB is actually from the read/write pools, the remaining overhead is created by other subsystems operating in LND, and is reflected in both measurements. Coupled with more efficient recycling of unused buffers, I'd expect the reduction in memory related to read/write buffers to be one or two orders of magnitude for most nodes.

Depends on:
 - [x] #2385 
 - [x] #2419

### DIFF
--- a/brontide/noise.go
+++ b/brontide/noise.go
@@ -8,15 +8,12 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"runtime"
 	"time"
 
 	"golang.org/x/crypto/chacha20poly1305"
 	"golang.org/x/crypto/hkdf"
 
 	"github.com/btcsuite/btcd/btcec"
-	"github.com/lightningnetwork/lnd/buffer"
-	"github.com/lightningnetwork/lnd/pool"
 )
 
 const (
@@ -60,14 +57,6 @@ var (
 	ephemeralGen = func() (*btcec.PrivateKey, error) {
 		return btcec.NewPrivateKey(btcec.S256())
 	}
-
-	// readBufferPool is a singleton instance of a buffer pool, used to
-	// conserve memory allocations due to read buffers across the entire
-	// brontide package.
-	readBufferPool = pool.NewReadBuffer(
-		pool.DefaultReadBufferGCInterval,
-		pool.DefaultReadBufferExpiryInterval,
-	)
 )
 
 // TODO(roasbeef): free buffer pool?
@@ -378,15 +367,6 @@ type Machine struct {
 	// next ciphertext header from the wire. The header is a 2 byte length
 	// (of the next ciphertext), followed by a 16 byte MAC.
 	nextCipherHeader [lengthHeaderSize + macSize]byte
-
-	// nextCipherText is a static buffer that we'll use to read in the
-	// bytes of the next cipher text message. As all messages in the
-	// protocol MUST be below 65KB plus our macSize, this will be
-	// sufficient to buffer all messages from the socket when we need to
-	// read the next one. Having a fixed buffer that's re-used also means
-	// that we save on allocations as we don't need to create a new one
-	// each time.
-	nextCipherText *buffer.Read
 }
 
 // NewBrontideMachine creates a new instance of the brontide state-machine. If
@@ -738,8 +718,28 @@ func (b *Machine) WriteMessage(w io.Writer, p []byte) error {
 // ReadMessage attempts to read the next message from the passed io.Reader. In
 // the case of an authentication error, a non-nil error is returned.
 func (b *Machine) ReadMessage(r io.Reader) ([]byte, error) {
-	if _, err := io.ReadFull(r, b.nextCipherHeader[:]); err != nil {
+	pktLen, err := b.ReadHeader(r)
+	if err != nil {
 		return nil, err
+	}
+
+	buf := make([]byte, pktLen)
+	return b.ReadBody(r, buf)
+}
+
+// ReadHeader attempts to read the next message header from the passed
+// io.Reader. The header contains the length of the next body including
+// additional overhead of the MAC. In the case of an authentication error, a
+// non-nil error is returned.
+//
+// NOTE: This method SHOULD NOT be used in the case that the io.Reader may be
+// adversarial and induce long delays. If the caller needs to set read deadlines
+// appropriately, it is preferred that they use the split ReadHeader and
+// ReadBody methods so that the deadlines can be set appropriately on each.
+func (b *Machine) ReadHeader(r io.Reader) (uint32, error) {
+	_, err := io.ReadFull(r, b.nextCipherHeader[:])
+	if err != nil {
+		return 0, err
 	}
 
 	// Attempt to decrypt+auth the packet length present in the stream.
@@ -747,34 +747,30 @@ func (b *Machine) ReadMessage(r io.Reader) ([]byte, error) {
 		nil, nil, b.nextCipherHeader[:],
 	)
 	if err != nil {
-		return nil, err
+		return 0, err
 	}
 
-	// If this is the first message being read, take a read buffer from the
-	// buffer pool. This is delayed until this point to avoid allocating
-	// read buffers until after the peer has successfully completed the
-	// handshake, and is ready to begin sending lnwire messages.
-	if b.nextCipherText == nil {
-		b.nextCipherText = readBufferPool.Take()
-		runtime.SetFinalizer(b, freeReadBuffer)
-	}
-
-	// Next, using the length read from the packet header, read the
-	// encrypted packet itself.
+	// Compute the packet length that we will need to read off the wire.
 	pktLen := uint32(binary.BigEndian.Uint16(pktLenBytes)) + macSize
-	if _, err := io.ReadFull(r, b.nextCipherText[:pktLen]); err != nil {
-		return nil, err
-	}
 
-	// TODO(roasbeef): modify to let pass in slice
-	return b.recvCipher.Decrypt(nil, nil, b.nextCipherText[:pktLen])
+	return pktLen, nil
 }
 
-// freeReadBuffer returns the Machine's read buffer back to the package wide
-// read buffer pool.
-//
-// NOTE: This method should only be called by a Machine's finalizer.
-func freeReadBuffer(b *Machine) {
-	readBufferPool.Return(b.nextCipherText)
-	b.nextCipherText = nil
+// ReadBody attempts to ready the next message body from the passed io.Reader.
+// The provided buffer MUST be the length indicated by the packet length
+// returned by the preceding call to ReadHeader. In the case of an
+// authentication eerror, a non-nil error is returned.
+func (b *Machine) ReadBody(r io.Reader, buf []byte) ([]byte, error) {
+	// Next, using the length read from the packet header, read the
+	// encrypted packet itself into the buffer allocated by the read
+	// pool.
+	_, err := io.ReadFull(r, buf)
+	if err != nil {
+		return nil, err
+	}
+
+	// Finally, decrypt the message held in the buffer, and return a
+	// new byte slice containing the plaintext.
+	// TODO(roasbeef): modify to let pass in slice
+	return b.recvCipher.Decrypt(nil, nil, buf)
 }
--- a/peer.go
+++ b/peer.go
@@ -44,8 +44,8 @@ const (
 	// idleTimeout is the duration of inactivity before we time out a peer.
 	idleTimeout = 5 * time.Minute
 
-	// writeMessageTimeout is the timeout used when writing a message to peer.
-	writeMessageTimeout = 50 * time.Second
+	// writeMessageTimeout is the timeout used when writing a message to a peer.
+	writeMessageTimeout = 10 * time.Second
 
 	// readMessageTimeout is the timeout used when reading a message from a
 	// peer.
--- a/peer.go
+++ b/peer.go
@@ -18,7 +18,6 @@ import (
 	"github.com/davecgh/go-spew/spew"
 
 	"github.com/lightningnetwork/lnd/brontide"
-	"github.com/lightningnetwork/lnd/buffer"
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/contractcourt"
@@ -26,6 +25,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnpeer"
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/pool"
 	"github.com/lightningnetwork/lnd/ticker"
 )
 
@@ -209,11 +209,11 @@ type peer struct {
 	// TODO(halseth): remove when link failure is properly handled.
 	failedChannels map[lnwire.ChannelID]struct{}
 
-	// writeBuf is a buffer that we'll re-use in order to encode wire
-	// messages to write out directly on the socket. By re-using this
-	// buffer, we avoid needing to allocate more memory each time a new
-	// message is to be sent to a peer.
-	writeBuf *buffer.Write
+	// writePool is the task pool to that manages reuse of write buffers.
+	// Write tasks are submitted to the pool in order to conserve the total
+	// number of write buffers allocated at any one time, and decouple write
+	// buffer allocation from the peer life cycle.
+	writePool *pool.Write
 
 	queueQuit chan struct{}
 	quit      chan struct{}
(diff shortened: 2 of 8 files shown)


Roasbeef:
All the dependent PRs have been merged now, so this needs a rebase!

peer.go
```diff
@@ -1366,32 +1361,32 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
...
-	b := bytes.NewBuffer(p.writeBuf[0:0:len(p.writeBuf)])
-
-	// With the temp buffer created and sliced properly (length zero, full
-	// capacity), we'll now encode the message directly into this buffer.
-	_, err := lnwire.WriteMessage(b, msg, 0)
-	if err != nil {
-		return err
-	}
+	var n int
+	err := p.writePool.Submit(func(buf *bytes.Buffer) error {
```
Roasbeef: I really dig this API!

peer.go
```diff
@@ -1366,34 +1361,37 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
...
 		return err
 	}
 
-	// Finally, write the message itself in a single swoop.
-	n, err := p.conn.Write(b.Bytes())
-
-	// Regardless of the error returned, record how many bytes were written
-	// to the wire.
-	atomic.AddUint64(&p.bytesSent, uint64(n))
+	// Record the number of bytes written on the wire, if any.
```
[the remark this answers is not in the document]
Roasbeef: Which err check?

server.go
```diff
@@ -262,15 +262,19 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
 
 	s := &server{
-		chanDB:          chanDB,
-		cc:              cc,
-		sigPool:         lnwallet.NewSigPool(runtime.NumCPU()*2, cc.signer),
-		writeBufferPool: writeBufferPool,
+		chanDB:    chanDB,
+		cc:        cc,
+		sigPool:   lnwallet.NewSigPool(runtime.NumCPU()*2, cc.signer),
+		writePool: writePool,
```
Roasbeef: Don't see the pool ever started.

brontide/read_pool.go
```diff
@@ -0,0 +1,179 @@
...
+	return <-req.errChan
+}
+
+// spawnWorker creates a worker goroutine that processes the initial task it is
+// given, then attempts to take new requests from the pool's work channel. If no
+// requests are received before the worker's timeout expires, the worker will
+// exit and return the buffer it was using to the buffer pool.
+//
+// NOTE: This method MUST be run as a goroutine.
+func (p *ReadPool) spawnWorker(req *readRequest) {
```
Roasbeef: The code overlap between this and the write pool is pretty substantial. I can think of a few ways to consolidate the two, but most of them involve either the addition of an `interface{}` here and there or additional type assertions. Both these methods would incur an additional performance hit as they need to be converted back into concrete types at the end points...

peer.go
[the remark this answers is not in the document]
cfromknecht: already been fixed in a prior force push

lnpeer/write_pool.go
```diff
@@ -0,0 +1,180 @@
...
+var ErrWritePoolExiting = errors.New("write pool shutting down")
+
+// writeRequest is tuple consisting of the function to execute and an error
+// channel for returning errors to the caller.
+type writeRequest struct {
+	fn      func(*bytes.Buffer) error
+	errChan chan error
+}
+
+// WritePool is a worker pool that can be used to serialize lnwire messages and
```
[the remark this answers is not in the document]
cfromknecht: currently we have `pool.WriteBuffer` and `pool.ReadBuffer`, so `pool.Write` is open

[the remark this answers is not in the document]
cfromknecht: seeing if this can be refactored to use a generic `pool.Worker` underneath, and then each passes in some closures that do the rest of the magic

lnpeer/write_pool.go
```diff
@@ -0,0 +1,180 @@
...
+	// Otherwise, submit the task to any of the active workers.
+	case p.work <- req:
+
+	case <-p.quit:
+		return ErrWritePoolExiting
+	}
+
+	// Wait for task to be processed. We don't select on quit here because
+	// we want to ensure we only return after the function has finished.
+	return <-req.errChan
```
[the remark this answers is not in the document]
cfromknecht: yes i think you're right, updating to select on quit as well

server.go
cfromknecht: added explicit start/stop methods

lnpeer/write_pool.go
```diff
@@ -0,0 +1,180 @@
...
+	"bytes"
+	"errors"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// DefaultWorkerTimeout is the default interval after which a worker will return
+// it's write buffer to the buffer pool if no write requests are submitted.
+const DefaultWorkerTimeout = 5 * time.Second
```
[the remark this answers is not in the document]
cfromknecht: not really, this dictates how long we will hold on to buffers taken from the pool if no new tasks are received, the other is how long we'll keep the returned buffers in the pool if no new workers are spawned. they can be fairly independent, though probably some room for tuning

cfromknecht:
@Roasbeef `pool.Read` and `pool.Write` now reside in the `pool` package, lmk how you like the new structure. some lingering godocs left to finish in the new unit tests, but o/w should be ready to another review, ptal

pool/worker.go
```diff
@@ -0,0 +1,250 @@
...
+				return
+			}
+
+		case <-w.quit:
+			return
+		}
+	}
+}
+
+// spawnWorker is used when the Worker pool wishes to create a new worker
```
Roasbeef: 👍 

Now that this has been abstracted even more it can be used in other areas of the codebase where we require a pool of works (with a burst limit) to process a series of tasks. With some tweaking, it'll likely also possible to convert the existing `lnwallet.SigPool` to use this new scaffolding.

pool/write.go
```diff
@@ -0,0 +1,100 @@
...
+func (w *Write) Stop() error {
+	return w.workerPool.Stop()
+}
+
+// Submit accepts a function closure that provides access to a fresh
+// bytes.Buffer backed by a buffer.Write object. The function's execution will
+// be allocated to one of the underlying Worker pool's goroutines.
+func (w *Write) Submit(inner func(*bytes.Buffer) error) error {
+	return w.workerPool.Submit(func(s WorkerState) error {
+		state := s.(*writeWorkerState)
```
Roasbeef: Just to be on the safe side, I think we should check the second value and error our here to avoid a panic if it isn't our expected type. This is unlikely to happen, but if we check the value, then a bug in a lower level of abstraction won't cause lnd to crash all together.

pool/read.go
```diff
@@ -0,0 +1,87 @@
...
+func (r *Read) Stop() error {
+	return r.workerPool.Stop()
+}
+
+// Submit accepts a function closure that provides access to the fresh
+// buffer.Read object. The function's execution will be allocated to one of the
+// underlying Worker pool's goroutines.
+func (r *Read) Submit(inner func(*buffer.Read) error) error {
+	return r.workerPool.Submit(func(s WorkerState) error {
+		state := s.(*readWorkerState)
```
Roasbeef: Similar comment here re checking the second type assertion value.

brontide/noise.go
```diff
@@ -68,8 +68,22 @@ var (
...
+	// read messages from brontide connections.
+	readPool = pool.NewRead(
+		readBufferPool, runtime.NumCPU(), pool.DefaultWorkerTimeout,
+	)
 )
 
+func init() {
+	// TODO(conner): pass read pool to listeners and dialers to avoid
+	// pkg-level read pool
+	if err := readPool.Start(); err != nil {
```
Roasbeef: So then the pool can never be stopped once the package is imported?

brontide/noise.go
```diff
@@ -750,31 +756,30 @@ func (b *Machine) ReadMessage(r io.Reader) ([]byte, error) {
...
-	}
-
-	// Next, using the length read from the packet header, read the
-	// encrypted packet itself.
+	// Compute the packet length that we will need to read off the wire.
 	pktLen := uint32(binary.BigEndian.Uint16(pktLenBytes)) + macSize
-	if _, err := io.ReadFull(r, b.nextCipherText[:pktLen]); err != nil {
+
+	var rawBytes []byte
+	err = readPool.Submit(func(buf *buffer.Read) error {
```
Roasbeef: Isn't it the case that this `io.Reader` above has already been obtained from the pool? At least in our current use case where this is _always_ called from within `peer.go`. So as is, this will two two items from the pool rather than a single item.

brontide/noise.go
```diff
@@ -743,7 +748,19 @@ func (b *Machine) WriteMessage(w io.Writer, p []byte) error {
 // ReadMessage attempts to read the next message from the passed io.Reader. In
 // the case of an authentication error, a non-nil error is returned.
 func (b *Machine) ReadMessage(r io.Reader) ([]byte, error) {
-	_, err := io.ReadFull(r, b.nextCipherHeader[:])
+	conn, ok := r.(net.Conn)
```
Roasbeef: This leaks the concrete implementation of the `io.Reader`. What's the rationale of pushing the deadline iteration down into the brontide layer rather than keeping it one step above at the peer?

brontide/noise.go
```diff
@@ -743,7 +748,19 @@ func (b *Machine) WriteMessage(w io.Writer, p []byte) error {
...
+	if !ok {
+		return nil, fmt.Errorf("net.Conn required to read message")
+	}
+
+	// Set the read timeout to zero, which will cause this method to block
```
Roasbeef: I'm not sure we actually want this. If the other party writes half the header, then we'll just sit here. Since this'll block forever, the idle timer in the peer also won't be able to tick on its own and disconnect the peer.

peer.go
```diff
@@ -43,8 +43,8 @@ const (
 	// idleTimeout is the duration of inactivity before we time out a peer.
 	idleTimeout = 5 * time.Minute
 
-	// writeMessageTimeout is the timeout used when writing a message to peer.
-	writeMessageTimeout = 50 * time.Second
+	// writeMessageTimeout is the timeout used when writing a message to a peer.
+	writeMessageTimeout = 10 * time.Second
```

brontide/noise.go
Roasbeef: Ok, I see the diff has been updated now to remove the pool usage from the peer and instead have it at the brontide layer. If we move it back into the peer, then we don't need to burden the brontide layer with things like read timeous. For example, as is now, I can no longer ask the `brontide.Machine` to read a message from a file on-disk that really just contains a set of regular packed brontide messages.

pool/write.go
cfromknecht: i figured that was the purpose of introducing the unit tests?

brontide/noise.go
cfromknecht: that is the current behavior

brontide/noise.go
cfromknecht: The reader isn't obtained from the pool, it's access to the read buffer that's scheduled. 

> Ok, I see the diff has been updated now to remove the pool usage from the peer and instead have it at the brontide layer.

read pool usage was always in the brontide layer, only write pool usage was in the peer.

this is more of an artifact of the framing protocol, since we can't read/write incrementally to the reader/writer. because of the framing, we have to do
writes:
 1. peer encodes full message (requires write pool)
 2. pass message.bytes() to conn
 3. brontide internally writes full bytes to the wire

reads:
 1. brontide reads and decrypt full message (requires read pool)
 2. retrieve message bytes out of conn
 3. peer decodes full message

step 2 in both cases is where the abstraction should logically happen, but the framing prevents that

brontide/noise.go
cfromknecht: yeah, not my favorite thing. my longer term solution is to initialize listeners and create a new `Dialer` object that has the same semantics of `brontide.Dial`, but which can get its own read pool. then the pool can be started stopped w/ more control. This would be a bigger change, but happy to make that

brontide/noise.go
cfromknecht: the abstraction is already somewhat broken imo bc it doesn't distinguish between adversarial vs non-adversarial `io.Reader` instances, e.g. connections vs files. I considered adding a separate `ReadConn` method to handle this

setting the timeout at the peer level will likely cause reads to timeout while the read is being scheduled, since it is unaware of when the task is being scheduled by the pool.

there may also be anan arbitrary amount of time between reading payload lengths (which we want to block for), though the header is received we expect the body of the message to follow shortly after. this was also less ugly than having to have something like this in the peer:
```golang
retry:
conn.SetReadDeadline(deadline)
_, err := conn.ReadMessage()
if nerr := err.(net.Error); nerr.Timeout() {
     goto retry
}
```
as this would mean the read timeout will be triggered multiple times if the peer doesn't send us anything for a minute and we want the read timeout to be something short like 5s

pool/worker.go
cfromknecht: indeed i'm interested to see how the `SigPool` gets adapted :)

brontide/noise.go
cfromknecht: also the idle timer fires a `Disconnect` via an `AfterFunc`, so blocking arbitrarily long doesn't prevent the timer from being triggered: https://github.com/lightningnetwork/lnd/blob/master/peer.go#L958

brontide/noise.go
cfromknecht: another consequence of setting the read deadline at the peer level is that we may successfully read the message header, but then timeout before/while reading the body. in this case, and w/ the retry logic, we will try again and start reading a body as if it were the header since we aren't distinguishing timeouts on header vs body

brontide/noise.go
cfromknecht: > For example, as is now, I can no longer ask the `brontide.Machine` to read a message from a file on-disk that really just contains a set of regular packed brontide messages.

we could instead have the type assertions suround the calls to `SetReadDeadline`, instead of erroring if the reader does not match?

cfromknecht: in fact, i think this is actually necessary for the unit tests to pass

brontide/noise.go
cfromknecht: we can push the read timeout abstraction up a bit by splitting `ReadMessage` into `ReadHeader` and `ReadBody`. this way the caller can block arbitrarily long on headers, and then set a real deadline for bodies. because of the scheduling tho, it would still be more prone to timing out while being scheduled than if the timeout is set once the task is scheduled

I think we could get around this a little more by having `ReadMessage` accept the buffer we want to read into, then from the peer we can do:
```
conn.SetReadDeadline(time.Time{})
conn.ReadHeader()
readPool.Submit(func(b *buffer.Read) error {
    conn.SetReadDeadline(time.Now().Add(readTimeout)
    rawBytes, err := conn.ReadNextBody(b) // internally calls Noise.ReadMessage(reader)
    return err
})
```
how does that sound?

cfromknecht: i think this also allows us to keep `ReadMessage` and note that it should only be used for non-adverserial readers, and the caller needs more control they should use the split header/body interface. interested in hearing your thoughts

cfromknecht:
@Roasbeef i've added some commits that implement a PoC split `ReadNextHeader`/`ReadNextBody`, while still retaining the unified `ReadMessage` method for non-adversarial cases. I must say, I much prefer this approach as it pushes all timeout handling into the peer and keeps the brontide package much simpler :)

brontide/noise.go
Roasbeef: Nice, I think that last fragment gives us the best of both worlds. Also nice to patch the issue that exists right now that allows a peer to block us by writing the brontide header and nothing else. 

Re the first comment, I don't think it's the case that the abstraction is broken if it doesn't distinguish between sockets, files, etc. On the contrary, that's the point of having something like an `io.Reader` in the first place: we don't care what the underlying implementation _is_.

capacitynetwork:
After experiencing https://github.com/lightningnetwork/lnd/issues/2726 I have changed `writeMessageTimeout` to `25` on my node and it appears to have helped.

@cfromknecht was there a reason for change from 50 to 10?

PR #2477, opened 2019-01-16
lnrpc: replace ChanPoint w/ OutPoint in ListUnspent resp

cfromknecht:
Follow up from #1984, this replaces the use of `lnrpc.ChannelPoint` with `lnrpc.OutPoint`. Currently the json key produced in the list unspent response will print `funding_txid_str`, which is now changed to just `txid_str`.


cfromknecht:
You’re right, i wonder then how it passed locally and on travis

cfromknecht:
@wpaulino figured it out, needed to update some grpc deps. PTAL

halseth:
Code LGTM, but question is whether we should wait with this for 0.6, as it is a semi-breaking change :p

cfromknecht:
`ListUnspent` has never been tagged in a release or rc, waiting another release cycle will only make this more breaking for more users ;)

halseth:
> ListUnspent has never been tagged in a release or rc

Aaah, in that case we're good 👍

lnrpc/rpc.proto
```diff
@@ -802,6 +800,19 @@ message ChannelPoint {
...
+                /// Raw bytes representing the transaction id.
+                bytes txid_bytes = 1 [json_name = "txid_bytes"];
+
+                /// Hex-encoded string representing the transaction id.
+                string txid_str = 2 [json_name = "txid_str"];
+        }
+
+        /// The index of the output on the transaction.
+        uint32 output_index = 3 [json_name = "output_index"];
+}
```
joostjager: Tabs off (8 instead of 4)?

lnrpc/rpc.proto
```diff
@@ -684,9 +684,7 @@ message Utxo {
 	string script_pubkey = 4 [json_name = "script_pubkey"];
 
 	/// The outpoint in format txid:n
-	/// Note that this reuses the `ChannelPoint` message but
-	/// is not actually a channel related outpoint, of course
-	ChannelPoint outpoint = 5 [json_name = "outpoint"];
+	OutPoint outpoint = 5 [json_name = "outpoint"];
```
joostjager: Can we remove message ChannelPoint completely? It it the same message now, and channel point is an outpoint too.

lnrpc/rpc.proto
cfromknecht: We have different tabwidths for protos? 🤦‍♂️ 

Starting to think we should commit a default vimrc in the tree...

Roasbeef:
This can land, but it would be in 0.6 rather than 0.5.2 as it's a new RPC all together.

cfromknecht:
Proto formatting fixed, @roasbeef @halseth @joostjager PTAL

lnrpc/rpc.proto
joostjager: Does it also break if you leave the `ChannelPoint` message in to not get errors from people building against the compiled proto, but really use the `OutPoint` type in all the calls? Not sure if the actual type is also serialized. If not, it may just work if they are identical.

halseth: Not sure what you mean, but projects importing the protos will have to do the name change (similar to what is done in `rpcserver` here). The serialized version will be the same as long as the field number (`5`) is the same IIRC.

halseth:
> This can land, but it would be in 0.6 rather than 0.5.2 as it's a new RPC all together.

Should be included in the first release that also includes #1984.

lnrpc/rpc.proto
joostjager: Yes, I meant just keeping the ChannelPoint message in so that dependents don't need to rename and just keep working because ChannelPoint and OutPoint are identical. But maybe not such a good idea.

cfromknecht:
@halseth @joostjager @Roasbeef @wpaulino updated  the `OutPoint` proto type to set both txid_bytes and txid_str, and then have lncli filter for only txid_str since the txid_bytes will be printed in base64 which is ugly.

This however, makes the rpc intreface more usable as people can copy the raw txid_bytes or display the txid_str depending on their needs. Previously we used a oneof and only set txid_str, so the client would be force to parse the txid_str, decode hex, and reverse the bytes if they wanted to obtain the raw bytes

cmd/lncli/types.go
```diff
@@ -0,0 +1,37 @@
...
+	AmountSat     int64             `json:"amount_sat"`
+	ScriptPubkey  string            `json:"script_pubkey"`
+	OutPoint      OutPoint          `json:"outpoint"`
+	Confirmations int64             `json:"confirmations"`
+}
+
+// NewUtxoFromProto creates a display Utxo from the Utxo proto. This filters out
+// the raw txid bytes from the provided outpoint, which will otherwise be
+// printed in base64.
+func NewUtxoFromProto(utxo *lnrpc.Utxo) *Utxo {
```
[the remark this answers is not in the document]
cfromknecht: because it will prevent code duplication if other commands need to do the same thing

cmd/lncli/types.go
```diff
@@ -0,0 +1,37 @@
...
+	OutputIndex uint32 `json:"output_index"`
+}
+
+// Utxo displays information about an unspent output, icluding its address,
+// amount, pkscript, and confirmations.
+type Utxo struct {
+	Type          lnrpc.AddressType `json:"address_type"`
+	Address       string            `json:"address"`
+	AmountSat     int64             `json:"amount_sat"`
+	ScriptPubkey  string            `json:"script_pubkey"`
```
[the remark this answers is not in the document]
cfromknecht: see #2556

cmd/lncli/types.go
```diff
@@ -0,0 +1,37 @@
...
+	OutPoint      OutPoint          `json:"outpoint"`
```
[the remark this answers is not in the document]
cfromknecht: this is how `listunspent` currently serializes the outpoints, but that can be changed

cmd/lncli/types.go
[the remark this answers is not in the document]
cfromknecht: figure it's easier to have one place where define what an `OutPoint` or `Utxo` looks like when displayed over lncli, than the reimplment it every time

cfromknecht:
@Roasbeef updated lncli output to print `lnrpc.Outpoint` in `<txid>:<output-index>` form, PTAL

joostjager:
I still have mixed feelings about the redundancy in returning a txid in two formats. What was again the reason for not just sticking to the most efficient format `bytes`? For `lncli` is doesn't matter, we can convert locally.

Also if we use `OutPoint` to pass data into `lnd` in the future, we need to make sure we always use a helper function to either take the raw bytes or parse the string.

lnrpc/rpc.proto
```diff
@@ -800,6 +798,17 @@ message ChannelPoint {
     uint32 output_index = 3 [json_name = "output_index"];
 }
 
+message OutPoint {
+    /// Raw bytes representing the transaction id.
+    bytes txid_bytes = 1 [json_name = "txid_bytes"];
+
+    /// Reversed, hex-encoded string representing the transaction id.
+    string txid_str = 2 [json_name = "txid_str"];
```
joostjager: I still have mixed feelings about the redundancy in returning a txid in two formats. What was again the reason for not just sticking to the most efficient format bytes? For lncli is doesn't matter, we can convert locally.

Also if we use OutPoint to pass data into lnd in the future, we need to make sure we always use a helper function to either take the raw bytes or parse the string.

Roasbeef: Mainly for when it's served over REST. If we don't encode ourselves, then it becomes base58 when one is expecting a hex encoded txid. In certain areas we don't encode raw bytes for ourselves manually, which causes confusion for users when for example, their payment preimage is base 58 encoded.

cmd/lncli/types.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// OutPoint displays an outpoint string in the form "<txid>:<output-index>".
+type OutPoint string
+
+// NewOutPointFromProto formats the lnrpc.OutPoint into an OutPoint for display.
+func NewOutPointFromProto(op *lnrpc.OutPoint) OutPoint {
+	return OutPoint(fmt.Sprintf("%s:%d", op.TxidStr, op.OutputIndex))
```
Roasbeef: /notbadface

cfromknecht:
Rebased to resolve conflict in `lnrpc/gen_protos.sh`. @Roasbeef @joostjager @wpaulino PTAL

rpcserver.go
```diff
@@ -717,10 +717,9 @@ func (r *rpcServer) ListUnspent(ctx context.Context,
 
 		// Now that we know we have a proper mapping to an address,
 		// we'll convert the regular outpoint to an lnrpc variant.
-		outpoint := &lnrpc.ChannelPoint{
-			FundingTxid: &lnrpc.ChannelPoint_FundingTxidStr{
-				FundingTxidStr: utxo.OutPoint.Hash.String(),
-			},
+		outpoint := &lnrpc.OutPoint{
+			TxidBytes:   utxo.OutPoint.Hash[:],
```
wpaulino: This is something I noticed the other day - this hash is actually reversed. The `String()` method is the one that actually reverses it. Not that it should/shouldn't be changed, but just something to note.

cfromknecht: @wpaulino yep, that's how bitcoind txids are serialized

wpaulino: I'm aware - have just seen users confused on slack about this.

cfromknecht: blame satoshi 🤷‍♂️not much we can do

wpaulino: I guess as a convenience we should just reverse them for the user anyway? Can be left as a follow-up though.

cfromknecht: why? these _are_ the raw bytes. the purpose of providing them is so that the user doesn't have to remember to reverse them. if it was simply a matter of decoding hex, then we wouldn't bother adding it

cfromknecht: you should be able to take this output and copy it into a `chainhash.Hash`. If we provide them reversed, now the user has to reverse them again

lnrpc/rpc.proto
joostjager: I see. It would be nicer if this was somehow transparently done in the proxy generation. Mark the txid fields with an attribute and in the proxy the conversion to reversed hex will be made.

PR #2485, opened 2019-01-16
[lnwire+funding] Don't send ErrorCode on wire

halseth:
Since the ErrorCodes are not part of the spec, they cannot be read by
other implementations.

Instead of only sending the error code we therefore send the complete
error message.


Roasbeef:
Soon (hopefully), we'll have proper error code support on the protocol level.

halseth:
Maybe keep everything around then, but send the full error description on the wire?

PR #2489, opened 2019-01-17
docs: better instructions for installing go-lang on Linux

lucash-dev:
Installation instructions for Linux were imprecise (don't work on all kinds of Linux, so should specify it's just for debian-based systems).
Also doesn't work in some versions of Ubuntu, so added link to go-lang documentation that has more details.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -31,10 +31,12 @@
     [`1.11`](https://blog.golang.org/go1.11).
 
 
-    On Linux:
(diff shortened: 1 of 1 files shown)


PatrickLemke:
This makes sense. If you want to install on a raspberry pi for example, there are no standard deb repositories for the newest golang versions.

Roasbeef:
See #2402.

lucash-dev:
That’s much better. Thanks!

PR #2490, opened 2019-01-17
Tor null auth and listen fixes

### LINKED ISSUE #2176: Can't connect to onion nodes when using remote Tor
### Background

I'm using Whonix, which operates as a workstation and separate gateway where Tor is run. In order to use an onion service for my node I need to listen on `0.0.0.0`. In order to do that I have to disable Tor in my conf. This causes me to not be able to make connections to other onion nodes. I believe the problem is here: https://github.com/lightningnetwork/lnd/blob/d4b042dc1946ece8b60d538ade8e912f035612fe/server.go#L1659-L1663 but I don't know `go` enough to fix it.

### LINKED ISSUE #2388: Support unauthenticated Tor control port
Currently `lnd` can only make use of Tor's control port if the user is using an authentication cookie.

https://github.com/lightningnetwork/lnd/blob/089d461552ffc05d7b1c22d603af0fa82e788e85/tor/controller.go#L53-L55

If the user has no authentication method set the exchange with the control port looks like this:

```
PROTOCOLINFO 1
250-PROTOCOLINFO 1
250-AUTH METHODS=NULL
250-VERSION Tor="0.3.4.9 (git-074ca2e0054fded1)"
250 OK
```

In that case `lnd` fails with this in the debug log:

```
unable to retrieve authentication cookie: cookie file path not found in reply
```

However with this response `250-AUTH METHODS=NULL` you can just go ahead and skip all the cookie stuff and simply authenticate. Here is an example on the same system as `lnd`:

```
user@host:~$ nc -n 127.0.0.1 9051
PROTOCOLINFO 1
250-PROTOCOLINFO 1
250-AUTH METHODS=NULL
250-VERSION Tor="0.3.4.9 (git-074ca2e0054fded1)"
250 OK
AUTHENTICATE
250 OK
ADD_ONION NEW:ED25519-V3 Port=9735,9735 
250-ServiceID=kxz6mk2kjqlwt6pmffbr23kr2lwhz4xcf5fq65aqvhopy4cd23g4hkqd
250-PrivateKey=ED25519-V3:GEE9Tesr1Z0OSWzHE70WCrLVa7GcVLFsgCGoc0cnb1OUTC38Pm5vTZyGqraZkbIHoeuxJpL587+1ZnJRcRB/iA==
250 OK
```

It might be worth noting that the authentication cookie method would have worked on this system had `lnd` checked the default location for the cookie. The reason the cookie location wasn't provided from the control port is the use of [`onion-grater`](https://github.com/Whonix/onion-grater), a whitelist filter for the control port's responses.

aakselrod:
This PR should fix #2388 and #2176 in the following ways:

* allowing NULL authentication when attempting to create a hidden service automatically
* when automatic hidden service configuration is disabled, defaulting to listening on localhost and allowing manual override for manual or alternative hidden service configuration

This allows `lnd` to work better in environments where the Tor daemon lives on a different machine, such as Whonix or OnionPi, and where certain Tor control protocol requests and responses are filtered.

Fixes #2388 
Fixes #2176

### DIFF
--- a/config.go
+++ b/config.go
@@ -502,12 +502,6 @@ func loadConfig() (*config, error) {
 	case cfg.DisableListen && (cfg.Tor.V2 || cfg.Tor.V3):
 		return nil, errors.New("listening must be enabled when " +
 			"enabling inbound connections over Tor")
-	case cfg.Tor.Active && (!cfg.Tor.V2 && !cfg.Tor.V3):
-		// If an onion service version wasn't selected, we'll assume the
-		// user is only interested in outbound connections over Tor.
-		// Therefore, we'll disable listening in order to avoid
-		// inadvertent leaks.
-		cfg.DisableListen = true
 	}
 
 	if cfg.Tor.PrivateKeyPath == "" {
@@ -866,9 +860,14 @@ func loadConfig() (*config, error) {
 
 	// Listen on the default interface/port if no listeners were specified.
 	// An empty address string means default interface/address, which on
-	// most unix systems is the same as 0.0.0.0.
+	// most unix systems is the same as 0.0.0.0. If Tor is active, we
+	// default to only listening on localhost for hidden service
+	// connections.
 	if len(cfg.RawListeners) == 0 {
 		addr := fmt.Sprintf(":%d", defaultPeerPort)
+		if cfg.Tor.Active {
+			addr = fmt.Sprintf("localhost:%d", defaultPeerPort)
+		}
 		cfg.RawListeners = append(cfg.RawListeners, addr)
 	}
 
@@ -945,20 +944,6 @@ func loadConfig() (*config, error) {
 					p2pListener)
 				return nil, err
 			}
-		}
-	}
-
-	// Ensure that we are only listening on localhost if Tor inbound support
-	// is enabled.
-	if cfg.Tor.V2 || cfg.Tor.V3 {
-		for _, addr := range cfg.Listeners {
-			if lncfg.IsLoopback(addr.String()) {
-				continue
-			}
-
-			return nil, errors.New("lnd must *only* be listening " +
-				"on localhost when running with Tor inbound " +
-				"support enabled")
 		}
 	}
 
--- a/tor/controller.go
+++ b/tor/controller.go
@@ -163,7 +163,7 @@ func parseTorReply(reply string) map[string]string {
 }
 
 // authenticate authenticates the connection between the controller and the
-// Tor server using the SAFECOOKIE authentication method.
+// Tor server using the SAFECOOKIE or NULL authentication method.
 func (c *Controller) authenticate() error {
 	// Before proceeding to authenticate the connection, we'll retrieve
 	// the authentication cookie of the Tor server. This will be used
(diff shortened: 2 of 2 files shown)


qubenix:
Awesome of you to work on this! I'm just testing this out on Whonix.

For me the null auth fix worked, but I'm still running into an issue listening on something other than localhost.

```sh
user@host:~$ lnd
lnd must *only* be listening on localhost when running with Tor inbound support enabled
```

I think it may have to do with these lines: https://github.com/lightningnetwork/lnd/blob/570b60e75b2822fc737b46a223c944ceadfedaab/config.go#L950-L962

aakselrod:
@qubenix, I think you might be correct. You can work around it by manually configuring the hidden service for now, and I'll fix the other issue momentarily.

aakselrod:
I've made the fix and rebased. I tested this on my Whonix system with a brand new `lnd` installation and was able to successfully auto-create a v3 service (with a new `onion-grater` merge file for the `lnd` request) while listening on `eth0`.

qubenix:
Tested ACK: 288870f043289282f8858775f350a6ee835583e7 66a150297a803d8ed85381a3b53397c80358100f

tor/controller.go
```diff
@@ -176,6 +176,13 @@ func (c *Controller) authenticate() error {
 			"%v", err)
 	}
 
+	// If cookie is empty and there's no error, we have a NULL
+	// authentication method that we should use instead.
+	if len(cookie) == 0 {
```
wpaulino: As we introduce more auth methods, perhaps we should consider refactoring extracting the supported auth methods from this method into another one? So the flow would be something like:

```go
func authenticate() {
    supportedAuthMethods := getSupportedAuthMethods()
    switch {
    case len(cookie) == 0 && supportedAuthMethods.Contains("NULL"):
        // Proceed with NULL auth.
    case len(cookie) != 0 && supportedAuthMethods.Contains("SAFECOOKIE"):
        // Proceed with cookie auth.
    }
}
```

NicolasDorier:
Concept ACK, I needed this.

config.go
```diff
@@ -866,9 +860,14 @@ func loadConfig() (*config, error) {
...
 	// Listen on the default interface/port if no listeners were specified.
 	// An empty address string means default interface/address, which on
-	// most unix systems is the same as 0.0.0.0.
+	// most unix systems is the same as 0.0.0.0. If Tor is active, we
+	// default to only listening on localhost for hidden service
+	// connections.
 	if len(cfg.RawListeners) == 0 {
 		addr := fmt.Sprintf(":%d", defaultPeerPort)
+		if cfg.Tor.Active {
+			addr = fmt.Sprintf("localhost:%d", defaultPeerPort)
```
NicolasDorier: I don't think it makes sense.

I want to use Tor with LND as an alternative way for people to connect to me.

wpaulino: This will only happen if you don't explicitly set the `listen` flag. It can be overridden.

tzarebczan:
I'm running into this same problem...anyone have a windows build to test this out?

tzarebczan:
Awesome @Roasbeef, can't wait to try this out!

PR #2493, opened 2019-01-17
Validate addresses before writing them to DB

halseth:
Addresses that don't pass these added checks, would previously cause us to fail reading the node from the DB.

### DIFF
--- a/channeldb/addr_test.go
+++ b/channeldb/addr_test.go
@@ -3,6 +3,7 @@ package channeldb
 import (
 	"bytes"
 	"net"
+	"strings"
 	"testing"
 
 	"github.com/lightningnetwork/lnd/tor"
@@ -15,7 +16,7 @@ func (t unknownAddrType) String() string  { return "unknown" }
 
 var addrTests = []struct {
 	expAddr net.Addr
-	serErr  error
+	serErr  string
 }{
 	{
 		expAddr: &net.TCPAddr{
@@ -43,7 +44,7 @@ var addrTests = []struct {
 	},
 	{
 		expAddr: unknownAddrType{},
-		serErr:  ErrUnknownAddressType,
+		serErr:  ErrUnknownAddressType.Error(),
 	},
 }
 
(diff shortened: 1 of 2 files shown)


channeldb/addr_test.go
```diff
@@ -43,7 +44,7 @@ var addrTests = []struct {
 	},
 	{
 		expAddr: unknownAddrType{},
-		serErr:  ErrUnknownAddressType,
+		serErr:  ErrUnknownAddressType.Error(),
```
Roasbeef: Hmm, this seems like step backwards...do we want to use a proper type instead?

halseth: In case of invalid addresses, we want the invalid address to be part of the returned error to ease debugging, so we cannot use a static error for these cases. We _could_ make  a wrapping error type (as we do with `ReservationError`) but it seemed a bit overkill just to accommodate testing.

halseth: PTAL @Roasbeef

wpaulino: As an alternative, we could replace `serErr` with a `valid bool` instead. Is there any reason we're interested in the exact error of the serialization?

channeldb/addr_test.go
```diff
@@ -55,11 +56,21 @@ func TestAddrSerialization(t *testing.T) {
...
+
+		case err != nil && !strings.Contains(err.Error(), test.serErr):
 			t.Fatalf("unexpected serialization err for addr %v, "+
-				"want: %v, got %v",
-				test.expAddr, test.serErr, err)
-		} else if test.serErr != nil {
+				"want: %v, got %v", test.expAddr, test.serErr,
+				err)
+
+		case err != nil:
```
wpaulino: Nit: this case does not need to be explicit.

channeldb/addr_test.go
halseth: It lets you check that a it fails in the expected way. I guess checking the string is a middle ground between only checkin if any error at all and checking the exact type.

PR #2495, opened 2019-01-17
[trivial] lnd_test: don't shadow predErr

halseth:


lnd_test.go
```diff
@@ -2724,10 +2724,11 @@ func testChannelForceClosure(net *lntest.NetworkHarness, t *harnessTest) {
 			return false
 		}
 
-		forceClose, predErr := findForceClosedChannel(
+		forceClose, err := findForceClosedChannel(
```
joostjager: Isn't it time to introduce a `WaitNoError`? It keeps trying the wait func in a loop until it returns no error. If the timer expires, it returns the last error, similar to what is happening here with `predErr`.

cfromknecht: Indeed, that would help. Having an api that returns the errors directly would be very nice

joostjager: It is small change

halseth: Feel free to add it :) I would prefer it being added as we go rather than a massive change in `lnd_test` though.

joostjager: I don't mean a massive change. Just add the method and use it only here. Make a start with 'added as we go' .

cfromknecht: i'd support adding it here

halseth: Added! PTAL @cfromknecht @joostjager

lntest/harness.go
```diff
@@ -1155,6 +1155,28 @@ func WaitPredicate(pred func() bool, timeout time.Duration) error {
...
+// this doesn't happen within the timeout.
+func WaitNoError(f func() error, timeout time.Duration) error {
+	var predErr error
+	pred := func() bool {
+		if err := f(); err != nil {
+			predErr = err
+			return false
+		}
+		return true
+	}
```
joostjager: Nice wrapper

lntest/harness.go
```diff
@@ -1155,6 +1155,28 @@ func WaitPredicate(pred func() bool, timeout time.Duration) error {
 	}
 }
 
+// WaitNoError is a wrapper around WaitPredicate that waits for the passed
+// method f to execute without error, and returns the last error encountered if
+// this doesn't happen within the timeout.
+func WaitNoError(f func() error, timeout time.Duration) error {
+	var predErr error
+	pred := func() bool {
+		if err := f(); err != nil {
```
joostjager: Can assign to `predErr` directly?

halseth:
> You left some predErr in the same unit test 

Yep, that was the intention for now. Can change the rest as we go :)

lntest/harness.go
halseth: We _could_. I think I prefer this style though, makes it clear that it is assigned only in the error case.

PR #2496, opened 2019-01-17
[trivial] lntest/node: print amounts in same format

halseth:


lntest/node.go
```diff
@@ -938,7 +938,8 @@ func (hn *HarnessNode) WaitForBalance(expectedBalance int64, confirmed bool) err
 	err := WaitPredicate(doesBalanceMatch, 30*time.Second)
 	if err != nil {
 		return fmt.Errorf("balances not synced after deadline: "+
-			"expected %v, only have %v", expectedBalance, lastBalance)
+			"expected %v, only have %v",
+			btcutil.Amount(expectedBalance), lastBalance)
```
joostjager: Alternative could be to change the func argument to `btcutil.Amount`.

joostjager: `fmt.Errorf` call not formatted according to guide lines.

halseth: > Alternative could be to change the func argument to btcutil.Amount.

Yeah, I would also prefer that. Turned out we are using the proto fields (which are `int64`) directly many places, so the diff became too big to make it worth it I think.

> fmt.Errorf call not formatted according to guide lines.

We usually try to keep logs and errors over as few lines as possible.

halseth: Actually, the diff didn't become that bad, so I'll do what you initially suggested :)

joostjager: >We usually try to keep logs and errors over as few lines as possible.

Ah, I see now that that exception indeed landed in the contribution guidelines.

PR #2500, opened 2019-01-18
server: require the DLP bit for all incoming/outgoing connections

Roasbeef:
In this commit, we modify our default local feature bits to require the
Data Loss Protection (DLP) feature to be active. Once full Static
Channel Backups are implemented, if we connect to a peer that doesn't
follow the DLP protocol, then the SCBs are useless, as we may not be
able to recover funds. As a result, in prep for full SCB deployment,
we'll now ensure that any peer we connect to, knows of the DLP bit. This
could be a bit more relaxed and allow _connections_ to non-DLP peers,
but reject channel requests to/from them. However, this implementation is
much simpler.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -958,9 +958,9 @@ func (p *peer) readHandler() {
 		p.Disconnect(err)
 	})
 
-	// Initialize our negotiated gossip sync method before reading
-	// messages off the wire. When using gossip queries, this ensures
-	// a gossip syncer is active by the time query messages arrive.
+	// Initialize our negotiated gossip sync method before reading messages
+	// off the wire. When using gossip queries, this ensures a gossip
+	// syncer is active by the time query messages arrive.
 	//
 	// TODO(conner): have peer store gossip syncer directly and bypass
 	// gossiper?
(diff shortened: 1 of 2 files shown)


cfromknecht:
We also need to modify `peer.handleInitMsg` to disconnect if the other peer doesn't advertise optional or required support DLP. Right now, we will rely on the other peer to DC if they aren't advertising, so a bug on the remote end could cause connections to be accepted that should have been disconnected.

Roasbeef:
Pushed up a commit to now disconnect the peers that don't have this new feature bit. Before we merge this, I want to refactor this area to be a bit more general so we have a single place in the `peer` or `server` where manage which feature bits we advertise/require.

cfromknecht:
@Roasbeef are you still planning to the refactoring here?

Roasbeef:
Ideally yes. If y'all don't consider it a blocker, then we can have this land now, and I can follow up with a refactor+tests.

kind84:
I think that this merge has some issues, my node cannot bootstrap connections saying "unable to start peer: data loss protection required".

cfromknecht:
@kind84 indeed, see #2570 for fix

PR #2501, opened 2019-01-18
htlcswitch: batch preimage writes/consistency fix

cfromknecht:
This PR adds batched writes for witnesses discovered in HTLC forwarding. At the same time, we correct a nuanced consistency issue related to a lack of synchronization with the channel state machine. Forcing the individual preimage writes to be synchronized with the link incurs a heavy performance penalty (about 80% from my measurements). Batching these allows us to minimize the number of db transactions required to write the preimages, allowing us to reinsert the batched write into the link's critical path and resolve the possible inconsistency. In fact, the benchmarks actually showed a slight performance improvement, even with the extra write in the critical path.

See commit messages for more details.

### DIFF
--- a/channeldb/witness_cache.go
+++ b/channeldb/witness_cache.go
@@ -70,12 +70,42 @@ func (d *DB) NewWitnessCache() *WitnessCache {
 	}
 }
 
-// AddWitness adds a new witness of wType to the witness cache. The type of the
-// witness will be used to map the witness to the key that will be used to look
-// it up.
+// witnessEntry is a key-value struct that holds each key -> witness pair, used
+// when inserting records into the cache.
+type witnessEntry struct {
+	key     []byte
+	witness []byte
+}
+
+// AddWitnesses adds a batch of new witnesses of wType to the witness cache. The
+// type of the witness will be used to map each witness to the key that will be
+// used to look it up. All witnesses should be of the same WitnessType.
 //
 // TODO(roasbeef): fake closure to map instead a constructor?
-func (w *WitnessCache) AddWitness(wType WitnessType, witness []byte) error {
+func (w *WitnessCache) AddWitnesses(wType WitnessType, witnesses ...[]byte) error {
+	// Optimistically compute the witness keys before attempting to start
+	// the db transaction.
+	entries := make([]witnessEntry, 0, len(witnesses))
+	for _, witness := range witnesses {
+		// Map each witness to its key by applying the appropriate
+		// transformation for the given witness type.
+		switch wType {
+		case Sha256HashWitness:
+			key := sha256.Sum256(witness)
+			entries = append(entries, witnessEntry{
+				key:     key[:],
+				witness: witness,
+			})
+		default:
+			return ErrUnknownWitnessType
+		}
+	}
+
+	// Exit early if there are no witnesses to add.
+	if len(entries) == 0 {
+		return nil
+	}
+
 	return w.db.Batch(func(tx *bbolt.Tx) error {
 		witnessBucket, err := tx.CreateBucketIfNotExists(witnessBucketKey)
 		if err != nil {
@@ -93,16 +123,14 @@ func (w *WitnessCache) AddWitness(wType WitnessType, witness []byte) error {
 			return err
 		}
 
-		// Now that we have the proper bucket for this witness, we'll map the
-		// witness type to the proper key.
-		var witnessKey []byte
-		switch wType {
-		case Sha256HashWitness:
-			key := sha256.Sum256(witness)
-			witnessKey = key[:]
+		for _, entry := range entries {
+			err = witnessTypeBucket.Put(entry.key, entry.witness)
+			if err != nil {
+				return err
+			}
 		}
 
-		return witnessTypeBucket.Put(witnessKey, witness)
+		return nil
 	})
 }
 
--- a/channeldb/witness_cache.go
+++ b/channeldb/witness_cache.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/lntypes"
 )
 
 var (
@@ -77,6 +78,24 @@ type witnessEntry struct {
 	witness []byte
 }
 
+// AddSha256Witnesses adds a batch of new sha256 preimages into the witness
+// cache. This is an alias for AddWitnesses that uses Sha256HashWitness as the
+// preimages' witness type.
+func (w *WitnessCache) AddSha256Witnesses(preimages ...lntypes.Preimage) error {
+	// Optimistically compute the preimages' hashes before attempting to
+	// start the db transaction.
+	entries := make([]witnessEntry, 0, len(preimages))
+	for i := range preimages {
+		hash := preimages[i].Hash()
+		entries = append(entries, witnessEntry{
+			key:     hash[:],
+			witness: preimages[i][:],
+		})
+	}
+
+	return w.addWitnessEntries(Sha256HashWitness, entries)
+}
+
 // AddWitnesses adds a batch of new witnesses of wType to the witness cache. The
 // type of the witness will be used to map each witness to the key that will be
 // used to look it up. All witnesses should be of the same WitnessType.
@@ -100,6 +119,15 @@ func (w *WitnessCache) AddWitnesses(wType WitnessType, witnesses ...[]byte) erro
 			return ErrUnknownWitnessType
 		}
 	}
+
+	return w.addWitnessEntries(wType, entries)
+}
+
+// addWitnessEntries inserts the witnessEntry key-value pairs into the cache,
+// using the appropriate witness type to segment the namespace of possible
+// witness types.
+func (w *WitnessCache) addWitnessEntries(wType WitnessType,
+	entries []witnessEntry) error {
 
 	// Exit early if there are no witnesses to add.
 	if len(entries) == 0 {
--- a/channeldb/witness_cache.go
+++ b/channeldb/witness_cache.go
@@ -1,7 +1,6 @@
 package channeldb
 
 import (
-	"crypto/sha256"
 	"fmt"
 
 	"github.com/coreos/bbolt"
@@ -96,33 +95,6 @@ func (w *WitnessCache) AddSha256Witnesses(preimages ...lntypes.Preimage) error {
 	return w.addWitnessEntries(Sha256HashWitness, entries)
 }
 
-// AddWitnesses adds a batch of new witnesses of wType to the witness cache. The
-// type of the witness will be used to map each witness to the key that will be
-// used to look it up. All witnesses should be of the same WitnessType.
-//
-// TODO(roasbeef): fake closure to map instead a constructor?
-func (w *WitnessCache) AddWitnesses(wType WitnessType, witnesses ...[]byte) error {
-	// Optimistically compute the witness keys before attempting to start
-	// the db transaction.
-	entries := make([]witnessEntry, 0, len(witnesses))
-	for _, witness := range witnesses {
-		// Map each witness to its key by applying the appropriate
-		// transformation for the given witness type.
-		switch wType {
-		case Sha256HashWitness:
-			key := sha256.Sum256(witness)
-			entries = append(entries, witnessEntry{
-				key:     key[:],
-				witness: witness,
-			})
-		default:
-			return ErrUnknownWitnessType
-		}
-	}
-
-	return w.addWitnessEntries(wType, entries)
-}
-
 // addWitnessEntries inserts the witnessEntry key-value pairs into the cache,
 // using the appropriate witness type to segment the namespace of possible
 // witness types.
@@ -162,10 +134,21 @@ func (w *WitnessCache) addWitnessEntries(wType WitnessType,
 	})
 }
 
-// LookupWitness attempts to lookup a witness according to its type and also
+// LookupSha256Witness attempts to lookup the preimage for a sha256 hash. If
+// the witness isn't found, ErrNoWitnesses will be returned.
+func (w *WitnessCache) LookupSha256Witness(hash lntypes.Hash) (lntypes.Preimage, error) {
+	witness, err := w.lookupWitness(Sha256HashWitness, hash[:])
+	if err != nil {
+		return lntypes.Preimage{}, err
+	}
+
+	return lntypes.MakePreimage(witness)
+}
+
+// lookupWitness attempts to lookup a witness according to its type and also
 // its witness key. In the case that the witness isn't found, ErrNoWitnesses
 // will be returned.
-func (w *WitnessCache) LookupWitness(wType WitnessType, witnessKey []byte) ([]byte, error) {
+func (w *WitnessCache) lookupWitness(wType WitnessType, witnessKey []byte) ([]byte, error) {
 	var witness []byte
 	err := w.db.View(func(tx *bbolt.Tx) error {
 		witnessBucket := tx.Bucket(witnessBucketKey)
@@ -199,8 +182,13 @@ func (w *WitnessCache) LookupWitness(wType WitnessType, witnessKey []byte) ([]by
 	return witness, nil
 }
 
-// DeleteWitness attempts to delete a particular witness from the database.
-func (w *WitnessCache) DeleteWitness(wType WitnessType, witnessKey []byte) error {
+// DeleteSha256Witness attempts to delete a sha256 preimage identified by hash.
+func (w *WitnessCache) DeleteSha256Witness(hash lntypes.Hash) error {
+	return w.deleteWitness(Sha256HashWitness, hash[:])
+}
+
+// deleteWitness attempts to delete a particular witness from the database.
+func (w *WitnessCache) deleteWitness(wType WitnessType, witnessKey []byte) error {
 	return w.db.Batch(func(tx *bbolt.Tx) error {
 		witnessBucket, err := tx.CreateBucketIfNotExists(witnessBucketKey)
 		if err != nil {
--- a/channeldb/witness_cache_test.go
+++ b/channeldb/witness_cache_test.go
@@ -25,7 +25,7 @@ func TestWitnessCacheRetrieval(t *testing.T) {
 	witnessKey := sha256.Sum256(witness)
 
 	// First, we'll attempt to add the witness to the database.
-	err = wCache.AddWitness(Sha256HashWitness, witness)
+	err = wCache.AddWitnesses(Sha256HashWitness, witness)
 	if err != nil {
 		t.Fatalf("unable to add witness: %v", err)
 	}
@@ -59,13 +59,13 @@ func TestWitnessCacheDeletion(t *testing.T) {
 	// We'll start by adding two witnesses to the cache.
 	witness1 := rev[:]
 	witness1Key := sha256.Sum256(witness1)
-	if err := wCache.AddWitness(Sha256HashWitness, witness1); err != nil {
+	if err := wCache.AddWitnesses(Sha256HashWitness, witness1); err != nil {
 		t.Fatalf("unable to add witness: %v", err)
 	}
 
 	witness2 := key[:]
 	witness2Key := sha256.Sum256(witness2)
-	if err := wCache.AddWitness(Sha256HashWitness, witness2); err != nil {
+	if err := wCache.AddWitnesses(Sha256HashWitness, witness2); err != nil {
 		t.Fatalf("unable to add witness: %v", err)
 	}
 
@@ -107,7 +107,7 @@ func TestWitnessCacheUnknownWitness(t *testing.T) {
 
 	// We'll attempt to add a new, undefined witness type to the database.
 	// We should get an error.
-	err = wCache.AddWitness(234, key[:])
+	err = wCache.AddWitnesses(234, key[:])
 	if err != ErrUnknownWitnessType {
 		t.Fatalf("expected ErrUnknownWitnessType, got %v", err)
 	}
--- a/channeldb/witness_cache_test.go
+++ b/channeldb/witness_cache_test.go
@@ -1,9 +1,12 @@
 package channeldb
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"reflect"
 	"testing"
+
+	"github.com/lightningnetwork/lnd/lntypes"
 )
 
 // TestWitnessCacheRetrieval tests that we're able to add and lookup new
@@ -123,5 +126,87 @@ func TestWitnessCacheUnknownWitness(t *testing.T) {
 	err = wCache.AddWitnesses(234, key[:])
 	if err != ErrUnknownWitnessType {
 		t.Fatalf("expected ErrUnknownWitnessType, got %v", err)
+	}
+}
+
+// TestAddSha256Witnesses tests that insertion using AddSha256Witnesses behaves
+// identically to the insertion via the generalized interface.
+func TestAddSha256Witnesses(t *testing.T) {
+	cdb, cleanUp, err := makeTestDB()
+	if err != nil {
+		t.Fatalf("unable to make test database: %v", err)
+	}
+	defer cleanUp()
+
+	wCache := cdb.NewWitnessCache()
+
+	// We'll start by adding a witnesses to the cache using the generic
+	// AddWitnesses method.
+	witness1 := rev[:]
+	witness1Key := sha256.Sum256(witness1)
+
+	witness2 := key[:]
+	witness2Key := sha256.Sum256(witness2)
+
+	var (
+		preimages = []lntypes.Preimage{rev, key}
+		witnesses = [][]byte{witness1, witness2}
+		keys      = [][]byte{witness1Key[:], witness2Key[:]}
+	)
+
+	err = wCache.AddWitnesses(Sha256HashWitness, witnesses...)
+	if err != nil {
+		t.Fatalf("unable to add witness: %v", err)
+	}
+
+	for i, key := range keys {
+		witness := witnesses[i]
+
+		dbWitness, err := wCache.LookupWitness(
+			Sha256HashWitness, key,
+		)
+		if err != nil {
+			t.Fatalf("unable to lookup witness: %v", err)
+		}
+
+		// Assert that the retrieved witness matches the original.
+		if bytes.Compare(dbWitness, witness) != 0 {
+			t.Fatalf("retrieved witness mismatch, want: %x, "+
+				"got: %x", witness, dbWitness)
+		}
+
+		// We'll now delete the witness, as we'll be reinserting it
+		// using the specialized AddSha256Witnesses method.
+		err = wCache.DeleteWitness(Sha256HashWitness, key)
+		if err != nil {
+			t.Fatalf("unable to delete witness: %v", err)
+		}
+	}
+
+	// Now, add the same witnesses using the type-safe interface for
+	// lntypes.Preimages..
+	err = wCache.AddSha256Witnesses(preimages...)
+	if err != nil {
+		t.Fatalf("unable to add sha256 preimage: %v", err)
+	}
+
+	// Finally, iterate over the keys and assert that the returned witnesses
+	// match the original witnesses. This asserts that the specialized
+	// insertion method behaves identically to the generalized interface.
+	for i, key := range keys {
+		witness := witnesses[i]
+
+		dbWitness, err := wCache.LookupWitness(
+			Sha256HashWitness, key,
+		)
+		if err != nil {
+			t.Fatalf("unable to lookup witness: %v", err)
+		}
+
+		// Assert that the retrieved witness matches the original.
+		if bytes.Compare(dbWitness, witness) != 0 {
+			t.Fatalf("retrieved witness mismatch, want: %x, "+
+				"got: %x", witness, dbWitness)
+		}
 	}
 }
--- a/channeldb/witness_cache_test.go
+++ b/channeldb/witness_cache_test.go
@@ -19,27 +19,38 @@ func TestWitnessCacheRetrieval(t *testing.T) {
 
 	wCache := cdb.NewWitnessCache()
 
-	// We'll be attempting to add then lookup a d simple hash witness
+	// We'll be attempting to add then lookup two simple hash witnesses
 	// within this test.
-	witness := rev[:]
-	witnessKey := sha256.Sum256(witness)
+	witness1 := rev[:]
+	witness1Key := sha256.Sum256(witness1)
 
-	// First, we'll attempt to add the witness to the database.
-	err = wCache.AddWitnesses(Sha256HashWitness, witness)
+	witness2 := key[:]
+	witness2Key := sha256.Sum256(witness2)
+
+	witnesses := [][]byte{witness1, witness2}
+	keys := [][]byte{witness1Key[:], witness2Key[:]}
+
+	// First, we'll attempt to add the witnesses to the database.
+	err = wCache.AddWitnesses(Sha256HashWitness, witnesses...)
 	if err != nil {
 		t.Fatalf("unable to add witness: %v", err)
 	}
 
-	// With the witness stored, we'll now attempt to look it up. We should
-	// get back the *exact* same witness as we originally stored.
-	dbWitness, err := wCache.LookupWitness(Sha256HashWitness, witnessKey[:])
-	if err != nil {
-		t.Fatalf("unable to look up witness: %v", err)
-	}
+	// With the witnesses stored, we'll now attempt to look them up.
+	for i, key := range keys {
+		witness := witnesses[i]
 
-	if !reflect.DeepEqual(witness, dbWitness[:]) {
-		t.Fatalf("witnesses don't match: expected %x, got %x",
-			witness[:], dbWitness[:])
+		// We should get back the *exact* same witness as we originally
+		// stored.
+		dbWitness, err := wCache.LookupWitness(Sha256HashWitness, key)
+		if err != nil {
+			t.Fatalf("unable to look up witness: %v", err)
+		}
+
+		if !reflect.DeepEqual(witness, dbWitness[:]) {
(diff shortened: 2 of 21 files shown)


htlcswitch/link.go
```diff
@@ -447,6 +453,22 @@ func (l *channelLink) Stop() {
 
 	close(l.quit)
 	l.wg.Wait()
+
+	// As a final precaution, we will attempt to flush any uncommitted
+	// preimages to the preimage cache. The preimages should be re-delivered
+	// after channel reestablishment, however this adds an extra layer of
+	// protection in case the peer never returns. Without this, we will be
+	// unable to settle any contracts depending on the preimages even though
+	// we had learned them at some point.
```
joostjager: Just to make sure that I understand this correctly: this is not to miss the opportunity to apply the preimages to the downstream node. But if this was not in, it would not lead to loss of funds would it? We would learn the preimage when the channel is force closed and then settle downstream?

htlcswitch/link.go
```diff
@@ -1473,6 +1489,41 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
...
+		// Since we may have learned new preimages for the first time,
+		// we'll add them to our preimage cache. By doing this, we
+		// ensure any contested contracts watched by any on-chain
+		// arbitrators can now sweep this HTLC on-chain. We delay
+		// committing the preimages until just before accepting the new
+		// remote commitment, as afterwards the peer won't resend the
+		// Settle messages on the next channel reestablishment. Doing so
+		// allows us to more effectively batch this operation, instead
+		// of doing a single write per preimage.
+		if len(l.uncommitedPreimages) > 0 {
```
joostjager: Is it right that this performance improvement is only observable if multiple htlcs are fulfilled before a new commitsig is sent? So on a very high frequency link?

cfromknecht: That's correct, yes

htlcswitch/link.go
cfromknecht: assuming the downstream node were to pull on chain, we may learn the preimage then. alternatively, if the peer reconnects and sends the same settle message and the commit sig, that would also suffice.

atm we spawn and forget a goroutine that is meant handle the write of each preimage, which in most cases would probably execute before shutdown/disconnect. however with the addition batching, we can't rely on these background tasks to execute for us, and in fact may not have attempted the write at all. so we add this to ensure that we always attempt to write any received preimages before the link exits, which is akin to what we do now, though the current impl has weaker consistency guarantees

joostjager: Ok clear. Yes that goroutine was not perfect indeed.

channeldb/witness_cache_test.go
```diff
@@ -107,7 +107,7 @@ func TestWitnessCacheUnknownWitness(t *testing.T) {
 
 	// We'll attempt to add a new, undefined witness type to the database.
 	// We should get an error.
-	err = wCache.AddWitness(234, key[:])
+	err = wCache.AddWitnesses(234, key[:])
 	if err != ErrUnknownWitnessType {
 		t.Fatalf("expected ErrUnknownWitnessType, got %v", err)
 	}
```
joostjager: Do you want to test adding an actual batch of preimages?

witness_beacon.go
```diff
@@ -101,16 +101,21 @@ func (p *preimageBeacon) LookupPreimage(payHash []byte) ([]byte, bool) {
...
+	}
+
 	p.Lock()
 	defer p.Unlock()
 
-	srvrLog.Infof("Adding preimage=%x to witness cache", pre[:])
-
 	// First, we'll add the witness to the decaying witness cache.
-	err := p.wCache.AddWitness(channeldb.Sha256HashWitness, pre)
+	err := p.wCache.AddWitnesses(channeldb.Sha256HashWitness, preimages...)
```
joostjager: Can this be lifted outside of the lock, as it is happening in a db tx already?

contractcourt/channel_arbitrator.go
```diff
@@ -62,8 +62,13 @@ type WitnessBeacon interface {
 	// True is returned for the second argument if the preimage is found.
 	LookupPreimage(payhash []byte) ([]byte, bool)
 
-	// AddPreImage adds a newly discovered preimage to the global cache.
-	AddPreimage(pre []byte) error
+	// AddPreimages adds a batch of newly discovered preimages to the global
+	// cache, and also signals any subscribers of the newly discovered
+	// witness.
+	//
+	// NOTE: The backing slice of MUST NOT be modified, otherwise the
```
joostjager: It is a risk. You could use the new `lntypes.Preimage` type and go with copies.

htlcswitch/link.go
```diff
@@ -331,6 +331,12 @@ type channelLink struct {
 	// commitment fee every time it fires.
 	updateFeeTimer *time.Timer
 
+	// uncommitedPreimages stores a list of all preimages that have been
+	// learned since receiving the last CommitSig from the remote peer. The
+	// batch will be flushed just before accepting the subsequent CommitSig
+	// or on shutdown to avoid doing a write for each preimage received.
+	uncommitedPreimages [][]byte
```
joostjager: Isn't commited spelled as committed?

htlcswitch/link.go
joostjager: Ok. But the main reason for this PR is the unsafety of writing the preimage?

cfromknecht: The 80% penalty comes when writing each preimage individually, using our async_payments_benchmark integration test as a yard stick (which does attempt to saturate the links). For the single payment case, there would be a negligible penalty since the write is now done in the critical path

contractcourt/channel_arbitrator.go
cfromknecht: Thought about this, but we’d still be forced to perform a deep copy to ensure the notifications can’t be corrupted

cfromknecht: Nvm I see your point now, using `lntypes.Hash` makes a shallow copy sufficient. Good suggestion, also increases type safety. Will fix!

htlcswitch/link.go
joostjager: Ok, got it.

contractcourt/channel_arbitrator.go
joostjager: yes, but then `lntypes.Preimage` for the preimages

cfromknecht: haha yes, that's what i meant to type :)

channeldb/witness_cache.go
```diff
@@ -70,12 +70,41 @@ func (d *DB) NewWitnessCache() *WitnessCache {
...
 // TODO(roasbeef): fake closure to map instead a constructor?
-func (w *WitnessCache) AddWitness(wType WitnessType, witness []byte) error {
+func (w *WitnessCache) AddWitnesses(wType WitnessType, witnesses ...[]byte) error {
+	// Define a key-value struct that holds each key -> witness pair.
+	type witnessEntry struct {
+		key     []byte
+		witness []byte
+	}
+
+	// Optimistically compute the witness keys before attempting to start
```
halseth: Awesome change!

channeldb/witness_cache_test.go
```diff
@@ -59,13 +59,13 @@ func TestWitnessCacheDeletion(t *testing.T) {
 	// We'll start by adding two witnesses to the cache.
 	witness1 := rev[:]
 	witness1Key := sha256.Sum256(witness1)
-	if err := wCache.AddWitness(Sha256HashWitness, witness1); err != nil {
+	if err := wCache.AddWitnesses(Sha256HashWitness, witness1); err != nil {
```
halseth: Nit: long line

contractcourt/channel_arbitrator.go
halseth: 👍

contractcourt/htlc_outgoing_contest_resolver.go
```diff
@@ -78,7 +78,7 @@ func (h *htlcOutgoingContestResolver) Resolve() (ContractResolver, error) {
 
 		// With the preimage obtained, we can now add it to the global
 		// cache.
-		if err := h.PreimageDB.AddPreimage(preimage[:]); err != nil {
+		if err := h.PreimageDB.AddPreimages(preimage[:]); err != nil {
```
halseth: Long line

witness_beacon.go
halseth: Since we never remove from the cache, I believe that's true

cfromknecht: great suggestion!

channeldb/witness_cache_test.go
cfromknecht: exactly 80

contractcourt/htlc_outgoing_contest_resolver.go
cfromknecht: i think this is 75 chars

channeldb/witness_cache_test.go
cfromknecht: indeed! i've also added a test in `link_test.go` to assert the link's behavior is correct

contractcourt/channel_arbitrator.go
cfromknecht: I generally dig the move to use a `Preimage` type for type safety. Unfortunately, it turns out that the `channeldb.WitnessCache` is written in a general sense, meaning that it only operates on byte slices. So even though we get type safety going into the `preimageBeacon`, we then have to make a second copy of all of the preimages before passing them into the `WitnessCache`, so it ends up being pretty ugly.

One thing I've considered is making a wrapper/specialized interface, maybe `AddSha256Preimages` which uses the proper types, then these are serialized into the witness cache directly and hopefully avoiding the redundant copying. Lmk if I should investigate this further

cfromknecht: ended up going with the a helper that does only sha256 preimages, makes the code much cleaner

cfromknecht:
@joostjager @halseth ready for another round of review. the changes sprawled quite a bit as you'll see. I've also moved the lntypes 

I plan to consolidate 696a5f9 (the move to lntypes) and d47aa70 (first commit), but wanted to get your feedback on the additional commits before doing so.

Along the way:
 - `lnmock` is born, avoiding the same mockPreimageCache implementation in 4 places in the code base.
 - `lntypes` is moved to `lnhash` 
 - 'channeldb.WitnessCache` gained the `AddSHA256Witnesses` helper, to avoid the extra copied in the `preimageBeacon`
 - modfiied the `channeldb/witness_cache_test` to test batch insertion
 - added a link-level test to assert that the changes in `htlcswitch/link.go` behave as expected.

ptal

htlcswitch/link_test.go
```diff
@@ -4500,6 +4634,170 @@ func TestChannelLinkWaitForRevocation(t *testing.T) {
...
+	// committed. On the second, we'll disconnect before allowing Alice to
+	// process the CommitSig, and ensure that the link writes the preimage
+	// batch before exiting.
+	var (
+		start      = 0
+		end        = batchSize
+		disconnect = false
+	)
+
+checkBatchConsistency:
```
joostjager: I think it would be more understandable to use one or two functions instead of the `goto`. Or alternatively make it a parameterized test (with and without disconnect) that is ran from a parent test using `t.Run`.

lntypes/preimage.go
```diff
@@ -9,28 +9,40 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &preimage, nil
+}
+
+// MakePreimage returns a new Preimage from a bytes slice. An error is returned
+// if the number of bytes passed in is not PreimageSize.
+func MakePreimage(newPreimage []byte) (Preimage, error) {
```
joostjager: Yes I also ran into that ref return from `NewPreimage`. In the hodl invoice branch, I chose to just change the return type to a value. I couldn't see a good use case for the ptr return.

lnwallet/channel.go
```diff
@@ -5394,7 +5395,7 @@ func newOutgoingHtlcResolution(signer input.Signer, localChanCfg *channeldb.Chan
 func newIncomingHtlcResolution(signer input.Signer, localChanCfg *channeldb.ChannelConfig,
 	commitHash chainhash.Hash, htlc *channeldb.HTLC, keyRing *CommitmentKeyRing,
 	feePerKw SatPerKWeight, dustLimit btcutil.Amount, csvDelay uint32,
-	localCommit bool, preimage [32]byte) (*IncomingHtlcResolution, error) {
+	localCommit bool) (*IncomingHtlcResolution, error) {
```
joostjager: Separate commit as it is more than changing types?

lnmock/preimage_cache.go
```diff
@@ -0,0 +1,45 @@
+package lnmock
```
joostjager: I am not sure about a mock package. Whether it will again introduce circular dep issues. In general I'd think it is best to define the mock in the same package as the real implementation, so that there can't be new circular dependencies.

channeldb/invoices.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	"github.com/btcsuite/btcd/wire"
 	"github.com/coreos/bbolt"
-	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnhash"
```
joostjager: `lnhash` is what I had initially, but after comments I changed it to `lntypes`.

channeldb/witness_cache.go
```diff
@@ -63,25 +64,44 @@ type WitnessCache struct {
...
 func (d *DB) NewWitnessCache() *WitnessCache {
 	return &WitnessCache{
 		db: d,
 	}
 }
 
+// AddSHA256Witnesses adds a batch of new SHA256 preimages into the witness
+// cache. This is an alias for AddWitnesses that uses Sha256HashWitness as the
+// preimages' witness type.
+func (w *WitnessCache) AddSHA256Witnesses(preimages ...lnhash.Preimage) error {
```
joostjager: I could only find a single witness type. Why don't we remove it alltogether? Looks like something was added prematurely that we may never need?

witness_beacon.go
```diff
@@ -111,22 +111,15 @@ func (p *preimageBeacon) LookupPreimage(
...
 func (p *preimageBeacon) AddPreimages(preimages ...lnhash.Preimage) error {
 	// Copy the preimages to ensure the backing area can't be modified by
-	// the caller when delivering notifications. We also copy the keys so
-	// that the they can be passed in as a slice of byte slices to the
-	// generic witness cache.
-	preimageCopies := make([]lnhash.Preimage, len(preimages))
-	preimageKeys := make([][]byte, len(preimages))
-	for i, preimage := range preimages {
+	// the caller when delivering notifications.
+	preimageCopies := make([]lnhash.Preimage, 0, len(preimages))
```
joostjager: Would reorder commits to avoid adding and removing code again in the same PR

channeldb/witness_cache.go
cfromknecht: remove the older interface you mean? we do need to keep the constants to access the witnesses on disk

joostjager: Yes, remove the old interface.

lnmock/preimage_cache.go
cfromknecht: all of the circular dependency issues wrt to mocks are solved by writing unit tests in `package name_test`. there can't be circular dependencies because `name_test` can't be imported by anything else, and the mocks aren't to be used in the real implementation.

if we add them in the same package but make them private, they'll be inaccessible from the tests that do so. if we make them public to access from tests, then we have to document them to satisfy the linter but they also show up on godocs.

lnwallet/channel.go
cfromknecht: hmm must have been a rebase issue, will address when i restructure the commits

lntypes/preimage.go
cfromknecht: sounds good, i'm okay with either approach. indeed i found i didn't need the pointer object in this PR, but i could imagine cases where it would be useful

htlcswitch/link_test.go
cfromknecht: good suggestion, i'm actually realizing now that the persistent link harness is unnecessary for this test, so splitting it into a table driven test would be easier now

lnmock/preimage_cache.go
joostjager: Ok. Yes, in another project I actually do the same. Created a separate package containing test code to be able to reuse it without exposing it all from the real implementation.

lntypes/preimage.go
joostjager: I think just one constructor is enough, but it is all not very important.

channeldb/witness_cache.go
cfromknecht: Just remembered that we will have other witness types when we move to discrete log  challenges, though seems like that can also be exposed only via the targeted interfaces instead of the general one. Also allows us to make `WitnessType` private

channeldb/invoices.go
cfromknecht: removed import move

channeldb/witness_cache.go
cfromknecht: decided to leave the existing interface for now, though i'm not opposed re: removing it in the future. main reason is to keep the test asserting the two are identical, as i did discover a bug in the initial implementation as a result of writing the test

cfromknecht:
@joostjager @halseth comments addressed and commits restructured, ptal

contractcourt/htlc_success_resolver.go
```diff
@@ -103,6 +103,18 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
...
 				h.payHash[:])
 
+			// The wallet returns resolutions with the zero
+			// preimage, as it doesn't have access to the preimage
+			// cache. Before constructing the input, we lookup the
+			// cache and apply it to our resolution.
+			preimage, ok := h.PreimageDB.LookupPreimage(h.payHash)
+			if !ok {
+				err := fmt.Errorf("unable to find preimage "+
+					"for hash=%x", h.payHash)
```
joostjager: These formatting string can now also be `%v`. End result is the same though.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -240,6 +215,15 @@ func (h *htlcIncomingContestResolver) AttachResolverKit(r ResolverKit) {
 	h.ResolverKit = r
 }
 
+// applyPreimage sets the preimage for the internal htlc resolution and logs the
+// event.
+func (h *htlcIncomingContestResolver) applyPreimage(preimage lntypes.Preimage) {
```
joostjager: Instead of this function, can the inner `htlcSuccessResolver.applyPreimage` be called?

lnwallet/channel.go
```diff
@@ -5490,7 +5480,10 @@ func newIncomingHtlcResolution(signer input.Signer, localChanCfg *channeldb.Chan
 	}
 
 	// Next, we'll construct the full witness needed to satisfy the input
-	// of the success transaction.
+	// of the success transaction. We'll populate the witness stack with a
+	// blank preimage, which must be overwritten using ApplyPreimage with
+	// the discovered preimage before the witness can be valid.
+	var preimage [32]byte
 	successWitness, err := input.ReceiverHtlcSpendRedeem(
 		htlc.Signature, preimage[:], signer, &successSignDesc, successTx,
```
joostjager: Maybe not for this PR, but could eventually this whole witness generation be moved to the resolvers? Not to create a template here and fill in the preimage later?

channeldb/witness_cache.go
```diff
@@ -70,12 +71,69 @@ func (d *DB) NewWitnessCache() *WitnessCache {
...
+	return w.addWitnessEntries(Sha256HashWitness, entries)
+}
+
+// AddWitnesses adds a batch of new witnesses of wType to the witness cache. The
+// type of the witness will be used to map each witness to the key that will be
+// used to look it up. All witnesses should be of the same WitnessType.
 //
 // TODO(roasbeef): fake closure to map instead a constructor?
-func (w *WitnessCache) AddWitness(wType WitnessType, witness []byte) error {
+func (w *WitnessCache) AddWitnesses(wType WitnessType, witnesses ...[]byte) error {
```
joostjager: >decided to leave the existing interface for now, though i'm not opposed re: removing it in the future. main reason is to keep the test asserting the two are identical, as i did discover a bug in the initial implementation as a result of writing the test

If this is dead code now and just needed for testing, I'd suggest moving it to the test.

contractcourt/htlc_incoming_contest_resolver.go
cfromknecht: i wanted to keep the log message that properly displays the resolver type via `%T`, but yes it is possible

channeldb/witness_cache.go
cfromknecht: good idea, will fix

lnwallet/channel.go
cfromknecht: indeed, i'm not sure how best to accomplish that atm, but i added a todo

cfromknecht: added todo

lnwallet/channel.go
```diff
@@ -5194,6 +5194,23 @@ type IncomingHtlcResolution struct {
...
+
+	// If this our commitment transaction, then we'll need to populate the
+	// witness for the second-level HTLC transaction.
+	if r.SignedSuccessTx != nil {
+		// Within the witness for the success transaction, the preimage
+		// is the 4th element as it looks like:
+		//
+		//  * <sender sig> <recvr sig> <preimage> <witness script>
+		r.SignedSuccessTx.TxIn[0].Witness[3] = preimage[:]
+	}
```
cfromknecht: is there anything we need to set in this case? @halseth @joostjager @Roasbeef

cfromknecht:
@joostjager revamped the witness cache to only expose the specific interface, makes everything much cleaner imo, ptal

cfromknecht:
note: planning to squash `channeldb+witness_beacon: use sha256 lookup+delete witness` down into `channeldb/witness_cache: create AddSHA256Witnesses helper + test`

contractcourt/htlc_outgoing_contest_resolver.go
```diff
@@ -58,38 +60,47 @@ func (h *htlcOutgoingContestResolver) Resolve() (ContractResolver, error) {
...
-			copy(preimage[:], spendingInput.Witness[1])
+			preimageBytes = spendingInput.Witness[1]
+		}
+
+		preimage, err := lntypes.MakePreimage(preimageBytes)
+		if err != nil {
+			return nil, err
 		}
 
 		log.Infof("%T(%v): extracting preimage=%x from on-chain "+
```
joostjager: preimage can have %v now that it uses the strong type.

contractcourt/htlc_success_resolver.go
```diff
@@ -191,6 +205,17 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
...
 
+	// The wallet returns resolutions with the zero preimage, as it doesn't
+	// have access to the preimage cache. Before constructing the input, we
+	// lookup the cache and apply it to our resolution.
+	preimage, ok := h.PreimageDB.LookupPreimage(h.payHash)
+	if !ok {
+		err := fmt.Errorf("unable to find preimage "+
+			"for hash=%v", h.payHash)
+		return nil, err
+	}
```
joostjager: You could move the lookup code into `applyPreimage` too and rename the method. Prevent duplication from above. But not critical.

htlcswitch/link_test.go
```diff
@@ -4500,6 +4628,135 @@ func TestChannelLinkWaitForRevocation(t *testing.T) {
...
+			name:       "flush on disconnect",
+			disconnect: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			testChannelLinkBatchPreimageWrite(t, test.disconnect)
+		})
+	}
```
joostjager: If there are only two cases (or even a few more), I tend to unroll the loop and just make that calls to `t.Run` one by one. Make it a little lighter. But it is a personal pref.

```
t.Run("a", func(t *testing.T) {test(t, false)}
t.Run("b", func(t *testing.T) {test(t, true)}
```

lntypes/preimage.go
halseth: Would prefer sticking to one constructor to avoid confusion :)

contractcourt/channel_arbitrator.go
```diff
@@ -1209,13 +1207,12 @@ func (c *ChannelArbitrator) checkChainActions(height uint32,
...
-		payHash := htlc.RHash
-
 		// If we have the pre-image, then we should go on-chain to
 		// redeem the HTLC immediately.
-		if _, ok := c.cfg.PreimageDB.LookupPreimage(payHash[:]); ok {
+		if _, ok := c.cfg.PreimageDB.LookupPreimage(htlc.RHash); ok {
 			log.Tracef("ChannelArbitrator(%v): preimage for "+
-				"htlc=%x is known!", c.cfg.ChanPoint, payHash[:])
+				"htlc=%x is known!", c.cfg.ChanPoint,
+				htlc.RHash[:])
```
halseth: I think the `Hash` type has a `String` method, so you can use "%v" to print

contractcourt/htlc_outgoing_contest_resolver.go
```diff
@@ -2,9 +2,11 @@ package contractcourt
 
 import (
 	"fmt"
-	"github.com/lightningnetwork/lnd/input"
 	"io"
 
+	"github.com/lightningnetwork/lnd/input"
```
halseth: nit: imports should be moved by other `lnd` imports

contractcourt/htlc_success_resolver.go
```diff
@@ -3,9 +3,11 @@ package contractcourt
 import (
 	"encoding/binary"
 	"fmt"
-	"github.com/lightningnetwork/lnd/input"
 	"io"
 
+	"github.com/lightningnetwork/lnd/input"
```
halseth: same wrt imports

witness_beacon.go
```diff
@@ -66,53 +66,60 @@ func (p *preimageBeacon) SubscribeUpdates() *contractcourt.WitnessSubscription {
...
 	if err != nil {
-		ltndLog.Errorf("unable to lookup witness: %v", err)
-		return nil, false
+		ltndLog.Errorf("Unable to lookup witness: %v", err)
+		return lntypes.Preimage{}, false
+	}
+
+	preimage, err := lntypes.MakePreimage(preimageBytes)
+	if err != nil {
+		ltndLog.Errorf("Unable to build preiamge: %v", err)
```
halseth: nit:  typo "preiamge"

htlcswitch/link.go
```diff
@@ -1475,6 +1491,40 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
...
+		// we'll add them to our preimage cache. By doing this, we
+		// ensure any contested contracts watched by any on-chain
+		// arbitrators can now sweep this HTLC on-chain. We delay
+		// committing the preimages until just before accepting the new
+		// remote commitment, as afterwards the peer won't resend the
+		// Settle messages on the next channel reestablishment. Doing so
+		// allows us to more effectively batch this operation, instead
+		// of doing a single write per preimage.
+		if len(l.uncommittedPreimages) > 0 {
+			err := l.cfg.PreimageCache.AddPreimages(
```
halseth: Does the `AddPreimages` method handle zero-length slices? If it did, we can avoid some indentation here...

lnmock/preimage_cache.go
halseth: I'm not sure about this either. Maybe the `preimage_cache` should have its own package, which can contain both the real, and the mock implementation?

Either way, I think this discussion can be made its own PR, as the actual consistency fix is what is important here.

contractcourt/htlc_success_resolver.go
```diff
@@ -103,6 +103,18 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
 				"incoming+remote htlc confirmed", h,
 				h.payHash[:])
 
+			// The wallet returns resolutions with the zero
+			// preimage, as it doesn't have access to the preimage
+			// cache. Before constructing the input, we lookup the
+			// cache and apply it to our resolution.
+			preimage, ok := h.PreimageDB.LookupPreimage(h.payHash)
```
halseth: I'm confused... is this another, unrelated fix to the consistency write fix?

lnwallet/channel.go
```diff
@@ -5591,19 +5584,12 @@ func extractHtlcResolutions(feePerKw SatPerKWeight, ourCommit bool,
...
-			// populate this resolution.
-			preimage, _ := pCache.LookupPreimage(htlc.RHash)
-
-			// Otherwise, we'll create an incoming HTLC resolution
-			// as we can satisfy the contract.
-			var pre [32]byte
-			copy(pre[:], preimage[:])
+			// We'll create an incoming HTLC resolution which allows
+			// us to satisfy the contract assuming the preimage is
+			// discovered before the timelock expires.
```
halseth: Am I wrong in thinking that the write consistency fix is unrelated to this change?

I'm a bit uneasy doing critical changes like this in a big PR. I would prefer having this in a separate PR where the proper discussion can be had.

channeldb/witness_cache.go
```diff
@@ -78,10 +77,10 @@ type witnessEntry struct {
 	witness []byte
 }
 
-// AddSHA256Witnesses adds a batch of new SHA256 preimages into the witness
+// AddSha256Witnesses adds a batch of new sha256 preimages into the witness
 // cache. This is an alias for AddWitnesses that uses Sha256HashWitness as the
 // preimages' witness type.
-func (w *WitnessCache) AddSHA256Witnesses(preimages ...lntypes.Preimage) error {
+func (w *WitnessCache) AddSha256Witnesses(preimages ...lntypes.Preimage) error {
```
halseth: 👍 I was a bit confused by the introduction of the two methods when reading this commit by commit, so would be nice to squash this with the first change :)

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -218,8 +218,8 @@ func (h *htlcIncomingContestResolver) AttachResolverKit(r ResolverKit) {
 // applyPreimage sets the preimage for the internal htlc resolution and logs the
 // event.
 func (h *htlcIncomingContestResolver) applyPreimage(preimage lntypes.Preimage) {
-	log.Infof("%T(%v): extracted preimage=%x from beacon!", h,
-		h.htlcResolution.ClaimOutpoint, preimage[:])
+	log.Infof("%T(%v): extracted preimage=%v from beacon!", h,
```
halseth: 👍

channeldb/witness_cache.go
halseth: Ok, I see now that this is planned 👍

lntypes/preimage.go
joostjager: In any case, please keep the non-pointer named `Make`. In another PR I just went through a rebase action to do that. I understand that there is a convention where `New` is for pointers and `Make` for values.

htlcswitch/link.go
cfromknecht: good suggestion, added a check in `AddPreimages` to exit early for zero length slices

lntypes/preimage.go
cfromknecht: My personal preference is typically to add both constructors, which allows the caller to use whatever is best as requiring an extra indirect or dereference can have performance impacts or make calling code more verbose. It seems like we don't really use the pointer version tho atm, so i just decided to leave `MakePreimage` that returns the non-pointer version. We can always add `NewPreimage` later if we want/need to

contractcourt/channel_arbitrator.go
cfromknecht: `htlc.RHash` is still a `[32]byte`, still need `%x`

cfromknecht:
@halseth @joostjager i took out the commits that removed the preimage cache from lnwallet, should be a little more contained now. it still includes the commits that make the switch to `lntypes` as requested, ptal

channeldb/witness_cache.go
```diff
@@ -77,6 +78,24 @@ type witnessEntry struct {
...
 
+// AddSha256Witnesses adds a batch of new sha256 preimages into the witness
+// cache. This is an alias for AddWitnesses that uses Sha256HashWitness as the
+// preimages' witness type.
+func (w *WitnessCache) AddSha256Witnesses(preimages ...lntypes.Preimage) error {
+	// Optimistically compute the preimages' hashes before attempting to
+	// start the db transaction.
+	entries := make([]witnessEntry, 0, len(preimages))
+	for i := range preimages {
+		hash := preimages[i].Hash()
```
halseth: any reason not to use `for _, p := range preimages {`?

contractcourt/channel_arbitrator.go
halseth: Ah, but it can still be used when calling `LookupPreimage(payhash lntypes.Hash)`? Did not know that :)

channeldb/witness_cache.go
cfromknecht: Indeed because each `p[:]` will point to the same preimage due to range var binding. This was the bug I had initially ;)

contractcourt/channel_arbitrator.go
cfromknecht: yep, since `lntypes.Preimage` is a named type and `[32]byte` is an unnamed type, they can be casted implicitly bc the underlying raw type is the same

contractcourt/htlc_success_resolver.go
cfromknecht: good suggestion, will carry this over to a subsequent PR that removes the preimage cache from lnwallet

htlcswitch/link_test.go
cfromknecht: ended up leaving as is so that the the booleans are named :)

PR #2503, opened 2019-01-18
watchtower/wtserver/server: return valid reward script

cfromknecht:
Small PR that ensures the wtserver will return the proper reward script for the client to include in it's transaction. Prior, we incorrectly used the address's `ScriptAddress`. This went unnoticed until doing the full e2e integration tests when no funds showed up in the tower's wallet.

### DIFF
--- a/watchtower/wtserver/server.go
+++ b/watchtower/wtserver/server.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/connmgr"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcutil"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/watchtower/blob"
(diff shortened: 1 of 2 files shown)


watchtower/wtserver/server.go
```diff
@@ -364,7 +365,15 @@ func (s *Server) handleCreateSession(peer Peer, id *wtdb.SessionID,
 		)
 	}
 
-	rewardAddrBytes := rewardAddress.ScriptAddress()
+	// Construct the pkscript the client should pay to when signing justice
+	// transactions for this session.
+	rewardScript, err := txscript.PayToAddrScript(rewardAddress)
```
Roasbeef: 👍

watchtower/wtserver/server.go
```diff
@@ -380,14 +389,14 @@ func (s *Server) handleCreateSession(peer Peer, id *wtdb.SessionID,
...
-		ID:            *id,
-		RewardAddress: rewardAddrBytes,
+		ID: *id,
 		Policy: wtpolicy.Policy{
 			BlobType:     req.BlobType,
 			MaxUpdates:   req.MaxUpdates,
 			RewardRate:   req.RewardRate,
 			SweepFeeRate: req.SweepFeeRate,
 		},
+		RewardAddress: rewardScript,
```
Roasbeef: To make things more consistent we should likely rename these fields to be `RewardPkScript`, just to be more consistent and eliminate any ambiguity. Not a blocker, though can be done in a follow up PR as the rename would likely sprawl multiple files.

PR #2504, opened 2019-01-18
build: update release.sh to contain vendor archive + source code

### LINKED ISSUE #1550: release: provide source archive with vendor directory
### Background

Hi. I'm looking into creating a port of lnd for OpenBSD. The OpenBSD ports framework does not allow any HTTP requests to fetch the go dependencies. Can you provide a source archive with the vendor directory included on every release?

Thanks you,
Fabian

### Expected behavior

Every release includes a source archive with the vendor directory already included.

### Actual behaviour

Every release includes a source archive with without the vendor directory.

Roasbeef:
In this commit, we update the release script to include a directory
containing all of our vendored dependencies. This makes out releases
more self contained as they can be built using `go build -mod=vendor`
without requiring any package fetching.

Additionally, we pacakge the entire source as well.

Fixes #1550.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -16,9 +16,21 @@ else
     TAG=$1
 fi
 
+go mod vendor
+tar -cvzf vendor.tar.gz vendor
+
 PACKAGE=lnd
 MAINDIR=$PACKAGE-$TAG
 mkdir -p $MAINDIR
+
+cp vendor.tar.gz $MAINDIR/
+rm vendor.tar.gz
+rm -r vendor
+
+PACKAGESRC="$MAINDIR/$PACKAGE-source-$TAG.tar"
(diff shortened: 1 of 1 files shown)


Roasbeef:
Sample release directory after this patch:
```
-rw-r--r--  1 roasbeef  staff    20M Jan 18 19:15 lnd-darwin-amd64-20190118-01.tar.gz
-rw-r--r--  1 roasbeef  staff   1.5M Jan 18 19:15 lnd-source-20190118-01.tar.gz
-rw-r--r--  1 roasbeef  staff   278B Jan 18 19:15 manifest-20190118-01.txt
-rw-r--r--  1 roasbeef  staff   5.0M Jan 18 19:15 vendor.tar.gz
```

PR #2505, opened 2019-01-18
build: more reliable coverage using go-acc

cfromknecht:
Testing out a different method of computing coverage results that hopefully is more accurate. The current script used by `unit-cover` does not correctly account for cross-package coverage. I found that https://github.com/ory/go-acc properly accounts for areas of code base that I know to be tested indirectly. As a result, our coverage gets a nice 2.6% bump.

This will allow us to make refactorings like #2481 simpler, and not have to worry as much about our code's location relative to the testing files. I would be interested in seeing how the coverage changes on that PR while using this as the coverage tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,11 +4,13 @@ ESCPKG := github.com\/lightningnetwork\/lnd
 BTCD_PKG := github.com/btcsuite/btcd
 GOVERALLS_PKG := github.com/mattn/goveralls
 LINT_PKG := gopkg.in/alecthomas/gometalinter.v2
+GOACC_PKG := github.com/ory/go-acc
 
 GO_BIN := ${GOPATH}/bin
 BTCD_BIN := $(GO_BIN)/btcd
 GOVERALLS_BIN := $(GO_BIN)/goveralls
 LINT_BIN := $(GO_BIN)/gometalinter.v2
+GOACC_BIN := $(GO_BIN)/go-acc
 
 BTCD_DIR :=${GOPATH}/src/$(BTCD_PKG)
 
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -36,6 +35,8 @@ require (
 	github.com/lightningnetwork/lightning-onion v0.0.0-20180605012408-ac4d9da8f1d6
 	github.com/ltcsuite/ltcd v0.0.0-20190101042124-f37f8bf35796
 	github.com/miekg/dns v0.0.0-20171125082028-79bfde677fa8
+	github.com/ory/go-acc v0.0.0-20181118080137-ddc355013f90 // indirect
```
joostjager: Is this right, that it is a module requirement of `lnd` now?

cfromknecht: this was the way go.mod and go.sum were left as a result of the coverage run. i don't think it's strictly necessary to include them, and am happy to remove them. for instance, i don't think they'd show up when just doing a regular `make && make install`

Makefile
```diff
@@ -88,6 +73,10 @@ $(LINT_BIN):
 	@$(call print, "Fetching gometalinter.v2")
 	GO111MODULE=off go get -u $(LINT_PKG)
 
+$(GOACC_BIN):
+	@$(call print, "Fetching go-acc")
+	$(GOINSTALL) $(GOACC_PKG)
```
joostjager: Is this what users expect? That they get a tool installed automatically?

joostjager: Or should we now start with `make toolchain` to install this and the grpc stuff as well. And then remove the autogen. code from the repo.

cfromknecht: this is really only for travis builds, so i think it's okay in this case

cfromknecht:
Removed the changes to go.mod and go.sum, it's now purely a makefile change

Makefile
joostjager: ok

go.mod
Roasbeef: Yeah I don't think this is needed, it's only a result of the script being run in the same directory.

PR #1164, opened 2018-05-01
"Production" Docker Image

tyzbit:
This PR satisfies the auto-build functionality desired in #722 without changing anything in the `docker` folder.  This allows development and testing to continue to build and use images locally, but external users get a very lean Docker image they can use.  The size difference is appreciable.

```
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
lnd                            latest              3634dcf5df68        17 seconds ago      1.19GB
lnd-single                     latest              b295cc248447        30 minutes ago      56MB
```

Right before merging this PR, an Automated Build repository should be configured on the appropriate account ([lightningnetwork](https://hub.docker.com/r/lightningnetwork/)?) on Docker Hub with the name of `lnd` in order to match documentation.

I see there are already some repositories created, however *I think* best practice for Docker is to have a single repository and different tags for different architectures.  The included image is based off of Alpine, but in the future we could extend into multiple Dockerfiles (example: ARM support, or being Ubuntu based).

So in this proposed scenario, `lightningnetwork/lnd:latest` would be `alpine`, but `lightningnetwork/lnd:ubuntu` would be an ubuntu-based image, for example.

(With thanks to @alingenhag for https://github.com/f-u-z-z-l-e/docker-lnd/ from which I borrowed heavily)

PR #697, opened 2018-01-30
Adds data volume to lnd docker service

Osso:
This is my proposed changes for issue #696

PR #2507, opened 2019-01-18
add volume to lnd to persist chain state across container changes

yancyribbens:
This PR adds a volume to the lnd `Dockerfile` to persist chain state.  I considered adding this to `docker-compose.yml` instead however the volume is only needed for use by `lnd` and isn't shared with other services.  The functionality can be verified by restarting the `lnd` container and observe it no longer needs to catch up to a particular chain height.

### DIFF
--- a/docker/lnd/Dockerfile
+++ b/docker/lnd/Dockerfile
@@ -2,6 +2,9 @@ FROM golang:1.11-alpine as builder
 
 MAINTAINER Olaoluwa Osuntokun <lightning.engineering>
 
+# Persistant volume store for lnd data
+VOLUME /root/.lnd/data
+
 # Copy in the local repository to build from.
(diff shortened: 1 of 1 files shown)


yancyribbens:
https://github.com/lightningnetwork/lnd/issues/2508

Roasbeef:
See #697 

Also this (https://github.com/lightningnetwork/lnd/pull/1164) added volumes to the "production" container.

yancyribbens:
@Roasbeef thanks! I wasn't aware of these existing PR's :)

Roasbeef:
That one seems to be abandoned though, so we might as well close that out and move forward with this one.

yancyribbens:
Closing this PR in favor of [2533](https://github.com/lightningnetwork/lnd/pull/2533)

PR #2515, opened 2019-01-19
watchtower/wtpolicy: altruist output calculation

cfromknecht:
This PR adds the necessary calculations for deriving a justice transaction that gives all funds back to the client, and no reward for the tower. 

This facilitates watchtower modes where the client and tower have an existing relation, e.g. you are running your own tower, and it doesn't make sense to split fees off for the tower. Towers operating only in this mode could even be created without an on-chain wallet, which makes them even more lightweight than those that collect rewards. Since the initial rollout is tailored to private towers, this gives operators and users another useful configuration for backing up revoked states.

Depends on:
 - [x] https://github.com/lightningnetwork/lnd/pull/2514
 - [x] https://github.com/lightningnetwork/lnd/pull/2513

### DIFF
--- a/watchtower/lookout/justice_descriptor.go
+++ b/watchtower/lookout/justice_descriptor.go
@@ -8,6 +8,7 @@ import (
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/txsort"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/watchtower/blob"
 	"github.com/lightningnetwork/lnd/watchtower/wtdb"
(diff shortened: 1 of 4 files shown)


watchtower/wtpolicy/policy.go
```diff
@@ -71,3 +73,30 @@ func (p Policy) String() string {
...
+
+	// TODO(conner): replace w/ configurable dust limit
+	dustLimit := lnwallet.DefaultDustLimit()
+
+	// Check that the created outputs won't be dusty.
+	if sweepAmt <= dustLimit {
+		return 0, ErrCreatesDust
+	}
+
+	// TODO(conner): also compare output value to fee?
```
Roasbeef: Compare to check that we're not throwing away most of it to the fee?

watchtower/lookout/justice_descriptor.go
```diff
@@ -166,42 +167,72 @@ func (p *JusticeDescriptor) assembleJusticeTxn(txWeight int64,
...
+			totalAmt, txWeight,
+		)
+		if err != nil {
+			return nil, err
+		}
 
-	// TODO(conner): abort/don't add if outputs are dusty
+		// Add the sweep and reward outputs to the justice transaction.
+		justiceTxn.AddTxOut(&wire.TxOut{
+			PkScript: p.JusticeKit.SweepAddress[:],
```
Roasbeef: IMO we should eventually re-name the justice kit elements to `*Script` instead of `*Address`.

watchtower/lookout/justice_descriptor_test.go
```diff
@@ -170,20 +195,44 @@ func TestJusticeDescriptor(t *testing.T) {
...
 		},
-		TxOut: []*wire.TxOut{
-			{
+	}
 
-				Value:    int64(sweepAmt),
-				PkScript: justiceKit.SweepAddress,
-			},
-			{
+	if blobType.Has(blob.FlagReward) {
```
Roasbeef: More or less the same block of code used in `assembleJusticeTxn`, I think we can consolidate a bit.

watchtower/wtpolicy/policy.go
cfromknecht: yep i left the todo here, but i'm thinking it should be moved to a higher level as this threshold wouldn't be a policy shared by both client and server. it would be purely client-level configuration

watchtower/lookout/justice_descriptor.go
cfromknecht: Agreed, i will add it to the list of todos

cfromknecht:
@Roasbeef comments addressed, PTAL

Issue #1221, opened 2018-05-10
breacharbiter: implement an optional scorched earth breach remedy policy 


Roasbeef:
The `breachArbiter` is a special sub-system within the daemon that's tasked with responding to potential contract breaches, and delivering justice to any cheating channel peers. Atm, it's possible for our justice transaction to get delayed (potentially indefinitely) if the initial fee rate we issue on the breach transaction is too low, or the mempool is very full at the time of broadcast. Due to the fact that we are entitled to all funds that the breaching party has fortified in the channel, we can actually slowly use those funds for fees. This will guarantee that either the remote party needs to pay more in fees that they have in the channel in order to make a successful breach, or we burn all their funds to miner's fees. As a result, we can implement a "scorched earth" policy when handling breaches with the end result in the state machine leading to the remote party forfeiting all their funds to miner's fees. 

# Steps To Completion 

- [ ] The [`exctractRetribution`](https://github.com/lightningnetwork/lnd/blob/master/breacharbiter.go#L303) method should be extracted to a _new_ persistent state machine within the `breachArbiter`. The role of this state machine is to monitor the number of blocks that has passed since we broadcasted the justice transaction. A heuristic should be developed such that, once N blocks has passed w/o our justice transaction being mined, then we allocate an additional `K%` of the funds of the remote party to miner's fees. 
- [ ] A series of tests should be written using the integration testing framework (and manually mined empty blocks) to exercise the newly added logic.

PR #2516, opened 2019-01-19
watchtower: client justice transaction creation + blob encryption

cfromknecht:
This PR adds one of the core components of the watchtower client, which handles the responsibility of taking a revoked state and generating the encrypted blob that is sent to the tower. To do so, we add an internal struct `wtclient.backupTask, which house the bulk of the logic surrounding justice transaction creation, signing, and encrypting the necessary portions for a selected tower.

```
// backupTask is an internal struct for computing the justice transaction for a
// particular revoked state. A backupTask functions as a scratch pad for storing
// computing values of the transaction itself, such as the final split in
// balance if the justice transaction will give a reward to the tower. The
// backup task has three primary phases:
//  1. Init: Determines which inputs from the breach transaction will be spent,
//     and the total amount contained in the inputs.
//  2. Bind: Asserts that the revoked state is eligible under a given session's
//     parameters. Certain states may be ineligible due to fee rates, too little
//     input amount, etc.. Backup of these states can be deferred to a later
//     time or session with more favorable parameters. If the session is bound
//     successfully, the final session-dependent values to the justice
//     transaction are solidified.
//  3. Send: Once the task is bound, it will be queued to send to a specific
//     tower corresponding to the session in which it was bound. The justice
//     transaction will be assembled by examining the parameters left as a
//     result of the binding. After the justice transaction is signed, the
//     necessary components are stripped out and encrypted before being sent to
//     the tower in a StateUpdate.
```
To elaborate further and get a sense for how this will function in the context of the rest of the client, the first stage happens immediately after submitting a revoked state to the client via an extrnal API such as:
```golang
type Client interface {
        BackupState(lnwire.ChannelID, lnwallet.BreachRetribution) error
}
```
All incoming tasks are aggregated into a pipeline so that they can be sequentially assigned to an active session with a tower.

Once the task has been pulled from the pipeline, the client will try to bind the `backupTask` to the given session using `backupTask.bindSession`. This procedure locks in the final form of the justice transaction by determining fee rates, reward rates, etc. using the negotiated session's policy.

If the bind succeeds, the task is queued for that particular tower. The resulting justice transaction will be constructed, signed, and encrypted using `backupTask.craftSessionPayload`, which is then included in a `wtwire.StateUpdate` and sent to the tower.

NOTE: The logic surrounding this pipeline _and making it reliable_ will be included in subsequent PRs, though this constitutes the majority of logic necessary to craft and send valid blobs as a client.

Depends on:
 - [x] #2514 
 - [x] #2513 
 - [x] #2515


watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+// backupTask is an internal struct for computing the justice transaction for a
+// particular revoked state. A backupTask functions as a scratch pad for storing
+// computing values of the transaction itself, such as the final split in
+// balance if the justice transaction will give a reward to the tower. The
+// backup task has three primary phases:
+//  1. Init: Determines which inputs from the breach transaction will be spent,
+//     and the total amount contained in the inputs.
+//  2. Bind: Asserts that the revoked state is eligible under a given session's
+//     parameters. Certain states may be ineligible due to fee rates, too little
+//     input amount, etc.. Backup of these states can be deferred to a later
```
Roasbeef: Extra period here after "etc"?

watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+//     necessary components are stripped out and encrypted before being sent to
+//     the tower in a StateUpdate.
+type backupTask struct {
+	chanID       lnwire.ChannelID
+	commitHeight uint64
+	breachInfo   *lnwallet.BreachRetribution
+
+	// state-dependent variables
+
+	toLocalInput  input.Input
```
Roasbeef: > `input.Input`

kek

watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+	// since the task has not been assigned to a session, at which point
+	// parameters such as fee rate, number of outputs, and reward rate will
+	// be finalized.
+	var (
+		totalAmt      int64
+		toLocalInput  input.Input
+		toRemoteInput input.Input
+	)
+
+	// If the to-local output is non-dust, add it to the set of candidate
```
Roasbeef: Comments and variable name here doesn't match. Comment references `to-local`, but the `RemoteOutputSignDesc` is checked. Similar applies to below.

watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+	// Next, add the contribution from the inputs that are present on this
+	// breach transaction.
+	if t.toLocalInput != nil {
+		weightEstimate.AddWitnessInput(input.ToLocalPenaltyWitnessSize)
+	}
+	if t.toRemoteInput != nil {
+		weightEstimate.AddWitnessInput(input.P2WKHWitnessSize)
+	}
+
+	// Now, compute the output values depending on whether FlagReward is set
```
Roasbeef: Carried over from the other PR, I think this can use that utility function referenced.

watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+	if t.toLocalInput != nil {
+		inputs[*t.toLocalInput.OutPoint()] = t.toLocalInput
+	}
+	if t.toRemoteInput != nil {
+		inputs[*t.toRemoteInput.OutPoint()] = t.toRemoteInput
+	}
+	return inputs
+}
+
+// bindSession determines if the backupTask is compatible with the passed
```
Roasbeef: The comment references validation, but I don't see any in the actual method. Perhaps it should be an error if _both_ the local+remote outputs aren't found?

watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+	// Now, begin construction of the justice transaction. We'll start with
+	// a version 2 transaction.
+	justiceTxn := wire.NewMsgTx(2)
+
+	// Next, add the non-dust inputs that were derived from the breach
+	// information. This will either be contain both the to-local and
+	// to-remote outputs, or only be the to-local output.
+	inputs := t.inputs()
+	for prevOutPoint := range inputs {
+		justiceTxn.AddTxIn(&wire.TxIn{
```
Roasbeef: We should modify the sequence numbers so they aren't all final so we can later extend the tower (and also eventually `lnd`) to be able to broadcast multiple versions of the transaction with varying fee rates a la #1221

watchtower/wtclient/backup_task.go
```diff
@@ -0,0 +1,344 @@
...
+		signature, err := lnwire.NewSigFromRawSignature(rawSignature)
+		if err != nil {
+			return hint, nil, err
+		}
+
+		// Finally, copy the serialized signature into the justice kit,
+		// using the input's witness type to select the appropriate
+		// field.
+		switch inp.WitnessType() {
+		case input.CommitmentRevoke:
```
Roasbeef: Again, shouldn't these be flipped? We revoke their output, but then just sweep ours w/ no delay.

watchtower/wtclient/backup_task_internal_test.go
```diff
@@ -0,0 +1,605 @@
...
+			KeyDesc: keychain.KeyDescriptor{
+				KeyLocator: revKeyLoc,
+				PubKey:     revPK,
+			},
+			Output: &wire.TxOut{
+				Value: toLocalAmt,
+			},
+			HashType: txscript.SigHashAll,
+		}
+		breachInfo.RemoteOutputSignDesc = toLocalSignDesc
```
Roasbeef: remote+local sign desc look to be swapped here, same below

watchtower/wtclient/backup_task_internal_test.go
```diff
@@ -0,0 +1,605 @@
...
+	blobTypeCommitReward = (blob.FlagCommitOutputs | blob.FlagReward).Type()
+
+	addr, _ = btcutil.DecodeAddress(
+		"mrX9vMRYLfVy1BnZbc5gZjuyaqH3ZW2ZHz", &chaincfg.TestNet3Params,
+	)
+
+	addrScript, _ = txscript.PayToAddrScript(addr)
+)
+
+var backupTaskTests = []backupTaskTest{
```
Roasbeef: /notbadface

once you table driven test...

watchtower/wtclient/backup_task.go
cfromknecht: heh, i've always thought it was correct to add another period at the end of a sentence, apparently not so. it is okay to add a comma after tho. english yo 🤷‍♂️

cfromknecht:
since we're dealing with the remote party's commitment, our local balance is store in an output with a to-remote script. similarly, the remote balance is stored in an output with a to-local script. this comment explains it https://github.com/lightningnetwork/lnd/pull/2516/files#diff-1b1a0991ab6305ea87930b85d6cf2486R126

thought i had moved that description into the backup_task file as well, will make that change

watchtower/wtclient/backup_task.go
cfromknecht: This is handled when computing the amounts, it will fail since the totalAmt is 0 and this isn't enough to cover the txn fee

cfromknecht:
@Roasbeef comments addressed, PTAL

PR #2517, opened 2019-01-19
Proper cleanup neutrino database in case of error

roeierez:
This PR implements proper "clean up" code in case there is a failure in the creation of neutrino ChainService.
More specifically, in case of error, close the database that was opened and stop the service if already started.


roeierez:
Rebased

PR #2520, opened 2019-01-21
specify a tag, branch or commit during docker prod build

yancyribbens:
Add `build-arg` to docker for choosing which branch, tag or commit to run during build time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,11 @@ FROM golang:1.13-alpine as builder
 # queries required to connect to linked containers succeed.
 ENV GODEBUG netdns=cgo
 
+# Pass a tag, branch or a commit using build-arg.  This allows a docker
+# image to be built from a specified Git state.  The default image
+# will use the Git tip of master by default.
+ARG checkout="master"
+
 # Install dependencies and build the binaries.
 RUN apk add --no-cache --update alpine-sdk \
     git \
@@ -11,6 +16,7 @@ RUN apk add --no-cache --update alpine-sdk \
     gcc \
 &&  git clone https://github.com/lightningnetwork/lnd /go/src/github.com/lightningnetwork/lnd \
 &&  cd /go/src/github.com/lightningnetwork/lnd \
+&&  git checkout $checkout \
 &&  make \
 &&  make install tags="signrpc walletrpc chainrpc invoicesrpc routerrpc"
 
--- a/docs/DOCKER.md
+++ b/docs/DOCKER.md
@@ -65,3 +65,30 @@ $ docker logs lnd-testnet
 
 This is a simple example, it is possible to use any command-line options necessary
 to expose RPC ports, use `btcd` or `bitcoind`, or add additional chains.
+
+## LND Development and Testing
+
+To test the Docker production image locally, run the following from
+the project root:
+
+```
+$ docker build . -t lnd:master
+```
+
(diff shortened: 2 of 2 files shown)


yancyribbens:
Replaced by: https://github.com/lightningnetwork/lnd/pull/2676

docs/DOCKER.md
```diff
@@ -65,3 +65,30 @@ $ docker logs lnd-testnet
...
 This is a simple example, it is possible to use any command-line options necessary
 to expose RPC ports, use `btcd` or `bitcoind`, or add additional chains.
+
+## LND Development and Testing
+
+To test the Docker production image locally, run the following from
+the project root:
+
+```
+$ docker build . -t lnd:1.0
```
guggero: What's the reason for the version `1.0`? Wouldn't it make more sense to call it something like `-t lnd:master`?

docs/DOCKER.md
```diff
@@ -65,3 +65,30 @@ $ docker logs lnd-testnet
...
+```
+
+To choose a specific branch or tag instead, use the "checkout" build-arg.  For example, to build the latest commits in master:
+
+```
+$ docker build . --build-arg checkout=master -t lnd:1.0
```
guggero: This will result in the exact same image as the example above. Why not give an example of a production tag (e.g. `v0.8.0-beta`) and then also name the image `lnd:0.8.0-beta`?

docs/DOCKER.md
```diff
@@ -65,3 +65,30 @@ $ docker logs lnd-testnet
...
+```
+
+To build the image using the most current tag:
+
+```
+$ docker build . --build-arg checkout=$(git describe --tags `git rev-list --tags --max-count=1`) -t lnd:1.0
```
guggero: Same as above, should use a different example name than `lnd:1.0`. Maybe `lnd:latest-tag` or something.

docs/DOCKER.md
yancyribbens: typically tag is a version number, although I like the idea of using `master` since `master` is the default.  Happy to make that change.

docs/DOCKER.md
yancyribbens: adding tag as an example is a good idea.  I don't have a problem adding that.

docs/DOCKER.md
yancyribbens: Cool.  happy to make that update.

Dockerfile
```diff
@@ -4,13 +4,17 @@ FROM golang:alpine as builder
 # queries required to connect to linked containers succeed.
 ENV GODEBUG netdns=cgo
 
+# pass tag, branch or commit during build
```
guggero: nit: Please write a full sentence with capitalization and full stop at the end.

yancyribbens: :+1:

yancyribbens:
@carlaKC I see lots of example commits with `docker`.  maybe that would be preferred over `dockerfile`?

guggero:
Yes, `docker: ` is fine too, thanks!

carlaKC:
> @carlaKC I see lots of example commits with `docker`. maybe that would be preferred over `dockerfile`?

Sorry I missed this, too many git emails 🙈 Thanks for the PR! 
`docker` is good as Oli says.

PR #2521, opened 2019-01-21
lnrpc: deprecate SendToRoute with more than one route

joostjager:
The current SendToRoute RPC call offers the option of passing in multiple routes. Especially in the streaming variant, this makes the structure of the response complicated, because it is double nested (multiple streamed requests each containing multiple routes for each of which a result needs to be reported). The advantage of this in terms of network round trips is negligible, so the proposal is to limit every SendToRoute request to a single route.


rpcserver.go
```diff
@@ -2545,6 +2527,43 @@ func (r *rpcServer) SendToRoute(stream lnrpc.Lightning_SendToRouteServer) error
...
+
+	switch {
+	case len(req.Routes) == 0 && req.Route == nil:
+		return nil, fmt.Errorf("unable to send, no routes provided")
+	case len(req.Routes) > 0 && req.Route != nil:
+		return nil, fmt.Errorf("cannot use both route and routes field")
+	}
+
+	var routes []*routing.Route
+	if len(req.Routes) > 0 {
```
Roasbeef: Why even distinguish these two cases? If the case of the `else` statement, it completes after a single loop iteration.

joostjager: I chose not to restrict the `Routes` field to a single element, but to add a new non-list field `Route`. Therefore the else construct here.

The intention behind this is to make one change to the interface and not two (first restrict to a single element and then later make it a non-list field).

joostjager:
ptal @Roasbeef

PR #1783, opened 2018-08-24
server: reorder start/stop order of subsystems

cfromknecht:
This commit restructures the start and stop order
of lnd's primary subsystems. The new ordering is
intended to be more aware of the dependencies
between various subsystems, and promote better
safety. The main things that are handled:

 - start connmgr last to prevent connections before init
 - shutdown connmgr immediately to prevent conns during shutdown
 - start all onchain subsystems before offchain ones
 - stop all offchain subsystems before all onchain ones
 - start CNCT before UTXN, BRAR, FMGR
 - stop CNCT after UXTN, BRAR, FMGR
 - stop all chain control services after server subsystems

PR #2522, opened 2019-01-21
Cleanup in case of server failed to start

roeierez:
This PR suggests a way to cleanup/release resources in case the server failed to start. 
In case the `Start` function of the server returns an error, there isn't any attempt to stop the sub systems that have been started before the error.
While this behavior doesn't create any major problem when LND is used as an executable, it is problematic when LND is used as a library and needs to be restarted.
For example currently if LND is started without internet connection (common in mobile) it exits and any attempt to run it again (as library) result in a  hang. This hang is due to some systems are still running, not properly stopped and as a result not releasing resources, db files, etc...
The suggested solution here is to trigger cleanup before returning the error, making sure the Start function doesn't leave anything running.

### DIFF
--- a/chainreg/chainregistry.go
+++ b/chainreg/chainregistry.go
@@ -226,7 +226,7 @@ type ChainControl struct {
 // full-node, another backed by a running bitcoind full-node, and the other
 // backed by a running neutrino light client instance. When running with a
 // neutrino light client instance, `neutrinoCS` must be non-nil.
-func NewChainControl(cfg *Config) (*ChainControl, error) {
+func NewChainControl(cfg *Config) (*ChainControl, func(), error) {
 
 	// Set the RPC config from the "home" chain. Multi-chain isn't yet
 	// active, so we'll restrict usage to a particular chain for now.
@@ -264,7 +264,7 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 			DefaultLitecoinStaticFeePerKW, 0,
 		)
 	default:
-		return nil, fmt.Errorf("default routing policy for chain %v is "+
+		return nil, nil, fmt.Errorf("default routing policy for chain %v is "+
 			"unknown", cfg.PrimaryChain())
 	}
 
@@ -294,7 +294,7 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 		heightHintCacheConfig, cfg.LocalChanDB,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("unable to initialize height hint "+
+		return nil, nil, fmt.Errorf("unable to initialize height hint "+
 			"cache: %v", err)
 	}
 
@@ -311,14 +311,14 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 		)
 		cc.ChainView, err = chainview.NewCfFilteredChainView(cfg.NeutrinoCS)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		// Map the deprecated neutrino feeurl flag to the general fee
 		// url.
 		if cfg.NeutrinoMode.FeeURL != "" {
 			if cfg.FeeURL != "" {
-				return nil, errors.New("feeurl and " +
+				return nil, nil, errors.New("feeurl and " +
 					"neutrino.feeurl are mutually exclusive")
 			}
 
@@ -358,7 +358,7 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 			// this back to the btcwallet/bitcoind port.
 			rpcPort, err := strconv.Atoi(cfg.ActiveNetParams.RPCPort)
 			if err != nil {
-				return nil, err
+				return nil, nil, err
 			}
 			rpcPort -= 2
 			bitcoindHost = fmt.Sprintf("%v:%d",
@@ -390,11 +390,11 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 			5*time.Second,
 		)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		if err := bitcoindConn.Start(); err != nil {
-			return nil, fmt.Errorf("unable to connect to bitcoind: "+
+			return nil, nil, fmt.Errorf("unable to connect to bitcoind: "+
 				"%v", err)
 		}
 
@@ -429,7 +429,7 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 				fallBackFeeRate.FeePerKWeight(),
 			)
 			if err != nil {
-				return nil, err
+				return nil, nil, err
 			}
 		} else if cfg.Litecoin.Active && !cfg.Litecoin.RegTest {
 			log.Infof("Initializing litecoind backed fee estimator in "+
@@ -445,7 +445,7 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 				fallBackFeeRate.FeePerKWeight(),
 			)
 			if err != nil {
-				return nil, err
+				return nil, nil, err
 			}
 		}
 
@@ -454,14 +454,14 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
 		// connection.
 		chainConn, err := rpcclient.New(rpcConfig, nil)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		// The api we will use for our health check depends on the
 		// bitcoind version.
 		cmd, err := getBitcoindHealthCheckCmd(chainConn)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		cc.HealthCheck = func() error {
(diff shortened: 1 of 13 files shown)


invoices/invoiceregistry.go
```diff
@@ -4,11 +4,12 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"fmt"
-	"github.com/btcsuite/btcd/chaincfg"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/btcsuite/btcd/chaincfg"
+
```
wpaulino: Nit: can remove this newline.

server.go
```diff
@@ -956,10 +970,24 @@ func (s *server) Start() error {
 		return nil
 	}
 
+	// If one sub system fails to start, the following code
+	// ensures that the  previous started ones are stopped.
```
wpaulino: Nit: Extend comments to 80 chars per line. Also there's an extra space after `the`.

roeierez:
> Would be nice to follow up with #1783 cc @cfromknecht.

I see, would you suggest to wait for that PR to be merged and then rebase this PR? Or any other suggestion how to act on this?

server.go
roeierez: fixed

invoices/invoiceregistry.go
roeierez: fixed

server.go
```diff
@@ -956,10 +970,23 @@ func (s *server) Start() error {
 		return nil
 	}
 
+	// If one sub system fails to start, the following code ensures that the
+	// previous started ones are stopped. It also ensures a proper wallet
+	// shutdown which is important for releasing its resources (boltdb, etc...)
+	cleanup := cleaner{
+		s.cc.chainView.Stop,
+		s.cc.feeEstimator.Stop,
+		func() error {
```
halseth: does this has to be wrapped in a `func` like this?

server.go
```diff
@@ -947,6 +947,20 @@ func (s *server) Started() bool {
 	return atomic.LoadInt32(&s.started) != 0
 }
 
+type cleaner []func() error
```
halseth: should add some comments to these...

server.go
roeierez: good catch! fixed

server.go
roeierez: Added.

invoices/invoiceregistry.go
```diff
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg"
-
```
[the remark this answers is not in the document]
roeierez: squashed

discovery/gossiper.go
```diff
@@ -426,9 +426,9 @@ func (d *AuthenticatedGossiper) Start() error {
 }
 
 // Stop signals any active goroutines for a graceful closure.
-func (d *AuthenticatedGossiper) Stop() {
+func (d *AuthenticatedGossiper) Stop() error {
 	if !atomic.CompareAndSwapUint32(&d.stopped, 0, 1) {
-		return
+		return nil
```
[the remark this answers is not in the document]
roeierez: @halseth I now combines all these error to one and return at the end.
LMK what you think.

discovery/gossiper.go
```diff
@@ -426,23 +427,30 @@ func (d *AuthenticatedGossiper) Start() error {
...
 	}
 
 	log.Info("Authenticated Gossiper is stopping")
 
 	d.blockEpochs.Cancel()
 
 	d.syncerMtx.RLock()
+	var errStrings []string
```
halseth: I think it's enough to log any error, an then just return the last encountered one.

invoices/invoiceregistry.go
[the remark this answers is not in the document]
halseth: Still not fixed :p

halseth:
Needs rebased!

invoices/invoiceregistry.go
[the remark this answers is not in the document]
roeierez: Sorry, now fixed :)

roeierez:
> Needs rebased!

rebased.

server.go
```diff
@@ -1008,10 +1026,21 @@ func (s *server) Start() error {
 		return nil
 	}
 
+	// If one sub system fails to start, the following code ensures that the
+	// previous started ones are stopped. It also ensures a proper wallet
+	// shutdown which is important for releasing its resources (boltdb, etc...)
+	cleanup := cleaner{
+		s.cc.chainView.Stop,
```
cfromknecht: these services are not started w/in the context of the server, thus there cleanup should occur outside the context of the server's stop. they should be cleaned up after all server subsystems have exited

server.go
```diff
@@ -999,6 +999,24 @@ func (s *server) Started() bool {
 	return atomic.LoadInt32(&s.started) != 0
 }
 
+// cleaner is used to aggregate "cleanup" functions during an operation that
+// starts several subsystems. In case one of the subsystem fails to start
+// and a proper resource cleanup is required, the "run" method achieves this
+// by running all these added "cleanup" functions
+type cleaner []func() error
+
+func (c cleaner) add(cleanup func() error) cleaner {
```
cfromknecht: missing comments on these methods

server.go
roeierez: @cfromknecht I agree, just keep in mind stopping these services was already handled in the `server.Stop` function before this PR.
Anyway I was thinking since they are created here: https://github.com/lightningnetwork/lnd/blob/master/chainregistry.go#L130 maybe good approach is to stop them as part of the `cleanup` function returned there.
Calling to `cleanup` already handled by `lnd.go` and then I can safely remove the logic that stops them in `server.Stop`
LMK what you think.

cfromknecht: @roeierez you are indeed correct, they are currently handled in the server even though they should probably be lifted out. I would support the approach you described and moving them to the lnd.go level cleanup assuming it doesn't introduce any weirdness :)

server.go
roeierez: Comments added.

server.go
roeierez: @cfromknecht see my last commit. I explained the reasons in the commit message. 
LMK you feedback please.

wpaulino:
I believe so but it needs a rebase. @roeierez if you're still up for it, we can try to get this in for the next major release (v0.13).

Roasbeef:
@roeierez can you re-request reviews on GH (top right?) when this is ready again? Thanks!

chainreg/chainregistry.go
```diff
@@ -652,18 +652,29 @@ func NewChainControl(cfg *Config) (*ChainControl, error) {
...
-		return nil, err
+		return nil, nil, err
 	}
 
 	log.Info("LightningWallet opened")
 
 	cc.Wallet = lnWallet
 
-	return cc, nil
+	ccCleanup := func() {
```
wpaulino, line 666: Looks like we'd also want to return a closure to shutdown any subsystems already started in the event of an error before getting here.

htlcswitch/htlcnotifier.go
```diff
@@ -87,12 +87,14 @@ func (h *HtlcNotifier) Start() error {
 }
 
 // Stop signals the notifier for a graceful shutdown.
-func (h *HtlcNotifier) Stop() {
+func (h *HtlcNotifier) Stop() error {
```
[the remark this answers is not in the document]
roeierez: Done

chainreg/chainregistry.go
roeierez: Right.
I removed the call to `cc.ChainView.Stop()` as its lifecycle is managed by the router and the code now returs the ccCleanup on previous errors.

server.go
```diff
@@ -1707,30 +1788,49 @@ func (s *server) Stop() error {
...
-		s.sweeper.Stop()
-		s.channelNotifier.Stop()
-		s.peerNotifier.Stop()
-		s.htlcNotifier.Stop()
-		if err := s.cc.Wallet.Shutdown(); err != nil {
-			srvrLog.Warnf("Unable to stop Wallet: %v", err)
-		}
-		if err := s.cc.ChainView.Stop(); err != nil {
-			srvrLog.Warnf("Unable to stop ChainView: %v", err)
+		if err := s.chanRouter.Stop(); err != nil {
```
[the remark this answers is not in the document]
roeierez: I agree in theory and aimed to do so as part of this PR. The reason I didn't went for this change is that I the order of Stop calls is not the reverse order of the Start calls (and also not the same order as the Start calls).
I am not confident that it won't introduce any issue but if one of the team supports this I will do it.

server.go
```diff
@@ -1432,17 +1432,45 @@ func (s *server) Started() bool {
...
+
+// add is used to add a cleanup function to be called when
+// the run function is executed
+func (c cleaner) add(cleanup func() error) cleaner {
+	return append(c, cleanup)
+}
+
+// run is used to run all the previousely added cleanup functions
+func (c cleaner) run() {
+	for _, cleanup := range c {
```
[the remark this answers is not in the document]
roeierez: Done

halseth (approved):

server.go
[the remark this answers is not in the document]
wpaulino: Sounds good, we could possibly replace https://github.com/lightningnetwork/lnd/pull/1783.

server.go
```diff
@@ -1432,17 +1432,45 @@ func (s *server) Started() bool {
...
+// the run function is executed
+func (c cleaner) add(cleanup func() error) cleaner {
+	return append(c, cleanup)
+}
+
+// run is used to run all the previousely added cleanup functions
+func (c cleaner) run() {
+	for i := len(c) - 1; i >= 0; i-- {
+		if err := c[i](); err != nil {
+			srvrLog.Infof("Cleanup failed: %v", err)
```
wpaulino, line 1451: Nit: use warn level for consistency with others below?

wpaulino (approved):

Roasbeef:
The `go-acc` test failure has been resolved in master.

PR #2533, opened 2019-01-23
add named volume to lnd service for persisting data accross container…

yancyribbens:
Add a named volume to `docker-compose.yml` for persisting lnd application volume state.  Without this volume, if the container is brought down (for example using `docker-compose down`) the state is lost, and lnd needs to resync, eg:

`lnd_btc    | 2019-01-18 16:38:41.154 [INF] LNWL: Caught up to height 10000`

Adding this volume will persist the state along with `btc` similar to how a local install would work.  The new volume can be inspected using `docker volume inspect docker_lnd `.  Using a named volume in `docker-compose` allows `docker` to re-attach to the named volume during `docker up`.  This is preferable to an anonymous volume which isn't recognized by `docker-compose` and so this PR replaces [#2507](https://github.com/lightningnetwork/lnd/pull/2507)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -9,8 +9,8 @@ services:
       build:
         context: btcd/
       volumes:
-            - shared:/rpc
-            - bitcoin:/data
+        - shared:/rpc
+        - bitcoin:/data
       environment:
         - RPCUSER
         - RPCPASS
@@ -41,8 +41,8 @@ services:
       build:
         context: ltcd/
       volumes:
-            - shared:/rpc
-            - litecoin:/data
+        - shared:/rpc
+        - litecoin:/data
       environment:
         - RPCUSER
         - RPCPASS
@@ -76,7 +76,8 @@ services:
           - CHAIN
           - DEBUG
         volumes:
-            - shared:/rpc
+          - shared:/rpc
+          - lnd:/root/.lnd
         entrypoint: ["./start-lnd.sh"]
 
     lnd_ltc:
@@ -100,6 +101,11 @@ volumes:
   # bitcoin volume is needed for maintaining blockchain persistence
   # during btcd container recreation.
   bitcoin:
+    driver: local
+
+  # lnd volume is used for persisting lnd application data and chain state
+  # during container lifecycle.
+  lnd:
     driver: local
 
   # litecoin volume is needed for maintaining blockchain persistence
--- a/docker/README.md
+++ b/docker/README.md
@@ -63,8 +63,12 @@ bitcoin into.
 # Init bitcoin network env variable:
 $ export NETWORK="simnet" 
 
+# Create persistent volumes for alice and bob.
+docker volume create simnet_lnd_alice
+docker volume create simnet_lnd_bob
+
 # Run the "Alice" container and log into it:
-$ docker-compose run -d --name alice lnd_btc
+$ docker-compose run -d --name alice --volume simnet_lnd_alice:/root/.lnd lnd_btc
 $ docker exec -i -t alice bash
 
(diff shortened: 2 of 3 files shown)


halseth:
Does this also replace #697?

yancyribbens:
Yes unless you prefer to cherry pick from #697 instead.  When I opened #2507 I wasn't aware of #697.

halseth:
This is a much simpler change, so i would prefer sticking to this.

yancyribbens:
SGTM

yancyribbens:
This PR can be found here: https://github.com/lightningnetwork/lnd/pull/2675

darwin:
This is definitely needed. I based my mainnet lnd setup on these docker scripts and did this change when tweaking it. But there is one additional subtle change you should make.

You should remove custom `--logdir`
https://github.com/lightningnetwork/lnd/blob/2b43da422093c077fd27270f0bc0f3d1ad4c166c/docker/lnd/start-lnd.sh#L54

By default, lnd logs into lnd's data folder `/root/.lnd/logs/...` which is what you want because that folder is mapped to host directory and persists. Logging into forced `/data` inside container would not persist logs between container rebuilds. Logs are not a big issue with simnet, but if someone took your setup (like I did) and based mainnet lnd setup on it, this could be an easy mistake.

yancyribbens:
Thanks @darwin.  I've updated the PR to use the default home directory.

docker/docker-compose.yml
```diff
@@ -100,6 +101,11 @@ volumes:
   bitcoin:
     driver: local
 
+  # lnd volume is used for persisting lnd application data and chain state
+  # during container lifecycle
```
guggero: nit: missing full stop at end of sentence

docker/docker-compose.yml
```diff
@@ -74,7 +74,8 @@ services:
           - CHAIN
           - DEBUG
         volumes:
-            - shared:/rpc
+          - shared:/rpc
+          - lnd:/root/.lnd
```
guggero: I just had a look at #697 which has the same goal.
Adding this persistent volume will create a problem for the tutorial in `README.md`.
Before this change, both nodes `alice` and `bob` stored their data in the non-persistent file system of their respective containers.
Now, they would both use the same volume and overwrite the data.
I don't really like like the approach in #697 that adds two extra configurations for `alice` and `bob` in the `docker-compose.yml`.
How about this instead (since users will copy/paste the commands from the tutorial anyway):
Changes to `docker/README.md`:

```
# Init bitcoin network env variable:
$ export NETWORK="simnet" 

# Create persistent volumes for alice and bob.
docker volume create simnet_lnd_alice
docker volume create simnet_lnd_bob

# Run the "Alice" container and log into it:
$ docker-compose run -d --name alice --volume simnet_lnd_alice:/root/.lnd lnd_btc
```

and further down:

```
# Run "Bob" node and log into it:
$ docker-compose run -d --name bob --volume simnet_lnd_bob:/root/.lnd lnd_btc
```

yancyribbens: good catch :fish:

docker/docker-compose.yml
yancyribbens: :+1:

guggero:
Thanks for the changes! I tested the tutorial again and it works as expected.

Something went wrong with the commits, now there are 5 instead of one. Feel free to ping me on Slack if you need help with git.

yancyribbens:
@guggero sure no problem! i'll rebase the commit history and ping you if needed.

yancyribbens:
@guggero comments have been squashed

PR #2535, opened 2019-01-23
multi: replace manual CAS with sync.Once in component start/stop

federicobond:
The change has been implemented only for `fundingManager` as its meant mostly to gather feedback. I noticed many modules using manual CAS operations to guarantee that the start/stop subroutines run at most once.

One characteristic of the current implementation is that subsequent callers will return immediately even if the first call has not completed yet. This is contrary to what one of the comments implies when it says: "This method will block until all goroutines have exited." causing a potential race condition for subsequent callers.

If there is interest in the change, I can replace other manual CAS operations with `sync.Once` blocks.

### DIFF
--- a/autopilot/agent.go
+++ b/autopilot/agent.go
@@ -6,7 +6,6 @@ import (
 	"math/rand"
 	"net"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -105,9 +104,8 @@ func (c channelState) ConnectedNodes() map[NodeID]struct{} {
 //
 // TODO(roasbeef): prob re-word
 type Agent struct {
-	// Only to be used atomically.
-	started uint32
-	stopped uint32
+	started sync.Once
+	stopped sync.Once
 
 	// cfg houses the configuration state of the Ant.
 	cfg Config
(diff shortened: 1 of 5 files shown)


Roasbeef:
>  One characteristic of the current implementation is that subsequent callers will return immediately even if the first call has not completed yet

How isn't this the case with the current implementation? This particular section doesn't use a CAS, instead the atomic increment serves as a serialization barrier, ensuring that only the "first" call succeeds. 

> causing a potential race condition for subsequent callers.

Race condition in starting/stopping?

cfromknecht:
> How isn't this the case with the current implementation? This particular section doesn't use a CAS, instead the atomic increment serves as a serialization barrier, ensuring that only the "first" call succeeds.

The distinction is that with `sync.Once` all subsequent calls block until the first finishes. So in the case of `Stop`, subsequent calls will return early w/o waiting

federicobond:
What Conner said. `fundingmanager` does not use CAS indeed, but every other module that needs a serialization barrier does.

cfromknecht:
@federicobond not forgotten, this is a good suggestion. it has been added to the 0.6.1 milestone. are you still interested in making these changes across the wider codebase?

federicobond:
@cfromknecht sure! I'm on vacation right now but next week I can take it.

cfromknecht:
@federicobond fwiw, i'd suggest breaking these changes up into smaller PRs that target a handful of subsystems. If we try to do it all in one, it will likely lead to lots of conflicts and need to stop-the-world while getting the full set of changes in

cfromknecht:
@federicobond any update on your end? i'm not opposed to getting this PR for 0.7 and starting the process (surprised there are no conflicts) and making an issue to track the remaining packages where this needs to be done

federicobond:
Hi @cfromknecht, sorry for the delay. I have rebased and updated the PR with a couple more components. It is nice to see the pattern has catched on at other parts of the codebase already.

PR #2538, opened 2019-01-25
config: increase default trickle delay from 30s to 1m30s

Roasbeef:
In this commit, we increase the default trickle delay from 30s to 1m30s.
We do this as before we implement the new INV gossip mechanism, we want
to de-emphasise the quick propagation of updates through the network
which eats up bandwidth.

### DIFF
--- a/config.go
+++ b/config.go
@@ -49,7 +49,7 @@ const (
 	defaultRPCHost             = "localhost"
 	defaultMaxPendingChannels  = 1
 	defaultNoSeedBackup        = false
-	defaultTrickleDelay        = 30 * 1000
+	defaultTrickleDelay        = 90 * 1000
 	defaultInactiveChanTimeout = 20 * time.Minute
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -49,7 +49,7 @@ const (
 	defaultRPCHost             = "localhost"
 	defaultMaxPendingChannels  = 1
 	defaultNoSeedBackup        = false
-	defaultTrickleDelay        = 30 * 1000
+	defaultTrickleDelay        = 90 * 1000
```
cfromknecht: `90 * time.Second`?

halseth: that would break old configs I guess, but I agree it would be nice to make all flags use `time.Duration`.

cfromknecht: hmm, yeah that is unfortunate that it doesn't. we could migrate to a `trickleinterval` that uses `time.Duration` and deprecate `trickledelay`, though that can be done later

PR #2542, opened 2019-01-26
Extended autocompletion in bash for lncli

dchirikov:
Autocomplete function for lncli suggest possible sub-commands
and arguments with <TAB> key in bash shell.
Supports all current and, probably, some future arguments, as it parse
`lncli -h` and `lncli subcommand -h` outputs to suggests possible
substitutions.


guggero:
This looks quite involved and specific. I'm personally not sure if this is worth the effort to maintain. Couldn't this be solved more generically by leveraging the completion feature of the library we use for `lncli`? https://github.com/urfave/cli/blob/master/docs/v1/manual.md#bash-completion

Roasbeef:
Closing due to inactivity.

PR #2548, opened 2019-01-27
Add support for Tor authentication via password

wereHamster:
Currently lnd only supports tor authentication via cookie. This requires that lnd has filesystem access to the location where tor writes the cookie. If the user runs lnd and tor in docker containers then sharing the filesystem in this way may not be feasible. Thankfully tor also offers password authentication.

This pull request adds support for this password authentication. If a tor password (`--tor.password=X`) is provided, and the tor server supports password authentication, then it is tried before cookie authentication.

### DIFF
--- a/config.go
+++ b/config.go
@@ -220,6 +220,7 @@ type torConfig struct {
 	StreamIsolation bool   `long:"streamisolation" description:"Enable Tor stream isolation by randomizing user credentials for each connection."`
 	Control         string `long:"control" description:"The host:port that Tor is listening on for Tor control connections"`
 	TargetIPAddress string `long:"targetipaddress" description:"IP address that Tor should use as the target of the hidden service"`
+	Password        string `long:"password" description:"Password for the control port. If provided it will be used instead of the SAFECOOKIE authentication method."`
 	V2              bool   `long:"v2" description:"Automatically set up a v2 onion service to listen for inbound connections"`
 	V3              bool   `long:"v3" description:"Automatically set up a v3 onion service to listen for inbound connections"`
 	PrivateKeyPath  string `long:"privatekeypath" description:"The path to the private key of the onion service being created"`
--- a/tor/controller.go
+++ b/tor/controller.go
@@ -79,6 +79,10 @@ type Controller struct {
 	// controller connections on.
 	controlAddr string
 
+	// Password that will be used for authentication, if the server supports
+	// HASHEDPASSWORD authentication.
+	password string
+
 	// version is the current version of the Tor server.
 	version string
 
@@ -90,8 +94,8 @@ type Controller struct {
 
 // NewController returns a new Tor controller that will be able to interact with
 // a Tor server.
-func NewController(controlAddr string, targetIPAddress string) *Controller {
-	return &Controller{controlAddr: controlAddr, targetIPAddress: targetIPAddress}
+func NewController(controlAddr string, targetIPAddress string, password string) *Controller {
+	return &Controller{controlAddr: controlAddr, targetIPAddress: targetIPAddress, password: password}
 }
 
 // Start establishes and authenticates the connection between the controller and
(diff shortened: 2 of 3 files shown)


wereHamster:
I'm running my node with this patch and the service is exposed through tor (the onion address is advertised and visible on 1ml.com).

tor/controller.go
```diff
@@ -162,15 +166,22 @@ func parseTorReply(reply string) map[string]string {
...
+
+	return nil
+}
+
+func (c *Controller) authenticateViaSafeCookie(info map[string]string) error {
 	// Before proceeding to authenticate the connection, we'll retrieve
 	// the authentication cookie of the Tor server. This will be used
 	// throughout the authentication routine. We do this before as once the
-	// authentication routine has begun, it is not possible to retrieve it
+	// authentication rout
```
wpaulino: Looks like this part of the comment was removed accidentally?

tor/controller.go
```diff
@@ -259,33 +270,52 @@ func (c *Controller) authenticate() error {
...
-	if !safeCookieSupport {
-		return nil, errors.New("the Tor server is currently not " +
-			"configured for cookie authentication")
+	// If the SAFECOOKIE method is supported, try that.
+	if Contains(authMethods, "SAFECOOKIE") {
+		return c.authenticateViaSafeCookie(info)
+	}
+
+	// Don't know how to authenticate. Fail.
+	return fmt.Errorf("[TOR] authenticate: exhausted all auth methods")
```
wpaulino: Would prefer this error message to be: `the Tor server must be configured with SAFECOOKIE or HASHEDPASSWORD authentication`.

tor/controller.go
```diff
@@ -350,46 +380,38 @@ func supportsV3(version string) error {
 
 // ProtocolInfo returns the different authentication methods supported by the
 // Tor server and the version of the Tor server.
-func (c *Controller) ProtocolInfo() ([]string, string, string, error) {
+func (c *Controller) ProtocolInfo() ([]string, string, map[string]string, error) {
```
wpaulino: Would prefer if this just returned all of the authentication methods supported rather than the whole info response.

wereHamster: The supported auth methods is the first return value (the `[]string`). But depending on the method, the actual auth implementation may require additional stuff from the PROTOCOLINFO response. For example the SAFECOOKIE method requires the COOKIEFILE part to know where to load the cookie from (search in the code for `info["COOKIEFILE"]`).

wereHamster:
Extending this PR to support auth-via-NULL should be trivial. See #2490, though that PR does a bit more than just support auth-via-NULL.

tor/controller.go
wpaulino: Gotcha, in that case I think we should bundle all of these return arguments into a struct for easier readability.

wereHamster:
What are the next steps here? How can I help to get this merged?

wereHamster:
There have been some changes to the tor_controller.go file in the meantime (support for NULL authentication). I've merged the changes.

wereHamster:
The tests appear to be flaky, the failures don't seem to be caused by my changes.

wereHamster:
Is there any hope to get this PR merged? Should I continue to update it to keep it mergeable?

halseth:
You don't have to keep to update it before has gotten review, mr hamster.

You should also _rebase_ it on master, instead of merging master into your branch.

meeDamian:
We would love to see that merged as well.

tor/controller.go
```diff
@@ -271,37 +277,57 @@ func (c *Controller) authenticate() error {
...
+	}
+
+	// If we have a password, and the HASHEDPASSWORD auth method is
+	// supported, then try that.
+	if Contains(protocolInfo.authMethods, "HASHEDPASSWORD") && c.password != "" {
+		return c.authenticateViaHashedPassword()
+	}
+
+	// If the SAFECOOKIE method is supported, try that.
+	if Contains(protocolInfo.authMethods, "SAFECOOKIE") {
```
carlaKC: Now that we have a few different auth methods, it would be nice to have `"SAFECOOKIE"` etc declared as constants.

tor/controller.go
```diff
@@ -360,48 +386,59 @@ func supportsV3(version string) error {
 	return nil
 }
 
+// The ProtocolInfo struct is used to wrap everything we get in the response to PROTOCOLINFO
+// into a neat package, so we can pass it around as a single object.
+type ProtocolInfo struct {
```
carlaKC: doesn't need to be exported

wereHamster: What do I need to do? I don't know go at all.

carlaKC: In go a capital letter indicates that something is exported, ie can be called from outside of the package. A lower case letter reserves it only for use within the package, which would be suitable in this case because we don't need it anywhere else. 

[Effective go](https://golang.org/doc/effective_go.html) is a good read for getting to know golang if you want to delve deeper :)

tor/controller.go
```diff
@@ -271,37 +277,57 @@ func (c *Controller) authenticate() error {
...
-		if authMethod == "SAFECOOKIE" {
-			safeCookieSupport = true
-		}
-		if authMethod == "NULL" {
-			return nil, nil
-		}
+	c.version = protocolInfo.version
+
+	// NULL authentication is the simplest, use it if it's available.
+	if Contains(protocolInfo.authMethods, "NULL") {
```
wpaulino: I'd assume we always want to use the most secure authentication method instead?

tor/controller.go
```diff
@@ -271,37 +277,57 @@ func (c *Controller) authenticate() error {
...
+	// If the SAFECOOKIE method is supported, try that.
+	if Contains(protocolInfo.authMethods, "SAFECOOKIE") {
+		return c.authenticateViaSafeCookie(protocolInfo)
 	}
 
-	if !safeCookieSupport {
-		return nil, errors.New("the Tor server is currently not " +
-			"configured for cookie or null authentication")
+	// Don't know how to authenticate. Fail.
+	return fmt.Errorf("the Tor server must be configured with NULL, SAFECOOKIE, or HASHEDPASSWORD authentication")
```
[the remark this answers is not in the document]
wpaulino: There are a few other cases throughout the diff as well.

config.go
```diff
@@ -220,6 +220,7 @@ type torConfig struct {
 	StreamIsolation bool   `long:"streamisolation" description:"Enable Tor stream isolation by randomizing user credentials for each connection."`
 	Control         string `long:"control" description:"The host:port that Tor is listening on for Tor control connections"`
 	TargetIPAddress string `long:"targetipaddress" description:"IP address that Tor should use as the target of the hidden service"`
+	Password        string `long:"password" description:"Password for the control port. If provided it will be used instead of the SAFECOOKIE authentication method."`
```
wpaulino: SAFECOOKIE and NULL*

wpaulino:
Thanks for getting this started @wereHamster! We've decided to get this in for our `v0.10.0` release, so we'll take it over from here. The new PR has you as a co-author to give you credit.

PR #2549, opened 2019-01-28
routing: ignore ChannelUpdates for unknown channels

halseth:
Since we should always get a `ChanInfo` before a channel update, we can exit early if we receive a `ChannelUpdate` for an edge not in the graph.

Earlier we would attempt to `GetUTXO` in the case we couldn't find it in the graph, which is a heavy operation for some backends, with the subsequent `UpdateEdgePolicy` failing regardless.

Broken off of #2279

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -1125,8 +1125,6 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 		}
 		r.rejectMtx.RUnlock()
 
-		channelID := lnwire.NewShortChanIDFromInt(msg.ChannelID)
-
 		// We make sure to hold the mutex for this channel ID,
 		// such that no other goroutine is concurrently doing
 		// database accesses for the same channel ID.
@@ -1140,6 +1138,15 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 			return errors.Errorf("unable to check for edge "+
 				"existence: %v", err)
 
+		}
+
+		// If the channel doesn't exist in our database, we cannot
+		// apply the updated policy.
+		if !exists {
+			return newErrf(ErrIgnored, "Ignoring update "+
+				"(flags=%v|%v) for unknown chan_id=%v",
+				msg.MessageFlags, msg.ChannelFlags,
+				msg.ChannelID)
 		}
 
 		// As edges are directional edge node has a unique policy for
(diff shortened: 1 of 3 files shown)


routing/router_test.go
```diff
@@ -1025,6 +1025,76 @@ func TestIgnoreNodeAnnouncement(t *testing.T) {
 	}
 }
 
+// TestIgnoreChannelEdgePolicyForUnknownChannel checks that a router will
+// ignore a channel policy for a channel not in the graph.
+func TestIgnoreChannelEdgePolicyForUnknownChannel(t *testing.T) {
+	t.Parallel()
+
+	const startingBlockHeight = 101
+	ctx, cleanUp, err := createTestCtxFromFile(startingBlockHeight,
```
wpaulino: Nit: function args should be formatted on their own line.

routing/router_test.go
```diff
@@ -1025,6 +1025,76 @@ func TestIgnoreNodeAnnouncement(t *testing.T) {
...
+
+	var pub1 [33]byte
+	copy(pub1[:], priv1.PubKey().SerializeCompressed())
+
+	var pub2 [33]byte
+	copy(pub2[:], priv2.PubKey().SerializeCompressed())
+
+	// Add the edge between the two unknown nodes to the graph, and check
+	// that the nodes are found after the fact.
+	fundingTx, _, chanID, err := createChannelEdge(ctx,
```
wpaulino: Same here.

routing/router_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2549/commits/29bd22862303c36fafdbcdd7df8620b752c8c8c2

routing/router_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2549/commits/29bd22862303c36fafdbcdd7df8620b752c8c8c2

halseth:
PTAL @wpaulino

routing/router_test.go
```diff
@@ -1025,6 +1025,77 @@ func TestIgnoreNodeAnnouncement(t *testing.T) {
 	}
 }
 
+// TestIgnoreChannelEdgePolicyForUnknownChannel checks that a router will
+// ignore a channel policy for a channel not in the graph.
+func TestIgnoreChannelEdgePolicyForUnknownChannel(t *testing.T) {
+	t.Parallel()
+
+	const startingBlockHeight = 101
+	ctx, cleanUp, err := createTestCtxFromFile(
```
joostjager: If you use `createTestGraphFromChannels`, you can create an empty graph and remove this (I think) unnecessary dependency on the test graph file.

routing/router.go
```diff
@@ -1142,6 +1140,15 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 
 		}
 
+		// If the channel doesn't exist in our database, we cannot
+		// apply the updated policy.
+		if !exists {
+			return newErrf(ErrIgnored, "Ignoring update "+
```
joostjager: We return a slightly different error object than before. Is it okay with the callers?

routing/router_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2549/commits/69ad94c68b696e79661fab322d6f0e90efdfb83d

halseth:
PTAL @joostjager  @wpaulino

routing/router.go
joostjager: Can the situation occur if something happens in between the two db queries for the channel?

wpaulino: Even if it does, the error is handled correctly by the caller here:
https://github.com/lightningnetwork/lnd/blob/351865ac6772e7885a64c7e0c1e154abade40c91/discovery/gossiper.go#L2034

joostjager: Ok, clear.

The other use of `UpdateEdge` is in `ChannelRouter.applyChannelUpdate`. But I think the behaviour improves there too?

halseth: Yes looks okay: https://github.com/lightningnetwork/lnd/blob/e10fe91f10efe46c540456ee5e22bfbd929fbe0d/routing/router.go#L2054

PR #2550, opened 2019-01-28
autopilot: don't block Stop on pending connection

halseth:
To ensure a call to ConnectToPeer doesn't block the agent from
shutting down, we'll launch it in a non-waitgrouped goroutine, that
will signal when a result is returned.

This could cause `disabling` autopilot to hang if there were pending connections.

### DIFF
--- a/autopilot/agent_test.go
+++ b/autopilot/agent_test.go
@@ -1546,3 +1546,156 @@ func TestAgentSkipPendingConns(t *testing.T) {
 		t.Fatalf("agent should have attempted connection")
 	}
 }
+
+// TestAgentQuitWhenPendingConns tests that we are able to stop the autopilot
+// agent even though there are pending connections to nodes.
+func TestAgentQuitWhenPendingConns(t *testing.T) {
+	t.Parallel()
+
+	// First, we'll create all the dependencies that we'll need in order to
+	// create the autopilot agent.
+	self, err := randKey()
+	if err != nil {
+		t.Fatalf("unable to generate key: %v", err)
+	}
+
+	quit := make(chan struct{})
+	defer close(quit)
+
+	heuristic := &mockHeuristic{
+		nodeScoresArgs:  make(chan directiveArg),
+		nodeScoresResps: make(chan map[NodeID]*NodeScore),
+		quit:            quit,
+	}
+	constraints := &mockConstraints{
+		moreChansResps: make(chan moreChansResp),
+		quit:           quit,
+	}
+
+	chanController := &mockChanController{
+		openChanSignals: make(chan openChanIntent),
+	}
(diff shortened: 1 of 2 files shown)


autopilot/agent_test.go
```diff
@@ -1546,3 +1546,157 @@ func TestAgentSkipPendingConns(t *testing.T) {
...
+	select {
+	case <-connect:
+	case <-time.After(time.Second * 10):
+		t.Fatalf("agent did not attempt connection")
+	}
+
+	// Make sure that we are able to stop the agent, even though there is a
+	// pending connection.
+	stopped := make(chan error)
+	go func() {
```
cfromknecht: style nit: this can all be oneline :)

autopilot/agent_test.go
```diff
@@ -1546,3 +1546,157 @@ func TestAgentSkipPendingConns(t *testing.T) {
...
+		t.Fatalf("unable to create agent: %v", err)
+	}
+
+	// To ensure the heuristic doesn't block on quitting the agent, we'll
+	// use the agent's quit chan to signal when it should also stop.
+	heuristic.quit = agent.quit
+
+	// With the autopilot agent and all its dependencies we'll start the
+	// primary controller goroutine.
+	// NOTE: We'll stop the agent manually below.
```
cfromknecht: should defer `Stop` anyway in case we fail before reaching the manual stop

autopilot/agent_test.go
```diff
@@ -1546,3 +1546,157 @@ func TestAgentSkipPendingConns(t *testing.T) {
 		t.Fatalf("agent should have attempted connection")
 	}
 }
+
+// TestAgentQuitWhenPendingConns tests that we are able to stop the autopilot
+// agent even though there are pending connections to nodes.
+func TestAgentQuitWhenPendingConns(t *testing.T) {
```
wpaulino: Not a blocker, but looks like we'd really benefit from abstracting some common patterns within the tests into helper functions for easier readability.

autopilot/agent_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2550/commits/1b26041e2a18afabddafb7a3f858e6661be807bf

autopilot/agent_test.go
halseth: Ah, like this? https://github.com/lightningnetwork/lnd/pull/2550/commits/eaf087ed89caac23fa0395863e3b7ab055c79e17

halseth:
PTAL @cfromknecht @wpaulino

wpaulino:
LGTM after squash.

halseth:
Squashed and rebased.

PR #2363, opened 2018-12-23
Deprecate IncorrectHtlcAmount

orbitalturtle:
Fixes #2321

PR #2554, opened 2019-01-29
multi: deprecate IncorrectHtlcAmount 

### LINKED ISSUE #2321: lnwire+routing: deprecate FinalIncorrectHtlcAmount in favor of unified UnknownPaymentHash
In order to fix a probing vector, [we've recently made a modification in the specification to phase out 
`FinalIncorrectHtlcAmount` in favor of adding an additional payload to `UnknownPaymentHash` of the final value the destination receives](https://github.com/lightningnetwork/lightning-rfc/commit/914ebab9080ccccb0ff176cb16b7a6ba21e23bbe). This change coalesces the two errors, making an unknown payment hash error indistinguishable from the case wherein an intermediate node modifies the value of the final HTLC. Otherwise, if this isn't patched, then a node can probe the final destination of an HTLC as the node that returns `FinalIncorrectHtlcAmount` reveals that the know the payment hash pre-image, but the value is incorrect. After this change, lack of knowledge of a payment hash extended, and an incorrect value for a known payment hash will be indistinguishable. 

# Steps To Completion 

- [ ] Extend [`UnknownPaymentHash`](https://github.com/lightningnetwork/lnd/blob/master/lnwire/onion_error.go#L297) include a new `lnwire.MiliSatoshi` field of the payment amount. At the end, the error type should properly implement the [`Serializable`](https://github.com/lightningnetwork/lnd/blob/master/lnwire/message.go#L139:6) interface. 

- [ ] Within the link, [whenever we send back an `UnknownPaymentHash` error, we should _always_ set the value of the incoming HTLC](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L2312). 

- [ ] Within the link, [stop returning `FailIncorrectPaymentAmount ` and instead return `UnknownPaymentHash` populated as we do above](https://github.com/lightningnetwork/lnd/blob/42c45979215e3242ef2e162ccd9de3cbb0ed28a1/htlcswitch/link.go#L2361).

Roasbeef:
Fixes #2321 (see issue for details). 

One implementation level thing to note is that we'll still expect this error on the network when decoding and during routing retry attempts. Otherwise, we won't be able to parse the errors from nodes in the wild that haven't upgraded to this new version. 

Replaces #2363.

### DIFF
--- a/lnd_test.go
+++ b/lnd_test.go
@@ -7878,10 +7878,11 @@ out:
 
 	// Next, we'll test the case of a recognized payHash but, an incorrect
 	// value on the extended HTLC.
+	htlcAmt := lnwire.NewMSatFromSatoshis(1000)
 	sendReq = &lnrpc.SendRequest{
 		PaymentHashString: hex.EncodeToString(carolInvoice.RHash),
 		DestString:        hex.EncodeToString(carol.PubKey[:]),
-		Amt:               1000, // 10k satoshis are expected.
+		Amt:               int64(htlcAmt.ToSatoshis()), // 10k satoshis are expected.
 	}
 	ctxt, _ = context.WithTimeout(ctxb, defaultTimeout)
 	resp, err = net.Alice.SendPaymentSync(ctxt, sendReq)
@@ -7895,10 +7896,17 @@ out:
 		t.Fatalf("payment should have been rejected due to wrong " +
 			"HTLC amount")
 	}
-	expectedErrorCode = lnwire.CodeIncorrectPaymentAmount.String()
+	expectedErrorCode = lnwire.CodeUnknownPaymentHash.String()
 	if !strings.Contains(resp.PaymentError, expectedErrorCode) {
 		t.Fatalf("payment should have failed due to wrong amount, "+
 			"instead failed due to: %v", resp.PaymentError)
+	}
+
+	// We'll also ensure that the encoded error includes the invlaid HTLC
+	// amount.
+	if !strings.Contains(resp.PaymentError, htlcAmt.String()) {
+		t.Fatalf("error didn't include expected payment amt of %v: "+
+			"%v", htlcAmt, resp.PaymentError)
 	}
 
 	// The balances of all parties should be the same as initially since
--- a/lnwire/onion_error.go
+++ b/lnwire/onion_error.go
@@ -359,7 +359,7 @@ func (f FailUnknownPaymentHash) Code() FailCode {
 //
 // NOTE: Implements the error interface.
 func (f FailUnknownPaymentHash) Error() string {
-	return f.Code().String()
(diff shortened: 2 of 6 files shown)


lnd_test.go
```diff
@@ -7895,12 +7896,19 @@ out:
...
-	expectedErrorCode = lnwire.CodeIncorrectPaymentAmount.String()
+	expectedErrorCode = lnwire.CodeUnknownPaymentHash.String()
 	if !strings.Contains(resp.PaymentError, expectedErrorCode) {
 		t.Fatalf("payment should have failed due to wrong amount, "+
 			"instead failed due to: %v", resp.PaymentError)
 	}
 
+	// We'll also ensure that the encoded error includes the invlaid HTLC
+	// amount.
+	if !strings.Contains(resp.PaymentError, htlcAmt.String()) {
```
cfromknecht: this will also pass for any value with 1000 as the prefix

routing/router.go
```diff
@@ -1849,6 +1849,11 @@ func (r *ChannelRouter) sendPayment(payment *LightningPayment,
 			case *lnwire.FailUnknownPaymentHash:
 				return preImage, nil, sendError
 
+			// If we sent the wrong amount to the destination, then
+			// we'll exit early.
+			case *lnwire.FailIncorrectPaymentAmount:
+				return preImage, nil, sendError
+
```
cfromknecht: should this be squashed with the prior commit?

lnd_test.go
cfromknecht: nvm, realized that `String` appends the unit so this shouldn't be a problem

routing/router.go
Roasbeef: Yeah I'll squash in all the fixups into the original commit that starts the switch over.

lnwire/onion_error.go
```diff
@@ -294,49 +294,73 @@ func (f FailUnknownNextPeer) Error() string {
...
+//
+// 2) if the amount paid is less than the amount expected, the final node MUST
+// fail the HTLC. If the amount paid is more than twice the amount expected,
+// the final node SHOULD fail the HTLC. This allows the sender to reduce
+// information leakage by altering the amount, without allowing accidental
+// gross overpayment.
 //
 // NOTE: May only be returned by the final node in the path.
-type FailIncorrectPaymentAmount struct{}
+type FailUnknownPaymentHash struct {
```
cfromknecht: we should consider adding a constructor for this error, and migrate all uses of this error in the code base to use it. this should help mitigate the chance of errors introduced by not setting `Amount`. In addition, I think it'd be worth making the amount private, so that it can only be accessed via `Amount()`

Roasbeef: Fixed in latest diff!

PR #2556, opened 2019-01-29
lnrpc+rpcserver: return all utxos by default from ListUnspent

cfromknecht:
This PR modifies the behavior of `ListUnspent` to return all utxos in the wallet when given all fields in the request are zero-valued. This makes things easier for users of the rpc interface, by not having to manually specify a `max_confs` parameter to see all utxos. Similarly, users of `lncli` will be able to query `listunspent` without any additional arguments. From lncli's perspective, some of the checks and default-setting has been moved into the rpcserver so that this applies to all rpc requests.

We implement this via a new `unconfirmed_only` flag on the `ListUnspentRequest` message. By default the value is false, and will cause a `max_confs` value of 0 to be overridden with `math.MaxInt32` before querying the wallet. If the user actually wishes to see only unconfirmed transactions (`min_confs=0` and `max_confs=0`), specifying `unconfirmed_only=true` will prevent the overriding from happening.

Finally, we add additional assertions to our integration tests to indirectly assert the correctness of `ListUnspent` when called with true and false values of `unconfirmed_only`.


cfromknecht:
@joostjager IMO this _is_ making the rpc interface cleaner, because it allows me to pass in a zero-value request and get the response that most would want/expect. To me it's better that it "just works" for 95% of cases out of the box w/o having to turn to documentation, but I do understand your point of keeping the actual interfaces minimal

cfromknecht:
@joostjager thanks for the feedback, i've removed the `unconfirmed_only` field from the rpc interface and made it purely an lncli thing. the itests have been updated to use the raw API, ptal

lnd_test.go
```diff
@@ -643,8 +644,8 @@ func testOnchainFundRecovery(net *lntest.NetworkHarness, t *harnessTest) {
 	// method takes the expected value of Carol's balance when using the
 	// given recovery window. Additionally, the caller can specify an action
 	// to perform on the restored node before the node is shutdown.
-	restoreCheckBalance := func(expAmount int64, recoveryWindow int32,
-		fn func(*lntest.HarnessNode)) {
+	restoreCheckBalance := func(expAmount int64, expectedNumUTXOs uint32,
```
joostjager: `expectedNumUTXOs` as `int` to prevent cast further down?

lntest/harness.go
```diff
@@ -1297,6 +1298,41 @@ func (n *NetworkHarness) sendCoins(ctx context.Context, amt btcutil.Amount,
 		return err
 	}
 
+	// Encode the pkScript in hex as this the format that it will be
+	// returned via rpc.
+	expPkScriptStr := hex.EncodeToString(addrScript)
+
+	// Now, wait for ListUnspent to show the unconfirmed transaction
+	// containing the correct pkscript.
+	err = WaitNoError(func() error {
```
joostjager: Usage of `WaitNoError` :+1:

lntest/harness.go
```diff
@@ -1,6 +1,7 @@
 package lntest
 
 import (
+	"encoding/hex"
```
joostjager: Commit message out of date

cmd/lncli/commands.go
```diff
@@ -321,15 +326,11 @@ func listUnspent(ctx *cli.Context) error {
...
-	default:
-		// No maxconfs was specified; we use max as flag;
-		// the default for ctx.Int64 (0) is *not* appropriate here.
-		maxConfirms = math.MaxInt32
 	}
 
-	if minConfirms < 0 || maxConfirms < minConfirms {
-		return fmt.Errorf("maximum confirmations must be greater or " +
-			"equal to minimum confirmations")
+	unconfirmedOnly = ctx.Bool("unconfirmed_only")
```
joostjager: force `minConfirms` to `0` when `unconfirmedOnly`?

joostjager: Maybe even give an error when `minConfirms` or `maxConfirms` is present in combination with `unconfirmedOnly`

cfromknecht:
Thanks for the comments @joostjager, also rebased on master. ptal

joostjager:
Happy that you agreed to leave the luxury flags out of the API

cfromknecht:
@joostjager luxury flags 😂indeed i agree it is cleaner

@Roasbeef ready for another pass, ptal

alexbosworth:
Sorry I missed the mention. In general I agree 100% with pushing UI knobs out to the edge but it's a wider design discussion

cfromknecht:
@Roasbeef rebased, ptal

PR #2557, opened 2019-01-29
Fix make rpc generating diffs on chainrpc.

Bluetegu:
At least on my machine make rpc on master would created proto diffs on chainrpc. Adding the go_package option seems to fix it (and align with other sub services proto files all including this option)

### DIFF
--- a/lnrpc/chainrpc/chainnotifier.proto
+++ b/lnrpc/chainrpc/chainnotifier.proto
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package chainrpc;
 
+option go_package = "github.com/lightningnetwork/lnd/lnrpc/chainrpc";
+
 message ConfRequest {
     /*
     The transaction hash for which we should request a confirmation notification


Bluetegu:
It failed twice on travis cover. There is no point in trying to wiggle through that in this kind of PR.

Bluetegu:
Actually, this is stranger than I thought. The reason the above change 'fixes' it for me is that make rpc generates the files under lnd/lnrpc/github.com/lightningnework/lnd/lnrpc/x for each x, instead in the tracked location.

So while the change above aligns the chainnotifier with the other sub services, it doesn't fix the original problem that make rpc creates diffs with each of the sub services.

joostjager:
The most likely cause is that your tools are not up to date. https://github.com/lightningnetwork/lnd/blob/master/lnrpc/README.md

Roasbeef:
Closing as I think it's an issue with your dependancies.

PR #2558, opened 2019-01-29
Channel restoration of old format fee update

### LINKED ISSUE #2546: segfault in lnwallet.removeUpdate()
### Background
A few days ago I noticed two channels had automatically force closed locally after some commitment disagreement (which I believed to be bug #1908), so I decided to stop running my node until a fix is released to prevent further channels from force closing.
Today I saw that a release candidate with the fix was released and decided to give it a try.

After syncing to the current height for a while, `lnd` crashed. These are the last two lines in the log:
```
2019-01-27 11:04:05.368 [INF] HSWC: ChannelPoint(065156341c98952a285b9b374dd2d3fd4d4040d4315268cf9c543819625205c0:0): updating commit fee to 2426 sat/kw
2019-01-27 11:04:14.370 [INF] HSWC: ChannelLink(065156341c98952a285b9b374dd2d3fd4d4040d4315268cf9c543819625205c0:0) has exited
```
And here's the stderr of the process from my systemd journal:
```
Jan 27 11:04:14 lnd[11218]: panic: runtime error: invalid memory address or nil pointer dereference
Jan 27 11:04:14 lnd[11218]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x4f88fc]
Jan 27 11:04:14 lnd[11218]: goroutine 8391 [running]:
Jan 27 11:04:14 lnd[11218]: container/list.(*List).Remove(...)
Jan 27 11:04:14 lnd[11218]:         /usr/local/go/src/container/list/list.go:123
Jan 27 11:04:14 lnd[11218]: github.com/lightningnetwork/lnd/lnwallet.(*updateLog).removeUpdate(...)
Jan 27 11:04:14 lnd[11218]:         /Users/roasbeef/gocode/src/github.com/lightningnetwork/lnd/lnwallet/channel.go:1170
Jan 27 11:04:14 lnd[11218]: github.com/lightningnetwork/lnd/lnwallet.compactLogs.func1(0x262f040, 0x262f180)
Jan 27 11:04:14 lnd[11218]:         /Users/roasbeef/gocode/src/github.com/lightningnetwork/lnd/lnwallet/channel.go:1238 +0x1b4
Jan 27 11:04:14 lnd[11218]: github.com/lightningnetwork/lnd/lnwallet.compactLogs(0x262f040, 0x262f180, 0x4e2, 0x0, 0x4e3, 0x0)
Jan 27 11:04:14 lnd[11218]:         /Users/roasbeef/gocode/src/github.com/lightningnetwork/lnd/lnwallet/channel.go:1252 +0x60
```

### Your environment

* `lnd` v0.5.2-beta-rc1 (GitHub armv7 pre-release)
* Raspberry Pi 3 B+
* Raspbian stretch (`Linux 4.14.79-v7+ #1159 SMP Sun Nov 4 17:50:20 GMT 2018 armv7l GNU/Linux`)
* `bitcoind` v0.17.0.1

halseth:
This PR fixes a bug that was introduced in #2397, where we started restoring `FeeUpdate`s from disk as other HTLC updates. The problem arose if the prior version of `lnd` wrote the updates to disk, and the new version attempted to restore. Since the old version did not write the `logIndex` for `FeeUpdate`s, this would be 0, causing it to be out of sync with the log update index, causing problems.

We fix this by setting the `logIndex` to the current log update index in case it is unset. A test that would previously cause this desync and following crash during `compactLogs` is included.

The last commit contains a sanity check that ensures all updates have their log index properly set, to catch these bugs in the future. It is not strictly necessary, but I thought it could be nice to catch it already here, and not in some obscure situation down the road.

Fixes #2546


lnwallet/channel.go
```diff
@@ -1804,6 +1804,24 @@ func (lc *LightningChannel) restoreStateLogs(
 			return err
 		}
 
+		// Earlier versions did not write the log index to disk for fee
+		// updates, so they will be unset. To account for this we set
+		// them to to current update log index.
+		if payDesc.EntryType == FeeUpdate && payDesc.LogIndex == 0 {
+			payDesc.LogIndex = lc.localUpdateLog.logIndex
+			walletLog.Debugf("Found FeeUpdate on "+
```
cfromknecht: do we also want to add `lc.localUpdateLog.logIndex > 0` to the condition above? otherwise it seems this log will be triggered if a fee update is the first entry in the log. There'd be no change in behavior, as the final log index will again be set to zero, though i wonder if the log message might confuse people

halseth: Agreed, fixed!

lnwallet/channel_test.go
```diff
@@ -4048,6 +4048,168 @@ func TestChannelRetransmissionFeeUpdate(t *testing.T) {
...
+		if _, err := bobChannel.ReceiveHTLC(htlc); err != nil {
+			t.Fatalf("unable to recv htlc: %v", err)
+		}
+		htlcs = append(htlcs, htlc)
+
+		if i != numHTLCs/2 {
+			continue
+		}
+
+		// When half of the HTLCs are sent, also include a fee update.
```
Roasbeef: Any reason why this just isn't split up to be outside the loop? As is, there'll be multiple fee updates FWIW.

lnwallet/channel_test.go
```diff
@@ -4048,6 +4048,168 @@ func TestChannelRetransmissionFeeUpdate(t *testing.T) {
...
+	if _, err := aliceChannel.AddHTLC(htlc, nil); err != nil {
+		t.Fatalf("unable to add htlc: %v", err)
+	}
+	if _, err := bobChannel.ReceiveHTLC(htlc); err != nil {
+		t.Fatalf("unable to recv htlc: %v", err)
+	}
+	if err := forceStateTransition(aliceChannel, bobChannel); err != nil {
+		t.Fatalf("unable to complete bob's state transition: %v", err)
+	}
+}
```
Roasbeef: Should assert that the logs are once again clear of all fee updates.

PR #2561, opened 2019-01-30
Revert "funding+lnwallet: ensure max_htlc_value_in_flight smaller tha…

Roasbeef:
…n capacity"

This reverts commit 4aa52d267f000f84caf912c62fc14a5b8e7cacb5.

It turns out that the other implementations set values for this field
which aren't based on the actual capacity of the channel. As a result,
we'll no reject most of their channel offerings, since they may offer a
value of a max `uint64` or something else hard coded that's above the
size of the channel. As a result, we're reverting this check for now to
maintain proper compatibility.


Roasbeef:
Removed.

EagleTM:
It'd be great to also cherry-pick this to a possible v0.5.2-beta-rc4 as I'm seeing on a good number of my channel open attempts on v0.5.2-beta-rc3

halseth:
@EagleTM this should be in the latest RC 👍

PR #2564, opened 2019-01-30
routing: move payment session to separate file

joostjager:
Preparation to keep better overview for upcoming routing related work.

### DIFF
--- a/routing/payment_session.go
+++ b/routing/payment_session.go
@@ -1,0 +1,175 @@
+package routing
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// paymentSession is used during an HTLC routings session to prune the local
+// chain view in response to failures, and also report those failures back to
+// missionControl. The snapshot copied for this session will only ever grow,
+// and will now be pruned after a decay like the main view within mission
+// control. We do this as we want to avoid the case where we continually try a
+// bad edge or route multiple times in a session. This can lead to an infinite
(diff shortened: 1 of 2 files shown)


routing/router.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/coreos/bbolt"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/go-errors/errors"
-	"github.com/lightningnetwork/lightning-onion"
+	sphinx "github.com/lightningnetwork/lightning-onion"
```
halseth: unrelated change? :p

joostjager: It's `goimports` again.

joostjager:
@halseth rebased, review dismissed, ptal

routing/payment_session.go
```diff
@@ -0,0 +1,175 @@
...
+func (p *paymentSession) ReportEdgeFailure(e *edgeLocator) {
+	log.Debugf("Reporting edge %v failure to Mission Control", e)
+
+	// First, we'll add the failed edge to our local prune view snapshot.
+	p.pruneViewSnapshot.edges[*e] = struct{}{}
+
+	// With the edge added, we'll now report back to the global prune view,
+	// with this new piece of information so it can be utilized for new
+	// payment sessions.
+	p.mc.Lock()
```
cfromknecht: Not related to this PR, tho we should abstract this operation behind a method to prevent us from modifying mission control’s internal state directly and externally acquiring its mutexes. Same for the above method

Better yet, mission control itself deserves its own interface so that access to internal variables is prevented entirely

joostjager: Yes, the design of mission control definitely needs to be stepped up.

PR #2568, opened 2019-01-31
Makefile: lint all travis builds

cfromknecht:
Ensures we fail faster if the source code doesn't
pass the lint checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,9 +134,12 @@ goveralls: $(GOVERALLS_BIN)
 	@$(call print, "Sending coverage report.")
 	$(GOVERALLS_BIN) -coverprofile=coverage.txt -service=travis-ci
 
-travis-race: btcd unit-race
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -134,9 +134,12 @@ goveralls: $(GOVERALLS_BIN)
...
 	$(GOVERALLS_BIN) -coverprofile=coverage.txt -service=travis-ci
 
-travis-race: btcd unit-race
 
-travis-cover: btcd lint unit-cover goveralls
+travis-race: lint btcd unit-race
+
+travis-cover: lint btcd unit-cover goveralls
+
+travis-itest: lint itest
```
halseth: travis.yml must be changed to use this?

cfromknecht: yep you're right

cfromknecht:
@halseth `make travis-test` is now called in `.travis.yaml`, PTAL

PR #2570, opened 2019-02-01
lnwire: ensure even/odd features have the same base name

Roasbeef:
In this commit, we fix a bug in the way we defined our even/odd features
for a particular feature. The check for if a feature bit is part of a
pair assumes that the pair bit has the exact same name as the bit being
queried. The way we defined our feature map didn't take note of this
assumption, as a result, any attempts to require a new bit moving from
optional to required would fail since the bit would be found, but the
names differed.


sangaman:
I was getting the issue of seeing `unable to start peer: data loss protection required` messages endlessly without lnd completing starting up and channels becoming active using the changes from #2500. With this branch that issue is gone and things seem to be operating normally.

PR #2593, opened 2019-02-06
fuzz: adding fuzz harnesses for acts 1-3, encryption+decryption

Crypt-iQ:
This PR adds fuzzing harnesses for all three acts of the brontide handshake.  Ideally, we would fuzz other things as well.  The harnesses are located in the `fuzz/brontide` directory.  The fuzzers should always fail the brontide MAC checks, barring some sort of bug.  

There are some limitations with go-fuzz compared to something like AFL: go-fuzz has no "persistent mode" where the target program isn't restarted between fuzzing runs.  This means that we can't properly fuzz encryption+decryption thousands of times and trigger the key rotation from occurring.  I have never used it before, but I think something like [smat](https://github.com/mschoch/smat) could be used for persistent fuzz runs.

TODO:
- [x] Log the machine states on panic (requires setting `Curve` parameter to nil) so we can reproduce crashes.
- [x] Fuzz acts 1-3
- [x] Fuzz harness for encryption and decryption of arbitrary data.
- [x] Fuzz encryption of arbitrary data.
- [x] Fuzz decryption of arbitrary data.

### DIFF
--- a/fuzz/brontide/fuzz_utils.go
+++ b/fuzz/brontide/fuzz_utils.go
@@ -1,0 +1,137 @@
+// +build gofuzz
+
+package brontidefuzz
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/davecgh/go-spew/spew"
+	"github.com/lightningnetwork/lnd/brontide"
+)
+
+var (
+	initBytes = []byte{
+		0x81, 0xb6, 0x37, 0xd8, 0xfc, 0xd2, 0xc6, 0xda,
+		0x63, 0x59, 0xe6, 0x96, 0x31, 0x13, 0xa1, 0x17,
+		0xd, 0xe7, 0x95, 0xe4, 0xb7, 0x25, 0xb8, 0x4d,
+		0x1e, 0xb, 0x4c, 0xfd, 0x9e, 0xc5, 0x8c, 0xe9,
+	}
+
+	respBytes = []byte{
+		0xaa, 0xb6, 0x37, 0xd9, 0xfc, 0xd2, 0xc6, 0xda,
+		0x63, 0x59, 0xe6, 0x99, 0x31, 0x13, 0xa1, 0x17,
+		0xd, 0xe7, 0x95, 0xe9, 0xb7, 0x25, 0xb8, 0x4d,
+		0x1e, 0xb, 0x4c, 0xf9, 0x9e, 0xc5, 0x8c, 0xe9,
+	}
+
+	// Returns the initiator's ephemeral private key.
+	initEphemeral = brontide.EphemeralGenerator(func() (*btcec.PrivateKey, error) {
+		e := "121212121212121212121212121212121212121212121212121212" +
+			"1212121212"
+		eBytes, err := hex.DecodeString(e)
+		if err != nil {
+			return nil, err
+		}
+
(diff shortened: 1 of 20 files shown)


Crypt-iQ:
I'm going to make two fuzzers for each target.  The first will use static keys (remote/local pub & ephemeral all static) and the second will use random keys (remote/local pub & ephemeral all random).  We need static keys so that the feedback loop isn't garbled by repeated key changes between execs.  We need random keys because maybe there is some property that the static keys do not have but a random key might have - this approach is "noisier".  

I ran four fuzzing runs with different targets to get an idea for speed.  The last three require going through the acts sequentially.  The first received random act one data, the second decrypted random data, the third checked for successful encryption + decryption, and the fourth encrypted + decrypted in a loop 2100 times to trigger key rotation.  These were run with 8 go-fuzz workers over 6 hours.  ActOne is noticeably faster because it doesn't require the handshake that the others require.  Given that key rotation is so slow, I think I'll leave it out in this PR.  Ideally, the fuzzer would give us (or we could derive) a "command sequence" consisting of key rotations, encryption calls, decryption calls and we would perform them but that's like the holy grail of fuzzing.

|                         |       ActOne     |Decrypt        | Enc + Dec           | Enc + Dec Loop  |
|---------------|:-------------:|:------------:|:---------------:|:------------------:|
| **execs/sec** |    11028/sec     |   2402/sec   |       2000/sec |         634/sec            |

Crypt-iQ:
Should be done now.  I have more ideas for fuzzing the brontide state machines like switching on gofuzz input and doing a command sequence (fail encryption+decryption, success encryption+decryption, decryption, encryption, rotate key) and seeing if that turns up anything.  The diff is a bit big considering these are just fuzzing tests so I could cut them down, but I wanted the tests to be as thorough as possible.  I included a corpus message for each harness even though it's pretty useless if there's only one -  I didn't really know where to put a corpus??? Could be useful to check it in this project if we want to run regression tests or something against the fuzz tests 🤷‍♂️

Crypt-iQ:
Should mention in `fuzz.md` (contained in #1895) that the brontide fuzzers need the `-timeout` flag set to 20 seconds or more since logging the machine state can take a while.

fuzz/brontide/fuzz_utils.go
```diff
@@ -0,0 +1,137 @@
...
+	return initiator, responder
+}
+
+// GetStaticBrontideMachines returns two brontide machines that use static keys
+// everywhere.
+func GetStaticBrontideMachines() (*brontide.Machine, *brontide.Machine) {
+	initPriv, _ := btcec.PrivKeyFromBytes(btcec.S256(), initBytes)
+	respPriv, respPub := btcec.PrivKeyFromBytes(btcec.S256(), respBytes)
+
+	initiator := brontide.NewBrontideMachine(true, initPriv, respPub,
```
Roasbeef: Style nit, line folding should resemble something like: 
```go
brontide.NewBrontideMachine(
    true, initPriv, respPub, initEphemeral,
)
```

Crypt-iQ:
Running:

1. `cd fuzz/brontide`

2. `find * -maxdepth 1 -regex '[A-Za-z0-9\-_.]'* -not -name fuzz_utils.go | sed 's/\.go$//1' | xargs -I % sh -c 'go-fuzz-build -func Fuzz_% -o brontide-%-fuzz.zip github.com/lightningnetwork/lnd/fuzz/brontide'`

3. `go-fuzz -bin <harness> -dumpcover -workdir <harness> -procs <num workers>`

PR #2600, opened 2019-02-06
fundingmanager: send out active channel ntfn on FundingLocked

Roasbeef:
In this commit, we fix a small gap in the existing notification logic.
Before this commit, if by the time we added a link to the switch, we
hadn't yet processed the `FundingLocked` message, then we wouldn't mark
the link as fully active. To remedy that, we extend the role of the
`fundingManager` to notify once the channel is "ready".

It's worth noting that with this fix, we can emit duplicate active
channel notifications. However, as this RPC is meant to be used primary
for UIs, I think it's an acceptable tradeoff, we can extend the
`ChannelNotifier` to maintain state to suppress duplicate notifications.

### DIFF
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -337,6 +337,11 @@ type fundingConfig struct {
 	// NotifyOpenChannelEvent informs the ChannelNotifier when channels
 	// transition from pending open to open.
 	NotifyOpenChannelEvent func(wire.OutPoint)
+
+	// NotifyReadyChannelEvent notifies an outside subs-system that a
+	// channel is "ready" for use. A channel only becomes ready once the
+	// FundingLocked message is received and processed.
+	NotifyReadyChannelEvent func(wire.OutPoint)
 }
 
 // fundingManager acts as an orchestrator/bridge between the wallet's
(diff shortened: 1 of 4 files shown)


cfromknecht:
travis failed, looks like we also need to initialize `NotifyReadyChannelEvent` in `test_utils.go`

Roasbeef:
itests fail now since a duplicate update can be sent: 
```
    --- FAIL: TestLightningNetworkDaemon/multiple_channel_creation_and_update_subscription (3.41s)
        lnd_test.go:75: Failed: (multiple channel creation and update subscription): exited with error: 
            *errors.errorString errored verifying close updates: channel update channel of wrong type, expected closed channel, got *lnrpc.ChannelEventUpdate_ActiveChannel
```

fundingmanager.go
```diff
@@ -2429,6 +2435,11 @@ func (f *fundingManager) handleFundingLocked(fmsg *fundingLockedMsg) {
 		return
 	}
 
+	// As this is the earliest possible time that the channel can be used
```
halseth: I'm just wondering.. should we really be adding the link to the switch if we haven't yet received `fundinglocked`?

Roasbeef: We need to in order to be able to process the chan reest message sent by the remote peer upon reconnection. We had cross-impl issues with this in the past until we moved to always loading the channel in the link pre funding locked.

halseth: Riight...

Roasbeef:
Review ping @valentinewallace

fundingmanager.go
```diff
@@ -2429,6 +2435,11 @@ func (f *fundingManager) handleFundingLocked(fmsg *fundingLockedMsg) {
...
+	// to create/receive commitment updates, we'll mark the channel has
+	// active now.
+	f.cfg.NotifyReadyChannelEvent(channel.FundingOutpoint)
```
cfromknecht: reminder that with #3369 this is no longer true, it should be done from inside the link

cfromknecht:
@Roasbeef do we want this for 0.11?

Roasbeef:
@cfromknecht with the latest changes in the switch, I think this should be done there instead as you mentioned above. I'm going to move this from the milestone for now, as we have higher priority items we'd like to land for the upcoming major release.

cfromknecht:
sgtm

PR #2602, opened 2019-02-07
build: add module support for ticker and queue packages

wpaulino:
In this PR, we begin our efforts towards enabling small utility packages that may be useful outside of `lnd` to be used as dependencies within other projects without needing to import all of `lnd`. For now, this is only done for the `ticker` and `queue` packages.

For future reference: https://github.com/go-modules-by-example/index/tree/master/009_submodules


ticker/ticker.go
```diff
@@ -60,10 +60,10 @@ type Ticker interface {
 	Stop()
 }
 
-// ticker is the production implementation of the resumable Ticker interface.
-// This allows various components to toggle their need for tick events, which
-// may vary depending on system load.
-type ticker struct {
+// T is the production implementation of the resumable Ticker interface. This
```
halseth: The name `T` is a bit confusing. What's the rationale behind the change?

go.mod
```diff
@@ -49,3 +51,7 @@ require (
 	gopkg.in/macaroon.v2 v2.0.0
 	gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce // indirect
 )
+
+replace github.com/lightningnetwork/lnd/ticker v1.0.0 => ./ticker
```
halseth: is this a standard way of using submodules?

halseth:
Hm, this also triggers a double Travis build? 😛

ticker/ticker.go
wpaulino: `ticker` should have previously been exported, but renaming it to `Ticker` would conflict with the interface, so @cfromknecht and I settled with `T`.

go.mod
wpaulino: I'll use an example to hopefully get the point across better:

Let's say we wish to rename the `Ticker` interface to something else. Without this replace directive, we would have to commit the changes to the `ticker` package, tag them, and update the version used within the project's root `go.mod` in order to use those changes elsewhere in the project. With the directive in place, we can avoid doing so.

I guess it's all a matter of preference. We can remove it and everything will still work fine, the extra hurdle will exist if  we decide to no include the replace directive.

wpaulino:
The branch was pushed as part of the repo rather than my own fork to avoid further replace directives to my fork.

Roasbeef:
Yeah double travis build is only since it was pushed as a direct branch of lnd in order to be able to do the tag on his end.

wpaulino:
There are some changes that'll need to be made to the `queue` package for the buffer pools, so this should wait until that lands.

ticker/go.mod
```diff
@@ -0,0 +1 @@
+module github.com/lightningnetwork/lnd/ticker
```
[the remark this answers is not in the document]
wpaulino: Yep, no external deps.

go.mod
wpaulino: Because the root project still depends on them. Also, a replace directive for a package can't exist without a require directive for the same package.

ticker/ticker.go
halseth: Why does it have to be exported? The commit message should have an explanation at least IMO.

wpaulino: Fixed.

halseth: Still don't see why exporting it is needed. Having (just) an exported interface nice since it hides the actual implementation details from the user of the module. Are we actually (planning to be) using the fact that the ticker implementation is now exported?

wpaulino: It's just good practice to return exported structs. This doesn't apply to this particular situation, but let's say there's the struct satisfies another interface which you'd like to use. If you return the interface and the struct remains unexported, there's no way of using that same struct to satisfy other interfaces.

> Having (just) an exported interface nice since it hides the actual implementation details from the user of the module.

How does exporting the struct change this? You still interact with it the same way.

cfromknecht: @halseth it also fixes the godocs https://godoc.org/github.com/lightningnetwork/lnd/ticker

PR #2604, opened 2019-02-07
watchtower/wtclient: weight + sweep pkscript fixups

cfromknecht:


watchtower/wtclient/backup_task.go
```diff
@@ -152,8 +158,8 @@ func (t *backupTask) bindSession(session *wtdb.SessionInfo,
 		return err
 	}
 
-	t.outputs = outputs
 	t.blobType = session.Policy.BlobType
+	t.outputs = outputs
```
Roasbeef: Any particular reason for this section of the diff? Just a line swap.

cfromknecht: just thought it was more logical to set blob type first, as the computed outputs depend on its value. but purely a stylistic preference

PR #2606, opened 2019-02-07
watchtower/multi: send connection features + chain hash in Init

cfromknecht:
This PR modifies the `wtwire.Init` message to send only connection-level features and the chain hash of the chain that each client is on. The previous message sent local and global features to mimic the design of the `lnwire.Init` message, though there has been some rough consensus that we will move away from that in favor of an `Init` message of this form.

This change fixes two things:
 - The global features don't need to be advertised at the connection level, as they have no impact on the connection. The move to calling the "connection features" is supposed to be more indicative that the feature bits only impact the behavior of the protocol over the lifetime of the established connection.
 - Adding the chainhash ensures that a client doesn't accidentally back up its state to server on the wrong chain, e.g. litecoin or testnet. This ensures that the tower at a particular address isn't watching the proper chain, or if a server on a different chain takes over the endpoint, the client will discontinue communication with the tower.

### DIFF
--- a/watchtower/wtwire/message.go
+++ b/watchtower/wtwire/message.go
@@ -88,7 +88,7 @@ func makeEmptyMessage(msgType MessageType) (Message, error) {
 
 	switch msgType {
 	case MsgInit:
-		msg = &Init{&lnwire.Init{}}
+		msg = &Init{}
 	case MsgCreateSession:
 		msg = &CreateSession{}
 	case MsgCreateSessionReply:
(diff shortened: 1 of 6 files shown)


watchtower/wtwire/message.go
```diff
@@ -88,7 +88,7 @@ func makeEmptyMessage(msgType MessageType) (Message, error) {
 
 	switch msgType {
 	case MsgInit:
-		msg = &Init{&lnwire.Init{}}
```
Roasbeef: Innit-ception 🧐

PR #2615, opened 2019-02-08
wtwire: add CheckRemoteInit helper

cfromknecht:
This PR is an extension of #2606 and #2605, adding a `CheckRemoteInit` helper method to the `wtwire.Init` message. Once the client is merged, this will checked in three different places with in the code base, so it makes sense to centralize this logic. We also add unit tests to ensure that the combination of features bits proposed via client and server will work together from the perspective of both the client and server.

Depends on:
 - [x] #2605 
 - [x] #2606

### DIFF
--- a/watchtower/config.go
+++ b/watchtower/config.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec"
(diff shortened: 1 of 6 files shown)


Roasbeef:
Needs a rebase! Should be g2g afterwards.

cfromknecht:
@Roasbeef rebased, ptal

PR #2621, opened 2019-02-09
better error message if config file does not exist

LightningHelper:
I assume if one will want to run lnd without a config (which is strange) this push should not be merged


config.go
```diff
@@ -359,6 +359,11 @@ func loadConfig() (*config, error) {
 			)
 		}
 	}
+	
+	// Check if config file exist
+        if _, err := os.Stat(configFilePath); os.IsNotExist(err) {
```
Roasbeef: A config file isn't required.

MrManPew: Ok, what about now? If a user  supplied a config file from the command line and it does not exist than we should raise an exception

config.go
```diff
@@ -359,6 +359,13 @@ func loadConfig() (*config, error) {
 			)
 		}
 	}
+	
+        // If user supplied alternative config, check if it exist
```
cfromknecht, line 363:
```suggestion
      // If user supplied alternative config, check if it exists.
```

config.go
```diff
@@ -359,6 +359,13 @@ func loadConfig() (*config, error) {
...
+	if configFilePath != defaultConfigFile {
+		if _, err := os.Stat(configFilePath); os.IsNotExist(err) {
+			return nil, fmt.Errorf("command line config file does not exist in %s", configFilePath)
```
cfromknecht, line 366: nit: line should be wrapped at 80 chars

MrManPew: fixed

cfromknecht:
@MrManPew the current version fails to compile with this error:
```
Error: ./config.go:500:23: undefined: defaultConfigFile
```

MrManPew:
Maybe now?

MrManPew:
@cfromknecht is that ok? I don't understand this error message:
```
config.go:506: File is not `goimports`-ed (goimports)
```

guggero:
The comment needs to be indented with a tab character, not 4 spaces. You can also locally run `make fmt` (to format) and then `make lint` to check if the linter is happy.

MrManPew:
What now?? I've fixed the lint..

guggero:
Replaced by https://github.com/lightningnetwork/lnd/pull/6080 to speed up review process. Original commit author was preserved, thank you for your contribution.

PR #2625, opened 2019-02-10
Holdinvoice redesign [wip]

joostjager:


lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -0,0 +1,408 @@
...
+		return nil, nil,
+			fmt.Errorf("cannot use all zeroes as a preimage")
+
+	case invoice.Preimage == nil && invoice.Hash == nil:
+		if _, err := rand.Read(paymentPreimage[:]); err != nil {
+			return nil, nil, err
+		}
+		paymentHash = paymentPreimage.Hash()
+
+	case invoice.Preimage == nil && invoice.Hash != nil:
```
cfromknecht: for this case, i think we also want to validate that the given hash isn't SHA256(000...), o/w the invoice is technically settled after accept w/o setting the preimage

htlcswitch/link.go
```diff
@@ -2601,30 +2684,43 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
...
+			hodlHtlcs,
+			hodlHtlc{
+				pd:         pd,
+				obfuscator: obfuscator,
+			},
+		)
+
+		// Subscribe to the hodl manager to get a notification when a
+		// resolution for this htlc is available.
+		l.hodlSubscription.Subscribe(hash)
```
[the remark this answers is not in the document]
cfromknecht: discussed offline, just leaving the note here

htlcswitch/link_test.go
```diff
@@ -5258,3 +5276,200 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
...
+	case h := <-receiver.registry.settleChan:
+		if hash != h {
+			t.Fatal("unexpect invoice settled")
+		}
+	}
+
+	// Unfortunately a slight delay is needed here, because otherwise the
+	// switch hasn't added the htlc to the circuit map.
+	//
+	// TODO: This may be a problem for rpc users too! Do we accept the
```
cfromknecht: can you clarify this race condition a bit? doesn't the link no send the exit hops through the circuit map?

invoices/interfaces.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// InvoiceDatabase is an interface which represents the persistent subsystem
+// which may search, lookup and settle invoices.
+type InvoiceDatabase interface {
```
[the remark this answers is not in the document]
joostjager: It is only a move, the interface already exists with this name. I agree that `invoices.InvoiceDatabase`  and `invoices.InvoiceRegistry` are not the best, but as I already have quite a few peripheral refactorings in this PR I prefer to leave it for later.

lnrpc/invoicesrpc/addinvoice.go
joostjager: ok, nice case. added

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -92,19 +99,38 @@ type AddInvoiceData struct {
...
-	} else {
+		paymentHash = sha256.Sum256(paymentPreimage[:])
+
+	case invoice.Preimage == nil && invoice.Hash != nil:
+		paymentPreimage = channeldb.UnknownPreimage
+		paymentHash = *invoice.Hash
+
+	case invoice.Preimage != nil && invoice.Hash == nil:
 		paymentPreimage = *invoice.Preimage
+		paymentHash = sha256.Sum256(invoice.Preimage[:])
```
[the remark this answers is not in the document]
joostjager: need to get used to the possibility, fixed

channeldb/invoices.go
```diff
@@ -654,6 +654,39 @@ func (d *DB) SettleInvoice(paymentHash [32]byte,
 	return settledInvoice, err
 }
 
+// SetPreimage sets the preimage of a hodl invoice.
+func (d *DB) SetPreimage(preimage lntypes.Preimage) (*Invoice, error) {
+	var updatedInvoice *Invoice
+	hash := preimage.Hash()
+	err := d.Update(func(tx *bbolt.Tx) error {
```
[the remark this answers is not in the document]
joostjager: If you mean setting batches of preimages, yes I think it is premature. Would also involve more changes up the call tree to actually be able to do this I think. Without accepting multiple preimages on the rpc level, we wouldn't be able to get any benefit out of batching?

invoices/invoiceregistry.go
```diff
@@ -480,6 +481,19 @@ func (i *InvoiceRegistry) SettleInvoice(rHash lntypes.Hash,
...
 
+// SetPreimage sets the preimage of a hodl invoice.
+func (i *InvoiceRegistry) SetPreimage(preimage lntypes.Preimage) error {
+
+	i.Lock()
+	defer i.Unlock()
+
+	_, err := i.cdb.SetPreimage(preimage)
+
+	log.Infof("Invoice %v preimage set", preimage.Hash())
```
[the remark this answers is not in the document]

htlcswitch/link_test.go
```diff
@@ -5258,3 +5276,200 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
...
+		n.firstBobChannelLink)
+
+	// Generate hold invoice preimage.
+	var preimage [sha256.Size]byte
+	r, err := generateRandomBytes(sha256.Size)
+	if err != nil {
+		t.Fatal(err)
+	}
+	copy(preimage[:], r)
+	hash := fastsha256.Sum256(preimage[:])
```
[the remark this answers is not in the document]

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -23,10 +23,75 @@ service Invoices {
...
+    A bare-bones invoice for a payment within the Lightning Network.  With the
+    details of the invoice, the sender has all the data necessary to send a
+    payment to the recipient.
+    */
+    string payment_request = 1 [json_name = "payment_request"];
+}
+
+message SettleInvoiceMsg {
+    /// Externally discovered pre-image that should be used to settle the hold invoice.
+    bytes pre_image = 1;
```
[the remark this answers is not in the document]
joostjager: agree

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -0,0 +1,408 @@
...
+			)
+			if amtMSat >= channel.LocalCommitment.RemoteBalance {
+				log.Debugf("Skipping channel %v due to "+
+					"not having enough remote balance",
+					chanPoint)
+				continue
+			}
+
+			// Make sure the channel is active.
+			link, err := cfg.Switch.GetLink(chanPoint)
```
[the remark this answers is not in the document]
joostjager: Do you think it should be done in this PR? I just moved the code, didn' t touch it.

htlcswitch/link_test.go
joostjager: Aha and that is the nice thing about the redesign. This is not needed anymore, because it is safe now!

invoices/hodl_manager.go
```diff
@@ -0,0 +1,175 @@
...
+		return
+	}
+
+	// Notify all interested subscribers and remove
+	// subscription from both maps.
+	for _, subscriber := range subscribers {
+		log.Debugf("HodlManager notify subscriber %v",
+			subscriber.id)
+
+		subscriber.SettleChan <- preimage
```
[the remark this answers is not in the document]
joostjager: good point, added

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -143,6 +148,20 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 			applyPreimage(preimage)
 			return &h.htlcSuccessResolver, nil
 
+		case preimage := <-hodlSubscription.SettleChan:
```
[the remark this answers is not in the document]
joostjager: sharp observation. fixed

lntypes/preimage.go
```diff
@@ -53,3 +53,8 @@ func NewPreimageFromStr(newPreimage string) (*Preimage, error) {
 func (p *Preimage) Hash() Hash {
 	return Hash(sha256.Sum256(p[:]))
 }
+
+// Matches returns whether this preimage is the preimage of the given hash.
+func (p *Preimage) Matches(h Hash) bool {
+	return h == p.Hash()
```
Roasbeef: 👍 nice little utility method

route/hophint.go
```diff
@@ -0,0 +1,36 @@
...
+
+import "github.com/btcsuite/btcd/btcec"
+
+const (
+	// DefaultFinalCLTVDelta is the default value to be used as the final
+	// CLTV delta for a route if one is unspecified.
+	DefaultFinalCLTVDelta = 9
+)
+
+// HopHint is a routing hint that contains the minimum information of a channel
```
Roasbeef: Would moving the hop hint struct into the `zpay32` package also address this?

Roasbeef: Or possible the introduction of a `path` package? So we'd have `path.Find`, etc in there and the objective of the package would be to house less LN specific graph operations. Things like fees and CLTV calculation, etc would still be in the main `routing` package. Already we have a separation between a path and a route, so we wouldn't be forcing any new delineation.

contractcourt/chain_arbitrator.go
```diff
@@ -137,10 +137,9 @@ type ChainArbitratorConfig struct {
 	// Sweeper allows resolvers to sweep their final outputs.
 	Sweeper *sweep.UtxoSweeper
 
-	// SettleInvoice attempts to settle an existing invoice on-chain with
-	// the given payment hash. ErrInvoiceNotFound is returned if an invoice
-	// is not found.
-	SettleInvoice func(lntypes.Hash, lnwire.MilliSatoshi) error
+	// Registry is the invoice database that is used by resolvers to lookup
+	// preimages and settle invoices.
+	Registry *invoices.InvoiceRegistry
```
Roasbeef: Rather than giving them the entire struct, how about introducing a new interface that specifies _only_ what the resolvers need? This will make writing unit tests easier as well, since they don't need to fully instantiate an `InvoiceRegistry`.

lntypes/hash.go
```diff
@@ -19,30 +19,30 @@ func (hash Hash) String() string {
 
 // NewHash returns a new Hash from a byte slice.  An error is returned if
 // the number of bytes passed in is not HashSize.
-func NewHash(newHash []byte) (*Hash, error) {
+func NewHash(newHash []byte) (Hash, error) {
```
Roasbeef: In a few other areas within the codebase, we use a `Make` prefix to indicate that we're returning a concrete type, and a `New` to indicate that a pointer will be returned. Not a blocker here to switch over, but something that we may want to consider standardizing throughout the codebase.

htlcswitch/interfaces.go
```diff
@@ -4,28 +4,9 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnpeer"
-	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
-// InvoiceDatabase is an interface which represents the persistent subsystem
```
Roasbeef: I don't think this move is warranted. As is, it allows the package to specify the requirement son the interface level it needs in order to carry out its duties. Keeping the interface in this package also allows us to avoid an additional package dependency as well.

htlcswitch/link.go
```diff
@@ -2292,228 +2292,20 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
...
-				)
-
-				needUpdate = true
-				continue
-			}
-
-			preimage := invoice.Terms.PaymentPreimage
-			err = l.channel.SettleHTLC(
-				preimage, pd.HtlcIndex, pd.SourceRef, nil, nil,
+			updated, err := l.processExitHop(
```
Roasbeef: Non blocker: let's also do the forwarding hop as well while we're at it?

htlcswitch/link.go
```diff
@@ -2601,23 +2599,37 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
...
-	if err != nil {
-		return false, fmt.Errorf("unable to settle htlc: %v", err)
-	}
-
 	// Notify the invoiceRegistry of the invoices we just settled (with the
 	// amount accepted at settle time) with this latest commitment update.
 	err = l.cfg.Registry.SettleInvoice(invoiceHash, pd.Amount)
+
+	// Reject htlcs for canceled invoices.
+	if err == channeldb.ErrInvoiceAlreadyCanceled {
```
Roasbeef: nice catch 👌

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -0,0 +1,365 @@
...
+
+// AddInvoiceConfig contains dependencies for invoice creation.
+type AddInvoiceConfig struct {
+	// InvoiceRegistry is a central registry of all the outstanding invoices
+	// created by the daemon.
+	InvoiceRegistry *invoices.InvoiceRegistry
+
+	// Switch is used by the invoices subserver to control acceptance and
+	// cancelation of invoices.
+	Switch *htlcswitch.Switch
```
Roasbeef: We only need `HasActiveLink` here not the _entire_ switch. Therefore we can get by with just defining a function closure for use here, or an interface that lets us know if a link is active or not.

Roasbeef: Similar comments apply to the usage of `ChanDB` below.

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -53,70 +53,95 @@ type AddInvoiceConfig struct {
 	ChanDB *channeldb.DB
 }
 
+// AddInvoiceData contains the required data to create a new invoice.
```
Roasbeef: (atm) we already import channeldb (might change though, see comment on earlier commit), can we just use that struct directly here? Either way we need to convert the types _somewhere_

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -92,19 +99,44 @@ type AddInvoiceData struct {
...
-// payment preimage. AddInvoice returns the payment hash and the invoice
-// structure as stored in the database.
+// duplicated invoices are rejected, therefore all invoices *must* have a
+// unique payment preimage.
 func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 	invoice *AddInvoiceData) (*lntypes.Hash, *channeldb.Invoice, error) {
 
-	var paymentPreimage lntypes.Hash
-	if invoice.Preimage == nil {
+	var paymentPreimage lntypes.Preimage
```
Roasbeef: Style nit: can be ground as:
```go
var (
    paymentPreimage lntypes.Preimage
    paymentHash lntypes.Hash
)
```

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -143,6 +148,24 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
 			return &h.htlcSuccessResolver, nil
 
+		case <-hodlSubscription.CancelChan:
+			// No action required, but we need to receive to prevent
+			// the channel from getting jammed.
+
+		case preimage := <-hodlSubscription.SettleChan:
+			// If this isn't our preimage, then we'll continue
+			// onwards. This should normally not happen.
+			if !preimage.Matches(h.payHash) {
```
Roasbeef: Non blocker: can use this above also in the `WitnessUpdates` case statement. Not yet convinced (still getting through the diff) that we actually need this as well though.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -114,6 +114,11 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 		blockEpochs.Cancel()
 	}()
 
+	hodlSubscription := h.HodlManager.Subscribe()
+	defer hodlSubscription.Cancel()
+
+	hodlSubscription.Subscribe(h.payHash)
```
Roasbeef: No error or feedback needed here?

invoices/hodl_manager.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/lightningnetwork/lnd/lntypes"
+)
+
+// HodlManager manages hold invoices.
+type HodlManager struct {
+
```
Roasbeef: Style nit: extra new line.

invoices/hodl_manager.go
```diff
@@ -0,0 +1,191 @@
...
+	lastSubscriberID uint64
+
+	quit chan struct{}
+}
+
+// HodlEventSubscription represents a subscription to hodl invoice events.
+type HodlEventSubscription struct {
+	manager    *HodlManager
+	id         uint64
+	SettleChan chan lntypes.Preimage
```
Roasbeef: Missing `godoc` comments.

invoices/hodl_manager.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/lightningnetwork/lnd/lntypes"
+)
+
+// HodlManager manages hold invoices.
+type HodlManager struct {
+
+	// Registry is a sub-system which responsible for managing the invoices
+	// in thread-safe manner.
+	registry InvoiceDatabase
```
Roasbeef: Circling back from the existing `htlcswitch` comment, I don't think it needs all these methods yet. Also, by using a slimmer interface, we can have the `SetPreimage` method (or w/e it's called) also insert into the pre-image cache.

Roasbeef: Ahh actually nvm, on that second portion. Doing so would start to intermingle locally created/sourced pre-images from those obtained via forwarding. I think this answers my questions in the contract court.

route/hophint.go
joostjager: Moving in the `zpay32` package works. It isn't too strange either. You could say that a hop hint isn't exclusively a property of an invoice though. Good suggestion, I went for this option.

Wouldn't want to pull in more restructuring like the `path` package now.

lntypes/hash.go
joostjager: Renamed to `Make`

htlcswitch/link.go
joostjager: It is a sensible idea, but I don't want to extend the PR further now.

lnrpc/invoicesrpc/addinvoice.go
joostjager: Do you want to do this in this PR? The dependency became more explicit because of the code move, but that is also all that happened. A move. I'd like to try to prevent this PR from expanding even further. Adding the interface can always be done later. It has no design impact at the moment imo.

lnrpc/invoicesrpc/addinvoice.go
joostjager: The goal here is to make a clean object that specifies only data necessary to add the invoice. Not reuse a db struct with many fields that have no meaning during creating time.

lnrpc/invoicesrpc/addinvoice.go
joostjager: fixed

htlcswitch/link_test.go
```diff
@@ -5265,3 +5275,197 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
...
+	case h := <-receiver.registry.settleChan:
+		if hash != h {
+			t.Fatal("unexpect invoice settled")
+		}
+	}
+
+	// Unfortunately a slight delay is needed here, because otherwise the
+	// switch hasn't added the htlc to the circuit map.
+	//
+	// TODO: This may be a problem for rpc users too! Do we accept the
```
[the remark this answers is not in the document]
joostjager: It is not a problem anymore. See https://github.com/lightningnetwork/lnd/pull/2625#discussion_r257420706

Forgot to remove it in this test.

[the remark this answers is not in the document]
joostjager: I think this particular test still may run in the race condition that conner pointed out with settling and subscribing to the hodl manager, but that will be fixed.

htlcswitch/link_test.go
```diff
@@ -5265,3 +5275,197 @@ func TestChannelLinkCanceledInvoice(t *testing.T) {
...
+		t.Fatalf("no payment result expected: %v", err)
+	case <-time.After(time.Second):
+		t.Fatal("timeout")
+	case h := <-receiver.registry.settleChan:
+		if hash != h {
+			t.Fatal("unexpect invoice settled")
+		}
+	}
+
+	err = n.bobServer.hodlManager.Cancel(hash)
```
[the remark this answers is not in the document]
joostjager: It does at this commit. Maybe github commit order?

channeldb/invoices.go
```diff
@@ -1054,7 +1054,10 @@ func cancelInvoice(invoices *bbolt.Bucket, invoiceNum []byte) (
 
 	switch invoice.Terms.State {
 	case ContractSettled:
-		return &invoice, ErrInvoiceAlreadySettled
+		// Only allow cancelation of hodl invoices.
```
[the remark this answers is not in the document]
joostjager: We are able to, but only if the haven't been settled yet. 

Canceling a settled invoice would mean that we'd stop accepting more payments to the same invoice. But probably also interferes with users expectations. 

I think that when we start to actually track the relation (multiple) htlc <-> invoice, we have a better opportunity to fix this.

contractcourt/chain_arbitrator.go
```diff
@@ -141,6 +141,8 @@ type ChainArbitratorConfig struct {
 	// preimages and settle invoices.
 	Registry *invoices.InvoiceRegistry
 
+	HodlManager *invoices.HodlManager
```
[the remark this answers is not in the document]
joostjager: I think you answered this yourself further down? It is to prevent mixing up invoice preimages with other preimages.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -143,6 +148,24 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 			applyPreimage(preimage)
 			return &h.htlcSuccessResolver, nil
 
+		case <-hodlSubscription.CancelChan:
+			// No action required, but we need to receive to prevent
+			// the channel from getting jammed.
+
+		case preimage := <-hodlSubscription.SettleChan:
```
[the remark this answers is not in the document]
joostjager: To prevent mixing up preimages of invoices and forwards. Otherwise I could just as well have used the preimage cache in the link, my initial approach.

contractcourt/htlc_incoming_contest_resolver.go
joostjager: Yes, it is a good suggestion. But I just don't want to touch more than I already do. It is a simple change, but any change incurs a slight risk. I'd rather leave it for the next who actually touches that code.

htlcswitch/link.go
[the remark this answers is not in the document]
joostjager: This issue has been addressed with your second suggestion. The registration is handled inside the invoice registry lock.

contractcourt/chain_arbitrator.go
joostjager: I could do that, but don't see too much benefit at this moment.

htlcswitch/link_test.go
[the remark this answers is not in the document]
joostjager: Which should be solved now.

contractcourt/htlc_incoming_contest_resolver.go
joostjager: Call has been removed and integrated with `NotifyExitHopHtlc`.

invoices/hodl_manager.go
joostjager: hodl manager removed

invoices/hodl_manager.go
joostjager: `HodlEventSubscription` removed

invoices/hodl_manager.go
joostjager: First part not applicable anymore, because no more hodl manager.

htlcswitch/interfaces.go
joostjager: I moved it to be able to use a mock for hodl manager (otherwise a circ. dep would occur). As hodl manager is non existent now, the move could be undone.

invoices/interfaces.go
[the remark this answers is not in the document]
joostjager: Also undid the move, so now I really didn't touch the code in this PR.

channeldb/invoices.go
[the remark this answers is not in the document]
joostjager: Converted to batch.

channeldb/invoices.go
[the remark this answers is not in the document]
joostjager: Commit bodies added

joostjager:
@Roasbeef @cfromknecht 

Processed and replied to your comments.

Closing this PR and continuing in the original hold invoice PR #2022.

Issue #2609, opened 2019-02-07
Autopilot uses the wrong amount for open channel


mcduck76:
The autopilot agent receives the total amount available for automatic channel opening, as well as the number of channels to open by a call to ChannelBudget:
https://github.com/lightningnetwork/lnd/blob/6c78fc574098d05be36e1423c0edd182aa13656c/autopilot/agent.go#L490

It then proceeds to open each channel with this amount. It is never divided by the number of channels. This immediately causes the threshold set (allocation) for the autopilot to be exceeded.

Example:
8 BTC wallet funds.
1 BTC allocated to 3 channels.
option Threshold set to 0.3.
option MaxChannels set to 5.

The ChannelBudget call will return (8+1) * 0.3 - 1 =1.7 available for funding 2 channels, as the current allocation is 1/(8+1) (0.11, less than 0.3)

The agent then opens two new channels with 1.7 BTC funding each. This leads to the current allocation being 4.4/(4.6+4.4)=0.49, which is way more than the 0.3 requested threshold.

The correct behavior would be if the 1.7 available funds were equally divided to the channels to be opened, in this case 1.7/2=0.85.

If that was the case, the new threshold would be 2.7/(6.3 + 2.7)=0.3, which would be correct according to the option allocation set in the configuration.

Note: The amount per channel does not have to be just funds/numofchannels, it could be splitted according to the channel score to be opened (as an extra feature), as long as the total amount opened does not exceed the total funds available.

halseth:
Are you saying you are seeing this overallocation with the current autopilot? 

The maximum network channel size is about 0.16 BTC, so it cannot open channels up to the size you mention.

The current implementation will open channels, one at the time of size maxchansize set in the config. Is this consistent with what you are seeing?

mcduck76:
My example was with crude numbers for easier back-of-the-envelope calculations.

The maxchansize (undocumented in the sample-lnd.conf file) is indeed going to cap the limit to whatever value it is set. This however has nothing to do with the reported issue.

Real example (how I discovered the problem in the first place):

0.48 BTC wallet funds.
0.0022 BTC allocated to 3 channels.
option Threshold set to 0.3.
option MaxChannels set to 5.
maxchansize not set, default 0.16777216.

Calculated budget: (0.48+0.0022)*0.3-0.0022=0.14246
Autopilot opens two new channels of its choice, with 0.14246 amount each (less than the 0.16777216 maximum).
[...]

PR #2633, opened 2019-02-12
[autopilot] distribute available funds among channels

halseth:
This commit fixes a regression in how we allocate funds to attempted
channels. We would earlier stay within the channel size limits, but we
wouldn't account for funds consumed by other channels being opened in
parallel.

We fix this by introducing a loop which greadily tries to distribute the
funds among the channels to open, and reduces the number of channels to
open in case not enough funds are available to satisfy the channel size
limits.

Fixes https://github.com/lightningnetwork/lnd/issues/2609

### DIFF
--- a/autopilot/agent.go
+++ b/autopilot/agent.go
@@ -567,8 +567,13 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 
 	// As channel size we'll use the maximum channel size available.
 	chanSize := a.cfg.Constraints.MaxChanSize()
-	if availableFunds-chanSize < 0 {
+	if availableFunds < chanSize {
 		chanSize = availableFunds
+	}
+
+	if chanSize < a.cfg.Constraints.MinChanSize() {
+		return fmt.Errorf("not enough funds available to open a " +
+			"single channel")
 	}
 
 	// Use the heuristic to calculate a score for each node in the
@@ -599,6 +604,17 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 		// so we'll skip it.
 		if len(addrs) == 0 {
 			continue
+		}
+
+		// Track the available funds we have left.
+		if availableFunds < chanSize {
+			chanSize = availableFunds
+		}
+		availableFunds -= chanSize
+
+		// If we run out of funds, we can break early.
+		if chanSize < a.cfg.Constraints.MinChanSize() {
+			break
 		}
 
 		chanCandidates[nID] = &AttachmentDirective{
@@ -725,8 +741,7 @@ func (a *Agent) executeDirective(directive AttachmentDirective) {
 	// fewer slots were available, and other successful attempts finished
 	// first.
 	a.pendingMtx.Lock()
-	if uint16(len(a.pendingOpens)) >=
-		a.cfg.Constraints.MaxPendingOpens() {
+	if uint16(len(a.pendingOpens)) >= a.cfg.Constraints.MaxPendingOpens() {
 		// Since we've reached our max number of pending opens, we'll
 		// disconnect this peer and exit. However, if we were
 		// previously connected to them, then we'll make sure to
@@ -799,6 +814,9 @@ func (a *Agent) executeDirective(directive AttachmentDirective) {
 
 	// Since the channel open was successful and is currently pending,
 	// we'll trigger the autopilot agent to query for more peers.
+	// TODO(halseth): this triggers a new loop before all the new channels
+	// are added to the pending channels map. Should add before executing
+	// directive in goroutine?
 	a.OnChannelPendingOpen()
 }
 
--- a/autopilot/agent_test.go
+++ b/autopilot/agent_test.go
@@ -58,7 +58,7 @@ func (m *mockConstraints) MaxPendingOpens() uint16 {
 }
 
 func (m *mockConstraints) MinChanSize() btcutil.Amount {
-	return 0
+	return 1e7
 }
 func (m *mockConstraints) MaxChanSize() btcutil.Amount {
 	return 1e8
@@ -85,13 +85,13 @@ func (m *mockHeuristic) Name() string {
 }
 
 func (m *mockHeuristic) NodeScores(g ChannelGraph, chans []Channel,
-	fundsAvailable btcutil.Amount, nodes map[NodeID]struct{}) (
+	chanSize btcutil.Amount, nodes map[NodeID]struct{}) (
 	map[NodeID]*NodeScore, error) {
 
 	if m.nodeScoresArgs != nil {
 		directive := directiveArg{
 			graph: g,
-			amt:   fundsAvailable,
+			amt:   chanSize,
 			chans: chans,
 			nodes: nodes,
 		}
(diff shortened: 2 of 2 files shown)


autopilot/agent.go
```diff
@@ -565,10 +565,29 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
 
-	// As channel size we'll use the maximum channel size available.
-	chanSize := a.cfg.Constraints.MaxChanSize()
-	if availableFunds-chanSize < 0 {
-		chanSize = availableFunds
+	// We'll keep decreasing the number of channel we will open until we
+	// have enough funds available for all of them.
+	var chanSize btcutil.Amount
+	for ; numChans > 0; numChans-- {
+		// We'll start by attempting to distribute the funds evenly
```
Roasbeef: The existing version of this loop would instead (more greedily) attempt to _maximize_ the size of each channel respecting the min and max channel limits. Pre-amp, I think this might be a better approach, as it'll mean they'll be able to send/receive more of their deposited funds in a single go. With the new atpl selection taking into account node availability data on mainnet, it'll also be less likely that they have a single larger channel to a lack luster node.

Roasbeef: I still also think we should _attach the recommended channel size_ to node score. The node score can then accept the constraints and use that to make decision w.r.t channel size. Externally, we can manage things like the total number of channels to open. It's still not clear to me _what_ exactly we'd add to the existing channel constraints to allow it to recommend channel sizes in a more intelligent manner. It seems much more natural that node selection be contingent on how large of channels we want to open, and how much we have available at our disposal.

halseth: Agree on trying to open as large channels as possible 👍 

When getting scores, it already takes the wanted channel size into account (that is, it is part of the API, the current heuristics ignore it). Isn't that what we want? 

Having the heuristic return return a score+size for each node makes it hard to reason about where to open channels, as the agent has two sets of variables to consider.

Roasbeef: > When getting scores, it already takes the wanted channel size into account (that is, it is part of the API, the current heuristics ignore it). Isn't that what we want?

Yeah, that's the current behavior, but then it leaves open the question of how do we determine what channels size we want in the first place? If the scoring is aware of the total budget, then based on that it may actually score nodes differently. A routing node looking to allocate 1 billion SAT within channels may warrant a distinct scoring from a client looking to open up 100 mil SAT worth of channels. Or as an example, if we're examining a node for scoring, and the max flow from/to the node is 1 mil SAT, then we may not want to open a 10 billion sat channel to it, as we'd only be able to ever effectively use a very small % of it. 

I brought this up when the new channel constraints struct was introduced, as it seemed to attempt to side step this fundamental problem rather than take it on head first. In a world of splicing, maybe this becomes less important, as we only care about our min channel size to establish connectivity, then we can slowly pump up the channels over time in a reactive manner. 

> Having the heuristic return return a score+size for each node makes it hard to reason about where to open channels, as the agent has two sets of variables to consider.

True it makes things a bit more involved. The question then is, is it better to actually confront this problem, or delegate things to a more advanced channel constraint system that has yet to be developed or researched? I don't have a clear idea of how that system would even work. Would it consider the level of utilization of the current channels, and attempt to select an ideal channel size from that? Would it aim only open channels that are powers of two? Would it first do a dry run node scoring and then tune the channel size from there on? 

Not saying we need to tackle this issue at this very moment. Just want to ensure we're not forgetting about the other important component aside from who to open channels to: how large should those channels be?

halseth: We can model the "perfect channel size" as a separate heuristic. That way the agent would go through steps 
`1) with my current setup determine a size for my next channel 2) given this size, determine which node to open a channel to.`

An alternative flow could be 
`1) Given my current channels, determine the best node on the network to open a channel to 2) given this node, what is the best channel size`

Modeling these decisions as a series of individual heuristics makes it easier to combine and improve on the various parts. In the current case it would simply be 1) max channel size 2) use bos score list to find best node.

Moving all this logic into the heuristic would totally be possible fwiw, then the API would look more like the previous `Select` API: give me a (set of) channel(s) to open. Behind this the scoring and channel size logic would need to be performed.

If we get these steps right, I imagine we can find the "optimal" channel by (binary/ternary) searching for a size, getting scores for this size, then choosing the size that gives the highest score. I think this will be further down the road though, as at this point I think choosing the channel size independently of the score is plenty good enough.

Roasbeef: Alright, with that all said and done, actionable step here seems to be: make the loop greedily try to maximize channel size as it did before the switch to scoring based heuristics.

halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2633/commits/ac1102ca109c19551a04c01606ccc4e60c3d9945

halseth:
PTAL @Roasbeef

autopilot/agent.go
```diff
@@ -799,6 +817,9 @@ func (a *Agent) executeDirective(directive AttachmentDirective) {
 
 	// Since the channel open was successful and is currently pending,
 	// we'll trigger the autopilot agent to query for more peers.
+	// TODO(halseth): this triggers a new loop before all the new channels
```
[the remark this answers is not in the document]
halseth: Yes, but this happens in a goroutine, and they all get spawned in parallel, so we have no guarantee that they are all added to the map before the first one finishes.

autopilot/agent_test.go
```diff
@@ -1699,3 +1699,390 @@ func TestAgentQuitWhenPendingConns(t *testing.T) {
...
+	select {
+	case constraints.moreChansResps <- moreChansResp{
+		numMore: 3,
+		amt:     channelBudget,
+	}:
+	case <-time.After(time.Second * 10):
+		t.Fatalf("heuristic wasn't queried in time")
+	}
+
+	// Since the maximum channel size is 1 BTC, the agent won't have enough
```
Roasbeef: I find these tests hard to follow at times in general. I think we can reduce some vertical space here with usage of some a helper functions to handle sending/receiving messages to/from the various mocked interfaces. I'd expect the logic to go "ok I have 1.5 BTC and the max size is 1 BTC, so I'll open to channels here for 1 and 0.5 BTC". Instead, as is, it'll instead execute the 1 BTC and the 0.5 BTC during distinct iterations. 

This is a regression compared to the behavior of the prior system, that would attempt to pipeline open as many channels as possible.

halseth: The rationale behind doing it this way was that you don't know if an opening attempt will always succeed, so I would wait until the next iteration to have more information about which ones succeed and potentially open a bigger channel.

Looking at it now though, I see that it won't matter much, since we will eventually end up with the smaller channel anyway, so whether it is being opened last or not makes no difference. Will fix!

Also agree on the verbosity of the tests, will clean they up a bit.

Roasbeef:
Pong.

halseth:
Comments addressed. PTAL @Roasbeef

autopilot/agent.go
```diff
@@ -567,10 +567,15 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 
 	// As channel size we'll use the maximum channel size available.
 	chanSize := a.cfg.Constraints.MaxChanSize()
-	if availableFunds-chanSize < 0 {
+	if availableFunds < chanSize {
 		chanSize = availableFunds
 	}
 
+	if chanSize < a.cfg.Constraints.MinChanSize() {
```
wpaulino: We could move this up to avoid iterating through the whole graph in the event that we don't have enough funds.

autopilot/agent_test.go
```diff
@@ -195,57 +222,95 @@ func TestAgentChannelOpenSignal(t *testing.T) {
...
+	}
 
-	// We'll send an initial "no" response to advance the agent past its
-	// initial check.
+	return ctx, cleanup
+}
+
+// respondMoreChans consumes the moreChanArgs element and responds to the agent
+// with the given moreChansResp.
+func respondMoreChans(t *testing.T, testCtx *testContext, resp moreChansResp) {
```
wpaulino: Nit: can use `t.Helper` here as well as in the rest of the assertion functions.

autopilot/agent_test.go
```diff
@@ -372,77 +377,24 @@ func TestAgentChannelFailureSignal(t *testing.T) {
...
 	}
 
-	select {
-	case heuristic.nodeScoresResps <- map[NodeID]*NodeScore{
-		NewNodeID(node): fakeDirective,
-	}:
-	case <-time.After(time.Second * 10):
-		t.Fatal("heuristic wasn't queried in time")
-	}
+	respondNodeScores(t, testCtx,
```
wpaulino: Nit: should be of the form
```
respondNodeScores(
    ..., ...,
)
```

autopilot/agent_test.go
```diff
@@ -150,10 +150,20 @@ func (m *mockChanController) SpliceOut(chanPoint *wire.OutPoint,
 
 var _ ChannelController = (*mockChanController)(nil)
 
-// TestAgentChannelOpenSignal tests that upon receipt of a chanOpenUpdate, then
-// agent modifies its local state accordingly, and reconsults the heuristic.
-func TestAgentChannelOpenSignal(t *testing.T) {
-	t.Parallel()
+type testContext struct {
```
wpaulino: 💯

autopilot/agent_test.go
```diff
@@ -1091,3 +1091,219 @@ func TestAgentQuitWhenPendingConns(t *testing.T) {
...
+		// All nodes in the graph should be potential channel
+		// candidates.
+		if len(req.nodes) != len(nodeScores) {
+			t.Fatalf("expected %v nodes, instead had %v",
+				len(nodeScores), len(req.nodes))
+		}
+
+		// One channel is already open.
+		if len(req.chans) != int(existingChans) {
+			t.Fatalf("expected 1 existing channel, got %v",
```
wpaulino: Should be expected `existingChans`?

autopilot/agent_test.go
```diff
@@ -1091,3 +1091,219 @@ func TestAgentQuitWhenPendingConns(t *testing.T) {
...
+			Score:  0.5,
+		}
+	}
+
+	// The agent should now query the heuristic in order to determine its
+	// next action as it local state has now been modified.
+	select {
+	case arg := <-testCtx.constraints.moreChanArgs:
+		if len(arg.chans) != 0 {
+			t.Fatalf("expected agent to have 1 channel already "+
```
wpaulino: Should be `expected agent to not have any open channels`?

autopilot/agent.go
halseth: It is already checked before calling `openChans`, this is just an additional sanity check :P

autopilot/agent_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2633/commits/8d148e78a7fcf865789724f573d97aa91ea1b537

autopilot/agent_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2633/commits/28f79b54069838fdabf89f069214cd203ed6c4bc

autopilot/agent_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2633/commits/c201211ed2772ecb6946e224177eed0812697405

autopilot/agent_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2633/commits/30bf23d9256a5d2dec7fb980b23bf9434c3196e0

halseth:
Comments addressed and rebase PTAL @wpaulino @Roasbeef

autopilot/agent_test.go
```diff
@@ -195,57 +222,95 @@ func TestAgentChannelOpenSignal(t *testing.T) {
...
+	select {
+	case testCtx.constraints.moreChansResps <- resp:
+	case <-time.After(time.Second * 10):
+		t.Fatalf("response wasn't sent in time")
+	}
+}
+
+// respondMoreChans consumes the nodeScoresArgs element and responds to the
+// agent with the given node scores.
+func respondNodeScores(t *testing.T, testCtx *testContext,
```
Roasbeef: 👍

halseth:
Rebased.

PR #2640, opened 2019-02-13
routing: add cltv limit

joostjager:
This PR adds a maximum cltv limit for payments, similar to the currently existing fee limit.


routing/pathfind_test.go
```diff
@@ -2010,3 +2010,99 @@ func TestRestrictOutgoingChannel(t *testing.T) {
...
+	t.Run("force high cost", func(t *testing.T) { testCltvLimit(t, 80, 2) })
+}
+
+func testCltvLimit(t *testing.T, limit uint32, expectedChannel uint64) {
+	t.Parallel()
+
+	// Set up a test graph with two possible paths to the target. The path
+	// through channel 2 is the highest cost path, but has the lowest cltv
+	// delta.
+	testChannels := []*testChannel{
```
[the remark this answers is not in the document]
joostjager: Would that test the cltv limit? If I understand you correctly, that would test another part of path finding: the fee function and the trade off between fees and time lock?

routing/pathfind_test.go
```diff
@@ -2010,3 +2010,99 @@ func TestRestrictOutgoingChannel(t *testing.T) {
...
+			ignoredNodes: ignoredVertexes,
+			ignoredEdges: ignoredEdges,
+			feeLimit:     noFeeLimit,
+			cltvLimit:    cltvLimit,
+		},
+		sourceNode, target, paymentAmt,
+	)
+	if expectedChannel == 0 {
+		// Finish test if we expect no route.
+		if err != nil {
```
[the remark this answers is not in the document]
joostjager: Isn't `!= nil` a stronger assertion than `== ErrNoPathFound`?

lnrpc/rpc.proto
```diff
@@ -780,7 +780,14 @@ message SendRequest {
     any channel may be used.
     */
     uint64 outgoing_chan_id = 9;
+
+    /** 
+    An optional maximum total time lock for the route. If zero, there is no
+    maximum enforced.
+    */
+    uint32 cltv_limit = 10;
```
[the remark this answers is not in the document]
joostjager: I didn't add it into `QueryRoutes` just yet, because I'd first like to get rid of the k-shortest algorithm there. Otherwise I need to spend time now to work the cltv limit into the partial paths of k-shortest, while in the end it will be removed completely.

https://github.com/lightningnetwork/lnd/pull/2497

joostjager:
> Somewhat unrelated -- I noticed that we check for the fee limit when constructing the route from the path. Is there a reason that check is still needed since we already do it within path finding?

I don't think there is. It is a sanity check, but I think it can be removed. I didn't bother adding another one for `cltvLimit`.

joostjager:
@wpaulino PTAL

routing/pathfind_test.go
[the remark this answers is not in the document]
wpaulino: `err != nil` means that anything could've gone wrong with path finding, while `err == ErrNoPathFound` is what we expect to happen, so it is a stricter assertion.

routing/pathfind_test.go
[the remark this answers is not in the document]
wpaulino: > Would that test the cltv limit?

Correct.

> If I understand you correctly, that would test another part of path finding: the fee function and the trade off between fees and time lock?

I would say this would test the tradeoff between path length and CLTV limit. Two paths (one longer than the other) have the same fee cost, but the shorter path has a CLTV higher than the limit, so the longer path should be chosen.

wpaulino:
> I don't think there is. It is a sanity check, but I think it can be removed. I didn't bother adding another one for `cltvLimit`.

Gotcha, would be nice to remove it in a follow-up PR then.

routing/pathfind_test.go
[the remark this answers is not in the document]
joostjager: Okay, looking from the inverted side. You are right ofc.

routing/pathfind_test.go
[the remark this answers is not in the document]
joostjager: Got it. Updated.

joostjager:
> > I don't think there is. It is a sanity check, but I think it can be removed. I didn't bother adding another one for `cltvLimit`.
> 
> Gotcha, would be nice to remove it in a follow-up PR then.

Yes, I can do that.

joostjager:
@wpaulino ptal

routing/router.go
```diff
@@ -1523,6 +1523,10 @@ type LightningPayment struct {
 	// if there isn't a route with lower fees than this limit.
 	FeeLimit lnwire.MilliSatoshi
 
+	// CltvLimit is the maximum time lock that is allowed for attempts to
+	// complete this payment.
+	CltvLimit *uint32
```
halseth: I encountered this pattern for the `outgoingChannelID *uint64` earlier. Could we instead let a zero-value indicating it is unset? It really makes serialization easier :)

(I changed this for `outgoingChannelID`  in my `router reliable payments` branch, as I need to serialize the payment attempts.)

routing/pathfind.go
```diff
@@ -624,14 +624,11 @@ func findPath(g *graphParams, r *restrictParams,
...
 		}
 
-		// If the edge has no time lock delta, the payment will always
-		// fail, so return.
-		//
-		// TODO(joostjager): Is this really true? Can't it be that
-		// nodes take this risk in exchange for a extraordinary high
-		// fee?
+		// Every edge should have a positive time lock delta. If we
+		// encounter a zero delta, log a warning line.
```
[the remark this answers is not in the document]
joostjager: I don't know all causes of the delta being zero. Don't even know if it happens at all. But I think there should be no check here and not even a log line. We just need to trust the systems that put the value in place. If we start validating data everywhere we use it, it will be never ending.

My idea was to put the log line here for now and see if that warnings comes up.

routing/router.go
joostjager: I am generally wary of magic values like 0. I use the pointer type be explicit about whether the option applies or not. I prevents people from forgetting to handle the 'not set' case. A channel id of 0 and a cltv limit of 0 are values that are valid. It is just that they will probably never be used in practice.

joostjager: For example with the fee limit it became confusing. A fee limit of zero means no limit. But a zero fee limit could also mean that we may only pay to a direct connection (fee zero).

joostjager:
@halseth  comments replied to, ptal

joostjager:
Rebased

joostjager:
> > I don't think there is. It is a sanity check, but I think it can be removed. I didn't bother adding another one for `cltvLimit`.
> 
> Gotcha, would be nice to remove it in a follow-up PR then.

It is happening in #2497

wpaulino:
Needs a rebase.

routing/router.go
halseth: Yep, it's a tradeoff. I definitely see the advantages of a value explicitly _not_ being set. I'm okay with this approach for now, as the serialization I mentioned earlier has been deferred.

routing/pathfind.go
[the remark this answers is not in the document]
halseth: The spec does allow a zero value it seems. I guess there might even exists cases where it would make sense with a 0 timelock delta?

Anyway, I agree that this is a good solution for now :) (since intermediate nodes having a 0 delta shouldn't negatively affect us)

rpcserver.go
```diff
@@ -2836,6 +2837,11 @@ func extractPaymentIntent(rpcPayReq *rpcPaymentRequest) (rpcPaymentIntent, error
 		payIntent.outgoingChannelID = &rpcPayReq.OutgoingChanId
 	}
 
+	// Take cltv limit from request if set.
+	if rpcPayReq.CltvLimit != 0 {
```
halseth: This check wouldn't be needed if we used 0 to indicate no limit.

cmd/lncli/commands.go
```diff
@@ -2267,6 +2279,11 @@ func payInvoice(ctx *cli.Context) error {
 		FeeLimit:       feeLimit,
 		OutgoingChanId: ctx.Uint64("outgoing_chan_id"),
 	}
+
+	if ctx.IsSet(cltvLimitFlag.Name) {
+		req.CltvLimit = uint32(ctx.Int(cltvLimitFlag.Name))
```
halseth: won't always assigning this value just default it to 0 if not present?

routing/pathfind.go
```diff
@@ -415,6 +415,10 @@ type RestrictParams struct {
 	// OutgoingChannelID is the channel that needs to be taken to the first
 	// hop. If nil, any channel may be used.
 	OutgoingChannelID *uint64
+
+	// CltvLimit is the maximum time lock of the route excluding the final
+	// ctlv.
```
halseth: Add a sentence explaining how the final cltv will affect the final route?

rpcserver.go
joostjager: It is the same discussion as above. The conversion is done as early as possible to prevent having to deal with the magic value further down the call stack.

cmd/lncli/commands.go
joostjager: Yes you are right. Simplified.

routing/pathfind.go
joostjager: Added

joostjager:
ptal

PR #2649, opened 2019-02-17
Dockerfile now builds on local state and fix docker documentation.

jlamur:
The Dockerfile was building the image from the master branch (ignoring
the local state) which is not expected behaviour. It is now building
from the local state.

The order of the Dockerfile instructions is now more optimized for caching,
and a .dockerignore file was added on top of that.

As a security measure, the lnd process now runs as the lnd user by
default. Since lnd is the default user of the image, the same goes lncli
and other executed commands.

The docker documentation no longer mention the legacy lightninglabs/lnd
dockerhub image that is a 9 months old outdated image of lnd.

In the future, one should find a way to get the version of lnd without
using git inside the makefile, such that it's possible to add the
.git folder to the .dockerignore list.


Dockerfile
```diff
@@ -4,34 +4,35 @@ FROM golang:alpine as builder
...
-
 # Expose lnd ports (p2p, rpc).
 EXPOSE 9735 10009
 
 # Specify the start command and entrypoint as the lnd daemon.
 ENTRYPOINT ["lnd"]
 CMD ["lnd"]
+
+# Add a user and use it by default.
+RUN adduser -D lnd
```
yancyribbens: Why add a user here?  Since this runs in a container and system access is provisioned by `chroot`, I don't know why adding this additional user provides better security.. could this be done as a separate PR with a better explanation if it's fixing a possible exploitation path?

Dockerfile
```diff
@@ -4,34 +4,35 @@ FROM golang:alpine as builder
 # queries required to connect to linked containers succeed.
 ENV GODEBUG netdns=cgo
 
-# Install dependencies and build the binaries.
+# Copy in the local repository to build from.
+COPY . /go/src/github.com/lightningnetwork/lnd
```
yancyribbens: There's a `Dockerfile` here https://github.com/lightningnetwork/lnd/blob/master/docker/lnd/Dockerfile which builds from the local repo.  Is there a reason you prefer to add this change to this `Dockerfile`?

.dockerignore
```diff
@@ -0,0 +1,24 @@
+.github
```
yancyribbens: I like the idea of using a dockerignore file to possibly reduce the image size 👍

Dockerfile
```diff
@@ -4,34 +4,35 @@ FROM golang:alpine as builder
...
 # Start a new, final image.
 FROM alpine as final
 
 # Define a root volume for data persistence.
 VOLUME /root/.lnd
 
-# Add bash and ca-certs, for quality of life and SSL-related reasons.
-RUN apk --no-cache add \
-    bash \
-    ca-certificates
```
yancyribbens: Why remove `ca-certificates` here?

Dockerfile
```diff
@@ -4,34 +4,35 @@ FROM golang:alpine as builder
...
 
 # Start a new, final image.
 FROM alpine as final
 
 # Define a root volume for data persistence.
 VOLUME /root/.lnd
 
-# Add bash and ca-certs, for quality of life and SSL-related reasons.
-RUN apk --no-cache add \
-    bash \
```
yancyribbens: I agree bash isn't actually needed to use the image, although as the comment says, it does improve life quality, especially when someone needs to debug the container :)

docs/DOCKER.md
```diff
@@ -7,19 +7,26 @@ check out the documentation at [docker/README.md](../docker/README.md).
...
-To use Docker in a production environment, you can run `lnd` by first creating
-a Docker container, adding the appropriate command-line options as parameters.
+To use Docker in a production environment, you can run `lnd` by creating a
+Docker container, adding the appropriate command-line options as parameters.
+
+You first need to build the lnd docker image:
 
 ```
-$ docker create --name=lnd lightninglabs/lnd [command-line options]
+$ git checkout <tag-or-branch>
```
yancyribbens: I created a PR to allow the docker image to be built with a tag or commit a while back: https://github.com/lightningnetwork/lnd/pull/2520/files.  Is there a reason you prefer a different PR and approach?

.dockerignore
jlamur: Actually that doesn't reduce the image size since the local directory is copied (with the `COPY` statement) only in the build stage.

jlamur: However that helps caching and preventing useless re-build of images. It's also considered a good practice.

Dockerfile
jlamur: I will move this change into another commit, and if requested by others, split it to another PR.

Regarding the purpose / usefulness of the change itself, [the docker best practices documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) says that:
> If a service can run without privileges, use `USER` to change to a non-root user. 

and that makes sense, to me, as it follows the principle of the least privilege.

Dockerfile
jlamur: The idea is to get rid of the dockerfile you mentionned. Once you have a dockerfile for the local state, then you also have a dockerfile for master (it's only a matter of checking out the right branch and syncing).

Also having a dockerfile for master is pretty useless, since the master branch should not be used in production.

.dockerignore
yancyribbens: Ah ok I see.  Good to know, thanks!

Dockerfile
jlamur: Because it's useless, AFAIK.

The same goes for bash (although it's debatable since there is [a completion script for bash](https://github.com/lightningnetwork/lnd/blob/master/contrib/lncli.bash-completion) that could be installed in this image).

Dockerfile
yancyribbens: I agree that it does follow principal of least privilege.  Having it in another PR will make the changes easier to test imo.

Dockerfile
jlamur: Well, there is already the `/bin/busybox` shell installed by default on alpine, that's enough if you need to debug the container.

Dockerfile
yancyribbens: Is it useless because of the copy change you made, or just useless in general.  Maybe this could be part of a separate cleanup PR?

docs/DOCKER.md
jlamur: Yes, there is a reason.

With your approach, you need two dockerfiles, one to build from local state, and another one to build for a tag / branch.

As everyone know, programmers are lazy and don't want to maintain two dockerfiles :)

Dockerfile
jlamur: Because there is nothing that relies on CA certificates in this container (again AFAIK).
Anyway, most of the CA certs are already included in the (installed by default) [`ca-certificates-cacert` package](https://pkgs.alpinelinux.org/package/v3.9/main/x86_64/ca-certificates-cacert).

Roasbeef:
Isn't this now the exact same as the other docker file?

Dockerfile
yancyribbens: By making this change you necessitate requiring the user to checkout and clone this repo, while as it stands this could be used as a stand-alone dockerfile.  I agree it's unlikely that someone will want to use master for "production", but instead will probably pass a [tag](https://github.com/lightningnetwork/lnd/pull/2520/files#diff-dad9d3cc0978f8b909f0876f7b120a68R87).

Ideally, we'd like to be able to just pass a flag to this dockerfile to indicate if we want to build locally or download from GitHub.

docs/DOCKER.md
yancyribbens: yes, but what if someone wishes to build from github without checking out this repo?

Dockerfile
yancyribbens: so no way to use `bash` then?  I guess `busybox` support bourne.

jlamur:
@Roasbeef Correct, they both build fron local state.

As I said in [another comment](https://github.com/lightningnetwork/lnd/pull/2649#discussion_r257848674), the idea is to get rid of the other dockerfile.
Here I clean this one, and on another PR I would delete the other.

Dockerfile
jlamur: There are multiple ways to have `bash` if you really want to install it. You could extend this image, write your own dockerfile, or even install it manually within a running container, but I honestly don't think we need `bash` by default. If that's only because `bash` is "better" (or more practical) than `busybox` then we could apply this argument to a lot of other tools. Docker images are supposed to be as lightweight as possible.

> I guess `busybox` support bourne.

No clue what supporting bourne mean, sorry.

Dockerfile
jlamur: > By making this change you necessitate requiring the user to checkout and clone this repo

Yes, that's their responsibility to download the code and verify that they got the good release (signed, up to date, etc…). At the end you still need to download the code, inside the dockerfile or elsewhere.

> Ideally, we'd like to be able to just pass a flag to this dockerfile to indicate if we want to build locally or download from GitHub.

It could be a solution if we can't reach a middle ground. However, it would complexify the dockerfile (again, in my opinion, for no practical improvement).

docs/DOCKER.md
jlamur: Let's continue on [this other comment thread](https://github.com/lightningnetwork/lnd/pull/2649#discussion_r257935874) where we discuss the same points.

Dockerfile
jlamur: Also, if the dockerfile gets updated, a checkout of the tag / commit would get you the "correct" dockerfile for that release.

Dockerfile
jlamur: Ok, I'm gonna split this to another commit and PR.

Dockerfile
yancyribbens: I think this is only needed in "the other" Dockerfile because [certificates](https://github.com/lightningnetwork/lnd/blob/master/docker/docker-compose.yml#L93) are used to secure communication between services.  I'd be in favor of removing as a different PR if it's not being used in this Dockerfile.

yancyribbens:
@jlamur it seems like the docs could be updated to reference the other Dockerfile if you wish to make a change.  Alternatively, we could get rid of the Dockerfile in the root directory I guess if others don't think referencing Github during build is useful.

Dockerfile
jlamur: It seems like you might be right. It's possible to add TLS to some btcd RPC api for example. I've got no idea how and if it would work with trusted certificatition authority --- ie. will they issue a certificate for such a service? is lnd going to check that certificates against the installed ca certificates?

I'll remove that change for now and double check if that's a needed / useful package. If not, I will create a new PR.

jlamur:
@yancyribbens What you propose is to move these changes to the `/docker/lnd/Dockerfile` file and remove the `/Dockerfile` one, right?

I'm not fundamentaly against that, however dockerfiles are almost always placed at the root of a project, that's where a dev would expect to see a dockerfile.
With that said, if the others want to get rid of this root dockerfile, let me know and I'll move my changes.

Dockerfile
yancyribbens: It looks like it's just a self-sighned cert (no certificate authority) such that if both lnd and btcr use the same cert than the authentication is successful:

lnd: https://github.com/lightningnetwork/lnd/blob/master/docker/lnd/start-lnd.sh#L58
btcr cert: https://github.com/lightningnetwork/lnd/blob/master/docker/btcd/start-btcd.sh#L58
btcr key: https://github.com/lightningnetwork/lnd/blob/master/docker/btcd/start-btcd.sh#L59

That package would be needed if `/Dockerfile` was connecting to a node that required it

yancyribbens:
@jlamur the reason there are two `Dockerfiles` is one copys in stuff locally and the other is standalone.  I have left them as two separate Dockerfiles because I was under the impression both serve different use cases.  Again, maybe pass a flag and combine them but the changes required seem not straightforward imo.

I agree usually `docker-compose.yml` or `Dockerfile` is at the root in many projects.

guggero:
Quite a bit has changed since this PR was created. Thank you for getting the ball rolling on the discussion about whether or not to build from local files or not. I've taken some of the work in this PR (with proper attribution of the author) and added my own opinion on what we should do on top: #4790 

Closing as partially replaced by #4790 and partially because we now have stronger opinions on how the docker builds should look like than at the time this PR was created.

PR #2661, opened 2019-02-20
lnwallet+lnrpc: add ability to fetch the last unused wallet addr 

Roasbeef:
In this PR, we add the ability to fetch the last unused wallet address over the RPC interface. This is useful for end-users UIs as they can spam this call at will without having to worry about potentially unbounded "address inflation". 

The current implementation simply adds two new (previously unused) address types. This makes it easy for any existing UIs, as they don't need to switch to a brand new call (or even update their protos), and can instead just pass in the new address type. The alternative was to add a new RPC call all together. I opted to not pursue this approach, as it seemed wasteful to add an entirely new RPC call for what is really a modification on an existing call. Also given that was have soooo many RPC methods, adding a new enum type seemed more conservative.


lnwallet/interface_test.go
```diff
@@ -2165,6 +2165,62 @@ func testChangeOutputSpendConfirmation(r *rpctest.Harness,
...
+		}
+		sendCoins(t, miner, bob, alice, output, feeRate)
+
+		// If we make a new address, then it should be brand new, as
+		// the prior address has been used.
+		addr3, err := alice.LastUnusedAddress(addrType)
+		if err != nil {
+			t.Fatalf("unable to get addr: %v", err)
+		}
+		if addr1.String() == addr3.String() {
```
bjarnemagnussen: Do you purposefully _prevent_ the scope of this unit test to make use of the `WalletController`'s access to `BtcWallet` and thereby the `CurrentAddress` method? Otherwise I would think one could additionally check that `addr3 == alice.CurrentAddress` (or something like that)?

Roasbeef: Yeah, it's designed to only used methods exposed to the `WalletController`.

Roasbeef:
Comments addressed, PTAL @cfromknecht!

PR #2669, opened 2019-02-20
netann+discovery+server: consolidate network announcements to netann pkg

cfromknecht:
Removes a lot of code duplication w/in the gossiper by consolidating all creation/signing of annoucnements into the netann pkg


Roasbeef:
Still relevant?

cfromknecht:
@Roasbeef yes, rebased

netann/common.go
```diff
@@ -0,0 +1,38 @@
...
+
+	switch m := msg.(type) {
+	case *lnwire.ChannelAnnouncement:
+		data, err = m.DataToSign()
+	case *lnwire.ChannelUpdate:
+		data, err = m.DataToSign()
+	case *lnwire.NodeAnnouncement:
+		data, err = m.DataToSign()
+	default:
+		return nil, errors.New("can't sign message " +
```
wpaulino: Nit: this line can be expanded a bit.

server.go
```diff
@@ -2015,34 +2015,15 @@ func (s *server) genNodeAnnouncement(refresh bool,
...
-
-	// Now that the announcement is fully updated, we'll generate a new
-	// signature over the announcement to ensure nodes on the network
-	// accepted the new authenticated announcement.
-	sig, err := netann.SignAnnouncement(
+	// Otherwise, we'll sign a new update after applying all of the passed
+	// modifiers.
+	err := netann.SignNodeAnnouncement(
 		s.nodeSigner, s.identityPriv.PubKey(), s.currentNodeAnn,
+		modifiers...,
```
wpaulino: The timestamp modifier doesn't seem to be included here. Is that intended?

netann/chan_status_manager.go
```diff
@@ -527,7 +527,8 @@ func (m *ChanStatusManager) signAndSendNextUpdate(outpoint wire.OutPoint,
 
 	err = SignChannelUpdate(
 		m.cfg.MessageSigner, m.cfg.OurPubKey, chanUpdate,
-		ChannelUpdateSetDisable(disabled),
+		ChanUpdSetDisable(disabled),
+		ChanUpdSetTimestamp,
```
wpaulino: Nit: this can go on the line above?

discovery/gossiper.go
```diff
@@ -2454,23 +2454,10 @@ func (d *AuthenticatedGossiper) updateChannel(info *channeldb.ChannelEdgeInfo,
...
-		HtlcMaximumMsat: edge.MaxHTLC,
-		BaseFee:         uint32(edge.FeeBaseMSat),
-		FeeRate:         uint32(edge.FeeProportionalMillionths),
-		ExtraOpaqueData: edge.ExtraOpaqueData,
-	}
-
 	// With the update applied, we'll generate a new signature over a
 	// digest of the channel announcement itself.
-	sig, err := netann.SignAnnouncement(
+	chanUpdate := netann.UnsignedChannelUpdateFromEdge(info, edge)
```
wpaulino: We update the channel update timestamp above, can we use the modifier there as well?

halseth: do we need the `unsigned` version? can't we call  `ChannelUpdateFromEdge` here directly, and then reset the sig?

cfromknecht: yeah don't know i ever went down this path lol

server.go
cfromknecht: added

discovery/gossiper.go
cfromknecht: fixed, we also needed to update the timestamp as well

cfromknecht: let me know if the new solution is acceptable. the issue was that the sig is empty and then `ChannelUpdateFromEdge` fails. i've changed `ChannelUpdateFromEdge` to return both the update and error so that the call site can just ignore. I have a slight preference for the having the unsigned version though, that way we don't accidentally ignore errors in the future if the method changes.

cfromknecht:
@halseth @wpaulino itests passing again, i think it had to do with not updating the timestamp on node announcements and them failing to propagate.

discovery/gossiper.go
halseth: Hm, yeah this looks a bit brittle. Maybe a simple unsigned version is preferred.

halseth: Did you decide on anything here? Possibly the signing/sigparsing can me made a modifier as well, such that we can avoid that brittle error handling

cfromknecht: reverted to use `UnsignedChannelUpdateFromEdge` bc i think it's the most explicit

cfromknecht:
@halseth ptal

PR #2674, opened 2019-02-21
Add instructions and docker-compose file for docker LND Development

yancyribbens:
Add `docker-compose` which allows for running an lnd docker development environment.  This should allow the dev team to use Docker as part of the normal day to day development process by allowing changes to be directly accessible inside the running container.

The `docker-compose` file points to `docker/lnd/Dockerfile` which copies local files into the container at build time.   I think this is preferable to `/Dockerfile` which pulls from Github since we want the current state of the dev environment when the container is created.  The compose file also persists two volumes, one for the local development root (for development changes going forward) and the other for the running lnd application files at `/root/.lnd`.

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -1,0 +1,4 @@
+STATE=bitcoin.active
+NETWORK=bitcoin.testnet
+BITCOIN_NODE=neutrino
+NEUTRINO_CONNECT=faucet.lightning.community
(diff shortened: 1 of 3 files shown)


yancyribbens:
was able to restore this one instead: https://github.com/lightningnetwork/lnd/pull/2559

PR #2690, opened 2019-02-23
peer: deprioritize gossip traffic

### LINKED ISSUE #2686: lnd outgoing gossip flood can cause message timeouts?
### Background

I run c-lightning under valgrind, which makes it pretty slow.  c-lightning also asks for *all* the gossip, which I suspect is causing part of the problem.

I tried to fund a channel to 038c72e69dda1966fea4b98f3724fc906a097f84bbf6cd4395cd285085db58b8f1 (which runs lnd, unknown version), and it seems to have not been able to send the funding_signed reply in time (through the flood of gossip?), eventually giving up and reconnecting:

```
2019-02-22T11:03:41.709Z lightningd(1212): lightning_openingd-038c72e69dda1966fea4b98f3724fc906a097f84bbf6cd4395cd285085db58b8f1 chan #7594: peer_out WIRE_OPEN_CHANNEL
2019-02-22T11:03:42.135Z lightningd(1212): lightning_openingd-038c72e69dda1966fea4b98f3724fc906a097f84bbf6cd4395cd285085db58b8f1 chan #7594: peer_in WIRE_ACCEPT_CHANNEL
2019-02-22T11:03:44.092Z lightningd(1212): lightning_openingd-038c72e69dda1966fea4b98f3724fc906a097f84bbf6cd4395cd285085db58b8f1 chan #7594: peer_out WIRE_FUNDING_CREATED
2019-02-22T11:08:45.983Z lightningd(1212): 038c72e69dda1966fea4b98f3724fc906a097f84bbf6cd4395cd285085db58b8f1 chan #7594: Killing openingd: Reconnected
```

As you can see, we get no response for 5 minutes, and lnd reconnects.  It then tries to REESTABLISH but we consider the channel not to have been started.

I fixed this by restarting without valgrind, but this might become an issue for other slow nodes, especially as the network gets bigger...

cfromknecht:
Should fix #2686

### DIFF
--- a/lnpeer/peer.go
+++ b/lnpeer/peer.go
@@ -12,10 +12,17 @@ import (
 // Peer is an interface which represents the remote lightning node inside our
 // system.
 type Peer interface {
-	// SendMessage sends a variadic number of message to remote peer. The
-	// first argument denotes if the method should block until the message
-	// has been sent to the remote peer.
-	SendMessage(sync bool, msg ...lnwire.Message) error
+	// SendMessage sends a variadic number of high-priority message to
+	// remote peer.  The first argument denotes if the method should block
+	// until the messages have been sent to the remote peer or an error is
+	// returned, otherwise it returns immediately after queuing.
+	SendMessage(sync bool, msgs ...lnwire.Message) error
+
+	// SendMessageLazy sends a variadic number of low-priority message to
+	// remote peer. The first argument denotes if the method should block
+	// until the messages have been sent to the remote peer or an error is
+	// returned, otherwise it returns immediately after queueing.
+	SendMessageLazy(sync bool, msgs ...lnwire.Message) error
 
 	// AddNewChannel adds a new channel to the peer. The channel should fail
 	// to be added if the cancel channel is closed.
--- a/peer.go
+++ b/peer.go
@@ -64,8 +64,9 @@ const (
 // a buffered channel which will be sent upon once the write is complete. This
 // buffered channel acts as a semaphore to be used for synchronization purposes.
 type outgoingMsg struct {
-	msg     lnwire.Message
-	errChan chan error // MUST be buffered.
+	priority bool
+	msg      lnwire.Message
+	errChan  chan error // MUST be buffered.
 }
 
 // newChannelMsg packages a channeldb.OpenChannel with a channel that allows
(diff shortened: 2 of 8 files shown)


lnpeer/peer.go
```diff
@@ -12,10 +12,17 @@ import (
...
 type Peer interface {
-	// SendMessage sends a variadic number of message to remote peer. The
-	// first argument denotes if the method should block until the message
-	// has been sent to the remote peer.
-	SendMessage(sync bool, msg ...lnwire.Message) error
+	// SendMessage sends a variadic number of high-priority message to
+	// remote peer.  The first argument denotes if the method should block
+	// until the messages have been sent to the remote peer or an error is
+	// returned, otherwise it returns immediately after queuing.
+	SendMessage(sync bool, msgs ...lnwire.Message) error
```
wpaulino: Any reason we don't just include a bool to denote the lazy message rather than having two separate methods?

peer.go
```diff
@@ -2348,6 +2348,9 @@ func (p *peer) sendMessage(sync, priority bool, msgs ...lnwire.Message) error {
 	// chans is populated for each message if the caller requested a sync
 	// send.
 	var errChans []chan error
+	if sync {
```
wpaulino: :+1:

server.go
```diff
@@ -1890,7 +1890,12 @@ func (s *server) BroadcastMessage(skips map[routing.Vertex]struct{},
 		// Dispatch a go routine to enqueue all messages to this peer.
 		wg.Add(1)
 		s.wg.Add(1)
-		go s.sendPeerMessages(sPeer, msgs, &wg)
+		go func(p lnpeer.Peer) {
+			defer s.wg.Done()
+			defer wg.Done()
+
+			p.SendMessageLazy(false, msgs...)
```
wpaulino: We should note in this method's godoc that the messages will be sent with low priority.

lnpeer/peer.go
cfromknecht: was thinking it'd be nice to have them separate for clarity and to prevent callers from accidentally swapping `sync` and `lazy` bools. certain subsytems may only need access to one or the other as well

server.go
cfromknecht: updated

cfromknecht:
> The only other case within the gossiper where we send messages to peers is in the reliable sender, but I think those can remain as is as there shouldn't be that many per peer.

Indeed, i think we want these to be prioritized also so that we can remove them from the message store as quickly as possible

Roasbeef:
Agreed that we'd want the direct chan ann/sig/update messages to be sent in prio mode.

PR #2691, opened 2019-02-23
cmd/lncli: increase default recovery window to 2.5k

cfromknecht:
Increases the default window from 250 to 25k. Many
users have reported attempting recovery with the
default value only to find an empty wallet. This
change should help ensure that the first recovery
attempt succeeds for the majority of nodes that
have modest load. It might prudent to consider
increasing this value further in the future if
the issue persists or average node age increases.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -33,7 +33,7 @@ import (
 
 // TODO(roasbeef): expose all fee conf targets
 
-const defaultRecoveryWindow int32 = 250
+const defaultRecoveryWindow int32 = 2500
 
 func printJSON(resp interface{}) {
 	b, err := json.Marshal(resp)


molxyz:
tACK! 
I have tried this on testnet and it works better than using the default `look-ahead` number (250) to search for addresses. 
On the slack a few days a user could not get his funds to show up in recovery by using the default `look-ahead`, so I suggested him to increase the number to 25000 and it worked! Here's a screenshot of what he said:

image

githorray:
Let's get this done. DOW 25000!!!!

bruceleeroy18:
Thanks again molxyz!

molxyz:
@Roasbeef  Ok here's what someone said on another thread:
```
I've had a similar issue and have used recovery window up to 5000 to no avail.
Only issues on LND github seem to point to that as a solution, but I'd be very surprised if any Blitz users have got through more addresses than the default recovery window covers.

https://github.com/rootzoll/raspiblitz/issues/278#issuecomment-463230615
```
If you want to reduce this number I don't think less than 10,000 can really help. 

Another issue is some people said they used less than 250 addresses but why this default number did not help them to recover their funds. I'm wondering if btcwallet his some issue that needs to be looked into?

halseth:
Should we limit/warn users when they create an excessive amount of unused addresses?

cfromknecht:
@halseth i think it'd be worth investigating such a mechanism, can also be used to diagnose any address inflation cases

@Roasbeef window reduced to 2.5k, indeed i think scaling up by 100x is a little ambitious for a single go :) If the issue persists we can revisit

PR #2697, opened 2019-02-25
channeldb: cache HasChannelEdge queries

cfromknecht:
Adds a small cache in front of `ChannelGraph.HasChannelEdge` in order to improve the performance of graph validation. This avoids hitting the database to check if an edge policy is stale, as well as a number of places in which `HasChannelEdge` is called as a subroutine to obtain the latest policy timestamps and check edge existence. `HasChannelEdge` is one of the primary CPU bottlenecks that remains on my node after merging the latest buffer pool optimizations. After running this PoC on my node, `HasChannelEdge` no longer appears on the flame graphs.

### DIFF
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -114,6 +114,7 @@ var bufPool = &sync.Pool{
 type DB struct {
 	*bbolt.DB
 	dbPath string
+	graph  *ChannelGraph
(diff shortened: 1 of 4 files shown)


cfromknecht:
Superceded by #2847

Issue #2509, opened 2019-01-18
it should be simple to start setup with `docker-compose up`


yancyribbens:
It should be simple to start lnd using `docker-compose up` and have everything "just work" however some services fail.  we could also split out a seperate `yml` file for use with `litecoin` in case someone wants to choose which type of node they run.  So for example, if someone did `docker-compose up` this would just run the `btc` setup, but if instead they did `docker-compose -f docker-compose.ltc.yml up` then it would run just litecoin or both perhaps.  splitting out a seperate `yml` file would also make testing and development simpler imo.

yancyribbens:
i'd be happy to be assigned this if it seems like a good idea.

darwin:
I put quite some effort into a project which can generate a simnet cluster which works out-of-the-box:
https://github.com/darwin/simverse

When you generate default one via `./sv create` you can look at generated `docker-compose.yml` and related docker files and scripts under `_workspace/default`. This could be some inspiration for implementing refresh of docker support here.

guggero:
Fixed by #2699.

Issue #2508, opened 2019-01-18
Docker LND doesn't persist state across reboots


yancyribbens:
I noticed the docker `lnd` service doesn't persist state across container restarts.  If we added a container this would allow more permanent storage for the container.

jrosco:
wouldn't this be up to the user to handle themselves?

yancyribbens:
@jrosco what do you mean exactly?  The volume can always be deleted if the user wants to remove the stored state.

jrosco:
It sounds to me like you need to add a persistent volume to your build configure. If so, this can easily be added using docker-cli, docker-compose, k8s etc. Although I could be incorrectly understanding you, in which case I can safely be ignored :)

yancyribbens:
@jrosco thanks for the feedback and actually I think you're right.  This should be not on the Dockerfile but instead in docker-compose to be run during run-time :/.   Testing that out now.

yancyribbens:
https://github.com/lightningnetwork/lnd/pull/2533

guggero:
Fixed by #2533.

PR #2699, opened 2019-02-25
docker-compose up

yancyribbens:
https://i.gifer.com/52an.gif

https://github.com/lightningnetwork/lnd/issues/2509

build and start both `lnd` and the `btcd` backend with a single command `docker-compose up`.  In order to enable docker-compose to bring up the services, I removed service definitions that aren't needed for running the environment.  The docs have been updated with new instructions for how to run `btcctl` now that it's not a separate service.

Also, this PR Moves `ltc` to a separate docker-compose file.  A future PR can allow `ltc` to be brought up with a single command and inherit from `docker-compose.yml`.

### DIFF
--- a/docker/docker-compose.ltc.yml
+++ b/docker/docker-compose.ltc.yml
@@ -1,27 +1,27 @@
 version: '2'
 services:
-    # btc is an image of bitcoin node which used as base image for btcd and
-    # btccli. The environment variables default values determined on stage of
+    # ltc is an image of litecoin node which used as base image for ltcd and
+    # ltcctl. The environment variables default values determined on stage of
     # container start within starting script.
-    btcd:
-      image: btcd
-      container_name: btcd
+    ltcd:
+      image: ltcd
+      container_name: ltcd
       build:
-        context: btcd/
+        context: ltcd/
       volumes:
         - shared:/rpc
-        - bitcoin:/data
+        - litecoin:/data
       environment:
         - RPCUSER
         - RPCPASS
         - NETWORK
         - DEBUG
         - MINING_ADDRESS
-      entrypoint: ["./start-btcd.sh"]
+      entrypoint: ["./start-ltcd.sh"]
 
     lnd:
       image: lnd
-      container_name: lnd
+      container_name: lnd_ltc
       build:
         context: ../
         dockerfile: docker/lnd/Dockerfile
@@ -33,23 +33,23 @@ services:
         - DEBUG
       volumes:
         - shared:/rpc
-        - lnd:/root/.lnd
+        - lnd_ltc:/root/.lnd
       entrypoint: ["./start-lnd.sh"]
       links:
-        - "btcd:blockchain"
+        - "ltcd:blockchain"
 
 volumes:
   # shared volume is need to store the btcd rpc certificates and use it within
-  # btcctl and lnd containers.
+  # ltcctl and lnd containers.
   shared:
     driver: local
 
-  # bitcoin volume is needed for maintaining blockchain persistence
-  # during btcd container recreation.
-  bitcoin:
+  # litecoin volume is needed for maintaining blockchain persistence
+  # during ltcd container recreation.
+  litecoin:
     driver: local
 
   # lnd volume is used for persisting lnd application data and chain state
   # during container lifecycle.
-  lnd:
+  lnd_ltc:
     driver: local
--- a/docker/ltcd/start-ltcctl.sh
+++ b/docker/ltcd/start-ltcctl.sh
@@ -52,7 +52,7 @@ PARAMS=$(echo $PARAMS \
     "--rpccert=/rpc/rpc.cert" \
     "--rpcuser=$RPCUSER" \
     "--rpcpass=$RPCPASS" \
-    "--rpcserver=rpcserver" \
+    "--rpcserver=localhost" \
 )
 
 PARAMS="$PARAMS $@"
--- a/docker/README.md
+++ b/docker/README.md
@@ -68,7 +68,7 @@ docker volume create simnet_lnd_alice
 docker volume create simnet_lnd_bob
 
 # Run the "Alice" container and log into it:
-$ docker-compose run -d --name alice --volume simnet_lnd_alice:/root/.lnd lnd_btc
+$ docker-compose run -d --name alice --volume simnet_lnd_alice:/root/.lnd lnd
 $ docker exec -i -t alice bash
 
 # Generate a new backward compatible nested p2sh address for Alice:
@@ -79,10 +79,10 @@ $ MINING_ADDRESS=<alice_address> docker-compose up -d btcd
 
 # Generate 400 blocks (we need at least "100 >=" blocks because of coinbase 
 # block maturity and "300 ~=" in order to activate segwit):
-$ docker-compose run btcctl generate 400
+$ docker exec -it btcd /start-btcctl.sh generate 400
 
 # Check that segwit is active:
-$ docker-compose run btcctl getblockchaininfo | grep -A 1 segwit
+$ docker exec -it btcd /start-btcctl.sh getblockchaininfo | grep -A 1 segwit
 ```
 
 Check `Alice` balance:
@@ -94,7 +94,7 @@ Connect `Bob` node to `Alice` node.
 
 ```bash
 # Run "Bob" node and log into it:
-$ docker-compose run -d --name bob --volume simnet_lnd_bob:/root/.lnd lnd_btc
+$ docker-compose run -d --name bob --volume simnet_lnd_bob:/root/.lnd lnd
 $ docker exec -i -t bob bash
 
 # Get the identity pubkey of "Bob" node:
@@ -163,7 +163,7 @@ Create the `Alice<->Bob` channel.
 alice$ lncli --network=simnet openchannel --node_key=<bob_identity_pubkey> --local_amt=1000000
 
 # Include funding transaction in block thereby opening the channel:
-$ docker-compose run btcctl generate 3
+$ docker exec -it btcd /start-btcctl.sh generate 3
 
 # Check that channel with "Bob" was opened:
 alice$ lncli --network=simnet listchannels
@@ -247,7 +247,7 @@ alice$ lncli --network=simnet listchannels
 alice$ lncli --network=simnet closechannel --funding_txid=<funding_txid> --output_index=<output_index>
 
 # Include close transaction in a block thereby closing the channel:
-$ docker-compose run btcctl generate 3
+$ docker exec -it btcd /start-btcctl.sh generate 3
 
 # Check "Alice" on-chain balance was credited by her settled amount in the channel:
 alice$ lncli --network=simnet walletbalance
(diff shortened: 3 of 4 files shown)


cfromknecht:
@yancyribbens might need a proper rebase?

yancyribbens:
@cfromknecht I thought I did a proper rebase.. any idea why the diff between this branch and master is so... different?

cfromknecht:
@yancyribbens it looks like a merge was applied instead of a rebase, judging by bfbe58b

yancyribbens:
@cfromknecht think it was caused by a pull after rebase but before push.

yancyribbens:
@valentinewallace I removed the `btc` specific contents from `docker-compose.ltc.yml` which reduced the diff a lot.  Litecoin can be run now using `docker-compose -f docker-compse.ltc.yml up`.  Does this work ok to reduce the diff size and keep Litcoin functionality?

yancyribbens:
Successfully tested this on mainnet to make a small transaction.  It took 4-5days to sync mainnet using BTCD on a dedicated machine with 8cores, 16gb of memory and a SSD drive.  I think if this is merged the next thing would be to add the option to use [bitcoind](https://github.com/lightningnetwork/lnd/issues/2827) instead to improve performance.

yancyribbens:
This PR closes issues: https://github.com/lightningnetwork/lnd/issues/2829, https://github.com/lightningnetwork/lnd/issues/2508, https://github.com/lightningnetwork/lnd/issues/2509

guggero:
@yancyribbens: We had some changes to these files recently. Could you please rebase then I'm going to have a look. Thanks!

yancyribbens:
@guggero sure i'll see if I can find some time this week to fix the merge conflicts.

yancyribbens:
@guggero the merge conflicts have been resolved.  sorry about the delay.

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,46 @@
...
+      environment:
+        - RPCUSER
+        - RPCPASS
+        - NETWORK
+        - DEBUG
+        - MINING_ADDRESS
+      entrypoint: ["./start-ltcd.sh"]
+
+    lnd:
+        image: lnd
```
guggero: nit: indentation is off here, should be the same as the elements in `ltcd`.

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,46 @@
...
+          dockerfile: docker/lnd/Dockerfile
+        environment:
+          - RPCUSER
+          - RPCPASS
+          - NETWORK
+          - CHAIN
+          - DEBUG
+        volumes:
+            - shared:/rpc
+        entrypoint: ["./start-lnd.sh"]
```
guggero: We need the `links` part here as well, otherwise lnd won't be able to connect to ltcd. The start script of lnd (`docker/lnd/start-lnd.sh`) connects to the host `blockchain`, which we need to create an alias for with the `links` entry.

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,46 @@
...
+          context: ../
+          dockerfile: docker/lnd/Dockerfile
+        environment:
+          - RPCUSER
+          - RPCPASS
+          - NETWORK
+          - CHAIN
+          - DEBUG
+        volumes:
+            - shared:/rpc
```
guggero: Add the persistent volume for lnd here too:
`          - lnd:/root/.lnd`

docker/docker-compose.yml
```diff
@@ -15,82 +15,27 @@ services:
...
-      links:
-          - "ltcd:blockchain"
-
-    lnd_btc:
-      extends: lnd
-      container_name: lnd_btc
+        - CHAIN
+        - DEBUG
+      volumes:
+        - shared:/rpc
```
guggero: You removed the persistent volume for the lnd data here, which probably happened on rebase.
Please add `          - lnd:/root/.lnd` again.

yancyribbens: good catch :fish:

yancyribbens:
@guggero thanks for the feedback.  I've updated the PR with your suggestions.

guggero:
Needs rebase.

yancyribbens:
@guggero done

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,48 @@
...
+        - RPCUSER
+        - RPCPASS
+        - NETWORK
+        - CHAIN
+        - DEBUG
+      volumes:
+        - shared:/rpc
+        - lnd:/root/.lnd
+      entrypoint: ["./start-lnd.sh"]
+        - "ltcd:blockchain"
```
guggero: `links:` key is missing.

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,48 @@
...
+
+volumes:
```
guggero: the volume for `lnd` is missing.

yancyribbens:
@guggero my bad.  I should have test the changes better.  I've updated with your suggestions and confirmed `docker-compose.ltc.yml` builds and syncs with mainnet:

`sudo NETWORK=mainnet docker-compose -f ./docker-compose.ltc.yml up`
`sudo docker exec -it ltcd /start-ltcctl.sh getblockchaininfo`

Also, if we want the experience to be similar for both, I recommend updating the docker docs to add instructions for using LTC.  https://github.com/lightningnetwork/lnd/issues/3808

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,54 @@
...
+        - RPCPASS
+        - NETWORK
+        - CHAIN
+        - DEBUG
+      volumes:
+        - shared:/rpc
+        - lnd:/root/.lnd
+      entrypoint: ["./start-lnd.sh"]
+      links:
+        - "ltcd:blockchain"
```
guggero: nit: there are a few trailing spaces on this line

yancyribbens: fixed

yancyribbens:
@guggero I squashed the commits and prefixed `docker` to the message.  I think it's preferable to include the readme changes in the same commit, so that if anyone is looking at the commit history can observe that the readme updates are tied together with the docker-compose file changes that go together.  I'm happy to split the readme update to a different commit if you have a strong opinion to do so.

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,54 @@
...
+        - RPCUSER
+        - RPCPASS
+        - NETWORK
+        - DEBUG
+        - MINING_ADDRESS
+      entrypoint: ["./start-ltcd.sh"]
+
+    lnd:
+      image: lnd
+      build:
```
guggero: add `container_name: lnd_ltc`

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,54 @@
...
+    driver: local
+
+  # litecoin volume is needed for maintaining blockchain persistence
+  # during ltcd container recreation.
+  litecoin:
+    driver: local
+
+  # lnd volume is used for persisting lnd application data and chain state
+  # during container lifecycle.
+  lnd:
```
guggero: To avoid collision with the other lnd, name this volume `lnd_ltc` (change also in the `volumes` section above).

docker/ltcd/start-ltcctl.sh
```diff
@@ -52,7 +52,7 @@ PARAMS=$(echo $PARAMS \
     "--rpccert=/rpc/rpc.cert" \
     "--rpcuser=$RPCUSER" \
     "--rpcpass=$RPCPASS" \
-    "--rpcserver=rpcserver" \
+    "--rpcserver=localhost"\
```
guggero: nit: missing space before backslash

yancyribbens:
@guggero agreed that would be ideal.  I think there's more work todo then your comments to make this happen.  For example, [L58](https://github.com/yancyribbens/lnd/blob/docker-compose-up/docker/lnd/start-lnd.sh#L58) of `start-lnd.sh` needs to be changed so that if both `docker-compose.yml` and `docker-compose.ltc.yml` are running they don't both reference `blockchain`.  In addition there are other changes since as renaming the `shared` volume.  Would you be OK with merging this PR since it solves the original issue and then we could create a new issue and PR to run both compose files on the same machine?  that way this won't continue to fall behind the master branch as well if we merge these changes as is.

docker/ltcd/start-ltcctl.sh
yancyribbens: fixed

docker/docker-compose.ltc.yml
yancyribbens: :+1: done

docker/docker-compose.ltc.yml
```diff
@@ -0,0 +1,55 @@
...
+        - RPCUSER
+        - RPCPASS
+        - NETWORK
+        - CHAIN
+        - DEBUG
+      volumes:
+        - shared:/rpc
+        - lnd:/root/.lnd
+      entrypoint: ["./start-lnd.sh"]
+      links:
```
johng: Should we remove links, which is deprecated?

> Links are a legacy option. We recommend using networks instead.

guggero:
I agree that more changes are necessary to get both bitcoin and litecoin running in parallel. In my opinion it's OK to postpone these changes (like switching to networks to get rid of the `blockchain` link/alias) to a separate PR.
But using the same volume for both lnds will lead to problems. So it would be nice if you could rename the ltc one. Then we can get this merged.

yancyribbens:
@guggero the ltc lnd volume name has been updated.

docker/docker-compose.ltc.yml
yancyribbens: @johng removing the legacy issues could take place in a separate PR.  I think there are enough changes in this PR and better to keep the PR simple and address the github issue imo.

carlaKC:
Also 😂 👏 
> https://i.gifer.com/52an.gif

yancyribbens:
@carlaKC thanks for testing out the changes.  If I understand your comment, I _think_ if you run `docker-compose build --no-cache` that should work, even if the images had been previously built.  Although, this should only be a temporary problem.

carlaKC:
> @carlaKC thanks for testing out the changes. If I understand your comment, I _think_ if you run `docker-compose build --no-cache` that should work, even if the images had been previously built. Although, this should only be a temporary problem.

Yeah, I ended up rebuilding the image and it was fine :)
This is a once off problem, but I'm wondering if we should make a comment to warn people that they need to run with `--node-cache` so they don't run into the same issue. Very small thing though, excited to get this in!

guggero:
@carlaKC I think we can just add that warning/hint to the release notes where we describe the changes of this PR. Then we can merge it for `0.9.0-beta-rc2`. What do you think?

carlaKC:
> @carlaKC I think we can just add that warning/hint to the release notes where we describe the changes of this PR. Then we can merge it for `0.9.0-beta-rc2`. What do you think?

Sure, sounds good to me! Happy to merge as is, it's just a small nit.

PR #2704, opened 2019-02-26
config: allow adding multiple tls ips and domains

### LINKED ISSUE #1492: Allow multiple domains being added to SAN via tlsextradomain(s)
### Background

tlsextradomain is only usable to add a single domain to the cert.

### Your environment

* version of `lnd`
https://github.com/lightningnetwork/lnd/commit/b8fecfca71507643e2ba3cc32b65c9de87790597
 4.15.0-1-ARCH (Archlinux)
bitcoind

### Steps to reproduce

Add 2 tlsextradomain entries.

### Expected behaviour

Use all tlsextradomain entries (same probably for tlsextraip I assume)

### Actual behaviour

Only last one will be used.

MDrollette:
Allows setting multiple `--tlsextraip` and `--tlsextradomain` to be used when generating tls certs.

fixes #1492

### DIFF
--- a/config.go
+++ b/config.go
@@ -172,20 +172,20 @@ type torConfig struct {
 type config struct {
 	ShowVersion bool `short:"V" long:"version" description:"Display version information and exit"`
 
-	LndDir         string `long:"lnddir" description:"The base directory that contains lnd's data, logs, configuration file, etc."`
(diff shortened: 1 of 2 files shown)


MDrollette:
This change would still be relevant, I think, since that PR doesn't remove these flags. It just allows for alternatively using Let's Encrypt certs.

Issue #1523, opened 2018-07-08
lnd updatechanpolicy has confusing --help


abrkn:
### Background

`lncli updatechanpolicy` states that the default `--time_lock_delta` is `0`:

```
--time_lock_delta value  the CLTV delta that will be applied to all forwarded HTLCs (default: 0)
```

However:

```
lncli updatechanpolicy --fee_rate 0.001 --base_fee_msat 100000 --time_lock_delta 0```

results in

```
[lncli] rpc error: code = Unknown desc = time lock delta of 0 is too small, minimum supported is 4
```

### Your environment

```
lncli version 0.4.2 commit=a0b2fadea35fa4642daf4e9f56e6ecfe31d22ce1
```

### Steps to reproduce

```
lncli updatechanpolicy --fee_rate 0.001 --base_fee_msat 100000 --time_lock_delta 0```

### Expected behaviour

Unsure

### Actual behaviour

Error. See above.

Roasbeef:
You can't have a value of 0 for the CLTV delta. If you did, then you'd be exposed to losing funds on each HTLC you forward, as a race condition is introduced if you ever need to go to the chain for an outgoing HTLC.

Roasbeef:
That's the default value for the _cli_, not lnd'` default values. You can see the default values for those fields at `lnd -h`:
```
      --bitcoin.minhtlc=                                      The smallest HTLC we are willing to forward on our channels, in millisatoshi (default: 1000)
      --bitcoin.basefee=                                      The base fee in millisatoshi we will charge for forwarding payments on our channels (default: 1000)
      --bitcoin.feerate=                                      The fee rate used when forwarding payments on our channels. The total fee charged is basefee + (amount * feerate / 1000000), where amount is the forwarded amount. (default: 1)
      --bitcoin.timelockdelta=                                The CLTV delta we will subtract from a forwarded HTLC's timelock value (default: 144)
```

We don't display these values there as it would require `lncli` to read the config for `lnd` to see if they user has modified these values.

Roasbeef:
You don't need to set a value, if you aren't actually modifying it.

ctrlbreak-:
Came here to report the following.  Unfortunately, I don't believe you *can* actually omit the argument:

```
lnd@LNDMainnet:~$ lncli updatechanpolicy --base_fee_msat=1999 --fee_rate=0.000042
[lncli] time_lock_delta argument missing
```

[...]

PR #2706, opened 2019-02-26
Close issue #1523 - Sane time_lock_delta default

lundy405:
Set Default to 144


cmd/lncli/commands.go
```diff
@@ -3251,6 +3251,7 @@ var updateChannelPolicyCommand = cli.Command{
 			Name: "time_lock_delta",
 			Usage: "the CLTV delta that will be applied to all " +
 				"forwarded HTLCs",
+			Value: 144,
```
Roasbeef: Would rather have this set as a constant in the `routing` package. Otherwise, we'll need to always change this value in _multiple_ places.

PR #2708, opened 2019-02-26
Add option to set remote channel reservation in open channel

roeierez:
This PR adds the option for a channel initiator to set the “channel_reserve_satoshis”, as defined in Bolt 2 for open_channel message.
Before this change, the field was fixed to 1% of the channel capacity (the recommended value in Bolt 2).
It seems reasonable to allow the channel initiator to change this value if it is willing to take the risk (or responsibility) of watching and monitoring the channel on-chain.
With the ability to dynamically set the channel reserve, nodes will be able to provide the other side better liquidity, enabling them to spend more funds


roeierez:
Do you think it worth scheduling this PR for 0.7.1 release now that the channel reserve fields are in the RPC interface (https://github.com/lightningnetwork/lnd/pull/3238)?

halseth:
Can target for 0.8 possibly.

roeierez:
> Can target for 0.8 possibly.

That would be great!

wpaulino:
Needs a rebase.

roeierez:
Rebased.

Roasbeef:
> With the ability to dynamically set the channel reserve

Likely just a choice of wording here, but isn't dynamic, even with this it would still be fixed. Also this isn't costless, by setting this value to zero or near zero (as I'm guessing you'd like to do), you introduce a costless attack for who ever you open a channel with.

roeierez:
> Likely just a choice of wording here, but isn't dynamic

Agree. Removed the word dynamically.

> you introduce a costless attack for who ever you open a channel with

Right. To me, though, it looks like it is a business decision that shouldn't be limited by the protocol itself. Default value may remain 1%, but it is up to the initiator to change the value based on its business model. For example, Bitrefill offers their users to buy channels. In this case, users already have skin in the game.

fundingmanager.go
```diff
@@ -2846,6 +2847,11 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		minHtlc = f.cfg.DefaultRoutingPolicy.MinHTLC
 	}
 
+	chanReserve := f.cfg.RequiredRemoteChanReserve(capacity, ourDustLimit)
```
cfromknecht: if i'm reading this correctly, we'll only use the user chosen reserve if it is at least the our dust limit. it doesn't seem obvious that it would be ignored if the value is too low. should we check this sooner in the pipeline and return an error?

roeierez: I agree. I have shifted the input validation code to rpcserver.go similar to the way other rpc fields are being validated: https://github.com/lightningnetwork/lnd/pull/2708/files#diff-01ada5680aaa18fc79cbf78b2d55a37bR1366

server.go
```diff
@@ -2982,8 +2982,9 @@ type openChanReq struct {
 
 	chainHash chainhash.Hash
 
-	subtractFees    bool
-	localFundingAmt btcutil.Amount
+	subtractFees      bool
+	localFundingAmt   btcutil.Amount
+	remoteChanReserve btcutil.Amount
```
wpaulino: Nit: move next to `remoteCsvDelay`?

roeierez:
> we should add some tests at the funding manager level with different values (valid and invalid)

Sounds good! I will work on that.

server.go
roeierez: fixed.

roeierez:
@wpaulino I added tests for:
1. custom valid value.
2. non valid value below dust.
3. non valid value over capacity.

For the functionality to be testable I needed to add validation for the channel reserve inside the fundingmanager.go. 
Although I didn't see other parameters validated there, I think it is good to have it there since there are other code flows that use the fundingmanager besides RPC interface.
Also rebased.

halseth:
Moving this to next release.

joostjager:
Marked blocked in anticipation of anchor output commitment format developments. Anchor outputs also have a value that is possibly taken out of a reserve.

halseth:
This is no longer blocked by anchor outputs, and can be rebased.

lnrpc/rpc.proto
```diff
@@ -1854,6 +1854,9 @@ message OpenChannelRequest {
     transaction.
     */
     uint32 max_local_csv = 17;
+
+    // The number of satoshis we require the remote peer to reserve.
+    int64 remote_chan_reserve_sat = 18;
```
wpaulino, line 1859: Nit: make uint64?

rpcserver.go
```diff
@@ -1897,22 +1905,43 @@ func (r *rpcServer) parseOpenChannelReq(in *lnrpc.OpenChannelRequest,
...
 }
 
+// validateChanReserve validates the remote channel reserve was set for this
+// channel in open channel request. This is mainly to ensure that reservation,
+// if set, is not bellow the dust limit and not above the local funding amount.
+func (r *rpcServer) validateChanReserve(chanReserve btcutil.Amount,
+	localFundingAmt btcutil.Amount) error {
+
+	dustLimit := lnwallet.DefaultDustLimit()
+	if chanReserve > 0 && chanReserve < dustLimit {
```
wpaulino, line 1932: Nit: could omit the > 0 check if the proto type is unsigned.

funding/manager.go
```diff
@@ -3259,6 +3259,16 @@ func (f *Manager) handleInitFundingMsg(msg *InitFundingMsg) {
 	if chanReserve == 0 {
 		chanReserve = f.cfg.RequiredRemoteChanReserve(capacity, ourDustLimit)
 	}
+	if chanReserve < ourDustLimit {
```
wpaulino, line 3262: We already have a similar check in `CommitConstraints`, though that doesn't check against the capacity.

funding/manager_test.go
```diff
@@ -3095,6 +3103,50 @@ func TestFundingManagerRejectPush(t *testing.T) {
...
+		}
+
+		alice.fundingMgr.InitFundingWorkflow(initReq)
+
+		// Alice should have sent the OpenChannel message to Bob.
+		select {
+		case <-alice.msgChan:
+			t.Fatalf("init funding workflow should fail with invalid chan"+
+				"reserve %v", chanReserve)
+		case <-initReq.Err:
```
wpaulino, line 3137: Should assert on the expected error here.

carlaKC:
@roeierez still interested in getting this in? I can pick up review once its rebased.

Bosch-0:
Would be great to see this in LND. LDK recently merged in a similar feature: https://github.com/lightningdevkit/rust-lightning/pull/1163 

> by setting this value to zero or near zero (as I'm guessing you'd like to do), you introduce a costless attack for who ever you open a channel with.

@Roasbeef you could also look at this in reverse if the channel opener had big trust issues with their partner they could increase the reserve so more is at stake. This would make sense if the local capacities of both parties differs substantially.

As @roeierez mentioned, this is more of a business feature (particularly for LSPs) but the default can stay at 1% and changes be opt-in only.

ellemouton:
@roeierez, Im gonna close this PR has it has been replaced by [this one](https://github.com/lightningnetwork/lnd/pull/6956). I have kept you as commit author where it applies. Thanks again for the og pr!

PR #2714, opened 2019-02-27
rpcserver: ensure accepted addrs bind to our active net

Roasbeef:
In this commit, we fix an existing bug that would allow an address with
a distinct network encoding from the current active network. With this
commit, we'll no reject for example a testnet address presented to a
mainnet node. Note that there's no real risk in performing this action,
as if the address are actually identical, then we'll send to the same
script.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -770,10 +770,45 @@ func (r *rpcServer) ListUnspent(ctx context.Context,
 	return resp, nil
 }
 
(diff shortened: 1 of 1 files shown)


Roasbeef:
#2715 is an alternative (made after this one).

wpaulino:
Closing in favor of https://github.com/lightningnetwork/lnd/pull/2715.

PR #2718, opened 2019-02-28
provide correct default values in updatechanpolicy help

Goddchen:


cmd/lncli/commands.go
```diff
@@ -3240,17 +3240,20 @@ var updateChannelPolicyCommand = cli.Command{
...
 				"of payment size",
+			Value: 1,
 		},
-		cli.StringFlag{
+		cli.Float64Flag{
 			Name: "fee_rate",
 			Usage: "the fee rate that will be charged " +
 				"proportionally based on the value of each " +
 				"forwarded HTLC, the lowest possible rate is 0.000001",
+			Value: 0.000001,
```
Roasbeef: Would rather have these set as a constant in the `routing` package. Otherwise, we'll need to always change this value in _multiple_ places.

Roasbeef: Also duplicate of #2706. We'll likely merge which ever one implements the feedback first.

Goddchen: I find `applyChannelUpdate(...)` in `router.go` but the actual default values are not used anywhere in the code. So where would those "multiple places" be?

Goddchen: I found some default values at the top of `chainregistry.go`. But I'm not sure those can be used in other packages since they are located in a `main` package :/

Goddchen: @Roasbeef I moved the values to the `lncfg` package and replaced all occurences to use the new defaults.

Goddchen:
@Roasbeef could you help me with this failing test? I'm totally lost :/ Didn't even change anything in this package. Nor can I find the specific test that fails somewhere in the code :/

lncfg/defaults.go
```diff
@@ -0,0 +1,14 @@
+package lncfg
+
+import (
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+const (
+	// DefaultBitcoinTimeLockDelta Default TLD for channel policy updates
```
guggero: Should be a full sentence, for example:
`DefaultBitcoinTimeLockDelta is the default TLD for channel policy updates.`
Applies to the other comments too.

Roasbeef:
Closing due to inactivity.

Issue #2726, opened 2019-03-03
Master has serious connectivity issues.


capacitynetwork:
### Background

Master branch resets connections with reason: unable to write message: write tcp 95.216.16.21:57444->167.86.78.209:9735: write: connection reset by peer.

I run on dedicated server and have log tail all day. I have seen increasing number of nodes with this issue on inbound connections to me the past few day. Now after upgrade to master I experience same issue. 

After about 20 minutes channels all start being disabled  "2019-03-03 16:08:03.032 [INF] NANN: Announcing channel(cde913642cc90d220850f2e850e431e9f1d45c0dcb2e8b7bfe74a499a096331d:1) disabled [detected]
2019-03-03 16:08:03.133 [INF] NANN: Announcing channel(c1bd6a9d416582fe47ce2e09c0b4d78c53105baa9da58ac9edd7fdb6215a0895:1) disabled [detected]
2019-03-03 16:08:03.215 [INF] NANN: Announcing channel(8fb02649faa6b97ded1ce6796b31d4b99d920dc28b2b01bff28f7673d4895745:0) disabled [detected]
2019-03-03 16:08:03.291 [INF] NANN: Announcing channel(30162d9a354c9f8bcd049a2462061cdabd1960e53acfa3606de4b9ffc3e6bafa:1) disabled [detected]"

Master branch is unusable.

### Your environment

* 0.5.2-99-beta commit=v0.5.1-beta-814-g2a652455aaea661b147b6adca0ff51edcd268508
* Ubuntu 16.04.5 LTS
* Bitcoin Core Daemon version v0.17.1.0-gef70f9b52b851c7997a9f1a0834714e3eebc1fd8
* Mainnet

### Steps to reproduce

Upgrade to master

### Expected behaviour

Connections should be maintained.

### Actual behaviour

Connections drop roughly every 90 seconds, repeatedly, until channels disabled. Trimmed log below.

2019-03-03 15:45:10.735 [INF] PEER: Disconnecting 79.62.233.36:9735, reason: unable to start peer: unable to read init msg: read tcp 95.216.16.21:40232->79.62.233.36:9735: read: connection reset by peer
2019-03-03 15:45:12.121 [INF] PEER: Disconnecting 87.78.117.245:9735, reason: unable to write message: write tcp 95.216.16.21:38732->87.78.117.245:9735: write: connection reset by peer
[...]

PR #2732, opened 2019-03-04
Revert "peer: reduce write timeout to 10s"

cfromknecht:
This reverts commit db2c104111c3c12e1651120fe5a51ce03dad8c58.

In #2474, a commit was added to reduce the write timeout from 50s to 5s as the existing timeout is rather long in the context of TCP. An issue arises when the remote peer is unable to validate gossip messages as quickly as we are sending them, causing 1) the remote peer's discovery message stream to backup, and 2) the backpressure to be applied to the sender side after any connection level buffers are also full. The result is that the sender disconnects prematurely, and causes connection loops when large amounts of gossip traffic need to be sent (e.g. on initial sync or if node is offline for a bit).

As a temporary measure, this PR reverts the commit and restores the 50s timeout. As of #2474, this 50s timeout is owned by the write pool instead of the peer, which can lead to less-than-optimal utilization of the available write buffers.

The longer-term solution I have in mind is to handle write timeout errors in the peer, and introduce a backoff procedure to dynamically tune the write load to the remote peer's message throughput. This will allow us to move back to a short write timeout at the connection level, ensuring the write pool maintains good utilization w/ less blocking and prevent the premature disconnections observed in master currently.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -44,8 +44,8 @@ const (
 	// idleTimeout is the duration of inactivity before we time out a peer.
 	idleTimeout = 5 * time.Minute
 
-	// writeMessageTimeout is the timeout used when writing a message to a peer.
-	writeMessageTimeout = 10 * time.Second
+	// writeMessageTimeout is the timeout used when writing a message to peer.
+	writeMessageTimeout = 50 * time.Second
 
 	// readMessageTimeout is the timeout used when reading a message from a
 	// peer.


cfromknecht:
Mitigates #2726

capacitynetwork:
Confirmed on live well connected instance, fixes #2726 for me

PR #2736, opened 2019-03-06
lncfg: add CLI worker configuration

cfromknecht:
This PR adds CLI configuration hooks for worker pools to allow users to set the desired number of works in the daemon-level read, write and sig pools. At the moment, these are chosen as a multiple of `runtime.NumCPU()`, which may provide the pool with too few resources on low end devices. At the same time, nodes that have a large number of peers may an inadequate number of workers even if they have a decent number of CPUs. In general, it is best to decouple these sorts of constants from the runtime, and let higher level applications decide these values as needed.
```
workers:
      --workers.read=     Maximum number of concurrent read pool workers. (default: 16)
      --workers.write=    Maximum number of concurrent write pool workers. (default: 16)
      --workers.sig=      Maximum number of concurrent sig pool workers. (default: 8)
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -252,6 +252,8 @@ type config struct {
 	net tor.Net
 
 	Routing *routing.Conf `group:"routing" namespace:"routing"`
+
+	Workers *lncfg.Workers `group:"workers" namespace:"workers"`
 }
 
 // loadConfig initializes and parses the config using a config file and command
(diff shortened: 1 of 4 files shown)


cfromknecht:
@wpaulino that sounds like a great idea, i can make an issue to track that

Roasbeef:
Will make a follow up PR to add an "auto" setting which once again uses num CPU.

PR #2740, opened 2019-03-07
discovery: introduce gossiper syncManager subsystem

wpaulino:
# Gossip Sync Manager

In this PR, we introduce a new subsystem for the gossiper: the SyncManager. This subsystem is a major overhaul on the way the daemon performs the graph query sync state machine with peers.

Along with this subsystem, we also introduce the concept of an active syncer. An active syncer is simply a GossipSyncer currently operating under an ActiveSync sync type. Before this commit, all GossipSyncer's would act as active syncers, which means that we were receiving new graph updates from all of them. This isn't necessary, as it greatly increases bandwidth usage as the network grows. The SyncManager changes this by requiring a specific number of active syncers. Once we reach this specified number, any future peers will have a GossipSyncer with a PassiveSync sync type.

## Responsibilities
It is responsible for three main things:
1. Choosing different peers randomly to receive graph updates from to ensure we don't only receive them from the same set of peers.
2. Choosing different peers to force a historical sync with to ensure we have as much of the public network as possible.
3. Managing an in-order queue of active syncers where the next cannot be started until the current one has completed its state machine to ensure they don't overlap and request the same set of channels, which significantly reduces bandwidth usage and addresses a number of issues.

# Gossip Syncer Sync Transitions

In order for it to achieve its first two responsibilities, GossipSyncer sync transitions were necessary.
In this PR, we also introduce the ability for them to do so. This allows us to be more flexible with our query graph sync logic, e.g., we can now request a GossipSyncer to force a historical sync of the graph with the remote peer to ensure we can converge on the same graph with them. 

We introduce two different sync types:

1. ActiveSync: gossip syncers operating under this sync type should exercise their default behavior. This includes converging on the set of missing graph updates with the remote peer _and_ receiving new updates from them.

2. PassiveSync: gossip syncers operating under this sync type should not attempt to converge on the set of missing graph updates with the remote peer or receive new updates from them, but it *should* respond to queries from the remote peer.

It's possible to transition from any of these types to any other. Certain transitions require some additional wire messages to be sent, like in the case of an ActiveSync GossipSyncer transitioning to a PassiveSync type.

# Gossip Syncer Synced Signal

In order for the SyncManager to achieve its third responsibility, we introduce another feature to the GossipSyncer in which it can deliver a signal to an external caller once it reaches its terminal chansSynced state.  This is necessary for our new round-robin sync mechanism, as we need to wait for one active syncer to finish syncing before moving on to the next.

# CLI Changes

Since the SyncManager now relies on being told how many active syncers it should strive to maintain, a new CLI flag has been added to allow users to manually tune this to their preference: `--numgraphsyncpeers`. This flag replaces the current `--nochanupdates` and can be used to specify the number of peers the daemon should receive new graph updates from.


discovery/syncer.go
```diff
@@ -472,13 +480,115 @@ func (g *gossipSyncer) channelGraphSyncer() {
...
+					// we're transitioning from an
+					// activeSync, since that's the only
+					// state that allows receiving new graph
+					// updates from the remote peer.
+					if syncType != activeSync {
+						atomic.StoreUint32(
+							&g.syncType,
+							uint32(newSyncType),
+						)
+						continue
```
halseth: should this be `break`?

wpaulino: Sure, `break` also seems to work here.

discovery/gossiper.go
```diff
@@ -1105,15 +1105,19 @@ func (d *AuthenticatedGossiper) InitSyncState(syncPeer lnpeer.Peer,
...
+		channelSeries: d.cfg.ChanSeries,
+		encodingType:  encoding,
+		chunkSize:     encodingTypeToChunkSize[encoding],
 		sendToPeer: func(msgs ...lnwire.Message) error {
 			return syncPeer.SendMessageLazy(false, msgs...)
 		},
 	}, nodeID)
+
+	if !recvUpdates {
+		syncer.syncType = uint32(passiveSync)
```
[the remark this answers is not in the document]
wpaulino: It's already done in commit https://github.com/lightningnetwork/lnd/pull/2740/commits/2968b8f05d8aa24c58eba8b2115a0edc45706d99.

discovery/syncer.go
```diff
@@ -1105,3 +1116,21 @@ func (g *gossipSyncer) SyncState() syncerState {
...
+// a signal for when the gossipSyncer has reached its chansSynced state.
+func (g *gossipSyncer) ResetSyncedSignal() chan struct{} {
+	g.Lock()
+	defer g.Unlock()
+
+	syncedSignal := make(chan struct{})
+
+	syncState := syncerState(atomic.LoadUint32(&g.state))
+	if syncState == chansSynced {
+		close(syncedSignal)
```
cfromknecht: also nil `g.syncedSignal`?

discovery/sync_manager.go
```diff
@@ -0,0 +1,562 @@
...
+		return
+	}
+
+	log.Infof("Removing gossipSyncer for peer=%x", peer)
+
+	// We'll start by stopping the gossipSyncer for the disconnected peer.
+	s.Stop()
+
+	// If it's a non-active syncer, then we can just exit now.
+	if s.SyncType() != activeSync {
```
cfromknecht: might be safe to just check
```golang
if _, ok := m.syncers[s.peerPub]; ok {
    ...
}
```
since the sync type could be mutated in another goroutine and isn't synchronized under this mutex

discovery/sync_manager.go
```diff
@@ -0,0 +1,562 @@
...
+		newSyncType, s.peerPub)
+
+	if newSyncType == activeSync {
+		m.activeSyncers[s.peerPub] = s
+		delete(m.syncers, s.peerPub)
+		go m.enqueueActiveSyncer(s, true)
+		return
+	}
+
+	s.ProcessSyncTransition(newSyncType)
```
cfromknecht: what happens if this block indefinitely? is there a way we can detect and recover?

discovery/gossiper_test.go
```diff
@@ -652,12 +653,16 @@ func createTestCtx(startHeight uint32) (*testCtx, func(), error) {
...
+		Router:                    router,
+		TrickleDelay:              trickleDelay,
+		RetransmitDelay:           retransmitDelay,
+		ProofMatureDelta:          proofMatureDelta,
+		WaitingProofStore:         waitingProofStore,
+		MessageStore:              newMockMessageStore(),
+		RotateTicker:              ticker.NewForce(DefaultSyncerRotationInterval),
+		FullSyncTicker:            ticker.NewForce(DefaultFullSyncInterval),
+		ActiveSyncerTimeoutTicker: ticker.NewForce(DefaultActiveSyncerTimeout),
+		NumActiveSyncers:          3,
```
cfromknecht: would you mind adding a commit to move node pubkey into Config?

discovery/syncer.go
```diff
@@ -438,13 +489,115 @@ func (g *gossipSyncer) channelGraphSyncer() {
 						"query: %v", err)
 				}
 
+			case newSyncType := <-g.syncTransitions:
+				log.Debugf("gossipSyncer(%x): transitioning "+
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,562 @@
...
+func (m *syncManager) PruneSyncState(peer routing.Vertex) {
+	m.syncersMtx.Lock()
+	defer m.syncersMtx.Unlock()
+
+	s, ok := m.gossipSyncer(peer)
+	if !ok {
+		return
+	}
+
+	log.Infof("Removing gossipSyncer for peer=%x", peer)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

config.go
```diff
@@ -245,7 +245,7 @@ type config struct {
 	Color       string `long:"color" description:"The color of the node in hex format (i.e. '#3399FF'). Used to customize node appearance in intelligence services"`
 	MinChanSize int64  `long:"minchansize" description:"The smallest channel size (in satoshis) that we should accept. Incoming channels smaller than this will be rejected"`
 
-	NoChanUpdates bool `long:"nochanupdates" description:"If specified, lnd will not request real-time channel updates from connected peers. This option should be used by routing nodes to save bandwidth."`
```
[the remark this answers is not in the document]
wpaulino: I'd imagine the flag isn't that popular and since this is meant to land in a major release that it should be somewhat fine to do?

discovery/syncer.go
```diff
@@ -446,24 +458,20 @@ func (g *gossipSyncer) channelGraphSyncer() {
...
-
-				g.localUpdateHorizon = &lnwire.GossipTimestampRange{
-					ChainHash:      g.cfg.chainHash,
-					FirstTimestamp: uint32(updateHorizon.Unix()),
-					TimestampRange: math.MaxUint32,
-				}
-				err := g.cfg.sendToPeer(g.localUpdateHorizon)
+
+				err := g.sendGossipTimestampRange(
+					updateHorizon, math.MaxUint32,
```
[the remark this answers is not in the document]
wpaulino: Yep! Fixed.

discovery/syncer.go
```diff
@@ -472,13 +480,115 @@ func (g *gossipSyncer) channelGraphSyncer() {
...
+
+				switch newSyncType {
+				// If an active sync has been requested, then we
+				// should resume receiving new graph updates
+				// from the remote peer.
+				case activeSync:
+					// If the gossip syncer currently has a
+					// passiveSync type, then we'll revert
+					// to our initial syncingChans state to
+					// ensure we can retaliate any channels
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -472,13 +480,115 @@ func (g *gossipSyncer) channelGraphSyncer() {
...
+				// the remote peer's entire graph.
+				case fullSync:
+					atomic.StoreUint32(
+						&g.state, uint32(syncingChans),
+					)
+
+				default:
+					log.Debugf("gossipSyncer(%x): "+
+						"unhandled sync transition %v",
+						g.peerPub, newSyncType)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -472,13 +480,115 @@ func (g *gossipSyncer) channelGraphSyncer() {
...
+					err := g.sendGossipTimestampRange(
+						time.Unix(0, 0), 0,
+					)
+					if err != nil {
+						log.Errorf("Unable to send "+
+							"update horizon to "+
+							"%x: %v", g.peerPub, err)
+					}
+
+					g.localUpdateHorizon = nil
```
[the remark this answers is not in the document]
wpaulino: Probably not, fixed.

discovery/syncer.go
wpaulino: Don't think there's a need to? If it's already set, then it'll be `close`d and `nil`d once reaching a `chansSynced` state.

discovery/sync_manager.go
wpaulino: Good point, fixed!

discovery/sync_manager.go
wpaulino: Fixed.

discovery/gossiper_test.go

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+	newActiveSyncers chan *activeSyncer
+
+	// staleActiveSyncers is a channel through which we'll send any stale
+	// active syncers that should be removed from the round-robin.
+	staleActiveSyncers chan routing.Vertex
+
+	// pendingActiveSyncers is the list of active syncers which are pending
+	// to be started. Syncers will be added to this list through the
+	// newActiveSyncers and staleActiveSyncers channels.
+	pendingActiveSyncers []*activeSyncer
```
Roasbeef: If this is made into a map, then slice tricks aren't required below when removing stale peers from this structure.

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+		// activeSyncer in the queue.
+		case peer := <-m.staleActiveSyncers:
+			if current != nil && peer == current.cfg.peerPub {
+				startNextSyncer()
+				continue
+			}
+
+			for i, activeSyncer := range m.pendingActiveSyncers {
+				if activeSyncer.cfg.peerPub == peer {
+					m.pendingActiveSyncers = append(
```
Roasbeef: See comment above re map usage.

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+						m.pendingActiveSyncers[i+1:]...,
+					)
+					break
+				}
+			}
+
+		// Our current active syncer has reached its terminal
+		// chansSynced state, so we'll proceed to starting the next
+		// pending active syncer if there is one.
+		case <-startNext:
```
Roasbeef: Why do we care if active syncers reach a synced state? They should start in this state more or less, since passive->active is just a matter of setting a new gossip timestamp.

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+	switch len(m.pendingActiveSyncers) {
+	case 0:
+		return nil
+	case 1:
+		s := m.pendingActiveSyncers[0]
+		m.pendingActiveSyncers = nil
+		return s
+	default:
+		s := m.pendingActiveSyncers[0]
+		m.pendingActiveSyncers = m.pendingActiveSyncers[1:]
```
Roasbeef: Should also set `pendingActiveSyncers[0] = nil` to release the dangling reference.

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+	if len(m.syncers) == 0 {
+		log.Debug("No candidates to rotate active syncers")
+		return
+	}
+
+	// Otherwise, we do so we should attempt to rotate as many eligible
+	// active syncers as possible.
+	numRotated := 0
+	blacklist := make(map[routing.Vertex]struct{})
+	for _, s := range m.activeSyncers {
```
Roasbeef: naming suggestion: `s` -> `currentSyncer`

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+			if err == nil {
+				return
+			}
+
+			log.Errorf("Unable to start gossipSyncer(%x): %v",
+				current.cfg.peerPub, err)
+
+			current = m.nextPendingActiveSyncer()
+		}
+
```
Roasbeef: We should log if we get here, as it means that we were unable to start any of the candidate active syncers.

discovery/syncer.go
```diff
@@ -950,7 +950,8 @@ func (g *gossipSyncer) FilterGossipMsgs(msgs ...msgWithSenders) {
 
 	passesFilter := func(timeStamp uint32) bool {
 		t := time.Unix(int64(timeStamp), 0)
-		return t.After(startTime) && t.Before(endTime)
+		return t.Equal(startTime) ||
```
Roasbeef: 👍

Roasbeef:
We''ll also need to test out this new behavior in the wild to ensure that the new operating modes play nicely with CL and eclair.

Roasbeef:
If we expose some more status information from the sync manager, then we can have the server be able to query for the gossip syncer status of each peer. With this information, we can add some annotations to `listpeers` in the form of new fields. Alternatively, the rpc server can query and populate this information itself.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: > Why do we care if active syncers reach a synced state?

This is how the round-robin works to ensure we only have one syncer negotiating at a time.

> They should start in this state more or less, since passive->active is just a matter of setting a new gossip timestamp.

`passiveSync` -> `activeSync` also triggers the state machine to start from `syncingChans`.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+//   reduces bandwidth usage.
+//
+// NOTE: This must be run as a goroutine.
+func (m *syncManager) syncerHandler() {
+	defer m.wg.Done()
+
+	// We'll start all of our relevant tickers, except for our
+	// ActiveSyncerTimeoutTicker, which should only be started when we start
+	// an activeSyncer.
+	m.cfg.RotateTicker.Resume()
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,619 @@
...
+
+const (
+	// DefaultSyncerRotationInterval is the default interval in which we'll
+	// rotate our active syncers for new ones.
+	DefaultSyncerRotationInterval = time.Hour
+
+	// DefaultFullSyncInterval is the default interval in which we'll force
+	// a full sync to ensure we have as much of the public network as
+	// possible.
+	DefaultFullSyncInterval = 24 * time.Hour
```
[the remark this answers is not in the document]
wpaulino: Went with the magic number 6 instead of 24. I've also added a check when `NumActiveSyncers == 0`, which should be the case for light clients, to do a full sync with the first peer they connect to.

discovery/syncer.go
```diff
@@ -973,7 +1023,105 @@ func (g *gossipSyncer) ProcessQueryMsg(msg lnwire.Message, peerQuit <-chan struc
...
+
+	// If a passiveSync transition has been requested, then we should no
+	// longer receive any new updates from the remote peer. We can do this
+	// by setting our update horizon to a range in the past (specifically,
+	// Jan 1 1970), ensuring no graph updates match the timestamp range.
+	case passiveSync:
+		// We'll only send the new update if we're transitioning from an
+		// activeSync, since that's the only state that allows receiving
+		// new graph updates from the remote peer.
+		if syncType != activeSync {
```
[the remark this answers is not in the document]
wpaulino: How does this apply to the case statement above?

discovery/syncer.go
```diff
@@ -973,7 +1023,105 @@ func (g *gossipSyncer) ProcessQueryMsg(msg lnwire.Message, peerQuit <-chan struc
 	}
 }
 
-// SyncerState returns the current syncerState of the target gossipSyncer.
+// ProcessSyncTransition sends a request to the gossip syncer to transition its
+// sync type to a new one.
+func (g *gossipSyncer) ProcessSyncTransition(newSyncType syncerType) error {
+	errChan := make(chan error, 1)
+	select {
+	case g.syncTransitions <- &syncTransitionReq{newSyncType, errChan}:
```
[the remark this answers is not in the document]
wpaulino: It didn't, but fixed.

discovery/syncer.go
```diff
@@ -127,15 +145,23 @@ var (
...
-	// our highest known channel ID.
-	chanRangeQueryBuffer = 144
+	// ErrSyncTransitionTimeout is an error returned when we've timed out
+	// attempting to perform a sync transition.
+	ErrSyncTransitionTimeout = errors.New("timed out attempting to " +
+		"transition sync type")
+
+	// zeroTimestamp is the timestamp we'll use when we want to indicate to
+	// peers that we do not want to receive any new graph updates.
+	zeroTimestamp = time.Unix(0, 0)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -117,6 +115,26 @@ func (s syncerState) String() string {
...
+	DefaultDelayedQueryReplyInterval = 5 * time.Second
+
+	// chanRangeQueryBuffer is the number of blocks back that we'll go when
+	// asking the remote peer for their any channels they know of beyond
+	// our highest known channel ID.
+	chanRangeQueryBuffer = 144
+
+	// syncTransitionTimeout is the default timeout in which we'll wait up
+	// to when attempting to perform a sync transition.
+	syncTransitionTimeout = 5 * time.Second
```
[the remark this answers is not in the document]
wpaulino: I think your comment is referring to the timeout used when starting the next syncer? That one is 5 minutes.

`syncTransitionTimeout` is just a timeout to ensure the `syncManager` doesn't get blocked when attempting to transaction a syncer that is not in a `chansSynced` state.

discovery/syncer.go
```diff
@@ -174,6 +203,12 @@ type gossipSyncer struct {
 	// NOTE: This variable MUST be used atomically.
 	state uint32
 
+	// syncType denotes the syncerType the gossip syncer is currently
+	// exercising.
+	//
+	// NOTE: This variable MUST be used atomically.
+	syncType uint32
```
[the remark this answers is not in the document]
wpaulino: Fixed.

wpaulino:
@Roasbeef comments addressed. I've also expanded the `listpeers` RPC with the current sync type.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -13,6 +13,35 @@ import (
 	"golang.org/x/time/rate"
 )
 
+// SyncerType encapsulates the different types of syncing mechanisms for a
+// gossip syncer.
+type SyncerType uint8
```
halseth: could also make this `uint32`

discovery/gossiper.go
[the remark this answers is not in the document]
halseth: I mean, it is `0 == activeSync` by default, but would be nice to actively set that in the config for readability.

discovery/sync_manager.go
```diff
@@ -0,0 +1,637 @@
...
+
+	s, ok := m.gossipSyncer(peer)
+	if !ok {
+		return
+	}
+
+	log.Infof("Removing GossipSyncer for peer=%v", peer)
+
+	// We'll start by stopping the GossipSyncer for the disconnected peer.
+	s.Stop()
```
halseth: should we free the lock while waiting for the syncer to stop?

discovery/sync_manager_test.go
```diff
@@ -0,0 +1,656 @@
...
+	s, ok := syncMgr.GossipSyncer(peer.PubKey())
+	if !ok {
+		return fmt.Errorf("gossip syncer for peer %x not found",
+			peer.PubKey())
+	}
+
+	// We'll check the status of our syncer within a WaitPredicate as some
+	// sync transitions might cause this to be racy.
+	var predErr error
+	err := lntest.WaitPredicate(func() bool {
```
halseth: nit: can use `WaitNoError`

config.go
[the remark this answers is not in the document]
halseth: I agree that this should be fine. Likely not many will run into this.

lnrpc/rpc.proto
```diff
@@ -1163,6 +1163,27 @@ message Peer {
...
+
+        /**
+        Denotes that we are performing a full sync of our graph since genesis
+        with the peer. We are also not receiving new graph updates from them.
+        */
+        FULL_SYNC = 2;
+    }
+
+    // The type of sync we are currently performing with this peer.
+    SyncType sync_type = 10 [json_name = "sync_type"];
```
halseth: Useful!

rpcserver.go
```diff
@@ -2000,16 +2001,38 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
 		}
 
-		nodePub := serverPeer.addr.IdentityKey.SerializeCompressed()
+		nodePub := serverPeer.PubKey()
+
+		syncer, ok := r.server.authGossiper.SyncManager().GossipSyncer(
+			nodePub,
+		)
+		if !ok {
+			return nil, errors.New("gossip syncer for peer not found")
```
halseth: Are we always expected to have a gossip syncer for each connected peer?

I'm particularly thinking about the window between the peer connects, and the gossip syncer has been created. Same for disconnects. If that's the case maybe wort just logging a warning without returning an error here?

discovery/gossiper.go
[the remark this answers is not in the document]
wpaulino: Gotcha, this code ends up being moved anyway.

discovery/syncer.go
```diff
@@ -28,6 +28,13 @@ const (
 	//   2. Should respond to queries from the remote peer.
 	//   3. Should not receive new updates from the remote peer.
 	PassiveSync
+
+	// FullSync denotes that a gossip syncer:
+	//   1. Should force a sync from the earliest height known in the
+	//   chain with the remote peer.
+	//   2. Should respond to queries from the remote peer.
+	//   3. Should not receive new updates from the remote peer.
+	FullSync
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -597,7 +644,10 @@ func (g *GossipSyncer) genChanRangeQuery() (*lnwire.QueryChannelRange, error) {
 	// we don't miss any channels. By default, we go back 1 day from the
 	// newest channel.
 	var startHeight uint32
+	SyncerType := SyncerType(atomic.LoadUint32(&g.syncType))
```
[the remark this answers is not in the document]
wpaulino: Hah, left over from a search and replace. Fixed.

discovery/syncer_test.go
```diff
@@ -1940,3 +1957,218 @@ func TestGossipSyncerAlreadySynced(t *testing.T) {
...
+			syncer.state = uint32(chansSynced)
+
+			// We'll set the intiial syncType to what the test demands.
+			syncer.syncType = uint32(test.entrySyncType)
+
+			// Then, we'll start the syncer in order to process the
+			// request and stop it right after.
+			syncer.Start()
+			syncer.ProcessSyncTransition(test.finalSyncType)
+			syncer.Stop()
```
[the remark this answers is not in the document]
wpaulino: Nope, fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,637 @@
...
+	syncersMtx sync.RWMutex
+
+	// activeSyncers is the set of all syncers for which we are currently
+	// receiving graph updates from. The number of possible active syncers
+	// is bounded by NumActiveSyncers.
+	activeSyncers map[routing.Vertex]*GossipSyncer
+
+	// syncers is the set of all syncers for which we are not currently
+	// receiving graph updates from.
+	syncers map[routing.Vertex]*GossipSyncer
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,637 @@
...
+}
+
+// nextPendingActiveSyncer returns the next activeSyncer pending to be started.
+// If there aren't any, then `nil` is returned.
+func (m *SyncManager) nextPendingActiveSyncer() *activeSyncer {
+	next := m.pendingActiveSyncerQueue.Front()
+	for next != nil {
+		s := m.pendingActiveSyncerQueue.Remove(next).(*activeSyncer)
+
+		if _, ok := m.pendingActiveSyncers[s.cfg.peerPub]; !ok {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,637 @@
...
+		// We can then transition each syncer to their respective new
+		// sync type.
+		err := m.transitionSyncer(currentSyncer, PassiveSync)
+		if err != nil {
+			log.Errorf("Unable to transition GossipSyncer(%x) to "+
+				"%v: %v", currentSyncer.cfg.peerPub,
+				PassiveSync, err)
+			continue
+		}
+		m.transitionSyncer(candidate, ActiveSync)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: Don't think it'll be much of a problem, but agree that it is nice. Fixed.

discovery/sync_manager_test.go
wpaulino: Fixed.

rpcserver.go
wpaulino: Good point. We should always have one, but since this can be racy it's indeed better if we just log it. Fixed.

wpaulino:
PTAL @halseth.

wpaulino:
Comments addressed, PTAL @halseth @cfromknecht.

peer.go
```diff
@@ -387,7 +387,7 @@ func (p *peer) initGossipSync() {
 		//
 		// TODO(roasbeef): craft s.t. we only get updates from a few
 		// peers
-		recvUpdates := !cfg.NoChanUpdates
+		recvUpdates := cfg.GraphUpdatePeers != 0
```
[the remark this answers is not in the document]
wpaulino: Fixed.

lnrpc/rpc.proto
```diff
@@ -1163,6 +1163,27 @@ message Peer {
...
+        /**
+        Denotes that we are not receiving new graph updates from the peer.
+        */
+        PASSIVE_SYNC = 1;
+
+        /**
+        Denotes that we are performing a full sync of our graph since genesis
+        with the peer. We are also not receiving new graph updates from them.
+        */
+        FULL_SYNC = 2;
```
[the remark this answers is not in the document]

rpcserver.go
```diff
@@ -2000,16 +2001,46 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+		if ok {
+			syncType = syncer.SyncType()
+		} else {
+			rpcsLog.Debugf("Gossip syncer for peer=%x not found",
+				nodePub)
+			syncType = discovery.PassiveSync
+		}
+
+		var lnrpcSyncType lnrpc.Peer_SyncType
+		switch syncType {
```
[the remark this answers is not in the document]

discovery/syncer.go
```diff
@@ -399,7 +437,8 @@ func (g *GossipSyncer) channelGraphSyncer() {
 			// If we haven't yet sent out our update horizon, and
 			// we want to receive real-time channel updates, we'll
 			// do so now.
-			if g.localUpdateHorizon == nil && g.cfg.syncChanUpdates {
+			syncType := SyncerType(atomic.LoadUint32(&g.syncType))
```
halseth: syncType already loaded above

discovery/syncer.go
```diff
@@ -592,7 +639,10 @@ func (g *GossipSyncer) genChanRangeQuery() (*lnwire.QueryChannelRange, error) {
 	// we don't miss any channels. By default, we go back 1 day from the
 	// newest channel.
 	var startHeight uint32
+	syncType := SyncerType(atomic.LoadUint32(&g.syncType))
 	switch {
+	case syncType == HistoricalSync:
```
halseth: add comment

discovery/sync_manager.go
halseth: Couldn't it be a problem if we start a new syncer that (almost) immediately goes to `chansSynced`, triggering us to start a new one, same thing happens again etc...? Possibly need an active syncer stay our chosen one for a while.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+		}
+
+		return s
+	}
+
+	return nil
+}
+
+// rotateActiveSyncers rotates all currently eligible active syncers.
+func (m *SyncManager) rotateActiveSyncers() {
```
halseth: Guess this should be named `rotateActiveSyncerCandidates` or similar, current name a bit confusing.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+func (m *SyncManager) transitionSyncer(s *GossipSyncer, newSyncType SyncerType) error {
+	log.Debugf("Requesting transition to %v for GossipSyncer(%x)",
+		newSyncType, s.cfg.peerPub)
+
+	// If we're transitioning to an ActiveSync type, we'll short-circuit the
+	// process by queueing the syncer in our round-robin. Once it's this
+	// syncer's turn, the transition will happen. If this transition fails,
+	// then any new gossip syncers will be used as active syncers until
+	// reaching NumActiveSyncers.
+	if newSyncType == ActiveSync {
```
halseth: Since this method doesn't actually transitions syncers to the active state, consider handling that transition by calling `enqueueActiveSyncer` directly.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+		// machine. This signal allows us to start the next pending
+		// activeSyncer, if any.
+		startNext chan struct{}
+	)
+
+	// startSyncer is a helper closure we'll use to start a given syncer. If
+	// necessary, a sync transition to an ActiveSync sync type will be
+	// performed to ensure we receive new graph updates from this peer.
+	startSyncer := func(s *activeSyncer) error {
+		if s.transition {
```
halseth: Is this `transition` bool really needed? We could rather start all syncers in the passive state, and transition them always to active when using them.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+		s.syncType = uint32(HistoricalSync)
+		m.inactiveSyncers[nodeID] = s
+		s.Start()
+		return
+	}
+
+	// If we've yet to reach our desired number of active syncers, then
+	// we'll use this one.
+	if len(m.activeSyncers) < m.cfg.NumActiveSyncers {
+		s.syncType = uint32(ActiveSync)
```
halseth: As mentioned, I think it is easier to just let this be `Passive`, then let the call to `enqueueActiveSyncer` transition it into `Active`.

Same for `HistoricalSync` above: let it start `Passive`, then signal to the `syncerHandler` to force a historical sync.

discovery/sync_manager.go
wpaulino: What's wrong with the current approach? Would rather have one call in which a syncer can perform any transition.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/syncer.go
wpaulino: Fixed.

discovery/syncer.go
wpaulino: Fixed.

rpcserver.go
```diff
@@ -2000,16 +2001,46 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+		// syncer for the peer, then we'll default to a passive sync.
+		// This can happen if the RPC is called while a peer is
+		// initializing.
+		syncer, ok := r.server.authGossiper.SyncManager().GossipSyncer(
+			nodePub,
+		)
+
+		var lnrpcSyncType lnrpc.Peer_SyncType
+		if !ok {
+			rpcsLog.Debugf("Gossip syncer for peer=%x not found",
```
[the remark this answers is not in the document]
wpaulino: FIxed.

peer.go
```diff
@@ -381,19 +381,16 @@ func (p *peer) initGossipSync() {
...
-		// peers
-		recvUpdates := cfg.NumGraphUpdatePeers != 0
-
 		// Register the this peer's for gossip syncer with the gossiper.
 		// This is blocks synchronously to ensure the gossip syncer is
 		// registered with the gossiper before attempting to read
 		// messages from the remote peer.
-		p.server.authGossiper.InitSyncState(p, recvUpdates)
+		//
+		// TODO(wilmer): Only sync updates from non-channel peers. This
```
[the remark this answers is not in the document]
wpaulino: They won't really block link updates anymore due to the recently added `SendMessageLazy` call for gossip messages.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+	delete(m.activeSyncers, s.cfg.peerPub)
+	m.dequeueActiveSyncer(s.cfg.peerPub)
+
+	newActiveSyncer := m.chooseRandomSyncer(nil)
+	if newActiveSyncer != nil {
+		_ = m.transitionSyncer(newActiveSyncer, ActiveSync)
+	}
+}
+
+// GossipSyncer returns the associated gossip syncer of a peer.
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+	if len(m.activeSyncers) < m.cfg.NumActiveSyncers {
+		s.syncType = uint32(ActiveSync)
+		m.activeSyncers[nodeID] = s
+		m.enqueueActiveSyncer(s, false)
+		return
+	}
+
+	// Otherwise, the syncer should be initialized as passive.
+	s.syncType = uint32(PassiveSync)
+	s.state = uint32(waitingQueryRangeReply)
```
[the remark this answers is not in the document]
wpaulino: `PassiveSync` syncers only respond to peer queries and do not ask remote peers for anything, so this is expected.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+func (m *SyncManager) chooseRandomSyncer(
+	blacklist map[routing.Vertex]struct{}) *GossipSyncer {
+
+	for _, s := range m.inactiveSyncers {
+		if blacklist != nil {
+			if _, ok := blacklist[s.cfg.peerPub]; ok {
+				continue
+			}
+		}
+		if s.syncState() != chansSynced {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+// round-robin manner to ensure we're not syncing with multiple peers at the
+// same time.
+type SyncManager struct {
+	start sync.Once
+	stop  sync.Once
+
+	cfg SyncManagerCfg
+
+	// syncersMtx guards the access to the variables activeSyncers and
+	// syncers below.
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+	})
+}
+
+// Stop stops the SyncManager from performing its duties.
+func (m *SyncManager) Stop() {
+	m.stop.Do(func() {
+		close(m.quit)
+		m.wg.Wait()
+
+		for _, syncer := range m.activeSyncers {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,640 @@
...
+			syncer.Stop()
```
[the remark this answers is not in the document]
wpaulino: `Start` and `Stop` don't return any errors at the moment, so just ended up removing them.

discovery/syncer.go
```diff
@@ -970,6 +1020,113 @@ func (g *GossipSyncer) ProcessQueryMsg(msg lnwire.Message, peerQuit <-chan struc
...
+				"horizon: %v", err)
+		}
+
+		// If the gossip syncer currently has a PassiveSync type, then
+		// we'll revert to our initial syncingChans state to ensure we
+		// can reconcile any channels we're missing that the remote peer
+		// has. This isn't needed when transitioning from a HistoricalSync, as
+		// both sides have synced their complete graphs with each other.
+		if syncType == PassiveSync {
+			atomic.StoreUint32(&g.state, uint32(syncingChans))
```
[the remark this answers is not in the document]
wpaulino: Not really other than some minimal bandwidth gains. Since `PassiveSync` is the only type that doesn't require sending out our set of channels, we make sure to do so when transitioning to another type.

discovery/syncer.go
```diff
@@ -127,15 +145,23 @@ var (
...
-	// our highest known channel ID.
-	chanRangeQueryBuffer = 144
+	// ErrSyncTransitionTimeout is an error returned when we've timed out
+	// attempting to perform a sync transition.
+	ErrSyncTransitionTimeout = errors.New("timed out attempting to " +
+		"transition sync type")
+
+	// zeroTimestamp is the timestamp we'll use when we want to indicate to
+	// peers that we do not want to receive any new graph updates.
+	zeroTimestamp time.Time
```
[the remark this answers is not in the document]
wpaulino: Not possible: `const initializer time.Time literal is not a constant`.

discovery/syncer.go
```diff
@@ -269,8 +304,11 @@ func (g *GossipSyncer) channelGraphSyncer() {
 
 	for {
 		state := atomic.LoadUint32(&g.state)
-		log.Debugf("GossipSyncer(%x): state=%v", g.cfg.peerPub[:],
-			syncerState(state))
+		syncType := atomic.LoadUint32(&g.syncType)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -107,9 +107,9 @@ const (
 	chanRangeQueryBuffer = 144
 )
 
-// gossipSyncerCfg is a struct that packages all the information a gossipSyncer
+// GossipSyncerCfg is a struct that packages all the information a GossipSyncer
 // needs to carry out its duties.
-type gossipSyncerCfg struct {
+type GossipSyncerCfg struct {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: I don't see a problem with that. The round-robin is just to ensure that we do the sync exchange with one active syncer at a time. New graph updates can be received from any active syncer, regardless of their order in the round-robin.

discovery/syncer.go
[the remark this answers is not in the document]
wpaulino: If the `GossipSyncer` transitions away from `chansSynced` for whatever reason between the time we're filtering for valid candidates and actually transitioning the syncer.

discovery/sync_manager.go
wpaulino: The transition requires the syncer to be in a `chansSynced` state, so if we go with this approach they'll have to go through the state machine twice (once the first time they're started, another after the transition).

discovery/syncer.go
```diff
@@ -970,6 +1020,113 @@ func (g *GossipSyncer) ProcessQueryMsg(msg lnwire.Message, peerQuit <-chan struc
...
+// transition.
+func (g *GossipSyncer) handleSyncTransition(req *syncTransitionReq) error {
+	syncType := SyncerType(atomic.LoadUint32(&g.syncType))
+	log.Debugf("GossipSyncer(%x): transitioning from %v to %v",
+		g.cfg.peerPub, syncType, req.newSyncType)
+
+	switch req.newSyncType {
+	// If an active sync has been requested, then we should resume receiving
+	// new graph updates from the remote peer.
+	case ActiveSync:
```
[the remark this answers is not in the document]
wpaulino: What about the transitions is hard to reason about? Is it just lacking some more documentation? Each type does its own thing and requires a minimal set of steps to transition to another type.

Would say I'm not a fan of mixing enums with bitfields, although I guess there is a use in allowing active syncers to also perform historical syncs when having <= NumActiveSyncers peers connected.

discovery/syncer.go
```diff
@@ -264,9 +299,11 @@ func (g *GossipSyncer) channelGraphSyncer() {
 	//  * needed if we want to sync chan state very few blocks?
 
 	for {
-		state := atomic.LoadUint32(&g.state)
-		log.Debugf("GossipSyncer(%x): state=%v", g.cfg.peerPub[:],
-			syncerState(state))
+		state := g.syncState()
```
halseth: 👍

discovery/sync_manager.go
halseth: Imagine there are no new updates, and all nodes are perfectly in sync. The problem now is that you'll choose an active syncer, but since you are already in sync you'll immediately move to `chansSynced` and trigger the `startNext` signal. Now you'll continue rapidly switching active syncers. Or is there something here that prevents that scenario?

discovery/sync_manager.go
```diff
@@ -0,0 +1,651 @@
...
+	// If we've yet to reach our desired number of active syncers, then
+	// we'll use this one.
+	if len(m.activeSyncers) < m.cfg.NumActiveSyncers {
+		s.setSyncType(ActiveSync)
+		m.activeSyncers[nodeID] = s
+		m.enqueueActiveSyncer(s, false)
+		return
+	}
+
+	// Otherwise, the syncer should be initialized as passive.
```
halseth: I think that what make this complex is the interaction between the syncer's state and the `syncerHandler`. Here we  modify the syncer state, start it an also modify the `SyncManager`s maps. I'm in favor of giving this responsibility completely to the `syncerHandler`, to avoid weird concurrency behavior.

In `InitSyncState` you would just send the new syncer to the `syncerHandler`, and it would decide which state to start it in. Would also trivially remove the `transition` boolean.

discovery/sync_manager.go
```diff
@@ -0,0 +1,651 @@
...
+
+	m.cfg.HistoricalSyncTicker.Resume()
+	defer m.cfg.HistoricalSyncTicker.Stop()
+
+	defer m.cfg.ActiveSyncerTimeoutTicker.Stop()
+
+	var (
+		// current will hold the current activeSyncer we're waiting for
+		// to complete its state machine.
+		current *activeSyncer
```
halseth: Doesn't this contradict the fact that we can have multiple active syncers?

discovery/sync_manager.go
halseth: What is wrong is that the `syncerHandler` is sending messages _to itself_ with the current approach, which complicates the logic.

Maybe I'm confused about the distinction between active and pending-active syncers, but could it be enough to just let `rotateActiveSyncerCandidates` shuffle the pending-active map without doing any transitioning?

discovery/sync_manager.go
wpaulino: Again, the round-robin is just to ensure we only query one peer at a time for _historical_ updates. Once a syncer reaches `chansSynced`, we can receive _new_ updates from the peer at any point.

discovery/sync_manager.go
wpaulino: I think there's some confusion about what the round-robin is meant to achieve. See my previous comment: https://github.com/lightningnetwork/lnd/pull/2740#discussion_r270156880

discovery/sync_manager.go
wpaulino: Even if we do that, we'll still need to transition it once the ticker fires, so it'll send itself a message anyway. I agree that the handler sending messages to itself can be somewhat weird, but it seems like the only alternative is to have a separate handler responsible for just the tickers.

discovery/sync_manager.go
wpaulino: Perhaps `syncerHandler` is not the right name for the method, which is causing some confusion? Its main purpose is just to handle the round-robin queue of active syncers, so it could be renamed to `activeSyncerHandler` and the tickers could be moved into a separate handler.

discovery/syncer.go
```diff
@@ -978,6 +1026,85 @@ func (g *GossipSyncer) syncState() syncerState {
...
+		newState       syncerState
+	)
+
+	switch req.newSyncType {
+	// If an active sync has been requested, then we should resume receiving
+	// new graph updates from the remote peer.
+	case ActiveSync:
+		firstTimestamp = time.Now()
+		timestampRange = math.MaxUint32
+		newState = syncingChans
```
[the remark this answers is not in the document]
wpaulino: Since we don't query the remote peer for their channels when under a passive sync, I figured it should happen once it transitions to active.

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+	numActiveSyncers := len(m.activeSyncers) + len(m.pendingActiveSyncers)
+	m.Unlock()
+
+	switch {
+	// If we don't have any active syncers, we'll force a historical sync
+	// with the first peer we connect to, to ensure we get as much of the
+	// graph as possible. This is mostly done for light clients, which
+	// shouldn't be receiving new updates anyway due to the high bandwidth
+	// consumption that stems from it.
+	case m.cfg.NumActiveSyncers == 0:
```
cfromknecht: imo this should be it's own config option, idt it's logical that if you set the number of active syncers to zero that it will go ahead and do a historical sync. i can also imagine cases where you want both

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+	// we don't happen to have any non-active syncers.
+	candidatesChosen := make(map[routing.Vertex]struct{})
+	s := m.chooseRandomSyncer(candidatesChosen, true)
+	for s != nil {
+		// Blacklist the candidate to ensure it's not chosen again.
+		candidatesChosen[s.cfg.peerPub] = struct{}{}
+
+		err := s.historicalSync()
+		if err == nil {
+			break
```
cfromknecht: style nit: i think return is clearer

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+
+			// Reset historicalSync to ensure it is tried again with
+			// a different peer.
+			m.historicalSync = sync.Once{}
+		}
+
+	// If we've yet to reach our desired number of active syncers, then
+	// we'll use this one.
+	case numActiveSyncers < m.cfg.NumActiveSyncers:
+		m.queueActiveSyncer(s)
```
halseth: nice, this is much easier to reason about!

I think we can make it even clearer: just queue the syncer here, and let the `syncerHandler` decide whether we should do a historical sync with the syncer.

discovery/gossiper_test.go
wpaulino: Should be this: https://github.com/lightningnetwork/lnd/pull/2740/commits/0de846370ebb8c3ae180068fbd86af92c0e7cc28

discovery/syncer.go
```diff
@@ -633,17 +648,19 @@ func (g *GossipSyncer) genChanRangeQuery() (*lnwire.QueryChannelRange, error) {
...
+	// miss any channels. By default, we go back 1 day from the newest
+	// channel, unless we're attempting a historical sync, where we'll
+	// actually start from the genesis block instead.
 	var startHeight uint32
 	switch {
 	case newestChan.BlockHeight <= chanRangeQueryBuffer:
 		fallthrough
 	case newestChan.BlockHeight == 0:
+		fallthrough
+	case g.genHistoricalChanRangeQuery:
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: Fixed.

discovery/gossiper_test.go
cfromknecht: That’s for the gossip syncer, not the gossipper

discovery/sync_manager.go
wpaulino: Fixed.

discovery/syncer.go
```diff
@@ -530,52 +628,54 @@ func (g *gossipSyncer) processChanRangeReply(msg *lnwire.ReplyChannelRange) erro
...
+	// of the channel, then subtract our default horizon to ensure we don't
+	// miss any channels. By default, we go back 1 day from the newest
+	// channel, unless we're attempting a historical sync, where we'll
+	// actually start from the genesis block instead.
 	var startHeight uint32
 	switch {
 	case newestChan.BlockHeight <= chanRangeQueryBuffer:
 		fallthrough
 	case newestChan.BlockHeight == 0:
+		fallthrough
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: I thought of doing this as well but seemed like there would be little to no gain since we'd still require acquiring the locks.

peer.go
```diff
@@ -402,7 +402,7 @@ func (p *peer) initGossipSync() {
 		//
 		// TODO(roasbeef): craft s.t. we only get updates from a few
 		// peers
-		recvUpdates := !cfg.NoChanUpdates
+		recvUpdates := cfg.NumGraphUpdatePeers != 0
```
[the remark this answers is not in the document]
wpaulino: It gets removed in the following commit when the integration is done :)

Roasbeef:
Two things:
  1. I think "initial graph sync" should always be set to `true` (so no flag), since after the first run, it's basically "free" since it'll just be a noop
  2. The default time of 6hrs for peer rotation is too long IMO, it can be much shorter (like 10-20 mins maybe?) given that after the first real exchange, it's just a spot check, and we want to ensure for light clients that we converge to the known graph as soon as possible"

wpaulino:
Comments addressed, PTAL @Roasbeef @halseth.

> The default time of 6hrs for peer rotation is too long IMO, it can be much shorter (like 10-20 mins maybe?) given that after the first real exchange, it's just a spot check, and we want to ensure for light clients that we converge to the known graph as soon as possible

I ran some manual light client tests and noticed historical syncs can take some time to completely finish (10-15 mins), so I settled with 20 mins.

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+
+const (
+	// DefaultSyncerRotationInterval is the default interval in which we'll
+	// rotate a single active syncer.
+	DefaultSyncerRotationInterval = 20 * time.Minute
+
+	// DefaultHistoricalSyncInterval is the default interval in which we'll
+	// force a historical sync to ensure we have as much of the public
+	// network as possible.
+	DefaultHistoricalSyncInterval = 20 * time.Minute
```
halseth: an idea to avoid the rotation and historical ticker ticking at the same time?

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+	cfg SyncManagerCfg
+
+	// historicalSync allows us to perform an initial historical sync only
+	// _once_ with a peer during the SyncManager's startup.
+	historicalSync sync.Once
+
+	// activeSyncers is the set of all syncers for which we are currently
+	// receiving graph updates from. The number of possible active syncers
+	// is bounded by NumActiveSyncers.
+	activeSyncers map[routing.Vertex]*GossipSyncer
```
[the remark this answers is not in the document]
wpaulino: The mutex is still needed between the `syncHandler` and `roundRobinHandler` in order to determine whether it should be queued or not.

> The pendingActiveSyncers map is not needed, as you can remove directly from the list.

This was suggested [here](https://github.com/lightningnetwork/lnd/pull/2740#discussion_r268017032) to avoid an O(N) lookup and slice tricks when removing a stale active syncer, but agree that it's not really needed.

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+
+	// Otherwise, we'll attempt to transition each syncer to their
+	// respective new sync type.
+	if err := m.transitionActiveSyncer(activeSyncer); err != nil {
+		log.Errorf("Unable to transition active "+
+			"GossipSyncer(%x): %v", activeSyncer.cfg.peerPub, err)
+		return
+	}
+	m.queueActiveSyncer(candidate)
+	m.signalNewActiveSyncer()
```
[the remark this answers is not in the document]
wpaulino: There needs to be a distinction since the `roundRobinHandler` also calls `queueActiveSyncer` but doesn't need the signal.

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+			return
+		}
+	}
+}
+
+// signalNewActiveSyncer sends a signal to the roundRobinHandler to ensure it
+// transitions any pending active syncers.
+func (m *SyncManager) signalNewActiveSyncer() {
+	select {
+	case m.newActiveSyncer <- struct{}{}:
```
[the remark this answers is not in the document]
wpaulino: I believe not making it block can introduce a slight race condition where the syncer isn't transitioned until a new one comes in.

discovery/sync_manager.go
wpaulino: Gotcha, the same pattern remains but the concurrency issue you pointed out should be fixed now.

discovery/syncer.go
```diff
@@ -1079,6 +1099,7 @@ func (g *GossipSyncer) handleSyncTransition(req *syncTransitionReq) error {
 		firstTimestamp = time.Now()
 		timestampRange = math.MaxUint32
 		newState = syncingChans
+		g.genHistoricalChanRangeQuery = false
```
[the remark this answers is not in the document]
wpaulino: Ended up removing the line completely. If it's done a historical sync before, then it shouldn't be a problem to do it again as it'll most likely be a NOP.

discovery/sync_manager.go
```diff
@@ -0,0 +1,673 @@
...
+	if !ok {
+		return
+	}
+
+	log.Infof("Removing GossipSyncer for peer=%v", peer)
+
+	// We'll start by stopping the GossipSyncer for the disconnected peer.
+	s.Stop()
+
+	// If it's a non-active syncer, then we can just exit now.
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/sync_manager.go
wpaulino: Discussed offline, most likely won't be an issue.

discovery/sync_manager.go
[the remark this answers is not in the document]
halseth: Resolved offline.

discovery/sync_manager.go
```diff
@@ -0,0 +1,735 @@
...
+		// pending active syncer if there is one.
+		case <-transitionNext:
+			transitionNextSyncer()
+
+		// We've timed out waiting for the current active syncer to
+		// reach its terminal chansSynced state, so we'll just
+		// move on to the next and avoid retrying as its already been
+		// transitioned.
+		case <-m.cfg.ActiveSyncerTimeoutTicker.Ticks():
+			log.Debugf("Timed out waiting for GossipSyncer(%x) "+
```
halseth: Warn?

discovery/sync_manager.go
```diff
@@ -0,0 +1,735 @@
...
+	// If we've yet to reach our desired number of active syncers, then
+	// we'll use this one.
+	numActiveSyncers := len(m.activeSyncers) + len(m.pendingActiveSyncers)
+	if numActiveSyncers < m.cfg.NumActiveSyncers {
+		m.queueActiveSyncer(s)
+		m.Unlock()
+		m.signalNewActiveSyncer()
+		return
+	}
+	m.Unlock()
```
halseth: can also just defer the unlock after locking

wpaulino: We don't want to hold it while `signalNewActiveSyncer` since it can potentially block for a bit if also instantiating many other syncers.

discovery/sync_manager.go
wpaulino: Fixed.

cfromknecht:
Needs rebase now that scb rest has been merged

PR #2744, opened 2019-03-08
cnct+chancloser: disable channel before closing

### LINKED ISSUE #2696: Unable to sign update disabling channel - Edge not found
### Background
Hello, I have several errors of this type:

2019-02-24 20:31:54.757 [INF] NANN: Announcing channel(3709624b745984791a08ee42faafed7a81a41135a37a754a3c5206234fffd4b2:1) disabled [detected]
2019-02-24 20:31:54.758 [ERR] NANN: Unable to sign update disabling channel(3709624b745984791a08ee42faafed7a81a41135a37a754a3c5206234fffd4b2:1): edge not found

The channel is closed for days and I'm not connected to that peer anymore.
Is this a bug?
Thanks in advice

### Your environment

lnd 0.5.1-beta-784
Linux
Bitcoin core 0.17.1

cfromknecht:
This PR reorders logic when closing channels to sign the disabling update before closing the channel in the database. Currently the disable is attempted asynchronously after the close, after which the channel may be removed from the graph and leads to logs generated in #2696.

The fix is simply to do the disables synchronously before closing the channels either cooperatively or unilaterally. The integration test framework has been updated to assert that a disable message is always sent after closing if the channel policy indicates it was enabled at the time of the closure.

The one exception to this was the reorg channel funding itest, for which this cannot happen because the channel has been removed from the channel graph. A separate assertion that ignores the disable assertion is added to handle this case.

Fixes #2696

### DIFF
--- a/chancloser.go
+++ b/chancloser.go
@@ -427,6 +427,14 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
 		}
 		c.closingTx = closeTx
 
+		// Before closing, we'll attempt to send a disable update for
+		// the channel. We do so before closing the channel as otherwise
+		// the current edge policy won't be retrievable from the graph.
+		if err := c.cfg.disableChannel(c.chanPoint); err != nil {
+			peerLog.Warnf("Unable to disable channel %v on "+
+				"close: %v", c.chanPoint, err)
+		}
+
 		// With the closing transaction crafted, we'll now broadcast it
 		// to the network.
 		peerLog.Infof("Broadcasting cooperative close tx: %v",
@@ -439,16 +447,6 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
 		if err := c.cfg.channel.MarkCommitmentBroadcasted(); err != nil {
 			return nil, false, err
 		}
-
-		// We'll attempt to disable the channel in the background to
-		// avoid blocking due to sending the update message to all
-		// active peers.
-		go func() {
-			if err := c.cfg.disableChannel(c.chanPoint); err != nil {
-				peerLog.Errorf("Unable to disable channel %v on "+
-					"close: %v", c.chanPoint, err)
-			}
-		}()
 
 		// Finally, we'll transition to the closeFinished state, and
 		// also return the final close signed message we sent.
--- a/contractcourt/chain_arbitrator.go
+++ b/contractcourt/chain_arbitrator.go
@@ -635,6 +635,14 @@ func (c *ChainArbitrator) ForceCloseContract(chanPoint wire.OutPoint) (*wire.Msg
 
 	log.Infof("Attempting to force close ChannelPoint(%v)", chanPoint)
 
+	// Before closing, we'll attempt to send a disable update for the
+	// channel. We do so before closing the channel as otherwise the current
+	// edge policy won't be retrievable from the graph.
+	if err := c.cfg.DisableChannel(chanPoint); err != nil {
+		log.Warnf("Unable to disable channel %v on "+
+			"close: %v", chanPoint, err)
+	}
+
 	errChan := make(chan error, 1)
 	respChan := make(chan *wire.MsgTx, 1)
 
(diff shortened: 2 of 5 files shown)


chancloser.go
```diff
@@ -427,6 +427,14 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
 		}
 		c.closingTx = closeTx
 
+		// We'll attempt to disable the channel in the background to
+		// avoid blocking due to sending the update message to all
+		// active peers.
```
wpaulino: The outdated comment can be removed.

contractcourt/chain_arbitrator.go
```diff
@@ -623,6 +623,11 @@ func (c *ChainArbitrator) ForceCloseContract(chanPoint wire.OutPoint) (*wire.Msg
 
 	log.Infof("Attempting to force close ChannelPoint(%v)", chanPoint)
 
+	if err := c.cfg.DisableChannel(chanPoint); err != nil {
```
wpaulino: If we fail to force close the channel below, then we'll leave the channel disabled. I guess this is fine since in most force close cases the peer is offline, so it'll be re-enabled once the peer reconnects.

lnd_test.go
```diff
@@ -223,16 +223,80 @@ func openChannelAndAssert(ctx context.Context, t *harnessTest,
...
+// closeReorgedChannelAndAssert attempts to close a channel identified by the
+// passed channel point owned by the passed Lightning node. A fully blocking
+// channel closure is attempted, therefore the passed context should be a child
+// derived via timeout from a base parent. Additionally, once the channel has
+// been detected as closed, an assertion checks that the transaction is found
+// within a block.
+//
+// NOTE: This method does not verify that the node sends a disable update for
+// the closed channel.
+func closeReorgedChannelAndAssert(ctx context.Context, t *harnessTest,
```
wpaulino: Heh... if the channel could be reorged as well then this wouldn't be needed. Maybe we can just use `AbandonChannel` instead of trying to force close it?

cfromknecht: we could, but this also cleans up the mempool and verifies that the node reverts to the proper state 🤷‍♂️

contractcourt/chain_arbitrator.go
cfromknecht: yep exactly

chancloser.go
cfromknecht: good catch

cfromknecht:
@wpaulino latest version should fix the travis failure, status manager was not started for the mock peer causing the operation to block

chancloser.go
```diff
@@ -427,6 +427,14 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
 		}
 		c.closingTx = closeTx
 
+		// Before closing, we'll attempt to send a disable update for
+		// the channel. We do so before closing the channel as otherwise
+		// the current edge policy won't be retrievable from the graph.
+		if err := c.cfg.disableChannel(c.chanPoint); err != nil {
```
wpaulino: Is there a reason for this not being able to stay in the same place (after `MarkCommitmentBroadcasted`)?

cfromknecht: the disable should occur before broadcasting the transaction, as there is an unlikely case where the txn gets instamined and the edge is no longer in our database to disable.

Looking at this though, I would say we should mark commitment broadcast before broadcasting the transaction, and then reliably publish on restarts. I would also venture to say this should be marked with a different status since we aren't actually broadcasting a commitment, cc @Roasbeef

cfromknecht:
@wpaulino new commit added that will properly remove the chanpoint from set of tracked outpoints when the remote party force closes, addresses an additional case where logs in #2696 can occur

chancloser.go
Roasbeef: I think this case is distinct from the force close case, as a force close means we don't want to continue with the channel at all (either HTLC timeout or we want to exit nao), vs a cooperative channel closure, which if really needed, should be re-attempted once both nodes are active again.

PR #2749, opened 2019-03-09
lnwallet: within ForceClose, mark the channel as borked, prevent mutations for borked channels

Roasbeef:
In this PR, we ensure that it isn't possible for a state mutation to take place _after_ the force close process has began. Before this commit, it may have been possible for a link to trigger/accept a new state update concurrently while a channel was being updated. We'll now mark the channel as borked in the `ForceClose` method which will:
  * Prevent all channel database commitment mutation calls from going through 
  * Prevent the target link reconnecting 

Along the way, another bug was fixed that would cause an error in the state machine from silently passing through, in the event that a call to `SignNextCommitment` failed at the database update state.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -107,6 +107,10 @@ var (
 	// mutate a channel that's been recovered.
 	ErrNoRestoredChannelMutation = fmt.Errorf("cannot mutate restored " +
 		"channel state")
+
+	// ErrChanBorked is returned when a caller attempts to mutate a borked
+	// channel.
+	ErrChanBorked = fmt.Errorf("channel mutate borked channel")
 )
 
 // ChannelType is an enum-like type that describes one of several possible
@@ -807,6 +811,20 @@ func (c *OpenChannel) MarkBorked() error {
 	return c.putChanStatus(ChanStatusBorked)
 }
 
+// isBorked returns true if the channel has been marked as borked in the
+// database. This requires an existing database transaction to already be
+// active.
+//
+// NOTE: The primary mutex should already be held before this method is called.
+func (c *OpenChannel) isBorked(chanBucket *bbolt.Bucket) (bool, error) {
+	channel, err := fetchOpenChannel(chanBucket, &c.FundingOutpoint)
+	if err != nil {
+		return false, err
+	}
+
+	return channel.chanStatus != ChanStatusDefault, nil
+}
+
 // MarkCommitmentBroadcasted marks the channel as a commitment transaction has
 // been broadcast, either our own or the remote, and we should watch the chain
 // for it to confirm before taking any further action.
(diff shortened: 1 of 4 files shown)


lnwallet/channel_test.go
```diff
@@ -6525,3 +6525,72 @@ func TestForceCloseFailLocalDataLoss(t *testing.T) {
 			"chan state")
 	}
 }
+
+// TestForceCloseBorkedState tests that once we force close a channel, it's
+// marked as barked in the database. Additionally, all calls to mutate channel
```
alexbosworth:
```suggestion
// marked as borked in the database. Additionally, all calls to mutate channel
```

channeldb/channel.go
```diff
@@ -807,6 +811,20 @@ func (c *OpenChannel) MarkBorked() error {
 	return c.putChanStatus(ChanStatusBorked)
 }
 
+// isBorked returns true if the channel has been marked as borked in the
+// database. This requires an existing database transaction to already be
+// active.
+//
+// NOTE: The primary mutex should already be held before this method is called.
+func (c *OpenChannel) isBorked(chanBucket *bbolt.Bucket) (bool, error) {
```
cfromknecht: should we make this stricter, and be checking `isDefault` and disallowing state transitions in all other states?

lnwallet/channel.go
```diff
@@ -5695,6 +5695,13 @@ func (lc *LightningChannel) ForceClose() (*LocalForceCloseSummary, error) {
 	// contested state.
 	lc.status = channelDispute
 
+	// Before we return the force close summary, we'll also mark the
+	// channel as borked to indicate that no new state transitions should
+	// be permitted.
+	if err := lc.channelState.MarkBorked(); err != nil {
```
cfromknecht: should be moved above `lc.status = channelDispute` to prevent in-memory mutations in failure case

contractcourt/chain_arbitrator.go
```diff
@@ -242,6 +242,10 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 				return nil, err
 			}
 
+			if err := channel.MarkBorked(); err != nil {
```
cfromknecht: add comment explaining that this permanently prevents the link from being readded to the switch after it's removed

contractcourt/chain_arbitrator.go
```diff
@@ -242,6 +242,10 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
...
+				return nil, err
+			}
+
```
cfromknecht: we should move the instantiation of the `chanMachine` until after the link has been removed from the switch, as it's possible it can be mutated while the link is shutting down

cfromknecht: currently it's possible that it could read a commitment, which gets replaced while link is shutting down, then plays the old one

channeldb/channel.go
Roasbeef: As is, the default state will always be set, and the remaining bits are interpreted as a bitvector.

contractcourt/chain_arbitrator.go
Roasbeef: Re-ordered, current order:
  * mark as borked
  * mark link inactive
  * fetch channel (peer not there anymore)
  * force close the channel

channeldb/channel.go
cfromknecht: the default state is the all-zero case, i'm suggest making the check test if _any_ bits are present

contractcourt/chain_arbitrator.go
cfromknecht: nice, ordering lgtm

channeldb/channel.go
Roasbeef: updated to disallow if non default (anything non default is borked)

PR #2750, opened 2019-03-09
htlcswitch/switch: log target node on insufficient cap error

cryptagoras:
Tiny addition. Added target node's pubkey on the insufficient capacity error.
It helps identify nodes that could use more inbound liquidity (i.e. potential fees).

### DIFF
--- a/htlcswitch/switch.go
+++ b/htlcswitch/switch.go
@@ -1038,7 +1038,8 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 
 			return s.failAddPacket(packet, failure, addErr)
 		}
-		interfaceLinks, _ := s.getLinks(targetLink.Peer().PubKey())
+		targetPeerKey := targetLink.Peer().PubKey()
+		interfaceLinks, _ := s.getLinks(targetPeerKey)
 		s.indexMtx.RUnlock()
 
 		// We'll keep track of any HTLC failures during the link
@@ -1105,7 +1106,7 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 
 			addErr := fmt.Errorf("unable to find appropriate "+
 				"channel link insufficient capacity, need "+
-				"%v", htlc.Amount)
+				"%v towards node=%x", htlc.Amount, targetPeerKey)
 
 			return s.failAddPacket(packet, failure, addErr)
 


htlcswitch/switch.go
```diff
@@ -1103,9 +1103,15 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 				failure = lnwire.NewTemporaryChannelFailure(update)
 			}
 
+			var nodeKey []byte
+			targetPeer := targetLink.Peer()
```
halseth: It is already assumed on line 1041  that this `Peer` is available. I would just extract the node key there :)

htlcswitch/mock.go
```diff
@@ -183,8 +183,8 @@ func newMockServer(t testing.TB, name string, startingHeight uint32,
 	db *channeldb.DB, defaultDelta uint32) (*mockServer, error) {
 
 	var id [33]byte
-	h := sha256.Sum256([]byte(name))
-	copy(id[:], h[:])
```
cryptagoras: Previously it was generating an invalid key which was throwing errors (invalid magic, square root) on some tests with the addition of `targetPeer.IdentityKey().SerializeCompressed()` at `htlcswitch/switch.go`.
So I changed it to generate the key using btcec using the provided name bytes, probably slower than before tho.

htlcswitch/switch.go
```diff
@@ -1038,7 +1038,9 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 
 			return s.failAddPacket(packet, failure, addErr)
 		}
-		interfaceLinks, _ := s.getLinks(targetLink.Peer().PubKey())
+		targetPeer := targetLink.Peer()
+		targetNodeKey := targetPeer.IdentityKey().SerializeCompressed()
```
halseth: Just call `targetPeer.PubKey()` directly?

cryptagoras: :man_facepalming: right, fixed and removed the extra var.

htlcswitch/switch.go
```diff
@@ -1038,7 +1038,8 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 
 			return s.failAddPacket(packet, failure, addErr)
 		}
-		interfaceLinks, _ := s.getLinks(targetLink.Peer().PubKey())
+		targetPeer := targetLink.Peer()
```
halseth: Even better:
```go
targetPeerKey := targetLink.Peer().PubKey()
interfaceLinks, _ := s.getLinks(targetPeerKey)
```

cryptagoras: :man_facepalming:² fixed, thx

htlcswitch/mock.go
halseth: Is this change still needed?

cryptagoras: Tested and looks fine, tests also fine. Removed, thx.

PR #2754, opened 2019-03-11
chancloser+rpcserver: proper error checking

chokoboko:


halseth:
nice catch! Did you use a tool to find these errors?

chokoboko:
@halseth Thanks! No tool is used. I saw this commit dac35c46f3ac976927fef0f26ab4913d044f559f from @roasbeef and used a "git grep" to check all code for similar cases.

PR #2755, opened 2019-03-11
Allow listening on non-localhost if TOR is enabled

NicolasDorier:
There is good reasons to allow both TOR and non localhost P2P connection.

* when you care about TOR to be compatible with other TOR nodes.
* when you care about TOR only as a solution for firewall punching

### DIFF
--- a/config.go
+++ b/config.go
@@ -962,20 +962,6 @@ func loadConfig() (*config, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Roasbeef:
Duplicate of https://github.com/lightningnetwork/lnd/pull/2490 ?

NicolasDorier:
yup, closing this, #2490 is better.

PR #2759, opened 2019-03-12
lnd: lower default CLTV delta from 144 to 40

Roasbeef:
In this commit, we lower the default CLTV delta from 144 (1 day) to 40
blocks (~7 hours). We do this as the initial value was purposefully very
high, and was only in place as place holder in earlier versions of the
software. Such a high value can also discourage path finding algorithms
from taking lnd paths due to the negative impact of CLTV value in widely
deployed heuristics.

### DIFF
--- a/chainregistry.go
+++ b/chainregistry.go
@@ -37,7 +37,7 @@ const (
 	defaultBitcoinMinHTLCMSat   = lnwire.MilliSatoshi(1000)
 	defaultBitcoinBaseFeeMSat   = lnwire.MilliSatoshi(1000)
 	defaultBitcoinFeeRate       = lnwire.MilliSatoshi(1)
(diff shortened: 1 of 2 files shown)


chainregistry.go
```diff
@@ -37,7 +37,7 @@ const (
 	defaultBitcoinMinHTLCMSat   = lnwire.MilliSatoshi(1000)
 	defaultBitcoinBaseFeeMSat   = lnwire.MilliSatoshi(1000)
 	defaultBitcoinFeeRate       = lnwire.MilliSatoshi(1)
-	defaultBitcoinTimeLockDelta = 144
+	defaultBitcoinTimeLockDelta = 40
```
halseth: decrease for LTC?

PR #2761, opened 2019-03-12
[reliable payments] router payment state machine

halseth:
This PR introduces a persistent state machine to the `ChannelRouter`'s payment flow, ensuring we can handle payment results that are received after a restart. It also paves they way for adding cancellation of payments, and resuming payment sessions across restarts.

### Problem
`lnd` currently runs into problems if it is restarted while an HTLC is in flight on the network. The primary reason for this is the way the `router` hands the HTLC to the `switch`.  The `router` persist no information about the payment, so when a result eventually comes back, we risk the information needed to properly populate the `OutgoingPayment` in the database is lost.

### Solution
With the `paymentStateMachine` introduced in this PR, we store two pieces of key information:
1. The HTLC `paymentID`. This is used to query the `Switch` whether the HTLC is still active. If not active we know that we can safely retry the payment attempt. If active the `router` will wait for the result to be available, and store the result to the DB.
2. The HTLC's `route`.  This is added to the DB  together with the preimage when the payment succeeds. 

#### Note
The `Switch` does not currently persist the pending payment attempts across restarts. This will be added in a follow-up PR.

Replaces #2475

Builds on 
- [x] #3065 
- [x] #3064 
- [x] #3063

PR #2762, opened 2019-03-12
[reliable payments] persist htlcswitch pending payments

halseth:
This PR is a follow-up from #2761. It adds a persistent `pendingPaymentStore` to the `Switch`, needed to store payment results when the `router` is not available to handle them.

### Problem
After a restart, there is no connection between an HTLC in flight on the network and the `router` payment flow that initially sent the HTLC. The `router` will eventually call `GetPaymentResult` to retrieve the result of the HTLC, but we have no guarantees this will be done before the result is received. This can lead to the result getting dropped, which is not ideal.

### Solution
We introduce a `pendingPayementStore` which has two primary functions:
1. It stores `PaymentResult`s when they are received, regardless of whether there are any subscribers (the `ChannelRouter`) to the corresponding HTLC. This ensures that the `router` gets handed the result when it calls `GetPaymentResult`.
2. It stores the information necessary to identify and handle a result that comes back after a restart. This is the `chanID` and `paymentID`, used to identify the payment in the circuit map.
3. On restarts it syncs the pending payments with the `CircuitMap`. We use the circuit map as the source of truth whether a payment is in flight on the network. Syncing it at startup, before we start handling HTLCs, is necessary to be able to decide whether an HTLC was successfully forwarded.

Builds on #2761 
Replaces #2265

### DIFF
--- a/htlcswitch/payment_result.go
+++ b/htlcswitch/payment_result.go
@@ -1,14 +1,24 @@
 package htlcswitch
 
 import (
+	"bytes"
+	"encoding/binary"
 	"errors"
 	"io"
+	"sync"
 
+	"github.com/coreos/bbolt"
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/multimutex"
 )
 
 var (
+
+	// networkResultStoreBucketKey is used for the root level bucket that
+	// stores the network result for each payment ID.
+	networkResultStoreBucketKey = []byte("network-result-store-bucket")
+
 	// ErrPaymentIDNotFound is an error returned if the given paymentID is
 	// not found.
 	ErrPaymentIDNotFound = errors.New("paymentID not found")
@@ -78,4 +88,173 @@ func deserializeNetworkResult(r io.Reader) (*networkResult, error) {
 	}
 
 	return n, nil
+}
+
+// networkResultStore is a persistent store that stores any results of HTLCs in
+// flight on the network. Since payment results are inherently asynchronous, it
+// is used as a common access point for senders of HTLCs, to know when a result
+// is back. The Switch will checkpoint any received result to the store, and
+// the store will keep results and notify the callers about them.
+type networkResultStore struct {
+	db *channeldb.DB
+
+	// results is a map from paymentIDs to channels where subscribers to
+	// payment results will be notified.
+	results    map[uint64][]chan *networkResult
+	resultsMtx sync.Mutex
+
+	// paymentIDMtx is a multimutex used to make sure the database and
+	// result subscribers map is consistent for each payment ID in case of
+	// concurrent callers.
+	paymentIDMtx *multimutex.Mutex
+}
+
+func newNetworkResultStore(db *channeldb.DB) *networkResultStore {
+	return &networkResultStore{
+		db:           db,
+		results:      make(map[uint64][]chan *networkResult),
+		paymentIDMtx: multimutex.NewMutex(),
+	}
+}
+
+// storeResult stores the networkResult for the given paymentID, and
+// notifies any subscribers.
+func (store *networkResultStore) storeResult(paymentID uint64,
+	result *networkResult) error {
+
+	// We get a mutex for this payment ID. This is needed to ensure
+	// consistency between the database state and the subscribers in case
+	// of concurrent calls.
+	store.paymentIDMtx.Lock(paymentID)
+	defer store.paymentIDMtx.Unlock(paymentID)
+
+	// Serialize the payment result.
+	var b bytes.Buffer
+	if err := serializeNetworkResult(&b, result); err != nil {
+		return err
+	}
+
+	var paymentIDBytes [8]byte
+	binary.BigEndian.PutUint64(paymentIDBytes[:], paymentID)
+
+	err := store.db.Batch(func(tx *bbolt.Tx) error {
+		networkResults, err := tx.CreateBucketIfNotExists(
+			networkResultStoreBucketKey,
+		)
+		if err != nil {
+			return err
+		}
+
+		return networkResults.Put(paymentIDBytes[:], b.Bytes())
+	})
+	if err != nil {
+		return err
+	}
+
+	// Now that the result is stored in the database, we can notify any
+	// active subscribers.
+	store.resultsMtx.Lock()
+	for _, res := range store.results[paymentID] {
+		res <- result
+	}
+	delete(store.results, paymentID)
+	store.resultsMtx.Unlock()
+
+	return nil
+}
+
+// subscribeResult is used to get the payment result for the given
+// payment ID. It returns a channel on which the result will be delivered when
+// ready.
+func (store *networkResultStore) subscribeResult(paymentID uint64) (
+	<-chan *networkResult, error) {
+
+	// We get a mutex for this payment ID. This is needed to ensure
+	// consistency between the database state and the subscribers in case
+	// of concurrent calls.
+	store.paymentIDMtx.Lock(paymentID)
+	defer store.paymentIDMtx.Unlock(paymentID)
+
+	var (
+		result     *networkResult
+		resultChan = make(chan *networkResult, 1)
+	)
+
+	err := store.db.View(func(tx *bbolt.Tx) error {
+		var err error
+		result, err = fetchResult(tx, paymentID)
+		switch {
+
+		// Result not yet available, we will notify once a result is
+		// available.
+		case err == ErrPaymentIDNotFound:
+			return nil
+
+		case err != nil:
+			return err
+
+		// The result was found, and will be returned immediately.
+		default:
+			return nil
+		}
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// If the result was found, we can send it on the result channel
+	// imemdiately.
+	if result != nil {
+		resultChan <- result
+		return resultChan, nil
+	}
+
+	// Otherwise we store the result channel for when the result is
+	// available.
+	store.resultsMtx.Lock()
+	store.results[paymentID] = append(
+		store.results[paymentID], resultChan,
+	)
+	store.resultsMtx.Unlock()
+
+	return resultChan, nil
+}
+
+// getResult attempts to immediately fetch the result for the given pid from
+// the store. If no result is available, ErrPaymentIDNotFound is returned.
+func (store *networkResultStore) getResult(pid uint64) (
+	*networkResult, error) {
+
+	var result *networkResult
+	err := store.db.View(func(tx *bbolt.Tx) error {
+		var err error
+		result, err = fetchResult(tx, pid)
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func fetchResult(tx *bbolt.Tx, pid uint64) (*networkResult, error) {
+	var paymentIDBytes [8]byte
+	binary.BigEndian.PutUint64(paymentIDBytes[:], pid)
+
+	networkResults := tx.Bucket(networkResultStoreBucketKey)
+	if networkResults == nil {
+		return nil, ErrPaymentIDNotFound
+	}
+
+	// Check whether a result is already available.
+	resultBytes := networkResults.Get(paymentIDBytes[:])
+	if resultBytes == nil {
+		return nil, ErrPaymentIDNotFound
+	}
+
+	// Decode the result we found.
+	r := bytes.NewReader(resultBytes)
+
+	return deserializeNetworkResult(r)
 }
--- a/htlcswitch/payment_result.go
+++ b/htlcswitch/payment_result.go
@@ -2,7 +2,9 @@ package htlcswitch
 
 import (
 	"errors"
+	"io"
 
+	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
@@ -45,4 +47,35 @@ type networkResult struct {
 	// isResolution indicates whether this is a resolution message, in
 	// which the failure reason might not be included.
 	isResolution bool
+}
+
+// serializeNetworkResult serializes the networkResult.
+func serializeNetworkResult(w io.Writer, n *networkResult) error {
+	if _, err := lnwire.WriteMessage(w, n.msg, 0); err != nil {
+		return err
+	}
+
+	return channeldb.WriteElements(w, n.unencrypted, n.isResolution)
+}
+
+// deserializeNetworkResult deserializes the networkResult.
+func deserializeNetworkResult(r io.Reader) (*networkResult, error) {
+	var (
+		err error
+	)
+
+	n := &networkResult{}
+
+	n.msg, err = lnwire.ReadMessage(r, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := channeldb.ReadElements(r,
+		&n.unencrypted, &n.isResolution,
+	); err != nil {
+		return nil, err
+	}
+
+	return n, nil
 }
--- a/htlcswitch/switch.go
+++ b/htlcswitch/switch.go
@@ -342,7 +342,7 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
 // result is received on the channel, the HTLC is guaranteed to no longer be in
 // flight. The switch shutting down is signaled by closing the channel. If the
 // paymentID is unknown, ErrPaymentIDNotFound will be returned.
-func (s *Switch) GetPaymentResult(paymentID uint64,
+func (s *Switch) GetPaymentResult(paymentID uint64, paymentHash lntypes.Hash,
 	deobfuscator ErrorDecrypter) (<-chan *PaymentResult, error) {
 
 	s.pendingMutex.Lock()
@@ -375,7 +375,7 @@ func (s *Switch) GetPaymentResult(paymentID uint64,
 
 		// Extract the result and pass it to the result channel.
 		result, err := s.extractResult(
-			deobfuscator, n, paymentID, payment.paymentHash,
+			deobfuscator, n, paymentID, paymentHash,
 		)
 		if err != nil {
 			e := fmt.Errorf("Unable to extract result: %v", err)
--- a/htlcswitch/switch.go
+++ b/htlcswitch/switch.go
@@ -64,16 +64,6 @@ var (
 	zeroPreimage [sha256.Size]byte
 )
 
-// pendingPayment represents the payment which made by user and waits for
-// updates to be received whether the payment has been rejected or proceed
-// successfully.
-type pendingPayment struct {
-	paymentHash lntypes.Hash
-	amount      lnwire.MilliSatoshi
-
-	resultChan chan *networkResult
-}
-
 // plexPacket encapsulates switch packet and adds error channel to receive
 // error from request handler.
 type plexPacket struct {
@@ -201,12 +191,12 @@ type Switch struct {
 	// service was initialized with.
 	cfg *Config
 
-	// pendingPayments stores payments initiated by the user that are not yet
-	// settled. The map is used to later look up the payments and notify the
-	// user of the result when they are complete. Each payment is given a unique
-	// integer ID when it is created.
-	pendingPayments map[uint64]*pendingPayment
-	pendingMutex    sync.RWMutex
+	// networkResults stores the results of payments initiated by the user.
+	// results. The store is used to later look up the payments and notify
+	// the user of the result when they are complete. Each payment attempt
+	// should be given a unique integer ID when it is created, otherwise
+	// results might be overwritten.
+	networkResults *networkResultStore
 
 	// circuits is storage for payment circuits which are used to
 	// forward the settle/fail htlc updates back to the add htlc initiator.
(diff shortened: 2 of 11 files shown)


channeldb/payments.go
```diff
@@ -103,6 +103,14 @@ type OutgoingPayment struct {
 // AddPayment saves a successful payment to the database. It is assumed that
 // all payment are sent using unique payment hashes.
 func (db *DB) AddPayment(payment *OutgoingPayment) error {
+	return db.Batch(func(tx *bbolt.Tx) error {
+		return db.AddPaymentTx(tx, payment)
+	})
+}
+
+// AddPaymentTx lets the caller call AddPayment with a given database tx to
+// ensure atimicity.
```
alexbosworth:
```suggestion
// ensure atomicity.
```

htlcswitch/pending_payment.go
```diff
@@ -441,3 +441,52 @@ func (store *pendingPaymentStore) fetchPendingPayments() ([]*PendingPayment,
...
+func (store *pendingPaymentStore) syncWithCircuitMap(
+	shouldDelete func(lnwire.ShortChannelID, uint64) bool) error {
+
+	return store.db.Update(func(tx *bbolt.Tx) error {
+		pendingPayments := tx.Bucket(pendingPaymentBucketKey)
+		if pendingPayments == nil {
+			return nil
+		}
+
+		pendingPayments.ForEach(func(k, v []byte) error {
```
Roasbeef: Should do:
```go
return pendingPayments.... 
```

htlcswitch/switch.go
```diff
@@ -1663,6 +1663,31 @@ func (s *Switch) Start() error {
 	}
 	s.blockEpochStream = blockEpochStream
 
+	// Before starting the HTLC forwarder, we sync our pending payment
+	// store with the circuit map. This must be done to ensure the store
+	// won't contain pending payments that was never forwarded on the
+	// network. By deleting those pending payments that are not found in
+	// the circuit map, we ensure that the router won't wait for a result
+	// to be returned for this payment ID.
+	// NOTE: this assumes that the circuit map has already been trimmed,
```
Roasbeef: I don't think this last assumption is sound with the current implementation. IIRC, the circuit map as is, can contain circuits for payments they have either already been completed or failed on the network. 

cc @cfromknecht

htlcswitch/pending_payment.go
```diff
@@ -0,0 +1,492 @@
...
+		return nil, err
+	}
+
+	return resultChan, nil
+}
+
+// fetchPendingPayments retrieves all payments from the store that are pending,
+// meaning no results are available yet, and can be re-forwarded on the
+// network.
+func (store *pendingPaymentStore) fetchPendingPayments() ([]*PendingPayment,
```
joostjager: Looks like this is test only code. Would look for a different way to assert behaviour (intercepting mock?)

htlcswitch/pending_payment.go
```diff
@@ -0,0 +1,492 @@
...
+//			return ErrPaymentIDNotFound
+//		}
+//
+//		return preimages.DeleteBucket(paymentIDBytes)
+//	})
+//}
+
+// getPayment is used to query the store for a pending payment for the given
+// payment ID. If the pid is not found, ErrPaymentIDNotFound will be returned.
+func (store *pendingPaymentStore) getPayment(pid uint64) (*PendingPayment,
```
joostjager: Is this function used?

htlcswitch/pending_payment.go
```diff
@@ -0,0 +1,492 @@
...
+		return nil, err
+	}
+
+	return ps, nil
+}
+
+// syncWithCircuitMap should be called at startup to delete the pending
+// payments which are no longer found in the circuit map. This removes pending
+// payments that was not successfully forwarded to the link before shutdown.
+func (store *pendingPaymentStore) syncWithCircuitMap(
```
joostjager: If we need to do this anyway, what is the use of persisting pending payments? Can't that list be reconstructed on startup from the circuit map?

So only persist the outcomes when they become available.

htlcswitch/pending_payment.go
```diff
@@ -0,0 +1,492 @@
...
+
+	if err != nil {
+		return err
+	}
+	return existingErr
+}
+
+// completePayment stores the PaymentResult for the given paymentID, and
+// notifies any subscribers.
+func (store *pendingPaymentStore) completePayment(paymentID uint64,
```
joostjager: Are those results left in forever?

htlcswitch/switch.go
```diff
@@ -973,10 +934,10 @@ func (s *Switch) parseFailedPayment(payment *pendingPayment, pkt *htlcPacket,
...
-	// If the provided payment is nil, we have discarded the error decryptor
-	// due to a restart. We'll return a fixed error and signal a temporary
-	// channel failure to the router.
-	case payment == nil:
+	// If the provided deobfuscator is nil, we have discarded the error
+	// decryptor due to a restart. We'll return a fixed error and signal a
+	// temporary channel failure to the router.
+	case deobfuscator == nil:
 		userErr := fmt.Sprintf("error decryptor for payment " +
 			"could not be located, likely due to restart")
```
joostjager: How is this still possible after this PR?

htlcswitch/payment_result.go
```diff
@@ -0,0 +1,268 @@
+package htlcswitch
```
joostjager: Rename file to `payment_result_store.go`, as we do with other stores too?

htlcswitch/payment_result.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// If there's no error, write 'false' and return.
+	if p.Error == nil {
+		return channeldb.WriteElements(w, false)
+	}
+
+	// Otherwise, write 'true' and the error fields.
+	msg := []byte(p.Error.ExtraMsg)
+	err := channeldb.WriteElements(w, true,
```
joostjager: Saving the human readable error `ExtraMsg` to the db is not ideal. There are only a few cases for `ExtraMsg`, those can probably be saved in a structured way.

htlcswitch/payment_result.go
```diff
@@ -0,0 +1,268 @@
...
+		p.Error.ErrorSource, msg,
```
joostjager: `ErrorSource` could be `nil` if the error source is unknown (for example after restart or when the onion decrypt fails). The store should probably prepare for that, to prevent a db migration.

htlcswitch/payment_result.go
joostjager: Upon further inspection, there may be a few more. But still doubt whether we should persist data just to get a nice error message. Maybe just logging when the error occurs is enough.

htlcswitch/payment_result.go
```diff
@@ -104,3 +114,155 @@ func deserializePaymentResult(r io.Reader) (*PaymentResult, error) {
...
+			return err
+		}
+
+		return paymentResults.Put(paymentIDBytes[:], b.Bytes())
+	})
+	if err != nil {
+		return err
+	}
+
+	// Now that the result is stored in the database, we can notify any
```
joostjager: If the subscribe happens at this point, there will be a duplicate notification?

htlcswitch/switch.go
```diff
@@ -205,12 +190,18 @@ type Switch struct {
...
+	// results.  The store is used to later look up the payments and notify
+	// the user of the result when they are complete. Each payment attempt
+	// should be given a unique integer ID when it is created, and only
+	// added once.
+	paymentResults *paymentResultStore
+
+	// deobfuscators is a map storing error decrypters we need to parse
+	// received payment failures. We don't store these in the database, as
+	// being unable to decrypt an error after restart is not critical.
+	deobfuscators    map[uint64]ErrorDecrypter
```
joostjager: How consistent is it that we encode the onion in the router and decode it in the switch? I tend to think it may be better to also move the decoding to the router.

The router already has the route stored, so to decode the failure it would only need to store the session key as an extra field in the control tower db.

This will also fix the problem that we can't decode a failure after restart of the switch.

htlcswitch/switch.go
```diff
@@ -345,15 +337,26 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+		HtlcID: paymentID,
+	}
 
-	if !ok {
-		return nil, ErrPaymentIDNotFound
+	// If the payment is not found in the circuit map, check whether a
+	// result is already available.
+	// Assumption: no one will add this payment ID other than the caller.
+	if s.circuits.LookupCircuit(outKey) == nil {
+		res, err := s.paymentResults.getPaymentResult(paymentID)
```
joostjager: `subscribePaymentResult` also returns the current state immediately, so is it necessary to have this separate block calling `getPaymentResult`?

htlcswitch/switch.go
```diff
@@ -848,31 +848,37 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
...
+
+	// Deliver the payment error and preimage to the application, if it is
+	// waiting for a response.
+	if payment != nil {
+		payment.resultChan <- result
+		s.removePendingPayment(pkt.incomingHTLCID)
+	}
+}
+
+func (s *Switch) extractResult(payment *pendingPayment,
```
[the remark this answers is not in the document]
halseth: Fixed.

htlcswitch/payment_result.go
halseth: Yeah, looks like they are only used locally for passing the case of error to the RPC. Would it be useful to define a field for `local error codes` that can be used to communicate local failures?

htlcswitch/payment_result.go
halseth: Using `nil` can leads to problems down the line, what do you think about defining a constant `UnknownPaymentSource`?

htlcswitch/payment_result.go
joostjager: I think that local internal errors (things that should never happen) don't need to be exposed over rpc. Logging and returning a generic internal error should be enough?

htlcswitch/payment_result.go
joostjager: What kind of problems? `nil` for unset fields is quite common? For serialize to the db we can of course write anything we want.

htlcswitch/switch.go
halseth: This is meant to handle the case where the paymentID is unknown, as we want to return an error immediately then, while `subscribePaymentResult` will wait for a result to appear.

htlcswitch/switch_test.go
```diff
@@ -1830,10 +1830,6 @@ func TestSwitchSendPayment(t *testing.T) {
 		t.Fatal("request was not propagated to destination")
 	}
 
-	if s.numPendingPayments() != 1 {
```
Roasbeef: Why remove rather than modify to check the new persistent store?

htlcswitch/payment_result.go
```diff
@@ -90,3 +99,155 @@ func deserializePaymentResult(r io.Reader) (*PaymentResult, error) {
...
+	// active subscribers.
+	store.resultsMtx.Lock()
+	for _, res := range store.results[paymentID] {
+		res <- result
```
Roasbeef: Expectation is that these channels are always buffered?

htlcswitch/payment_result.go
```diff
@@ -90,3 +99,155 @@ func deserializePaymentResult(r io.Reader) (*PaymentResult, error) {
...
+		resultChan <- result
+	}
+
+	return resultChan, nil
+}
+
+func (store *paymentResultStore) getPaymentResult(pid uint64) (
+	*PaymentResult, error) {
+
+	var result *PaymentResult
```
Roasbeef: Can combine under a single `var` statement.

htlcswitch/payment_result.go
```diff
@@ -90,3 +99,155 @@ func deserializePaymentResult(r io.Reader) (*PaymentResult, error) {
...
+	var err error
+
+	err = store.db.View(func(tx *bbolt.Tx) error {
+		result, err = fetchPaymentResult(tx, pid)
```
Roasbeef: Error being over shadowed here?

htlcswitch/switch.go
```diff
@@ -385,12 +368,7 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
 		htlc:           htlc,
 	}
 
-	if err := s.forward(packet); err != nil {
-		s.removePendingPayment(paymentID)
-		return err
-	}
-
-	return nil
+	return s.forward(packet)
```
Roasbeef: When do pending payments need to be removed due to failures? Never now under the new assumptions?

lnd_test.go
```diff
@@ -13939,6 +13941,259 @@ func testChannelBackupRestore(net *lntest.NetworkHarness, t *harnessTest) {
...
+		if err != nil {
+			t.Fatalf("unable to generate preimage: %v", err)
+		}
+
+		preimages[preimage] = struct{}{}
+	}
+
+	// Let Carol create hold-invoices for all the payments.
+	payAmt := btcutil.Amount(4)
+	var payReqs []string
```
Roasbeef: `var` statements can be unified under a single one

lnd_test.go
```diff
@@ -13939,6 +13941,259 @@ func testChannelBackupRestore(net *lntest.NetworkHarness, t *harnessTest) {
...
+		}
+
+		return nil
+	}, time.Second*15)
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	// TODO(halseth): add a failed payment.
+	// TODO(halseth): check payment stream when got idempotent payments.
```
Roasbeef: Yeah should definitely be done before this lands.

htlcswitch/payment_result.go
halseth: Fixed by using multimutex on the payment ID.

htlcswitch/payment_result.go
halseth: Yes, they are always created with a 1 element buffer, and deleted after they are first sent on.

htlcswitch/switch.go
halseth: Fixed by #3063

htlcswitch/switch_test.go
halseth: No concept of pending payments anymore.

htlcswitch/payment_result.go
halseth: Fixed.

lnd_test.go
halseth: Done.

htlcswitch/payment_result.go
halseth: Hm, not sure, but fixed to be sure.

halseth:
PTAL @joostjager @Roasbeef

Roasbeef:
Can be rebased now that https://github.com/lightningnetwork/lnd/pull/2761 is in.

halseth:
Rebased.

Roasbeef:
First commit can be removed during the next rebase.

htlcswitch/payment_result.go
```diff
@@ -46,3 +48,34 @@ type networkResult struct {
...
+	)
+
+	n := &networkResult{}
+
+	n.msg, err = lnwire.ReadMessage(r, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := channeldb.ReadElements(r,
```
Roasbeef: `r` on next line?

htlcswitch/payment_result.go
```diff
@@ -1,14 +1,24 @@
...
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/multimutex"
 )
 
 var (
+
+	// networkResultStoreBucketKey is used for the root level bucket that
+	// stores the network result for each payment ID.
+	networkResultStoreBucketKey = []byte("network-result-store-bucket")
```
Roasbeef: Such bucket name, wow.

htlcswitch/payment_result.go
```diff
@@ -79,3 +89,172 @@ func deserializeNetworkResult(r io.Reader) (*networkResult, error) {
...
+	results    map[uint64][]chan *networkResult
+	resultsMtx sync.Mutex
+
+	// paymentIDMtx is a multimutex used to make sure the database and
+	// result subscribers map is consistent for each payment ID in case of
+	// concurrent callers.
+	paymentIDMtx *multimutex.Mutex
+}
+
+func newNetworkResultStore(db *channeldb.DB) *networkResultStore {
```
Roasbeef: Missing minimal `godoc` comment.

htlcswitch/payment_result.go
```diff
@@ -79,3 +89,172 @@ func deserializeNetworkResult(r io.Reader) (*networkResult, error) {
...
+	// Serialize the payment result.
+	var b bytes.Buffer
+	if err := serializeNetworkResult(&b, result); err != nil {
+		return err
+	}
+
+	var paymentIDBytes [8]byte
+	binary.BigEndian.PutUint64(paymentIDBytes[:], paymentID)
+
+	err := store.db.Batch(func(tx *bbolt.Tx) error {
```
Roasbeef: May want to consider making this an update instead, if only for the sake of the integration tests. In the past, then extra 30ms or so jitter here really accumulated making many tests time out when we tried `Batch`ing everywhere before.

htlcswitch/switch.go
```diff
@@ -345,12 +335,33 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+		if err != nil {
+			return nil, err
+		}
+		c := make(chan *networkResult, 1)
+		c <- res
+		nChan = c
+	} else {
+		// The payment was committed to the circuits, subscribe for a
+		// result.
+		nChan, err = s.networkResults.subscribeResult(paymentID)
```
Roasbeef: With this new approach, it seems that a caller that passes in an unknown `paymentID` will block indefinitely, rather than receiving an immediate error. I'm guessing it was done this way in order to avoid storing mostly duplicate data here and also in the router?

halseth: If the payment ID is unknown, `s.circuits.LookupCircuit(outKey) == nil` above will be true, and we attempt to fetch the result directly. `ErrUnknownPaymetnID` will be returned.

htlcswitch/payment_result.go
halseth: This matches the style used in many other parts of the codebase

htlcswitch/payment_result.go
halseth: Yeah, I'm def in favor of using `Update` as long as we don't have any clear indication that it is a performance bottleneck.

cfromknecht: if we do decide to use `Update`, then we probably don't need a multimutex since the db txns can't execute concurrently. `storeResult` is called in separate goroutines, so under heavy load i suspect `Batch` would help, but would need to benchmark to be certain

htlcswitch/switch.go
```diff
@@ -989,6 +976,7 @@ func (s *Switch) parseFailedPayment(deobfuscator ErrorDecrypter,
 	// the first hop. In this case, we'll report a permanent
 	// channel failure as this means us, or the remote party had to
 	// go on chain.
+	// TODO: check reason length instead
```
cfromknecht: what is meant by this/when do we plan to do this?

htlcswitch/switch_test.go
```diff
@@ -2125,3 +2126,115 @@ func TestUpdateFailMalformedHTLCErrorConversion(t *testing.T) {
...
+	// store.
+	lookup <- nil
+	resultChan, err = s.GetPaymentResult(
+		paymentID, lntypes.Hash{}, newMockDeobfuscator(),
+	)
+	if err != nil {
+		t.Fatalf("unable to get payment result: %v", err)
+	}
+
+	select {
```
cfromknecht: not blocking, but this segment is an exact copy of the logic above, maybe add a `checkResult` closure?

cfromknecht:
Really happy with how this refactor turned out, this method of storing the payment results i think is much cleaner than some of the avenues we explored earlier! Will need rebase now that #3087 has landed

lnd_test.go
```diff
@@ -13099,6 +13101,262 @@ func testChannelBackupRestore(net *lntest.NetworkHarness, t *harnessTest) {
...
+		if err != nil {
+			t.Fatalf("unable to subscribe to invoice: %v", err)
+		}
+
+		invoiceStreams = append(invoiceStreams, stream)
+		payReqs = append(payReqs, resp.PaymentRequest)
+	}
+
+	// Wait for all the invoices to reach the OPEN state.
+	for _, stream := range invoiceStreams {
```
cfromknecht: at some point these helper routines for hodl invoices could be extracted, but not blocker on this PR

htlcswitch/switch.go
```diff
@@ -856,15 +844,34 @@ func (s *Switch) handleLocalDispatch(pkt *htlcPacket) error {
...
+	paymentID := pkt.incomingHTLCID
+
+	// The error reason will be unencypted in case this a local
+	// failure or a converted error.
+	unencrypted := pkt.localFailure || pkt.convertedError
+	n := &networkResult{
+		msg:          pkt.htlc,
+		unencrypted:  unencrypted,
+		isResolution: pkt.isResolution,
+	}
```
joostjager: Add timestamp to record when failure was received. This is one building block of black hole defense

joostjager: Preferably `time.UnixNano` for sub-second resolution

Roasbeef:
Can now be updated to use the latest RPC calls in the new sub-server.

htlcswitch/switch.go
halseth: Removed.

halseth:
Alright, itest using the new payment API is pushed. PTAL @joostjager

Issue #271, opened 2017-08-27
Fuzz the wire parser in the lnwire package


Roasbeef:
# Issue

Currently, within the `lnwire` package, we have a set that uses the [`testing/quick`](https://golang.org/pkg/testing/quick/) package to create a set of block-block tests that use the quickcheck testing paradigm to exercise our wire parser. This set of tests uses property based testing to ensure that the parser is able to adhere to the [`mainScenario`](https://github.com/lightningnetwork/lnd/blob/master/lnwire/lnwire_test.go#L95), amongst other things, this ensures that we'll always be able to consume a message that we produced, ensuring correctness of our wire parsing implementation. 

This set of tests are nice, but insufficient. What we _really_ want to ensure (for safety/security properties) is that: no sequence or random bytes, partially correct wire encodings, or mutated wire encodings case `lnd` to _crash_. We
ll need to ensure that under the presence of arbitrary malicious input, `lnd` partially parses, them ignores them if invalid, and keeps on trucking. This can be achieved by what's known as [fuzz testing](https://en.wikipedia.org/wiki/Fuzzing). 

There exists a popular fuzz-testing library written in golang, called [`go-fuzz`](https://github.com/dvyukov/go-fuzz), we can use this to being the process of first fuzzing encoding/decoding code within the `lnwire` package. In the future, we can expand our usage to cover other relevant areas of the codebase. 

# Steps To Completion 

- [x] Get familiar with how the `go-fuzz` package is used by [reading the documentation](https://github.com/dvyukov/go-fuzz). 

- [x] Create an applicable fuzz function within the `lnwire` package. This may closely resemble the `mainScenario` used within the tests, but should call `panic` rather than failing the test. 

- [x] Next, gather an initial testing corpus. This should consist of known valid serialized wire messages. At the very least, there should be an entry for _each_ of the known wire messages defined within the current [specification](https://github.com/lightningnetwork/lnd/blob/master/lnwire/message.go#L23). A diversity of input is desirable, so these can be obtained by either adding some code to `lnd` to dump valid messages into a file, or by generating them manually. 

- [x] Execute the fuzzer with the defined fuzz function, and corpus. 

- [x] Monitor the execution of the fuzzer, noting which inputs cause the fuzz function to panic. Examine the inputs, and the current parsing code, creating PR's fixing the panic for each input found. 

- [x] ????

- [x] Profit 🎉

Crypt-iQ:
[...]

PR #2770, opened 2019-03-13
Pass context.Context to GetUtxo

halseth:
Alternative to #2716

### DIFF
--- a/lnwallet/wallet.go
+++ b/lnwallet/wallet.go
@@ -2,6 +2,7 @@ package lnwallet
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"errors"
 	"fmt"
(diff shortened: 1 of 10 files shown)


lnwallet/wallet.go
```diff
@@ -972,10 +973,26 @@ func (l *LightningWallet) handleFundingCounterPartySigs(msg *addCounterPartySigs
 			)
 			if err != nil {
 			}
+
+			ctx, cancel := context.WithCancel(context.Background())
```
cfromknecht: since our goal is simply to break on shutdown, we should only need to make one `context.Context` for the whole `LightningWallet`. the `ctx` and `cancel` can be tracked as member variables, and `cancel` only needs to invoked when `Stop`is called. similar comment applies for router

halseth: We could, but that seems like a big deviation from our normal pattern? I like sticking with the simple `cancel chan` tbh.

cfromknecht: what's the normal pattern? having one context for all concurrent wallet/router `GetUtxo` requests seems more in line with how we would otherwise pass the object's `quit` channel, like in #2716?

wpaulino: I think having a `context.Background` makes sense for the struct itself. We can then use specific deadlines/cancels by creating a context specific for that call, like we do with the integration tests.

cfromknecht: > I like sticking with the simple cancel chan tbh.

Do you mean `quit` chans or `cancel()` funcs?

> I think having a context.Background makes sense for the struct itself. We can then use specific deadlines/cancels by creating a context specific for that call, like we do with the integration tests.

Agreed, though it's not necessary to create a cancellable context for each call when our only need is to cancel them all simultaneously. If in the future we want to be able to cancel/timeout individual calls, then creating a subcontext would be appropriate. In the meantime it is extra goroutine overhead

atm i'm kind of in favor of #2716 for simplicity, it matches the existing patterns w/in the codebase, and accomplishes the same task w/ less overhead

halseth: > Do you mean quit chans or cancel() funcs?

I meant as in #2716 

> what's the normal pattern?

That is " it matches the existing patterns w/in the codebase", and I'm also in favor of #271 for now.

cfromknecht:
Closing since #2716 was merged

PR #2773, opened 2019-03-13
channeldb: convert all Update calls to use Batch

Roasbeef:
In this commit, we convert all the `Update` calls which are serial, to
use `Batch` calls which are optimistically batched together for
concurrent writers. This should increase performance slightly during the
initial graph sync, and also updates at tip as we can coalesce more of
these individual transactions into a single transaction.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -1027,7 +1027,7 @@ func (c *OpenChannel) UpdateCommitment(newCommitment *ChannelCommitment) error {
 		return ErrNoRestoredChannelMutation
 	}
 
-	err := c.Db.Update(func(tx *bbolt.Tx) error {
+	err := c.Db.Batch(func(tx *bbolt.Tx) error {
 		chanBucket, err := fetchChanBucket(
 			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
 		)
@@ -1465,7 +1465,7 @@ func (c *OpenChannel) AppendRemoteCommitChain(diff *CommitDiff) error {
 		return ErrNoRestoredChannelMutation
 	}
 
-	return c.Db.Update(func(tx *bbolt.Tx) error {
+	return c.Db.Batch(func(tx *bbolt.Tx) error {
 		// First, we'll grab the writable bucket where this channel's
 		// data resides.
 		chanBucket, err := fetchChanBucket(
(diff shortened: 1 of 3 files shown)


channeldb/graph.go
```diff
@@ -697,7 +697,7 @@ func (c *ChannelGraph) PruneGraph(spentOutputs []*wire.OutPoint,
 
 	var chansClosed []*ChannelEdgeInfo
 
-	err := c.db.Update(func(tx *bbolt.Tx) error {
+	err := c.db.Batch(func(tx *bbolt.Tx) error {
```
cfromknecht: `chansClosed` needs to be nilled during each invocation

channeldb/graph.go
```diff
@@ -946,7 +946,7 @@ func (c *ChannelGraph) DisconnectBlockAtHeight(height uint32) ([]*ChannelEdgeInf
 	// Keep track of the channels that are removed from the graph.
 	var removedChans []*ChannelEdgeInfo
 
-	if err := c.db.Update(func(tx *bbolt.Tx) error {
+	if err := c.db.Batch(func(tx *bbolt.Tx) error {
```
cfromknecht: sam here for `removedChans`

channeldb/graph.go
Roasbeef: Nice catch! Fixed here and below, reminds me to do a quick sweep to see if we have any other possible bugs here due to the potential of batch commit retries.

channeldb/invoices.go
```diff
@@ -654,7 +654,7 @@ func (d *DB) SettleInvoice(paymentHash [32]byte,
 // payment hash.
 func (d *DB) CancelInvoice(paymentHash lntypes.Hash) (*Invoice, error) {
 	var canceledInvoice *Invoice
-	err := d.Update(func(tx *bbolt.Tx) error {
+	err := d.Batch(func(tx *bbolt.Tx) error {
```
cfromknecht: all changes in this file are side-effect prone

channeldb/channel.go
```diff
@@ -1608,7 +1608,7 @@ func (c *OpenChannel) AdvanceCommitChainTail(fwdPkg *FwdPkg) error {
 
 	var newRemoteCommit *ChannelCommitment
 
-	err := c.Db.Update(func(tx *bbolt.Tx) error {
+	err := c.Db.Batch(func(tx *bbolt.Tx) error {
```
cfromknecht: side-effects

halseth:
To avoid unintended side-effects, maybe we should only change calls known to be performance critical for now?

Roasbeef:
PTAL @cfromknecht

PR #2775, opened 2019-03-13
 discovery: check if stale within isMsgStale for ChannelUpdate messages

wpaulino:
In this commit, we address an assumption of the gossiper's recently introduce reliable sender. The reliable sender is currently only used for messages of unannounced channels. This makes sense as peers should be able to retrieve messages from the network if they've previously announced. However, within isMsgStale, we assumed that the reliable sender would be used for every ChannelUpdate being sent, even if the channel is already announced. Due to this, checking if the policy is stale was unnecessary. But since this isn't the case, we should actually be checking whether it is stale to prevent sending it later on.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -221,7 +221,10 @@ type AuthenticatedGossiper struct {
 	peerSyncers map[routing.Vertex]*gossipSyncer
 
 	// reliableSender is a subsystem responsible for handling reliable
-	// message send requests to peers.
+	// message send requests to peers. This should only be used for channels
+	// that are unadvertised at the time of handling the message since if it
+	// is advertised, then peers should be able to get the message from the
+	// network.
 	reliableSender *reliableSender
 
 	sync.Mutex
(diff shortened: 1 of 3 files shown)


discovery/gossiper.go
```diff
@@ -2321,16 +2324,10 @@ func (d *AuthenticatedGossiper) isMsgStale(msg lnwire.Message) bool {
...
-		// never be stale. We should still however check if the channel
-		// is part of our graph. If it's not, we can mark it as stale.
-		_, _, _, err := d.cfg.Router.GetChannelByID(msg.ShortChannelID)
-		if err != nil && err != channeldb.ErrEdgeNotFound {
-			log.Debugf("Unable to retrieve channel=%v from graph: "+
-				"%v", err)
-		}
-		return err == channeldb.ErrEdgeNotFound
+		timestamp := time.Unix(int64(msg.Timestamp), 0)
+		return d.cfg.Router.IsStaleEdgePolicy(
```
halseth: are we sure we will never overwrite the policy with a new one before we have been able to send it to the peer?

wpaulino: Even if we overwrite it, then it means we have a newer version which we should be sending. In this case, we'll only reliably send if the channel is still unannounced, because we assume any peer can receive the message as long as the channel is announced.

discovery/gossiper_test.go
```diff
@@ -2782,10 +2781,151 @@ func TestSendChannelUpdateReliably(t *testing.T) {
...
+			assertMessage(t, batch.localProofAnn, msg)
+		default:
+			t.Fatalf("send unexpected %v message", msg.MsgType())
+		}
+	}
+
+	// Since the messages above are now deemed as stale, they should be
+	// removed from the message store.
+	var predErr error
+	err = lntest.WaitPredicate(func() bool {
```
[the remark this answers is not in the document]

discovery/reliable_sender.go
```diff
@@ -181,9 +181,19 @@ out:
...
-		case <-peerMgr.msgs:
+		case msg := <-peerMgr.msgs:
+			// Retrieve the short channel ID for which this message
+			// applies for logging purposes. The error can be
+			// ignored as the store can only contain messages which
+			// have a ShortChannelID field.
+			shortChanID, _ := msgShortChanID(msg)
+			log.Debugf("Received request to send %v message for "+
+				"channel=%v while peer=%x is offline",
+				shortChanID, msg.MsgType(), peerPubKey)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

discovery/gossiper.go
halseth: The scenario I'm thinking about is that the channel stay unannounced (private?) and we for some reason create a new edge policy that is added to the graph (but not to the message store!). In this scenario the policy found in the message store is considered stale.

We need to make sure we either add the updated policy to the message store in this case, or keep the update around until the channel has been announced.

wpaulino: Not sure I follow. If the edge for the policy that we're applying is still unannounced, then we'll also add it to the message store. See:
https://github.com/lightningnetwork/lnd/blob/e3c65e1a4935dbf9ba91cb326ea78b5161748dad/discovery/gossiper.go#L1961

discovery/reliable_sender.go
```diff
@@ -225,6 +225,14 @@ out:
 		// can only contain messages which have a ShortChannelID field.
 		shortChanID, _ := msgShortChanID(msg)
 
+		// Ensure the peer is still online right before sending the
+		// message.
+		select {
+		case <-offlineChan:
```
Roasbeef: Why's this needed, if we go to send the message below, we'll get an error and go back into `waitUntilOnline` anyway. Or is there a possibility of deadlock?

wpaulino: The send below is asynchronous, which I believe won't return an error back if the peer disconnects.

wpaulino: Could alternatively make the send synchronous instead.

discovery/gossiper.go
halseth: Resolved offline.

PR #2777, opened 2019-03-13
channeldb+routing+discovery: reject zombie announcements

wpaulino:
In this PR, we introduce a new persistent index for zombie edges to quickly reject announcements for edges we've previously deemed as zombies. Care has been taken to ensure we don't reject fresh updates for edges we've considered zombies.

### DIFF
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -123,9 +123,6 @@ var (
 	// case we'll remove all entries from the prune log with a block height
 	// that no longer exists.
 	pruneLogBucket = []byte("prune-log")
-
-	edgeBloomKey = []byte("edge-bloom")
-	nodeBloomKey = []byte("node-bloom")
 )
 
 const (
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -595,17 +595,20 @@ func (c *ChannelGraph) addChannelEdge(tx *bbolt.Tx, edge *ChannelEdgeInfo) error
 // HasChannelEdge returns true if the database knows of a channel edge with the
 // passed channel ID, and false otherwise. If an edge with that ID is found
 // within the graph, then two time stamps representing the last time the edge
-// was updated for both directed edges are returned along with the boolean.
-func (c *ChannelGraph) HasChannelEdge(chanID uint64) (time.Time, time.Time, bool, error) {
-	// TODO(roasbeef): check internal bloom filter first
+// was updated for both directed edges are returned along with the boolean. If
+// it is not found, then the zombie index is checked and its result is returned
+// as the second boolean.
+func (c *ChannelGraph) HasChannelEdge(chanID uint64,
+) (time.Time, time.Time, bool, bool, error) {
 
 	var (
 		node1UpdateTime time.Time
 		node2UpdateTime time.Time
 		exists          bool
+		isZombie        bool
 	)
 
-	if err := c.db.View(func(tx *bbolt.Tx) error {
+	err := c.db.View(func(tx *bbolt.Tx) error {
 		edges := tx.Bucket(edgeBucket)
 		if edges == nil {
 			return ErrGraphNoEdgesFound
@@ -617,12 +620,21 @@ func (c *ChannelGraph) HasChannelEdge(chanID uint64) (time.Time, time.Time, bool
 
 		var channelID [8]byte
 		byteOrder.PutUint64(channelID[:], chanID)
+
+		// If the edge doesn't exist, then we'll also check our zombie
+		// index.
 		if edgeIndex.Get(channelID[:]) == nil {
 			exists = false
+			zombieIndex := edges.Bucket(zombieBucket)
+			if zombieIndex != nil {
+				isZombie, _, _ = isZombieEdge(zombieIndex, chanID)
+			}
+
 			return nil
 		}
 
 		exists = true
+		isZombie = false
 
 		// If the channel has been found in the graph, then retrieve
 		// the edges itself so we can return the last updated
@@ -648,11 +660,9 @@ func (c *ChannelGraph) HasChannelEdge(chanID uint64) (time.Time, time.Time, bool
 		}
 
 		return nil
-	}); err != nil {
-		return time.Time{}, time.Time{}, exists, err
-	}
+	})
 
-	return node1UpdateTime, node2UpdateTime, exists, nil
+	return node1UpdateTime, node2UpdateTime, exists, isZombie, err
 }
 
 // UpdateChannelEdge retrieves and update edge of the graph database. Method
@@ -2537,7 +2547,8 @@ func (c *ChannelEdgePolicy) Signature() (*btcec.Signature, error) {
 // found, then ErrEdgeNotFound is returned. A struct which houses the general
 // information for the channel itself is returned as well as two structs that
 // contain the routing policies for the channel in either direction.
-func (c *ChannelGraph) FetchChannelEdgesByOutpoint(op *wire.OutPoint) (*ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy, error) {
+func (c *ChannelGraph) FetchChannelEdgesByOutpoint(op *wire.OutPoint,
+) (*ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy, error) {
 
 	var (
 		edgeInfo *ChannelEdgeInfo
@@ -2615,7 +2626,12 @@ func (c *ChannelGraph) FetchChannelEdgesByOutpoint(op *wire.OutPoint) (*ChannelE
 // ErrEdgeNotFound is returned. A struct which houses the general information
 // for the channel itself is returned as well as two structs that contain the
 // routing policies for the channel in either direction.
-func (c *ChannelGraph) FetchChannelEdgesByID(chanID uint64) (*ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy, error) {
+//
+// ErrZombieEdge an be returned if the edge is currently marked as a zombie
+// within the database. In this case, the ChannelEdgePolicy's will be nil, and
+// the ChannelEdgeInfo will only include the public keys of each node.
+func (c *ChannelGraph) FetchChannelEdgesByID(chanID uint64,
+) (*ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy, error) {
 
 	var (
 		edgeInfo  *ChannelEdgeInfo
@@ -2646,13 +2662,48 @@ func (c *ChannelGraph) FetchChannelEdgesByID(chanID uint64) (*ChannelEdgeInfo, *
 
 		byteOrder.PutUint64(channelID[:], chanID)
 
+		// Now, attempt to fetch edge.
 		edge, err := fetchChanEdgeInfo(edgeIndex, channelID[:])
+
+		// If it doesn't exist, we'll quickly check our zombie index to
+		// see if we've previously marked it as so.
+		if err == ErrEdgeNotFound {
+			// If the zombie index doesn't exist, or the edge is not
+			// marked as a zombie within it, then we'll return the
+			// original ErrEdgeNotFound error.
+			zombieIndex := edges.Bucket(zombieBucket)
+			if zombieIndex == nil {
+				return ErrEdgeNotFound
+			}
+
+			isZombie, pubKey1, pubKey2 := isZombieEdge(
+				zombieIndex, chanID,
+			)
+			if !isZombie {
+				return ErrEdgeNotFound
+			}
+
+			// Otherwise, the edge is marked as a zombie, so we'll
+			// populate the edge info with the public keys of each
+			// party as this is the only information we have about
+			// it and return an error signaling so.
+			edgeInfo = &ChannelEdgeInfo{
+				NodeKey1Bytes: pubKey1,
+				NodeKey2Bytes: pubKey2,
+			}
+			return ErrZombieEdge
+		}
+
+		// Otherwise, we'll just return the error if any.
 		if err != nil {
 			return err
 		}
+
 		edgeInfo = &edge
 		edgeInfo.db = c.db
 
+		// Then we'll attempt to fetch the accompanying policies of this
+		// edge.
 		e1, e2, err := fetchChanEdgePolicies(
 			edgeIndex, edges, nodes, channelID[:], c.db,
 		)
@@ -2664,6 +2715,9 @@ func (c *ChannelGraph) FetchChannelEdgesByID(chanID uint64) (*ChannelEdgeInfo, *
 		policy2 = e2
 		return nil
 	})
+	if err == ErrZombieEdge {
+		return edgeInfo, nil, nil, err
+	}
 	if err != nil {
 		return nil, nil, nil, err
 	}
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -728,6 +728,10 @@ func (c *ChannelGraph) PruneGraph(spentOutputs []*wire.OutPoint,
 		if nodes == nil {
 			return ErrSourceNodeNotSet
 		}
+		zombieIndex, err := edges.CreateBucketIfNotExists(zombieBucket)
+		if err != nil {
+			return err
+		}
 
 		// For each of the outpoints that have been spent within the
 		// block, we attempt to delete them from the graph as if that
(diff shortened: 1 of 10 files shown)


discovery/gossiper.go
```diff
@@ -1787,9 +1803,37 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
+				nMsg.err <- fmt.Errorf("rejecting "+
+					"ChannelUpdate(%v) for zombie channel",
+					msg.ShortChannelID)
+				return nil
+			}
+
+			// Otherwise, we'll mark the edge as live within the
+			// router to ensure we can process any future updates
+			// for this edge until it is considered zombie again.
+			err := d.cfg.Router.MarkEdgeLive(msg.ShortChannelID)
```
cfromknecht: we should defer marking an edge as live only after we are sure  the signature validates, otherwise someone could blast updates w/ fake signatures that bring about a zombie uprising

discovery/gossiper.go
```diff
@@ -1787,9 +1803,37 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 		}
 
 		// Before we perform any of the expensive checks below, we'll
-		// make sure that the router doesn't already have a fresher
-		// announcement for this edge.
+		// make sure that we don't process any messages for known zombie
+		// edges.
 		timestamp := time.Unix(int64(msg.Timestamp), 0)
+		if d.cfg.Router.IsZombieEdge(msg.ShortChannelID) {
```
cfromknecht: similar comment here with unifying `IsZombieEdge` w/ `IsStaleEdgePolicy` to avoid adding another db tx

routing/router.go
```diff
@@ -435,6 +423,29 @@ func (r *ChannelRouter) Start() error {
 		}
 	}
 
+	// If AssumeChannelValid is present, then we'll only prune channels once
+	// we detect them as zombies.
```
cfromknecht: comment doesn't seem to match code, this looks like we prune them immediately

cfromknecht: i see now, would update comment to say if assume channel valid is present, we won't start a chain view to track spentness or something

wpaulino: Fixed.

discovery/gossiper.go
wpaulino: Fixed.

discovery/gossiper.go
wpaulino: Since the `GetChannelByID` call below will fail due to it being a zombie channel, we'll go ahead and stash it so that it gets reprocessed once we receive its corresponding channel announcement. Without the announcement, we're unable to verify signatures since we don't have the public keys of each edge. If and when we receive the announcement, we won't be able to re-process it because it'll still be marked as a zombie.

discovery/gossiper.go
```diff
@@ -1571,8 +1579,16 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 			return nil
 		}
 
-		// At this point, we'll now ask the router if this is a stale
-		// update. If so we can skip all the processing below.
+		// At this point, we'll now ask the router if this is a
+		// zombie/known edge. If so we can skip all the processing
+		// below. In the case of a zombie edge, we'll reject it
+		// completely.
+		if d.cfg.Router.IsZombieEdge(msg.ShortChannelID) {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

channeldb/graph.go
```diff
@@ -2778,6 +2784,68 @@ func (c *ChannelGraph) NewChannelEdgePolicy() *ChannelEdgePolicy {
...
+		return zombieIndex.Delete(k[:])
+	})
+}
+
+// IsZombieEdge returns whether the edge is considered zombie.
+func (c *ChannelGraph) IsZombieEdge(chanID uint64) bool {
+	isZombie := false
+
+	// It's safe to ignore the error being returned, if any, as we'll just
+	// return false anyway.
```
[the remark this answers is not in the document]
wpaulino: Fixed.

routing/router.go
```diff
@@ -734,10 +734,17 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
 			chanToPrune)
 
-		err := r.cfg.Graph.DeleteChannelEdge(&chanToPrune)
+		err := r.cfg.Graph.MarkEdgeZombie(chanToPrune.ChannelID)
+		if err != nil {
+			return fmt.Errorf("unable to mark ChannelPoint(%v) "+
+				"as zombie: %v", chanToPrune.ChannelPoint, err)
+		}
+
+		err = r.cfg.Graph.DeleteChannelEdge(&chanToPrune.ChannelPoint)
```
Roasbeef: Perhaps we should combine both operations into a single call? If we do, then this means that someone can't give us a super old channel and make us re-scan the graph with GetUTXO or the like.

routing/router.go
```diff
@@ -433,6 +421,30 @@ func (r *ChannelRouter) Start() error {
 		}
 	}
 
+	// If AssumeChannelValid is present, then we won't rely on pruning
+	// channels from the graph based on their spentness, but whether they
+	// are considered zombies or not.
+	if r.cfg.AssumeChannelValid {
+		if err := r.pruneZombieChans(); err != nil {
```
Roasbeef: This skips calling `PruneGraphNodes`.

routing/router.go
cfromknecht: yeah the more i think about this, i think its a good idea to have the transition from live -> zombie be atomic

routing/router.go
wpaulino: Will be including these changes in another PR.

routing/router.go
wpaulino: Good point! Fixed.

discovery/gossiper_test.go
```diff
@@ -2196,6 +2198,233 @@ func TestForwardPrivateNodeAnnouncement(t *testing.T) {
...
+	)
+
+	select {
+	case <-ctx.broadcastedMessage:
+		t.Fatal("expected to not broadcast live channel update " +
+			"without announcement")
+	case <-time.After(2 * trickleDelay):
+	}
+
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
wpaulino: Since we have to re-process it, if we return to the caller immediately then they won't be informed of any errors that occurred while re-processing it. Turns out the sleep wasn't needed though.

routing/router.go
```diff
@@ -76,14 +76,25 @@ type ChannelGraphSource interface {
...
+
+	// MarkEdgeZombie marks an edge as a zombie within our zombie index.
+	MarkEdgeZombie(chanID lnwire.ShortChannelID) error
+
+	// MarkEdgeLive clears an edge from our zombie index, deeming it as
+	// live.
+	MarkEdgeLive(chanID lnwire.ShortChannelID) error
+
+	// IsZombieEdge returns whether the edge is considered zombie.
+	IsZombieEdge(chanID lnwire.ShortChannelID) bool
```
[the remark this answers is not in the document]
wpaulino: Slimmed it down a bit since the gossiper now only requires knowing of `MarkEdgeLive`.

discovery/gossiper.go
wpaulino: Fixed!

wpaulino:
Comments addressed, PTAL @cfromknecht @Roasbeef.

routing/ann_validation.go
```diff
@@ -132,20 +133,28 @@ func ValidateChannelUpdateAnn(pubKey *btcec.PublicKey, capacity btcutil.Amount,
...
+}
+
+// VerifyChannelUpdateSignature verifies that the channel update message was
+// signed by the party with the given node public key.
+func VerifyChannelUpdateSignature(msg *lnwire.ChannelUpdate,
+	pubKey *btcec.PublicKey) error {
+
+	data, err := msg.DataToSign()
 	if err != nil {
-		return errors.Errorf("unable to reconstruct message: %v", err)
```
cfromknecht: 👍 one day, we will kill go-errors

channeldb/graph.go
```diff
@@ -1631,11 +1664,21 @@ func delChannelByEdge(edges *bbolt.Bucket, edgeIndex *bbolt.Bucket,
...
-	return chanIndex.Delete(b.Bytes())
+	if err := chanIndex.Delete(b.Bytes()); err != nil {
+		return err
+	}
+
+	var pubKey1, pubKey2 [33]byte
+	copy(pubKey1[:], nodeKeys[:33])
+	copy(pubKey2[:], nodeKeys[33:])
+
+	return markEdgeZombie(
```
Roasbeef: We may want to distinguish between _deleting_ and edge, and marking it as a zombie...

wpaulino: Fixed.

channeldb/graph.go
```diff
@@ -2782,6 +2793,107 @@ func (c *ChannelGraph) NewChannelEdgePolicy() *ChannelEdgePolicy {
...
+
+// markEdgeZombie marks an edge as a zombie within our zombie index. The public
+// keys should represent the node public keys of the two parties involved in the
+// edge.
+func markEdgeZombie(zombieIndex *bbolt.Bucket, chanID uint64, pubKey1,
+	pubKey2 [33]byte) error {
+
+	var k [8]byte
+	byteOrder.PutUint64(k[:], chanID)
+	v := make([]byte, 66)
```
halseth: `var v [66]byte`

channeldb/graph.go
```diff
@@ -1638,12 +1653,29 @@ func delChannelByEdge(edges *bbolt.Bucket, edgeIndex *bbolt.Bucket,
...
 	}
-	return chanIndex.Delete(b.Bytes())
+	if err := chanIndex.Delete(b.Bytes()); err != nil {
+		return err
+	}
+
+	// Finally, we'll mark the edge as a zombie within our index if it's
+	// being removed due to the channel becoming a zombie. We do this to
+	// ensure we don't store unnecessary data for spent channels.
+	if !isZombie {
```
halseth: huh, missing param?

routing/router.go
```diff
@@ -1009,11 +1009,14 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
...
+		_, _, exists, isZombie, err := r.cfg.Graph.HasChannelEdge(
+			msg.ChannelID,
+		)
 		if err != nil && err != channeldb.ErrGraphNoEdgesFound {
 			return errors.Errorf("unable to check for edge "+
 				"existence: %v", err)
-		} else if exists {
+		}
+		if exists || isZombie {
 			return newErrf(ErrIgnored, "Ignoring msg for known "+
```
halseth: Would be useful to return a slightly different msg for zombies.

routing/router.go
```diff
@@ -1141,7 +1143,8 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 
 		// If the channel doesn't exist in our database, we cannot
 		// apply the updated policy.
-		if !exists {
+		isStaleUpdate := time.Since(msg.LastUpdate) > r.cfg.ChannelPruneExpiry
```
halseth: This looks like new behavior (other than zombie check)?

I think it would be worth documenting what is going on here, returning different error messages (see comment above), and possibly make this its own commit (or describe in the commit message what this is good for).

discovery/gossiper.go
```diff
@@ -1808,56 +1817,100 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
+				"graph(shortChanID=%v), saving for "+
+				"reprocessing later", shortChanID)
+
+			// NOTE: We don't return anything on the error channel
+			// for this message, as we expect that will be done when
+			// this ChannelUpdate is later reprocessed.
+			return nil
+
+		case nil:
+			// No error, fallthrough.
```
halseth: you mean `break`? 😛

discovery/gossiper.go
```diff
@@ -1808,56 +1817,100 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
-				d.rejectMtx.Lock()
-				d.recentRejects[msg.ShortChannelID.ToUint64()] = struct{}{}
-				d.rejectMtx.Unlock()
 				return nil
 			}
+
+			log.Debugf("Removed edge with chan_id=%v from zombie "+
+				"index", msg.ShortChannelID)
+
+			fallthrough
```
halseth: Add a comment explaining why this falltrhoughs. Also update comment below?

discovery/gossiper.go
```diff
@@ -1808,56 +1817,100 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
-				)
-				d.pChanUpdMtx.Unlock()
+		switch err {
+		case channeldb.ErrZombieEdge:
+			// If we've previously deemed the channel for this
+			// update a zombie and we came across a fresh update,
+			// then we'll mark it as live to ensure we can process
+			// any future updates for this edge until it is
+			// considered zombie again.
+			if time.Since(timestamp) > d.cfg.LiveEdgeHorizon {
```
halseth: seems like this check is already done by `IsStaleEdgePolicy`

discovery/gossiper_test.go
```diff
@@ -2191,6 +2193,255 @@ func TestForwardPrivateNodeAnnouncement(t *testing.T) {
...
+	// is a zombie or not.
+	processAnnouncement := func(ann lnwire.Message, isZombie, expectsErr bool) {
+		t.Helper()
+
+		errChan := ctx.gossiper.ProcessRemoteAnnouncement(
+			ann, remotePeer,
+		)
+		select {
+		case err := <-errChan:
+			if isZombie && (expectsErr && err == nil ||
```
halseth: this is kinda hard to comprehend 😛  Better to divide into distinct cases I think

discovery/gossiper_test.go
```diff
@@ -2191,6 +2193,255 @@ func TestForwardPrivateNodeAnnouncement(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("expected to process live channel update: %v",
+				err)
+		}
+	case <-time.After(time.Second):
+		t.Fatal("expected to process announcement")
+	}
+
+	select {
+	case <-ctx.broadcastedMessage:
```
halseth: same, check expected message.

discovery/gossiper_test.go
```diff
@@ -2191,6 +2193,255 @@ func TestForwardPrivateNodeAnnouncement(t *testing.T) {
...
+			}
+			if !isZombie && err != nil {
+				t.Fatalf("expected to process live channel "+
+					"message: %v", err)
+			}
+		case <-time.After(time.Second):
+			t.Fatal("expected to process announcement")
+		}
+		select {
+		case <-ctx.broadcastedMessage:
```
halseth: should check that this is the expected message.

discovery/gossiper_test.go

discovery/gossiper_test.go

discovery/gossiper.go
wpaulino: Fixed.

discovery/gossiper.go
wpaulino: Fixed.

discovery/gossiper.go
wpaulino: Fixed.

routing/router.go
wpaulino: Fixed.

routing/router.go
wpaulino: Fixed.

channeldb/graph.go
wpaulino: Fixed.

channeldb/graph.go
wpaulino: Fixed.

discovery/gossiper.go
```diff
@@ -1808,56 +1809,99 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
+		case nil:
+			break
+
+		case channeldb.ErrZombieEdge:
+			// Since we've deemed the update as not stale above,
+			// before marking it live, we'll make sure it has been
+			// signed by the correct party. The least-significant
+			// bit in the flag on the channel update tells us which
+			// edge is being updated.
+			var pubKey *btcec.PublicKey
```
cfromknecht: not a blocker, just noting that this same block of code exists in many places w/in the code base. would be great to have a method like
```golang
func (c *ChannelEdgeInfo) DirectionKey(flags lnwire.ChannelFlags) *btcec.PublicKey
```

discovery/gossiper.go
```diff
@@ -1808,56 +1809,99 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
...
+			d.prematureChannelUpdates[shortChanID] = append(
+				d.prematureChannelUpdates[shortChanID], nMsg,
+			)
+			d.pChanUpdMtx.Unlock()
+
+			log.Debugf("Got ChannelUpdate for edge not found in "+
+				"graph(shortChanID=%v), saving for "+
+				"reprocessing later", shortChanID)
+
+			// NOTE: We don't return anything on the error channel
```
halseth: Can this potentially cause the gossiper to deadlock if the announcement never comes through? 🤔

wpaulino: No because we return. The `errChan` in which we're sending the error back is given to the caller.

halseth: Was just wondering who waits for the value on the channel, but looks like we never actually read it for remote annonucements...

PR #2781, opened 2019-03-14
routerrpc: adapt to changed interfaces of routing subsystem

joostjager:


lnrpc/routerrpc/config_active.go
```diff
@@ -40,4 +40,8 @@ type Config struct {
 	//
 	// TODO(roasbeef): assumes router handles saving payment state
 	Router *routing.ChannelRouter
+
+	// RouterBackend contains shared logic between this sub server and the
+	// main rpc server.
+	RouterBackend *RouterBackend
```
Roasbeef: We only need this for the self node pubkey right? I think we can pass in just that instead.

joostjager: In this PR it is just that one, but my follow up https://github.com/lightningnetwork/lnd/pull/1662 requires access to other parts of it.

Roasbeef:
Needs a rebase!

joostjager:
@Roasbeef rebased ptal

Roasbeef:
Fails travis: 
```
    lnd_test.go:107: lnd finished with error (stderr):
        exit status 2
        panic: reflect: call of reflect.Value.Set on zero Value
```

joostjager:
@Roasbeef bug fixes, ptal

PR #2783, opened 2019-03-16
watchtower/wtdb: add bbolt-backed tower database

cfromknecht:
This PR adds an initial implementation of a persistent watchtower server-side database. To do so, we extend the current channeldb code with `wtdb` specific types and use quick tests to assert the encoding/decoding of the primary db objects. We then construct a separate `wtdb.TowerDB` using a similar architecture to `channeldb.DB` that implements the `watchtower.DB` interface. A suite of unit tests have been added to assert that the real implementation behaves identically to the mock implementation. The `wtdb.TowerDB` supports database versioning, and includes the necessary pieces to handle any future migrations.


watchtower/wtdb/tower_db.go
```diff
@@ -0,0 +1,577 @@
...
+		return nil, err
+	}
+
+	towerDB := &TowerDB{
+		db:     bdb,
+		dbPath: dbPath,
+	}
+
+	if firstInit {
+		// The file didn't exist before, initialize a fresh database
```
Roasbeef: Seems possible for the file to have been created, but the bucket structure left blank. Therefore, we should just check the bucket directly.

watchtower/wtdb/tower_db.go
```diff
@@ -0,0 +1,577 @@
...
+	})
+}
+
+// InsertStateUpdate stores an update sent by the client after validating that
+// the update is well-formed in the context of other updates sent for the same
+// session. This include verifying that the sequence number is incremented
+// properly and the last applied values echoed by the client are sane.
+func (t *TowerDB) InsertStateUpdate(update *SessionStateUpdate) (uint16, error) {
+	var lastApplied uint16
+	err := t.db.Update(func(tx *bbolt.Tx) error {
```
Roasbeef: `Batch`?

Roasbeef: Applies to write transactions in the entire file as well.

watchtower/wtdb/tower_db.go
```diff
@@ -0,0 +1,577 @@
...
+			return err
+		}
+
+		// Remove the target session.
+		err = sessions.Delete(target[:])
+		if err != nil {
+			return err
+		}
+
+		return updates.ForEach(func(hint, v []byte) error {
```
Roasbeef: Perhaps we should have an additional index mapping a session to the state hints it's stored? This would allow us to avoid scanning over the _entire database of session updates_ for a delete. Instead, we'd iterate over the index and delete the individual sessions. If we go this route, then we might also need a reference counter somewhere as well.

watchtower/wtdb/tower_db_test.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+
+	return lastApplied
+}
+
+// deleteSession attempts to delete the session identified by id and asserts
+// that the error returned from DeleteSession matches the expected error.
+func (h *towerDBHarness) deleteSession(id wtdb.SessionID, expErr error) {
+
+	h.t.Helper()
```
Roasbeef: Style nit: extra new line above.

watchtower/wtdb/tower_db_test.go
```diff
@@ -0,0 +1,712 @@
...
+	h.getSession(id0, wtdb.ErrSessionNotFound)
+
+	// No matches should exist for this hint.
+	matches = h.queryMatches(hint)
+	if len(matches) != 0 {
+		h.t.Fatalf("expected zero updates, found: %d", len(matches))
+	}
+}
+
+type stateUpdateTest = struct {
```
Roasbeef: Extraneous `=` here? Didn't even know something like this would compile lol

watchtower/wtdb/tower_db_test.go
```diff
@@ -0,0 +1,712 @@
...
+			run:  testMultipleMatches,
+		},
+		{
+			name: "lookout tip",
+			run:  testLookoutTip,
+		},
+	}
+
+	for _, db := range dbs {
+		t.Run(db.name, func(t *testing.T) {
```
Roasbeef: If you add a `t.Parallel()` here (before the second loop), then all the DB tests will execute in parallel.

watchtower/wtdb/tower_db.go
cfromknecht: I think we should hold off on adding `Batch` just yet. We have to consider that the inputs are adversarial, and specially crafted inputs can cause the methods below to produce errors and induce even more latency since everything in the batch must be replayed (and this method will execute again in isolation). To do batching properly, we would want `Update` return nil in all error cases that could be adversarially triggered, and then capture the specific error in the surrounding scope. This allows `Batch` to see the operations as successful, even if they trigger no db mutations, but still return the appropriate error to the caller

watchtower/wtdb/codec.go
```diff
@@ -0,0 +1,141 @@
...
+			return err
+		}
+
+	case *BreachHint:
+		if _, err := io.ReadFull(r, e[:]); err != nil {
+			return err
+		}
+
+	case *wtpolicy.Policy:
+		var blobType uint16
```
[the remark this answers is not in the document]
cfromknecht: fixed 79fc15f

watchtower/wtdb/tower_db.go
cfromknecht: wdym by check the bucket directly?

I reworked the logic slightly in e5c34751bad0efd0fd9448a0f59946530fc739b7  to handle setting/migration versions first, then initializing all top-level buckets

watchtower/wtdb/tower_db_test.go
cfromknecht: heh that's strange. i suppose it's just an alias for an unnamed struct, which seems valid. can't think off the top of my head how that'd be useful on its own tho

fixed e098bd44ef8d66aaa0a94ee72d1fa710e88727b6

watchtower/wtdb/tower_db_test.go
cfromknecht: fixed e098bd44ef8d66aaa0a94ee72d1fa710e88727b6

watchtower/wtdb/tower_db_test.go
cfromknecht: interesting. adding this surfaced some weird test interactions that only happen when run in parallel, they're fixed in 01948226bf34da79718842e9bbaa17f9e54b9afa

watchtower/wtdb/tower_db.go
```diff
@@ -0,0 +1,577 @@
...
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/chainntnfs"
+	"github.com/lightningnetwork/lnd/channeldb"
+)
+
+const (
+	// dbName is the filename of tower database.
+	dbName = "watchtower.db"
```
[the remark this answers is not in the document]
cfromknecht: good point! had not considered that, but certainly is useful when we roll out a standalone binary!

watchtower/wtdb/codec.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/watchtower/blob"
+	"github.com/lightningnetwork/lnd/watchtower/wtpolicy"
+)
+
+// UnknownElementType is an alias for channeldb.UnknownElementType.
+type UnknownElementType = channeldb.UnknownElementType
+
+// ReadElement deserializes a single element from the provided io.Reader.
+func ReadElement(r io.Reader, element interface{}) error {
```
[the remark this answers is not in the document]
cfromknecht: thanks! should help conserve a fair amount of code.

one thing that came to mind was to use the wt-specific types first, and only then fall back to the channeldb codec. in the end, i went this this ordering so that we could easily see, via coverage reports, any extraneous definitions in the wt-codec that could be removed. as things solidify, it may be useful to reverse this ordering

watchtower/wtdb/tower_db.go
```diff
@@ -0,0 +1,577 @@
...
+	updatesBkt = []byte("updates-bucket")
+
+	// lookoutTipBkt is a bucket containing the last block epoch processed
+	// by the lookout subsystem. It has one key, lookoutTipKey.
+	//   lookoutTipKey -> block epoch
+	lookoutTipBkt = []byte("lookout-tip-bucket")
+
+	// lookoutTipKey is a static key used to retrieve lookout tip's block
+	// epoch from the lookoutTipBkt.
+	lookoutTipKey = []byte("lookout-tip")
```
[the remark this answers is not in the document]
cfromknecht: not sure i follow, `*bbolt.Tx` doesn't have a `Get` method, so it necessitates having a top-level bucket. maybe i'm missing something tho?

watchtower/wtdb/tower_db.go
cfromknecht: Initially thought we could do a migration that would add this info, since it's possible to do a complete scan and reconstruct these indexes. I went ahead and added them tho and it wasn't too much additional code, lmk what you think f6c9bd6

cfromknecht:
@Roasbeef comments addressed, ptal

watchtower/wtdb/tower_db.go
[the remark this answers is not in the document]
Roasbeef: Ah yeah, was thinking this was `lnd` where everything already sort of has a top-level bucket.

watchtower/wtdb/tower_db.go
Roasbeef: I mean just always call the init logic as if the bucket exists, this it's a noop.

watchtower/wtdb/tower_db.go
Roasbeef: Can't the same be said for any node announcement gossip message? Rationale behind optimistically batching is to ensure a tower can keep up with the level of traffic at scale. We'll either need to do it here, do it at a higher level (coalesce several updates within the server), or switch to something like `leveldb` which is write optimized.

watchtower/wtdb/tower_db.go
```diff
@@ -98,23 +98,36 @@ func OpenTowerDB(dbPath string) (*TowerDB, error) {
...
+		})
 		if err != nil {
 			bdb.Close()
 			return nil, err
 		}
 	} else {
-		// Otherwise, ensure that any migrations are applied before
-		// returning the opened database.
+		// Otherwise, ensure that any migrations are applied to ensure
+		// the data is in the format expected by the latest version.
```
Roasbeef: If the database file was created, we crash before the if statement body above is executed, then this will fail since there's no metadata bucket.

cfromknecht: gotcha, i see what you're saying now. `firstInit` should be `file DNE || metadataBkt DNE`

cfromknecht: should be fixed by fe64b8c

Issue #1611, opened 2018-07-23
Feature request: Add programmatic way of unlocking wallet


slush0:
### Background

Currently there's no easy way of unlocking wallet in automatic manner, making it hard to automate services for managing lnd. Although I understand passing passphrase over command line is not safe, reading passphrase over stdin or launching external command for fetching passphrase would make it possible.

### Expected behaviour

Starting lnd and unlocking wallet with passphrase from stdin:

`cat passphrase.txt | lnd --stdin-passphrase`

or

`lnd --unlock-command unlock.sh`

### Actual behaviour

There's no way to "un-encrypt" existing wallet (or at least I didn't find a way), only create new one with `--noencryptwallet`, so it is not possible to automate lnd with systemd.

Currently I use this workaround (may be useful for other people), but it is ugly hack, because lnd need to be spawn as separate process before, making systemd scripts non-trivial.

```
# cat lnd-start.sh
#!/bin/bash
/usr/local/bin/lnd-unlockwallet.sh &
/home/bitcoin/go/bin/lnd
```

```
# cat lnd-unlockwallet.sh
#!/usr/bin/expect
sleep 3
spawn /home/bitcoin/go/bin/lncli "unlock"
expect {
	"Input wallet password:" {
		send -- "password\n"
	}
}
expect eof
exit
```

Roasbeef:
> Currently there's no easy way of unlocking wallet in automatic manner, making it hard to automate services for managing lnd.

There's an gRPC method: https://api.lightning.community/#unlockwallet 
Also available over REST: https://api.lightning.community/rest/index.html#v1-unlockwallet

> There's no way to "un-encrypt" existing wallet (or at least I didn't find a way), only create new one with --noencryptwallet, so it is not possible to automate lnd with systemd.

Yeah there's no way to "unencrypt" as it defers to on-disk encryption. The `--noencryptwallet` option actually just uses a hard coded key for the encryption, and atm also doesn't give you a mnemonic seed or anything like that.

slush0:
Thank you for pointing me to REST API, it may be more reliable than `expect` workaround I'm using now. It is still workaround though, as it need invoking another script/process to start lnd, although it could be done easily and in clean way by adding new parameter.

mariodian:
@slush0 did you manage to figure it out? I'm struggling to run lncli remotely while automatically providing the password.

[...]

PR #2791, opened 2019-03-18
lncli: add password arg to unlock

pinheadmz:
This PR adds a new command line argument  `password` to the `lncli unlock` command.

Example usage:
```
lncli unlock --password=hunter2
```

The motivation is to unlock the lnd wallet with a single command instead of interacting with `lncli` after running `lncli unlock`. More specifically, I am interested in running lnd as a service on my server with auto-restart enabled. Ordinarily, a restart would require the password to be re-entered. I very much appreciate the security concern of storing a password in plaintext or printing it on the command line in plaintext but for my use case at least, lnd is a hot wallet and when its running I am already taking on the usual hot-wallet risks.

Normal functionality has been altered slightly with regard to the existing optional param `recovery_window`. This param was expected at index 0 but now will be at index 1 and password will be at 0. Not sure how we want to deal with two optional params, but without the flags the command will work like this:
```
lncli unlock <password> <recovery_window>
```
e.g.
```
lncli unlock hunter2 10
```

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -1548,6 +1548,10 @@ var unlockCommand = cli.Command{
 	--noseedbackup, then a default passphrase will be used.
 	`,
 	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "password",
+			Usage: "the wallet encryption password",
+		},
 		cli.IntFlag{
 			Name: "recovery_window",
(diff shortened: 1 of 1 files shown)


halseth:
See discussion in https://github.com/lightningnetwork/lnd/issues/1611. If you feel this functionality is still needed, please open an issue describing your use case.

PR #2797, opened 2019-03-18
[autopilot] penalize small channels in preferantial attachment heuristic

halseth:
To avoid assigning a high score to nodes with a large number of small
channels, we only count channels at least as large as a given fraction of
the graph's median channel size.


autopilot/prefattach.go
```diff
@@ -72,12 +77,63 @@ func (p *PrefAttachment) NodeScores(g ChannelGraph, chans []Channel,
...
+	// the graph.
 	var maxChans int
 	nodeChanNum := make(map[NodeID]int)
 	if err := g.ForEachNode(func(n Node) error {
 		var nodeChans int
-		err := n.ForEachChannel(func(_ ChannelEdge) error {
+		err := n.ForEachChannel(func(e ChannelEdge) error {
+			// Since connecting to nodes with a lot of small
+			// channels actually worsens our connectivity in the
+			// graph (we will potentially waste time tryng to use
```
alexbosworth:
```suggestion
			// graph (we will potentially waste time trying to use
```

rpcserver.go
```diff
@@ -3961,6 +3967,20 @@ func (r *rpcServer) GetNetworkInfo(ctx context.Context,
 		return nil, err
 	}
 
+	// Sort the channels by capacity, and find the median.
```
alexbosworth: Maybe there should be a cap here in case the median is super high for some reason?

halseth: You mean cap the number of channels we'll sort?

alexbosworth: Cap the median value amount

Roasbeef: This is just for the RPC call, why should we cap it?

rpcserver.go
```diff
@@ -3961,6 +3967,20 @@ func (r *rpcServer) GetNetworkInfo(ctx context.Context,
...
+	sort.Slice(allChans, func(i, j int) bool {
+		return allChans[i] < allChans[j]
+	})
+
+	if numChannels == 0 {
```
Roasbeef: Switch statement instead perhaps?

autopilot/prefattach.go
```diff
@@ -72,12 +77,63 @@ func (p *PrefAttachment) NodeScores(g ChannelGraph, chans []Channel,
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	sort.Slice(allChans, func(i, j int) bool {
```
Roasbeef: Duplicate code to what was added in the server, we can add an abstract method in the `router` package perhaps? Or a "static method" (lol) on the `ChannelGraph`

autopilot/prefattach.go
```diff
@@ -72,12 +77,63 @@ func (p *PrefAttachment) NodeScores(g ChannelGraph, chans []Channel,
...
 	if err := g.ForEachNode(func(n Node) error {
 		var nodeChans int
-		err := n.ForEachChannel(func(_ ChannelEdge) error {
+		err := n.ForEachChannel(func(e ChannelEdge) error {
+			// Since connecting to nodes with a lot of small
+			// channels actually worsens our connectivity in the
+			// graph (we will potentially waste time tryng to use
+			// these useless channels in path finding), we decrease
+			// the counter for such channels.
+			if e.Capacity < medianChanSize/4 {
```
Roasbeef: `4` should be moved up into a constant

autopilot/prefattach.go
halseth: Fixed.

rpcserver.go
halseth: Question is whether to cap it for prefattach?

rpcserver.go
halseth: Fixed.

autopilot/prefattach.go
halseth: added `autopilot.Median()`

autopilot/prefattach.go
halseth: Done.

halseth:
PTAL @alexbosworth @Roasbeef

autopilot/prefattach.go
```diff
@@ -72,12 +82,50 @@ func (p *PrefAttachment) NodeScores(g ChannelGraph, chans []Channel,
...
 
-	// Count the number of channels for each particular node in the graph.
+	// We first run though the graph once in order to find the median
+	// channel size.
+	var (
+		allChans  []btcutil.Amount
+		seenChans = make(map[uint64]struct{})
+	)
+	if err := g.ForEachNode(func(n Node) error {
+		err := n.ForEachChannel(func(e ChannelEdge) error {
```
alexbosworth: It may be safer to only consider channels that have recently been updated

alexbosworth:
lgtm although I still feel iffy about situations where the median could actually yield unexpected results. The code in comments says "large" but median doesn't necessarily imply large

halseth:
@alexbosworth "large-ish channels" 😛

PR #2799, opened 2019-03-18
router: Optimize pruning zombie channels.

roeierez:
This commit introduces a different, more optimized, approach to prune zombie channels.
"zombie channels" stands for channels that their last channel policy update time was earlier than "channel expiry", configured by default to a period of 14 days.
The current approach iterates all channels in the graph in order to filter those in need. This approach is time consuming, several seconds on my mobile device for ~40,000 channels, while during this time the
db is locked in a transaction.

The proposed change is to use an existing functionality that utilize the fact that channel update are saved indexed by date. This way enables us to go over only a small subset of the channels, only those that were updated before the "channel expiry" time and further filter them for our need.
The same graph that took several seconds to prune was pruned, after the change in several milliseconds.

Currently the `pruneZombieChans` is running every hour, I think that the performance improvement will safely allow to run it on every restart and make sure LND starts up with no zombie channels.

My measurements for ~40,000 channels.

**Samsung galaxy s7**
prune time before change: 5-12 seconds, depends on the current load.
prune time after change: 15 milliseconds.

**Mac book pro 2.8 GHz Intel Core i7**
prune time before change: ~400 milliseconds
prune time after change: 15 milliseconds.

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -675,10 +675,15 @@ func (r *ChannelRouter) pruneZombieChans() error {
 	r.rejectMtx.Lock()
 	defer r.rejectMtx.Unlock()
 
-	err := r.cfg.Graph.ForEachChannel(filterPruneChans)
+	startTime := time.Unix(0, 0)
+	endTime := time.Now().Add(-1 * chanExpiry)
+	updates, err := r.cfg.Graph.ChanUpdatesInHorizon(startTime, endTime)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


halseth:
Related: https://github.com/lightningnetwork/lnd/pull/2777

roeierez:
Replaced by https://github.com/lightningnetwork/lnd/pull/3197

PR #2802, opened 2019-03-19
routing: probability based path finding

joostjager:
In this PR, several improvements to path finding and mission control are made with the goal to improve the reliability of Lightning payments.

# Changes

## Probability based routing

The previously used edge and node ignore lists in path finding are replaced by a probability based system. It modifies path finding so that it not only compares routes on fee and time lock, but also takes route success probability into account.

Allowing routes to be compared based on success probability is achieved by introducing a 'virtual' cost of a payment attempt and using that to translate probability into an extra cost factor.

## Failed route decay time

This PR also extends the decay time of mission control data. It used to be very short which hardly allowed a payment to benefit from outcomes of previous payment attempts. With this PR, the decay time becomes configurable.

## Minimum route probability

The payment process stop condition has been adapted to probability based routing. Previously the payment processed stopped when no more routes could be found. This typically happened when all tried edges had been added to the ignore list. With probabilities, every route will always have a non-zero success probability. That probability can be small, but the payment may still succeed via this route. 

To prevent the payment process ending up in a loop trying routes that are highly unlikely to succeed until it times out, a minimum required success probability is introduced. Only routes with a probability above this threshold are considered and if there are none, the payment process is stopped.

## Query mission control state

A new rpc call has been added to query the internal mission control state. This should give advanced users a tool to better understand why certain routes are chosen.

# Usage
## Config
The main parameters influencing the payment process within `lnd` are exposed as command line or config flags. For these flags to be available, `lnd` needs to be build with `tags="routerrpc"`.

`--routerrpc.minrtprob=`

Minimum required route success probability to attempt the payment (default: 0.01)

`--routerrpc.apriorihopprob=`

Assumed success probability of a hop in a route when no other                                                        information is available. (default: 0.95)

`--routerrpc.penaltyhalflife=`

Defines the duration in minutes after which a penalized node or channel is back at 50% probability (default: 60)

`--routerrpc.attemptcost=`

The (virtual) cost in sats of a failed payment attempt (default: 100)

## lncli
Querying mission control is available through `lncli` (also requiring the `routerrpc` tag):

`lncli querymc`


routing/pathfind.go
```diff
@@ -614,5 +634,44 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
 	return pathEdges, nil
 }
+
+// getProbabilityBasedDist converts a weight into a distance that takes into
+// account the success probability and the (virtual) cost of a failed payment
+// attempt.
+//
+// Derivation:
+//
+// Suppose there are two routes A and B with fees Fa and Fb and success
```
halseth: Great comment! 👍

routing/payment_session.go
```diff
@@ -161,11 +175,11 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
...
-			IgnoredNodes:      pruneView.vertexes,
-			IgnoredEdges:      pruneView.edges,
-			FeeLimit:          payment.FeeLimit,
-			OutgoingChannelID: payment.OutgoingChannelID,
-			CltvLimit:         cltvLimit,
+			ProbabilitySource:     p.getEdgeProbability,
+			FeeLimit:              payment.FeeLimit,
+			OutgoingChannelID:     payment.OutgoingChannelID,
+			CltvLimit:             cltvLimit,
+			PaymentAttemptPenalty: DefaultPaymentAttemptPenalty,
```
halseth: Nice! This can let us dynamically increase this each time a route is requested (if we want).

routing/missioncontrol.go
```diff
@@ -287,7 +220,108 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	}
+
+	lastEdgeFailure, ok := nodeHistory.channelLastFail[channelID]
+	if ok {
+		if lastEdgeFailure.After(lastFailure) {
+			lastFailure = lastEdgeFailure
+		}
+	}
+
+	// If we are still in the hard prune window, return probability 0.
```
halseth: could we consider letting this happen, and instead rely on the payment timing out instead? That would allow long-lived path finding.

joostjager: Changed this part of the PR. Currently mission control isn't concerned with 'hard pruning' anymore. Instead, a minimum probability restriction is added to path finding. Setting this to a low value should help prevent trying unlikely routes in a short loop.

joostjager: For long payment timeouts it may actually be desired that we go over previously failed routes again to see if conditions improved.

joostjager:
ptal @halseth

routing/pathfind.go
```diff
@@ -614,5 +649,44 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
+//
+// Divide by Pa*Pb:
+// c/Pa - Fb < c/Pb - Fa
+//
+// Move terms around:
+// Fa + c/Pa < Fb + c/Pb
+//
+// So the value of F + c/P can be used to compare routes.
+func getProbabilityBasedDist(weight int64, probability float64, penalty int64) int64 {
+	return weight + int64(float64(penalty)/probability)
```
cfromknecht: do we want to clamp the output? for very small probabilities, this _could_ overflow or be very large or cause overflows later on

At some point we should consider moving purely to `uint64` weights. At least then if an overflow happens, it can't produce a negative weight edge.

routing/pathfind_test.go
```diff
@@ -2035,3 +2040,142 @@ func testCltvLimit(t *testing.T, limit uint32, expectedChannel uint64) {
 			route.Hops[0].ChannelID)
 	}
 }
+
+// TestProbabilityRouting asserts that path finding not only takes into account
+// fees but also success probability.
+func TestProbabilityRouting(t *testing.T) {
```
cfromknecht: solid tests!

routing/pathfind.go
```diff
@@ -40,6 +40,18 @@ type pathFinder = func(g *graphParams, r *RestrictParams,
...
+var (
+	// DefaultPaymentAttemptPenalty is the virtual cost in path finding weight
+	// units of executing a payment attempt that fails. It is used to trade
+	// off potentially better routes against their probability of
+	// succeeding.
+	DefaultPaymentAttemptPenalty = lnwire.NewMSatFromSatoshis(100)
+
+	// DefaultMinProbability is the default minimum probability for routes
+	// returned from findPath.
+	DefaultMinProbability = float64(0.05)
```
cfromknecht: this value seems a little high to use an initial minimum, though I haven't worked with the PR as much as yourself. it might be useful to start low and then bump after we are confident in the behavior in the wild. mainly thinking about the situation where last resort paths exist, but mission control prevents us from exploring them

routing/router.go
```diff
@@ -1916,13 +1868,11 @@ func (r *ChannelRouter) processSendError(paySession *paymentSession,
...
 	// continue with the rest of the routes.
 	case *lnwire.FailPermanentChannelFailure:
-		paySession.ReportEdgeFailure(&EdgeLocator{
-			ChannelID: failedEdge.ChannelID,
-			Direction: 0,
-		})
-		paySession.ReportEdgeFailure(&EdgeLocator{
-			ChannelID: failedEdge.ChannelID,
-			Direction: 1,
+		paySession.ReportEdgeFailure(failedEdge)
```
cfromknecht: interesting that a permanent failure would have a half-life wrt to success probability. down the road are we planning to make probabilities more failure-specific?

for example, permanent failures could update the `nodeHistory` to then always report the probability as 0 for that channel

another thing that comes to mind is that a bandwidth failure for 100000 sats may cause a channel to be ignored when trying to route 1 sat payment directly after. seems we would need to track the size of the payment that generated the failure (at least in this case).

routing/route/route.go
```diff
@@ -175,3 +177,19 @@ func (r *Route) ToSphinxPath() (*sphinx.PaymentPath, error) {
 
 	return &path, nil
 }
+
+// String returns a human readable representation of the route.
+func (r *Route) String() string {
+	var b strings.Builder
+
+	b.WriteString(strconv.FormatUint(r.Hops[0].ChannelID, 10))
```
cfromknecht: do we want to put this in a conditional, so that it can't panic if the route is empty?

routing/router.go
cfromknecht: thinking more on the bandwidth failures, i think we would only need to track the minimum payment size that caused the failure. could then define a piecewise probability function:
```
Pr[success] = 1               if amt < min_failed_amt
Pr[success] = 1 - 2^exp       if amt >= min_failed_amt
```

routing/route/route.go
joostjager: good point, fixed

routing/pathfind.go
```diff
@@ -483,16 +516,18 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
-			fee:             fee,
 			incomingCltv:    incomingCltv,
+			probability:     probability,
 		}
 
 		next[fromVertex] = edge
 
 		// Add this new node to our heap as we'd like to further
 		// explore backwards through this edge.
+
```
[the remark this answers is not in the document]
joostjager: no, removed

routing/pathfind.go
joostjager: Added clamp

routing/payment_session.go
joostjager: Yes, it will be interesting to see how this will develop further.

routing/pathfind.go
joostjager: Yes I wasn't sure about this value either. I will lower it to 1%. The worst that can happen is that a route is tried in a loop until the payment attempt timeout.

lntest/node.go
```diff
@@ -243,6 +244,8 @@ type HarnessNode struct {
 	lnrpc.WalletUnlockerClient
 
 	invoicesrpc.InvoicesClient
+
+	RouterClient routerrpc.RouterClient
```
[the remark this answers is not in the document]
joostjager: There is a name collision `SendPayment` between the main rpc and routerrpc :(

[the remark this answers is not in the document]
joostjager: Comment added

routing/missioncontrol.go
```diff
@@ -279,9 +242,152 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	}
+
+	timeSinceLastFailure := now.Sub(lastFailure)
+
+	// Calculate coefficient to set the half life time.
+	coeff := math.Log10(2) / m.penaltyHalfLife.Hours()
+
+	// Calculate success probability.
+	probability := 1 - (1 /
+		(math.Pow(10, coeff*timeSinceLastFailure.Hours())))
```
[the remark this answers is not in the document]
joostjager: Yes better.

routing/missioncontrol_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// TestMissionControl tests mission control probability estimation.
+func TestMissionControl(t *testing.T) {
+	now := testTime
```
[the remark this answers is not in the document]
joostjager: It was in `notifications_test.go`, but converted to a local var to keep the tests independent.

lnrpc/routerrpc/router.proto
```diff
@@ -213,6 +213,40 @@ message ResetMissionControlRequest{}
...
+    repeated NodeHistory nodes = 1;
+}
+
+/// NodeHistory contains the mission control state for a particular node.
+message NodeHistory {
+    /// Node pubkey
+    bytes pubkey = 1;
+
+    /// Time stamp of last failure. Set to zero if no failure happened yet.
+    int64 last_fail = 2;
```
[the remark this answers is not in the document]
joostjager: Added

lnrpc/routerrpc/router.proto
```diff
@@ -209,6 +209,43 @@ message ChannelUpdate {
...
+
+    /// Estimation of success probability for channels not in the channel array.
```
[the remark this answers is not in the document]
joostjager: The array only contains channels for which we have specific information. For channels that we never tried, we can only look at node level failures. Those are assumed to have happened regardless of the outgoing channel.

routing/missioncontrol.go
```diff
@@ -303,3 +343,51 @@ func (m *MissionControl) reportEdgeFailure(failedEdge edge) {
...
+	m.Lock()
+	defer m.Unlock()
+
+	log.Debugf("Requesting history snapshot from mission control: "+
+		"node_count=%v", len(m.history))
+
+	nodes := make([]*MissionControlNodeSnapshot, 0, len(m.history))
+
+	for v, h := range m.history {
+		channelSnapshot := make([]*MissionControlChannelSnapshot,
```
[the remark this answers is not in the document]
joostjager: What exact format do you suggest?

server.go
```diff
@@ -609,13 +611,74 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
 	}
 	s.currentNodeAnn = nodeAnn
 
+	queryBandwidth := func(edge *channeldb.ChannelEdgeInfo) lnwire.MilliSatoshi {
+		// If we aren't on either side of this edge, then we'll
+		// just thread through the capacity of the edge as we
+		// know it.
+		if !bytes.Equal(edge.NodeKey1Bytes[:], selfNode.PubKeyBytes[:]) &&
```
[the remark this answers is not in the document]
joostjager: Agree. Removed.

routing/missioncontrol.go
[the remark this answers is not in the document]
joostjager: So currently the probabilities of a channel are not symmetrical. This is something that I'd like to change in follow up prs, but for this pr I'd like to stay close to the previous version of mission control.

routing/router.go
joostjager: Yes, so the question is how permanent a permanent failure is. Do we want to take into account that this result could have been sent because of a bug and give another chance?

For this PR, the probability calculation is very basic. I plan to do a follow up with much better interpretation of failures. Not only the permanent failures, but also other errors for which the error is currently not attributed to the right channel. 

The main goal for this pr is to make mission control global and have it remember things longer than it used to do, without regression.

Your comment about bandwidth failures is valid in that regard. Previously we tracked those failures only locally for a single payment. All routes for a payment used roughly the same amount (ignoring fees). Now we may prune channels because we tried a high amount, while they may still work for a low amount (in the next payment).

I think there are two ways of addressing this:

1. In line with your suggestion. Try to factor the amount into the probability. I like your idea because it is simple and will work almost the same as in master. For concurrent payments there will be interference between the different amounts probably, but I feel that is ok for now. Later we could go with multiple segments

2. Be more aggressive and expect nodes to keep their channels balanced. Only pick channels during path finding for which the payment amount is <30% of the channel capacity and immediately prune the channel if it fails.

joostjager: Also what could happen is that if we decide to be more relaxed about balance issues, nodes may just replace their actual error by a temp channel failure. To get penalized less. I tend to be cautious with coupling too specific behaviour to a particular fail code.

joostjager: Discussed with @Roasbeef as well. Option 2 is likely too early for the network in its current state.

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
cfromknecht: gotcha, i'm wondering then if naming it `node_success_prob` is more clear. up to you

routing/pathfind.go
cfromknecht: sgtm

routing/router.go
cfromknecht: > Do we want to take into account that this result could have been sent because of a bug and give another chance?

Typically this would indicate the channel has already closed (via timeout resolutions). I suppose it doesn't matter too much, by the time the penalty has decayed enough to make the route attractive, we will have likely pruned the channel from our graph.

That being said, do we also want to prune mission control state when pruning the graph?

> Also what could happen is that if we decide to be more relaxed about balance issues, nodes may just replace their actual error by a temp channel failure. To get penalized less. I tend to be cautious with coupling too specific behaviour to a particular fail code.

Good point, we can never be 100% certain. If we want to treat all of the errors similarly, then I suppose we should track the failed payment amount for all failure types and always apply the piece-wise heuristic?

> Option 2 is likely too early for the network in its current state.

Agreed. I'd be in favor of seeing what option 1 looks like so that we maintain a close-ish behavior to current master

routing/missioncontrol.go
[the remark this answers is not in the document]
cfromknecht: i just mean to define it as []MissionControlSnapshot since the slice will already be on the heap, and then the MissionControlSnapshots will occupy a contiguous block of memory rather than doing one allocation per entry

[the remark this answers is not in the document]
joostjager: Done

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
joostjager: Yes, the thing is that it is not a node probability. It is an estimate for channels that we don't have any other info on. In the future, this could also be (for example) an average of all channels that we do have info on. I found the concept of node probability and how to blend that with channel probabilities not very clear.

routing/router.go
joostjager: Implemented option 1

joostjager: >That being said, do we also want to prune mission control state when pruning the graph?

I think I am more comfortable with pruning mission control independently. Just take a max age and remove when it is exceeded.

Also, observations may still be valuable after the channel is pruned. It indicates how the node behaved historically.

In this PR, mission control state grows in memory. I don't think it is a problem, as it it bounded by the total number of channels and nodes in the graph.

In the follow up where persistence is added, I will add pruning.

joostjager:
Added config flags.

PTAL @halseth @cfromknecht

lnrpc/routerrpc/config_active.go
```diff
@@ -19,6 +19,19 @@ type Config struct {
 	// directory, named DefaultRouterMacFilename.
 	RouterMacPath string `long:"routermacaroonpath" description:"Path to the router macaroon"`
 
+	// MinProbability is the minimum required route probability to attempt
+	// the payment.
+	MinProbability float64 `long:"minprobability" description:"Minimum required route probability to attempt the payment"`
+
+	// PenaltyHalfLife defines after how much time a penalized node or
+	// channel is back at 50% probability.
+	PenaltyHalfLife int `long:"penaltyhalflife" description:"Defines the duration in minutes after which a penalized node or channel is back at 50% probability"`
```
cfromknecht: should be okay to use `time.Duration` directly?

server.go
```diff
@@ -632,8 +633,32 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
+
+	mcRefl := reflect.ValueOf(cfg.SubRPCServers.RouterRPC).Elem()
+	if mcRefl.NumField() != 0 {
+		mcCfg.MinProbability = mcRefl.FieldByName("MinProbability").
+			Float()
+
+		mcCfg.PaymentAttemptPenalty = lnwire.NewMSatFromSatoshis(
+			btcutil.Amount(mcRefl.FieldByName("AttemptCost").Int()),
+		)
+		mcCfg.PenaltyHalfLife = time.Hour *
```
cfromknecht: config says the value is in minutes, though this converts to hours? would be irrelevant tho if we can use `time.Duration` on the config

routing/missioncontrol.go
```diff
@@ -281,7 +211,108 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+	now := m.now()
+
+	m.Lock()
+	defer m.Unlock()
+
+	history := m.createHistoryIfNotExists(failedEdge.from)
+	history.channelLastFail[failedEdge.channel] = &channelHistory{
+		lastFail:       now,
+		minPenalizeAmt: minPenalizeAmt,
```
cfromknecht: should this compute the min of the new value and the old? do we need a way to clear these when we get a success?

lnrpc/routerrpc/config_active.go
joostjager: didn't know that. works ok indeed. changed.

routing/missioncontrol.go
joostjager: Clearing on success is planned for a follow up pr. this main requirement of this pr is no regression. Assuming that payments don't happen concurrently, the current logic works roughly the same as before. The first failure sets the minPenalizeAmt and subsequent attempts will use around the same value, so their probability will be low. When a new attempt starts with a lower value, probability 1 is returned. A new chance, just as with the previous payment session scoped mission control.

joostjager:
@cfromknecht ptal

routing/missioncontrol.go
```diff
@@ -281,7 +211,108 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+	return m.getEdgeProbabilityForNode(nodeHistory, edge.ChannelID, amt)
+}
+
+// getEdgeProbabilityForNode estimates the probability of successfully
+// traversing a channel based on the node history.
+func (m *MissionControl) getEdgeProbabilityForNode(nodeHistory *nodeHistory,
+	channelID uint64, amt lnwire.MilliSatoshi) float64 {
+
+	now := m.now()
```
halseth: nit: can just call when it is first used.

routing/missioncontrol.go
```diff
@@ -70,83 +54,33 @@ type MissionControl struct {
 	// TODO(roasbeef): also add favorable metrics for nodes
 }
 
+// nodeHistory contains a summary of payment attempt outcomes involving a
+// particular node.
+type nodeHistory struct {
+	lastFail        *time.Time
+	channelLastFail map[uint64]*channelHistory
```
halseth: should have a comment.

routing/missioncontrol.go
```diff
@@ -70,83 +54,33 @@ type MissionControl struct {
...
+}
+
+// channelHistory contains a summary of payment attempt outcomes involving a
+// particular channel.
+type channelHistory struct {
+	lastFail       time.Time
+	minPenalizeAmt lnwire.MilliSatoshi
```
halseth: should have a comment.

routing/router.go
```diff
@@ -2331,3 +2332,10 @@ func (r *ChannelRouter) IsStaleEdgePolicy(chanID lnwire.ShortChannelID,
 func (r *ChannelRouter) MarkEdgeLive(chanID lnwire.ShortChannelID) error {
 	return r.cfg.Graph.MarkEdgeLive(chanID.ToUint64())
 }
+
+// edge is a combination of a channel and the node pubkeys of both of its
+// endpoints.
+type edge struct {
```
halseth: define above first use?

Also, `edge` is possibly a confusing name since it used for all sorts of things. Do we need to introduce yet another edge/channel definition?

routing/missioncontrol.go
```diff
@@ -316,3 +358,58 @@ func (m *MissionControl) reportEdgeFailure(failedEdge edge,
...
+			)
+
+			channelSnapshot[chanIdx] = MissionControlChannelSnapshot{
+				ChannelID:      id,
+				LastFail:       lastFail.lastFail,
+				MinPenalizeAmt: lastFail.minPenalizeAmt,
+				SuccessProb:    prob,
+			}
+
+			chanIdx++
```
halseth: can use `append`

routing/missioncontrol.go
```diff
@@ -316,3 +358,58 @@ func (m *MissionControl) reportEdgeFailure(failedEdge edge,
...
+		otherProb := m.getEdgeProbabilityForNode(h, 0, 0)
+
+		nodes[nodeIdx] = MissionControlNodeSnapshot{
+			Node:                 v,
+			LastFail:             h.lastFail,
+			OtherChanSuccessProb: otherProb,
+			Channels:             channelSnapshot,
+		}
+
+		nodeIdx++
```
halseth: same, `append`

routing/missioncontrol.go
```diff
@@ -316,3 +358,58 @@ func (m *MissionControl) reportEdgeFailure(failedEdge edge,
...
+	}
+
+	snapshot := MissionControlSnapshot{
+		Nodes: nodes,
+	}
+
+	log.Debugf("History snapshot created")
```
halseth: Not a very useful log? Given that nothing can fail  before returning.

server.go
```diff
@@ -632,8 +633,31 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
+	// TODO(joostjager): When we are further in the process of moving to sub
+	// servers, the mission control instance itself can be moved there too.
+	mcCfg := &routing.MissionControlConfig{
+		MinProbability:        routing.DefaultMinProbability,
+		PaymentAttemptPenalty: routing.DefaultPaymentAttemptPenalty,
+		PenaltyHalfLife:       routing.DefaultPenaltyHalfLife,
+	}
+
+	mcRefl := reflect.ValueOf(cfg.SubRPCServers.RouterRPC).Elem()
+	if mcRefl.NumField() != 0 {
```
halseth: possible to define a method `GetMCConfig` in `routerRPC` instead, to avoid the reflection?

routing/pathfind.go
```diff
@@ -380,21 +382,25 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
-
+		// Calculate amount that the candidate node would have to sent
+		// out.
 		toNodeDist := distance[toNode]
-
 		amountToSend := toNodeDist.amountToReceive
 
+		// Request the success probability for this edge.
+		edgeProbability := float64(1)
+		if r.ProbabilitySource != nil {
```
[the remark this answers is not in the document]
joostjager: It is a convenience so that the default behaves as expected. Saves setting this to a dummy function in many tests.

routing/pathfind_test.go
```diff
@@ -2045,3 +2040,122 @@ func testCltvLimit(t *testing.T, limit uint32, expectedChannel uint64) {
...
+	probabilitySource := func(node route.Vertex, edge EdgeLocator) float64 {
+
+		switch edge.ChannelID {
+		case 10:
+			return p10
+		case 11:
+			return p11
+		case 20:
+			return p20
+		default:
```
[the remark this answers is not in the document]
joostjager: Yes, there are the channels from self for which a probability is requested too.

routing/pathfind_test.go
```diff
@@ -2045,3 +2040,122 @@ func testCltvLimit(t *testing.T, limit uint32, expectedChannel uint64) {
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	const (
+		startingHeight = 100
+		finalHopCLTV   = 1
+	)
+	route, err := newRoute(
```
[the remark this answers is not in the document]
joostjager: good suggestion, fixed.

routing/pathfind_test.go
```diff
@@ -2051,10 +2051,24 @@ func TestProbabilityRouting(t *testing.T) {
...
+	// excluded and the two hop route to be picked.
+	t.Run("probability limit", func(t *testing.T) {
+		testProbabilityRouting(t, 0.8, 0.5, 0.7, 0.5, 20)
+	})
+
+	// With a probability limit above the probability of both routes, we
+	// expect no route to be returned. This expectation is signaled by using
+	// expected channel 0.
+	t.Run("probability limit no routes", func(t *testing.T) {
+		testProbabilityRouting(t, 0.8, 0.5, 0.7, 0.8, 0)
```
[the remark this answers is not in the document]
joostjager: Convert to table driven, better indeed.

routing/missioncontrol.go
```diff
@@ -281,7 +211,108 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	if ok && channelHistory.minPenalizeAmt <= amt {
+		if channelHistory.lastFail.After(lastFailure) {
+			lastFailure = channelHistory.lastFail
+		}
+	}
+
+	timeSinceLastFailure := now.Sub(lastFailure)
+
+	// Calculate success probability.
+	exp := -timeSinceLastFailure.Hours() / m.penaltyHalfLife.Hours()
```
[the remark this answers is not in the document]
joostjager: Added

routing/router.go
joostjager: Moved up.

Yes it is another variation indeed, but it only exists temporary until we move to passing the full route into mission control in #3077.

routing/missioncontrol.go
joostjager: Indeed, removed.

server.go
joostjager: Ok, good suggestion. Although in general I still haven't heard a convincing argument for the conditional compilation. I only know that it cost me a lot time in total dealing with it, all spread out over hundreds of micro-penalties. I am sure users are experiencing the same.

joostjager:
@halseth ptal

routing/pathfind.go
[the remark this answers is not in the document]
halseth: I would prefer not doing this check, and rather setting it in the tests to avoid this hiding some issue in production.

server.go
halseth: I'm not fundamentally opposed to conditional compilation, but reflection degrades our beautiful language to javascript.

routing/pathfind.go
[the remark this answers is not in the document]
joostjager: Removed the check. It is better indeed.

server.go
joostjager: I am neither, but in the way we use it, I don't see what we gain. I hope in the future there will be a majority to remove it in favor of clearly marking things as experimental in the rpc docs.

routing/pathfind.go
```diff
@@ -228,13 +236,9 @@ type graphParams struct {
...
-	// IgnoredNodes is an optional set of nodes that should be ignored if
-	// encountered during path finding.
-	IgnoredNodes map[route.Vertex]struct{}
-
-	// IgnoredEdges is an optional set of edges that should be ignored if
-	// encountered during path finding.
-	IgnoredEdges map[EdgeLocator]struct{}
+	// ProbabilitySource is a callback that is expected to return the
+	// success probability of traversing the channel from the node.
+	ProbabilitySource func(route.Vertex, EdgeLocator) float64
```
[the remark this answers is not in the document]
joostjager: It is indeed a hard restriction anymore, but a cost factor. Although probabilities can still be 0, in which case it would be a restriction again. To me it is mainly semantics. I prefer to leave it here and if more cost factor controls are added, extract it then.

routing/pathfind.go
```diff
@@ -464,6 +472,13 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
 		// of the route must succeed.
 		probability := toNodeDist.probability * edgeProbability
 
+		// If the probability is below the specified lower bound, we can
+		// abandon this direction. Adding further nodes can only lower
+		// the probability more.
+		if probability < r.MinProbability {
```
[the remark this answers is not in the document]
joostjager: In case `MinProbability` is 0, it still has a function.

joostjager:
@halseth ptal

routing/pathfind.go
```diff
@@ -453,20 +456,32 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
 			return
 		}
 
+		// Calculate total probability of successfully reaching target
+		// by multiplying the probabilities. Both this edge and the rest
+		// of the route must succeed.
+		probability := toNodeDist.probability * edgeProbability
```
cfromknecht: is there a point to continuing here if `probability` is < min allowed prob?

lnrpc/routerrpc/config_active.go
```diff
@@ -45,3 +63,24 @@ type Config struct {
 	// main rpc server.
 	RouterBackend *RouterBackend
 }
+
+// DefaultConfig defines the config defaults.
+var DefaultConfig = Config{
```
cfromknecht: would recommend defining this a function that returns the default config, o/w it can be modified by importers arbitrarily

joostjager: I am not too worried about that, but it is an improvement. Fixed.

joostjager:
@cfromknecht ptal

Roasbeef:
Tried this PR out on mainnet, was unable to send a payment, and the node ended up OOMing, investigating.

routing/missioncontrol.go
```diff
@@ -279,9 +282,166 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	}
+
+	timeSinceLastFailure := m.now().Sub(lastFailure)
+
+	// Calculate success probability. It is an exponential curve that brings
+	// the probability down to zero when a failure occurs. From there it
+	// recovers asymptotically back to 1. The rate at which this happens is
+	// controlled by the penaltyHalfLife parameter.
+	exp := -timeSinceLastFailure.Hours() / m.cfg.PenaltyHalfLife.Hours()
+	probability := 1 - math.Pow(2, exp)
```
Roasbeef: May be useful to log the probability on the trace level for debugging purposes (with some of the surrounding context).

Roasbeef:
Heap profile of the run:
```
         .          .    631:
         .          .    632:           // Add the next hop to the list of path edges.
    1.80GB     1.80GB    633:           pathEdges = append(pathEdges, nextNode)
         .          .    634:
```

Seems like it's entered an infinite loop in `findPath`, the interesting thing is: it never even dispatches an outgoing payment.

Roasbeef:
Looks like the route that it's trying to reverse has a repeated node, causing an infinite loop.

Roasbeef:
Was able to attempt payments after applying this commit f84cecb7d2a614108375b37bae3824ba11c759b4

Roasbeef:
The first payment that worked after applying that commit was 9 hops, and each of the hops had zero fees. It's interesting that the prior algorithm wouldn't have taken this path since it was more aggressive, but what about this algo causes it to explore more routes before converging on this "shortest route".

cmd/lncli/cmd_query_mission_control.go
```diff
@@ -0,0 +1,35 @@
...
+	client := routerrpc.NewRouterClient(conn)
+
+	req := &routerrpc.QueryMissionControlRequest{}
+	rpcCtx := context.Background()
+	snapshot, err := client.QueryMissionControl(rpcCtx, req)
+	if err != nil {
+		return err
+	}
+
+	printRespJSON(snapshot)
```
Roasbeef: We should have a custom encoding here for the node pubkey, otherwise it shows up as base58.

lnrpc/routerrpc/router.proto
```diff
@@ -209,6 +209,46 @@ message ChannelUpdate {
...
+
+message QueryMissionControlRequest {}
+
+/// QueryMissionControlResponse contains mission control state per node.
+message QueryMissionControlResponse {
+    repeated NodeHistory nodes = 1;
+}
+
+/// NodeHistory contains the mission control state for a particular node.
+message NodeHistory {
```
Roasbeef: These protos should set `json_name` so it's consistent with the display of the rest of our fields on the command line.

routing/pathfind.go
```diff
@@ -331,26 +350,18 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
 		incomingCltv:    0,
+		probability:     1,
 	}
 
 	// We'll use this map as a series of "next" hop pointers. So to get
 	// from `Vertex` to the target node, we'll take the edge that it's
 	// mapped to within `next`.
 	next := make(map[route.Vertex]*channeldb.ChannelEdgePolicy)
 
-	ignoredEdges := r.IgnoredEdges
```
Roasbeef: So no more edge/node black listing?

routing/missioncontrol.go
joostjager: That will be a lot of logging. One line per evaluated edge.

joostjager:
>Looks like the route that it's trying to reverse has a repeated node, causing an infinite loop.

Infinite loop was caused by a regression. A node was also replaced in the distance map when the distance was equal, while it should only be replaced when the distance is lower. Fixed.

lnrpc/routerrpc/router.proto
joostjager: Isn't this only relevant for the rest interface, which we don't have yet?

cmd/lncli/cmd_query_mission_control.go
joostjager: Added custom encoding. But in general I don't like using two types of json encoders. I think we should consider making a tiny fork of `protobuf` to modify `jsonpb` so that it duplicates slice fields with a `Hex` version.

lnrpc/routerrpc/router.proto
joostjager: Ok, see what you mean now. But this is not possible anymore because we use the custom encoder now.

joostjager:
@roasbeef ptal

routing/pathfind.go
```diff
@@ -453,20 +456,35 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
+		// probability.
+		tempDist := getProbabilityBasedDist(
+			tempWeight, probability, int64(r.PaymentAttemptPenalty),
+		)
+
+		// If the current best route is better than this candidate
+		// route, return. It is important to also return if the distance
+		// is equal, because otherwise the algorithm could run into an
+		// endless loop.
+		if distance[fromVertex].dist != infinity &&
```
halseth: Is this infinity check necessary?

halseth: Is this infinity check necessary?

routing/missioncontrol.go
halseth: I think that's okay on trace

lnrpc/routerrpc/router.proto
```diff
@@ -213,6 +213,43 @@ message ResetMissionControlRequest{}
...
+/// NodeHistory contains the mission control state for a particular channel.
+message ChannelHistory {
+    /// Short channel id
+    uint64 channel_id = 1;
+    
+    /// Time stamp of last failure.
+    int64 last_fail_time = 2;
+
+    /// Minimum penalization amount.
+    int64 min_penalize_amt = 3;
```
halseth: should have `sat` or `msat` suffix

lnrpc/routerrpc/router.proto
```diff
@@ -213,6 +213,43 @@ message ResetMissionControlRequest{}
...
+    repeated NodeHistory nodes = 1;
+}
+
+/// NodeHistory contains the mission control state for a particular node.
+message NodeHistory {
+    /// Node pubkey
+    bytes pubkey = 1;
+
+    /// Time stamp of last failure. Set to zero if no failure happened yet.
+    int64 last_fail_time = 2;
```
halseth: `_unix` suffix perhaps?

routing/pathfind.go
joostjager: Yes, not anymore indeed. I can't remember why I added it in the first place. But it seems it would actually work worse with the infinity check, because it would add a useless edge to the map.

lnrpc/routerrpc/router.proto
joostjager: Do we have a final guideline on that? We currently use the suffix pretty randomly

lnrpc/routerrpc/router.proto
joostjager: I don't know. I think we may first want to discuss what our standards should be regarding suffixes. It isn't consistent at the moment.

routing/missioncontrol.go
joostjager: Log added

routing/pathfind.go
joostjager: Fixed

joostjager:
@halseth ptal

lnrpc/routerrpc/router.proto
joostjager: These fields can always be renamed without breaking backwards compatibility

lnrpc/routerrpc/router.proto
cfromknecht: for amounts the standard is to always have the suffix, so it's clear w/o needing to turn to documentation. some of them do not have it since they were preexisting, and it would break the API (though not the wire serialization)

lnrpc/routerrpc/router.proto
cfromknecht: idt we have a standard for timestamps. note that renaming the field does break API compatibility, but not serialization

lnrpc/routerrpc/router.proto
joostjager: `sat` added

lnrpc/routerrpc/router.proto
joostjager: Oh, that is true indeed. If people rebuild their apps with the new proto, it will break.

lnrpc/routerrpc/router.proto
joostjager: Added `json_name` anyway, although still not sure who is going to benefit from this.

lntest/node.go
```diff
@@ -248,6 +249,10 @@ type HarnessNode struct {
 	lnrpc.WalletUnlockerClient
 
 	invoicesrpc.InvoicesClient
+
+	// RouterClient cannot be embedded, because a name collision would occur
+	// on the main rpc SendPayment.
+	RouterClient routerrpc.RouterClient
```
halseth: We have the same problem on mobile actually, it creates an API conflict if it is compiled with the router subserver. Should we consider changing the API name? Especially if we want to to enable it by default at some point.

routing/missioncontrol.go
```diff
@@ -298,7 +235,115 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+	m.Lock()
+	defer m.Unlock()
+
+	// Get the history for this node. If there is no history available,
+	// assume that it's success probability is 1. After the attempt new
+	// information becomes available to adjust this probability.
+	nodeHistory, ok := m.history[fromNode]
+	if !ok {
+		return 1
```
halseth: should we use the default/configurable value her?

routing/missioncontrol.go
```diff
@@ -298,7 +235,115 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	// Calculate the last failure of the given edge. A node failure is
+	// considered a failure that would have affected every edge. Therefore
+	// we insert a node level failure into the history of every channel.
+	lastFailure := nodeHistory.lastFail
+
+	// Take into account a minimum penalize amount. For balance errors, a
+	// failure may be reported with such a minimum to prevent too aggresive
+	// penalization.
+	channelHistory, ok := nodeHistory.channelLastFail[channelID]
+	if ok && channelHistory.minPenalizeAmt <= amt {
```
halseth: This logic is a bit hard to follow. Can you add some comments?

routing/payment_session.go
```diff
@@ -28,15 +27,15 @@ type PaymentSession interface {
...
+	ReportEdgeFailure(failedEdge edge, minPenalizeAmt lnwire.MilliSatoshi)
 
 	// ReportEdgePolicyFailure reports to the PaymentSession that we
 	// received a failure message that relates to a channel policy. For
 	// these types of failures, the PaymentSession can decide whether to to
 	// keep the edge included in the next attempted route. The
 	// PaymentSession will use this information to produce a better next
 	// route.
-	ReportEdgePolicyFailure(errSource route.Vertex, failedEdge *EdgeLocator)
+	ReportEdgePolicyFailure(failedEdge edge)
```
halseth: Does this change penalization from node to "node pair"?

routing/payment_session.go
```diff
@@ -28,15 +27,15 @@ type PaymentSession interface {
 	// channel failed to route the previous payment attempt. The
 	// PaymentSession will use this information to produce a better next
 	// route.
-	ReportEdgeFailure(e *EdgeLocator)
+	ReportEdgeFailure(failedEdge edge, minPenalizeAmt lnwire.MilliSatoshi)
```
halseth: should have a corresponding doc change.

lnrpc/routerrpc/router_server.go
```diff
@@ -453,3 +457,48 @@ func (s *Server) ResetMissionControl(ctx context.Context,
...
+				SuccessProb: float32(channel.SuccessProb),
+			}
+		}
+
+		var lastFail int64
+		if node.LastFail != nil {
+			lastFail = node.LastFail.Unix()
+		}
+
+		rpcNode := NodeHistory{
```
halseth: style nit: 
```go
rpcNodes[i] = &NodeHistory{
```

cmd/lncli/cmd_query_mission_control.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	type displayNodeHistory struct {
+		Pubkey               string
+		LastFailTime         int64
+		OtherChanSuccessProb float32
+		Channels             []*routerrpc.ChannelHistory
+	}
+
+	displayResp := struct {
```
halseth: style nit: I doubt we would see any performance downsides by just doing 
```go
displayResp := struct {
        Nodes []displayNodeHistory
}{}

	for i, n := range snapshot.Nodes {
		displayResp.Nodes = append(displayResp.Nodes,  displayNodeHistory{
```

server.go
```diff
@@ -17,6 +17,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
```
halseth: import should be moved.

routing/missioncontrol.go
halseth: Ok, this is done in later commit. Would be nice if the apriori value was defined before this, but not a blocker.

routing/pathfind.go
```diff
@@ -614,5 +635,51 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
 			"too many hops")
 	}
 
+	log.Infof("Found route with probability: %v\n", distance[source].probability)
```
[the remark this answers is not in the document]
joostjager: Added

routing/missioncontrol.go
joostjager: Moved into this commit.

lntest/node.go
joostjager: Maybe, but definitely not in this PR :)

routing/missioncontrol.go
joostjager: added

routing/payment_session.go
joostjager: No there is no change here, just a different struct that contains both parameters.

cmd/lncli/cmd_query_mission_control.go
joostjager: simplified

joostjager:
@halseth ptal

PR #2806, opened 2019-03-20
chainntnfs: fix RegisterSpend height hint logs

wpaulino:


chainntnfs/txnotifier.go
```diff
@@ -840,7 +840,7 @@ func (n *TxNotifier) RegisterSpend(ntfn *SpendNtfn) (*HistoricalSpendDispatch,
 	if err == nil {
 		if hint > startHeight {
 			Log.Debugf("Using height hint %d retrieved from cache "+
-				"for %v", startHeight, ntfn.SpendRequest)
+				"for %v", hint, ntfn.SpendRequest)
```
halseth: Would still be useful to log the OG height I think.

PR #2813, opened 2019-03-21
Tor option to override target IP address

wereHamster:
This option is required when the Tor server is not running on the same host as lnd. Otherwise peers can't connect to lnd through the onion service.

For example let's say you have a Kubernetes cluster and have a single Tor server which can be used by all the internal services. Then you can start lnd with `--tor.active --tor.v3 --tor.sock=tor:9050 --tor.control=tor:9051 --tor.targetipaddress=$MY_POD_ID`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -219,6 +219,7 @@ type torConfig struct {
 	DNS             string `long:"dns" description:"The DNS server as host:port that Tor will use for SRV queries - NOTE must have TCP resolution enabled"`
 	StreamIsolation bool   `long:"streamisolation" description:"Enable Tor stream isolation by randomizing user credentials for each connection."`
 	Control         string `long:"control" description:"The host:port that Tor is listening on for Tor control connections"`
+	TargetIPAddress string `long:"targetipaddress" description:"IP address that Tor should use as the target of the hidden service"`
 	V2              bool   `long:"v2" description:"Automatically set up a v2 onion service to listen for inbound connections"`
 	V3              bool   `long:"v3" description:"Automatically set up a v3 onion service to listen for inbound connections"`
 	PrivateKeyPath  string `long:"privatekeypath" description:"The path to the private key of the onion service being created"`
--- a/tor/controller.go
+++ b/tor/controller.go
@@ -81,12 +81,17 @@ type Controller struct {
 
 	// version is the current version of the Tor server.
 	version string
+
+	// The IP address which we tell the Tor server to use to connect to the LND node.
+	// This is required when the Tor server runs on another host, otherwise the service
+	// will not be reachable.
+	targetIPAddress string
 }
 
 // NewController returns a new Tor controller that will be able to interact with
 // a Tor server.
-func NewController(controlAddr string) *Controller {
-	return &Controller{controlAddr: controlAddr}
+func NewController(controlAddr string, targetIPAddress string) *Controller {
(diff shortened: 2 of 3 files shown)


config.go
```diff
@@ -168,6 +168,7 @@ type torConfig struct {
 	DNS             string `long:"dns" description:"The DNS server as host:port that Tor will use for SRV queries - NOTE must have TCP resolution enabled"`
 	StreamIsolation bool   `long:"streamisolation" description:"Enable Tor stream isolation by randomizing user credentials for each connection."`
 	Control         string `long:"control" description:"The host:port that Tor is listening on for Tor control connections"`
+	TargetIPAddress string `long:"targetipaddress" description:"IP address that Tor should use as the target of the hidden service"`
```
wpaulino: Can we actually just use the host as part of the SOCKS flag rather than introducing a new flag?

1043465747:
ترجمة


في الجمعة، ٣٠ أغسطس، ٢٠١٩ ١٢:١٧ ص Wilmer Paulino <notifications@github.com>
كتب:

wereHamster:
I think you misunderstood the `TargetIPAddress` setting. It's the IP address of lnd from the point of view of the tor node.

If you have a setup where the lnd and tor nodes have a separate ip address (eg. when running in separate pods in a kubernetes cluster), then `--tor.socks` and `--tor.control` contain the IP address of the tor node. And `--tor.targetipaddress` would contain the ip address of the lnd node.

In my setup, I start lnd with something like this: `lnd … --tor.socks=tor:9050 --tor.control=tor:9051 --tor.targetipaddress=$MY_POD_IP`. `tor` is the hostname of the tor node, which is reachable from anywhere in the cluster, and `$MY_POD_IP` contains the cluster-wide public IP of the container where lnd is running.

tor/controller.go
```diff
@@ -81,12 +81,19 @@ type Controller struct {
 
 	// version is the current version of the Tor server.
 	version string
+
+	// The IP address which we tell the Tor server to use to connect to the LND node.
+	// This option optional. It is set through the '--tor.targetipaddress' commandline
```
wpaulino: We can omit this line and the one below because this can be used elsewhere without the context of the CLI flag.

wpaulino:
@wereHamster could you also rebase this on the current master branch?

wereHamster:
@wpaulino done

gkrizek:
@wereHamster I came across this PR when trying to do a similar thing. Curious, does your Tor service running on Kubernetes need to be exposed on a public IP? Are you exposing it via `NodePort` with the host's IP?

wereHamster:
@gkrizek my tor service is a … service. Accessible through a cluster-wide IP address (10.63.x.y), as all services are, not through a public IP.

gkrizek:
@wereHamster I was meaning the communication between the Tor Service and the Tor Network. Does your Tor Service need to be publicly accessible for communication with the Tor network or do you have NATs in place?

I understand how the internal Kubernetes communication works.

wereHamster:
@gkrizek I'm not using any NAT, the tor node does not use a public IP address. Incoming connections should work nonetheless. This is one of the points of tor, that you don't have to expose your public address even if you want to run a onion service.

gkrizek:
@wereHamster right, was just curious how network communication was working from your cluster out to the internet for Tor connectivity. But this makes sense. Thanks for the help.

PR #2819, opened 2019-03-22
peer: catch write timeouts, retry with backoff

### LINKED ISSUE #2784: Peer connections flapping
### Background

Updated to 0.5.2-99-beta commit=queue/v1.0.1-23-g36cc1da8ea0eec6be4559c673bc07bd0ae9be66a

And started back up.  It can't seem to sync, and keeps disconnecting 10-20 peers every few minutes while 'applying gossip update'

### Your environment

* version of `lnd` : 0.5.2-99-beta commit=queue/v1.0.1-23-g36cc1da8ea0eec6be4559c673bc07bd0ae9be66a

* which operating system (`uname -a` on *Nix) Rasbian

### Steps to reproduce

Start node

### Expected behaviour
Runs normally

### Actual behaviour
"Applying update" stalls for 10+ seconds... then spam of peer connections failing.  
All of these peers were working find before the upgrade.  There seems to be an endless loop of trying to get the gossip... causes peers to disconnect... when reconnect tries to get gossip... which causes peers to disconnect.
It's like the gossiper is causing peers to time out... which is causing the gossiper to start over when the channel comes up and rinse/repeat.

2019-03-15 23:27:41.020 [INF] DISC: gossipSyncer(03abf6f44c355dec0d5aa155bdbdd6e0c8fefe318eff402de65c6eb2e1be55dc3e): applying new update horizon: start=2019-03-15 22:26:59 +0700 +07, end=2155-04-22 04:55:14 +0700 +07, backlog_size=21999
2019-03-15 23:27:55.228 [INF] DISC: Broadcasting batch of 905 new announcements
2019-03-15 23:27:55.336 [INF] PEER: Disconnecting 213.174.156.76:9735, reason: unable to write message: write tcp 10.34.0.85:43314->213.174.156.76:9735: write: connection reset by peer
2019-03-15 23:27:55.348 [INF] PEER: Disconnecting 46.229.165.138:9735, reason: unable to write message: write tcp 10.34.0.85:55744->46.229.165.138:9735: write: connection reset by peer
2019-03-15 23:27:55.360 [INF] PEER: Disconnecting 46.229.165.133:9735, reason: unable to write message: write tcp 10.34.0.85:39280->46.229.165.133:9735: write: connection reset by peer
2019-03-15 23:27:55.430 [INF] PEER: Disconnecting 46.229.165.154:9735, reason: unable to write message: write tcp 10.34.0.85:49280->46.229.165.154:9735: write: connection reset by peer

cfromknecht:
Fixes #2784.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -45,7 +45,7 @@ const (
 	idleTimeout = 5 * time.Minute
 
 	// writeMessageTimeout is the timeout used when writing a message to peer.
-	writeMessageTimeout = 50 * time.Second
+	writeMessageTimeout = 5 * time.Second
 
 	// readMessageTimeout is the timeout used when reading a message from a
 	// peer.
--- a/peer.go
+++ b/peer.go
@@ -1005,7 +1005,12 @@ func (p *peer) readHandler() {
 out:
 	for atomic.LoadInt32(&p.disconnect) == 0 {
 		nextMsg, err := p.readNextMessage()
-		idleTimer.Stop()
+		if !idleTimer.Stop() {
+			select {
+			case <-idleTimer.C:
+			default:
+			}
+		}
 		if err != nil {
 			peerLog.Infof("unable to read message from %v: %v",
 				p, err)
@@ -1427,6 +1432,14 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
 //
 // NOTE: This method MUST be run as a goroutine.
 func (p *peer) writeHandler() {
+	// We'll stop the timer after a new messages is sent, and also reset it
+	// after we process the next message.
+	idleTimer := time.AfterFunc(idleTimeout, func() {
+		err := fmt.Errorf("Peer %s no write for %s -- disconnecting",
+			p, idleTimeout)
+		p.Disconnect(err)
+	})
+
 	var exitErr error
 
 	const (
@@ -1500,6 +1513,16 @@ out:
 
 				goto retryWithDelay
 			}
+
+			// The write succeeded, reset the idle timer to prevent
+			// us from disconnecting the peer.
+			if !idleTimer.Stop() {
+				select {
+				case <-idleTimer.C:
+				default:
+				}
+			}
+			idleTimer.Reset(idleTimeout)
 
 			// If the peer requested a synchronous write, respond
 			// with the error.
--- a/peer.go
+++ b/peer.go
@@ -1429,27 +1429,80 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
 func (p *peer) writeHandler() {
 	var exitErr error
 
+	const (
+		minRetryDelay = 5 * time.Second
+		maxRetryDelay = time.Minute
+	)
+
 out:
 	for {
 		select {
 		case outMsg := <-p.sendQueue:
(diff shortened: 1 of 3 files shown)


peer.go
```diff
@@ -1005,7 +1005,9 @@ func (p *peer) readHandler() {
 out:
 	for atomic.LoadInt32(&p.disconnect) == 0 {
 		nextMsg, err := p.readNextMessage()
-		idleTimer.Stop()
+		if !idleTimer.Stop() {
```

peer.go
```diff
@@ -1427,29 +1429,97 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
...
+				if retryDelay == 0 {
+					retryDelay = minRetryDelay
+				} else {
+					retryDelay *= 2
+					if retryDelay > maxRetryDelay {
+						retryDelay = maxRetryDelay
+					}
+				}
+
+				peerLog.Infof("Write timeout detected for "+
```
Roasbeef: Should probably demote this to debug logging level. It can be spammy with the current set of network peers.

cfromknecht: tbh peer debug logs are useless for just viewing, they're only useful when grepping because there's 1000s of logs/s. what if we also demote gossip send/recv logs to trace? stilling keeping important ones like htlc forwards in debug?

wpaulino: Also log peer public key?

peer.go
```diff
@@ -1429,27 +1431,72 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
...
+			// Initialize a retry delay of zero, which will be
+			// increased if we encounter a write timeout on the
+			// send.
+			var retryDelay time.Duration
+		retryWithDelay:
+			if retryDelay > 0 {
+				select {
+				case <-time.After(retryDelay):
+				case <-p.quit:
+					exitErr = ErrPeerExiting
```
wpaulino: Shouldn't we be sending this error back to the caller as well if they asked for a synchronized send?

wpaulino: Ah never mind, I see it in the follow up commit.

peer.go
```diff
@@ -1429,6 +1429,14 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
 //
 // NOTE: This method MUST be run as a goroutine.
 func (p *peer) writeHandler() {
+	// We'll stop the timer after a new messages is received, and also
```
wpaulino: Nit: s/received/sent/

peer.go
cfromknecht: Fixed and fixed

peer.go
```diff
@@ -1429,27 +1429,80 @@ func (p *peer) writeMessage(msg lnwire.Message) error {
 func (p *peer) writeHandler() {
 	var exitErr error
 
+	const (
+		minRetryDelay = 5 * time.Second
```
Roasbeef: I think we can remove this in favor of just using `writeMessageTimeout` here since it's added in the next commit and don't need to rack the value in two places in the same file.

cfromknecht: they're different values really, just coincidence that they were chosen to be the same value. in my mind it's not clear that they should always assume the same?

PR #2820, opened 2019-03-22
wtclient: session private key derivation

cfromknecht:
In this PR, we address some of the last major architectural changes planned for the `wtclient` before it is ready to be integrated with its database and into the lnd daemon.  The primary component is encompassed by deriving the session private keys from the user's seed, which are used to encrypt brontide traffic with the tower and authenticate requests. To do so, we add a new key family:
```golang
 	// KeyFamilyTowerSession is the family of keys that will be used to
	// derive session keys when negotiation sessions with watchtowers. The
	// session keys are limited to the lifetime of the session and are used
	// to increase privacy in the watchtower protocol.
	KeyFamilyTowerSession KeyFamily = 8
```

The initial version of the wtclient #2618 generated these randomly and stored them in plaintext in the mock database to keep the scope of the initial PR from growing too much. Here we resolve this, ensuring we don't need to store the private keys on disk.

In addition, we also improve the `CreateSession` flow, namely by:
 1. Returning the current `LastApplied` value for the session if it exists and has a non-zero number of updates.
 1. Allowing clients to overwrite an existing session if the session has no submitted updates

The first improvement is useful in detecting cases of data loss, where a client tries to create a session with the same tower but that session already has updates. When keys are generated randomly, the probability of this occurring is negligible. However with deterministic derivation, this could happen if the client uses the same tower and the indexes collide. 

The rationale behind the second is more subtle, but it permits more asynchrony between the client and server during session creation, and permits the client to tolerate edge cases with less persistent state.

The current flow will reject if a session already exists. However, if the client is unable to receive the `CreateSessionReply`, the client must remember what policy and session key indexes it used so that it can continue using the session if the tower says it already exists. Doing so requires that the client store a tuple of (tower, session key index, policy). If the tower claims the session already exists and the `LastApplied` value is zero, then the client would continue using the tower.

The difficulty here is that failure to use the exact same policy would result in the signatures in the updates being invalid, and only detectable at the time of the broadcast. By allowing the session to be overwritten while the session has no updates, the tower is able to use first state update as an acknowledgment that the client has successfully persisted the session. This allows the client to only store a (tower, session key index) tuple for each attempted session, as the policy will implicitly resolve to whichever it uses last since the tower will return `CodeOK` for the new session. This is true even if the client crashes and is restarted with a different policy, assuming the new policy is also supported by the tower. Furthermore, the client only ever has to reserve one session key index at a time for a given tower, as opposed to one for each (tower, policy) pair.

Along the way, we make some miscellaneous improvements:
 1. Checking blob type before attempting to create a reward address on the tower side
 2. Printing error codes with their human readable name instead of their number on the tower side
 3. Add the `DeleteSessionCodeNotFound` string to the `wtwire.ErrorCode` list, as a follow up from #2779 
 4. Remove storage of the `Tower` and `SessionPrivKey` from the mock db, since these are now generated on startup. The existing unit tests in the client package assert that they are regenerated on startup during the client restart test
 5. Fix a bug that could cause the client not to force quit during session creation.

NOTE: If #2783 is merged before this, the `wtdb.TowerDB` unit tests will be need modified to test the session overwriting.

### DIFF
--- a/watchtower/wtclient/client_test.go
+++ b/watchtower/wtclient/client_test.go
@@ -430,10 +430,11 @@ func newHarness(t *testing.T, cfg harnessCfg) *testHarness {
 		Dial: func(string, string) (net.Conn, error) {
 			return nil, nil
 		},
-		DB:           clientDB,
-		AuthDial:     mockNet.AuthDial,
-		PrivateTower: towerAddr,
-		Policy:       cfg.policy,
+		DB:            clientDB,
+		AuthDial:      mockNet.AuthDial,
+		SecretKeyRing: wtmock.NewSecretKeyRing(),
+		PrivateTower:  towerAddr,
+		Policy:        cfg.policy,
 		NewAddress: func() ([]byte, error) {
 			return addrScript, nil
 		},
--- a/watchtower/wtclient/client_test.go
+++ b/watchtower/wtclient/client_test.go
@@ -379,10 +379,11 @@ type testHarness struct {
 }
 
 type harnessCfg struct {
-	localBalance    lnwire.MilliSatoshi
-	remoteBalance   lnwire.MilliSatoshi
-	policy          wtpolicy.Policy
-	noRegisterChan0 bool
+	localBalance       lnwire.MilliSatoshi
+	remoteBalance      lnwire.MilliSatoshi
+	policy             wtpolicy.Policy
+	noRegisterChan0    bool
+	noAckCreateSession bool
 }
 
 func newHarness(t *testing.T, cfg harnessCfg) *testHarness {
@@ -414,6 +415,7 @@ func newHarness(t *testing.T, cfg harnessCfg) *testHarness {
 		NewAddress: func() (btcutil.Address, error) {
 			return addr, nil
 		},
+		NoAckCreateSession: cfg.noAckCreateSession,
 	}
 
 	server, err := wtserver.New(serverCfg)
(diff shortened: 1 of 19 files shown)


watchtower/wtclient/client_test.go
```diff
@@ -1099,6 +1101,111 @@ var clientTests = []clientTest{
...
+			// creation.
+			h.server.Stop()
+			h.serverCfg.NoAckCreateSession = false
+			h.startServer()
+			defer h.server.Stop()
+
+			// Restart the client with a new policy, which will
+			// immediately try to overwrite the prior session with
+			// the old policy.
+			h.clientCfg.Policy.SweepFeeRate = 2
```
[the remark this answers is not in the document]
cfromknecht: Fixed

watchtower/wtserver/create_session.go
```diff
@@ -54,22 +54,26 @@ func (s *Server) handleCreateSession(peer Peer, id *wtdb.SessionID,
...
-	if err != nil {
-		log.Errorf("unable to generate reward script for %s", id)
-		return s.replyCreateSession(
-			peer, id, wtwire.CodeTemporaryFailure, nil,
-		)
+	var rewardScript []byte
+	if req.BlobType.Has(blob.FlagReward) {
+		rewardAddress, err := s.cfg.NewAddress()
+		if err != nil {
+			log.Errorf("unable to generate reward addr for %s", id)
```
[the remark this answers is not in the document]
cfromknecht: Also updated to log errors in both instances

cfromknecht:
comments addressed, @wpaulino ptal

watchtower/wtmock/client_db.go
```diff
@@ -110,7 +110,7 @@ func (m *ClientDB) CreateClientSession(session *wtdb.ClientSession) error {
 		Policy:           session.Policy,
 		SeqNum:           session.SeqNum,
 		TowerLastApplied: session.TowerLastApplied,
-		RewardPkScript:   session.RewardPkScript,
+		RewardPkScript:   cloneBytes(session.RewardPkScript),
```
Roasbeef: Is this diff a result of a bug that occurs if the bytes aren't copied?

watchtower/wtserver/create_session.go
```diff
@@ -21,11 +21,15 @@ func (s *Server) handleCreateSession(peer Peer, id *wtdb.SessionID,
 	existingInfo, err := s.cfg.DB.GetSessionInfo(id)
 	switch {
 
+	// We already have a session, though it is currently unused. We'll allow
+	// the client to recommit the session if it wanted to change the policy.
+	case err == nil && existingInfo.LastApplied == 0:
```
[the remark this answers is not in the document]
Roasbeef: Add a break here maybe to make it a bit more clear that this case is just meant to be instructive to the reader?

watchtower/wtdb/client_session.go
```diff
@@ -29,6 +29,15 @@ var (
 	// LastApplied value greater than any allocated sequence number.
 	ErrUnallocatedLastApplied = errors.New("tower echoed last appiled " +
 		"greater than allocated seqnum")
+
+	// ErrNoReservedKeyIndex signals that a client session could not be
+	// created because no session key index was reserved.
+	ErrNoReservedKeyIndex = errors.New("key index not reserved")
```
Roasbeef: Don't see any new tests for this added though the mock has been updated to return this new value.

watchtower/wtmock/client_db.go
cfromknecht: not that i am aware of, but i think it better mirrors the behavior of the real db. we do this in a few other places when copying byte slices, and figured i'd add it as well for reward pkscripts

watchtower/wtdb/client_session.go
cfromknecht: the mock client db doesn't have any unit tests at this point, they will be added in the follow up client db PR. these should never be hit when testing the client as a block box, which is why the current set of high-level client tests can't/don't trigger it

PR #2821, opened 2019-03-22
routerrpc: share payment parsing code from main rpc

joostjager:
This PR is a preparation for the new reliable send payment rpc in routerrpc.

Last debug commit will be dropped after review.


joostjager:
Closing this PR, because the new API introduced in #2973 is too different from the existing one to benefit from code reuse.

Issue #2823, opened 2019-03-22
Channel from listchannels has chan_id as "0"


LNBIG-COM:
### Background

I found strange channels - they are not active but they are in listchannels and they are missing in closedchannels

### Your environment

* version of `lnd` 0.5.2-99-beta commit=queue/v1.0.1-106-g9143067014bcde14f1e8f7eb515bae1ccccb97c5
* which operating system (`uname -a` on *Nix) CentOS 6
* version of `btcd`, `bitcoind`, or other backend 0.17.1

### Actual behaviour

For example lnd-41

    {
        "active": false,
        "remote_pubkey": "028c2f98b483bccf9e2634d6d5dca5cc6606ed6afd17d209da1a95f4b138d11a87",
        "channel_point": "c47516fcca67ea4501ac05b2e3eeae3ad20ef1a7e178dcb3ee47b2bc1ffc1acb:0",
        "chan_id": "0",
        "capacity": "2000000",
        "local_balance": "1997276",
        "remote_balance": "0",
        "commit_fee": "2724",
        "commit_weight": "600",
        "fee_per_kw": "3763",
        "unsettled_balance": "0",
        "total_satoshis_sent": "0",
        "total_satoshis_received": "0",
        "num_updates": "0",
        "pending_htlcs": [
        ],
        "csv_delay": 240,
        "private": false,
        "initiator": true
    },

This channel is missing in feereport, closedchannels. What is this? And i have second there channel with chan_id as 0

    {
        "active": false,
        "remote_pubkey": "02005dcac896582db30f259c17bba1849a9e1127f60d120ec7db69c17ce64d749d",
        "channel_point": "0078eb89e787a676afa940e36c8871fa0f7bb5a21ad00a886e12c59b83c16fe1:0",
        "chan_id": "0",
        "capacity": "2000000",
        "local_balance": "1997276",
        "remote_balance": "0",
        "commit_fee": "2724",
        "commit_weight": "600",
        "fee_per_kw": "3763",
        "unsettled_balance": "0",
        "total_satoshis_sent": "0",
        "total_satoshis_received": "0",
        "num_updates": "0",
        "pending_htlcs": [
        ],
        "csv_delay": 240,
        "private": false,
        "initiator": true
    },

I will send now the full encrypted logs. Common feature of all these channels in logs look like `...too-long-mempool-chain, too many descendants...`

wpaulino:
[...]

PR #2824, opened 2019-03-22
rpcserver: retrieve Channel ChanId from OpenChannel record

wpaulino:
This is a better alternative than retrieving it from the graph as it's
possible that the channel is pruned from it doing to not having any
updates within the past two weeks.

Fixes https://github.com/lightningnetwork/lnd/issues/2823.


cfromknecht:
@wpaulino race test needs a restart

PR #2842, opened 2019-03-27
signal: catch all termination signals by default

Roasbeef:
In this commit, we modify the primary `signal` package to instead catch
all signals. Before this commit, it would only catch the interrupt
signal sent from the kernel. With this new commit, we'll now also catch
(or attempt to catch): `SIGABRT`, `SIGTERM`, `SIGSTOP`, and `SIGQUIT`.

### DIFF
--- a/signal/signal.go
+++ b/signal/signal.go
@@ -8,6 +8,7 @@ package signal
 import (
 	"os"
 	"os/signal"
+	"syscall"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


Roasbeef:
Before this PR, we didn't attempt to catch `SIGTERM` which is customarily used by `systemd` to shutdown active services. Many of the raspberry pi guides circulating out there use a `ssytemd` set up to handle creation/destruction of the `lnd` process.

signal/signal.go
```diff
@@ -26,7 +27,11 @@ var (
 )
 
 func init() {
-	signal.Notify(interruptChannel, os.Interrupt)
+	signalsToCatch := []os.Signal{
+		os.Interrupt, os.Kill, syscall.SIGABRT, syscall.SIGTERM,
+		syscall.SIGSTOP, syscall.SIGQUIT,
```
cfromknecht: style nit: new line for each signal?

Roasbeef: Fixed, also we'll now log the signal itself.

vamp111:
expected update reliese anytime soon ?

PR #2845, opened 2019-03-27
routerrpc: register router sub server logger

joostjager:
This PR register the `routerrpc` sub server logger so that its log level can be set.

### DIFF
--- a/log.go
+++ b/log.go
@@ -10,7 +10,7 @@ import (
 	"github.com/btcsuite/btclog"
 	"github.com/jrick/logrotate/rotator"
 	"github.com/lightninglabs/neutrino"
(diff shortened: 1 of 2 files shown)


log.go
```diff
@@ -108,6 +109,16 @@ func init() {
 	chainrpc.UseLogger(ntfrLog)
 	invoicesrpc.UseLogger(irpcLog)
 	channelnotifier.UseLogger(chnfLog)
+
+	addSubLogger(routerrpc.Subsystem, routerrpc.UseLogger)
+}
+
+// addSubLogger is a helper method to conveniently register the logger of a sub
+// system.
+func addSubLogger(subsystem string, useLogger func(btclog.Logger)) {
```
halseth: 😍

PR #2847, opened 2019-03-28
channeldb: add reject and channel caches

cfromknecht:
This PR adds two caches housed within the channeldb to optimize two existing hot spots related to gossip traffic.

## Reject Cache

The first is dubbed a _reject cache_, whose entries contain a small amount of information critical to determining if we should spend resources validating a particular channel announcement or channel update. This improves the performance of [`HasChannelEdge`](https://github.com/lightningnetwork/lnd/blob/master/channeldb/graph.go#L601), which is a subroutine of [`KnownEdge`](https://github.com/lightningnetwork/lnd/blob/master/routing/router.go#L2272) and [`IsStaleEdgePolicy`](https://github.com/lightningnetwork/lnd/blob/master/routing/router.go#L2281).

Each entry in the reject cache stores the following:
```golang
type rejectCacheEntry struct {
    upd1Time int64
    upd2Time int64
    flags    rejectFlags
}
```
where `flags` is packed bitfield containing, for now, the `exists` and `isZombie` booleans. We store the time as an unix integer as opposed to the `time.Time` directly, since `time.Time`'s [internal pointer](https://golang.org/src/time/time.go#L147) for storing timezone info would force the garbage collector to traverse these long-lived entries. They are subsequently reconstructed during calls to `HasChannelEdge` using the integral value.

The addition of the reject cache greatly improves LND's ability to efficiently filter gossip traffic, and results in a significantly lower number of database accesses for this purpose. Users should notice the gossip syncers terminating much quicker (due to the absence of db hits) and overall result in snappier connections.

## Channel Cache
The second is _channel cache_ which caches `ChannelEdge` values, and used to reduce memory allocations stemming from [`ChanUpdatesInHorizon`](https://github.com/lightningnetwork/lnd/blob/master/channeldb/graph.go#L1236). A `ChannelEdge` has the following structure:
```golang
type ChannelEdge struct {
    Info    *ChannelEdgeInfo
    Policy1 *ChannelEdgePolicy
    Policy2 *ChannelEdgePolicy
}
```
Currently, each call to `ChanUpdatesInHorizon` will seek and deserialize all `ChannelEdge` values in the requested range. When connected to large number of peers, this can result an excessive amount of memory that must be 1) allocated, and 2) cleaned up by the garbage collector. The values are intended to be read only, and are discarded as soon as the relevant information is written out on the wire to the peers.

As a result, the channel cache can greatly reduce the amount of wasted allocations, especially if a large percentage of the requested range is held in memory or peers request similar time slices of the graph.

## Eviction
Both caches employ randomized eviction when inserting an element would cause the cache to exceed its configured capacity. The rationale stems from the fact that the access pattern for these caches is dictated entirely by our peers. Assuming the entire working set cannot fit in memory, a deterministic caching strategy would ease a malicious peer's ability to craft access patterns that incur a worst-case hit frequency (close-to-or-equal-to 0%). The resulting effect would that we equivalent to having no cache at all, and force us to hit disk for each rejection check and/or deserialize each requested `ChannelEdge`. The randomized eviction strategy thus provides some level of DOS protection, while also being simple to and efficient to implement in go (because map iteration is randomized by default).

## Lazy Consistency
For some cases, keeping the cache in sync with the on-disk state requires reading and deserializing extra data from the db that is not deducible from the inputs. However, at the time the entry is modified, it's not certain that the entry will be accessed again, meaning that extraneous allocations and deserializations may be performed, even though the entry could be evicted before that data is ever used.

For this reason, both the reject and channel caches remove entries from cache whenever an operation dirties an entry and then lazily loads them on the next access. The lone exception is `UpdateChannelPolicy`, where we write through info to the caches if those entries are already present because it is the most frequently used operation. If the entries are not present, then they are lazily loaded on the next access for the reasons stated above.

There are other possible places we could add write through, though removing the entry is by far the safest alternative. We can proceed in doing so in other places if they prove to be a bottleneck.

## CLI configuration
Each cache can be configured by way of the new `lncfg.Caches` subconfig, allowing users to set the maximum number of cache entries for both the reject and channel caches. The configuration options appear as:
```
caches:
      --caches.reject-cache-size=    Maximum number of entries contained in the reject cache, which is used to speed up filtering of new channel announcements and channel updates from peers. Each entry requires 25 bytes. (default: 50000)
      --caches.channel-cache-size=   Maximum number of entries contained in the channel cache, which is used to reduce memory allocations from gossip queries from peers. Each entry requires 2Kb. (default: 20000)
```

At the default values provided, the reject cache occupies 1.2MB and easily holds an entry for today's entire graph. The channel cache occupies about 40MB, holding about half of the channels in memory. The majority of peers query for values near the tip of the graph, allowing gossip queries to be satisfied almost entirely from the in-memory contents.

There are certain peers that request a full dump of all known channels, which will require going to disk (unless of course the channel cache is configured to hold the entire graph in memory). AFAIK, CL currently queries for the entire range on each connection, and after #2740 LND nodes will begin doing so roughly once every six hours to a random peer to ensure that any holes in its routing table are filled. Inherently no caching algorithm can be optimal under such circumstances, though empirically, the channel cache quickly converges back to having most of the elements at tip for subsequent queries on the hot spots.

I'm open to other opinions on the default cache sizes, please discuss if people have others they prefer!

Depends on:
 - [x] #2846

### DIFF
--- a/channeldb/channel_cache.go
+++ b/channeldb/channel_cache.go
@@ -1,0 +1,50 @@
+package channeldb
+
+// channelCache is an in-memory cache used to improve the performance of
+// ChanUpdatesInHorizon. It caches the chan info and edge policies for a
+// particular channel.
+type channelCache struct {
+	n        int
+	channels map[uint64]ChannelEdge
+}
+
+// newChannelCache creates a new channelCache with maximum capacity of n
+// channels.
+func newChannelCache(n int) *channelCache {
+	return &channelCache{
+		n:        n,
+		channels: make(map[uint64]ChannelEdge),
+	}
+}
+
+// get returns the channel from the cache, if it exists.
+func (c *channelCache) get(chanid uint64) (ChannelEdge, bool) {
+	channel, ok := c.channels[chanid]
+	return channel, ok
+}
+
+// insert adds the entry to the channel cache. If an entry for chanid already
+// exists, it will be replaced with the new entry. If the entry doesn't exist,
+// it will be inserted to the cache, performing a random eviction if the cache
+// is at capacity.
+func (c *channelCache) insert(chanid uint64, channel ChannelEdge) {
+	// If entry exists, replace it.
+	if _, ok := c.channels[chanid]; ok {
+		c.channels[chanid] = channel
+		return
+	}
+
+	// Otherwise, evict an entry at random and insert.
+	if len(c.channels) == c.n {
+		for id := range c.channels {
+			delete(c.channels, id)
+			break
+		}
+	}
+	c.channels[chanid] = channel
+}
+
+// remove deletes an edge for chanid from the cache, if it exists.
+func (c *channelCache) remove(chanid uint64) {
+	delete(c.channels, chanid)
+}
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -114,13 +114,18 @@ type DB struct {
 
 // Open opens an existing channeldb. Any necessary schemas migrations due to
 // updates will take place as necessary.
-func Open(dbPath string) (*DB, error) {
+func Open(dbPath string, modifiers ...OptionModifier) (*DB, error) {
 	path := filepath.Join(dbPath, dbName)
 
 	if !fileExists(path) {
 		if err := createChannelDB(dbPath); err != nil {
 			return nil, err
 		}
+	}
+
+	opts := DefaultOptions()
+	for _, modifier := range modifiers {
+		modifier(&opts)
 	}
 
 	bdb, err := bbolt.Open(path, dbFilePermission, nil)
@@ -132,7 +137,9 @@ func Open(dbPath string) (*DB, error) {
 		DB:     bdb,
 		dbPath: dbPath,
 	}
-	chanDB.graph = newChannelGraph(chanDB)
+	chanDB.graph = newChannelGraph(
+		chanDB, opts.RejectCacheSize, opts.ChannelCacheSize,
+	)
 
 	// Synchronize the version of database and apply migrations if needed.
 	if err := chanDB.syncVersions(dbVersions); err != nil {
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -992,7 +992,7 @@ func (d *DB) RestoreChannelShells(channelShells ...*ChannelShell) error {
 				chanEdge.ChannelFlags |= lnwire.ChanUpdateDirection
 			}
 
-			err = updateEdgePolicy(tx, &chanEdge)
+			_, err = updateEdgePolicy(tx, &chanEdge)
 			if err != nil {
 				return err
 			}
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -899,7 +899,12 @@ type ChannelShell struct {
 func (d *DB) RestoreChannelShells(channelShells ...*ChannelShell) error {
 	chanGraph := d.ChannelGraph()
 
-	return d.Update(func(tx *bbolt.Tx) error {
+	// TODO(conner): find way to do this w/o accessing internal members?
+	chanGraph.cacheMu.Lock()
+	defer chanGraph.cacheMu.Unlock()
+
+	var chansRestored []uint64
+	err := d.Update(func(tx *bbolt.Tx) error {
 		for _, channelShell := range channelShells {
 			channel := channelShell.Chan
 
@@ -984,10 +989,22 @@ func (d *DB) RestoreChannelShells(channelShells ...*ChannelShell) error {
 			if err != nil {
 				return err
 			}
+
+			chansRestored = append(chansRestored, edgeInfo.ChannelID)
 		}
 
 		return nil
 	})
+	if err != nil {
+		return err
+	}
+
+	for _, chanid := range chansRestored {
+		chanGraph.rejectCache.remove(chanid)
+		chanGraph.chanCache.remove(chanid)
+	}
+
+	return nil
 }
 
 // AddrsForNode consults the graph and channel database for all addresses known
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -109,6 +109,7 @@ var (
 type DB struct {
 	*bbolt.DB
 	dbPath string
+	graph  *ChannelGraph
 }
 
 // Open opens an existing channeldb. Any necessary schemas migrations due to
@@ -131,6 +132,7 @@ func Open(dbPath string) (*DB, error) {
 		DB:     bdb,
 		dbPath: dbPath,
 	}
+	chanDB.graph = newChannelGraph(chanDB)
 
 	// Synchronize the version of database and apply migrations if needed.
 	if err := chanDB.syncVersions(dbVersions); err != nil {
@@ -895,7 +897,7 @@ type ChannelShell struct {
 // well. This method is idempotent, so repeated calls with the same set of
 // channel shells won't modify the database after the initial call.
 func (d *DB) RestoreChannelShells(channelShells ...*ChannelShell) error {
-	chanGraph := ChannelGraph{d}
+	chanGraph := d.ChannelGraph()
 
 	return d.Update(func(tx *bbolt.Tx) error {
 		for _, channelShell := range channelShells {
@@ -1107,7 +1109,7 @@ func (d *DB) syncVersions(versions []version) error {
 
 // ChannelGraph returns a new instance of the directed channel graph.
 func (d *DB) ChannelGraph() *ChannelGraph {
-	return &ChannelGraph{d}
+	return d.graph
 }
 
 func getLatestDBVersion(versions []version) uint32 {
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -7,7 +7,6 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -103,10 +102,6 @@ var (
 	// integer keys iterating in order.
 	byteOrder = binary.BigEndian
 )
-
-var bufPool = &sync.Pool{
-	New: func() interface{} { return new(bytes.Buffer) },
-}
 
 // DB is the primary datastore for the lnd daemon. The database stores
 // information related to nodes, routing data, open/closed channels, fee
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"math"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -157,15 +158,18 @@ const (
 type ChannelGraph struct {
 	db *DB
 
-	// TODO(roasbeef): store and update bloom filter to reduce disk access
-	// due to current gossip model
-	//  * LRU cache for edges?
+	cacheMu     sync.RWMutex
+	rejectCache *rejectCache
+	chanCache   *channelCache
 }
 
-// newChannelGraph allocates a new ChannelGraph backed by a DB instance.
+// newChannelGraph allocates a new ChannelGraph backed by a DB instance. The
+// returned instance has its own unique reject cache and channel cache.
 func newChannelGraph(db *DB) *ChannelGraph {
 	return &ChannelGraph{
-		db: db,
+		db:          db,
+		rejectCache: newRejectCache(50000),
+		chanCache:   newChannelCache(20000),
 	}
 }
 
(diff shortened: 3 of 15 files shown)


Roasbeef:
Now that the dependent PR has been merged, this can be rebased!

channeldb/channel_cache.go
```diff
@@ -0,0 +1,50 @@
+package channeldb
+
+// channelCache is an in-memory cache used to improve the performance of
+// ChanUpdatesInHorizon. It caches the chan info and edge policies for a
+// particular channel.
+type channelCache struct {
+	n        int
+	channels map[uint64]ChannelEdge
```
Roasbeef: Why not a pointer value here? Less GC pressure with a large cache?

channeldb/graph.go
```diff
@@ -1703,9 +1799,20 @@ func delChannelByEdge(edges, edgeIndex, chanIndex, zombieIndex,
...
+	defer c.cacheMu.Unlock()
+
+	err := c.db.Batch(func(tx *bbolt.Tx) error {
 		return updateEdgePolicy(tx, edge)
 	})
+	if err != nil {
+		return err
+	}
+
+	c.rejectCache.remove(edge.ChannelID)
```
Roasbeef: Alternatively, we can insert the new copy of the edge policy into the cache. I see no issue in delaying this to a distinct change though once we see how this fares in the wild once most of the network has updated.

channeldb/db.go
```diff
@@ -114,7 +114,7 @@ type DB struct {
 
 // Open opens an existing channeldb. Any necessary schemas migrations due to
 // updates will take place as necessary.
-func Open(dbPath string) (*DB, error) {
+func Open(dbPath string, modifiers ...OptionModifier) (*DB, error) {
```
Roasbeef: Nice! We can also eventually maybe expose some `boltdb` options here as well.

config.go
```diff
@@ -975,9 +982,12 @@ func loadConfig() (*config, error) {
 			"minbackoff")
 	}
 
-	// Assert that all worker pools will have a positive number of
-	// workers, otherwise the pools will rendered useless.
-	if err := cfg.Workers.Validate(); err != nil {
+	// Validate the subconfigs for workers and caches.
+	err = lncfg.Validate(
+		cfg.Workers,
```
Roasbeef: 👍

channeldb/db.go
```diff
@@ -104,10 +103,6 @@ var (
 	byteOrder = binary.BigEndian
 )
 
-var bufPool = &sync.Pool{
```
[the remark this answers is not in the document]
cfromknecht: lol indeed :)

channeldb/channel_cache.go
cfromknecht: yes, and because a `ChannelEdge` itself consists of only 3 pointers and so are relatively small

channeldb/channel_cache.go
```diff
@@ -0,0 +1,50 @@
+package channeldb
+
+// channelCache is an in-memory cache used to improve the performance of
+// ChanUpdatesInHorizon. It caches the chan info and edge policies for a
+// particular channel.
+type channelCache struct {
```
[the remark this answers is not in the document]
cfromknecht: generally would agree, though the actual amount of duplicate code is about 15 lines. would say the performance benefits of using explicit types outweigh the downsides in our hottest code paths

channeldb/graph.go
```diff
@@ -660,16 +685,24 @@ func (c *ChannelGraph) HasChannelEdge(chanID uint64,
...
+			upd1Time = e1.LastUpdate
 		}
 		if e2 != nil {
-			node2UpdateTime = e2.LastUpdate
+			upd2Time = e2.LastUpdate
 		}
 
 		return nil
+	}); err != nil {
+		return time.Time{}, time.Time{}, false, false, err
```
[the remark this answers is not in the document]
cfromknecht: indeed, to match the old behavior it's probably worth returning `exists` and `isZombie`

config.go
cfromknecht: hoping that we can eventually move more of our subconfigs into `lncfg` and make use of this pattern, would make `config.go` quite a bit slimmer :)

channeldb/channel_cache.go
```diff
@@ -0,0 +1,50 @@
...
+	n        int
+	channels map[uint64]ChannelEdge
+}
+
+// newChannelCache creates a new channelCache with maximum capacity of n
+// channels.
+func newChannelCache(n int) *channelCache {
+	return &channelCache{
+		n:        n,
+		channels: make(map[uint64]ChannelEdge),
```
Roasbeef: Can also be pre-sized.

channeldb/graph.go
```diff
@@ -693,7 +693,7 @@ func (c *ChannelGraph) HasChannelEdge(
 
 		return nil
 	}); err != nil {
-		return time.Time{}, time.Time{}, false, false, err
+		return time.Time{}, time.Time{}, exists, isZombie, err
```
Roasbeef: If it's returning an error anyway, why's this matter?

channeldb/channel_cache.go
Roasbeef: Or maybe it's better not to, since in the near future, ideally most node won't really request a backlog.

channeldb/graph.go
```diff
@@ -491,9 +502,20 @@ func (c *ChannelGraph) deleteLightningNode(nodes *bbolt.Bucket,
...
+
+	err := c.db.Update(func(tx *bbolt.Tx) error {
 		return c.addChannelEdge(tx, edge)
 	})
+	if err != nil {
+		return err
+	}
+
+	c.rejectCache.remove(edge.ChannelID)
+	c.chanCache.remove(edge.ChannelID)
```
Roasbeef: Since we update the cache now when we add/update edge policies, we may need to also add the edge info to the cache here. Otherwise, it seems possible that we add two edges for a `ChannelEdge`, but then never update the edge info as well, causing us to send a blank edge info.

Roasbeef: Scratch that: items are only added for the first time within `ChanUpdatesInHorizon`, so when we add the edge policy, only if it's already present will we update the edge policies in the cache.

channeldb/graph.go
```diff
@@ -498,9 +502,20 @@ func (c *ChannelGraph) deleteLightningNode(nodes *bbolt.Bucket,
 // the channel supports. The chanPoint and chanID are used to uniquely identify
 // the edge globally within the database.
 func (c *ChannelGraph) AddChannelEdge(edge *ChannelEdgeInfo) error {
-	return c.db.Update(func(tx *bbolt.Tx) error {
+	c.cacheMu.Lock()
+	defer c.cacheMu.Unlock()
```
halseth: could we just move the mutex inside the cache access methods? Seems like it wouldn't be a problem if the cache and DB was slightly out of sync for a split second.

channeldb/graph.go
```diff
@@ -605,17 +620,27 @@ func (c *ChannelGraph) addChannelEdge(tx *bbolt.Tx, edge *ChannelEdgeInfo) error
 // was updated for both directed edges are returned along with the boolean. If
 // it is not found, then the zombie index is checked and its result is returned
 // as the second boolean.
-func (c *ChannelGraph) HasChannelEdge(chanID uint64,
-) (time.Time, time.Time, bool, bool, error) {
+func (c *ChannelGraph) HasChannelEdge(
+	chanID uint64) (time.Time, time.Time, bool, bool, error) {
+
+	c.cacheMu.Lock()
```
halseth: `RLock` for the exist-check?

lncfg/caches.go
```diff
@@ -0,0 +1,42 @@
...
+	// reject cache, which is used for efficiently rejecting gossip updates.
+	// Memory usage is roughly 25b per entry.
+	RejectCacheSize int `long:"reject-cache-size" description:"Maximum number of entries contained in the reject cache, which is used to speed up filtering of new channel announcements and channel updates from peers. Each entry requires 25 bytes."`
+
+	// ChannelCacheSize is the maximum number of entries stored in lnd's
+	// channel cache, which is used reduce memory allocations in reply to
+	// peers querying for gossip traffic. Memory usage is roughly 2Kb per
+	// entry.
+	ChannelCacheSize int `long:"channel-cache-size" description:"Maximum number of entries contained in the channel cache, which is used to reduce memory allocations from gossip queries from peers. Each entry requires roughly 2Kb."`
+}
```
halseth: Add interface satisfy check.

config.go
```diff
@@ -254,6 +255,8 @@ type config struct {
 	Routing *routing.Conf `group:"routing" namespace:"routing"`
 
 	Workers *lncfg.Workers `group:"workers" namespace:"workers"`
+
+	Caches *lncfg.Caches `group:"caches" namespace:"caches"`
```
halseth: 👍

channeldb/graph.go
```diff
@@ -1798,34 +1798,52 @@ func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
...
+		c.chanCache.insert(edge.ChannelID, channel)
+	}
 
 	return nil
 }
 
 // updateEdgePolicy attempts to update an edge's policy within the relevant
 // buckets using an existing database transaction.
-func updateEdgePolicy(tx *bbolt.Tx, edge *ChannelEdgePolicy) error {
+func updateEdgePolicy(tx *bbolt.Tx, edge *ChannelEdgePolicy) (bool, error) {
```
halseth: document bool?

channeldb/graph.go
```diff
@@ -1798,34 +1798,52 @@ func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
...
+	if entry, ok := c.rejectCache.get(edge.ChannelID); ok {
+		if isUpdate1 {
+			entry.upd1Time = edge.LastUpdate.Unix()
+		} else {
+			entry.upd2Time = edge.LastUpdate.Unix()
+		}
+		c.rejectCache.insert(edge.ChannelID, entry)
+	}
+
+	if channel, ok := c.chanCache.get(edge.ChannelID); ok {
```
halseth: Can we insert without checking existence?

channeldb/graph.go
```diff
@@ -1798,34 +1798,52 @@ func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
...
+		isUpdate1, err = updateEdgePolicy(tx, edge)
+		return err
 	})
 	if err != nil {
 		return err
 	}
 
-	c.rejectCache.remove(edge.ChannelID)
-	c.chanCache.remove(edge.ChannelID)
+	if entry, ok := c.rejectCache.get(edge.ChannelID); ok {
```
halseth: Better to just remove IMO, we have no guarantees that the updated policy doesn't change the value of the `rejectFlags`.

channeldb/graph.go
```diff
@@ -1696,26 +1795,55 @@ func delChannelByEdge(edges, edgeIndex, chanIndex, zombieIndex,
 // determined by the lexicographical ordering of the identity public keys of
 // the nodes on either side of the channel.
 func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
-	return c.db.Batch(func(tx *bbolt.Tx) error {
-		return updateEdgePolicy(tx, edge)
+	c.cacheMu.Lock()
```
halseth: Also need to refresh cache after calls to `MarkEdgeLive`? Any other methods that can alter the any of the cached information?

channeldb/graph.go
cfromknecht: I'd be okay with doing so. We'd have to check twice, once w/ shared lock and then again after acquiring the exclusive, but likely better performance

channeldb/graph.go
cfromknecht: the reject flags are only changed wrt to channel existence/zombie pruning. updating the edge policy shouldn't modify `rejectFlags`

channeldb/graph.go
cfromknecht: not without also loading the channel info and the other edge policy in the transaction above. there's no way to know this edge will actually be read from the cache, so lazily populating the cache avoids unnecessary allocation/deserialization/db seeks

channeldb/graph.go
cfromknecht: good cache! i did a scan of all the methods when first implementing, wouldn't hurt to do another to be safe

channeldb/graph.go
cfromknecht: mainly did it to keep existing behavior, but it shouldn't

channeldb/graph.go
cfromknecht: yep exactly!

channeldb/graph.go
cfromknecht: `MarkEdgeZombie` also needed to be modified, fixed

channeldb/channel_cache.go
cfromknecht: sgtm

cfromknecht:
@Roasbeef @wpaulino @halseth latest version is up and comments addressed, ptal

PR #2855, opened 2019-03-30
contractcourt: detect local force closes based on commitment outputs

Roasbeef:
In this commit, we modify the way we detect local force closes. Before
this commit, we would directly check the broadcast commitment's txid
against what we know to be our best local commitment. In the case of DLP
recovery from an SCB, it's possible that the user force closed, _then_
attempted to recover their channels. As a result, we need to check the
outputs directly in order to also handle this rare, but
possible recovery scenario.

The new detection method uses the outputs to detect if it's a local
commitment or not. Based on the state number, we'll re-derive the
expected scripts, and check to see if they're on the commitment. If not,
then we know it's a remote force close. A new test has been added to
exercise this new behavior, ensuring we catch local closes where we have
and don't have a direct output.

### DIFF
--- a/contractcourt/chain_watcher.go
+++ b/contractcourt/chain_watcher.go
@@ -1,6 +1,7 @@
 package contractcourt
 
 import (
+	"bytes"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -16,6 +17,7 @@ import (
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/shachain"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


contractcourt/chain_watcher_test.go
```diff
@@ -409,3 +421,101 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
+			return true
+
+		case <-time.After(time.Second * 5):
+			t.Errorf("didn't get local for close for state #%v",
+				numUpdates)
+			return false
+		}
+	}
+
+	err := quick.Check(localForceCloseScenario, &quick.Config{
```
cfromknecht: perhaps we can enumerate some table driven tests as opposed to using quick tests for this. this tests alone adds over three minutes to the race test execution time

```
=== RUN   TestChainWatcherLocalForceCloseDetect
--- PASS: TestChainWatcherLocalForceCloseDetect (215.03s)
```

contractcourt/chain_watcher.go
```diff
@@ -271,6 +273,74 @@ func (c *chainWatcher) SubscribeChannelEvents() *ChainEventSubscription {
...
+// funding transaction using our commitment transaction (a local force close).
+// In order to do this in a state agnostic manner, we'll make our decisions
+// based off of only the set of outputs included.
+func isOurCommitment(localChanCfg, remoteChanCfg channeldb.ChannelConfig,
+	commitSpend *chainntnfs.SpendDetail, broadcastStateNum uint64,
+	revocationProducer shachain.Producer) (bool, error) {
+
+	// First, we'll re-derive our commitment point for this state since
+	// this is what we use to randomize each of the keys for this state.
+	commitSecret, err := revocationProducer.AtIndex(broadcastStateNum)
```
halseth: Most of the following key derivation is already done in `lnwallet`, can it be re-used?

contractcourt/chain_watcher_test.go
halseth: y u called `quick`

cfromknecht:
needs rebase

contractcourt/chain_watcher_test.go
Roasbeef: Modified to use table driven tests with sub-tests. Each of the sub-tests run in parallel.

contractcourt/chain_watcher.go
Roasbeef: The code there is different in that's it's more generic. IMO duplicating a small portion of that code in order to focus on which keys are specifically needed makes this section of the code more readable.

Roasbeef:
Rebased with comments addressed. Planning on making a follow up PR to the DLP chain watcher tests to improve the execution time of the tests as I did these ones.

Roasbeef:
PTAL @cfromknecht @halseth

contractcourt/chain_watcher_test.go
```diff
@@ -409,3 +422,134 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
+				numUpdates)
+			return false
+		}
+	}
+
+	// For our test cases, we'll ensure that we test having a remote output
+	// present and absent with non or some number of updates in the channel.
+	testCases := []struct {
+		numUpdates       uint8
+		remoteOutputOnly bool
```
cfromknecht: do we also want to test the case where there is only a `localOutput`? currently we test local+remote and remote-only

Roasbeef: added!

contractcourt/chain_watcher.go
```diff
@@ -320,34 +390,38 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
 
-			if isOurCommitment {
-				if err := c.dispatchLocalForceClose(
-					commitSpend, *localCommit,
-				); err != nil {
-					log.Errorf("unable to handle local"+
-						"close for chan_point=%v: %v",
-						c.cfg.chanState.FundingOutpoint, err)
-				}
-				return
```
halseth: should be returning here?

contractcourt/chain_watcher_test.go
```diff
@@ -409,3 +422,148 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
 	}
 }
+
+// TestChainWatcherLocalForceCloseDetect tests we're able to always detect our
+// commitment output based on only the outputs present on the transaction.
+func TestChainWatcherLocalForceCloseDetect(t *testing.T) {
+	t.Parallel()
+
+	// localForceCloseScenario is the primary test we'll use to create a
+	// quickcheck test. We'll assert that for any number of state updates,
```
halseth: nit: no longer `quickcheck`

contractcourt/chain_watcher_test.go
```diff
@@ -409,3 +422,148 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
+
+		// We should get a local force close event from Alice as she
+		// should be able to detect the close based on the commitment
+		// outputs.
+		select {
+		case <-chanEvents.LocalUnilateralClosure:
+			return true
+
+		case <-time.After(time.Second * 5):
+			t.Errorf("didn't get local for close for state #%v",
```
halseth: Why not just use `t.Fatalf` everywhere?

contractcourt/chain_watcher_test.go
```diff
@@ -409,3 +422,148 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
+	}
+	for _, testCase := range testCases {
+		testName := fmt.Sprintf(
+			"num_updates=%v,remote_output=%v,local_output=%v",
+			testCase.numUpdates, testCase.remoteOutputOnly,
+			testCase.localOutputOnly,
+		)
+
+		t.Run(testName, func(t *testing.T) {
+			t.Parallel()
```
halseth: same comment here about not capturing loop variable.

contractcourt/chain_watcher.go
Roasbeef: Nice catch! Fixed.

contractcourt/chain_watcher_test.go
Roasbeef: This lets all the tests execute without stoping the entire group of tests. I don't think it makes a big difference either way.

contractcourt/chain_watcher_test.go
Roasbeef: Fixed

Roasbeef:
PTAL @halseth

Roasbeef:
Rebased after the other related PR landed before this one.

Roasbeef:
PTAL @halseth

PR #2863, opened 2019-03-31
[WIP] Fix incorrect http status code

### LINKED ISSUE #2332: Incorrect HTTP status code when macaroon is not provided over REST API
### Background

The HTTP error code returned by lnd using REST API is not correct. It returns 500 Internal server error instead of 4xx. This is confusing when debugging.

### Your environment

* version of `lnd` 0.5.1
* which operating system (`uname -a` on *Nix) `Linux <redacted> 3.16.0-4-amd64 #1 SMP Debian 3.16.51-3 (2017-12-13) x86_64 GNU/Linux`

### Steps to reproduce

* Configure `restlisten=127.0.0.1:9737`
* Run `wget --content-on-error --no-check-certificate -O - https://127.0.0.1:9737/v1/getinfo`

### Expected behaviour

Error 403 Forbidden or other reasonable 4xx error is returned.

### Actual behaviour

500 Internal server error is returned

Ironically, since error 500 means there's something broken with the server, it's strictly true (server sending bad status code), but it doesn't help at all.

philippgille:
Closes #2332.

Currently WIP because there are some remaining questions:

1. Other than my comments on the issue, I'm not so sure anymore if it's the correct way to overwrite the grpc-gateway error response to include the `WWW-Authenticate` header. It should be the default behavior of grpc-gateway, or maybe there's a good reason not to include the header. A short web search didn't turn up any good results. I asked in the gRPC gitter ([link to message](https://gitter.im/grpc/grpc?at=5ca0e2c1016a930a453cdc6d)), but didn't get a response so far.
2. I didn't find a place to add / extend an integration test for the gateway response status code. On the one hand, it might be seen as redundant because it would just check if grpc-gateway works correctly (properly converts the error), meaning the test that I _did_ extend is enough because it makes sure the proper gRPC error is returned, but on the other hand who knows what layers of code get added in the future, maybe wrapping the returned error and thus making it incompatible for the gateway conversion again.  
  I _did_ make sure _manually_ that the gateway now returns the correct HTTP status codes though.

#### Pull Request Checklist

- [x] All changes are Go version 1.11 compliant
- [x]  The code being submitted is commented according to the
  [Code Documentation and Commenting](#CodeDocumentation) section
- [x]  For new code: Code is accompanied by tests which exercise both
- [ ]  For bug fixes: Code is accompanied by new tests which trigger
- [x]  Any new logging statements use an appropriate subsystem and
  logging level
- [x]  Code has been formatted with `go fmt`
- [x]  For code and documentation: lines are wrapped at 80 characters
  per default)
- [ ]  Running `make check` does not fail any tests
- [x]  Running `go vet` does not report any issues
- [x]  Running `make lint` does not report any **new** issues that did not
  already exist
- [x] Commits have a logical structure ([see section 4.5, of the Code Contribution Guidelines])

Regarding this list:

- Regarding the bug fix test code: See above open question (whether to implement an integration test or not)
- Regarding `make check`: It resultet in a single error, which I don't think is related to the code changes in this PR:  
  ```
  --- FAIL: TestLightningWallet (5.47s)
    interface_test.go:2803: unable to establish connection to bitcoind: -28: Loading banlist...
  ```

### DIFF
--- a/macaroons/service.go
+++ b/macaroons/service.go
@@ -8,7 +8,9 @@ import (
 
 	"github.com/coreos/bbolt"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 
 	"gopkg.in/macaroon-bakery.v2/bakery"
 	"gopkg.in/macaroon-bakery.v2/bakery/checkers"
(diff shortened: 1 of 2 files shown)


philippgille:
Maybe useful in case an intgration test should be implemented: https://github.com/grpc-ecosystem/grpc-gateway/issues/699

philippgille:
Some news regarding the missing `WWW-Authenticate` header: I opened a GitHub issue in the grpc-gateway project and it looks like they're willing to accept a PR that fixes the issue: https://github.com/grpc-ecosystem/grpc-gateway/issues/917

I could work on that as soon as I have time. But is there a chance to have that dependency updated to the most recent version (after the fix) at all? Or does lnd require a specific older version and is only willing to update with extensive additional testing?

halseth:
@philippgille We don't need an old version, we uptime from time to time :)

Roasbeef:
Closing due to inactivity. 

(I think this has also mostly been resolved with all the chagnes we've made to the way the REST proxy works over the last year or two)

PR #2866, opened 2019-04-02
pool/worker: increase worker timeout to 90s

cfromknecht:
This commit increases the default worker timeout currently backing the
read and write pools. This allows the read and write pools to sustain
regular bursty traffic such as ping/pong without releasing their buffers
back to the underlying gc queue. In the future, jitter can be added to
our ping and/or gossip messages to reduce the concurrent usage of read
and write pools, which will make this change even more effective.

### DIFF
--- a/pool/worker.go
+++ b/pool/worker.go
@@ -13,7 +13,7 @@ var ErrWorkerPoolExiting = errors.New("worker pool exiting")
 // DefaultWorkerTimeout is the default duration after which a worker goroutine
 // will exit to free up resources after having received no newly submitted
 // tasks.
-const DefaultWorkerTimeout = 5 * time.Second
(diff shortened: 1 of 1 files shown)


halseth:
I would assume releasing a buffer every 5 seconds wouldn't be a big deal for the GC, is that not the case?

cfromknecht:
@halseth it's not the end of the world, certainly. however we have a lot messages that have very bursty sending/receiving patterns, which would cause all of these buffers be allocated/gc'd every 60-90s. Allocating/gc'ing numpeers*65kb every minute isn't free, and this change is meant to ensure that the waste incurred for regular patterns is minimized, while still allowing bursts from outliers to converge back to the transient working set

PR #2868, opened 2019-04-02
chainntnfs: Always commit bitcoind spendhints

cfromknecht:
This PR fixes a bug that would cause the
notifier not to commit spend hints for items
that are not found. This is done by calling
UpdateSpendDetails with a nil detail, permitting
the notifier to begin updating the spend hints
with new blocks that arrive at tip. The change
is designed to mimic the behavior for historical
confirmation dispatch.

The symptom of this bug is needing to do many
long rescans on startup, even if new blocks
arrive after the rescan had completed. With
this change, nodes will have to do the scans
once more before their hints will be properly
updated. Restarts from then on should not
have this behavior.

### DIFF
--- a/chainntnfs/bitcoindnotify/bitcoind.go
+++ b/chainntnfs/bitcoindnotify/bitcoind.go
@@ -25,13 +25,6 @@ const (
 	notifierType = "bitcoind"
 )
 
-var (
-	// ErrTransactionNotFound is an error returned when we attempt to find a
-	// transaction by manually scanning the chain within a specific range
-	// but it is not found.
-	ErrTransactionNotFound = errors.New("transaction not found within range")
-)
-
 // chainUpdate encapsulates an update to the current main chain. This struct is
 // used as an element within an unbounded queue in order to avoid blocking the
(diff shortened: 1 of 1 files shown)


chainntnfs/bitcoindnotify/bitcoind.go
```diff
@@ -828,7 +850,8 @@ func (b *BitcoindNotifier) RegisterSpendNtfn(outpoint *wire.OutPoint,
 // script. If one is found, it's spending details are sent to the TxNotifier,
 // which will then dispatch the notification to all of its clients.
 func (b *BitcoindNotifier) dispatchSpendDetailsManually(
-	historicalDispatchDetails *chainntnfs.HistoricalSpendDispatch) error {
+	historicalDispatchDetails *chainntnfs.HistoricalSpendDispatch) (
```
wpaulino: Should rename the method and update the `godoc` as it's no longer dispatching the details.

cfromknecht: fixed

PR #2885, opened 2019-04-04
server: stagger initial reconnects

cfromknecht:
This commit adds optional jitter to our initial reconnection to our
persistent peers. Currently we will attempt reconnections to all peers
simultaneously, which results in large amount of contention as the
number of channels a node has grows.

We resolve this by adding a randomized delay between 0 and 30 seconds
for all persistent peers. This spreads out the load and contention to
resources such as the database, read/write pools, and memory
allocations. On my node, this allows to start up with about 80% of the
memory burst compared to the all-at-once approach.

This also has a second-order effect in better distributing messages sent
at constant intervals, such as pings. This reduces the concurrent jobs
submitted to the read and write pools at any given time, resulting in
better reuse of read/write buffers and fewer bursty allocation and
garbage collection cycles.

The behavior can be enabled via the `--stagger-initial-reconnect` flag.

### DIFF
--- a/config.go
+++ b/config.go
@@ -252,6 +252,8 @@ type config struct {
 
 	RejectPush bool `long:"rejectpush" description:"If true, lnd will not accept channel opening requests with non-zero push amounts. This should prevent accidental pushes to merchant nodes."`
 
+	StaggerInitialReconnect bool `long:"stagger-initial-reconnect" description:"If true, will apply a randomized staggering between 0s and 30s when reconnecting to persistent peers on startup. The first 10 reconnections will be attempted instantly, regardless of the flag's value"`
+
 	net tor.Net
 
 	Routing *routing.Conf `group:"routing" namespace:"routing"`
--- a/server.go
+++ b/server.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"image/color"
 	"math/big"
+	prand "math/rand"
 	"net"
 	"path/filepath"
 	"regexp"
@@ -60,6 +61,18 @@ const (
 	// durations exceeding this value will be eligible to have their
 	// backoffs reduced.
 	defaultStableConnDuration = 10 * time.Minute
+
+	// numInstantInitReconnect specifies how many persistent peers we should
+	// always attempt outbound connections to immediately. After this value
+	// is surpassed, the remaining peers will be randomly delayed using
+	// maxInitReconnectDelay.
+	numInstantInitReconnect = 10
+
+	// maxInitReconnectDelay specifies the maximum delay in seconds we will
+	// apply in attempting to reconnect to persistent peers on startup. The
+	// value used or a particular peer will be chosen between 0s and this
+	// value.
+	maxInitReconnectDelay = 30
 )
 
 var (
@@ -1931,6 +1944,7 @@ func (s *server) establishPersistentConnections() error {
 
 	// Iterate through the combined list of addresses from prior links and
 	// node announcements and attempt to reconnect to each node.
+	var numOutboundConns int
 	for pubStr, nodeAddr := range nodeAddrsMap {
 		// Add this peer to the set of peers we should maintain a
 		// persistent connection with.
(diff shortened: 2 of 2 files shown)


config.go
```diff
@@ -252,6 +252,8 @@ type config struct {
 
 	RejectPush bool `long:"rejectpush" description:"If true, lnd will not accept channel opening requests with non-zero push amounts. This should prevent accidental pushes to merchant nodes."`
 
+	StaggerInitialReconnect bool `long:"stagger-initial-reconnect" description:"If true, will randomly stagger reconnections to persistent peers on startup between 0s and 30s. Nodes with a large number of channels should use this to reduce large memory bursts when restarting."`
```
wpaulino: Any reason for including the flag and not just making this the default behavior?

cfromknecht: I think this should be default behavior, I have a commit locally that I will push up that makes it so. 

We probably don’t want this behavior on mobile, as it may take 30s to reconnect. Mobile devs can set this to false since they’ll probably have far fewer channels as well. We also want to disable this for integration tests, so the more recent version will also set this to false.

cfromknecht: latest version makes this the default behavior and disables it for itests

config.go
```diff
@@ -252,6 +252,8 @@ type config struct {
 
 	RejectPush bool `long:"rejectpush" description:"If true, lnd will not accept channel opening requests with non-zero push amounts. This should prevent accidental pushes to merchant nodes."`
 
+	NoStaggerInitialReconnect bool `long:"no-stagger-initial-reconnect" description:"If false, will randomly stagger reconnections to persistent peers on startup between 0s and 30s. Nodes with a large number of channels should set this value to true to reduce large memory bursts when restarting."`
```
wpaulino: Would prefer noting what happens when the flag is true instead. Also the second sentence doesn't seem to be relevant anymore, should perhaps mention this should be set when running on mobile.

cfromknecht: sgtm, fixed

server.go
```diff
@@ -60,6 +61,18 @@ const (
 	// durations exceeding this value will be eligible to have their
 	// backoffs reduced.
 	defaultStableConnDuration = 10 * time.Minute
+
+	// numInsantInitReconnect specifies how many persistent peers we should
+	// always attempt outbound connections to immediately. After this value
+	// is surpassed, the remaining peers will be randomly delayed using
+	// maxInitReconnectDelay.
+	numInsantInitReconnect = 10
```
joostjager: typo

server.go
```diff
@@ -60,6 +61,18 @@ const (
 	// durations exceeding this value will be eligible to have their
 	// backoffs reduced.
 	defaultStableConnDuration = 10 * time.Minute
+
+	// numInstantInitReconnect specifies how many persistent peers we should
+	// always attempt outbound connections to immediately. After this value
+	// is surpassed, the remaining peers will be randomly delayed using
+	// maxInitReconnectDelay.
+	numIntsantInitReconnect = 10
```
joostjager: Typo

config.go
```diff
@@ -252,6 +252,8 @@ type config struct {
 
 	RejectPush bool `long:"rejectpush" description:"If true, lnd will not accept channel opening requests with non-zero push amounts. This should prevent accidental pushes to merchant nodes."`
 
+	StaggerInitialReconnect bool `long:"stagger-initial-reconnect" description:"If true, will apply a randomized staggering between 0s and 30s when reconnecting to persistent peers on startup. The first 10 reconnections will be attempted instantly, regardless of the flag's value"`
```
wpaulino: Why is it no longer enabled by default? With `numInstantInitReconnect` you could even say that this flag is not needed anymore.

cfromknecht:
AFAIK it’s been tested on several nodes with 50+ channels, hoping to get feedback from even larger nodes after the next tag

config.go
cfromknecht: decided to make the behavior opt-in given that it's close to the release, decision was to make it default in a later release if the majority of users report it useful.

PR #2879, opened 2019-04-03
htlcswitch: revert forwarding policy block delta requirements

joostjager:
After the merge of #2839, a minimum cltv delta of 14 blocks is enforced for outgoing htlcs. This is a problem when forwarding to the final node which has an invoice with a cltv < 14. This happens in the wild.

In this case, `lnd` doesn't forward and returns an `ExpiryTooSoon` failure to the sender.

This PR reverts the parameters to their original values. 

It does not fully revert #2839, but instead introduces a clearer way to configure the relevant parameters and sets them to their original values. It replaces the original `broadcastRedeemMultiplier` and `expiryGraceDelta`.

In follow up #2887 the parameters are tuned further to improve the go-to-chain behaviour and forwarding policies.

PR #2887, opened 2019-04-04
config: lower outgoing htlc broadcast delta

joostjager:
Builds on top of #2879.

We create outgoing htlcs and accept incoming htlcs within the go-to-chain window, risking triggering a force close. This PR updates the block deltas to make sure that we don't trigger a force close ourselves.

The main change is lowering the broadcast delta for outgoing htlcs. It can be lower than the delta for incoming htlcs, as we don't risk not being able to settle with the preimage if the tx(es) don't confirm in time.

The spec even suggests a delta of -1 for outgoing htlcs (https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#cltv_expiry_delta-selection). Meaning that we only go to chain in the block after the block where the htlc expires.

This PR doesn't lower it to -1, but uses 0 instead to prevent confusion around the negative number. This means that we go to chain when an outgoing htlc expires. 

`OutgoingCltvRejectDelta` is a few blocks more than `OutgoingBroadcastDelta`. This is to prevent us forwarding an htlc that would bring us inside the broadcast window. We would probably receive a settle or fail quickly, but if a new block arrives in the mean time, we'd force close the channel.

`FinalCltvRejectDelta` could have been equal to `IncomingBroadcastDelta`, because as an exit hop we can immediately decide what to do with the incoming htlc.

This PR does increase `FinalCltvRejectDelta` a bit because for hodl invoices we do not decide immediately. Eventually, for hodl invoices, there will be automatic cancellation to prevent channel closure. This automatic cancellation should trigger some blocks before we go to chain. We can probably use the same value `FinalCltvRejectDelta` and the delta at which we cancel back hodl invoice htlcs.


joostjager:
Comments from @halseth 
- Add unit test to assert spec compliance of delta constant
- Reword comment on `IncomingBroadcastDelta` in `chain_arbitrator.go`

joostjager:
Reminder: Also need to check the outgoing expiry when we are the sender of the payment. In that case, `HtlcSatisfyPolicy` isn't called.

cfromknecht:
```
# github.com/lightningnetwork/lnd [github.com/lightningnetwork/lnd.test]
./lnd_test.go:9473:33: constant -4 overflows uint32
```

joostjager:
@cfromknecht ptal

Implemented policy check for local htlcs.

Didn't implement unit test to assert spec compliance. It doesn't look very useful to me to just check that a constant has a value. Instead I added extensive comments to the config constants which should future devs make think about what these values mean.

htlcswitch/switch_test.go
```diff
@@ -1382,8 +1397,9 @@ func TestSkipIneligibleLinksLocalForward(t *testing.T) {
 	chanID1, _, aliceChanID, _ := genIDs()
 
 	aliceChannelLink := newMockChannelLink(
-		s, chanID1, aliceChanID, alicePeer, false,
+		s, chanID1, aliceChanID, alicePeer, eligable,
```
cfromknecht: s/eligable/eligible

joostjager:
@cfromknecht ptal

config.go
```diff
@@ -68,11 +68,50 @@ const (
...
+	// are not in a hurry to force close, because there is nothing to claim
+	// for us. We do need to time the htlc out, because there may be an
+	// incoming htlc that will time out too (albeit later). Bolt #2 suggests
+	// a value of -1 here, but we allow one block less to prevent potential
+	// confusion around the negative value. It means we force close the
+	// channel at exactly the htlc expiry height.
+	defaultOutgoingBroadcastDelta = 0
+
+	// defaultOutgoingCltvRejectDelta defines the number of blocks before
+	// the expiry of an outgoing htlc at which we don't want to it offer to
```
halseth: s/it offer/offer it

htlcswitch/link.go
```diff
@@ -2195,6 +2165,36 @@ func (l *channelLink) HtlcSatifiesPolicy(payHash [32]byte,
 		return &lnwire.FailExpiryTooFar{}
 	}
 
+	// Next, using the amount of the incoming HTLC, we'll calculate the
```
halseth: commit message should explain the rationale behind this change.

config.go
```diff
@@ -68,11 +68,50 @@ const (
 	defaultTorV2PrivateKeyFilename = "v2_onion_private_key"
 	defaultTorV3PrivateKeyFilename = "v3_onion_private_key"
 
-	defaultIncomingBroadcastDelta = 20
-	defaultFinalCltvRejectDelta   = 2
-
-	defaultOutgoingBroadcastDelta  = 10
-	defaultOutgoingCltvRejectDelta = 0
+	// defaultIncomingBroadcastDelta defines the number of blocks before the
```
halseth: excellent documentation 👍

htlcswitch/link.go
joostjager: added

joostjager:
@halseth ptal

PR #2889, opened 2019-04-04
Remove equal sign on verifychanbackup cmd argument

grunch:
This is specifically on multi_file argument description


cfromknecht:
@grunch can you rebase on latest master? i think two of the three lines were already fixed by #2888

PR #2892, opened 2019-04-04
rpc: modify VerifyChanBackup to take either a Single or Multi

wpaulino:


lnrpc/rpc.proto
```diff
@@ -733,10 +733,8 @@ service Lightning {
     };
 
     /**
-    VerifyChanBackup allows a caller to verify the integrity of a channel
-    backup snapshot. This method will accept both a packed Single, and also a
-    Packed multi. Two bools are returned which indicate if the passed Single
-    (if present) is valid and also if the passed Multi (if present) is valid.
+    VerifyChanBackup allows a caller to verify the integrity of a channel backup
+    snapshot. This method will accept either a packed Single or a Packed multi.
```
cfromknecht: comment should mention that specifying both will return an error

rpcserver.go
```diff
@@ -4766,63 +4766,61 @@ func (r *rpcServer) ExportChannelBackup(ctx context.Context,
...
 			"backup must be specified")
-	}
 
-	// If a Single is specified then we'll only accept one of them to allow
-	// the caller to map the valid/invalid state for each individual
-	// Single.
-	if in.GetSingleChanBackups() != nil &&
-		len(in.GetSingleChanBackups().ChanBackups) != 1 {
+	// Either a Single or a Multi must be specified, but not both.
+	case in.GetSingleChanBackups() == nil && in.GetMultiChanBackup() == nil:
```
cfromknecht: identical to first case

rpcserver.go
```diff
@@ -4766,63 +4766,61 @@ func (r *rpcServer) ExportChannelBackup(ctx context.Context,
...
 		packedMultiBackup := in.GetMultiChanBackup().MultiChanBackup
 		packedMulti := chanbackup.PackedMulti(packedMultiBackup)
 
-		// We'll now attempt to unpack the Multi. If this fails, then
-		// we know it's invalid.
+		// We'll now attempt to unpack the Multi. If this fails, then we
+		// know it's invalid.
 		_, err := packedMulti.Unpack(r.server.cc.keyRing)
-		resp.MultiValid = err == nil
+		valid = err == nil
```
cfromknecht: might be better to just return the error directly, and the user can infer that it's valid if there was no error

rpcserver.go

lnrpc/rpc.proto

Roasbeef:
@joostjager the separation allows distinct packing/encoding formats for a Single vs a Multi. A raw Single is also smaller than a Multi with only one channel.

PR #2893, opened 2019-04-05
channeldb/graph: filter zombie channels in FilterKnownChanIDs

cfromknecht:
This commit modifies FilterKnownChanIDs to skip edges that
we ourselves have deemed zombies. This prevents us from requesting
the updates from the remote peer, as this wastes bandwidth and cpu
cycles.

With the addition of regular historical spot checks in #2740, this will
prevent upgraded LND nodes from rerequesting these channels from
unupgraded LND nodes or implementations that don't filter out 
zombies. There are currently 3.3k zombie channels, so this would save
at least 3MB on historical syncs with such peers.

### DIFF
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -1513,10 +1513,10 @@ func (c *ChannelGraph) NodeUpdatesInHorizon(startTime, endTime time.Time) ([]Lig
 }
 
 // FilterKnownChanIDs takes a set of channel IDs and return the subset of chan
-// ID's that we don't know of in the passed set. In other words, we perform a
-// set difference of our set of chan ID's and the ones passed in. This method
-// can be used by callers to determine the set of channels ta peer knows of
-// that we don't.
+// ID's that we don't know and are not known zombies of the passed set. In other
+// words, we perform a set difference of our set of chan ID's and the ones
+// passed in. This method can be used by callers to determine the set of
+// channels another peer knows of that we don't.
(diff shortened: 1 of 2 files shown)


channeldb/graph.go
```diff
@@ -1515,7 +1515,7 @@ func (c *ChannelGraph) NodeUpdatesInHorizon(startTime, endTime time.Time) ([]Lig
 // FilterKnownChanIDs takes a set of channel IDs and return the subset of chan
 // ID's that we don't know of in the passed set. In other words, we perform a
 // set difference of our set of chan ID's and the ones passed in. This method
-// can be used by callers to determine the set of channels ta peer knows of
+// can be used by callers to determine the set of channels another peer knows of
```
halseth: nit: mention zombie behavior in the doc?

channeldb/graph.go
```diff
@@ -1529,16 +1529,29 @@ func (c *ChannelGraph) FilterKnownChanIDs(chanIDs []uint64) ([]uint64, error) {
 		if edgeIndex == nil {
 			return ErrGraphNoEdgesFound
 		}
+		zombieIndex := edges.Bucket(zombieBucket)
+		if edgeIndex == nil {
```
wpaulino: Should be checking `zombieIndex` here. Also we don't want to return an error if it doesn't exist, as it's possible that a node hasn't marked any edges as zombies and so the bucket won't exist.

cfromknecht: heh good catch. we should really consider having migrations to initialize new buckets, or always initializing all top-level buckets on start up...

PR #2900, opened 2019-04-06
Accept external URL for neutrino fee estimation.

valentinewallace:
In this commit, we enable neutrino users to specify a URL where fee
information can be retrieved for on-chain transactions.

The fee information must be a JSON object mapping `fee_by_block_target`
to fee information, e.g:
```
{
	fee_by_block_target: {
		2: 105804,
		4: 91720,
		...
		604: 1012
	}
}
```
The object nested in `fee_by_block_target` maps a number of confs to a
fee (in satoshis per kilovbyte). Note that the API's fee information need
not start at zero confs, and only fees up to 1009 confirmations are
cached/used. The reason for this is because 1009 confirmations is the 
maximum number of confs that bitcoind will return an estimate for.

Also, there may be gaps in the fee information: if the API's response
contains a fee for 2 confs and then a fee for 6 confs, it is assumed that
the fee for 3 through 5 confs is the same as the fee for 2 confs.


chainregistry.go
```diff
@@ -211,6 +211,25 @@ func newChainControlFromConfig(cfg *config, chanDB *channeldb.DB,
...
+				cc.chainNotifier,
+				defaultBitcoinStaticFeePerKW,
+			)
+
+			if err := estimator.Start(); err != nil {
+				cleanUp()
+				return nil, nil, err
+			}
+			cc.feeEstimator = estimator
+			walletConfig.FeeEstimator = estimator
```
halseth: is this necessary? Doesn't look like it is set for the other backends.

lnwallet/fee_estimator.go
```diff
@@ -458,3 +475,270 @@ func (b *BitcoindFeeEstimator) fetchEstimate(confTarget uint32) (SatPerKWeight,
 // A compile-time assertion to ensure that BitcoindFeeEstimator implements the
 // FeeEstimator interface.
 var _ FeeEstimator = (*BitcoindFeeEstimator)(nil)
+
+// WebAPIFeeEstimator is an implementation of the FeeEstimator interface that
+// queries an HTTP-based fee estimation from an existing web API.
+type WebAPIFeeEstimator struct {
```
halseth: Awesome! 😀 

Would be cool if you could add this fee estimator in its own commit.

lnwallet/fee_estimator.go
```diff
@@ -458,3 +475,270 @@ func (b *BitcoindFeeEstimator) fetchEstimate(confTarget uint32) (SatPerKWeight,
...
+// confirmation and returns the estimated fee expressed in sat/kw.
+//
+// NOTE: This method is part of the FeeEstimator interface.
+func (w *WebAPIFeeEstimator) EstimateFeePerKW(numBlocks uint32) (SatPerKWeight, error) {
+	if numBlocks > MaxBlockTarget {
+		numBlocks = MaxBlockTarget
+	}
+
+	w.feesMtx.RLock()
+	feePerKb := w.feeByBlockTarget[numBlocks]
```
halseth: handle the case where `numBlocks≈ is not found.

lnwallet/fee_estimator.go
```diff
@@ -458,3 +475,270 @@ func (b *BitcoindFeeEstimator) fetchEstimate(confTarget uint32) (SatPerKWeight,
...
+		return nil, fmt.Errorf("API fee estimator returned empty map")
+	}
+
+	return resp.FeeByBlockTarget, nil
+}
+
+// formatFeeEstimates fills in missing fee estimates in the JSON object generated
+// by the API. E.g., if the object contains a fee for 2 confs and a fee for 6
+// confs, fill in confs 3 through 5 with the value for 2.
+func (w *WebAPIFeeEstimator) formatFeeEstimates(json map[uint32]uint32) [MaxBlockTarget + 1]uint32 {
```
halseth: add a test for this?

lnwallet/fee_estimator.go
```diff
@@ -458,3 +475,270 @@ func (b *BitcoindFeeEstimator) fetchEstimate(confTarget uint32) (SatPerKWeight,
...
+//
+// NOTE: This method is part of the FeeEstimator interface.
+func (w *WebAPIFeeEstimator) RelayFeePerKW() SatPerKWeight {
+	return FeePerKwFloor
+}
+
+// parseAPIResponse attempts to parse the body of the response generated by the
+// API, into the JSON format: { fee_by_block_target: { a: b, c: d, ... y: z }}
+// where the keys are block targets and the values are satPerKB fee estimates.
+func (w *WebAPIFeeEstimator) parseAPIResponse(r io.Reader) (map[uint32]uint32, error) {
```
halseth: Test for this would be nice 👍

chainregistry.go
valentinewallace: fixed!

lnwallet/fee_estimator.go
valentinewallace: fixed heh

lnwallet/fee_estimator.go
valentinewallace: we error if it isn't now

lnwallet/fee_estimator.go
valentinewallace: removed this helper

lnwallet/fee_estimator.go
valentinewallace: fixed :) the testing could still be stronger though. i'm still thinking about how (might add a ChainNotifier harness and an HTTP API harness...?)

Roasbeef: Can mock out the `ChainNotifier` interface (just send blocks), then also mock out the HTTP request itself since we can use the interface to just return the value directly.

lnwallet/fee_estimator.go
```diff
@@ -474,3 +477,50 @@ type WebAPIFeeSource interface {
 	// specifics are left to the WebAPIFeeSource implementation.
 	ParseResponse(r io.Reader) (map[uint32]uint32, error)
 }
+
+// ExternalFeeSource is an implementation of the WebAPIFeeSource that utilizes
+// a user-specified fee estimation API for Bitcoin. It expects the response
+// to be in the JSON format: `fee_by_block_target: { ... }` where the value maps
+// block targets to fee estimates (in sat per kilovbyte).
+type ExternalFeeSource struct {
```
Roasbeef: Naming suggestion: `SparseConfFeeSource`? Sparse in that it only includes non-redundant confirmation targets.

Roasbeef: That it's external is already a given since the interface itself is called `WebAPIFeeSource`.

lnwallet/fee_estimator.go
```diff
@@ -524,3 +542,205 @@ func (e ExternalFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, error)
...
+	started uint32 // To be used atomically.
+	stopped uint32 // To be used atomically.
+
+	// apiSource is the backing web API source we'll use for our queries.
+	apiSource WebAPIFeeSource
+
+	// chainNotifier is used by the WebAPIFeeEstimator to determine when to
+	// re-query the API for fresh fees. We re-query whenever a new block
+	// comes in.
+	chainNotifier chainntnfs.ChainNotifier
```
Roasbeef: Does it need both the notifier and a block epoch attribute? Seems like it could do with one or the other.

lnwallet/fee_estimator.go
```diff
@@ -524,3 +542,205 @@ func (e ExternalFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, error)
...
+
+	return satPerKw, nil
+}
+
+// Start signals the FeeEstimator to start any processes or goroutines it needs
+// to perform its duty.
+//
+// NOTE: This method is part of the FeeEstimator interface.
+func (w *WebAPIFeeEstimator) Start() error {
+	if !atomic.CompareAndSwapUint32(&w.started, 0, 1) {
```
Roasbeef: We've largely moved to using a [`sync.Once`](https://golang.org/pkg/sync/#Once) across the codebase.

lnwallet/fee_estimator.go
```diff
@@ -524,3 +542,205 @@ func (e ExternalFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, error)
...
+		return nil
+	}
+
+	if err := w.chainNotifier.Start(); err != nil {
```
Roasbeef: The notifier should already be started at this point, so we'd only need to register for the epoch notifications.

lnwallet/fee_estimator.go
```diff
@@ -524,3 +542,205 @@ func (e ExternalFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, error)
...
+
+	return nil
+}
+
+// Stop stops any spawned goroutines and cleans up the resources used by the
+// fee estimator.
+//
+// NOTE: This method is part of the FeeEstimator interface.
+func (w *WebAPIFeeEstimator) Stop() error {
+	if !atomic.CompareAndSwapUint32(&w.stopped, 0, 1) {
```
Roasbeef: Similar comment here re `sync.Once`. We'd then be able to get rid of the atomic variables.

lnwallet/fee_estimator.go
```diff
@@ -524,3 +542,205 @@ func (e ExternalFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, error)
...
+// confirmation and returns an estimated fee (if one was returned by the API). If
+// the fee was not previously cached, we cache it here. Note that the feesMtx
+// should be held while calling this function, to avoid a race condition.
+func (w *WebAPIFeeEstimator) getCachedFee(numBlocks uint32) (uint32, error) {
+	for target := numBlocks; target >= MinBlockTarget; target-- {
+		fee, ok := w.feeByBlockTarget[target]
+		if !ok {
+			continue
+		}
+		w.feeByBlockTarget[target] = fee
```
Roasbeef: Style nit: missing a space here.

lnwallet/fee_estimator.go
```diff
@@ -524,3 +542,205 @@ func (e ExternalFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, error)
...
+// NOTE: This method is part of the FeeEstimator interface.
+func (w *WebAPIFeeEstimator) RelayFeePerKW() SatPerKWeight {
+	return FeePerKwFloor
+}
+
+// getCachedFee takes in a target for the number of blocks until an initial
+// confirmation and returns an estimated fee (if one was returned by the API). If
+// the fee was not previously cached, we cache it here. Note that the feesMtx
+// should be held while calling this function, to avoid a race condition.
+func (w *WebAPIFeeEstimator) getCachedFee(numBlocks uint32) (uint32, error) {
```
Roasbeef: Should have tests for. Also I think we can just obtain the mutex within here rather than externally at the call site.

lnwallet/fee_estimator.go
Roasbeef: This won't cache the original value, instead it'll cache the highest value that's present, which is already in the map.

lnwallet/fee_estimator.go
valentinewallace: done :)

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: n/a since we just hit the API on a random timer now

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: n/a since we just hit the API on a random timer now

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: oo... fixed

lnwallet/fee_estimator.go
valentinewallace: it's tested indirectly now, let me know if it should be tested even more?

lnwallet/fee_estimator.go
```diff
@@ -474,3 +477,50 @@ type WebAPIFeeSource interface {
...
+	kresp, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	type jsonResp struct {
+		FeeByBlockTarget map[uint32]uint32 `json:"fee_by_block_target"`
+	}
+
+	resp := jsonResp{FeeByBlockTarget: make(map[uint32]uint32)}
```
halseth: style: newline before setting fields.

lnwallet/fee_estimator.go
```diff
@@ -474,3 +477,50 @@ type WebAPIFeeSource interface {
...
+	dec := bytes.NewReader(kresp)
```
halseth: why not use `r` directly?

lnwallet/fee_estimator_test.go
```diff
@@ -89,3 +91,60 @@ func TestStaticFeeEstimator(t *testing.T) {
...
+	url := "test"
+	feeSource := lnwallet.SparseConfFeeSource{URL: url}
+	queryURL := feeSource.GenQueryURL()
+	if queryURL != url {
+		t.Fatalf("expected query URL of %v, got %v", url, queryURL)
+	}
+
+	// Test parsing a properly formatted JSON API response.
+	// First, create the response as a bytes.Reader.
+	testFees := map[uint32]uint32{1: 12345, 2: 42, 3: 54321}
```
halseth: style nit: space out this initialization with some newlines

lnwallet/fee_estimator_test.go
```diff
@@ -89,3 +91,60 @@ func TestStaticFeeEstimator(t *testing.T) {
...
+		t.Fatalf("unable to marshal JSON API response: %v", err)
+	}
+	reader := bytes.NewReader(jsonResp)
+
+	// Finally, ensure the expected map is returned without error.
+	fees, err := feeSource.ParseResponse(reader)
+	if err != nil {
+		t.Fatalf("unable to parse API response: %v", err)
+	}
+	if len(fees) != len(testFees) {
```
halseth: can use `reflect.DeepEqual` maybe?

lnwallet/fee_estimator.go
```diff
@@ -524,3 +549,206 @@ func (s SparseConfFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, erro
...
+	w.feesMtx.Unlock()
+}
+
+// feeUpdateManager updates the fee estimates whenever a new block comes in.
+func (w *WebAPIFeeEstimator) feeUpdateManager() {
+	defer w.wg.Done()
+
+	for {
+		select {
+		case <-w.updateFeeTimer.C:
```
halseth: could we just update the fee estimate on demand instead of having a goroutine running to keep it up to date?

lnwallet/fee_estimator_test.go
```diff
@@ -3,12 +3,28 @@ package lnwallet_test
...
+	"io"
+	"strings"
 	"testing"
 
 	"github.com/btcsuite/btcutil"
+
 	"github.com/lightningnetwork/lnd/lnwallet"
 )
 
+type mockSparseConfFeeSource struct {
```
halseth: I'm reading `coffee` everytime :)

lnwallet/fee_estimator.go
```diff
@@ -15,6 +21,25 @@ const (
 	// FeePerKwFloor is the lowest fee rate in sat/kw that we should use for
 	// determining transaction fees.
 	FeePerKwFloor SatPerKWeight = 253
+
+	// MaxBlockTarget is the highest number of blocks confirmations that
+	// a WebAPIFeeEstimator will cache fees for. This number is chosen
+	// because it's the highest number of confs bitcoind will return a fee
+	// estimate for.
+	MaxBlockTarget uint32 = 1009
```
halseth: looks like these constants don't have to be exported?

config.go
```diff
@@ -129,6 +129,7 @@ type neutrinoConfig struct {
 	MaxPeers     int           `long:"maxpeers" description:"Max number of inbound and outbound peers"`
 	BanDuration  time.Duration `long:"banduration" description:"How long to ban misbehaving peers.  Valid time units are {s, m, h}.  Minimum 1 second"`
 	BanThreshold uint32        `long:"banthreshold" description:"Maximum allowed ban score before disconnecting and banning misbehaving peers."`
+	FeeURL       string        `long:"feeurl" description:"Use this URL to for fee estimation."`
```
halseth: Description should note is is optional.

chainregistry.go
```diff
@@ -132,7 +132,7 @@ type chainControl struct {
 func newChainControlFromConfig(cfg *config, chanDB *channeldb.DB,
 	privateWalletPw, publicWalletPw []byte, birthday time.Time,
 	recoveryWindow uint32, wallet *wallet.Wallet,
-	neutrinoCS *neutrino.ChainService) (*chainControl, func(), error) {
```
halseth: 👍

chainregistry.go
halseth: Hm, I'm wondering... maybe it is a bug that it is not set for any backend?

We do set `walletConfig.FeeEstimator = cc.estimator` initially, but not sure if changing `cc.estimator` afterwards update the original reference? 🤔

valentinewallace: i checked and the original reference is not updated :( but also, I can't see `walletConfig.FeeEstimator` being used anywhere IIRC. I'll double check and prob add a commit removing it.

lnwallet/fee_estimator.go
valentinewallace: Hm, what do you think about updating it on demand, but only if X minutes have passed since it was last updated? Could be worthwhile to avoid the API getting overwhelmed?

chainregistry.go
valentinewallace: That field isn't used anywhere so I added a commit removing it. Lmk if that's acceptable :thinking:

lnwallet/fee_estimator_test.go
valentinewallace: lolol.. can't unsee

lnwallet/fee_estimator_test.go
valentinewallace: ahhh true. fixed

lnwallet/fee_estimator_test.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: wow, this improved the function a lot. fixed

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: fixed :sweat_smile:

config.go
valentinewallace: fixed

lnwallet/fee_estimator.go
halseth: Yeah, idea is to request the fee and cache it first time the estimator is queried, then request it again only after some time has passed. Would that lead to more requests to the API than the current approach?

chainregistry.go
halseth: Ah, even better! 💯

lnwallet/fee_estimator.go
Roasbeef: The timer approach means that the load on w/e API is bounded and can be extrapolated ahead of time (based on the number of clients). Otherwise, (within lnd) routine things like sweeping multiple HTLCs or updating the commitment fee rate will trigger far more load if we query directly each time we need an estimate.

halseth: It would go like this:
```go
func EstimateFee(blocks int) {
    _, ok := cache[blocks]
    if time.Now() > lastTime.Add(cacheInvalidationTimeout) || !ok {
          fetchAndAddToCache(...)
    }

    return cache[blocks]
```
}

lnwallet/fee_estimator_test.go
```diff
@@ -144,3 +160,71 @@ func TestSparseConfFeeSource(t *testing.T) {
 		t.Fatalf("expected ParseResponse to fail")
 	}
 }
+
+func TestWebAPIFeeEstimator(t *testing.T) {
```
Roasbeef: Missing `godoc` comment for the test.

lnwallet/fee_estimator.go
```diff
@@ -518,3 +543,206 @@ func (s SparseConfFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, erro
...
+type WebAPIFeeEstimator struct {
+	started sync.Once
+	stopped sync.Once
+
+	// apiSource is the backing web API source we'll use for our queries.
+	apiSource WebAPIFeeSource
+
+	// updateFeeTimer is the timer responsible for updating the Estimator's
+	// fee estimates every time it fires.
+	updateFeeTimer *time.Timer
```
Roasbeef: If this uses a `time.Ticker` then it doesn't need to be reset. Going further, it can use our `ticker.Ticker` to make it easier to mock out in unit tests.

lnwallet/fee_estimator.go
```diff
@@ -518,3 +543,206 @@ func (s SparseConfFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, erro
...
+// Start signals the FeeEstimator to start any processes or goroutines it needs
+// to perform its duty.
+//
+// NOTE: This method is part of the FeeEstimator interface.
+func (w *WebAPIFeeEstimator) Start() error {
+	var err error
+	w.started.Do(func() {
+		walletLog.Infof("Starting web API fee estimator")
+
+		w.updateFeeTimer = time.NewTimer(w.randomFeeUpdateTimeout())
```
Roasbeef: See comment above about using a ticker instead.

lnwallet/fee_estimator.go
```diff
@@ -518,3 +543,206 @@ func (s SparseConfFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, erro
...
+}
+
+// getCachedFee takes in a target for the number of blocks until an initial
+// confirmation and returns an estimated fee (if one was returned by the API). If
+// the fee was not previously cached, we cache it here. Note that the feesMtx
+// should be held while calling this function, to avoid a race condition.
+func (w *WebAPIFeeEstimator) getCachedFee(numBlocks uint32) (uint32, error) {
+	w.feesMtx.Lock()
+	defer w.feesMtx.Unlock()
+	for target := numBlocks; target >= minBlockTarget; target-- {
```
Roasbeef: Style nit: lacks new line above, and comment for broader loop.

lnwallet/fee_estimator.go
```diff
@@ -518,3 +543,206 @@ func (s SparseConfFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, erro
...
+		fee, ok := w.feeByBlockTarget[target]
+		if !ok {
+			continue
+		}
+
+		w.feeByBlockTarget[numBlocks] = fee
```
Roasbeef: We only need to override the entry if it wasn't already found in the map.

lnwallet/fee_estimator.go
```diff
@@ -518,3 +543,206 @@ func (s SparseConfFeeSource) ParseResponse(r io.Reader) (map[uint32]uint32, erro
...
+}
+
+// feeUpdateManager updates the fee estimates whenever a new block comes in.
+func (w *WebAPIFeeEstimator) feeUpdateManager() {
+	defer w.wg.Done()
+
+	for {
+		select {
+		case <-w.updateFeeTimer.C:
+			w.updateFeeTimer.Reset(w.randomFeeUpdateTimeout())
```
Roasbeef: No need to reset if a ticker is used.

lnwallet/fee_estimator_test.go
```diff
@@ -144,3 +160,71 @@ func TestSparseConfFeeSource(t *testing.T) {
...
+		blockTargetSmallFee: 42,
+		validBlockTarget:    54321,
+	}
+	feeSource := mockSparseConfFeeSource{
+		url:  "https://www.github.com",
+		fees: testFees,
+	}
+	estimator := lnwallet.NewWebAPIFeeEstimator(feeSource, 10)
+
+	// Test that requesting a fee for a too-low block target will error.
```
Roasbeef: Many of these assertions below can be converted to use [sub-tests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks).

lnwallet/fee_estimator_test.go
Roasbeef: Not clear that this test covers the case of falling back to a lower estimate if the target num blocks conf isn't found in the map. The `getCachedFee` method may need to be test directly in order to achieve this level of coverage.

valentinewallace: fixed & fixed

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator.go
valentinewallace: fixed

lnwallet/fee_estimator_test.go
valentinewallace: fixed

PR #2901, opened 2019-04-06
peer+brontide: dc partial writes

cfromknecht:
See commit messages for more details

### DIFF
--- a/brontide/noise.go
+++ b/brontide/noise.go
@@ -703,7 +703,7 @@ func (b *Machine) WriteMessage(w io.Writer, p []byte) error {
 
 	// First, write out the encrypted+MAC'd length prefix for the packet.
 	cipherLen := b.sendCipher.Encrypt(nil, nil, pktLen[:])
-	if _, err := w.Write(cipherLen); err != nil {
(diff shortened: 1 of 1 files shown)


brontide/noise.go
```diff
@@ -703,15 +703,30 @@ func (b *Machine) WriteMessage(w io.Writer, p []byte) error {
...
 	// single packet, as any fragmentation should have taken place at a
 	// higher level.
 	cipherText := b.sendCipher.Encrypt(nil, nil, p)
-	_, err := w.Write(cipherText)
+	return safeWrite(w, cipherText)
+}
+
+// safeWrite writes the message p to the given io.Writer. An error is returned
+// if the message was partially written, i.e. the number of bytes written was
+// in the range [1, len(p)). This allows us to propagate timeout errors that
```
halseth: Ah, so for `n == 0` we want to return the error unaltered?

wpaulino: Yeah because we never wrote anything, so it was most likely a timeout or EOF.

cfromknecht:
closing in favor of #2924

PR #2913, opened 2019-04-08
cnct+htlcswitch+invoices: move invoice parameter check to registry

joostjager:
This PR simplifies the link implementation by moving out all references to invoices. As an exit hop, the link offers its htlc to the invoice registry. The invoice registry decides how and when the link should respond to the htlc.

Checking of the cltv delta and amount is now performed in the invoice registry. This concentrates the logic around invoice acceptance and settling in the invoice registry. It increases maintainability and is a also preparatory step for implementation of multi path payments.

In addition to that, moving the htlc parameter check to the invoice registry allows us to apply the exact same logic for on-chain resolution of an incoming htlc. This fixes some unaddressed edge cases:

* On-chain resolution of an incoming htlc for a canceled invoice. Previously the htlc would still be settled even though the invoice database indicates that we should reject payments to the invoice. The htlc was pulled but the invoice stayed in the canceled stayed.

* On-chain resolution of an incoming htlc for an open invoice. Previously the invoice was marked settled as long as the htlc hadn't expired yet. But if the number of remaining blocks until expiry is low, it could 
happen that we couldn't pull the htlc in the end.

* On-chain resolution of an incoming htlc for a hodl invoice. Previously the hodl invoice would move to the accepted state regardless of the number of remaining blocks until htlc expiry. It could be that invoices move to the accepted state without meeting the invoice min cltv delta requirement. This violates the assumption of the user that there are still enough blocks left when the accepted state transition is received.

To reduce the number of execution paths for incoming htlc resolution, this PR modifies the channel arbitrator to always send htlcs via the incoming contest resolver. This allows us to perform the check with the invoice registry in a single location. 

Furthermore this change makes it unnecessary to already determine earlier in the process whether an incoming contest or a success resolver is needed. This simplifies `channel` and `channel_arbitrator`.


joostjager:
Linking https://github.com/lightningnetwork/lnd/pull/2935. This may make it easier to handle hodl invoices and incoming preimages in the contract resolvers.

htlcswitch/link.go
```diff
@@ -2868,8 +2782,12 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 	// Notify the invoiceRegistry of the exit hop htlc. If we crash right
 	// after this, this code will be re-executed after restart. We will
 	// receive back a resolution event.
+	blocksToExpiry := int32(pd.Timeout) - int32(heightNow)
```
cfromknecht: not sure that converting to `int32` helps here. it's still susceptible to underflow, which would later satisfy either of the equations: `blocksTilExpiry >= finalExpiry` or `blocksTilExpiry >= finalRejectDelta`. however, checking that `pd.Timeout > heightNow` only needs to be on the first accept, which goes in line with my other comment passing both distinctly into `NotifyExitHopHtlc` and validating this in the invoice registry

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -57,8 +57,11 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 		return nil, err
 	}
 
-	// If we're past the point of expiry of the HTLC, then at this point
-	// the sender can sweep it, so we'll end our lifetime.
+	// If we're past the point of expiry of the HTLC, then at this point the
+	// sender can sweep it, so we'll end our lifetime. Here we deliberately
+	// forego the chance that the sender doesn't sweep and we already have
+	// or will learn the preimage. Otherwise the resolver could potentially
+	// stay active indefinitely and the channel will never close properly.
```
cfromknecht: are we sure we want to do this? seems like if the htlc is unswept by either party the channel _should_ remain open. perhaps there's an argument to be made that in this case we should offload to the stray output pool, to decouple it from the closing of the channel. but that seems easier to add on later if the contract hasn't been closed, than trying to resurrect everything after the resolver state has been cleaned up

joostjager: It was already like that, I just clarified the rationale in the comment. Would you like to change that behaviour in this PR? I am not sure how users will react to these stuck channels, as long as we don't have a stray output pool.

Btw, for exit hops this check is actually unnecessary, because further down `NotifyExitHopHtlc` will also return a cancel action. For exit hops we really want to close the contract, because accepting it would violate the invoice final cltv requirement.

lnwallet/channel.go
```diff
@@ -1309,13 +1309,6 @@ type LightningChannel struct {
 	// signatures, of which there may be hundreds.
 	sigPool *SigPool
 
-	// pCache is the global preimage cache shared across all other
-	// LightningChannel instance. We'll use this cache either when we force
-	// close, or we detect that the remote party has force closed. If the
-	// preimage for an incoming HTLC is found in the cache, then we'll try
-	// to claim it on chain.
-	pCache PreimageCache
```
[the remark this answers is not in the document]
joostjager: Yes it can indeed. Removed

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -117,14 +131,7 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 	// ensure the preimage can't be delivered between querying and
 	// registering for the preimage subscription.
 	preimageSubscription := h.PreimageDB.SubscribeUpdates()
-	blockEpochs, err := h.Notifier.RegisterBlockEpochNtfn(nil)
-	if err != nil {
-		return nil, err
-	}
-	defer func() {
-		preimageSubscription.CancelSubscription()
-		blockEpochs.Cancel()
```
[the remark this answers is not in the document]
joostjager: Yes I noticed that too indeed. Probably never happened, but still nice to be correct.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -160,19 +163,27 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
 	}
 
 	for {
 
 		select {
 		case preimage := <-preimageSubscription.WitnessUpdates:
-			if !tryApplyPreimage(preimage) {
+			// We receive all new preimages, so we need to ignore
+			// all except the preimage we are waiting for.
+			if !preimage.Matches(h.payHash) {
```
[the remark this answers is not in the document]
joostjager: Yes I considered that, but didn't like to rely on the implementation detail that there is just a single error case. Actually that check in `applyPreimage` is just a sanity check, I was in doubt whether to actually add it at all.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -186,9 +197,10 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 				continue
 			}
 
-			if !tryApplyPreimage(*hodlEvent.Preimage) {
-				continue
```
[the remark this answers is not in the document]
joostjager: Yes exactly. To be stricter about this and catch potential bugs.

contractcourt/channel_arbitrator.go
```diff
@@ -1122,7 +1122,30 @@ func (c *ChannelArbitrator) checkChainActions(height uint32,
...
+		// Start by checking the preimage cache for preimages of
+		// forwarded HTLCs.
+		_, preimageAvailable := c.cfg.PreimageDB.LookupPreimage(
+			htlc.RHash,
+		)
+
+		// Then check if we have an invoice that can be settled by this
+		// HTLC.
+		//
+		// TODO: Check that there are still more blocks remaining than
```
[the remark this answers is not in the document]
joostjager: I was wondering, why do we add our names there actually? Git blame shows the person that added it. In my IDE hover over shows it.

contractcourt/channel_arbitrator.go
```diff
@@ -1122,7 +1122,30 @@ func (c *ChannelArbitrator) checkChainActions(height uint32,
...
+		// the invoice cltv delta. We don't want to go to chain only to
+		// have the incoming contest resolver decide that we don't want
+		// to settle this invoice.
+		invoice, _, err := c.cfg.Registry.LookupInvoice(htlc.RHash)
```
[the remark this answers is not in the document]
joostjager: Extracted into function and also added stricter `err` handling

witness_beacon.go
```diff
@@ -72,25 +69,6 @@ func (p *preimageBeacon) LookupPreimage(
 	p.RLock()
 	defer p.RUnlock()
 
-	// First, we'll check the invoice registry to see if we already know of
-	// the preimage as it's on that we created ourselves.
-	invoice, _, err := p.invoices.LookupInvoice(payHash)
```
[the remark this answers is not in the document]
joostjager: Yes, now that we sometimes not want to settle invoices that we could, we needed this.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -134,26 +134,47 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
+		if err := applyPreimage(*e.Preimage); err != nil {
+			return nil, err
+		}
+
+		return &h.htlcSuccessResolver, nil
+	}
+
 	// Create a buffered hodl chan to prevent deadlock.
 	hodlChan := make(chan interface{}, 1)
+	defer h.Registry.HodlUnsubscribeAll(hodlChan)
```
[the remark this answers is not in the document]
joostjager: It got a bit more complicated with the `ErrInvoiceNotFound` case. But refactored the code and made it correct using a `switch` statement.

htlcswitch/link.go
```diff
@@ -2739,94 +2749,14 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		return false, nil
 	}
 
-	// First, we'll check the expiry of the HTLC itself against, the current
```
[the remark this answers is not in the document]
joostjager: Yes, we need to divide and conquer.

htlcswitch/link.go
```diff
@@ -2868,8 +2782,12 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 	// Notify the invoiceRegistry of the exit hop htlc. If we crash right
 	// after this, this code will be re-executed after restart. We will
 	// receive back a resolution event.
+	blocksToExpiry := int32(pd.Timeout) - int32(heightNow)
+	invoiceHash := lntypes.Hash(pd.RHash)
```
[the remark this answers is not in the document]
joostjager: I would think also, but if I change it it does complain `cannot use pd.RHash (type lnwallet.PaymentHash) as type lntypes.Hash in argument to l.cfg.Registry.NotifyExitHopHtlc`

invoices/invoiceregistry.go
```diff
@@ -27,12 +28,57 @@ var (
...
+
+	// CancelInvoiceCanceled is returned when the invoice is already
+	// canceled and can't be paid to anymore.
+	CancelInvoiceCanceled
+
+	// CancelAmountTooLow is returned when the amount paid is too low.
+	CancelAmountTooLow
+)
+
+func (r HtlcCancelReason) String() string {
```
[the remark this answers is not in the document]
joostjager: added

invoices/invoiceregistry.go
```diff
@@ -27,12 +28,57 @@ var (
...
+	ErrInvoiceAmountTooLow = errors.New("paid amount less than invoice amount")
+)
+
 // HodlEvent describes how an htlc should be resolved. If HodlEvent.Preimage is
 // set, the event indicates a settle event. If Preimage is nil, it is a cancel
 // event.
 type HodlEvent struct {
-	Preimage *lntypes.Preimage
-	Hash     lntypes.Hash
+	Preimage     *lntypes.Preimage
```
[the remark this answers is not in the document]
joostjager: added

contractcourt/htlc_incoming_resolver_test.go
```diff
@@ -0,0 +1,272 @@
...
+	successResolver, ok := i.nextResolver.(*htlcSuccessResolver)
+	if !ok {
+		i.t.Fatal("expected succesresolver")
+	}
+
+	if successResolver.htlcResolution.Preimage != testResPreimage {
+		i.t.Fatal("invalid preimage")
+	}
+
+	if successResolver.htlcResolution.SignedSuccessTx != nil &&
```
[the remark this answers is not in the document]
joostjager: Yes, better

htlcswitch/link_test.go
```diff
@@ -5786,15 +5785,102 @@ func TestChannelLinkHoldInvoiceCancel(t *testing.T) {
...
-	case <-time.After(5 * time.Second):
-		t.Fatal("timeout")
+		t.Fatal("expected unknown payment hash")
+	}
+}
+
+// TestChannelLinkHoldInvoiceRestart asserts hodl htlcs are held after blocks
+// are mined and the link is restarted. The initial expiry checks should not
+// apply to hodl htlcs after restart.
+func TestChannelLinkHoldInvoiceRestart(t *testing.T) {
```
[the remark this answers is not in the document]
joostjager: Cool. Yes especially with lots of channel action, it removes the cluttering boiler plate.

htlcswitch/link.go
```diff
@@ -2739,94 +2749,14 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
...
 	// the HTLC to ensure that it was crafted correctly by the sender and
 	// matches the HTLC we were extended.
-	//
-	// NOTE: We make an exception when the value requested by the invoice is
-	// zero. This means the invoice allows the payee to specify the amount
-	// of satoshis they wish to send.  So since we expect the htlc to have a
-	// different amount, we should not fail.
-	if !l.cfg.DebugHTLC && invoice.Terms.Value > 0 &&
-		fwdInfo.AmountToForward < invoice.Terms.Value {
+	if pd.Amount != fwdInfo.AmountToForward {
```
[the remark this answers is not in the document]
joostjager: added

htlcswitch/link.go
joostjager: Suppose `pd.Timeout` is `3` and `heightNow` is `5`. `blocksToExpiry` would then be `-2` and `blocksTilExpiry >= finalExpiry` = `-2 >= finalExpiry` would be false?

htlcswitch/link.go
```diff
@@ -2868,8 +2782,12 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 	// Notify the invoiceRegistry of the exit hop htlc. If we crash right
 	// after this, this code will be re-executed after restart. We will
 	// receive back a resolution event.
+	blocksToExpiry := int32(pd.Timeout) - int32(heightNow)
+	invoiceHash := lntypes.Hash(pd.RHash)
+
 	event, err := l.cfg.Registry.NotifyExitHopHtlc(
-		invoiceHash, pd.Amount, l.hodlQueue.ChanIn(),
+		invoiceHash, pd.Amount, blocksToExpiry,
```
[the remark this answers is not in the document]
joostjager: Refactored now.

htlcswitch/link.go
joostjager: But this check has now indeed been moved to `invoice registry`.

joostjager:
@cfromknecht ptal

joostjager:
Integration test `testMultiHopReceiverChainClaim` is failing with these changes. The cause is interesting: Alice sends a payment to Carol with a final cltv expiry of 40. Carol hodls the htlc and the channel is closed on chain. Then the commitment tx is published and mined, after which the resolver takes over. At that point there are just 39 blocks left. According to the new (consistent) logic introduced in this PR, that is not enough and the invoice is canceled.

Wondering more in general, don't we allow a one block grace period in case a new block comes in between creation of the route by the sender and receiving the htlc by the exit hop?

htlcswitch/link.go
cfromknecht: The case I'm thinking about is when `pd.Timeout` is very high (as a unit32), like `0x80000000`. In this example, `blocksTilExpiry` is now `2147483643` which exceeds any reasonable `finalExpriy`

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
cfromknecht: can't say this is the only reason, but probably so that one doesn't need an IDE to be able to discern that. some of us peruse on mobile :)

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
cfromknecht: much clearer 🎸

htlcswitch/link.go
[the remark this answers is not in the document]
cfromknecht: ah you're right, was thinking `pd.RHash` was `[32]byte`, but indeed it is a named type so the conversion cannot be done automatically

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: Added my name

htlcswitch/link.go
joostjager: Rewritten the comparison without subtraction

joostjager:
Integration test fixed. Created issue for invoice final expiry grace: #3055

Roasbeef:
> Then the commitment tx is published and mined, after which the resolver takes over. At that point there are just 39 blocks left. According to the new (consistent) logic introduced in this PR, that is not enough and the htlc is canceled.

Why should the HTLC be cancelled? It should be swept since Carol has the preimage.

lnd_multihoplocalchainclaim_test.go
```diff
@@ -0,0 +1,352 @@
+// +build rpctest
```
Roasbeef: How about `lnd_multi_hop_local_chain_claim_test.go ` instead of `lnd_multihoplocalchainclaim_test.go `? The former is far more readable IMO. Also applies to all the other files created in this commit.

lnd_multihoplocalchainclaim_test.go
```diff
@@ -23,7 +23,8 @@ func testMultiHopHtlcLocalChainClaim(net *lntest.NetworkHarness, t *harnessTest)
 	// First, we'll create a three hop network: Alice -> Bob -> Carol, with
 	// Carol refusing to actually settle or directly cancel any HTLC's
 	// self.
-	aliceChanPoint, bobChanPoint, carol := createThreeHopHodlNetwork(t, net)
+	aliceChanPoint, bobChanPoint, carol :=
```
Roasbeef: I think this can just be folded as: 
```go
aliceChanPoint, bobChanPoint, carol := createThreeHopNetwork(
    t, net, true, 
)
```

Same applies to all other instances in this file.

channeldb/invoices.go
```diff
@@ -631,8 +631,12 @@ func (d *DB) QueryInvoices(q InvoiceQuery) (InvoiceSlice, error) {
 //
 // When the preimage for the invoice is unknown (hold invoice), the invoice is
 // marked as accepted.
+//
+// TODO: Store invoice cltv as separate field in database so that it doesn't
+// need to be decoded from the payment request.
```
Roasbeef: I've also noticed that at times the pubkey recovery from the signature can show up in profiles when benchmarking payments. We may also want to add a flag to the decode routing that lets it skip obtaining the public key as well.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -134,28 +134,52 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
+	processHodlEvent := func(e invoices.HodlEvent) (ContractResolver,
+		error) {
+
+		if e.Preimage == nil {
+			log.Infof("%T(%v): Exit hop HTLC canceled "+
+				"(expiry=%v, height=%v), abandoning", h,
+				h.htlcResolution.ClaimOutpoint,
+				h.htlcExpiry, currentHeight)
+
+			h.resolved = true
```
Roasbeef: It's a change in behavior to checkpoint and mark the contract has being resolved. It isn't _fully_ resolved here in that it still needs to turn into a regular success resolver.

channeldb/invoice_test.go
```diff
@@ -99,7 +99,10 @@ func TestInvoiceWorkflow(t *testing.T) {
 	// now have the settled bit toggle to true and a non-default
 	// SettledDate
 	payAmt := fakeInvoice.Terms.Value * 2
-	if _, err := db.AcceptOrSettleInvoice(paymentHash, payAmt); err != nil {
+	_, err = db.AcceptOrSettleInvoice(
```
Roasbeef: I think this commit could've easily been split up into multiple commits. There's a lot going on in this commit across multiple files and packages.

invoices/invoiceregistry.go
```diff
@@ -27,12 +28,64 @@ var (
...
+	// ErrInvoiceAmountTooLow is returned  when an invoice is attempted to be
+	// accepted or settled with an amount that is too low.
+	ErrInvoiceAmountTooLow = errors.New("paid amount less than invoice amount")
+)
+
 // HodlEvent describes how an htlc should be resolved. If HodlEvent.Preimage is
 // set, the event indicates a settle event. If Preimage is nil, it is a cancel
 // event.
 type HodlEvent struct {
+	// Preimage is the htlc preimage. It's value is nil in case of a cancel.
```
Roasbeef: It's -> Its

invoices/invoiceregistry.go
```diff
@@ -460,6 +521,33 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
 	return invoice, expiry, nil
 }
 
+func (i *InvoiceRegistry) checkHtlcParameters(invoice *channeldb.Invoice,
```
Roasbeef: Missing a `godoc` comment.

lnd_multihopreceiverchainclaim_test.go
```diff
@@ -22,52 +24,63 @@ import (
 func testMultiHopReceiverChainClaim(net *lntest.NetworkHarness, t *harnessTest) {
 	ctxb := context.Background()
 
-	// First, we'll create a three hop network: Alice -> Bob -> Carol, with
-	// Carol refusing to actually settle or directly cancel any HTLC's
-	// self.
+	// First, we'll create a three hop network: Alice -> Bob -> Carol.
 	aliceChanPoint, bobChanPoint, carol :=
-		createThreeHopNetwork(t, net, true)
+		createThreeHopNetwork(t, net, false)
```
Roasbeef: Where is this ever set to `true` now?

lnd_multihopreceiverchainclaim_test.go
```diff
@@ -22,52 +24,63 @@ import (
...
-		t.Fatalf("unable to create payment stream for alice: %v", err)
-	}
-	err = alicePayStream.Send(&lnrpc.SendRequest{
-		PaymentRequest: carolInvoice.PaymentRequest,
-	})
-	if err != nil {
-		t.Fatalf("unable to send payment: %v", err)
-	}
+	payErr := make(chan error, 1)
+	go func() {
```
Roasbeef: Is this change necessary? It emulates the prior behavior, but with an additional goroutine.

Roasbeef: A similar comment applies to the instance below.

lnwallet/channel.go
```diff
@@ -5526,10 +5516,12 @@ func newIncomingHtlcResolution(signer input.Signer, localChanCfg *channeldb.Chan
...
 
-	// Next, we'll construct the full witness needed to satisfy the input
-	// of the success transaction.
+	// Next, we'll construct the full witness needed to satisfy the input of
+	// the success transaction. Don't specify the preimage yet. The preimage
+	// will be supplied by the contract resolver, either directly or when it
+	// becomes known.
 	successWitness, err := input.ReceiverHtlcSpendRedeem(
-		htlc.Signature, preimage[:], signer, &successSignDesc, successTx,
+		htlc.Signature, nil, signer, &successSignDesc, successTx,
```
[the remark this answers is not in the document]
joostjager: It already needed to be aware in the case that the preimage was discovered later. Previously the resolution wasn't always fully populated either at this point (error ignored):

https://github.com/lightningnetwork/lnd/blob/f8c824fb1d6c5ef8524148f59ea5650af65af98b/lnwallet/channel.go#L5634

In this PR it is always populated in the `incomingContestResolver`.

contractcourt/channel_arbitrator.go
```diff
@@ -1122,7 +1125,13 @@ func (c *ChannelArbitrator) checkChainActions(height uint32,
 		// know the pre-image and it's close to timing out. We need to
 		// ensure that we claim the funds that our rightfully ours
 		// on-chain.
-		if _, ok := c.cfg.PreimageDB.LookupPreimage(htlc.RHash); !ok {
+
+		preimageAvailable, err := c.isPreimageAvailable(htlc.RHash)
```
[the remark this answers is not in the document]
joostjager: Yes here it should still check both sources. But in the `incomingContestResolver` the logic for both sources is different. Invoices are not settled 'greedy' anymore, but according to the same rules that apply in the link.

lnd_multihopreceiverchainclaim_test.go
joostjager: In all the unchanged tests still in `lnd_test.go` that use a hodl network.

contractcourt/htlc_incoming_contest_resolver.go
joostjager: The change in behaviour is that we sometimes don't pull htlcs for invoices for which we do have the preimage. We don't pull if there are not enough blocks left to satisfy the final cltv expiry requirement of the invoice. In exact the same way as we do it in the link.

Suppose we have a hodl invoice with final cltv 100 blocks. An htlc comes in with just 20 blocks left. This htlc would normally be canceled in the link. But if the link goes down after lock-in but before canceling the htlc and we go to chain, we'll end up here in the resolver.

Without this change, we would just mark that invoice as accepted (and notify listeners), even if there were only still 20 blocks left. Sending the accepted event under those conditions would probably violate the caller's assumptions.

lnd_multihoplocalchainclaim_test.go
joostjager: Fixed

lnd_multihoplocalchainclaim_test.go
joostjager: Renamed

channeldb/invoices.go
joostjager: Ok, but out of scope for this PR? Also the fix direction could either be to create a light payment decoding function or move the final cltv value in its own db key.

lnd_multihopreceiverchainclaim_test.go
joostjager: Reverted change. It was a left over from a previous different approach to fixing the itests.

joostjager:
> > Then the commitment tx is published and mined, after which the resolver takes over. At that point there are just 39 blocks left. According to the new (consistent) logic introduced in this PR, that is not enough and the htlc is canceled.
> 
> Why should the HTLC be cancelled? It should be swept since Carol has the preimage.

See explanation in comments above. Having the preimage for an invoice doesn't always mean we should settle the htlc. At least that is the idea introduced in this PR.

contractcourt/htlc_incoming_contest_resolver.go
joostjager: Another consistency that existed previously: an invoice with final ctlv 20 and an htlc is locked-in with 30 blocks to expiry. In the link the invoice would then be marked settled and the htlc settled too. But if the same thing would happen in a resolver, we would all of a sudden postpone marking the invoice as settled until the final sweep completed. During this time, the sender of the payment is wondering why his payment is stuck.

channeldb/invoice_test.go
joostjager: How would you do this easily? This commit is a set of changes that are all interconnected. Leaving some out will not result in a working `lnd`. I tried to split off all independent changes already. Also, I want to fix tests in the same commit that breaks them, so those can't be moved out.

joostjager:
@Roasbeef @cfromknecht ptal

channeldb/invoices.go
Roasbeef: Oh yeah it's totally out of the scope, just a drive by comment.

channeldb/invoice_test.go
Roasbeef: Yeah there's a tradeoff w.r.t having them always compile vs having them make smaller changes at a time with an eye for ease of review. IMO things are more likely to slip by with larger commits that make modifications over several packages/files. Isolated commits also allow the committer to tell a narrative in a sense, with the progression of each commit. 

As it's already received some attention as is, I think it's too late at this point. However, as an example, the `AcceptOrSettleInvoice` change could've been added in a distinct commit that just left all the logic in place, but used a `nil` value for the function closure in the  link. A follow up commit could then remove the old verification logic and add the height based verification to that new closure.

joostjager: Ok, that is a good example indeed. Could have been a separate commit. Not sure about committing non-building code though.

contractcourt/htlc_incoming_contest_resolver.go
cfromknecht: @joostjager doesn't need to be done in this pr, that is a bigger change and currently we are keeping the existing behavior. 

> I am not sure how users will react to these stuck channels

i'm not sure either, though imo it'd be preferable to know that we are possibly leaving money on the table, especially since the preimage could come in shortly after the expiry. it's also possible that this whole flow can be delayed due to fee spikes, and in that case it seems like it'd be premature to give up on the HTLC, though i admit i haven't worked through all the edge cases

> accepting it would violate the invoice final cltv requirement

i'm not positive what the correct behavior is either. in the case of a regular invoice, we will already know the preimage and the resolver will convert itself to a success resolver and we no longer have this issue. if we have a hold invoice, "learning" the preimage after the expiry could only be done by the preimage being provided externally via rpc, so perhaps that constitutes some explicit consent that the user in fact wants to settle the invoice even if the expiry has expired. what do you think?

contractcourt/htlc_incoming_contest_resolver.go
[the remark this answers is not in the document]
cfromknecht: i suppose we can leave this as is then, as it's not incorrect and also isn't expensive to duplicate the check

lnd_multi-hop_htlc_local_chain_claim_test.go
```diff
@@ -0,0 +1,352 @@
+// +build rpctest
```
cfromknecht: imo it'd be preferable to start moving these all into the `lntest` package, tho i suspect there are a bunch of dependencies either in `lnd` or just helper functions that would break if we don't do the move all at once. is that the case?

lnd_multi-hop_htlc_remote_chain_claim_test.go
```diff
@@ -59,20 +77,33 @@ func testMultiHopHtlcRemoteChainClaim(net *lntest.NetworkHarness, t *harnessTest
...
 	}, time.Second*15)
 	if err != nil {
-		t.Fatalf("htlc mismatch: %v", err)
+		t.Fatalf("htlc mismatch: %v", predErr)
+	}
+
+	// Wait for carol to mark invoice as accepted. There is a small gap to
+	// bridge between adding the htlc to the channel and executing the exit
+	// hop logic.
+	for {
```
cfromknecht: this block might make a good helper function, since it's used 3+ times

contractcourt/htlc_incoming_contest_resolver.go
joostjager: >this whole flow can be delayed due to fee spikes, and in that case it seems like it'd be premature to give up on the HTLC

Maybe, but in the other hand, no one is losing money. The htlc is canceled back and we never marked it as settled, so no pizza has been delivered yet.

>if we have a hold invoice, "learning" the preimage after the expiry could only be done by the preimage being provided externally via rpc, so perhaps that constitutes some explicit consent that the user in fact wants to settle the invoice even if the expiry has expired.

It is about accepting the invoice. For hodl invoices, the preimage isn't needed to accept. If we would accept, the user should check the height when the `accepted` event comes because there may be fewer blocks left than expected.

Also for regular invoices, there is a risk in marking the invoice settled when there are not enough blocks left and we still need to wait for confirmation of a tx spending the commit output. We normally assume that we need the final cltv delta blocks to get those confirmed. This is also the criterium for settling invoices in the link. 

Previously we would only mark the invoice as settled when the success resolver completes. This is safer, but can also be an unnecessary delay in case there are plenty of blocks left. It will take longer for the pizza to be delivered because of a channel failure that the payer has nothing to do with.

contractcourt/htlc_incoming_contest_resolver.go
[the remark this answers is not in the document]
joostjager: Added a clearer comment explaining that it is a sanity check.

lnd_multi-hop_htlc_remote_chain_claim_test.go
joostjager: extracted

lnd_multi-hop_htlc_local_chain_claim_test.go
joostjager: It needs to be done all at once, commit added.

joostjager:
@cfromknecht ptal

PR #2916, opened 2019-04-08
discovery: make gossip replies synchronous

cfromknecht:
This PR modifies the `discovery.GossipSyncer` to send all replies and and gossip filter backlogs synchronously. This serves to throttle the rate at which the serving node writes messages to the wire, which has been observed to result to a disconnect if the remote node's discovery stream is filled up. Testing on my node has shown overall more stable connections, as this helps in more closely aligning our sending rate to the throughput of the other node. 

In the process, we make a pair of syncer's state machines totally isolated, which facilitates a nice removal of duplicate test code in `discovery/syncer_test.go` since the requesting side can now be tested independently of the remote peer's state machine.

See commit messages for more details.

### DIFF
--- a/discovery/syncer.go
+++ b/discovery/syncer.go
@@ -219,6 +219,16 @@ type gossipSyncerCfg struct {
 	// responding to gossip queries after replying to
 	// maxUndelayedQueryReplies queries.
 	delayedQueryReplyInterval time.Duration
+
+	// noSyncChannels will prevent the GossipSyncer from spawning a
+	// channelGraphSyncer, meaning we will not try to reconcile unknown
+	// channels with the remote peer.
+	noSyncChannels bool
+
+	// noReplyQueries will prevent the GossipSyncer from spawning a
+	// replyHandler, meaning we will not reply to queries from our remote
+	// peer.
+	noReplyQueries bool
 }
 
 // GossipSyncer is a struct that handles synchronizing the channel graph state
@@ -271,9 +281,14 @@ type GossipSyncer struct {
 	// PassiveSync to ActiveSync.
 	genHistoricalChanRangeQuery bool
 
-	// gossipMsgs is a channel that all messages from the target peer will
-	// be sent over.
+	// gossipMsgs is a channel that all responses to our queries from the
+	// target peer will be sent over, these will be read by the
+	// channelGraphSyncer.
 	gossipMsgs chan lnwire.Message
+
+	// queryMsgs is a channel that all queries from the remote peer will be
+	// received over, these will be read by the replyHandler.
+	queryMsgs chan lnwire.Message
 
 	// bufferedChanRangeReplies is used in the waitingQueryChanReply to
 	// buffer all the chunked response to our query.
@@ -332,6 +347,7 @@ func newGossipSyncer(cfg gossipSyncerCfg) *GossipSyncer {
 		syncTransitionReqs: make(chan *syncTransitionReq),
 		historicalSyncReqs: make(chan *historicalSyncReq),
 		gossipMsgs:         make(chan lnwire.Message, 100),
+		queryMsgs:          make(chan lnwire.Message, 100),
 		quit:               make(chan struct{}),
 	}
 }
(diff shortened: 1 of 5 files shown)


discovery/syncer.go
```diff
@@ -327,8 +343,17 @@ func (g *GossipSyncer) Start() {
 	g.started.Do(func() {
 		log.Debugf("Starting GossipSyncer(%x)", g.cfg.peerPub[:])
 
-		g.wg.Add(1)
-		go g.channelGraphSyncer()
+		// TODO(conner): only spawn channelGraphSyncer if remote
```
Roasbeef: Won't these two new variables handle this? Or is this referring to handling it at the server/p2p level?

discovery/syncer.go
```diff
@@ -894,23 +903,22 @@ func (g *GossipSyncer) replyShortChanIDs(query *lnwire.QueryShortChanIDs) error
...
-	if len(replyMsgs) == 0 {
-		return g.cfg.sendToPeer(&lnwire.ReplyShortChanIDsEnd{
-			ChainHash: query.ChainHash,
-			Complete:  1,
-		})
+	// Reply with any messages related to those channel ID's, we'll write
+	// each one individually and synchronously to throttle the sends and
+	// perform buffering of responses in the syncer as opposed to the peer.
+	for _, msg := range replyMsgs {
+		err := g.cfg.sendToPeerSync(msg)
```
Roasbeef: This should help with some of the write timeout issues as well since it should move more or less at the pace of their processing.

discovery/syncer.go
cfromknecht: it's possible to do so with the new variables, but this TODO is referring to hooking them up to the negotiated feature bits

discovery/syncer.go
cfromknecht: that's the hope! there's still a case which cannot be handled by this, which the write+flush fix solves completely. mainly that even if these sends are throttled, it can still get jammed if the remote peer is slow to process after reading and we try to send one more message. with this, we are less likely to hit that case, but it will go away completely with #2924

discovery/syncer_test.go
```diff
@@ -1180,216 +1143,148 @@ func TestGossipSyncerDelayDOS(t *testing.T) {
...
-				_, ok := msg.(*lnwire.QueryShortChanIDs)
-				if !ok {
-					t.Fatalf("wrong message: expected "+
-						"QueryShortChanIDs for %T", msg)
-				}
-
-				select {
-				case <-time.After(time.Second * 2):
-					t.Fatalf("node 2 didn't read msg")
+		expDelayResponse := i >= numUndelayedQueries
```
wpaulino: Don't know if this was the case already before these changes, but this test is actually delaying three queries -- you can confirm this if you enable logging. The reason for this is because the first query (QueryChannelRange) is also being taken into consideration in the rate limiter because all the queries occur within the same `delayedQueryInterval`.

discovery/syncer.go
```diff
@@ -183,11 +184,16 @@ type gossipSyncerCfg struct {
 	// encoding type that we can fit into a single message safely.
 	chunkSize int32
 
-	// sendToPeer is a function closure that should send the set of
-	// targeted messages to the peer we've been assigned to sync the graph
-	// state from.
+	// sendToPeer sends a variadic number of messages to the remote peer.
```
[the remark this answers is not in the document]
cfromknecht: initially i thought it might be nice to keep the async version for use in our own `channelGraphSyncer`, since it would be less blocking and more available to receive commands from the the `SyncManager`. it also seemed a more minimal change, though i think it's worth exploring making them all synchronous, especially since we already handle the errors on most of the call sites

discovery/syncer_test.go
cfromknecht: looked into this, and yes it was a preexisting issue. basically there was an off by one in the calculated parameters at the beginning, which would cause an extra query at the end. see the last commit

cfromknecht: one bonus is that is test's execution is now 33% shorter :)

cfromknecht:
@Roasbeef @wpaulino ptal

Roasbeef:
Needs a rebase now that #2932 has landed.

wpaulino:
Need another rebase 😬

cfromknecht:
@wpaulino rebased

PR #2918, opened 2019-04-09
htlcswitch: ensure commit tx gets updated after settle hodl invoice

joostjager:
This PR is a temporary fix for hodl invoices not being settled correctly.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -945,6 +945,13 @@ out:
 			break out
 		}
 
+		// If the previous event resulted in a non-empty
+		// batch, reinstate the batch ticker so that it can be
+		// cleared.
+		if l.batchCounter > 0 {
+			l.cfg.BatchTicker.Resume()
+		}
+
 		select {
 		// Our update fee timer has fired, so we'll check the network
 		// fee to see if we should adjust our commitment fee.
@@ -1047,13 +1054,6 @@ out:
 
 			l.handleDownStreamPkt(packet, true)
 
-			// If the downstream packet resulted in a non-empty
-			// batch, reinstate the batch ticker so that it can be
-			// cleared.
-			if l.batchCounter > 0 {
-				l.cfg.BatchTicker.Resume()
-			}
-
 		// A message from the switch was just received. This indicates
 		// that the link is an intermediate hop in a multi-hop HTLC
 		// circuit.
@@ -1075,13 +1075,6 @@ out:
 			}
 
 			l.handleDownStreamPkt(pkt, false)
-
-			// If the downstream packet resulted in a non-empty
-			// batch, reinstate the batch ticker so that it can be
-			// cleared.
-			if l.batchCounter > 0 {
-				l.cfg.BatchTicker.Resume()
-			}
 
 		// A message from the connected peer was just received. This
 		// indicates that we have a new incoming HTLC, either directly
@@ -1197,6 +1190,8 @@ func (l *channelLink) processHodlEvent(hodlEvent invoices.HodlEvent,
 		if err := hodlAction(htlc); err != nil {
 			return err
 		}
+
+		l.batchCounter++
 	}
 
 	return nil
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -5790,3 +5790,141 @@ func TestChannelLinkHoldInvoiceCancel(t *testing.T) {
 		t.Fatal("timeout")
 	}
 }
+
+// TestChannelLinkRevocationWindowHodl asserts that htlcs paying to a hodl
+// invoice are settled even if the revocation window gets exhausted.
+func TestChannelLinkRevocationWindowHodl(t *testing.T) {
+	t.Parallel()
+
+	const (
+		chanAmt = btcutil.SatoshiPerBitcoin * 5
+	)
+
+	// We'll start by creating a new link with our chanAmt (5 BTC). We will
+	// only be testing Alice's behavior, so the reference to Bob's channel
+	// state is unnecessary.
+	aliceLink, bobChannel, batchTicker, start, cleanUp, _, err :=
+		newSingleLinkTestHarness(chanAmt, 0)
+	if err != nil {
+		t.Fatalf("unable to create link: %v", err)
+	}
+	defer cleanUp()
+
+	if err := start(); err != nil {
+		t.Fatalf("unable to start test harness: %v", err)
+	}
+
+	var (
+		coreLink  = aliceLink.(*channelLink)
+		registry  = coreLink.cfg.Registry.(*mockInvoiceRegistry)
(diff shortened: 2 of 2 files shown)


joostjager:
Question that I have about this is whether the value of `batchCounter` cannot be retrieved from the channel, instead of keeping a separate counter for it.

joostjager:
Also investigate why `logTicker` didnt trigger

halseth:
Needs a testcase to catch the bug being fixed.

cfromknecht:
@joostjager good catch, previously we only used the batch counter to signal the presence of downstream htlcs from the switch, though i think this is the correct fix. prior i believe the assumption is that the log commit ticker would handle this, though it seems that may not always be true and this in fact is more accurate. should be ready to go with an accompanying unit test as @halseth suggests

halseth:
Is the log commit ticker even necessary at this point?

joostjager:
@cfromknecht I started writing unit tests and the case with regular invoices was still failing despite the initial `batchCounter` fix. So I went on at made more modifications to pass both tests.

joostjager:
Moved larger change into new PR #2927.

joostjager:
>Also investigate why logTicker didnt trigger
Answering my own question: logTicker may trigger, but if the rev window is then still not open, there is no retry.

joostjager:
>Question that I have about this is whether the value of batchCounter can be retrieved from the channel, instead of keeping a separate counter for it.

This idea has been implemented in #2927

joostjager:
The reason I am no fan of using `batchCounter` for fixing the hodl settle problem is that it retries updating the commit tx in a loop until it succeeds, while the real trigger for retrying should be the revokeAndAck because there the window is reopened.

htlcswitch/link_test.go
```diff
@@ -5790,3 +5790,141 @@ func TestChannelLinkHoldInvoiceCancel(t *testing.T) {
...
+	registry.SettleHodlInvoice(preimage2)
+
+	// Expect alice to send a settle for htlc 2.
+	receiveSettleAliceToBob(t, aliceMsgs, aliceLink, bobChannel)
+
+	// At this point, Alice cannot send a new commit sig to bob because the
+	// revocation window is exhausted.
+
+	// Sleep to let timer(s) expire.
+	time.Sleep(time.Second)
```
halseth: can we force tick the timers instead?

htlcswitch/link_test.go
```diff
@@ -5790,3 +5790,141 @@ func TestChannelLinkHoldInvoiceCancel(t *testing.T) {
...
+	// Allow some time for it to be processed by the link.
+	time.Sleep(time.Second)
+
+	// Trigger the batch timer as this may trigger Alice to send a commit
+	// sig.
+	batchTicker <- time.Time{}
+
+	// After the revocation, it is again possible for Alice to send a commit
+	// sig no more htlcs. Bob acks the update.
+	receiveCommitSigAliceToBob(t, aliceMsgs, aliceLink, bobChannel, 0)
```
halseth: so this is where it would fail previously ?

htlcswitch/link_test.go
```diff
@@ -5790,3 +5790,141 @@ func TestChannelLinkHoldInvoiceCancel(t *testing.T) {
...
+
+	// Bob sends revocation and signs commit with htlc 1 settled.
+	sendRevAndAckBobToAlice(t, aliceLink, bobChannel)
+
+	// Allow some time for it to be processed by the link.
+	time.Sleep(time.Second)
+
+	// Trigger the batch timer as this may trigger Alice to send a commit
+	// sig.
+	batchTicker <- time.Time{}
```
halseth: This is interesting, as we force tick the timer here, but we have no guarantees it is actually not paused. So it could still hide a bug where it would never tick in production.

Maybe we can add functionality to the mock ticker, such that we can trigger a tick only if active?

htlcswitch/link.go
```diff
@@ -1197,6 +1190,8 @@ func (l *channelLink) processHodlEvent(hodlEvent invoices.HodlEvent,
 		if err := hodlAction(htlc); err != nil {
 			return err
 		}
+
+		l.batchCounter++
```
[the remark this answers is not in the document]
joostjager: In `handleDownStreamPkt` calls to the switch are made. Hodl invoices never reach the switch.

htlcswitch/link_test.go
```diff
@@ -5790,3 +5790,141 @@ func TestChannelLinkHoldInvoiceCancel(t *testing.T) {
...
+	// Lock in htlc 2 on both sides.
+	sendHtlcBobToAlice(t, aliceLink, bobChannel, htlc2)
+	sendCommitSigBobToAlice(t, aliceLink, bobChannel, 2)
+	receiveRevAndAckAliceToBob(t, aliceMsgs, aliceLink, bobChannel)
+	receiveCommitSigAliceToBob(t, aliceMsgs, aliceLink, bobChannel, 2)
+	sendRevAndAckBobToAlice(t, aliceLink, bobChannel)
+
+	select {
+	case <-registry.settleChan:
+	case <-time.After(15 * time.Second):
```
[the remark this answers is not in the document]
joostjager: This is the timeout used in other calls that are used in this test. Like `receiveRevAndAckAliceToBob`. 

In general, I don't like the extra lines of timeout handling code added for every receive. I prefer using a helper like `invoices.timeout`. It is a test level timeout which also works fine for relatively short tests. And I also get the stack traces of all goroutines. It is also easier to change the timeout to a long value if I am running the test in a debugger.

htlcswitch/link_test.go
joostjager: The log commit timer isn't currently exposed. It means more changes to the link to do that and I wanted to keep changes to the link minimal from a risk perspective.

Also if we indeed remove the log commit timer in a follow up pr, the sleep or force tick isn't needed anymore in the test.

htlcswitch/link_test.go
joostjager: Yes, Alice would do nothing previously.

htlcswitch/link_test.go
joostjager: I have been looking at this before. I wouldn't want to change that in this PR.

I am not so sure that a normal ticker with a Force channel is the best way to do it. I think I'd prefer a purely forced ticker, not to also get some automatic ticks in the test. It makes it less deterministic. I can imagine having only the force ticks is sufficient for most tests and if something periodic, non-deterministic is still needed, it may be better to create it explicitly in that unit test.

halseth: Yes, I'm suggesting adding a metthod `TickIfActive` to the mock ticker, which will be a noop if it is paused.

joostjager: I got that. I just don't want to 'improve' the current mock ticker because I think this hybrid idea is not what we should be using in the future.

PR #2921, opened 2019-04-10
channeldb/graph: skip unknown edges in FetchChanInfos

cfromknecht:
This commit modifies FetchChanInfos to skip any channels that are not in
the graph at the time of the call. Currently the entire call will fail
if the edge is not found, which stalls a gossip sync in the following
scenario:

 1. Remote peer queries for a channel range
 2. We return the set of channel ids in that range
 3. A channel from that set is removed from the graph, e.g. via close.
 4. Remote peer queries for removed edge, causing the query to fail.

To remedy this, we will now skip any edges that are not known in the
database at the time of the query. This prevents the syncer state
machines from halting, which otherwise could only be resolved by
disconnecting and reconnecting.

### DIFF
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -3065,36 +3065,6 @@ func (c *ChannelGraph) NewChannelEdgePolicy() *ChannelEdgePolicy {
 	return &ChannelEdgePolicy{db: c.db}
 }
 
-// MarkEdgeZombie marks an edge as a zombie within the graph's zombie index.
-// The public keys should represent the node public keys of the two parties
-// involved in the edge.
-func (c *ChannelGraph) MarkEdgeZombie(chanID uint64, pubKey1,
-	pubKey2 [33]byte) error {
-
-	c.cacheMu.Lock()
-	defer c.cacheMu.Unlock()
-
-	err := c.db.Update(func(tx *bbolt.Tx) error {
-		edges := tx.Bucket(edgeBucket)
-		if edges == nil {
-			return ErrGraphNoEdgesFound
-		}
-		zombieIndex, err := edges.CreateBucketIfNotExists(zombieBucket)
-		if err != nil {
-			return err
-		}
-		return markEdgeZombie(zombieIndex, chanID, pubKey1, pubKey2)
-	})
-	if err != nil {
-		return err
-	}
-
-	c.rejectCache.remove(chanID)
-	c.chanCache.remove(chanID)
-
-	return nil
-}
-
 // markEdgeZombie marks an edge as a zombie within our zombie index. The public
 // keys should represent the node public keys of the two parties involved in the
 // edge.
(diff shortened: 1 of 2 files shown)


channeldb/graph.go
```diff
@@ -1664,11 +1666,16 @@ func (c *ChannelGraph) FetchChanInfos(chanIDs []uint64) ([]ChannelEdge, error) {
...
-			// First, we'll fetch the static edge information.
+			// First, we'll fetch the static edge information. If
+			// the edge is unknown, we will skip the edge and
+			// continue gathering all known edges.
 			edgeInfo, err := fetchChanEdgeInfo(
 				edgeIndex, cidBytes[:],
 			)
-			if err != nil {
+			switch {
+			case err == ErrEdgeNotFound:
```
wpaulino: Also check `ErrZombieEdge`?

cfromknecht: `fetchChanEdgeInfo` does not return `ErrZombieEdge`

cfromknecht: extended test to assert that queries for zombies don't have any impact

Roasbeef:
Test failure:
```
--- FAIL: TestGraphZombieIndex (0.01s)
    graph_test.go:2890: expected pubKey1 02a8bf2c03bdd588a988476d8d78fb3460b53a187f0d2454b39eb69e388170eca3, got 02a33752f7e5f0608f14ff61580de8a022dceba202e2cc3ffae9402fd30ad7ee04
```

Roasbeef:
Also shows on the other build:
```
--- FAIL: TestGraphZombieIndex (0.03s)
    graph_test.go:2890: expected pubKey1 029c0b08d76743584ea5fe561a167e0b8256486db1e6a25577bcb543adf3ca52ef, got 0220df24d52878cf459d700d5b2509524dc3c910df449ab4336d310e1cc61f9185
```

cfromknecht:
@Roasbeef @wpaulino ptal, unit tests have been fixed

channeldb/graph_test.go
```diff
@@ -2846,7 +2863,17 @@ func TestGraphZombieIndex(t *testing.T) {
...
 	}
+
+	// Swap the nodes if the second's pubkey is smaller than the first.
+	// Without this, the comparisons at the end will fail probabilistically.
+	if bytes.Compare(node2.PubKeyBytes[:], node1.PubKeyBytes[:]) < 0 {
+		node1, node2 = node2, node1
+	}
+
 	edge, _, _ := createChannelEdge(db, node1, node2)
+	if err := graph.AddChannelEdge(edge); err != nil {
```
Roasbeef: The check below assumes that the graph doesn't yet know of this edge. So why does adding it to the graph not cause the test to fail?

cfromknecht: the comment is now incorrect. since the edge is known, it shouldn't be zombie

PR #2923, opened 2019-04-10
multi: peer related SCB fixes 

Roasbeef:
In this PR, we fix a few issues that would cause the itests for the new SCB system to fail randomly based on timing:
  1. Ensure we send the chan sync message in a synchronous manner to the remote peer. Otherwise, if we fail early, then we'll exit w/o ever sending the message to the peer. 
  2. Don't exit during the restoration process if we're already connected to the remote peer.


Roasbeef:
PTAL @cfromknecht

PR #2925, opened 2019-04-10
[trivial] rpcserver: demote newaddress log to debug, log addr type

halseth:


cfromknecht:
Should we demote all rpc command logging to debug?

halseth:
I think it is better to change the current only needed, but can be conscious about it when adding new logs.

PR #2926, opened 2019-04-10
routing: tracking visited nodes in findPath

Crypt-iQ:
These commits add a `visitedNodes` map that tracks if the Djikstra's algorithm in `findPath` has already visited this node before.  Previously, when evaluating the best node, we would look at all of the incoming edges and repeat the `processEdge` flow for already visited nodes.  The second commit enables the `TestNewRoutePathTooLong` test by fixing the excessive hops graph to correctly be parsed.

### DIFF
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -236,6 +236,28 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
 	})
 }
 
+// ForEachNodeChannel iterates through all channels of a given node, executing the
+// passed callback with an edge info structure and the policies of each end
+// of the channel. The first edge policy is the outgoing edge *to* the
+// the connecting node, while the second is the incoming edge *from* the
+// connecting node. If the callback returns an error, then the iteration is
+// halted with the error propagated back up to the caller.
+//
+// Unknown policies are passed into the callback as nil values.
+//
+// If the caller wishes to re-use an existing boltdb transaction, then it
+// should be passed as the first argument.  Otherwise the first argument should
+// be nil and a fresh transaction will be created to execute the graph
+// traversal.
+func (c *ChannelGraph) ForEachNodeChannel(tx *bbolt.Tx, nodePub []byte,
+	cb func(*bbolt.Tx, *ChannelEdgeInfo, *ChannelEdgePolicy,
+		*ChannelEdgePolicy) error) error {
+
+	db := c.db
+
+	return nodeTraversal(tx, nodePub, db, cb)
+}
+
 // ForEachNode iterates through all the stored vertices/nodes in the graph,
 // executing the passed callback with each node encountered. If the callback
 // returns an error, then the transaction is aborted and the iteration stops
@@ -2183,23 +2205,10 @@ func (c *ChannelGraph) HasLightningNode(nodePub [33]byte) (time.Time, bool, erro
 	return updateTime, exists, nil
 }
 
-// ForEachChannel iterates through all channels of this node, executing the
-// passed callback with an edge info structure and the policies of each end
-// of the channel. The first edge policy is the outgoing edge *to* the
-// the connecting node, while the second is the incoming edge *from* the
-// connecting node. If the callback returns an error, then the iteration is
-// halted with the error propagated back up to the caller.
-//
-// Unknown policies are passed into the callback as nil values.
-//
-// If the caller wishes to re-use an existing boltdb transaction, then it
-// should be passed as the first argument.  Otherwise the first argument should
-// be nil and a fresh transaction will be created to execute the graph
-// traversal.
-func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
+// nodeTraversal is used to traverse all channels of a node given by its
+// public key and passes channel information into the specified callback.
+func nodeTraversal(tx *bbolt.Tx, nodePub []byte, db *DB,
 	cb func(*bbolt.Tx, *ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy) error) error {
-
-	nodePub := l.PubKeyBytes[:]
 
 	traversal := func(tx *bbolt.Tx) error {
 		nodes := tx.Bucket(nodeBucket)
@@ -2241,7 +2250,7 @@ func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
 			if err != nil {
 				return err
 			}
-			edgeInfo.db = l.db
+			edgeInfo.db = db
 
 			outgoingPolicy, err := fetchChanEdgePolicy(
 				edges, chanID, nodePub, nodes,
@@ -2256,7 +2265,7 @@ func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
 			}
 
 			incomingPolicy, err := fetchChanEdgePolicy(
-				edges, chanID, otherNode, nodes,
+				edges, chanID, otherNode[:], nodes,
 			)
 			if err != nil {
 				return err
@@ -2275,12 +2284,34 @@ func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
 	// If no transaction was provided, then we'll create a new transaction
 	// to execute the transaction within.
 	if tx == nil {
-		return l.db.View(traversal)
+		return db.View(traversal)
 	}
 
 	// Otherwise, we re-use the existing transaction to execute the graph
 	// traversal.
 	return traversal(tx)
+}
+
+// ForEachChannel iterates through all channels of this node, executing the
+// passed callback with an edge info structure and the policies of each end
+// of the channel. The first edge policy is the outgoing edge *to* the
+// the connecting node, while the second is the incoming edge *from* the
+// connecting node. If the callback returns an error, then the iteration is
+// halted with the error propagated back up to the caller.
+//
+// Unknown policies are passed into the callback as nil values.
+//
+// If the caller wishes to re-use an existing boltdb transaction, then it
+// should be passed as the first argument.  Otherwise the first argument should
+// be nil and a fresh transaction will be created to execute the graph
+// traversal.
+func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
+	cb func(*bbolt.Tx, *ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy) error) error {
+
+	nodePub := l.PubKeyBytes[:]
+	db := l.db
+
+	return nodeTraversal(tx, nodePub, db, cb)
 }
 
 // ChannelEdgeInfo represents a fully authenticated channel along with all its
@@ -2450,15 +2481,15 @@ func (c *ChannelEdgeInfo) BitcoinKey2() (*btcec.PublicKey, error) {
 // OtherNodeKeyBytes returns the node key bytes of the other end of
 // the channel.
 func (c *ChannelEdgeInfo) OtherNodeKeyBytes(thisNodeKey []byte) (
-	[]byte, error) {
+	[33]byte, error) {
 
 	switch {
 	case bytes.Equal(c.NodeKey1Bytes[:], thisNodeKey):
-		return c.NodeKey2Bytes[:], nil
+		return c.NodeKey2Bytes, nil
 	case bytes.Equal(c.NodeKey2Bytes[:], thisNodeKey):
-		return c.NodeKey1Bytes[:], nil
+		return c.NodeKey1Bytes, nil
 	default:
-		return nil, fmt.Errorf("node not participating in this channel")
+		return [33]byte{}, fmt.Errorf("node not participating in this channel")
 	}
 }
 
--- a/routing/heap.go
+++ b/routing/heap.go
@@ -3,6 +3,7 @@ package routing
 import (
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
 )
 
 // nodeWithDist is a helper struct that couples the distance from the current
@@ -12,9 +13,9 @@ type nodeWithDist struct {
 	// current context.
 	dist int64
 
-	// node is the vertex itself. This pointer can be used to explore all
-	// the outgoing edges (channels) emanating from a node.
-	node *channeldb.LightningNode
+	// node is the vertex itself. This can be used to explore all the
+	// outgoing edges (channels) emanating from a node.
+	node route.Vertex
 
 	// amountToReceive is the amount that should be received by this node.
 	// Either as final payment to the final node or as an intermediate
--- a/routing/heap.go
+++ b/routing/heap.go
@@ -1,6 +1,8 @@
 package routing
 
 import (
+	"container/heap"
+
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing/route"
@@ -40,6 +42,21 @@ type nodeWithDist struct {
 // algorithm to keep track of the "closest" node to our source node.
 type distanceHeap struct {
 	nodes []nodeWithDist
+
+	// pubkeyIndices maps public keys of nodes to their respective index in
+	// the heap. This is used as a way to avoid db lookups by using heap.Fix
+	// instead of having duplicate entries on the heap.
+	pubkeyIndices map[route.Vertex]int
+}
+
+// newDistanceHeap initializes a new distance heap. This is required because
+// we must initialize the pubkeyIndices map for path-finding optimizations.
+func newDistanceHeap() distanceHeap {
+	distHeap := distanceHeap{
+		pubkeyIndices: make(map[route.Vertex]int),
+	}
+
+	return distHeap
 }
 
 // Len returns the number of nodes in the priority queue.
@@ -60,13 +77,17 @@ func (d *distanceHeap) Less(i, j int) bool {
 // NOTE: This is part of the heap.Interface implementation.
 func (d *distanceHeap) Swap(i, j int) {
 	d.nodes[i], d.nodes[j] = d.nodes[j], d.nodes[i]
+	d.pubkeyIndices[d.nodes[i].node] = i
+	d.pubkeyIndices[d.nodes[j].node] = j
 }
 
 // Push pushes the passed item onto the priority queue.
 //
 // NOTE: This is part of the heap.Interface implementation.
 func (d *distanceHeap) Push(x interface{}) {
-	d.nodes = append(d.nodes, x.(nodeWithDist))
+	n := x.(nodeWithDist)
+	d.nodes = append(d.nodes, n)
+	d.pubkeyIndices[n.node] = len(d.nodes) - 1
 }
 
 // Pop removes the highest priority item (according to Less) from the priority
@@ -77,7 +98,27 @@ func (d *distanceHeap) Pop() interface{} {
 	n := len(d.nodes)
 	x := d.nodes[n-1]
 	d.nodes = d.nodes[0 : n-1]
+	delete(d.pubkeyIndices, x.node)
 	return x
+}
+
+// PushOrFix attempts to adjust the position of a given node in the heap.
+// If the vertex already exists in the heap, then we must call heap.Fix to
+// modify its position and reorder the heap. If the vertex does not already
+// exist in the heap, then it is pushed onto the heap. Otherwise, we will end
+// up performing more db lookups on the same node in the pathfinding algorithm.
+func (d *distanceHeap) PushOrFix(dist nodeWithDist) {
+	index, ok := d.pubkeyIndices[dist.node]
+	if !ok {
+		heap.Push(d, dist)
+		return
+	}
+
+	// Change the value at the specified index.
+	d.nodes[index] = dist
+
+	// Call heap.Fix to reorder the heap.
+	heap.Fix(d, index)
 }
 
 // path represents an ordered set of edges which forms an available path from a
--- a/routing/heap_test.go
+++ b/routing/heap_test.go
@@ -6,7 +6,8 @@ import (
 	"reflect"
 	"sort"
 	"testing"
-	"time"
+
+	"github.com/lightningnetwork/lnd/routing/route"
 )
 
 // TestHeapOrdering ensures that the items inserted into the heap are properly
@@ -16,20 +17,33 @@ func TestHeapOrdering(t *testing.T) {
 
 	// First, create a blank heap, we'll use this to push on randomly
 	// generated items.
-	var nodeHeap distanceHeap
+	nodeHeap := newDistanceHeap()
 
-	prand.Seed(time.Now().Unix())
+	prand.Seed(1)
 
 	// Create 100 random entries adding them to the heap created above, but
 	// also a list that we'll sort with the entries.
 	const numEntries = 100
 	sortedEntries := make([]nodeWithDist, 0, numEntries)
 	for i := 0; i < numEntries; i++ {
+		var pubKey [33]byte
+		prand.Read(pubKey[:])
+
 		entry := nodeWithDist{
+			node: route.Vertex(pubKey),
 			dist: prand.Int63(),
 		}
 
-		heap.Push(&nodeHeap, entry)
+		// Use the PushOrFix method for the initial push to test the scenario
+		// where entry doesn't exist on the heap.
+		nodeHeap.PushOrFix(entry)
+
+		// Re-generate this entry's dist field
+		entry.dist = prand.Int63()
+
+		// Reorder the heap with a PushOrFix call.
+		nodeHeap.PushOrFix(entry)
+
 		sortedEntries = append(sortedEntries, entry)
 	}
 
@@ -45,6 +59,13 @@ func TestHeapOrdering(t *testing.T) {
 	for nodeHeap.Len() != 0 {
 		e := heap.Pop(&nodeHeap).(nodeWithDist)
 		poppedEntries = append(poppedEntries, e)
+	}
+
+	// Assert that the pubkeyIndices map is empty after popping all of the
+	// items off of it.
+	if len(nodeHeap.pubkeyIndices) != 0 {
+		t.Fatalf("there are still %d pubkeys in the pubkeyIndices map",
+			len(nodeHeap.pubkeyIndices))
 	}
 
 	// Finally, ensure that the items popped from the heap and the items we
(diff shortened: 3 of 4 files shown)


halseth:
I think this will not have the desired effect, as we are not guaranteed that the first time we visit a node is the "shortest path" getting there. We must therefore evaluate all other incoming edges before we can consider the node properly "visited".

Crypt-iQ:
@halseth But we are popping the best entry from the heap so we should have the shortest path from source to this node, no?

halseth:
Hm, but aren't we actually attempting to visit the node one the _other end_ of the edge than the best node?

halseth:
Ah, okay I see what it achieves now. So it is an optimization to avoid the extra calculation before reaching the line 
```go
if tempDist >= distance[fromVertex].dist {
```
?

I think you might be right then, since we assume we have processed all edges from that node at that point. The only exception is maybe the additional eddges?

Crypt-iQ:
Yeah good catch, the `visitedNodes[pivot] = struct{}{}` should be below the additional edges call, mb.  And yeah since we've processed all the edges from the node and it was the best node, there's no need to examine it in later calls.

Crypt-iQ:
I suppose instead of having the `visitedNodes` map we can just add the visited nodes to `ignoredNodes` and it would achieve the same thing.

halseth:
Do you actually see any speedup by using this change?

Crypt-iQ:
I think I will need to create a test graph and test this out.  I think if the check is done before `edgeInfo.FetchOtherNode`, there should be some efficiency gain.  Like this:
```
var fromBytes [33]byte
switch {
case bytes.Equal(edgeInfo.NodeKey1Bytes[:], pivot[:]):
	fromBytes = edgeInfo.NodeKey2Bytes
case bytes.Equal(edgeInfo.NodeKey2Bytes[:], pivot[:]):
	fromBytes = edgeInfo.NodeKey1Bytes
default:
	return fmt.Errorf("node not participating in this channel")
}

if _, ok := visitedNodes[Vertex(fromBytes)]; ok {
	return nil
}

channelSource, err := edgeInfo.FetchOtherNode(
	tx, pivot[:],
)
if err != nil {
	return err
}
```

Crypt-iQ:
I was thinking about the findPath code today and I discovered that a particular node can be the "best node" in the heap multiple times.  Luckily, we will always evaluate the shortest distance to this node first and subsequent iterations can be ignored by the same `visitedNodes` map.  If we have a graph like this:
```
//		    ┌──────────┐250 ┌──────────┐
//		    │    D     │─── │    B     │
//                  └──────────┘    └──────────┘   
//	     100  /                       |     \  50    
//		 /                        |      \      
//  ┌──────────┐/                         |       ┌──────────┐
//  │     S    │                          | 30    │     T    │
//  └──────────┘\                         |       └──────────┘
//		 \                        |      / 
//	     100  \                       |     /  100
//		    ┌──────────┐250 ┌──────────┐   
//		    │    C     │─── │    A     │
//	            └──────────┘    └──────────┘
```
then the current findPath algorithm will start with T, evaluate its neighbors (A, B) and add them both to the heap.  However, it will add `A` to the heap with a distance of 100 (for example's sake... not an accurate edge weight per the edgeWeight func) and B with a distance of 50.  Then the best node is B so we take it.  It evaluates its neighbors (A, T, D) and adds `A` to the heap with a distance of 80 (since it's less than 100) and D to the heap with a distance of 300.  The next best node is A with a distance of 80, so it's taken.  After that, the next best node is also A but with a distance of 100.  We can get efficiency gains if we check the best node before iterating over all channels with `ForEachChannel` (if we haven't iterated over the best node, we would also check the neighbors to see if we've already iterated over any of them as well - keeping the changes mentioned in the previous comments).

routing/pathfind.go
```diff
@@ -743,6 +739,10 @@ func findPath(g *graphParams, r *RestrictParams, source, target Vertex,
 			processEdge(reverseEdge.sourceNode, reverseEdge.edge,
 				bandWidth, pivot)
 		}
+
+		// We have visited the node and explored all incoming edges so
+		// we add it to our visited map.
+		visitedNodes[pivot] = struct{}{}
```
joostjager: This could be moved here in the commit where `visitedNodes` is introduced.

routing/pathfind.go
```diff
@@ -684,6 +680,16 @@ func findPath(g *graphParams, r *RestrictParams, source, target Vertex,
 		// examine all the incoming edges (channels) from this node to
 		// further our graph traversal.
 		pivot := Vertex(bestNode.PubKeyBytes)
+
+		// But before that, we'll check that we are not visiting the same
+		// node again which can happen because of the way our pathfinding
+		// algorithm works. Fortunately, only the shortest path to this
+		// node will be evaluated and all subsequent iterations over
+		// this node can safely be ignored.
+		if _, ok := visitedNodes[pivot]; ok {
```
joostjager: I am also thinking about a different fix, where the node is actually replaced on the heap instead of added again. There already exists a `heap.Fix` function to reposition an item. Maybe a direction is to extend `distanceHeap` with another field that is a map to lookup the heap index by pubkey.

Crypt-iQ: Time complexity of `heap.Fix` is `O(log n)` as is `heap.Push`.  So I don't think we save any on time when replacing the node on the heap (we do save space).  Plus we would also still have the space requirement of having a map of heap indices.  Also it still wouldn't address the problem of re-iterating over previously visited nodes in the `bestNode.ForEachChannel` loop, so we would still need the `visitedNodes` map.  Thoughts?

Crypt-iQ: My previous comment was a little wrong. `heap.Fix` fixes having duplicate items on the heap and probably has a similar space constraint as the `visitedNodes` map.  The only problem I see with it is that the nodes in the `ForEachChannel` loop will be visited again.  Sorry about the inaccuracy.

routing/pathfind.go
Crypt-iQ: Fixed

routing/pathfind.go
joostjager: I think the problem with visiting again is that it needs that db lookup? Other than that it would be pretty fast, perhaps neglible to visit again?

This can also be fixed in a different way by just not doing the db lookup when fetching the other end of the channel. Pass in a `Vertex` without the associated db fields to `processEdge`. Then create a different `ForEachChannel` that just operates on a `tx` and a `Vertex`.

What do you think about that direction? It may also be an additional speed up if nodes are visited multiple times before ending up in the `visitedNodes` map.

routing/pathfind.go
```diff
@@ -700,6 +715,22 @@ func findPath(g *graphParams, r *RestrictParams, source, target Vertex,
 				)
 			}
 
+			var fromBytes [33]byte
```
halseth: can use `edgeInfo.OtherNodeKeyBytes()`?

routing/pathfind.go
```diff
@@ -506,6 +507,10 @@ func findPath(g *graphParams, r *RestrictParams, source, target Vertex,
 		ignoredNodes = make(map[Vertex]struct{})
 	}
 
+	// We use this map to track which nodes have already been visited so
```
halseth: I would have said something like "explored all possible incoming channels", since "visited" is ambiguous .

routing/pathfind_test.go
```diff
@@ -1203,7 +1203,7 @@ func TestNewRoute(t *testing.T) {
 }
 
 func TestNewRoutePathTooLong(t *testing.T) {
-	t.Skip()
+	t.Parallel()
```
halseth: This needs an explanation in the commit message, why this can be enabled now.

routing/pathfind.go
Crypt-iQ: Hey I need some clarification on this.  Are you suggesting adding a `ForEachChannel` to the `Vertex` which takes a `tx` + callback function which emulates the current `ForEachChannel` being used?  This requires a change to the heap to only store `Vertex` right?  This is good because then there's no db lookup required in the callback itself before `processEdge`.  Combining this with `heap.Fix`, we wouldn't really need a `visitedNodes` map since there would be no db call there.

joostjager: Yes that is the suggestion.

routing/pathfind_test.go
Crypt-iQ: Fixed

routing/pathfind.go
Crypt-iQ: The new commits implement this, I tested with the graph mentioned in a previous comment.  The `pivot` node isn't added to the heap again.

joostjager:
Ran a test with `QueryRoutes` in a loop. Performance improvement seems to be around 15%

channeldb/graph.go
```diff
@@ -236,6 +236,93 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+				edges, chanID, otherNode, nodes,
+			)
+			if err != nil {
+				return err
+			}
+
+			err = cb(&edgeInfo, outgoingPolicy, incomingPolicy)
+			if err != nil {
+				return err
+			}
```
joostjager: Can this code be reused from `LightningNode.ForEachChannel`? Possibly that function calling into here.

routing/heap.go
```diff
@@ -53,13 +59,21 @@ func (d *distanceHeap) Less(i, j int) bool {
...
+	d.pubkeyIndices[d.nodes[i].node] = i
+	d.pubkeyIndices[d.nodes[j].node] = j
 }
 
 // Push pushes the passed item onto the priority queue.
 //
 // NOTE: This is part of the heap.Interface implementation.
 func (d *distanceHeap) Push(x interface{}) {
-	d.nodes = append(d.nodes, x.(nodeWithDist))
+	if d.pubkeyIndices == nil {
```
joostjager: This check is now executed with every push while it could also go into a constructor?

routing/heap.go
```diff
@@ -53,13 +59,21 @@ func (d *distanceHeap) Less(i, j int) bool {
 // NOTE: This is part of the heap.Interface implementation.
 func (d *distanceHeap) Swap(i, j int) {
 	d.nodes[i], d.nodes[j] = d.nodes[j], d.nodes[i]
+	d.pubkeyIndices[d.nodes[i].node] = i
+	d.pubkeyIndices[d.nodes[j].node] = j
 }
```
joostjager: Would it be useful to add a basic test around this heap now that it does a bit more than what all the examples show?

routing/pathfind.go
```diff
@@ -483,14 +478,25 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
+		// If fromVertex exists in the heap, then we must call heap.Fix
+		// to adjust its position in the heap. Otherwise, we will end up
+		// performing more db lookups on the same node.
+		if index, ok := nodeHeap.pubkeyIndices[fromVertex]; ok {
+			// Change the value at the specified index.
+			nodeHeap.nodes[index] = distance[fromVertex]
+			// Call heap.Fix to reorder the heap.
+			heap.Fix(&nodeHeap, index)
+			return
+		}
```
joostjager: This could go into a method of `distanceHeap`?

routing/pathfind.go
```diff
@@ -507,21 +513,19 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
 			break
 		}
 
 		// Now that we've found the next potential step to take we'll
 		// examine all the incoming edges (channels) from this node to
 		// further our graph traversal.
-		pivot := route.Vertex(bestNode.PubKeyBytes)
-		err := bestNode.ForEachChannel(tx, func(tx *bbolt.Tx,
-			edgeInfo *channeldb.ChannelEdgeInfo,
+		err := g.graph.ForEachNodeChannel(pivot[:], tx, func(edgeInfo *channeldb.ChannelEdgeInfo,
```
joostjager: Consider splitting db lookup change and heap fix in separate commits.

channeldb/graph.go
Crypt-iQ: I think `ForEachNodeChannel` doesn't even need to be used.  If we were to create a `LightningNode` struct containing only the public key and then call `ForEachChannel`, I think it would still have the same effect? We don't need to add the `db` field either, since the `ChannelEdgeInfo` never uses it in this context.

routing/heap.go
Crypt-iQ: Will do

channeldb/graph.go
joostjager: It is a bit a matter of personal preference, but I normally try not to use structs in a way they were not intended to be used. If a function only accepts the parameters (including the struct fields it is a method of) it really requires, it is easier for others to understand its behavior. 

But in any case, copy/paste doesn't seem to be necessary.

Crypt-iQ: Gotcha, I will move the `traversal` outside of `ForEachChannel` and have it accept a boltdb transaction, public key bytes, and the db field.  That should dedupe the code well I think?

Crypt-iQ: I actually cannot do what I suggested above because of the traversal closure is used here:

https://github.com/lightningnetwork/lnd/blob/642fed53a3f0cefc750d79eaf0ae1d31d2217381/channeldb/graph.go#L2349-L2351

`View` takes a `func(*Tx) error)`, so the closure is actually necessary and can't really be moved out of the function.

routing/pathfind.go
Crypt-iQ: Done

routing/pathfind.go
Crypt-iQ: Done

routing/heap.go
Crypt-iQ: Sort of addressed, I don't think it's possible to compare `pubkeyIndices` maps between a mock heap (an array sorted by `sort.Slice`) and the actual heap.  Instead, I added a little code snippet that in `TestHeapOrdering` that validates that `pubkeyIndices` has the correct indices for `d.nodes`

channeldb/graph.go
Crypt-iQ: Done

channeldb/graph.go
```diff
@@ -236,6 +236,24 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+// ForEachNodeChannel iterates through all channels of a given node, executing the
+// passed callback with an edge info structure and the policies of each end
+// of the channel. The first edge policy is the outgoing edge *to* the
+// the connecting node, while the second is the incoming edge *from* the
+// connecting node. If the callback returns an error, then the iteration is
+// halted with the error propagated back up to the caller.
+//
+// Unknown policies are passed into the callback as nil values.
+//
+// The caller must provide an existing boltdb transaction.
```
joostjager: If not, a new tx is created still?

routing/heap_test.go
```diff
@@ -39,6 +45,16 @@ func TestHeapOrdering(t *testing.T) {
...
+	// Test that the pubkeyIndices map accurately tracks each node's index
+	// in the nodes array.
+	for i, n := range nodeHeap.nodes {
+		pkIndex := nodeHeap.pubkeyIndices[n.node]
+		if pkIndex != i {
+			t.Fatalf("Pubkey indices do not match. Pubkey index: %d,"+
+				" node index: %d", pkIndex, i)
+		}
+	}
+
```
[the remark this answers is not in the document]
Crypt-iQ: Lmk what you think about my test approach

routing/heap_test.go
```diff
@@ -36,6 +35,12 @@ func TestHeapOrdering(t *testing.T) {
 		}
 
 		heap.Push(&nodeHeap, entry)
+
+		// Re-generate this entry's dist field
+		entry.dist = prand.Int63()
+		// Reorder the heap with a PushOrFix call.
```
joostjager: Add newline

routing/heap_test.go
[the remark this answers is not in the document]
joostjager: Looks good. Maybe also use `PushOrFix` for the initial push, to test the non-exist scenario. `PushOrFix` is the only thing we call anyway from path finding. And possibly an assert to verify that the pubkey map is empty at the end of the test.

channeldb/graph.go
```diff
@@ -236,6 +236,24 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+func (c *ChannelGraph) ForEachNodeChannel(tx *bbolt.Tx, nodePub []byte,
+	cb func(*bbolt.Tx, *ChannelEdgeInfo, *ChannelEdgePolicy, *ChannelEdgePolicy) error) error {
```
halseth: long line

channeldb/graph.go
```diff
@@ -2275,14 +2280,36 @@ func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
...
+// connecting node. If the callback returns an error, then the iteration is
+// halted with the error propagated back up to the caller.
+//
+// Unknown policies are passed into the callback as nil values.
+//
+// If the caller wishes to re-use an existing boltdb transaction, then it
+// should be passed as the first argument.  Otherwise the first argument should
+// be nil and a fresh transaction will be created to execute the graph
+// traversal.
+func (l *LightningNode) ForEachChannel(tx *bbolt.Tx,
```
halseth: Did you explore whether we can remove this method now?

routing/pathfind.go
```diff
@@ -368,12 +366,9 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
 
 	// processEdge is a helper closure that will be used to make sure edges
 	// satisfy our specific requirements.
-	processEdge := func(fromNode *channeldb.LightningNode,
-		edge *channeldb.ChannelEdgePolicy,
+	processEdge := func(fromVertex route.Vertex, edge *channeldb.ChannelEdgePolicy,
```
halseth: nit: line a tiny bit too long?

routing/pathfind.go
```diff
@@ -548,21 +543,19 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
 			break
 		}
 
 		// Now that we've found the next potential step to take we'll
 		// examine all the incoming edges (channels) from this node to
 		// further our graph traversal.
-		pivot := route.Vertex(bestNode.PubKeyBytes)
-		err := bestNode.ForEachChannel(tx, func(tx *bbolt.Tx,
-			edgeInfo *channeldb.ChannelEdgeInfo,
+		err := g.graph.ForEachNodeChannel(tx, pivot[:], func(_ *bbolt.Tx, edgeInfo *channeldb.ChannelEdgeInfo,
```
halseth: long line

routing/pathfind.go
```diff
@@ -595,16 +588,17 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
...
-			channelSource, err := edgeInfo.FetchOtherNode(
-				tx, pivot[:],
-			)
+			channelSource, err := edgeInfo.OtherNodeKeyBytes(pivot[:])
 			if err != nil {
 				return err
 			}
 
+			var source [33]byte
+			copy(source[:], channelSource)
```
halseth: I would prefer just changing the return type of `OtherNodeKeyBytes` to `[33]byte` if possible (and no significant perf impact)

routing/heap.go
```diff
@@ -40,6 +42,21 @@ type nodeWithDist struct {
 // algorithm to keep track of the "closest" node to our source node.
 type distanceHeap struct {
 	nodes []nodeWithDist
+
+	// pubkeyIndices maps public keys of nodes to their respective index in
+	// the heap. This is used as a way to avoid db lookups by using heap.Fix
+	// instead of having duplicate entries on the heap.
+	pubkeyIndices map[route.Vertex]int
```
halseth: hmm... does this change have an actual performance impact? I would avoid adding this complexity if not.

routing/pathfind_test.go
halseth: So enabling this test is unrelated to the other changes in this PR?

routing/heap.go
Crypt-iQ: If `pubkeyIndices` isn't added, then if we have a duplicate entry to add to the heap (the second example I outlined way above), we can't call `heap.Fix` on it via `distanceHeap.PushOrFix`. If we get rid of that function, then the entry could potentially be the best entry in the heap and the `ForEachNodeChannel` loop is called on it.

channeldb/graph.go
Crypt-iQ: No, but from a cursory glance it seems that it can't be removed since not all callers to `ForEachChannel` will have a `*ChannelGraph` instance handy.

routing/pathfind.go
Crypt-iQ: I agree

routing/pathfind_test.go
Crypt-iQ: Right, it doesn't belong in this PR

routing/heap.go
halseth: Does avoiding duplicates lead to a significant performance gain?

Crypt-iQ: I think the performance gain is noticeable, but these are milliseconds after all.
I'm colorblind so the non-red points are with these changes and the red points are on master without these changes.  Roughly 25% improvement in pathfinding.  The graph was exported from testnet with about 2500 nodes and about 8000 channels.  Then found a path to each node on the graph by modifying the json unmarshalling in `pathfind_test.go`.  Joost noticed a 15% improvement on QueryRoutes, but I think that includes some other sources of noise so I opted for this testing approach.

pathfinding improvements

Crypt-iQ: Here the red is master, the purple is with 446e873 and no later commits, and the non-red non-purple is with all the changes in the branch (removing duplicates). So there are incremental performance improvements when combining both approaches (converting the `LightningNode` to `Vertex` and removing duplicates)

comparing with duplicates

Crypt-iQ: I tested out different source nodes and db sizes (minimal test db, medium 300mb, large 4gb) and concluded that both variables have the expected effect on time.  All the commits in the branch regularly outperform both master and commit 446e873.  The time gain varies depending on various conditions as well - heap memory usage etc but I found this to be the case across the board.  

Here in the first graph, the red is master and the non-red is with all the commits in the branch.  The master *mostly* stays within a band of 45ms - 90ms and seems to track both the 45ms and 90ms lines, std dev is high.  The non-red concentrates on a band between 30ms - 50ms with far lower std dev.  
2926-plot-1

In the second graph, the purple represents 446e873 and clusters on the bottom around 30ms and at the top around 80ms but with std dev than the non-red non-purple points.  The purple points above the 50ms band are the points that get improved upon when deleting duplicates.
446e873_visited-2926

It is hard to say specifically what the % improvement is over master and over not taking duplicates into account since there are many variables that can affect this.  However, I do think that this performance improvement is worth the change.

joostjager: So this PR contains two speed ups:
1. Remove `OtherNode` db lookup
2. Don't process nodes from the heap that we already processed before. This saves an unnecessary lookup of channel edges from the db.

For 2, two options were tried:
2a. Add a visited map
2b. Use `heap.Fix`

I think 2a and 2b have more or less the same effect on performance, but it seems 2b is more efficient in terms of memory usage.

I also prefer 2b because it stays closer to the general descriptions given of the Dijkstra algorithm.

routing/pathfind_test.go
Crypt-iQ: Removed the commit that added this test

routing/heap_test.go
Crypt-iQ: Done

channeldb/graph.go
Crypt-iQ: Looked at other callbacks in `graph.go` as a guide and none of them split lines, so I didn't know how to split this one.

routing/pathfind.go
Crypt-iQ: Fixed this by switching arguments to make the lines be under <= 80 chars

routing/pathfind.go
Crypt-iQ: Done, moved the callback above the `ForEachNodeChannel` call

routing/pathfind.go
Crypt-iQ: Done

channeldb/graph.go
joostjager:
```
func (c *ChannelGraph) ForEachNodeChannel(tx *bbolt.Tx, nodePub []byte,
	cb func(*bbolt.Tx, *ChannelEdgeInfo, *ChannelEdgePolicy,
		*ChannelEdgePolicy) error) error {
```

Crypt-iQ: Fixed

routing/heap_test.go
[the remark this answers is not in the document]
joostjager: @Crypt-iQ do you think this suggestion makes sense or do you think the test as is is sufficient?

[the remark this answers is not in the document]
Crypt-iQ: I think it makes sense since `Push` is never called anyways, updated the test

joostjager:
LGTM. You can squash the fixup commits.

Crypt-iQ:
@joostjager squashed

routing/heap.go
halseth: @Crypt-iQ Thanks for the detailed performance stats! 😄

routing/pathfind.go
```diff
@@ -595,18 +585,21 @@ func findPath(g *graphParams, r *RestrictParams, source, target route.Vertex,
 			// the node on the _other_ end of this channel as we
 			// may later need to iterate over the incoming edges of
 			// this node if we explore it further.
-			channelSource, err := edgeInfo.FetchOtherNode(
-				tx, pivot[:],
-			)
+			chanSource, err := edgeInfo.OtherNodeKeyBytes(pivot[:])
```
halseth: I think is where the optimization really happens, no? Commit message should be updated to explain that.

routing/heap.go
```diff
@@ -77,9 +98,28 @@ func (d *distanceHeap) Pop() interface{} {
...
+func (d *distanceHeap) PushOrFix(dist nodeWithDist) {
+	index, ok := d.pubkeyIndices[dist.node]
+	if !ok {
+		heap.Push(d, dist)
+		return
+	}
+
+	// Change the value at the specified index.
+	d.nodes[index] = dist
+	// Call heap.Fix to reorder the heap.
```
halseth: nit: newline above

routing/pathfind.go
Crypt-iQ: Updated - lmk what you think about the new one.  I think it's descriptive enough?

routing/heap.go
Crypt-iQ: fixed

routing/pathfind.go
halseth: Perfect! 👌

Issue #1927, opened 2018-09-17
Empty `commit_sig`


pm47:
### Background

This looks a lot like #1127. I don't know what is the version of `lnd` but considering the node first appeared a month ago I suppose it should have the fix, so maybe this is something different?

### Your environment

* version of `lnd`: unknown, node is `fairly.cheap` `03cb7983dc247f9f81a0fa2dfa3ce1c255365f7279c8dd143e086ca333df10e278`

### Steps to reproduce

```
 eclair               lnd
   |                   |
   |        ...        | 
   |                   |  <--- 4 cross-signed htlcs eclair->lnd
   |                   |
   |<----- ful-1 ------|
   |<------ sig -------|
   |<----- ful-2 ------|
   |<----- ful-3 ------|
   |<----- ful-4 ------|
   |------- rev ------>| <--- eclair acks ful-1
   |------- sig ------>| <--- eclair signs ful-1
   |<------ sig -------| <--- lnd signs ful-2 + full-3 + ful-4
   |------- rev ------>| <--- eclair acks ful-2 + full-3 + ful-4
   |------- sig ------>| <--- eclair signs ful-2 + full-3 + ful-4
   |<------ sig -------| <--- this signature doesn't include anything new
   |------- err ------>| <--- cannot sign without changes!
   |                   |
```

Roasbeef:
Not much we can go off here if we don't know which version the node was running on, as we've fixed issues similar to this in the past. Although, on the implementers call today, I think Rusty was saying that they found a version of lnd would still send them under stress tests? However, again there, we're not sure what version they were testing agaisnt.

Roasbeef:
It _does_ look like we should've just sent a revoke there at the end though...there're generally some corner cases in the state machine w.r.t concurrent updates which we may not implement identically right now.

PR #2927, opened 2019-04-10
htlcswitch: fix empty commit sig and no commit sig

joostjager:
Builds on top of #3564.

This PR fixes two known bugs in the commitment tx update sequence:
* Sending an commit sig message that didn't sign any new updates ("empty commit sig")
* Not sending a commit sig message when it is expected.

In addition to that, several simplifications are carried through where values were tracked in the link that can also be obtained from the channel.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -742,19 +742,15 @@ func (l *channelLink) resolveFwdPkgs() error {
 
 	l.log.Debugf("loaded %d fwd pks", len(fwdPkgs))
 
-	var needUpdate bool
 	for _, fwdPkg := range fwdPkgs {
-		hasUpdate, err := l.resolveFwdPkg(fwdPkg)
-		if err != nil {
+		if err := l.resolveFwdPkg(fwdPkg); err != nil {
 			return err
 		}
-
-		needUpdate = needUpdate || hasUpdate
 	}
 
 	// If any of our reprocessing steps require an update to the commitment
 	// txn, we initiate a state transition to capture all relevant changes.
-	if needUpdate {
+	if l.channel.PendingLocalUpdateCount() > 0 {
 		return l.updateCommitTx()
 	}
 
@@ -764,7 +760,7 @@ func (l *channelLink) resolveFwdPkgs() error {
 // resolveFwdPkg interprets the FwdState of the provided package, either
 // reprocesses any outstanding htlcs in the package, or performs garbage
 // collection on the package.
-func (l *channelLink) resolveFwdPkg(fwdPkg *channeldb.FwdPkg) (bool, error) {
+func (l *channelLink) resolveFwdPkg(fwdPkg *channeldb.FwdPkg) error {
 	// Remove any completed packages to clear up space.
 	if fwdPkg.State == channeldb.FwdStateCompleted {
 		l.log.Debugf("removing completed fwd pkg for height=%d",
@@ -774,7 +770,7 @@ func (l *channelLink) resolveFwdPkg(fwdPkg *channeldb.FwdPkg) (bool, error) {
 		if err != nil {
 			l.log.Errorf("unable to remove fwd pkg for height=%d: "+
 				"%v", fwdPkg.Height, err)
-			return false, err
+			return err
 		}
 	}
 
@@ -793,7 +789,7 @@ func (l *channelLink) resolveFwdPkg(fwdPkg *channeldb.FwdPkg) (bool, error) {
 		if err != nil {
 			l.log.Errorf("unable to process remote log updates: %v",
 				err)
-			return false, err
+			return err
 		}
 		l.processRemoteSettleFails(fwdPkg, settleFails)
 	}
@@ -802,7 +798,6 @@ func (l *channelLink) resolveFwdPkg(fwdPkg *channeldb.FwdPkg) (bool, error) {
 	// downstream logic is able to filter out any duplicates, but we must
 	// shove the entire, original set of adds down the pipeline so that the
 	// batch of adds presented to the sphinx router does not ever change.
-	var needUpdate bool
 	if !fwdPkg.AckFilter.IsFull() {
 		adds, err := lnwallet.PayDescsFromRemoteLogUpdates(
 			fwdPkg.Source, fwdPkg.Height, fwdPkg.Adds,
@@ -810,20 +805,20 @@ func (l *channelLink) resolveFwdPkg(fwdPkg *channeldb.FwdPkg) (bool, error) {
 		if err != nil {
 			l.log.Errorf("unable to process remote log updates: %v",
 				err)
-			return false, err
+			return err
 		}
-		needUpdate = l.processRemoteAdds(fwdPkg, adds)
+		l.processRemoteAdds(fwdPkg, adds)
 
 		// If the link failed during processing the adds, we must
 		// return to ensure we won't attempted to update the state
 		// further.
 		if l.failed {
-			return false, fmt.Errorf("link failed while " +
+			return fmt.Errorf("link failed while " +
 				"processing remote adds")
 		}
 	}
 
-	return needUpdate, nil
+	return nil
 }
 
 // fwdPkgGarbager periodically reads all forwarding packages from disk and
@@ -1873,7 +1868,7 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 		}
 
 		l.processRemoteSettleFails(fwdPkg, settleFails)
-		needUpdate := l.processRemoteAdds(fwdPkg, adds)
+		l.processRemoteAdds(fwdPkg, adds)
 
 		// If the link failed during processing the adds, we must
 		// return to ensure we won't attempted to update the state
@@ -1882,7 +1877,11 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 			return
 		}
 
-		if needUpdate {
+		// If there are pending local updates, try to update the commit
+		// tx. Pending updates could already have been present because
+		// of a previously failed update to the commit tx or freshly
+		// added by processRemoteAdds.
+		if l.channel.PendingLocalUpdateCount() > 0 {
 			if err := l.updateCommitTx(); err != nil {
 				l.fail(LinkFailureError{code: ErrInternalError},
 					"unable to update commitment: %v", err)
@@ -2532,7 +2531,7 @@ func (l *channelLink) processRemoteSettleFails(fwdPkg *channeldb.FwdPkg,
 // whether we are reprocessing as a result of a failure or restart. Adds that
 // have already been acknowledged in the forwarding package will be ignored.
 func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
-	lockedInHtlcs []*lnwallet.PaymentDescriptor) bool {
+	lockedInHtlcs []*lnwallet.PaymentDescriptor) {
 
 	l.log.Tracef("processing %d remote adds for height %d",
 		len(lockedInHtlcs), fwdPkg.Height)
@@ -2571,13 +2570,10 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 	if sphinxErr != nil {
 		l.fail(LinkFailureError{code: ErrInternalError},
 			"unable to decode hop iterators: %v", sphinxErr)
-		return false
+		return
 	}
 
-	var (
-		needUpdate    bool
-		switchPackets []*htlcPacket
-	)
+	var switchPackets []*htlcPacket
 
 	for i, pd := range lockedInHtlcs {
 		idx := uint16(i)
@@ -2614,7 +2610,6 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 			// sender.
 			l.sendMalformedHTLCError(pd.HtlcIndex, failureCode,
 				onionBlob[:], pd.SourceRef)
-			needUpdate = true
 
 			l.log.Errorf("unable to decode onion hop "+
 				"iterator: %v", failureCode)
@@ -2633,7 +2628,6 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 			l.sendMalformedHTLCError(
 				pd.HtlcIndex, failureCode, onionBlob[:], pd.SourceRef,
 			)
-			needUpdate = true
 
 			l.log.Errorf("unable to decode onion "+
 				"obfuscator: %v", failureCode)
@@ -2664,7 +2658,6 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 				lnwire.NewInvalidOnionPayload(failedType, 0),
 				obfuscator, pd.SourceRef,
 			)
-			needUpdate = true
 
 			l.log.Errorf("unable to decode forwarding "+
 				"instructions: %v", err)
@@ -2675,7 +2668,7 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 
 		switch fwdInfo.NextHop {
 		case hop.Exit:
-			updated, err := l.processExitHop(
+			err := l.processExitHop(
 				pd, obfuscator, fwdInfo, heightNow, pld,
 			)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


halseth:
For reference: https://github.com/lightningnetwork/lnd/issues/1927

joostjager:
@halseth's repro of empty commitSig: #3012

htlcswitch/link.go
```diff
@@ -1079,35 +1059,11 @@ out:
 
 			break out
 
-		case <-l.logCommitTick:
```
cfromknecht: does this mean that in certain scenarios the commit dance will be faster?

htlcswitch/link.go
```diff
@@ -349,14 +340,6 @@ type channelLink struct {
 	// sub-systems with the latest set of active HTLC's on our channel.
 	htlcUpdates chan *contractcourt.ContractUpdate
 
-	// logCommitTimer is a timer which is sent upon if we go an interval
-	// without receiving/sending a commitment update. It's role is to
-	// ensure both chains converge to identical state in a timely manner.
-	//
-	// TODO(roasbeef): timer should be >> then RTT
-	logCommitTimer *time.Timer
```
cfromknecht: nice to see these go!

htlcswitch/link_isolated_test.go
```diff
@@ -0,0 +1,268 @@
...
+	"crypto/sha256"
+	"testing"
+	"time"
+
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+type linkTestContext struct {
```
cfromknecht: 😍

joostjager: Yes, every pr should have a small clean up component...

build/prefix_log.go
```diff
@@ -0,0 +1,104 @@
...
+// Debug formats message using the default formats for its operands and writes
+// to log with LevelDebug.
+func (p *PrefixLog) Debug(v ...interface{}) {
+	p.log.Debug(v...)
+}
+
+// Info formats message using the default formats for its operands and writes to
+// log with LevelInfo.
+func (p *PrefixLog) Info(v ...interface{}) {
+	p.log.Info(v...)
```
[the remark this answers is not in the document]
joostjager: no, forgotten. fixed.

lnwallet/channel.go
```diff
@@ -1416,6 +1422,7 @@ func NewLightningChannel(signer input.Signer,
 		Capacity:          state.Capacity,
 		LocalFundingKey:   state.LocalChanCfg.MultiSigKey.PubKey,
 		RemoteFundingKey:  state.RemoteChanCfg.MultiSigKey.PubKey,
+		log:               build.NewPrefixLog(logPrefix, walletLog),
```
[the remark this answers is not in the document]
joostjager: Hehe. Ok, I know that I sometimes comment on commit structure in PRs and suggest some of them to be merged. This is particular the case for structs that are introduced in a separate commit but it is hard to review the code without any further context. In this case the prefix logger is an object that doesn't need much context to review (although this is a sliding scale) and therefore I made it a separate commit.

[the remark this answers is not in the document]
joostjager: Added commits to also make the link logging consistent. Open question is what we think is the best prefix to use for link and channel? uint64, shortchan, chanpoint?

lnwallet/channel.go
```diff
@@ -4151,24 +4151,29 @@ func (lc *LightningChannel) ReceiveNewCommitment(commitSig lnwire.Sig,
...
+
+	localUpdatesPending := lc.localUpdateLog.logIndex !=
+		lastRemoteCommit.ourMessageIndex
 
-	localUpdatesSynced := (lastLocalCommit.ourMessageIndex ==
-		lastRemoteCommit.ourMessageIndex)
+	remoteUpdatesPending := lastLocalCommit.theirMessageIndex !=
+		lastRemoteCommit.theirMessageIndex
+
+	allSigned := !localUpdatesPending && !remoteUpdatesPending
```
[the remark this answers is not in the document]
joostjager: But it isn't duplicated. Here we check if we owe a signature, while in `ReceiveNewCommitment`, we check whether the remote party owed anything.

htlcswitch/link.go
joostjager: Yes indeed!

htlcswitch/link.go
joostjager: Yes!

htlcswitch/link_isolated_test.go
```diff
@@ -0,0 +1,253 @@
...
+	"crypto/sha256"
+	"testing"
+	"time"
+
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// sendHtlcBobToAlice sends an HTLC from Bob to Alice, that pays to a preimage
```
halseth: why this move?

htlcswitch/link.go
```diff
@@ -502,11 +501,11 @@ func (l *channelLink) Start() error {
 // NOTE: Part of the ChannelLink interface.
 func (l *channelLink) Stop() {
 	if !atomic.CompareAndSwapInt32(&l.shutdown, 0, 1) {
-		log.Warnf("channel link(%v): already stopped", l)
+		l.log.Warn("already stopped")
```
halseth: Looks like the logging changes can perfectly fine go in its own PR separate from the far more critical commit sig changes.

htlcswitch/link_test.go
```diff
@@ -4696,11 +4696,12 @@ func TestChannelLinkNoEmptySig(t *testing.T) {
 	ctx.receiveRevAndAckAliceToBob()
 	ctx.sendRevAndAckBobToAlice()
 
-	// The situation now is that Alice still doesn't have her two htlcs on
-	// the local commit tx. Bob needs to send a new signature and Alice can
-	// only wait for that. However, Alice's log commit timer fires and Alice
-	// sends a commitment tx containing no updates. THIS SHOULD NOT HAPPEN!
-	ctx.receiveCommitSigAliceToBob(2)
+	// Allow some time for the log commit ticker to trigger for Alice.
+	time.Sleep(time.Second)
```
halseth: can also add this `time` as an argument to `assertNoMsgFromAlice`, then we can select on it instead of sleeping here.

htlcswitch/link_isolated_test.go
joostjager: It is a preparation from creating a link test context. I wanted to keep the move to a different file isolated, to make it easy to see the real changes in the next commit.

htlcswitch/link.go
joostjager: Separate prs created and linked.

htlcswitch/link_test.go

lnwallet/channel.go
```diff
@@ -4152,24 +4152,29 @@ func (lc *LightningChannel) ReceiveNewCommitment(commitSig lnwire.Sig,
...
+// yet.
+func (lc *LightningChannel) AllUpdatesSigned() bool {
 	lc.RLock()
 	defer lc.RUnlock()
 
-	lastLocalCommit := lc.localCommitChain.tip()
 	lastRemoteCommit := lc.remoteCommitChain.tip()
+	lastLocalCommit := lc.localCommitChain.tip()
+
+	localUpdatesPending := lc.localUpdateLog.logIndex !=
```
halseth: Sprinkle some comments in here.

lnwallet/channel.go
```diff
@@ -4152,24 +4152,29 @@ func (lc *LightningChannel) ReceiveNewCommitment(commitSig lnwire.Sig,
...
-// FullySynced returns a boolean value reflecting if both commitment chains
-// (remote+local) are fully in sync. Both commitment chains are fully in sync
-// if the tip of each chain includes the latest committed changes from both
-// sides.
-func (lc *LightningChannel) FullySynced() bool {
+// AllUpdatesSigned returns a boolean value reflecting whether we need to send
+// out a commitment signature because there are outstanding local updates and/or
+// updates in the local commit tx that aren't reflected in the remote commit tx
+// yet.
+func (lc *LightningChannel) AllUpdatesSigned() bool {
```
halseth: Name `OweCommitment`?

htlcswitch/link.go
```diff
@@ -1009,11 +1009,14 @@ out:
 			break out
 		}
 
-		// If the previous event resulted in a non-empty
-		// batch, reinstate the batch ticker so that it can be
-		// cleared.
+		// If the previous event resulted in a non-empty batch, resume
+		// the batch ticker so that it can be cleared. Otherwise pause
+		// the ticker to prevent waking up the htlcManager while the
+		// batch is empty.
```
halseth: Should we be using a timer instead?

htlcswitch/link_test.go
```diff
@@ -5890,6 +5890,104 @@ func TestChannelLinkHoldInvoiceRestart(t *testing.T) {
...
+
+	// Allow some time for the log commit ticker to trigger for Alice.
+	time.Sleep(time.Second)
+
+	// Now that Bob revoked, Alice should send the sig she owes.
+	//
+	// THIS SHOULD NOT HAPPEN.
+	ctx.assertNoMsgFromAlice()
+
+	aliceLink.Stop()
```
halseth: nit: defer stop

lnwallet/channel.go
```diff
@@ -3989,6 +3989,20 @@ func (lc *LightningChannel) ReceiveNewCommitment(commitSig lnwire.Sig,
 	localACKedIndex := lc.remoteCommitChain.tail().ourMessageIndex
 	localHtlcIndex := lc.remoteCommitChain.tail().ourHtlcIndex
 
+	// Check for empty commit sig.
```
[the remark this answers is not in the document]
joostjager: added

lnwallet/channel.go
joostjager: renamed

htlcswitch/link.go
joostjager: It could be a timer, but also the ticker is currently used pretty heavily in the unit tests that cover this code. So we'd need something similar to `time.After` that also allows manual triggering. I don't really want to extend the scope of this pr to include that.

htlcswitch/link.go
```diff
@@ -1853,11 +1814,14 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
...
-		// added by processRemoteAdds.
-		if l.channel.PendingLocalUpdateCount() > 0 {
+		// The revocation window opened up. If there are pending local
+		// updates, try to update the commit tx. Pending updates could
+		// already have been present because of a previously failed
+		// update to the commit tx or freshly added in by
+		// processRemoteAdds. Also in case there are no local updates,
+		// but there are still remote updates that are not in the remote
+		// commit tx yet, send out an update.
+		if l.channel.OweCommitment() {
```
halseth: nit: should this change really be in a separate commit? Seems orthogonal to removing the ticker.

joostjager: It is in the same commit because the log commit ticker was 'fixing' the problem caused by just checking `l.channel.PendingLocalUpdateCount() > 0`.

halseth: Definitely not a blocker, was just curious.

htlcswitch/link.go
```diff
@@ -1014,11 +991,14 @@ out:
...
 
-		// If the previous event resulted in a non-empty
-		// batch, reinstate the batch ticker so that it can be
-		// cleared.
-		if l.batchCounter > 0 {
+		// If the previous event resulted in a non-empty batch, resume
+		// the batch ticker so that it can be cleared. Otherwise pause
+		// the ticker to prevent waking up the htlcManager while the
+		// batch is empty.
+		if l.channel.PendingLocalUpdateCount() > 0 {
```
cfromknecht: this is certainly correct, but does this need to be executed on every iteration? not a blocker

lnwallet/channel.go
```diff
@@ -4156,24 +4156,36 @@ func (lc *LightningChannel) ReceiveNewCommitment(commitSig lnwire.Sig,
...
-		lastRemoteCommit.ourMessageIndex)
+	// There are remote updates pending if their remote commitment tx (our
+	// local commitment tx) contains updates that we don't have added to our
+	// remote commitment tx yet.
+	remoteUpdatesPending := lastLocalCommit.theirMessageIndex !=
+		lastRemoteCommit.theirMessageIndex
+
+	// If any of the conditions above is true, we owe a commitment
+	// signature.
+	oweCommitment := localUpdatesPending || remoteUpdatesPending
```
cfromknecht: i see now, the local and remote PoV are reversed. is it possible to consolidate this with the empty commit sig check via a helper method, e.g. `oweCommitment(local bool) bool`?

joostjager:
PR is approved, but need to do interop testing with CL en Eclair

htlcswitch/link.go
joostjager: Maybe not, but it I put it there to be sure that the flow can never get stuck if we forget to set the timer somewhere. It may be a little defensive, but I think it is justified given the importance of this logic. I don't think there is significant performance impact of executing the check with every iteration.

lnwallet/channel.go
joostjager: consolidate in accordance with your suggestion

joostjager:
Same test for Eclair 0.3.2-5ad3944, no problems observed.

```
#!/bin/bash

while true
do
        LNCLI="$HOME/lightninglabs/lnd/lncli-debug --network=regtest"
        ECLAIRCLI="$HOME/Downloads/eclair-cli.sh -a localhost:8081 -p foobar"
        ID1=`od -vAn -N4 -tu4 < /dev/urandom`
        ID2=`od -vAn -N4 -tu4 < /dev/urandom`

        BOLT1=`$ECLAIRCLI createinvoice --amountMsat=10000 --description="test" | jq -r .serialized`
        BOLT2=`$LNCLI addinvoice 10 | jq -r .pay_req`

        $LNCLI --network=regtest payinvoice -f $BOLT1
        $ECLAIRCLI payinvoice --invoice=$BOLT2
done
```

joostjager:
Finally interop with `lnd 0.8`. Works too.

htlcswitch/link.go
cfromknecht: sounds good, agree with those comments

lnwallet/channel.go
cfromknecht: looks great, thank you!

joostjager:
Rebased

lnwallet/channel.go
```diff
@@ -4140,6 +4150,65 @@ func (lc *LightningChannel) ReceiveNewCommitment(commitSig lnwire.Sig,
 	return nil
 }
 
+// OweCommitment returns a boolean value reflecting whether we need to send
+// out a commitment signature because there are outstanding local updates and/or
+// updates in the local commit tx that aren't reflected in the remote commit tx
+// yet.
+func (lc *LightningChannel) OweCommitment(local bool) bool {
+	lc.RLock()
```
Roasbeef: Perhaps we should instead obtain the mutex once within `ReceiveNewCommitment`, then either make this a private method, or update the `godoc` to note that it expects the mutex to already be held?

htlcswitch/link.go
```diff
@@ -1109,19 +1112,10 @@ out:
 			}
 
 		case <-l.cfg.BatchTicker.Ticks():
-			// If the current batch is empty, then we have no work
```
Roasbeef: Nice, this should also lower idle CPU usage for routing nodes with tons of channels.

lnwallet/channel.go
joostjager: It is also called without a lock from `ChannelLink`. I now created a private no-lock version and a public method that wraps it in a lock.

I am wondering whether our channel manipulations are always safe though. Something like this for example:

```
if lc.OweCommitment(true) {
			commitSig, htlcSigs, _, err := lc.SignNextCommitment()
```

I think the channel isn't modified concurrently, but why it the lock there then? Is it a safety precaution? Or for reading channel state in another subsystem?

joostjager:
Test from #3012 added to this PR. It needed a bit of modification, but the idea of the scenario is the same.

For review, this is the delta with the previous revision of this pr:
https://github.com/joostjager/lnd/compare/hodl-drop-fix-better-ref..joostjager:hodl-drop-fix-better

lnwallet/channel.go
Roasbeef: The lock was mostly a safety precaution when the struct was first created just to ensure things were goroutine safe, so all other code added afterwards followed that trend. In the past there were also instances of grabbing live state machines from the peer, and using that to populate RPC information.

PR #2830, opened 2019-03-24
docs: update contribution checklist

philippgille:
- Update Go version 1.11 to 1.12
- Fix broken link in GitHub PR template
- Fix inconsistent number of spaces after checkbox
- Add proper link to "Ideal Git Commit structure"

#### Pull Request Checklist

- [x] If this is your first time contributing, we recommend you read the [Code
  Contribution Guidelines](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md)
- [x] All changes are Go version 1.11 compliant
- [x]  The code being submitted is commented according to the
  [Code Documentation and Commenting](#CodeDocumentation) section
- [x]  For new code: Code is accompanied by tests which exercise both
  the positive and negative (error paths) conditions (if applicable)
- [x]  For bug fixes: Code is accompanied by new tests which trigger
  the bug being fixed to prevent regressions
- [x]  Any new logging statements use an appropriate subsystem and
  logging level
- [x]  Code has been formatted with `go fmt`
- [x]  For code and documentation: lines are wrapped at 80 characters
  (the tab character should be counted as 8 characters, not 4, as some IDEs do
  per default)
- [x]  Running `make check` does not fail any tests
- [x]  Running `go vet` does not report any issues
- [x]  Running `make lint` does not report any **new** issues that did not
  already exist
- [x] All commits build properly and pass tests. Only in exceptional
  cases it can be justifiable to violate this condition. In that case, the
  reason should be stated in the commit message.
- [x] Commits have a logical structure ([see section 4.5, of the Code Contribution Guidelines])

PR #2928, opened 2019-04-10
Update Go version from 1.11 to 1.12

philippgille:
This is a follow-up PR to #2830, which already updated the Go version in the contribution checklist.

Pinging @cfromknecht, who mentioned in [this PR comment](https://github.com/lightningnetwork/lnd/pull/2830#pullrequestreview-223480557) that it would be nice to also have the installation docs updated before the final 0.6 release.

A few things to note here:

1. I'm not sure about the macOS installation. On https://formulae.brew.sh/formula/go it's mentioned that the `go` package is "also known as `go@1.12`", but while https://formulae.brew.sh/formula/go@1.11 is a valid URL, https://formulae.brew.sh/formula/go@1.12 is not. So maybe someone with a Mac can test that `brew install go@1.12` works at all?
2. Similar to a Mac, I don't have a FreeBSD installation, so I can't test that either. I didn't change the docs for FreeBSD and when searching for a Go package with https://www.freebsd.org/cgi/ports.cgi?query=%5Ego&stype=name&sektion=lang there's only `go-1.12.2,1`. I'm not sure if `go` works as an alias for that package though.

Another option would be to not advise to use a package manager at all, and use https://dl.google.com/go/go1.12.3.darwin-amd64.pkg for macOS and https://dl.google.com/go/go1.12.3.freebsd-amd64.tar.gz for FreeBSD instead, or even just point to https://golang.org/doc/install for both.

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/docker/btcd/Dockerfile
+++ b/docker/btcd/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.11-alpine as builder
+FROM golang:1.12-alpine as builder
 
 MAINTAINER Olaoluwa Osuntokun <laolu@lightning.network>
 
--- a/docker/lnd/Dockerfile
+++ b/docker/lnd/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.11-alpine as builder
+FROM golang:1.12-alpine as builder
 
 MAINTAINER Olaoluwa Osuntokun <lightning.engineering>
 
--- a/docker/ltcd/Dockerfile
+++ b/docker/ltcd/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.11-alpine as builder
+FROM golang:1.12-alpine as builder
 
 MAINTAINER Olaoluwa Osuntokun <lightning.engineering>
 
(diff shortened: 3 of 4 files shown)


molxyz:
I've seen more and more Windows users showing up asking questions about LND. So I'm thinking maybe it's time to share the bits and pieces of what I know how to run Go on windows.  Also, the last time when I downloaded Go1.12.3 it triggered a trojan alarm (which I concluded as a false positive), you can see my issue here:  https://github.com/golang/go/issues/30969 . But because of this I learned how to compile Go on Windows 10 which to me is extremely easy, also helps not to worry about viruses. 

So I've written this guide how to compile Go, for whoever that would like to try it out. And it's up to you the dev team if you'd like to include it in the LND doc:   

https://privatebin.net/?7bcf76e4881c6db6#Rnon/QaZFe1jKzqDRw4PRrbwRErIHJenISTWBl564L0=

Thanks for all your hard work!  :)

PR #2932, opened 2019-04-11
discovery: SyncManager improvements

wpaulino:
In this PR, we make some improvements to the [recently introduced `SyncManager`](#2740). These improvements include:

* Active syncers will no longer attempt to synchronize our graph with remote peers. Instead, we'll now rely on synchronizing our graph with remote peers through the routine historical syncs performed by the `SyncManager`. Since active syncers will no longer attempt this synchronization, the `SyncManager`'s round-robin is no longer needed. 

* Handling initial historical sync disconnections: Every time `lnd` starts up, it attempts an initial historical graph sync with the first peer that connects. If the peer ends up disconnecting, then we wouldn't handle finding a replacement to continue performing the initial historical sync.

* Queueing active syncers until the initial historical sync completes: We do this to ensure we can properly handle any new channel updates at tip. This is required for fresh nodes that are syncing the channel graph for the first time. If we begin accepting updates at tip while the initial historical sync is still ongoing, then we risk not processing certain updates since we've yet to learn of the channels themselves.

**NOTE**: To test, start up a fresh mainnet node and attempt to sync the graph. Bumping up `DISC` logs to `debug` will serve useful. You should not see any of your peers in an `ActiveSync` state until the initial historical sync has completed. You can also disconnect the peer we're performing the initial historical sync with to ensure we find a replacement (assuming you have other peers connected).


discovery/syncer.go
```diff
@@ -18,14 +18,21 @@ import (
...
-	// behavior. This includes reconciling the set of missing graph updates
-	// with the remote peer _and_ receiving new updates from them.
+	// ActiveSync denotes that a gossip syncer:
+	//
+	//   1. Should not attempt to synchronize with the remote peer for
+	//   missing channels.
+	//   2. Should respond to queries from the remote peer.
+	//   3. Should receive new updates from the remote peer.
+	//
+	// They are started in a chansSynced state in order to accomplish their
```
halseth: can we remove `chansSynced` now?

discovery/sync_manager.go
```diff
@@ -205,297 +186,301 @@ func (m *SyncManager) syncerHandler() {
...
+
+	// If it's a non-active syncer, then we can just exit now.
+	if _, ok := m.inactiveSyncers[peer]; ok {
+		delete(m.inactiveSyncers, peer)
+		return
+	}
+
+	// Otherwise, we'll need find a new one to replace it, if any.
+	delete(m.activeSyncers, peer)
+	newActiveSyncer := chooseRandomSyncer(
```
halseth: not used?

discovery/sync_manager.go
```diff
@@ -205,297 +186,301 @@ func (m *SyncManager) syncerHandler() {
...
-		for current != nil {
-			// Ensure we properly handle a shutdown signal.
-			select {
-			case <-m.quit:
-				return
-			default:
+			// We'll force a historical sync with the first peer we
+			// connect to, to ensure we get as much of the graph as
+			// possible.
+			if !attemptInitialHistoricalSync {
```
halseth: looks like we can get rid of `pendingActiveSyncers` now. We can always add new syncers to the `inactiveSyncers` map, and let the syncer handler choose syncers from here after the historical sync is done :)

discovery/syncer.go
wpaulino: What do you mean by remove? Just remove that part of the comment?

discovery/sync_manager.go
wpaulino: It is. If you look at `chooseRandomSyncer` you can provide a callback to perform a certain action on each syncer. So by the time this returns, it either returns `nil` because it couldn't perform the action on any syncers, or it returns the first syncer with which we successfully performed the action.

discovery/sync_manager.go
wpaulino: Good point! Fixed.

discovery/sync_manager.go
halseth: Ah, so it is only used to print following debug log?

discovery/syncer.go
halseth: Was just wondering whether the `chansSynced` variable is possible to remove, if they always start in this state now.

discovery/sync_manager.go
wpaulino: Correct.

discovery/syncer.go
wpaulino: Don't think so as this is the only state in which we can process sync transitions.

discovery/sync_manager.go
```diff
@@ -518,86 +503,60 @@ func (m *SyncManager) transitionPassiveSyncer(s *GossipSyncer) error {
...
-			"GossipSyncer(%x): %v", s.cfg.peerPub, err)
-
-		s = m.chooseRandomSyncer(candidatesChosen, true)
+func (m *SyncManager) forceHistoricalSync() *GossipSyncer {
+	m.syncersMu.Lock()
+	defer m.syncersMu.Unlock()
+
+	// We'll sample from both sets of active and inactive syncers in the
+	// event that we don't have any inactive syncers.
+	numCandidates := len(m.inactiveSyncers) + len(m.activeSyncers)
```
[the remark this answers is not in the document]
wpaulino: Good catch, done.

routing/router.go
```diff
@@ -1242,7 +1242,7 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 			return err
 		}
 
-		log.Tracef("New channel update applied: %v",
+		log.Debugf("New channel update applied: %v",
```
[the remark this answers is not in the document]
wpaulino: Was part of me debugging some test flakes. Has been removed in the current version.

discovery/sync_manager.go
```diff
@@ -108,6 +103,18 @@ type SyncManager struct {
...
+	// GossipSyncers for newly connected peers.
+	newSyncers chan *newSyncer
+
+	// staleSyncers is a channel we'll use to process requests to tear down
+	// GossipSyncers for disconnected peers.
+	staleSyncers chan *staleSyncer
+
+	// syncersMu guards the read and write access to the activeSyncers and
+	// inactiveSyncers maps below.
+	syncersMu sync.Mutex
```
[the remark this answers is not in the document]
wpaulino: Correct. I went with the mutex to avoid the additional internal messages within the event loop, but I'm not opposed to adding them in order to remove it.

discovery/sync_manager.go
```diff
@@ -426,72 +411,48 @@ func (m *SyncManager) forceHistoricalSync() {
...
+	numCandidates := len(m.inactiveSyncers) + len(m.activeSyncers)
+	candidates := make(map[routing.Vertex]*GossipSyncer, numCandidates)
+	for _, s := range m.inactiveSyncers {
+		candidates[s.cfg.peerPub] = s
+	}
+	for _, s := range m.activeSyncers {
+		candidates[s.cfg.peerPub] = s
 	}
+
+	_ = chooseRandomSyncer(candidates, func(s *GossipSyncer) error {
```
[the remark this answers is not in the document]
wpaulino: It's not used in that commit, but it is later on.

discovery/sync_manager.go
```diff
@@ -340,113 +505,53 @@ func (m *SyncManager) chooseRandomSyncer(blacklist map[routing.Vertex]struct{},
...
+	done := make(chan struct{})
+
+	// We avoid returning an error when the SyncManager is stopped since the
+	// GossipSyncer will be stopped then anyway.
+	select {
+	case m.staleSyncers <- &staleSyncer{
+		peer:     peer,
+		doneChan: done,
+	}:
+	case <-m.quit:
```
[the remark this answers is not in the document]
wpaulino: Done.

discovery/sync_manager.go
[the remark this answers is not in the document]
cfromknecht: i'm fine as is

PR #2945, opened 2019-04-12
routing: disable route cache for non-default requests

joostjager:
The PR attempts to fix the following problem reported by a user:

>Hi, I'm trying the new queryroutes command and I'm using it via grpc (python). I'm seeing weird time-dependent behavior. Specifically I'm trying to get routes from queryroutes, try the route and after failure putting the failed channel on the blacklist. However, if I request then another path applying the channel blacklist, I'm getting back the same route as before. With the same code, only by putting a sleep statement,  sometimes I get a different path (but not always), where the blacklisted channel was excluded. Is there some sort of caching active? Thank you!

This fix doesn't address the changing bandwidth hints which may still cause routes to be returned of which we could already know that they'd fail.

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -1365,18 +1365,30 @@ func (r *ChannelRouter) FindRoutes(source, target Vertex,
 	// cache
 	//
 	// TODO: Route cache should store all request parameters instead of just
-	// amt and target. Currently false positives are returned if just the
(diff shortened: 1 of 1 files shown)


joostjager:
I think it is better to remove the route cache alltogether: #2950

PR #2955, opened 2019-04-15
cnct: fix on-chain hodl invoice handling and simplify incoming resolver flow

joostjager:
One of the first things the incoming contest resolver does is checking if the preimage is available and if it is, convert itself into a success resolver.

This behaviour makes it unnecessary to already determine earlier in the process whether an incoming contest or a success resolver is needed.

By having all incoming htlcs go through the incoming contest resolver, the number of execution paths is reduced and it becomes easier to ascertain that the implemented logic is correct.


joostjager:
Integrated in #2913

PR #1529, opened 2018-07-10
tests: extend integration tests to grab snapshot of active goroutines on failure

taketa:
Closes: https://github.com/lightningnetwork/lnd/issues/1219
Short description: upon test failures, scan through all active nodes, hit their profile page, write it to disk using a similar naming scheme as to the regular set of logs.

PR #2957, opened 2019-04-15
[lnd test] write goroutine snapshot on integration test failure

### LINKED ISSUE #1219: tests: extend integration tests to grab snapshot of active goroutines on failure
Currently, if we encounter deadlocks when modifying or adding new integration tests, it's a bit difficult to pinpoint exactly what the flaw in the new test/code is. All `lnd` instances running in the test now also [expose the profile port which allows us to obtain valuable debug information](https://github.com/lightningnetwork/lnd/blob/master/lntest/node.go#L262). It would be really useful if upon test failure, the framework hit the profile page over HTTP (http://localhost:port/debug/pprof/goroutine?debug=1) for each active node in the harness, and then write that out on disk in an appropriately name filed. 

# Steps To Completion 

- [ ] Upon test failures, scan through all active nodes and hit their profile page (`/debug/profile/goroutine`) in order to obtain a snapshot of the set of active goroutines. 

- [ ] Once the goroutine snapshot has been obtained, a new file should be written to disk for each node using a similar naming scheme as to the one we use currently for the regular set of logs.

halseth:
Fixup of https://github.com/lightningnetwork/lnd/pull/1529

Fixes #1219
Replaces #1529

### DIFF
--- a/lntest/harness.go
+++ b/lntest/harness.go
@@ -652,6 +652,11 @@ func (n *NetworkHarness) StopNode(node *HarnessNode) error {
 // SaveProfilesPages hits profiles pages of all active nodes and writes it to
 // disk using a similar naming scheme as to the regular set of logs.
 func (n *NetworkHarness) SaveProfilesPages() {
+	// Only write gorutine dumps if flag is active.
+	if !(*goroutineDump) {
+		return
+	}
+
 	for _, node := range n.activeNodes {
 		if err := saveProfilesPage(node); err != nil {
 			fmt.Println(err)
--- a/lntest/harness.go
+++ b/lntest/harness.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


lntest/node.go
```diff
@@ -266,6 +268,8 @@ func newNode(cfg nodeConfig) (*HarnessNode, error) {
 	cfg.P2PPort, cfg.RPCPort, cfg.RESTPort = generateListeningPorts()
 
 	nodeNum := numActiveNodes
+	cfg.ProfilePort = 9000 + nodeNum
```
wpaulino: Nit: include in `generateListeningPorts`?

lntest/harness.go
```diff
@@ -648,6 +649,57 @@ func (n *NetworkHarness) StopNode(node *HarnessNode) error {
...
+	for _, node := range n.activeNodes {
+		resp, err := http.Get(
+			fmt.Sprintf(
+				"http://localhost:%d/debug/pprof/goroutine?debug=1",
+				node.cfg.ProfilePort,
+			),
+		)
+		if err != nil {
+			fmt.Printf("Failed to get profile page: "+
+				"NodeId is %d; Node name is %s. "+
```
wpaulino: Nit: would make this `id=%d, name=%s, pubkey=%x`

Roasbeef:
Needs rebase.

halseth:
Rebased.

make/testing_flags.mk
```diff
@@ -65,4 +65,4 @@ else
 ITEST_TAGS += btcd
 endif
 
-ITEST := rm lntest/itest/output*.log; date; $(GOTEST) ./lntest/itest -tags="$(ITEST_TAGS)" $(TEST_FLAGS) -logoutput
+ITEST := rm lntest/itest/output*.log; rm lntest/itest/pprof*.log; date; $(GOTEST) ./lntest/itest -tags="$(ITEST_TAGS)" $(TEST_FLAGS) -logoutput -goroutinedump
```
wpaulino: I think we'll also want to upload the goroutine dumps along with the logs in our travis builds.

lntest/harness.go
cfromknecht: could also apply to the other error messages in this method

lntest/harness.go
```diff
@@ -648,6 +649,57 @@ func (n *NetworkHarness) StopNode(node *HarnessNode) error {
 	return node.stop()
 }
 
+// SaveProfilesPages hits profiles pages of all active nodes and writes it to
+// disk using a similar naming scheme as to the regular set of logs.
+func (n *NetworkHarness) SaveProfilesPages() {
+	for _, node := range n.activeNodes {
+		resp, err := http.Get(
```
cfromknecht: not blocking, but might help with indentation to define a `saveProfilePage` helper

make/testing_flags.mk
halseth: Since they end with `*.log` they will :)

lntest/harness.go
halseth: Done

lntest/harness.go
halseth: Done

lntest/node.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/2957/commits/98712a1fc88ac8de4bf0ceab54c3da44a26d2f2a

PR #2959, opened 2019-04-15
[trivial] routing/router: define ErrRouterShuttingDown

halseth:


routing/router.go
```diff
@@ -44,6 +44,10 @@ var (
 	// construct a new sphinx packet, but provides an empty set of hops for
 	// each route.
 	ErrNoRouteHopsProvided = fmt.Errorf("empty route hops provided")
+
+	// ErrRouterShuttingDown is returned if the router is in the process of
+	// shutting down.
+	ErrRouterShuttingDown = fmt.Errorf("router shutting down")
```
halseth: Reminder: make part of `routing.Error` type?

joostjager: If we merge this PR, how is this comment going to remind you?

halseth: Don't merge just yet 😛

halseth:
Not ready for merge.

routing/router.go
halseth: Did a sweep of the `fundingmgr` and only `routing.ErrOutdated` and `routing.ErrIgnored` is ignored there, so this should be safe to merge.

PR #2967, opened 2019-04-17
[trivial] multi: fix various typos in comments and docs

chokoboko:


cfromknecht:
@chokoboko most of the corrections look pretty good to me. the `iff -> if` modifications should be rolled back tho, since `iff != if`

PR #2973, opened 2019-04-18
routerrpc: async payment rpc

joostjager:
This PR adds a new `SendPayment` payment call to the `routerrpc` sub server. Notable changes:

* Configurable payment timeout (this is not to "cancel" htlcs in flight, which is impossible, but controls what the point is after which a new payment attempt isn't started anymore)

* Fee limit can only be specified as an absolute value. There is no default anymore. Specifying a value of zero means that no fees may be paid. In that case, only a payment to a direct peer would succeed.

* Only payment level failures are reported. This is currently limited to `timeout` and `no_routes`, but will be extended in a follow up pr #3156. The goal is to report terminal failures like `payment_hash_unknown`, but not to expose the result of the last payment _attempt_ anymore. 

In addition to that, a new rpc call `TrackPayment` is added. Using `TrackPayment` an in-flight payment can be picked up after the rpc connection was lost on SendPayment. When #2762 is merged, this also allows picking up payments after a restart of `lnd`.


Roasbeef:
Shouldn't this be branched off of #2761?

lnrpc/routerrpc/router_backend.go
```diff
@@ -29,7 +29,7 @@ type RouterBackend struct {
 	// routes.
 	FindRoutes func(source, target routing.Vertex,
 		amt lnwire.MilliSatoshi, restrictions *routing.RestrictParams,
-		numPaths uint32, finalExpiry ...uint16) (
+		numPaths uint32, finalExpiry ...int32) (
```
Roasbeef: Any strong reason to move to signed integers here? Virtually none of these fields are permitted to be negative. Historically, we've also used `uint16` for time lock deltas as we typically don't need anything more than a few weeks into the future.

routing/router.go
```diff
@@ -1494,10 +1494,8 @@ type LightningPayment struct {
 
 	// FinalCLTVDelta is the CTLV expiry delta to use for the _final_ hop
 	// in the route. This means that the final hop will have a CLTV delta
-	// of at least: currentHeight + FinalCLTVDelta. If this value is
-	// unspecified, then a default value of DefaultFinalCLTVDelta will be
-	// used.
-	FinalCLTVDelta *int32
+	// of at least: currentHeight + FinalCLTVDelta.
+	FinalCLTVDelta int32
```
Roasbeef: 👍

lnrpc/routerrpc/router.proto
```diff
@@ -2,42 +2,59 @@ syntax = "proto3";
...
+option go_package = "github.com/lightningnetwork/lnd/lnrpc/routerrpc";
+
 message PaymentRequest {
-    /**
-    A serialized BOLT-11 payment request that contains all information
-    required to dispatch the payment. If the pay req is invalid, or expired,
-    an error will be returned.
-    */
-    string pay_req = 1;
+    /// The identity pubkey of the payment recipient
```
Roasbeef: This information is already contained in the payment request, why do we need to duplicate it?

Roasbeef: There's no commit body message explaining rationale or motivation.

lnrpc/routerrpc/router_backend.go
```diff
@@ -236,3 +255,165 @@ func (r *RouterBackend) MarshallRoute(route *routing.Route) *lnrpc.Route {
 
 	return resp
 }
+
+// ExtractIntentFromSendRequest attempts to parse the SendRequest details
+// required to dispatch a client from the information presented by an RPC
+// client.
+func (r *RouterBackend) ExtractIntentFromSendRequest(rpcPayReq *PaymentRequest) (
```
Roasbeef: Seems to duplicate an existing method in the RPC server without removing the existing one. Also if this is to be done, then I think it can be done in a distinct commit.

cmd/lncli/commands.go
```diff
@@ -16,6 +16,8 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
```
Roasbeef: Lack of commit message body for such a significant change.

cmd/lncli/commands.go
```diff
@@ -2259,40 +2279,73 @@ func sendPaymentRequest(ctx *cli.Context, req *lnrpc.SendRequest) error {
...
+	req.FeeLimit = feeLimit
 
-	if err := paymentStream.Send(req); err != nil {
+	_, err = routerClient.SendPayment(context.Background(), req)
+	if err != nil {
 		return err
 	}
 
-	resp, err := paymentStream.Recv()
+	fmt.Printf("Payment in flight, track status by running:\n")
```
Roasbeef: I'm not sure we want to change the behavior of the CLI command over night like this. Instead we should add the RPC level change first, then follow up with the CLI changes. Otherwise, we may break many bash scripts that rely on `payinvoice` or `sendpayment` returning as soon as the payment is complete.

cmd/lncli/commands.go
```diff
@@ -2259,40 +2279,73 @@ func sendPaymentRequest(ctx *cli.Context, req *lnrpc.SendRequest) error {
...
-	// with a non-zero exit code.
-	if resp.PaymentError != "" {
-		return errors.New(resp.PaymentError)
+	stream, err := client.LookupPayment(context.Background(), req)
+	if err != nil {
+		return err
 	}
 
-	return nil
+	for {
```
Roasbeef: The help for the command should detail that this command may not return immediately. Also we may want to consider a mode where simply prints the latest state and exists early. In this case, something along the lines of a `--watch` flag would be added to allow the user to request on going notifications.

lnrpc/routerrpc/router_server.go
```diff
@@ -37,11 +40,15 @@ var (
...
 	macPermissions = map[string][]bakery.Op{
-		"/routerpc.Router/SendPayment": {{
+		"/routerrpc.Router/SendPayment": {{
 			Entity: "offchain",
 			Action: "write",
 		}},
-		"/routerpc.Router/EstimateRouteFee": {{
+		"/routerrpc.Router/LookupPayment": {{
+			Entity: "offchain",
+			Action: "write",
```
Roasbeef: Why does this require write permissions?

joostjager:
> Shouldn't this be branched off of #2761?

Yes, this is still wip. Should have added `[wip]` to the title.

lnrpc/routerrpc/router_backend.go
joostjager: Got bitten a few times now with negative numbers after subtraction (deltas can be negative) and want to prevent it from happening again to me or a future dev working with heights.

lnrpc/routerrpc/router.proto
joostjager: I should rename the message to `SendPaymentRequest`. It is not an encoded invoice.

lnrpc/routerrpc/router_backend.go
joostjager: I had a common method before, but then we introduced some differences that made not easily possible to reuse. The rpc request message is different. Fee limit is expressed differently and a custom timeout field is added.

cmd/lncli/commands.go
joostjager: This PR wasn't ready for review yet, commit messages not final.

cmd/lncli/commands.go
joostjager: If we don't want to break bash scripts, we'd better not change `lncli` at all. We are changing the payment result data as well. The string based "last error" isn't there anymore. No more reporting on payment attempt level outcomes.

lnrpc/routerrpc/router_backend.go
joostjager: In general, I think it is good to consistently use `int32` for heights and block deltas everywhere throughout the code base.

lnrpc/routerrpc/router_server.go
joostjager: Changed to read.

cmd/lncli/commands.go
joostjager: Not applicable if we leave `lncli unchanged`.

lnrpc/routerrpc/router.proto
joostjager: Renamed

cmd/lncli/commands.go
juscamarena: We're in beta, scripts will be updated. Most should probably be using the up to date payment rpc anyway.

Roasbeef:
Can be rebased now that https://github.com/lightningnetwork/lnd/pull/2761 is in.

joostjager:
rebase, ptal

routing/router.go
```diff
@@ -1560,10 +1560,8 @@ type LightningPayment struct {
 
 	// FinalCLTVDelta is the CTLV expiry delta to use for the _final_ hop
 	// in the route. This means that the final hop will have a CLTV delta
-	// of at least: currentHeight + FinalCLTVDelta. If this value is
-	// unspecified, then a default value of DefaultFinalCLTVDelta will be
-	// used.
-	FinalCLTVDelta *uint16
+	// of at least: currentHeight + FinalCLTVDelta.
+	FinalCLTVDelta int16
```
halseth: 👍 

Can we do this for the cltv limit also?

lnrpc/routerrpc/router.proto
```diff
@@ -7,41 +7,56 @@ package routerrpc;
...
 
     /**
-    An absolute limit on the cumulative CLTV value along the route for this
-    payment. Routes with total CLTV values higher than this will be ignored,
-    if there are no routes with a CLTV value below this amount, an error will
-    be returned.
+    A bare-bones invoice for a payment within the Lightning Network.  With the
+    details of the invoice, the sender has all the data necessary to send a
+    payment to the recipient. The amount in the payment request may be zero. In
+    that case it is required to set the amt field as well.
```
halseth: can it be an empty string?

lnrpc/routerrpc/router_backend.go
```diff
@@ -1,18 +1,27 @@
...
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing"
 	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/lightningnetwork/lnd/zpay32"
 	context "golang.org/x/net/context"
 )
 
+var (
+	zeroHash [32]byte
```
halseth: can use `lntypes.Hash{}`

halseth: Maybe we should even define a constant `lntypes.ZeroHash`? :)

lnrpc/routerrpc/router_backend.go
```diff
@@ -36,6 +45,16 @@ type RouterBackend struct {
 	FindRoute func(source, target route.Vertex,
 		amt lnwire.MilliSatoshi, restrictions *routing.RestrictParams,
 		finalExpiry ...uint16) (*route.Route, error)
+
+	// Activate the debug htlc mode. With the debug HTLC mode, all payments
+	// sent use a pre-determined payment hash.
+	DebugHTLC bool
```
halseth: is this necessary?

lnrpc/routerrpc/router_backend.go
```diff
@@ -321,3 +340,166 @@ func (r *RouterBackend) UnmarshallRoute(rpcroute *lnrpc.Route) (
...
+					"specified when paying a zero amount " +
+					"invoice")
+			}
+
+			payIntent.Amount = lnwire.NewMSatFromSatoshis(
+				btcutil.Amount(rpcPayReq.Amt),
+			)
+		} else {
+			if rpcPayReq.Amt != 0 {
+				return nil, errors.New("amount must not be " +
```
halseth: We can let user overpay if they want, but this seems safer indeed.

lnrpc/routerrpc/router_backend.go
```diff
@@ -321,3 +340,166 @@ func (r *RouterBackend) UnmarshallRoute(rpcroute *lnrpc.Route) (
...
+			r.MaxPaymentMSat)
+
+	}
+
+	return payIntent, nil
+}
+
+// validatePayReqExpiry checks if the passed payment request has expired. In
+// the case it has expired, an error will be returned.
+func validatePayReqExpiry(payReq *zpay32.Invoice) error {
```
halseth: export s.t. can be used by the rpcserver also?

channeldb/control_tower.go
```diff
@@ -92,6 +116,9 @@ func (p *paymentControl) InitPayment(paymentHash lntypes.Hash,
 	}
 	infoBytes := b.Bytes()
 
+	p.hashMutex.Lock(paymentHash)
```
halseth: should have a comment

channeldb/control_tower.go
```diff
@@ -389,6 +460,23 @@ func ensureInFlight(bucket *bbolt.Bucket) error {
...
+// fetchPaymentAttempt fetches the payment attempt from the bucket.
+func fetchPaymentAttempt(bucket *bbolt.Bucket) (*PaymentAttemptInfo, error) {
+	attemptData := bucket.Get(paymentAttemptInfoKey)
+	if attemptData == nil {
+		return nil, fmt.Errorf("unable to find attempt " +
+			"info for inflight payment")
+	}
+
+	r := bytes.NewReader(attemptData)
+	attempt, err := deserializePaymentAttemptInfo(r)
```
halseth: nit: can return directly

channeldb/control_tower.go
```diff
@@ -401,6 +489,131 @@ type InFlightPayment struct {
...
+		switch status {
+
+		// Payment is currently in flight. Register this subscriber and
+		// return without writing a result to the channel yet.
+		case StatusInFlight:
+			p.subscribersMtx.Lock()
+
+			subscribers, ok := p.subscribers[paymentHash]
+			if !ok {
+				subscribers = []chan PaymentEvent{c}
```
halseth: this is equivalent to `subscribers = append(subscribers, c)`

channeldb/control_tower.go
```diff
@@ -401,6 +489,131 @@ type InFlightPayment struct {
...
+				return errors.New("missing preimage")
+			}
+			var err error
+			event.Preimage, err = lntypes.MakePreimage(preimageSlice)
+			if err != nil {
+				return err
+			}
+
+			a, err := fetchPaymentAttempt(bucket)
+			event.Route = &a.Route
```
halseth: must check error before using result.

channeldb/control_tower.go
```diff
@@ -440,15 +653,10 @@ func (p *paymentControl) FetchInFlightPayments() ([]*InFlightPayment, error) {
...
-			// available.
-			attempt := bucket.Get(paymentAttemptInfoKey)
-			if attempt != nil {
-				r = bytes.NewReader(attempt)
-				inFlight.Attempt, err = deserializePaymentAttemptInfo(r)
-				if err != nil {
-					return err
-				}
+			// Now get the attempt info.
+			inFlight.Attempt, err = fetchPaymentAttempt(bucket)
```
halseth: attempts aren't always available here.

channeldb/control_tower.go
```diff
@@ -401,6 +489,131 @@ type InFlightPayment struct {
...
+		c <- event
+		close(c)
+
+		return nil
+	})
+	if err != nil {
+		return false, nil, err
+	}
+
+	return inFlight, c, nil
```
halseth: why is it necessary to distinguish a inflight payment here? Don't we know that it is either not initiated (which returns an error), or a payment is in flight and will be notified eventually?

lnrpc/routerrpc/router.proto
```diff
@@ -59,24 +59,54 @@ message PaymentRequest {
...
     */
-    bytes pre_image = 2;
+    SUCCEEDED = 1;
 
     /**
-    If not an empty string, then a string representation of the payment error.
+    There are more routes to try, but the payment timeout was exceeded.
     */
-    string payment_err = 3;
+    FAILED_TIMEOUT = 2;
```
halseth: We could also do `FAILED` with extra `Reason` field?

lnrpc/routerrpc/router.proto
```diff
@@ -228,12 +258,18 @@ message ChannelUpdate {
 service Router {
     /**
     SendPayment attempts to route a payment described by the passed
-    PaymentRequest to the final destination. If we are unable to route the
-    payment, or cannot find a route that satisfies the constraints in the
-    PaymentRequest, then an error will be returned. Otherwise, the payment
-    pre-image, along with the final route will be returned.
+    PaymentRequest to the final destination. The call returns immediately.
```
halseth: maybe specify that it returns after the payment has been successfully registered with the router,

lnrpc/routerrpc/router_server.go
```diff
@@ -178,22 +185,26 @@ func (s *Server) RegisterWithRootServer(grpcServer *grpc.Server) error {
...
-	preImage, _, err := s.cfg.Router.SendPayment(payment)
-	if err != nil {
-		return nil, err
-	}
+	// Launch the payment in a goroutine. It is no problem that we discard
+	// the error, because errors will also be recorded by control tower and
+	// available through LookupPayment. Exception to this are internal
+	// errors, but those should not happen.
+	go func() {
+		_, _, err := s.cfg.Router.SendPayment(payment)
```
halseth: I think we should rather make `SendPayment` return when it has registered the payment with the control tower. That way the caller know that after this RPC call returns without an error, the payment is being handled.

halseth: This is also racy, since we have no guarantee this goroutine will run before we call `LookupPayment`.

cmd/lncli/commands.go
```diff
@@ -2038,7 +2066,7 @@ var sendPaymentCommand = cli.Command{
 	destination.
 	`,
 	ArgsUsage: "dest amt payment_hash final_cltv_delta | --pay_req=[payment request]",
-	Flags: []cli.Flag{
+	Flags: append(paymentFlags,
```
halseth: 😱 can this modify the flag slice?

routing/router.go
joostjager: No, because the cltv limit really is optional, while final cltv is required.

lnrpc/routerrpc/router.proto
joostjager: Comment was copied from main rpc. Empty payment request explanation added.

lnrpc/routerrpc/router_backend.go
joostjager: good idea, created

lnrpc/routerrpc/router_backend.go
joostjager: I added it with the idea that we eventually are going to remove the main rpc `SendPayment` and that `DebugHTLC` is still useful. Do you think it should be removed?

lnrpc/routerrpc/router_backend.go
joostjager: Yes it seems safer to me like this.

lnrpc/routerrpc/router_backend.go
joostjager: exported and reused

channeldb/control_tower.go
joostjager: Added

joostjager: Also removed locks that were present at init and registerattempt. I don't think they were necessary.

channeldb/control_tower.go
joostjager: added extra condition

channeldb/control_tower.go
joostjager: Then on the rpc level, we would always first return an inflight, followed by a succeeded. Even if the payment succeed already a long time ago.

lnrpc/routerrpc/router.proto
joostjager: Like this it isn't possible to mistakingly return `FAILED` without a reason.

lnrpc/routerrpc/router.proto
joostjager: added

cmd/lncli/commands.go
joostjager: yes, we are wiser now. fixed.

lnrpc/routerrpc/router_server.go
joostjager: Yes you are totally right. I had it like you suggested, but rebased it out mistakingly thinking it was equivalent.

joostjager:
@halseth ptal

lnrpc/routerrpc/router_backend.go
halseth: I don't know if anybody uses this anymore. Introducing a new API feels like a good opportunity to get rid of it. Easier to add it later if someone really still want it

channeldb/control_tower.go
halseth: Which wouldn't _reeeeally_ make a difference for the caller, would it?

joostjager: It is a matter of preference. Imo it isn't the cleanest thing to start every lookup of a (historical) payment with an inflight notification

lnrpc/routerrpc/router_backend.go
joostjager: Removed

lntypes/hash.go
```diff
@@ -8,6 +8,9 @@ import (
 // HashSize of array used to store hashes.
 const HashSize = 32
 
+// ZeroHash is a predefined hash containing all zeroes.
+var ZeroHash = Hash{}
```
Roasbeef: 👍 

Alternatively, it can be declared as just: 
```go
var ZeroHash Hash
```

There're likely many other places that can use this new value, but we can defer updating all those locations for another change.

multimutex/multimutex_hash.go
```diff
@@ -0,0 +1,90 @@
+package multimutex
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/lightningnetwork/lnd/lntypes"
+)
+
+// HashMutex is a struct that keeps track of a set of mutexes with
```
Roasbeef: Entire file appears to be folded to less than 80 characters for each column.

Roasbeef: An alternative to this approach could be to:  
  1. Modify the base `multimutex` to be based on a `string` key
  2. Update the current integer based version to map to a string
  3. Use the base version in place of this 

However, I think we have a few callers of this package in `neutrino`, so then we'd need to update them all as well.

channeldb/control_tower.go
```diff
@@ -64,18 +71,39 @@ type ControlTower interface {
 
 	// FetchInFlightPayments returns all payments with status InFlight.
 	FetchInFlightPayments() ([]*InFlightPayment, error)
+
+	// SubscribePayment subscribes to updates for the payment with the given
+	// hash. It returns a boolean indicating whether the payment is still in
+	// flight and a channel that provides the final outcome of the payment.
+	SubscribePayment(paymentHash lntypes.Hash) (bool, chan PaymentEvent,
```
Roasbeef: How are subscriptions cancelled? To me, this feel like something that should be done on a higher level (in the router for example by intercepting the calls to the interface). Nowhere else do we have subscription/event logic _within_ the primary database package.

channeldb/control_tower.go
```diff
@@ -64,18 +71,39 @@ type ControlTower interface {
...
+
+	subscribers    map[lntypes.Hash][]chan PaymentEvent
+	subscribersMtx sync.Mutex
+	hashMutex      *multimutex.HashMutex
+}
+
+// PaymentEvent is the struct describing the events received by payment
+// subscribers.
+type PaymentEvent struct {
+	Status        PaymentStatus
```
Roasbeef: Missing `godoc` comments.

channeldb/control_tower.go
```diff
@@ -288,6 +353,16 @@ func (p *paymentControl) Fail(paymentHash lntypes.Hash,
 		return err
 	}
 
+	if updateErr == nil {
+		// Notify subscribers of fail event.
+		p.notifyFinalEvent(
```
Roasbeef: So there can only be a single subscriber per payment? Consider the following case: two RPC clients race to create a payment only one of them succeeds. IMO, in this scenario, the second should _still_ gain a subscription internally to be notified once the final payment has completed.

channeldb/control_tower.go
```diff
@@ -401,6 +487,129 @@ type InFlightPayment struct {
...
+
+		var event PaymentEvent
+
+		status := fetchPaymentStatus(bucket)
+		switch status {
+
+		// Payment is currently in flight. Register this subscriber and
+		// return without writing a result to the channel yet.
+		case StatusInFlight:
+			p.subscribersMtx.Lock()
```
Roasbeef: Grabbing mutexes in a database transaction makes be uneasy...

Alternatively, I think we can lift most of this logic _outside_ of the transaction by optimistically fetching everything we _might_ need. Returning back to my prior comment, if the subscription handling wasn't so coupled to the underlying database logic, there would be no concern here.

routing/router.go
```diff
@@ -1595,14 +1595,50 @@ type LightningPayment struct {
...
+
+	// Since this is the first time this payment is being made, we pass nil
+	// for the existing attempt.
+	return r.sendPayment(nil, payment, paySession)
+}
+
+// SendPaymentAsync is the non-blocking version of SendPayment. The payment
+// result needs to be retrieved via the control tower.
+func (r *ChannelRouter) SendPaymentAsync(payment *LightningPayment) error {
+	paySession, err := r.preparePayment(payment)
```
Roasbeef: I think this can just be:
```go
go func() {
    if _, _, err := r.SendPayment(payment); err != nil {
			log.Errorf("Payment with hash %v failed: %v",
				payment.PaymentHash, err)
    }
}

return nil
```

Roasbeef: If so, then we also wouldn't need to extract logic into `preparePayment`.

lnrpc/routerrpc/router.proto
```diff
@@ -60,24 +60,54 @@ message PaymentRequest {
...
-message PaymentResponse {
+message SendPaymentResponse {
+}
+
+message LookupPaymentRequest {
+    /// The hash of the payment to look up.
+    bytes payment_hash = 1;
+}
+
+enum PaymentState {
```
Roasbeef: What about "terminal" failures like unknown payment hash?

lnrpc/routerrpc/router_server.go
```diff
@@ -396,3 +407,81 @@ func marshallChannelUpdate(update *lnwire.ChannelUpdate) *ChannelUpdate {
 		FeeRate:         update.FeeRate,
 	}
 }
+func (s *Server) LookupPayment(request *LookupPaymentRequest,
```
Roasbeef: Missing `godoc` comment as well as some breathing room between the function above.

lnrpc/routerrpc/router_backend.go
```diff
@@ -45,6 +46,8 @@ type RouterBackend struct {
 	// that we receive payment requests that send to destinations on our
 	// network.
 	ActiveNetParams *chaincfg.Params
+
+	Tower channeldb.ControlTower
```
Roasbeef: If we only need payment subscription, then we can use a much slimmer interface here: an interface with only `SubscribePayment`.

lnrpc/routerrpc/router_server.go
```diff
@@ -396,3 +407,81 @@ func marshallChannelUpdate(update *lnwire.ChannelUpdate) *ChannelUpdate {
...
+
+			default:
+				return errors.New("unknown failure reason")
+			}
+
+		case channeldb.StatusSucceeded:
+			status.State = PaymentState_SUCCEEDED
+			status.Preimage = event.Preimage[:]
+			status.Route = s.cfg.RouterBackend.
+				MarshallRoute(event.Route)
```
Roasbeef: Funky line wrapping here.

lnrpc/routerrpc/router_server.go
```diff
@@ -396,3 +407,81 @@ func marshallChannelUpdate(update *lnwire.ChannelUpdate) *ChannelUpdate {
...
+			switch event.FailureReason {
+
+			case channeldb.FailureReasonTimeout:
+				status.State = PaymentState_FAILED_TIMEOUT
+
+			case channeldb.FailureReasonNoRoute:
+				status.State = PaymentState_FAILED_NO_ROUTE
+
+			default:
+				return errors.New("unknown failure reason")
```
Roasbeef: I feel like we're losing _too much_ context here. For example, without unknown payment hash, this this API can't really be used for probing. Alternatively, without the final expiry too soon error, then it makes it harder to debug issues like the sending node being behind the main chain.

cmd/lncli/commands.go
Roasbeef: Have our thoughts on this matter changed?

channeldb/control_tower.go
joostjager: `notifyFinalEvent` notifies a list of subscribers (multiple)

routing/router.go
joostjager: We don't want to return before the payment is registered with the control tower. When the rpc call returns, the caller should have the guarantee that their payment is started and will be resumed on restart.

lnrpc/routerrpc/router_backend.go
joostjager: I prefer to only add those interfaces when we actually need them.

cmd/lncli/commands.go
joostjager: This is a debug commit that will be taken off before merge. I'd rather update `lncli` later.

channeldb/control_tower.go
joostjager: Subscriptions don't need to be canceled. There is only ever a single notification and the channel is buffered.

lntypes/hash.go
joostjager: Declared like suggested. Indeed not updating everywhere to keep this pr contained.

channeldb/control_tower.go
joostjager: Added

channeldb/control_tower.go
joostjager: You are absolutely right about the abstraction. Created a split between db and subscriber handling.

channeldb/control_tower.go
joostjager: Addressed by splitting off subscriber logic.

multimutex/multimutex_hash.go
joostjager: Folding fixed.

Converted to string to prevent more of these types emerging in the future. Leaving int based one in place. It is probably more efficient and also to not let this pr sprawl into unrelated dependents of multimutex.

joostjager:
After discussion with @halseth, the decision was made to remove the `IN_FLIGHT` status from the former `LookupPayment` call. This state isn't essential and it allowed us convert `LookupPayment` into a regular blocking call that blocks until the payment completes (succes or fail). The name has been changed to `TrackPayment` to better convey the blocking character of it.

`SendPayment` has seen a change too: it is now blocking too and its return value is identical to `TrackPayment`. In a previous iteration, the response was given already after the payment was initiated. This intermediate state isn't essential either and even then the caller still needs to handle the case where an application crashes after calling `SendPayment` but before receiving the response. After a restart it needs to be aware that the payment may be in flight already.

The main downside of `SendPayment` returning immediately is that in typical use this call is usually followed by a `TrackPayment` call. So basically always two calls to make a payment. By having `SendPayment` return the payment outcome, only a single call is needed.

Another consideration is that `SendPayment` in `routerrpc` uses the same blocking model as the main rpc, facilitating application code migration.

routing/router.go
joostjager: Comment above isn't applicable anymore because `SendPayment` is blocking now.

lnrpc/routerrpc/router.proto
joostjager: The plan is to add that in a follow up pr. `Router` isn't yet storing those outcomes. The follow up pr would add storing them and passing them through over rpc.

lnrpc/routerrpc/router_server.go
joostjager: Yes, that information is definitely useful. Terminal errors need to be exposed. See comment above about follow up pr.

joostjager:
@halseth @Roasbeef ptal

cmd/lncli/commands.go
```diff
@@ -2012,6 +2012,34 @@ var cltvLimitFlag = cli.UintFlag{
 		"this payment",
 }
 
+var paymentFlags = []cli.Flag{
```
cfromknecht: not blocking, but could make this a function to avoid the double appends later on

routing/router.go
joostjager: Fixed

lnrpc/routerrpc/router.proto
```diff
@@ -7,41 +7,57 @@ package routerrpc;
...
-    int32 timeout_seconds = 4;
+    int32 timeout_seconds = 6;
+
+    /**
+    The maximum number of satoshis that will be paid as a fee of the payment.
+    If this field is left to the default value of 0, only zero-fee routes will
+    be considered. This usually means single hop routes connecting directly to
+    the destination. To send the payment without a fee limit, use max int here.
+    */
+    int64 fee_limit = 7;
```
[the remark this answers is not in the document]
joostjager: Added

cmd/lncli/commands.go
joostjager: I was looking for a way. Good suggestion, done.

routing/control_tower.go
```diff
@@ -85,10 +141,115 @@ func (p *controlTower) Success(paymentHash lntypes.Hash,
...
+	return nil
 }
 
 // FetchInFlightPayments returns all payments with status InFlight.
 func (p *controlTower) FetchInFlightPayments() ([]*channeldb.InFlightPayment, error) {
 	return p.db.FetchInFlightPayments()
 }
+
+// SubscribePayment subscribes to updates for the payment with the given hash.
+// It returns a boolean indicating whether the payment is still in flight and a
```
halseth: comment needs update

routing/control_tower.go
```diff
@@ -85,10 +141,115 @@ func (p *controlTower) Success(paymentHash lntypes.Hash,
...
+		event.Success = true
+		event.Preimage = *payment.PaymentPreimage
+		event.Route = &payment.Attempt.Route
+
+	// Payment already failed. It is not necessary to register as a
+	// subscriber, because we can send the result on the channel
+	// immediately.
+	case channeldb.StatusFailed:
+		event.Success = false
+		event.FailureReason = *payment.Failure
```
halseth: don't we want to set the route for failures?

routing/control_tower.go
```diff
@@ -85,10 +141,115 @@ func (p *controlTower) Success(paymentHash lntypes.Hash,
 func (p *controlTower) Fail(paymentHash lntypes.Hash,
 	reason channeldb.FailureReason) error {
 
-	return p.db.Fail(paymentHash, reason)
+	// Lock hash to prevent the race condition described in the Success
+	// method.
+	p.hashMutex.Lock(string(paymentHash[:]))
+	defer p.hashMutex.Unlock(string(paymentHash[:]))
```
halseth: same, I think it can be locked after the db tx is over. 

Would this also let us use a regular mutex? As the mutex would no longer encompass the critical path (multiple payments recording attempts and results with the DB).

routing/control_tower_test.go
```diff
@@ -0,0 +1,253 @@
...
+		Hops: []*route.Hop{
+			testHop,
+			testHop,
+		},
+	}
+)
+
+// TestPaymentControlSubscribeUnknown tests that subscribing to an unknown
+// payment fails.
+func TestPaymentControlSubscribeUnknown(t *testing.T) {
```
halseth: tests should be named `TestControlTower...` since we are now separating `ControlTower` and `PaymentControl` :)

routing/control_tower_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// Register a third subscriber after the payment succeeded.
+	subscriber3, err := pControl.SubscribePayment(info.PaymentHash)
+	if err != nil {
+		t.Fatalf("expected subscribe to succeed, but got: %v", err)
+	}
+
+	// We expect all subscribers to now report the final outcome followed by
+	// no other events.
+	for _, s := range []chan PaymentResult{
```
halseth: define the slice before the loop to avoid awkward line wrap.

routing/control_tower_test.go
```diff
@@ -0,0 +1,253 @@
...
+		subscriber1, subscriber2, subscriber3,
+	} {
+		event := <-s
```
halseth: should have a timeout on the channel receive.

routing/control_tower_test.go
```diff
@@ -0,0 +1,253 @@
...
+		if event.Preimage != preimg {
+			t.Fatal("unexpected preimage")
+		}
+		if !reflect.DeepEqual(event.Route, &attempt.Route) {
+			t.Fatal("unexpected route")
+		}
+
+		// After the final event, we expect the channel to be closed.
+		select {
+		case _, ok := <-s:
```
halseth: does the go runtime guarantee that the close event will arrive immediately?

routing/control_tower_test.go
```diff
@@ -0,0 +1,253 @@
...
+			}
+		default:
+			t.Fatal("expected channel to be closed")
+		}
+	}
+}
+
+// TestPaymentControlSubscribeFail tests that payment updates for a
+// failed payment are properly sent to subscribers.
+func TestPaymentControlSubscribeFail(t *testing.T) {
```
halseth: Looks like this test could be merged with the previous by initiating a slice where, say, half payments succeed, half fails.

server.go
```diff
@@ -650,12 +650,14 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB, cc *chainControl,
...
 
+	paymentControl := channeldb.NewPaymentControl(chanDB)
+
 	s.chanRouter, err = routing.New(routing.Config{
 		Graph:              chanGraph,
 		Chain:              cc.chainIO,
 		ChainView:          cc.chainView,
 		Payer:              s.htlcSwitch,
-		Control:            channeldb.NewPaymentControl(chanDB),
+		Control:            routing.NewControlTower(paymentControl),
```
halseth: nit: looks like should be in the "move control tower" commit?

routing/router.go
```diff
@@ -1595,14 +1595,50 @@ type LightningPayment struct {
...
+	if err != nil {
+		return err
+	}
+
+	// Since this is the first time this payment is being made, we pass nil
+	// for the existing attempt.
+	go func() {
```
halseth: needs waitgroup

lnrpc/routerrpc/router.proto
```diff
@@ -60,24 +60,49 @@ message PaymentRequest {
     int32 cltv_limit = 9;
 }
 
-message PaymentResponse {
+message SendPaymentResponse {
```
halseth: not used?

lnrpc/routerrpc/router_server.go
```diff
@@ -10,15 +10,18 @@ import (
...
-	"github.com/lightningnetwork/lnd/lntypes"
-
-	"github.com/btcsuite/btcutil"
 	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing"
 	"github.com/lightningnetwork/lnd/routing/route"
+
+	"github.com/btcsuite/btcutil"
```
halseth: can go up together with other github imports

lnrpc/routerrpc/router_server.go
```diff
@@ -53,6 +56,10 @@ var (
 			Entity: "offchain",
 			Action: "write",
 		}},
+		"/routerrpc.Router/TrackPayment": {{
```
halseth: Oh, yeah this is another benefit of splitting up the APIs that we can offer a way to only look at outcomes without initiating new payments :)

lnrpc/routerrpc/router_server.go
```diff
@@ -178,22 +185,31 @@ func (s *Server) RegisterWithRootServer(grpcServer *grpc.Server) error {
...
 	if err != nil {
+		log.Debugf("SendPayment async result for hash %x: %v",
+			payment.PaymentHash, err)
+
+		// Transform user errors to grpc code.
+		if err == channeldb.ErrPaymentInFlight ||
+			err == channeldb.ErrAlreadyPaid {
+
+			return nil, status.Error(
+				codes.AlreadyExists, err.Error(),
```
halseth: Do we need these grpc error codes now that this call is sync?

lnrpc/routerrpc/router_server.go
```diff
@@ -396,3 +412,68 @@ func marshallChannelUpdate(update *lnwire.ChannelUpdate) *ChannelUpdate {
...
+// waitForPayment waits for the result of the payment to be available and
+// returns it as an rpc type.
+func (s *Server) waitForPayment(ctx context.Context, paymentHash lntypes.Hash) (
+	*PaymentResponse, error) {
+
+	// Subscribe to the outcome of this payment.
+	resultChan, err := s.cfg.RouterBackend.Tower.SubscribePayment(
+		paymentHash,
+	)
+	switch err {
```
halseth: style nit:
```go
switch {
case err == channeldb.ErrPaymentNotInitiated:
     return nil, status.Error(codes.NotFound, err.Error())
case err != nil {
     return nil, err
}
```

lnrpc/routerrpc/router_server.go
```diff
@@ -396,3 +412,68 @@ func marshallChannelUpdate(update *lnwire.ChannelUpdate) *ChannelUpdate {
...
+	case nil:
+	case channeldb.ErrPaymentNotInitiated:
+		return nil, status.Error(codes.NotFound, err.Error())
+	default:
+		return nil, err
+	}
+
+	// Wait for the outcome of the payment. For payments that have already
+	// completed, the event should already be waiting on the channel.
```
halseth: Maybe add a log here before wait started?

lnrpc/routerrpc/router_server.go
```diff
@@ -396,3 +412,68 @@ func marshallChannelUpdate(update *lnwire.ChannelUpdate) *ChannelUpdate {
...
+	select {
+	case result := <-resultChan:
+		// Marshall event to rpc type.
+		var response PaymentResponse
+
```
halseth: log outcome.

rpcserver.go
```diff
@@ -476,6 +476,7 @@ func newRPCServer(s *server, macService *macaroons.Service,
 		},
 		FindRoute:       s.chanRouter.FindRoute,
 		ActiveNetParams: activeNetParams.Params,
+		Tower:           s.controlTower,
```
halseth: can we pass the controltower as a param instead (or make an `rpcserverCfg` to pass in)? Seems unnecessary to increase the server<->rpcserver coupling further, and avoids the server to hold a reference to it.

cmd/lncli/commands.go
```diff
@@ -2038,7 +2070,7 @@ var sendPaymentCommand = cli.Command{
 	destination.
 	`,
 	ArgsUsage: "dest amt payment_hash final_cltv_delta | --pay_req=[payment request]",
-	Flags: []cli.Flag{
+	Flags: append(paymentFlags(),
```
halseth: 👍

lntypes/hash.go
```diff
@@ -8,6 +8,9 @@ import (
 // HashSize of array used to store hashes.
 const HashSize = 32
 
+// ZeroHash is a predefined hash containing all zeroes.
+var ZeroHash Hash
```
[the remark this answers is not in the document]
joostjager: I don't think it will work as const, for example when comparing to `ZeroHash`. Also that would allocate it over and over again?

routing/control_tower.go
```diff
@@ -35,18 +40,48 @@ type ControlTower interface {
...
+		error)
+}
+
+// PaymentResult is the struct describing the events received by payment
+// subscribers.
+type PaymentResult struct {
+	// Success indicates whether the payment was successful.
+	Success bool
+
+	// Route is the (last) route attempted to send the HTLC.
```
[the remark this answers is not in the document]
joostjager: I thought it can never be nil? Every failure or success should have a route?

routing/control_tower.go
```diff
@@ -75,7 +110,28 @@ func (p *controlTower) RegisterAttempt(paymentHash lntypes.Hash,
 func (p *controlTower) Success(paymentHash lntypes.Hash,
 	preimage lntypes.Preimage) error {
 
-	return p.db.Success(paymentHash, preimage)
+	// Lock hash to prevent the following race: payment status is updated to
+	// succeeded, new subscriber is registered and will immediately receive
+	// the succeeded status, this function sends out the succeeded
+	// notification again to the same subscriber.
+	p.hashMutex.Lock(string(paymentHash[:]))
```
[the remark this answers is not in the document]
joostjager: The reason the lock needs to be acquired before the db transaction is to prevent a new subscription happening after the db transaction, but before `notifyFinalEvent`, as is described in the comment above. Maybe I misunderstand you suggestion?

routing/control_tower.go
joostjager: See comment on `Success`

routing/control_tower.go
joostjager: Not yet. With `timeout` and `no_route` the route itself isn't relevant. I am planning to expose the route only on failures where it makes sense, like `unknown_pay_hash`.

routing/control_tower_test.go
joostjager: good suggestion, done

routing/control_tower_test.go
joostjager: Added

routing/control_tower_test.go
joostjager: Good point, added timeout

routing/control_tower_test.go
joostjager: It could be merged, but I prefer to keep them apart. Not to test too many things in a single test.

routing/router.go
joostjager: added

lnrpc/routerrpc/router.proto
joostjager: hmm leftover indeed, removed

lnrpc/routerrpc/router_server.go
joostjager: Yes, we sometimes don't want to start if not started yet.

lnrpc/routerrpc/router_server.go
joostjager: Yes, otherwise we can only do string matching again at the client

lnrpc/routerrpc/router_server.go
joostjager: Okay, nicer. I never felt good about `case nil:`, but this simple improvement didn't occur to me.

lnrpc/routerrpc/router_server.go
joostjager: Added

lnrpc/routerrpc/router_server.go
joostjager: added

rpcserver.go
joostjager: I don't see much harm in this here. Maybe it is good to do some hiding, but I prefer to do it outside of this PR. `r.server` is referenced 113 times in `rpcserver.go`, so it will me a large diff I expect.

server.go
joostjager: yes, that's right

joostjager:
@halseth ptal

routing/control_tower.go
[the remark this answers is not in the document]
halseth: Failure can have no attempt if no route is found.

lnrpc/routerrpc/router_server.go
halseth: I meant we can hand out a macaroon that only allows tracking, but not initiating payments :)

routing/control_tower.go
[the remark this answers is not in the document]
halseth: From reading the comment that sounds alright though?

Payment status is updated to succeeded, new subscriber comes in and (acquire lock) retrieves succeed status and sends notification (release lock), this function (acquire lock)  finds no one to send notification to (release lock).

Or is there another order of actions that require us to hold it while we update the DB?

lnrpc/routerrpc/router_server.go
joostjager: Ah yes, didn't think of that, but indeed that is a benefit.

routing/control_tower.go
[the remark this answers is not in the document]
joostjager: Actually it currently is nil for every failure. Comment updated.

routing/control_tower.go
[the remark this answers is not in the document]
joostjager: Updated code to only use a global lock. It indeed seems to be sufficient.

lnrpc/routerrpc/router_backend.go
Roasbeef: Rephrased the router backend now depends on a number of methods that it doesn't actually need. Any thing that mocks this `Tower` variable now needs to unnecessarily create a bunch of empty methods, when we only need a method that exposes the subscription functionality.

joostjager: Yes, I understand that in general, but I mean that we currently don't mock that `Tower` variable. I'd like to postpone creating that interface until we need to mock that. But if it is a blocker for you, let me know. I don't mind adding it.

routing/control_tower.go
```diff
@@ -85,10 +131,108 @@ func (p *controlTower) Success(paymentHash lntypes.Hash,
...
+	}
+
+	var event PaymentResult
+
+	switch payment.Status {
+
+	// Payment is currently in flight. Register this subscriber and
+	// return without writing a result to the channel yet.
+	case channeldb.StatusInFlight:
+		subscribers := p.subscribers[paymentHash]
```
halseth: style nit: can be written 
```go
p.subscribers[paymentHash] = append(p.subscribers[paymentHash], c)
```

routing/control_tower.go
```diff
@@ -85,10 +131,108 @@ func (p *controlTower) Success(paymentHash lntypes.Hash,
...
+	delete(p.subscribers, paymentHash)
+	p.subscribersMtx.Unlock()
+
+	// Notify all subscribers of the event. The subscriber channel is
+	// buffered, so it cannot block here.
+	for _, subscriber := range list {
+		subscriber <- event
+		close(subscriber)
+	}
+
```
halseth: nit: extra newline

lnrpc/routerrpc/router.proto
```diff
@@ -60,24 +60,46 @@ message PaymentRequest {
...
+    FAILED_TIMEOUT = 1;
+    
+    /**
+    All possible routes were tried and failed permanently. Or were no
+    routes to the destination at all.
     */
-    bytes pre_image = 2;
+    FAILED_NO_ROUTE = 2;
+}
+
```
halseth: nit: extra newline

lnrpc/routerrpc/router_server.go
```diff
@@ -178,22 +185,31 @@ func (s *Server) RegisterWithRootServer(grpcServer *grpc.Server) error {
...
 
 	payment, err := s.cfg.RouterBackend.extractIntentFromSendRequest(req)
 	if err != nil {
 		return nil, err
 	}
 
-	preImage, _, err := s.cfg.Router.SendPayment(payment)
+	err = s.cfg.Router.SendPaymentAsync(payment)
 	if err != nil {
+		log.Debugf("SendPayment async result for hash %x: %v",
```
halseth: `Errorf`?

cmd/lncli/commands.go
```diff
@@ -2098,7 +2100,28 @@ var sendPaymentCommand = cli.Command{
 
 // retrieveFeeLimit retrieves the fee limit based on the different fee limit
 // flags passed.
-func retrieveFeeLimit(ctx *cli.Context) (*lnrpc.FeeLimit, error) {
+func retrieveFeeLimit(ctx *cli.Context, amt int64) (int64, error) {
```
halseth: reminder to remove this commit.

lnrpc/routerrpc/router_server.go
joostjager: Fixed using different log level for expected and unexpected errors.

joostjager:
Brought back the payment update stream and the `IN_FLIGHT` payment state.

ptal @halseth @Roasbeef

channeldb/payment_control.go
```diff
@@ -187,9 +192,12 @@ func (p *PaymentControl) RegisterAttempt(paymentHash lntypes.Hash,
 // duplicate payments to the same payment hash. The provided preimage is
 // atomically saved to the DB for record keeping.
 func (p *PaymentControl) Success(paymentHash lntypes.Hash,
-	preimage lntypes.Preimage) error {
+	preimage lntypes.Preimage) (*route.Route, error) {
```
halseth: To avoid having to do this change, one could instead provinde the route as a param to the `Success` method in the routing `ControlTower`.

routing/control_tower.go
```diff
@@ -85,10 +132,108 @@ func (p *controlTower) Success(paymentHash lntypes.Hash,
...
+	switch payment.Status {
+
+	// Payment is currently in flight. Register this subscriber and
+	// return without writing a result to the channel yet.
+	case channeldb.StatusInFlight:
+		p.subscribers[paymentHash] = append(
+			p.subscribers[paymentHash], c,
+		)
+
+		return true, c, nil
```
halseth: Ok, this is much better now that we don't have to return this from the DB.

channeldb/payment_control.go
joostjager: I am not sure if that is better. It puts a new requirement on the caller of `ControlTower`.

routing/control_tower.go
joostjager: It isn't new, I just restored code that I had previously. Do you mean it is better because the bool isn't set in the db tx?

routing/control_tower.go
```diff
@@ -1,7 +1,11 @@
 package routing
 
 import (
+	"sync"
+
+	"github.com/go-errors/errors"
```
cfromknecht: regular "errors" package should suffice?

joostjager: fixed

PR #2974, opened 2019-04-18
Build LND subservers when building docker image

OpinionatedGeek:
Now that the official releases are built with subservers, can/should we build docker images with subservers by default too?

This one-line change is all I needed to do to build LND with subservers so loop would work with the docker image.

(Checklist not applicable - it's a one-line change to a docker file rather than a go file.)


wpaulino:
> Now that the official releases are built with subservers, can/should we build docker images with subservers by default too?

Doing so makes sense to me. Could you also update [`lnd/Dockerfile`](https://github.com/lightningnetwork/lnd/blob/master/Dockerfile)?

OpinionatedGeek:
lnd/Dockerfile updated as requested. (Apologies - I didn't even know it was there.)

PR #2982, opened 2019-04-19
contractcourt: speed up TestChainWatcherDataLossProtect by ditching q…

Roasbeef:
…uickcheck

In this commit, we speed up the `TestChainWatcherDataLossProtect`
_considerably_ by enumerating relevant tests using table driven tests
rather than generating random tests via the `testing/quick` package.
Each of these test cases are also run in parallel bringing down the
execution time of this test from a few minutes, to a few seconds.

Companion to #2855.

### DIFF
--- a/contractcourt/chain_watcher_test.go
+++ b/contractcourt/chain_watcher_test.go
@@ -3,11 +3,8 @@ package contractcourt
 import (
 	"bytes"
 	"crypto/sha256"
-	"math"
-	"math/rand"
-	"reflect"
+	"fmt"
 	"testing"
-	"testing/quick"
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -254,7 +251,7 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
 	// was broadcast, while numUpdates is the actual number of updates
 	// we'll execute. Both of these will be random 8-bit values generated
 	// by testing/quick.
-	dlpScenario := func(testCase dlpTestCase) bool {
+	dlpScenario := func(t *testing.T, testCase dlpTestCase) bool {
 		// First, we'll create two channels which already have
 		// established a commitment contract between themselves.
 		aliceChannel, bobChannel, cleanUp, err := lnwallet.CreateTestChannels()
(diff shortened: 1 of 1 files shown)


contractcourt/chain_watcher_test.go
```diff
@@ -369,43 +366,48 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
-			broadcastNum := uint8(stateNum + 1 + uint8(broadcastRange))
-
-			testCase := dlpTestCase{
-				BroadcastStateNum: broadcastNum,
-				NumUpdates:        stateNum,
+		t.Run(testName, func(t *testing.T) {
+			t.Parallel()
+
+			if !dlpScenario(t, testCase) {
+				t.Fatalf("test %v failed", testName)
```
cfromknecht: will this ever be executed? it seems the test would fail inside `dlpScenario` now

Roasbeef: `Errorf` fails at the very end, rather then immediately.

cfromknecht: gotcha, looks like there're some other calls to `Fatalf` in `dlpScenario`, though they have more to do with initialization rather than the assertions themselves. probably fine to leave it then

contractcourt/chain_watcher_test.go
```diff
@@ -369,43 +366,48 @@ func TestChainWatcherDataLossProtect(t *testing.T) {
...
-			// between the state and 255, ensuring that it' at
-			// least one state beyond the target stateNum.
-			broadcastRange := rand.Int31n(int32(math.MaxUint8 - stateNum))
-			broadcastNum := uint8(stateNum + 1 + uint8(broadcastRange))
-
-			testCase := dlpTestCase{
-				BroadcastStateNum: broadcastNum,
-				NumUpdates:        stateNum,
+		t.Run(testName, func(t *testing.T) {
+			t.Parallel()
```
halseth: This might be a problem since we are not capturing the loop variable. See https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

Roasbeef: Fixed

Roasbeef:
PTAL @halseth @cfromknecht

PR #2996, opened 2019-04-23
lntest: expect one graph update for node's own channels

wpaulino:
In this PR, we address a bug within the test harness that would arise due to race conditions between two peers opening a channel. Peers that are not a part of a given channel should expect to receive the
policies for both sides of the channel. This is not always the case however if the peer is part of the channel. In this case, a peer would only receive one policy from the remote party, since it already knows of its own. To fix this, we modify each node within the test harness to only expect one policy when waiting for a channel announcement of their own channel.

### DIFF
--- a/lntest/harness.go
+++ b/lntest/harness.go
@@ -809,7 +809,7 @@ func (n *NetworkHarness) OpenChannel(ctx context.Context,
 			"alice and bob: %v", err)
 	}
 
-	chanOpen := make(chan struct{})
+	chanPending := make(chan *lnrpc.PendingUpdate)
 	errChan := make(chan error)
 	go func() {
 		// Consume the "channel pending" update. This waits until the node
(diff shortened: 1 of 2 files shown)


wpaulino:
>        --- FAIL: TestLightningNetworkDaemon/channel_backup_restore/restore_from_backup_file (40.64s)
>            lnd_test.go:81: Error outside of test: *errors.errorString dave didn't report channel: channel not opened before timeout
>                /home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd_test.go:13610 (0xd114bc)
>                	testChanRestoreScenario: t.Fatalf("dave didn't report channel: %v", err)
>                /home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd_test.go:13896 (0xd28e64)
>                	testChannelBackupRestore.func5: testChanRestoreScenario(h, net, &testCase, password)
>                /home/travis/.gimme/versions/go1.12.4.linux.amd64/src/testing/testing.go:865 (0x4fbe50)
>                	tRunner: fn(t)
>                /home/travis/.gimme/versions/go1.12.4.linux.amd64/src/runtime/asm_amd64.s:1337 (0x45e711)
>                	goexit: BYTE	$0x90	// NOP

This failure is not related and is due to a race condition between calling `SubscribeChannelGraph` and the `ChannelRouter` starting when creating a node with `NewNodeWithSeed`.

halseth:
Doesn't this sound like a bug in the graph subscription API, really? I would expect the policy update to be notified also for local channels, or at least for the API to be consistent.

wpaulino:
Good point @halseth.

PR #3010, opened 2019-04-25
lntest/itest: prevent direct connection within testNodeAnnouncement

wpaulino:
Alice and Dave don't need to be connected in order to receive the node announcement as we assume that she can receive it from Bob because they are connected at the beginning of every test.

### DIFF
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -9334,6 +9334,17 @@ func testNodeAnnouncement(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("unable to connect bob to carol: %v", err)
 	}
 
+	// Alice shouldn't receive any new updates yet since the channel has yet
+	// to be opened.
+	select {
+	case <-aliceSub.updateChan:
+		t.Fatalf("received unexpected update from dave")
+	case <-time.After(time.Second):
+	}
+
+	// We'll then go ahead and open a channel between Bob and Dave. This
+	// ensures that Alice receives the node announcement from Bob as part of
+	// the announcement broadcast.
 	ctxt, _ = context.WithTimeout(ctxb, channelOpenTimeout)
 	chanPoint := openChannelAndAssert(
 		ctxt, t, net, net.Bob, dave,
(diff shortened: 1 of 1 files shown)


halseth:
@wpaulino Are we still seeing this issue?

wpaulino:
@halseth yes, the error is `did not receive node ann update`.

lntest/itest/lnd_test.go
```diff
@@ -8751,6 +8751,24 @@ func testNodeAnnouncement(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("unable to connect bob to carol: %v", err)
 	}
 
+	// We'll connect Alice with Dave now to ensure she receives the node
```
halseth: Feels a bit like working around the actual problem. Does this mean a node connecting to another node never will receive its node announcement if no new channel is opened? Shouldn't this be sent during the historical sync?

wpaulino: > Does this mean a node connecting to another node never will receive its node announcement if no new channel is opened? 

That is part of the spec, see https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#the-node_announcement-message.

> Shouldn't this be sent during the historical sync?

Alice has been connected to other nodes prior to this test, so a historical sync does not cover it.

halseth: > Alice has been connected to other nodes prior to this test, so a historical sync does not happen.

Should Alice receive the announcement from the node she is already connected to (Bob?) then. If she has no connections before connecting to Dave, she should perform a historical sync.

wpaulino: You're right, Alice ends up not receiving Dave's announcement from Bob due to https://github.com/lightningnetwork/lnd/issues/3058.

wpaulino: I think we still want a connection to be made explicitly between Alice and Bob at this point however to avoid relying on the test harness connecting them before the start of every test.

halseth:
Can be closed now that https://github.com/lightningnetwork/lnd/pull/3219 is in?

wpaulino:
@halseth see https://github.com/lightningnetwork/lnd/pull/3010#discussion_r295074152.

PR #3013, opened 2019-04-25
[dependency update] update to latest btcwallet, btcutil, btcd, neutrino

halseth:
Resolves a crash when using neutrino backend.


halseth:
Done! PTAL @Roasbeef

halseth:
Allso updated `neutrino` and `btcd` deps when at it.

PR #3016, opened 2019-04-26
[contractcourt] Republish close tx on startup

halseth:
This PR makes us store the closing TX we propose to the database, such that the ChainArbitrator can re-publish them at startup for channels in state `waiting close`.

This is meant to handle the case were we attempt to close the channel, but go down before the tx has propagated to the network. Now we'll instead store this tx the moment it is ready, such that we can re-publish it.

Builds on https://github.com/lightningnetwork/lnd/pull/3510

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -59,6 +59,10 @@ var (
 	// remote peer during a channel sync in case we have lost channel state.
 	dataLossCommitPointKey = []byte("data-loss-commit-point-key")
 
+	// closingTxKey points to a the closing tx that we broadcasted when
+	// moving the channel to state CommitBroadcasted.
+	closingTxKey = []byte("closing-tx-key")
+
 	// commitDiffKey stores the current pending commitment state we've
 	// extended to the remote party (if any). Each time we propose a new
 	// state, we store the information necessary to reconstruct this state
@@ -878,12 +882,24 @@ func (c *OpenChannel) isBorked(chanBucket *bbolt.Bucket) (bool, error) {
 
 // MarkCommitmentBroadcasted marks the channel as a commitment transaction has
 // been broadcast, either our own or the remote, and we should watch the chain
-// for it to confirm before taking any further action.
-func (c *OpenChannel) MarkCommitmentBroadcasted() error {
+// for it to confirm before taking any further action. It takes as argument the
+// closing tx _we believe_ will appear in the chain. This is only used to
+// republish this tx at startup to ensure propagation, and we should still
+// handle the case where a different tx actually hits the chain.
+func (c *OpenChannel) MarkCommitmentBroadcasted(closeTx *wire.MsgTx) error {
 	c.Lock()
 	defer c.Unlock()
 
-	return c.putChanStatus(ChanStatusCommitBroadcasted)
+	var b bytes.Buffer
+	if err := WriteElement(&b, closeTx); err != nil {
+		return err
+	}
+
+	putClosingTx := func(chanBucket *bbolt.Bucket) error {
+		return chanBucket.Put(closingTxKey, b.Bytes())
+	}
+
+	return c.putChanStatus(ChanStatusCommitBroadcasted, putClosingTx)
 }
 
 // putChanStatus appends the given status to the channel. fs is an optional
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -16,6 +16,7 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/shachain"
@@ -808,6 +809,85 @@ func (c *OpenChannel) MarkBorked() error {
 	defer c.Unlock()
 
 	return c.putChanStatus(ChanStatusBorked)
+}
+
+// ChanSyncMsg returns the ChannelReestablish message that should be sent upon
+// reconnection with the remote peer that we're maintaining this channel with.
+// The information contained within this message is necessary to re-sync our
+// commitment chains in the case of a last or only partially processed message.
+// When the remote party receiver this message one of three things may happen:
+//
+//   1. We're fully synced and no messages need to be sent.
+//   2. We didn't get the last CommitSig message they sent, to they'll re-send
+//      it.
+//   3. We didn't get the last RevokeAndAck message they sent, so they'll
+//      re-send it.
+//
+// The isRestoredChan bool indicates if we need to craft a chan sync message
+// for a channel that's been restored. If this is a restored channel, then
+// we'll modify our typical chan sync  message to ensure they force close even
+// if we're on the very first state.
+func (c *OpenChannel) ChanSyncMsg(
+	isRestoredChan bool) (*lnwire.ChannelReestablish, error) {
+
+	c.Lock()
+	defer c.Unlock()
+
+	// The remote commitment height that we'll send in the
+	// ChannelReestablish message is our current commitment height plus
+	// one. If the receiver thinks that our commitment height is actually
+	// *equal* to this value, then they'll re-send the last commitment that
+	// they sent but we never fully processed.
+	localHeight := c.LocalCommitment.CommitHeight
+	nextLocalCommitHeight := localHeight + 1
+
+	// The second value we'll send is the height of the remote commitment
+	// from our PoV. If the receiver thinks that their height is actually
+	// *one plus* this value, then they'll re-send their last revocation.
+	remoteChainTipHeight := c.RemoteCommitment.CommitHeight
+
+	// If this channel has undergone a commitment update, then in order to
+	// prove to the remote party our knowledge of their prior commitment
+	// state, we'll also send over the last commitment secret that the
+	// remote party sent.
+	var lastCommitSecret [32]byte
+	if remoteChainTipHeight != 0 {
+		remoteSecret, err := c.RevocationStore.LookUp(
+			remoteChainTipHeight - 1,
+		)
+		if err != nil {
+			return nil, err
+		}
+		lastCommitSecret = [32]byte(*remoteSecret)
+	}
+
+	// Additionally, we'll send over the current unrevoked commitment on
+	// our local commitment transaction.
+	currentCommitSecret, err := c.RevocationProducer.AtIndex(
+		localHeight,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// If we've restored this channel, then we'll purposefully give them an
+	// invalid LocalUnrevokedCommitPoint so they'll force close the channel
+	// allowing us to sweep our funds.
+	if isRestoredChan {
+		currentCommitSecret[0] ^= 1
+	}
+
+	return &lnwire.ChannelReestablish{
+		ChanID: lnwire.NewChanIDFromOutPoint(
+			&c.FundingOutpoint,
+		),
+		NextLocalCommitHeight:  nextLocalCommitHeight,
+		RemoteCommitTailHeight: remoteChainTipHeight,
+		LastRemoteCommitSecret: lastCommitSecret,
+		LocalUnrevokedCommitPoint: input.ComputeCommitmentPoint(
+			currentCommitSecret[:],
+		),
+	}, nil
 }
 
 // isBorked returns true if the channel has been marked as borked in the
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -726,44 +726,16 @@ func (c *OpenChannel) MarkDataLoss(commitPoint *btcec.PublicKey) error {
 	c.Lock()
 	defer c.Unlock()
 
-	var status ChannelStatus
-	if err := c.Db.Update(func(tx *bbolt.Tx) error {
-		chanBucket, err := fetchChanBucket(
-			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
-		)
-		if err != nil {
-			return err
-		}
-
-		channel, err := fetchOpenChannel(chanBucket, &c.FundingOutpoint)
-		if err != nil {
-			return err
-		}
-
-		// Add status LocalDataLoss to the existing bitvector found in
-		// the DB.
-		status = channel.chanStatus | ChanStatusLocalDataLoss
-		channel.chanStatus = status
-
-		var b bytes.Buffer
-		if err := WriteElement(&b, commitPoint); err != nil {
-			return err
-		}
-
-		err = chanBucket.Put(dataLossCommitPointKey, b.Bytes())
-		if err != nil {
-			return err
-		}
-
-		return putOpenChannel(chanBucket, channel)
-	}); err != nil {
+	var b bytes.Buffer
+	if err := WriteElement(&b, commitPoint); err != nil {
 		return err
 	}
 
-	// Update the in-memory representation to keep it in sync with the DB.
-	c.chanStatus = status
+	putCommitPoint := func(chanBucket *bbolt.Bucket) error {
+		return chanBucket.Put(dataLossCommitPointKey, b.Bytes())
+	}
 
-	return nil
+	return c.putChanStatus(ChanStatusLocalDataLoss, putCommitPoint)
 }
 
 // DataLossCommitPoint retrieves the stored commit point set during
@@ -914,7 +886,12 @@ func (c *OpenChannel) MarkCommitmentBroadcasted() error {
 	return c.putChanStatus(ChanStatusCommitBroadcasted)
 }
 
-func (c *OpenChannel) putChanStatus(status ChannelStatus) error {
+// putChanStatus appends the given status to the channel. fs is an optional
+// list of closures that are given the chanBucket in order to atomically add
+// extra information together with the new status.
+func (c *OpenChannel) putChanStatus(status ChannelStatus,
+	fs ...func(*bbolt.Bucket) error) error {
+
 	if err := c.Db.Update(func(tx *bbolt.Tx) error {
 		chanBucket, err := fetchChanBucket(
 			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
@@ -932,7 +909,17 @@ func (c *OpenChannel) putChanStatus(status ChannelStatus) error {
 		status = channel.chanStatus | status
 		channel.chanStatus = status
 
-		return putOpenChannel(chanBucket, channel)
+		if err := putOpenChannel(chanBucket, channel); err != nil {
+			return err
+		}
+
+		for _, f := range fs {
+			if err := f(chanBucket); err != nil {
+				return err
+			}
+		}
+
+		return nil
 	}); err != nil {
 		return err
 	}
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -514,16 +514,6 @@ type OpenChannel struct {
 	sync.RWMutex
 }
 
-// FullSync serializes, and writes to disk the *full* channel state, using
-// both the active channel bucket to store the prefixed column fields, and the
-// remote node's ID to store the remainder of the channel state.
-func (c *OpenChannel) FullSync() error {
-	c.Lock()
-	defer c.Unlock()
-
-	return c.Db.Update(c.fullSync)
-}
-
 // ShortChanID returns the current ShortChannelID of this channel.
 func (c *OpenChannel) ShortChanID() lnwire.ShortChannelID {
 	c.RLock()
@@ -648,9 +638,8 @@ func fetchChanBucket(tx *bbolt.Tx, nodeKey *btcec.PublicKey,
 	return chanBucket, nil
 }
 
-// fullSync is an internal version of the FullSync method which allows callers
-// to sync the contents of an OpenChannel while re-using an existing database
-// transaction.
+// fullSync syncs the contents of an OpenChannel while re-using an existing
+// database transaction.
 func (c *OpenChannel) fullSync(tx *bbolt.Tx) error {
 	// First fetch the top level bucket which stores all data related to
 	// current, active channels.
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -803,13 +803,10 @@ func (c *OpenChannel) MarkBorked() error {
 //   3. We didn't get the last RevokeAndAck message they sent, so they'll
 //      re-send it.
 //
-// The isRestoredChan bool indicates if we need to craft a chan sync message
-// for a channel that's been restored. If this is a restored channel, then
-// we'll modify our typical chan sync  message to ensure they force close even
-// if we're on the very first state.
-func (c *OpenChannel) ChanSyncMsg(
-	isRestoredChan bool) (*lnwire.ChannelReestablish, error) {
-
+// If this is a restored channel, having status ChanStatusRestored, then we'll
+// modify our typical chan sync message to ensure they force close even if
+// we're on the very first state.
+func (c *OpenChannel) ChanSyncMsg() (*lnwire.ChannelReestablish, error) {
 	c.Lock()
 	defer c.Unlock()
 
@@ -853,7 +850,7 @@ func (c *OpenChannel) ChanSyncMsg(
 	// If we've restored this channel, then we'll purposefully give them an
 	// invalid LocalUnrevokedCommitPoint so they'll force close the channel
 	// allowing us to sweep our funds.
-	if isRestoredChan {
+	if c.hasChanStatus(ChanStatusRestored) {
 		currentCommitSecret[0] ^= 1
 	}
 
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -108,6 +108,10 @@ var (
 	// in the database.
 	ErrNoCommitPoint = fmt.Errorf("no commit point found")
 
+	// ErrNoCloseTx is returned when no closing tx is found for a channel
+	// in the state CommitBroadcasted.
+	ErrNoCloseTx = fmt.Errorf("no closing tx found")
+
 	// ErrNoRestoredChannelMutation is returned when a caller attempts to
 	// mutate a channel that's been recovered.
 	ErrNoRestoredChannelMutation = fmt.Errorf("cannot mutate restored " +
@@ -900,6 +904,37 @@ func (c *OpenChannel) MarkCommitmentBroadcasted(closeTx *wire.MsgTx) error {
 	}
 
 	return c.putChanStatus(ChanStatusCommitBroadcasted, putClosingTx)
+}
+
+// BroadcastedCommitment retrieves the stored closing tx set during
+// MarkCommitmentBroadcasted. If not found ErrNoCloseTx is returned.
+func (c *OpenChannel) BroadcastedCommitment() (*wire.MsgTx, error) {
+	var closeTx *wire.MsgTx
+
+	err := c.Db.View(func(tx *bbolt.Tx) error {
+		chanBucket, err := fetchChanBucket(
+			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
+		)
+		switch err {
+		case nil:
+		case ErrNoChanDBExists, ErrNoActiveChannels, ErrChannelNotFound:
+			return ErrNoCloseTx
+		default:
+			return err
+		}
+
+		bs := chanBucket.Get(closingTxKey)
+		if bs == nil {
+			return ErrNoCloseTx
+		}
+		r := bytes.NewReader(bs)
+		return ReadElement(r, &closeTx)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return closeTx, nil
 }
 
 // putChanStatus appends the given status to the channel. fs is an optional
--- a/contractcourt/chain_arbitrator.go
+++ b/contractcourt/chain_arbitrator.go
@@ -412,6 +412,36 @@ func (c *ChainArbitrator) Start() error {
 		}
 
 		c.activeChannels[chanPoint] = channelArb
+
+		// If the channel has had its commitment broadcasted already,
+		// republish it in case it didn't propagate.
+		if !channel.HasChanStatus(
+			channeldb.ChanStatusCommitBroadcasted,
+		) {
+			continue
+		}
+
+		closeTx, err := channel.BroadcastedCommitment()
+		switch {
+
+		// This can happen for channels that had their closing tx
+		// published before we started storing it to disk.
+		case err == channeldb.ErrNoCloseTx:
+			log.Warnf("Channel %v is in state CommitBroadcasted, "+
+				"but no closing tx to re-publish...", chanPoint)
+			continue
+
+		case err != nil:
+			return err
+		}
+
+		log.Infof("Re-publishing closing tx(%v) for channel %v",
+			closeTx.TxHash(), chanPoint)
+		err = c.cfg.PublishTx(closeTx)
+		if err != nil && err != lnwallet.ErrDoubleSpend {
+			log.Warnf("Unable to broadcast close tx(%v): %v",
+				closeTx.TxHash(), err)
+		}
 	}
 
 	// In addition to the channels that we know to be open, we'll also
--- a/contractcourt/chain_arbitrator_test.go
+++ b/contractcourt/chain_arbitrator_test.go
@@ -1,1 +1,117 @@
 package contractcourt
+
+import (
+	"io/ioutil"
+	"net"
+	"os"
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwallet"
+)
+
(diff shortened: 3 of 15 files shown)


channeldb/channel.go
```diff
@@ -107,6 +107,10 @@ var (
 	// in the database.
 	ErrNoCommitPoint = fmt.Errorf("no commit point found")
 
+	// ErrNoCloseTx is returned when no clsing tx is found for a channel in
```
Roasbeef: clsing -> closing

contractcourt/chain_arbitrator_test.go
```diff
@@ -1 +1,277 @@
...
+	key = [chainhash.HashSize]byte{
+		0x81, 0xb6, 0x37, 0xd8, 0xfc, 0xd2, 0xc6, 0xda,
+		0x68, 0x59, 0xe6, 0x96, 0x31, 0x13, 0xa1, 0x17,
+		0xd, 0xe7, 0x93, 0xe4, 0xb7, 0x25, 0xb8, 0x4d,
+		0x1e, 0xb, 0x4c, 0xf9, 0x9e, 0xc5, 0x8c, 0xe9,
+	}
+
+	privKey, pubKey = btcec.PrivKeyFromBytes(btcec.S256(), key[:])
+	testTx          = &wire.MsgTx{
+		Version: 1,
```
Roasbeef: FWIW, the `codec.go` file in `channeldb` assumes that all transactions are version 2.

contractcourt/chain_arbitrator_test.go
```diff
@@ -1 +1,277 @@
...
+					0xa6, // 65-byte signature
+					0xac, // OP_CHECKSIG
+				},
+			},
+		},
+		LockTime: 5,
+	}
+)
+
+func createTestChannelState(cdb *channeldb.DB) (*channeldb.OpenChannel, error) {
```
Roasbeef: I think this can use [`lnwallet.CreateTestChannels`](https://godoc.org/github.com/lightningnetwork/lnd/lnwallet#CreateTestChannels).

contractcourt/chain_arbitrator.go
```diff
@@ -397,6 +397,32 @@ func (c *ChainArbitrator) Start() error {
...
+			log.Warn("Channel %v is in state CommitBroadcasted, "+
+				"but no closing tx to re-publish...", chanPoint)
+			continue
+
+		case err != nil:
+			return err
+		}
+
+		log.Infof("Re-publishing closing tx for channel %v", chanPoint)
+		err = c.cfg.PublishTx(closeTx)
```
Roasbeef: Do we also need to manually set the channel arb's state to be `StateCommitmentBroadcasted `?

halseth: This shouldn't be necessary, since the `chanArb` is already handling any closing tx appearing in chain regardless of the current state.

Worst case we save the tx before committing to `StateCommitmentBroadcasted`, and re-publish on startup. When the tx eventually confirms, we will treat it the same way as a "surprising" remote force close.

contractcourt/chain_arbitrator_test.go
halseth: Interesting, any reason we are still using version 1 some places? (e.g. https://github.com/lightningnetwork/lnd/blob/985902be2779a5dcc7ac1d5b3d7eb7929608816d/lnwallet/wallet.go#L688)

channeldb/channel.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/d64361369a6f103e7d1a2699e968dbc2b37942c9

contractcourt/chain_arbitrator_test.go
halseth: Needed to make a slight change to it (https://github.com/lightningnetwork/lnd/pull/3016/commits/55212de7e19aa19e91136a1b8294e2dcd225f900), and was then able to use it: https://github.com/lightningnetwork/lnd/pull/3016/commits/4686e5c6511efddaa7a2200fedd44277164d6251

halseth:
Comments addressed. PTAL @Roasbeef

contractcourt/chain_arbitrator.go
```diff
@@ -397,6 +397,32 @@ func (c *ChainArbitrator) Start() error {
...
+
+		closeTx, err := channel.BroadcastedCommitment()
+		switch {
+
+		// This can happen for channels that had their closing tx
+		// published before we started storing it to disk.
+		case err == channeldb.ErrNoCloseTx:
+			log.Warn("Channel %v is in state CommitBroadcasted, "+
+				"but no closing tx to re-publish...", chanPoint)
+			continue
```
wpaulino: Any reason we don't craft a new closing transaction in this case?

lnwallet/test_utils.go
```diff
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"io"
 	"io/ioutil"
+	rrand "math/rand"
```
wpaulino: Nit: this is usually `prand` within the codebase.

contractcourt/chain_arbitrator.go
halseth: We don't really know which tx was originally broadcasted.

That being said, it wouldn't hurt to attempt re-creating and broadcasting a force close transaction. Would that solve any edge case you think?

lnwallet/test_utils.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/d390428983891065fd8c6cbb2bd3af625b548e08

contractcourt/chain_arbitrator_test.go
Roasbeef: Mostly just legacy code really. We should likely update the `wire` package to always use version 2.

contractcourt/chain_arbitrator.go
Roasbeef: This is a state that should never really happen. It seems a bit dangerous to attempt to craft a new transaction as we may not be 100% consistent here.

contractcourt/channel_arbitrator.go
```diff
@@ -721,6 +721,15 @@ func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
 		}
 		closeTx = closeSummary.CloseTx
 
+		// Before publishing the closing tx, we persist it to the
```
Roasbeef: I think there's a gap we're still missing. When we call the `ForceCloseChan` method above, we'll go to mark the channel as borked: https://github.com/lightningnetwork/lnd/blob/master/contractcourt/chain_arbitrator.go#L236. However, we don't specify any additional context here to convey that we're about to force close the channel. As a result, when we come back up, we won't know to trigger the force close again as there's no force close transaction present. 

I'm on the fence w.r.t if we need to care about this scenario atm. This starts to get into the greater issue w.r.t reliably carrying out the DLP steps, since it also applies to all the calls to `MarkBorked` in this method: https://github.com/lightningnetwork/lnd/blob/master/lnwallet/channel.go#L3211 

If we provide the context there, then we'll be able to handle failures at each step of the force close (and also DLP) flow:
  * mark channel action (borked) -> (1) mark commitment broadcast and commit transaction -> (2) broadcast transaction -> (3) update state 

The state written at step 1 should allow us to determine if we need to act (remote party data loss DLP and our local force close), or just halt and wait (local DLP loss). Any auxiliary data such as a channel close summary, or the last chan-reest message sent should also be written at that point. In the DLP case, we need to write the message up front as otherwise when we go to look for it, it won't be present until the channel has been close on chain fully.

Roasbeef:
> Is it possible to fix this issue without adding more persistent state? Maybe by doing a bit more republishing, possibly redundant, like in the sweeper?

Yes, but it needs to store this transaction _somewhere_. For channels that are closing or closed, it currently isn't stored anywhere. We can't reconstruct this data by ourselves since we need the signature of the remote node along side it.

halseth:
Yes, the idea behind storing the tx was to handle all possible closes uniformly. Our own force close tx we can re-derive, but not coop closes. It would also allow us in the future to store the remote force close tx if we retrieve it.

Roasbeef:
So with that said, I think we can move forward on this? Or are we still unclear w.r.t the goals here?

joostjager:
My question about why db is answered.

contractcourt/channel_arbitrator.go
halseth: You're right, there are several cases like this, and feels like they should be handled in a general way.

We could make all the `MarkChanX` methods take an optional close tx, such that we can provide it for the cases where we know of a safe tx to publish. Alternatively we could make the borked state imply that we deem it safe to publish our latest commitment.

Roasbeef:
> Is it possible to fix this issue without adding more persistent state? Maybe by doing a bit more republishing, possibly redundant, like in the sweeper?

We can reconstruct our commitment transaction for force closes, but we can't reconstruct the co-op close transaction since we need a new signature from the remote party. So then the question is should we have a unified location to check for closing transaction, or special case the possible closing paths. At the very least, we need a new key in the database to know what _type_ of close was attempted and where we are in that process it seems.

Roasbeef:
Another open question is if we deem the level of fix implement in this PR acceptable to move forward with, or if we'd like to step back and resolve the other issues as well I outlined in an earlier comment.

contractcourt/channel_arbitrator.go
```diff
@@ -724,6 +724,15 @@ func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
...
+		// database, such that it can be republished if something goes
+		// wrong.
+		if err := c.cfg.MarkCommitmentBroadcasted(closeTx); err != nil {
+			log.Errorf("ChannelArbitrator(%v): unable to "+
+				"mark commitment broadcasted: %v",
+				c.cfg.ChanPoint, err)
```
cfromknecht: feels like the error be returned here?

halseth:
Pushed a new version which makes sure we store the context necessary when changing the channel status in the DB. Instead of calling `MarkBorked` when we attempt to force close, we'll now call `MarkCommitmentBroadcasted` immediately, with out closing tx, such that it will be re-published on startup.

Calling `MarkBorked` should now only be done in cases where we don't know if it is safe to force close the channel.

PTAL @Roasbeef @cfromknecht @wpaulino

htlcswitch/link.go
```diff
@@ -879,6 +879,13 @@ func (l *channelLink) htlcManager() {
 			// what they sent us before.
 			// TODO(halseth): ban peer?
 			case err == lnwallet.ErrInvalidLocalUnrevokedCommitPoint:
+				err = l.channel.MarkBorked()
+				if err != nil {
+					log.Errorf("Unable to mark channel "+
+						"borked: %v", err)
+					return
```
wpaulino: Why do we return here? Shouldn't we proceed to fail the link?

htlcswitch/link.go
```diff
@@ -887,20 +908,34 @@ func (l *channelLink) htlcManager() {
...
-			// when the remote closes the channel.
+			// latest unrevoked commitment point, and we'll store
+			// it in the database, such that we can attempt to
+			// recover the funds if the remote force closes the
+			// channel.
 			// TODO(halseth): mark this, such that we prevent
 			// channel from being force closed by the user or
 			// contractcourt etc.
-			case err == lnwallet.ErrCommitSyncLocalDataLoss:
+			case localDataLoss:
```
wpaulino: I think the TODO above can be removed since `LightningChannel.ForceClose` will check whether it has a `ChanStatusLocalDataLoss` status before doing so.

htlcswitch/link.go
```diff
@@ -874,6 +879,22 @@ func (l *channelLink) htlcManager() {
...
 			// TODO(halseth): ban peer?
 			case err == lnwallet.ErrInvalidLocalUnrevokedCommitPoint:
+				forceClose, err := l.channel.ForceClose()
+				if err != nil {
+					log.Errorf("Unable to force close "+
+						"channel: %v", err)
+					return
+				}
+
+				err = l.channel.MarkCommitmentBroadcasted(
```
wpaulino: Do we need to do all of this now given that we fail the link and signal that it should be force closed below? So it'll do all of this, but at the contract court level.

lnwallet/channel.go
```diff
@@ -101,6 +91,24 @@ var (
...
+// channel and MUST NOT force close it. Instead we should wait for the remote
+// to force close it, such that we can attempt to sweep our funds. The
+// commitment point needed to sweep the remote's force close is encapsuled.
+type ErrCommitSyncLocalDataLoss struct {
+	// CommitPoint is the last unrevoked commit point, sent to us by the
+	// remote when we determined we had lost state.
+	CommitPoint *btcec.PublicKey
+}
+
+func (e *ErrCommitSyncLocalDataLoss) Error() string {
```
Roasbeef: Missing `godoc` comment. Can also log the serialized commit point in this error as well, and maybe also the channel point while we're at it?

htlcswitch/link.go
Roasbeef: Yeah, within `handleLinkFailure` (peer.go) we handle all the clean up. As a result, we only need to modify this closure: https://github.com/lightningnetwork/lnd/blob/master/contractcourt/chain_arbitrator.go#L234

Recall that care was taken to ensure that no link is active, and no new link can be loaded in while we're carrying out the force close, to avoid an accidental self breach. We may need to modify the ordering here slightly,  to ensure that we're able to carry out the restart properly. Otherwise, if we crash after that fist call, then the channel is marked as borked, so it won't be resumed by the chain arb.

contractcourt/chain_arbitrator.go
```diff
@@ -232,12 +232,10 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
...
-			// First, we mark the channel as borked, this ensure
-			// that no new state transitions can happen, and also
-			// that the link won't be loaded into the switch.
-			if err := channel.MarkBorked(); err != nil {
-				return nil, err
-			}
+
+			// TODO: Must BORK first to avoid any updates? Or get a
+			// lock ensuring no updates won't be made before
+			// marking commitment broadcasted.
```
Roasbeef: We did this so the link wouldn't be loaded in again while we were force closing. If we remove this, then as we're force closing, then it's possible that the peer re-connects (after the disable below), and initiates a state transition.

contractcourt/chain_watcher.go
```diff
@@ -908,6 +908,8 @@ func (c *chainWatcher) dispatchContractBreach(spendEvent *chainntnfs.SpendDetail
 		"ChannelPoint(%v). Revoked state #%v was broadcast!!!",
 		c.cfg.chanState.FundingOutpoint, broadcastStateNum)
 
+	// TODO: store remote breach tx here also?
+	// TODO: needed? do the same for remote force closes?
```
Roasbeef: Why would we need to do this?

Roasbeef:
OP bump.

contractcourt/chain_arbitrator.go
```diff
@@ -412,6 +412,32 @@ func (c *ChainArbitrator) Start() error {
...
+		case err == channeldb.ErrNoCloseTx:
+			log.Warn("Channel %v is in state CommitBroadcasted, "+
+				"but no closing tx to re-publish...", chanPoint)
+			continue
+
+		case err != nil:
+			return err
+		}
+
+		log.Infof("Re-publishing closing tx for channel %v", chanPoint)
```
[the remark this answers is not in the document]
halseth: Done.

htlcswitch/link.go
halseth: Yep, done!

htlcswitch/link.go
halseth: Done!

lnwallet/channel.go
halseth: Done

contractcourt/chain_watcher.go
halseth: Removed.

htlcswitch/link.go
halseth: Removed and added comment

contractcourt/chain_arbitrator.go
halseth: This is a bit tricky, since the two obvious ways here both have a gap:
1. First mark the channel borked, get the force close tx and write it to the DB:
* If we go down after marking it borked then we don't know at startup that we want to force close it.
2. Get the force close tx, then store it atomically to the db. 
* We don't really have a guarantee that no state updates happen between getting the tx and storing it to the db.

The first way to solve this that comes to mind, is for `MarkBorked` to take an "intention" argument (or just add a new state `MarkLocalForceClose`) such that we know on restart that we are to force close the channel.

halseth:
Pushed a new version, where the main difference is that we will resend `channel reestablish` messages for all borked channels with a peer on connect. 

> 1. The peer that lost data should still re-send their chan reest message upon each connection until the channel has been force closed. Otherwise, it's possible to recognize we lost data, but the remote peer (say it's unupgraded) crashes before it processes our message. In this case, we've marked it as borked locally, so we won't load it back into the link

With the above mentioned change, any channel having state `LocalDataLoss` we'll resend the chan sync message for when connecting to the peer.

> 2. We should write the chan reest message as soon as we initiate a DLP for an eroded client (lost data). This way we can send the message upon their next reconnection, rather than until the channel is fully confirmed.

Exactly as above, for channels where we have recognized remote data loss, we'll resend the chan sync message on reconnect.

Main question left os how to handle the atomic transition when force closing a channel (in `chain_arbitrator`). See discussion there.

contractcourt/chain_arbitrator.go
halseth: Go with new state

channeldb/channel.go
```diff
@@ -810,6 +811,85 @@ func (c *OpenChannel) MarkBorked() error {
...
+//      it.
+//   3. We didn't get the last RevokeAndAck message they sent, so they'll
+//      re-send it.
+//
+// The isRestoredChan bool indicates if we need to craft a chan sync message
+// for a channel that's been restored. If this is a restored channel, then
+// we'll modify our typical chan sync  message to ensure they force close even
+// if we're on the very first state.
+func (c *OpenChannel) ChanSyncMsg(
+	isRestoredChan bool) (*lnwire.ChannelReestablish, error) {
```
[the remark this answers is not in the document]
halseth: Yep, good catch! https://github.com/lightningnetwork/lnd/pull/3016/commits/20eb4b7722314b80aa9e8f3bacb910d771ada50b

peer.go
```diff
@@ -359,10 +359,20 @@ func (p *peer) Start() error {
...
+	msgs, err := p.loadActiveChannels(activeChans)
+	if err != nil {
 		return fmt.Errorf("unable to load channels: %v", err)
 	}
 
+	if len(msgs) > 0 {
+		peerLog.Infof("Sending %d channel sync messages to peer after "+
+			"loading active channels", len(msgs))
+		if err := p.SendMessage(true, msgs...); err != nil {
+			return fmt.Errorf("Failed sending channel sync "+
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/631cb6483d6f217e1c9aa456c61157e299369cdc

peer.go
```diff
@@ -437,6 +453,21 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
...
+			// To help our peer recover from a potential data loss,
+			// we resend our channel reestablish message if the
+			// channel is in a borked state. We won't process any
+			// channel reestablish message sent from the peer, but
+			// that's okay since the assumption is that we did when
+			// marking the channel borked.
+			chanSync, err := dbChan.ChanSyncMsg(false)
+			if err != nil {
+				peerLog.Errorf("Unable to create channel "+
+					"reestablish message: %v", err)
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/631cb6483d6f217e1c9aa456c61157e299369cdc

contractcourt/chain_arbitrator.go
```diff
@@ -412,6 +423,34 @@ func (c *ChainArbitrator) Start() error {
...
+			continue
+		}
+
+		closeTx, err := channel.BroadcastedCommitment()
+		switch {
+
+		// This can happen for channels that had their closing tx
+		// published before we started storing it to disk.
+		case err == channeldb.ErrNoCloseTx:
+			log.Warn("Channel %v is in state CommitBroadcasted, "+
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/17cb9448073716b6425e51b33b5bf7a974ab32ae

contractcourt/chain_arbitrator_test.go
```diff
@@ -1 +1,113 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwallet"
+)
+
+// TestChainArbitratorRepulishCommitment testst that the chain arbitrator will
+// republish closing transactions for channels marked CommitementBroadcast in
+// the database at startup.
+func TestChainArbitratorRepublishCommitment(t *testing.T) {
```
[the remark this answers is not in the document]
halseth: I prefer keeping them separate  both for commit size and easy cherry-picking.

lnwallet/channel.go
```diff
@@ -101,6 +91,30 @@ var (
...
+	ChannelPoint wire.OutPoint
+
+	// CommitPoint is the last unrevoked commit point, sent to us by the
+	// remote when we determined we had lost state.
+	CommitPoint *btcec.PublicKey
+}
+
+// Error returns a string representation of the local data loss error.
+func (e *ErrCommitSyncLocalDataLoss) Error() string {
+	return fmt.Sprintf("ChannelPoint(%v) had possible local commitment "+
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/5656a7a2cfe796484227a804a3e999d787b44a3d

contractcourt/channel_arbitrator.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3016/commits/37253551e8b6278b33863a7d50566182f21701e7

contractcourt/chain_arbitrator.go
halseth: When adding this new state, I realized that it is actually already encoded within the `ChannelArbitrator`'s state machine. When we trigger a force close it will commit to `StateBroadcastCommit`, and in this state mark the channel borked, get the force close tx, and then finally mark the commitment broadcasted. 

This means that the scenario 1) I outlined above is handled by the `ChannelArbitrator` itself, since after a restart it will be in `StateBroadcastCommit` and republish the force close tx. So we should be okay.

wpaulino:
Travis fails with:
```
        panic: runtime error: invalid memory address or nil pointer dereference
        [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xac4d07]
        
        goroutine 1 [running]:
        github.com/lightningnetwork/lnd/contractcourt.(*ChannelArbitrator).stateStep(0xc0000c41e0, 0x5db, 0x0, 0xc0002d1530, 0x3, 0x3, 0x14f7ec0)
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/contractcourt/channel_arbitrator.go:806 +0x287
        github.com/lightningnetwork/lnd/contractcourt.(*ChannelArbitrator).advanceState(0xc0000c41e0, 0x5db, 0x0, 0xc0000a7c01, 0x3, 0x3, 0x700000000000000)
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/contractcourt/channel_arbitrator.go:1042 +0x19d
        github.com/lightningnetwork/lnd/contractcourt.(*ChannelArbitrator).Start(0xc0000c41e0, 0xc0000181e0, 0xc0009373d8)
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/contractcourt/channel_arbitrator.go:387 +0x4a5
        github.com/lightningnetwork/lnd/contractcourt.(*ChainArbitrator).Start(0xc000144360, 0x0, 0x0)
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/contractcourt/chain_arbitrator.go:543 +0x12e6
        github.com/lightningnetwork/lnd.(*server).Start.func1()
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/server.go:1215 +0x430
        sync.(*Once).Do(0xc000146488, 0xc000937880)
        	/home/travis/.gimme/versions/go1.12.9.linux.amd64/src/sync/once.go:44 +0xb3
        github.com/lightningnetwork/lnd.(*server).Start(0xc000146480, 0x1396a30, 0xc0000cd040)
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/server.go:1151 +0x6c
        github.com/lightningnetwork/lnd.Main(0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/lnd.go:616 +0x1c92
        main.main()
        	/home/travis/gopath/src/github.com/lightningnetwork/lnd/cmd/lnd/main.go:14 +0x32
```

Looks like we'll need to set the `ForceCloseChan` field of a `ChannelArbitrator` for pending close channels.

peer.go
```diff
@@ -362,6 +363,17 @@ func (p *peer) Start() error {
 	go p.channelManager()
 	go p.pingHandler()
 
+	// Now that the peer has started up, we send any channel sync messages
+	// that must be resent for borked channels.
+	if len(msgs) > 0 {
+		peerLog.Infof("Sending %d channel sync messages to peer after "+
+			"loading active channels", len(msgs))
+		if err := p.SendMessage(true, msgs...); err != nil {
```
wpaulino: We should remove `resendChanSyncMsg` to prevent sending it twice upon reconnections.

halseth: `resendChanSyncMsg` is only used for closed channels, here we send for borked, open channels.

go.mod
```diff
@@ -58,6 +58,8 @@ replace github.com/lightningnetwork/lnd/ticker => ./ticker
 
 replace github.com/lightningnetwork/lnd/queue => ./queue
 
+replace github.com/btcsuite/btcwallet => github.com/halseth/btcwallet v0.0.0-20190919123305-ca06abc72b9c
```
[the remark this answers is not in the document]
halseth: Not yet, see #3510

lnwallet/btcwallet/btcwallet.go
```diff
@@ -429,47 +429,29 @@ func (b *BtcWallet) ListUnspentWitness(minConfs, maxConfs int32) (
...
 
-		case *chain.NeutrinoClient:
-			if strings.Contains(err.Error(), "already spent") {
-				// Output was already spent.
-				return lnwallet.ErrDoubleSpend
-			}
+		switch {
+		// If the wallet reports a double spend, convert it to our
+		// internal ErrDoubleSpend and return.
+		case errors.As(err, &doubleSpendErr):
```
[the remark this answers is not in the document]
halseth: See #3510

htlcswitch/link.go
```diff
@@ -592,21 +592,6 @@ func (l *channelLink) syncChanStates() error {
...
 	}
 
-	// If we have a restored channel, we'll delay sending our channel
-	// reestablish message briefly to ensure we first have a stable
-	// connection. Sending the message will cause the remote peer to force
-	// close the channel, which currently may not be resumed reliably if the
-	// connection is being torn down simultaneously. This delay can be
-	// removed after the force close is reliable, but in the meantime it
-	// improves the reliability of successfully closing out the channel.
-	if chanState.HasChanStatus(channeldb.ChanStatusRestored) {
```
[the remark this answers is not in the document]
halseth: Flake fixed by https://github.com/lightningnetwork/lnd/pull/3016/commits/70a0c546bb70664c59b7bfceca5996b4bd690be0 😄

lntest/itest/lnd_test.go
```diff
@@ -13526,6 +13528,21 @@ func testChanRestoreScenario(t *harnessTest, net *lntest.NetworkHarness,
...
+		balReq := &lnrpc.WalletBalanceRequest{}
+		daveBalResp, err := dave.WalletBalance(ctxt, balReq)
+		if err != nil {
+			return false
+		}
+
+		return daveBalResp.ConfirmedBalance > 0
+	}, defaultTimeout)
+	if err != nil {
+		t.Fatalf("On-chain balance not restored: %v", err)
```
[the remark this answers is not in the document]
halseth: It will most likely just be zero, but fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/f4786abc54e3af067cc677701777eedb5e4faa27 😅

channeldb/channel.go
```diff
@@ -902,6 +906,41 @@ func (c *OpenChannel) MarkCommitmentBroadcasted(closeTx *wire.MsgTx) error {
...
+		default:
+			return err
+		}
+
+		bs := chanBucket.Get(closingTxKey)
+		if bs == nil {
+			return ErrNoCloseTx
+		}
+		r := bytes.NewReader(bs)
+		if err := ReadElements(r, &closeTx); err != nil {
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3016/commits/9482d7197f3345932e2a5bb70e2acd093013079f

peer.go
wpaulino: Any reason to not unify those with these?

peer.go
```diff
@@ -401,13 +413,19 @@ func (p *peer) QuitSignal() <-chan struct{} {
 
 // loadActiveChannels creates indexes within the peer for tracking all active
 // channels returned by the database.
-func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
+func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) (
```
Roasbeef: `godoc` should be updated to detail the new return value

peer.go
```diff
@@ -427,6 +445,22 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) error {
...
 			peerLog.Warnf("ChannelPoint(%v) has status %v, won't "+
 				"start.", chanPoint, dbChan.ChanStatus())
+
+			// To help our peer recover from a potential data loss,
+			// we resend our channel reestablish message if the
+			// channel is in a borked state. We won't process any
+			// channel reestablish message sent from the peer, but
+			// that's okay since the assumption is that we did when
+			// marking the channel borked.
+			chanSync, err := dbChan.ChanSyncMsg(false)
```
Roasbeef: Why `false` here? We don't have enough context to know if it was recovered or not. 

Nvm scratch that.

peer.go
halseth: Done

halseth:
> However, neutrino itest fails with:
> 
> ```
>     --- FAIL: TestLightningNetworkDaemon/data_loss_protection (13.37s)
>         lnd_test.go:100: Failed: (data loss protection): exited with error: 
>             *errors.errorString unable to connect Dave to carol: rpc error: code = Unknown desc = dial tcp 127.0.0.1:19799: connect: connection refused
>             /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:8213 (0xd4c4a3)
> ```
> 
> Is this unrelated, do we have any idea what is causing the failure? It seems distinct from the prior lack of force close related errors due to the inconsistency/race-condition.

Yeah, I saw this flake pretty regularly when working on the initial neutrino itests. It was caused by some connection replacement in the peer/connmgr logic. It went away mostly with the connection manager changes, but maybe it is still not totally robust yet.

I gave it a squash, but still needs #3510 before final swuash an merge

halseth:
> > However, neutrino itest fails with:
> > ```
> >     --- FAIL: TestLightningNetworkDaemon/data_loss_protection (13.37s)
> >         lnd_test.go:100: Failed: (data loss protection): exited with error: 
> >             *errors.errorString unable to connect Dave to carol: rpc error: code = Unknown desc = dial tcp 127.0.0.1:19799: connect: connection refused
> >             /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:8213 (0xd4c4a3)
> > ```
> > 
> > 
> > Is this unrelated, do we have any idea what is causing the failure? It seems distinct from the prior lack of force close related errors due to the inconsistency/race-condition.
> 
> Yeah, I saw this flake pretty regularly when working on the initial neutrino itests. It was caused by some connection replacement in the peer/connmgr logic. It went away mostly with the connection manager changes, but maybe it is still not totally robust yet.

Just hit this also on the travis run for #3510  https://travis-ci.org/lightningnetwork/lnd/jobs/589294404, so it is existing.

lntest/itest/lnd_test.go
```diff
@@ -13559,6 +13561,27 @@ func testChanRestoreScenario(t *harnessTest, net *lntest.NetworkHarness,
...
+	err = wait.NoError(func() error {
+		ctxt, _ := context.WithTimeout(ctxb, defaultTimeout)
+		balReq := &lnrpc.WalletBalanceRequest{}
+		daveBalResp, err := dave.WalletBalance(ctxt, balReq)
+		if err != nil {
+			return err
+		}
+
+		daveBal := daveBalResp.ConfirmedBalance
+		if daveBal <= 0 {
```
wpaulino: Can we provide a better bound to ensure Dave has recovered all of his balance?

Issue #986, opened 2018-03-31
Autopilot can create impossible channel requests, which fail due to occupying funds without room for transaction fees


Empact:
### Background

If you configure autopilot to control the entire wallet, it will attempt to open channels with all available funds, which will result in destined-to-fail openChannel directives due to the need to pay fees on opening

### Your environment

* lnd 0.4.0, but also present in master
* mac os
* bitcoind

### Steps to reproduce

Set `autopilot.allocation=1`
Have enough funds for < `autopilot.maxchannels` channels in the wallet.

### Expected behaviour

Autopilot should be able to control the full wallet and account for fees in the final channel opening directive

### Actual behaviour

```
2018-03-31 04:04:09.657 [INF] ATPL: Triggering attachment directive dispatch, total_funds=0.02493046 BTC
2018-03-31 04:04:10.729 [INF] ATPL: Attempting to execute channel attachment directives: ([]autopilot.AttachmentDirective) (len=1 cap=1) {
 (autopilot.AttachmentDirective) {
  PeerKey: (*btcec.PublicKey)(0xXXXX)({
   Curve: (elliptic.Curve) <nil>,
   X: (*big.Int)(0x)(XXXXX),
   Y: (*big.Int)(0x)(YYYYY)
  }),
  ChanAmt: (btcutil.Amount) 0.02493046 BTC,
  Addrs: ([]net.Addr) (len=1 cap=1) {
   (*net.TCPAddr)(0x)(X.X.X.X)
  }
 }
}

2018-03-31 04:04:11.856 [INF] FNDG: Initiating fundingRequest(localAmt=0.02493046 BTC, remoteAmt=0 mSAT, capacity=0.02493046 BTC, chainhash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f, addr=X.X.X.X, dustLimit=0.00000573 BTC)
2018-03-31 04:04:11.858 [INF] LNWL: Performing funding tx coin selection using 4 sat/vbyte as fee rate
2018-03-31 04:04:11.858 [WRN] ATPL: Unable to open channel to XXX of 0.02493046 BTC: not enough witness outputs to create funding transaction, need 0.02493658 BTC only have 0.02493046 BTC  available
```

Empact:
Seems like the Heuristics need to try to account for fees, or LND should have a convention whereby the fees are taken out of the channel fund allocation, rather than added to it. The latter seems simpler to me.

[...]

PR #3017, opened 2019-04-26
[autopilot] Fee aware autopilot

halseth:
Previously the autopilot wouldn't take fees into account when opening
channels, causing it to have insufficient funds when trying to utilize
close to all funds in the wallet.

This commit uses the FeeEstimate method to get a rough estimate (we
allow some wiggle room) for the fees needed to open the channels. Since
the total fee will change with the number of channels being opened
(potentially we can open channels in the same tx in the future), we
re-calculate the total fee each time we find a new potential channel
candidate, and resize the last channel to fit it within the budget with
the fee.

Fixes https://github.com/lightningnetwork/lnd/issues/986

TODO:
- [x] unit test for autopilot agent

### DIFF
--- a/autopilot/agent.go
+++ b/autopilot/agent.go
@@ -596,6 +596,7 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 	}
 
 	chanCandidates := make(map[NodeID]*AttachmentDirective)
+	var chanSizes []btcutil.Amount
 	for nID := range scores {
 		// Add addresses to the candidates.
 		addrs := addresses[nID]
@@ -606,22 +607,52 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 			continue
 		}
 
-		// Track the available funds we have left.
-		if availableFunds < chanSize {
-			chanSize = availableFunds
+		for {
+			// Get the fee estimated we need to open this channel
+			// in addition to the already selected ones.
+			feeEstimate, err := a.cfg.ChanController.FeeEstimate(
+				append(chanSizes, chanSize),
+			)
+			if err != nil {
+				return err
+			}
+
+			// Give us some wiggle room.
+			feeEstimate *= 3
+
+			// If we have enough balance to cover the fee, we can
+			// break.
+			if chanSize+feeEstimate <= availableFunds {
+				break
+			}
+
+			// Otherwise we need to decrease our channels size and
+			// get another estimate. We fetch another estimate
+			// since a lower channel size potentially lets us use
+			// less inputs and get away with a smaller fee.
+			chanSize = availableFunds - feeEstimate
+			log.Debugf("Resizing channel size to %v - %v = %v "+
+				"to account for fees", availableFunds,
+				feeEstimate, chanSize)
 		}
-		availableFunds -= chanSize
 
 		// If we run out of funds, we can break early.
 		if chanSize < a.cfg.Constraints.MinChanSize() {
 			break
 		}
 
+		// Track the available funds we have left.
+		availableFunds -= chanSize
+
 		chanCandidates[nID] = &AttachmentDirective{
 			NodeID:  nID,
 			ChanAmt: chanSize,
 			Addrs:   addrs,
 		}
+
+		// We track the size of channels we intend to open, to make fee
+		// estimation as we go.
+		chanSizes = append(chanSizes, chanSize)
 	}
 
 	if len(chanCandidates) == 0 {
--- a/pilot.go
+++ b/pilot.go
@@ -3,6 +3,7 @@ package lnd
 import (
 	"errors"
 	"fmt"
+	"math"
 	"net"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -10,6 +11,8 @@ import (
 	"github.com/btcsuite/btcutil"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/lightningnetwork/lnd/autopilot"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/tor"
 )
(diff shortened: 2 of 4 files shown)


autopilot/agent.go
```diff
@@ -606,22 +607,51 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
+			if chanSize+feeEstimate <= availableFunds {
+				break
+			}
+
+			// Otherwise we need to decrease our channels size and
+			// get another estimate.
+			chanSize = availableFunds - feeEstimate
+			log.Debugf("Resizing channel size to %v - %v = %v "+
+				"to account for fees", availableFunds,
+				feeEstimate, chanSize)
```
wpaulino: Would add a comment here to make it clear we continue to ensure we retry the fee estimate as it's possible that we can use less inputs due to the `chanSize` being lower.

autopilot/agent.go
```diff
@@ -606,22 +607,51 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
+			chans := append(chanSizes, chanSize)
+			feeEstimate, err := a.cfg.ChanController.FeeEstimate(
+				chans,
+			)
+			if err != nil {
+				return err
+			}
+
+			// Give us some wiggle room.
+			feeEstimate *= 3
```
wpaulino: Why not 2X? 😛

pilot.go
```diff
@@ -134,6 +137,74 @@ func (c *chanController) SpliceOut(chanPoint *wire.OutPoint,
...
+			coin := coins[currentCoin]
+			currentCoin++
+
+			switch coin.AddressType {
+			case lnwallet.WitnessPubKey:
+				weightEstimate.AddP2WKHInput()
+			case lnwallet.NestedWitnessPubKey:
+				weightEstimate.AddNestedP2WKHInput()
+			default:
+				return 0, fmt.Errorf("Unsupported address: %v",
```
wpaulino: Nit: returned error should start with lowercase.

autopilot/agent.go
halseth: NO2X!!

pilot.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3017/commits/ca4fc621f3a8bd818f981b4262c251f357e1cc61

autopilot/agent.go
```diff
@@ -606,22 +607,51 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
+
+			// If we have enough balance to cover the fee, we can
+			// break.
+			if chanSize+feeEstimate <= availableFunds {
+				break
+			}
+
+			// Otherwise we need to decrease our channels size and
+			// get another estimate.
+			chanSize = availableFunds - feeEstimate
```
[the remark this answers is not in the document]
halseth: `chans` is just a local variable that was never reused. Removed it entirely to make it more clear: https://github.com/lightningnetwork/lnd/pull/3017/commits/7c2c77176fc7564c44413dc2ab21280ebc899353

autopilot/agent.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3017/commits/ed7b36bfc86a13b78e91ff28ff4eaef4a9a12d63

halseth:
> Not really a fan of estimating fees this way as it can be inaccurate, but can see how it's an intermediate solution until things are refactored better.

Yeah, it is not the best at the moment since we cannot lock in UTXOs and channels are being opened in parallel. I expect this to be improved when we start batching channel openings. At least this let us work around this issue for now :)

PTAL @wpaulino

wpaulino:
LGTM. There's a minor linter failure: `autopilot/agent_test.go:1284:30:warning: should drop = 0 from declaration of var balance; it is the zero value (golint)`

halseth:
> There's a minor linter failure: `autopilot/agent_test.go:1284:30:warning: should drop = 0 from declaration of var balance; it is the zero value (golint)`

Fixed!

wpaulino:
Fixup commits can be squashed now :)

halseth:
Rebased and squashed. PTAL @wpaulino

Roasbeef:
Thinking about this from a different angle, an alternative would be to implement the "fund all" command internally. This edge case only comes up if we're making channels below the max size, and don't have any funds left over for fees. Therefore if we're making a channel below the funding max size, we just need to tell the funding manager to create a channel using all available funds. However this doesn't quite work once the current funding limit is lifted and/or other soft caps are in place.

pilot.go
```diff
@@ -134,6 +137,74 @@ func (c *chanController) SpliceOut(chanPoint *wire.OutPoint,
...
+	if err != nil {
+		return 0, err
+	}
+
+	// Since we currently open each channel in a separate tx, the combined
+	// fee estimate will just be the sum of the individual ones. We attempt
+	// to mimic the coin selection done in the wallet. This won't be
+	// totally accurate (coins can be selected differently when the
+	// channels are actually being opened), but this is just an estimate
+	// soooo ¯\_(ツ)_/¯
```
Roasbeef: kek

Roasbeef: In order to make it a bit more accurate, this entire method should be nested within a call to [`WithCoinSelectLock`](https://github.com/lightningnetwork/lnd/blob/master/lnwallet/wallet.go#L1261).

pilot.go
```diff
@@ -134,6 +137,74 @@ func (c *chanController) SpliceOut(chanPoint *wire.OutPoint,
...
+// FeeEstimate returns the total estimated fee needed for the autopilot to open
+// channels of the given sizes. Since we currently open each channel in a
+// separate tx, the combined fee estimate will just be the sum of the
+// individual ones.
+func (c *chanController) FeeEstimate(chanSizes []btcutil.Amount) (
+	btcutil.Amount, error) {
+
+	// Find all unlocked unspent witness outputs that satisfy the minimum
+	// number of confirmations required.
+	coins, err := c.server.cc.wallet.ListUnspentWitness(
```
Roasbeef: Why this compared to making a dry run transaction?

autopilot/agent.go
```diff
@@ -606,22 +607,52 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
+			// in addition to the already selected ones.
+			feeEstimate, err := a.cfg.ChanController.FeeEstimate(
+				append(chanSizes, chanSize),
+			)
+			if err != nil {
+				return err
+			}
+
+			// Give us some wiggle room.
+			feeEstimate *= 3
```
Roasbeef: FWIW this can result in a very small channel (right above the min channel size) being created compared to the "fund all" approach.

pilot.go
```diff
@@ -134,6 +137,74 @@ func (c *chanController) SpliceOut(chanPoint *wire.OutPoint,
...
+
+	// Since we currently open each channel in a separate tx, the combined
+	// fee estimate will just be the sum of the individual ones. We attempt
+	// to mimic the coin selection done in the wallet. This won't be
+	// totally accurate (coins can be selected differently when the
+	// channels are actually being opened), but this is just an estimate
+	// soooo ¯\_(ツ)_/¯
+	var totalWeight int64
+	currentCoin := 0
+	for _, amt := range chanSizes {
```
Roasbeef: This assumes a _single_ transaction for all channels that has a funding and change address. As a result, it'll underestimate since it's missing the signalling overhead for teach of the individual transactions.

Roasbeef:
Ping!

pilot.go
wpaulino: What overhead are you referring to? This is how we do estimates on transaction weights in other parts of the codebase.

Roasbeef: Overhead as in things like the transaction version and lock time. In other parts, when we do estimation it's for a single transaction. This attempts to estimate the weight of a single mega funding transaction, but ultimately a series of funding transactions will be created instead. 

In terms of fixing the issue at hand (more breathing room for fees), this won't materially affect that, but it will cause the estimates to be off.

Roasbeef:
Pong.

autopilot/agent.go
```diff
@@ -606,22 +607,52 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
...
 		}
 
-		// Track the available funds we have left.
-		if availableFunds < chanSize {
-			chanSize = availableFunds
+		for {
+			// Get the fee estimated we need to open this channel
+			// in addition to the already selected ones.
+			feeEstimate, err := a.cfg.ChanController.FeeEstimate(
+				append(chanSizes, chanSize),
```
halseth: reminder: this is dangerous.

halseth:
Closing in favor of #3165

PR #3019, opened 2019-04-27
invoicesrpc: SubscribeSingleInvoice check RHashStr

sangaman:
This fixes a bug where the `RHashStr` field on the `PaymentHash` request for `SubscribeSingleInvoice` was being ignored. This brings the behavior of `SubscribeSingleInvoice` in line with `LookupInvoice`, where `RHashStr` is checked and decoded first before falling back to the `RHash` bytes field.


cfromknecht:
@sangaman good catch! i think this is worth including in 0.6.1

sangaman:
Right, the alternative is to just use a single `bytes` field the way `CancelInvoice` and `SettleInvoice` do. I can't really say which is better (and whether `RHashStr` has a real purpose here), but it should be fixed one way or the other and I don't mind refactoring this to take the single bytes field approach if that's what you guys think is best.

It also makes me wonder though, if you're right about the string field being intended for non-streaming calls over the REST interface, maybe it would make sense for the `SettleInvoice` and `CancelInvoice` calls to take the `PaymentHash` message type with both fields. At least if there's ever the need to support REST for those calls.

cfromknecht:
@joostjager i don't believe `RHashStr` is required for REST, i think it would just require the user to encode the value in base64 before making the request (which is quite cumbersome).

that being said, is it really worth introducing a breaking rpc change? as it stands in 0.6, this does appear to be a bug with an easy fix. if we want to propose a breaking change it can't be deployed until the next major version at the earliest, but we'd also introduce version incompatibilities between lnd and the loop client, for example. from my pov i think it still makes sense to get the fix in whether or not we proceed in altering the rpc

Roasbeef:
The primary rationale for any of the `Str` fields in the RPC protocol today is to allow certain REST endpoints to be things like: `/v1/channels/close/<txid>` without having to accept base64. However that only works for very simple calls, and most of them end up needing to have a JSON body in the post request, so they can handle the encoding themelves manually.

sangaman:
> think it would just require the user to encode the value in base64 before making the request (which is quite cumbersome).

The reason I encountered this bug is in fact because I was setting the RHashStr field out of convenience. And it's a good point that it'd be nice to fix this in a backwards compatible way, even code that's currently setting the RHash bytes field would need to be refactored if the request were changed to match CancelInvoice and SettleInvoice

cfromknecht:
looking at it now, it might have been an oversight to not make `SubscribeSingleInvoice` accept a `SubscribeSignleInvoiceRequest` with an embedded `lnrpc.PaymentHash` (or a single `RHash` field). as it stands, we are unable to extend the request with additional information without also extending `lnrpc.PaymentHash` or introducing a breaking change to move to a `SubscribeSingleInvoiceRequest`

joostjager:
@cfromknecht if we create `SubscribeSingleInvoiceRequest` with a single field `RHash` that has the same id 2 as it has in `lnrpc.PaymentHash`, I think it will be backwards compatible.

cfromknecht:
@joostjager only one way to find out :)

joostjager:
I already tried it before I posted that comment. It is true :)

sangaman:
Let me know how I should proceed with this PR.

Another idea is to make a `SubscribeSingleInvoiceRequest` message that has field 1 as `RHashStr` and field 2 as `RHash`. Then you can extend it in the future however you please, and if I understand @joostjager correctly then it wouldn't be breaking.

joostjager:
Yes, it didn't break backwards compatibility in the quick test that I did.

I think it should be a new message `SubscribeSingleInvoiceRequest` with a single parameter `payment_hash` of type `bytes`.

sangaman:
> I think it should be a new message SubscribeSingleInvoiceRequest with a single parameter payment_hash of type bytes.

If that's the consensus I'll refactor this PR tomorrow as per above.

cfromknecht:
@sangaman sounds good to me, let's move forward with `SubscribeSingleInvoiceRequest`

sangaman:
@cfromknecht @joostjager I've updated the PR to use `SubscribeSingleInvoiceRequest` and I tested this against code that was using the old proto definition and it worked, confirming what @joostjager found.

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -99,3 +99,8 @@ message SettleInvoiceMsg {
 } 
 
 message SettleInvoiceResp {}
+
+message SubscribeSingleInvoiceRequest {
+    /// Hash corresponding to the (hold) invoice to subscribe to.
+    bytes r_hash = 2 [json_name = "r_hash"];
```
halseth: should set `reserved  1` to indicate field 1 should not be used.

sangaman: Thanks, done.

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -99,3 +99,9 @@ message SettleInvoiceMsg {
 } 
 
 message SettleInvoiceResp {}
+
+message SubscribeSingleInvoiceRequest {
+    reserved 1;
+    /// Hash corresponding to the (hold) invoice to subscribe to.
```
joostjager: nit: newline after first field

sangaman: t

PR #3022, opened 2019-04-28
invoicesrpc: specify private hops for addinvoice, addholdinvoice

### LINKED ISSUE #2963: Specify individual private channels in addinvoice call
### Background

I have multiple private channels open and I want to issue an invoice via the `addinvoice` lncli call.  I want to include _one_ of these channels as a routing hint via the `--private` flag and not all of them.  Therefore, I propose adding a new flag to `addinvoice` that allows a user to specify individual private channels for routing hints instead of exposing all of their private channels in one invoice.
### Your environment

* version of `lnd`
lnd version 0.6.0-beta commit=v0.6-beta-rc3-19-g6cfb02813b0a1b470cf2fd738043f8cb20ac1569
Darwin

Crypt-iQ:
Fixes #2963

This PR allows a user to manually specify private channels to be used as hop hints in both the `addinvoice` and `addholdinvoice` RPC calls via the new `--private_hops` flag which accepts a JSON array of private channel ids.  The new flag must be used in conjunction with the `--private` flag.  In the current implementation, users will advertise all private channels which isn't ideal for privacy-conscious users.

I also changed the `--private` flag from `BoolTFlag` to `BoolFlag` so that users aren't always leaking their private channels if they don't want to.

Usage:
- addinvoice
`lncli addinvoice --private_hops '[467292441870336, 447501232570368, 487083651170304]' --private`

- addholdinvoice
`lncli addholdinvoice <hash> --private_hops '[467292441870336, 447501232570368, 487083651170304]' --private`


cmd/lncli/commands.go
```diff
@@ -2566,12 +2566,17 @@ var addInvoiceCommand = cli.Command{
...
+		cli.BoolFlag{
 			Name: "private",
 			Usage: "encode routing hints in the invoice with " +
 				"private channels in order to assist the " +
 				"payer in reaching you",
 		},
+		cli.StringFlag{
+			Name: "private_hops",
+			Usage: "a json array string of private channel ids " +
+				"to use as hop hints",
```
joostjager: Why not the simpler format `--private_hops 100,200,300`?

cmd/lncli/invoicesrpc_active.go
```diff
@@ -179,12 +180,17 @@ var addHoldInvoiceCommand = cli.Command{
...
 			Name: "private",
 			Usage: "encode routing hints in the invoice with " +
 				"private channels in order to assist the " +
 				"payer in reaching you",
 		},
+		cli.StringFlag{
+			Name: "private_hops",
+			Usage: "a json array string of private channel ids " +
+				"to use as hop hints",
+		},
```
joostjager: If you want, you can do some reuse here. See `commands.go`, `paymentFlags()`.

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -253,6 +258,11 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		options = append(options, zpay32.CLTVExpiry(uint64(defaultDelta)))
 	}
 
+	if len(invoice.PrivateHops) > 0 && !invoice.Private {
+		return nil, nil, fmt.Errorf("Must set private flag in conjunction " +
```
joostjager: errors are generally non-capitalized

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -279,6 +289,25 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 				continue
 			}
 
+			// If the user has manually defined private hop hints, ignore
+			// any channels not in the array.
+			if len(invoice.PrivateHops) > 0 {
+				match := false
+				for _, privateChan := range invoice.PrivateHops {
```
joostjager: This may be better to implement using a `map[uint64]struct{}`

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -82,6 +82,11 @@ message AddHoldInvoiceRequest {
 
     /// Whether this invoice should include routing hints for private channels.
     bool private = 9 [json_name = "private"];
+
+    /**
+    The list of private channel ids that will be used as hop hints.
+    */
+    repeated uint64 private_hops = 10 [json_name = "private_hops"];
```
joostjager: Wondering whether name should be `private_channels`? But we also have hop _hints_...

joostjager:
Can you also give a bit more background on how exactly you'd be using this? Is the goal to only expose one private channel for every peer connected to, so that the invoice doesn't disclose the combined capacity of all channels to a peer?

Or do you maybe have info on how the sender is going to approach you?

Crypt-iQ:
Currently the commands include all private channels as hop hints by default.  Users may not even know that this is happening (I didn't), so I disabled that in this PR.  If I only wanted to specify one, or a few, for the sender to use in path finding, this gives me more granular control.  Perhaps some of these private channels are unbalanced and could use some balancing.  Maybe there is one "very secret" private channel that I don't want exposed via the regular add(hold)invoice command(s) and I don't really care if my other private channels get exposed to the sender.

andronoob:
@joostjager I'm looking for this feature as well. My Internet connection comes with carrier-grade NAT, so simply setting NAT=true didn't seem to work. I have to embed routing hints of the channel counterparty to receive funds. Eclair mobile has done this properly, while LN-App still generate invoices without any routing hints.
I think maybe I should open private channels instead of public channels? In my opinion, there should be some protocol layer workaround, which could work like `ssh -R 0.0.0.0:9735:127.0.0.1:9735`

halseth:
@andronoob Yes, for non-routing nodes you should be using private channels.

andronoob:
@halseth In other words, it's a limitation of LN protocol that NAT can be a problem for routing nodes, therefore routing nodes must figure out how to overcome this problem by themselves, right?

halseth:
A routing node must be connectable, and how that is achieved is not a protocol problem.

> therefore routing nodes must figure out how to overcome this problem by themselves, right?

So yes :)

cmd/lncli/commands.go
Crypt-iQ: Done

lnrpc/invoicesrpc/addinvoice.go
Crypt-iQ: Agree, done

lnrpc/invoicesrpc/invoices.proto
Crypt-iQ: Changed to `private_channels` since I think it's more descriptive

lnrpc/invoicesrpc/invoices.proto
```diff
@@ -82,6 +82,11 @@ message AddHoldInvoiceRequest {
 
     /// Whether this invoice should include routing hints for private channels.
     bool private = 9 [json_name = "private"];
+
+    /**
+    The list of private channel ids that will be used as hop hints.
```
carlaKC: nit: add that these are short channel IDs?
It's apparent from the uint64 when you're used to the rpc, but might be helpful for newer users :)

cmd/lncli/invoicesrpc_active.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"strings"
```
carlaKC: Might as well blitz this unnecessary newline while you're here?

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -253,6 +258,16 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		options = append(options, zpay32.CLTVExpiry(uint64(defaultDelta)))
 	}
 
+	if len(invoice.PrivateChannels) > 0 && !invoice.Private {
```
carlaKC: Wouldn't the presence of a list of private channels imply that the user wants to have a private invoice? 
Non-critical, but I can definitely see myself calling with a list of channels and forgetting the private flag and wanting it to be set anyway.

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -279,6 +294,18 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
...
 
+			// If the user has manually defined private hop hints, ignore
+			// any channels not in the array.
+			if len(invoice.PrivateChannels) > 0 {
+				// Check to see whether the ShortChannelID exists in the private
+				// channel map. If it isn't, don't add this private channel to
+				// the list of hop hints.
+				sid := channel.ShortChannelID.ToUint64()
+				if _, ok := privateChanMap[sid]; !ok {
+					continue
```
carlaKC: As is, users can provide public or already closed channels in this list and they will just be filtered out by this + the isPublic check, which could result in there being no hints added when the user gives bad input?

The extreme case would be to error if they provide hops and we end up using none, but maybe just add some logging along the lines of `private invoice: received {len} suggested channels, created: {numHints} hints`?

lnrpc/rpc.proto
```diff
@@ -2242,6 +2242,11 @@ message Invoice {
 
     /// List of HTLCs paying to this invoice [EXPERIMENTAL].
     repeated InvoiceHTLC htlcs = 22 [json_name = "htlcs"];
+
+    /**
+    The list of private channel ids that will be used as hop hints.
```
carlaKC: Add a comment that hop hints will be strictly limited to these channels if they are specified?

lnrpc/invoicesrpc/addinvoice.go
Crypt-iQ: Gotcha, agree

alexbosworth:
A potential workaround for this issue now exists with the signMessage RPC API, you can encode your own BOLT 11 payment request however you like and then use that API to craft the signature for the payment request

Crypt-iQ:
> A potential workaround for this issue now exists with the signMessage RPC API, you can encode your own BOLT 11 payment request however you like and then use that API to craft the signature for the payment request

I agree with this, but I think that might be complicated for new users? Also at a minimum I think the private hops should *not* be specified by default if possible

halseth:
@Crypt-iQ Can we close in favor of #3672?

PR #3026, opened 2019-04-30
sweep: add support for multiple fee preferences to UtxoSweeper

wpaulino:
In this PR, we introduce support for arbitrary client fee preferences when accepting input sweep requests. This is possible with the addition of fee rate buckets. Fee rate buckets are buckets that contain inputs with similar fee rates within a specific range, e.g., 1-10 sat/vbyte, 11-20 sat/vbyte, etc. Having these buckets allows us to batch and sweep inputs from different clients with similar fee rates within a single transaction, allowing us to save on chain fees. Implementing these fee rate buckets also lays down the groundwork required for being able to bump the fee rates for particular inputs.

We also get rid of the UtxoSweeper's default fee preference. Any clients using it to sweep inputs specify the same default fee preference to not change their behavior. Each of these can be fine-tuned later on given their use cases.

### DIFF
--- a/contractcourt/commit_sweep_resolver.go
+++ b/contractcourt/commit_sweep_resolver.go
@@ -5,10 +5,16 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/lightningnetwork/lnd/input"
-
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/sweep"
+)
+
+const (
+	// commitOutputConfTarget is the default confirmation target we'll use
+	// for sweeps of commit outputs that belong to us.
+	commitOutputConfTarget = 6
 )
 
 // commitSweepResolver is a resolver that will attempt to sweep the commitment
@@ -98,7 +104,8 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 		// sweeper.
 		log.Infof("%T(%v): sweeping commit output", c, c.chanPoint)
 
-		resultChan, err := c.Sweeper.SweepInput(&inp)
+		feePref := sweep.FeePreference{ConfTarget: commitOutputConfTarget}
+		resultChan, err := c.Sweeper.SweepInput(&inp, feePref)
 		if err != nil {
 			log.Errorf("%T(%v): unable to sweep input: %v",
 				c, c.chanPoint, err)
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -488,8 +489,17 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 			s.timer = nil
 
 			// We'll attempt to cluster all of our inputs with
-			// similar fee rates.
-			for _, cluster := range s.clusterBySweepFeeRate() {
+			// similar fee rates. Before attempting to sweep them,
+			// we'll sort them in descending fee rate order. We do
+			// this to ensure any inputs which have had their fee
+			// rate bumped are broadcast first in order enforce the
+			// RBF policy.
+			inputClusters := s.clusterBySweepFeeRate()
+			sort.Slice(inputClusters, func(i, j int) bool {
+				return inputClusters[i].sweepFeeRate >
+					inputClusters[j].sweepFeeRate
+			})
+			for _, cluster := range inputClusters {
 				// Examine pending inputs and try to construct
 				// lists of inputs.
 				inputLists, err := s.getInputLists(
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -3,6 +3,7 @@ package sweep
 import (
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"sync"
 	"sync/atomic"
@@ -14,6 +15,25 @@ import (
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
+)
+
+const (
+	// DefaultMaxFeeRate is the default maximum fee rate allowed within the
+	// UtxoSweeper. The current value is equivalent to a fee rate of 10,000
+	// sat/vbyte.
+	DefaultMaxFeeRate lnwallet.SatPerKWeight = 250 * 1e4
+
+	// DefaultFeeRateBucketSize is the default size of fee rate buckets
+	// we'll use when clustering inputs into buckets with similar fee rates
+	// within the UtxoSweeper.
+	//
+	// Given a minimum relay fee rate of 1 sat/vbyte, a multiplier of 10
+	// would result in the following fee rate buckets up to the maximum fee
+	// rate:
+	//
+	//   #1: min = 1 sat/vbyte, max = 10 sat/vbyte
+	//   #2: min = 11 sat/vbyte, max = 20 sat/vbyte...
+	DefaultFeeRateBucketSize = 10
 )
 
 var (
@@ -53,6 +73,22 @@ type pendingInput struct {
 	// publishAttempts records the number of attempts that have already been
 	// made to sweep this tx.
 	publishAttempts int
+
+	// feePreference is the fee preference of the client who requested the
+	// input to be swept. If a confirmation target is specified, then we'll
+	// map it into a fee rate whenever we attempt to cluster inputs for a
+	// sweep.
+	feePreference FeePreference
+}
+
+// pendingInputs is a type alias for a set of pending inputs.
+type pendingInputs = map[wire.OutPoint]*pendingInput
+
+// inputCluster is a helper struct to gather a set of pending inputs that should
+// be swept with the specified fee rate.
+type inputCluster struct {
+	sweepFeeRate lnwallet.SatPerKWeight
+	inputs       pendingInputs
 }
 
 // UtxoSweeper is responsible for sweeping outputs back into the wallet
@@ -65,7 +101,9 @@ type UtxoSweeper struct {
 	newInputs chan *sweepInputMessage
 	spendChan chan *chainntnfs.SpendDetail
 
-	pendingInputs map[wire.OutPoint]*pendingInput
+	// pendingInputs is the total set of inputs the UtxoSweeper has been
+	// requested to sweep.
+	pendingInputs pendingInputs
 
 	// timer is the channel that signals expiry of the sweep batch timer.
 	timer <-chan time.Time
@@ -74,7 +112,7 @@ type UtxoSweeper struct {
 
 	currentOutputScript []byte
 
-	relayFeePerKW lnwallet.SatPerKWeight
+	relayFeeRate lnwallet.SatPerKWeight
 
 	quit chan struct{}
 	wg   sync.WaitGroup
@@ -114,10 +152,6 @@ type UtxoSweeperConfig struct {
 	// time the incubated outputs need to be spent.
 	Signer input.Signer
 
-	// SweepTxConfTarget assigns a confirmation target for sweep txes on
-	// which the fee calculation will be based.
-	SweepTxConfTarget uint32
-
 	// MaxInputsPerTx specifies the default maximum number of inputs allowed
 	// in a single sweep tx. If more need to be swept, multiple txes are
 	// created and published.
@@ -131,6 +165,22 @@ type UtxoSweeperConfig struct {
 	// NextAttemptDeltaFunc returns given the number of already attempted
 	// sweeps, how many blocks to wait before retrying to sweep.
 	NextAttemptDeltaFunc func(int) int32
+
+	// MaxFeeRate is the the maximum fee rate allowed within the
+	// UtxoSweeper.
+	MaxFeeRate lnwallet.SatPerKWeight
+
+	// FeeRateBucketSize is the default size of fee rate buckets we'll use
+	// when clustering inputs into buckets with similar fee rates within the
+	// UtxoSweeper.
+	//
+	// Given a minimum relay fee rate of 1 sat/vbyte, a fee rate bucket size
+	// of 10 would result in the following fee rate buckets up to the
+	// maximum fee rate:
+	//
+	//   #1: min = 1 sat/vbyte, max = 10 sat/vbyte
+	//   #2: min = 11 sat/vbyte, max = 20 sat/vbyte...
+	FeeRateBucketSize int
 }
 
 // Result is the struct that is pushed through the result channel. Callers can
@@ -149,19 +199,19 @@ type Result struct {
 // sweepInputMessage structs are used in the internal channel between the
 // SweepInput call and the sweeper main loop.
 type sweepInputMessage struct {
-	input      input.Input
-	resultChan chan Result
+	input         input.Input
+	feePreference FeePreference
+	resultChan    chan Result
 }
 
 // New returns a new Sweeper instance.
 func New(cfg *UtxoSweeperConfig) *UtxoSweeper {
-
 	return &UtxoSweeper{
 		cfg:           cfg,
 		newInputs:     make(chan *sweepInputMessage),
 		spendChan:     make(chan *chainntnfs.SpendDetail),
 		quit:          make(chan struct{}),
-		pendingInputs: make(map[wire.OutPoint]*pendingInput),
+		pendingInputs: make(pendingInputs),
 	}
 }
 
@@ -199,7 +249,7 @@ func (s *UtxoSweeper) Start() error {
 
 	// Retrieve relay fee for dust limit calculation. Assume that this will
 	// not change from here on.
-	s.relayFeePerKW = s.cfg.FeeEstimator.RelayFeePerKW()
+	s.relayFeeRate = s.cfg.FeeEstimator.RelayFeePerKW()
 
 	// Register for block epochs to retry sweeping every block.
 	bestHash, bestHeight, err := s.cfg.ChainIO.GetBestBlock()
@@ -252,24 +302,33 @@ func (s *UtxoSweeper) Stop() error {
 }
 
 // SweepInput sweeps inputs back into the wallet. The inputs will be batched and
-// swept after the batch time window ends.
+// swept after the batch time window ends. A custom fee preference can be
+// provided, otherwise the UtxoSweeper's default will be used.
 //
 // NOTE: Extreme care needs to be taken that input isn't changed externally.
 // Because it is an interface and we don't know what is exactly behind it, we
 // cannot make a local copy in sweeper.
-func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
+func (s *UtxoSweeper) SweepInput(input input.Input,
+	feePreference FeePreference) (chan Result, error) {
+
 	if input == nil || input.OutPoint() == nil || input.SignDesc() == nil {
 		return nil, errors.New("nil input received")
 	}
 
+	// Ensure the client provided a sane fee preference.
+	if _, err := s.feeRateForPreference(feePreference); err != nil {
+		return nil, err
+	}
+
 	log.Infof("Sweep request received: out_point=%v, witness_type=%v, "+
-		"time_lock=%v, size=%v", input.OutPoint(), input.WitnessType(),
-		input.BlocksToMaturity(),
-		btcutil.Amount(input.SignDesc().Output.Value))
+		"time_lock=%v, amount=%v, fee_preference=%v", input.OutPoint(),
+		input.WitnessType(), input.BlocksToMaturity(),
+		btcutil.Amount(input.SignDesc().Output.Value), feePreference)
 
 	sweeperInput := &sweepInputMessage{
-		input:      input,
-		resultChan: make(chan Result, 1),
+		input:         input,
+		feePreference: feePreference,
+		resultChan:    make(chan Result, 1),
 	}
 
 	// Deliver input to main event loop.
@@ -280,6 +339,27 @@ func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
 	}
 
 	return sweeperInput.resultChan, nil
+}
+
+// feeRateForPreference returns a fee rate for the given fee preference. It
+// ensures that the fee rate respects the bounds of the UtxoSweeper.
+func (s *UtxoSweeper) feeRateForPreference(
+	feePreference FeePreference) (lnwallet.SatPerKWeight, error) {
+
+	feeRate, err := DetermineFeePerKw(s.cfg.FeeEstimator, feePreference)
+	if err != nil {
+		return 0, err
+	}
+	if feeRate < s.relayFeeRate {
+		return 0, fmt.Errorf("fee preference resulted in invalid fee "+
+			"rate %v, mininum is %v", feeRate, s.relayFeeRate)
+	}
+	if feeRate > s.cfg.MaxFeeRate {
+		return 0, fmt.Errorf("fee preference resulted in invalid fee "+
+			"rate %v, maximum is %v", feeRate, s.cfg.MaxFeeRate)
+	}
+
+	return feeRate, nil
 }
 
 // collector is the sweeper main loop. It processes new inputs, spend
@@ -315,6 +395,7 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 				listeners:        []chan Result{input.resultChan},
 				input:            input.input,
 				minPublishHeight: bestHeight,
+				feePreference:    input.feePreference,
 			}
 			s.pendingInputs[outpoint] = pendInput
 
@@ -406,29 +487,30 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 			// be started when new inputs arrive.
 			s.timer = nil
 
-			// Retrieve fee estimate for input filtering and final
-			// tx fee calculation.
-			satPerKW, err := s.cfg.FeeEstimator.EstimateFeePerKW(
-				s.cfg.SweepTxConfTarget,
-			)
-			if err != nil {
-				log.Errorf("estimate fee: %v", err)
-				continue
-			}
-
-			// Examine pending inputs and try to construct lists of
-			// inputs.
-			inputLists, err := s.getInputLists(bestHeight, satPerKW)
-			if err != nil {
-				log.Errorf("get input lists: %v", err)
-				continue
-			}
-
-			// Sweep selected inputs.
-			for _, inputs := range inputLists {
-				err := s.sweep(inputs, satPerKW, bestHeight)
+			// We'll attempt to cluster all of our inputs with
+			// similar fee rates.
+			for _, cluster := range s.clusterBySweepFeeRate() {
+				// Examine pending inputs and try to construct
+				// lists of inputs.
+				inputLists, err := s.getInputLists(
+					cluster, bestHeight,
+				)
 				if err != nil {
-					log.Errorf("sweep: %v", err)
+					log.Errorf("Unable to examine pending "+
+						"inputs: %v", err)
+					continue
+				}
+
+				// Sweep selected inputs.
+				for _, inputs := range inputLists {
+					err := s.sweep(
+						inputs, cluster.sweepFeeRate,
+						bestHeight,
+					)
+					if err != nil {
+						log.Errorf("Unable to sweep "+
+							"inputs: %v", err)
+					}
 				}
 			}
 
@@ -441,7 +523,7 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 
 			bestHeight = epoch.Height
 
-			log.Debugf("New blocks: height=%v, sha=%v",
+			log.Debugf("New block: height=%v, sha=%v",
 				epoch.Height, epoch.Hash)
 
 			if err := s.scheduleSweep(bestHeight); err != nil {
@@ -454,6 +536,62 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 	}
 }
 
+// bucketForFeeReate determines the proper bucket for a fee rate. This is done
+// in order to batch inputs with similar fee rates together.
+func (s *UtxoSweeper) bucketForFeeRate(
+	feeRate lnwallet.SatPerKWeight) lnwallet.SatPerKWeight {
+
+	minBucket := s.relayFeeRate + lnwallet.SatPerKWeight(s.cfg.FeeRateBucketSize)
+	return lnwallet.SatPerKWeight(
+		math.Ceil(float64(feeRate) / float64(minBucket)),
+	)
+}
+
+// clusterBySweepFeeRate takes the set of pending inputs within the UtxoSweeper
+// and clusters those together with similar fee rates. Each cluster contains a
+// sweep fee rate, which is determined by calculating the average fee rate of
+// all inputs within that cluster.
+func (s *UtxoSweeper) clusterBySweepFeeRate() []inputCluster {
+	bucketInputs := make(map[lnwallet.SatPerKWeight]pendingInputs)
+	inputFeeRates := make(map[wire.OutPoint]lnwallet.SatPerKWeight)
+
+	// First, we'll group together all inputs with similar fee rates. This
+	// is done by determining the fee rate bucket they should belong in.
+	for op, input := range s.pendingInputs {
+		feeRate, err := s.feeRateForPreference(input.feePreference)
+		if err != nil {
+			log.Warnf("Skipping input %v: %v", op, err)
+			continue
+		}
+		bucket := s.bucketForFeeRate(feeRate)
+
+		inputs, ok := bucketInputs[bucket]
+		if !ok {
+			inputs = make(pendingInputs)
+			bucketInputs[bucket] = inputs
+		}
+		inputs[op] = input
+		inputFeeRates[op] = feeRate
+	}
+
+	// We'll then determine the sweep fee rate for each set of inputs by
+	// calculating the average fee rate of the inputs within each set.
+	inputClusters := make([]inputCluster, 0, len(bucketInputs))
+	for _, inputs := range bucketInputs {
+		var sweepFeeRate lnwallet.SatPerKWeight
+		for op := range inputs {
+			sweepFeeRate += inputFeeRates[op]
+		}
+		sweepFeeRate /= lnwallet.SatPerKWeight(len(inputs))
+		inputClusters = append(inputClusters, inputCluster{
+			sweepFeeRate: sweepFeeRate,
+			inputs:       inputs,
+		})
+	}
+
+	return inputClusters
+}
+
 // scheduleSweep starts the sweep timer to create an opportunity for more inputs
 // to be added.
 func (s *UtxoSweeper) scheduleSweep(currentHeight int32) error {
@@ -464,27 +602,25 @@ func (s *UtxoSweeper) scheduleSweep(currentHeight int32) error {
 		return nil
 	}
 
-	// Retrieve fee estimate for input filtering and final tx fee
-	// calculation.
-	satPerKW, err := s.cfg.FeeEstimator.EstimateFeePerKW(
-		s.cfg.SweepTxConfTarget,
-	)
-	if err != nil {
-		return fmt.Errorf("estimate fee: %v", err)
+	// We'll only start our timer once we have inputs we're able to sweep.
+	startTimer := false
+	for _, cluster := range s.clusterBySweepFeeRate() {
+		// Examine pending inputs and try to construct lists of inputs.
+		inputLists, err := s.getInputLists(cluster, currentHeight)
+		if err != nil {
+			return fmt.Errorf("get input lists: %v", err)
+		}
+
+		log.Infof("Sweep candidates at height=%v with fee_rate=%v, "+
+			"yield %v distinct txns", currentHeight,
+			cluster.sweepFeeRate, len(inputLists))
+
+		if len(inputLists) != 0 {
+			startTimer = true
+			break
+		}
 	}
-
-	// Examine pending inputs and try to construct lists of inputs.
-	inputLists, err := s.getInputLists(currentHeight, satPerKW)
-	if err != nil {
-		return fmt.Errorf("get input lists: %v", err)
-	}
-
-	log.Infof("Sweep candidates at height=%v, yield %v distinct txns",
-		currentHeight, len(inputLists))
-
-	// If there are no input sets, there is nothing sweepable and we can
-	// return without starting the timer.
-	if len(inputLists) == 0 {
+	if !startTimer {
 		return nil
 	}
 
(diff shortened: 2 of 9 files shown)


sweep/sweeper.go
```diff
@@ -252,12 +270,15 @@ func (s *UtxoSweeper) Stop() error {
 }
 
 // SweepInput sweeps inputs back into the wallet. The inputs will be batched and
-// swept after the batch time window ends.
+// swept after the batch time window ends. A custom fee preference can be
+// provided, otherwise the UtxoSweeper's default will be used.
```
joostjager: Why would we offer a default? I think it is good to make it a required parameter and force callers to think about it.

sweep/sweeper.go
```diff
@@ -318,6 +340,37 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+			// If the caller didn't specify a fee preference, we'll
+			// use the sweeper's default.
+			if input.feePreference == nil {
+				pendInput.feePreference = FeePreference{
+					ConfTarget: s.cfg.SweepTxConfTarget,
+				}
+			} else {
+				// Otherwise, we'll validate it to ensure the
+				// caller provided a sane value.
+				_, err := DetermineFeePerKw(
```
joostjager: This can be moved into `SweepInput`, before we hand off to this goroutine.

sweep/sweeper.go
```diff
@@ -464,27 +521,25 @@ func (s *UtxoSweeper) scheduleSweep(currentHeight int32) error {
...
-		return fmt.Errorf("get input lists: %v", err)
-	}
+		log.Infof("Sweep candidates at height=%v, fee_preference=%v, "+
+			"yield %v distinct txns", currentHeight, feePref,
+			len(inputLists))
 
-	log.Infof("Sweep candidates at height=%v, yield %v distinct txns",
-		currentHeight, len(inputLists))
+		if len(inputLists) != 0 {
+			startTimer = true
```
joostjager: I think timers should be separate per fee pref. The idea is to allow time for more inputs to come in, but if they are in a different bucket, that won't work as expected.

sweep/sweeper.go
```diff
@@ -531,15 +586,23 @@ func (s *UtxoSweeper) signalAndRemove(outpoint *wire.OutPoint, result Result) {
 
 	// Inputs are no longer pending after result has been sent.
 	delete(s.pendingInputs, *outpoint)
+	delete(s.inputsByFeePref[pendInput.feePreference], *outpoint)
```
joostjager: Is the map itself for a particular fee pref ever deleted? Not really important though.

sweep/sweeper.go
```diff
@@ -603,21 +670,28 @@ func (s *UtxoSweeper) getInputLists(currentHeight int32,
...
+		pkScript, err := s.cfg.GenSweepScript()
 		if err != nil {
 			return fmt.Errorf("gen sweep script: %v", err)
 		}
+		s.scriptsByFeePref[feePreference] = pkScript
+	}
+
+	// Retrieve the fee rate we'll attempt to use for the sweep transaction
+	// based of the given fee preference.
+	satPerKW, err := DetermineFeePerKw(s.cfg.FeeEstimator, feePreference)
```
joostjager: Is it necessary to call this again? We already made the call when we created the input set.

sweep/sweeper.go
wpaulino: Was thinking of the case where in the future we allow a flag that specifies the default fee preference of users. The flag doesn't exist yet, but don't think it'd hurt to keep this as is for now.

sweep/sweeper.go
wpaulino: This validation is now handled by the `bucketManager`.

sweep/sweeper.go
wpaulino: Isn't that the case with this change already? If any of the buckets has inputs, then we'll start the timer to collect more.

sweep/sweeper.go
wpaulino: No, but it is now.

sweep/sweeper.go
wpaulino: Not really, removed.

wpaulino:
PTAL @joostjager.

sweep/bucket.go
```diff
@@ -0,0 +1,186 @@
...
+// rate allowed for any input within the bucket.
+type bucketID = lnwallet.SatPerKWeight
+
+// bucket represents a fee rate bucket which coalesces inputs with similar fee
+// rates in order to sweep them together within a single transaction to save on
+// fees.
+type bucket struct {
+	// id is the bucket's identifier, which also serves as the maximum fee
+	// rate allowed for any input within the bucket.
+	id bucketID
```
joostjager: Isn't it better to name this field after what it really is, fee rate?

sweep/sweeper.go
joostjager: It isn't an important detail, but there is an assumption in it that there will be a single default fee preference for all input sources. It may also be that we will assign different fee rates to different sources and then the defaults need to move to the callers.

sweep/sweeper.go
```diff
@@ -259,23 +276,40 @@ func (s *UtxoSweeper) Stop() error {
...
+	// fee preference.
+	if feePreference == nil {
+		feePreference = &FeePreference{
+			ConfTarget: s.cfg.SweepTxConfTarget,
+		}
+	}
+	feeRate, err := DetermineFeePerKw(s.cfg.FeeEstimator, *feePreference)
+	if err != nil {
+		return nil, err
+	}
```
joostjager: Could the whole idea of conf targets be moved outside of `UtxoSweeper`?

sweep/sweeper.go
joostjager: There is still a single timer. The idea of the timer is that once the first input for a tx comes in, we start a timer to accumulate more inputs for that tx. Because of the buckets, we are now accumulating inputs for multiple txes in parallel. So suppose input A with fee rate 10 comes in. This will start the timer. Then after 29 seconds, input B with fee rate 15 comes in. One second later, two txes are published, but the fee rate 15 tx didn't get its 30 seconds to accumulate more inputs.

sweep/sweeper_test.go
```diff
@@ -402,9 +399,10 @@ func TestNegativeInput(t *testing.T) {
 	ctx.backend.mine()
 
 	ctx.expectResult(secondLargeInputResult, nil)
-	ctx.expectResult(negInputResult, nil)
 
-	ctx.finish(1)
+	// The negative input's waitForSpend goroutine should still be lingering
+	// since it hasn't been swept.
+	ctx.finish(2)
```
joostjager: If this commit is fixing a failing test (?), I think it should be grouped with the code that introduces the failure.

sweep/sweeper.go
```diff
@@ -322,9 +356,17 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+				feeRate:          input.feeRate,
 			}
 			s.pendingInputs[outpoint] = pendInput
 
+			// We'll add the input to a fee rate bucket based on the
+			// fee preference specified by the client so that it can
+			// be batched with the other inputs with similar fee
+			// rates. We can ignore the error as the fee rate has
+			// already been validated.
+			s.bucketManager.addInput(pendInput)
```
joostjager: Is this the best split of responsibilities between `UtxoSweeper` and `bucketManager`? Both keeping sets of inputs looks like it isn't as clear as it could be.

If you are keeping a global set of inputs anyway, you may just as well determine the buckets on the fly when you need them. It would basically be a function that takes a list of (input) fee rates and returns a mapping of those input fee rates to the fee rate to be used for the tx. Clustering the input fee rates identical to how this currently happens in the bucketManager.

sweep/sweeper.go
```diff
@@ -413,29 +455,31 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
-				log.Errorf("get input lists: %v", err)
-				continue
-			}
-
-			// Sweep selected inputs.
-			for _, inputs := range inputLists {
-				err := s.sweep(inputs, satPerKW, bestHeight)
+			// We'll attempt to batch all of our inputs by their fee
+			// preference.
+			for _, bucket := range s.bucketManager.buckets {
```
joostjager: So this would be the place where you'd call `s.Cluster(s.pendingInputs)`

sweep/sweeper.go
joostjager: But it could also be argued that all inputs typically come in in a 30 second window after a block is mined anyway, regardless of the fee rate. In that case it wouldn't matter whether you have one or multiple timers, as they all expire at the same moment.

sweep/sweeper_test.go
wpaulino: I opted for fixing the test in a separate commit to make it more clear as to why the change was needed. FWIW, it still compiles, so it only fails if you try to run it.

sweep/sweeper.go
wpaulino: As in allow the sweeper to only take in fee rates, so the caller would be responsible for mapping a target to a fee rate?

sweep/sweeper.go
wpaulino: > But it could also be argued that all inputs typically come in in a 30 second window after a block is mined anyway, regardless of the fee rate. In that case it wouldn't matter whether you have one or multiple timers, as they all expire at the same moment.

Yeah this is how I was thinking of the timer. Seems fine to leave as is then?

sweep/sweeper_test.go
joostjager: Every commit builds is :2nd_place_medal: but getting every commit pass tests is :1st_place_medal:  At least that is what I do :)

sweep/sweeper.go
Roasbeef: IMO it make sense to have it live here. Callers are expressing their time preference w.r.t their expectation of when the sweep would happen. Pushing this mapping on the callers would just add code duplication at each call site, when it can instead all be consolidated here.

wpaulino: Yeah it's necessary to have it done at the `UtxoSweeper` level anyway in order to get the confirmation target.

sweep/bucket.go
```diff
@@ -0,0 +1,186 @@
...
+	// DefaultMaxFeeRate is the default maximum fee rate we'll allow when
+	// sweeping inputs. This value corresponds to a 1e4 sat/vbyte fee rate.
+	DefaultMaxFeeRate = DefaultMinFeeRate + (DefaultMinFeeRate * 1e4)
+
+	// DefaultFeeRateSpacing is the default distribution of each fee rate
+	// bucket. The current value allows for a distribution as follows:
+	//
+	//   bucket #1: min=1 sat/vbyte, max=20 sat/vbyte
+	//   bucket #2: min=21 sat/vbyte, max=40 sat/vbyte...
+	DefaultFeeRateSpacing = DefaultMinFeeRate * 20
```
Roasbeef: I think 20 might be too large a distance between buckets. Especially at the current fee levels on the chain today. Perhaps 5 is more reasonable?

sweep/bucket.go
```diff
@@ -0,0 +1,186 @@
...
+	// the inputs.
+	sweepScript []byte
+}
+
+// newBucket initializes a new bucket with the given identifier.
+func newBucket(id bucketID) *bucket {
+	return &bucket{
+		id:          id,
+		inputs:      make(map[wire.OutPoint]*pendingInput),
+		sweepScript: nil,
```
Roasbeef: Style nit: no need to set to nil.

Roasbeef: Won't we also know the target script at time of bucket creation also?

sweep/bucket.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+}
+
+// feeRate determines the fee rate of a bucket. This is done by calculating the
+// average fee rate of all inputs within it.
+func (b *bucket) feeRate() lnwallet.SatPerKWeight {
```
Roasbeef: Likely can be deferred as a follow up, but we may also eventually want to introduce the concept of "strict buckets". As an example, a strict bucket of 1 sat/byte would allow a caller to express that they _only_ want that input swept at that specified fee rate and care less about batching.

sweep/bucket.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, input := range b.inputs {
+		total += input.feeRate
+	}
+	return total / lnwallet.SatPerKWeight(len(b.inputs))
+}
+
+// bucketManager is responsible for managing a collection of fee rate buckets in
+// which we'll bundle inputs with similar fee rates together in order to sweep
+// them under a single transaction to save on fees.
+type bucketManager struct {
```
Roasbeef: prefix `fee` for all `bucket*`?

sweep/sweeper.go
Roasbeef: There's still a single input instance, but multiple references to a given input now exist. I think your second comment is a matter of the degree of encapsulation. As is now, the buckets are a black box to the sweeper until it actually needs to sweep something, but then even from it's PoV they're just another layer above the input lists. Additionally IMO having the new behavior be encapsulated as it is now makes it easier to unit test and also modify exactly _what_ the buckets consist of (for example, confs vs fee rates) in the future. 

I'd also expect the re-clustering logic (when we need to re-target due to a missed confirmation target) to live in this new sub-system as well.

sweep/sweeper.go
```diff
@@ -413,29 +455,31 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+					log.Errorf("Unable to examine pending "+
+						"inputs: %v", err)
+					continue
+				}
+
+				// Sweep selected inputs.
+				for _, inputs := range inputLists {
+					err := s.sweep(
+						inputs, feeRate,
+						bucket.sweepScript, bestHeight,
```
Roasbeef: Where is `sweepScript` ever set for a bucket?

wpaulino: It'll be set within `sweep()` at the time of sweeping.

sweep/sweeper.go
joostjager: I was thinking about a wrapper struct around the `UtxoSweeper` for callers that want to specify a conf target. So callers still don't need to do the mapping themselves.

`UtxoSweeper` isn't the easiest code to understand, so splitting off what's possible may help.

But not a big issue.

sweep/sweeper.go
joostjager: I don't see a downside to determining the clusters only when you need them and it allows more dynamic schemes.

If you want to stay with the buckets as they currently are, I'd at least separate concerns more. If a bucket is more or less operating independently, there must be code inside `UtxoSweeper` that can be moved to the bucket. For example creating the txes for a bucket.

sweep/bucket.go
joostjager: Suppose the fee landscape changes drastically, are we going to make a new release that changes this constant?

I am still not convinced that the bucket system is better than just grouping inputs based on their exact fee rate. Inputs that are offered with the same conf target will end up in the same bucket anyway and for bumping non-lnd inputs (not the main use case as far as I can see), there needs to be some awareness with the caller that grouping only happens with the same fee rate. A UI could offer a limited set of options to make that clear.

sweep/sweeper.go
joostjager: I think we don't need a sweep script per bucket, but can just track the last unused address sweeper-wide.

sweep/bucket.go
wpaulino: Agree 20 is probably too large. Ended up going with 10.

> Suppose the fee landscape changes drastically, are we going to make a new release that changes this constant?

sweep/sweeper.go

sweep/sweeper.go

sweep/sweeper_test.go
wpaulino: Done.

sweep/sweeper.go
wpaulino: Yeah agree that it would be a nice separation of concerns. Seems like a good candidate for a follow-up PR though?

sweep/sweeper.go

wpaulino:
PTAL @joostjager @Roasbeef.

contractcourt/commit_sweep_resolver.go
```diff
@@ -5,10 +5,16 @@ import (
...
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/lightningnetwork/lnd/sweep"
+)
+
+const (
+	// commitOutputConfTarget is the default confirmation target we'll use
+	// for sweeps of commit outputs that belong to us.
+	commitOutputConfTarget = 6
```
joostjager: I like it, explicit without a default. Later we could move those settings for different classes of inputs to a shared location possibly. Or add some logic around it to set the conf target based on the value.

joostjager: I like it, explicit without a default. Later we could move those settings for different classes of inputs to a shared location possibly. Or add some logic around it to set the conf target based on the value.

sweep/sweeper.go
```diff
@@ -282,6 +374,26 @@ func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
...
+// the UtxoSweeper's config.
+func (s *UtxoSweeper) validateFeePreference(
+	feePreference FeePreference) (lnwallet.SatPerKWeight, error) {
+
+	feeRate, err := DetermineFeePerKw(s.cfg.FeeEstimator, feePreference)
+	if err != nil {
+		return 0, err
+	}
+
+	if feeRate < s.cfg.MinFeeRate {
```
joostjager: Hmm, now we may reject an input when it is submitted, pushing for example a resolver into an error state. While the next block the estimate may be different and within bounds.

Alternatively, we could just cap the fee with the max.

sweep/sweeper.go
```diff
@@ -441,8 +574,65 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+				if numBlocksLeft < 1 {
+					numBlocksLeft = 1
+				}
+				threshold := int32(
+					float64(confTarget) * confTargetRatio,
+				)
+
+				if numBlocksLeft > threshold {
+					continue
+				}
```
joostjager: I am not sure if we should do this already now. I'd be worried that on average users are going to end up paying more. Is this an actual problem to solve? To me it seems that being able to adjust the conf target or the fee rate directly is enough control.

Suppose the conf target is 6. We published with a fee that should get is confirmed after max 6 blocks. Suppose it would happen after exactly 6 blocks, which is within what is specified. We wouldn't need to do anything. But this logic starts bumping the fee up already after 1 block to the 5 block fee estimate.

sweep/sweeper.go
joostjager: I think we have to be realistic at follow up, pure refactor prs. They are mostly not going to happen. Also we would incur a risk twice. First changes in this pr and when that is all good, we change the system again with new risks.

Imo it is better to try to do it first time right. Less overall costs. We surely don't want to over-engineer, but the size of the sweeper as it currently is justifies some effort in separation responsibilities. Costs over the life time of the software of future devs trying to understand code adds up.

An alternative design that could be good is to offer inputs to `UtxoSweeper` with a `GetFeeRate` callback. That way callers can pass in a function that either returns a static fee rate, a fee rate calculated from a conf target or something more advanced that also takes into account the value of the input for example. For `UtxoSweeper` it'll be all the same. When it sweeps, it just collects fee rates and does its clustering magic. 

For conf targets, there could be an object `feeEstimationCache` that caches the fee rate estimates for the block. The `getFeeRate()` callback for an input could be something like this closure:
`func() {return feeEstimationCache.GetFeeRate(defaultCommitSweepTarget)}`

sweep/sweeper.go
```diff
@@ -454,6 +644,60 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+// clusterBySweepFeeRate takes the set of pending inputs within the UtxoSweeper
+// and clusters those together with similar fee rates. Each cluster contains a
+// sweep fee rate, which is determined by calculating the average fee rate of
+// all inputs within that cluster.
+func (s *UtxoSweeper) clusterBySweepFeeRate() []inputCluster {
+	// First, we'll group together all inputs with similar fee rates. This
+	// is done by determining the fee rate bucket they should belong in.
+	bucketInputs := make(map[lnwallet.SatPerKWeight]pendingInputs)
+	for _, input := range s.pendingInputs {
+		bucket := s.bucketForFeeRate(input.feeRate)
```
joostjager: With the fee rate callback suggestion above, this would become `bucket := s.bucketForFeeRate(input.getFeeRate())`

sweep/sweeper.go
joostjager: Yes I think it is fine

sweep/sweeper.go
```diff
@@ -441,8 +574,65 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+				}
+
+				log.Debugf("Applying new fee rate of %v for "+
+					"input %v in order to confirm by "+
+					"height %v", newFeeRate,
+					pendInput.input.OutPoint(),
+					expectedConfHeight)
+
+				pendInput.feeRate = newFeeRate
+			}
```
joostjager: With the `getFeeRate()` suggestion, this would all be moved behind that call back and abstracted away from `UtxoSweeper`.

sweep/sweeper.go
wpaulino: Currently this value is set to the fee floor, which all estimators ensure we don't return a fee rate below. If this was to change though, then the case you bring up could definitely happen. Seems like it might be better to not even make this value configurable then?

sweep/sweeper.go
wpaulino: It's definitely a bit aggressive at the moment, but the functionality itself is still useful. Perhaps a better approach is to only start aggressively bumping after a positive delta (e.g., 2 blocks) from the intended confirmation height?

joostjager: Yes, so I am arguing that it may be too aggressive. If I pay for 6 confs and after 1 block I already become nervous and up the fee, isn't that strange?

We should also not lose sight of the problem to solve. Is it really this strict enforcement of a deadline or do we just want to get some more control over the sweeper process without hard guarantuees?

wpaulino: > If I pay for 6 confs and after 1 block I already become nervous and up the fee, isn't that strange?

That's why the threshold exists. The point is to increase the fee rate _only_ if we need to in order to meet the confirmation target. If the fee rate doesn't change, then we won't update it.

The problem to solve is that confirmation targets won't be retried when scheduling/performing a sweep. If fees rise, the sweeper would handle that. With the buckets however, confirmation targets aren't retried. We could keep the behavior the same by querying for the same confirmation target like it previously would, but it's also about enforcing what "confirmation target" actually means. The current behavior doesn't hold the guarantees the term implies.

sweep/sweeper.go
Roasbeef: IMO we shouldn't push all this burden on the caller. The goal is to be able to allow the caller to express a fee preference. In the end, that preference may not be _exactly_ met, but within a ball park range, particularly if there's a slightly different mapping that results in `lnd` paying lower fees over all. The code in question as is is also just a few lines: calling to the estimator. If we want to add some more magic between the caller and the estimator, then we can do that at the estimator interface level. 

If we look at the current use cases within `lnd`, every time a caller goes to request that something be swept, it has a rough _time frame_ w.r.t when it should be swept. Scenarios such as sweeping an HTLC have a very clear time frame (to avoid the race), while presences such as sweeping a CSV output on a confirmed output don't (as long as it isn't a revoked commitment ;)). 

> Yeah agree that it would be a nice separation of concerns. 
> but the size of the sweeper as it currently is justifies some effort in separation responsibilities.

The prior approach had an even more direct separation of concerns, but here we are now. Let's not forget the reality of things as they exist today: chain fess are rising, and users need to be able to specify a time preference. Even our own systems need this capability in order to scale and be sustainable. We have an opportunity to get something incrementally better in (atm we basically have nothing) before the next scheduled major release. Lets focus on the use cases at hand that we've experienced the need for first had (and has also been requested time after time by users of `lnd`). 

> An alternative design that could be good is to offer inputs to UtxoSweeper with a GetFeeRate callback

In the near future, do we ever see this being anything other than the default? All our current use cases just want to express a preference, and have the fee rate be _loosely_ adjusted (there're no guarantees after all) if that is at risk of not being achieved. Down the line if we have a compelling use case, this can be added in: for these special inputs, the sweeper skips its regular mapping and uses that function instead.

sweep/bucket.go
Roasbeef: > I am still not convinced that the bucket system is better than just grouping inputs based on their exact fee rate.

It's a tradeoff between 1 transaction per satoshi fee rate unit compared to grouping them optimistically to batch the transactions. The former approach will result in more transaction on chain, and likely amount of satoshis spent on fees by `lnd` which may offset any precision gains. The buckets are about "satisficing": the caller specifies a fee rate and we do our best to meet that match, but if we can bundle them, then we do as it leads to less fees globally across all callers.

Roasbeef: We can add a strict mode in the future if there're certain cases that are unable to compromise w.r.t their target fee rate.

sweep/sweeper.go
joostjager: >The code in question as is is also just a few lines: calling to the estimator. 

It is more than a few lines, it is also the re-targeting part. I think an object shouldn't have too many responsibilities and the conf target mapping is something that can be extracted. But in the end it comes down to a believe about what is the best design.

>In the near future, do we ever see this being anything other than the default? 

As I now understand it, there are two implementations of this callback: either conf target based (returns a diff. fee rate every block) and a static fee rate (passed in on the cmd line).

sweep/bucket.go
joostjager: Conf based sweeps will end up in the same group anyway, because they have the exact same fee rate. And for inputs specified on the command line, I think a user would have a few favorite fee rates that they pick from. I don't see it happening that a user bumps fees for several inputs in the same time window with just slightly different fee rates. So I don't expect there to be more chain txes in real life use of this feature.

But similar to the design discussion, this is also based on an expectation of how the function will be used.

sweep/sweeper.go
joostjager: Ok, so this check is in for manually bumped inputs then?

sweep/bucket.go
Roasbeef: Regardless of how it ends up getting used (user access patterns), it's a fact that more chain transactions will be generated if no sort of bucketization is used. As an example, today we use 3 or 6 confs pretty widely within the codebase. With the current proposed buckets those would be grouped together meaning a single transaction rather than two.

joostjager: Why would they be grouped together? It could be that the fee rates for 3 and 6 confs from the estimator at some point end up in a different bucket?

sweep/sweeper.go
halseth: I believe this could be both simplified and made more safe for many of our use cases by specifying an _absolute block height conf target_ instead the current conf target in number of blocks. Say we need a sweep to confirm by a specific height, the sweeper would start out low and increase the feerate as the target approaches.

Most important IMO is to adjust the fee rate dynamically, which could be one of the really powerful usecases of a sweeper like this.

sweep/sweeper.go
```diff
@@ -454,6 +541,56 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+	minBucket := s.cfg.MinFeeRate + step
+	for bucket := minBucket; bucket <= s.cfg.MaxFeeRate; bucket += step {
+		if feeRate <= bucket {
+			return bucket
+		}
+	}
+
+	// This should never happen as we assume that the fee rate has already
+	// been validated so that it fits within our constraints.
+	return minBucket
```
halseth: return error?

sweep/sweeper.go
```diff
@@ -454,6 +541,56 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 	}
 }
 
+// bucketForFeeReate determines the proper bucket for a fee rate. This is done
+// in order to batch inputs with similar fee rates together.
+func (s *UtxoSweeper) bucketForFeeRate(
+	feeRate lnwallet.SatPerKWeight) lnwallet.SatPerKWeight {
+
+	step := s.cfg.FeeRateSpacing
```
halseth: Using fixed buckets seems inherently limiting. For instance would you never sweep two inputs having feerates of 9 and 11 sat/b together, even though it could be cheaper to group them together in one 11 sat/b transaction.

Another approach would be to look at the lowest feerate input, and keep clustering it with other inputs as long as that is cheaper than sweeping them separately.

sweep/sweeper.go
```diff
@@ -528,8 +564,65 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 
 			bestHeight = epoch.Height
 
-			log.Debugf("New block: height=%v, sha=%v",
-				epoch.Height, epoch.Hash)
+			log.Debugf("New block: height=%v, sha=%v", epoch.Height,
+				epoch.Hash)
+
+			// In order to correctly meet the confirmation targets
```
halseth: Ok, this addresses some of my other comments :)

sweep/sweeper.go
halseth: It could be useful to distinguish inputs that we _want_ to be swept within X blocks, and inputs that _has_ to be swept within X blocks (absolute block height).

sweep/sweeper.go
halseth: Could also be a method on the `FeePreference` struct.

sweep/sweeper.go
```diff
@@ -528,8 +529,17 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 			s.timer = nil
 
 			// We'll attempt to cluster all of our inputs with
-			// similar fee rates.
-			for _, cluster := range s.clusterBySweepFeeRate() {
+			// similar fee rates. Before attempting to sweep them,
+			// we'll sort them in descending fee rate order. We do
```
halseth: 👍

contractcourt/commit_sweep_resolver.go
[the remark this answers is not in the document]
wpaulino: I prefer this way in order to use the `const`.

sweep/sweeper.go
```diff
@@ -252,23 +308,33 @@ func (s *UtxoSweeper) Stop() error {
...
-func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
+func (s *UtxoSweeper) SweepInput(input input.Input,
+	feePreference FeePreference) (chan Result, error) {
+
 	if input == nil || input.OutPoint() == nil || input.SignDesc() == nil {
 		return nil, errors.New("nil input received")
 	}
 
+	// Ensure the client provided a sane fee preference.
+	feeRate, err := s.validateFeePreference(feePreference)
```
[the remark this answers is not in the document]
wpaulino: Yeah this is already the case if the client specifies a confirmation target. If a fee rate is specified, then it indicates the client would like to sweep only at that fee rate (or somewhere close due to the bucket structure).

sweep/sweeper.go
wpaulino: Yeah there are definitely multiple ways of clustering inputs that would produce more efficient sweeps depending on the scenario -- the current implementation only serves as a starting point. In the future, it'd be nice if the sweeper was aware of the best way to cluster inputs given different scenarios.

sweep/sweeper.go
wpaulino: It will be yes.

sweep/bucket.go
Roasbeef: Ah sorry, there's an implicit assumption there for the example that those would fall in similar buckets. Main point is that we can create less on chain transactions by satisficing across fee rates.

sweep/sweeper.go
Roasbeef: One other relevant mempool level detail is that we can't add _new_ unconfirmed inputs to an existing transaction. As a result, the transaction level clustering can't vary _too much_ between retargets. With the bucket approach the transactions in each bucket can stay relatively stable, but say if our 1 sat/byte input gets bumped to 5 sat/byte, then it'll still be in the same transaction cluster but will push the actual fee rate of that bucket up a bit.

sweep/bucket.go
joostjager: Yes I got that. My point is that if we only ever pick a few fee rates (for example 3 confs, 6 confs (for internal sweeps) and 10,20,50 sat/byte (for external bumps)), we don't need to cluster and can just publish a tx per distinct fee rate.

Roasbeef: Yes if there's really no variation in fee preferences we'll end up publishing more transactions, but may save on fee cost globally.

sweep/sweeper.go
```diff
@@ -547,8 +574,65 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+			// NOTE: Meeting the confirmation target is a bit
+			// inaccurate at the moment due to the sweeper randomly
+			// rescheduling publish attempts for inputs.
+			for input, confTarget := range s.confTargetInputs {
+				pendInput := s.pendingInputs[input]
+
+				// We'll only proceed to query for a new fee
+				// rate if we don't exceed the threshold of
+				// confirmations left in order to do so.
+				expectedConfHeight := pendInput.initialHeight +
```
Roasbeef: I think we need to reset `initialHeight` here. Otherwise, after the first retarget period, we'll continually retarget it each block.

wpaulino:
Pushed a new version that removes re-evaluating fee rates for inputs with confirmation targets.

wpaulino:
PTAL @Roasbeef @joostjager @halseth.

sweep/sweeper.go
```diff
@@ -282,6 +341,27 @@ func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
 	return sweeperInput.resultChan, nil
 }
 
+// validateFeePreference validates a client's fee preference in accordance to
+// the UtxoSweeper's config.
+func (s *UtxoSweeper) validateFeePreference(
+	feePreference FeePreference) (lnwallet.SatPerKWeight, error) {
+
+	feeRate, err := DetermineFeePerKw(s.cfg.FeeEstimator, feePreference)
```
joostjager: `DetermineFeePerKw` internally uses a hard coded `FeePerKwFloor`, while we also have a min relay fee in the sweeper. These are the same thing, aren't they? If so, shouldn't the same source be used?

sweep/sweeper.go
```diff
@@ -282,6 +341,27 @@ func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
 	return sweeperInput.resultChan, nil
 }
 
+// validateFeePreference validates a client's fee preference in accordance to
+// the UtxoSweeper's config.
+func (s *UtxoSweeper) validateFeePreference(
```
joostjager: Name of the function only partially covers what it does. It is also used to pick a fee rate.

sweep/sweeper.go
```diff
@@ -252,24 +302,33 @@ func (s *UtxoSweeper) Stop() error {
...
-func (s *UtxoSweeper) SweepInput(input input.Input) (chan Result, error) {
+func (s *UtxoSweeper) SweepInput(input input.Input,
+	feePreference FeePreference) (chan Result, error) {
+
 	if input == nil || input.OutPoint() == nil || input.SignDesc() == nil {
 		return nil, errors.New("nil input received")
 	}
 
+	// Ensure the client provided a sane fee preference.
+	if _, err := s.validateFeePreference(feePreference); err != nil {
```
joostjager: Maybe it is clearer if there is a separate validate function that just checks the static fee rate.

sweep/sweeper.go
```diff
@@ -454,6 +545,73 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+// in order to batch inputs with similar fee rates together.
+func (s *UtxoSweeper) bucketForFeeRate(
+	feeRate lnwallet.SatPerKWeight) (lnwallet.SatPerKWeight, error) {
+
+	step := lnwallet.SatPerKWeight(s.cfg.FeeRateBucketSize)
+	for bucket := s.relayFeeRate + step; bucket <= s.cfg.MaxFeeRate; bucket += step {
+		if feeRate <= bucket {
+			return bucket, nil
+		}
+	}
```
joostjager: This could also be a simple divide with rounding?

sweep/sweeper.go
```diff
@@ -454,6 +545,73 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+	inputFeeRates := make(map[wire.OutPoint]lnwallet.SatPerKWeight)
+
+	// First, we'll group together all inputs with similar fee rates. This
+	// is done by determining the fee rate bucket they should belong in.
+	for op, input := range s.pendingInputs {
+		feeRate, err := s.validateFeePreference(input.feePreference)
+		if err != nil {
+			log.Warnf("Skipping input %v: %v", op, err)
+			continue
+		}
```
joostjager: I understood this should never be possible? The static fee rate has been validated already when the input was submitted to the sweeper and the fee estimate should never go out of the ranges?

sweep/sweeper.go
```diff
@@ -454,6 +545,73 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
...
+// sweep fee rate, which is determined by calculating the average fee rate of
+// all inputs within that cluster.
+func (s *UtxoSweeper) clusterBySweepFeeRate() []inputCluster {
+	bucketInputs := make(map[lnwallet.SatPerKWeight]pendingInputs)
+	inputFeeRates := make(map[wire.OutPoint]lnwallet.SatPerKWeight)
+
+	// First, we'll group together all inputs with similar fee rates. This
+	// is done by determining the fee rate bucket they should belong in.
+	for op, input := range s.pendingInputs {
+		feeRate, err := s.validateFeePreference(input.feePreference)
```
joostjager: Re-suggesting to add `input.GetFeeRate()`. Sweeper is already a large struct and imo it should be made smaller rather than larger if we change things.

sweep/sweeper.go
wpaulino: Nodes can have custom minimum relay fee rates, so it's possible that it's actually above the fee floor we enforce. In any case, we really care about the minimum relay fee rate in this case.

sweep/sweeper.go
wpaulino: Fixed.

sweep/sweeper.go
wpaulino: Why only the static fee rate? An estimate could give us something above our max fee rate for whatever reason.

sweep/sweeper.go
wpaulino: Good point, done.

sweep/sweeper.go
wpaulino: Yes so we can ignore the error, but I opted to log it in case something actually goes wrong for whatever reason.

sweep/sweeper.go
joostjager: Ok, I thought you wrote earlier that estimates would never exceed that.

sweep/sweeper.go
```diff
@@ -252,24 +303,33 @@ func (s *UtxoSweeper) Stop() error {
...
+func (s *UtxoSweeper) SweepInput(input input.Input,
+	feePreference FeePreference) (chan Result, error) {
+
 	if input == nil || input.OutPoint() == nil || input.SignDesc() == nil {
 		return nil, errors.New("nil input received")
 	}
 
+	// Ensure the client provided a sane fee preference.
+	if _, err := s.feeRateForPreference(feePreference); err != nil {
+		return nil, err
```
joostjager: If the fee estimate is currently above the maximum but will possible drop below later, we still want to return an error here? It means that resolvers go into an error state until `lnd` is restarted.

sweep/sweeper.go
```diff
@@ -454,6 +546,62 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 	}
 }
 
+// bucketForFeeReate determines the proper bucket for a fee rate. This is done
+// in order to batch inputs with similar fee rates together.
+func (s *UtxoSweeper) bucketForFeeRate(
+	feeRate lnwallet.SatPerKWeight) lnwallet.SatPerKWeight {
```
joostjager: Return value could just as well be an int or other identifier, as the actual value isn't used other than for grouping.

PR #2455, opened 2019-01-11
[WIP] multi: add new draft key send mode for spontaneous payments 

Roasbeef:
In this commit, we add new feature to `lncli sendpayment` (and the RPC): **the ability to send a payment to a destination _without_ first needing to have an invoice**. This allows users to start exploring a new set of use cases that benefits from this type of spontaneous payment. The payment can also carry additional application-specific data such as an account ID, API call, etc. This new feature requires the [new EOB (extra onion blob parsing) implemented here](https://github.com/lightningnetwork/lightning-onion/pull/31). The coolest part about this new feature is that **it can be used _today_ in the wild as long as both nodes are updated to this branch**! 

An example run on the cli: 
```
⛰   lncli --network=simnet sendpayment --dest=02788242941915ed5a101511b8dfeb6db81e0fcd7546f6a55ef4dedf590a7d7ff4 --final_cltv_delta=144 --amt=100 --key_send
{
    "payment_error": "",
    "payment_preimage": "0b1661fb77071fa5efd2217e380c93068bc3dc6e40e302673ff6373ecd8997b5",
    "payment_route": {
        "total_time_lock": 3299,
        "total_amt": 100,
        "hops": [
            {
                "chan_id": 3610796185681920,
                "chan_capacity": 15000000,
                "amt_to_forward": 100,
                "expiry": 3299,
                "amt_to_forward_msat": 100000,
                "pub_key": "02788242941915ed5a101511b8dfeb6db81e0fcd7546f6a55ef4dedf590a7d7ff4"
            }
        ],
        "total_amt_msat": 100000
    }
}
```
**NOTE**: This is only a draft implementation and while it works today on mainnet out of the box (if both sides are upgraded) much this will likely change. Notably the modifications in the link are very hacky and will likely be refactored in preparation of AMP and the like. 

Fixes #132.

PR #3027, opened 2019-04-30
router+build: update to the latest version of lightning-onion

Roasbeef:
In this commit, as a prep for the latest iteration of the EOB design within the spec, we update the `lightning-onion` repo to the latest version. The second commit in this PR updates any relevant callers to use the new API. This commit was broken off of #2455. The ultimate spontaneous payment implementation will still proceed within #2455 once the re-worked code to encode/decode EOBs from the onion is finished.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -2369,8 +2369,22 @@ func (l *channelLink) processRemoteSettleFails(fwdPkg *channeldb.FwdPkg,
 				outgoingHTLCID: pd.ParentIndex,
 				destRef:        pd.DestRef,
 				htlc: &lnwire.UpdateFailHTLC{
-					Reason: lnwire.OpaqueReason(pd.FailReason),
+					Reason: lnwire.OpaqueReason(
+						pd.FailReason,
+					),
 				},
+			}
+
+			// If the failure message lacks an HMAC (but includes
+			// the 4 bytes for encoding the message and padding
+			// lengths, then this means that we received it as an
+			// UpdateFailMalformedHTLC. As a result, we'll signal
+			// that we need to convert this error within the switch
+			// to an actual error, by encrypting it as if we were
+			// the originating hop.
+			convertedErrorSize := lnwire.FailureMessageLength + 4
+			if len(pd.FailReason) == convertedErrorSize {
+				failPacket.convertedError = true
 			}
 
 			// Add the packet to the batch to be forwarded, and
--- a/htlcswitch/switch_test.go
+++ b/htlcswitch/switch_test.go
@@ -2033,3 +2033,76 @@ func TestMultiHopPaymentForwardingEvents(t *testing.T) {
 		}
 	}
 }
+
+// TestUpdateFailMalformedHTLCErrorConversion tests that we're able to properly
+// convert malformed HTLC errors that originate at the direct link, as well as
+// during multi-hop HTLC forwarding.
+func TestUpdateFailMalformedHTLCErrorConversion(t *testing.T) {
+	t.Parallel()
+
+	// First, we'll create our traditional three hop network.
+	channels, cleanUp, _, err := createClusterChannels(
+		btcutil.SatoshiPerBitcoin*3, btcutil.SatoshiPerBitcoin*5,
+	)
+	if err != nil {
+		t.Fatalf("unable to create channel: %v", err)
+	}
+	defer cleanUp()
+
+	n := newThreeHopNetwork(
+		t, channels.aliceToBob, channels.bobToAlice,
+		channels.bobToCarol, channels.carolToBob, testStartingHeight,
+	)
+	if err := n.start(); err != nil {
+		t.Fatalf("unable to start three hop network: %v", err)
+	}
+
+	assertPaymentFailure := func(t *testing.T) {
+		// With the decoder modified, we'll now attempt to send a
+		// payment from Alice to carol.
+		finalAmt := lnwire.NewMSatFromSatoshis(100000)
+		htlcAmt, totalTimelock, hops := generateHops(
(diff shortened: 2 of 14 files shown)


htlcswitch/link.go
```diff
@@ -2369,10 +2369,24 @@ func (l *channelLink) processRemoteSettleFails(fwdPkg *channeldb.FwdPkg,
...
 
+			// If the failure message lacks an HMAC (but includes
+			// the 4 bytes for encoding the message and padding
+			// lengths, then this means that we received it as an
+			// UpdateFailMalformedHTLC. As a result, we'll signal
+			// that we need to convert this error within the switch
+			// to an actual error, by encrypting it as if we were
+			// the originating hop.
+			convertedErrorSize := lnwire.FailureMessageLength + 4
+			if len(pd.FailReason) == convertedErrorSize {
```
joostjager: This looks like a short cut, checking the length to find out if it is encrypted or not. Isn't there a cleaner way?

If was thinking of encrypting the error already in `handleDownStreamPkt` when the `UpdateFailMalformedHTLC` comes in. Just calling `EncryptFirstHop` there.

joostjager: Maybe that doesn't work if we are the sender.

joostjager:
With regards to the converted error in case we receive an unknown code in `UpdateFailMalformedHTLC`: shouldn't we return a `PermanentChannelFailure` and force close the channel to the peer?

If we keep it open, we may be forced to send out more of these failures, pulling down our reputation with senders.

htlcswitch/link.go
Roasbeef: Currently we always let the switch handle the onion encryption for errors as it holds the circuit which holds the shared secret. 

One alternative path is to modify the channel state machine itself to differentiate between a regular fail, and a malformed fail. We'd then need to also store the malformed fail message on disk within commitment diffs as well as forwarding packages in the database. Additionally, we'd need a new message within the channel state machine as well to maintain the 1:1 mapping between what's stored and the contents of the update log. Overall, it's a much bigger change than what's being proposed in this PR as it'll eliminate this new if statement from this diff, implement the above, then also modify the link to understand this message. 

I'm fond of the fix described above as IMO it implements a more robust solution (recognizes the scenario rather than relying on a length check), however it would increase the scope of this PR and surface area of modified code. If we like the alternative fixed described above, I propose we move forward with it, but also merge this as is since it fixes an existing issue and is small enough to be included in the next release cycle.

Roasbeef:
Right now the returned error when decoding hop iterators doesn't distinguish between a replay and an actual error when decrypting the packet. In both the case of modified packet and an attempted replay, it may actually be the fault of the sender (re-using a packet or messing up with the sphinx packet construction) and not actually the direct peer. Even in the case of a direct peer, force closing inconveniences us more than the peer if it is indeed nefarious.

joostjager:
>Right now the returned error when decoding hop iterators doesn't distinguish between a replay and an actual error when decrypting the packet. In both the case of modified packet and an attempted replay, it may actually be the fault of the sender (re-using a packet or messing up with the sphinx packet construction) and not actually the direct peer. Even in the case of a direct peer, force closing inconveniences us more than the peer if it is indeed nefarious.

I mean just in the case the we receive a `UpdateFailMalformedHTLC` message that has a failure code other than invalid onion version, hmac or key. That isn't something that can be triggered by the sender.

htlcswitch/link.go
joostjager: And how about fixing it like this? (last commit)
https://github.com/lightningnetwork/lnd/compare/master...joostjager:new-onion-structs-alt

To me, it seems that fixing the malformed htlc case is not necessarily needed in 0.6.1. Unless I miss an urgent reason to do it now, I think it should be fixed properly without the len check.

Roasbeef: That diff is incomplete, it only implements part of the change I mentioned above. We must fix it now as otherwise the integration tests fail.

htlcswitch/switch_test.go
```diff
@@ -2033,3 +2033,76 @@ func TestMultiHopPaymentForwardingEvents(t *testing.T) {
...
+
+		fwdingErr := err.(*ForwardingError)
+		failureMsg := fwdingErr.FailureMessage
+		if _, ok := failureMsg.(*lnwire.FailTemporaryChannelFailure); !ok {
+			t.Fatalf("expected temp chan failure instead got: %v",
+				fwdingErr.FailureMessage)
+		}
+	}
+
+	t.Run("multi-hop error conversion", func(t *testing.T) {
```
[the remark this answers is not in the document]
Roasbeef: Since they use `Fatalf` and don't run concurrently, if one of them fails the other one doesn't execute.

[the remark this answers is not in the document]
joostjager: I mean if the first one succeeds, but modifies state in an unexpected way and makes the second one fail.

htlcswitch/link.go
joostjager: From the change you described:

1. Modify the channel state machine itself to differentiate between a regular fail, and a malformed fail. 
2. We'd then need to also store the malformed fail message on disk within commitment diffs as well as forwarding packages in the database.
3. Additionally, we'd need a new message within the channel state machine as well to maintain the 1:1 mapping between what's stored and the contents of the update log. 

Looking through `channel.go`, it does seem that most of this is already in. Which bigger change is still remaining after this diff?

joostjager: With regards to the failing itest: we could postpone fixing the expected message in the itest.

Roasbeef: Fixing the expected message without fixing handling of malformed errors won't resolve anything.

Issue #3003, opened 2019-04-24
peer: enforce pong responses 


Roasbeef:
We send a ping message very 30 seconds or so via the `pingHandler` in order to ensure the connection isn't broken. However, we don't currently _enforce_ that the remote peer sends us a pong back in a timely manner. Doing so might allow us to detect dead connections sooner and remove then, which at times can hold back a stale connection from being re-established. 

# Steps To Completion 

- [ ] [Extend the `pingHandler`](https://github.com/lightningnetwork/lnd/blob/master/peer.go#L1636) to also expect a Pong message reply within a timely manner. The reply interval should be << than the ping interval.

cgst:
I can take a look at this. Quick questions:

1) Should we start the clock from the moment the ping is enqueued in pingHandler or from the moment it's [flushed down the wire](https://github.com/lightningnetwork/lnd/blob/master/peer.go#L1520)? The latter is more accurate but might be overcomplicated.

2) How many consecutive pong timeouts should we allow? Just one?

Roasbeef:
> I can take a look at this. Quick questions:


1. I'd say after we know the full message has been flushed. We maintain two message queues now, and pings are on the non-lazy message queue. 
2. Only a single one. If we send a Ping and they don't reply with a pong, then we ideally shouldn't set any others.

cgst:
Please see #3035.

cgst:
👋

ProofOfKeags:
Alright I'm tackling this as we speak but I want to clear the approach before I begin working on it. I also will use this opportunity to try and upgrade the code modularity around this region.

First off, this seems like a good time to address [this TODO](https://github.com/lightningnetwork/lnd/blob/master/peer/brontide.go#L2234-L2235). Perhaps randomly selecting a value near the message size of `update_add_htlc` or another common message is an appropriate pick.

Secondly, we will store the `numPongBytes :: Maybe u16` selected on the `Brontide` and during the ping launch we will set it to the number of bytes we requested in the most recent ping. We will also extend `readHandler` to check received pong messages against this value, if it is correct, it will reset the value to some semantically null value. During the *queuing* (if we want it on write as a previous iteration of this attempted, we will need to do some more refactoring to keep things clean) of the  message we launch a future action that when it times out it will check this numPongBytes value. If it is a semantic null, it is considered successful, if it is not a semantic null it means the timeout handler beat the pong response and we need to handle the logic for failing a ping.

[...]

PR #3035, opened 2019-05-03
peer: enforce pong responses in a timely manner

cgst:
Make peerHandler expect and enforce timely pong response for every
outgoing ping message.

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -1707,7 +1707,8 @@ out:
 		case <-pongTimeoutChan:
 			err := fmt.Errorf(
 				"timeout of %s reached while waiting"+
-					"for pong message from peer %s", pongTimeout, p,
+					"for pong message from peer %s",
+				pongTimeout, p,
 			)
 			p.Disconnect(err)
 			break out
--- a/peer.go
+++ b/peer.go
@@ -38,6 +38,10 @@ const (
 	// pingInterval is the interval at which ping messages are sent.
 	pingInterval = 1 * time.Minute
 
+	// pongTimeout is the timeout used when waiting for a pong message in
+	// response to a ping. It must be smaller than pingInterval.
+	pongTimeout = pingInterval / 2
+
 	// idleTimeout is the duration of inactivity before we time out a peer.
 	idleTimeout = 5 * time.Minute
 
@@ -226,6 +230,10 @@ type peer struct {
 	// buffer allocation from the peer life cycle.
 	writePool *pool.Write
 
+	// pingPong is a channel that gets written to whenever we successfully
+	// send out a ping message to the peer, or when we receive a pong.
+	pingPong chan lnwire.MessageType
+
 	readPool *pool.Read
 
 	queueQuit chan struct{}
@@ -281,6 +289,7 @@ func newPeer(conn net.Conn, connReq *connmgr.ConnReq, server *server,
 
 		queueQuit: make(chan struct{}),
 		quit:      make(chan struct{}),
+		pingPong:  make(chan lnwire.MessageType, 1),
 	}
 	copy(p.pubKeyBytes[:], nodePub.SerializeCompressed())
 
@@ -375,7 +384,7 @@ func (p *peer) Start() error {
 	go p.writeHandler()
 	go p.readHandler()
 	go p.channelManager()
-	go p.pingHandler()
+	go p.pingHandler(pingInterval, pongTimeout)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


cgst:
Addresses #3003

cgst:
👋

peer.go
```diff
@@ -645,13 +654,20 @@ func (p *peer) Disconnect(reason error) {
...
+	if p.conn != nil {
+		p.conn.Close()
+	}
 
 	close(p.quit)
 }
 
 // String returns the string representation of this peer.
 func (p *peer) String() string {
+	if p.conn == nil {
```
halseth: Ideally we should set it in the test instead, to avoid special casing this.

peer_test.go
```diff
@@ -588,3 +588,55 @@ func TestPeerChannelClosureFeeNegotiationsInitiator(t *testing.T) {
 		t.Fatalf("closing tx not broadcast")
 	}
 }
+
+func TestPeerPingPong(t *testing.T) {
```
halseth: Should have a comment

peer_test.go
```diff
@@ -588,3 +588,55 @@ func TestPeerChannelClosureFeeNegotiationsInitiator(t *testing.T) {
...
+
+	responder, _, _, cleanUp, err := createTestPeer(notifier, broadcastTxChan)
+	if err != nil {
+		t.Fatalf("unable to create test peer: %v", err)
+	}
+	defer cleanUp()
+
+	// Start pingHandler to process internal ping/pong notifications.
+	responder.wg.Add(1)
+	go responder.pingHandler(time.Second/2, time.Second/4)
```
halseth: extract these values into constants.

peer.go
```diff
@@ -1555,6 +1578,16 @@ out:
 				break out
 			}
 
+			// Notify pingHandler of outgoing ping messages.
+			if _, ok := outMsg.msg.(*lnwire.Ping); ok {
```
halseth: can this be done in the `pinhandler` instead?

peer.go
```diff
@@ -1076,6 +1092,13 @@ out:
 			delay := (time.Now().UnixNano() - pingSendTime) / 1000
 			atomic.StoreInt64(&p.pingTime, delay)
 
+			// Notify pingHandler we've received a pong message.
```
halseth: can we just forward ping/pong messages to the pinghandler, and move all the associated logic there?

cfromknecht: i'm okay with it here, as the ping time calculation will be more accurate

peer.go
cfromknecht: would agree if it doesn't add too much code, otherwise we can remove this when we expand our mock connections and use a proper interface inside the peer

Crypt-iQ:
If possible, I think this PR should:
- disconnect with unsolicited pongs
- disconnect if more than X pings are received in Y time

Crypt-iQ:
Followup of my above comment - you could make something like:
```
type pingPongMachine struct {
       state pingPongState
}

type pingPongState uint8

const (
      pingSent pingPongState = 1 << iota
      
      pongReceived

      pingReceived

      pingReceivedTwo

      pingReceivedThree
)

func (pp *pingPongMachine) stateStep(msg lnwire.Message, sent) bool {
       var nextState pingPongState
       switch msg.(type) {
       case *lnwire.Ping:
            if sent {
                  nextState = pingSent
            } else {
                  nextState = pingReceived
            }        
       case *lnwire.Pong:
            // only accept Pongs that we received
            nextState = pongReceived  
       }
       
       switch nextState {
       case pingSent:
             // Check if a pong was received and the pinghandler has ticked at least once before.
             if pp.state&pingSent == 1 && pp.state&pongReceived == 0 {
                  return false
             }
             // Else we received a pong for this and can reset the state
             pp.state = pingSent
       case pongReceived:
             if pp.state&pingSent == 0 {
                   // Received a pong before the ping handler ticked for the first time
                   return false
             } else if pp.state&pongReceived == 1 {
                   // This is an unsolicited pong
                   return false
             }
             pp.state |= pongReceived
       case pingReceived:
             if pp.state&pingReceived == 0 {
                    // Received a ping for the first time this cycle
                    pp.state |= pingReceived
             } else if pp.state&pingReceivedTwo == 0 {
                    // Received a ping for the second time this cycle
                    pp.state |= pingReceivedTwo
             } else if pp.state&pingReceivedThree == 0 {
                    // Received a ping for the third time this cycle
                    pp.state |= pingReceivedThree
             } else {
                    // Received four pings this cycle, fail.
                    return false
             }
       default:
             return false
       }

       return true
}
```

This is a little messy since I made it in the GitHub comment, but basically you get rid of `time.After(pongTimeout)`. If you send a ping out, receive a ping, receive a pong, you forward it over the `pingPong` channel you had and just feed the message and "direction" into the state machine outlined above.  If it returns false, you fail the connection.  The logic permits for the pong to come in at any time between two pings (not just half the time) and permits 3 pings to be received within the two-ping-interval.  It also disallows unsolicited pongs.  Could probably be cleaned up, but this is just like a basic outline of what I was thinking.

Crypt-iQ:
Ping @cgst !

Issue #3005, opened 2019-04-24
rpc+autopilot: allow for arbitrary predicate based channel acceptance criteria


Roasbeef:
Today in the network, the only value a node has that can let it reject a channel is the min channel size parameter. However, there're many other constraints routing nodes may want to apply for accepting/rejecting inbound channels. Accepting an inbound channel in a single funder world is _mostly_ "free", in that you don't need to put any funds up. However, as the network matures, nodes may start to be more discerning w.r.t _which_ channels they accept. For example, a node that may want to position itself as a reliable backbone node, may only want to accept channels from other nodes with a certain channel age, or based on their connections to other node in the network. 

# Steps To Completion 

- [ ] Rather than hard code each of this individual scenarios, we should create a generic system to allow routing node operators to express any sort of criteria they want. At the `fundingManager` level, we can extend it with an interface that resembles something like:
```go
type ChannelAcceptor interface {
    Accept(node *btcec.PublicKey, req *lnwire.OpenChannel) boo
}
```

- [ ] The initial implementation would just always return `true`. As a follow up, we would then refactor the min channel size validation logic to stand behind the `Accept` method. 

- [ ] Next, a new bi-directional streaming RPC should be created. Within the default `Accept` method, if a client is active (there can only be one at a time?), then the default method should be ignored and a client query issued instead. The query should proxy the arguments to `Accept` to the client, who will then return a `bool` indicating if the channel should be accepted or not.

Crypt-iQ:
I can take this if nobody's working on it

Crypt-iQ:
What exactly does this have to do with autopilot (looking at the title)? I thought that autopilot just opened outgoing channels, and all incoming open channel messages were processed by the fundingmanager.  Is that correct?

Crypt-iQ:
Can close this #3039 was merged

PR #3039, opened 2019-05-04
rpc: bi-directional streaming for predicate-based channel acceptance

Crypt-iQ:
My attempt at #3005.  I made sure not to have any deadlocks like the `SendPayment` bi-directional stream has.  The funding key, revocation point, and other points probably aren't necessary for the RPC client, but I sent everything from the `*lnwire.OpenChannel` message for the sake of completeness.  Only one client can be active at a time. Added integration tests.


lnrpc/rpc.proto
```diff
@@ -908,6 +918,85 @@ message SendToRouteRequest {
     Route route = 4;
 }
 
+message ChannelAcceptRequest {
```
Roasbeef: I don't think we need to include _all_ of this information. The keys on our side are auto generated, and don't provide any additional context. Minimally, we can leave off all the keying and revocation information.

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ChannelAcceptor is an interface that represents a predicate on the
+// OpenChannel message and the peer who sent the message.
+type ChannelAcceptor interface {
+	Accept(node *btcec.PublicKey, req *lnwire.OpenChannel) bool
+}
+
+// DefaultChannelAcceptor implements the ChannelAcceptor interface
+// and will be used as the default in the funding manager.
+type DefaultChannelAcceptor struct {
```
Roasbeef: Is this actually the default? This seems like a special case, RPC controlled variant. 

As far as other default, we may also eventually (in distinct PRs), want to provide a set of acceptors that implement thresholds based off of:
  * mission control state
  * autopilot scoring state (can leverage the scoring system to make decisions off of node connectivity, total channel age, etc)
  * reject nodes we already have channels to
  * etc, etc

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
...
+	clientMtx sync.Mutex
+	client    *ChanAcceptorClient
+}
+
+// OpenChannelRequest is a struct containing the requesting node's
+// public key along with the lnwire.OpenChannel message that they
+// sent when requesting to open an inbound channel. This is sent
+// to the RPC streaming client.
+type OpenChannelRequest struct {
+	node        *btcec.PublicKey
```
Roasbeef: Public yet the fields are private?

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
...
+	"sync"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// ChannelAcceptor is an interface that represents a predicate on the
+// OpenChannel message and the peer who sent the message.
+type ChannelAcceptor interface {
+	Accept(node *btcec.PublicKey, req *lnwire.OpenChannel) bool
```
Roasbeef: Why not use `OpenChannelRequest` here?

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
...
+// Accept returns true if no RPC streaming client is specified.
+// Otherwise, if a client is specified, the node's public key along with
+// the OpenChannel message are sent to the client who will make their own
+// decision and return a bool.
+func (d *DefaultChannelAcceptor) Accept(node *btcec.PublicKey,
+	req *lnwire.OpenChannel) bool {
+
+	// If the client is not nil, proxy the arguments to Accept to the RPC
+	// client.
+	if d.client != nil && d.client.active {
```
Roasbeef: I think the "default" can be even simpler, and just return "true". We could then create a higher level acceptor (the RPC one) that _embeds_ the default acceptor. Taking a step even further, we could allow _chaining_ acceptors, returning the conjunction of the registered acceptors back to the client.

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
...
+		case d.client.sendChan <- chanReq:
+		default:
+			// Could not send the OpenChannelRequest, so we return false.
+			return false
+		}
+
+		// Now we wait for a response. If the channel is closed, return false.
+		for {
+			select {
+			case accept, ok := <-d.client.recvChan:
```
Roasbeef: Should listen on some sort of quit channel (or context).

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
...
+				if !ok {
```
Roasbeef: IIRc, when a channel is closed, the default value is returned, so unless we want to add additional logic here, we can just return `accept` directly.

channel_acceptor.go
Roasbeef: What about time out logic? Do we want to return `false` if our request isn't responded to within period `T`? Would seem that processing time will depend on the client side implementation (can it do concurrent requests, is it FIFO, etc).

lnrpc/rpc.proto
```diff
@@ -908,6 +918,85 @@ message SendToRouteRequest {
...
+
+    /// The first commitment point for the initiator.
+    bytes first_commitment_point = 18;
+
+    /// A bit-field which the initiator uses to specify proposed channel behavior.
+    uint32 channel_flags = 19;
+}
+
+message ChannelAcceptResponse {
+    /// Whether or not the client accepts the channel.
```
Roasbeef: If we add a new request/response ID, then the client's implementation will be able to respond to requests out of order, and also ensure we're able to properly dispatch responses to the correct peer. Atm, the funding mgr is single threaded, but that may change in the future as we add more advanced features like internal channel batching.

lnd_test.go
```diff
@@ -13939,6 +13939,91 @@ func testChannelBackupRestore(net *lntest.NetworkHarness, t *harnessTest) {
...
+		// If a response was passed, we send it to the stream.
+		if response != nil {
+			if err := predStream.Send(response); err != nil {
+				t.Fatalf("unable to accept open channel request: %v", err)
+			}
+		}
+	}()
+}
+
+func testPredicateChannelAccept(net *lntest.NetworkHarness, t *harnessTest) {
```
Roasbeef: Integration tests have been moved to a new directory since this PR was created.

channel_acceptor.go
```diff
@@ -0,0 +1,111 @@
+package lnd
```
halseth: possible to move this out of the `lnd` package?

Crypt-iQ: Yup will do

halseth:
Needs rebase!

channel_acceptor.go
Crypt-iQ: I think that's a great idea!  It think all the acceptors should be separated though.  I think there are several approaches to this idea of chaining acceptors:

- If the bitwise `&` is not 1, then reject the open channel request.
- Optionally have some sort of "learning" in which a few acceptors return `false` but the majority return `true` but the node still accepts the open channel request.  I'm thinking that over time this model can be improved where there is some total score based on many different factors like you mentioned above.  If the score is above a threshold, accept.
- If the RPC acceptor is specified, this gives "veto" power over all other acceptors.  I think if the RPC acceptor is defined, the other acceptors can be run to send the result and the data used to reach the result back to the client.  This would allow a client to utilize their own score-based approach.

Kind of a brain-dump and maybe quite complicated but I think that would be an ideal direction for these acceptors.  I think it could be complicated for a user to manually specify what their thresholds / scores for various acceptors are and it would make sense for `lnd` to provide this functionality automagically.  For more advanced routing nodes, they could ultimately determine scores.

lnrpc/rpc.proto
Crypt-iQ: Done

channel_acceptor.go
Crypt-iQ: Done

channel_acceptor.go
Crypt-iQ: Done

channel_acceptor.go
Crypt-iQ: Done

Crypt-iQ: I just did a default of 15 seconds, could change this and have this user-specified.

channel_acceptor.go
Crypt-iQ: Fixed

lnrpc/rpc.proto
Crypt-iQ: Fixed

Crypt-iQ:
PTAL @halseth @Roasbeef 

I changed a bit of the logic and added a `ChainedAcceptor` which can combine 2 or more different `ChannelAcceptor` implementations.  Let me know what you think!

channelacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,180 @@
...
+	responseChan := make(chan bool)
+	r.acceptClients[req.OpenChanMsg.PendingChannelID] = responseChan
+
+	select {
+	case r.acceptChan <- req:
+	case <-r.quit:
+		return false
+	}
+
+	for {
```
cfromknecht: can this ever loop?

channelacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,180 @@
...
+		case rpcResp := <-r.client.RecvChan:
+			// This is an OpenChannelResponse received from the RPC server. We
+			// must forward this to the correct Accept method referenced by the
+			// PendingChannelID.
+			responseChan := r.acceptClients[rpcResp.PendingChannelID]
+
+			select {
+			case responseChan <- rpcResp.Accept:
+			default:
+				break
```
cfromknecht: `break` is unnecessary

channelacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+		case chanReq := <-r.acceptChan:
+			// This is an *OpenChannelRequest received from the Accept method.
+			// We forward this to the RPC server.
+			select {
+			case r.client.SendChan <- chanReq:
+			case <-r.client.closeChan:
+				break
```
cfromknecht: similar comment here wrt to `break`

lnrpc/rpc.proto
```diff
@@ -905,6 +915,58 @@ message SendToRouteRequest {
...
+    bytes node_pubkey = 1;
+
+    /// The hash of the genesis block that the proposed channel resides in.
+    bytes chain_hash = 2;
+
+    /// The pending channel id.
+    bytes pending_chan_id = 3;
+
+    /// The funding amount in satoshis that initiator wishes to use in the channel.
+    int64 funding_amt = 4;
```
cfromknecht: any reason that `int64` was chosen instead of `uint64`? none of these values can be negative

lntest/itest/lnd_test.go
```diff
@@ -13928,6 +13928,93 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+
+	// Open up a bidirectional stream so that Bob can respond to incoming
+	// open channel requests.
+	bobPredStream, err := net.Bob.ChannelAcceptor(ctxt)
+	if err != nil {
+		t.Fatalf("unable to create channel acceptor stream for bob: %v", err)
+	}
+
+	// Accept the channel request.
+	acceptReponse := &lnrpc.ChannelAcceptResponse{
```
cfromknecht: is the pending chanid on all these responses zero?

channelacceptor/rpcacceptor.go
Crypt-iQ: Ah no

Crypt-iQ: Ah no

lnrpc/rpc.proto
Crypt-iQ: Yup they should be uint64

lntest/itest/lnd_test.go
Crypt-iQ: Nope it's set in the `respondChannelAccept` method

channelacceptor/chainedacceptor.go
```diff
@@ -0,0 +1,36 @@
+package channelacceptor
```
Roasbeef: Could shorten the package name to just: `chanacceptor`.

channelacceptor/interface.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// OpenChannelRequest is a struct containing the requesting node's public key
+// along with the lnwire.OpenChannel message that they sent when requesting an
+// inbound channel. This information is provided to each acceptor so that they
+// can each leverage their own decision-making with this information.
+type OpenChannelRequest struct {
+	Node        *btcec.PublicKey
```
Roasbeef: Missing `godoc` comments for these two.

channelacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,180 @@
...
+	// DefaultRPCTimeout is the default time that the Accept method will
+	// listen for responses from the RPC server, after which it will timeout.
+	DefaultRPCTimeout = 15 * time.Second
+)
+
+// OpenChannelResponse is a struct containing the response from the RPC client
+// which includes whether or not they accept the open channel request and to
+// which PendingChannelID the response is to.
+type OpenChannelResponse struct {
+	Accept           bool
```
Roasbeef: Missing `godoc` comments here and below.

channelacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,180 @@
...
+		RecvChan:  make(chan OpenChannelResponse),
+		closeChan: make(chan struct{}),
+		active:    true,
+	}
+
+	return r.client, nil
+}
+
+// UnregisterClient unregisters the RPC streaming client.
+func (r *RPCAcceptor) UnregisterClient() {
```
Roasbeef: Shouldn't this be on the `ChanAcceptorClient` struct? Otherwise, we can't individually unregister clients.

Roasbeef: Nvm, I see how the `RegisterClient` method works now, no strong preference, but it could also be done with an explicit constructor that instead passes in the `ChanAcceptorClient`.

channelacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		// If a client does not exist, continue.
+		if !r.active() {
+			continue
+		}
+
+		select {
+		case rpcResp := <-r.client.RecvChan:
+			// This is an OpenChannelResponse received from the RPC server. We
```
Roasbeef: Would lift the comment here and below to reside right on top of the top-level `select` case.

rpcserver.go
```diff
@@ -5047,3 +5055,110 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+
+// ChannelAcceptor dispatches a bi-directional streaming RPC in which
+// OpenChannel requests are sent to the client and the client responds with
+// a boolean that tells LND whether or not to accept the channel. This allows
+// node operators to specify their own criteria for accepting inbound channels
+// through a single persistent connection. There may only be one client at a
+// time for this RPC.
+func (r *rpcServer) ChannelAcceptor(stream lnrpc.Lightning_ChannelAcceptorServer) error {
+
+	var acceptorClient *channelacceptor.ChanAcceptorClient
```
Roasbeef: Can combine both of these declarations under a single `var` statement:
```go
var (
...
)
```

rpcserver.go
```diff
@@ -5047,3 +5055,110 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+	var rpcAcceptor *channelacceptor.RPCAcceptor
+
+	// If the OpenChannelPredicate is the ChainedAcceptor, locate the RPCAcceptor
+	// and register this RPC client.
+	openChanPredicate := r.server.fundingMgr.cfg.OpenChannelPredicate
+	chainedAcceptor, ok := openChanPredicate.(*channelacceptor.ChainedAcceptor)
+	if !ok {
+		// The funding manager is not using the ChainedAcceptor.
```
Roasbeef: We'll _always_ have the chained acceptor active if this RPC method is also activated?

channelacceptor/chainedacceptor.go
Roasbeef: Would be nice to have some basic coverage tests for this package now that it no longer lives in the main package.

lntest/itest/lnd_test.go
```diff
@@ -13928,6 +13928,93 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+	respondChannelAccept(bobPredStream, acceptReponse, t)
+
+	// Here we are going to test an accepted channel open.
+	ctxt, _ = context.WithTimeout(ctxb, channelOpenTimeout)
+	_ = openChannelAndAssert(
+		ctxt, t, net, net.Alice, net.Bob,
+		lntest.OpenChannelParams{
+			Amt: 100000,
+		},
+	)
```
Roasbeef: All channels that end up being created should be closed at the end of the test.

rpcserver.go
Crypt-iQ: Since the funding manager accepts a `ChannelAcceptor`, I thought that there could be a case where other implementations besides the `ChainedAcceptor` could also be passed in.  But it seems to me with the current design, `ChainedAcceptor` should be the only implementation that should be passed in since it allows more flexibility.  I can change this and the fundingmanager's `start` and `stop` methods.

rpcserver.go
```diff
@@ -5067,3 +5075,108 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+
+	var (
+		acceptorClient *chanacceptor.Client
+		rpcAcceptor    *chanacceptor.RPCAcceptor
+		ok             bool
+	)
+
+	// Locate the RPCAcceptor within the ChainedAcceptor and register this RPC client.
+	openChanPredicate := r.server.fundingMgr.cfg.OpenChannelPredicate
+	chainedAcceptor, _ := openChanPredicate.(*chanacceptor.ChainedAcceptor)
```
Roasbeef: Whenever a type assertion is done, the second variable should _always_ be checked. 

It's also a code smell that we're accessing the _config_ of the funding manager here in order to access the RPC acceptor. Atm, the rpc acceptor is a singleton, so we could add a package level method to obtain the latest version of the acceptor.

rpcserver.go
```diff
@@ -5067,3 +5075,108 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+
+	for _, acceptor := range chainedAcceptor.Acceptors {
```
Roasbeef: Similar comment here re just obtaining the global rpc acceptor instance.

rpcserver.go
```diff
@@ -5067,3 +5075,108 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+			case acceptorClient.RecvChan <- openChanResp:
+			case <-r.quit:
+				return
+			}
+		}
+	}()
+
+	for {
+		select {
+		case req := <-acceptorClient.SendChan:
```
Roasbeef: These channel attributes (to be morphed into methods based on the comments above), would be better named as response/request rather than send/recv. IMO the intent is more clear with the request/response naming.

fundingmanager.go
```diff
@@ -487,6 +493,17 @@ func (f *fundingManager) Start() error {
 func (f *fundingManager) start() error {
 	fndgLog.Tracef("Funding manager running")
 
+	// Locate the RPCAcceptor within the ChainedAcceptor and start it if it exists.
+	openChanPredicate := f.cfg.OpenChannelPredicate
+	chainedAcceptor, _ := openChanPredicate.(*chanacceptor.ChainedAcceptor)
+	for _, acceptor := range chainedAcceptor.Acceptors {
+		rpcAcceptor, ok := acceptor.(*chanacceptor.RPCAcceptor)
+		if ok {
+			rpcAcceptor.Start()
```
Roasbeef: Shouldn't the RPC acceptor _already_ be started if it has been passed in at this point?

fundingmanager.go
```diff
@@ -738,6 +755,17 @@ func (f *fundingManager) Stop() error {
 func (f *fundingManager) stop() error {
 	fndgLog.Infof("Funding manager shutting down")
 
+	// Locate the RPCAcceptor within the ChainedAcceptor and stop it if it exists.
+	openChanPredicate := f.cfg.OpenChannelPredicate
+	chainedAcceptor, _ := openChanPredicate.(*chanacceptor.ChainedAcceptor)
+	for _, acceptor := range chainedAcceptor.Acceptors {
+		rpcAcceptor, ok := acceptor.(*chanacceptor.RPCAcceptor)
+		if ok {
+			rpcAcceptor.Stop()
```
Roasbeef: Abstraction leak here, also same comment about it already being stopped (or stopped elsewhere) at this point.

chanacceptor/chainedacceptor.go
```diff
@@ -0,0 +1,36 @@
...
+	Acceptors []ChannelAcceptor
+}
+
+// Accept evaluates the results of all Acceptors and returns the conjunction
+// of all of these predicates.
+func (c *ChainedAcceptor) Accept(req *OpenChannelRequest) bool {
+	result := true
+
+	for _, acceptor := range c.Acceptors {
+		switch a := acceptor.(type) {
```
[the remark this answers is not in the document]
Crypt-iQ: I think this will be needed in the future when other acceptors are added.

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,199 @@
...
+
+	quit chan struct{}
+	wg   sync.WaitGroup
+}
+
+// Accept is a predicate on the OpenChannelRequest which is sent to the RPC
+// client who will respond with the ultimate decision. This assumes a registered
+// client exists.
+func (r *RPCAcceptor) Accept(req *OpenChannelRequest) bool {
+	// Register a new channel for the PendingChannelID associated with this
```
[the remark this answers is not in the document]
Crypt-iQ: Good point!

chanacceptor/chainedacceptor.go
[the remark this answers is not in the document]
Crypt-iQ: My bad, you're right.  Should all be abstracted away

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,199 @@
...
+		RecvChan:  make(chan OpenChannelResponse),
+		closeChan: make(chan struct{}),
+		active:    true,
+	}
+	r.clientMtx.Unlock()
+
+	return r.client, nil
+}
+
+// UnregisterClient unregisters the RPC streaming client.
```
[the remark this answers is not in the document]
Crypt-iQ: Will there ever be a case where there is more than one client?  How would that work?  If so, this method can then just accept a `ClientID` and call the `Unregister` method on the client whose `ClientID` matches the argument.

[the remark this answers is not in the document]
halseth: I assumed more clients would just mean more chained acceptors. So if 4 clients are active, they must all accept for the channel to be accepted. This would allow more general operation, and also some interesting usecases around distributed acceptance.

rpcserver.go
```diff
@@ -5067,3 +5075,108 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+		rpcAcceptor    *chanacceptor.RPCAcceptor
+		ok             bool
+	)
+
+	// Locate the RPCAcceptor within the ChainedAcceptor and register this RPC client.
+	openChanPredicate := r.server.fundingMgr.cfg.OpenChannelPredicate
+	chainedAcceptor, _ := openChanPredicate.(*chanacceptor.ChainedAcceptor)
+
+	for _, acceptor := range chainedAcceptor.Acceptors {
+		rpcAcceptor, ok = acceptor.(*chanacceptor.RPCAcceptor)
```
halseth: To more easily support multiple RPC clients, and also simplify the registration code a little bit, you could create a `NewRPCAcceptor` method that can be called every time this RPC is called. This acceptor would be strictly tied to this rpc context (so no registration/unregistration would be necessary), and would be added to the current chain of acceptors. When the RPC context is done, it would be removed from the chain.

fundingmanager.go
```diff
@@ -487,6 +493,17 @@ func (f *fundingManager) Start() error {
 func (f *fundingManager) start() error {
 	fndgLog.Tracef("Funding manager running")
 
+	// Locate the RPCAcceptor within the ChainedAcceptor and start it if it exists.
+	openChanPredicate := f.cfg.OpenChannelPredicate
+	chainedAcceptor, _ := openChanPredicate.(*chanacceptor.ChainedAcceptor)
+	for _, acceptor := range chainedAcceptor.Acceptors {
```
halseth: The fundingmanager shouldn't have to be aware of the individual acceptors.

chanacceptor/rpcacceptor.go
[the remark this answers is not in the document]
Crypt-iQ: Yeah like threshold acceptance of channels in a sense.

halseth:
ping @Crypt-iQ  :)

rpcserver.go
Crypt-iQ: Yeah I agree with this approach, I started implementing the other way (multiple clients per RPCAcceptor) and found it more difficult to reason about.

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,199 @@
...
+func (r *RPCAcceptor) processRequests() {
+	defer r.wg.Done()
+	for {
+		select {
+		case <-r.quit:
+			return
+		default:
+		}
+
+		// If a client does not exist, continue.
```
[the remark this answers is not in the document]
Crypt-iQ: Don't think this is necessary since now a client is *always* active per @halseth suggestion

Crypt-iQ:
PTAL @halseth @Roasbeef @cfromknecht 

Addressed comments - each `RPCAcceptor` now includes its own client by default per @halseth and there is now a `ChainedAcceptor` singleton which has an expanded API than it previously did.  Travis coverage is failing (haven't had time to look at this).  Could have expanded on the `acceptor_test` unit tests now that the usage of the `RPCAcceptor` and `ChainedAcceptor` has changed, but figured it was more prudent to get a quicker review on these changes.

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Register a new channel for the PendingChannelID associated with this
+	// OpenChannelRequest.
+	responseChan := make(chan bool)
+	r.acceptClientsMtx.Lock()
+	r.acceptClients[req.OpenChanMsg.PendingChannelID] = responseChan
+	r.acceptClientsMtx.Unlock()
+
+	select {
+	case r.acceptChan <- req:
```
halseth: I find this way of communicating with the RPCserver a bit convulated. 

Not totally sure what would be a better way thought. Would it work having the rpcserver define a closure that would do the demultiplexing? 

```go



NewAcceptor(
    func(req *OpenChannelRequest) bool {
             respChan := make chan
             newRequets <- {req, respChan}

             return <-respChan
    },
)

go func() {
     for {
         resp := stream.Recv()
          resps <- resp
      }
} 

requests := make(map[pendingID]chan response)
for {
     select{
     case req := <-newRequests:
              requests[pendingid] = respChan
             stream.Send(req)
     case resp := <-resps:
              respChan, ok := requests[pendindid]
              if !ok {continue}

             respChan <- resp
             delete(requets, pendingID)
     }

}
```

lntest/itest/lnd_test.go
```diff
@@ -14106,6 +14106,97 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+		ctxt, net.Alice, net.Bob,
+		lntest.OpenChannelParams{
+			Amt: 100000,
+		},
+	)
+	if err == nil || grpc.Code(err) != 202 {
+		t.Fatalf("failed to reject channel: %v", err)
+	}
+
+	// Timeout the channel request.
```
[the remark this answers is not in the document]
Crypt-iQ: `DefaultRPCTimeout` = 15 seconds

chanacceptor/chainedacceptor.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+var (
+	once sync.Once
+
+	// chainedSingleton is the singleton ChainedAcceptor used in the main server.
+	// It is possible to add/remove acceptors to/from this ChainedAcceptor while
+	// it's running. It is also possible to have multiple acceptors of the same
+	// type.
+	chainedSingleton *ChainedAcceptor
```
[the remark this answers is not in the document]
Crypt-iQ: We still need to grab the `ChainedAcceptor` via `GetChainedAcceptor` when adding the `RPCAcceptor` to the acceptors map, so I think it's needed?

chanacceptor/chainedacceptor.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Accept evaluates the results of all ChannelAcceptors in the acceptors map
+// and returns the conjunction of all these predicates.
+func (c *ChainedAcceptor) Accept(req *OpenChannelRequest) bool {
+	result := true
+
+	c.acceptorsMtx.RLock()
+	for _, acceptor := range c.acceptors {
+		result = result && acceptor.Accept(req)
```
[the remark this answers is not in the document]
Crypt-iQ: I don't disagree but what's the rationale here? If the result will be false anyways, what's the point of calling the remaining acceptors?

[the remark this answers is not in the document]
halseth: If we have multiple RPC acceptors, they might be used as some kind of notification that channels are being requested.

chanacceptor/chainedacceptor.go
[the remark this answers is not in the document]
halseth: The RPCserver will call `NewRPCAcceptor` to create the acceptor, then add it to the `ChainedAcceptor`, which has been injected from the server

lntest/itest/lnd_test.go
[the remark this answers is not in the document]
halseth: That's a bit long for a integration test, since it will just sit here?

chanacceptor/chainedacceptor.go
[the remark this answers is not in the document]
Crypt-iQ: Yeah that's what I thought, I'll change to `acceptor.Accept(req) && result`

chanacceptor/chainedacceptor.go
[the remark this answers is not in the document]
Crypt-iQ: Yeah I guess it would be better to access the `server` from `rpcserver`

lntest/itest/lnd_test.go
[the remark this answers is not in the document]
Crypt-iQ: I think the only way to configure this and have the same code coverage would be to specify the timeout when calling the `ChannelAcceptor` rpc method.  I do think it gives the user more control over this new functionality, what do you think?

lnrpc/rpc.proto
```diff
@@ -430,6 +430,16 @@ service Lightning {
...
 
+    /**
+    ChannelAcceptor dispatches a bi-directional streaming RPC in which
+    OpenChannel requests are sent to the client and the client responds with
+    a boolean that tells LND whether or not to accept the channel. This allows
+    node operators to specify their own criteria for accepting inbound channels
+    through a single persistent connection. There may only be one client at a
+    time for this RPC.
+    */
+    rpc ChannelAcceptor (stream ChannelAcceptResponse) returns (stream ChannelAcceptRequest);
```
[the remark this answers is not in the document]
Crypt-iQ: This compiles on my machine but I might be missing something (running `go1.13`)?

chanacceptor/rpcacceptor.go
Crypt-iQ: This would also let us remove `Start` & `Stop` in the `RPCAcceptor`

lnrpc/rpc.proto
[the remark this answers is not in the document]
halseth: It won't compile if you checkout this exact commit :) But no blocker

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,39 @@
...
+
+// RPCAcceptor represents the RPC-controlled variant of the ChannelAcceptor.
+// One RPCAcceptor allows one RPC client.
+type RPCAcceptor struct {
+	acceptClosure func(req *OpenChannelRequest) bool
+}
+
+// Accept is a predicate on the OpenChannelRequest which is sent to the RPC
+// client who will respond with the ultimate decision. This assumes an accept
+// closure has been specified during creation.
```
halseth: add note that this is part of the interface

rpcserver.go
```diff
@@ -5065,3 +5077,160 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+	var acceptRequestsMtx sync.RWMutex
+	acceptRequests := make(map[[32]byte]chan bool)
+
+	for {
+		select {
+		case newRequest := <-newRequests:
+
+			req := newRequest.chanReq
+
+			acceptRequestsMtx.Lock()
```
halseth: No mutex needed since only accessed from this goroutine :)

lntest/itest/lnd_test.go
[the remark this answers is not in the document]
halseth: Yeah, would that work, or would it be hard because of the bidirectional stream? 

I think either way that it should be configurable, as imagine you want to run some fancy algorithm to determine whether to accept the channel, that is super heavy :P

rpcserver.go
Crypt-iQ: Righto!

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,198 @@
...
+	// DefaultRPCTimeout is the default time that the Accept method will
+	// listen for responses from the RPC server, after which it will timeout.
+	DefaultRPCTimeout = 15 * time.Second
+)
+
+// ErrNoChainedAcceptor is returned when the RPCAcceptor is unable to retrieve
+// the ChainedAcceptor singleton.
+var ErrNoChainedAcceptor = errors.New("unable to get chained acceptor")
+
+// OpenChannelResponse is a struct containing the response from the RPC client
```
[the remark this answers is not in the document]
Crypt-iQ: I went with `OpenChannelResponse` since it mirrors `OpenChannelRequest`, but I'm totally open to changing it.  I think `ChannelAcceptResponse` is a better name but it doesn't like implicitly convey that it's the `OpenChannelRequest`s opposite if you catch my drift.

lnrpc/rpc.proto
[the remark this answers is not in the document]
Crypt-iQ: Ahhhh, I'll fix it. Good catch!

lntest/itest/lnd_test.go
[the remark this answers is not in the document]
Crypt-iQ: Should be possible w/ the bidi stream, I agree it should be configurable.  Will report back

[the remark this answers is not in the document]
Crypt-iQ: Since there can only be one input and one output, we can't really specify the timeout unless it's part of the `OpenChannelResponse`, at which point the timeout would apply to the *next* `OpenChannelRequest` message.  I think this approach is confusing.  The timeout could alternatively be specified as a separate rpc or as a command line option when starting lnd.  The downside of the command line option is that it's not mutable unless the user restarts their lnd instance.

rpcserver.go
```diff
@@ -5065,3 +5065,8 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
 		}
 	}
 }
+
+// ChannelAcceptor method stub.
```
halseth: 👍

chanacceptor/rpcacceptor.go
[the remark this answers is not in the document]
halseth: Reason I don't want to use `OpenChannelRequest`/`Response` is that we already have `OpenChannelRequest` and `openChanReq` in used in the codebase iirc 😛

rpcserver.go
```diff
@@ -5066,7 +5078,153 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+// ChannelAcceptor dispatches a bi-directional streaming RPC in which
+// OpenChannel requests are sent to the client and the client responds with
+// a boolean that tells LND whether or not to accept the channel. This allows
+// node operators to specify their own criteria for accepting inbound channels
+// through a single persistent connection.
 func (r *rpcServer) ChannelAcceptor(stream lnrpc.Lightning_ChannelAcceptorServer) error {
-	return nil
+
+	// Locate the ChainedAcceptor within the fundingmanager.
+	openChanPredicate := r.server.fundingMgr.cfg.OpenChannelPredicate
```
halseth: instead of acccessing the funding config here, you can store the predicate as a member of the rpc server. Then you don't need the type assertion below either.

lnrpc/rpc.proto
```diff
@@ -912,6 +921,58 @@ message SendToRouteRequest {
...
+    /// The total number of incoming HTLC's that the initiator will accept.
+    uint32 max_accepted_htlcs = 12;
+
+    /// A bit-field which the initiator uses to specify proposed channel behavior.
+    uint32 channel_flags = 13;
+}
+
+message ChannelAcceptResponse {
+    /// Whether or not the client accepts the channel.
+    bool accept = 1;
```
halseth: it could be useful to return a reject reason here, but can be left for future extension.

server.go
```diff
@@ -896,6 +897,11 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 	if _, err := rand.Read(chanIDSeed[:]); err != nil {
 		return nil, err
 	}
+
+	// Initialize the ChainedAcceptor and add a DefaultAcceptor.
+	chainedAcceptor := chanacceptor.NewChainedAcceptor()
+	chainedAcceptor.AddAcceptor(&chanacceptor.DefaultAcceptor{})
```
halseth: I don't think we need the default acceptor. The chained acceptor can just return true as long as there are no added acceptor.

chanacceptor/acceptor_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func acceptAndIncrementCtr(rpc ChannelAcceptor, req *OpenChannelRequest,
+	ctr *uint32, success chan struct{}) {
+
+	result := rpc.Accept(req)
+	if !result {
+		return
+	}
+
+	atomic.AddUint32(ctr, 1)
```
halseth: you must add and load value in the same operation

rpcserver.go
Crypt-iQ: Lmk what you think about my approach

server.go
```diff
@@ -294,7 +295,8 @@ func noiseDial(idPriv *btcec.PrivateKey) func(net.Addr) (net.Conn, error) {
 func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 	towerClientDB *wtdb.ClientDB, cc *chainControl,
 	privKey *btcec.PrivateKey,
-	chansToRestore walletunlocker.ChannelsToRecover) (*server, error) {
+	chansToRestore walletunlocker.ChannelsToRecover,
+	chanPredicate chanacceptor.ChannelAcceptor) (*server, error) {
```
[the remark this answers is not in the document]
Crypt-iQ: From what I can tell it is? Compiled on my machine when I checked out this commit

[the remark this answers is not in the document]
halseth: Ah, not sure why I thought otherwise 😅

Crypt-iQ:
My only complaint with this PR is that the timeout isn't currently configurable.  If somebody wanted to make this interact with a mobile app, a user would only have 15 seconds to decide if they want this open channel.  So it only really works with people who have automated scripts ready-to-go.  On the other hand, the funding manager is currently single-threaded so if a user took a long time to accept/reject channel, it may block funding flows for other channels.

chanacceptor/rpcacceptor.go
```diff
@@ -0,0 +1,41 @@
+package chanacceptor
+
+// ChannelAcceptResponse is a struct containing the response from the RPC client
+// which includes whether or not they accept the open channel request and to
+// which PendingChannelID the response is to.
+type ChannelAcceptResponse struct {
```
halseth: Looks like this is a simple type that now can be defined locally in the rpcserver, where it is used.

rpcserver.go
```diff
@@ -76,6 +77,10 @@ var (
 	// It is set to the value under the Bitcoin chain as default.
 	MaxPaymentMSat = maxBtcPaymentMSat
 
+	// defaultAcceptorTimeout is the time after which an RPCAcceptor will time
+	// out and return false if it hasn't yet received a response.
```
halseth: add a TODO to make this configurable?

rpcserver.go
```diff
@@ -5066,7 +5084,163 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+			resp, err := stream.Recv()
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			var pendingID [32]byte
+			copy(pendingID[:], resp.PendingChanId)
+
+			openChanResp := chanacceptor.ChannelAcceptResponse{
```
halseth: actually, looks like we can send the `lnrpc` proto on `responses` directly, and remove `ChannelAcceptResponse`

Crypt-iQ: Good point

chanacceptor/chainedacceptor.go
```diff
@@ -0,0 +1,65 @@
...
+//
+// NOTE: Part of the ChannelAcceptor interface.
+func (c *ChainedAcceptor) Accept(req *ChannelAcceptRequest) bool {
+	result := true
+
+	c.acceptorsMtx.RLock()
+	for _, acceptor := range c.acceptors {
+		// We call Accept first in case any acceptor (perhaps an RPCAcceptor)
+		// wishes to be notified about ChannelAcceptRequest.
+		result = acceptor.Accept(req) && result
```
Roasbeef: Non-blocking style nit: can be
```go
result &= acceptor.Accept(req)
```

halseth: I don't think we can, since that would short cut the call to `Accept`, see discussion https://github.com/lightningnetwork/lnd/pull/3039#discussion_r322605966

rpcserver.go
```diff
@@ -5052,7 +5072,165 @@ func (r *rpcServer) SubscribeChannelBackups(req *lnrpc.ChannelBackupSubscription
...
+
+			openChanResp := lnrpc.ChannelAcceptResponse{
+				Accept:        resp.Accept,
+				PendingChanId: pendingID[:],
+			}
+
+			// Now that we have the response from the RPC client, send it to
+			// the responses chan.
+			select {
+			case responses <- openChanResp:
```
halseth: realized we also need to select on quit here, to handle the case where we shut down the acceptor right as the response comes in.

halseth: I think when we have that in this is ready to land!

Crypt-iQ: @halseth donezo. Also thanks for the extensive review on this PR!

PR #3046, opened 2019-05-06
htlcswitch+routing: handle unknown failure source

joostjager:
This PR modifies the switch to properly signal the case where the error source is unknown. Previously the self node was reported as the source of the error.

Handling of an unknown error source in router is modified to prune all channel edges along the route, as the cause of this error could be any node.


joostjager:
Combined with #3188

PR #3054, opened 2019-05-07
lnrpc+routing: remove k shortest path finding

joostjager:
This PR is the follow up of #2497. After deprecating the `num_routes` parameter in `lnd` version 0.6, the parameter and all k-shortest related code can now be removed.


joostjager:
> LGTM - only minor thing is that the RPC is still named `QueryRoutes`.

I didn't think about that, but I changing that would break all clients, even the ones that only request a single route already now. Not renaming it requires users that rely on multiple routes to notice this change, because their application may silently behave differently.

C-Otto:
Could you add some information for those users who need more than a single route?

joostjager:
@C-Otto you should be able to implement this client side by using the ignore parameters of queryroutes.

C-Otto:
@joostjager Yes, I learned about the Router RPC calls that give the information I need. Attempt a payment, figure out why it fails, ignore the culprit node/edge, try again. A bit cumbersome, but it might work. See the discussion/progress in https://github.com/C-Otto/rebalance-lnd/issues/51 and https://github.com/lightningnetwork/lnd/issues/3206 in case you're interested.

PR #3059, opened 2019-05-08
lnd: close the wallet unlock grpc server

champo:
The server was kept alive long after it stopped being used. This caused
problems for services using long-lived GRPC connections which might be
created before wallet unlocked. They got stuck connected to the wallet
unlock service needing a restart.

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/lnd.go
+++ b/lnd.go
@@ -879,6 +879,7 @@ func waitForWalletPassword(grpcEndpoints, restEndpoints []net.Addr,
 	// Set up a new PasswordService, which will listen for passwords
 	// provided over RPC.
 	grpcServer := grpc.NewServer(serverOpts...)
+	defer grpcServer.GracefulStop()
 
 	chainConfig := cfg.Bitcoin
 	if registeredChains.PrimaryChain() == litecoinChain {


lnd.go
```diff
@@ -713,6 +713,8 @@ func waitForWalletPassword(grpcEndpoints, restEndpoints []net.Addr,
 	// Set up a new PasswordService, which will listen for passwords
 	// provided over RPC.
 	grpcServer := grpc.NewServer(serverOpts...)
+	// This server is not used outside this function
```
wpaulino: Nit: I think the `defer` already implies this comment, so could be removed.

halseth:
Do the same to the main `grpcServer`?

lnd.go
```diff
@@ -713,6 +713,8 @@ func waitForWalletPassword(grpcEndpoints, restEndpoints []net.Addr,
...
+	defer grpcServer.GracefulStop()
```
halseth: should we use `Stop()` instead? From the doc it says that 
```
GracefulStop stops the gRPC server gracefully. It stops the server from accepting new connections and RPCs and blocks until all the pending RPCs are finished.
```
which sounds like could lock shutdown.

champo: My first try was with stop, as a side effect the wallet unlock call fails in the eyes of the caller since their grpc connection is dropped mid call. graceful stop will wait until all rpc finished so the unlocker gets his response correctly. 

regarding locking, it can lock waiting for existing calls to finish. I'm not sure that's a problem in practice here since the wallet unlocker service exposes simple calls (nothing long lived like sendpayment). If you feel this is a potential issue, the defer can run a goroutine.

halseth: Ah, yeah in that case it makes sense to do it gracefully. Would possibly be more risky to do graceful stop of the main grpc server.

champo: I guess it's a design choice for the main grpc server. Do you want to shutdown immediately? Or is it a good idea to wait while a payment is executing (which could possible take a while)?

AFAIK in the current state of LND in terms of continuing a payment across restarts I feel a GracefulShutdown would be best. But once the whole reliable payments family of PRs land it stops being that critical to be graceful about it.

federicobond:
@halseth hi! Any updates on this?

lnd.go
halseth: Yeah, comment can be removed.

champo:
Sorry about the delay, I overlooked this. I've removed the comment and rebased against master.

PR #3061, opened 2019-05-09
tlv: add library for new message/payload serialization format

cfromknecht:
An implementation of https://github.com/lightningnetwork/lightning-rfc/pull/607/


tlv/varint.go
```diff
@@ -0,0 +1,109 @@
...
+
+	case val <= 0xffffffff:
+		buf[0] = uint8(0xfe)
+		binary.LittleEndian.PutUint32(buf[1:5], uint32(val))
+		_, err := w.Write(buf[:5])
+		return err
+
+	default:
+		buf[0] = uint8(0xff)
+		_, err := w.Write(buf[:1])
```
joostjager: Do we need a 9-byte buffer?

tlv/stream_test.go
```diff
@@ -0,0 +1,408 @@
...
+				tlv.MakePrimitiveRecord(5, &o.u64),
+			)
+			return o, s
+		},
+		expObj: &testObj{
+			u16: 2,
+			u64: 7,
+		},
+	},
+	{
```
joostjager: Another test case may be that there is no overlap at all between encode and decode types

tlv/stream_test.go
```diff
@@ -0,0 +1,408 @@
...
+		expObj: &testObj{
+			u16: 2,
+			u32: 3,
+			u64: 7,
+			buf: bytes.Repeat([]byte{0x05}, 3),
+		},
+	},
+}
+
+func TestTLV(t *testing.T) {
```
joostjager: Function description missing.

tlv/stream_test.go
```diff
@@ -0,0 +1,408 @@
...
+		t.Fatalf("unable to decode tlv stream: %v", err)
+	}
+
+	if !reflect.DeepEqual(test.expObj, obj2) {
+		t.Fatalf("mismatch object, want: %v, got: %v",
+			test.expObj, obj2)
+	}
+}
+
+type thing struct {
```
joostjager: This thing thing, can it be moved to a separate benchmark file?

tlv/stream_test.go
```diff
@@ -0,0 +1,408 @@
...
+		&t.known,
+		&t.field1,
+		&t.field9,
+		&t.field10,
+		&t.field100,
+		&t.field101,
+	)
+}
+func (t *thing) Decode(r io.Reader) error {
+	err := wtwire.ReadElement(r, &t.known, 0)
```
joostjager: This could use some explanation why `known` is not in the tlv stream.

tlv/stream_test.go
```diff
@@ -0,0 +1,408 @@
...
+	if err != nil {
+		return err
+	}
+
+	return t.tlvStream.Decode(r)
+}
+
+func BenchmarkEncodeTLV(t *testing.B) {
```
joostjager: Test description would be good here to explain the different cases.

joostjager:
Nice work in this PR. It shows careful engineering.

tlv/varint.go
cfromknecht: i considered it but decided against it to promote cache alignment. in practice we will probably never read/write a number this big, so seemed okay to special case

tlv/varint.go
```diff
@@ -0,0 +1,109 @@
...
+	"encoding/binary"
+	"errors"
+	"io"
+)
+
+// ErrCanonicalVarInt signals that the decoded varint was not minimally encoded.
+var ErrCanonicalVarInt = errors.New("decoded varint is not canonical")
+
+// ReadVarInt reads a variable length integer from r and returns it as a uint64.
+func ReadVarInt(r io.Reader, buf *[8]byte) (uint64, error) {
```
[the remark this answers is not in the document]
cfromknecht: yes, and quite a significant one. passing `buf` into an `io.Writer` will cause it to escape to the heap. each [de]-serialization of a varint would then allocate 8-bytes to the heap , rather than one for the entire stream that can be reused for encoding/decoding.

watchtower/wtwire/wtwire.go
```diff
@@ -33,27 +34,55 @@ func WriteElement(w io.Writer, element interface{}) error {
 			return err
 		}
 
+	case *uint16:
+		var b [2]byte
+		binary.BigEndian.PutUint16(b[:], *e)
+		if _, err := w.Write(b[:]); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
cfromknecht: commit removed

watchtower/wtwire/wtwire.go
```diff
@@ -33,27 +34,55 @@ func WriteElement(w io.Writer, element interface{}) error {
...
 		if _, err := w.Write(b[:]); err != nil {
 			return err
 		}
 
+	case *uint64:
+		var b [8]byte
+		binary.BigEndian.PutUint64(b[:], *e)
+		if _, err := w.Write(b[:]); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
cfromknecht: ended up making this change in the varint encoding

tlv/stream_test.go
```diff
@@ -0,0 +1,408 @@
...
+			)
+			return o, s
+		},
+		expObj: &testObj{
+			u16: 2,
+			u32: 3,
+			u64: 7,
+			buf: bytes.Repeat([]byte{0x05}, 3),
+		},
+	},
```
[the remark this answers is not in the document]
cfromknecht: i've added a few more cases, still looking to add a few more for incorrect lengths

tlv/record.go
```diff
@@ -0,0 +1,99 @@
...
+	case *uint64, *int64:
+		staticSize = 8
+	case *[32]byte:
+		staticSize = 32
+	case *[33]byte, **btcec.PublicKey:
+		staticSize = 33
+	case *lnwire.Sig:
+		staticSize = 64
+	case *[]byte:
+		sizeFunc = SizeVarBytes(e)
```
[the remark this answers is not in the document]
cfromknecht: changes are up

tlv/stream.go
```diff
@@ -0,0 +1,251 @@
...
+		min = typ + 1
+	}
+}
+
+// getRecord searches for a record matching typ known to the stream. The boolean
+// return value indicates whether the record is known to the stream. The integer
+// return value carries the index from where getRecord should be invoked on the
+// subsequent call. The first call to getRecord should always use an idx of 0.
+func (s *Stream) getRecord(typ Type, idx int) (Record, int, bool) {
+	for idx < len(s.records) {
```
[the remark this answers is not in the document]
joostjager: Ok, clear

tlv/record.go
```diff
@@ -0,0 +1,147 @@
...
+	var (
+		staticSize uint16
+		sizeFunc   SizeFunc
+		encoder    Encoder
+		decoder    Decoder
+	)
+	switch e := val.(type) {
+	case *uint8, *int8:
+		staticSize = 1
+		encoder = EUint8
```
Roasbeef: I think the commits were re-ordered? This type is used before its defined in the series.

tlv/record.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Encoder is a signature for methods that can encode TLV values. An error
+// should be returned if the Encoder cannot support the underlying type of val.
+// The provided scratch buffer must be non-nil.
+type Encoder = func(w io.Writer, val interface{}, buf *[8]byte) error
+
+// Decoder is a signature for methods that can decode TLV values. An error
+// should be returned if the Decoder cannot support the underlying type of val.
+// The provided scratch buffer must be non-nil.
+type Decoder = func(r io.Reader, val interface{}, buf *[8]byte, l uint64) error
```
Roasbeef: Unnecessary equals here?

tlv/stream.go
```diff
@@ -0,0 +1,259 @@
...
+func (t ErrUnknownRequiredType) Error() string {
+	return fmt.Sprintf("unknown required type: %d", t)
+}
+
+// RecordEncoder specifies a codec for serializing a given record.
+type RecordEncoder func(io.Writer, interface{}) error
+
+// RecordDecoder specifies a codec for deserializing given record. The final
+// argument is the expected length, and the method should return an error if
+// thee number of bytes read does not match this value.
```
Roasbeef: thee -> the

tlv/stream.go
```diff
@@ -0,0 +1,259 @@
...
+	records []Record
+	buf     [8]byte
+}
+
+// NewStream creates a new TLV Stream given an encoding codec, a decoding codec,
+// and a set of known records.
+func NewStream(records ...Record) *Stream {
+	// Assert that the ordering of the Records is canonical and appear in
+	// ascending order of type.
+	var min Type
```
Roasbeef: Variables can be defined under the same `var` block.

tlv/stream.go
```diff
@@ -0,0 +1,259 @@
...
+			panic(ErrStreamNotCanonical.Error())
+		}
+		if record.encoder == nil {
+			panic(ErrNoEncoder.Error())
+		}
+		if record.decoder == nil {
+			panic(ErrNoDecoder.Error())
+		}
+		if record.typ == math.MaxUint64 {
+			overflow = true
```
Roasbeef: If the final type is what overflows, then it appears we won't panic at all?

tlv/stream.go
```diff
@@ -0,0 +1,259 @@
...
+	var overflow bool
+	for _, record := range records {
+		if overflow || record.typ < min {
+			panic(ErrStreamNotCanonical.Error())
+		}
+		if record.encoder == nil {
+			panic(ErrNoEncoder.Error())
+		}
+		if record.decoder == nil {
+			panic(ErrNoDecoder.Error())
```
Roasbeef: Depending on the context in which this is used within `lnd` itself, we may actually want to return an error instead.

tlv/stream.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	return nil
+}
+
+// Decode deserializes TLV Stream from the passed io.Reader. The Stream will
+// inspect each record that is parsed and check to see if it has a corresponding
+// Record to facilitate deserialization of that field. If the record is unknown,
+// the Stream will discard the record's bytes and proceed to the subsequent
+// record.
```
Roasbeef: > If the record is unknown,
> // the Stream will discard the record's bytes and proceed to the subsequent
> // record.

Is this subject to the even/odd rule? Also perhaps we should allow a "strict" stream that'll always error our rather than silently go to the next record to be decoded.

tlv/stream.go
```diff
@@ -0,0 +1,259 @@
...
+//    (t: 0x02, l: 0x01, v: 0x01)
+//
+// This method asserts that the byte stream is canonical, namely that each
+// record is unique and that all records are sorted in ascending order. An
+// ErrNotCanonicalStream error is returned if the encoded TLV stream is not.
+//
+// We permit an io.EOF error only when reading the type byte which signals that
+// the last record was read cleanly and we should stop parsing. All other io.EOF
+// or io.ErrUnexpectedEOF errors are returned.
+func (s *Stream) Decode(r io.Reader) error {
```
Roasbeef: Not a blocker for this PR, but this method would make for a good fuzz testing target. Even before the, we can assert compliance of our encoder by using `testing/quick` here to ensure we can always ingest what we produce, and don't inadvertently create any invalid streams.

tlv/record.go
```diff
@@ -0,0 +1,147 @@
...
+
+// SizeVarBytes is a SizeFunc that can compute the length of a byte slice.
+func SizeVarBytes(e *[]byte) SizeFunc {
+	return func() uint16 {
+		return uint16(len(*e))
+	}
+}
+
+// Record holds the required information to encode or decode a TLV record.
+type Record struct {
```
Roasbeef: I don't think it matters if the padding is at the middle or at the end, but `structlayout` recommends a slightly different ordering: 
```
⛰   structlayout github.com/lightningnetwork/lnd/tlv Record
Record.typ github.com/lightningnetwork/lnd/tlv.Type: 0-8 (size 8, align 8)
Record.staticSize uint16: 8-10 (size 2, align 2)
padding: 10-16 (size 6, align 0)
Record.value interface{}: 16-32 (size 16, align 8)
Record.sizeFunc github.com/lightningnetwork/lnd/tlv.SizeFunc: 32-40 (size 8, align 8)
Record.encoder func(w io.Writer, val interface{}, buf *[8]byte) error: 40-48 (size 8, align 8)
Record.decoder func(r io.Reader, val interface{}, buf *[8]byte, l uint64) error: 48-56 (size 8, align 8)
```

```
⛰   structlayout -json github.com/lightningnetwork/lnd/tlv Record | structlayout-optimize
Record.value interface{}: 0-16 (size 16, align 8)
Record.typ github.com/lightningnetwork/lnd/tlv.Type: 16-24 (size 8, align 8)
Record.sizeFunc github.com/lightningnetwork/lnd/tlv.SizeFunc: 24-32 (size 8, align 8)
Record.encoder func(w io.Writer, val interface{}, buf *[8]byte) error: 32-40 (size 8, align 8)
Record.decoder func(r io.Reader, val interface{}, buf *[8]byte, l uint64) error: 40-48 (size 8, align 8)
Record.staticSize uint16: 48-50 (size 2, align 8)
padding: 50-56 (size 6, align 0)
```

tlv/primitive.go
```diff
@@ -0,0 +1,272 @@
...
+
+	return ErrTypeForDecoding{
+		val:       val,
+		expType:   expType,
+		valLength: valLength,
+		expLength: expLength,
+	}
+}
+
+func (e ErrTypeForDecoding) Error() string {
```
Roasbeef: Missing `godoc` comment.

Roasbeef:
Reviewee bump!

tlv/stream.go
```diff
@@ -0,0 +1,283 @@
...
+			case err == io.EOF:
+				return io.ErrUnexpectedEOF
+
+			// Other unexpected errors.
+			case err != nil:
+				return err
+			}
+		}
+
+		// Update our record index so that we can begin our next search
```
halseth: Clever! 😀

tlv/primitive.go
```diff
@@ -0,0 +1,272 @@
...
+		return err
+	}
+	return ErrTypeForDecoding{val, "[64]byte", l, 64}
+}
+
+// EPubKey is an Encoder for *btcec.PublicKey values. An error is returned if
+// val is not a **btcec.PublicKey.
+func EPubKey(w io.Writer, val interface{}, _ *[8]byte) error {
+	if pk, ok := val.(**btcec.PublicKey); ok {
+		_, err := w.Write((*pk).SerializeCompressed())
```
halseth: Do we really need methods for `PublicKey`, or could we just require the caller to use `[33]byte`?

tlv/primitive.go
```diff
@@ -73,6 +73,15 @@ func EUint8(w io.Writer, val interface{}, buf *[8]byte) error {
 	return ErrTypeForEncoding{val, "uint8"}
 }
 
+// EUint8T encodes a uint8 val to the provided io.Writer. This method is exposed
```
halseth: Commit message should have an explanation why these methods are needed. Not obvious to me.

tlv/stream.go
cfromknecht: if the final type is `math.MaxUint64` that's okay, so long as there isn't a number after it. `overflow` in this case signifies that a record hit `math.MaxUint64`, and any other records should be forbidden

tlv/stream.go
cfromknecht: agreed, i'll change this to use `Must` paradigm employed by the go std library

tlv/primitive.go
cfromknecht: the test vectors in the spec require us to fail when reading an invalid point, so only parsing `[33]byte` isn't strict enough. arguably that check could be deferred to a higher level, but it this is the safest approach

tlv/record.go
cfromknecht: good suggestion, done!

tlv/stream.go
```diff
@@ -0,0 +1,283 @@
...
+// encoding, and nothing else. The ordering of the Records is asserted upon the
+// creation of a Stream, and thus the output will be by definition canonical.
+func (s *Stream) Encode(w io.Writer) error {
+	// Iterate through all known records, if any, serializing each record's
+	// type, length and value.
+	for i := range s.records {
+		rec := &s.records[i]
+
+		// Write the record's type as a varint.
+		err := WriteVarInt(w, uint64(rec.typ), &s.buf)
```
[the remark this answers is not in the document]
cfromknecht: it gets allocated to the heap regardless, so i thought it made sense to just do it in the same allocation as the stream. it also allows the same buffer to be reused for both encoding and decoding

tlv/stream.go
cfromknecht: yes a strict mode would definitely be useful. the original version of this pr included that, but i removed it for now to simplify the diff

tlv/stream.go
cfromknecht: hmm, got a better name than `tlv.MustStream`?

tlv/record.go
cfromknecht: i think it might just be a gh thing? the commits look in order locally

tlv/stream_test.go
cfromknecht: removed entirely as it was just initial testing code

tlv/stream_test.go
cfromknecht: added

tlv/stream_test.go
cfromknecht: also removed, mostly just a PoC to show that you could extend known message fields with a tlv stream

tlv/stream.go
cfromknecht: ooo good idea, maybe we can add it to the fuzzing harnesses that @Crypt-iQ has been working on

tlv/tlv_test.go
```diff
@@ -0,0 +1,552 @@
...
+	{
+		name:   "type wraparound",
+		bytes:  []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00, 0x00},
+		expErr: tlv.ErrStreamNotCanonical,
+	},
+}
+
+func TestTLVDecodingFailures(t *testing.T) {
+	for _, test := range tlvDecodingFailureTests {
+		t.Run(test.name, func(t *testing.T) {
```
joostjager: The bolt spec has numbered the test cases. I was thinking, would it be worth to prefix `test.name` with an index in this loop?

tlv/varint_test.go
```diff
@@ -0,0 +1,225 @@
...
+	},
+	{
+		Name:  "eight byte high",
+		Value: math.MaxUint64,
+		Bytes: []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff},
+	},
+	{
+		Name:   "two byte not canonical",
+		Bytes:  []byte{0xfd, 0x00, 0xfc},
+		ExpErr: tlv.ErrVarIntNotCanonical.Error(),
```
joostjager: Can `Error()` be called when the comparison is done?

tlv/stream.go
Roasbeef: Also I actually rely on the `decoder` being nil for my TLV-EOB PR within the router. When we receive a fully serialized TLV record stream over the RPC from the user (custom route for `SendToRoute`) I attach a shim encoder that just writes the raw bytes, and I have no need for a decoder since we'll never decode into the record. As a work around, I guess I can provide one that just does nothing instead since it's a special case.

cfromknecht: perhaps instead we could make a nop encoder/decoder and automatically set them if no encoder or decoder is provided

tlv/varint_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"testing"
+
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+type HexBytes []byte
+
+func (b HexBytes) MarshalJSON() ([]byte, error) {
+	return json.Marshal(hex.EncodeToString(b))
+}
```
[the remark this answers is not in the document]
cfromknecht: good catch, was left over from creating the test vectors

tlv/bench_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Decode reads the CreateSessionTLV from the passed io.Reader.
+func (c *CreateSessionTLV) Decode(r io.Reader) error {
+	return c.tlvStream.Decode(r)
+}
+
+// BenchmarkEncodeCreateSession benchmarks encoding of the non-TLV
+// CreateSession.
+func BenchmarkEncodeCreateSession(t *testing.B) {
```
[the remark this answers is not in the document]
cfromknecht: yes there's a good bit of optimization that could be done to non-tlv messages. amongst other things, this was one of the optimizations included in https://github.com/btcsuite/btcd/pull/1426. i don't think we'd see as significant of gains for `lnwire` messages since the primary speed up was due to reducing RTTs with the buffer pool, but this certainly helps remove unnecessary allocations and gc pressure

tlv/tlv_test.go
cfromknecht: the formatting in the spec is a little weird, indeed it is numbered, but each case is actually two lines so they won't line up unless we number them as `2*i+1`

tlv/truncated.go
```diff
@@ -0,0 +1,167 @@
...
+
+// numLeadingZeroBytes16 computes the number of leading zeros for a uint16.
+func numLeadingZeroBytes16(v uint16) uint64 {
+	var mask = uint16(0xff << 8)
+	for i := uint64(0); i < 2; i++ {
+		if mask&v > 0 {
+			return i
+		}
+		mask >>= 8
+	}
```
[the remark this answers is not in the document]
cfromknecht: ended up unrolling all of them, f901b9a8d5eca2541eaa091d2e58a1d745db4622

tlv/varint_test.go
```diff
@@ -0,0 +1,225 @@
...
+			testReadVarInt(t, test)
+		})
+	}
+}
+
+func testReadVarInt(t *testing.T, test varIntTest) {
+	var buf [8]byte
+	r := bytes.NewReader(test.Bytes)
+	val, err := tlv.ReadVarInt(r, &buf)
+	if err != nil && err.Error() != test.ExpErr {
```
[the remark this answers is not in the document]
cfromknecht: fixed 8d16117d06722c2908f10d37e28d471e398a39a2

tlv/varint_test.go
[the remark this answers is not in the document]
cfromknecht: fixed 8d16117d06722c2908f10d37e28d471e398a39a2

tlv/tlv_test.go
```diff
@@ -0,0 +1,552 @@
...
+		}
+		if err := tlv.DUint64(r, &t.amt1, buf, 8); err != nil {
+			return err
+		}
+		return tlv.DUint64(r, &t.amt2, buf, 8)
+	}
+	return tlv.NewTypeForDecodingErr(val, "nodeAmts", l, 49)
+}
+
+type N1 struct {
```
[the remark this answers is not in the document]
cfromknecht: test files can't be imported so it shouldn't make a difference, originally i defined it this way so that variable names could be `n1` or `n2`.

tlv/tlv_test.go
```diff
@@ -0,0 +1,552 @@
...
+}
+
+func TestTLVDecodingFailures(t *testing.T) {
+	for _, test := range tlvDecodingFailureTests {
+		t.Run(test.name, func(t *testing.T) {
+			n1 := NewN1()
+			r := bytes.NewReader(test.bytes)
+
+			err := n1.Decode(r)
+			if err.Error() != test.expErr.Error() {
```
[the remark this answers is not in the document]
cfromknecht: fixed 2792cab2e

tlv/varint_test.go
cfromknecht: fixed 8d16117d06722c2908f10d37e28d471e398a39a2

tlv/truncated.go
```diff
@@ -0,0 +1,167 @@
...
+	"encoding/binary"
+	"errors"
+	"io"
+)
+
+// ErrTUintNotMinimal signals that decoding a truncated uint failed because the
+// value was not minimally encoded.
+var ErrTUintNotMinimal = errors.New("truncated uint not minimally encoded")
+
+// numLeadingZeroBytes16 computes the number of leading zeros for a uint16.
```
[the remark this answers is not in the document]
cfromknecht: the spec tests the 32 and 64 bit variants, tho not the uint16

tlv/tlv_test.go
```diff
@@ -0,0 +1,552 @@
...
+
+func (n *N2) Decode(r io.Reader) error {
+	return n.stream.Decode(r)
+}
+
+var tlvDecodingFailureTests = []struct {
+	name   string
+	bytes  []byte
+	expErr error
+	skipN2 bool
```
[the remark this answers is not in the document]
cfromknecht: fixed aa2f1d9ed

tlv/bench_test.go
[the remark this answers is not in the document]
joostjager: Ok, interesting. Yes, my main thinking was that for a fair comparison, both non-tlv and tlv should be as optimized as possible. But then there is also the "custom tlv encoder/decoder" method, where (unrolled) code is generated for a specific stream. From a theoretical pov it would be nice to see how all three of them compare. What the absolute minimum overhead of tlv is. Definitely not the most important question atm.

tlv/tlv_test.go
joostjager: Yes, I saw the strange numbering. It was just an idea, nothing important.

tlv/bench_test.go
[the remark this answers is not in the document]
cfromknecht: agreed, i too would be interested to see that. my goal here was to squeeze some low hanging fruit out of tlv, though as you said the actual overhead might be a little higher if we optimize the traditional encoding. there's also the question of even if those optimizations were possible, would they ever be deployed. if we think tlv will be the base encoding for our messages going forward, we may have to just accept whatever it is :P

Chinwendu20:
Do you think it would be a good idea to include a boolean to the primitives?

PR #3062, opened 2019-05-09
lnd: add litecoin support for Neutrino

Wowee0:
Please could you review the following proposal for enabling litecoin support in LND/Neutrino :

LND/config.go :
allow the configuration of litecoin.node=neutrino

LND/chainparams.go/applyLitecoinParams :
apply some litecoin chainparams settings that were missed in the original implementation, namely:
GenesisBlock information
PowLimitBits and PowLimit

Ref: lightninglabs/neutrino#142

### DIFF
--- a/chainparams.go
+++ b/chainparams.go
@@ -105,6 +105,34 @@ func applyLitecoinParams(params *bitcoinNetParams, litecoinParams *litecoinNetPa
 	params.CoinbaseMaturity = litecoinParams.CoinbaseMaturity
 
 	copy(params.GenesisHash[:], litecoinParams.GenesisHash[:])
+	copy(params.GenesisBlock.Header.MerkleRoot[:],
+		litecoinParams.GenesisBlock.Header.MerkleRoot[:])
+	params.GenesisBlock.Header.Version =
+		litecoinParams.GenesisBlock.Header.Version
+	params.GenesisBlock.Header.Timestamp =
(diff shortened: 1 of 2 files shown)


wakiyamap:
Isn't it better to copy DNSseed too?
```
	dnsSeeds := make([]chaincfg.DNSSeed, len(litecoinParams.DNSSeeds))
	for i := 0; i < len(litecoinParams.DNSSeeds); i++ {
		dnsSeeds[i] = chaincfg.DNSSeed{
			Host:         litecoinParams.DNSSeeds[i].Host,
			HasFiltering: litecoinParams.DNSSeeds[i].HasFiltering,
		}
	}
	params.DNSSeeds = dnsSeeds
```

Wowee0:
@wakiyamap yes, in general it would be best if the entire litecoinParams is copied over, but I don't know a way to do that without copying each data element explicitly.  By the way, LND uses its own seeding as defined in lnd/chainregistry.go (line 561 - 595)

wakiyamap:
@Wowee0 thanks reply!
lnd/chainregistry.go (line 561 - 595) is dnsseed of layer2(lnd).
Dnsseed of layer1 is different.

So I think this would load the default BTCtestnet dnsseed.

Wowee0:
@wakiyamap ok, yes your code above should indeed be included - done.  Good find!

Roasbeef:
Closing due to inactivity.

PR #3063, opened 2019-05-09
[reliable payments] Move payment error deobfuscation to router

halseth:
This PR consists of two important pieces of #2761: 

### Make `router`<->`switch` interaction async
We define `PaymentResult` and the  method `GetPaymentResult`
exposed by the `switch`.

This lets us distinguish an critical error from a actual payment result
(success or failure). This is important since we know that we can only
attempt another payment when a final result from the previous payment
attempt is received.

In order to achieve this we let the `router` handling generating `paymentID`s,
since they are used to later query the `switch` for the final result.

### Move payment error deobfuscation
In this commit we move handing the deobfuscator from the router to the
switch from when the payment is initiated, to when the result is
queried.

We do this because only the router can recreate the deobfuscator after a
restart, and we are preparing for being able to handle results across
restarts.

Since the deobfuscator cannot be nil anymore, we can also get rid of
that special case.

Split off from #2761

### DIFF
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -54,6 +54,8 @@ func newConcurrentTester(t *testing.T) *concurrentTester {
 }
 
 func (c *concurrentTester) Fatalf(format string, args ...interface{}) {
+	c.T.Helper()
+
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
@@ -1108,13 +1110,30 @@ func TestChannelLinkMultiHopUnknownPaymentHash(t *testing.T) {
 	}
 
 	// Send payment and expose err channel.
-	_, err = n.aliceServer.htlcSwitch.SendHTLC(
+	err = n.aliceServer.htlcSwitch.SendHTLC(
 		n.firstBobChannelLink.ShortChanID(), pid, htlc,
 		newMockDeobfuscator(),
 	)
-	if !strings.Contains(err.Error(), lnwire.CodeUnknownPaymentHash.String()) {
-		t.Fatalf("expected %v got %v", err,
-			lnwire.CodeUnknownPaymentHash)
+	if err != nil {
+		t.Fatalf("unable to get send payment: %v", err)
+	}
+
+	resultChan, err := n.aliceServer.htlcSwitch.GetPaymentResult(pid)
+	if err != nil {
+		t.Fatalf("unable to get payment result: %v", err)
+	}
+
+	var result *PaymentResult
+	select {
+
+	case result = <-resultChan:
+	case <-time.After(5 * time.Second):
+		t.Fatalf("no result arrive")
+	}
+
+	fErr := result.Error
+	if !strings.Contains(fErr.Error(), lnwire.CodeUnknownPaymentHash.String()) {
+		t.Fatalf("expected %v got %v", lnwire.CodeUnknownPaymentHash, fErr)
 	}
 
 	// Wait for Alice to receive the revocation.
@@ -3867,7 +3886,7 @@ func TestChannelLinkAcceptDuplicatePayment(t *testing.T) {
 	// With the invoice now added to Carol's registry, we'll send the
 	// payment. It should succeed w/o any issues as it has been crafted
 	// properly.
-	_, err = n.aliceServer.htlcSwitch.SendHTLC(
+	err = n.aliceServer.htlcSwitch.SendHTLC(
 		n.firstBobChannelLink.ShortChanID(), pid, htlc,
 		newMockDeobfuscator(),
 	)
@@ -3875,9 +3894,23 @@ func TestChannelLinkAcceptDuplicatePayment(t *testing.T) {
 		t.Fatalf("unable to send payment to carol: %v", err)
 	}
 
+	resultChan, err := n.aliceServer.htlcSwitch.GetPaymentResult(pid)
+	if err != nil {
+		t.Fatalf("unable to get payment result: %v", err)
+	}
+
+	select {
+	case result := <-resultChan:
+		if result.Error != nil {
+			t.Fatalf("payment failed: %v", result.Error)
+		}
+	case <-time.After(5 * time.Second):
+		t.Fatalf("payment result did not arrive")
+	}
+
 	// Now, if we attempt to send the payment *again* it should be rejected
 	// as it's a duplicate request.
-	_, err = n.aliceServer.htlcSwitch.SendHTLC(
+	err = n.aliceServer.htlcSwitch.SendHTLC(
 		n.firstBobChannelLink.ShortChanID(), pid, htlc,
 		newMockDeobfuscator(),
 	)
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -1100,15 +1100,16 @@ func TestChannelLinkMultiHopUnknownPaymentHash(t *testing.T) {
 	// Generate payment invoice and htlc, but don't add this invoice to the
 	// receiver registry. This should trigger an unknown payment hash
 	// failure.
-	_, htlc, err := generatePayment(amount, htlcAmt, totalTimelock,
-		blob)
+	_, htlc, pid, err := generatePayment(
+		amount, htlcAmt, totalTimelock, blob,
+	)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	// Send payment and expose err channel.
 	_, err = n.aliceServer.htlcSwitch.SendHTLC(
-		n.firstBobChannelLink.ShortChanID(), htlc,
+		n.firstBobChannelLink.ShortChanID(), pid, htlc,
 		newMockDeobfuscator(),
 	)
 	if !strings.Contains(err.Error(), lnwire.CodeUnknownPaymentHash.String()) {
@@ -1909,7 +1910,9 @@ func TestChannelLinkBandwidthConsistency(t *testing.T) {
 	// a switch initiated payment.  The resulting bandwidth should
 	// now be decremented to reflect the new HTLC.
 	htlcAmt := lnwire.NewMSatFromSatoshis(btcutil.SatoshiPerBitcoin)
-	invoice, htlc, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
+	invoice, htlc, _, err := generatePayment(
+		htlcAmt, htlcAmt, 5, mockBlob,
+	)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -1989,7 +1992,7 @@ func TestChannelLinkBandwidthConsistency(t *testing.T) {
 
 	// Next, we'll add another HTLC initiated by the switch (of the same
 	// amount as the prior one).
-	invoice, htlc, err = generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
+	invoice, htlc, _, err = generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -2075,8 +2078,9 @@ func TestChannelLinkBandwidthConsistency(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to gen route: %v", err)
 	}
-	invoice, htlc, err = generatePayment(htlcAmt, htlcAmt,
-		totalTimelock, blob)
+	invoice, htlc, _, err = generatePayment(
+		htlcAmt, htlcAmt, totalTimelock, blob,
+	)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -2183,7 +2187,9 @@ func TestChannelLinkBandwidthConsistency(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to gen route: %v", err)
 	}
-	invoice, htlc, err = generatePayment(htlcAmt, htlcAmt, totalTimelock, blob)
+	invoice, htlc, _, err = generatePayment(
+		htlcAmt, htlcAmt, totalTimelock, blob,
+	)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -2314,7 +2320,9 @@ func TestChannelLinkBandwidthConsistencyOverflow(t *testing.T) {
 
 	var htlcID uint64
 	addLinkHTLC := func(id uint64, amt lnwire.MilliSatoshi) [32]byte {
-		invoice, htlc, err := generatePayment(amt, amt, 5, mockBlob)
+		invoice, htlc, _, err := generatePayment(
+			amt, amt, 5, mockBlob,
+		)
 		if err != nil {
 			t.Fatalf("unable to create payment: %v", err)
 		}
@@ -2580,7 +2588,7 @@ func TestChannelLinkTrimCircuitsPending(t *testing.T) {
 	// message for the test.
 	var mockBlob [lnwire.OnionPacketSize]byte
 	htlcAmt := lnwire.NewMSatFromSatoshis(btcutil.SatoshiPerBitcoin)
-	_, htlc, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
+	_, htlc, _, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -2860,7 +2868,7 @@ func TestChannelLinkTrimCircuitsNoCommit(t *testing.T) {
 	// message for the test.
 	var mockBlob [lnwire.OnionPacketSize]byte
 	htlcAmt := lnwire.NewMSatFromSatoshis(btcutil.SatoshiPerBitcoin)
-	_, htlc, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
+	_, htlc, _, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -3113,7 +3121,7 @@ func TestChannelLinkBandwidthChanReserve(t *testing.T) {
 	// a switch initiated payment.  The resulting bandwidth should
 	// now be decremented to reflect the new HTLC.
 	htlcAmt := lnwire.NewMSatFromSatoshis(3 * btcutil.SatoshiPerBitcoin)
-	invoice, htlc, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
+	invoice, htlc, _, err := generatePayment(htlcAmt, htlcAmt, 5, mockBlob)
 	if err != nil {
 		t.Fatalf("unable to create payment: %v", err)
 	}
@@ -3844,8 +3852,9 @@ func TestChannelLinkAcceptDuplicatePayment(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	invoice, htlc, err := generatePayment(amount, htlcAmt, totalTimelock,
-		blob)
+	invoice, htlc, pid, err := generatePayment(
+		amount, htlcAmt, totalTimelock, blob,
+	)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -3859,7 +3868,7 @@ func TestChannelLinkAcceptDuplicatePayment(t *testing.T) {
 	// payment. It should succeed w/o any issues as it has been crafted
 	// properly.
 	_, err = n.aliceServer.htlcSwitch.SendHTLC(
-		n.firstBobChannelLink.ShortChanID(), htlc,
+		n.firstBobChannelLink.ShortChanID(), pid, htlc,
 		newMockDeobfuscator(),
 	)
 	if err != nil {
@@ -3869,7 +3878,7 @@ func TestChannelLinkAcceptDuplicatePayment(t *testing.T) {
 	// Now, if we attempt to send the payment *again* it should be rejected
 	// as it's a duplicate request.
 	_, err = n.aliceServer.htlcSwitch.SendHTLC(
-		n.firstBobChannelLink.ShortChanID(), htlc,
+		n.firstBobChannelLink.ShortChanID(), pid, htlc,
 		newMockDeobfuscator(),
 	)
 	if err != ErrAlreadyPaid {
@@ -4255,7 +4264,7 @@ func generateHtlcAndInvoice(t *testing.T,
 		t.Fatalf("unable to generate route: %v", err)
 	}
 
-	invoice, htlc, err := generatePayment(
+	invoice, htlc, _, err := generatePayment(
 		htlcAmt, htlcAmt, uint32(htlcExpiry), blob,
 	)
 	if err != nil {
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -1112,21 +1112,26 @@ func TestChannelLinkMultiHopUnknownPaymentHash(t *testing.T) {
 	// Send payment and expose err channel.
 	err = n.aliceServer.htlcSwitch.SendHTLC(
 		n.firstBobChannelLink.ShortChanID(), pid, htlc,
-		newMockDeobfuscator(),
 	)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


routing/router.go
```diff
@@ -180,7 +179,15 @@ type Config struct {
...
 		htlcAdd *lnwire.UpdateAddHTLC,
-		circuit *sphinx.Circuit) ([sha256.Size]byte, error)
+		circuit *sphinx.Circuit) error
+
+	// GetPaymentResult returns the the result of the payment attempt with
+	// the given paymentID. The method returns a channel where the payment
+	// result will be sent when available, or an error is encountered. If
+	// the paymentID is unknown, htlcswitch.ErrPaymentIDNotFound will be
+	// returned.
+	GetPaymentResult func(paymentID uint64) (
```
Roasbeef: Not a blocker, but at this point it may make sense to extract this method and the one above it into a new interface.

channeldb/graph_test.go
```diff
@@ -3005,3 +3009,54 @@ func compareEdgePolicies(a, b *ChannelEdgePolicy) error {
 	}
 	return nil
 }
+
+// TestLightningNodeSigVerifcation checks that we can use the LightningNode's
+// pubkey to verify signatures.
```
Roasbeef: This commit can be removed now.

htlcswitch/link_test.go
```diff
@@ -1110,7 +1110,6 @@ func TestChannelLinkMultiHopUnknownPaymentHash(t *testing.T) {
 	// Send payment and expose err channel.
 	err = n.aliceServer.htlcSwitch.SendHTLC(
 		n.firstBobChannelLink.ShortChanID(), pid, htlc,
-		newMockDeobfuscator(),
```
Roasbeef: Is this movement really needed in order to complete the reliable payments feature? The onion decryption/encryption portion is a _link level_ detail, meaning that the router doesn't need to be aware of it at all. The current placement is by design, in that the router crafts the route, but all error encryption/decryption is handled by the switch. As a result, the router gets back a nice concrete error, and doesn't need to know about the details of error encryption/decryption. 

IMO, it's a code smell that the router is now _creating_ then _returning_ an `htlcswitch.ForwardingError` back to itself. It's also the case that we already store the items required to reconstruct the circuit (for encrypting and decrypting errors) in the switch's circuit map. Additionally, there's some particular context in the switch w.r.t things like if this is a locally sourced error, a resolution, and so on that are required in order to properly handle all classes of errors. Overall, I don't see why this change is warranted.

halseth: As pointed out in https://github.com/lightningnetwork/lnd/pull/2762/#discussion_r281499203, the current placement is awkward, as _both_ the `router` (encryption) and `switch` (decryption) is aware of the onion format now. By moving it all into either one, it should be less shared responsibility.

I moved it all into the `router` since it has everything that is needed to decrypt the error (route + session key) by recreating the `ErrorDecryptor` after a restart. We could possibly move it all into the `switch`, but that would require us to store this information across restarts, as it is currently not (the curcuit map only stores onion blobs from what I can tell). If that's what we want then a natural place to put it would be the circuit map, as part of the `paymentCircuit`?

And finally, the reason this is needed for reliable payments is to handle received errors across restarts. The `switch` will already handle persisting received results (see #2762), but it cannot deobfuscate the errors. This has to be done by the router, either by getting the obfuscated data from the switch, or by giving it the `deobfuscator` to the switch after a restart (but this is a bit tricky timingwise).

channeldb/graph_test.go
halseth: Done

routing/router.go
halseth: Made interface

htlcswitch/link_test.go
halseth: > or by giving it the deobfuscator to the switch after a restart (but this is a bit tricky timingwise).
I made a rough attempt doing this, and the diff is not that bad:  https://github.com/lightningnetwork/lnd/commit/9e2c428f34c6de9486fceb4d09a376671b4e578c

lmk what is preferred.

halseth: Went with the smallest diff case.

halseth:
PTAL @Roasbeef

routing/router.go
```diff
@@ -1694,9 +1704,67 @@ func (r *ChannelRouter) sendToSwitch(route *route.Route, paymentHash [32]byte) (
...
+	err = r.cfg.Payer.SendHTLC(
+		firstHop, paymentID, htlcAdd,
+	)
+	if err != nil {
+		log.Errorf("Failed sending attempt %d for payment %x to "+
+			"switch: %v", paymentID, paymentHash, err)
+
+		// We must inspect the error to know whether it was critical or
+		// not, to decide whether we should continue trying.
+		finalOutcome := r.processSendError(
```
joostjager: This is not the type of error that is suitable for `processSendError`. It will never be a wire error at this stage and `finalOutcome` is always `true`?

htlcswitch/switch.go
```diff
@@ -353,35 +340,80 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+			return
+		}
+
+		// Extract the result and pass it to the result channel.
+		result, err := s.extractResult(
+			deobfuscator, n, payment.paymentHash,
+		)
+		if err != nil {
+			log.Errorf("Unable to extract result: %v", err)
+			return
```
joostjager: This will leave the `router` blocked?

htlcswitch/switch.go
```diff
@@ -889,64 +903,80 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
 	// has been restarted since sending the payment.
 	payment := s.findPayment(pkt.incomingHTLCID)
```
joostjager: If `payment` is nil, we can return already here.

htlcswitch/switch.go
```diff
@@ -990,29 +1019,16 @@ func (s *Switch) parseFailedPayment(payment *pendingPayment, pkt *htlcPacket,
...
-	// channel failure to the router.
-	case payment == nil:
-		userErr := fmt.Sprintf("error decryptor for payment " +
-			"could not be located, likely due to restart")
-		failure = &ForwardingError{
-			ErrorSource:    s.cfg.SelfKey,
-			ExtraMsg:       userErr,
-			FailureMessage: lnwire.NewTemporaryChannelFailure(nil),
-		}
-
```
joostjager: :+1:

htlcswitch/switch.go
```diff
@@ -956,14 +986,13 @@ func (s *Switch) parseFailedPayment(payment *pendingPayment, pkt *htlcPacket,
...
 		var userErr string
 		r := bytes.NewReader(htlc.Reason)
 		failureMsg, err := lnwire.DecodeFailure(r, 0)
 		if err != nil {
-			userErr = fmt.Sprintf("unable to decode onion failure, "+
-				"htlc with hash(%x): %v",
-				pkt.circuit.PaymentHash[:], err)
+			userErr = fmt.Sprintf("unable to decode onion "+
+				"failure: %v", err)
 			log.Error(userErr)
```
joostjager: Is the log line useful without any identifying info? Same for other statement in this function.

routing/mock_test.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, htlcswitch.ErrPaymentIDNotFound
+	}
+	c <- res
+
+	return c, nil
+
+}
+
+func (m *mockPaymentAttemptDispatcher) setPaymentResult(f func(firstHop lnwire.ShortChannelID) ([32]byte, error)) {
+
```
joostjager: newline

routing/router.go
```diff
@@ -1694,9 +1704,67 @@ func (r *ChannelRouter) sendToSwitch(route *route.Route, paymentHash [32]byte) (
 	firstHop := lnwire.NewShortChanIDFromInt(
 		route.Hops[0].ChannelID,
 	)
-	return r.cfg.SendToSwitch(
-		firstHop, htlcAdd, circuit,
+
+	// Using the created circuit, initialize the error
+	// decrypter so we can parse+decode any failures
+	// incurred by this payment within the switch.
+	errorDecryptor := &htlcswitch.SphinxErrorDecrypter{
```
joostjager: Can be moved down to where it is used.

htlcswitch/switch.go
```diff
@@ -889,64 +903,80 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
...
+		unencrypted:  unencrypted,
+		isResolution: pkt.isResolution,
+	}
 
-	switch htlc := pkt.htlc.(type) {
+	// Deliver the payment error and preimage to the application, if it is
+	// waiting for a response.
+	if payment != nil {
+		payment.resultChan <- n
+		s.removePendingPayment(pkt.incomingHTLCID)
```
cfromknecht: seems like there's a possible race where:
1. `SendHTLC` is called, and the pending payment is added to the map
2. The HTLC is settled/failed (most likely, failed locally) and the pending payment is removed from the map
3. `GetPaymentResult` is called but the pending payment is no longer found

Perhaps we can just opt to not remove pending payments from the map? Would think the flow would be more like:
1. If pending payment exists, wait for ntfn to be delivered via channel
2. If no pending payment exists, retrieve from database or cached in-memory response

Roasbeef: Yeah this doesn't yet resolve races like this or gaps in persistence. Those will be taken care of in  #2762 (the pending payments map goes away all together).

htlcswitch/payment_result.go
```diff
@@ -0,0 +1,47 @@
...
+	// msg is the received result. This should be of type UpdateFulfillHTLC
+	// or UpdateFailHTLC.
+	msg lnwire.Message
+
+	// unencrypted indicates whether the failure encoded in the message is
+	// unencrypted, and hence doesn't need to be decrypted.
+	unencrypted bool
+
+	// isResolution indicates whether this is a resolution message, in
+	// which the failure reason might not be included.
```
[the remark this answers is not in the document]
Roasbeef: IIRC we return "permanent chan failure" if we need to sweep it on chain, then cancel back along the incoming link.

htlcswitch/switch.go
Roasbeef: Agree we should still log the payment hash at the very least.

routing/router.go
Roasbeef: Yeah I think this will be the case, as if we get an error here, then it never left the switch. Will need to double check before we modify this at all. But in the end, having this statement here may actually be harmless.

htlcswitch/switch.go
Roasbeef: Yeah we need to return an error result in this case. Also above when we exit due to a quit channel.

htlcswitch/switch.go
```diff
@@ -353,35 +340,80 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
 	return nil
 }
 
+// GetPaymentResult returns the the result of the payment attempt with the
+// given paymentID. The method returns a channel where the payment result will
+// be sent when available, or an error is encountered. If the paymentID is
+// unknown, ErrPaymentIDNotFound will be returned.
+func (s *Switch) GetPaymentResult(paymentID uint64,
```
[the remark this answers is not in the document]
Roasbeef: More than once as in multiple subscribers?

htlcswitch/switch.go
cfromknecht: gotcha, okay looking through #2762 this race does seem to be eliminated

routing/mock_test.go
```diff
@@ -0,0 +1,30 @@
+package routing
```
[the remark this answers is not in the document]
halseth: Fixed.

routing/mock_test.go
```diff
@@ -0,0 +1,30 @@
...
+	_ htlcswitch.ErrorDecrypter) ([sha256.Size]byte, error) {
+
+	if m.onPayment != nil {
+		return m.onPayment(firstHop)
+	}
+
+	return [sha256.Size]byte{}, nil
+}
+
+func (m *mockPaymentAttemptDispatcher) setPaymentResult(f func(firstHop lnwire.ShortChannelID) ([32]byte, error)) {
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3063/commits/6e28fb8d374e818b8d8cb7b305ce5ec54dbf2190

routing/mock_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3063/commits/6e28fb8d374e818b8d8cb7b305ce5ec54dbf2190

routing/router.go
```diff
@@ -1694,8 +1703,16 @@ func (r *ChannelRouter) sendToSwitch(route *route.Route, paymentHash [32]byte) (
 	firstHop := lnwire.NewShortChanIDFromInt(
 		route.Hops[0].ChannelID,
 	)
-	return r.cfg.SendToSwitch(
-		firstHop, htlcAdd, circuit,
+
+	// Using the created circuit, initialize the error
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3063/commits/9df97c56ddd378448def5065002123e91df2909d

routing/router.go
halseth: Not sure what you mean?

htlcswitch/switch.go
halseth: In upcoming PRs we won't always have the payment hash available at this point (we would need to persist it). Added payment ID to the logs: https://github.com/lightningnetwork/lnd/pull/3063/commits/0062686efade9ac80de973caeb0b6b1f9181360a

htlcswitch/switch.go
halseth: Parse failed payment will actually log. Also this will change in upcoming PR.

routing/router.go
halseth: We will return a `ForwardingError` for some local cases, like if no Link has capacity.

htlcswitch/switch.go
halseth: It's a good point, it does introduce a race. This is fixed in #2762, but it is a bummer that this PR introduces the race. Temp fixed it by not removing the pending payment at all: https://github.com/lightningnetwork/lnd/pull/3063/commits/1f9b4fb260979920cb62eda3e9940384becdc1bd

htlcswitch/switch.go
```diff
@@ -889,64 +903,80 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
...
+	switch htlc := n.msg.(type) {
 
 	// We've received a settle update which means we can finalize the user
 	// payment and return successful response.
 	case *lnwire.UpdateFulfillHTLC:
 		// Persistently mark that a payment to this payment hash
 		// succeeded. This will prevent us from ever making another
 		// payment to this hash.
-		err := s.control.Success(pkt.circuit.PaymentHash)
+		err := s.control.Success(paymentHash)
```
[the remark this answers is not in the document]
halseth: Yeah, when the control tower is removed, it will be more aptly named.

htlcswitch/switch.go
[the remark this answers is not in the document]
halseth: It will support multiple callers, but I guess it won't happen.

htlcswitch/payment_result.go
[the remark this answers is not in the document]
halseth: I see at least one place where the failure reason is non-nil: https://github.com/lightningnetwork/lnd/blob/a8fa4094ba61e26bd9855e9743883e3a234d383a/htlcswitch/switch.go#L1184

cc @cfromknecht

htlcswitch/switch.go
```diff
@@ -353,35 +340,80 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+		return nil, ErrPaymentIDNotFound
+	}
+
+	resultChan := make(chan *PaymentResult, 1)
+
+	// Since the payment was known, we can start a goroutine that can
+	// extract the result when it is available, and pass it on to the
+	// caller.
+	s.wg.Add(1)
+	go func() {
```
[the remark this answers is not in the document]
halseth: `GetPaymentID` needs to return for the router to know whether the payment ID existed (also to not deadlock the interaction router<->switch on shutdown).

htlcswitch/switch.go
halseth: Hm, yeah this is interesting. When calling `GetPaymentResult` we need to distinguish between 3 types of errors:

1. An immediate error whether the subscription was successful (currently only error is in case of unknown payment ID).
2. Critical error that happened when the result was received (this case). This is not a `ForwardingError` (though we could wrap it in one?).
3. A forwarding error.

I see two ways of designing this:

* Make `PaymentResult` also hold critical errors:
```go
type PaymentResult struct {
	// Preimage is set by the switch in case a sent HTLC was settled.
	Preimage [32]byte

 	// Error is non-nil in case a HTLC send failed, and the HTLC is now
	// irrevocably cancelled.
	Error error
}
```
The reason this could only be `ForwardingErrors` is that in an earlier iteration this whole struct was serialized.
* Wrap the critical error in a `ForwardingError`.

htlcswitch/switch.go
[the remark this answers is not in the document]
joostjager: Yes, calling with the same pid twice.

htlcswitch/payment_result.go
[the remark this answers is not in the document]
joostjager: In that line, it isn't a locally sourced payment. So that reason can not end up as a payment result in the switch with `isResolution` true?

htlcswitch/switch.go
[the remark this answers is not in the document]
joostjager: But `GetPaymentResult` without the goroutine would still return early with `ErrPaymentIDNotFound` if it doesn't exist?

Does a dead lock happen if both router and switch wait for closure of the same quit chan?

htlcswitch/switch.go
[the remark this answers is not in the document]
joostjager: How would it support multiple callers if there is only one `resultChan`? Only the first callers receives the result and the next will wait forever?

htlcswitch/switch.go
halseth: I realized that we cannot return any result in case of quit, so maybe we can just close the channel in that case.

routing/router.go
joostjager: Hmm I see. That is a whole new chapter in failure message interpretation then. I am wondering if it would be better in the end to gradually move those over to local errors that describe the exact problem, instead of a text string `ExtraMsg`.

routing/router.go
joostjager: Could be instantiated right above `GetPaymentResult` where it is used. To prevent unnecessary execution if it doesnt get there

halseth: gotcha, fixed here: https://github.com/lightningnetwork/lnd/pull/3063/commits/0a2ff1cb8da7a0e89ca24c8e2efa94e9fa42552f

htlcswitch/switch.go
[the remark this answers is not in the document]
halseth: making the method return a result async makes it more natural that this is a long-running process. The deadlock happens if the router is waiting for the method to return on shutdown, while the switch will be shutdown after the router.

htlcswitch/switch.go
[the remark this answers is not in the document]
halseth: Yeah, it is only supported in the upcoming PRs

htlcswitch/switch.go
halseth: Made a thing: https://github.com/lightningnetwork/lnd/pull/3063/commits/1fa25c266a357dae5f6d758a968df44b5d2eacce

htlcswitch/payment_result.go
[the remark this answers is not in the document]
halseth: I think it would be safe to remove the `isResolution` bool, and just leave the `htlc.Reason==nil` check (or safer `len(htlc.Reason)==0` to handle serialization`)? If there are other cases where the reason can be nil, that is not resolution, decryption would already be failing...

htlcswitch/switch.go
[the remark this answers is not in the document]
halseth: It was sort of pre-existing, but it wouldn't deadlock on shutdown. Since `router.SendPayment` wouldn't return when shutting down, but the goroutines calling it aren't waitgrouped.

Passing in a quit channel is possible, but not the nicest API imo, as we are not actually cancelling the HTLC or anything, only triggering the call to return.

Having the decryption being done in the router was what the previous iteration of the PR did, but the diff became much larger (especially because the tests and mocks must be changed to return an encrypted error), and we have to pass the error metadata to the router, so we decided to go this route. See: https://github.com/lightningnetwork/lnd/commit/d9ccc3431f47c120d71049ea04f462e4f6d9b7e2

In short terms, it is mostly done to keep the boundary close to the current one, where the `ForwardingError` is returned directly to the router. I do agree it is not optional, but it can be changed later.

htlcswitch/switch.go
```diff
@@ -353,35 +340,90 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+		// Extract the result and pass it to the result channel.
+		result, err := s.extractResult(
+			deobfuscator, n, paymentID, payment.paymentHash,
+		)
+		if err != nil {
+			e := fmt.Errorf("Unable to extract result: %v", err)
+			log.Error(e)
+			resultChan <- &PaymentResult{
+				Error: e,
+			}
```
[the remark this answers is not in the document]
halseth: We know the channel is buffered here.

htlcswitch/switch.go
halseth: Yes, we would need to pass it in, but not sure if it is valuable, for instance in an post-AMP world? (where the switch would only know about payment shards, not the "payment hash").

htlcswitch/switch.go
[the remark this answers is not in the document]
joostjager: Ok, my suggestion was not to move decryption into the router, but to just to make it a separate call from router to switch. So `switch.SendHTLC` -> `switch.GetPaymentResult` -> `switch.Decrypt` (so slightly different from previous iteration which I remember). But can understand that in this variation the diff in tests may still be large.

htlcswitch/switch.go
joostjager: I don't know. But if I today see a log statement with a pid in it, does it tell me anything more than no extra info at all? Externally I only know the hash.

htlcswitch/switch.go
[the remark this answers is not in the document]
joostjager: Ok

htlcswitch/switch.go
halseth: Added hash back to the log.

htlcswitch/payment_result.go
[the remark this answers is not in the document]
halseth: Decided to keep this as is for now, and decide in the persistence PR.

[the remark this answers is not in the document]
joostjager: :+1:

htlcswitch/switch.go
cfromknecht: @halseth nice, i think that is okay. if for some reason this happens twice then it block everything since the channel is only buffered with 1 item. could also just temporarily buffer them at 100 to be safe, but i think the deduping by the circuit map should prevent this.

htlcswitch/switch.go
```diff
@@ -353,35 +340,90 @@ func (s *Switch) ProcessContractResolution(msg contractcourt.ResolutionMsg) erro
...
+		defer s.wg.Done()
+
+		var n *networkResult
+		select {
+		case n = <-payment.resultChan:
+		case <-s.quit:
+			// We close the result channel to signal a shutdown. We
+			// don't send any result in this case since the HTLC is
+			// still in flight.
+			close(resultChan)
```
cfromknecht: coudl also make this return `ErrSwitchExiting`, since that is indeed the failure

cfromknecht:
needs rebase as well :)

htlcswitch/switch.go
halseth: Wanted to keep the result channel reserved for the cases where the HTLC is no longer in flight. Felt it was best to make that distinction clear.

halseth:
Rebased :) PTAL @joostjager @cfromknecht

PR #3064, opened 2019-05-09
[Reliable payments] add routing.Route (de)serialization + test

halseth:
We will store the routes used during payment attempts, so we need
serialization code.

Split off from #2761

### DIFF
--- a/channeldb/payments.go
+++ b/channeldb/payments.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/coreos/bbolt"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


channeldb/payments.go
```diff
@@ -328,3 +329,84 @@ func deserializeOutgoingPayment(r io.Reader) (*OutgoingPayment, error) {
...
+	); err != nil {
+		return nil, err
+	}
+
+	return h, nil
+}
+
+func serializeRoute(w io.Writer, r route.Route) error {
+	if err := WriteElements(w,
+		r.TotalTimeLock, r.TotalFees, r.TotalAmount, r.SourcePubKey[:],
```
joostjager: `TotalFees` is redundant info and can be derived by `TotalAmount-r.Hops[len(r.Hops)-1].AmtToForward`

halseth: https://github.com/lightningnetwork/lnd/pull/3081

Issue #3058, opened 2019-05-08
NodeAnnouncement not being sent


Crypt-iQ:
If we open a channel between two nodes `A` and `B`, the node that receives the `AnnounceSignatures` message second will likely not send out its `NodeAnnouncement` message to the opposite peer.  This can be seen by running lnd with `-d PEER=debug` and examining that one of the peers does not have the log message starting with `Sending NodeAnnouncement`

Let's say `B` sends out `AnnounceSignatures` first.  This issue then arises because when `B` [sends out their `AnnounceSignatures` message followed by their `NodeAnnouncement` message](https://github.com/lightningnetwork/lnd/blob/master/fundingmanager.go#L2679), their node is not public according to this check here since it hasn't yet received the other half of the channel proof:
https://github.com/lightningnetwork/lnd/blob/c4415f04008c70be20b42f85e1ceff3ef99614c2/discovery/gossiper.go#L1526

Node `A` on the other hand, will most likely (barring the same race condition happening) send out their `NodeAnnouncement` given that it has the full channel proof before sending it out.

A side effect of this is that peers connected to `A` may not see the `NodeAnnouncement` from `B`.  I.e. if we have a graph that looks like
```
Z ---- A ---- B
```
then Z will not receive B's node announcement.

wpaulino:
This would only happen in the case of a node opening it's first public channel, so it's not too crucial of an issue. https://github.com/lightningnetwork/lnd/pull/2084 would also allow us to rebroadcast it if the node decides not to open any other channels.

Crypt-iQ:
Yeah good point on the first public channel.  Although it's possible they open another channel and the announcement isn't broadcast b/c it's a coin flip but it gets less and less likely with more open channels.

Crypt-iQ:
Update: My original comment wasn't entirely accurate. If ANY lnd node receives the remote `AnnounceSignatures` message after already sending it out, no `NodeAnnouncement` message will be sent.  This is because of the previously mentioned check `IsPublicNode` and also because https://github.com/lightningnetwork/lnd/blob/c4415f04008c70be20b42f85e1ceff3ef99614c2/discovery/gossiper.go#L2326-L2330 where source is set to the `AnnounceSignatures` sender.  So when a node attempts to broadcast these messages, it won't send to the channel peer since they were set as the sender of these messages:
[...]

PR #3066, opened 2019-05-09
lnd_test: fixing testGraphTopologyNotifications flake

Crypt-iQ:
This commit fixes a race condition where Alice's gossip filter would be set after the announcement generation between Bob & Carol resulting in Alice receiving no channel updates or node announcements.  Now, Alice and Bob connect before the channel open between Bob and Carol. I also changed the check here:
https://github.com/lightningnetwork/lnd/blob/c4415f04008c70be20b42f85e1ceff3ef99614c2/lnd_test.go#L8609
to `for numChannelUpds < 2 || numNodeAnns < 1` so it actually waits for both channel updates and node announcements. I think it should wait for an additional node announcement from Carol, but that requires a fix of #3058 

This should also be changed in the same manner but would require a fix of #3058 as well:
https://github.com/lightningnetwork/lnd/blob/c4415f04008c70be20b42f85e1ceff3ef99614c2/lnd_test.go#L8460

This flake is identical to #3010

### DIFF
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -8905,7 +8905,7 @@ func testGraphTopologyNotifications(net *lntest.NetworkHarness, t *harnessTest) 
 	// and two node announcements.
 	var numChannelUpds int
 	var numNodeAnns int
-	for numChannelUpds < 2 && numNodeAnns < 2 {
+	for numChannelUpds < 2 || numNodeAnns < 2 {
 		select {
 		// Ensure that a new update for both created edges is properly
 		// dispatched to our registered client.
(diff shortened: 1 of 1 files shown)


wpaulino:
Needs a rebase since the tests have moved over to `lntest/itest`.

Crypt-iQ:
rebased @wpaulino

Crypt-iQ:
I think it would be best if this wasn't merged in until #3219 was.  Then I can update the code to properly check for node announcements.

Crypt-iQ:
Rebased since #3219 was merged in

halseth:
@Crypt-iQ Can you update the PR description to explain why this is still needed now that #3219 is merged?

Crypt-iQ:
@halseth #3219 makes sure each node gets the node announcement. There are two problems present in the current `testGraphTopologyNotifications` itest. 

1. It only waits for 2 updates OR 2 node announcements before breaking out of the for loops described in the description. It needs an `||` instead of `&&` to properly wait for both types.
2. There is also a race condition with Alice's gossip filter being set after announcement generation b/t Bob & Carol.

The reason I waited until #3219 was merged in was so that the for loops in case 1 actually receive 2 node announcements.  This way the test actually represents what it says it does.  Should I update the PR description w/ this?

wpaulino:
> There is also a race condition with Alice's gossip filter being set after announcement generation b/t Bob & Carol which causes the flake.

@Crypt-iQ this should actually be fine after all since Alice and Bob are already connected before each test case starts.

Crypt-iQ:
@wpaulino needed because Alice and Bob disconnect with each other here
https://github.com/lightningnetwork/lnd/blob/2711b7c7f622d879c5991968c4be207c55c5e13a/lntest/itest/lnd_test.go#L9021-L9023

lntest/itest/lnd_test.go
```diff
@@ -9031,30 +9031,41 @@ out:
...
-	)
 
 	// Reconnect Alice and Bob. This should result in the nodes syncing up
 	// their respective graph state, with the new addition being the
 	// existence of Carol in the graph, and also the channel between Bob
 	// and Carol. Note that we will also receive a node announcement from
 	// Bob, since a node will update its node announcement after a new
-	// channel is opened.
+	// channel is opened. We connect Alice and Bob here to ensure her
+	// gossip filter isn't set after the announcements between Bob and Carol
```
halseth: This kinda changes what this test is checking. Previously it checked that a new node's announcement was synced on connection, now it connects first then the announcement is created.

lntest/itest/lnd_test.go
```diff
@@ -9031,30 +9031,41 @@ out:
...
+			Amt: chanAmt,
+		},
+	)
+
 	// We should receive an update advertising the newly connected node,
 	// Bob's new node announcement, and the channel between Bob and Carol.
 	numNodeAnns = 0
 	numChannelUpds = 0
-	for numChannelUpds < 2 && numNodeAnns < 1 {
+	for numChannelUpds < 2 || numNodeAnns < 2 {
```
halseth: 👍 

If you do this (and the other loop condition change) without moving the `openChannel` call, are tests still passing?

Crypt-iQ: It passes most of the time, but still flakes

halseth:
@Crypt-iQ can be closed?

Roasbeef:
Still relevant?

yyforyongyu:
replaced by #6825

PR #3075, opened 2019-05-13
routing/router: nil path pubkeys before spewing

cfromknecht:


willcl-ark:
Hello Conner @cfromknecht, unfortunately building with this patch does not seem to have fixed the excessive debug output on my side... still getting the same ~10MB output file filled with approx 41,000 extra lines of curve 'spew' per RPC. Let me know if you want another debug.log or me to do anything else to help!

routing/router.go
```diff
@@ -1405,7 +1405,11 @@ func generateSphinxPacket(rt *route.Route, paymentHash []byte) ([]byte,
 
 	log.Tracef("Constructed per-hop payloads for payment_hash=%x: %v",
 		paymentHash[:], newLogClosure(func() string {
-			return spew.Sdump(sphinxPath[:sphinxPath.TrueRouteLength()])
+			path := sphinxPath[:sphinxPath.TrueRouteLength()]
+			for _, hop := range path {
+				hop.NodePub.Curve = nil
```
halseth: are we sure this won't crash if we try to use these pubkeys later on? I saw such a crash that was fixed by https://github.com/lightningnetwork/lnd/pull/3065

cfromknecht:
@willcl-ark you're right, had a chance to investigate and should now be resolved. let me know if you disagree :)

routing/router.go
cfromknecht: should be fine, the only operations done on these pubkeys afaict is `ScalarMult` and `ScalarBaseMult` when computing shared secrets. lightning-onion calls these methods directly on `btcec.S256()` so should be okay

cfromknecht: i ran all of the tests with `nolog trace` and nothing crashed

willcl-ark:
@cfromknecht all working as expected my side again now; thanks!

PR #3077, opened 2019-05-14
Probability followup [no review]

joostjager:


joostjager:
Reminder: store route and outcome as separate events, to allow for (temp) black hole penalties

PR #3081, opened 2019-05-15
routing/route+multi: remove redundant TotalFees field

halseth:
Instead get it on demand using method TotalFees().

As pointed out in https://github.com/lightningnetwork/lnd/pull/3064#discussion_r283911175

### DIFF
--- a/channeldb/payments.go
+++ b/channeldb/payments.go
@@ -361,7 +361,7 @@ func deserializeHop(r io.Reader) (*route.Hop, error) {
 
 func serializeRoute(w io.Writer, r route.Route) error {
 	if err := WriteElements(w,
-		r.TotalTimeLock, r.TotalFees, r.TotalAmount, r.SourcePubKey[:],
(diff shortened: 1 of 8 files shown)


channeldb/payments.go
```diff
@@ -361,7 +361,7 @@ func deserializeHop(r io.Reader) (*route.Hop, error) {
 
 func serializeRoute(w io.Writer, r route.Route) error {
 	if err := WriteElements(w,
-		r.TotalTimeLock, r.TotalFees, r.TotalAmount, r.SourcePubKey[:],
+		r.TotalTimeLock, r.TotalAmount, r.SourcePubKey[:],
```
halseth: note: this is a safe change since the code is not yet used anywhere.

wpaulino: Phew, was wondering why we weren't handling that 😅

halseth:
Pushed a fixup commit to address an empty route struct. PTAL

cfromknecht:
needs rebase tho

halseth:
Rebased.

PR #1701, opened 2018-08-08
build+lnrpc: bump gRPC + protobuf related dependencies to latest version

wpaulino:
Went through the current docs on my new setup and realized some deps were missing, so we now explicitly use `go get`.

Fixes #1699.

PR #3082, opened 2019-05-15
lnrpc: upgrade to protobuf v1.3.1

joostjager:
This commit upgrades the protobuf version. Compared to the previous
v1.2.0 it generates smaller diffs in generated code. This change was
introduced in:

https://github.com/golang/protobuf/commit/fffb0f78287ac709b1395be9708df0b579a385f7


joostjager:
Blocked on same issue as https://github.com/lightningnetwork/lnd/pull/1701, doc parsing

wpaulino:
I think this is fine to land. The doc parsing is only affected by the swagger JSON definitions, which are not altered by this change.

PR #3087, opened 2019-05-16
lntest: move itest into package

joostjager:
This PR moves the itests from the root into their own package. We started giving every itest its own file and a separate itest package is good not to clutter the root with those individual test files.

### DIFF
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -1,6 +1,6 @@
 // +build rpctest
 
-package lnd
+package itest
 
 import (
 	"bytes"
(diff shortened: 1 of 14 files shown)


joostjager:
Offline discussion outcome: `lntest/itest`

lntest/itest/lnd_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/lightningnetwork/lnd/lntest"
```
halseth: Unnecessary import move

PR #3089, opened 2019-05-17
sweep+rpc+cmd/lncli: expose pending input sweeps over RPC + CLI

wpaulino:
In this PR, we expose the set of pending inputs that the UtxoSweeper is currently attempting to sweep. This will serve as preparatory work for allowing users to perform fee bumps. Due to the UtxoSweeper's asynchronous nature, we are not able to quickly report on whether a fee bump succeeded or not, so we resort towards exposing its internal fee rate bucketing system over RPC to give users an indication of what's currently being swept and with what fee rate.

I opted to only expose the sweeping fee rate for each bucket, along with all of the inputs within each bucket and their individual fee rates for now. Looking for feedback on whether we could be exposing this information better.

Depends on #3026.


sweep/sweeper.go
```diff
@@ -520,6 +548,11 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
 				log.Errorf("schedule sweep: %v", err)
 			}
 
+		// A new external request has been received to retrieve all of
+		// the inputs we're currently attempting to sweep.
+		case req := <-s.pendingSweepsReqs:
+			req.respChan <- s.handlePendingSweepsReq(req)
```
joostjager: I am wondering how this approach compares to just locking the pending inputs while the summary is generated.

sweep/sweeper.go
```diff
@@ -111,6 +111,28 @@ type inputCluster struct {
...
+// attempting to sweep.
+type pendingSweepsReq struct {
+	respChan chan []PendingSweep
+}
+
+// PendingInput represents an input the UtxoSweeper is currently attempting to
+// sweep.
+type PendingInput struct {
+	OutPoint wire.OutPoint
+	FeeRate  lnwallet.SatPerKWeight
```
joostjager: Should we expose more sweeper state, as this is a pretty technical rpc already? `minPublishHeight` and `publishAttempts`, `witnessType`. Would be useful for investigating problems in the field too. In any case, `value` seems like a useful one to add (`input.SignDesc().Output.Value`).

sweep/sweeper.go
```diff
@@ -111,6 +111,28 @@ type inputCluster struct {
...
+}
+
+// PendingSweep represents a set of inputs that are grouped together according
+// to some heuristic in order to sweep them within a single transaction with a
+// specific sweep fee rate.
+type PendingSweep struct {
+	FeeRate lnwallet.SatPerKWeight
```
joostjager: I am not sure what the sweep fee rate is actually going to tell you. If a tx were to be published now, it would be published with that fee rate. But it could very well be that we already published a tx with this input in it at a different fee rate within the same bucket range. The sweep fee rate fluctuates because it is an average that may change.

sweep/sweeper.go
wpaulino: I think we generally favor this pattern if possible over using a mutex.

sweep/sweeper.go
wpaulino: Good idea, done.

sweep/sweeper.go
wpaulino: Removed. I think it would still be useful to show which inputs are grouped together though, no? They're currently sorted in ascending fee rate order.

wpaulino:
PTAL @joostjager.

sweep/sweeper.go
joostjager: Maybe a comment should be added saying that those fields that are dependent on the internal sweeper algorithm (like `publishAttempts`) can change with every release. So that we don't need to go through the deprecate cycle if we want to change anything internally.

sweep/sweeper.go
joostjager: I think we could do that, but not sure if you should really group the inputs. There are also other ways of grouping, like broadcast attempts, next height, etc. Why group in this way? It isn't even guaranteed that those inputs actually end up in the same tx.

You could also add two fields `bucketMin` and `bucketMax` to show the bucket it is in. Or a separate list in the response that contains the static bucket boundaries.

I would keep it simple and flat initially and see how it is used.

lnrpc/invoicesrpc/invoices.pb.go
```diff
@@ -37,7 +37,7 @@ func (m *CancelInvoiceMsg) Reset()         { *m = CancelInvoiceMsg{} }
 func (m *CancelInvoiceMsg) String() string { return proto.CompactTextString(m) }
 func (*CancelInvoiceMsg) ProtoMessage()    {}
 func (*CancelInvoiceMsg) Descriptor() ([]byte, []int) {
-	return fileDescriptor_invoices_b95f8bc86c534ce9, []int{0}
```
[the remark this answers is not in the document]
wpaulino: A wrong version was used in a commit already in master.

rpcserver.go
```diff
@@ -840,6 +844,81 @@ func (r *rpcServer) ListUnspent(ctx context.Context,
...
+		opHashBytes, _ := hex.DecodeString(opHashStr)
+		amountSat := uint32(pendingInput.SignDesc().Output.Value)
+		satPerByte := uint32(pendingInput.FeeRate.FeePerKVByte() / 1000)
+		broadcastAttempts := uint32(pendingInput.PublishAttempts)
+		nextBroadcastHeight := uint32(pendingInput.MinPublishHeight)
+
+		rpcPendingSweeps = append(rpcPendingSweeps, &lnrpc.PendingSweep{
+			Outpoint: &lnrpc.OutPoint{
+				TxidStr:     opHashStr,
+				TxidBytes:   opHashBytes,
```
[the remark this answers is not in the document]
wpaulino: We still need to populate both from the server's perspective as we don't know ahead of time which one the client is interested in.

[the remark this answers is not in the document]
joostjager: I mean just don't offer the string based one. This is a machine to machine interface. But can also understand that this may not be this pr's problem to address.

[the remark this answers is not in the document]
wpaulino: Ah yes, I'm not a fan of the duplicate fields either. If we're worried about showing raw bytes as base64 in `lncli`, we can just map that to hex in the response.

lnrpc/invoicesrpc/invoices.pb.go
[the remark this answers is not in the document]
Roasbeef: IIRC every time you compile it gets regenerated now as it seems to include a timestamp somewhere?

rpcserver.go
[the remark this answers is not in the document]
Roasbeef: Atm we only use the str variants if we're accepting them within the path URL for the REST interface.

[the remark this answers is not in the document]
Roasbeef: So since this is just an output, I think we only need to set one of them.

sweep/sweeper.go

lnrpc/invoicesrpc/invoices.pb.go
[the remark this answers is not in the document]
wpaulino: Possible that it includes a timestamp, but only if any changes were actually made to the relevant protos. In this case the diff was attempting to fix an incorrect version used, but this is no longer needed since it's been addressed in master.

lnrpc/rpc.proto
```diff
@@ -1002,6 +1015,121 @@ message ListUnspentResponse {
...
+    OutPoint outpoint = 1 [json_name = "outpoint"];
+
+    // The witness type of the output we're attempting to sweep.
+    WitnessType witness_type = 2 [json_name = "witness_type"];
+
+    // The value of the output we're attempting to sweep.
+    uint32 amount_sat = 3 [json_name = "amount_sat"];
+
+    // The fee rate we'll use to sweep the output.
+    uint32 sat_per_byte = 4 [json_name = "sat_per_byte"];
```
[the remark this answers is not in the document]
wpaulino: This is the current fee rate of the input that was either a mapped conf target or a manual fee rate. `sat_per_byte` already describes this is a fee rate so including it would seem redundant?

rpcserver.go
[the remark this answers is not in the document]
wpaulino: Removed the string one.

lnrpc/rpc.proto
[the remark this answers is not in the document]
joostjager: It isn't important, but you do have `amount_sat`...

sweep/sweeper.go
```diff
@@ -996,6 +1019,43 @@ func (s *UtxoSweeper) waitForSpend(outpoint wire.OutPoint,
...
+	pendingInputs := make([]*PendingInput, 0, len(s.pendingInputs))
+	for _, pendingInput := range s.pendingInputs {
+		pendingInputs = append(pendingInputs, pendingInput)
+	}
+
+	// We'll sort the inputs in ascending fee rate order as a convenience to
+	// the caller.
+	sort.Slice(pendingInputs, func(i, j int) bool {
+		return pendingInputs[i].FeeRate < pendingInputs[j].FeeRate
+	})
```
joostjager: This may be more a thing to do in `rpcserver.go`. It is mainly presentation. (non-blocking)

wpaulino: Good point, it's only done within `lncli` now.

lnrpc/rpc.proto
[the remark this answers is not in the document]
wpaulino: My understanding is that there's a general pattern for amount fields throughout the RPC where we label them with `sat`/`msat` accordingly.

lnrpc/rpc.proto
```diff
@@ -1002,6 +1018,121 @@ message ListUnspentResponse {
...
+
+    // The number of broadcast attempts we've made to sweep the output.
+    uint32 broadcast_attempts = 5 [json_name = "broadcast_attempts"];
+
+    /*
+    The next height of the chain at which we'll attempt to broadcast the
+    sweep transaction of the output.
+    */
+    uint32 next_broadcast_height = 6 [json_name = "next_broadcast_height"];
+}
```
Roasbeef: Once we circle back and add retargeting, it may be useful to expose what the _next_ fee rate would be.

rpcserver.go
```diff
@@ -790,19 +811,11 @@ func (r *rpcServer) ListUnspent(ctx context.Context,
...
-			TxidStr:     utxo.OutPoint.Hash.String(),
-			OutputIndex: utxo.OutPoint.Index,
-		}
-
 		utxoResp := lnrpc.Utxo{
 			Type:          addrType,
 			AmountSat:     int64(utxo.Value),
 			PkScript:      hex.EncodeToString(utxo.PkScript),
-			Outpoint:      outpoint,
+			Outpoint:      marshallOutPoint(&utxo.OutPoint),
```
Roasbeef: Alternatively, you can just pass the outpoint as is, and then within `lncli` use `chainhash.NewFromStr` (or w/e it's called). This'll result in the bytes only being reversed when they need to be displayed to the user.

cmd/lncli/types.go
```diff
@@ -11,7 +11,7 @@ type OutPoint string
 
 // NewOutPointFromProto formats the lnrpc.OutPoint into an OutPoint for display.
 func NewOutPointFromProto(op *lnrpc.OutPoint) OutPoint {
-	return OutPoint(fmt.Sprintf("%s:%d", op.TxidStr, op.OutputIndex))
+	return OutPoint(fmt.Sprintf("%x:%d", op.TxidBytes, op.OutputIndex))
```
Roasbeef: Will this change break any of the other cli commands?

cmd/lncli/types.go
```diff
@@ -38,3 +38,27 @@ func NewUtxoFromProto(utxo *lnrpc.Utxo) *Utxo {
 		Confirmations: utxo.Confirmations,
 	}
 }
+
+// PendingSweep displays information about a pending sweep within lnd's
+// UtxoSweeper.
+type PendingSweep struct {
+	OutPoint            OutPoint          `json:"outpoint"`
+	WitnessType         lnrpc.WitnessType `json:"witness_type"`
```
Roasbeef: If you change this to be a `string` and use the `String()` method on the enum, then the output will be `COMMITMENT_TIME_LOCK` rather than zero.

sweep/sweeper.go
```diff
@@ -979,6 +1001,37 @@ func (s *UtxoSweeper) waitForSpend(outpoint wire.OutPoint,
...
+		return nil, ErrSweeperShuttingDown
+	}
+}
+
+// handlePendingSweepsReq handles a request to retrieve all pending inputs the
+// UtxoSweeper is attempting to sweep.
+func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) []*PendingInput {
+	pendingInputs := make([]*PendingInput, 0, len(s.pendingInputs))
+	for _, pendingInput := range s.pendingInputs {
+		pendingInputs = append(pendingInputs, pendingInput)
```
joostjager: Should a deep copy be made here? To really get a snapshot? And also to prevent possible concurrency issues when the caller starts doing things with the inputs? (maybe just unsafe read if that is possible)

cmd/lncli/types.go
wpaulino: Don't think so, I've made sure that the `TxidBytes` is always set for `lnrpc.OutPoint`s.

cmd/lncli/types.go
wpaulino: True, though they currently already show the string value. Done anyway.

rpcserver.go
wpaulino: We can't use the string if the response doesn't include it, which we decided in a previous comment? For what it's worth, I've also seen some confusion with regular users of the RPC expecting the bytes to not be in the reversed order.

sweep/sweeper.go
```diff
@@ -79,6 +79,10 @@ type PendingInput struct {
 	// map it into a fee rate whenever we attempt to cluster inputs for a
 	// sweep.
 	FeePreference FeePreference
+
+	// LastFeeRate is the most recent fee rate used for this input within a
+	// transaction broadcast to the network.
+	LastFeeRate lnwallet.SatPerKWeight
```
wpaulino: Had to add this field in order to display fee rates on the RPC for inputs with confirmation targets. Is it worth setting this value to something once the input comes in to the sweeper to not have it be 0? The definitive fee rate won't be known until we actually try to sweep the input.

joostjager: I think 0 is okay to signal that it hasn't been in a tx yet.

sweep/sweeper.go
```diff
@@ -880,6 +909,45 @@ func (s *UtxoSweeper) waitForSpend(outpoint wire.OutPoint,
...
+}
+
+// handlePendingSweepsReq handles a request to retrieve all pending inputs the
+// UtxoSweeper is attempting to sweep.
+func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) []*PendingInput {
+	pendingInputs := make([]*PendingInput, 0, len(s.pendingInputs))
+	for _, pendingInput := range s.pendingInputs {
+		// Only the exported fields are set, as we expect the response
+		// to only be consumed externally.
+		pendingInputs = append(pendingInputs, &PendingInput{
```
joostjager: I had the same thing with creating a snapshot in mission control and introduced separate exported structs for the snapshot. I think it is cleaner. (non blocking)

sweep/sweeper.go
wpaulino: Sounds good, added a note to explain the behavior.

sweep/sweeper.go
wpaulino: Yeah that was how the initial design of the PR approached it, but I opted for reusing the same struct to result in less code, though the difference is not much.

rpcserver.go
Roasbeef: The response has the bytes, why do you need the string? This is about _where_ do we reverse the bytes. What txids aren't displayed in reversed order today? Perhaps they get confused if something is in base58, then they need to decode _and_ reverse it manually.

Roasbeef: Ah I see what you mean about the string, even then it can just be reversed on the cli, we don't reverse anywhere else atm.

lnrpc/rpc.swagger.json
```diff
@@ -3420,6 +3482,25 @@
...
+        "HTLC_ACCEPTED_REVOKE",
+        "HTLC_OFFERED_TIMEOUT_SECOND_LEVEL",
+        "HTLC_ACCEPTED_SUCCESS_SECOND_LEVEL",
+        "HTLC_OFFERED_REMOTE_TIMEOUT",
+        "HTLC_ACCEPTED_REMOTE_SUCCESS",
+        "HTLC_SECOND_LEVEL_REVOKE",
+        "WITNESS_KEY_HASH",
+        "NESTED_WITNESS_KEY_HASH"
+      ],
+      "default": "COMMITMENT_TIME_LOCK",
```
cfromknecht: perhaps we should also define an UNKNOWN_WITNESS type so that it can be the default.

rpcserver.go
```diff
@@ -839,6 +843,80 @@ func (r *rpcServer) ListUnspent(ctx context.Context,
...
+				pendingInput.WitnessType(),
+				pendingInput.OutPoint())
+		}
+
+		op := &lnrpc.OutPoint{
+			TxidBytes:   pendingInput.OutPoint().Hash[:],
+			OutputIndex: pendingInput.OutPoint().Index,
+		}
+		amountSat := uint32(pendingInput.SignDesc().Output.Value)
+		satPerByte := uint32(pendingInput.LastFeeRate.FeePerKVByte() / 1000)
```
cfromknecht: i'd be in favor exposing sat/kb instead to not lose the precision

lnrpc/rpc.swagger.json
wpaulino: Done.

rpcserver.go
wpaulino: Seems better to keep the consistency with the rest of our APIs? Nothing else uses sat/kb to my knowledge.

cfromknecht: hmm yeah i guess most things are sat/b, we do have a couple things in sat/kw though

cfromknecht:
@wpaulino needs rebase

cmd/lncli/types.go
```diff
@@ -11,7 +12,9 @@ type OutPoint string
 
 // NewOutPointFromProto formats the lnrpc.OutPoint into an OutPoint for display.
 func NewOutPointFromProto(op *lnrpc.OutPoint) OutPoint {
-	return OutPoint(fmt.Sprintf("%s:%d", op.TxidStr, op.OutputIndex))
```
cfromknecht: are we certain this doesn't break existing rpcs?

wpaulino: I went through all of the RPCs that return an `OutPoint` and they all set the bytes.

cfromknecht: just checked, looks like listunspent sets both so we're good

rpcserver.go
```diff
@@ -178,6 +178,10 @@ var (
 			Entity: "onchain",
 			Action: "read",
 		}},
+		"/lnrpc.Lightning/PendingSweeps": {{
+			Entity: "onchain",
+			Action: "read",
+		}},
```
joostjager: If we decide in the followup to offer bumpfee in a sub server, maybe it should be done with `PendingSweeps` too.

sweep/sweeper.go
```diff
@@ -880,6 +909,45 @@ func (s *UtxoSweeper) waitForSpend(outpoint wire.OutPoint,
...
+}
+
+// handlePendingSweepsReq handles a request to retrieve all pending inputs the
+// UtxoSweeper is attempting to sweep.
+func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) pendingInputs {
+	pendingInputs := make(pendingInputs, len(s.pendingInputs))
+	for _, pendingInput := range s.pendingInputs {
+		// Only the exported fields are set, as we expect the response
+		// to only be consumed externally.
+		pendingInputs[*pendingInput.OutPoint()] = &PendingInput{
```
joostjager: I still feel that usage of `PendingInput` and the embedded `input.Input` to expose the sweeper state is not good from an interface segregation perspective. A new struct may just be 10 lines of code and will take away any doubts about what is and isn't populated / meant to be read.

rpcserver.go
joostjager: Ok, you made that comment yourself in #3140 already

wpaulino: Moved to `WalletKit`.

sweep/sweeper.go

wpaulino:
PTAL @Roasbeef @joostjager.

sweep/sweeper.go
joostjager: I think it looks better

wpaulino:
> Only minor comment is that the field name sat_per_byte remains a little vague without reading the comment.

Agreed. This is why `bitcoin-cli` also includes a description of the command's response. Possibly something we can look into later down the road.

Rebased to address a `cmd/lncli` conflict.

cmd/lncli/types.go
```diff
@@ -3,38 +3,38 @@ package main
...
+	var hash chainhash.Hash
+	copy(hash[:], op.TxidBytes)
+	return fmt.Sprintf("%v:%d", hash, op.OutputIndex)
 }
 
-// Utxo displays information about an unspent output, including its address,
+// utxo displays information about an unspent output, including its address,
 // amount, pkscript, and confirmations.
-type Utxo struct {
+type utxo struct {
```
cfromknecht: nit: don't see the benefit in making these private, currently it provides a library to other applications that may want to format data in the same way as lncli. also not much use in having a private struct with public fields

wpaulino: Reverted the private change.

PR #3095, opened 2019-05-17
multi: always fail pending htlcs on commitment confirmation

### LINKED ISSUE #2045: HTLC not getting failed upstream after they timeout downstream

Roasbeef:
In this commit, we address a TOOD, and end up fixing a lingering bug related to commitments on chain. Before this commit, it was possible that `lnd` would unnecessarily cause an incoming link to force close on chain if the outgoing link was in an indeterminate state. We fix this issue by always checking for HTLCs that we need to time out on the incoming link based on all valid commits. With this fix, we'll no longer go to chain unnecessarily which wastes chain fees and may also increase off-chain HTLC fail delays. 

Fixes #2045.

### DIFF
--- a/contractcourt/briefcase.go
+++ b/contractcourt/briefcase.go
@@ -9,6 +9,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
 )
@@ -80,6 +81,24 @@ type ArbitratorLog interface {
 	// FetchContractResolutions fetches the set of previously stored
 	// contract resolutions from persistent storage.
 	FetchContractResolutions() (*ContractResolutions, error)
+
+	// InsertConfirmedCommitSet stores the known set of active HTLCs at the
+	// time channel closure. We'll use this to reconstruct our set of chain
+	// actions anew based on the confirmed and pending commitment state.
+	InsertConfirmedCommitSet(c *CommitSet) error
+
+	// FetchConfirmedCommitSet fetches the known confirmed active HTLC set
+	// from the database.
+	FetchConfirmedCommitSet() (*CommitSet, error)
+
+	// FetchChainActions attempts to fetch the set of previously stored
+	// chain actions. We'll use this upon restart to properly advance our
+	// state machine forward.
+	//
+	// NOTE: This method only exists in order to be able to serve nodes had
+	// channels in the process of closing before the CommitSet struct was
+	// introduced.
+	FetchChainActions() (ChainActionMap, error)
 
 	// WipeHistory is to be called ONLY once *all* contracts have been
 	// fully resolved, and the channel closure if finalized. This method
@@ -247,6 +266,11 @@ var (
 	// actionsBucketKey is the key under the logScope that we'll use to
 	// store all chain actions once they're determined.
 	actionsBucketKey = []byte("chain-actions")
+
+	// commitSetKey is the primary key under the logScope that we'll use to
+	// store the confirmed active HTLC sets once we learn that a channel
+	// has closed out on chain.
+	commitSetKey = []byte("commit-set")
 )
 
 var (
@@ -265,6 +289,11 @@ var (
 	// errNoActions is retuned when the log doesn't contain any stored
 	// chain actions.
 	errNoActions = fmt.Errorf("no chain actions exist")
+
+	// errNoCommitSet is return when the log doesn't contained a CommitSet.
+	// This can happen if the channel hasn't closed yet, or a client is
+	// running an older version that didn't yet write this state.
+	errNoCommitSet = fmt.Errorf("no commit set exists")
 )
 
 // boltArbitratorLog is an implementation of the ArbitratorLog interface backed
@@ -708,6 +737,103 @@ func (b *boltArbitratorLog) FetchContractResolutions() (*ContractResolutions, er
 	return c, err
 }
 
+// FetchChainActions attempts to fetch the set of previously stored chain
+// actions. We'll use this upon restart to properly advance our state machine
+// forward.
+//
+// NOTE: Part of the ContractResolver interface.
+func (b *boltArbitratorLog) FetchChainActions() (ChainActionMap, error) {
+	actionsMap := make(ChainActionMap)
+
+	err := b.db.View(func(tx *bbolt.Tx) error {
+		scopeBucket := tx.Bucket(b.scopeKey[:])
+		if scopeBucket == nil {
+			return errScopeBucketNoExist
+		}
+
+		actionsBucket := scopeBucket.Bucket(actionsBucketKey)
+		if actionsBucket == nil {
+			return errNoActions
+		}
+
+		return actionsBucket.ForEach(func(action, htlcBytes []byte) error {
+			if htlcBytes == nil {
+				return nil
+			}
+
+			chainAction := ChainAction(action[0])
+
+			htlcReader := bytes.NewReader(htlcBytes)
+			htlcs, err := channeldb.DeserializeHtlcs(htlcReader)
+			if err != nil {
+				return err
+			}
+
+			actionsMap[chainAction] = htlcs
+
+			return nil
+		})
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return actionsMap, nil
+}
+
+// InsertConfirmedCommitSet stores the known set of active HTLCs at the time
+// channel closure. We'll use this to reconstruct our set of chain actions anew
+// based on the confirmed and pending commitment state.
+//
+// NOTE: Part of the ContractResolver interface.
+func (b *boltArbitratorLog) InsertConfirmedCommitSet(c *CommitSet) error {
+	return b.db.Update(func(tx *bbolt.Tx) error {
+		scopeBucket, err := tx.CreateBucketIfNotExists(b.scopeKey[:])
+		if err != nil {
+			return err
+		}
+
+		var b bytes.Buffer
+		if err := encodeCommitSet(&b, c); err != nil {
+			return err
+		}
+
+		return scopeBucket.Put(commitSetKey, b.Bytes())
+	})
+}
+
+// FetchConfirmedCommitSet fetches the known confirmed active HTLC set from the
+// database.
+//
+// NOTE: Part of the ContractResolver interface.
+func (b *boltArbitratorLog) FetchConfirmedCommitSet() (*CommitSet, error) {
+	var c *CommitSet
+	err := b.db.View(func(tx *bbolt.Tx) error {
+		scopeBucket := tx.Bucket(b.scopeKey[:])
+		if scopeBucket == nil {
+			return errScopeBucketNoExist
+		}
+
+		commitSetBytes := scopeBucket.Get(commitSetKey)
+		if commitSetBytes == nil {
+			return errNoCommitSet
+		}
+
+		commitSet, err := decodeCommitSet(bytes.NewReader(commitSetBytes))
+		if err != nil {
+			return err
+		}
+
+		c = commitSet
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
 // WipeHistory is to be called ONLY once *all* contracts have been fully
 // resolved, and the channel closure if finalized. This method will delete all
 // on-disk state within the persistent log.
@@ -956,4 +1082,77 @@ func decodeCommitResolution(r io.Reader,
 	}
 
 	return binary.Read(r, endian, &c.MaturityDelay)
+}
+
+func encodeHtlcSetKey(w io.Writer, h *HtlcSetKey) error {
+	err := binary.Write(w, endian, h.IsRemote)
+	if err != nil {
+		return err
+	}
+	return binary.Write(w, endian, h.IsPending)
+}
+
+func encodeCommitSet(w io.Writer, c *CommitSet) error {
+	if err := encodeHtlcSetKey(w, c.ConfCommitKey); err != nil {
+		return err
+	}
+
+	numSets := uint8(len(c.HtlcSets))
+	if err := binary.Write(w, endian, numSets); err != nil {
+		return err
+	}
+
+	for htlcSetKey, htlcs := range c.HtlcSets {
+		htlcSetKey := htlcSetKey
+		if err := encodeHtlcSetKey(w, &htlcSetKey); err != nil {
+			return err
+		}
+
+		if err := channeldb.SerializeHtlcs(w, htlcs...); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func decodeHtlcSetKey(r io.Reader, h *HtlcSetKey) error {
+	err := binary.Read(r, endian, &h.IsRemote)
+	if err != nil {
+		return err
+	}
+
+	return binary.Read(r, endian, &h.IsPending)
+}
+
+func decodeCommitSet(r io.Reader) (*CommitSet, error) {
+	c := &CommitSet{
+		ConfCommitKey: &HtlcSetKey{},
+		HtlcSets:      make(map[HtlcSetKey][]channeldb.HTLC),
+	}
+
+	if err := decodeHtlcSetKey(r, c.ConfCommitKey); err != nil {
+		return nil, err
+	}
+
+	var numSets uint8
+	if err := binary.Read(r, endian, &numSets); err != nil {
+		return nil, err
+	}
+
+	for i := uint8(0); i < numSets; i++ {
+		var htlcSetKey HtlcSetKey
+		if err := decodeHtlcSetKey(r, &htlcSetKey); err != nil {
+			return nil, err
+		}
+
+		htlcs, err := channeldb.DeserializeHtlcs(r)
+		if err != nil {
+			return nil, err
+		}
+
+		c.HtlcSets[htlcSetKey] = htlcs
+	}
+
+	return c, nil
 }
--- a/contractcourt/briefcase.go
+++ b/contractcourt/briefcase.go
@@ -9,7 +9,6 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/coreos/bbolt"
-	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
 )
@@ -81,17 +80,6 @@ type ArbitratorLog interface {
 	// FetchContractResolutions fetches the set of previously stored
 	// contract resolutions from persistent storage.
 	FetchContractResolutions() (*ContractResolutions, error)
-
-	// LogChainActions stores a set of chain actions which are derived from
-	// our set of active contracts, and the on-chain state. We'll write
-	// this et of cations when: we decide to go on-chain to resolve a
-	// contract, or we detect that the remote party has gone on-chain.
-	LogChainActions(ChainActionMap) error
-
-	// FetchChainActions attempts to fetch the set of previously stored
-	// chain actions. We'll use this upon restart to properly advance our
-	// state machine forward.
-	FetchChainActions() (ChainActionMap, error)
 
 	// WipeHistory is to be called ONLY once *all* contracts have been
 	// fully resolved, and the channel closure if finalized. This method
@@ -720,88 +708,6 @@ func (b *boltArbitratorLog) FetchContractResolutions() (*ContractResolutions, er
 	return c, err
 }
 
-// LogChainActions stores a set of chain actions which are derived from our set
-// of active contracts, and the on-chain state. We'll write this et of cations
-// when: we decide to go on-chain to resolve a contract, or we detect that the
-// remote party has gone on-chain.
-//
-// NOTE: Part of the ContractResolver interface.
-func (b *boltArbitratorLog) LogChainActions(actions ChainActionMap) error {
-	return b.db.Batch(func(tx *bbolt.Tx) error {
-		scopeBucket, err := tx.CreateBucketIfNotExists(b.scopeKey[:])
-		if err != nil {
-			return err
-		}
-
-		actionsBucket, err := scopeBucket.CreateBucketIfNotExists(
-			actionsBucketKey,
-		)
-		if err != nil {
-			return err
-		}
-
-		for chainAction, htlcs := range actions {
-			var htlcBuf bytes.Buffer
-			err := channeldb.SerializeHtlcs(&htlcBuf, htlcs...)
-			if err != nil {
-				return err
-			}
-
-			actionKey := []byte{byte(chainAction)}
-			err = actionsBucket.Put(actionKey, htlcBuf.Bytes())
-			if err != nil {
-				return err
-			}
-		}
-
-		return nil
-	})
-}
-
-// FetchChainActions attempts to fetch the set of previously stored chain
-// actions. We'll use this upon restart to properly advance our state machine
-// forward.
-//
-// NOTE: Part of the ContractResolver interface.
-func (b *boltArbitratorLog) FetchChainActions() (ChainActionMap, error) {
-	actionsMap := make(ChainActionMap)
-
-	err := b.db.View(func(tx *bbolt.Tx) error {
-		scopeBucket := tx.Bucket(b.scopeKey[:])
-		if scopeBucket == nil {
-			return errScopeBucketNoExist
-		}
-
-		actionsBucket := scopeBucket.Bucket(actionsBucketKey)
-		if actionsBucket == nil {
-			return errNoActions
-		}
-
-		return actionsBucket.ForEach(func(action, htlcBytes []byte) error {
-			if htlcBytes == nil {
-				return nil
-			}
-
-			chainAction := ChainAction(action[0])
-
-			htlcReader := bytes.NewReader(htlcBytes)
-			htlcs, err := channeldb.DeserializeHtlcs(htlcReader)
-			if err != nil {
-				return err
-			}
-
-			actionsMap[chainAction] = htlcs
-
-			return nil
-		})
-	})
-	if err != nil {
-		return nil, err
-	}
-
-	return actionsMap, nil
-}
-
 // WipeHistory is to be called ONLY once *all* contracts have been fully
 // resolved, and the channel closure if finalized. This method will delete all
 // on-disk state within the persistent log.
@@ -835,7 +741,6 @@ func (b *boltArbitratorLog) WipeHistory() error {
 			return err
 		}
 		if err := scopeBucket.DeleteBucket(contractsBucketKey); err != nil {
-			fmt.Println("nah")
 			return err
 		}
 
--- a/contractcourt/chain_watcher.go
+++ b/contractcourt/chain_watcher.go
@@ -54,7 +54,7 @@ type RemoteUnilateralCloseInfo struct {
 	*lnwallet.UnilateralCloseSummary
 
 	// CommitSet is the set of known valid commitments at the time the
-	// remote party's commitemnt hit the chain.
+	// remote party's commitment hit the chain.
 	CommitSet CommitSet
 }
 
--- a/contractcourt/chain_watcher.go
+++ b/contractcourt/chain_watcher.go
@@ -30,18 +30,75 @@ const (
 	maxCommitPointPollTimeout = 10 * time.Minute
 )
 
-// LocalUnilateralCloseInfo encapsulates all the informnation we need to act
-// on a local force close that gets confirmed.
+// LocalUnilateralCloseInfo encapsulates all the information we need to act on
+// a local force close that gets confirmed.
 type LocalUnilateralCloseInfo struct {
 	*chainntnfs.SpendDetail
 	*lnwallet.LocalForceCloseSummary
 	*channeldb.ChannelCloseSummary
+
+	// CommitSet is the set of known valid commitments at the time the
+	// remote party's commitment hit the chain.
+	CommitSet CommitSet
 }
 
-// CooperativeCloseInfo encapsulates all the informnation we need to act
-// on a cooperative close that gets confirmed.
+// CooperativeCloseInfo encapsulates all the information we need to act on a
+// cooperative close that gets confirmed.
 type CooperativeCloseInfo struct {
 	*channeldb.ChannelCloseSummary
+}
+
+// RemoteUnilateralCloseInfo wraps the normal UnilateralCloseSummary to couple
+// the CommitSet at the time of channel closure.
+type RemoteUnilateralCloseInfo struct {
+	*lnwallet.UnilateralCloseSummary
+
+	// CommitSet is the set of known valid commitments at the time the
+	// remote party's commitemnt hit the chain.
+	CommitSet CommitSet
+}
+
+// CommitSet is a collection of the set of known valid commitments at a given
+// instant. If ConfCommitKey is set, then the commitment identified by the
+// HtlcSetKey has hit the chain. This struct will be used to examine all live
+// HTLCs to determine if any additional actions need to be made based on the
+// remote party's commitments.
+type CommitSet struct {
+	// ConfCommitKey if non-nil, identifies the commitment that was
+	// confirmed in the chain.
+	ConfCommitKey *HtlcSetKey
+
+	// HtlcSets stores the set of all known active HTLC for each active
+	// commitment at the time of channel closure.
+	HtlcSets map[HtlcSetKey][]channeldb.HTLC
+}
+
+// IsEmpty returns true if there are no HTLCs at all within all commitments
+// that are a part of this commitment diff.
+func (c *CommitSet) IsEmpty() bool {
+	if c == nil {
+		return true
+	}
+
+	for _, htlcs := range c.HtlcSets {
+		if len(htlcs) != 0 {
+			return false
+		}
+	}
+
+	return true
+}
+
+// toActiveHTLCSets returns the set of all active HTLCs across all commitment
+// transactions.
+func (c *CommitSet) toActiveHTLCSets() map[HtlcSetKey]htlcSet {
+	htlcSets := make(map[HtlcSetKey]htlcSet)
+
+	for htlcSetKey, htlcs := range c.HtlcSets {
+		htlcSets[htlcSetKey] = newHtlcSet(htlcs)
+	}
+
+	return htlcSets
 }
 
 // ChainEventSubscription is a struct that houses a subscription to be notified
@@ -55,7 +112,7 @@ type ChainEventSubscription struct {
 
 	// RemoteUnilateralClosure is a channel that will be sent upon in the
 	// event that the remote party's commitment transaction is confirmed.
-	RemoteUnilateralClosure chan *lnwallet.UnilateralCloseSummary
+	RemoteUnilateralClosure chan *RemoteUnilateralCloseInfo
 
 	// LocalUnilateralClosure is a channel that will be sent upon in the
 	// event that our commitment transaction is confirmed.
@@ -249,7 +306,7 @@ func (c *chainWatcher) SubscribeChannelEvents() *ChainEventSubscription {
 
 	sub := &ChainEventSubscription{
 		ChanPoint:               c.cfg.chanState.FundingOutpoint,
-		RemoteUnilateralClosure: make(chan *lnwallet.UnilateralCloseSummary, 1),
+		RemoteUnilateralClosure: make(chan *RemoteUnilateralCloseInfo, 1),
 		LocalUnilateralClosure:  make(chan *LocalUnilateralCloseInfo, 1),
 		CooperativeClosure:      make(chan *CooperativeCloseInfo, 1),
 		ContractBreach:          make(chan *lnwallet.BreachRetribution, 1),
@@ -373,6 +430,30 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 			return
 		}
 
+		// Fetch the current known commit height for the remote party,
+		// and their pending commitment chain tip if it exist.
+		remoteStateNum := remoteCommit.CommitHeight
+		remoteChainTip, err := c.cfg.chanState.RemoteCommitChainTip()
+		if err != nil && err != channeldb.ErrNoPendingCommit {
+			log.Errorf("unable to obtain chain tip for "+
+				"ChannelPoint(%v): %v",
+				c.cfg.chanState.FundingOutpoint, err)
+			return
+		}
+
+		// Now that we have all the possible valid commitments, we'll
+		// make the CommitSet the ChannelArbitrator will need it in
+		// order to carry out its duty.
+		commitSet := CommitSet{
+			HtlcSets: make(map[HtlcSetKey][]channeldb.HTLC),
+		}
+		commitSet.HtlcSets[LocalHtlcSet] = localCommit.Htlcs
+		commitSet.HtlcSets[RemoteHtlcSet] = remoteCommit.Htlcs
+		if remoteChainTip != nil {
+			htlcs := remoteChainTip.Commitment.Htlcs
+			commitSet.HtlcSets[RemotePendingHtlcSet] = htlcs
+		}
+
 		// We'll not retrieve the latest sate of the revocation store
 		// so we can populate the information within the channel state
 		// object that we have.
@@ -411,8 +492,10 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 		// as we don't have any further processing we need to do (we
 		// can't cheat ourselves :p).
 		if isOurCommit {
+			commitSet.ConfCommitKey = &LocalHtlcSet
+
 			if err := c.dispatchLocalForceClose(
-				commitSpend, *localCommit,
+				commitSpend, *localCommit, commitSet,
 			); err != nil {
 				log.Errorf("unable to handle local"+
 					"close for chan_point=%v: %v",
@@ -439,17 +522,6 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 		log.Warnf("Unprompted commitment broadcast for "+
 			"ChannelPoint(%v) ", c.cfg.chanState.FundingOutpoint)
 
-		// Fetch the current known commit height for the remote party,
-		// and their pending commitment chain tip if it exist.
-		remoteStateNum := remoteCommit.CommitHeight
-		remoteChainTip, err := c.cfg.chanState.RemoteCommitChainTip()
-		if err != nil && err != channeldb.ErrNoPendingCommit {
-			log.Errorf("unable to obtain chain tip for "+
-				"ChannelPoint(%v): %v",
-				c.cfg.chanState.FundingOutpoint, err)
-			return
-		}
-
 		// If this channel has been recovered, then we'll modify our
 		// behavior as it isn't possible for us to close out the
 		// channel off-chain ourselves. It can only be the remote party
@@ -465,9 +537,10 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 		// we'll trigger the unilateral close signal so subscribers can
 		// clean up the state as necessary.
 		case broadcastStateNum == remoteStateNum && !isRecoveredChan:
+			commitSet.ConfCommitKey = &RemoteHtlcSet
 
 			err := c.dispatchRemoteForceClose(
-				commitSpend, *remoteCommit,
+				commitSpend, *remoteCommit, commitSet,
 				c.cfg.chanState.RemoteCurrentRevocation,
 			)
 			if err != nil {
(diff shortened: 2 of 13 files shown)


contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,78 @@ const (
...
+	RemotePendingCommit *channeldb.ChannelCommitment
+}
+
+// toActiveHTLCSets returns the set of all active HTLCs across all commitment
+// transactions.
+func (c *CommitSet) toActiveHTLCSets() map[HtlcSetKey]htlcSet {
+	htlcSets := make(map[HtlcSetKey]htlcSet)
+
+	htlcSets[HtlcSetKey{
+		IsRemote: false, IsPending: false,
```
cfromknecht: would help readability to use the defined keys, e.g. `LocalHtlcSetKey`

contractcourt/chain_watcher.go
```diff
@@ -373,6 +431,28 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 			return
 		}
 
+		// Fetch the current known commit height for the remote party,
+		// and their pending commitment chain tip if it exist.
+		remoteStateNum := remoteCommit.CommitHeight
+		remoteChainTip, err := c.cfg.chanState.RemoteCommitChainTip()
+		if err != nil && err != channeldb.ErrNoPendingCommit {
+			log.Errorf("unable to obtain chain tip for "+
```
cfromknecht: nit: capitalize log message

contractcourt/chain_watcher.go
```diff
@@ -373,6 +431,28 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
+				"ChannelPoint(%v): %v",
+				c.cfg.chanState.FundingOutpoint, err)
+			return
+		}
+
+		// Now that we have all the possible valid commitments, we'll
+		// make the CommitSet s the ChannelArbitrator will need it in
```
cfromknecht: stray "s"

contractcourt/channel_arbitrator.go
```diff
@@ -191,6 +191,30 @@ func newHtlcSet(htlcs []channeldb.HTLC) htlcSet {
 	}
 }
 
+// HtlcSetKey is a two-tuple that uniquely identifies a set of HTLCs on a
+// commitment transaction.
+type HtlcSetKey struct {
+	// IsResolved denotes if the HTLCs are on the remote commitment
```
cfromknecht: `IsResolved` -> `IsRemote`

contractcourt/channel_arbitrator.go
```diff
@@ -191,6 +191,30 @@ func newHtlcSet(htlcs []channeldb.HTLC) htlcSet {
...
+	// transaction.
+	IsRemote bool
+
+	// IsDangling denotes if the commitment transaction that HTLCS are on
```
cfromknecht: `IsDangling` -> `IsPending`

contractcourt/channel_arbitrator.go
```diff
@@ -209,7 +233,7 @@ type ChannelArbitrator struct {
 
 	// activeHTLCs is the set of active incoming/outgoing HTLC's on the
 	// commitment transaction.
-	activeHTLCs htlcSet
+	activeHTLCs map[HtlcSetKey]htlcSet
```
cfromknecht: godoc needs updating

contractcourt/channel_arbitrator.go
```diff
@@ -802,7 +837,19 @@ func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
...
+		// TODO(roasbeef): not needed since we must have a commitment
+		// output in order to make an outgoing HTLC?
+		noCommitHtlcs := true
+		if confCommitSet != nil {
+			localHTLCs := len(confCommitSet.LocalCommit.Htlcs) > 0
+			remoteHTLCs := len(confCommitSet.RemoteCommit.Htlcs) > 0
+			danglingHtlcs := (confCommitSet.RemotePendingCommit != nil &&
+				len(confCommitSet.RemotePendingCommit.Htlcs) > 0)
+
+			noCommitHtlcs = !localHTLCs && !remoteHTLCs && !danglingHtlcs
```
cfromknecht: adding a helper method to compute this would be useful, e.g. `confCommitSet.IsEmpty()`

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1321,171 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+	localCommitActions, err := c.checkCommitChainActions(
+		height, trigger, activeHTLCs[LocalHtlcSet],
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Next, we'll examine the remote commitment (and maybe a dangling one)
+	// to see if the set difference of our HTLCs is non-empty. If so, then
+	// we may need to cnacel back some HTLCs if we decide go to chain.
```
cfromknecht: cnacel -> cancel

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1321,171 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+	for htlcSetKey, htlcs := range activeHTLCs {
+		if htlcSetKey.IsRemote {
+			continue
+		}
+
+		for _, htlc := range htlcs.outgoingHTLCs {
+			localHTLCs[htlc.HtlcIndex] = htlc
+		}
+	}
+	for htlcSetKey, htlcs := range activeHTLCs {
```
cfromknecht: the local and remote htlcs can be computed in a single pass

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1321,171 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+// that are close to timing out. If we find any, then we'll return a set of
+// chain actions for HTLCs that are on our commitment, but not theirs to cancel
+// immediately.
+func (c *ChannelArbitrator) checkRemoteDanglingActions(
+	height uint32, activeHTLCs map[HtlcSetKey]htlcSet,
+	commitsConfirmed bool) ChainActionMap {
+
+	var (
+		pendingRemoteHTLCs []channeldb.HTLC
+		localHTLCs         = make(map[uint64]channeldb.HTLC)
```
cfromknecht: looks like this can be made a `map[uint64]struct{}`

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1321,171 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+	// remote commitment transaction, but not on ours.
+	for htlcIndex, htlc := range remoteHTLCs {
+		if _, ok := localHTLCs[htlcIndex]; ok {
+			continue
+		}
+
+		pendingRemoteHTLCs = append(pendingRemoteHTLCs, htlc)
+	}
+
+	// Finally, we'll examine al the pending remote HTLCs for those that
```
cfromknecht: al -> all

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1321,171 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+	// are close to expiring. If we find any, then we'll recommend that
```
cfromknecht: should this read "have expired" instead of "close to expiring", since our outgoing broadcast delta is now zero?

htlcswitch/link.go
```diff
@@ -1720,8 +1720,11 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 		// Since we just revoked our commitment, we may have a new set
 		// of HTLC's on our commitment, so we'll send them over our
 		// HTLC update channel so any callers can be notified.
+		newHTLCS := map[contractcourt.HtlcSetKey][]channeldb.HTLC{
```
cfromknecht: defining a new type, e.g. `contractcourt.HtlcSets` would help cut down on the verbosity here

lnwallet/channel.go
```diff
@@ -3035,8 +3035,10 @@ func (lc *LightningChannel) createCommitDiff(
 // The first return parameter is the signature for the commitment transaction
 // itself, while the second parameter is a slice of all HTLC signatures (if
 // any). The HTLC signatures are sorted according to the BIP 69 order of the
-// HTLC's on the commitment transaction.
-func (lc *LightningChannel) SignNextCommitment() (lnwire.Sig, []lnwire.Sig, error) {
+// HTLC's on the commitment transaction. Finally, the new set of pending HTLCs
+// for the remote party's commitment are also returned.
+func (lc *LightningChannel) SignNextCommitment() (lnwire.Sig, []lnwire.Sig, []channeldb.HTLC, error) {
```
cfromknecht: nit: long line

contractcourt/channel_arbitrator.go
```diff
@@ -851,11 +899,14 @@ func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
 
 		// With the commitment broadcast, we'll then send over all
 		// messages we can send immediately.
-		err = c.cfg.DeliverResolutionMsg(pktsToSend...)
-		if err != nil {
-			// TODO(roasbeef): make sure packet sends are idempotent
-			log.Errorf("unable to send pkts: %v", err)
-			return StateError, closeTx, err
+		if len(pktsToSend) != 0 {
```
cfromknecht: `DeliverResolutionMsg` is already a NOP for zero-length slices. if this is only for satisfying the tests, perhaps we update the tests to ignore empty slices instead of vice versa?

Roasbeef: Yeah this was added just so the test could pass. Once could argue that that current check is an underlying detail of the default `DeliverResolutionMsg` closure, so it's simply a defensive check on the part of the channel arb.

contractcourt/chain_watcher.go
Roasbeef: This existed before this diff, the log messages are also inconsistent in this file as is.

contractcourt/chain_watcher.go
Roasbeef: Fixed, forgot to update this after I aded the definitions.

Roasbeef:
Updated with a new commit structure to ensure each of them compile.

contractcourt/channel_arbitrator.go
cfromknecht: On the other hand, one could argue that we lose a valuable testing hook, that could otherwise be used to assert the internal state machine is progressing as expected ;)

contractcourt/chain_watcher.go
```diff
@@ -373,6 +427,28 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
+		// Now that we have all the possible valid commitments, we'll
+		// make the CommitSet the ChannelArbitrator will need it in
+		// order to carry out its duty.
+		commitSet := CommitSet{
+			LocalCommit:  *localCommit,
+			RemoteCommit: *remoteCommit,
+		}
+		if remoteChainTip != nil {
+			commitSet.RemotePendingCommit = &remoteChainTip.Commitment
+		}
```
joostjager: Shouldn't c.cfg.chanState.LatestCommitments() return the remotePendingCommit too, instead of fetching it separately? Or expose it as a field on OpenChannel?

contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,74 @@ const (
...
+// the CommitSet at the time of channel closure.
+type RemoteUnilateralCloseInfo struct {
+	*lnwallet.UnilateralCloseSummary
+
+	// CommitSet is the set of known valid commitments at the time the
+	// remote party's commitemnt hit the chain.
+	CommitSet CommitSet
+}
+
+// CommitSet is a collection of the set of known validate commitments at a
```
joostjager: s/validate/valid

contractcourt/channel_arbitrator.go
```diff
@@ -191,6 +191,30 @@ func newHtlcSet(htlcs []channeldb.HTLC) htlcSet {
...
+	// LocalHtlcSet is the HtlcSetKey used for local commitments.
+	LocalHtlcSet = HtlcSetKey{IsRemote: false, IsPending: false}
+
+	// RemoteHtlcSet is the HtlcSetKey used for remote commitments.
+	RemoteHtlcSet = HtlcSetKey{IsRemote: true, IsPending: false}
+
+	// RemotePendingHtlcSet is the HtlcSetKey used for dangling remote
+	// commitment transactions.
+	RemotePendingHtlcSet = HtlcSetKey{IsRemote: true, IsPending: true}
+)
```
joostjager: Isn't it clearer to just define this as an enum? There may be an extra `||` in some if statement, but it can be worth it.

contractcourt/channel_arbitrator.go
```diff
@@ -600,8 +626,9 @@ func (t transitionTrigger) String() string {
 // the appropriate state transition if necessary. The next state we transition
 // to is returned, Additionally, if the next transition results in a commitment
 // broadcast, the commitment transaction itself is returned.
-func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
-	trigger transitionTrigger) (ArbitratorState, *wire.MsgTx, error) {
+func (c *ChannelArbitrator) stateStep(
+	triggerHeight uint32, trigger transitionTrigger,
+	confCommitSet *CommitSet) (ArbitratorState, *wire.MsgTx, error) {
```
joostjager: `confCommitSet` unused (in this commit).

contractcourt/channel_arbitrator.go
joostjager: Also these `HtlcSetKey` fields are not used in this commit.

contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,74 @@ const (
 	maxCommitPointPollTimeout = 10 * time.Minute
 )
 
-// LocalUnilateralCloseInfo encapsulates all the informnation we need to act
-// on a local force close that gets confirmed.
+// LocalUnilateralCloseInfo encapsulates all the informnation we need to act on
```
joostjager: s/informnation/information

contractcourt/chain_watcher.go
```diff
@@ -54,7 +54,7 @@ type RemoteUnilateralCloseInfo struct {
 	*lnwallet.UnilateralCloseSummary
 
 	// CommitSet is the set of known valid commitments at the time the
-	// remote party's commitemnt hit the chain.
+	// remote party's commitment hit the chain.
```
joostjager: Typo fixed in wrong commit

contractcourt/chain_watcher.go
```diff
@@ -74,7 +74,7 @@ type CommitSet struct {
 	// RemoteCommit is the latest valid commitment for the remote party.
 	RemoteCommit channeldb.ChannelCommitment
 
-	// RemotePendingCommit if non nil, is the pending commmitment for the
+	// RemotePendingCommit if non nil, is the pending commitment for the
```
joostjager: Typo fixed in wrong commit

contractcourt/channel_arbitrator.go
```diff
@@ -1298,6 +1307,118 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+		// been confirmed, then we can move on. Otherwise, if
+		// commitments have been confirmed, then we need to cancel back
+		// *all* of the pending remote HTLCS.
+		if !goToChain && !commitsConfirmed {
+			continue
+		}
+
+		log.Tracef("ChannelArbitrator(%v): immediately "+
+			"htlc=%x about to timeout on remote commitment",
+			c.cfg.ChanPoint, htlc.RHash[:])
```
joostjager: Log text reads not as a sentence.

contractcourt/channel_arbitrator.go
```diff
@@ -824,7 +824,7 @@ func (c *ChannelArbitrator) stateStep(
 	case StateContractClosed:
 		// First, we'll fetch our chain actions, and both sets of
 		// resolutions so we can process them.
-		chainActions, err := c.log.FetchChainActions()
+		_, err := c.log.FetchChainActions()
```
joostjager: Commit message says we reconstruct chain actions, but here they are still fetched from the log.

contractcourt/channel_arbitrator.go
joostjager: `CommitSet.IsEmpty` is defined, but not called here?

contractcourt/channel_arbitrator.go
```diff
@@ -424,18 +424,24 @@ func (c *ChannelArbitrator) relaunchResolvers() error {
...
 	// db migration.
 	htlcMap := make(map[wire.OutPoint]*channeldb.HTLC)
-	chainActions, err := c.log.FetchChainActions()
-	if err != nil {
-		log.Errorf("unable to fetch chain actions: %v", err)
-		return err
-	}
-	for _, htlcs := range chainActions {
-		for _, htlc := range htlcs {
+	for _, htlcs := range c.activeHTLCs {
```
joostjager: Make this change in the same commit where we stop writing chain actions?

contractcourt/briefcase.go
```diff
@@ -82,17 +81,6 @@ type ArbitratorLog interface {
...
-	// LogChainActions stores a set of chain actions which are derived from
-	// our set of active contracts, and the on-chain state. We'll write
-	// this et of cations when: we decide to go on-chain to resolve a
-	// contract, or we detect that the remote party has gone on-chain.
-	LogChainActions(ChainActionMap) error
-
-	// FetchChainActions attempts to fetch the set of previously stored
-	// chain actions. We'll use this upon restart to properly advance our
-	// state machine forward.
-	FetchChainActions() (ChainActionMap, error)
```
joostjager: I would remove this in the same commit where we stop using them, to prevent having dead code in master.

contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,91 @@ const (
...
-// on a local force close that gets confirmed.
+// LocalUnilateralCloseInfo encapsulates all the informnation we need to act on
+// a local force close that gets confirmed.
 type LocalUnilateralCloseInfo struct {
 	*chainntnfs.SpendDetail
 	*lnwallet.LocalForceCloseSummary
 	*channeldb.ChannelCloseSummary
+
+	// CommitSet is the set of known valid commitments at the time the
+	// remote party's commitemnt hit the chain.
```
alexbosworth:
```suggestion
	// remote party's commitment hit the chain.
```

contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,91 @@ const (
...
+	CommitSet CommitSet
 }
 
-// CooperativeCloseInfo encapsulates all the informnation we need to act
-// on a cooperative close that gets confirmed.
+// CooperativeCloseInfo encapsulates all the informnation we need to act on a
```
alexbosworth:
```suggestion
// CooperativeCloseInfo encapsulates all the information we need to act on a
```

contractcourt/chain_watcher.go
```diff
@@ -373,6 +444,28 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
...
+		remoteChainTip, err := c.cfg.chanState.RemoteCommitChainTip()
+		if err != nil && err != channeldb.ErrNoPendingCommit {
+			log.Errorf("unable to obtain chain tip for "+
+				"ChannelPoint(%v): %v",
+				c.cfg.chanState.FundingOutpoint, err)
+			return
+		}
+
+		// Now that we have all the possible valid commitments, we'll
+		// make the CommitSet the ChannelArbitrator will need it in
```
alexbosworth:
```suggestion
		// make the CommitSet the ChannelArbitrator will need in
```

contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,91 @@ const (
...
+	// LocalCommit is the latest valid commitment for the local party.
+	LocalCommit channeldb.ChannelCommitment
+
+	// RemoteCommit is the latest valid commitment for the remote party.
+	RemoteCommit channeldb.ChannelCommitment
+
+	// RemotePendingCommit if non nil, is the pending commitment for the
+	// remote party. This is a commitment that we've sent to them, but have
+	// not yet received a revocation for.
+	RemotePendingCommit *channeldb.ChannelCommitment
```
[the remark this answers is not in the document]
Roasbeef: I think in the case of HTLCSets a map is more apt given the type of manipulation we do to interact with the items. We use pointers here to signal a field that may not always be populated. This struct also has the extra `ConfCommitKey` value as well.

contractcourt/chain_watcher.go
```diff
@@ -411,8 +504,10 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 		// as we don't have any further processing we need to do (we
 		// can't cheat ourselves :p).
 		if isOurCommit {
+			commitSet.ConfCommitKey = &LocalHtlcSet
+
 			if err := c.dispatchLocalForceClose(
-				commitSpend, *localCommit,
+				commitSpend, commitSet,
```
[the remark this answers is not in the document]
Roasbeef: Yes, it's possible that:
  * we go to force close a channel and have what we beielve to be an update to date HTLC set
  * we write that to disk (not done anymore, but provided for the sake of the scenario)]
  * we transition states in the channel arb
  * during that we revoke a commitment and block on sending it to the channel arb
  * channel arb force closes

At that point, if we use only the in-memory state, we may miss out on some context, or in the worse case if the entire set of HTLCs has changed, then we're using stale data.

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1321,164 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+		htlcActions, err = c.checkLocalChainActions(
+			height, trigger, htlcSets, true,
+		)
+
+	// If the remote commitment confirmed, then we only need to examine the
+	// remote set for any HTLCs.
+	case RemoteHtlcSet:
+		htlcActions, err = c.checkCommitChainActions(
+			height, trigger, htlcSets[RemoteHtlcSet],
+		)
```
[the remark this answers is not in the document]
Roasbeef: That would be the case if there's no incoming HTLC at all, so it's a direct payment. In that case we could go to chain (as written in order to fully resolve it). However since we have an outgoing broadcast delta of zero atm, we would've went to chain anyway in order to resolve it.

contractcourt/channel_arbitrator.go
```diff
@@ -1856,7 +2053,7 @@ func (c *ChannelArbitrator) channelAttendant(bestHeight int32) {
 			// a terminal state.
 			_, _, err = c.advanceState(
 				uint32(uniClosure.SpendingHeight),
-				remoteCloseTrigger,
+				remoteCloseTrigger, &uniClosure.CommitSet,
```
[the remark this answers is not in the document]
Roasbeef: We pass in all 3 as our in memory state and what actually confirmed in the chain may have diverged.

contractcourt/channel_arbitrator.go
```diff
@@ -802,7 +837,19 @@ func (c *ChannelArbitrator) stateStep(triggerHeight uint32,
 		// If the resolution is empty, then we're done here. We don't
 		// need to launch any resolvers, and can go straight to our
 		// final state.
-		if contractResolutions.IsEmpty() {
+		//
+		// TODO(roasbeef): not needed since we must have a commitment
+		// output in order to make an outgoing HTLC?
```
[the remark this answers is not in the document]
Roasbeef: Yes that if we've made an outgoing HTLC, then due to the reserve requirements, we _must_ also have an output since we need funds to make that HTLC. However, thinking about it more: it's possible that we do have enough funds and have met the reserve, but the output is now considered dust.

contractcourt/chain_watcher.go
```diff
@@ -30,20 +30,91 @@ const (
...
 }
 
+// RemoteUnilateralCloseInfo wraps the normal UnilateralCloseSummary to couple
+// the CommitSet at the time of channel closure.
+type RemoteUnilateralCloseInfo struct {
+	*lnwallet.UnilateralCloseSummary
+
+	// CommitSet is the set of known valid commitments at the time the
+	// remote party's commitment hit the chain.
+	CommitSet CommitSet
```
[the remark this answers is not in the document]
Roasbeef: Just the HTLCs atm, but I don't think there's any harm in passing the rest in. It's also nice since the channel arb now gains _full_ context on the valid commitment state at time of channel closure.

contractcourt/channel_arbitrator_test.go
```diff
@@ -144,27 +134,33 @@ func (*mockChainIO) GetBlock(blockHash *chainhash.Hash) (*wire.MsgBlock, error)
 }
 
 func createTestChannelArbitrator(log ArbitratorLog) (*ChannelArbitrator,
-	chan struct{}, error) {
+	chan struct{}, chan []ResolutionMsg, chan *chainntnfs.BlockEpoch, error) {
```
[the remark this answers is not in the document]
Roasbeef: It builds, but the tests don't run. I only went for building.

contractcourt/chain_watcher.go
Roasbeef: Possibly, they could very well be combined but I don't think we need to do that in this PR (this is just code that was moved around). It isn't a field since this needs to be read from disk and my not be present at all.

contractcourt/channel_arbitrator.go
Roasbeef: I don't think there's a big win as an enum or a struct. The struct is attractive IMO as we can add additional fields and still have the map operations work properly. They're introduced since the concept of the `HtlcSet` is also introduced in this commit.

contractcourt/channel_arbitrator.go
Roasbeef: Adding it in this commit made it easier to construct the commits that came after it. I don't think there's much to be gained by moving around this small portion of the diff.

contractcourt/chain_watcher.go
Roasbeef: This was already there, but fixed.

contractcourt/channel_arbitrator.go
Roasbeef: I don't think we have such a hard requirement. If so many other log messages would need to be modified. This is an existing log message with a swap swapped in.

contractcourt/channel_arbitrator.go
Roasbeef: They're still fetched, but we don't use them.

contractcourt/channel_arbitrator.go
Roasbeef: Ah messed up in the rebase somehow, should be fixed in the latest diff.

contractcourt/channel_arbitrator.go
Roasbeef: I think it stands on its own as: if we've already written chain actions for this channel, then the return parameter will be properly populated.

Roasbeef:
PTAL @cfromknecht @joostjager!

contractcourt/chain_watcher.go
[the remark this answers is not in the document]
joostjager: In this PR, almost in everywhere a specific set is explictly picked (local, remote, remotepending). It seems that only when something comes in through `htlcUpdates`, the manipulation is easier with a map. If combined with my other comment about just storing the htlcs in `channelArbitrator` rather than the full commit tx, I think the code would be simpler, easier to understand.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: I am thinking about the case where we've sent commitSig only for a forwarded htlc. That htlc is then in the remote pending commit tx only. Then the remote party force closes the channel with the remote commit (not including that htlc). What makes us cancel the incoming htlc in that case?

contractcourt/chain_watcher.go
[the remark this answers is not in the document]
joostjager: In general, I think there is harm in passing in more info than needed. The extreme is making every variable global so all code can access everything.

My preference would be to pass in just what is needed and only change that when there is a real need for the channel arbitrator to know more.

contractcourt/channel_arbitrator_test.go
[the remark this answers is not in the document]
joostjager: Ok. I saw some tests are updated in commits though. What is the reason for leaving this build error until the end of the commit set?

contractcourt/channel_arbitrator.go
joostjager: Ok, not always clear to see in the diff what was pre-existing.

contractcourt/channel_arbitrator.go
joostjager: To me that makes the commit not completely logical. Why would you fetch them if not used?

contractcourt/channel_arbitrator.go
joostjager: But doesn't that mean that the commit where you stopped writing chain actions broke this part of the code?

contractcourt/chain_watcher.go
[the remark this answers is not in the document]
Roasbeef: Yeah we could grab it then as well. However we'd need to then store the commitment set in a private attribute in order to have it be available for the next call to `stateStep` so we can prep the contract resolvers. Functionally, it's the same path, but IMO having the chain watcher provide this state (as it's what tells the channel arb what confirmed on chain) is more natural within this context.

contractcourt/channel_arbitrator.go
```diff
@@ -1794,7 +1986,7 @@ func (c *ChannelArbitrator) channelAttendant(bestHeight int32) {
 			// a terminal state.
 			_, _, err = c.advanceState(
 				uint32(closeInfo.SpendingHeight),
-				localCloseTrigger,
+				localCloseTrigger, &closeInfo.CommitSet,
```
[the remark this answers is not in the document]
Roasbeef: The commit set is only obtained when the chain watcher sends us message that something has confirmed on chain. We only use the commit set when prepping contract resolutions (and check for nil-ness in other areas). We only prep resolutions once a commitment has confirmed. Therefore we'll only access this when it's non-nil and a commitment has confirmed.

contractcourt/channel_arbitrator.go
Roasbeef: The diff is to be taken as a whole (all the commits), if a user cherry picks some commits, but not others, then behavior will be broken. As I said earlier, the commits are constructed so things compile, but if you run things like the itests at any commit but the end, they may fail.

contractcourt/channel_arbitrator.go
Roasbeef: It makes the commit diffs smaller, of course I could just do everything in a single commit. If you follow the commits step by step, it shows that we start to reconstruct, then no longer write them all together.

contractcourt/channel_arbitrator_test.go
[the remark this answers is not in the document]
Roasbeef: Simply preference (I went for just building), I don't think leaving it till the end invalidates the diff or makes things any harder to review.

contractcourt/chain_watcher.go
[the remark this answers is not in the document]
Roasbeef: I don't think we're anywhere near that extreme. I don't consider this a blocking item, there're many other areas in the codebase where we use an entire struct for convenience when we only need a few fields. It's also the case that the chain watcher already passes in the entire commitment confirmed, we just modify it to also pass in all valid commitments at time of confirmation.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
Roasbeef: Ahh I see what you're getting at! Updated the tests to cover: "remote commit conf, with remote pending htlc" (htlc expired and not).

contractcourt/chain_watcher.go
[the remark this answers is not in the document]
Roasbeef: > In this PR, almost in everywhere a specific set is explictly picked (local, remote, remotepending).

There're also areas where pass all 3 and examine 2 of them. The two representations are isomorphic, as a result it's up to preference which of them are used.

Roasbeef:
PTAL @joostjager @cfromknecht

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: But if we crash at this point, will we get the commitment confirmed signal again after restart from chain watcher?

contractcourt/channel_arbitrator.go
joostjager: I do it differently and only create commits that have no broken behaviour. I also try to get the unit tests and itests running. If that seems to be work that is reverted again in later commits, it is an indication that my commit structure isn't right. But it is something up to preference too.

contractcourt/channel_arbitrator.go
joostjager: My preference (it is a preference) is to have one logical commit that makes the change, keeping behaviour in tact.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: And how about the other way around, remote pending confirms, with htlc still on the remote (htlc already failed on the remote pending)?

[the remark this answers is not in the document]
Roasbeef: If I'm following then then that's not possible. The only way the remote pending commit can have something that a remote commit doesn't is if it was manually failed back, and the only party that can fail back HTLCs is the one that received the HTLC.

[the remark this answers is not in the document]
Roasbeef: If the remote pending confirms, then we just play that as is. If the remote pending has the HTLC failed, then the remote party would've needed to send us a sig, then we revoke, meaning that we also don't have the HTLC locally on our commitment transaction.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
Roasbeef: Hmm yeah at an initial glance, I thought it would continue to send this information until we mark the channel as closed, which is the case, but we do so before we start to advance our state. Atm, we do so only after we finish writing the contract resolutions here: https://github.com/lightningnetwork/lnd/blob/master/contractcourt/channel_arbitrator.go#L1837

It appears that we may also need to start to write this commitment transaction information, as this state isn't stored in the `ChannelCloseSummary` which is what's available to us once we mark a channel as closed.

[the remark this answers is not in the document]
Roasbeef: To make this backwards compatible for old nodes (that are for the case of this scenario, "frozen" in this limbo state), then if this new piece of information isn't present, we just only play the commitment at hand.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: Isn't it in that case that we only cancel back our incoming htlc once we receive the revoke from the remote party? So:
* Remote party sends us a sig over a commit tx that doesn't contain the htlc anymore
* We send the revoke the to remote party
* We remove htlc from our remote tx and send a sig (it is now removed from remote pending)
* The remote party publishes that remote pending tx without sending a revoke
* What mechanism now lets us now cancel back the incoming?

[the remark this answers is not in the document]
joostjager: The only tx where the htlc is still in is the remote tx, but we don't inspect that

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: Can this then combined with that other comment about getting the final commit sets during force close?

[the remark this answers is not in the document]
joostjager: So not passing them in from chain watcher at all anymore

contractcourt/briefcase.go
```diff
@@ -802,6 +781,59 @@ func (b *boltArbitratorLog) FetchChainActions() (ChainActionMap, error) {
...
+		if err != nil {
+			return err
+		}
+
+		var b bytes.Buffer
+		if err := encodeCommitSet(&b, commits); err != nil {
+			return err
+		}
+
+		return scopeBucket.Put(commitSetKey, b.Bytes())
```
joostjager: It still feels like redundant data. The commit sets are already persisted by the channel and now here again. Couldn't storing just the type of commit (local, remote, remote pending) be enough?

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
Roasbeef: I implemented what I described above in this latest diff. I still prefer passing the information from the chain watcher as IMO it's easier to reason about the consistency of the commitment states at that point since the channel has been closed.

contractcourt/briefcase.go
Roasbeef: Once the channel has been marked as pending closed, we no longer have access to that data since the channel close summary doesn't store the full commitment information. Once this is written, you're correct that the data is redundant, but once we proceed to the next call to mark the channel as pending close, it's the only area in the database that we store this information. Looking back upon it, it was a mistake to remove all the channel state once we moved from open -> pending close. A few times now, we've had to supplement the pending close information in order to carry out tasks like what's implemented in this PR.

joostjager: So shouldn't that data be kept a bit longer in the channel buckets then? To prevent serializing them there when the channel is open and storing it in the new store when it is pending close?

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: Ok, I thought maybe this last change changed your view on this.

contractcourt/briefcase.go
Roasbeef: It's possible for us to head back in this direction, but it's a larger change over all. We'd either need to move all that state into the current channel close summary, or modify `CloseChannel` to hold onto all the state or "move" all the channel state into a different bucket since if it lingers there, then many other sub-systems/callers will think the channel is still open.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
Roasbeef: Extended the tests to cover this case.

Roasbeef:
PTAL @cfromknecht @joostjager

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1340,254 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+		height, activeHTLCs, commitsConfirmed,
+	)
+
+	// Finally, we'll merge the two set of chain actions.
+	localCommitActions.Merge(remoteDanglingActions)
+
+	return localCommitActions, nil
+}
+
+// checkRemoteChainActions examines the set of remote commitments for any HTLCs
```
cfromknecht: godoc/name mismatch

contractcourt/briefcase.go
```diff
@@ -82,15 +82,22 @@ type ArbitratorLog interface {
...
+	// from the database.
+	FetchConfirmedCommitSet() (*CommitSet, error)
 
 	// FetchChainActions attempts to fetch the set of previously stored
 	// chain actions. We'll use this upon restart to properly advance our
 	// state machine forward.
+	//
+	// NOTE: This method only exists in order to be able to serve nodes had
+	// channels in the process of closing before the CommitSet struct was
+	// introduced.
```
joostjager: What is the plan to remove this eventually? Or will we never able to?

contractcourt/briefcase.go
```diff
@@ -802,6 +781,59 @@ func (b *boltArbitratorLog) FetchChainActions() (ChainActionMap, error) {
...
+		commitSet, err := decodeComitSet(bytes.NewReader(commitSetBytes))
+		if err != nil {
+			return err
+		}
+
+		c = commitSet
+		return nil
+	})
+	if err != nil {
+		return nil, nil
```
joostjager: `return nil, err` ?

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1340,254 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+	return remoteCommitActions, nil
+}
+
+// checkRemoteDiffActions checks the set difference of the HTLCs on the remote
+// confirmed commit and remote dangling commit for HTLCS that we need to cancel
+// back. If we find any HTLCs on the remote pending but not the remote, then
+// we'll mark them to be failed immediately.
+func (c *ChannelArbitrator) checkRemoteDiffActions(height uint32,
+	activeHTLCs map[HtlcSetKey]htlcSet,
+	pendingConf bool) ChainActionMap {
```
joostjager: I think it would be clearer to pass in the two set keys instead of the boolean. Make it a more generic diff "htlcs in B but not in A". Then what A and B are can be specified in `checkRemoteChainActions`.

contractcourt/channel_arbitrator.go
```diff
@@ -1269,15 +1340,254 @@ func (c *ChannelArbitrator) isPreimageAvailable(hash lntypes.Hash) (bool,
...
+	// the CommitSet was introduced. In this case, we'll just return the
+	// existing ChainActionMap they had on disk.
+	if confCommitSet == nil {
+		return c.log.FetchChainActions()
+	}
+
+	// Otherwise we have the full commitment set written to disk, and can
+	// proceed as normal.
+	htlcSets := confCommitSet.toActiveHTLCSets()
+	switch *confCommitSet.ConfCommitKey {
```
joostjager: Make it is an idea to take out the call to `checkCommitChainActions` from `checkLocalChainActions` and `checkRemoteChainActions` and put it here. It is clear then that we always check the confirmed commit and use the switch only to merge the 'extra' actions.

contractcourt/briefcase.go
joostjager: If it touches many other sub-systems, I can imagine it is a bigger change. It seems to me that only persisting commit txes in the channel state (where they already are stored) is a cleaner solution, but probably to big of a refactoring for now.

contractcourt/briefcase.go
Roasbeef: Remove the method? All new channels don't need it, but prior channels do. Once those old channels have been fully resolved (only those in the process of being closed as this PR lands), then it's no longer needed. No current plan to remove it atm (we're able to do so), but we can discuss such a plan offline.

contractcourt/channel_arbitrator.go
Roasbeef: It's just a matter of following the function to see that `checkCommitChainActions` is called first. It also states something along these lines in the `godoc` comments of each of the methods.

contractcourt/channel_arbitrator.go
Roasbeef: If we actually need to generalize it in the future, then it's a small diff away. Atm, we only care about pending or not.

PR #3101, opened 2019-05-21
Allow unit-cover target for make to filter by package

AdamISZ:
Previous to this commit, running `make unit-cover pkg=xx`
would ignore the selected package and run unit tests and
coverage for all packages.
After this commit, the package selected with pkg= is the
only one that is tested and coverage output generated for.

Note: I kept the lnrpc filter for this case too (not 100% sure I understand it), also I think it's a bit suboptimal using `grep` here, so if someone prefers to make a better version than this I'm all for it! :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,8 @@ unit: btcd
 
 unit-cover: $(GOACC_BIN)
 	@$(call print, "Running unit coverage tests.")
-	$(GOACC_BIN) $$(go list ./... | grep -v lnrpc) -- -test.tags="$(DEV_TAGS) $(LOG_TAGS)"
+	$(GOACC_BIN) $(COVER_PKG) -- -test.tags="$(DEV_TAGS) $(LOG_TAGS)"
+
 
 unit-race:
 	@$(call print, "Running unit race tests.")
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -126,7 +126,11 @@ unit: btcd
 
 unit-cover: $(GOACC_BIN)
 	@$(call print, "Running unit coverage tests.")
-	$(GOACC_BIN) $$(go list ./... | grep -v lnrpc) -- -test.tags="$(DEV_TAGS) $(LOG_TAGS)"
+	if [ ! "$(pkg)" = "" ]; then \
```
cfromknecht: this definitely works, tho i have a simpler fix in mind :)

most of the conditional flag parsing is done `make/testing_flags.mk`. here's an example of one that uses a default if none is provided https://github.com/lightningnetwork/lnd/blob/master/make/testing_flags.mk#L33

what do you think of defining a `COVER_PKG` variable there, with a default of `./...`, and using that as the argument to `go list`?

cfromknecht: fwiw, the argument to go list is absolute, so we'd want `COVER_PKG := $(PKG)/pkg` when an arg is provided, which will prepend the full lnd package name

cfromknecht: actually, scratch the last part, `./...` is obviously relative, so we can just prepend a `./` 😅

AdamISZ: Yes, understood, this makes sense. Fixed it up will push here in a mo'

PR #3103, opened 2019-05-21
discovery/sync_manager: restart historical sync on first connected peer

halseth:
To handle the case where we have been without peers, and get a new
connection, we reset the historical scan booleans when the first active
syncer is connected to trigger another historical sync.

This would cause integration tests to be flaky (especially with Neutrino), since we would risk not getting graph updates after having being disconnected.

### DIFF
--- a/discovery/sync_manager.go
+++ b/discovery/sync_manager.go
@@ -186,10 +186,6 @@ func (m *SyncManager) syncerHandler() {
 	defer m.cfg.HistoricalSyncTicker.Stop()
 
 	var (
-		// attemptInitialHistoricalSync determines whether we should
-		// attempt an initial historical sync when a new peer connects.
-		attemptInitialHistoricalSync = true
-
 		// initialHistoricalSyncCompleted serves as a barrier when
 		// initializing new active GossipSyncers. If false, the initial
 		// historical sync has not completed, so we'll defer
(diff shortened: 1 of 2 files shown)


discovery/sync_manager.go
```diff
@@ -225,21 +221,38 @@ func (m *SyncManager) syncerHandler() {
...
-			// it when the initial historical sync completes.
+			// If the initial historical sync has yet to complete,
+			// then we'll declare is as passive and attempt to
+			// transition it when the initial historical sync
+			// completes.
 			case !initialHistoricalSyncCompleted:
 				s.setSyncType(PassiveSync)
 				m.inactiveSyncers[s.cfg.peerPub] = s
 
+			// Otherwise, it should be initialized as active.
```
wpaulino: Nit: seems like a redundant comment.

discovery/sync_manager_test.go
```diff
@@ -219,6 +219,41 @@ func TestSyncManagerInitialHistoricalSync(t *testing.T) {
 	assertNoMsgSent(t, extraPeer)
 }
 
+// TestSyncManagerHistoricalSyncOnReconnect tests that the sync manager will
+// re-trigger a historical sync when a new peer connects after a historical
+// sync has completed, but we have lost all peers.
+func TestSyncManagerHistoricalSyncOnReconnect(t *testing.T) {
```
wpaulino: Nit: would prefer if this commit was squashed with what introduced the behavior.

discovery/sync_manager.go
halseth: Removed: https://github.com/lightningnetwork/lnd/pull/3103/commits/9c9ac36aa9ae7e53af1480cfe2513b041c6a1632

discovery/sync_manager_test.go
halseth: I prefer splitting a bugfix and the test reproducing the bug in separate commits. Especially when I'm reviewing this makes it easy to cherry-pick the test commit on master to exercise the failing scenario.

wpaulino: Sure but that wasn't the case here. The non-test changes alone didn't have an impact on the tests 😛

PR #3105, opened 2019-05-22
fix totalFunds calculation by adding LocalFundedAmt

### LINKED ISSUE #2608: Autopilot miscalculates the total funds
The autopilot agent calculates the total funds available by adding the on chain wallet funds to the total channel allocation.

However, the channel allocation is erroneously calculated by adding the channel capacities instead of the local balance:
https://github.com/lightningnetwork/lnd/blob/a0c0e8edc9a58e8833fd32119964207205f9c764/autopilot/agent_constraints.go#L103

As the threshold is checked against this value, it can lead to unexpected results, especially as remote funding grows larger (which should be excluded from the calculation of the total funds).

Apparently there was a provision for this amount to be calculated properly, but the corresponding field (FundedAmt) is not populated accordingly:
https://github.com/lightningnetwork/lnd/blob/e2285732e47bacd74206f0bbf2c42c6ad8825548/autopilot/graph.go#L103

yashbhutwala:
Fixes #2608 

#### Pull Request Checklist

  logging level
  per default)
  already exist


chanbackup/single.go
```diff
@@ -75,6 +75,10 @@ type Single struct {
 	// Capacity is the size of the original channel.
 	Capacity btcutil.Amount
 
+	// LocalFundedAmt is the amount the local node funded into the target
+	// channel.
+	LocalFundedAmt btcutil.Amount
```
halseth: why is this change needed?

channeldb/channel.go
```diff
@@ -444,6 +444,10 @@ type OpenChannel struct {
 	// Capacity is the total capacity of this channel.
 	Capacity btcutil.Amount
 
+	// LocalFundedAmt is the amount the local node funded into the target
+	// channel.
+	LocalFundedAmt btcutil.Amount
```
halseth: Fields cannot be arbitrarly added like this, since they will also need to be supported in serialization.

autopilot/agent_constraints.go
```diff
@@ -100,7 +100,7 @@ func (h *agentConstraints) ChannelBudget(channels []Channel,
 	// present within the set of active channels.
 	var totalChanAllocation btcutil.Amount
 	for _, channel := range channels {
-		totalChanAllocation += channel.Capacity
+		totalChanAllocation += channel.LocalFundedAmt
```
halseth: Instead of the funded amount, I think we should use the current channel balance?

halseth:
@yashbhutwala Ping.

halseth:
@yashbhutwala Any updates?

halseth:
Closing due to inactivity.

PR #3107, opened 2019-05-22
Makefile: allow passing of tags to integration tests

cfromknecht:


Roasbeef:
Still relevant?

Roasbeef:
A version of this is now present in the makefile as is.

PR #3118, opened 2019-05-25
neutrino: enable user to assert filter header chain.

valentinewallace:
In the case that neutrino's filter header chain is inaccurate, the user can pass in a filter header to neutrino that causes it to reset its chain and resync.


go.mod
```diff
@@ -50,3 +50,5 @@ require (
 replace github.com/lightningnetwork/lnd/ticker => ./ticker
 
 replace github.com/lightningnetwork/lnd/queue => ./queue
+
+replace github.com/lightninglabs/neutrino => github.com/Roasbeef/neutrino v0.0.0-20190523013801-affe42973e63
```
wpaulino: Reminder to remove this once the dependent PR has landed in master.

chainregistry.go
```diff
@@ -751,3 +760,43 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
 
 	return neutrinoCS, cleanUp, nil
 }
+
+func parseHeaderStateAssertion(state map[string]string) (
```
wpaulino: Any reason we use a map here? While there could be a use for allowing multiple header state assertions, the underlying API doesn't allow doing so.

chainregistry.go
```diff
@@ -751,3 +760,43 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+	*headerfs.FilterHeader, error) {
+
+	if len(state) == 0 || state == nil {
```
wpaulino: Nit: both of these checks are essentially the same, a `nil` map has a `0` length.

chainregistry.go
```diff
@@ -751,3 +760,43 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+		return nil, nil
+	}
+
+	var headerStateAssertion = &headerfs.FilterHeader{}
```
wpaulino: Nit: can drop `var`.

chainregistry.go
```diff
@@ -751,3 +760,43 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+	heightArg, ok := state["height"]
+	if !ok {
+		return nil, fmt.Errorf("unable to create neutrino " +
+			"light client: required parameter height " +
+			"missing from header state assertion")
```
wpaulino: Nit: the error here shouldn't contain context of callers, so we should drop `unable to create...` from the error string.

chainregistry.go
wpaulino: Nit: missing `godoc`.

chainregistry.go
```diff
@@ -751,3 +760,43 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+			"light client: required parameter hash " +
+			"missing from header state assertion")
+	}
+	hash, err := chainhash.NewHashFromStr(hashArg)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create neutrino "+
+			"light client: errored parsing hash from "+
+			"header state assertion: %v", err)
+	}
+	headerStateAssertion.FilterHash = *hash
```
cfromknecht: style nit: initialize and return the struct inline at the end of the method, rather than setting each field independently :)

chainregistry.go
valentinewallace: fixed

config.go
```diff
@@ -164,12 +164,13 @@ type chainConfig struct {
...
-	FeeURL       string        `long:"feeurl" description:"Optional URL for fee estimation. If a URL is not specified, static fees will be used for estimation."`
+	AddPeers           []string          `short:"a" long:"addpeer" description:"Add a peer to connect with at startup"`
+	ConnectPeers       []string          `long:"connect" description:"Connect only to the specified peers at startup"`
+	MaxPeers           int               `long:"maxpeers" description:"Max number of inbound and outbound peers"`
+	BanDuration        time.Duration     `long:"banduration" description:"How long to ban misbehaving peers.  Valid time units are {s, m, h}.  Minimum 1 second"`
+	BanThreshold       uint32            `long:"banthreshold" description:"Maximum allowed ban score before disconnecting and banning misbehaving peers."`
+	FeeURL             string            `long:"feeurl" description:"Optional URL for fee estimation. If a URL is not specified, static fees will be used for estimation."`
+	AssertFilterHeader map[string]string `long:"assertfilterheader" description:"Optional filter header in { height: <height> hash: <hash> } format to assert the state of neutrino's filter header chain."`
```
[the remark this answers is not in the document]
valentinewallace: switched :)

chainregistry.go
valentinewallace: fixed

chainregistry.go
```diff
@@ -751,3 +760,34 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+// parseHeaderStateAssertion parses the user-specified neutrino header state
+// into a headerfs.FilterHeader.
+func parseHeaderStateAssertion(state string) (*headerfs.FilterHeader, error) {
+	if len(state) == 0 {
+		return nil, nil
+	}
+
+	split := strings.Split(state, ":")
+	if len(split) != 2 {
+		return nil, fmt.Errorf("header state assertion %v in " +
```
cfromknecht: builds showing a vet error:
```
chainregistry.go:773:15:error: Errorf format %v reads arg #1, but call has 0 args (vet)
```

chainregistry.go
```diff
@@ -751,3 +760,34 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+
+	height, err := strconv.ParseUint(split[0], 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("height parameter from header state "+
+			"assertion is of wrong type: %v", err)
+	}
+
+	hash, err := chainhash.NewHashFromStr(split[1])
+	if err != nil {
+		return nil, fmt.Errorf("errored parsing hash from header state "+
```
wpaulino: Nit: would make the error more clear with `invalid filter header hash: %v`

chainregistry.go
```diff
@@ -751,3 +760,34 @@ func initNeutrinoBackend(chainDir string) (*neutrino.ChainService, func(), error
...
+
+	split := strings.Split(state, ":")
+	if len(split) != 2 {
+		return nil, fmt.Errorf("header state assertion %v in " +
+			"unexpected format")
+	}
+
+	height, err := strconv.ParseUint(split[0], 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("height parameter from header state "+
```
wpaulino: Nit: same here, `invalid filter header height: %v`.

chainregistry.go
wpaulino: Also would be nice to have the error return the expected format.

valentinewallace: :woman_facepalming: thanks, fixed!

cfromknecht:
@wpaulino ready for dep update

config.go
```diff
@@ -169,12 +169,13 @@ type chainConfig struct {
...
-	FeeURL       string        `long:"feeurl" description:"Optional URL for fee estimation. If a URL is not specified, static fees will be used for estimation."`
+	AddPeers           []string      `short:"a" long:"addpeer" description:"Add a peer to connect with at startup"`
+	ConnectPeers       []string      `long:"connect" description:"Connect only to the specified peers at startup"`
+	MaxPeers           int           `long:"maxpeers" description:"Max number of inbound and outbound peers"`
+	BanDuration        time.Duration `long:"banduration" description:"How long to ban misbehaving peers.  Valid time units are {s, m, h}.  Minimum 1 second"`
+	BanThreshold       uint32        `long:"banthreshold" description:"Maximum allowed ban score before disconnecting and banning misbehaving peers."`
+	FeeURL             string        `long:"feeurl" description:"Optional URL for fee estimation. If a URL is not specified, static fees will be used for estimation."`
+	AssertFilterHeader string        `long:"assertfilterheader" description:"Optional filter header in height:hash format to assert the state of neutrino's filter header chain."`
```
wpaulino: Would extend this a bit with "... filter header chain on startup. If the assertion does not hold, then the filter header chain will be re-synced from the genesis block."

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/juju/utils v0.0.0-20180820210520-bf9cc5bdd62d // indirect
 	github.com/juju/version v0.0.0-20180108022336-b64dbd566305 // indirect
 	github.com/kkdai/bstream v0.0.0-20181106074824-b3251f7901ec
-	github.com/lightninglabs/neutrino v0.0.0-20190521105038-03deaa76ddc2
+	github.com/lightninglabs/neutrino v0.0.0-20190605105923-ccf209926edf
```
wpaulino: Could you update this dependency again to the current master?

PR #3131, opened 2019-05-28
[reliable payments] switch result store clean-up

halseth:
~~This commit makes GetPaymentResult return a channel that should be closed by the caller when the result from the result channel is handled. This instructs the Switch that it is safe to delete the result from its
database.~~

~~We make the router close this channel after it has successfully recorded the result, or decided it can re-attempt~~

This is a follow-up from #2762 and is not critical for the operation of `lnd`, but the network results would never be deleted from the Switch's store, causing them to take up space. This fixes that by deleting them when we *know for sure* that the router has handled the result, and won't need it anymore.

~~There is still a gap that can happen if we crash after the router has acked the result, but before the switch has deleted the payment, but in the worst case it will just leave a result lingering in the store.~~

2020 update: we do the clean up simply by syncing the control tower with the network result store on startup, garbage collecting the never-to-be-read-again results.

### DIFF
--- a/htlcswitch/payment_result.go
+++ b/htlcswitch/payment_result.go
@@ -258,3 +258,48 @@ func fetchResult(tx kvdb.RTx, pid uint64) (*networkResult, error) {
 
 	return deserializeNetworkResult(r)
 }
+
+// cleanStore removes all entries from the store, except the payment IDs given.
+// NOTE: Since every result not listed in the keep map will be deleted, care
+// should be taken to ensure no new payment attempts are being made
+// concurrently while this process is ongoing, as its result might end up being
+// deleted.
+func (store *networkResultStore) cleanStore(keep map[uint64]struct{}) error {
+	return kvdb.Update(store.db.Backend, func(tx kvdb.RwTx) error {
+		networkResults, err := tx.CreateTopLevelBucket(
+			networkResultStoreBucketKey,
+		)
+		if err != nil {
+			return err
+		}
+
+		// Iterate through the bucket, deleting all items not in the
+		// keep map.
+		var toClean [][]byte
+		if err := networkResults.ForEach(func(k, _ []byte) error {
+			pid := binary.BigEndian.Uint64(k)
+			if _, ok := keep[pid]; ok {
+				return nil
+			}
+
+			toClean = append(toClean, k)
+			return nil
+		}); err != nil {
+			return err
+		}
+
+		for _, k := range toClean {
+			err := networkResults.Delete(k)
+			if err != nil {
+				return err
+			}
+		}
+
+		if len(toClean) > 0 {
+			log.Infof("Removed %d stale entries from network "+
+				"result store", len(toClean))
+		}
+
+		return nil
+	})
+}
--- a/htlcswitch/payment_result_test.go
+++ b/htlcswitch/payment_result_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/stretchr/testify/require"
 )
 
 // TestNetworkResultSerialization checks that NetworkResults are properly
@@ -180,7 +181,6 @@ func TestNetworkResultStore(t *testing.T) {
 
 	// Since we don't delete results from the store (yet), make sure we
 	// will get subscriptions for all of them.
-	// TODO(halseth): check deletion when we have reliable handoff.
 	for i := uint64(0); i < numResults; i++ {
 		sub, err := store.subscribeResult(i)
 		if err != nil {
@@ -192,5 +192,26 @@ func TestNetworkResultStore(t *testing.T) {
 		case <-time.After(1 * time.Second):
 			t.Fatalf("no result received")
 		}
+	}
+
+	// Clean the store keeping the first two results.
+	toKeep := map[uint64]struct{}{
+		0: {},
+		1: {},
+	}
+	// Finally, delete the result.
+	err = store.cleanStore(toKeep)
+	require.NoError(t, err)
+
+	// Payment IDs 0 and 1 should be found, 2 and 3 should be deleted.
+	for i := uint64(0); i < numResults; i++ {
+		_, err = store.getResult(i)
+		if i <= 1 {
+			require.NoError(t, err, "unable to get result")
+		}
+		if i >= 2 && err != ErrPaymentIDNotFound {
+			t.Fatalf("expected ErrPaymentIDNotFound, got %v", err)
+		}
+
 	}
 }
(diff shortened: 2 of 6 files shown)


joostjager:
Is it worth rebasing this and see how much code is left?

halseth:
Should maybe wait for MPP changes to determine whether this is still needed.

Roasbeef:
Do we still care about this?

halseth:
As stated in #3703 the bucket takes up a good chunk of space unnecessarily, so I think it would make sense to finish this one.

Roasbeef:
@halseth put it back on the milestone, looks like it's gonna take quite a rebase...

halseth:
Rebased. Ended up doing it in a much simpler way, that also gets rid of the gap from the original PR...

htlcswitch/payment_result.go
```diff
@@ -258,3 +258,48 @@ func fetchResult(tx kvdb.RTx, pid uint64) (*networkResult, error) {
 
 	return deserializeNetworkResult(r)
 }
+
+// cleaStore removes all entries from the store, except the payment IDs given.
+// NOTE: Since eery result not listed in the keep map will be deleted, care
```
bhandras, line 263: nit: s/eery/every

htlcswitch/payment_result.go
```diff
@@ -258,3 +258,48 @@ func fetchResult(tx kvdb.RTx, pid uint64) (*networkResult, error) {
...
+	return kvdb.Update(store.db.Backend, func(tx kvdb.RwTx) error {
+		networkResults, err := tx.CreateTopLevelBucket(
+			networkResultStoreBucketKey,
+		)
+		if err != nil {
+			return err
+		}
+
+		var toClean [][]byte
+		if err := networkResults.ForEach(func(k, _ []byte) error {
```
bhandras, line 277: This could be further simplified by using a `kvdb.RwCursor`. Using the cursor there's no need to collect and delete but items can be deleted on the fly.

htlcswitch/payment_result_test.go
```diff
@@ -193,4 +192,26 @@ func TestNetworkResultStore(t *testing.T) {
...
+	// Finally, delete the result.
+	if err := store.cleanStore(toKeep); err != nil {
+		t.Fatal(err)
+	}
+
+	// Payment IDs 0 and 1 should be found, 2 and 3 should be deleted.
+	for i := uint64(0); i < numResults; i++ {
+		_, err = store.getResult(i)
+		if i <= 1 && err != nil {
+			t.Fatalf("unable to get result: %v", err)
```
bhandras, line 210: You could use `require.NoError` instead.

htlcswitch/payment_result_test.go
```diff
@@ -193,4 +192,26 @@ func TestNetworkResultStore(t *testing.T) {
...
+		}
+		if i >= 2 && err != ErrPaymentIDNotFound {
+			t.Fatalf("expected ErrPaymentIDNotFound, got %v", err)
```
bhandras, line 213: Similarly, you could use `require.Error`.

htlcswitch/payment_result_test.go
```diff
@@ -193,4 +192,26 @@ func TestNetworkResultStore(t *testing.T) {
...
 	}
+
+	// Clean the store keeping the first two results.
+	toKeep := map[uint64]struct{}{
+		0: struct{}{},
+		1: struct{}{},
+	}
+	// Finally, delete the result.
+	if err := store.cleanStore(toKeep); err != nil {
+		t.Fatal(err)
```
bhandras, line 203: You could use `require.NoError` here.

routing/router.go
```diff
@@ -538,6 +546,30 @@ func (r *ChannelRouter) Start() error {
...
+	toKeep := make(map[uint64]struct{})
+	for _, p := range payments {
+		payment, err := r.cfg.Control.FetchPayment(
+			p.Info.PaymentHash,
+		)
+		if err != nil {
+			return err
+		}
+
+		for _, a := range payment.HTLCs {
```
bhandras, line 563: It seems to me that `p` which is an `InFlightPayment` has an `Attempts` member which holds the same info? (Could be I'm wrong here lacking deeper knowledge of the underlying code.)

bhandras (changes requested):

routing/router.go
halseth: I deleted that field in the first commit, since it was not used for anything else :)

htlcswitch/payment_result.go
halseth: Nice, didn't know about that! Done 👍

htlcswitch/payment_result_test.go
halseth: I don't know if that can be used to match on the exact error?

htlcswitch/payment_result.go
```diff
@@ -258,3 +258,44 @@ func fetchResult(tx kvdb.RTx, pid uint64) (*networkResult, error) {
...
+				continue
+			}
+
+			numDeleted++
+			if err := cursor.Delete(); err != nil {
+				return err
+			}
+		}
+
+		if numDeleted == 0 {
```
bhandras, line 292: nit: could be
```
if numDeleted > 0 {
        log.Infof("Removed %d...)
}
```

htlcswitch/payment_result_test.go
bhandras: this should do it: `require.Error(t, ErrPaymentIDNotFound, err)`

routing/router.go
bhandras: Need to upgrade my memory to be able the handle three consecutive commits :)

bhandras (approved):
A few nits, otherwise LGTM :bulb:

htlcswitch/payment_result.go
halseth: Done!

htlcswitch/payment_result_test.go
halseth: Couldn't get it to work. Looks like `require.Error` only checks that it is non-nil.

htlcswitch/payment_result.go
```diff
@@ -258,3 +258,42 @@ func fetchResult(tx kvdb.RTx, pid uint64) (*networkResult, error) {
...
+		var numDeleted int
+		cursor := networkResults.ReadWriteCursor()
+		for k, _ := cursor.First(); k != nil; k, _ = cursor.Next() {
+			pid := binary.BigEndian.Uint64(k)
+			if _, ok := keep[pid]; ok {
+				continue
+			}
+
+			numDeleted++
+			if err := cursor.Delete(); err != nil {
```
cfromknecht, line 287: didn't we have an issue in the past where deleting while iterating caused issues? maybe i'm misremembering the issue tho...

cfromknecht (approved):
LGTM, only one comment about deleting under a cursor. i think this is okay, just wanted to bring it up just in case

htlcswitch/payment_result.go
halseth: You're right, this is still a bug: https://github.com/etcd-io/bbolt/issues/146

When you mention it I remember we had to work around it in `btcwallet`

halseth: Reverted back to using the original `ForEach`: https://github.com/lightningnetwork/lnd/pull/3131/commits/1ff47c65e1c789f7468a1e5dc172d4cda5dfc270

cc @bhandras

PR #3134, opened 2019-05-29
Add gRPC performance Prometheus exporter.

valentinewallace:
Optionally export lnd gRPC performance metrics for use by a Prometheus server through gRPC interceptors.

If the user compiles `lnd` with the `monitoring` tag, the exporter will automatically start listening for Prometheus requests when `lnd` starts up. The user has the option of specifying an address for the exporter to listen on, with the default being `localhost:8989`.


monitoring/config.go
```diff
@@ -0,0 +1,9 @@
+package monitoring
+
+// PrometheusConfig is the set of configuration data that specifies
```
Roasbeef: Alternatively, this can live in `lncfg` and have two versions: one with w/ the build tag, and one w/o it. This way, if the user doesn't have this tag enabled, then it won't show up in `lnd -h`.

monitoring/monitoring_on.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Enabled signifies whether the monitoring tag was specified when building lnd
+// and whether to automatically export Prometheus metrics.
+const Enabled = true
+
+// Start registers gRPC metrics and launches the Prometheus exporter on the
+// specified address.
+func Start(grpcServer *grpc.Server, listenAddr string) {
```
Roasbeef: Naming suggestion: `ExportPrometheusMetrics`.

rpcserver.go
```diff
@@ -531,17 +534,33 @@ func newRPCServer(s *server, macService *macaroons.Service,
 
 	// If macaroons aren't disabled (a non-nil service), then we'll set up
 	// our set of interceptors which will allow us handle the macaroon
-	// authentication in a single location .
-	if macService != nil {
-		unaryInterceptor := grpc.UnaryInterceptor(
+	// authentication in a single location. If monitoring is enabled, we'll
+	// also set up interceptors for the Prometheus exporter. If both are
+	// enabled, we'll chain the interceptors together.
+	var unaryInterceptor grpc.UnaryServerInterceptor
```
Roasbeef: Can be collapsed under a single `var` statement.

rpcserver.go
```diff
@@ -531,17 +534,33 @@ func newRPCServer(s *server, macService *macaroons.Service,
...
 		)
-		streamInterceptor := grpc.StreamInterceptor(
+		streamInterceptor = grpc_middleware.ChainStreamServer(
 			macService.StreamServerInterceptor(permissions),
+			grpc_prometheus.StreamServerInterceptor,
 		)
+	case macService != nil:
+		unaryInterceptor = macService.UnaryServerInterceptor(permissions)
+		streamInterceptor = macService.StreamServerInterceptor(permissions)
+	case monitoring.Enabled:
```
Roasbeef: Rather than check this manually here independent of the build tag, I think this can be melded in with the current `monitoring.Start` method.

monitoring/config.go
valentinewallace: IIUC, fixed.

rpcserver.go
valentinewallace: Hm, server options need to be set before any services can register with the grpc server. Lmk if my solution works.

monitoring/monitoring_on.go
```diff
@@ -0,0 +1,67 @@
...
+// (if enabled) and Prometheus monitoring.
+func GetServerOpts(macService *macaroons.Service, perms map[string][]bakery.Op,
+	serverOpts []grpc.ServerOption) []grpc.ServerOption {
+
+	var (
+		unaryInterceptor  grpc.UnaryServerInterceptor
+		streamInterceptor grpc.StreamServerInterceptor
+	)
+
+	// If macaroons aren't disabled (a non-nil service), then we'll set up
```
Roasbeef: I don't think we need to force the monitoring package to be responsible for setting up the macaroon interceptors. Instead, can't we just through through solely the `serverOpts` then chain them as normal? Applies to the alternate build tag as well.

lncfg/monitoring_off.go
```diff
@@ -0,0 +1,9 @@
+// !build monitoring
+
+package lncfg
+
+// PrometheusConfig configures the Prometheus exporter when monitoring is
+// enabled. Monitoring is currently disabled.
+type PrometheusConfig struct {
+	ListenAddr string `long:"listenaddr" description:"the interface we should listen on for prometheus"`
```
Roasbeef: If this is the version with prometheus off, then why do we need to expose this `ListenAddr` field?

monitoring/monitoring_on.go
```diff
@@ -0,0 +1,67 @@
...
+	// authentication in a single location. If monitoring is enabled, we'll
+	// also set up interceptors for the Prometheus exporter. If both are
+	// enabled, we'll chain the interceptors together.
+	switch macService {
+	case nil:
+		unaryInterceptor = grpc_prometheus.UnaryServerInterceptor
+		streamInterceptor = grpc_prometheus.StreamServerInterceptor
+	default:
+		unaryInterceptor = grpc_middleware.ChainUnaryServer(
+			macService.UnaryServerInterceptor(perms),
```
Roasbeef: This file fails to compile for me.

lncfg/monitoring_off.go
cfromknecht: i think because [this](https://github.com/lightningnetwork/lnd/pull/3134/files#diff-01ada5680aaa18fc79cbf78b2d55a37bR607) will fail to compile

@valentinewallace we can get around this by defining a function, which when `_on` returns the `ListenAddr` and when `_off` returns `""`. that will allow us to vary the fields without breaking compilation

fwiw, most CLI args take `listen` as the arg as opposed to `listenaddr`

monitoring/monitoring_on.go
cfromknecht: if we gather the unary and stream interceptors separately, we can join them all at once in the rpcserver via: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/chain.go#L175

monitoring/monitoring_on.go
```diff
@@ -0,0 +1,67 @@
...
+		grpc.UnaryInterceptor(unaryInterceptor),
+		grpc.StreamInterceptor(streamInterceptor),
+	)
+}
+
+// ExportPrometheusMetrics sets server options, registers gRPC metrics and
+// launches the Prometheus exporter on the specified address.
+func ExportPrometheusMetrics(grpcServer *grpc.Server, listenAddr string) {
+	if listenAddr == "" {
+		listenAddr = "localhost:8989"
```
cfromknecht: would think that this should be set where the other defaults are handled?

if we define an `lncfg.DefaultPrometheus()` for both on and off, one can be empty and the other can contain this value

lncfg/monitoring_off.go
valentinewallace: Solved by passing the entire Prometheus config to `ExportPrometheusMetrics`.

monitoring/monitoring_on.go
valentinewallace: IIUC fixed :grimacing: I now gather the unary/stream interceptors separately, and then join them in the rpcserver.

monitoring/monitoring_on.go
```diff
@@ -0,0 +1,46 @@
...
+	"sync"
+
+	"google.golang.org/grpc"
+
+	"github.com/grpc-ecosystem/go-grpc-prometheus"
+	"github.com/lightningnetwork/lnd/lncfg"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Enabled signifies whether the monitoring tag was specified when building lnd
```
Roasbeef: I think this can be lifted into the config above, so it's possible to build with the tag, but not always have the extra HTTP server listening.

monitoring/monitoring_on.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	return unaryInterceptors, streamInterceptors
+}
+
+// ExportPrometheusMetrics sets server options, registers gRPC metrics and
+// launches the Prometheus exporter on the specified address.
+func ExportPrometheusMetrics(grpcServer *grpc.Server, cfg lncfg.Prometheus) {
+	started.Do(func() {
+		grpc_prometheus.Register(grpcServer)
+
```
Roasbeef: Should adds some logging here to users can be sure that things have actually activated.

rpcserver.go
```diff
@@ -531,19 +533,36 @@ func newRPCServer(s *server, macService *macaroons.Service,
...
+
+	// Get interceptors for Prometheus to gather gRPC performance metrics.
+	// If monitoring is disabled, GetPromInterceptors() will return empty
+	// slices.
+	promUnaryInterceptors, promStrmInterceptors := monitoring.GetPromInterceptors()
 
-		serverOpts = append(serverOpts,
-			unaryInterceptor, streamInterceptor,
+	// Concatenate the slices of unary and stream interceptors respectively.
+	unaryInterceptors := append(macUnaryInterceptors, promUnaryInterceptors...)
```
Roasbeef: 👍

lncfg/monitoring_on.go
```diff
@@ -0,0 +1,27 @@
...
+
+package lncfg
+
+// Prometheus is the set of configuration data that specifies the listening
+// address of the Prometheus exporter.
+type Prometheus struct {
+	// Listen is the listening address that we should use to allow the main
+	// Prometheus server to scrape our metrics.
+	Listen string `long:"listen" description:"the interface we should listen on for Prometheus"`
+	Enable bool   `long:"enable" description:"enable Prometheus exporting of lnd gRPC performance metrics."`
```
Roasbeef: Missing a `godoc` comment, also new line above.

monitoring/monitoring_off.go
```diff
@@ -0,0 +1,20 @@
...
+// GetPromInterceptors returns the set of interceptors for Prometheus
+// monitoring if monitoring is enabled, else empty slices. Monitoring is
+// currently disabled.
+func GetPromInterceptors() ([]grpc.UnaryServerInterceptor, []grpc.StreamServerInterceptor) {
+	return []grpc.UnaryServerInterceptor{}, []grpc.StreamServerInterceptor{}
+}
+
+// ExportPrometheusMetrics is required for lnd to compile so that Prometheus
+// metric exporting can be hidden behind a build tag.
+func ExportPrometheusMetrics(_ *grpc.Server, _ lncfg.Prometheus) {}
```
halseth: should this method possibly return an error if called on `!monitoring` build?

PR #3139, opened 2019-05-29
rpc: update updatechanpolicy to allow zero fees

### LINKED ISSUE #3138: Fee zero - is it possible to change in LND?
Now the LND has minimum fee rate as 1/1,000,000
If i try to use zero i get error "fee rate of 0 is too small, min fee rate is 1e-06" (v0.6.1-beta)

But other nodes already has fee as 0 / 0: [ACINQ channel example](https://1ml.com/channel/635796996973068288) (and capacity 1 BTC too! ;-)

It turns out that all nodes based on LND software will be noncompetitive with Eclair software... Is it possible to allow fee rate & base fee as **0**?

Roasbeef:
Fixes #3138.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -4429,15 +4429,17 @@ func (r *rpcServer) UpdateChannelPolicy(ctx context.Context,
 		return nil, fmt.Errorf("unknown scope: %v", scope)
 	}
 
-	// As a sanity check, we'll ensure that the passed fee rate is below
-	// 1e-6, or the lowest allowed fee rate, and that the passed timelock
-	// is large enough.
-	if req.FeeRate < minFeeRate {
+	switch {
+	// As a sanity check, if the fee isn't zero, we'll ensure that the
+	// passed fee rate is below 1e-6, or the lowest allowed non-zero fee
+	// rate expressible within the protocol.
+	case req.FeeRate != 0 && req.FeeRate < minFeeRate:
 		return nil, fmt.Errorf("fee rate of %v is too small, min fee "+
 			"rate is %v", req.FeeRate, minFeeRate)
-	}
 
-	if req.TimeLockDelta < minTimeLockDelta {
+	// We'll also ensure that the user isn't setting a CLTV delta that
+	// won't give outgoing HTLCs enough time to fully resolve if needed.
+	case req.TimeLockDelta < minTimeLockDelta:
 		return nil, fmt.Errorf("time lock delta of %v is too small, "+
 			"minimum supported is %v", req.TimeLockDelta,
 			minTimeLockDelta)


rpcserver.go
```diff
@@ -4429,15 +4429,17 @@ func (r *rpcServer) UpdateChannelPolicy(ctx context.Context,
 		return nil, fmt.Errorf("unknown scope: %v", scope)
 	}
 
+	switch {
 	// As a sanity check, we'll ensure that the passed fee rate is below
 	// 1e-6, or the lowest allowed fee rate, and that the passed timelock
 	// is large enough.
-	if req.FeeRate < minFeeRate {
+	case req.FeeRate != 0 && req.FeeRate < minFeeRate:
```
cfromknecht: comment looks out of date?

Roasbeef:
PTAL @cfromknecht

rpcserver.go
cfromknecht: kek should have been more clear that it also says "we'll ensure that the passed fee rate is below 1e-6", instead of above the minimum.

ZapUser77:
Doesn't a fee of 0, 0 break the pathfinding algo?

cfromknecht:
@ZapUser77 nope, fees aren’t the only consideration in determining distance. That will be even more true with #2802

ZapUser77:
I remember running into this specific problem.  The EdgeWeight can return zero if fee = zero, because INT64 is used instead of float:

func edgeWeight(lockedAmt lnwire.MilliSatoshi, fee lnwire.MilliSatoshi,
	timeLockDelta uint16) int64 {
	// timeLockPenalty is the penalty for the time lock delta of this channel.
	// It is controlled by RiskFactorBillionths and scales proportional
	// to the amount that will pass through channel. Rationale is that it if
	// a twice as large amount gets locked up, it is twice as bad.
	timeLockPenalty := int64(lockedAmt) * int64(timeLockDelta) *
		RiskFactorBillionths / 1000000000

	return int64(fee) + timeLockPenalty
}

If TimelockPenalty results in number less than one, the return would be fees only... which was almost always.
This resulted in an infinite loop between two nodes if the fees were zero/zero, regardless of the Timelock.
I suggested changing several variable types.

Though, I think that was addressed by changing:
if tempDist >= distance[fromVertex].dist {
			return
}
From ">" to ">="

LNBIG-COM:
@alexbosworth wrote to me in LN chat that: "If you set your fees to zero programmatically you need to restart your node otherwise forwarding will break"
@Roasbeef , can you confirm or deny this? I changed fees through gRPC API. Should i restart LND daemons after setting zero fees in channels? And if yes, why?

Roasbeef:
Hi,

That's correct, we became aware of this issue last week. We have a fix, and
it should be implemented before our scheduled 0.8 release. When the fix is
available in master, we'll comment on this issue.

gcomte:
Hi @Roasbeef 

So I'd assume the fix is available in master as of now?

cfromknecht:
@gcomte given the merge dates this looks like it shipped in 0.7

Issue #2977, opened 2019-04-18
Feature Request: Allow Spending Unconfirmed Funding Transaction Change Output to CPFP it with sendcoins


juscamarena:
### Background

Describe your issue here.

Currently have a channel open where I cannot spend the change output with sendcoins to increase fees for channel to confirm quicker. 

Would be nice to have a flag where all coins even unconfirmed can be spent.

Have not checked any of the build flags if I can do this manually.

wpaulino:
See https://github.com/lightningnetwork/lnd/pull/1746.

juscamarena:
Nice, didn't search for PRs, but issues.

Issue #609, opened 2018-01-15
[Request] Command to anchor funding transactions with extra fees


junderw:
Problem: Funded channel with small fee. Confirmation doesn't occur for weeks (some block explorers re-broadcast transactions too)... This command will allow a user to add more fee after the fact.

Ideal (Does this require change in protocol spec? idk):

1. Set sequence for all inputs in the funding transaction as < MAX - 1.
2. Fund normally.
3. Fee too low.
4. Run this command, and it will use the same inputs, and decrease the change output to increase fees. The partner will cooperate with signing the refund transaction for the same channel. (Will this fail? Does it break protocol spec?)

Less than ideal:

1. When possible ensure that all funding transactions will always have a change output.
2. Fund normally.
3. Fee too low.
4. Run this command which will calculate the missing fees from the funding transaction to match the given sat_per_byte, then it will create a transaction with that change output, and add the missing fees from the funding transaction on top of the dependent transaction. (causing CPFP)

- This will allow the user to speed up confirmation on funding transactions.
- This will allow the user to risk using lower confirmation targets since they can use this afterward when necessary.
- This will NOT allow the user to revert a funding transaction. (the output scripts must remain the same.)

junderw:
ACK CPFP approach for now.

RBF is kind of messy...

Roasbeef:
Agreed that the CPFP route is easier for now, and also has less edge cases. Rather than make it specific to funding transactions, I think we can instead make it fully generalized. So maybe a command like: `lncli feebump cpfp --outpoints=op1, op2, op3 --fee_rate=X`. Notice that it takes a list of outpoints, so one could even anchor down a series of transactions with a single child. 

One Q with this method would be how to interpret the `--fee_rate` param. One interoperation would be the new _cumulative_ package should have that particular fee rate. Another would be that this should just be the feerate of the child transaction, so it's up to the caller to properly compute this value.

jamesob:
Yup, agree that a more general `feebump` mechanism that'd allow multiple parent inputs seems very useful. I'll go that route.

rashlab:
Great, and important! any estimate as to when such feature could be available, and what could be a temporary solution?? (my funding transaction is stuck, miners "low priority"...). Thanks,

jamesob:
@rashlab I'm working on this at the moment, but I'm still getting to know lnd so I might be kinda slow. I'd say I'll have a PR up in the next two weeks?

jamesob:
[...]

PR #3140, opened 2019-05-29
sweep+rpc: add support to bump fee of inputs/transactions

wpaulino:
In this PR, we add support to the `UtxoSweeper` to bump the fee of an input it's currently attempting to sweep and to bump the fee of a transaction through CPFP. We also add a new RPC and a corresponding `lncli` command to expose the functionality to users.

This RPC takes a different approach than `bitcoind`'s bumpfee command. `lnd` has a central batching engine, known as the `UtxoSweeper`, in which inputs with similar fee rates are batched together to save on transaction fees. Due to this, we cannot rely on bumping the fee on a specific transaction, since transactions can change at any point with the addition of new inputs. The list of inputs that currently exist within the `UtxoSweeper` can be retrieved through the `PendingSweeps` RPC.

When bumping the fee of an input that currently exists within the `UtxoSweeper`, a higher fee transaction will be created that replaces the lower fee transaction through the Replace-By-Fee (RBF) policy.

This RPC also serves useful when wanting to perform a Child-Pays-For-Parent (CPFP), where the child transaction pays for its parent's fee. This can be done by specifying an outpoint within the low fee transaction that is under the control of the wallet.

Note that this RPC currently doesn't perform any validation checks on the fee preference being provided. For now, the responsibility of ensuring that the new fee preference is sufficient is delegated to the user.

Depends on https://github.com/lightningnetwork/lnd/pull/3089.

Fixes https://github.com/lightningnetwork/lnd/issues/609.
Fixes https://github.com/lightningnetwork/lnd/issues/2977.


sweep/sweeper.go
```diff
@@ -949,6 +981,135 @@ func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) []*PendingIn
...
+
+// handleRBFBump handles a bump fee request by simply updating the inputs fee
+// preference. Currently, no validation is done on the new fee preference to
+// ensure it will properly create a replacement transaction.
+//
+// TODO(wilmer):
+//   * Validate fee preference to ensure we'll create a valid replacement
+//     transaction to allow the new fee rate to propagate throughout the
+//     network.
+//   * Ensure we don't combine this input with any other unconfirmed inputs that
```
Roasbeef: Are there any situations that exist today, taking into account our current use cases, wherein improper usage of this new RPC can cause this to happen?

sweep/sweeper.go
```diff
@@ -949,6 +981,135 @@ func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) []*PendingIn
...
+//
+// TODO(wilmer): Validate fee preference to ensure we're creating a child
+// transaction with a high enough fee rate to cover its parents.
+func (s *UtxoSweeper) handleCPFPBump(req *bumpFeeReq,
+	bestHeight int32) (chan Result, error) {
+
+	resultChan := make(chan Result, 1)
+	inputReq := &sweepInputMessage{
+		input:         req.input,
+		feePreference: req.feePreference,
```
Roasbeef: So it's the job of the caller to compute the proper fee rate that'll cause the entire package to be bumped by a satisfactory amount?

wpaulino: Yes. I'm not a fan of this, but from our previous discussions offline it seemed like this was the approach to take for now. W should be able to compute this quite easily for full nodes, but things get a bit trickier with light clients.

sweep/sweeper.go
wpaulino: It can happen if the sweeper increases the fee rate of an input and a CPFP is made where the input being swept is within the same bucket as the increased fee rate input.

cmd/lncli/commands.go
```diff
@@ -302,6 +302,117 @@ func sendCoins(ctx *cli.Context) error {
...
+		return errors.New("outpoint should be of the form txid:index")
+	}
+	txid := parts[0]
+	if hex.DecodedLen(len(txid)) != chainhash.HashSize {
+		return fmt.Errorf("invalid hex-encoded txid %v", txid)
+	}
+	outputIndex, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return fmt.Errorf("invalid output index: %v", err)
+	}
```
joostjager: Extract to helper func (if not already present?)

lnrpc/rpc.proto
```diff
@@ -250,6 +250,40 @@ service Lightning {
...
+
+    This RPC also serves useful when wanting to perform a Child-Pays-For-Parent
+    (CPFP), where the child transaction pays for its parent's fee. This can be
+    done by specifying an outpoint within the low fee transaction that is under
+    the control of the wallet.
+
+    The fee preference can be expressed either as a specific fee rate or a delta
+    of blocks in which the output should be swept on-chain within. If a fee
+    preference is not explicitly specified, then a fee rate corresponding to a
+    confirmation target of 6 blocks will be used.
```
joostjager: I think we shouldn't use defaults on the rpc level and make a fee preference required. Force callers to think about what they want. Also that default can't really be changed anymore, because it may break assumptions.

rpcserver.go
```diff
@@ -174,6 +174,10 @@ var (
 			Entity: "onchain",
 			Action: "write",
 		}},
+		"/lnrpc.Lightning/BumpFee": {{
+			Entity: "onchain",
+			Action: "write",
+		}},
```
joostjager: We are gradually working towards moving rpc calls into sub servers and split up `lnd`. Should we still add something new to the main rpc server?

rpcserver.go
```diff
@@ -970,6 +974,111 @@ func (r *rpcServer) EstimateFee(ctx context.Context,
 	return resp, nil
 }
 
+// unmarshallOutPoint converts an outpoint from its lnrpc type to its canonical
+// type.
+func unmarshallOutPoint(op *lnrpc.OutPoint) (*wire.OutPoint, error) {
```
joostjager: Wasn't this function yet present for another rpc that parses outpoints?

rpcserver.go
```diff
@@ -970,6 +974,111 @@ func (r *rpcServer) EstimateFee(ctx context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Construct the request's fee preference.
+	satPerKw := lnwallet.SatPerKVByte(in.SatPerByte * 1000).FeePerKWeight()
+	feePreference := sweep.FeePreference{
+		ConfTarget: uint32(in.TargetConf),
+		FeeRate:    satPerKw,
+	}
```
joostjager: Validation here in case both or none are set?

rpcserver.go
```diff
@@ -970,6 +974,111 @@ func (r *rpcServer) EstimateFee(ctx context.Context,
...
+	// Before attempting to bump the fee of the output, we'll first attempt
+	// to retrieve it from our wallet.
+	txOut, err := r.server.cc.wallet.FetchInputInfo(op)
+	switch {
+	// It's possible for ErrNotMine to be returned if the output exists
+	// within the wallet, but it is not an output for which we have control
+	// of, e.g., CommitmentTimeLock, CommitmentNoDelay outputs, so we'll
+	// still attempt to send it over to the UtxoSweeper to see if it's
+	// currently attempting to sweep it.
+	case err == lnwallet.ErrNotMine:
```
joostjager: How does this compare to first checking if the sweeper knows it (now that it exposes its inputs) and only if it doesn't, check the wallet? Are there other wallet inputs that we have no control over but are also not in the sweeper (utxonursery maybe?)

sweep/sweeper.go
```diff
@@ -948,6 +981,135 @@ func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) []*PendingIn
...
+func (s *UtxoSweeper) handleBumpFeeReq(req *bumpFeeReq,
+	bestHeight int32) (chan Result, error) {
+
+	// If the UtxoSweeper is already trying to sweep this input, then we can
+	// simply just increase its fee rate. This will allow the input to be
+	// batched with others which also have a similar fee rate, creating a
+	// higher fee rate transaction that replaces the original input's
+	// sweeping transaction.
+	if pendingInput, ok := s.pendingInputs[*req.input.OutPoint()]; ok {
+		return s.handleRBFBump(req, pendingInput, bestHeight)
```
joostjager: It feels like the `Input` struct isn't the best fit for rbf bump. Maybe `RBFBump` should be exposed on the sweeper separately. In the `rpcserver` we already determined whether it was rbf or cpfp?

wpaulino: Agreed. Having one call to bump fees is only necessary at the RPC level, we can abstract things better within the sweeper.

rpcserver.go
wpaulino: > How does this compare to first checking if the sweeper knows it (now that it exposes its inputs) and only if it doesn't, check the wallet?

Same thing really, I opted to do it this way to prevent more code within the sweeper.

> Are there other wallet inputs that we have no control over but are also not in the sweeper (utxonursery maybe?)

The wallet only has the keys for addresses it's derived. If the input exists within the nursery, then that would mean it's not ready to be swept yet, no? So there'd be no point in attempting to bump its fee.

rpcserver.go
wpaulino: Validation is done within the `UtxoSweeper.BumpFee` call.

rpcserver.go
wpaulino: I don't think so? No other RPCs take `lnrpc.OutPoint` as an input, but they return them.

rpcserver.go
wpaulino: I wouldn't be opposed to doing so - seems like this should apply to the `pendingsweeps` RPC as well? Would we want to add these to `WalletKit` or have a new `Sweeper` sub-server?

cmd/lncli/commands.go
wpaulino: Can do so, but this parsing is somewhat specific to this command as it assumes that the outpoint will be the first argument provided.

sweep/sweeper.go
wpaulino: Done. `UtxoSweeper.BumpFee` is now only meant to be used for inputs already within the sweeper.  To handle CPFPs, we can just go through the usual `UtxoSweeper.SweepInput` path.

rpcserver.go
wpaulino: Updated so that it checks the sweeper for the input first.

lnrpc/rpc.proto
wpaulino: Done.

wpaulino:
PTAL @joostjager.

cmd/lncli/commands.go
joostjager: Yes, so I mean not the part that takes the `outpoint` string from the arg list, but converting that string into `lnrpc.OutPoint`

rpcserver.go
joostjager: I would think `WalletKit`. @Roasbeef probably has an idea about it too

rpcserver.go
joostjager: `CloseChannel` takes an outpoint. But the bigger question there is why we have a `ChannelPoint` and an `OutPoint`.

rpcserver.go
joostjager: Just checking that exactly one of those fields is set could may be done in a `sweeper.NewFeePreference(confTarget, feeRate)` constructor. It simplifies consumers of this struct, because they don't all need to do this check.

sweep/sweeper.go
```diff
@@ -958,6 +990,97 @@ func (s *UtxoSweeper) handlePendingSweepsReq(req *pendingSweepsReq) pendingInput
...
+	log.Debugf("Updating fee preference for %v from %v to %v",
+		*input.OutPoint(), input.FeePreference, req.feePreference)
+
+	input.FeePreference = req.feePreference
+
+	// We'll reset the input's publish height to the current so that a new
+	// transaction can be created that replaces the transaction currently
+	// spending the input. We only do this for inputs that have been
+	// broadcast at least once to ensure we don't spend an input before its
+	// maturity height.
```
joostjager: I don't think the sweeper looks at maturity height at all. It expects any input offered to be no longer locked?

wpaulino: No, but I figured still having the check here is nice in case it's forgotten if we ever decide to offer time-locked inputs to the sweeper.

rpcserver.go
wpaulino: Some other RPCs rely on a fee preference not being set to choose the default. Adding a constructor that applies additional validation checks would lead to some confusion IMO. Either we should unify the behavior of `FeePreference` all around, or we introduce a new `SweepPreference` type that applies the validation needed for sweeps only.

rpcserver.go
wpaulino: Seems to just be a legacy thing since it was introduced before `OutPoint`.

rpcserver.go
wpaulino: Moved to `WalletKit` after offline discussion.

cmd/lncli/commands.go
wpaulino: Done.

wpaulino:
PTAL @joostjager.

Roasbeef:
Can be rebased now that #3089 is in.

sweep/sweeper.go
joostjager: Maybe then add a comment that this is not yet the case. Like it is, it may be confusing and make people think we already await locked outputs.

wpaulino:
Rebased!

rpcserver.go
Roasbeef: Yeah `ChannelPoint` existed first, then `OutPoint` was added later, IIRC, around the same time that `ListUnspent` was added.

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -409,3 +416,107 @@ func (w *WalletKit) PendingSweeps(ctx context.Context,
...
+	switch {
+	case txscript.IsPayToWitnessPubKeyHash(txOut.PkScript):
+		witnessType = input.WitnessKeyHash
+	case txscript.IsPayToScriptHash(txOut.PkScript):
+		witnessType = input.NestedWitnessKeyHash
+	default:
+		return nil, fmt.Errorf("unknown input witness %v", op)
+	}
+
+	signDesc := &input.SignDescriptor{
```
Roasbeef: Don't we need to provide a more populated sign descriptor? At the very least, the pubkey information.

cmd/lncli/types.go
```diff
@@ -1,14 +1,18 @@
...
+	"strconv"
+	"strings"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/lightningnetwork/lnd/lnrpc"
 )
 
 // OutPoint displays an outpoint string in the form "<txid>:<output-index>".
-type OutPoint string
+type OutPoint = string
```
Roasbeef: ?

lnrpc/walletrpc/walletkit_server.go
wpaulino: The only field required seems to be the output. This is also what `CraftSweepAllTx` does.

lntest/node.go
```diff
@@ -250,9 +250,10 @@ type HarnessNode struct {
 
 	invoicesrpc.InvoicesClient
 
-	// RouterClient cannot be embedded, because a name collision would occur
-	// on the main rpc SendPayment.
-	RouterClient routerrpc.RouterClient
+	// RouterClient and WalletKitClient cannot be embedded, because a name
+	// collision would occur with LightningClient.
```
joostjager: Is this also true for `WalletKitClient`? `LightningClient` and `RouterClient` both have a `SendPayment` call, that is where the collision comes from.

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -409,3 +416,107 @@ func (w *WalletKit) PendingSweeps(ctx context.Context,
...
+	// We'll retrieve the set of inputs the UtxoSweeper is currently
+	// attempting to sweep. If the input is found within it, then we can
+	// simply bump its fee, which will result in a replacement transaction
+	// (RBF) being broadcast.
+	pendingInputs, err := w.cfg.Sweeper.PendingInputs()
+	if err != nil {
+		return nil, err
+	}
+
+	if _, ok := pendingInputs[*op]; ok {
```
joostjager: Is this necessary or can you rely on `ErrNotMine` returned from `BumpFee` ?

sweep/sweeper.go
```diff
@@ -979,6 +1021,100 @@ func (s *UtxoSweeper) handlePendingSweepsReq(
...
+
+	// If the UtxoSweeper is already trying to sweep this input, then we can
+	// simply just increase its fee rate. This will allow the input to be
+	// batched with others which also have a similar fee rate, creating a
+	// higher fee rate transaction that replaces the original input's
+	// sweeping transaction.
+	pendingInput, ok := s.pendingInputs[req.input]
+	if !ok {
+		return nil, lnwallet.ErrNotMine
+	}
```
joostjager: Is it not too verbose to have an additional function just for this check? (vs integrating with `handleRBFBump`)

sweep/sweeper.go
```diff
@@ -979,6 +1021,100 @@ func (s *UtxoSweeper) handlePendingSweepsReq(
...
+//
+// TODO(wilmer):
+//   * Validate fee preference to ensure we'll create a valid replacement
+//     transaction to allow the new fee rate to propagate throughout the
+//     network.
+//   * Ensure we don't combine this input with any other unconfirmed inputs that
+//     did not exist in the original sweep transaction, resulting in an invalid
+//     replacement transaction.
+func (s *UtxoSweeper) handleRBFBump(req *bumpFeeReq, input *pendingInput,
+	bestHeight int32) (chan Result, error) {
```
joostjager: Why return a `chan Result`? It is not used other than in the test it seems and maybe there you can use the original input registration result channel? (yagni?)

sweep/sweeper.go
```diff
@@ -979,6 +1021,100 @@ func (s *UtxoSweeper) handlePendingSweepsReq(
...
+// transaction (RBF) of the original sweeping transaction, if any.
+//
+// NOTE: This currently doesn't do any fee rate validation to ensure that a bump
+// is actually successful. The responsibility of doing so should be handled by
+// the caller.
+func (s *UtxoSweeper) BumpFee(input wire.OutPoint,
+	feePreference FeePreference) (chan Result, error) {
+
+	// Ensure the client provided a sane fee preference.
+	if _, err := s.feeRateForPreference(feePreference); err != nil {
```
joostjager: Still feels a bit strange that we validate against a moving fee rate here.

lntest/node.go
wpaulino: Yes, the collision happens with `EstimateFee`.

sweep/sweeper.go
wpaulino: I was thinking of the case where a new caller wants to bump the fee of the input, but they weren't the one who submitted the input request, so they wouldn't be notified of when the input gets swept.

sweep/sweeper.go
wpaulino: The fee preference won't always map to a moving fee rate. I see this more as just a sanity check to ensure it respects the min and max fee rate we'll allow.

sweep/sweeper.go

lnrpc/walletrpc/walletkit_server.go
wpaulino: Done.

wpaulino:
PTAL @Roasbeef @joostjager.

lntest/node.go
joostjager: Ah, ok.

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -409,3 +416,106 @@ func (w *WalletKit) PendingSweeps(ctx context.Context,
...
+	// If it is currently attempting to sweep the input, then it'll simply
+	// bump its fee, which will result in a replacement transaction (RBF)
+	// being broadcast. If it is not aware of the input however,
+	// lnwallet.ErrNotMine is returned.
+	_, err = w.cfg.Sweeper.BumpFee(*op, feePreference)
+	switch err {
+	case nil:
+		return &BumpFeeResponse{}, nil
+	case lnwallet.ErrNotMine:
+		break
```
joostjager: `break` can be removed

sweep/sweeper.go
joostjager: Ok, I can see the use case. I think I would have added it only when actually needed.

sweep/sweeper.go
joostjager: I think I made the point before, but what I find strange is that we may reject an input here when the fee estimate (let's say X) exceeds the max. While if the fee estimate was below max, we would have added and when - later - the fee estimate would increase to X we would just wait until it goes down again. Seems inconsistent, but no blocker.

wpaulino: I understand your point, but I see this more as a sanity check to ensure we don't ever see some kind of insane fee rate.

PR #3143, opened 2019-05-30
htlcswitch: pipeline settles to switch

### LINKED ISSUE #3069: htlcswitch: pipeline pre-images to the incoming link as soon as it's obtained
Today, once we receive a pre-image from the remote party, we don't immediately pipeline it to the incoming link. Instead, we'll wait until we receive a revocation from the remote party as that's when [we examine the commitments for updates that have been fully locked in which we can process](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L1762). [If we start to properly pipeline these pre-images](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L1583), then we'll be able to save some perceived latency of payments (1.5 RTT vs 0.5 RTT), as well as address some scenarios where lost messages may cause us to never settle the HTLC on the incoming link (#2045 ). 

# Steps To Completion 
- [ ] Right after we receive the settle, [send the pre-image to the switch using the proper circuit identifiers](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L1583). This achieves pipelining, but introduces other loose ends that need to be tied up. 

- [ ] With the above change, it's possible to send a duplicate settle to the incoming link. As implement atm, this would cause an error on the incoming link, making it reject that new update. Atm, we don't distinguish between a double settle error, and an invalid pre-image. We'll need to make concrete errors for both these cases to ensure we don't cause any force closes or link failures due to duplicate settles.

Crypt-iQ:
This PR makes the outgoing link pipeline the settle to the
switch as soon as it receives it. Previously, it would wait for a
revocation before sending it, which caused increased latency on
payments as well as possibly never settling on the incoming link.
A duplicate settle is still sent to the switch, but it is handled
gracefully. A new AckEventTicker was added to the switch which
acknowledges any pending settle / fail entries in an outgoing
link's fwd pkgs in batch. This was needed in order to reduce the
number of db txn's which would have been incurred by acking whenever
we receive a duplicate settle without batching.

Fixes #3069

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -1608,13 +1608,22 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 			return
 		}
 
-		// TODO(roasbeef): pipeline to switch
+		settlePacket := &htlcPacket{
+			outgoingChanID: l.ShortChanID(),
+			outgoingHTLCID: idx,
+			htlc: &lnwire.UpdateFulfillHTLC{
+				PaymentPreimage: pre,
+			},
+		}
 
 		// Add the newly discovered preimage to our growing list of
 		// uncommitted preimage. These will be written to the witness
 		// cache just before accepting the next commitment signature
 		// from the remote peer.
 		l.uncommittedPreimages = append(l.uncommittedPreimages, pre)
+
+		// Pipeline this settle, send it to the switch.
+		go l.forwardBatch(settlePacket)
 
 	case *lnwire.UpdateFailMalformedHTLC:
 		// Convert the failure type encoded within the HTLC fail
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -232,10 +232,10 @@ func TestChannelLinkSingleHopPayment(t *testing.T) {
 		t.Fatalf("unable to make the payment: %v", err)
 	}
 
-	// Wait for Bob to receive the revocation.
+	// Wait for Alice to receive the revocation.
 	//
 	// TODO(roasbeef); replace with select over returned err chan
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(2 * time.Second)
 
 	// Check that alice invoice was settled and bandwidth of HTLC
 	// links was changed.
@@ -494,8 +494,8 @@ func testChannelLinkMultiHopPayment(t *testing.T,
 		t.Fatalf("unable to send payment: %v", err)
 	}
 
-	// Wait for Bob to receive the revocation.
-	time.Sleep(100 * time.Millisecond)
+	// Wait for Alice and Bob's second link to receive the revocation.
+	time.Sleep(2 * time.Second)
 
 	// Check that Carol invoice was settled and bandwidth of HTLC
 	// links were changed.
@@ -3977,7 +3977,8 @@ func TestChannelLinkAcceptOverpay(t *testing.T) {
 		t.Fatalf("unable to send payment: %v", err)
 	}
 
-	time.Sleep(100 * time.Millisecond)
+	// Wait for Alice and Bob's second link to receive the revocation.
+	time.Sleep(2 * time.Second)
 
 	// Even though we sent 2x what was asked for, Carol should still have
 	// accepted the payment and marked it as settled.
@@ -5801,7 +5802,12 @@ func TestChannelLinkHoldInvoiceSettle(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Wait for Bob to receive the revocation.
+	// Wait for Alice to receive the revocation. This is needed
+	// because the settles are pipelined to the switch and otherwise
+	// the bandwidth won't be updated by the time Alice receives a
+	// response here.
+	time.Sleep(2 * time.Second)
+
 	if ctx.startBandwidthAlice-ctx.amount !=
 		ctx.n.aliceChannelLink.Bandwidth() {
 
--- a/htlcswitch/switch.go
+++ b/htlcswitch/switch.go
@@ -30,6 +30,10 @@ const (
 	// DefaultLogInterval is the duration between attempts to log statistics
 	// about forwarding events.
 	DefaultLogInterval = 10 * time.Second
+
+	// DefaultAckInterval is the duration between attempts to ack any settle
+	// fails in a forwarding package.
+	DefaultAckInterval = 15 * time.Second
 )
 
 var (
@@ -159,6 +163,10 @@ type Config struct {
 	// aggregate stats about it's forwarding during the last interval.
 	LogEventTicker ticker.Ticker
 
+	// AckEventTicker is a signal instructing the htlcswitch to ack any settle
+	// fails in forwarding packages.
+	AckEventTicker ticker.Ticker
+
 	// NotifyActiveChannel and NotifyInactiveChannel allow the link to tell
 	// the ChannelNotifier when channels become active and inactive.
 	NotifyActiveChannel   func(wire.OutPoint)
@@ -259,6 +267,11 @@ type Switch struct {
 	// active ChainNotifier instance. This will be used to retrieve the
 	// lastest height of the chain.
 	blockEpochStream *chainntnfs.BlockEpochEvent
+
+	// pendingSettleFails is the set of settle/fail entries that we need to
+	// ack in the forwarding package of the outgoing link. This was added to
+	// make pipelining settles more efficient.
+	pendingSettleFails []channeldb.SettleFailRef
 }
 
 // New creates the new instance of htlc switch.
@@ -1347,11 +1360,10 @@ func (s *Switch) closeCircuit(pkt *htlcPacket) (*PaymentCircuit, error) {
 			pkt.outgoingHTLCID)
 		log.Error(err)
 
-		// TODO(conner): ack settle/fail
 		if pkt.destRef != nil {
-			if err := s.ackSettleFail(*pkt.destRef); err != nil {
-				return nil, err
-			}
+			// Add this SettleFailRef to the set of pending settle/fail entries
+			// awaiting acknowledgement.
+			s.pendingSettleFails = append(s.pendingSettleFails, *pkt.destRef)
 		}
 
 		return nil, err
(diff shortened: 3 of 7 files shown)


htlcswitch/link.go
```diff
@@ -1590,7 +1590,16 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
...
+		settlePacket := &htlcPacket{
+			outgoingChanID: l.ShortChanID(),
+			outgoingHTLCID: idx,
+			htlc: &lnwire.UpdateFulfillHTLC{
+				PaymentPreimage: pre,
+			},
+		}
+
+		// Pipeline this settle, send it to the switch.
+		go l.forwardBatch(settlePacket)
```
cfromknecht: might suggest reordering this to pipeline after adding the preimage to the list of uncommitted preimages, that way if anything panics we will still write the preimage

Crypt-iQ: Done

Crypt-iQ:
I have been running flakehunter unit tester on the failed test in travis and it hasn't failed... so I think it's unrelated.  I have seen that error on my mac though (unrelated to this branch).

htlcswitch/switch.go
```diff
@@ -1717,6 +1738,28 @@ out:
...
+			if err := s.ackSettleFail(s.pendingSettleFails...); err != nil {
+				log.Errorf("Unable to ack batch of settle/fails: %v", err)
+			}
+
+			log.Debugf("Acked %d settle fails: %v", len(s.pendingSettleFails),
+				spew.Sdump(s.pendingSettleFails))
+
+			// Reset the pendingSettleFails buffer while keeping acquired
+			// memory.
+			s.pendingSettleFails = s.pendingSettleFails[:0]
```
cfromknecht: should we only clear this if the write was successful? don't think it's too big of a deal tho, it will eventually be retried on restart or reconnect

htlcswitch/switch.go
```diff
@@ -1717,6 +1738,28 @@ out:
 			totalSatSent += diffSatSent
 			totalSatRecv += diffSatRecv
 
+		// The ack ticker has fired so if we have any settle/fail entries
+		// for a forwarding package to ack, we will do so here in a batch
+		// db call.
+		case <-s.cfg.AckEventTicker.Ticks():
+			// If the current set is empty, pause the ticker.
+			if len(s.pendingSettleFails) == 0 {
```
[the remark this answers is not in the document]
cfromknecht: had considered this, but as is this will prevent tearing down and reinitializing the ticker under load, since o/w we would expect to tear it down after every tick

htlcswitch/switch.go
Crypt-iQ: Agree - good catch

htlcswitch/switch.go
```diff
@@ -1717,6 +1738,28 @@ out:
...
+				s.cfg.AckEventTicker.Pause()
+				continue
+			}
+
+			// Batch ack the settle/fail entries.
+			if err := s.ackSettleFail(s.pendingSettleFails...); err != nil {
+				log.Errorf("Unable to ack batch of settle/fails: %v", err)
+			}
+
+			log.Debugf("Acked %d settle fails: %v", len(s.pendingSettleFails),
```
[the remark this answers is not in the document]
Crypt-iQ: Yeah I think trace is a better option

Crypt-iQ:
@wpaulino I wasn't entirely sure what to do about the extra error logs - I could remove it if the packet has a `destRef` field?  What was once erroneous behavior is the norm so I think that'd make sense.  What do you think?

htlcswitch/link_test.go
```diff
@@ -5801,7 +5801,12 @@ func TestChannelLinkHoldInvoiceSettle(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Wait for Bob to receive the revocation.
+	// Wait for Alice to receive the revocation. This is needed
+	// because the settles are pipelined to the switch and otherwise
+	// the bandwidth won't be updated by the time Alice receives a
+	// response here.
+	time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
Crypt-iQ: I'm a little confused about how this would work.  We can feed ticks to the `AckEventTicker`, but that won't help since we need to actually wait for the revocation from Bob (the duplicate settle doesn't matter here) and there doesn't seem to be a straightforward way to be notified upon receiving a revocation.  We could loop and timeout after some timeout interval, but that's not so clean either...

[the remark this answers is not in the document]
Crypt-iQ: I think we could actually *consume* the `AckEventTicker` tick because then that would let us know a revocation has been received.  Something like:
```
ticker := ctx.n.aliceServer.htlcSwitch.cfg.AckEventTicker
select {
case <-ticker.Ticks():
case <-time.After(20 * time.Second):
	t.Fatalf("alice did not receive revocation from bob")
}
```
Might be a bit hacky though? Either the test timeout would need to be bumped up or the mock interval for the ticker would need to be decreased.  If the revocation was received before this check and the tick was already consumed, it doesn't matter since we will get a duplicate tick.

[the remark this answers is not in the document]
Crypt-iQ: It seems `TestChannelLinkSingleHopPayment` fails for a similar reason and can be updated with this.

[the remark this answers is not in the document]
wpaulino: Ah sorry, you're right. I see why the sleep is indeed needed, seems fine to me as is for now.

wpaulino:
> @wpaulino I wasn't entirely sure what to do about the extra error logs - I could remove it if the packet has a destRef field? What was once erroneous behavior is the norm so I think that'd make sense. What do you think?

Sounds good to me. I think @cfromknecht had some thoughts on possibly refactoring that portion of code to improve clarity of the new behavior, though I'd say it's not a blocker for this PR.

htlcswitch/link_test.go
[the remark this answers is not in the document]
Crypt-iQ: Hm, the `TestChannelLinkSingleHopPayment` no longer fails on my machine with `flake-unit`, so maybe it's just my Mac.

Crypt-iQ:
> > @wpaulino I wasn't entirely sure what to do about the extra error logs - I could remove it if the packet has a destRef field? What was once erroneous behavior is the norm so I think that'd make sense. What do you think?
> 
> Sounds good to me. I think @cfromknecht had some thoughts on possibly refactoring that portion of code to improve clarity of the new behavior, though I'd say it's not a blocker for this PR.

If we don't log the error, we should still return the error and it will still get reported in the link though.  So idk if the change is worth it?

htlcswitch/switch.go
```diff
@@ -1704,6 +1725,29 @@ out:
...
+			}
+
+			// Batch ack the settle/fail entries.
+			if err := s.ackSettleFail(s.pendingSettleFails...); err != nil {
+				log.Errorf("Unable to ack batch of settle/fails: %v", err)
+				continue
+			}
+
+			log.Tracef("Acked %d settle fails: %v", len(s.pendingSettleFails),
+				spew.Sdump(s.pendingSettleFails))
```
cfromknecht: small nit: for spews we typically wrap them with `newLogClosure` to prevent `Sdump` from being evaluated if we don't have trace logs enabled

htlcswitch/switch.go
```diff
@@ -1540,8 +1552,17 @@ func (s *Switch) htlcForwarder() {
...
 	defer s.cfg.FwdEventTicker.Stop()
 
+	defer s.cfg.AckEventTicker.Stop()
+
 out:
 	for {
+
+		// If the set of pending settle/fail entries is non-zero,
+		// reinstate the ack ticker so we can batch ack them.
+		if len(s.pendingSettleFails) > 0 {
```
cfromknecht: is it possible to move this statement after `handlePacketForward` in the `htlcPlex` case? should be fine since resolutions will never have a destref

htlcswitch/link_test.go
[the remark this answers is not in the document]
Crypt-iQ: Nevermind, it flakes!

htlcswitch/switch.go
Crypt-iQ: Do you mean here:
https://github.com/lightningnetwork/lnd/blob/9f065dccb1e1bee4c468d27098f41371695a7450/htlcswitch/switch.go#L1634-L1635
because I checked and couldn't find any resolution messages passing through that channel.  If we moved it, then the ticker would only wake once a new htlcPacket came across the channel after there were pendingSettleFails which seems fine but I prefer the current approach since it's more straightforward when the ticker will wake up.

cfromknecht: correct, the resolutions are processed in a different case that calls `handlePacketForward`, but they shouldn't have any destrefs and so don't influence the set of things that need to be acked. the current approach is fine by me, it is more conservative but certainly correct :)

htlcswitch/switch.go
Crypt-iQ: fixed

htlcswitch/switch.go
```diff
@@ -31,6 +31,10 @@ const (
 	// DefaultLogInterval is the duration between attempts to log statistics
 	// about forwarding events.
 	DefaultLogInterval = 10 * time.Second
+
+	// DefaultAckInterval is the duration between attempts to ack any settle
+	// fails in a forwarding package.
+	DefaultAckInterval = 15 * time.Second
```
Roasbeef: I would this we want this value to be much lower, like in the `ms`? The two above are for non-critical operations like logging.

lntest/itest/lnd_test.go
```diff
@@ -9384,6 +9384,11 @@ func testAsyncPayments(net *lntest.NetworkHarness, t *harnessTest) {
 	// All payments have been sent, mark the finish time.
 	timeTaken := time.Since(now)
 
+	// Wait for the revocation to be received so alice no longer has
+	// pending htlcs listed. This is needed due to the fact that we
+	// now pipeline the settles.
+	time.Sleep(2 * time.Second)
```
Roasbeef: Seems like we could instead wrap the block below in a `WaitPredicate` variant? Also I don't follow why this is needed, things should be _faster_ now, not slower.

Roasbeef: Ah ok, faster in the prefix of the route, but we still need to do a full dance in the remainder.

htlcswitch/switch.go
cfromknecht: this isn't too critical, it just needs to happen at some point so we stop resending the htlc internally. if it doesn't happen before shutdown it will be done when on the next connection

lntest/itest/lnd_test.go
Crypt-iQ: Done

htlcswitch/link_test.go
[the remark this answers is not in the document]
halseth: Can use `lntest.WaitPredicate instead`?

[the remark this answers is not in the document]
Crypt-iQ: Get an import cycle then, but we could just copy the function?

Roasbeef:
Needs a rebase @Crypt-iQ.

Crypt-iQ:
Rebased and squashed fixups @Roasbeef

cfromknecht:
@Crypt-iQ Cheers to lower latency for anyone routing through LND!! 🍾🎉🔥

joostjager:
Created issue for unjustified error: https://github.com/lightningnetwork/lnd/pull/3143

Not ideal that there is always an error added to the log during the very basic operation of a successful payment.

PR #3147, opened 2019-06-01
Add KeyForAddress to KeyRing interface and walletrpc call

AdamISZ:
This commit enables callers over grpc to access a KeyDescriptor
(including a raw pubkey) for an address if it is owned by the
wallet. This enables signing calls (via SignOutputRaw) by grpc
callers, if they only know the address and utxo information.

Some notes:

This was motivated by wanting to investigate possible coinjoin solutions (I won't get sidetracked with a long discussion here) but in any case, the idea is to enable signing functionality from outside the wallet; neither the `ComputeInputScript` nor the `SignOutputRaw` via the signrpc sub-rpc-server support this currently, based on the information available over the Lightning RPC interface; in particular, `listunspent` only provides the address and scriptPubkey (equivalent, obviously) and of course the utxo information, but not the underlying pubkey nor the key's HD derivation, at least one of which is required to populate a `KeyDescriptor`, which itself is required to populate a `SignDescriptor` object for signing to be possible.

Adding the `KeyForAddress` (could change name if you like) to the KeyRing interface seemed most logical to allow this. This is then exposed via the `walletrpc` rpc sub-server.

I realise I should write a test for the KeyForAddress functionality, I will add that.

Also, I had some minor uncertainties about the structure of the KeyLocator; technically, to achieve the above mentioned goal I can just populate the KeyDescriptor with the raw pubkey, but it's silly not to fill it out with a KeyLocator also (i.e. the HD path info), so I've done that, but it's a bit weird to me that (a) the KeyLocator doesn't include branch information (here it's just 0 i.e. external) and that the Account here is the btcwallet "defaultAccount" (see lnwallet/btcwallet/btcwallet.go), which is zero, that seems fine, but it's technically not part of the "KeyFamily" list, which is part of the Lightning wallet (I think this doesn't matter).


AdamISZ:
After a chat with Conner on IRC I think I missed something important re: the last paragraph above, the KeyLocator appears to be specific for the Lightning wallet key management, that's why it doesn't have a branch specified, so it may need a different struct type instead of KeyLocator passed through? (since this is supposed to be for on-chain keys).

Roasbeef:
> so it may need a different struct type instead of KeyLocator passed through? (since this is supposed to be for on-chain keys).

Yeah, so atm it's the case that the key locators assumes a given _scope_ (BIP 43 purpose). In order to support the use case you have in mind an additional field to specify the _scope_ needs to be added.

Roasbeef:
Actually I think ultimately once we implement PSBT support, your original use case is handled?

AdamISZ:
Well first, feel free to close this as I don't seem to find time to take this little project up again, at the moment. I can always come back to it.

> Actually I think ultimately once we implement PSBT support, your original use case is handled?

But more directly to your question: I wouldn't think so? The idea was something like: could I implement [SNICKER](https://gist.github.com/AdamISZ/2c13fb5819bd469ca318156e2cf25d79) receivers on an lnd wallet, which would require access to the public key of a utxo.

And although I haven't kept up with recent lnd developments, at the time I was looking at this I had the feeling that this wouldn't necessarily be the only scenario in which access to public keys might be of use. But that kind of thing is very much your call.

Roasbeef:
Closing due to inactivity.

Issue #386, opened 2017-10-20
Handling non-confirmed funding transactions


halseth:
### Issue
During the current funding workflow, the responder should forget the channel if the funding transaction is not confirmed [within a reasonable timeout](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#requirements-4). For `lnd` this timeout is [currently set to 288 blocks](https://github.com/lightningnetwork/lnd/blob/master/fundingmanager.go#L47).

The funder (initiator) of the channel cannot forget about the channel if the funding transaction is broadcast, since he has funded the transaction with coins. Therefore, we should make sure the money is returned to the funder's wallet.

### Steps to completion:

- [ ] Prevent the funder from forgetting about the channel if the `fundingManager` gets restarted while waiting for the funding transaction to confirm. Currently, on restart it will [launch a `goroutine` that will timeout after 288 blocks](https://github.com/lightningnetwork/lnd/blob/master/fundingmanager.go#L387), but this should only be done for the channel responder.

- [ ] Determine how to reclaim the funds. We can either double spend the input to the funding transaction back into our wallet, or broadcast the commitment transaction and sweep the funds using CPFP.

Extra credits: We might also be able to bump the fee of the funding transaction to get it confirmed, but this would possibly need some more advanced scripting features, due to changing the amounts int the funding tx changes the `txid`, which the commitment tx relies on.

johynpapin:
I'll try to resolve this issue.

johng:
I've done the first part here #514

jemilsson:
It would be really neat to also be able to "cancel" the opening of a channel and reclaim the funds while the funding transaction is  unconfirmed, even if the timout hasn't expired. This way you could avoid having to wait for the timeout.

Roasbeef:
First portion fixed by #514.

Crypt-iQ:
Hi if nobody is working on this, I'd like to take it

Crypt-iQ:
Since I haven't gotten a reply I'll assume nobody is working on this and work on it myself

halseth:
I believe @wpaulino has started on this one.


On Wed, Jul 11, 2018 at 4:47pm, Eugene < notifications@github.com [notifications@github.com] > wrote:
Since I haven't gotten a reply I'll assume nobody is working on this and work on it myself

johynpapin:
I'm really sorry I didn't give any news. I no longer work on this problem indeed.

[...]

PR #1755, opened 2018-08-20
[WIP] Funding detect double spends

halseth:
This PR makes the `fundingmanager` aware of transaction spending inputs  to the funding transaction, invalidating the funding tx. In this case we will cancel the funding flow.

TODO:
- [ ] handle this also after startup by extracting common logic
- [ ] properly set `pkScript` for funding tx inputs
- [ ] add scenario to integration tests

Fixes #1494
Fixes #1623

~~Blocked by #1910~~

PR #3149, opened 2019-06-02
Funding TX Recovery

johng:
For the final part of #386 

Comments:
* Uses CPFP, spending from our change funding point output
* Done some refactoring in fundingmanager + lnd test to prevent code dup 
* Have only been testing this in simnet so far, but the workflow seems correct

This PR is mainly to get initial feedback on how this mechanism will work, would be great to get some comments on my initial ideas.

#### Pull Request Checklist

- [ ] If this is your first time contributing, we recommend you read the [Code
- [ ] All changes are Go version 1.12 compliant
- [ ] The code being submitted is commented according to [Code Documentation and Commenting](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#CodeDocumentation)
- [ ] For new code: Code is accompanied by tests which exercise both
- [ ] For bug fixes: Code is accompanied by new tests which trigger
- [ ] Any new logging statements use an appropriate subsystem and
  logging level
- [ ] Code has been formatted with `go fmt`
- [ ] For code and documentation: lines are wrapped at 80 characters
  per default)
- [ ] Running `make check` does not fail any tests
- [ ] Running `go vet` does not report any issues
- [ ] Running `make lint` does not report any **new** issues that did not
  already exist
- [ ] All commits build properly and pass tests. Only in exceptional
- [ ] Commits have a logical structure according to [Ideal Git Commit Structure](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#IdealGitCommitStructure)


halseth:
How do we handle stuck funding transactions with no change output?

Also, note that there is a fundingmanager refactor PR up: #1910

johng:
Then we should not entering recovery mode, and just continue to note. Will need to implement that

johng:
> How do we handle stuck funding transactions with no change output?

Currently we fail the recovery workflow, added a change to prevent entering it in the first place

johng:
@halseth I've rebased ontop of your changes in #1910

halseth:
Now that we have fee bumping added, we should explore reusing that logic to bump the funding tx fee: https://github.com/lightningnetwork/lnd/pull/3140

halseth:
After discussing this a bit we have come to the conclusion we _do not_ want to introduce more persistent state during the funding process. Instead we should let the fundingmanager 
1) unlock the coins after a timeout, making them spendable in other transactions. 
2) In combination with something like #1755  this should allow it to gracefully fail the channel.

To bump the fee on an unconfirmed channel manually, the current bump fee logic can be used.

I will add this to the issue, feel free to open a new PR for 1).

Issue #3150, opened 2019-06-02
queryroutes uses unrouteable final_cltv_delta value by default


Crypt-iQ:
### Background

The `queryroutes` command uses the `zpay32.DefaultFinalCltvDelta` value to extend to the final hop if none is specified with the request:

https://github.com/lightningnetwork/lnd/blob/4068e78af690f9b4a598de1f3f0b21b5560dd146/routing/router.go#L1398-L1403

However, using `sendtoroute` will fail on the exit hop with `FinalExpiryTooSoon` because of the check here:
https://github.com/lightningnetwork/lnd/blob/4068e78af690f9b4a598de1f3f0b21b5560dd146/invoices/invoiceregistry.go#L534-L540

The `finalCltvRejectDelta` is 13 which is greater than the final cltv default of 9, so this will fail.  Additionally, the default `--bitcoin.timelockdelta` config option is 40, so the final cltv should be _at least_ 40 or else the second check will probably fail given a user's configuration. This isn't a problem with the `addinvoice` command since that uses the `--bitcoin.timelockdelta` value.  This is a quick fix, so I can fix it once we decide on a fix.

### Your environment

* version of `lnd`: master

joostjager:
> Additionally, the default --bitcoin.timelockdelta config option is 40, so the final cltv should be at least 40 or else the second check will probably fail given a user's configuration.

I don't think this is true. The second check compares the htlc expiry with the invoice cltv delta requirement.

I indeed think the fix is to raise `zpay32.DefaultFinalCLTVDelta` to a larger number.

Crypt-iQ:
The only way the invoice cltv delta of an invoice can be configured with the cli with the addinvoice command is through the `--bitcoin.timelockdelta` flag since the `CltvExpiry` field doesn't appear enabled.  I tested raising `zpay.DefaultFinalCLTVDelta` above 13 and the sendtoroute call still failed on the exit hop because the default value of 40 was used as the invoice cltv delta when creating an invoice.  Am I missing something?

joostjager:
If the goal is for `QueryRoutes` to come up with a valid route assuming the default `lnd` invoice expiry, it should indeed be raised to `40`. But invoices could also have been created (using a direct rpc call) with a different expiry. And `QueryRoutes` can also be used to pay to other implementations that have different defaults.

Thinking about it now, maybe the problem really is that the default doesn't make sense? That it should be a required parameter to force callers to think about it?

[...]

PR #3153, opened 2019-06-03
routing: add block padding to prevent inflight HTLC failures

### LINKED ISSUE #3055: routing: payment fails if block mined while htlc is in flight
### Background

Currently the sender of a payment builds a route that delivers a final htlc to the receiver at exactly the invoice final cltv delta. This assumes that while the htlc is on its way towards the receiver, no block is mined. If a new block does arrive, the payment fails with a `FinalExpiryTooSoon` error.

Assuming it takes on average 5 seconds for an htlc to arrive at the receiver and blocks come in every 600 secs, about 1% of the payments are expected to fail because of this.

### Your environment

* lnd 0.6.1-rc2
* linux
* bitcoind v0.17

### Steps to reproduce

* Setup a three node network Alice -> Bob -> Charlie
* Modify Bob so that it delays forwarding the htlc for 30 seconds
* Send a payment Alice -> Charlie via Bob
* When Bob enters the delay, mine a block

### Expected behaviour

Payment succeeded.

### Actual behaviour

Payment fails with `FinalExpiryTooSoon`.

### Fix suggestions
1. Pad the route that Alice constructs with one block.
2. Alternatively, retry the payment.

Crypt-iQ:
Fixes #3055 

Fixed some integration tests + unit tests.  Pretty self-explanatory.  Wasn't sure if I should have added an integration test?


routing/router.go
```diff
@@ -1402,6 +1402,10 @@ func (r *ChannelRouter) FindRoute(source, target route.Vertex,
 		finalCLTVDelta = finalExpiry[0]
 	}
 
+	// Add BlockPadding to the finalCltvDelta so that the receiving node
+	// does not reject the HTLC if a block is mined while its in-flight.
+	finalCLTVDelta += BlockPadding
```
joostjager: I am in doubt whether we should also pad when the user specified an explicit final cltv expiry for `queryroutes`. It is a more low level function and it may be confusing if it doesn't returns what was requested?

Maybe the same thing holds for `sendpayment` with a set expiry? Perhaps the padding should just happen when only an encoded payment request is supplied.

Will get back to you.

joostjager: @Crypt-iQ it seems best to only apply the padding if no explicit final cltv delta is set in the rpc request.

routing/payment_session.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
+// BlockPadding is used to increment the finalCltvDelta value for the last hop
+// to prevent an HTLC being failed if a block is mined while it's in-flight.
+var BlockPadding uint16 = 1
```
cfromknecht: make `const`? also discussed with @Roasbeef and decided maybe bump to 3 to be conservative

routing/router.go
cfromknecht: looks like this still needs to be moved

lntest/itest/lnd_test.go
```diff
@@ -3044,8 +3045,8 @@ func testChannelForceClosure(net *lntest.NetworkHarness, t *harnessTest) {
 	// output spending from the commitment txn, so we must deduct the number
 	// of blocks we have generated since adding it to the nursery, and take
 	// an additional block off so that we end up one block shy of the expiry
-	// height.
-	cltvHeightDelta := defaultCLTV - defaultCSV - 2 - 1
+	// height, and add one additional block for the block padding.
+	cltvHeightDelta := uint32(routing.BlockPadding) + defaultCLTV - defaultCSV - 2 - 1
```
cfromknecht: perhaps define a small helper for padding cltvs, e.g. `padCLTV(x) { return x+routing.BlockPadding }`?

routing/payment_session_test.go
```diff
@@ -18,8 +18,8 @@ func TestRequestRoute(t *testing.T) {
 		[]*channeldb.ChannelEdgePolicy, error) {
 
 		// We expect find path to receive a cltv limit excluding the
-		// final cltv delta.
-		if *r.CltvLimit != 22 {
+		// final cltv delta (including the block padding).
+		if *r.CltvLimit != 21 {
```
cfromknecht: might be clearer to write this as is done below, e.g. `22 - padding`

routing/router.go
Crypt-iQ: Will do this week

Crypt-iQ: Addressed the comments.  This is related to #3150 since if there is no expiry set, the payment request will likely fail because of sender-side checks (at least if the node is an lnd node).  The result of that discussion was to *require* the final cltv delta.  Thoughts?

routing/payment_session.go
```diff
@@ -154,6 +158,13 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
 		return nil, fmt.Errorf("pre-built route already tried")
 	}
 
+	// Add BlockPadding to the finalCltvDelta so that the receiving node
+	// does not reject the HTLC if a block is mined while its in-flight.
+	// This is only added if no finalCltvDelta was set.
+	if finalCltvDelta == 0 {
```
Roasbeef: If we only add block padding in this case, then it doesn't address the common case where a users payment fails due to a block being mined _right_ as they send. We _always_ set our `finalCltvDelta` if the payment goes through `SendPayment`.

Crypt-iQ: Ahhhh! I missed that.  Joost had a comment above where he said the block padding should *only* be specified iff no `finalCltvDelta` was supplied.  But it seems that has some problems as well.
see: https://github.com/lightningnetwork/lnd/pull/3153#discussion_r302687572

Roasbeef: From my reading, his comment only applied to `queryroutes`. IMO, we should also apply the padding there, as users are known to execute commands that resemble: `queryroutes | sendtoroute`.

Crypt-iQ: Yeah that's true, should it be set any time a route is requested then or perhaps this could be the default with an option to turn the padding off?

Roasbeef:
SGTM

routing/pathfind_test.go
```diff
@@ -1707,14 +1709,14 @@ func TestPathFindSpecExample(t *testing.T) {
 		t.Fatalf("wrong hop fee: got %v, expected %v", fee, 0)
 	}
 
-	// The CLTV expiry should be the current height plus 9 (the expiry for
-	// the B -> C channel.
+	// The CLTV expiry should be the current height plus 9 (the expiry for the
+	// B -> C channel) plus the block padding.
 	if route.TotalTimeLock !=
-		startingHeight+zpay32.DefaultFinalCLTVDelta {
+		startingHeight+zpay32.DefaultFinalCLTVDelta+uint32(BlockPadding) {
```
cfromknecht: `noPadding` is `false`, so why does this change?

routing/pathfind_test.go
```diff
@@ -1755,15 +1757,15 @@ func TestPathFindSpecExample(t *testing.T) {
...
 	// The total amount should factor in a fee of 10199 and also use a CLTV
-	// delta total of 29 (20 + 9),
+	// delta total of 29 plus the block padding.
 	expectedAmt := lnwire.MilliSatoshi(5010198)
 	if route.TotalAmount != expectedAmt {
 		t.Fatalf("wrong amount: got %v, expected %v",
 			route.TotalAmount, expectedAmt)
 	}
-	if route.TotalTimeLock != startingHeight+29 {
+	if route.TotalTimeLock != startingHeight+29+uint32(BlockPadding) {
```
cfromknecht: similar comment here

routing/payment_session.go
joostjager: Another way to see this is, is from the perspective of keeping the api surface as small as possible. `QueryRoutes` (the rpc) is for advanced api users and you could also say that they should just supply just an absolute block height. This should include the final cltv expiry of the invoice and block padding. It would give more control over the time locks of the route returned, which may be required if there is chaining involved with time locks of other payments.

`QueryRoutes` (the rpc) is different from `queryroutes` (the cli command). We could choose to have `lncli` do some of the calculation to still be able to `queryroutes | sendtoroute`.

halseth:
@Crypt-iQ  Decided to make the changes only apply to `SendPayment` for now, and then we can move the `queryroutes`/`sendToRoute` discussion to a new PR.

Crypt-iQ:
> @Crypt-iQ Decided to make the changes only apply to `SendPayment` for now, and then we can move the `queryroutes`/`sendToRoute` discussion to a new PR.

See#3339

wpaulino:
Needs a rebase!

joostjager:
Closing until we have good reasons to also add implicit padding for the low-level calls `SendToRoute` and `QueryRoutes`.

PR #3154, opened 2019-06-04
neutrinonotify: async call GetUtxo from RegisterSpendNtfn

halseth:
Since GetUtxo is a potentially long running call, we would see
RegisterSpendNtfn block in some cases, in particular on starting the
chain watcher, causing lnd to hang on startup.

This commit makes the call to GetUtxo in a go routine, letting
RegisterSpendNtfn return immediately when the notification events are
created.


wpaulino:
Why is this necessary if we already pass the `quit` chan through a functional option?

wpaulino:
Ah I see, it can stall while starting the chain watchers, so if we shutdown before that, then `Stop` won't be called and the `quit` chan won't be closed.

Issue #3146, opened 2019-06-01
Payments made with sendtoroute do not reflect the paid amounts in listpayments


alexbosworth:
In `19fafd7a9a5fe4ba05a6250aae748f2b7465ea05` the behavior of listpayments changes to no longer return the sent value of payments made with sendtoroute

Prior to this commit, sendtoroute payments values could be accessed via listpayments

After this commit, the values are returned as 0 instead of their actual amount

The expected value for these payments is the route total amount - total fees

ksedgwic:
I think I'm seeing this:

```
        {
            "payment_hash": "352a2df60a19175f06cb5fa9993436e3a0f16cee5b32cfba8dd8b1fcf82ccef5",
            "value": "0",
            "creation_date": "1559353582",
            "path": [
                "0234d70cf65a4e8974d3d49e23ec7b9792877f783f2f5dfd3882422aada18a916c",
                "02a5fa844d310f582d209fe649352b225440b8a54e77361f229bb92ee263c87e6f",
                "02e9046555a9665145b0dbd7f135744598418df7d61d3660659641886ef1274844",
                "02a5fa844d310f582d209fe649352b225440b8a54e77361f229bb92ee263c87e6f"
            ],
            "fee": "3",
            "payment_preimage": "16bd366b68d34960f16436fd5ca93cf67745c40d7c143204534a5e6f5a8599da",
            "value_sat": "0",
            "value_msat": "0"
        },
￼```

Sending with:
```
		sendRsp, err := router.SendToRoute(ctxt, &routerrpc.SendToRouteRequest{
			PaymentHash: invoiceRsp.RHash,
			Route: route,
		})
```
Version:
```
lnd version 0.6.1-beta commit=v0.6.1-beta-181-g1b5655f7fab463558b0e6a4120339581e810c096
```

ksedgwic:
### Background

While rebalancing channels using loops created with QueryRoutes and
routerrpc SendToRoute sends are not credited to the
destination channel.

### Your environment

* mainnet
* lnd version 0.6.1-beta commit=v0.6.1-beta-181-g1b5655f7fab463558b0e6a4120339581e810c096
* Linux home2 4.15.0-48-generic #51~16.04.1-Ubuntu SMP Fri Apr 5 12:01:12 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
[...]

PR #3155, opened 2019-06-04
routing/router: correct SendToRoute's amount record in DB

halseth:
Previously we would mistakenly use the payment value from the dummy
LightningPayment struct, which would obviously be 0 always. Now we
instead calculate the value from the given route.

This regression was introduced in https://github.com/lightningnetwork/lnd/pull/2761/commits/de1bf8a518145e7b33a6af6d64b05d560bf2cf73#diff-01ada5680aaa18fc79cbf78b2d55a37b, where we changed from using the route to calculate the amount, to using the LightningPayment.

Fixes https://github.com/lightningnetwork/lnd/issues/3146.

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -1637,24 +1637,29 @@ func (r *ChannelRouter) SendToRoute(hash lntypes.Hash, route *route.Route) (
 	// Create a payment session for just this route.
 	paySession := r.cfg.MissionControl.NewPaymentSessionForRoute(route)
 
-	// Create a (mostly) dummy payment, as the created payment session is
-	// not going to do path finding.
-	payment := &LightningPayment{
-		PaymentHash: hash,
-	}
+	// Calculate amount paid to receiver.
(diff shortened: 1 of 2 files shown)


halseth:
cc @alexbosworth

routing/router.go
```diff
@@ -1637,26 +1637,31 @@ func (r *ChannelRouter) SendToRoute(hash lntypes.Hash, route *route.Route) (
...
 
-	err := r.cfg.Control.InitPayment(payment.PaymentHash, info)
+	err := r.cfg.Control.InitPayment(hash, info)
 	if err != nil {
 		return [32]byte{}, err
 	}
 
+	// Create a (mostly) dummy payment, as the created payment session is
+	// not going to do path finding.
+	// TODO(halseth): sendPayment doesn't relly need LightningPayment, make
```
wpaulino: 👍

PR #3156, opened 2019-06-04
routerrpc: add more failure reasons and route hints

### LINKED ISSUE #3202: Allow hop hints to be specified in routerrpc sendPayment
In the routerrpc API, the following fields are supported for specifying payment details if not using a payment request

- dest
- amt
- payment_hash
- final_cltv_delta

These fields are insufficient to pay a private invoice because they are missing a way to pass hop hints

It should be the case that private invoices are payable with the API without using a BOLT11 payment request to pass payment details

joostjager:
Changes:
* `FailureReasonError` added. This is returned in case we get an unexpected response to our payment attempt for which it does not make sense to try different routes.
* `FailureReasonIncorrectPaymentDetails` added. This is returned in case the payment hash is unknown, the amount is incorrect and/or the final cltv expiry is incorrect. The route that triggered this failure is returned along side the failure message itself.
* Route hints can be passed into `routerrpc.SendPayment`. Previously this was only possible when the route hints were encoded in a payment request. Fixes #3202

These changes allow probing destinations (that may require route hints) with an unknown payment hash. By inspecting the route when a probe succeeds, a realistic fee and time lock estimate can be obtained.


cmd/lncli/commands.go
```diff
@@ -2012,6 +2012,38 @@ var cltvLimitFlag = cli.UintFlag{
 		"this payment",
 }
 
+// paymentFlags returns common flags for sendpayment and payinvoice.
+func paymentFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:  "pay_req",
+			Usage: "a zpay32 encoded payment request to fulfill",
```
alexbosworth: why still zpay32

joostjager: commit is removed from pr

channeldb/payments.go
```diff
@@ -108,11 +115,21 @@ func (r FailureReason) String() string {
...
+	case FailureReasonIncorrectPaymentDetails:
+		return "incorrect_payment_details"
 	}
 
 	return "unknown"
 }
 
+// Ref returns a reference to the failure reason. This is useful in a nilable
+// context.
+func (r FailureReason) Ref() *FailureReason {
```
Roasbeef: Why this over just having the caller take the address of the struct they hold?

routing/payment_lifecycle.go
```diff
@@ -338,12 +338,12 @@ func (p *paymentLifecycle) sendPaymentAttempt(firstHop lnwire.ShortChannelID,
...
 // whether we should make another payment attempt.
 func (p *paymentLifecycle) handleSendError(sendErr error) error {
-	var finalOutcome bool
+	var finalOutcome *channeldb.FailureReason
 
 	// If an internal, non-forwarding error occurred, we can stop trying.
 	fErr, ok := sendErr.(*htlcswitch.ForwardingError)
 	if !ok {
-		finalOutcome = true
+		finalOutcome = channeldb.FailureReasonError.Ref()
```
Roasbeef: Hmm, will the compiler allow one to do `*finalOutcome = somethingElse=? Given that you're obtaining a pointer to a constant.

lnrpc/routerrpc/router.proto
```diff
@@ -86,6 +86,17 @@ enum PaymentState {
...
+    /**
+    A non-recoverable error has occured.
+    */
+    FAILED_ERROR = 4;
+
+    /**
+    Payment details incorrect (unknown hash, invalid amt or
+    invalid final cltv delta)
+    */
+    FAILED_INCORRECT_PAYMENT_DETAILS = 5;
```
Roasbeef: Will we also eventually surface this additional meta data? So the payment hash extended, amount, etc?

routing/control_tower.go
```diff
@@ -191,23 +203,22 @@ func (p *controlTower) SubscribePayment(paymentHash lntypes.Hash) (
...
-		event.FailureReason = *payment.Failure
+		event = createFailedResult(*payment.Failure)
 
 	default:
 		return false, nil, errors.New("unknown payment status")
 	}
 
 	// Write immediate result to the channel.
-	c <- event
+	c <- *event
```
Roasbeef: Why not leave the declaration above is a regular type since we immediately deference this pointer?

lnrpc/routerrpc/router_server.go
```diff
@@ -546,6 +546,12 @@ func (s *Server) trackPayment(paymentHash lntypes.Hash,
 				return err
 			}
 			status.State = state
+			if result.Route != nil {
+				status.Route = s.cfg.RouterBackend.
```
Roasbeef: Odd wrapping here. Maybe just capture `s.cfg.RouterBackend` above as a `router` variable?

lnrpc/routerrpc/router.proto
```diff
@@ -58,6 +58,11 @@ message SendPaymentRequest {
     maximum enforced.
     */
     int32 cltv_limit = 9;
+
+    /**
+    Optional route hints to reach the destination through private channels.
+    */
+    repeated lnrpc.RouteHint route_hints = 10 [json_name = "route_hints"];
```
Roasbeef: Since we already have the payment request, which itself can include routing hints, why do we need a distinct field?

Roasbeef: I would think we'd want this on the `SendToRoute` request rather than here?

lnrpc/routerrpc/router_backend.go
```diff
@@ -382,6 +384,15 @@ func (r *RouterBackend) extractIntentFromSendRequest(
 	payIntent.PayAttemptTimeout = time.Second *
 		time.Duration(rpcPayReq.TimeoutSeconds)
 
+	// Route hints.
```
Roasbeef: Doesn't attempt to also parse the route hints from the regular payment request.

channeldb/payments.go
joostjager: This is a utility function to get a pointer to value defined by const.

routing/payment_lifecycle.go
joostjager: It is not a pointer to a constant. The `Ref()` function creates a new variable with the same value as the const and returns a pointer to that variable. Pointer to constant if impossible afaik.

lnrpc/routerrpc/router.proto
joostjager: We can't. Because of the anti-probe measure we took on the protocol level, we don't know which of the parameters is incorrect.

routing/control_tower.go
joostjager: Fixed

lnrpc/routerrpc/router_server.go
joostjager: Fixed

lnrpc/routerrpc/router.proto
joostjager: In general I think that anything possible via a payment request, should also be possible by specifying the payment parameters separately.

I use this to pre-pay probe a payment for which I have a payment request. I decode the payment request, modify the payment hash to a random value and pass in the separate parameters, including hop hints, to `SendPayment`.

For `SendToRoute` it wouldn't be necessary to pass in hop hints, as those are already part of the supplied route.

lnrpc/routerrpc/router_backend.go
joostjager: Payment request route hints are appended to the list of hints further down. This allows you to augment the encoded hints with a few more.

routing/payment_lifecycle.go
joostjager: Changed `processSendError` to return a `final` bool and the reason. On second thought the `Ref()` thing was a bit too weird.

channeldb/payments.go
joostjager: Removed now

joostjager:
@Roasbeef ptal

routing/router.go
```diff
@@ -1918,7 +1936,7 @@ func (r *ChannelRouter) processSendError(paySession PaymentSession,
 		processChannelUpdateAndRetry(
 			&onionErr.Update, errSource,
 		)
-		return false
+		return false, 0
```
cfromknecht: are all of these 0's meant to be `channeldb.FailureReasonTimeout`?

channeldb/payments.go
```diff
@@ -94,10 +94,17 @@ const (
 	// destination was found during path finding.
 	FailureReasonNoRoute FailureReason = 1
 
+	// FailureReasonError indicates that an unexpected error happened during
+	// payment.
+	FailureReasonError FailureReason = 2
+
+	// FailureReasonIncorrectPaymentDetails indicates that either the hash
+	// is unknown or the final cltv delta or amount is incorrect.
+	FailureReasonIncorrectPaymentDetails FailureReason = 3
```
cfromknecht: not a blocker (related to my other comment tho), but are these serialized anywhere? if not, it might be useful to define 0 as the unknown case so that forgetting to set it results in an unknown error rather than the an incorrect-but-known error

routing/payment_lifecycle.go
```diff
@@ -338,39 +338,40 @@ func (p *paymentLifecycle) sendPaymentAttempt(firstHop lnwire.ShortChannelID,
...
+	}
+
+	log.Debugf("Payment %x failed: final_outcome=%v, raw_err=%v",
+		p.payment.PaymentHash, reason, sendErr)
 
-		// Terminal state, return the error we encountered.
-		return sendErr
+	// Mark the payment failed with no route.
+	// TODO(halseth): make payment codes for the actual reason we
+	// don't continue path finding.
```
cfromknecht: nit: short line wrap

routing/payment_lifecycle.go
```diff
@@ -338,39 +338,40 @@ func (p *paymentLifecycle) sendPaymentAttempt(firstHop lnwire.ShortChannelID,
...
-		// Save the forwarding error so it can be returned if this turns
-		// out to be the last attempt.
-		p.lastError = fErr
-	}
-
-	if finalOutcome {
-		log.Errorf("Payment %x failed with final outcome: %v",
-			p.payment.PaymentHash, sendErr)
+		if !final {
+			// Save the forwarding error so it can be returned if this turns
```
cfromknecht: nit: long line wrap?

routing/router.go
joostjager: No, the `0` values are irrelevant, because the first return parameter is `false`. I first had a single return parameter `*channeldb.FailureReason`, but it got ugly with obtaining pointers to consts.

channeldb/payments.go
joostjager: We started serializing this to the db with the reliable payments PRs. It cannot be changed without a db migration.

That was the advantage of the pointer approach. You were forced to handle it.

routing/payment_lifecycle.go
joostjager: fixed

routing/payment_lifecycle.go
joostjager: fixed

Issue #1874, opened 2018-09-10
Database update error after updating to lnd v0.5-beta-rc2


peername:
This is the error I get during the start process:

2018-09-10 16:32:14.944 [INF] LTND: Version 0.5.0-beta commit=1941353fb28755a170793e43595601d75c8f3dda
2018-09-10 16:32:14.944 [INF] LTND: Active chain: Bitcoin (network=mainnet)
2018-09-10 16:32:14.944 [INF] CHDB: Checking for schema update: latest_version=6, db_version=5
2018-09-10 16:32:14.944 [INF] CHDB: Performing database schema migration
2018-09-10 16:32:14.944 [INF] CHDB: Applying migration #6
2018-09-10 16:32:14.945 [INF] CHDB: Migrating database to properly prune edge update index
2018-09-10 16:32:14.964 [INF] LTND: Shutdown complete
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt.(*leafPageElement).value(...)
	/home/peernode/go/src/github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt/page.go:126
github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt.(*Bucket).Stats.func1(0x7f344127a000, 0x3)
	/home/peernode/go/src/github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt/bucket.go:440 +0x34a
github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt.(*Tx).forEachPage(0xc420590000, 0x527a, 0x3, 0xc4206d3660)
	/home/peernode/go/src/github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt/tx.go:607 +0x81
github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt.(*Tx).forEachPage(0xc420590000, 0x4d23, 0x2, 0xc4206d3660)
	/home/peernode/go/src/github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt/tx.go:613 +0xd7
github.com/lightningnetwork/lnd/vendor/github.com/coreos/bbolt.(*Tx).forEachPage(0xc420590000, 0x1942e, 0x1, 0xc4206d3660)
[...]

PR #3164, opened 2019-06-05
routing: persistent mission control

joostjager:
This PR adds persistence to mission control by storing the raw payment results and reprocessing them on startup.

### DIFF
--- a/lnrpc/routerrpc/config.go
+++ b/lnrpc/routerrpc/config.go
@@ -3,26 +3,25 @@ package routerrpc
 import (
 	"time"
 
-	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/btcsuite/btcutil"
 )
 
 // RoutingConfig contains the configurable parameters that control routing.
 type RoutingConfig struct {
-	// PenaltyHalfLife defines after how much time a penalized node or
-	// channel is back at 50% probability.
-	PenaltyHalfLife time.Duration
-
-	// PaymentAttemptPenalty is the virtual cost in path finding weight
-	// units of executing a payment attempt that fails. It is used to trade
-	// off potentially better routes against their probability of
-	// succeeding.
-	PaymentAttemptPenalty lnwire.MilliSatoshi
-
-	// MinProbability defines the minimum success probability of the
-	// returned route.
-	MinRouteProbability float64
+	// MinRouteProbability is the minimum required route success probability
+	// to attempt the payment.
+	MinRouteProbability float64 `long:"minrtprob" description:"Minimum required route success probability to attempt the payment"`
 
 	// AprioriHopProbability is the assumed success probability of a hop in
 	// a route when no other information is available.
-	AprioriHopProbability float64
+	AprioriHopProbability float64 `long:"apriorihopprob" description:"Assumed success probability of a hop in a route when no other information is available."`
+
+	// PenaltyHalfLife defines after how much time a penalized node or
+	// channel is back at 50% probability.
+	PenaltyHalfLife time.Duration `long:"penaltyhalflife" description:"Defines the duration after which a penalized node or channel is back at 50% probability"`
+
+	// AttemptCost is the virtual cost in path finding weight units of
+	// executing a payment attempt that fails. It is used to trade off
+	// potentially better routes against their probability of succeeding.
+	AttemptCost btcutil.Amount `long:"attemptcost" description:"The (virtual) cost in sats of a failed payment attempt"`
 }
--- a/lnrpc/routerrpc/config.go
+++ b/lnrpc/routerrpc/config.go
@@ -24,4 +24,8 @@ type RoutingConfig struct {
 	// executing a payment attempt that fails. It is used to trade off
 	// potentially better routes against their probability of succeeding.
 	AttemptCost btcutil.Amount `long:"attemptcost" description:"The (virtual) cost in sats of a failed payment attempt"`
+
+	// MaxMcHistory defines the maximum number of payment results that
+	// are held on disk by mission control.
+	MaxMcHistory int `long:"maxmchistory" description:"the maximum number of payment results that are held on disk by mission control"`
 }
--- a/lnrpc/routerrpc/config_active.go
+++ b/lnrpc/routerrpc/config_active.go
@@ -3,10 +3,6 @@
 package routerrpc
 
 import (
-	"time"
-
-	"github.com/btcsuite/btcutil"
-	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/macaroons"
 	"github.com/lightningnetwork/lnd/routing"
 )
@@ -17,27 +13,12 @@ import (
 // options, while if able to be populated, the latter fields MUST also be
 // specified.
 type Config struct {
+	RoutingConfig
+
 	// RouterMacPath is the path for the router macaroon. If unspecified
 	// then we assume that the macaroon will be found under the network
 	// directory, named DefaultRouterMacFilename.
 	RouterMacPath string `long:"routermacaroonpath" description:"Path to the router macaroon"`
-
-	// MinProbability is the minimum required route success probability to
-	// attempt the payment.
-	MinRouteProbability float64 `long:"minrtprob" description:"Minimum required route success probability to attempt the payment"`
-
-	// AprioriHopProbability is the assumed success probability of a hop in
-	// a route when no other information is available.
-	AprioriHopProbability float64 `long:"apriorihopprob" description:"Assumed success probability of a hop in a route when no other information is available."`
-
-	// PenaltyHalfLife defines after how much time a penalized node or
-	// channel is back at 50% probability.
-	PenaltyHalfLife time.Duration `long:"penaltyhalflife" description:"Defines the duration after which a penalized node or channel is back at 50% probability"`
-
-	// AttemptCost is the virtual cost in path finding weight units of
-	// executing a payment attempt that fails. It is used to trade off
-	// potentially better routes against their probability of succeeding.
-	AttemptCost int64 `long:"attemptcost" description:"The (virtual) cost in sats of a failed payment attempt"`
 
 	// NetworkDir is the main network directory wherein the router rpc
 	// server will find the macaroon named DefaultRouterMacFilename.
@@ -62,13 +43,16 @@ type Config struct {
 
 // DefaultConfig defines the config defaults.
 func DefaultConfig() *Config {
-	return &Config{
+	defaultRoutingConfig := RoutingConfig{
 		AprioriHopProbability: routing.DefaultAprioriHopProbability,
 		MinRouteProbability:   routing.DefaultMinRouteProbability,
 		PenaltyHalfLife:       routing.DefaultPenaltyHalfLife,
-		AttemptCost: int64(
-			routing.DefaultPaymentAttemptPenalty.ToSatoshis(),
-		),
+		AttemptCost: routing.DefaultPaymentAttemptPenalty.
+			ToSatoshis(),
+	}
+
+	return &Config{
+		RoutingConfig: defaultRoutingConfig,
 	}
 }
 
@@ -77,9 +61,7 @@ func GetRoutingConfig(cfg *Config) *RoutingConfig {
 	return &RoutingConfig{
 		AprioriHopProbability: cfg.AprioriHopProbability,
 		MinRouteProbability:   cfg.MinRouteProbability,
-		PaymentAttemptPenalty: lnwire.NewMSatFromSatoshis(
-			btcutil.Amount(cfg.AttemptCost),
-		),
-		PenaltyHalfLife: cfg.PenaltyHalfLife,
+		AttemptCost:           cfg.AttemptCost,
+		PenaltyHalfLife:       cfg.PenaltyHalfLife,
 	}
 }
--- a/lnrpc/routerrpc/config_active.go
+++ b/lnrpc/routerrpc/config_active.go
@@ -49,6 +49,7 @@ func DefaultConfig() *Config {
 		PenaltyHalfLife:       routing.DefaultPenaltyHalfLife,
 		AttemptCost: routing.DefaultPaymentAttemptPenalty.
 			ToSatoshis(),
+		MaxMcHistory: routing.DefaultMaxMcHistory,
 	}
 
 	return &Config{
@@ -63,5 +64,6 @@ func GetRoutingConfig(cfg *Config) *RoutingConfig {
 		MinRouteProbability:   cfg.MinRouteProbability,
 		AttemptCost:           cfg.AttemptCost,
 		PenaltyHalfLife:       cfg.PenaltyHalfLife,
+		MaxMcHistory:          cfg.MaxMcHistory,
 	}
 }
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -395,7 +395,7 @@ func (m *MissionControl) GetHistorySnapshot() *MissionControlSnapshot {
 // returns a bool indicating whether this error is a final error. If it is
 // final, a failure reason is returned and no further payment attempts need to
 // be made.
-func (m *MissionControl) ReportPaymentFail(rt *route.Route,
+func (m *MissionControl) ReportPaymentFail(paymentID uint64, rt *route.Route,
 	failureSourceIdx *int, failure lnwire.FailureMessage) (bool,
 	channeldb.FailureReason) {
 
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -147,6 +147,17 @@ type MissionControlChannelSnapshot struct {
 	SuccessProb float64
 }
 
+// paymentResult is the information that becomes available when a payment
+// attempt completes.
+type paymentResult struct {
+	id                 uint64
+	timeFwd, timeReply time.Time
+	route              *route.Route
+	success            bool
+	failureSourceIdx   *int
+	failure            lnwire.FailureMessage
+}
+
 // NewMissionControl returns a new instance of missionControl.
 func NewMissionControl(cfg *MissionControlConfig) *MissionControl {
 	log.Debugf("Instantiating mission control with config: "+
@@ -399,25 +410,33 @@ func (m *MissionControl) ReportPaymentFail(paymentID uint64, rt *route.Route,
 
 	timestamp := m.now()
 
+	// TODO(joostjager): Use actual payment initiation time for timeFwd.
+	result := &paymentResult{
+		success:          false,
+		timeFwd:          timestamp,
+		timeReply:        timestamp,
+		id:               paymentID,
+		failureSourceIdx: failureSourceIdx,
+		failure:          failure,
+		route:            rt,
+	}
+
 	// Apply result to update mission control state.
-	return m.applyPaymentResult(
-		timestamp, paymentID, rt, failureSourceIdx, failure,
-	)
+	return m.applyPaymentResult(result)
 }
 
 // applyPaymentResult applies a payment result as input for future probability
 // estimates. It returns a bool indicating whether this error is a final error
 // and no further payment attempts need to be made.
-func (m *MissionControl) applyPaymentResult(timeReply time.Time,
-	paymentID uint64, rt *route.Route, failureSourceIdx *int,
-	failureMessage lnwire.FailureMessage) (bool, channeldb.FailureReason) {
+func (m *MissionControl) applyPaymentResult(result *paymentResult) (
+	bool, channeldb.FailureReason) {
 
 	var (
 		failureSourceIdxInt int
 		failure             lnwire.FailureMessage
 	)
 
-	if failureSourceIdx == nil {
+	if result.failureSourceIdx == nil {
 		// If the failure message could not be decrypted, attribute the
 		// failure to our own outgoing channel.
 		//
@@ -425,24 +444,24 @@ func (m *MissionControl) applyPaymentResult(timeReply time.Time,
 		failureSourceIdxInt = 0
 		failure = lnwire.NewTemporaryChannelFailure(nil)
 	} else {
-		failureSourceIdxInt = *failureSourceIdx
-		failure = failureMessage
+		failureSourceIdxInt = *result.failureSourceIdx
+		failure = result.failure
 	}
 
 	var failureVertex route.Vertex
 
 	if failureSourceIdxInt > 0 {
-		failureVertex = rt.Hops[failureSourceIdxInt-1].PubKeyBytes
+		failureVertex = result.route.Hops[failureSourceIdxInt-1].PubKeyBytes
 	} else {
-		failureVertex = rt.SourcePubKey
+		failureVertex = result.route.SourcePubKey
 	}
 	log.Tracef("Node %x (index %v) reported failure when sending htlc",
-		failureVertex, failureSourceIdx)
+		failureVertex, result.failureSourceIdx)
 
 	// Always determine chan id ourselves, because a channel update with id
 	// may not be available.
 	failedEdge, failedAmt := getFailedEdge(
-		rt, failureSourceIdxInt,
+		result.route, failureSourceIdxInt,
 	)
 
 	switch failure.(type) {
@@ -501,7 +520,7 @@ func (m *MissionControl) applyPaymentResult(timeReply time.Time,
 	// that sent us this error, as it doesn't now what the
 	// correct block height is.
 	case *lnwire.FailExpiryTooSoon:
-		m.reportVertexFailure(timeReply, failureVertex)
+		m.reportVertexFailure(result.timeReply, failureVertex)
 		return false, 0
 
 	// If we hit an instance of onion payload corruption or an invalid
@@ -521,49 +540,49 @@ func (m *MissionControl) applyPaymentResult(timeReply time.Time,
 	// amount, we'll apply the new minimum amount and retry
 	// routing.
 	case *lnwire.FailAmountBelowMinimum:
-		m.reportEdgePolicyFailure(timeReply, failedEdge)
+		m.reportEdgePolicyFailure(result.timeReply, failedEdge)
 		return false, 0
 
 	// If we get a failure due to a fee, we'll apply the
 	// new fee update, and retry our attempt using the
 	// newly updated fees.
 	case *lnwire.FailFeeInsufficient:
-		m.reportEdgePolicyFailure(timeReply, failedEdge)
+		m.reportEdgePolicyFailure(result.timeReply, failedEdge)
 		return false, 0
 
 	// If we get the failure for an intermediate node that
 	// disagrees with our time lock values, then we'll
 	// apply the new delta value and try it once more.
 	case *lnwire.FailIncorrectCltvExpiry:
-		m.reportEdgePolicyFailure(timeReply, failedEdge)
+		m.reportEdgePolicyFailure(result.timeReply, failedEdge)
 		return false, 0
 
 	// The outgoing channel that this node was meant to
 	// forward one is currently disabled, so we'll apply
 	// the update and continue.
 	case *lnwire.FailChannelDisabled:
-		m.reportEdgeFailure(timeReply, failedEdge, 0)
+		m.reportEdgeFailure(result.timeReply, failedEdge, 0)
 		return false, 0
 
 	// It's likely that the outgoing channel didn't have
 	// sufficient capacity, so we'll prune this edge for
 	// now, and continue onwards with our path finding.
 	case *lnwire.FailTemporaryChannelFailure:
-		m.reportEdgeFailure(timeReply, failedEdge, failedAmt)
+		m.reportEdgeFailure(result.timeReply, failedEdge, failedAmt)
 		return false, 0
 
 	// If the send fail due to a node not having the
 	// required features, then we'll note this error and
 	// continue.
 	case *lnwire.FailRequiredNodeFeatureMissing:
-		m.reportVertexFailure(timeReply, failureVertex)
+		m.reportVertexFailure(result.timeReply, failureVertex)
 		return false, 0
 
 	// If the send fail due to a node not having the
 	// required features, then we'll note this error and
 	// continue.
 	case *lnwire.FailRequiredChannelFeatureMissing:
-		m.reportVertexFailure(timeReply, failureVertex)
+		m.reportVertexFailure(result.timeReply, failureVertex)
 		return false, 0
 
 	// If the next hop in the route wasn't known or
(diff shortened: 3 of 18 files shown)


joostjager:
Reminder: extract error source index as an int, in case a node appears in a route twice.

joostjager:
note to self: handle private channels

joostjager:
Ran a small performance test to check startup time impact of rebuilding mission control state. On my machine reconstructing from 1000 historical payment results took 15 ms.

Roasbeef:
No longer WIP and ready to review?

joostjager:
wip label removed.

routing/payment_session.go
```diff
@@ -16,27 +16,6 @@ type PaymentSession interface {
 	// specified HTLC payment to the target node.
 	RequestRoute(payment *LightningPayment,
 		height uint32, finalCltvDelta uint16) (*route.Route, error)
-
```
Roasbeef: 👍 clean separation of concerns

routing/missioncontrol.go
```diff
@@ -369,3 +369,215 @@ func (m *MissionControl) GetHistorySnapshot() *MissionControlSnapshot {
...
+	var (
+		failureVertex route.Vertex
+	)
+	if failureSourceIdx > 0 {
+		failureVertex = rt.Hops[failureSourceIdx-1].PubKeyBytes
+	} else {
+		failureVertex = rt.SourcePubKey
+	}
+
+	// Always determine chan id ourselves, because a channel
```
Roasbeef: All the line wrapping in this method looks like it can be updated to fill out each line more.

routing/missioncontrol.go
```diff
@@ -373,7 +373,7 @@ func (m *MissionControl) GetHistorySnapshot() *MissionControlSnapshot {
 // reportPaymentFail reports a failed payment to mission control as input for
 // future probability estimates. It returns a bool indicating whether this error
 // is a final error and no further payment attempts need to be made.
-func (m *MissionControl) reportPaymentFail(rt *route.Route,
+func (m *MissionControl) reportPaymentFail(paymentID uint64, rt *route.Route,
```
Roasbeef: Why does it need the payment ID? Atm this is a lower level identifier that shouldn't really impact the operation of mission control.

routing/missioncontrol.go
```diff
@@ -126,6 +126,17 @@ type MissionControlChannelSnapshot struct {
 	SuccessProb float64
 }
 
+// paymentResult is the information that becomes available when a payment
```
Roasbeef: Commit title+body are pretty bare, just mentions payment results just then makes independent changes to other methods.

routing/missioncontrol.go
```diff
@@ -18,6 +19,9 @@ const (
 	// minSecondChanceInterval is the minimum time required between
 	// second-chance failures.
 	minSecondChanceInterval = time.Minute
+
+	// DefaultMaxMcHistory is the default maximum history size.
```
Roasbeef: Does this include failed attempts as well?

lnrpc/routerrpc/config.go
```diff
@@ -25,4 +25,8 @@ type RoutingConfig struct {
 	// AprioriHopProbability is the assumed success probability of a hop in
 	// a route when no other information is available.
 	AprioriHopProbability float64
+
+	// MaxMcHistory defines the maximum number of payment results that
+	// are held on disk by mission control.
+	MaxMcHistory int `long:"maxmchistory" description:"the maximum number of payment results that are held on disk by mission control"`
```
Roasbeef: Also this struct has many fields that are duplicated below.

routing/missioncontrol.go
```diff
@@ -138,29 +148,67 @@ type paymentResult struct {
...
+	if err := mc.init(); err != nil {
+		return nil, err
+	}
+
+	return mc, nil
+}
+
+// init initializes mission control with historical data.
+func (m *MissionControl) init() error {
+	log.Debugf("Mission control state reconstruction started")
```
Roasbeef: Perhaps log timing information on the `Tracef` log level along with the number of entries loaded? I think we can even just do this asynchronously, blocking other prospective callers to mission control if their requests arrive before we've finished loading all the prior history.

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,232 @@
...
+		if err := tx.DeleteBucket(resultsKey); err != nil {
+			return err
+		}
+
+		_, err := tx.CreateBucket(resultsKey)
+		return err
+	})
+}
+
+// fetch returns all results currently stored in the database.
```
Roasbeef: `fetchAll`? 

`retrieveAll`? 

`loadAll`?

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,232 @@
...
+func (b *missionControlStore) AddResult(rp *paymentResult) error {
+	return b.db.Update(func(tx *bbolt.Tx) error {
+		bucket := tx.Bucket(resultsKey)
+
+		// Prune oldest entries.
+		if b.maxRecords > 0 {
+			count := bucket.Stats().KeyN
+			for count >= b.maxRecords {
+				cursor := bucket.Cursor()
+				cursor.First()
```
Roasbeef: Errors not checked.

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,232 @@
...
+			&b,
+			uint64(rp.timeFwd.UnixNano()),
+			uint64(rp.timeReply.UnixNano()),
+			rp.success, dbFailureSourceIdx,
+		)
+		if err != nil {
+			return err
+		}
+
+		// Write route.
```
Roasbeef: [No comments are better than these bare ones which don't add any additional context or commentary.](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#43-code-documentation-and-commenting)

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,232 @@
...
+		if err := channeldb.SerializeRoute(&b, *rp.route); err != nil {
+			return err
+		}
+
+		// Write failure.
```
Roasbeef: Possibly use `WriteVarBytes` here instead? The failure will just be blank upon read if nothing was specified.

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		// Put into results bucket.
+		return bucket.Put(getKey(rp), b.Bytes())
+	})
+}
+
+// idToKey returns a byte slice representing a unique key for this payment
+// attempt.
+func getKey(rp *paymentResult) []byte {
```
Roasbeef: Mismatched function name and `godoc` comment. Also `getKey` is pretty vague.

joostjager:
> IMO the scope of this PR doesn't match the title nor the body. Reading the title I'd expect the bulk of the diff to be focused on persistence and resurrection logic. Instead, this PR contains a number of changes, many of which can stand as isolated PRs to be evaluated independently.

Yes, there is definitely a lot of preparation before the actual persistence is added. I needed to have all mission control logic (second chance logic, failure attribution) in one place to be able to record and playback payment results. 

I can take the last commit `routing: persist mission control data` off if you want?

lnrpc/routerrpc/config.go
```diff
@@ -0,0 +1,28 @@
+package routerrpc
+
+import (
+	"time"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// RoutingConfig contains the configurable parameters that control routing.
+type RoutingConfig struct {
```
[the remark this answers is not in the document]
joostjager: It would be a circular dependency.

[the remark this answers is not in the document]
joostjager: Ah no, that is not the reason. There is no circular dependency.

The reason that it is here, is because it is a helper struct that only exists to deal with the conditional compilation and different variants of the sub server config struct. Without conditional compilation, we could have just taken those values from `routerrpc.Config`. 

I have expressed my opinion on conditional compilation multiple times before. In short, I advocate for dropping conditional compilation and clearly marking experimental rpc calls as experimental in the docs. 

This code and discussion is again something that for me counts towards unnecessary complications without upside.

lnrpc/routerrpc/config_active.go
```diff
@@ -72,10 +72,9 @@ func DefaultConfig() *Config {
 	}
 }
 
-// GetMissionControlConfig returns the mission control config based on this sub
-// server config.
-func GetMissionControlConfig(cfg *Config) *routing.MissionControlConfig {
-	return &routing.MissionControlConfig{
+// GetRoutingConfig returns the routing config based on this sub server config.
```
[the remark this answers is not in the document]
joostjager: Do you mean the command line parameters? They are here because they also will be here eventually when there is no more main rpc.

routing/payment_session_source.go
```diff
@@ -0,0 +1,150 @@
...
+	SelfNode *channeldb.LightningNode
+
+	// MissionControl is a shared memory of sorts that executions of payment
+	// path finding use in order to remember which vertexes/edges were
+	// pruned from prior attempts. During payment execution, errors sent by
+	// nodes are mapped into a vertex or edge to be pruned. Each run will
+	// then take into account this set of pruned vertexes/edges to reduce
+	// route failure and pass on graph information gained to the next
+	// execution.
+	MissionControl missionControlInterface
```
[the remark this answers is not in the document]
joostjager: Is there a reason to make it public? There is no use of it outside of `routing`.

routing/router.go
```diff
@@ -171,6 +171,18 @@ type PaymentSessionSource interface {
 	NewPaymentSessionEmpty() PaymentSession
 }
 
+// missionControlInterface is an interface that exposes failure reporting and
```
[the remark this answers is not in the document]
joostjager: Yes, I know. I didn't succeed in coming up with different names for mission control the implementation and mission control the interface. What would you suggest?

routing/router_test.go
```diff
@@ -263,7 +263,7 @@ func TestSendPaymentRouteFailureFallback(t *testing.T) {
 	var payHash [32]byte
 	paymentAmt := lnwire.NewMSatFromSatoshis(1000)
 	payment := LightningPayment{
-		Target:      ctx.aliases["luoji"],
+		Target:      ctx.aliases["sophon"],
```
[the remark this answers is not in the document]
joostjager: I am aware of these cases. But I still think we shouldn't rely on mission control to handle them. In my model of how this should work, such an attempt will fail and by the time we start another path finding round, the local channel data is updated and that channel won't be used anymore.

routing/missioncontrol.go
```diff
@@ -16,6 +14,10 @@ const (
 	// half-life duration defines after how much time a penalized node or
 	// channel is back at 50% probability.
 	DefaultPenaltyHalfLife = time.Hour
+
+	// minSecondChanceInterval is the minimum time required between
+	// second-chance failures.
```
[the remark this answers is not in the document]
joostjager: Added

routing/missioncontrol.go
```diff
@@ -209,6 +216,39 @@ func (m *MissionControl) getEdgeProbabilityForNode(nodeHistory *nodeHistory,
...
+// at providing a channel update for its channel with toNode.
+func (m *MissionControl) requestSecondChance(timestamp time.Time,
+	fromNode, toNode route.Vertex) bool {
+
+	// Look up previous second chance time.
+	pair := DirectedNodePair{
+		From: fromNode,
+		To:   toNode,
+	}
+	lastSecondChance := m.lastSecondChance[pair]
```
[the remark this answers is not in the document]
joostjager: fixed

routing/missioncontrol.go
```diff
@@ -209,6 +216,39 @@ func (m *MissionControl) getEdgeProbabilityForNode(nodeHistory *nodeHistory,
...
+
+		log.Debugf("Second chance granted for %v->%v", fromNode, toNode)
+
+		return true
+	}
+
+	// Otherwise penalize the channel, because we don't allow channel
+	// updates that are that frequent. This is to prevent nodes from keeping
+	// us busy by continuously sending new channel updates.
+	log.Debugf("Second chance denied for %v->%v", fromNode, toNode)
```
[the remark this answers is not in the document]
joostjager: Added

routing/missioncontrol.go
joostjager: Fixed

routing/missioncontrol.go
joostjager: It is to prevent the same payment result from being recorded multiple times because of `lnd` restarts. It makes the call idempotent.

routing/missioncontrol.go
joostjager: Commit split in two

lnrpc/routerrpc/config.go
joostjager: This is again a consequence of the conditional compilation. Suggestions to remove duplication are welcome.

routing/missioncontrol.go
joostjager: In this PR we only store failed attempts.

routing/missioncontrol.go
joostjager: Log added. Kept it on debug level as that felt more in line with other log statements.

Async loading is definitely a good idea, but I wanted to do that outside of this PR.

routing/missioncontrol_store.go
joostjager: `fetchAll`

routing/missioncontrol_store.go
joostjager: Check added for `Delete()`

routing/missioncontrol_store.go
joostjager: I see these as 'chapters'. For `SerializeRoute` in isolation I wouldn't have added it, but leaving it out in this context makes it less readable imo.

routing/missioncontrol_store.go
joostjager: Converted to `WriteVarBytes`. I also discovered that I was serializing data that included length and padding. Created a separate commit that defines slimmer encode/decode functions in `lnwire`.

routing/missioncontrol_store.go
joostjager: Fixed and renamed

joostjager:
@roasbeef ptal

joostjager:
Another reason why it may be good to take the persistence commit off for the first merge is to test the other changes first and see if that somehow triggers a change in the db structure.

routing/router.go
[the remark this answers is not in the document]
Roasbeef: `RouteMemory`? lol 

or `PathfindingMemory` 

`RoutingCommand` 

`PaymentHistory`

lnrpc/routerrpc/config.go
Roasbeef: Create a struct with the shared fields, and embed it into both of them.

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,235 @@
...
+			return err
+		}
+
+		_, err := tx.CreateBucket(resultsKey)
+		return err
+	})
+}
+
+// fetch returns all results currently stored in the database.
+func (b *missionControlStore) fetchAll() ([]*paymentResult, error) {
```
Roasbeef: Method name out of sync with the `godoc` comment.

routing/missioncontrol_store.go
Roasbeef: Don't think this one in particular adds anything but an extra line in the diff.

routing/payment_session_source.go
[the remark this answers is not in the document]
joostjager: Exported in #3298

routing/router.go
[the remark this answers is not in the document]
joostjager: `MissionController` it is now.

routing/missioncontrol_store.go
joostjager: fixed

routing/missioncontrol_store.go
joostjager: I still like it the way it is, but it is unimportant. Merged comment with one above.

lnrpc/routerrpc/config.go
joostjager: Shared struct created.

routing/router.go
```diff
@@ -1873,38 +1873,40 @@ func (r *ChannelRouter) tryApplyChannelUpdate(rt *route.Route,
...
+			)
+			if err != nil {
+				return true, channeldb.FailureReasonError
+			}
 		}
+
+		log.Tracef("node=%v reported failure when sending htlc",
+			*failureSourceIdx)
+	} else {
+		log.Tracef("unreadable failure when sending htlc")
```
cfromknecht: nit: log mess should be capitalized

routing/router.go
```diff
@@ -1873,38 +1873,40 @@ func (r *ChannelRouter) tryApplyChannelUpdate(rt *route.Route,
...
+		if failureMessage != nil {
+			err := r.tryApplyChannelUpdate(
+				rt, *failureSourceIdx, failureMessage,
+			)
+			if err != nil {
+				return true, channeldb.FailureReasonError
+			}
 		}
+
+		log.Tracef("node=%v reported failure when sending htlc",
```
cfromknecht: capitalization

lnrpc/routerrpc/config.go
```diff
@@ -3,26 +3,25 @@ package routerrpc
...
+	AprioriHopProbability float64 `long:"apriorihopprob" description:"Assumed success probability of a hop in a route when no other information is available."`
+
+	// PenaltyHalfLife defines after how much time a penalized node or
+	// channel is back at 50% probability.
+	PenaltyHalfLife time.Duration `long:"penaltyhalflife" description:"Defines the duration after which a penalized node or channel is back at 50% probability"`
+
+	// AttemptCost is the virtual cost in path finding weight units of
+	// executing a payment attempt that fails. It is used to trade off
+	// potentially better routes against their probability of succeeding.
+	AttemptCost btcutil.Amount `long:"attemptcost" description:"The (virtual) cost in sats of a failed payment attempt"`
```
cfromknecht: what's the rationale for changing this to sats, when we later convert it back to msats? this will make anyone's current config values be off by a factor of 1000 after updating

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// AddResult adds a new result to the db.
+func (b *missionControlStore) AddResult(rp *paymentResult) error {
+	return b.db.Update(func(tx *bbolt.Tx) error {
+		bucket := tx.Bucket(resultsKey)
+
+		// Prune oldest entries.
+		if b.maxRecords > 0 {
+			count := bucket.Stats().KeyN
```
cfromknecht: we've had issues in the past using `Stats()`, perhaps instead we can count the number of entries in the db on startup and track the current count in memory?

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,235 @@
...
+	)
+
+	err := channeldb.ReadElements(
+		r, &timeFwd, &timeReply, &result.success, &dbFailureSourceIdx,
+	)
+	if err != nil {
+		return nil, err
+	}
+	result.timeFwd = time.Unix(0, int64(timeFwd)).UTC()
+	result.timeReply = time.Unix(0, int64(timeReply)).UTC()
```
cfromknecht: do we use this `UTC()` conversion anywhere else in the codebase?

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,235 @@
...
+		// Write failure. If there is no failure message, write an empty
+		// byte slice.
+		var failureBytes bytes.Buffer
+		if rp.failure != nil {
+			err := lnwire.EncodeFailureMessage(&failureBytes, rp.failure, 0)
+			if err != nil {
+				return err
+			}
+		}
+		err = wire.WriteVarBytes(&b, 0, failureBytes.Bytes())
```
cfromknecht: definitely not a blocker, but could be nice to have an equivalent `serializeResult` method so that's it's easier to examine to encoding and decoding logic side-by-side

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,235 @@
...
+	// stored.
+	resultsKey = []byte("missioncontrol-results")
+
+	// Big endian is the preferred byte order, due to cursor scans over
+	// integer keys iterating in order.
+	byteOrder = binary.BigEndian
+
+	// unknownFailureSourceIdx is the database encoding of an unknown error
+	// source.
+	unknownFailureSourceIdx = -1
```
cfromknecht: where is this used?

routing/missioncontrol.go
```diff
@@ -1,6 +1,7 @@
 package routing
 
 import (
+	"github.com/lightningnetwork/lnd/channeldb"
```
halseth: import order.

routing/router.go
```diff
@@ -1909,195 +1907,9 @@ func (r *ChannelRouter) processSendError(rt *route.Route, sendErr error) (
...
-			channel: failedEdge.channel,
-		}, 0)
-		return false, 0
-
-	// Any other failure or an empty failure will get the node pruned.
-	default:
-		r.cfg.MissionControl.ReportVertexFailure(failureVertex)
-		return false, 0
-	}
+	return r.cfg.MissionControl.ReportPaymentFail(
```
halseth: Nice 👍

routing/missioncontrol.go
```diff
@@ -393,26 +393,30 @@ func (m *MissionControl) GetHistorySnapshot() *MissionControlSnapshot {
 // future probability estimates. It returns a bool indicating whether this error
 // is a final error and no further payment attempts need to be made.
 func (m *MissionControl) ReportPaymentFail(rt *route.Route,
-	failureSourceIdx int, failure lnwire.FailureMessage) (bool,
+	failureSourceIdx *int, failure lnwire.FailureMessage) (bool,
```
halseth: the pointer argument needs godoc

routing/missioncontrol.go
```diff
@@ -393,26 +393,30 @@ func (m *MissionControl) GetHistorySnapshot() *MissionControlSnapshot {
...
 	channeldb.FailureReason) {
 
-	var (
-		failureVertex route.Vertex
-	)
+	var failureSourceIdxInt int
+	if failureSourceIdx == nil {
```
halseth: this needs explanation/comments

routing/router.go
```diff
@@ -1873,38 +1873,40 @@ func (r *ChannelRouter) tryApplyChannelUpdate(rt *route.Route,
...
+		failureMessage   lnwire.FailureMessage
+		failureSourceIdx *int
+	)
 
-	// If an internal, non-forwarding error occurred, we can stop trying.
-	fErr, ok := sendErr.(*htlcswitch.ForwardingError)
-	if !ok {
-		return true, channeldb.FailureReasonError
-	}
+	if sendErr != htlcswitch.ErrUnreadableFailureMessage {
```
halseth: would this be easier to follow using a switch statement?

routing/router.go
```diff
@@ -178,7 +178,7 @@ type MissionController interface {
 	// input for future probability estimates. It returns a bool indicating
 	// whether this error is a final error and no further payment attempts
 	// need to be made.
-	ReportPaymentFail(rt *route.Route,
+	ReportPaymentFail(paymentID uint64, rt *route.Route,
```
halseth: what's the purpose of this change?

routing/missioncontrol.go
```diff
@@ -307,32 +307,30 @@ func (m *MissionControl) reportEdgePolicyFailure(failedEdge edge) {
...
 	history := m.createHistoryIfNotExists(failedEdge.from)
-	history.lastFail = &now
+	history.lastFail = &timestamp
 }
 
 // reportEdgeFailure reports a channel level failure.
 //
 // TODO(roasbeef): also add value attempted to send and capacity of channel
-func (m *MissionControl) reportEdgeFailure(failedEdge edge,
+func (m *MissionControl) reportEdgeFailure(timestamp time.Time, failedEdge edge,
```
halseth: maybe we could add timestamp to the `edge` struct?

lnrpc/routerrpc/config.go
joostjager: It was already in sats. The diff may be a bit confusing, because the command line options struct and the internal config struct are unified in this pr.

routing/missioncontrol_store.go
joostjager: I did a quick find and it seems we don't. The reason I do this is to get UTC timezones for log messages that contain these times.

routing/router.go
joostjager: To make reporting to mission control idempotent. Without this id, it wouldn't be able to recognize when it is offered the same result again after restart.

joostjager:
> Since persisting mission control data is a big change, I would suggesting breaking this up into two or more PRs:
> 
> 1. Simple changes (exporting Decode/Serialization methods, embedding configs)
> 2. Changing params passed to mission control, and interfaces
> 3. Writing data to disk

We could move the last commit that finally adds the persistence to its own pr. But not sure if it makes sense at this stage of the review process.

routing/missioncontrol.go
joostjager: more godoc added explaining the ptr

routing/missioncontrol.go
joostjager: comments got lost when I moved this code. readded

routing/router.go
joostjager: not sure exactly how to convert to a switch, but i took another look at this part and simplified it in another way.

routing/missioncontrol.go
joostjager: I prefer to keep them separately. The `edge` is instantiated in `getFailedEdge` and this function has no idea of time.

routing/missioncontrol_store.go
joostjager: it should have been used instead of constants in the code. Fixed.

routing/missioncontrol_store.go
joostjager: Do you have more details on what issues those were? We can work around it, but it would be nice to have some confirmation that `Stats().KeyN` is indeed unreliable.

routing/missioncontrol_store.go
joostjager: added

routing/missioncontrol_store.go
cfromknecht: UTC is is the default https://play.golang.org/p/e_-nuERpM66

routing/missioncontrol_store.go
cfromknecht: it caused a panic in one of our prior migrations:
 * issue: https://github.com/lightningnetwork/lnd/issues/1874#issuecomment-420022023
 * fix: https://github.com/lightningnetwork/lnd/pull/1898

the root cause was never determined, so we've been avoiding `Stats()` like the plague

lnrpc/routerrpc/config.go
cfromknecht: ah i see, thanks for explaining

routing/missioncontrol_store.go
wpaulino: I think that's just because the server being used for the playground is on UTC. Running this locally results in my timezone being shown instead.

cfromknecht: hmm, that's obnoxious.

but still the bigger question is why wouldn't we print this in the user's local timezone? nothing else in our logs are UTCified

joostjager: I assumed we want UTC in order to be able to easily compare/merge logs from different machines without needing to think about physical location. But if we don't do it anywhere else in the code base, we may not want to do it here either.

routing/missioncontrol_store.go
joostjager: Ah, I see. So there is no other way to get a count than iterating through all items with a cursor?

joostjager: Ah, I see. So there is no other way to get a count than iterating through all items with a cursor?

joostjager: The issue that you mention, could it be that it was caused by a database that was already corrupt?

cfromknecht: > Ah, I see. So there is no other way to get a count than iterating through all items with a cursor?

yeah that's the simplest way

> The issue that you mention, could it be that it was caused by a database that was already corrupt?

we saw it happen on perfectly good dbs. once the fix landed, people were able to resume w/o issue afaiaa

routing/missioncontrol_store.go
joostjager: UTC removed

routing/missioncontrol_store.go
joostjager: `Stats()` call removed. Small complication because store needs to be idempotent, but solved that with an additional bucket get.

How did you actually know these were perfectly good dbs?

routing/router_test.go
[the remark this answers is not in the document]
joostjager: Although this change is part of a commit that is now removed from the pr, I found a case where we are unfortunately still forced to penalize local channels.

When the error source is unknown, it may or may not be our peer. In that case we can only penalize all channels along the route including our own.

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,271 @@
...
+	})
+}
+
+// fetchAll returns all results currently stored in the database.
+func (b *missionControlStore) fetchAll() ([]*paymentResult, error) {
+	var results []*paymentResult
+
+	err := b.db.View(func(tx *bbolt.Tx) error {
+		resultBucket := tx.Bucket(resultsKey)
+		results = make([]*paymentResult, resultBucket.Stats().KeyN)
```
cfromknecht: still calling `Stats().KeyN` here. i think it's fine if we just don't allocate it up front and append. if this isn't used outside of testing then it shouldn't be a performance issue

routing/missioncontrol_store.go
```diff
@@ -0,0 +1,271 @@
...
+func (b *missionControlStore) AddResult(rp *paymentResult) error {
+	return b.db.Update(func(tx *bbolt.Tx) error {
+		bucket := tx.Bucket(resultsKey)
+
+		// Prune oldest entries.
+		if b.maxRecords > 0 {
+			for b.numRecords >= b.maxRecords {
+				cursor := bucket.Cursor()
+				cursor.First()
+				if err := cursor.Delete(); err != nil {
```
cfromknecht: in the past i think we observed a performance issue in calling `First()` then `Delete()` repetitively, since it actually results in a quadratic operation. fortunately i think we only expect this to be called once here, but something to keep in mind in the future. @Roasbeef was this not one of the root causes in that migration performance issue?

routing/missioncontrol_store.go
joostjager: Yes my mistake. I had fixed that already on a different computer, but hadn't pushed.

routing/missioncontrol_store.go
joostjager: Ok interesting. Yes, it is expected to be called only once.

Roasbeef: IIRC, in that case, we would have repeated calls to `Delete()`, then `First()`, as `Delete` would move the cursor one record after the deleted key. We ended up resolving it by using a two pass approach: https://github.com/lightningnetwork/lnd/commit/974ec02d249e6a77712e24233692e4e08c8bf31b#diff-99f3013528876f55a0e46ff7014e8a19

PR #3165, opened 2019-06-05
[funding+autopilot] Make the funding manager and autopilot fee aware

### LINKED ISSUE #986: Autopilot can create impossible channel requests, which fail due to occupying funds without room for transaction fees

halseth:
This is an alternative to #3017.

Since the funding manager and wallet can't know up front the how much it will cost us in fees to fund a channel, it will always cost us more than `chanSize` in total. This makes it hard to spend all the funds remaining in the wallet into a channel, since we won't have enough to cover the additional fee.

This PR attempts to fix this by letting one specify a "spend amount" instead of a "channel size" when opening a channel. This makes the wallet do coin selection that results in a decrease in balance equal to this spend amount, instead of spending `chanSize+fees`.

We make the autopilot use this new option, making it better at staying within its budget, and also able to open channels that utilizes the whole wallet balance.

Fixes #986

### DIFF
--- a/autopilot/agent.go
+++ b/autopilot/agent.go
@@ -584,6 +584,7 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 
 	// Use the heuristic to calculate a score for each node in the
 	// graph.
+	log.Debugf("Scoring %d nodes for chan_size=%v", len(nodes), chanSize)
 	scores, err := a.cfg.Heuristic.NodeScores(
 		a.cfg.Graph, totalChans, chanSize, nodes,
 	)
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -1084,16 +1084,16 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	// funds to the channel ourselves.
 	chainHash := chainhash.Hash(msg.ChainHash)
 	req := &lnwallet.InitFundingReserveMsg{
-		ChainHash:       &chainHash,
-		NodeID:          fmsg.peer.IdentityKey(),
-		NodeAddr:        fmsg.peer.Address(),
-		FundingAmount:   0,
-		Capacity:        amt,
-		CommitFeePerKw:  lnwallet.SatPerKWeight(msg.FeePerKiloWeight),
-		FundingFeePerKw: 0,
-		PushMSat:        msg.PushAmount,
-		Flags:           msg.ChannelFlags,
-		MinConfs:        1,
+		ChainHash:        &chainHash,
+		NodeID:           fmsg.peer.IdentityKey(),
+		NodeAddr:         fmsg.peer.Address(),
+		LocalFundingAmt:  0,
+		RemoteFundingAmt: amt,
+		CommitFeePerKw:   lnwallet.SatPerKWeight(msg.FeePerKiloWeight),
+		FundingFeePerKw:  0,
+		PushMSat:         msg.PushAmount,
+		Flags:            msg.ChannelFlags,
+		MinConfs:         1,
 	}
 
 	reservation, err := f.cfg.Wallet.InitChannelReservation(req)
@@ -2783,16 +2783,16 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	// wallet doesn't have enough funds to commit to this channel, then the
 	// request will fail, and be aborted.
 	req := &lnwallet.InitFundingReserveMsg{
-		ChainHash:       &msg.chainHash,
-		NodeID:          peerKey,
-		NodeAddr:        msg.peer.Address(),
-		FundingAmount:   localAmt,
-		Capacity:        capacity,
-		CommitFeePerKw:  commitFeePerKw,
-		FundingFeePerKw: msg.fundingFeePerKw,
-		PushMSat:        msg.pushAmt,
-		Flags:           channelFlags,
-		MinConfs:        msg.minConfs,
+		ChainHash:        &msg.chainHash,
+		NodeID:           peerKey,
+		NodeAddr:         msg.peer.Address(),
+		LocalFundingAmt:  localAmt,
+		RemoteFundingAmt: 0,
+		CommitFeePerKw:   commitFeePerKw,
+		FundingFeePerKw:  msg.fundingFeePerKw,
+		PushMSat:         msg.pushAmt,
+		Flags:            channelFlags,
+		MinConfs:         msg.minConfs,
 	}
 
 	reservation, err := f.cfg.Wallet.InitChannelReservation(req)
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -337,6 +337,14 @@ type fundingConfig struct {
 	// due to fees.
 	MinChanSize btcutil.Amount
 
+	// MaxPendingChannels is the maximum number of pending channels we
+	// allow for each peer.
+	MaxPendingChannels int
+
+	// RejectPush is set true if the fundingmanager should reject any
+	// incoming channels having a non-zero push amount.
+	RejectPush bool
+
 	// NotifyOpenChannelEvent informs the ChannelNotifier when channels
 	// transition from pending open to open.
 	NotifyOpenChannelEvent func(wire.OutPoint)
@@ -1012,7 +1020,7 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 
 	// TODO(roasbeef): modify to only accept a _single_ pending channel per
 	// block unless white listed
-	if numPending >= cfg.MaxPendingChannels {
+	if numPending >= f.cfg.MaxPendingChannels {
 		f.failFundingFlow(
 			fmsg.peer, fmsg.msg.PendingChannelID,
 			lnwire.ErrMaxPendingChannels,
@@ -1057,7 +1065,7 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 
 	// If request specifies non-zero push amount and 'rejectpush' is set,
 	// signal an error.
-	if cfg.RejectPush && msg.PushAmount > 0 {
+	if f.cfg.RejectPush && msg.PushAmount > 0 {
 		f.failFundingFlow(
 			fmsg.peer, fmsg.msg.PendingChannelID,
 			lnwallet.ErrNonZeroPushAmount())
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -2742,8 +2742,7 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	var (
 		peerKey        = msg.peer.IdentityKey()
 		localAmt       = msg.localFundingAmt
-		remoteAmt      = msg.remoteFundingAmt
-		capacity       = localAmt + remoteAmt
+		capacity       = localAmt
 		minHtlc        = msg.minHtlc
 		remoteCsvDelay = msg.remoteCsvDelay
 	)
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -1722,8 +1722,8 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 			return
 		case shortChanID, ok = <-confChan:
 			if !ok {
-				fndgLog.Errorf("waiting for funding confirmation" +
-					" failed")
+				fndgLog.Errorf("waiting for funding " +
+					"confirmation failed")
 				return
 			}
 		}
@@ -1893,8 +1893,9 @@ func makeFundingScript(channel *channeldb.OpenChannel) ([]byte, error) {
 // when a channel has become active for lightning transactions.
 // The wait can be canceled by closing the cancelChan. In case of success,
 // a *lnwire.ShortChannelID will be passed to confChan.
-func (f *fundingManager) waitForFundingConfirmation(completeChan *channeldb.OpenChannel,
-	cancelChan <-chan struct{}, confChan chan<- *lnwire.ShortChannelID) {
+func (f *fundingManager) waitForFundingConfirmation(
+	completeChan *channeldb.OpenChannel, cancelChan <-chan struct{},
+	confChan chan<- *lnwire.ShortChannelID) {
 
 	defer close(confChan)
 
@@ -1904,16 +1905,19 @@ func (f *fundingManager) waitForFundingConfirmation(completeChan *channeldb.Open
 	fundingScript, err := makeFundingScript(completeChan)
 	if err != nil {
 		fndgLog.Errorf("unable to create funding script for "+
-			"ChannelPoint(%v): %v", completeChan.FundingOutpoint, err)
+			"ChannelPoint(%v): %v", completeChan.FundingOutpoint,
+			err)
 		return
 	}
 	numConfs := uint32(completeChan.NumConfsRequired)
 	confNtfn, err := f.cfg.Notifier.RegisterConfirmationsNtfn(
-		&txid, fundingScript, numConfs, completeChan.FundingBroadcastHeight,
+		&txid, fundingScript, numConfs,
+		completeChan.FundingBroadcastHeight,
 	)
 	if err != nil {
 		fndgLog.Errorf("Unable to register for confirmation of "+
-			"ChannelPoint(%v): %v", completeChan.FundingOutpoint, err)
+			"ChannelPoint(%v): %v", completeChan.FundingOutpoint,
+			err)
 		return
 	}
 
@@ -1928,14 +1932,17 @@ func (f *fundingManager) waitForFundingConfirmation(completeChan *channeldb.Open
 	select {
 	case confDetails, ok = <-confNtfn.Confirmed:
 		// fallthrough
+
 	case <-cancelChan:
 		fndgLog.Warnf("canceled waiting for funding confirmation, "+
 			"stopping funding flow for ChannelPoint(%v)",
 			completeChan.FundingOutpoint)
 		return
+
 	case <-f.quit:
 		fndgLog.Warnf("fundingManager shutting down, stopping funding "+
-			"flow for ChannelPoint(%v)", completeChan.FundingOutpoint)
+			"flow for ChannelPoint(%v)",
+			completeChan.FundingOutpoint)
 		return
 	}
 
@@ -1948,6 +1955,18 @@ func (f *fundingManager) waitForFundingConfirmation(completeChan *channeldb.Open
 
 	fundingPoint := completeChan.FundingOutpoint
 	chanID := lnwire.NewChanIDFromOutPoint(&fundingPoint)
+	if int(fundingPoint.Index) >= len(confDetails.Tx.TxOut) {
+		fndgLog.Warnf("Funding point index does not exist for "+
+			"ChannelPoint(%v)", completeChan.FundingOutpoint)
+		return
+	}
+
+	outputAmt := btcutil.Amount(confDetails.Tx.TxOut[fundingPoint.Index].Value)
+	if outputAmt != completeChan.Capacity {
+		fndgLog.Warnf("Invalid output value for ChannelPoint(%v)",
+			completeChan.FundingOutpoint)
+		return
+	}
 
 	fndgLog.Infof("ChannelPoint(%v) is now active: ChannelID(%x)",
 		fundingPoint, chanID[:])
@@ -2742,7 +2761,6 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	var (
 		peerKey        = msg.peer.IdentityKey()
 		localAmt       = msg.localFundingAmt
-		capacity       = localAmt
 		minHtlc        = msg.minHtlc
 		remoteCsvDelay = msg.remoteCsvDelay
 	)
@@ -2756,10 +2774,11 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		ourDustLimit = defaultLitecoinDustLimit
 	}
 
-	fndgLog.Infof("Initiating fundingRequest(localAmt=%v, remoteAmt=%v, "+
-		"capacity=%v, chainhash=%v, peer=%x, dustLimit=%v, min_confs=%v)",
-		localAmt, msg.pushAmt, capacity, msg.chainHash,
-		peerKey.SerializeCompressed(), ourDustLimit, msg.minConfs)
+	fndgLog.Infof("Initiating fundingRequest(local_amt=%v "+
+		"(subtract_fees=%v), push_amt=%v, chain_hash=%v, peer=%x, "+
+		"dust_limit=%v, min_confs=%v)", localAmt, msg.subtractFees,
+		msg.pushAmt, msg.chainHash, peerKey.SerializeCompressed(),
+		ourDustLimit, msg.minConfs)
 
 	// First, we'll query the fee estimator for a fee that should get the
 	// commitment transaction confirmed by the next few blocks (conf target
@@ -2786,6 +2805,7 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		ChainHash:        &msg.chainHash,
 		NodeID:           peerKey,
 		NodeAddr:         msg.peer.Address(),
+		SubtractFees:     msg.subtractFees,
 		LocalFundingAmt:  localAmt,
 		RemoteFundingAmt: 0,
 		CommitFeePerKw:   commitFeePerKw,
@@ -2800,6 +2820,12 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		msg.err <- err
 		return
 	}
+
+	// Now that we have successfully reserved funds for this channel in the
+	// wallet, we can fetch the final channel capacity. This is done at
+	// this point since the final capacity might change in case of
+	// SubtractFees=true.
+	capacity := reservation.Capacity()
 
 	// Obtain a new pending channel ID which is used to track this
 	// reservation throughout its lifetime.
--- a/fundingmanager_test.go
+++ b/fundingmanager_test.go
@@ -236,7 +236,8 @@ func createTestWallet(cdb *channeldb.DB, netParams *chaincfg.Params,
 }
 
 func createTestFundingManager(t *testing.T, privKey *btcec.PrivateKey,
-	addr *lnwire.NetAddress, tempTestDir string) (*testNode, error) {
+	addr *lnwire.NetAddress, tempTestDir string,
+	options ...cfgOption) (*testNode, error) {
 
 	netParams := activeNetParams.Params
 	estimator := lnwallet.NewStaticFeeEstimator(62500, 0)
@@ -282,7 +283,7 @@ func createTestFundingManager(t *testing.T, privKey *btcec.PrivateKey,
 
 	var chanIDSeed [32]byte
 
-	f, err := newFundingManager(fundingConfig{
+	fundingCfg := fundingConfig{
 		IDKey:        privKey.PubKey(),
 		Wallet:       lnw,
 		Notifier:     chainNotifier,
@@ -363,8 +364,15 @@ func createTestFundingManager(t *testing.T, privKey *btcec.PrivateKey,
 		},
 		ZombieSweeperInterval:  1 * time.Hour,
 		ReservationTimeout:     1 * time.Nanosecond,
+		MaxPendingChannels:     DefaultMaxPendingChannels,
 		NotifyOpenChannelEvent: func(wire.OutPoint) {},
-	})
+	}
+
+	for _, op := range options {
+		op(&fundingCfg)
+	}
+
+	f, err := newFundingManager(fundingCfg)
 	if err != nil {
 		t.Fatalf("failed creating fundingManager: %v", err)
 	}
@@ -468,12 +476,10 @@ func recreateAliceFundingManager(t *testing.T, alice *testNode) {
 	}
 }
 
-func setupFundingManagers(t *testing.T, maxPendingChannels int) (*testNode, *testNode) {
-	// We need to set the global config, as fundingManager uses
-	// MaxPendingChannels, and it is usually set in lndMain().
-	cfg = &config{
-		MaxPendingChannels: maxPendingChannels,
-	}
+type cfgOption func(*fundingConfig)
+
+func setupFundingManagers(t *testing.T,
+	options ...cfgOption) (*testNode, *testNode) {
 
 	aliceTestDir, err := ioutil.TempDir("", "alicelnwallet")
 	if err != nil {
@@ -481,7 +487,7 @@ func setupFundingManagers(t *testing.T, maxPendingChannels int) (*testNode, *tes
 	}
 
 	alice, err := createTestFundingManager(
-		t, alicePrivKey, aliceAddr, aliceTestDir,
+		t, alicePrivKey, aliceAddr, aliceTestDir, options...,
 	)
 	if err != nil {
 		t.Fatalf("failed creating fundingManager: %v", err)
@@ -492,7 +498,9 @@ func setupFundingManagers(t *testing.T, maxPendingChannels int) (*testNode, *tes
 		t.Fatalf("unable to create temp directory: %v", err)
 	}
 
-	bob, err := createTestFundingManager(t, bobPrivKey, bobAddr, bobTestDir)
+	bob, err := createTestFundingManager(
+		t, bobPrivKey, bobAddr, bobTestDir, options...,
+	)
 	if err != nil {
 		t.Fatalf("failed creating fundingManager: %v", err)
 	}
@@ -1029,7 +1037,7 @@ func assertHandleFundingLocked(t *testing.T, alice, bob *testNode) {
 }
 
 func TestFundingManagerNormalWorkflow(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// We will consume the channel updates as we go, so no buffering is needed.
@@ -1102,7 +1110,7 @@ func TestFundingManagerNormalWorkflow(t *testing.T) {
 }
 
 func TestFundingManagerRestartBehavior(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// Run through the process of opening the channel, up until the funding
@@ -1240,7 +1248,7 @@ func TestFundingManagerRestartBehavior(t *testing.T) {
 // server to notify when the peer comes online, in case sending the
 // fundingLocked message fails the first time.
 func TestFundingManagerOfflinePeer(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// Run through the process of opening the channel, up until the funding
@@ -1374,7 +1382,7 @@ func TestFundingManagerOfflinePeer(t *testing.T) {
 // will properly clean up a zombie reservation that times out after the
 // initFundingMsg has been handled.
 func TestFundingManagerPeerTimeoutAfterInitFunding(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// We will consume the channel updates as we go, so no buffering is needed.
@@ -1434,7 +1442,7 @@ func TestFundingManagerPeerTimeoutAfterInitFunding(t *testing.T) {
 // will properly clean up a zombie reservation that times out after the
 // fundingOpenMsg has been handled.
 func TestFundingManagerPeerTimeoutAfterFundingOpen(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// We will consume the channel updates as we go, so no buffering is needed.
@@ -1503,7 +1511,7 @@ func TestFundingManagerPeerTimeoutAfterFundingOpen(t *testing.T) {
 // will properly clean up a zombie reservation that times out after the
 // fundingAcceptMsg has been handled.
 func TestFundingManagerPeerTimeoutAfterFundingAccept(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// We will consume the channel updates as we go, so no buffering is needed.
@@ -1577,7 +1585,7 @@ func TestFundingManagerPeerTimeoutAfterFundingAccept(t *testing.T) {
 }
 
 func TestFundingManagerFundingTimeout(t *testing.T) {
-	alice, bob := setupFundingManagers(t, DefaultMaxPendingChannels)
+	alice, bob := setupFundingManagers(t)
 	defer tearDownFundingManagers(t, alice, bob)
 
 	// We will consume the channel updates as we go, so no buffering is needed.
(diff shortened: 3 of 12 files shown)


fundingmanager_test.go
```diff
@@ -467,20 +475,19 @@ func recreateAliceFundingManager(t *testing.T, alice *testNode) {
...
-func setupFundingManagers(t *testing.T, maxPendingChannels int) (*testNode, *testNode) {
-	// We need to set the global config, as fundingManager uses
-	// MaxPendingChannels, and it is usually set in lndMain().
-	cfg = &config{
-		MaxPendingChannels: maxPendingChannels,
-	}
+type cfgOption func(*fundingConfig)
+
+func setupFundingManagers(t *testing.T, maxPendingChannels int,
+	options ...cfgOption) (*testNode, *testNode) {
```
wpaulino: If we're going this route, why not also make `maxPendingChannels` a `cfgOption`?

lnwallet/reservation.go
```diff
@@ -163,7 +163,7 @@ func NewChannelReservation(capacity, fundingAmt btcutil.Amount,
 		// TODO(roasbeef): need to rework fee structure in general and
 		// also when we "unlock" dual funder within the daemon
 
-		if capacity == fundingAmt {
+		if capacity == localFundingAmt {
```
wpaulino: Unrelated to this diff (https://github.com/lightningnetwork/lnd/pull/3165/commits/949783387a3b8f170e362c1edef45a45bb1578e6) but noticed while reviewing the whole method: there's a dust check below and above for 0 when it should really be `2 * DefaultDustLimit()`?

lnwallet/wallet.go
```diff
@@ -484,15 +484,16 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 			req.FundingFeePerKw, req.LocalFundingAmt, req.MinConfs,
-			reservation.ourContribution,
 		)
 		if err != nil {
 			req.err <- err
 			req.resp <- nil
 			return
 		}
+		reservation.ourContribution.Inputs = coins
+		reservation.ourContribution.ChangeOutputs = change
```
wpaulino: Unrelated: any reason this is a slice rather than just one?

lnwallet/wallet.go
```diff
@@ -1302,52 +1304,64 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
...
-		contribution.ChangeOutputs[0] = &wire.TxOut{
+		changeOutputs = make([]*wire.TxOut, 1)
+		changeOutputs[0] = &wire.TxOut{
 			Value:    int64(changeAmt),
 			PkScript: changeScript,
 		}
 	}
 
-	return nil
+	unlock := func() {
```
wpaulino: Seems like we'd want to call this in the error cases above after the inputs are locked.

lnwallet/wallet.go
```diff
@@ -457,8 +457,36 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
+
+	// If we're on the receiving end of a single funder channel then we
+	// don't need to perform any coin selection, and the remote contributes
+	// all funds.
+	capacity := req.RemoteFundingAmt
+	localFundingAmt := btcutil.Amount(0)
+
+	// Otherwise, attempt to obtain enough coins to meet the required
+	// funding amount.
+	if req.LocalFundingAmt != 0 {
```
wpaulino: Add TODO to handle dual funder case?

lnwallet/wallet.go
```diff
@@ -494,66 +495,88 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 		req.PushMSat, l.Cfg.NetParams.GenesisHash, req.Flags,
 	)
 	if err != nil {
+		unlock()
 		req.err <- err
 		req.resp <- nil
 		return
 	}
 
+	err = l.initOurContribution(
```
wpaulino: :+1:

lnwallet/wallet.go
```diff
@@ -432,13 +439,21 @@ func (l *LightningWallet) InitChannelReservation(
...
-	if req.LocalFundingAmt+req.RemoteFundingAmt == 0 {
+	if req.LocalFundingAmt+req.RemoteFundingAmt+req.LocalSpendAmt == 0 {
 		err := ErrZeroCapacity()
 		req.err <- err
 		req.resp <- nil
 		return
 	}
 
+	if req.LocalFundingAmt+req.RemoteFundingAmt > 0 && req.LocalSpendAmt > 0 {
+		err := fmt.Errorf("Local and remote fund amt must be zero " +
```
wpaulino: Nit: errors returned should begin lowercase.

There are some other cases as well.

lnwallet/wallet.go
```diff
@@ -69,6 +69,13 @@ type InitFundingReserveMsg struct {
 	// workflow.
 	NodeAddr net.Addr
 
+	// LocalSpendAmt should be set to the amount we intend to spend when
+	// opening the channel. This can be used for instance to use all our
+	// remaining funds to open the channel, since it will take fees into
+	// account. Can currently only be used when LocalFundingAmt and
+	// RemoteFundingAmt are set to 0.
+	LocalSpendAmt btcutil.Amount
```
wpaulino: WTB enum structs. An alternative to make it clear only one of these can be specified would be something similar to what's shown in https://play.golang.org/p/jfgsCBxb9q.

lnwallet/wallet.go
```diff
@@ -1303,6 +1336,10 @@ func (l *LightningWallet) WithCoinSelectLock(f func() error) error {
 // successful/possible, then the selected coins and change outputs are
 // returned. This method locks the selected outputs, and a function closure to
 // unlock them in case of an error is returned.
+//
+// NOTE: This method is different from selectCoinsToSpend in the way that it is
+// targetting a specific funding output size, instead of spending a given
```
wpaulino: Nit: s/targetting/targeting/

There are some other cases as well.

lnwallet/wallet.go
```diff
@@ -1380,6 +1417,119 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
...
+// 'toSpend' creating the funding tx. If necessary, a change address will also
+// be generated. If coin selection is successful/possible, then the selected
+// coins and change outputs are returned, and the value of the resulting
+// funding output. This method locks the selected outputs, and a function
+// closure to unlock them in case of an error is returned.
+//
+// NOTE: This method is different from selectCoinsAndChange in the way that it
+// spends the given amount when selecting coins, instead of targetting a
+// specific funding output size.
+func (l *LightningWallet) selectCoinsToSpend(feeRate SatPerKWeight,
```
wpaulino: There's a good bit of duplication here with `coinSelect` and `selectCoinsAndChange`. Seems like it'd be possible to refactor some things around.

lnwallet/wallet.go
wpaulino: Another approach would be to make this a `SubtractFees bool` and keep `LocalFundingAmt` the same. Then we can just thread the boolean down to coin selection.

fundingmanager_test.go
```diff
@@ -2724,3 +2724,95 @@ func TestFundingManagerMaxConfs(t *testing.T) {
...
+		}
+
+		if !test.change && len(fundingTx.TxOut) != 1 {
+			t.Fatalf("expected 1 output, had %v",
+				len(fundingTx.TxOut))
+		}
+
+		// Inputs should be all funds in the wallet.
+		if len(fundingTx.TxIn) != len(allCoins) {
+			t.Fatalf("Fad %d inputs, expected %d",
```
wpaulino: Nit: s/fad/had/

lnwallet/reservation.go
halseth: I think that is covered by the dustcheck online 198?

lnwallet/wallet.go
halseth: Not sure. Legacy reasons I guess

lnwallet/wallet.go
halseth: I think dual funder is a known TODO tbh :p

fundingmanager_test.go
halseth: Yes, good idea! Done.

lnwallet/wallet.go
halseth: Good catch! Fixed by moving locking to below the error cases.

lnwallet/wallet.go
halseth: Fixed.

fundingmanager_test.go
halseth: Fixed.

lnwallet/wallet.go
halseth: Done.

lnwallet/wallet.go
halseth: Fixed

lnwallet/wallet.go
halseth: Yeah, a enum/union holding the amount and the type of amount would be ideal here. Defining a new type seemed a bit overkill.

Using the bool would work, would you prefer it over the current approach? (code wise it would be more or less the same).

halseth:
PTAL @wpaulino

lnwallet/wallet.go
```diff
@@ -449,9 +457,12 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 		return
 	}
 
+	capacity := req.LocalFundingAmt + req.RemoteFundingAmt
+	localFundingAmt := req.LocalFundingAmt
+
 	id := atomic.AddUint64(&l.nextFundingID, 1)
 	reservation, err := NewChannelReservation(
-		req.Capacity, req.FundingAmount, req.CommitFeePerKw, l, id,
+		capacity, localFundingAmt, req.CommitFeePerKw, l, id,
```
wpaulino: Nit: could also make `NewChannelReservation` take the remote funding amount as well.

lnwallet/reservation.go
wpaulino: That check doesn't cover the remote case. It does cover the initiator case though, so one of those checks is redundant.

lnwallet/wallet.go
```diff
@@ -1320,48 +1320,76 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
...
+	return &wire.TxOut{
+		Value:    int64(changeAmt),
+		PkScript: changeScript,
+	}, nil
+}
+
+// lockCoins locks the given UTXOs such that they will be reserved and
+// ineglible for use by other transactions. A function closure to unlock them
+// is returned.
+func (l *LightningWallet) lockCoins(coins []*Utxo) (func(), error) {
```
wpaulino: Add a note that this requires the `coinSelectMtx` lock to be held.

wpaulino: Also seems like this method doesn't need to return an error.

lnwallet/wallet.go
wpaulino: Yeah I think the bool approach is cleaner.

lnwallet/wallet.go
wpaulino: Hmm I was thinking of actually reusing the existing `selectCoins` method. We'd pass a `subtractFees bool` to it and, if set, run the special case logic.

halseth: Done.

halseth:
Pushed a new version using the `subtractFees` bool. Also increased the test covereage of `coinSelect`,  it should be handling more edge cases now.

PTAL @wpaulino

lnwallet/wallet.go
```diff
@@ -1392,19 +1436,27 @@ func selectInputs(amt btcutil.Amount, coins []*Utxo) (btcutil.Amount, []*Utxo, e
...
+// returns the size of the created funding ouput, as well as the size of the
+// change output if any. The specified fee rate should be expressed in sat/kw
+// for coin selection to function properly, and the passed dustLimit is used to
+// clamp the outputs created.
+//
+// If subtractFees is true, then the estimated fee will be taken from the
+// output amount, meaning that the resuling funding output might be smaller
+// than the passed amt. The resulting funding output value will be returned,
+// along with the change amount if any.
+func coinSelect(feeRate SatPerKWeight, amt, dustLimit btcutil.Amount,
```
wpaulino: It might just be me, but I find this method a bit difficult to follow in its final form due to all of the different variables involved, though avoiding the code duplication is certainly nice.

halseth: Yeah, I agree. In the process of trying to share the code between the two cases, it kinda got out of hand. The two cases (`subtractFee=true/false`) are different enough to make it tricky to handle edge cases between them.

Code harder to reason about is more prone for errors than a little bit of code duplication.

halseth: I created a version where the two cases are separated, lmk what you think: https://github.com/lightningnetwork/lnd/compare/master...halseth:autopilot-spend-coins-subtract-fees-distinct-coinselect?expand=1

wpaulino: Separating them LGTM. Would also be nice to get someone else's opinion.

lnwallet/wallet_test.go
```diff
@@ -0,0 +1,152 @@
...
+		// Optionally count a change output.
+		if change {
+			weightEstimate.AddP2WKHOutput()
+		}
+
+		totalWeight := int64(weightEstimate.Weight())
+		return feeRate.FeeForWeight(totalWeight)
+	}
+
+	type testCase struct {
```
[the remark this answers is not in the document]
halseth: Done.

lnwallet/wallet.go
```diff
@@ -1290,11 +1301,12 @@ func (l *LightningWallet) WithCoinSelectLock(f func() error) error {
 // outputs which sum to at least 'amt' amount of satoshis. If necessary,
 // a change address will also be generated. If coin selection is
 // successful/possible, then the selected coins and change outputs are
-// returned. This method locks the selected outputs, and a function closure to
-// unlock them in case of an error is returned.
+// returned, and the value of the resulting funding output. This method locks
```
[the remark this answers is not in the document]
halseth: It will always select coins targetting a multisig output, so I think it is safe to assume it is only used for funding txs for now.

lnwallet/wallet.go
```diff
@@ -1453,6 +1466,9 @@ func coinSelect(feeRate SatPerKWeight, amt, dustLimit btcutil.Amount,
 		// selected inputs. We'll remember the resulting values for
 		// now, while we try to add a change output.
 		changeAmt := btcutil.Amount(0)
+		if subtractFees {
+			outputAmt += rem
```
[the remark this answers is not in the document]
halseth: Should be better now after the split.

fundingmanager_test.go
```diff
@@ -2727,3 +2727,94 @@ func TestFundingManagerMaxConfs(t *testing.T) {
...
+			t.Fatalf("expected 2 outputs, had %v",
+				len(fundingTx.TxOut))
+		}
+
+		if !test.change && len(fundingTx.TxOut) != 1 {
+			t.Fatalf("expected 1 output, had %v",
+				len(fundingTx.TxOut))
+		}
+
+		// Inputs should be all funds in the wallet.
```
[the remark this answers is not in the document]
halseth: Done.

lnwallet/reservation.go
halseth: You're right, the below check covers the case were we are the initiator, but not when the remote gets a dust balance. I think it is okay for now though, since the remote is the one that should reject it because of the dust (it doesn't hurt us).

lnwallet/wallet.go
halseth: Agreed, but seemed like a change too big to be done in this PR. Added a TODO.

halseth:
Coin select separated to make logic more straightforward and comments addressed. PTAL @wpaulino

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+	// wil go to change.
+	newChange := totalSat - amt
+	newOutput := amt - requiredFee
+
+	// If adding a change output lead to both outputs being above
+	// the dust limit, we'll add the change ouput. Otherwise we'll
+	// go with the no change tx we originally found.
+	if newChange > dustLimit && newOutput > dustLimit {
+		outputAmt = newOutput
+		changeAmt = newChange
```
wpaulino: If we can proceed with the change output, wouldn't we want to subtract the fee from the change output instead if possible? The current logic subtracts it from the non-change output.

fundingmanager.go
```diff
@@ -2756,10 +2755,11 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		ourDustLimit = defaultLitecoinDustLimit
 	}
 
-	fndgLog.Infof("Initiating fundingRequest(localAmt=%v, remoteAmt=%v, "+
-		"capacity=%v, chainhash=%v, peer=%x, dustLimit=%v, min_confs=%v)",
-		localAmt, msg.pushAmt, capacity, msg.chainHash,
-		peerKey.SerializeCompressed(), ourDustLimit, msg.minConfs)
+	fndgLog.Infof("Initiating fundingRequest(localAmt=%v "+
+		"(subtractFees=%v), pushAmt=%v, chainhash=%v, peer=%x, "+
+		"dustLimit=%v, min_confs=%v)", localAmt, msg.subtractFees,
```
wpaulino: Style nit: make `min_confs` camelCase for consistency?

Roasbeef:
Will this result in 1-in-1-out channels if using the entire wallet balance? I bring this up as ideally we can still support CPFP'ing these channels for confirmation purposes even if we're using _most_ of the balance of the wallet. However, in that case, we'd need to start leaving a "fee purse" on-chain at all times. In the future, once we move to pure CPFP or sighash single fees for channel force closes, then such a coin tranche will be required for operation.

fundingmanager_test.go
```diff
@@ -2391,6 +2420,8 @@ func TestFundingManagerCustomChannelParameters(t *testing.T) {
 // TestFundingManagerMaxPendingChannels checks that trying to open another
 // channel with the same peer when MaxPending channels are pending fails.
 func TestFundingManagerMaxPendingChannels(t *testing.T) {
+	t.Parallel()
```
Roasbeef: IIRC there's some lingering global variable race condition that prevented us from doing this in the past?

lnwallet/wallet_test.go
```diff
@@ -0,0 +1,176 @@
...
+	// Optionally count a change output.
+	if change {
+		weightEstimate.AddP2WKHOutput()
+	}
+
+	totalWeight := int64(weightEstimate.Weight())
+	return feeRate.FeeForWeight(totalWeight)
+}
+
+// TestCoinSelect tests that we pick coins adding up to the expected amount
```
Roasbeef: 👌

lnwallet/wallet_test.go
```diff
@@ -0,0 +1,176 @@
...
+// when creating a funding transaction, and that the calculated change is the
+// expected amount.
+// NOTE: coinSelect will always attempt to add a change output, so we must
```
Roasbeef: Style nit: missing newline above.

lnwallet/wallet_test.go
```diff
@@ -0,0 +1,176 @@
...
+			// We have just enough left to pay the fee, so there is
+			// nothing left for change.
+			// TODO(halseth): currently coinselect estimates fees
+			// assuming a change output.
+			expectedChange: 0,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
```
Roasbeef: Can bind the loop variable `test := test`, then use `t.Paralle()` within the `Run` closure.

lnwallet/wallet.go
```diff
@@ -1274,13 +1275,14 @@ func (l *LightningWallet) WithCoinSelectLock(f func() error) error {
...
-// also be generated.
+// outputs which sum to at least 'amt' amount of satoshis. If necessary,
+// a change address will also be generated. If coin selection is
+// successful/possible, then the selected coins and change outputs are
+// returned. This method locks the selected outputs, and a function closure to
+// unlock them in case of an error is returned.
 func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
-	amt btcutil.Amount, minConfs int32,
-	contribution *ChannelContribution) error {
+	amt btcutil.Amount, minConfs int32) ([]*wire.TxIn, []*wire.TxOut,
```
Roasbeef: Would be more readable if the return parameters were lifted into a struct that looks something like: 
```go
type selectedCoins struct {
    Coins []*wire.TxIn
    Change []*wire.TxOut
    ReleaseCoins func() 
}
```

lnwallet/interface_test.go
```diff
@@ -772,7 +772,7 @@ func testReservationInitiatorBalanceBelowDustCancel(miner *rpctest.Harness,
 		LocalFundingAmt:  fundingAmount,
 		RemoteFundingAmt: 0,
 		CommitFeePerKw:   feePerKw,
-		FundingFeePerKw:  feePerKw,
+		FundingFeePerKw:  1000,
```
Roasbeef: Why was this needed?

lnwallet/wallet.go
```diff
@@ -483,66 +484,88 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
-		req.err <- err
-		req.resp <- nil
-		return
+		return err
 	}
 	reservation.ourContribution.DelayBasePoint, err = l.DeriveNextKey(
 		keychain.KeyFamilyDelayBase,
 	)
 	if err != nil {
-		req.err <- err
```
Roasbeef: Nice to get rid of all these error sends as well!

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
 		return selectedUtxos, changeAmt, nil
 	}
 }
+
+// coinSelectSubtractFees attempts to select coins such that we'll spend up to
```
Roasbeef: we'll -> will ?

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+		return nil, 0, 0, fmt.Errorf("output amount(%v) after "+
+			"subtracting fees(%v) below dust limit(%v)", outputAmt,
+			requiredFee, dustLimit)
+	}
+
+	// We were able to create a transaction with no change from the
+	// selected inputs. We'll remember the resulting values for
+	// now, while we try to add a change output. Assume that change output
+	// is a P2WKH output.
+	weightEstimate.AddP2WKHOutput()
```
Roasbeef: Don't see an area where this is ever undone, if we don't need a change output. So we'll still factor this output into the fees we pay, resulting in us over paying.

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+		changeAmt = newChange
+	}
+
+	// Sanity check the resulting ouput values to make sure we
+	// don't burn a great part to fees.
+	totalOut := outputAmt + changeAmt
+	fee := totalSat - totalOut
+
+	// Fail if more than 20% goes to fees.
+	if fee > totalOut/5 {
```
Roasbeef: Possibly make this 20% value a config parameter? In certain situations like testnet or simnet, the user may not care.

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+
+	// If adding a change output lead to both outputs being above
+	// the dust limit, we'll add the change ouput. Otherwise we'll
+	// go with the no change tx we originally found.
+	if newChange > dustLimit && newOutput > dustLimit {
+		outputAmt = newOutput
+		changeAmt = newChange
+	}
+
+	// Sanity check the resulting ouput values to make sure we
```
Roasbeef: ouput -> output

lnwallet/wallet_test.go
```diff
@@ -174,3 +174,173 @@ func TestCoinSelect(t *testing.T) {
 		})
 	}
 }
+
+// TestCoinSelectSubtractFees tests that we pick coins adding up to the expected amount
```
Roasbeef: Funky line wrapping?

fundingmanager.go
Roasbeef: We kinda use both in the codebase for logging atm. Personally, I prefer snake case for stuff like this, and camel case when referring to a sub-system within the codebase, similar to Python naming conventions.

autopilot/agent.go
```diff
@@ -584,6 +584,7 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 
 	// Use the heuristic to calculate a score for each node in the
 	// graph.
+	log.Debugf("Scoring %d nodes for chansize=%v", len(nodes), chanSize)
```
Roasbeef: `chan_size`?

fundingmanager_test.go
wpaulino: I believe that was addressed by https://github.com/lightningnetwork/lnd/pull/3165/commits/fe7d9143d8e15bfa5f57793e52c7110a643faf04.

wpaulino:
> Will this result in 1-in-1-out channels if using the entire wallet balance? I bring this up as ideally we can still support CPFP'ing these channels for confirmation purposes even if we're using most of the balance of the wallet. However, in that case, we'd need to start leaving a "fee purse" on-chain at all times. In the future, once we move to pure CPFP or sighash single fees for channel force closes, then such a coin tranche will be required for operation.

Seems like we can work around this with the autopilot allocation setting?

fundingmanager_test.go
halseth: Yep exactly :)

lnwallet/interface_test.go
halseth: The test is attempting to test that an insane commit fee is pushing our balance into the negative, but without this change it would fail even to select coins for the funding tx (this was not a problem before since we first selected coins after making the reservation).

lnwallet/wallet.go
halseth: we'll is correct

lnwallet/wallet.go
halseth: The `weightEstimate` is only used again on the line below, so no reason to undo. 

At this point we have found a valid coin selection (`outputAmt` and `changeAmt=0`), so we just check if adding a change output still is valid (if not we return the first valid selection we found).

lnwallet/wallet_test.go
halseth: fixed: https://github.com/lightningnetwork/lnd/pull/3165/commits/53f5e10b856323356115c9ac8df9792b869f4c4a

lnwallet/wallet_test.go
halseth: fixed: https://github.com/lightningnetwork/lnd/pull/3165/commits/53f5e10b856323356115c9ac8df9792b869f4c4a

lnwallet/wallet.go
halseth: Good idea! Done: https://github.com/lightningnetwork/lnd/pull/3165/commits/b136a27562cfa91d569f8a7588d1c42dbaca09d6

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+
+	var weightEstimate input.TxWeightEstimator
+	for _, utxo := range selectedUtxos {
+		switch utxo.AddressType {
+		case WitnessPubKey:
+			weightEstimate.AddP2WKHInput()
+		case NestedWitnessPubKey:
+			weightEstimate.AddNestedP2WKHInput()
+		default:
+			return nil, 0, 0, fmt.Errorf("Unsupported "+
```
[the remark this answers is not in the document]
halseth: done: https://github.com/lightningnetwork/lnd/pull/3165/commits/a01182ae51fa85b5c9b3be8e64ad4e2e203fa243

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+	// is a P2WKH output.
+	weightEstimate.AddP2WKHOutput()
+
+	// Now that we have added the change output, redo the fee
+	// estimate.
+	totalWeight = int64(weightEstimate.Weight())
+	requiredFee = feeRate.FeeForWeight(totalWeight)
+
+	// For a transaction with a change output, everything we don't spend
+	// wil go to change.
```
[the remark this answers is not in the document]
halseth: Fixed

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+	newChange := totalSat - amt
+	newOutput := amt - requiredFee
+
+	// If adding a change output lead to both outputs being above
```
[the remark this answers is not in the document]
halseth: Fixed

lnwallet/wallet.go
```diff
@@ -1485,3 +1485,93 @@ func coinSelect(feeRate SatPerKWeight, amt btcutil.Amount,
...
+	// the dust limit, we'll add the change ouput. Otherwise we'll
```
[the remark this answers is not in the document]
halseth: fixed

lnwallet/wallet.go
halseth: fixed.

lnwallet/wallet_test.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3165/commits/bb47413e52538b1ddffe9bec133e247824537dd7

lnwallet/wallet.go
halseth: Since this method is meant to target an exact _spend amount_, we want to ensure that the total we spend is `fundingAmt+fee`. Subtracting it from the change would cause us to spend a bit more than intended.

fundingmanager.go
halseth: Made them all snake case 🐍

lnwallet/wallet.go
halseth: Are you thinking a user configurable setting? 

Since the user can already control fees for manually opened channels, maybe it makes sense to make it an `autopilot` setting?

halseth:
> > Will this result in 1-in-1-out channels if using the entire wallet balance? I bring this up as ideally we can still support CPFP'ing these channels for confirmation purposes even if we're using most of the balance of the wallet. However, in that case, we'd need to start leaving a "fee purse" on-chain at all times. In the future, once we move to pure CPFP or sighash single fees for channel force closes, then such a coin tranche will be required for operation.
> 
> Seems like we can work around this with the autopilot allocation setting?

Decided to go with the 1-output funding txes for now, as this mostly only targetting the "fund all" case, where it makes sense to not leave any tiny change output.

PTAL @wpaulino @Roasbeef

wpaulino:
>>> Will this result in 1-in-1-out channels if using the entire wallet balance? I bring this up as ideally we can still support CPFP'ing these channels for confirmation purposes even if we're using most of the balance of the wallet. However, in that case, we'd need to start leaving a "fee purse" on-chain at all times. In the future, once we move to pure CPFP or sighash single fees for channel force closes, then such a coin tranche will be required for operation.

>> Seems like we can work around this with the autopilot allocation setting?

> Decided to go with the 1-output funding txes for now, as this mostly only targetting the "fund all" case, where it makes sense to not leave any tiny change output.

Another approach we could take it to keep things as is in this PR and introduce a new autopilot flag that ensures a tiny anchor output is created for the purpose of bumping fees. We'd then tune the autopilot allocation to leave X% for the "fee purse" and use that when bumping.

halseth:
Squashed.

fundingmanager.go
```diff
@@ -2801,6 +2802,12 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		return
 	}
 
+	// Now that we have successfully reserved funds for this channel in the
+	// wallet, we can fetch the final channel capacity. This is done at
+	// this point since the final capacity might change in case of
+	// fundAll=true.
```
cfromknecht: fundAll -> subtractFees?

lnwallet/wallet.go
cfromknecht: agree on making this check configurable! however i don't think a percentage makes much sense here, since the fees paid are independent of the amount being sent. 20% might make sense for $1 channel, but not a $100 or $1000. perhaps this is better expressed as a max fee rate?

lnwallet/wallet.go
```diff
@@ -1291,52 +1330,93 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
...
-	for i, coin := range selectedCoins {
-		outpoint := &coin.OutPoint
-		l.lockedOutPoints[*outpoint] = struct{}{}
-		l.LockOutpoint(*outpoint)
+	switch {
+	// In case this request want the fees subtracted from the local amount,
+	// we'll call the specialized method for that. This ensures that we
+	// won't deduct more that the specified balance from our wallet.
+	case subtractFees:
+		dustLimit := DefaultDustLimit()
```
cfromknecht: will this be accurate on ltc?

lnwallet/wallet.go
```diff
@@ -1291,52 +1330,93 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
...
+			feeRate, amt, dustLimit, coins,
+		)
+		if err != nil {
+			return nil, err
+		}
 
-		// Empty sig script, we'll actually sign if this reservation is
-		// queued up to be completed (the other side accepts).
-		contribution.Inputs[i] = wire.NewTxIn(outpoint, nil, nil)
+	// Ótherwise to a normal coin selection where we target a given funding
```
cfromknecht: to -> do

lnwallet/wallet.go
```diff
@@ -1291,52 +1330,93 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate SatPerKWeight,
...
 
-	return nil
+	// Lock the selected coins. These coins are now "reserved", this
+	// prevents concurrent funding requests from referring to and this
+	// double-spending the same set of coins.
+	inputs := make([]*wire.TxIn, len(selectedCoins))
+	for i, coin := range selectedCoins {
+		outpoint := &coin.OutPoint
+		l.lockedOutPoints[*outpoint] = struct{}{}
+		l.LockOutpoint(*outpoint)
```
cfromknecht: isn't this a little racey since two calls to this method can get the same utxos before locking them?

wpaulino: The `coinSelectMtx` being acquired throughout this method should prevent that, no?

cfromknecht: ahh cool missed that

lnwallet/wallet.go
halseth: I added the check in the first place to avoid cases where a user would deposit say $5 into the wallet, and autopilot opened a channel where $3 went to fees. Setting a max fee rate in that scenario wouldn't help, since $3 fee might be okay if you are opening large channels.

We are targeting app users primarily here, so maybe we should discuss it in that context.

cfromknecht: > Setting a max fee rate in that scenario wouldn't help, since $3 fee might be okay if you are opening large channels.

Why doesn't it help? To me it seems more effective than setting a percentage, since we're directly limiting how much we allocate to fees rather than assuming people are okay paying higher fees simply because they have more money.

> We are targeting app users primarily here, so maybe we should discuss it in that context.

We should still be cognizant of non-app users of autopilot or developers using lnd+autopilot in their own applications, of which there are many on mainnet.

Roasbeef: > Why doesn't it help?

Because it's absolute. In that "deposit $5, but get $3" scenario, what max fee should be set there? As this example is denominated in fiat, one would also need to factor in the current exchange rate as well. Alternatively, we could have "fee bands", kinda like a piece-wise function to decide what the proper % limit should be given the channel size. So for a 30k satoshi channel, 10k satoshis may be unacceptable as a total fee, while for a 300k satoshi channel 10k satoshis isn't so bad.

cfromknecht: > Because it's absolute.

Not sure I follow, aren't both absolute? One is don't go above this %, the other is don't go above this sat/byte.

In that "deposit $5, but get $3" scenario, what max fee should be set there? As this example is denominated in fiat, one would also need to factor in the current exchange rate as well.

Was thinking that a user would just tell autopilot to never use more than, e.g., 5 sat/byte for funding. Exchange rates wouldn't come into play at all

> Alternatively, we could have "fee bands", kinda like a piece-wise function to decide what the proper % limit should be given the channel size. So for a 30k satoshi channel, 10k satoshis may be unacceptable as a total fee, while for a 300k satoshi channel 10k satoshis isn't so bad.

To me it'd be less intuitive to specify a number of satoshis, because then you have to convert into dollar amounts. When people see a fee rate, they now roughly know the difference between a good and a bad sat/byte value. Perhaps fee rate bands? :)

halseth: I don't think this question has a "right" answer, and should be getting more discussion, at least before we start implementing a more complex heuristic.

I'm okay just removing this sanity check for now, even though app users might end up paying high percentage wise fees for opening channels (earlier we had this capped as a result of targeting 95% autopilot allocation in the app, which would leave only 5% to fees). Or just set it at 50% or something, with a TODO of improving it.

lnwallet/wallet_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/input"
+)
+
+// fee is a helper method that returns the fee estimate used for a tx with the
+// given number of inputs and the optional change output. This matches the
+// estimate done by the wallet.
+func fee(feeRate SatPerKWeight, numInput int, change bool) btcutil.Amount {
```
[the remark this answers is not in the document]
halseth: fixed: 9d1404bfb1ad02dcb72d4e65af514283dae7d05a

lnwallet/wallet.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3165/commits/e74c13c935052638fff291b4e60c779e05935a36

lnwallet/wallet.go
halseth: No. Changed to using `DefaultConstraints.DustLimit` instead, which is properly set also for LTC: :https://github.com/lightningnetwork/lnd/pull/3165/commits/a1e826d721bcf799d52490bc7633e775ead073dd

It is a bit hacky since the constraints are really meant for commitment outputs, but I guess we can assume that dustlimit and the regular tx dustlimit is the same (they are currently)-

fundingmanager.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3165/commits/c04ce8727f5c2a4ddaf42fca89db31fd374b5c2b

lnwallet/wallet.go
cfromknecht: @halseth sounds good to me, in that case i'd just leave the 20% heuristic with a todo

halseth: TODO added

PR #3172, opened 2019-06-07
routing+lnrpc: move default payment timeout out of router

joostjager:
This commit moves the default timeout out of router and thereby fixes a bug that caused SendToRoute to not return the actual error, but a timeout result instead. SendToRoute only tries a single route, so a timeout should never happen.

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -29,10 +29,10 @@ import (
 )
 
 const (
-	// defaultPayAttemptTimeout is a duration that we'll use to determine
-	// if we should give up on a payment attempt. This will be used if a
-	// value isn't specified in the LightningNode struct.
-	defaultPayAttemptTimeout = time.Duration(time.Second * 60)
+	// DefaultPayAttemptTimeout is the default payment attempt timeout. The
+	// payment attempt timeout defines the duration after which we stop
+	// trying more routes for a payment.
+	DefaultPayAttemptTimeout = time.Duration(time.Second * 60)
 
 	// DefaultChannelPruneExpiry is the default duration used to determine
 	// if a channel should be pruned or not.
@@ -529,6 +529,9 @@ func (r *ChannelRouter) Start() error {
 			// We create a dummy, empty payment session such that
 			// we won't make another payment attempt when the
 			// result for the in-flight attempt is received.
+			//
+			// PayAttemptTime doesn't need to be set, as there is
+			// only a single attempt.
 			paySession := r.cfg.MissionControl.NewPaymentSessionEmpty()
 
 			lPayment := &LightningPayment{
@@ -1572,7 +1575,8 @@ type LightningPayment struct {
 	// PayAttemptTimeout is a timeout value that we'll use to determine
 	// when we should should abandon the payment attempt after consecutive
 	// payment failure. This prevents us from attempting to send a payment
-	// indefinitely.
+	// indefinitely. A zero value means the payment will never time out.
+	//
 	// TODO(halseth): make wallclock time to allow resume after startup.
 	PayAttemptTimeout time.Duration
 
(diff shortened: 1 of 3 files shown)


routing/router.go
```diff
@@ -1767,29 +1767,23 @@ func (r *ChannelRouter) sendPayment(
...
 		paySession:     paySession,
-		timeoutChan:    timeoutChan,
 		currentHeight:  currentHeight,
 		finalCLTVDelta: uint16(payment.FinalCLTVDelta),
 		attempt:        existingAttempt,
 		circuit:        nil,
 		lastError:      nil,
 	}
 
+	if payment.PayAttemptTimeout != time.Duration(0) {
```
cfromknecht: is this cast necessary?

rpcserver.go
```diff
@@ -64,6 +64,11 @@ const (
 	// permitted.
 	maxLtcPaymentMSat = lnwire.MilliSatoshi(math.MaxUint32) *
 		btcToLtcConversionRate
+
+	// defaultPayAttemptTimeout is a duration that we'll use to determine
+	// if we should give up on a payment attempt. This will be used if a
+	// value isn't specified in the LightningNode struct.
```
cfromknecht: Should `LightningNode` be `LightningPayment`?

routing/router.go
joostjager: no

rpcserver.go
joostjager: fixed

joostjager:
@cfromknecht ptal

rpcserver.go
```diff
@@ -64,6 +64,11 @@ const (
 	// permitted.
 	maxLtcPaymentMSat = lnwire.MilliSatoshi(math.MaxUint32) *
 		btcToLtcConversionRate
+
+	// defaultPayAttemptTimeout is a duration that we'll use to determine
+	// if we should give up on a payment attempt. This will be used if a
+	// value isn't specified in the request.
+	defaultPayAttemptTimeout = time.Duration(time.Second * 60)
```
halseth: would it make more sense to define this in `routing.DefaultPayAttemptTimeout`?

halseth: Basically just exporting the constant that was already there.

routing/router.go
```diff
@@ -1767,29 +1767,23 @@ func (r *ChannelRouter) sendPayment(
...
 		paySession:     paySession,
-		timeoutChan:    timeoutChan,
 		currentHeight:  currentHeight,
 		finalCLTVDelta: uint16(payment.FinalCLTVDelta),
 		attempt:        existingAttempt,
 		circuit:        nil,
 		lastError:      nil,
 	}
 
+	if payment.PayAttemptTimeout != 0 {
```
halseth: Add a comment.

routing/router.go
```diff
@@ -1699,9 +1697,11 @@ func (r *ChannelRouter) SendToRoute(hash lntypes.Hash, route *route.Route) (
 	}
 
 	// Create a (mostly) dummy payment, as the created payment session is
-	// not going to do path finding.
-	// TODO(halseth): sendPayment doesn't relly need LightningPayment, make
-	// it take just needed fields instead.
+	// not going to do path finding. TODO(halseth): sendPayment doesn't
```
halseth: nit: TODO should be on newline

rpcserver.go
joostjager: Yes, I think that is better. fixed.

joostjager:
@halseth ptal

joostjager:
Reminder: handle timeout 0 if passed in via rpc. Reject? Or only do one attempt?

routing/router.go
```diff
@@ -29,10 +29,10 @@ import (
 )
 
 const (
-	// defaultPayAttemptTimeout is a duration that we'll use to determine
-	// if we should give up on a payment attempt. This will be used if a
-	// value isn't specified in the LightningNode struct.
-	defaultPayAttemptTimeout = time.Duration(time.Second * 60)
+	// DefaultPayAttemptTimeout is the default payment attempt timeout. The
+	// payment attempt timeout defines the duration after which  we stop
```
alexbosworth:
```suggestion
	// payment attempt timeout defines the duration after which we stop
```

PR #3173, opened 2019-06-07
routing/router: log aggregate graph processing stats

cfromknecht:
This PR puts all of the CRTR logs during IGD on trace, and logs aggregate stats every 30s if there is anything new discovered

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -25,6 +25,7 @@ import (
 	"github.com/lightningnetwork/lnd/multimutex"
 	"github.com/lightningnetwork/lnd/routing/chainview"
 	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/lightningnetwork/lnd/ticker"
 	"github.com/lightningnetwork/lnd/zpay32"
 )
 
@@ -37,6 +38,11 @@ const (
 	// DefaultChannelPruneExpiry is the default duration used to determine
 	// if a channel should be pruned or not.
 	DefaultChannelPruneExpiry = time.Duration(time.Hour * 24 * 14)
+
+	// defaultStatInterval governs how often the router will log non-empty
+	// stats related to processing new channels, updates, or node
+	// announcements.
+	defaultStatInterval = 30 * time.Second
 )
 
 var (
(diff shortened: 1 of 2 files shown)


wpaulino:
Needs a rebase!

routing/router.go
```diff
@@ -1155,10 +1189,12 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 			if err := r.cfg.Graph.AddChannelEdge(msg); err != nil {
 				return fmt.Errorf("unable to add edge: %v", err)
 			}
-			log.Infof("New channel discovered! Link "+
+			log.Tracef("New channel discovered! Link "+
```
halseth: Maybe a bit aggressive to demote to `Trace`? I find these logs useful sometimes running itests, where we use debug level.

cfromknecht: hmm my hope was that we could get rid of these at the `Debug` level since they're superrr spammy and aren't that useful in production. I run my node with `CRTR=debug` so just moving them to debug wouldn't change anything.

routing/router.go
```diff
@@ -1108,7 +1141,8 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 				"graph: %v", msg.PubKeyBytes, err)
 		}
 
-		log.Infof("Updated vertex data for node=%x", msg.PubKeyBytes)
+		log.Tracef("Updated vertex data for node=%x", msg.PubKeyBytes)
+		r.stats.numNodesUpdates++
```
Roasbeef: This is called concurrently with other goroutines, so we either need a wrapper method with a mutex, or usage of an `atomic` increment and laod.

Roasbeef: Applies to all other spots where we increment/read the new `stats` attribute.

routing/stats.go
```diff
@@ -0,0 +1,36 @@
...
+// Reset clears any router stats and sets the lastReset field to now.
+func (g *routerStats) Reset() {
+	*g = routerStats{
+		lastReset: time.Now(),
+	}
+}
+
+// String returns a human-readable description of the router stats.
+func (g *routerStats) String() string {
+	return fmt.Sprintf("Processed channels=%d updates=%d nodes=%d in "+
```
Roasbeef: Either needs a mutex, or needs to use `atomic.LoadInt32`.

cfromknecht:
concurrency issues addressed, @Roasbeef ptal

routing/router.go
cfromknecht: good catch, fixed

PR #3177, opened 2019-06-08
[autopilot] make confirmation target configurable

halseth:
This PR adds a new configuration option `conftarget` for channels opened by autopilot.

### DIFF
--- a/autopilot/interface.go
+++ b/autopilot/interface.go
@@ -9,6 +9,11 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
+// DefaultConfTarget is the default confirmation target for autopilot channels.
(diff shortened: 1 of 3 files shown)


autopilot/interface.go
```diff
@@ -9,6 +9,11 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
+// DefaultConfTarget is the default confirmation target for autopilot channels.
+// TODO(halseth): possibly make dynamic, going aggressive->lax as more channels
+// are opened.
+const DefaultConfTarget = 3
```
wpaulino: Once https://github.com/lightningnetwork/lnd/pull/3140 is in, we could even make this higher and users can bump manually if desired.

PR #3178, opened 2019-06-09
multi: replace manual CAS with sync.Once in several more modules

federicobond:


channelnotifier/channelnotifier.go
```diff
@@ -57,26 +57,19 @@ func New(chanDB *channeldb.DB) *ChannelNotifier {
...
-	log.Tracef("ChannelNotifier %v starting", c)
-
-	if err := c.ntfnServer.Start(); err != nil {
-		return err
-	}
-
-	return nil
+	var err error
+	c.started.Do(func() {
+		log.Tracef("ChannelNotifier %v starting", c)
```
wpaulino: While we're here, I don't think `c` needs to be logged here.

federicobond:
Updated: added a missing return a the end of a function and removed the ChannelNotifier instance from the log string.

Issue #3001, opened 2019-04-23
invoices+rpc: disallow creating an invoice if the node lacks insufficient incoming bandwidth


Roasbeef:
Today it's possible to create an invoice that can't actually ever be fulfilled. For example, if the maximum inbound bandwidth for all my channels is 100k satoshi, then I can't receive a 300k satoshi payment. An optional mode should be added to `AddInvoice` to _reject_ invoice creation like this, notifying the caller that they can't receive a payment of that size. Eventually, we may also want to make this mode the default in order to slowly educate users w.r.t the flow constraints of channels. Even after AMP, this will still be relevant, but instead we'd sum inbound bandwidth across all channels.

xsb:
I can pick this one

PR #3179, opened 2019-06-09
invoices+rpc: disallow creating an invoice if the node lacks sufficient incoming bandwidth

xsb:
Reject invoice creation if node lacks sufficient inbound bandwidth in any of its channels. This new mode is disabled by default.

Also checked that no existing tests are broken if we enable this mode globally. So a future change making this the default would be easy to implement. The changes required taking AMP into account seem also very small.

Fixes https://github.com/lightningnetwork/lnd/issues/3001


lntest/itest/lnd_test.go
```diff
@@ -5387,6 +5387,87 @@ func testInvoiceRoutingHints(net *lntest.NetworkHarness, t *harnessTest) {
 	cleanupForceClose(t, net, net.Alice, chanPointEve)
 }
 
+// testInvoiceCheckInboundBandwidth tests that creating a new invoice in a node
+// without sufficient inbound bandwidth fails when the check for it is enabled.
+//
+// The scenario in which there is enough bandwidth is also tested.
+func testInvoiceCheckInboundBandwidth(net *lntest.NetworkHarness,
```
halseth: Instead of adding an integration test, check if it can be made a unit test on `AddInvoice` 😄

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -180,6 +184,33 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		)
 	}
 
+	// Check if any open channel has enough inbound bandwidth to receive an
+	// invoice of the specified amount
+	if invoice.CheckInboundBandwidth {
+		inboundBandwidth := false
+
+		openChannels, err := cfg.ChanDB.FetchAllOpenChannels()
```
halseth: we might also consider only checking active channels, but I think this is okay (less aggressive).

lntest/itest/lnd_test.go
xsb: This makes a lot of sense, thanks @halseth. I am checking the code and I am a bit confused so a bit of help would be very appreciated.

My changes are on `lnrpc/invoicesrpc` where there are not unit tests and as far as I understand the behaviour is covered with integration tests. I see how the unit test could in theory cover `channeldb`'s `AddInvoice` but I did not modify that part. I tried to do the check sooner than later so the db does not handle this.

lnrpc/invoicesrpc/addinvoice.go
xsb: I did not think about this but it's an easy change. As this check is optional I guess we can be strict here and make it require an active channel.

lntest/itest/lnd_test.go
halseth: You can add a new file `lnrpc/invoicesrpc/addinvoice_test.go` for the unit test.

joostjager:
From the perspective of keeping the api surface small, is this functionality really needed? With `listchannels` as a workaround.

joostjager:
Or otherwise adding a `maxreceiveamount` field to the `channelbalance` call.

lntest/itest/lnd_test.go
xsb: @halseth I've been reviewen this again after a few days I am not sure how this could be a unit test because to check if there are enough inbound funds we would need to first create channels so it would require to touch channeldb.

On top of that I also found out that the contribution guide asks that anything that changes behaviour within in the RPC be accompanied by integration tests. So I am a bit lost here right now. Any hint on what I may be missing would be very very helpful.

xsb:
@joostjager thanks for your comments. I can think of another solution, which is to implement this in `lncli` but not on the RPC. This would keep the API surface small as you said, but also would provide the functionality to anyone that needs it in the command line. Workarounds are way easier when using LND programatically, just 2 RPC calls instead of 1.

This is a solution that I've seen in kubernetes. And the idea is that, whith time, if the functionality is adopted and people ask for it in the API, the discussion can be opened again.

lnrpc/invoicesrpc/addinvoice.go
xsb: just added this on a fixup commit

joostjager:
@roasbeef what is your opinion on this as the original creator of the issue?

lntest/itest/lnd_test.go
halseth: If you look at `AddInvoiceConfig` you can mock out several of the dependencies. Since the actual logic here is pretty self contained you could also create a method such as `checkInboundBandwidht(chans *OpenChannel, amt Millisatoshi) (bool, error)` and unit test that method directly.

xsb: Small update: I moved the logic to `checkInboundBandwidth(...)` as you proposed and unit tested it.

Roasbeef:
Closing due to inactivity.

PR #3182, opened 2019-06-10
Initialize addressType to UnknownAddressType

yaslama:
Initialize addressType to UnknownAddressType instead of the default (null value) which is lnwallet.WitnessPubKey.
Without this change, if both IsPayToWitnessPubKeyHash and IsPayToScriptHash returns false, the address is considered to be WitnessPubKey which is obviously false.

### DIFF
--- a/lnwallet/btcwallet/btcwallet.go
+++ b/lnwallet/btcwallet/btcwallet.go
@@ -378,7 +378,7 @@ func (b *BtcWallet) ListUnspentWitness(minConfs, maxConfs int32) (
 			return nil, err
 		}
 
-		var addressType lnwallet.AddressType
+		addressType := lnwallet.UnknownAddressType
 		if txscript.IsPayToWitnessPubKeyHash(pkScript) {
 			addressType = lnwallet.WitnessPubKey
 		} else if txscript.IsPayToScriptHash(pkScript) {


yaslama:
@wpaulino I agree that it's safe to also modify the default value because the value of this lnwallet.AddressType is not stored anywhere nor is used in rpc. I opened a PR for that (https://github.com/lightningnetwork/lnd/pull/3187).  But this specific PR is a lot more important, because without the initialization to UnknownAddressType, when ImportWitnessTransaction (from https://github.com/btcsuite/btcwallet/pull/625) is used, lnd panics.

cfromknecht:
good catch @yaslama, yeah it's unfortunate the default value wasn't made to be the zero case which prevents things like this from happening. could be worth investigating further if we can actually change the definition down the road

halseth:
With #3187 I don't think this PR is needed?

yaslama:
> With #3187 I don't think this PR is needed?

Technically it's not needed, but without initialization it's not clear what is the value when reading the code. A mental effort needs to be done to consider UnknownAddressType as a null/empty value.

halseth:
> Technically it's not needed, but without initialization it's not clear what is the value when reading the code.

Good point, making it easy on the reader should always be a goal :)

PR #3188, opened 2019-06-11
htlcswitch+routing+routerrpc: error source as index and expose decrypt failure

joostjager:
This PR modifies the switch behavior to return an integer indicating the error source instead of a pub key. This allows for a node occurring in the path multiple times. It also changes `routerrpc.SendToRoute` to return the failure source as an integer.

A second change introduces the `ErrUnreadableFailure` result for payments. This is a preparation for this case to be processed properly in `router`. Also an unknown failure (this is different from an unreadable failure) is now distinguishable.

Builds onto https://github.com/lightningnetwork/lightning-onion/pull/37

This PR prepares for persistent mission control #3164


htlcswitch/switch.go
```diff
@@ -999,7 +999,7 @@ func (s *Switch) parseFailedPayment(deobfuscator ErrorDecrypter,
 				paymentHash, paymentID, err)
 			log.Error(userErr)
 			failure = &ForwardingError{
-				ErrorSource:    s.cfg.SelfKey,
+				ErrorSourceIdx: 0,
```
halseth: can we get rid of `SelfKey` now?

lnrpc/routerrpc/router.proto
```diff
@@ -203,6 +199,12 @@ message Failure {
 
     /// A failure type-dependent flags value.
     uint32 flags = 7;
+
+    /**
+    The position in the path of the intermediate or final node that generated
+    the failure message. Position zero is the sender node.
+    **/
+    int32 failure_source_index = 8;
```
halseth: should we be consistent in naming, `error source ` or `failure source`?

htlcswitch/failure.go
```diff
@@ -247,7 +247,7 @@ var _ ErrorEncrypter = (*SphinxErrorEncrypter)(nil)
 // SphinxErrorDecrypter wraps the sphinx data SphinxErrorDecrypter and maps the
 // returned errors to concrete lnwire.FailureMessage instances.
 type SphinxErrorDecrypter struct {
```
halseth: Could just be an interface now?

htlcswitch/switch.go
```diff
@@ -59,6 +59,10 @@ var (
 	// active links in the switch for a specific destination.
 	ErrNoLinksFound = errors.New("no channel links found")
 
+	// ErrUnreadableFailureMessage is returned when the failure message
+	// cannot be decrypted.
+	ErrUnreadableFailureMessage = errors.New("unreadable failure message")
```
halseth: 👍

htlcswitch/failure.go
```diff
@@ -262,10 +262,14 @@ func (s *SphinxErrorDecrypter) DecryptError(reason lnwire.OpaqueReason) (*Forwar
 		return nil, err
 	}
 
+	// Decode the failure. If an error occurs, we leave the failure message
```
halseth: why?

joostjager: Because it is not an internal error. We do have an error source and we can only process this outcome in a proper way in mission control if we can distinguish this case from internal errors.

htlcswitch/switch.go
joostjager: Yes! Removed

lnrpc/routerrpc/router.proto
joostjager: Yes good point. I renamed everything to failure. That is what it is called in the spec and also prevents confusion with internal errors.

htlcswitch/failure.go
joostjager: You mean `type SphinxErrorDecrypter interface`? But what is implementing this interface then? `SphinxErrorDecrypter` offers a method `DecryptError` that depends on an underlying decryptor function `Decrypt`

joostjager:
@halseth ptal

routing/router.go
```diff
@@ -1822,19 +1822,32 @@ func (r *ChannelRouter) sendPayment(
...
+	log.Tracef("node=%x reported failure when sending htlc", failureSourceIdx)
 
-	log.Tracef("node=%x reported failure when sending htlc", errVertex)
+	var (
+		failureVertex route.Vertex
+		failureSource *btcec.PublicKey
+	)
+	if failureSourceIdx > 0 {
+		var err error
+		failureVertex = rt.Hops[failureSourceIdx-1].PubKeyBytes
```
Roasbeef: The failure index passed up from the onion layer isn't zero-indexed? If not IMO it should be.

routing/router.go
```diff
@@ -2036,45 +2049,38 @@ func (r *ChannelRouter) processSendError(paySession PaymentSession,
...
-	}
-
-	return edge{}, 0, fmt.Errorf("cannot find error source node in route")
+	// Determine if we have a failure from the final hop. If it is, we
+	// assume that the failing channel is the incoming channel.
+	//
+	// TODO(joostjager): In this case, certain types of failures are not
+	// expected. For example FailUnknownNextPeer. This could be a reason to
+	// prune the node?
+	if failureSource == len(route.Hops) {
```
Roasbeef: Similar question here re zero indexing.

htlcswitch/failure.go
Roasbeef: The `OnionErrorDecrypter` would implement this interface. A struct with a single field that is a function closure aspires to be an interface.

Roasbeef: Actually, we already have `ErrorDecrypter `, why is that insufficient?

routing/payment_lifecycle.go
```diff
@@ -17,7 +17,7 @@ import (
 type errNoRoute struct {
 	// lastError is the error encountered during the last payment attempt,
 	// if at least one attempt has been made.
-	lastError *htlcswitch.ForwardingError
+	lastError error
```
Roasbeef: Why should concrete type be abandoned here? We'll just need to type assert it elsewhere to uncover it again.

routing/router.go
joostjager: I propose to define index zero as the self node in the route. That indeed means that any failure received over the network will have at least index 1. 

There currently are locally generated failures. They occur even before we send out the network packet and get index zero. How would we deal with those if the onion layer would be zero-indexed?

Using two different definitions of failure index may be confusing.

routing/payment_lifecycle.go
joostjager: Because it can also be `ErrUnreadableFailureMessage` which is not of type `ForwardingError`. That explicit type assert might be good, to force users to not forget this case.

htlcswitch/failure.go
joostjager: `ErrorDecrypter` is a higher level interface that returns a `ForwardingError`.

Converted function closure to interface.

joostjager:
@roasbeef ptal

routing/router.go
```diff
@@ -1820,7 +1820,24 @@ func (r *ChannelRouter) sendPayment(
...
 // return value.
 func (r *ChannelRouter) processSendError(paySession PaymentSession,
-	rt *route.Route, fErr *htlcswitch.ForwardingError) bool {
+	rt *route.Route, sendErr error) bool {
+
+	// If the failure message could not be decrypted, attribute the failure
+	// to our own outgoing channel.
+	//
+	// TODO(joostager): Penalize all channels in the route.
+	if sendErr == htlcswitch.ErrUnreadableFailureMessage {
```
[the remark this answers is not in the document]
joostjager: Yes. But that is currently already the case, as a decryption error is transformed into a self node failure.

In the follow up pr with proper failure attribution, this will be fixed.

lnrpc/routerrpc/router.proto
```diff
@@ -183,11 +183,7 @@ message Failure {
     /// Failure code as defined in the Lightning spec
     FailureCode code = 1;
 
-    /**
-    The node pubkey of the intermediate or final node that generated the failure
-    message.
-    **/
-    bytes failure_source_pubkey = 2;
+    reserved 2;
```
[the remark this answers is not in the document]
joostjager: afaik the idea of the sub servers is that they are experimental and we can change interfaces without going through the deprecation cycle.

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
joostjager: updated

routing/router.go
```diff
@@ -1822,19 +1822,32 @@ func (r *ChannelRouter) sendPayment(
...
+	failureSourceIdx := fErr.FailureSourceIdx
+	log.Tracef("node=%x reported failure when sending htlc", failureSourceIdx)
 
-	log.Tracef("node=%x reported failure when sending htlc", errVertex)
+	var (
+		failureVertex route.Vertex
+		failureSource *btcec.PublicKey
+	)
+	if failureSourceIdx > 0 {
+		var err error
```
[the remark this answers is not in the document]
joostjager: moved above. not 100% sure if it is an improvement because it unnecessarily extends the scope of `err` too.

htlcswitch/switch.go
```diff
@@ -986,21 +990,17 @@ func (s *Switch) parseFailedPayment(deobfuscator ErrorDecrypter,
 		var err error
 		// We'll attempt to fully decrypt the onion encrypted
 		// error. If we're unable to then we'll bail early.
-		failure, err = deobfuscator.DecryptError(htlc.Reason)
+		failure, err := deobfuscator.DecryptError(htlc.Reason)
```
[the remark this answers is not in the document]
joostjager: removed

joostjager:
@cfromknecht ptal

htlcswitch/switch.go
```diff
@@ -983,24 +987,19 @@ func (s *Switch) parseFailedPayment(deobfuscator ErrorDecrypter,
...
 			userErr := fmt.Sprintf("unable to de-obfuscate "+
 				"onion failure (hash=%v, pid=%d): %v",
 				paymentHash, paymentID, err)
 			log.Error(userErr)
-			failure = &ForwardingError{
-				FailureSourceIdx: 0,
-				ExtraMsg:         userErr,
-				FailureMessage:   lnwire.NewTemporaryChannelFailure(nil),
-			}
+			return ErrUnreadableFailureMessage
```
cfromknecht: nit: since `userErr` is no longer returned in the failure, unable should be capitalized.

joostjager: fixed

joostjager:
The itest `single_hop_invoice` was failing because of a subtle bug caused by mixing values and pointers for failure messages. Added commit to fix that.

go.sum
```diff
@@ -137,8 +137,10 @@ github.com/lightninglabs/neutrino v0.0.0-20190313035638-e1ad4c33fb18 h1:lxD7RgKY
 github.com/lightninglabs/neutrino v0.0.0-20190313035638-e1ad4c33fb18/go.mod h1:v6tz6jbuAubTrRpX8ke2KH9sJxml8KlPQTKgo9mAp1Q=
 github.com/lightninglabs/neutrino v0.0.0-20190629001446-52dd89dd1aaa h1:vLsdHrqHKNDShoSRkrFEM0FsNt5waIKS87IgP1YLz+g=
 github.com/lightninglabs/neutrino v0.0.0-20190629001446-52dd89dd1aaa/go.mod h1:vzLU75ll8qbRJIzW5dvK/UXtR9c2FecJ6VNOM8chyVM=
-github.com/lightningnetwork/lightning-onion v0.0.0-20190430041606-751fb4dd8b72 h1:KgmypyQfJnEf2vhwboKCtTp4mHxIcLeXPBPWDbPuzFQ=
-github.com/lightningnetwork/lightning-onion v0.0.0-20190430041606-751fb4dd8b72/go.mod h1:Sooe/CoCqa85JxqHV+IBR2HW+6t2Cv+36awSmoccswM=
+github.com/lightningnetwork/lightning-onion v0.0.0-20190430041606-ecc936dc56c9ed8167a560b9a1fed0f9653abfc0 h1:RMBbf6V9Y12KhhZVTt74MJ8XQ2tINE1vk2SbLz/CPvc=
```
Roasbeef: This second one will disappear if you run `go mod tidy`.

routing/router.go
```diff
@@ -1822,19 +1822,31 @@ func (r *ChannelRouter) sendPayment(
 func (r *ChannelRouter) processSendError(paySession PaymentSession,
 	rt *route.Route, fErr *htlcswitch.ForwardingError) bool {
 
-	errSource := fErr.ErrorSource
-	errVertex := route.NewVertex(errSource)
+	failureSourceIdx := fErr.FailureSourceIdx
+	log.Tracef("node=%x reported failure when sending htlc", failureSourceIdx)
```
Roasbeef: IMO we should still log the pubkey as it makes it easier to obtain information from the logs. Otherwise, with only the index you'll need to  scroll back up many lines to count the nodes in the route, but even then the pubkey may be logged in a `btcec.PublicKey` only exposing the raw `big.Int`.

go.sum
joostjager: Fixed

routing/router.go
joostjager: Agreed and added.

routing/router.go
```diff
@@ -2036,45 +2052,38 @@ func (r *ChannelRouter) processSendError(paySession PaymentSession,
...
-	}
-
-	return edge{}, 0, fmt.Errorf("cannot find error source node in route")
+	// Determine if we have a failure from the final hop. If it is, we
+	// assume that the failing channel is the incoming channel.
+	//
+	// TODO(joostjager): In this case, certain types of failures are not
+	// expected. For example FailUnknownNextPeer. This could be a reason to
+	// prune the node?
+	if failureSource == len(route.Hops) {
```
halseth: Is this expected to happen? Should have a comment.

Roasbeef: Existing comment above stands, this is an old (?) case of the last hop sending an error. We decrement the failure source as our zero index is self, so if we have 2 hops, and a source index of 2, we need to access the second (`1`) element in the slice. I guess this portion could use a comment.

routing/router.go
```diff
@@ -1822,19 +1822,35 @@ func (r *ChannelRouter) sendPayment(
 func (r *ChannelRouter) processSendError(paySession PaymentSession,
 	rt *route.Route, fErr *htlcswitch.ForwardingError) bool {
 
-	errSource := fErr.ErrorSource
-	errVertex := route.NewVertex(errSource)
+	failureSourceIdx := fErr.FailureSourceIdx
```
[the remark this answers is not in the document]
joostjager: Fixed, but not sure if it is an improvement.

routing/router.go
```diff
@@ -1822,19 +1822,35 @@ func (r *ChannelRouter) sendPayment(
...
+	var (
+		failureVertex route.Vertex
+		failureSource *btcec.PublicKey
+		err           error
+	)
+	if failureSourceIdx > 0 {
```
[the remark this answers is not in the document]
joostjager: added

routing/router.go
joostjager: Yes, this is still something that could happen. This PR only adds the conversion to failure indices, it doesn't improve the interpretation itself. The follow up pr #3256 will deal with this.

Added explanation of the minus one.

joostjager:
PR rebased after failure reason changes were merged in #3156

lnwire/onion_error_test.go
```diff
@@ -180,7 +180,7 @@ func TestFailUnknownPaymentHashOptionalAmount(t *testing.T) {
 	onionError := FailUnknownPaymentHash{}
 
 	var b bytes.Buffer
-	if err := EncodeFailure(&b, onionError, 0); err != nil {
+	if err := EncodeFailure(&b, &onionError, 0); err != nil {
```
halseth: nit: take reference when var is declared instead

joostjager: Fixed. Also in the `reflect.DeepEqual` below, the same objects were compared.

PR #3191, opened 2019-06-11
Add option for deleting failed-only payments.

halseth:
Since we now store failed and in-flight payments as well in the database, this gives `DeleteAllPayments` the option to delete only failed payments.


halseth:
Should extend this to also delete failed routes for successful payments (configurable)

lnrpc/rpc.proto
```diff
@@ -2166,6 +2166,8 @@ message ListPaymentsResponse {
 }
 
 message DeleteAllPaymentsRequest {
+    /// Only delete failed payments.
```
guggero, line 2169: nit: tripple slash comment not necessary for API docs anymore.

guggero (commented):
Code looks good, needs a rebase though.

An additional flag `prune_htlcs` could indeed be useful. Could also be applied to failed payments then the two flags could operate independently.

halseth:
Rebased and added flag `failed_htlcs_only` that can be used to prune HTLCs.

lnrpc/rpc.proto
```diff
@@ -3223,6 +3223,11 @@ message ListPaymentsResponse {
 message DeleteAllPaymentsRequest {
     // Only delete failed payments.
     bool failed_payments_only = 1;
+
+    /*
+    Only delete failed HTLCs from payments, not not the payment itself.
```
guggero, line 3228: nit: double "not".

guggero (approved):
tACK, LGTM :tada: 

Maybe we could also add a CLI command for this RPC? We could set it to `Hidden: true` to avoid users deleting them by accident. But would be useful to have.
Or alternatively add a `gc-failed-payments-on-startup` config option?

rpcserver.go
```diff
@@ -5703,11 +5703,16 @@ func (r *rpcServer) ListPayments(ctx context.Context,
 
 // DeleteAllPayments deletes all outgoing payments from DB.
 func (r *rpcServer) DeleteAllPayments(ctx context.Context,
-	_ *lnrpc.DeleteAllPaymentsRequest) (*lnrpc.DeleteAllPaymentsResponse, error) {
+	req *lnrpc.DeleteAllPaymentsRequest) (
```
bhandras, line 5706: nit: consider squashing with previous commit (proto changes). (similarly for failed htlcs)

rpcserver.go
```diff
@@ -5706,11 +5706,12 @@ func (r *rpcServer) DeleteAllPayments(ctx context.Context,
 	req *lnrpc.DeleteAllPaymentsRequest) (
 	*lnrpc.DeleteAllPaymentsResponse, error) {
 
-	rpcsLog.Debugf("[DeleteAllPayments] failed_payments_only=%v",
-		req.FailedPaymentsOnly)
+	rpcsLog.Debugf("[DeleteAllPayments] failed_payments_only=%v, "+
```
bhandras, line 5709: I think this could be raised to info level since it's not common and only one line per RPC call.

bhandras (approved):
LGTM :1st_place_medal:

rpcserver.go
halseth: Squashed proto changes

rpcserver.go
halseth: Agreed 👍

PR #3190, opened 2019-06-11
ListPayments: filter out non-succeeded payments, include payment status

halseth:
Earlier versions of ListPayments only included completed payments. We
return to this behavior by ignore all other payments if the `nonSucceeded`
boolean is not set in the request

PR #3194, opened 2019-06-12
[integration tests]: restart Alice after on-chain settle

halseth:
To ensure lnd is able to pick up an on-chain preimage properly after a
restart, we suspend Alice and check that the payment is listed correctly
as succeeded after a restart.

Builds on #3190

### DIFF
--- a/lntest/itest/lnd_multi-hop_htlc_local_chain_claim_test.go
+++ b/lntest/itest/lnd_multi-hop_htlc_local_chain_claim_test.go
@@ -195,6 +195,13 @@ func testMultiHopHtlcLocalChainClaim(net *lntest.NetworkHarness, t *harnessTest)
 		}
 	}
 
+	// At this point we suspend Alice to make sure she'll handle the
+	// on-chain settle after a restart.
+	restartAlice, err := net.SuspendNode(net.Alice)
+	if err != nil {
+		t.Fatalf("unable to suspend alice: %v", err)
+	}
+
 	// Mine a block to confirm the two transactions (+ the coinbase).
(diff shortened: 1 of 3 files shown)


lntest/itest/lnd_multi-hop_htlc_local_chain_claim_test.go
```diff
@@ -420,3 +427,44 @@ func waitForInvoiceAccepted(t *harnessTest, node *lntest.HarnessNode,
 		}
 	}
 }
+
+// checkPaymentSucceeded asserts that the given node list a payment with the
+// given preimage as succeeded.
+func checkPaymentSucceeded(ctxt context.Context, node *lntest.HarnessNode,
```
wpaulino: Seems better to pass in the payment status here instead to make it more general?

halseth: Done: https://github.com/lightningnetwork/lnd/pull/3194/commits/b9a4971a964538b7eb9b97404f7648499b630727

halseth:
PTAL @wpaulino

cfromknecht:
Do we still test that the preimage is properly recovered w/o a restart?

Roasbeef:
Still relevant?

halseth:
> Still relevant?

Yes, will update this the coming week.

halseth:
Updated and rebase. PTAL.

> Do we still test that the preimage is properly recovered w/o a restart?

Now only doing this change for one of the tests, such that the other tests still will do the non-restart version.

PR #3197, opened 2019-06-12
Optimize prune zombie channels

roeierez:
This PR replaces https://github.com/lightningnetwork/lnd/pull/2799.
Since there were some major changes regarding zombie channels, I preferred to create a new PR instead of rebasing. It seemed easier to implement and review this way.

The PR introduces a different, more optimized, approach to prune zombie channels.
"zombie channels" stands for channels that their last channel policy update time was earlier than "channel expiry", configured by default to a period of 14 days or, when AssumeChannelValid=false, those that have both ChannelEdgePolicy with their disable bit on.
From my understanding the second part (disabled channels) of the prune serves as a heuristic to prune on-chain closed channels on devices with low resources.

The current approach iterates all channels in the graph in order to filter those in need. This approach is time consuming, several seconds on my mobile device for ~40,000 channels, while during this time the
db is locked in a transaction.

The proposed change is split into two parts:

**Prune expired channels**
Use an existing functionality that utilize the fact that channel update are saved indexed by date. This way enables us to go over only a small subset of the channels, only those that were updated before the "channel expiry" time and further filter them for our need.

**Prune disabled channels (AssumeChannelValid=false)**
Here I added an index bucket for disabled ChannelEdgePolicy that allows iterating over it when prune is needed. 

**Results**
The same graph that took several seconds to prune was pruned, after the change in several milliseconds.
Recently a progression was pushed to run the pruneZombieChans on startup (When AssumeChannelValid=false), as a result of that I think that the performance improvement is now more important and should take a higher priority.

My measurements for ~40,000 channels.

Samsung galaxy s7
prune time before change: 5-12 seconds, depends on the current load.
prune time after change: 15 milliseconds.

Mac book pro 2.8 GHz Intel Core i7
prune time before change: ~400 milliseconds
prune time after change: 15 milliseconds.

### DIFF
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -118,6 +118,18 @@ var (
 	// edge's participants.
 	zombieBucket = []byte("zombie-index")
 
+	// disabledEdgePolicyBucket is a sub-bucket of the main edgeBucket bucket
+	// responsible for maintaining an index of disabled edge policies. Each
+	// entry exists within the bucket as follows:
+	//
+	// maps: <chanID><direction> -> []byte{}
+	//
+	// The chanID represents the channel ID of the edge and the direction is
+	// one byte representing the direction of the edge. The main purpose of
+	// this index is to allow pruning disabled channels in a fast way without
+	// the need to iterate all over the graph.
+	disabledEdgePolicyBucket = []byte("disabled-edge-policy-index")
+
 	// graphMetaBucket is a top-level bucket which stores various meta-deta
 	// related to the on-disk channel graph. Data stored in this bucket
 	// includes the block to which the graph has been synced to, the total
@@ -256,6 +268,46 @@ func (c *ChannelGraph) ForEachNodeChannel(tx *bbolt.Tx, nodePub []byte,
 	db := c.db
 
 	return nodeTraversal(tx, nodePub, db, cb)
+}
+
+// DisabledChannelIDs returns the channel ids of disabled channels.
+// A channel is disabled when two of the associated ChanelEdgePolicies
+// have their disabled bit on.
+func (c *ChannelGraph) DisabledChannelIDs() ([]uint64, error) {
+	var disabledChanIDs []uint64
+	chanEdgeFound := make(map[uint64]struct{})
+
+	err := c.db.View(func(tx *bbolt.Tx) error {
+		edges := tx.Bucket(edgeBucket)
+		if edges == nil {
+			return ErrGraphNoEdgesFound
+		}
+
+		disabledEdgePolicyIndex := edges.Bucket(disabledEdgePolicyBucket)
+		if disabledEdgePolicyIndex == nil {
+			return nil
+		}
+
+		// We iterate over all disabled policies and we add each channel that
+		// has more than one disabled policy to disabledChanIDs array.
+		return disabledEdgePolicyIndex.ForEach(func(k, v []byte) error {
+			chanID := byteOrder.Uint64(k[:8])
+			_, edgeFound := chanEdgeFound[chanID]
+			if edgeFound {
+				delete(chanEdgeFound, chanID)
+				disabledChanIDs = append(disabledChanIDs, chanID)
+				return nil
+			}
+
+			chanEdgeFound[chanID] = struct{}{}
+			return nil
+		})
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return disabledChanIDs, nil
 }
 
 // ForEachNode iterates through all the stored vertices/nodes in the graph,
@@ -1821,6 +1873,11 @@ func delChannelEdge(edges, edgeIndex, chanIndex, zombieIndex,
 			return err
 		}
 	}
+
+	// As part of deleting the edge we also remove all disabled entries
+	// from the edgePolicyDisabledIndex bucket. We do that for both directions.
+	updateEdgePolicyDisabledIndex(edges, cid, false, false)
+	updateEdgePolicyDisabledIndex(edges, cid, true, false)
 
 	// With the edge data deleted, we can purge the information from the two
 	// edge indexes.
@@ -3647,7 +3704,45 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 		return err
 	}
 
+	updateEdgePolicyDisabledIndex(
+		edges, edge.ChannelID,
+		edge.ChannelFlags&lnwire.ChanUpdateDirection > 0,
+		edge.IsDisabled(),
+	)
+
 	return edges.Put(edgeKey[:], b.Bytes()[:])
+}
+
+// updateEdgePolicyDisabledIndex is used to update the disabledEdgePolicyIndex
+// bucket by either add a new disabled ChannelEdgePolicy or remove an existing
+// one.
+// The direction represents the direction of the edge and disabled is used for
+// deciding whether to remove or add an entry to the bucket.
+// In general a channel is disabled if two entries for the same chanID exist
+// in this bucket.
+// Maintaining the bucket this way allows a fast retrieval of disabled
+// channels, for example when prune is needed.
+func updateEdgePolicyDisabledIndex(edges *bbolt.Bucket, chanID uint64,
+	direction bool, disabled bool) error {
+
+	var disabledEdgeKey [8 + 1]byte
+	byteOrder.PutUint64(disabledEdgeKey[0:], chanID)
+	if direction {
+		disabledEdgeKey[8] = 1
+	}
+
+	disabledEdgePolicyIndex, err := edges.CreateBucketIfNotExists(
+		disabledEdgePolicyBucket,
+	)
+	if err != nil {
+		return err
+	}
+
+	if disabled {
+		return disabledEdgePolicyIndex.Put(disabledEdgeKey[:], []byte{})
+	}
+
+	return disabledEdgePolicyIndex.Delete(disabledEdgeKey[:])
 }
 
 // putChanEdgePolicyUnknown marks the edge policy as unknown
--- a/channeldb/graph_test.go
+++ b/channeldb/graph_test.go
@@ -2680,6 +2680,102 @@ func TestNodeIsPublic(t *testing.T) {
 	)
 }
 
+// TestDisabledChannelIDs ensures that the disabled channels within the
+// disabledEdgePolicyBucket are managed properly and the list returned from
+// DisabledChannelIDs is correct.
+func TestDisabledChannelIDs(t *testing.T) {
+	t.Parallel()
+
+	db, cleanUp, err := makeTestDB()
+	if err != nil {
+		t.Fatalf("unable to make test database: %v", err)
+	}
+	defer cleanUp()
+
+	graph := db.ChannelGraph()
+
+	// Create first node and add it to the graph.
+	node1, err := createTestVertex(db)
+	if err != nil {
+		t.Fatalf("unable to create test node: %v", err)
+	}
+	if err := graph.AddLightningNode(node1); err != nil {
+		t.Fatalf("unable to add node: %v", err)
+	}
+
+	// Create second node and add it to the graph.
+	node2, err := createTestVertex(db)
+	if err != nil {
+		t.Fatalf("unable to create test node: %v", err)
+	}
+	if err := graph.AddLightningNode(node2); err != nil {
+		t.Fatalf("unable to add node: %v", err)
+	}
+
+	// Adding a new channel edge to the graph.
+	edgeInfo, edge1, edge2 := createChannelEdge(db, node1, node2)
+	if err := graph.AddLightningNode(node2); err != nil {
+		t.Fatalf("unable to add node: %v", err)
+	}
+
+	if err := graph.AddChannelEdge(edgeInfo); err != nil {
+		t.Fatalf("unable to create channel edge: %v", err)
+	}
+
+	// Ensure no disabled channels exist in the bucket on start.
+	disabledChanIds, err := graph.DisabledChannelIDs()
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) > 0 {
+		t.Fatalf("expected empty disabled channels, got %v disabled channels",
+			len(disabledChanIds))
+	}
+
+	// Add one disabled policy and ensure the channel is still not in the
+	// disabled list.
+	edge1.ChannelFlags |= lnwire.ChanUpdateDisabled
+	if err := graph.UpdateEdgePolicy(edge1); err != nil {
+		t.Fatalf("unable to update edge: %v", err)
+	}
+	disabledChanIds, err = graph.DisabledChannelIDs()
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) > 0 {
+		t.Fatalf("expected empty disabled channels, got %v disabled channels",
+			len(disabledChanIds))
+	}
+
+	// Add second disabled policy and ensure the channel is now in the
+	// disabled list.
+	edge2.ChannelFlags |= lnwire.ChanUpdateDisabled
+	if err := graph.UpdateEdgePolicy(edge2); err != nil {
+		t.Fatalf("unable to update edge: %v", err)
+	}
+	disabledChanIds, err = graph.DisabledChannelIDs()
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) != 1 || disabledChanIds[0] != edgeInfo.ChannelID {
+		t.Fatalf("expected disabled channel with id %v, "+
+			"got %v", edgeInfo.ChannelID, disabledChanIds)
+	}
+
+	// Delete the channel edge and ensure it is removed from the disabled list.
+	if err = graph.DeleteChannelEdges(edgeInfo.ChannelID); err != nil {
+		t.Fatalf("unable to delete channel edge: %v", err)
+	}
+	disabledChanIds, err = graph.DisabledChannelIDs()
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) > 0 {
+		t.Fatalf("expected empty disabled channels, got %v disabled channels",
+			len(disabledChanIds))
+	}
+}
+
 // TestEdgePolicyMissingMaxHtcl tests that if we find a ChannelEdgePolicy in
 // the DB that indicates that it should support the htlc_maximum_value_msat
 // field, but it is not part of the opaque data, then we'll handle it as it is
--- a/routing/pathfind_test.go
+++ b/routing/pathfind_test.go
@@ -314,6 +314,7 @@ type testChannelPolicy struct {
 	FeeRate     lnwire.MilliSatoshi
 	LastUpdate  time.Time
 	Disabled    bool
+	Direction   bool
 }
 
 type testChannelEnd struct {
@@ -346,6 +347,9 @@ func symmetricTestChannel(alias1 string, alias2 string, capacity btcutil.Amount,
 		id = chanID[0]
 	}
 
+	node2Policy := *policy
+	node2Policy.Direction = !policy.Direction
+
 	return &testChannel{
 		Capacity: capacity,
 		Node1: &testChannelEnd{
@@ -354,7 +358,7 @@ func symmetricTestChannel(alias1 string, alias2 string, capacity btcutil.Amount,
 		},
 		Node2: &testChannelEnd{
 			Alias:             alias2,
-			testChannelPolicy: policy,
+			testChannelPolicy: &node2Policy,
 		},
 		ChannelID: id,
 	}
@@ -527,6 +531,9 @@ func createTestGraphFromChannels(testChannels []*testChannel, source string) (
 			if testChannel.Node1.Disabled {
 				channelFlags |= lnwire.ChanUpdateDisabled
 			}
+			if testChannel.Node1.Direction {
+				channelFlags |= lnwire.ChanUpdateDirection
+			}
 			edgePolicy := &channeldb.ChannelEdgePolicy{
 				SigBytes:                  testSig.Serialize(),
 				MessageFlags:              msgFlags,
@@ -549,9 +556,12 @@ func createTestGraphFromChannels(testChannels []*testChannel, source string) (
 			if testChannel.Node2.MaxHTLC != 0 {
 				msgFlags |= lnwire.ChanUpdateOptionMaxHtlc
 			}
-			channelFlags := lnwire.ChanUpdateDirection
+			channelFlags := lnwire.ChanUpdateChanFlags(0)
 			if testChannel.Node2.Disabled {
 				channelFlags |= lnwire.ChanUpdateDisabled
+			}
+			if testChannel.Node2.Direction {
+				channelFlags |= lnwire.ChanUpdateDirection
 			}
 			edgePolicy := &channeldb.ChannelEdgePolicy{
 				SigBytes:                  testSig.Serialize(),
--- a/routing/router.go
+++ b/routing/router.go
@@ -734,7 +734,7 @@ func (r *ChannelRouter) syncGraphWithChain() error {
 // usually signals that a channel has been closed on-chain. We do this
 // periodically to keep a healthy, lively routing table.
 func (r *ChannelRouter) pruneZombieChans() error {
-	var chansToPrune []uint64
+	chansToPrune := make(map[uint64]struct{})
 	chanExpiry := r.cfg.ChannelPruneExpiry
 
 	log.Infof("Examining channel graph for zombie channels")
@@ -743,6 +743,11 @@ func (r *ChannelRouter) pruneZombieChans() error {
 	// collection due to being zombies.
 	filterPruneChans := func(info *channeldb.ChannelEdgeInfo,
 		e1, e2 *channeldb.ChannelEdgePolicy) error {
+
+		// Exit early in case this channel is already marked to be pruned
+		if _, markedToPrune := chansToPrune[info.ChannelID]; markedToPrune {
+			return nil
+		}
 
 		// We'll ensure that we don't attempt to prune our *own*
 		// channels from the graph, as in any case this should be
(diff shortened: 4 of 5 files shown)


roeierez:
10 seconds delay on every startup is a killer for every mobile app that hosts LND. Don’t know exactly what P3 means but I am guessing it is not going to be reviewed soon.
IMHO the goal of this PR is important enough to put it in higher priority.

Crypt-iQ:
@roeierez  I was looking at code similar to this and another change that can be done is here during the `syncGraphWithChain` function:
https://github.com/lightningnetwork/lnd/blob/0e28ecd6164e77e86f38bc3ea692a4a60896acdf/routing/router.go#L719-L724

`PruneGraph` could instead be called outside of the loop after processing all of the blocks.  When `pruneGraphNodes` is finally called, it allocates a very large map: 
https://github.com/lightningnetwork/lnd/blob/e65b8e38056425d336950eeff326067469ad03cc/channeldb/graph.go#L928

If there are a lot of blocks to be processed, this map will be allocated for every block processed and on my lnd node, this allocated about 200MB on the heap with my ~2k nodes and ~8k channels.  More blocks also correlates with more time calling `PruneGraph`.  When the call is moved out of the loop, it only takes up 1MB on the heap and takes considerably less time after a long time offline.

roeierez:
@Crypt-iQ I think it is certainly a good point! 
As this PR is dealing with optimizing the case for light node (assumechannelvalid=true) and the PruneGraph is not called there I tend to think this change deserve its own discussion in a different PR.
What do you think?

Crypt-iQ:
> @Crypt-iQ I think it is certainly a good point!
> As this PR is dealing with optimizing the case for light node (assumechannelvalid=true) and the PruneGraph is not called there I tend to think this change deserve its own discussion in a different PR.
> What do you think?

Ah yeah didn't know this was exclusively for light clients, I'll make my own PR

channeldb/graph.go
```diff
@@ -236,6 +248,44 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
 	})
 }
 
+// DisabledChannelIDs returns the channel ids of disabled channels.
+// A channel is disabled when two of the associated ChanelEdgePolicies
+// have their disabled bit on.
+func (c *ChannelGraph) DisabledChannelIDs() ([]uint64, error) {
+
```
cfromknecht: nit: extra new line

channeldb/graph.go
```diff
@@ -236,6 +248,44 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+	var disabledChanIDs []uint64
+	chanEdgeFound := make(map[uint64]bool)
```
cfromknecht: looks like this only needs to be a set of the channel ids, so we can make this `map[uint64]struct{}`

channeldb/graph.go
```diff
@@ -236,6 +248,44 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+				delete(chanEdgeFound, chanID)
+				disabledChanIDs = append(disabledChanIDs, chanID)
+				return nil
+			}
+
+			chanEdgeFound[chanID] = true
+			return nil
+		})
+		return nil
+	})
```
cfromknecht: preferably we would return the `nil, err` if an error was encountered, and `disabledChanIDs, nil` below if no failure was encountered to prevent returning two non-nil values.

channeldb/graph.go
```diff
@@ -3616,9 +3669,34 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
+		disabledEdgeKey[8] = 1
+	}
+
+	disabledEdgePolicyIndex, err := edges.CreateBucketIfNotExists(disabledEdgePolicyBucket)
+	if err != nil {
+		return err
+	}
+
+	if disabled {
+		err = disabledEdgePolicyIndex.Put(disabledEdgeKey[:], nil)
```
cfromknecht: we never want to pass `nil` as the value to bolt, instead we should pass `[]byte{}`. bolt represents buckets as keys with nil values, which can cause undefined behavior if values are also nil. it is accepted in order to maintain backwards compatibility, but is generally not advised

channeldb/graph.go
```diff
@@ -3616,9 +3669,34 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 		return err
 	}
 
+	updateEdgePolicyDisabledIndex(edges, edge.ChannelID,
+		edge.ChannelFlags&lnwire.ChanUpdateDirection > 0, edge.IsDisabled())
+
 	return edges.Put(edgeKey[:], b.Bytes()[:])
 }
 
+func updateEdgePolicyDisabledIndex(edges *bbolt.Bucket, chanID uint64,
```
cfromknecht: needs godoc

channeldb/graph.go
```diff
@@ -3616,9 +3669,34 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
+	initiator bool, disabled bool) error {
```
cfromknecht: is `initiator` the appropriate name? seems instead like this just the direction, which is determined by comparing the node pubkeys

channeldb/graph.go
```diff
@@ -1800,6 +1850,9 @@ func delChannelEdge(edges, edgeIndex, chanIndex, zombieIndex,
 		}
 	}
 
+	updateEdgePolicyDisabledIndex(edges, cid, false, false)
```
cfromknecht: some comments would help explain what's happening here, e.g. we remove the entries from the disabled index since we are deleting the channel

channeldb/graph.go
```diff
@@ -118,6 +118,18 @@ var (
 	// edge's participants.
 	zombieBucket = []byte("zombie-index")
 
+	// disabledEdgePolicyBucket is a sub-bucket of the main edgeBucket bucket
+	// responsible for maintaining an index of disabled edge policies.
+	// Each entry exists within the bucket as follows:
+	//
+	// maps: <chanID><direction> -> nil
```
cfromknecht: as per my other comment, `nil` here should become `[]byte{}`

channeldb/graph.go
```diff
@@ -3616,9 +3669,34 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
+	if err != nil {
+		return err
+	}
+
+	if disabled {
+		err = disabledEdgePolicyIndex.Put(disabledEdgeKey[:], nil)
+	} else {
+		err = disabledEdgePolicyIndex.Delete(disabledEdgeKey[:])
+	}
+	return err
```
cfromknecht: style nit: return directly each case
```golang
if disabled {
     return put()
}

return delete()
```

routing/pathfind_test.go
```diff
@@ -518,10 +522,13 @@ func createTestGraphFromChannels(testChannels []*testChannel) (*testGraphInstanc
 			if testChannel.Node1.MaxHTLC != 0 {
 				msgFlags |= lnwire.ChanUpdateOptionMaxHtlc
 			}
-			var channelFlags lnwire.ChanUpdateChanFlags
+			channelFlags := lnwire.ChanUpdateChanFlags(0)
```
cfromknecht: reason for this change? it looks equivalent?

routing/pathfind_test.go
```diff
@@ -314,6 +314,7 @@ type testChannelPolicy struct {
 	FeeRate     lnwire.MilliSatoshi
 	LastUpdate  time.Time
 	Disabled    bool
+	Initiator   bool
```
cfromknecht: I think this is better named `Direction`, since it doesn't reflect who initiated the channel

routing/router.go
```diff
@@ -817,25 +817,48 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
+		chansToPrune[info.ChannelID] = true
 
 		return nil
 	}
 
-	err := r.cfg.Graph.ForEachChannel(filterPruneChans)
+	startTime := time.Unix(0, 0)
+	endTime := time.Now().Add(-1 * chanExpiry)
+	updates, err := r.cfg.Graph.ChanUpdatesInHorizon(startTime, endTime)
+
```
cfromknecht: nit: extra new line

routing/router.go
```diff
@@ -742,7 +742,7 @@ func (r *ChannelRouter) syncGraphWithChain() error {
 // usually signals that a channel has been closed on-chain. We do this
 // periodically to keep a healthy, lively routing table.
 func (r *ChannelRouter) pruneZombieChans() error {
-	var chansToPrune []uint64
+	chansToPrune := make(map[uint64]bool)
```
cfromknecht: similar comment here, this can just be a set of `map[uint64]struct{}`

routing/router.go
```diff
@@ -817,25 +817,48 @@ func (r *ChannelRouter) pruneZombieChans() error {
 			info.ChannelID)
 
 		// TODO(roasbeef): add ability to delete single directional edge
-		chansToPrune = append(chansToPrune, info.ChannelID)
+		chansToPrune[info.ChannelID] = true
```
cfromknecht: can this map also be used to exit this function early if the channel has already been processed?

routing/router_test.go
```diff
@@ -1957,7 +1957,7 @@ func TestPruneChannelGraphStaleEdges(t *testing.T) {
 	t.Parallel()
 
 	freshTimestamp := time.Now()
-	staleTimestamp := time.Time{}
+	staleTimestamp := time.Unix(0, 0)
```
cfromknecht: reason for this change?

routing/router.go
```diff
@@ -817,25 +817,48 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
+
+	edges, err := r.cfg.Graph.FetchChanInfos(disabledChanIDs)
 	if err != nil {
-		return fmt.Errorf("unable to filter local zombie channels: "+
-			"%v", err)
+		return fmt.Errorf("Unable to filter local zombie "+
+			"chans: %v", err)
+	}
+
+	updates = append(updates, edges...)
```
cfromknecht: nit: kinda confused me for a while that updates and edges could be appended together, then i realized they're both `[]ChannelEdges`. perhaps `oldEdges` and `disabledEdges` would help convey what the two sets contain?

routing/router_test.go
roeierez: @cfromknecht `time.Time{}.Unix()` results in a negative number. 
While querying boltdb in ChanUpdatesInHorizon (https://github.com/lightningnetwork/lnd/blob/master/channeldb/graph.go#L1380) there is a conversion to uint64 which results in a very large number serialized to `startTimeBytes`
I think this is OK and only the test should be updated to better reflect the reality.

channeldb/graph.go
roeierez: removed.

channeldb/graph.go
roeierez: yes, that's better. changed.

channeldb/graph.go
roeierez: fixed.

channeldb/graph.go
roeierez: I am aware of that now. fixed.

channeldb/graph.go
roeierez: Added.

channeldb/graph.go
roeierez: Agree. Fixed.

channeldb/graph.go
roeierez: Fixed.

channeldb/graph.go
roeierez: Fixed.

routing/pathfind_test.go
roeierez: I guess no reason. reverted.

routing/pathfind_test.go
roeierez: Agree. Changed to use Direction.

routing/router.go
roeierez: Line removed.

routing/router.go
roeierez: Done.

routing/router.go
roeierez: Good one. added logic at the beginning of the function.

routing/router.go
roeierez: Naming changed.

roeierez:
@cfromknecht thanks for your valuable review.
I believed I addressed the feedback. Ready for another pass.

routing/router_test.go
cfromknecht: Very interesting, TIL!

channeldb/graph.go
```diff
@@ -236,6 +248,47 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+				delete(chanEdgeFound, chanID)
+				disabledChanIDs = append(disabledChanIDs, chanID)
+				return nil
+			}
+
+			chanEdgeFound[chanID] = struct{}{}
+			return nil
+		})
+	})
+
```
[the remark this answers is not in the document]
roeierez: fixed.

channeldb/graph.go
```diff
@@ -236,6 +248,47 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
...
+// DisabledChannelIDs returns the channel ids of disabled channels.
+// A channel is disabled when two of the associated ChanelEdgePolicies
+// have their disabled bit on.
+func (c *ChannelGraph) DisabledChannelIDs() ([]uint64, error) {
+	var disabledChanIDs []uint64
+	chanEdgeFound := make(map[uint64]struct{})
+
+	err := c.db.View(func(tx *bbolt.Tx) error {
+		edges := tx.Bucket(edgeBucket)
+		if edges == nil {
```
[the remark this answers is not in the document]
roeierez: fixed to return `ErrGraphNoEdgesFound` error

channeldb/graph.go
```diff
@@ -1800,6 +1853,11 @@ func delChannelEdge(edges, edgeIndex, chanIndex, zombieIndex,
 		}
 	}
 
+	// as part of deleting the edge we also remove all disabled entries
```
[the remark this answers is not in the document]
roeierez: fixed.

channeldb/graph.go
```diff
@@ -118,6 +118,18 @@ var (
 	// edge's participants.
 	zombieBucket = []byte("zombie-index")
 
+	// disabledEdgePolicyBucket is a sub-bucket of the main edgeBucket bucket
+	// responsible for maintaining an index of disabled edge policies.
+	// Each entry exists within the bucket as follows:
```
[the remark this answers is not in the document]
roeierez: @wpaulino these three lines are bellow 80 characters length. I am not sure I understand the feedback here. Did you mean to move the "Each" one line above?

channeldb/graph.go
```diff
@@ -3616,9 +3674,42 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 		return err
 	}
 
+	updateEdgePolicyDisabledIndex(edges, edge.ChannelID,
+		edge.ChannelFlags&lnwire.ChanUpdateDirection > 0, edge.IsDisabled())
```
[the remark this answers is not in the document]
roeierez: fixed.

channeldb/graph.go
```diff
@@ -3616,9 +3674,42 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
+
 	return edges.Put(edgeKey[:], b.Bytes()[:])
 }
 
+// updateEdgePolicyDisabledIndex is used to update the disabledEdgePolicyIndex
+// bucket by either add a new disabled ChannelEdgePolicy or remove an existing
+// one.
+// The direction represents the direction of the edge and disabled is used for
+// deciding whether to remvoe or add an entry to the bucket.
```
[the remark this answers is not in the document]
roeierez: fixed.

routing/router.go
```diff
@@ -752,6 +752,11 @@ func (r *ChannelRouter) pruneZombieChans() error {
 	filterPruneChans := func(info *channeldb.ChannelEdgeInfo,
 		e1, e2 *channeldb.ChannelEdgePolicy) error {
 
+		// exit early in case this channel is already marked to be pruned
```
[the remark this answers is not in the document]
roeierez: fixed.

routing/router.go
```diff
@@ -817,25 +822,47 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
 
 		return nil
 	}
 
-	err := r.cfg.Graph.ForEachChannel(filterPruneChans)
+	startTime := time.Unix(0, 0)
+	endTime := time.Now().Add(-1 * chanExpiry)
+	oldEdges, err := r.cfg.Graph.ChanUpdatesInHorizon(startTime, endTime)
+	if err != nil {
+		return fmt.Errorf("Unable to filter local zombie "+
```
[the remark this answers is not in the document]
roeierez: fixed for all cases.

channeldb/graph.go
```diff
@@ -236,6 +248,47 @@ func (c *ChannelGraph) ForEachChannel(cb func(*ChannelEdgeInfo, *ChannelEdgePoli
 	})
 }
 
+// DisabledChannelIDs returns the channel ids of disabled channels.
+// A channel is disabled when two of the associated ChanelEdgePolicies
+// have their disabled bit on.
+func (c *ChannelGraph) DisabledChannelIDs() ([]uint64, error) {
```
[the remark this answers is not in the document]
roeierez: Sure thing! test added in graph_test.go `TestDisableChannelIDs`

channeldb/graph_test.go
```diff
@@ -2680,6 +2680,103 @@ func TestNodeIsPublic(t *testing.T) {
...
 }
 
+// TestDisableChannelIDs ensures that the disabled channels within the
+// disabledEdgePolicyBucket are managed properly and the list returned from
+// DisabledChannelIDs is correct.
+func TestDisableChannelIDs(t *testing.T) {
+	t.Parallel()
+
+	db, cleanUp, err := makeTestDB()
+	defer cleanUp()
```
cfromknecht: `cleanUp()` should only be deferred if `err != nil`

channeldb/graph_test.go
```diff
@@ -2680,6 +2680,103 @@ func TestNodeIsPublic(t *testing.T) {
 	)
 }
 
+// TestDisableChannelIDs ensures that the disabled channels within the
+// disabledEdgePolicyBucket are managed properly and the list returned from
+// DisabledChannelIDs is correct.
+func TestDisableChannelIDs(t *testing.T) {
```
cfromknecht: great test!!

routing/router.go
```diff
@@ -817,25 +822,47 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
+	for _, u := range edgesToFilter {
+		filterPruneChans(u.Info, u.Policy1, u.Policy2)
 	}
 
 	log.Infof("Pruning %v zombie channels", len(chansToPrune))
 
 	// With the set of zombie-like channels obtained, we'll do another pass
 	// to delete them from the channel graph.
-	for _, chanID := range chansToPrune {
+	var toPrune []uint64
```
cfromknecht: do we need to duplicate this slice? seems like the old logic would work fine?

channeldb/graph_test.go
```diff
@@ -2680,6 +2680,103 @@ func TestNodeIsPublic(t *testing.T) {
...
+	edge1.ChannelFlags |= lnwire.ChanUpdateDisabled
+	if err := graph.UpdateEdgePolicy(edge1); err != nil {
+		t.Fatalf("unable to update edge: %v", err)
+	}
+	disabledChanIds, err = graph.DisabledChannelIDs()
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) > 0 {
+		t.Fatalf("expected empty disabled channels, got %v disabled channels", len(disabledChanIds))
```
cfromknecht: nit: long line

routing/router.go
cfromknecht: ah you're right, good point

channeldb/graph.go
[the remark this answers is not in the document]
roeierez: Fixed.

channeldb/graph.go
```diff
@@ -3616,9 +3673,44 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
...
+func updateEdgePolicyDisabledIndex(edges *bbolt.Bucket, chanID uint64,
+	direction bool, disabled bool) error {
+
+	var disabledEdgeKey [8 + 1]byte
+	byteOrder.PutUint64(disabledEdgeKey[0:], chanID)
+	if direction {
+		disabledEdgeKey[8] = 1
+	}
+
+	disabledEdgePolicyIndex, err := edges.CreateBucketIfNotExists(disabledEdgePolicyBucket)
```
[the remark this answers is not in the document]
roeierez: Fixed.

channeldb/graph.go
```diff
@@ -3616,9 +3673,44 @@ func putChanEdgePolicy(edges, nodes *bbolt.Bucket, edge *ChannelEdgePolicy,
 		return err
 	}
 
+	updateEdgePolicyDisabledIndex(
+		edges, edge.ChannelID,
+		edge.ChannelFlags&lnwire.ChanUpdateDirection > 0,
+		edge.IsDisabled())
```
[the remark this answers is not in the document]
roeierez: Fixed.

channeldb/graph_test.go
roeierez: Fixed.

channeldb/graph_test.go
roeierez: Fixed.

channeldb/graph_test.go
roeierez: Fixed.

channeldb/graph_test.go
```diff
@@ -2680,6 +2680,103 @@ func TestNodeIsPublic(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) != 1 || disabledChanIds[0] != edgeInfo.ChannelID {
+		t.Fatalf("expected disabled channel with id %v, "+
+			"got %v", edgeInfo.ChannelID, disabledChanIds)
+	}
+
+	// Delete the channel edge and ensure it is removed from the disabled list.
+	graph.DeleteChannelEdges(edgeInfo.ChannelID)
```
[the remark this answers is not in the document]
roeierez: Fixed.

channeldb/graph_test.go
```diff
@@ -2680,6 +2680,103 @@ func TestNodeIsPublic(t *testing.T) {
...
+	disabledChanIds, err = graph.DisabledChannelIDs()
+	if err != nil {
+		t.Fatalf("unable to get disabled channel ids: %v", err)
+	}
+	if len(disabledChanIds) > 0 {
+		t.Fatalf(
```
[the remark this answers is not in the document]
roeierez: Fixed.

roeierez:
> it seems like the fixups and test addition were added to the wrong commit

Right. I rebased with separated commits.

Roasbeef:
Needs a rebase, but then ready to land afterwards.

channeldb/graph.go
```diff
@@ -118,6 +118,18 @@ var (
 	// edge's participants.
 	zombieBucket = []byte("zombie-index")
 
+	// disabledEdgePolicyBucket is a sub-bucket of the main edgeBucket bucket
```
Roasbeef: No migration to populate the bucket for all the _existing_ disabled channels? Once nodes update, even if all the channels in the graph are currently disabled, the `DisabledChannelIDs` method will return an empty slice. 

This also means that any double disable channels which exist (lets say the node is dormant right now but has that double disable state, and hasn't yet ticked to prune again), will only be pruned once they get old enough, rather than immediately.

routing/router.go
```diff
@@ -817,25 +822,47 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
-	err := r.cfg.Graph.ForEachChannel(filterPruneChans)
+	startTime := time.Unix(0, 0)
+	endTime := time.Now().Add(-1 * chanExpiry)
+	oldEdges, err := r.cfg.Graph.ChanUpdatesInHorizon(startTime, endTime)
+	if err != nil {
+		return fmt.Errorf("unable to filter local zombie "+
+			"chans: %v", err)
+	}
+
+	disabledChanIDs, err := r.cfg.Graph.DisabledChannelIDs()
```
wpaulino: Looking at this again, we should only be calling this when `AssumeChannelValid` is set. We can also remove the double disabled logic within the filtering closure above and avoid filtering these as we already know they're double disabled.

channeldb/graph.go
wpaulino: Nice catch! I wonder if writing this migration, even if it's a simple one, will actually be worth it though. It seems like this would only really affect nodes that haven't started up in a while. Path finding already ignores disabled edges, but the larger graph will likely incur a performance hit (though we don't have any concrete results).

routing/router.go
roeierez: @wpaulino it will be nice to be able to populate the disabledChanIDs straight into the map without loading the ChannelEdge from the db and passing it through the filtering closure.
One caveat though is that there is a check there to skip self channels. Any idea on how to do that filtering with the channelID alone? Or if that filtering is needed at the disabled case at all?

wpaulino: We could provide an optional compressed public key to `DisabledChannelIDs` and do the filtering there.

roeierez: @wpaulino it seems that loading the ChannelEdgeInfo is inevitable in order to get the pubkey for filtering, so I added a small helper function `isSelfChannelEdge` and used it to filter the disabled channels only if AssumeChannelValid=true.
Did this in a separate commit so you can easily review. Once the review is finished I will sqash.
Also rebased.

channeldb/graph.go
cfromknecht: This seems okay, new nodes will see the effects immediately while existing nodes will see them fully after ~2 weeks.

routing/router.go
```diff
@@ -772,62 +781,69 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
+	for _, u := range oldEdges {
+		filterPruneChans(u.Info, u.Policy1, u.Policy2)
 	}
 
 	log.Infof("Pruning %v zombie channels", len(chansToPrune))
 
 	// With the set of zombie-like channels obtained, we'll do another pass
 	// to delete them from the channel graph.
-	for _, chanID := range chansToPrune {
+	var toPrune []uint64
```
wpaulino: Nit: can initialize this to the size of the `chansToPrune` map.

routing/router.go
```diff
@@ -772,62 +781,69 @@ func (r *ChannelRouter) pruneZombieChans() error {
...
 	}
 
-	err := r.cfg.Graph.ForEachChannel(filterPruneChans)
+	// If AssumeChannelValid is present we'll look at the disabled bit for both
+	// edges. If they're both disabled, then we can interpret this as the 
+	// channel being closed and can prune it from our graph.	
+	if r.cfg.AssumeChannelValid {
+		disabledChanIDs, err := r.cfg.Graph.DisabledChannelIDs()
+		if err != nil {
+			return fmt.Errorf("unable to filter local zombie "+
```
wpaulino: Nit: we should differentiate the errors being returned from this and the other method calls below.

routing/router.go
roeierez: fixed.

routing/router.go
roeierez: fixed.

PR #3198, opened 2019-06-12
Avoid modifying bucket during ForEach loop, and Cursor traversal

halseth:
bbolt documentation states that this is not safe and leads to undefined
behavior.

`channeldb/migrations` and `contractcourt/briefcase` are the only places in the codebase where the `ForEach` modification was done. `

`channeldb` had a few cases of cursor modification.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -2087,10 +2087,6 @@ func (c *OpenChannel) CloseChannel(summary *ChannelCloseSummary) error {
 		// information stored within the revocation log.
 		logBucket := chanBucket.Bucket(revocationLogBucket)
 		if logBucket != nil {
-			err := wipeChannelLogEntries(logBucket)
-			if err != nil {
-				return err
-			}
 			err = chanBucket.DeleteBucket(revocationLogBucket)
 			if err != nil {
 				return err
@@ -2712,17 +2708,4 @@ func fetchChannelLogEntry(log *bbolt.Bucket,
 
 	commitReader := bytes.NewReader(commitBytes)
 	return deserializeChanCommit(commitReader)
-}
-
-func wipeChannelLogEntries(log *bbolt.Bucket) error {
-	// TODO(roasbeef): comment
-
-	logCursor := log.Cursor()
-	for k, _ := logCursor.First(); k != nil; k, _ = logCursor.Next() {
-		if err := logCursor.Delete(); err != nil {
-			return err
-		}
-	}
-
-	return nil
 }
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -1061,6 +1061,9 @@ func (c *ChannelGraph) DisconnectBlockAtHeight(height uint32) ([]*ChannelEdgeInf
 
 		// Scan from chanIDStart to chanIDEnd, deleting every
 		// found edge.
+		// NOTE: we must delete the edges after the cursor loop, since
+		// modifying the bucket while traversing is not safe.
+		var keys [][]byte
 		cursor := edgeIndex.Cursor()
 		for k, v := cursor.Seek(chanIDStart[:]); k != nil &&
 			bytes.Compare(k, chanIDEnd[:]) <= 0; k, v = cursor.Next() {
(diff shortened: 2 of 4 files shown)


channeldb/channel.go
```diff
@@ -2717,9 +2717,16 @@ func fetchChannelLogEntry(log *bbolt.Bucket,
 func wipeChannelLogEntries(log *bbolt.Bucket) error {
 	// TODO(roasbeef): comment
 
+	var keys [][]byte
 	logCursor := log.Cursor()
 	for k, _ := logCursor.First(); k != nil; k, _ = logCursor.Next() {
```
wpaulino: Seems cleaner to just do a `ForEach` on the bucket here?

channeldb/graph.go
```diff
@@ -1061,24 +1061,34 @@ func (c *ChannelGraph) DisconnectBlockAtHeight(height uint32) ([]*ChannelEdgeInf
...
+			copy(key[:], k[:])
+			keys = append(keys, key)
+
 			edgeInfoReader := bytes.NewReader(v)
 			edgeInfo, err := deserializeChanEdgeInfo(edgeInfoReader)
 			if err != nil {
 				return err
 			}
+
+			removedChans = append(removedChans, &edgeInfo)
```
wpaulino: Any reason for moving this up here?

channeldb/migrations.go
```diff
@@ -226,12 +231,28 @@ func migrateInvoiceTimeSeries(tx *bbolt.Tx) error {
 			return err
 		}
 
-		return invoices.Put(invoiceNum, b.Bytes())
+		// Make a copy of the key since it is only valid within the
+		// function closure.
```
wpaulino: Why is this the case?

cfromknecht: https://github.com/etcd-io/bbolt#foreach
> Please note that keys and values in ForEach() are only valid while the transaction is open. If you need to use a key or value outside of the transaction, you must use copy() to copy it to another byte slice.

wpaulino: Doesn't seem necessary then since everything's happening within a single database transaction.

cfromknecht: mhm see my other comment

channeldb/channel.go
```diff
@@ -2717,9 +2717,16 @@ func fetchChannelLogEntry(log *bbolt.Bucket,
...
-		if err := logCursor.Delete(); err != nil {
+		key := make([]byte, len(k))
+		copy(key[:], k[:])
```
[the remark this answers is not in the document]
cfromknecht: on second thought, this whole method can be replaced by `DeleteBucket(log)`

channeldb/migrations.go
halseth: You're right. I read the documentation as only valid within the ForEach loop, but it is indeed the database tx itself. Fixed: https://github.com/lightningnetwork/lnd/pull/3198/commits/5a4057978d6e34d9d2bdad18a8358db95854e102

channeldb/channel.go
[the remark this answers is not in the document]
halseth: That's indeed better. Not sure why this method was here in the first place: https://github.com/lightningnetwork/lnd/pull/3198/commits/526e059c2e9e8774be3f8dfab6648041e1ba5f5c

(or am I missing something?)

channeldb/graph.go
halseth: It's not really moved, it just looks that way because the `delChannelEdge` calls between are moved down.

halseth:
PTAL @cfromknecht @wpaulino

channeldb/channel.go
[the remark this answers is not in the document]
cfromknecht: There are a few places in the codebase where we do this unnecessarily :)

channeldb/channel.go
```diff
@@ -2087,10 +2087,6 @@ func (c *OpenChannel) CloseChannel(summary *ChannelCloseSummary) error {
 		// information stored within the revocation log.
 		logBucket := chanBucket.Bucket(revocationLogBucket)
 		if logBucket != nil {
-			err := wipeChannelLogEntries(logBucket)
-			if err != nil {
-				return err
-			}
```
cfromknecht: Can we delete the `revocationLogBucket` without the above conditional?

cfromknecht: Ah I see, this makes it unnecessary to check the bucket not found case

PR #3211, opened 2019-06-17
channeldb/migrations: skip migration for empty outgoing payments bucket

halseth:
Previously the migration would fail if the source node was not set in
the database. Since we know that the source node must have been set
before making any payments, we check whether there actually are any
payments to migrate, and return early if not.

### DIFF
--- a/channeldb/migrations.go
+++ b/channeldb/migrations.go
@@ -746,9 +746,14 @@ func migrateGossipMessageStoreKeys(tx *bbolt.Tx) error {
 // payments) we delete those statuses, so only Completed payments remain in the
 // new bucket structure.
 func migrateOutgoingPayments(tx *bbolt.Tx) error {
-	oldPayments, err := tx.CreateBucketIfNotExists(paymentBucket)
-	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


channeldb/migrations.go
```diff
@@ -746,9 +746,14 @@ func migrateGossipMessageStoreKeys(tx *bbolt.Tx) error {
...
 func migrateOutgoingPayments(tx *bbolt.Tx) error {
-	oldPayments, err := tx.CreateBucketIfNotExists(paymentBucket)
-	if err != nil {
-		return err
+	log.Infof("Migrating outgoing payments to new bucket structure")
+
+	oldPayments := tx.Bucket(paymentBucket)
+
+	// Return early if there are no payments to migrate.
+	if oldPayments == nil || oldPayments.Stats().KeyN == 0 {
```
cfromknecht: Historically we've avoided using the `Stats` since it can have undefined behavior in some cases

halseth: Probably best to avoid then. Pushed fixup with a diff approach.

cfromknecht:
@halseth latest fix looks good to me, is it possible to have a unit test for this?

PR #3212, opened 2019-06-17
rpcserver: hide channels in getnodeinfo by default

### LINKED ISSUE #3204: New output format for `getnodeinfo` is too long!
### Background

Describe your issue here.
LND master has a new output format for `getnodeinfo` that shows info of a node which often slides off the screen and you still don't know the key info of the node. Often times we only look up a node to find their IP or alias. It's fine to add extra info but can we make a better design for this? 

What I would like to suggest is we just keep the existing format, and add a flag for `--more` if users want to see more info of that node, then LND can just show the channels the node has. 

So for example, this is what I would like for us to keep:
```
$ tlncli getnodeinfo 0270685ca81a8e4d4d01beec5781f4cc924684072ae52c507f8ebe9daf0caaab7b
{
    "node": {
        "last_update": 1560506625,
        "pub_key": "0270685ca81a8e4d4d01beec5781f4cc924684072ae52c507f8ebe9daf0caaab7b",
        "alias": "0270685ca81a8e4d4d01",
        "addresses": [
            {
                "network": "tcp",
                "addr": "159.203.125.125:9735"
            }
        ],
        "color": "#3399ff"
    },
    "num_channels": 175,
    "total_capacity": "1141251451"
}
```
And if someone wants to know more info of a node:
```
$ tlncli getnodeinfo --more 0270685ca81a8e4d4d01beec5781f4cc924684072ae52c507f8ebe9daf0caaab7b
```
then they can have all the extra info like what's done in lnd master currently. 

Maybe someone can code this up for us? Please? I would appreciate it. Otherwise I think it would be very irritating and frustrating with `getnodeinfo` in the next release...  😿

cfromknecht:
Fixes #3204


cmd/lncli/commands.go
```diff
@@ -2925,6 +2925,11 @@ var getNodeInfoCommand = cli.Command{
 			Usage: "the 33-byte hex-encoded compressed public of the target " +
 				"node",
 		},
+		cli.BoolFlag{
+			Name: "include_channels",
+			Usage: "if true, will return all known channels " +
+				"associated with the node.",
```
wpaulino: Nit: we usually don't include periods for these.

PR #3221, opened 2019-06-19
routing+routerrpc+lnrpc: add option to use mc in queryroutes

joostjager:
A boolean flag is added to the QueryRoutes rpc that allows feeding mission control probabilities into path finding.

This pr builds on #3298


lnrpc/rpc.proto
```diff
@@ -1634,6 +1634,12 @@ message QueryRoutesRequest {
     self is assumed.
     */
     string source_pub_key = 8;
+
+    /**
+    If set to true, edge probabilities from mission control will be used to get
+    to the optimal route.
```
C-Otto: to?

joostjager:
>I just gave it a try. The first route had a temporary channel failure. Immediately afterwards the same route was returned by QueryRoutes, so I didn't see any change by mission control. Please tell me how to debug this, I'd be happy to help.

What I think happened is: It did work, but the reason the same route was returned was because it was still the route with the highest probability (although very low).

In `SendPayment` we apply a minimum probability. If the best route is below that, we give up. 

Modified this PR to also apply that minimum probability to `QueryRoutes` with `UseMissionControl` is `true`.

C-Otto:
Same issue. With the code from this PR I get a route, resulting in a temporary channel failure. Running QueryRoutes again returns the same route.

With my code, ignoring based on the channel failure, I get a different route.

joostjager:
I failed to reproduce in the following set up on regtest: A -> B -> C.  The B has 100k sat balance on the B->C channel. 
* `queryroutes` for 200k sat from A -> C. 
* `sendtoroute` using that route -> fails because `TempChannelFailure`
* `queryroutes` again for 200k from A -> C: `unable to find path to destination`.

This is the behavior that I'd expect.

Do you have more detailed steps? One thing about `TempChannelFailure` is that it only takes out the channel only for amounts that are at least the failing amount. So if it fails for 200k sat, it is only blacklisted for values >=200 k sat.

joostjager:
Note to self: `use_mc` doesn't work fully until all mission control state is globalized. Currently the second try logic still resides in the payment session.

joostjager:
Rebased onto first part of #3164

cmd/lncli/commands.go
```diff
@@ -2997,6 +2997,10 @@ var queryRoutesCommand = cli.Command{
 			Usage: "(optional) number of blocks the last hop has to reveal " +
 				"the preimage",
 		},
+		cli.BoolFlag{
+			Name:  "use_mc",
+			Usage: "use mission control probabilities",
```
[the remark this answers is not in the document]
cfromknecht: Are we okay modifying the default behavior in a point release? Should we wait until 0.8 to change the default?

routing/router.go
```diff
@@ -181,7 +181,7 @@ type missionControlInterface interface {
 
 	reportVertexFailure(v route.Vertex)
 
-	getEdgeProbability(fromNode route.Vertex, edge EdgeLocator,
+	GetEdgeProbability(fromNode route.Vertex, edge EdgeLocator,
```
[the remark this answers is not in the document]
joostjager: Full interface exported

cmd/lncli/commands.go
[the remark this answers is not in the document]
joostjager: Agree with @cfromknecht on not changing in point release.

Roasbeef:
Needs a rebase.

joostjager:
rebased

lnrpc/routerrpc/router_backend.go
```diff
@@ -55,6 +55,22 @@ type RouterBackend struct {
 	Tower routing.ControlTower
 }
 
+// MissionControl defines the mission control dependencies of routerrpc.
+type MissionControl interface {
```
halseth: couldn't we define this interface in `routing`?

joostjager: We could, but these are `routerrpc`'s specific requirements for a mission control.

halseth:
Yeah, makes sense. Could move it later if we need to mock it in the router
package.

lnrpc/routerrpc/router_backend.go
cfromknecht: yeah i think it makes sense to do it in `routerrpc`, so it can minimally define what it requires to operate

halseth:
Needs rebase

joostjager:
rebased

PR #3224, opened 2019-06-20
watchtower: use separate key family for wtserver public key, add watchtower subserver

cfromknecht:
This PR modifies the watchtower's public key to be derived from a separate key family from the node's id. This prevents a private tower from being discoverable simply through knowledge of the node's public used for the base LN protocol. For now, this will offer the ability to have a soft whitelist before more advanced whitelisting features are added.

In order to expose this separate public key to users, a new `watchtowerrpc` subserver is added exposing a `GetInfo` rpc. This rpc can be hit programmatically by interfacing directly via gRPC, or can be queried using `lncli tower info`, producing an output such as:
```
$ lncli tower info
{
        "pubkey": "03281d603b2c5e19b8893a484eb938d7377179a9ef1a6bca4c0bcbbfc291657b63",
        "listeners": [
                "[::]:9911"
        ],
        "uris": [
                "03281d603b2c5e19b8893a484eb938d7377179a9ef1a6bca4c0bcbbfc291657b63@X.X.X.X:9911"
        ]
}
```


lncfg/address.go
```diff
@@ -243,6 +243,10 @@ func ParseLNAddressString(strAddress string, defaultPort string,
 // a port, then we'll assume that the user is using the short cut to specify a
 // localhost:port address.
 func verifyPort(address string, defaultPort string) string {
+	if address == "" {
```
Roasbeef: Reminder to revert as this may get in the way of listening on both v4 and v6 for dual stack hosts.

lnrpc/watchtowerrpc/interface.go
```diff
@@ -0,0 +1,23 @@
...
+	// authentication and encrypt traffic with clients.
+	PubKey() *btcec.PublicKey
+
+	// ListeningAddrs returns the listening addresses where the watchtower
+	// server can accept client connections.
+	ListeningAddrs() []net.Addr
+
+	// ExternalIPs returns the addresses where the watchtower can be reached
+	// by clients externally.
+	ExternalIPs() []net.Addr
```
wpaulino: Nit: should be `ExternalAddrs` instead?

lntest/node.go
```diff
@@ -717,6 +720,7 @@ func (hn *HarnessNode) stop() error {
 	hn.processExit = nil
 	hn.LightningClient = nil
 	hn.WalletUnlockerClient = nil
+	hn.WatchtowerClient = nil
```
wpaulino: Didn't realize this was a thing. Should we also do this for `RouterClient` and `WalletKitClient`?

lntest/itest/lnd_test.go
```diff
@@ -7635,28 +7637,57 @@ func testRevokedCloseRetributionAltruistWatchtower(net *lntest.NetworkHarness,
...
+			len(willyInfo.Uris))
+	}
+	if !strings.Contains(willyInfo.Uris[0], externalIP) {
+		t.Fatalf("expected uri with %v, got %v",
+			externalIP, willyInfo.Uris[0])
+	}
+
+	// Construct a URI from listening port and public key, since aren't
+	// actually connecting remotely.
+	willyTowerURI := fmt.Sprintf("%x@%s", willyInfo.Pubkey, listener)
```
wpaulino: Nit: can't wer just use `willyInfo.Uris[0]` instead of this?

cmd/lncli/watchtower_active.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/lightningnetwork/lnd/lnrpc/watchtowerrpc"
+	"github.com/urfave/cli"
+)
+
+func watchtowerCommands() []cli.Command {
+	return []cli.Command{
+		{
+			Name:     "tower",
+			Usage:    "Interact with the watchtower.",
```
wpaulino: Nit: note that we're interacting with the watchtower server?

lntest/itest/lnd_test.go
cfromknecht: listening addrs aren't included in the URIs, those are meant to be what you would give to clients to connect externally. In `lncli getinfo`, we don't show these either

PR #3227, opened 2019-06-20
lnd: reject custom fee when force closing channel

### LINKED ISSUE #3074: Prevent force closing channels with fee rate/conf target
### Background

When i use lnd API for force-closing of channel with `target_conf: 36` i see that LND makes transaction with big fee (as target_conf would be 1-2 only).

### Your environment

* version of `lnd` **0.6.1-beta commit=v0.6.1-beta-rc2**
* which operating system (`uname -a` on *Nix) CentOS
* version of `btcd`, `bitcoind`, or other backend - 0.17.1

### Actual behaviour

For example i do (node.js):

    CloseChannel({ channel_point: { funding_txid_str: 'cfcf31ce231918fb69b36c64345b424cd3aa9b2ef38e32160ffa7ed1b58a1082', output_index: 1 }, force: true, target_conf: 36 })
    
In that time at target bitcoin node i have a fee for 36 blocks as:
    
    $ bitcoin-cli estimatesmartfee 36
    {
      "feerate": 0.00011096,
      "blocks": 36
    }

So it's about 11 sat/byte

But after the LND does this transaction:

https://www.smartbit.com.au/tx/ee9cac6a8cfed7a37f8929794fe3b21daf781d06b7541b093a399d9a4a66e549

There is 133 sat/byte. I’m sure that in this case the LND does not explicitly follow the `target_conf`

And i see many same cases now (i did some force closings and checked them)

2xic:
This fixes #3074

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1674,6 +1674,12 @@ func (r *rpcServer) CloseChannel(in *lnrpc.CloseChannelRequest,
 		return fmt.Errorf("must specify channel point in close channel")
 	}
 
+	// If force closing a channel, the fee set in the commitment transaction
+	// is used.
+	if in.Force && (in.SatPerByte != 0 || in.TargetConf != 0) {
+		return fmt.Errorf("force closing a channel uses a pre-defined fee")
+	}
+
 	force := in.Force
 	index := in.ChannelPoint.OutputIndex
 	txid, err := GetChanPointFundingTxid(in.GetChannelPoint())


cfromknecht:
one small thing, you'll need to run `go fmt -w -s rpcserver.go`, then push again
```
rpcserver.go:1::warning: file is not gofmted with -s (gofmt)
```

2xic:
Fixed 👍

PR #3233, opened 2019-06-21
[trivial] rpcserver+router: log terminal sendPayment error

halseth:
Increase visibility into why payments fail, especially in the app logs.

### DIFF
--- a/routing/payment_lifecycle.go
+++ b/routing/payment_lifecycle.go
@@ -218,6 +218,9 @@ func (p *paymentLifecycle) createNewPaymentAttempt() (lnwire.ShortChannelID,
 		p.payment, uint32(p.currentHeight), p.finalCLTVDelta,
 	)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


routing/payment_lifecycle.go
```diff
@@ -218,6 +218,9 @@ func (p *paymentLifecycle) createNewPaymentAttempt() (lnwire.ShortChannelID,
 		p.payment, uint32(p.currentHeight), p.finalCLTVDelta,
 	)
 	if err != nil {
+		log.Errorf("Failed to find route for payment %x: %v",
+			p.payment.PaymentHash, err)
```
joostjager: I think we should reserve `Error` for internal errors. This would be just `Info` I guess, not even warning?

halseth: `warn` is a nice middle ground maybe?

Roasbeef:
Committed as `8b83f03308b45a3b118b0157ddc71c873209c811` using `Warnf`.

PR #3245, opened 2019-06-25
rpc: fix calculating RPC permissions.

roeierez:
This PR fixes a bug that prevents restart LND, after it has stopped, at the same process.
The problem resides in the fact that an array of permissions is initialized as a package variable and when creating the RPCServer, all subserver permissions are appended to this variable, while checking for duplicates.
On subsequent restart this array is left with the data from the previous run and being populated again with the same permissions causing a duplicate error.
The solution I propose is simple, just to extract out the initial permissions to
a separate function and call it from the context is is needed.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -168,9 +168,12 @@ var (
 			Action: "write",
 		},
 	}
+)
 
-	// permissions maps RPC calls to the permissions they require.
-	permissions = map[string][]bakery.Op{
+// mainRPCServerPermissions returns a mapping of the main RPC server calls to
+// the permissions they require.
+func mainRPCServerPermissions() map[string][]bakery.Op {
+	return map[string][]bakery.Op{
 		"/lnrpc.Lightning/SendCoins": {{
 			Entity: "onchain",
 			Action: "write",
@@ -381,7 +384,7 @@ var (
 			Action: "read",
 		}},
 	}
-)
+}
 
 // rpcServer is a gRPC, RPC front end to the lnd daemon.
 // TODO(roasbeef): pagination support for the list-style calls
@@ -520,6 +523,7 @@ func newRPCServer(s *server, macService *macaroons.Service,
 	// Next, we need to merge the set of sub server macaroon permissions
 	// with the main RPC server permissions so we can unite them under a
 	// single set of interceptors.
+	permissions := mainRPCServerPermissions()
 	for _, subServerPerm := range subServerPerms {
 		for method, ops := range subServerPerm {
 			// For each new method:ops combo, we also ensure that


rpcserver.go
```diff
@@ -517,6 +519,7 @@ func newRPCServer(s *server, macService *macaroons.Service,
 		subServerPerms = append(subServerPerms, macPerms)
 	}
 
+	permissions := initialPermissions()
```
halseth: nit: move to below the comment

rpcserver.go
```diff
@@ -168,9 +168,60 @@ var (
 			Action: "write",
 		},
 	}
+)
+
+// rpcServer is a gRPC, RPC front end to the lnd daemon.
+// TODO(roasbeef): pagination support for the list-style calls
+type rpcServer struct {
```
halseth: Is moving this struct necessary? Diff would be much smaller if not.

rpcserver.go
```diff
@@ -168,9 +168,60 @@ var (
...
+}
+
+// A compile time check to ensure that rpcServer fully implements the
+// LightningServer gRPC service.
+var _ lnrpc.LightningServer = (*rpcServer)(nil)
 
-	// permissions maps RPC calls to the permissions they require.
-	permissions = map[string][]bakery.Op{
+// permissions maps RPC calls to the permissions they require.
+func initialPermissions() map[string][]bakery.Op {
```
halseth: comment should be slightly updated to describe what this returns.

halseth: should be named `mainRpcServerPermissions` or similar?

roeierez: updated.

roeierez: changed to `mainRPCServerPermissions`.

rpcserver.go
roeierez: moved.

rpcserver.go
roeierez: @halseth  I haven't moved the struct, it is the diff algo probably. Struggled with it and couldn't force it to be smaller than this.

cfromknecht:
@roeierez `rpcserver.go` needs a go fmt to pass the linter

roeierez:
> @roeierez `rpcserver.go` needs a go fmt to pass the linter

My bad. formatted.

rpcserver.go
wpaulino: It seems like we can avoid the bigger diff by moving `mainRPCServerPermissions` above `type rpcServer struct`.

halseth: Yeah, keeping the struct below `mainRPCServerPermissions` should fix it?

roeierez: @halseth indeed. fixed.

Issue #3240, opened 2019-06-24
ForwardingHistory: intelligently handle vague requests


tyzbit:
### Background

If no parameters are given, the `ForwardingHistory` RPC queries the `ChannelDB` for the last 24 hours of forwarding events.  This behaviour may not be expected considering the name of the RPC; one might expect that a full history, or the most recent chunk of forwarding events be returned rather than a time slice of 24 hours.

I suggest the default behavior of `ForwardingHistory` be changed to look for the 100 most recent forwarding events.  Also, if a user provides only one of StartTime or EndTime, `ForwardingHistory` should assume a default StartTime of the seed birthday, or a default EndTime of `time.Now()`.  This might also help in cases such as #2585.

Alternatively (and more simply), we can default end_time to time.Now() so users can omit it in favor of just providing a start_time.

This is separate from #1233.

### Steps to reproduce

1. Record a specific forwarding event
2. Wait 24 hours
3. Run `$ lncli fwdinghistory`

### Expected behaviour

All forwarding events are shown, or the most recent X forwarding events are shown.

### Actual behaviour

The specific forwarding event is not shown.  It must be found by specifying StartTime and EndTime together.

tyzbit:
If the idea sticks, I'd like to take this on.

Roasbeef:
Fixed by #3246.

PR #3246, opened 2019-06-26
rpcserver: ForwardingHistory end_time defaults to time.Now()

tyzbit:
Closes simple version of #3240.

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -4600,14 +4600,19 @@ func (r *rpcServer) ForwardingHistory(ctx context.Context,
 		numEvents uint32
 	)
 
-	// If the start and end time were not set, then we'll just return the
-	// records over the past 24 hours.
-	if req.StartTime == 0 && req.EndTime == 0 {
+	// If the start time wasn't specified, we'll default to 24 hours ago.
+	if req.StartTime == 0 {
 		now := time.Now()
 		startTime = now.Add(-time.Hour * 24)
-		endTime = now
 	} else {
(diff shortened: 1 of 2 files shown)


rpcserver.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/btcsuite/btcwallet/wallet/txauthor"
 	"github.com/coreos/bbolt"
 	"github.com/davecgh/go-spew/spew"
-	"github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
halseth: unnecessary change?

rpcserver.go
```diff
@@ -4602,12 +4602,18 @@ func (r *rpcServer) ForwardingHistory(ctx context.Context,
 
 	// If the start and end time were not set, then we'll just return the
 	// records over the past 24 hours.
-	if req.StartTime == 0 && req.EndTime == 0 {
+	if req.StartTime == 0 {
```
halseth: comment should be updated

rpcserver.go
tyzbit: For some reason my editor (Atom) wanted that there.  I removed it.

rpcserver.go
tyzbit: Updated comment.

PR #3253, opened 2019-06-28
multi: add secondary DNS seed for bootstrap, add exponential back off to initial bootstrap

Roasbeef:
In this commit we add a second DNS seed to make lnd's process of bootstrapping an initial set of peers more robust. Along the way, we modify the primary DNS `SampleNodeAddrs` method to no longer retry, and also simply try the next seed if it fails to connect to one. Additionally, it is now no longer required that a DNS seed uses our "secondary fall back" mechanism. 

Reviewers can test that this PR will fall back to the newly added seed if the primary is down by modifying the domain in `chainregistry.go`, and ensuring that a new `lnd` instance is still able to find peers.

### DIFF
--- a/discovery/bootstrapper.go
+++ b/discovery/bootstrapper.go
@@ -379,135 +379,146 @@ func (d *DNSSeedBootstrapper) SampleNodeAddrs(numAddrs uint32,
 
 	var netAddrs []*lnwire.NetAddress
 
-	// We'll continue this loop until we reach our target address limit.
-	// Each SRV query to the seed will return 25 random nodes, so we can
-	// continue to query until we reach our target.
+	// We'll try all the registered DNS seeds, exiting early if one of them
+	// gives us all the peers we need.
+	//
+	// TODO(roasbeef): should combine results from both
 search:
-	for uint32(len(netAddrs)) < numAddrs {
-		for _, dnsSeedTuple := range d.dnsSeeds {
-			// We'll first query the seed with an SRV record so we
-			// can obtain a random sample of the encoded public
-			// keys of nodes. We use the lndLookupSRV function for
-			// this task.
-			primarySeed := dnsSeedTuple[0]
-			_, addrs, err := d.net.LookupSRV("nodes", "tcp", primarySeed)
-			if err != nil {
-				log.Tracef("Unable to lookup SRV records via "+
-					"primary seed: %v", err)
+	for _, dnsSeedTuple := range d.dnsSeeds {
+		// We'll first query the seed with an SRV record so we can
+		// obtain a random sample of the encoded public keys of nodes.
+		// We use the lndLookupSRV function for this task.
+		primarySeed := dnsSeedTuple[0]
+		_, addrs, err := d.net.LookupSRV("nodes", "tcp", primarySeed)
+		if err != nil {
+			log.Tracef("Unable to lookup SRV records via "+
+				"primary seed (%v): %v", primarySeed, err)
 
-				log.Trace("Falling back to secondary")
+			log.Trace("Falling back to secondary")
 
-				// If the host of the secondary seed is blank,
-				// then we'll bail here as we can't proceed.
-				if dnsSeedTuple[1] == "" {
-					return nil, fmt.Errorf("Secondary seed is blank")
-				}
-
-				// If we get an error when trying to query via
-				// the primary seed, we'll fallback to the
-				// secondary seed before concluding failure.
-				soaShim := dnsSeedTuple[1]
-				addrs, err = d.fallBackSRVLookup(
-					soaShim, primarySeed,
-				)
-				if err != nil {
-					return nil, err
-				}
-				log.Tracef("Successfully queried fallback DNS seed")
+			// If the host of the secondary seed is blank, then
+			// we'll bail here as we can't proceed.
+			if dnsSeedTuple[1] == "" {
+				log.Tracef("DNS seed %v has no secondary, "+
+					"skipping fallback", primarySeed)
+				continue
 			}
 
-			log.Tracef("Retrieved SRV records from dns seed: %v",
-				spew.Sdump(addrs))
+			// If we get an error when trying to query via the
+			// primary seed, we'll fallback to the secondary seed
+			// before concluding failure.
+			soaShim := dnsSeedTuple[1]
+			addrs, err = d.fallBackSRVLookup(
+				soaShim, primarySeed,
+			)
+			if err != nil {
+				log.Tracef("Unable to query fall "+
+					"back dns seed (%v): %v", soaShim, err)
+				continue
+			}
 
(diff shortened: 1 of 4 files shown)


discovery/bootstrapper.go
```diff
@@ -379,135 +379,135 @@ func (d *DNSSeedBootstrapper) SampleNodeAddrs(numAddrs uint32,
...
 
-				// If the host of the secondary seed is blank,
-				// then we'll bail here as we can't proceed.
-				if dnsSeedTuple[1] == "" {
-					return nil, fmt.Errorf("Secondary seed is blank")
-				}
+			// If the host of the secondary seed is blank, then
+			// we'll bail here as we can't proceed.
+			if dnsSeedTuple[1] == "" {
+				log.Errorf("dns seed %v has no secondary, "+
```
wpaulino: Nit: s/dns/DNS/

wpaulino: Also don't see why this needs to be logged as an error? Maybe just keep trace as the one above?

discovery/bootstrapper.go
```diff
@@ -379,135 +379,135 @@ func (d *DNSSeedBootstrapper) SampleNodeAddrs(numAddrs uint32,
...
-				log.Tracef("Successfully queried fallback DNS seed")
+			// If we get an error when trying to query via the
+			// primary seed, we'll fallback to the secondary seed
+			// before concluding failure.
+			soaShim := dnsSeedTuple[1]
+			addrs, err = d.fallBackSRVLookup(
+				soaShim, primarySeed,
+			)
+			if err != nil {
+				log.Errorf("unable to query fall "+
```
wpaulino: Nit: logs should start capitalized. Can also be trace?

discovery/bootstrapper.go
```diff
@@ -379,135 +379,135 @@ func (d *DNSSeedBootstrapper) SampleNodeAddrs(numAddrs uint32,
...
+					"back dns ssed (%v): %v", soaShim, err)
```
wpaulino: Nit: s/ssed/seed/

server.go
```diff
@@ -1723,14 +1723,35 @@ const bootstrapBackOffCeiling = time.Minute * 5
 func (s *server) initialPeerBootstrap(ignore map[autopilot.NodeID]struct{},
 	numTargetPeers uint32, bootstrappers []discovery.NetworkPeerBootstrapper) {
 
+	// We'll start off by waiting 2 seconds between failed attempts, then
+	// double each time we fail until we hit the bootstrapBackOffCeiling.
+	delayTime := time.Second * 2
+	delayTicker := time.NewTicker(delayTime)
+	defer delayTicker.Stop()
```
wpaulino: Since we're re-initializing this ticker below, are we sure that this `Stop` won't apply to the first initialization?

server.go
```diff
@@ -1741,8 +1762,20 @@ func (s *server) initialPeerBootstrap(ignore map[autopilot.NodeID]struct{},
...
+		// We failed to reach our target, so we'll back off a bit to
+		// avoid hammering any of the bootstrap sources.
+		if attempts > 0 {
+			delayTime *= 2
+			if delayTime > backOffCeiling {
+				delayTime = backOffCeiling
+			}
+
+			delayTicker.Stop()
+			delayTicker = time.NewTicker(delayTime)
```
cfromknecht: might be simpler to keep track of the delay time and then use a `time.After(delayTime)` above, otherwise this ticker may never be stopped

server.go
```diff
@@ -1752,6 +1751,7 @@ func (s *server) initialPeerBootstrap(ignore map[autopilot.NodeID]struct{},
 
 		// Then, we'll attempt to establish a connection to the
 		// different peer addresses retrieved by our bootstrappers.
+		var wg sync.WaitGroup
```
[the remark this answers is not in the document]
cfromknecht: shouldn't have any affect afaict, but better to avoid reuse

discovery/bootstrapper.go
```diff
@@ -379,135 +379,135 @@ func (d *DNSSeedBootstrapper) SampleNodeAddrs(numAddrs uint32,
...
 
-			// Next, we'll need to issue an A record request for
-			// each of the nodes, skipping it if nothing comes
-			// back.
-			for _, nodeSrv := range addrs {
-				if uint32(len(netAddrs)) >= numAddrs {
-					break search
-				}
+		log.Tracef("Retrieved SRV records from dns seed: %v",
+			spew.Sdump(addrs))
```
cfromknecht: naked spew

discovery/bootstrapper.go
```diff
@@ -379,135 +379,135 @@ func (d *DNSSeedBootstrapper) SampleNodeAddrs(numAddrs uint32,
...
-				if ignore != nil {
-					nID := autopilot.NewNodeID(nodeKey)
-					if _, ok := ignore[nID]; ok {
-						continue
-					}
-				}
+			// If we have a set of valid addresses, then we'll need
+			// to parse the public key from the original bech32
+			// encoded string.
+			bechNode := strings.Split(bechNodeHost, ".")
```
cfromknecht: assert length is greater than 0? o/w a bug in the host could take down the daemon

watchtower/wtclient/session_queue.go
```diff
@@ -557,7 +557,7 @@ func (q *sessionQueue) sendStateUpdate(conn wtserver.Peer,
 		// TODO(conner): borked watchtower
 		err = fmt.Errorf("unable to ack seqnum=%d: %v",
 			stateUpdate.SeqNum, err)
-		log.Errorf("SessionQueue(%s) failed to ack update: %v", err)
+		log.Errorf("SessionQueue(%v) failed to ack update: %v", q.ID(), err)
```
cfromknecht: 👍

server.go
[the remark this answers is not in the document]
Roasbeef: If this loop didn't succeed after the first time, then IIRC we may panic as you can't call `Add` on a wait group after you call `Done`

server.go
Roasbeef: You mean it'll still be pointing to the original ticker? We use this pattern a few places within the codebase, FWIW.

Roasbeef: Switched to manually manage a `time.After` instance.

discovery/bootstrapper.go
Roasbeef: Was already present, but fixed.

discovery/bootstrapper.go
Roasbeef: Done, also added a check to ensure it has a `.` within it as well. Aside from this, we should do some fuzzing in general on our `bech32` implementation.

Roasbeef:
PTAL @wpaulino @cfromknecht

PR #3254, opened 2019-06-28
chainntnfs: stricter historical confirmation matching

cfromknecht:
This PR refactors the dispatch logic for historical confirmation notifications on backends with txindex enabled. Currently these dispatches are done on txid which should be sufficient, but we modify this to also assert that the transaction contains the correct scriptpubkey in the output. This better aligns the dispatch logic with all other dispatch paths, which already dispatch on scripts in order to support neutrino. Under some nuanced circumstances, this prevents false positive dispatches when interacting with `loopd`.

### DIFF
--- a/chainntnfs/interface.go
+++ b/chainntnfs/interface.go
@@ -1,8 +1,11 @@
 package chainntnfs
 
 import (
+	"bytes"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/btcsuite/btcd/btcjson"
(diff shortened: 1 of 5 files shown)


chainntnfs/txnotifier.go
```diff
@@ -192,18 +192,22 @@ func (r ConfRequest) ConfHintKey() ([]byte, error) {
...
+		return false
 	}
 
-	pkScript := r.PkScript.Script()
-	for _, txOut := range tx.TxOut {
-		if bytes.Equal(txOut.PkScript, pkScript) {
-			return true
-		}
+	if r.TxID != ZeroHash {
+		return r.TxID == tx.TxHash() && scriptMatches()
```
Roasbeef: +1 for a stricter check

Will also need to do a sweep to ensure that we're properly setting the script everywhere. I _think_ we did this properly once we switched to the current script based dispatch, but won't hurt to double check.

cfromknecht: double checked all the call sites and appears we are

chainntnfs/interface.go
```diff
@@ -583,3 +586,103 @@ func getMissedBlocks(chainConn ChainConn, startingHeight,
 
 	return missedBlocks, nil
 }
+
+// TxIndexConn abstracts an RPC backend with txindex enabled.
+type TxIndexConn interface {
```
Roasbeef: 👍

PR #3256, opened 2019-06-28
routing: failure attribution

joostjager:
This PR overhauls the interpretation of failed payments. It changes the interpretation rules so that we always apply the strongest possible set of penalties, without making assumptions that would hurt good nodes.

Main changes are:

* Apply different rule sets for intermediate and final nodes. Both types of nodes have different sets of failures that we expect. Penalize nodes that send unexpected failure messages.

* Distinguish between direct payments and multi-hop payments. For direct payments, we can infer more about the performance of our peer because we trust ourselves.

* In many cases it is impossible for the sender to determine which of the two nodes in a pair is responsible for the failure. In this situation, we now penalize bidirectionally. This does not hurt the good node of the pair, because only its connection to a bad node is penalized.

* Previously we always penalized the outgoing connection of the reporting node. This is incorrect for policy related failures. For policy related failures, it could also be that the reporting node received a wrongly crafted htlc from its predecessor. By penalizing the incoming channel, we surely hit the responsible node.

* FailExpiryTooSoon is a failure that could have been caused by any node up to the reporting node by delaying forwarding of the htlc. We don't know which node is responsible, therefore we now penalize all node pairs in the route. By doing this, we do hurt good nodes. But unfortunately there is currently no way to pinpoint just a single pair to penalize.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -1637,8 +1637,21 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 				OnionSHA256: msg.ShaOnionBlob,
 			}
 		default:
-			log.Errorf("Unknown failure code: %v", msg.FailureCode)
-			failure = &lnwire.FailTemporaryChannelFailure{}
+			log.Warnf("Unexpected failure code received in "+
+				"UpdateFailMailformedHTLC: %v", msg.FailureCode)
+
+			// We don't just pass back the error we received from
+			// our successor. Otherwise we might report a failure
+			// that penalizes us more than needed. If the onion that
+			// we forwarded was correct, the node should have been
+			// able to send back its own failure. The node did not
+			// send back its own failure, so we assume there was a
+			// problem with the onion and report that back. We reuse
+			// the invalid onion key failure because there is no
+			// specific error for this case.
+			failure = &lnwire.FailInvalidOnionKey{
+				OnionSHA256: msg.ShaOnionBlob,
+			}
 		}
 
 		// With the error parsed, we'll convert the into it's opaque
--- a/routing/result_interpretation.go
+++ b/routing/result_interpretation.go
@@ -7,6 +7,11 @@ import (
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
+// Instantiate variables to allow taking a reference from the failure reason.
+var (
+	reasonError = channeldb.FailureReasonError
+)
+
 // interpretedResult contains the result of the interpretation of a payment
 // outcome.
 type interpretedResult struct {
@@ -52,21 +57,14 @@ func (i *interpretedResult) processFail(
 	rt *route.Route, errSourceIdx *int,
 	failure lnwire.FailureMessage) (bool, channeldb.FailureReason) {
 
-	var failureSourceIdxInt int
-
 	if errSourceIdx == nil {
-		// If the failure message could not be decrypted, attribute the
-		// failure to our own outgoing channel.
-		//
-		// TODO(joostager): Penalize all channels in the route.
-		failureSourceIdxInt = 0
-		failure = lnwire.NewTemporaryChannelFailure(nil)
-	} else {
-		failureSourceIdxInt = *errSourceIdx
+		i.processPaymentOutcomeUnknown(rt)
+		return false, 0
 	}
 
 	var failureVertex route.Vertex
 
+	failureSourceIdxInt := *errSourceIdx
 	if failureSourceIdxInt > 0 {
 		failureVertex = rt.Hops[failureSourceIdxInt-1].PubKeyBytes
 	} else {
@@ -239,6 +237,74 @@ func (i *interpretedResult) processFail(
 		i.nodeFailure = &failureVertex
 		return false, 0
 	}
+}
+
+// processPaymentOutcomeUnknown processes a payment outcome for which no failure
+// message or source is available.
+func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
+	n := len(route.Hops)
+
+	// If this is a direct payment, the destination must be at fault.
+	if n == 1 {
+		i.failNode(route, n)
+		i.finalFailureReason = &reasonError
+		return
+	}
+
+	// Otherwise penalize all channels in the route to make sure the
+	// responsible node is at least hit too. We even penalize the connection
+	// to our own peer, because that peer could also be responsible.
+	i.failPairRange(route, 0, n-1)
+}
+
+// failNode marks the node indicated by idx in the route as failed. This
+// function intentionally panics when the self node is failed.
+func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	i.nodeFailure = &rt.Hops[idx-1].PubKeyBytes
+}
+
+// failPairRange marks the node pairs from node fromIdx to node toIdx as failed.
+func (i *interpretedResult) failPairRange(
+	rt *route.Route, fromIdx, toIdx int) {
+
+	for idx := fromIdx; idx <= toIdx; idx++ {
+		i.failPair(rt, idx)
+	}
+}
+
+// failPair marks a pair as failed in both directions.
+func (i *interpretedResult) failPair(
+	rt *route.Route, idx int) {
+
+	pair, _ := getPair(rt, idx)
+
+	// Report pair in both directions without a minimum penalization amount.
+	i.pairResults[pair] = 0
+	i.pairResults[pair.Reverse()] = 0
+}
+
+// getPair returns a node pair from the route and the amount passed between that
+// pair.
+func getPair(rt *route.Route, channelIdx int) (DirectedNodePair,
+	lnwire.MilliSatoshi) {
+
+	nodeTo := rt.Hops[channelIdx].PubKeyBytes
+	var (
+		nodeFrom route.Vertex
+		amt      lnwire.MilliSatoshi
+	)
+
+	if channelIdx == 0 {
+		nodeFrom = rt.SourcePubKey
+		amt = rt.TotalAmount
+	} else {
+		nodeFrom = rt.Hops[channelIdx-1].PubKeyBytes
+		amt = rt.Hops[channelIdx-1].AmtToForward
+	}
+
+	pair := NewDirectedNodePair(nodeFrom, nodeTo)
+
+	return pair, amt
 }
 
 // getFailedPair tries to locate the failing pair given a route and the pubkey
--- a/routing/result_interpretation.go
+++ b/routing/result_interpretation.go
@@ -2,14 +2,14 @@ package routing
 
 import (
 	"github.com/lightningnetwork/lnd/channeldb"
-
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
 // Instantiate variables to allow taking a reference from the failure reason.
 var (
-	reasonError = channeldb.FailureReasonError
+	reasonError            = channeldb.FailureReasonError
+	reasonIncorrectDetails = channeldb.FailureReasonIncorrectPaymentDetails
 )
 
 // interpretedResult contains the result of the interpretation of a payment
@@ -44,10 +44,7 @@ func interpretResult(rt *route.Route, failureSrcIdx *int,
 		pairResults: make(map[DirectedNodePair]lnwire.MilliSatoshi),
 	}
 
-	final, reason := i.processFail(rt, failureSrcIdx, failure)
-	if final {
-		i.finalFailureReason = &reason
-	}
+	i.processFail(rt, failureSrcIdx, failure)
 
 	return i
 }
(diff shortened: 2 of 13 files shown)


joostjager:
note to self: handle private channels

routing/result_interpretation.go
```diff
@@ -31,211 +31,211 @@ func interpretResult(rt *route.Route, failureSrcIdx *int,
...
+	// If we get a permanent channel, we'll prune the channel set in both
+	// directions and continue with the rest of the routes.
 	case *lnwire.FailPermanentChannelFailure:
-		i.pairResults[failedPair] = 0
-		i.pairResults[failedPair.Reverse()] = 0
-		return false, 0
+		reportOutgoing()
+
+	// If we get a failure due to violating the channel policy, we request a
+	// second chance because our graph may be out of date. An attached
```
Roasbeef: Comment looks out of date. Are we doing away with the second chance logic?

routing/result_interpretation.go
```diff
@@ -31,211 +31,211 @@ func interpretResult(rt *route.Route, failureSrcIdx *int,
...
+		reportIncoming()
+
+	// If the outgoing channel doesn't have enough capacity, we penalize.
+	// But we penalize only in a single direction and only for amounts
+	// greater than the attempted amount.
+	case *lnwire.FailTemporaryChannelFailure:
+		reportOutgoingBalance()
 
-	// Any other failure or an empty failure will get the node pruned.
+	// If FailExpiryTooSoon is received, there must have been some delay
```
Roasbeef: Alternatively, this could actually be us being behind the chain (not yet synced).

routing/result_interpretation.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// interpretedResult contains the result of the interpretation of a payment
+// outcome.
+type interpretedResult struct {
+	nodeFailures       map[route.Vertex]struct{}
+	pairResults        map[DirectedNodePair]lnwire.MilliSatoshi
+	finalFailureReason *channeldb.FailureReason
+	policyFailure      *DirectedNodePair
```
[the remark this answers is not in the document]
joostjager: added

routing/result_interpretation.go
```diff
@@ -0,0 +1,275 @@
...
+		i.finalFailureReason = &reason
+	}
+
+	return i
+}
+
+// processFail processes a failed payment attempt.
+func (i *interpretedResult) processFail(
+	rt *route.Route, errSourceIdx *int,
+	failure lnwire.FailureMessage) (bool, channeldb.FailureReason) {
```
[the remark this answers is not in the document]
joostjager: For the same reason it was originally introduced: because we can't take references to the constant returned in this function. But this will be further cleaned up in following commits.

routing/result_interpretation.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Always determine chan id ourselves, because a channel update with id
+	// may not be available.
+	failedPair, failedAmt := getFailedPair(
+		rt, failureSourceIdxInt,
+	)
+
+	switch failure.(type) {
+
+	// If the end destination didn't know the payment
```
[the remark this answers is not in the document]
joostjager: fixed

routing/result_interpretation_test.go
```diff
@@ -0,0 +1,88 @@
...
+	if i.pairResults[NewDirectedNodePair(hops[0], hops[1])] != 100 {
+		t.Fatal("wrong pair result")
+	}
+
+	if i.finalFailureReason != nil {
+		t.Fatal("expected attempt to be non-final")
+	}
+}
+
+func TestResultInterpretationFailExpiryTooSoon(t *testing.T) {
```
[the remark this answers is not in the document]
joostjager: added

routing/result_interpretation.go
```diff
@@ -241,6 +239,75 @@ func (i *interpretedResult) processFail(
...
+func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	i.nodeFailures[rt.Hops[idx-1].PubKeyBytes] = struct{}{}
+}
+
+// failPairRange marks the node pairs from node fromIdx to node toIdx as failed.
+func (i *interpretedResult) failPairRange(
+	rt *route.Route, fromIdx, toIdx int) {
+
+	// Start at one because we don't penalize our own channels.
+	for idx := fromIdx; idx <= toIdx; idx++ {
```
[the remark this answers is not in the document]
joostjager: Comment removed, outdated. Explanation of ranges is given at the call site.

routing/result_interpretation.go
```diff
@@ -31,211 +31,211 @@ func interpretResult(rt *route.Route, failureSrcIdx *int,
...
 
-	var failureVertex route.Vertex
+	switch *errSourceIdx {
 
-	failureSourceIdxInt := *errSourceIdx
-	if failureSourceIdxInt > 0 {
-		failureVertex = rt.Hops[failureSourceIdxInt-1].PubKeyBytes
-	} else {
-		failureVertex = rt.SourcePubKey
+	// We don't keep a reputation for ourselves, as information about
```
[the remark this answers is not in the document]
joostjager: Thanks. I am happy that after so much preparatory refactoring in the routing area, this finally becomes mergeable.

routing/result_interpretation.go
joostjager: Interesting. The question is how to deal with this? Isn't the solution to not try to send payments if we are not yet synced? Because we can't calculate the proper htlc expiries?

routing/result_interpretation.go
joostjager: No, it is still there, but up in mission control. This file is just dealing with the result in isolation, without looking at any history.

Comments clarified.

routing/result_interpretation.go
```diff
@@ -31,211 +31,211 @@ func interpretResult(rt *route.Route, failureSrcIdx *int,
...
+		// includes our own outgoing connection.
+		i.failPairRange(
+			route, 0, errorSourceIdx-1,
+		)
+	}
+
+	switch failure.(type) {
+
+	// If a hop reports onion payload corruption or an invalid version, we
+	// will report the outgoing channel set of that node. It may be either
```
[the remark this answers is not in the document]
joostjager: At this point, it is not possible that the node's predecessor corrupted the onion blob. If the predecessor would have corrupted the payload, the error source wouldn't have been able to encrypt this failure message for us.

Extended comment to make this more clear.

routing/result_interpretation.go
```diff
@@ -31,211 +31,211 @@ func interpretResult(rt *route.Route, failureSrcIdx *int,
...
+
+	reportOutgoingBalance := func() {
+		i.failPairBalance(
+			route, errorSourceIdx,
+		)
+	}
+
+	reportIncoming := func() {
+		// We trust ourselves. If the error comes from the first hop, we
+		// can penalize the whole node. In that case there is no
```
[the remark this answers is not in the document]
joostjager: If the error comes from the first hop, the whole (tried) route is just us and our error sending peer.  We assume we ourselves are bug-free, so there must be something wrong with our peer. By reporting the whole node (instead of just the pair us<->them), we also won't use that node again in different paths.

routing/result_interpretation.go
```diff
@@ -0,0 +1,338 @@
+package routing
+
+import (
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// Instantiate variables to allow taking a reference from the failure reason.
+var (
```
cfromknecht: do these need to be mutable?

routing/result_interpretation.go
```diff
@@ -0,0 +1,338 @@
...
+	i.failPairRange(route, 0, n-1)
+}
+
+// failNode marks the node indicated by idx in the route as failed. This
+// function intentionally panics when the self node is failed.
+func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	i.nodeFailure = &rt.Hops[idx-1].PubKeyBytes
+}
+
+// failPairRange marks the node pairs from node fromIdx to node toIdx as failed.
```
cfromknecht: comment should note that the whole range is failed in both directions?

routing/result_interpretation.go
```diff
@@ -0,0 +1,338 @@
...
+	if n == 1 {
+		i.failNode(route, n)
+		i.finalFailureReason = &reasonError
+		return
+	}
+
+	// Otherwise penalize all channels in the route to make sure the
+	// responsible node is at least hit too. We even penalize the connection
+	// to our own peer, because that peer could also be responsible.
+	i.failPairRange(route, 0, n-1)
```
cfromknecht: if the path were to return to the source node, e.g. `sendtoroute`, this would also blacklist us? how should we handle that case?

routing/result_interpretation.go
joostjager: I intentionally made them mutable, so that I can take the reference of it. They are already defined as `const` in `channeldb`. Or is it possible to have an immutable (non-const) variable? One that can only be initialized and is final thereafter.

routing/result_interpretation.go
joostjager: It wouldn't blacklist us completely, but only our connection with the last hop before the htlc is returned to us. It is not ideal, but I think there is no satisfying solution without having more information on the error source. Just hope that this is not going to become something that happens regularly.

routing/result_interpretation_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	routeTwoHop = route.Route{
+		Hops: []*route.Hop{
+			{PubKeyBytes: hops[0], AmtToForward: 100},
+			{PubKeyBytes: hops[1]},
+		},
+	}
+
+	routeThreeHop = route.Route{
```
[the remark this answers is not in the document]
joostjager: left over from something that is split into follow up prs. removed here

routing/result_interpretation.go
joostjager: added

routing/result_interpretation.go
cfromknecht: i see now, you can't take a reference to a const. makes sense when i think about it, but have never came across it. TIL

> Or is it possible to have an immutable (non-const) variable? One that can only be initialized and is final thereafter.

not that i'm aware

Roasbeef:
Travis is funky due to coveralls being down right now, but needs a rebase!

routing/result_interpretation.go
Roasbeef: Yeah we do this atm if we're still syncing, but if we fall behind for w/e reason (RPC connection dies or something), then it isn't covered yet. Not saying we need to handle this here, just pointing it out for future consideration.

joostjager:
Rebased

routing/result_interpretation.go
joostjager: Comment extended to describe this case and mention that we ignore it for now.

joostjager:
Refactored test into parameterized test.

Roasbeef:
On travis, the integration tests seem to stall out after: 
```
=== RUN   TestLightningNetworkDaemon/sphinx_replay_persistence
```

joostjager:
This was a very useful test failure. Previously I assumed that any failures originating from self would be temporary and caused by the gap between path finding and payment execution. But there is also the case where our peer sends us a malformed htlc failure that is converted into a local error.

I changed the failure that is returned in that case (commit "htlcswitch: always assume an onion error for malformed htlc failures"), see commit message for explanation. And I added this exception for handling self failures during result interpretation.

joostjager:
`private_channels` itest now failing, going to look into that.

htlcswitch/link.go
```diff
@@ -1637,8 +1637,21 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
...
+			// We don't just pass back the error we received from
+			// our successor. Otherwise we might report a failure
+			// that penalizes us more than needed. If the onion that
+			// we forwarded was correct, the node should have been
+			// able to send back its own failure. The node did not
+			// send back its own failure, so we assume there was a
+			// problem with the onion and report that back. We reuse
+			// the invalid onion key failure because there is no
+			// specific error for this case.
+			failure = &lnwire.FailInvalidOnionKey{
```
Roasbeef: IIRC, in the past, we decided to send the temp chan failure as we didn't want to give away any further information detailing that we rejected the packet due to it being a replay. However, I think this is ok as well since it doesn't give away any additional information.

joostjager: One thing that I was wondering is why a replay always leads to failing back the htlc. Shouldn't the receiving node just ignore the duplicate package so that the htlc can live on?

joostjager:
`private_channels` was another interesting failing test that surfaced that we only query bandwidth hints once per payment. In this test, two concurrent payments are executed. Because we stop tracking our own reputation in this pr, not updating the bandwidth hints led to an endless loop. Added a commit to update bandwidth hints with every attempt.

htlcswitch/link.go
cfromknecht: change to invalid key makes sense to me.

> Shouldn't the receiving node just ignore the duplicate package so that the htlc can live on?

If the packet is replayed it will result in two distinct htlcs (they'll have different htlc ids), so the original one lives on independently

Roasbeef:
>  In this test, two concurrent payments are executed. Because we stop tracking our own reputation in this pr,

Doesn't this series of bugs introduced show that maybe we _shouldn't_ ignore our own channels at the router level? These isolated cases in our existing integration tests have been resolved, but whose to say there aren't other regressions awaiting?

Roasbeef:
>  Because we stop tracking our own reputation in this pr, not updating the bandwidth hints led to an endless loop. 

Why does it loop endlessly rather than just fail?

htlcswitch/link.go
joostjager: Ok, clear

joostjager:
>Is this infinite loop still possible if an outdated node sends back tempchanfailure instead of invalidonionkey?

No, because the infinite loop could only happen when we got a `tempchanfail` from the self node. The self node cannot be outdated. For all other nodes on the route, we apply penalties.

>Doesn't this series of bugs introduced show that maybe we shouldn't ignore our own channels at the router level? These isolated cases in our existing integration tests have been resolved, but whose to say there aren't other regressions awaiting?

Imo it shows that bugs have been able to go undetected because we penalize our own channels to create a vaguely defined safety net. Fixing those bugs doesn't just serve the changes in this pr, but makes the performance of `lnd` really better. For example, querying bandwidth hints only once per payment may have been causing problems with payments sometimes failing in a setting with concurrent payments.

I am still convinced that we should move forward with not penalizing ourselves. Imo there isn't a fundamental reason why would need to do that and it does create a complex interaction between switch, (global) mission control and concurrent payments that is hard to understand.

I do agree that other latent bugs may surface because of this change. In that case we will log a warning, but it could be that an endless loop occurs. Not the worst thing that can happen and with a one minute payment loop timeout it doesn't last that long, but it isn't pretty.

We could add some kind of circuit breaker that for example terminates the payment loop if more than three times the same route is attempted and log an error. But it needs to be kept in mind that three consecutive failures with identical routes could also happen without any bugs: If every time a channel has balance during path finding, but not anymore during execution (because of concurrent payments) AND a payment comes in thereafter to refill the channel before the next path finding round. 

Also, the circuit breaker itself may again hide bugs because users ignore the error in the log. And it may never trip because there are no more bugs.

I tend towards reviewing the local dispatch execution path once more for any missed self-failure cases and then merging this without protection. What are your thoughts?

joostjager:
>Why does it loop endlessly rather than just fail? (tentative LGTM...but want to know why this happened)

It queried the bandwidth hints once at the start of the payment process. Bandwidth was sufficient at that point. Then a concurrent payment took the balance. The first payment process then started its attempt loop. An endless loop occurred because it tried the same depleted channel over and over again without updating the bandwidth hints. If it would have done so, it would have known that the channel had been emptied and would have given up.

joostjager:
Traced the local dispatch/response path and checked all locally generated onion failure messages in `switch.handleLocalDispatch`, `link.handleDownStreamPkt`, `link.handleUpstreamMsg` and `switch.GetPaymentResult` (and the functions they call). Error cases can be divided in three groups:

1. Errors that fail the link -> will be picked up in next path finding round
2. Errors that are checked for during path finding too (for example `max_htlc`) -> will be picked up in next path finding round
3. Unexpected errors that should not happen. Examples:
    * We can't deserialize our own locally serialized failure
    * `channel.AddHtlc` fails (no overflow)
    * Can not get our own channel policy
-> may trigger retry or an endless loop
4. Errors that could happen, but aren't picked up during path finding -> I found none of those

Assuming my assessment of this is correct, a potential endless loop can only occur for errors that shouldn't happen (3). I think that is acceptable. An endless loop (until timeout) isn't the worst that can happen and in those cases, there may be instability on other areas of `lnd` as well.  

Further into the future, we may want to reevaluate how we handle those unexpected errors. Better define how fail-over behavior should be implemented, for example by failing the link instead of just logging an error. Or alternatively don't support fail-over at all and exit `lnd` completely in case of an undefined state.

Roasbeef:
Thanks for investigating this in detail @joostjager! In the wild, it seems that even if we encounter something within the 3rd category, it won't loop _endlessly_ as we still have the payment timeout parameter. In any case, as you pointed out, re-evaluating the bandwidth hints each time should address any issues that can arise today with multiple concurrent payments, which'll really help multi-user `lnd` instances such as some of the popular custodial wallets out there.

subhramazumdar:
Hi @joostjager. I was going through the penalization mechanism of failed payments. I don't find any mention of the same BOLT (sorry may be I have missed it). I see this has already been pushed in the lnd codebase so is it somehow working inherently? Can you elaborate on how it works? How are you deciding the penalty value?

joostjager:
There is no BOLT for the penalization mechanism. It is up to the implementation team to decide.

subhramazumdar:
@joostjager do you have any documentation on how this would work, if implemented? And how does penalty value gets decided for handling each category of failure?

joostjager:
There is just the code and the comments in the code. You'll have to figure it out from there.

joostjager:
Another resource is my presentation at lnconf 2019 in Berlin, https://twitter.com/joostjgr/status/1186177262238031872

subhramazumdar:
Thanks for sharing.

Issue #3180, opened 2019-06-09
lnd segfaults when I run restorechanbackup


bob-333:
### Background

I've been having an issue with a bad channel since February. See my comments in #2582. For a while I was unable to run lnd any more - it would crash every time I started it, but I worked around that by blocking the IP of the peer on the other end of the bad channel at the firewall level.

Then yesterday I noticed I was unable to make a particular payment using my node. I tried 10 different routes and they all failed. I hadn't updated lnd from git for a while, so tried that. After updating I was unable to run lnd. It was failing with the same error as in #2582, even though the IP address was still blocked on the firewall. Luckily I had kept a backup of the full .lnd folder from just before updating lnd, so was able to rebuild the old code (commit 3a19afe4) and get lnd running again.

I was advised on the slack that I should shut down all my existing channels and start fresh, to avoid anyone having to mess about with the db to fix this bug, so that's what I did. I've closed all the channels except the broken one.

I tried exporting backup data from the broken channel, making a new node with the same wallet seed phrase and importing it there, but that causes the new lnd node to crash. I tried again, this time waiting a few hours for the new node to scan the whole blockchain to get all the onchain transactions into its db, but it made no difference; the restorechanbackup still crashes lnd.

On the old node I ran:

```
$ lncli exportchanbackup --chan_point=067119f4d38dec2d15fb1e1fc88176fc0504801245f11d5bd74371e137c2502c:0
{
	"chan_point": "067119f4d38dec2d15fb1e1fc88176fc0504801245f11d5bd74371e137c2502c:0",
[...]

PR #3257, opened 2019-06-30
chanrestore: return error instead of nil if failed derivation

Crypt-iQ:
Fixes the panic in #3180


Roasbeef:
Correct that an error here wold be unexpected. If something _is_ popping up however, then we'll be actually able to pin point the issue now that the error is actually being returned.

PR #3258, opened 2019-07-01
chainntnfs: allow registerconf without pkscript [wip]

joostjager:


chainntnfs/txnotifier.go
```diff
@@ -148,9 +148,13 @@ type ConfRequest struct {
 // dispatch the confirmation notification on the script.
 func NewConfRequest(txid *chainhash.Hash, pkScript []byte) (ConfRequest, error) {
 	var r ConfRequest
-	outputScript, err := txscript.ParsePkScript(pkScript)
-	if err != nil {
-		return r, err
+
+	if len(pkScript) > 0 {
```
wpaulino: This isn't possible now that https://github.com/lightningnetwork/lnd/pull/3254 has been merged.

joostjager: Ok. This was a draft pr to get a simple tx lookup working. It seems that we need to use a separate backend connection then.

PR #3264, opened 2019-07-03
lnwallet/fee_estimator: shorten TLS + dial timeouts for web API fee e…

valentinewallace:
…stimator

The current long timeouts can cause problems on devices that tend to have
internet connection flappiness, e.g. mobile devices. Fee estimation failures
can happen when devices switch from 4G to wifi, vice versa, and during other
such disruptions.

### DIFF
--- a/lnwallet/fee_estimator.go
+++ b/lnwallet/fee_estimator.go
@@ -711,12 +711,12 @@ func (w *WebAPIFeeEstimator) updateFeeEstimates() {
 	// overloaded, we can exit early and use our default fee.
 	netTransport := &http.Transport{
 		Dial: (&net.Dialer{
-			Timeout: 5 * time.Second,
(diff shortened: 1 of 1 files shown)


lnwallet/fee_estimator.go
```diff
@@ -711,12 +711,12 @@ func (w *WebAPIFeeEstimator) updateFeeEstimates() {
 	// overloaded, we can exit early and use our default fee.
 	netTransport := &http.Transport{
 		Dial: (&net.Dialer{
-			Timeout: 5 * time.Second,
+			Timeout: 1 * time.Second,
```
cfromknecht: if users are in a place with high latency, does this mean they may never be able to get a fee estimate? or at least with less success than they would with 5 seconds?

perhaps we should have some sort of exponential backoff?

PR #3269, opened 2019-07-03
CODEOWNERS: create initial code owners config

Roasbeef:
In this commit, we create the initial CODEOWNERS file for `lnd`. This
file is a new-ish feature of Github that allows the maintains of a
project to define the individuals who are deemed to be experts of a
particular section of the codebase. Automated review assignment systems
can then use this file to generate required reviewers for a section of
the project. The CODEOWNERS file isn't meant to be a strict guideline,
but instead can be used to ease the burden of finding reviewers for a
particular section of the codebase.

It's also possible to enforce that the CODEOWNERS for a particular
file/package sign-off on relevant changes, thereby blocking review
unless they're involved. However, for now we'll simply use it to guide
the review assignment project, leaving creating blocking review
constraints until a later point if deemed useful.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,81 @@
+* @roasbeef @cfromknecht @valentinewallace @joostjager @wpaulino @halseth
+
+azeeed/* @roasbeef @cfromknecht
+
+autopilot/* @halseth @roasbeef
+
+brontide/* @roasbeef @cfromknecht
+
+buffer/* @cfromknecht
+
+build/* @cfromknecht
+
+chainntnfs/* @halseth @wpaulino
+
+chanbackup/* @roasbeef
+
+channeldb/channel.go @roasbeef
+
+channeldb/forwarding_log.go @roasbeef
+
+channeldb/graph.go @roasbeef @halseth
+
+channeldb/invoices.go @roasbeef @joostjager
+
+channeldb/payment_control.go @halseth @joostjager
+
+channeldb/payments.go @halseth @joostjager
+
+channeldb/reject_cache.go @cfromknecht
+
+channelnotifier/* @valentinewallace
+
+contractcourt/* @roasbeef @joostjager
+
+discovery/* @halseth @wpaulino
+
+htlcswitch/* @roasbeef @cfromknecht
+
+input/* @joostjager @roasbeef
+
+keychain/* @roasbeef
+
+lnwallet/* @roasbeef @halseth
+
+macaroons/* @roasbeef
+
+monitoring/* @valentinewallace
+
+nat/* @wpaulino
+
+pool/* @cfromknecht
+
+queue/* @cfromknecht
+
+routing/* @joostjager @roasbeef
+
+sweep/* @joostjager
+
+tor/* @wpaulino
+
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,81 @@
...
+
+buffer/* @cfromknecht
+
+build/* @cfromknecht
+
+chainntnfs/* @halseth @wpaulino
+
+chanbackup/* @roasbeef
+
+channeldb/channel.go @roasbeef
```
cfromknecht: do we need a `channeldb` catch all for files other than those listed?

.github/CODEOWNERS
```diff
@@ -0,0 +1,81 @@
...
+
+contractcourt/* @roasbeef @joostjager
+
+discovery/* @halseth @wpaulino
+
+htlcswitch/* @roasbeef @cfromknecht
+
+input/* @joostjager @roasbeef
+
+input/* @joostjager
```
cfromknecht: duplicate

.github/CODEOWNERS
```diff
@@ -0,0 +1,81 @@
...
+
+tor/* @wpaulino
+
+walletunlocker/* @roasbeef
+
+watchtower/* @cfromknecht @roasbeef
+
+zpay32/* @halseth @roasbeef
+
+breacharbiter.go @cfromknecht @roasbeef
```
cfromknecht: also need catch all for `lnd` pkg?

.github/CODEOWNERS
Roasbeef: I don't think so? The post mentions [a concept of `global owners`](https://help.github.com/en/articles/about-code-owners), but it's unclear if that's a default, or if we'd need to make it ourselves.

.github/CODEOWNERS
Roasbeef: See comment above.

.github/CODEOWNERS
cfromknecht: it looks like we need to define them first:
```
# These owners will be the default owners for everything in
# the repo. Unless a later match takes precedence,
# @global-owner1 and @global-owner2 will be requested for
# review when someone opens a pull request.
*       @global-owner1 @global-owner2
```
so then if something matches after it overrides the global owners

Roasbeef: Added

.github/CODEOWNERS
```diff
@@ -0,0 +1,81 @@
+* @global-owners
```
cfromknecht: i think these need to be usernames iiuc

Roasbeef: It can be a team within the org.

cfromknecht: is @global-owners a team?

.github/CODEOWNERS
```diff
@@ -0,0 +1,81 @@
+* @roasbeef @cfromknecht @valentinewallace @joostjager @wpaulino @halseth
+
+azeeed/* @roasbeef @cfromknecht
```
Crypt-iQ: s/azeeed/aezeed

PR #3274, opened 2019-07-05
Fix typo in doc

esneider:


lnwallet/reservation.go
```diff
@@ -286,8 +286,8 @@ func (r *ChannelReservation) CommitConstraints(c *channeldb.ChannelConstraints)
 		return ErrCsvDelayTooLarge(c.CsvDelay, maxDelay)
 	}
 
-	// The dust limit should always be greater or equal to the channel
-	// reserve. The reservation request should be denied if otherwise.
+	// The channel reserve should always be greater or equal to the dust
```
halseth: should be "strictly greater"

wpaulino: I believe the change proposed is correct, it's not strict.

halseth: Yep, I read the `if` backwards 😝

PR #3278, opened 2019-07-08
channeldb: specify freelist bbolt options by default

### LINKED ISSUE #3241: Specify options to better handle boltdb fragmentation
### Background

When profiling my lnd node on testnet, I noticed a rather high heap memory with most of it due to boltdb:
```
Showing nodes accounting for 67.91MB, 98.54% of 68.92MB total
Showing top 10 nodes out of 64
      flat  flat%   sum%        cum   cum%
   24.33MB 35.30% 35.30%    24.33MB 35.30%  github.com/coreos/bbolt.(*freelist).reindex
   14.36MB 20.84% 56.14%    14.36MB 20.84%  github.com/coreos/bbolt.pgids.merge
    8.04MB 11.67% 67.80%     8.04MB 11.67%  github.com/lightningnetwork/lnd/pool.NewReadBuffer.func1
    7.20MB 10.45% 78.26%     7.20MB 10.45%  github.com/coreos/bbolt.(*DB).allocate
```
digging deeper, I saw that `UpdateEdgePolicy` was using a lot of memory in the `db.Update` function.
```
         0    22.07MB (flat, cum) 32.02% of Total
         .          .   1836:func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
         .          .   1837:	c.cacheMu.Lock()
         .          .   1838:	defer c.cacheMu.Unlock()
         .          .   1839:
         .          .   1840:	var isUpdate1 bool
         .    22.07MB   1841:	err := c.db.Update(func(tx *bbolt.Tx) error {
         .          .   1842:		var err error
         .          .   1843:		isUpdate1, err = updateEdgePolicy(tx, edge)
         .          .   1844:		return err
         .          .   1845:	})
         .          .   1846:	if err != nil {
```
digging even deeper we see that ~7MB is being allocated for every `Update` call because it is commiting a new freelist to disc every time:
```
ROUTINE ======================== github.com/coreos/bbolt.(*DB).allocate in /Users/nsa/go/pkg/mod/github.com/coreos/bbolt@v1.3.2/db.go
    7.20MB     7.20MB (flat, cum) 10.45% of Total
         .          .    914:	// Allocate a temporary buffer for the page.
         .          .    915:	var buf []byte
         .          .    916:	if count == 1 {
         .          .    917:		buf = db.pagePool.Get().([]byte)
         .          .    918:	} else {
    7.20MB     7.20MB    919:		buf = make([]byte, count*db.pageSize)
         .          .    920:	}
         .          .    921:	p := (*page)(unsafe.Pointer(&buf[0]))
         .          .    922:	p.overflow = uint32(count - 1)
         .          .    923:
         .          .    924:	// Use pages from the freelist if they are available.
```
By disabling the committing of the freelist to disk and setting the underlying freelist to hashmap-based, we see the memory usage drop since `merge` calls are cheaper with a hashmap and because 7MB isn't being allocated every `Update` call.  If we disable committing the freelist to disk, we still have to reindex the freelist upon opening the db, but for my 7MB freelist this time was negligible.
```
Showing nodes accounting for 41976.83kB, 98.79% of 42488.85kB total
Showing top 10 nodes out of 85
      flat  flat%   sum%        cum   cum%
24912.89kB 58.63% 58.63% 24912.89kB 58.63%  github.com/coreos/bbolt.(*freelist).reindex
 7168.88kB 16.87% 75.51%  7168.88kB 16.87%  github.com/lightningnetwork/lnd/htlcswitch.(*circuitMap).decodeCircuit
 2337.56kB  5.50% 81.01%  2337.56kB  5.50%  github.com/lightningnetwork/lnd/channeldb.newRejectCache
 2337.56kB  5.50% 86.51%  9506.44kB 22.37%  github.com/lightningnetwork/lnd/htlcswitch.(*circuitMap).restoreMemState.func1.1
 1646.53kB  3.88% 90.38%  1646.53kB  3.88%  github.com/lightningnetwork/lnd/pool.NewReadBuffer.func1
 1089.33kB  2.56% 92.95%  1089.33kB  2.56%  github.com/lightningnetwork/lnd/pool.NewWriteBuffer.func1
  902.59kB  2.12% 95.07%   902.59kB  2.12%  github.com/lightningnetwork/lnd/routing/chainview.(*BtcdFilteredChainView).chainFilterer
  557.26kB  1.31% 96.38%   557.26kB  1.31%  crypto/elliptic.initTable
  512.16kB  1.21% 97.59%   512.16kB  1.21%  crypto/aes.(*aesCipherGCM).NewGCM
  512.07kB  1.21% 98.79%   512.07kB  1.21%  github.com/lightningnetwork/lnd/channeldb.deserializeChanEdgePolicy
```

As a result, UpdateEdgePolicy calls go from 7MB to 513KB:
```
UpdateEdgePolicy goes from 7MB to 513KB
ROUTINE ======================== github.com/lightningnetwork/lnd/channeldb.(*ChannelGraph).UpdateEdgePolicy in /Users/nsa/go/src/github.com/lightningnetwork/lnd/channeldb/graph.go
         0   513.12kB (flat, cum) 0.064% of Total
         .          .   1836:func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
         .          .   1837:	c.cacheMu.Lock()
         .          .   1838:	defer c.cacheMu.Unlock()
         .          .   1839:
         .          .   1840:	var isUpdate1 bool
         .   513.12kB   1841:	err := c.db.Update(func(tx *bbolt.Tx) error { // 21.55MB of in-use space?
         .          .   1842:		var err error
         .          .   1843:		isUpdate1, err = updateEdgePolicy(tx, edge)
         .          .   1844:		return err
         .          .   1845:	})
         .          .   1846:	if err != nil {
```

The options can be specified in `channeldb/db.go` when opening the db like so:
```
        options := &bbolt.Options{
                NoFreelistSync: true,
                FreelistType: "hashmap"
        }
	bdb, err := bbolt.Open(path, dbFilePermission, options)
	if err != nil {
		return nil, err
	}
```

### Your environment

* version of `lnd`
master
Mac
* version of `boltdb`
1.3.2

### Steps to reproduce

Have a fragmented `channel.db`. You can check if this is the case by running:
`bolt pages channel.db` and counting the number of free pages (should be very high).  This happens when boltdb frees up a lot of pages by deleting something.  I did run `bolt compact` on my channel.db and it did get rid of *most* of the pages on my freelist, but:
1) a user shouldn't have to continually run `bolt compact` to have perf improvements on the heap
2) it doesn't always work, see: https://github.com/boltdb/bolt/issues/640

This could easily be a boltdb flag in lnd to minimize heap usage.
### Related:
https://github.com/etcd-io/bbolt/pull/141

Crypt-iQ:
Fixes #3241.

These options could really be specified anywhere that bbolt is opened, so if that's of interest I can also do that.  But the only other large database I had on my machine was `wallet.db` which is the `btcwallet` repo.  Perhaps watchtowers might use a lot of space, or there could be some design parameters that change in the future where the other databases start using more space (macaroons.db etc).  Should I add these options to all the tests so they reflect the new behavior?

Updated bbolt dependency to v1.3.3 per https://github.com/etcd-io/bbolt/pull/153
The only downside is that if the db txn has to be rolled back in case of failure, the freelist must be built again temporarily which can be heavy with big databases.  I think this is an acceptable tradeoff since this shouldn't be happening too often.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
 	github.com/btcsuite/btcwallet v0.0.0-20190628225330-4a9774585e57
 	github.com/btcsuite/fastsha256 v0.0.0-20160815193821-637e65642941
-	github.com/coreos/bbolt v1.3.2
+	github.com/coreos/bbolt v1.3.3
 	github.com/davecgh/go-spew v1.1.1
 	github.com/go-errors/errors v1.0.1
 	github.com/golang/protobuf v1.3.1
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -135,7 +135,14 @@ func Open(dbPath string, modifiers ...OptionModifier) (*DB, error) {
 		modifier(&opts)
 	}
 
-	bdb, err := bbolt.Open(path, dbFilePermission, nil)
+	// Specify bbolt freelist options to reduce heap pressure in case the
+	// freelist grows to be very large.
+	options := &bbolt.Options{
+		NoFreelistSync: true,
+		FreelistType:   bbolt.FreelistMapType,
+	}
+
+	bdb, err := bbolt.Open(path, dbFilePermission, options)
 	if err != nil {
 		return nil, err
 	}
--- a/go.sum
+++ b/go.sum
@@ -64,6 +64,8 @@ github.com/coreos/bbolt v0.0.0-20180223184059-7ee3ded59d4835e10f3e7d0f7603c42aa5
 github.com/coreos/bbolt v0.0.0-20180223184059-7ee3ded59d4835e10f3e7d0f7603c42aa5e83820/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
 github.com/coreos/bbolt v1.3.2 h1:wZwiHHUieZCquLkDL0B8UhzreNWsPHooDAG3q34zk0s=
 github.com/coreos/bbolt v1.3.2/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
+github.com/coreos/bbolt v1.3.3 h1:n6AiVyVRKQFNb6mJlwESEvvLoDyiTzXX7ORAUlkeBdY=
+github.com/coreos/bbolt v1.3.3/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
 github.com/davecgh/go-spew v0.0.0-20171005155431-ecdeabc65495/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
--- a/watchtower/wtdb/db_common.go
+++ b/watchtower/wtdb/db_common.go
@@ -63,7 +63,14 @@ func createDBIfNotExist(dbPath, name string) (*bbolt.DB, bool, error) {
 		}
 	}
 
-	bdb, err := bbolt.Open(path, dbFilePermission, nil)
+	// Specify bbolt freelist options to reduce heap pressure in case the
+	// freelist grows to be very large.
+	options := &bbolt.Options{
+		NoFreelistSync: true,
+		FreelistType:   bbolt.FreelistMapType,
+	}
+
+	bdb, err := bbolt.Open(path, dbFilePermission, options)
 	if err != nil {
 		return nil, false, err
 	}


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
 	github.com/btcsuite/btcwallet v0.0.0-20190628225330-4a9774585e57
 	github.com/btcsuite/fastsha256 v0.0.0-20160815193821-637e65642941
-	github.com/coreos/bbolt v1.3.2
+	github.com/coreos/bbolt v1.3.3
```
cfromknecht: small nit: swap the order of the commits so bbolt is updated first

Crypt-iQ:
@cfromknecht Addressed comment and added the options to watchtower client + server.  Can make a PR for btcwallet next.  For testing, I'm wondering if the options should also be specified in unit/itests?  If so, in order to reduce code duplication, I suppose I could add the options to `channeldb/options.go` similar to the way the cache sizes are setup?

cfromknecht:
@Crypt-iQ since the options are set inside the `Open` method, wouldn't these already be used in testing as well? Which cases do you think are still missing?

Crypt-iQ:
> @Crypt-iQ since the options are set inside the `Open` method, wouldn't these already be used in testing as well? Which cases do you think are still missing?

Hey you're right, I meant the other databases (sphinxreplay, macaroons) and some of the temporary test databases.  I didn't look at the results of my search very closely, mb

cfromknecht:
> Hey you're right, I meant the other databases (sphinxreplay, macaroons) and some of the temporary test databases. I didn't look at the results of my search very closely, mb

okay cool then i think we're good. sphinxreplay and macaroon dbs are less of a priority as they're much smaller, but not opposed to adding them as well

Crypt-iQ:
> > Hey you're right, I meant the other databases (sphinxreplay, macaroons) and some of the temporary test databases. I didn't look at the results of my search very closely, mb
> 
> okay cool then i think we're good. sphinxreplay and macaroon dbs are less of a priority as they're much smaller, but not opposed to adding them as well

The freelist on my 8MB sphinxreplay db has 5 free pages on it... so I don't think I'll include it here.  And my macaroons.db is tiny.

cfromknecht:
@Crypt-iQ decided we should push this off to 0.8, since it's a pretty big change for a point release and want to leave ample time for testing

Crypt-iQ:
@cfromknecht no problem, I'm looking into the InitialMmapSize option as well, so might be good to hold off until I can determine if it's worth adding

cfromknecht:
@Crypt-iQ sounds good, yeah we haven't done much exploration wrt bbolt options, would be interested to hear what you find!

Crypt-iQ:
Looked at `InitialMmapSize` and some of the related bbolt code.  Bbolt uses mmap to map the db file and sometimes it needs to be remapped via `munmap` and `mmap`.  In these cases, any write transactions that are referencing the old mmap need to copy these references to the heap.  So IMO it seems like we should just set a high `InitialMmapSize` and never have to remap.  Thoughts on size?

Crypt-iQ:
I'm kind of indifferent on the `InitialMmapSize` parameter.  I tested and couldn't, in any reasonable (non-test) circumstance, produce any huge memory allocations when resizing the mmap.  It is also not resized very often so it's not likely to occur that long-running reads (which there aren't that many of) block special write txn's.

Roasbeef:
> I'm kind of indifferent on the InitialMmapSize parameter.

I don't think we need to set this value in this PR, and can defer tuning it to a later PR. 

> But the only other large database I had on my machine was wallet.db which is the btcwallet repo.

Once this goes in, we can follow up with the change on the `btcwallet` side as well.

Issue #2344, opened 2018-12-18
REST API CORS headers


wbobeirne:
### Background

If a web application wanted to use an LND node, requests would be blocked even if the REST API were exposed due to the lack of a `Access-Control-Allow-Origin` header. I don't know that there's much harm in setting it to `*` by default, since you'd need macaroons for the node to do pretty much anything anyway.

This is somewhat of a duplicate of #1405, however that issue went a little off-topic with a fork of LND.

### Your environment

* version of `lnd` - 0.5.1
* which operating system (`uname -a` on *Nix) - Ubuntu Server 16.04
* version of `btcd`, `bitcoind`, or other backend - bitcoind

tiero:
As far I know, shouldn't be needed a `grpc-web` proxy running to serve browsers?

wbobeirne:
> As far I know, shouldn't be needed a `grpc-web` proxy running to serve browsers?

Not sure I understand what you're asking, but there's a REST API is provided by `grpc-gateway`, which is what browsers can use to communicate.

It looks like #1433 was an attempt at adding this. However, it looks like that's gone a bit stale. It also added a config option, rather than CORS headers by default, but I think that'd be satisfactory.

PR #1433, opened 2018-06-22
Allow CORS

phillyfan1138:
This PR allows CORS IP addresses to be entered in the lnd.conf similar to rpclisten and restlisten.  However, the protocol (HTTP/HTTPS) needs to be specified.

PR #3281, opened 2019-07-09
rpcserver: add cors support

AnthonyRonning:
This adds cors support to the LND http proxy, ticket created on it here: https://github.com/lightningnetwork/lnd/issues/2344

Basing it off of this stale PR: https://github.com/lightningnetwork/lnd/pull/1433
Key differences: 
-cors automatically
-doing so without pulling in a separate library

This should allow browser based wallets to access LND nodes via http, since browsers cannot access via grcp. Otherwise, browsers need to run with security disabled which do not check cors.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -690,31 +690,27 @@ func (r *rpcServer) Start() error {
 	return nil
 }
 
-// checkCors allows Cross Origin Resoruce Sharing from configured origins.
+// checkCors allows Cross Origin Resource Sharing from configured origins.
 func checkCors(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		origin := r.Header.Get("Origin")
 
 		// Skip CORS check if request did not send origin.
 		if origin != "" {
-			// Check if all origins are allowed or requested origin is in list.
+			// Check if origin is allowed.
 			for _, allowedOrigin := range cfg.RestCORS {
-				if allowedOrigin == "*" || origin == allowedOrigin {
-					// Only set ACAO header for requested origin.
-					w.Header().Set("Access-Control-Allow-Origin", origin)
+				if allowedOrigin == "*" ||
+					origin == allowedOrigin {
 
-					// Set necessary headers for preflight requests.
-					if r.Method == "OPTIONS" &&
-						r.Header.Get("Access-Control-Request-Method") != "" {
+					// Only set ACAO for requested origin.
+					acaoStr := "Access-Control-Allow-Origin"
+					w.Header().Set(acaoStr, origin)
 
-						w.Header().Set("Access-Control-Allow-Headers",
-							"Content-Type, Accept, grpc-metadata-macaroon")
-						w.Header().Set("Access-Control-Allow-Methods",
-							"GET, POST, DELETE")
-
-						// Nothing else needs to be served for OPTIONS.
+					// Preflights only need headers back.
+					if checkPreflight(w, r) {
 						return
 					}
+
 					break
 				}
 			}
@@ -722,6 +718,23 @@ func checkCors(h http.Handler) http.Handler {
 
 		h.ServeHTTP(w, r)
 	})
+}
+
+// checkPreflight will check if the request is an appropriate preflight
+// request and add additional access control headers if so.
+func checkPreflight(w http.ResponseWriter, r *http.Request) bool {
+	if r.Method == "OPTIONS" &&
+		r.Header.Get("Access-Control-Request-Method") != "" {
+
+		w.Header().Set("Access-Control-Allow-Headers",
+			"Content-Type, Accept, grpc-metadata-macaroon")
+		w.Header().Set("Access-Control-Allow-Methods",
+			"GET, POST, DELETE")
+
+		return true
+	}
+
+	return false
 }
 
 // Stop signals any active goroutines for a graceful closure.
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -697,7 +697,7 @@ func allowCORS(h http.Handler) http.Handler {
 			w.Header().Set("Access-Control-Allow-Headers",
 				"Content-Type, Accept, grpc-metadata-macaroon")
 			w.Header().Set("Access-Control-Allow-Methods",
-				"GET, POST")
+				"GET, POST, DELETE")
 			return
 		}
 
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -679,26 +679,44 @@ func (r *rpcServer) Start() error {
 
 		go func() {
 			rpcsLog.Infof("gRPC proxy started at %s", lis.Addr())
-			http.Serve(lis, allowCORS(mux))
+			if len(cfg.RestCORS) == 0 {
+				http.Serve(lis, mux)
+			} else {
+				http.Serve(lis, checkCors(mux))
+			}
 		}()
 	}
 
 	return nil
 }
 
-// allowCORS allows Cross Origin Resoruce Sharing from any origin.
-func allowCORS(h http.Handler) http.Handler {
+// checkCors allows Cross Origin Resoruce Sharing from configured origins.
+func checkCors(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Access-Control-Allow-Origin", "*")
+		origin := r.Header.Get("Origin")
 
-		// Set necessary headers for preflight requests
-		if r.Method == "OPTIONS" &&
-			r.Header.Get("Access-Control-Request-Method") != "" {
-			w.Header().Set("Access-Control-Allow-Headers",
-				"Content-Type, Accept, grpc-metadata-macaroon")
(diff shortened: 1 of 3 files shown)


rpcserver.go
```diff
@@ -679,13 +679,32 @@ func (r *rpcServer) Start() error {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+
+		// Set necessary headers for preflight requests
+		if r.Method == "OPTIONS" &&
+			r.Header.Get("Access-Control-Request-Method") != "" {
+			w.Header().Set("Access-Control-Allow-Headers",
+				"Content-Type, Accept, grpc-metadata-macaroon")
+			w.Header().Set("Access-Control-Allow-Methods",
+				"GET, POST")
```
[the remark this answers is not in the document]
AnthonyRonning: Oh nice catch, will add `DELETE`

AnthonyRonning:
@MDrollette Thanks for the comment/review. 

I'm not opposed to adding it as a config value, though I do want to mitigate any concerns there might be by having it on by default, if possible. And to see if I'm missing something as well.

CORS is a request from a server to the browser asking it to obey its rules. It does not have to be obeyed, a user can turn off this setting and access from any origin anyways. 

A leaked macaroon can be remotely exploitable whether or not cors is on or off. Cors only stops client side code in a browser (w/ default cors settings) from accessing resources/api's on another webserver somewhere else. 

That said, I think the only way for something to take place if cors is on vs off is if a website not meant for accessing LND got ahold of someone's macaroon, whether by user file upload or something, and decided to execute the macaroon right there in the user's browser. So let's say cors is off and the execution didn't work in the user's browser, then the website uploads it to a server and the server admin can try to execute requests to LND. Only thing that would prevent that would be IP-locks/firewalls, right? 

So essentially, malicious client side code in a browser that can't get passed CORS, gets access to a user's macaroon and knows the user's LND IP/Port to access his IP locked LND node and executes commands right there in the browser. Or the LND node is not IP locked but the malicious code can't upload the macaroon to the malicious actor so it needs to execute commands in the browser.

That could very well be reason enough to have it off by default, but I did want to dispel the concern of simply 'macaroon is leaked, it would be remotely exploitable if cors is on'. Thoughts?

MDrollette:
Assuming you had an otherwise secure setup where the gRPC/HTTP listeners are limited to `localhost` and with cors disabled, even if I was able to get a readonly macaroon of yours (say from an errant git commit) I wouldn't be able to do much of anything with it. But, if you had cors enabled and allowed all origins (`*`), I would be able to send you a link to my own website which has some malicious javascript that would read out all your node info and post it back to my server. A restrictive cors policy, or having it off by default, would prevent a leaked macaroon from being exploitable via a users browser in that scenario.

I don't know how big of a concern that is, but it's worth pointing out. I would prefer the option, at least, and off by default would be even better, imo.

AnthonyRonning:
Makes sense. I appreciate the dialog! I'll get the config in there defaulted to off this weekend.

cfromknecht:
@AnthonyRonning @MDrollette definitely agree that we probably don't want to enable wildcard cors by default, would much rather we have that opt in.

I'm thinking it'd be nice to be able to specify either `*` or a list of hostnames that are whitelisted. The list of hostnames is a little trickier, as we'll have to inject the requesting host into the header if it passes the whitelist, but it should still be possible to accomplish via middleware as you've already done. As far as a config name, maybe something like `restcors`?

> -doing so without pulling in a separate library

Nice! We try to keep our deps slim :)

AnthonyRonning:
Good suggestions @cfromknecht, I added it as an option, defaulted off, where you can add whitelisted ips/domains or `*` via `restcors`.

rpcserver.go
```diff
@@ -679,13 +679,50 @@ func (r *rpcServer) Start() error {
...
 }
 
+// checkCors allows Cross Origin Resoruce Sharing from configured origins.
+func checkCors(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		origin := r.Header.Get("Origin")
+
+		// Skip CORS check if request did not send origin
+		if origin == "" {
+			return
```
MDrollette: Should you still call `h.ServeHTTP(w, r)` for these non-browser http requests?

AnthonyRonning: Yeah, nice catch. Updated.

rpcserver.go
```diff
@@ -679,13 +679,50 @@ func (r *rpcServer) Start() error {
...
 
+// checkCors allows Cross Origin Resoruce Sharing from configured origins.
+func checkCors(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		origin := r.Header.Get("Origin")
+
+		// Skip CORS check if request did not send origin
+		if origin != "" {
+			// Check if all origins are allowed or requested origin is in list
+			for _, allowedOrigin := range cfg.RestCORS {
```
cfromknecht: to avoid going over the 80-char line limit, it might help to break this loop out into a helper method

rpcserver.go
```diff
@@ -679,13 +679,50 @@ func (r *rpcServer) Start() error {
...
 
 	return nil
 }
 
+// checkCors allows Cross Origin Resoruce Sharing from configured origins.
+func checkCors(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		origin := r.Header.Get("Origin")
+
+		// Skip CORS check if request did not send origin
```
cfromknecht: nit: inline comments are missing punctuation

rpcserver.go
```diff
@@ -679,13 +679,50 @@ func (r *rpcServer) Start() error {
...
+			// Check if all origins are allowed or requested origin is in list
+			for _, allowedOrigin := range cfg.RestCORS {
+				if allowedOrigin == "*" || origin == allowedOrigin {
+					// Only set ACAO header for requested origin
+					w.Header().Set("Access-Control-Allow-Origin", origin)
+
+					// Set necessary headers for preflight requests
+					if r.Method == "OPTIONS" &&
+						r.Header.Get("Access-Control-Request-Method") != "" {
+						w.Header().Set("Access-Control-Allow-Headers",
```
cfromknecht: nit: add new line between condition and code block

rpcserver.go
AnthonyRonning: Got em all

rpcserver.go
AnthonyRonning: Good catch, that was looking a little weird.

rpcserver.go
AnthonyRonning: Right now it's under 80, longest line is the line 700 with 79 (substituting tabs w/ spaces and then checking count). Would need to restructured this logic a little bit if pulling out into another method, with the return that's in there and all, but think it's still suggested even if it's under 80? 

Is it the tabs making it look longer than 80 in github? I notice the project uses tabs instead of spacing.

cfromknecht: > Is it the tabs making it look longer than 80 in github? I notice the project uses tabs instead of spacing.

All go code uses tabs since `go fmt` will convert spaces into tabs. Our editors are configured to use 8-space tabs to match the formatting on github, which makes some of these lines 100+ characters.

> Would need to restructured this logic a little bit if pulling out into another method, with the return that's in there and all, but think it's still suggested even if it's under 80?

Yeah i think returning a bool indicating if it's a preflight request would be sufficient

AnthonyRonning: Ohhh gotcha, okay yeah it's definitely too long, will work on that!

The contrib guide may need updating to match the 8-space rule: https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#48-additional-style-constraints-on-top-of-gofmt

cfromknecht: Good catch! That's definitely out of date 😂

AnthonyRonning: latest refactoring should put it behind 80 now

cfromknecht:
@AnthonyRonning latest changes look good to me! Will need to be squash, perhaps down to two commits (1 for config, then 1 for lnd+rpcserver). Can probably defer that though until we get a second reviewer

Roasbeef:
Rather than exposing so much control, how about we just target this particular use case? So a mode that _only_ allows select macaroons to be served, rather than giving access to the entire on-disk state. That would be easier to use, more directed, and also more secure since it exposes a smaller surface area.

Roasbeef:
Also the current implementation doesn't seem very user friendly: for each website I want to allow, i need to _restart_ my `lnd` node and edit its config.

AnthonyRonning:
@Roasbeef 

> a mode that only allows select macaroons to be served

So a mode that turns on cors and restricts macaroons? By "select macaroons", do you mean allowing users to select the macaroons they want to allow in this mode, or only allow certain types of macaroons? IE only allow read-only macaroons? 

> i need to restart my lnd node and edit its config

Then maybe expose a new endpoint to allow hosts to be added/removed from the list of allowed sites? In addition to the starting list set by the config?

Roasbeef:
will be done in https://github.com/lightningnetwork/lnd/pull/4141

PR #3283, opened 2019-07-09
multi: move several wait group done calls to defer statements

federicobond:
This change provides stronger guarantees that the wait group counter will be decremented when the goroutine returns or panics.


guggero:
Taken over and replaced by #4790, original commit attributed to this PR's author.

PR #3284, opened 2019-07-09
build: update dependencies to prevent unnecessary bitcoind ZMQ timeouts

wpaulino:
* We address an issue where we would unnecessarily time out bitcoind ZMQ connections if there are no messages to be read. This would cause connections to be torn down, which would then trigger the reconnection logic and prevent us from detecting this issue.

* A new sanity check within the wallet is done to prevent adding unconfirmed transactions that the wallet has already recognized as confirmed.


halseth:
Can be updated to the latest `btcwallet` commit now?

wpaulino:
It already is.

halseth:
Looks like https://github.com/btcsuite/btcwallet/pull/633 was merged after.

PR #3292, opened 2019-07-10
multi: fix various typos in comments and docs

chokoboko:


halseth:
Thanks for making this PR, but it unfortunately touches too many files to not cause conflicts if it gets merged. It is better to create smaller PRs (one for each subdirectory perhaps) to make it both easier to review and reduce the risk of conflicts with other PRs.

Issue #3209, opened 2019-06-16
Use mission control information in QueryRoutes


C-Otto:
According to the discussion in https://github.com/lightningnetwork/lnd/issues/3206, lnd does not make use of mission control information about failing routes/edges/nodes when returning a route for a QueryRoutes request. I'd like lnd to return a different route after a failed payment when using QueryRoutes.

Roasbeef:
It _does_ take them into account (same code pathway for path finding), failures will also be incorporated.

joostjager:
`QueryRoutes` does not take into account mission control information.

https://github.com/lightningnetwork/lnd/blob/d6d87e12feeb197f599a329c079ecef17a4f2b4a/lnrpc/routerrpc/router_backend.go#L140

There it passes in a probability source that just returns binary probabilities based on caller supplied ignore lists.

joostjager:
@C-Otto There are currently basically two models of using `lnd` for payments:

* `SendPayment`: `lnd` handles everything. It launches multiple payment attempts if needed and keeps track of node reputations in mission control.

* `QueryRoutes` + `routerrpc.SendToRoute`: these are lower level calls that require the caller to have its own reputation management system (mission control). Callers can use the detailed error returned from `SendToRoute` to update their own reputation data. Then use the updated node reputations to build new ignore lists to pass into `QueryRoutes`. Basically replicating the logic that is behind `SendPayment` in `lnd`.

   A useful follow up could be to allow probabilities to be passed into `QueryRoutes` instead of the ignore lists.

The only point where both models touch is `SendToRoute` passing every outcome into `lnd`'s mission control. The idea is that if a user uses both models, future `SendPayment` calls can benefit from what was learnt through `SendToRoute`.

Your request in this issue is mix both models. How exactly would you be using this if it were possible?

C-Otto:
I don't know a lot about lnd internals, and I don't want to interfere with plans or mix code that shouldn't be mixed. But I can explain my perspective.

I'd like to rebalance a specific channel. For that I need to create my own route, i.e. extend some route by one additional hop back to my own node. To find this route I need to use QueryRoutes.

However, I'm not interested in adding any mission control logic in my code. My very naive assumption is that lnd is able to figure out which nodes/edges/... are broken, and ignore them based on some clever heuristics. As such, I expect routes returned by QueryRoutes to be usable (at least when trying multiple times).

[...]

PR #3298, opened 2019-07-11
routing: globalize mc

joostjager:
This PR moves the last payment-specific mission control state into the global mission control instances.

Builds off of #3188 

The PR is a base for #3221 and #3164

### DIFF
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -133,40 +133,6 @@ func NewMissionControl(cfg *MissionControlConfig) *MissionControl {
 	}
 }
 
-// generateBandwidthHints is a helper function that's utilized the main
-// findPath function in order to obtain hints from the lower layer w.r.t to the
-// available bandwidth of edges on the network. Currently, we'll only obtain
-// bandwidth hints for the edges we directly have open ourselves. Obtaining
-// these hints allows us to reduce the number of extraneous attempts as we can
-// skip channels that are inactive, or just don't have enough bandwidth to
-// carry the payment.
-func generateBandwidthHints(sourceNode *channeldb.LightningNode,
-	queryBandwidth func(*channeldb.ChannelEdgeInfo) lnwire.MilliSatoshi) (map[uint64]lnwire.MilliSatoshi, error) {
-
-	// First, we'll collect the set of outbound edges from the target
-	// source node.
-	var localChans []*channeldb.ChannelEdgeInfo
-	err := sourceNode.ForEachChannel(nil, func(tx *bbolt.Tx,
-		edgeInfo *channeldb.ChannelEdgeInfo,
-		_, _ *channeldb.ChannelEdgePolicy) error {
-
-		localChans = append(localChans, edgeInfo)
-		return nil
-	})
-	if err != nil {
-		return nil, err
-	}
-
-	// Now that we have all of our outbound edges, we'll populate the set
-	// of bandwidth hints, querying the lower switch layer for the most up
-	// to date values.
-	bandwidthHints := make(map[uint64]lnwire.MilliSatoshi)
-	for _, localChan := range localChans {
-		bandwidthHints[localChan.ChannelID] = queryBandwidth(localChan)
-	}
-
-	return bandwidthHints, nil
-}
 
 // ResetHistory resets the history of MissionControl returning it to a state as
 // if no payment attempts have been made.
--- a/routing/missioncontrol.go
+++ b/routing/missioncontrol.go
@@ -5,12 +5,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/btcsuite/btcd/btcec"
 	"github.com/coreos/bbolt"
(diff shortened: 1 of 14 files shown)


lnwire/onion_error_test.go
```diff
@@ -180,7 +180,7 @@ func TestFailUnknownPaymentHashOptionalAmount(t *testing.T) {
 	onionError := &FailUnknownPaymentHash{}
 
 	var b bytes.Buffer
-	if err := EncodeFailure(&b, onionError, 0); err != nil {
+	if err := EncodeFailure(&b, &onionError, 0); err != nil {
```
cfromknecht: commit message indicates that there were two cases, but i only see one. was the other fixed somewhere else?

routing/payment_session_source.go
```diff
@@ -0,0 +1,150 @@
...
+	SelfNode *channeldb.LightningNode
+
+	// MissionControl is a shared memory of sorts that executions of payment
+	// path finding use in order to remember which vertexes/edges were
+	// pruned from prior attempts. During payment execution, errors sent by
+	// nodes are mapped into a vertex or edge to be pruned. Each run will
+	// then take into account this set of pruned vertexes/edges to reduce
+	// route failure and pass on graph information gained to the next
+	// execution.
+	MissionControl missionControlInterface
```
cfromknecht: it's a little strange to have a public struct export a field with a private interface, but not a huge deal. it just won't render properly on the godoc site. does this field need to be exported?

routing/router.go
```diff
@@ -1822,6 +1824,40 @@ func (r *ChannelRouter) sendPayment(
...
+	}
+
+	// Parse pubkey to allow validation of the channel update.
+	errVertex := rt.Hops[errorSourceIdx-1].PubKeyBytes
+	errSource, err := btcec.ParsePubKey(
+		errVertex[:], btcec.S256(),
+	)
+	if err != nil {
+		log.Debugf("Cannot parse pubkey: idx=%v, pubkey=%v",
+			errorSourceIdx, errVertex)
```
cfromknecht: exit after error?

routing/missioncontrol.go
```diff
@@ -16,6 +14,30 @@ const (
...
+	// second chance is sufficiently long ago. This is what
+	// minSecondChanceInterval defines. If a second policy failure comes in
+	// within that interval, we will apply a penalty.
+	//
+	// Second chances granted are tracked on the level of node pairs. This
+	// means that if a node has multiple channels to the same peer, they
+	// will only get a single second chance to route to that peer again.
+	// Nodes forward non-strict, so it isn't necessary to apply a less
+	// restrictive channel level tracking scheme here.
+	minSecondChanceInterval = time.Minute
```
cfromknecht: should this be made public so that it can be configured later?

lnwire/onion_error_test.go
joostjager: This commit has already been merged as part of #3188. What is left here is actually incorrect. Commit dropped.

routing/payment_session_source.go
joostjager: It is populated in `server.go`, so the field needs to be exported. The interface itself is more like a package-internal way to be able to inject a mock. I can make the interface itself public, but it wouldn't serve a purpose at the moment.

routing/router.go
joostjager: Good catch. Actually this is an internal error that should never happen. Changed code to propagate the error up the call tree.

routing/missioncontrol.go
joostjager: To me there are two logical choices here:

* leave as is
* make public and also expose it as a config option. 

I think that the value here is not very critical. It is mainly to have some sort of delay to prevent a node from endlessly sending slightly different channel updates that make us keep trying the same route. It is also difficult to explain all those fine tune parameters to users.

For this moment, I tend to leaving it as is. But let me know if you see this differently.

joostjager:
Tacked on one more commit that cleans up the payment session, making the scope of this PR a bit more complete.

joostjager:
@roasbeef you originally reviewed these commits in #3164. Replies to those comments are in that pr.

routing/missioncontrol.go
cfromknecht: sounds good to me 👍

routing/payment_session_source.go
cfromknecht: i think it'd be fine to export the interface, a la `MissionController`. Because it has private methods, only types defined in the `routing` package will be able to satisfy it, which fits our use case without breaking godoc

joostjager: Ack on the name `MissionController`. It remains difficult sometimes to find a good name. Interface exported.

joostjager: Exported all methods, as we already start exporting one in the follow up pr #3209.

cfromknecht: Awesome, latest diff lgtm!

PR #3303, opened 2019-07-11
cmd: Add passphrase confirmation when creating wallet from existing seed

### LINKED ISSUE #1507: lncli should ask to repeat passphrase
### Background

`lncli` should ask to repeat passphrase for improved user experience/security

### Your environment

`lnd version 0.4.2-beta commit=a0b2fadea35fa4642daf4e9f56e6ecfe31d22ce1`

### Steps to reproduce

`lncli create`

Use existing seed. Add passphrase.

### Expected behaviour

Should ask to repeat passphrase

### Actual behaviour

Only asks once.

carlaKC:
Require that users confirm their passphrase when providing their own seed with a passphrase.
Currently implemented as an infinite loop until passwords match, can be limited to a set number of retries if that's the UX wanted. 

Fixes #1507

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -1398,32 +1398,11 @@ func create(ctx *cli.Context) error {
 	client, cleanUp := getWalletUnlockerClient(ctx)
 	defer cleanUp()
 
-	// First, we'll prompt the user for their passphrase twice to ensure
-	// both attempts match up properly.
-	fmt.Printf("Input wallet password: ")
-	pw1, err := terminal.ReadPassword(int(syscall.Stdin))
+	walletPassword, err := capturePassword(
+		"Input wallet password: ", false, walletunlocker.ValidatePassword,
+	)
 	if err != nil {
 		return err
-	}
-	fmt.Println()
-
-	fmt.Printf("Confirm wallet password: ")
-	pw2, err := terminal.ReadPassword(int(syscall.Stdin))
-	if err != nil {
-		return err
-	}
-	fmt.Println()
-
-	// If the passwords don't match, then we'll return an error.
-	if !bytes.Equal(pw1, pw2) {
-		return fmt.Errorf("passwords don't match")
-	}
-
-	// If the password length is less than 8 characters, then we'll
-	// return an error.
-	pwErr := walletunlocker.ValidatePassword(pw1)
-	if pwErr != nil {
-		return pwErr
 	}
 
 	// Next, we'll see if the user has 24-word mnemonic they want to use to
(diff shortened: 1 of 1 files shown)


carlaKC:
Okay got it, I wasn't super clear on what needed to be done from the original issue, will update :)

While I'm at it, I'm wondering if it's worth adding similar password retry to creating the wallet password? Not sure if endless password loops everywhere are ideal, but adding it there would make the whole flow consistent?

cfromknecht:
@carlaKC sure that sounds great! should be able to reuse a good amount of code between the two :)

cmd/lncli/commands.go
```diff
@@ -1650,6 +1612,50 @@ mnemonicCheck:
 	return nil
 }
 
+func capturePassword(instruction, noun string, optional bool,
```
halseth: should have a godoc

cmd/lncli/commands.go
```diff
@@ -1539,47 +1517,31 @@ mnemonicCheck:
...
 		// command.
 		fmt.Println("Your cipher seed can optionally be encrypted.")
-		fmt.Printf("Input your passphrase if you wish to encrypt it " +
+
+		instruction := "Input your passphrase if you wish to encrypt it " +
 			"(or press enter to proceed without a cipher seed " +
-			"passphrase): ")
-		aezeedPass1, err := terminal.ReadPassword(int(syscall.Stdin))
+			"passphrase): "
+		aezeedPass, err = capturePassword(instruction, "azseed",
```
halseth: s/azseed/aezeed

cmd/lncli/commands.go
```diff
@@ -1650,6 +1612,50 @@ mnemonicCheck:
...
+
+		// If the password provided is not valid, restart
+		// password capture process from the beginning.
+		if err := validate(password); err != nil {
+			fmt.Println(err.Error())
+			fmt.Println()
+			continue
+		}
+
+		fmt.Printf("Confirm %v password: ", noun)
```
halseth: I think the noun is unnecessary here.

cmd/lncli/commands.go
```diff
@@ -1650,6 +1612,50 @@ mnemonicCheck:
...
+			return nil, err
+		}
+		fmt.Println()
+
+		if bytes.Equal(password, passwordConfirmed) {
+			return password, nil
+		}
+
+		fmt.Printf("%v passwords don't match, "+
+			"please try again", noun)
```
halseth: Same, don't need to take the noun.

cmd/lncli/commands.go
```diff
@@ -1398,33 +1398,11 @@ func create(ctx *cli.Context) error {
...
-	fmt.Printf("Input wallet password: ")
-	pw1, err := terminal.ReadPassword(int(syscall.Stdin))
-	if err != nil {
-		return err
-	}
-	fmt.Println()
-
-	fmt.Printf("Confirm wallet password: ")
-	pw2, err := terminal.ReadPassword(int(syscall.Stdin))
+	walletPassword, err := capturePassword("Input wallet password: ",
```
cfromknecht: nit: suggested wrapping
```golang
walletPassword, err := capturePassword(
        "Input wallet password: ", false, walletunlocker.ValidatePassword,
)
```
as a general rule if function calls can't fit on one line the preferred wrapping is:
```
function(
        arg1, arg2, arg3,
)
```

cmd/lncli/commands.go
```diff
@@ -1539,47 +1517,31 @@ mnemonicCheck:
...
 		// command.
 		fmt.Println("Your cipher seed can optionally be encrypted.")
-		fmt.Printf("Input your passphrase if you wish to encrypt it " +
+
+		instruction := "Input your passphrase if you wish to encrypt it " +
 			"(or press enter to proceed without a cipher seed " +
-			"passphrase): ")
-		aezeedPass1, err := terminal.ReadPassword(int(syscall.Stdin))
+			"passphrase): "
+		aezeedPass, err = capturePassword(instruction, true,
```
cfromknecht: similar wrapping comment:
```golang
aezeedPass, err = capturePassword(
        instruction, true, func(password []byte) error {
	        return nil
	},
)
```
also since we don't use the password argument, we can replace `password` with `_` if space is tight :)

cmd/lncli/commands.go
```diff
@@ -1650,6 +1612,56 @@ mnemonicCheck:
 	return nil
 }
 
+// capturePassword returns a password value that has been
```
cfromknecht: Excellent description! The line wrapping on the godoc looks like it breaks before 80 chars, but not a blocker

carlaKC: Thanks!
I was wrapping at 72 because I remembered seeing it somewhere in the contributions guide. but that was for commit messages 🤦‍♀

cmd/lncli/commands.go
```diff
@@ -1650,6 +1612,55 @@ mnemonicCheck:
...
+
+		// Do not require users to repeat password if
+		// it is optional and they are not using one.
+		if len(password) == 0 && optional {
+			return nil, nil
+		}
+
+		// If the password provided is not valid, restart
+		// password capture process from the beginning.
+		if err := validate(password); err != nil {
```
halseth: Good call to do validation before asking to repeat! 😀

PR #3307, opened 2019-07-13
aezeed: publicly export the `word` field  in `ErrUnknownMnenomicWord`

Roasbeef:
In this commit, we publicly export the `word` field as it makes it
easier to programmatically interact with the package when attempting to
re-derive proper `cipherseed` instances.

### DIFF
--- a/aezeed/cipherseed.go
+++ b/aezeed/cipherseed.go
@@ -510,10 +510,13 @@ func (m *Mnemonic) Decipher(pass []byte) ([DecipheredCipherSeedSize]byte, error)
 	// Before we attempt to map the mnemonic back to the original
 	// ciphertext, we'll ensure that all the word are actually a part of
 	// the current default word list.
-	for _, word := range m {
+	for i, word := range m {
 		if !strings.Contains(englishWordList, word) {
 			emptySeed := [DecipheredCipherSeedSize]byte{}
-			return emptySeed, ErrUnknownMnenomicWord{word}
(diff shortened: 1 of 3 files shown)


aezeed/cipherseed.go
```diff
@@ -513,7 +513,7 @@ func (m *Mnemonic) Decipher(pass []byte) ([DecipheredCipherSeedSize]byte, error)
 	for _, word := range m {
 		if !strings.Contains(englishWordList, word) {
 			emptySeed := [DecipheredCipherSeedSize]byte{}
-			return emptySeed, ErrUnknownMnenomicWord{word}
+			return emptySeed, ErrUnknownMnenomicWord{Word: word}
```
cfromknecht: i wonder if it also makes sense to export the index of the failed word, just to alleviate any ambiguity if invalid word looks a lot like another valid one. but doesn't seem critical to this change

cfromknecht:
```
aezeed/cipherseed_test.go:546:12:error: wordErr.word undefined (type ErrUnknownMnenomicWord has no field or method word, but does have Word) (vet)
aezeed/cipherseed_test.go:547:64:error: wordErr.word undefined (type ErrUnknownMnenomicWord has no field or method word, but does have Word) (vet)
```
missing some changes in the tests

aezeed/cipherseed.go
Roasbeef: In my current use case, I use the `string` package to find the index of that word to swap it out. However, may be insufficient in many scenarios.

PR #3309, opened 2019-07-13
Improve error message of ErrChanTooSmall

guycall:
Error message was ambiguous about what is imposing the channel size limit.

### DIFF
--- a/lnwallet/errors.go
+++ b/lnwallet/errors.go
@@ -137,8 +137,8 @@ func ErrNumConfsTooLarge(numConfs, maxNumConfs uint32) error {
 // configured value for the min channel size we'll accept.
 func ErrChanTooSmall(chanSize, minChanSize btcutil.Amount) ReservationError {
 	return ReservationError{
-		fmt.Errorf("Your %v channel is below the %v minimum size "+
-			"set on the connecting node", chanSize, minChanSize),
+		fmt.Errorf("your channel size of %v is below the remote "+
+			"node's %v minimum size", chanSize, minChanSize),
 	}
 }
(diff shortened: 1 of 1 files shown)


lnwallet/errors.go
```diff
@@ -137,8 +137,8 @@ func ErrNumConfsTooLarge(numConfs, maxNumConfs uint32) error {
 // configured value for the min channel size we'll accept.
 func ErrChanTooSmall(chanSize, minChanSize btcutil.Amount) ReservationError {
 	return ReservationError{
-		fmt.Errorf("chan size of %v is below min chan size of %v",
-			chanSize, minChanSize),
+		fmt.Errorf("Your %v channel is below the %v minimum size "+
```
tyzbit: Would render like:
```
Unable to open channel: Your 0.1BTC channel is below the 0.5BTC minimum size set on the connecting node
```
I think we want to start lowercase here (though other error messages are inconsistent).  How about this:
```
"channel size of %v is below the remote node's minimum size of %v"
```
Which would render like:
```
Unable to open channel: channel size of 0.1BTC is below the remote node's minimum size of 0.2BTC
```

guycall:
Made a further change. It would render as:

```
Unable to open channel: your channel size of 0.01BTC is below the remote node's 0.05BTC minimum size
```

halseth:
Thanks for the PR, I agree that it is ambiguous. However, this is an error sent to the remote peer, so it should be up to the remote peer how to show this error.

Feel free to open a PR that changes how `lncli` displays error coming from the remote!

PR #3315, opened 2019-07-15
discovery: don't reply with historical updates to gossip filters

cfromknecht:
This PR removes the existing behavior that responds with all known
updates when a remote peer sends a gossip_timestamp_filter. When a peer
sets this with a very old start timestamp, this can cause the local node
to dump the entire graph to the peer, which is functionally equivalent
to the original `intitial_routing_sync` feature. That feature was
deprecated because it is expensive and can force a node to use a
disproportionate amount of resources, especially on flapping
connections. Now, the filter is only applied to new gossip messages when
deciding whether or not to forward them to a particular peer.


cfromknecht:
This is an example of the changes proposed in https://github.com/lightningnetwork/lightning-rfc/pull/641

cfromknecht:
we would also be able to remove `channeldb/channel_cache.go` as its primary purpose is to reply to `ChanUpdatesInHorizon`, or possibly start using it to speed up the router.

cfromknecht:
parking this until 0.9 cycle, will be superceded by #3359 in the meantime

Roasbeef:
#3359 has been since merged, so I think this is no longer relevant?

saubyk:
Closing due to inactivity.

PR #3331, opened 2019-07-22
[build, tooling] change linter gometalinter->golangci-lint

halseth:
`gometalinter` has been deprecated: https://github.com/alecthomas/gometalinter/issues/590

Line length check is NOT enabled, since it will fail many places throughout the codebase. It looks like it was never working properly with the old linter.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,18 @@
+run:
+  # timeout for analysis
+  deadline: 4m
+
+linters-settings:
+  govet:
+    # Don't report about shadowed variables
+    check-shadowing: false
+  gofmt:
+    # simplify code: gofmt with `-s` option, true by default
+    simplify: true
+
+linters:
+  disable-all: true
+  enable:
+    - gofmt
+    - golint
+    - govet
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,13 @@ ESCPKG := github.com\/lightningnetwork\/lnd
 
 BTCD_PKG := github.com/btcsuite/btcd
 GOVERALLS_PKG := github.com/mattn/goveralls
-LINT_PKG := gopkg.in/alecthomas/gometalinter.v2
+LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
 GOACC_PKG := github.com/ory/go-acc
 
 GO_BIN := ${GOPATH}/bin
 BTCD_BIN := $(GO_BIN)/btcd
 GOVERALLS_BIN := $(GO_BIN)/goveralls
-LINT_BIN := $(GO_BIN)/gometalinter.v2
+LINT_BIN := $(GO_BIN)/golangci-lint
 GOACC_BIN := $(GO_BIN)/go-acc
 
 BTCD_DIR :=${GOPATH}/src/$(BTCD_PKG)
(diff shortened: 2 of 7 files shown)


wpaulino:
> Line length check is NOT enabled, since it will fail many places throughout the codebase. It looks like it was never working properly with the old linter.

Why don't we begin enforcing it then? The line length should be 80 anyway, but that'll probably still fail in many parts of the codebase.

Makefile
```diff
@@ -72,8 +66,8 @@ $(GOVERALLS_BIN):
 	go get -u $(GOVERALLS_PKG)
 
 $(LINT_BIN):
-	@$(call print, "Fetching gometalinter.v2")
-	GO111MODULE=off go get -u $(LINT_PKG)
+	@$(call print, "Fetching linter")
+	go get -u $(LINT_PKG)
```
wpaulino: The current method of installation ends up producing a `go.mod` diff when running `make lint`. This is probably because `go get` thinks we want to add this as a dependency to the project rather than just install the binary. We can avoid this by prepending `GO111MODULE=off`.

halseth: You're right, good catch!

halseth:
> > Line length check is NOT enabled, since it will fail many places throughout the codebase. It looks like it was never working properly with the old linter.
> 
> Why don't we begin enforcing it then? The line length should be 80 anyway, but that'll probably still fail in many parts of the codebase.

Enforcing 80 on the existing codebase is not really an option, since there are (and probably will still be) many places where more than 80 makes sense (like test data strings, diagrams etc).

`golangci-lint` supports linting only changes happened after a specific commit, so we could try starting to enforce it from the current commit hash 😅

halseth:
Enabled line length checks for new code only. Let's see how that goes 😬

htlcswitch/switch.go
```diff
@@ -1194,11 +1194,7 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 				fail.Reason = circuit.ErrorEncrypter.EncryptMalformedError(
 					fail.Reason,
 				)
-				if err != nil {
```
cfromknecht: 🤔

halseth:
```
htlcswitch/switch.go:1197:12: nilness: impossible condition: nil != nil (govet)
				if err != nil {
```

Roasbeef:
> Enabled line length checks for new code only. Let's see how that goes

It's usually on a best effort basis, there're likely some instances in the codebase where 81 or 82 characters are used.

.golangci.yml
```diff
@@ -0,0 +1,31 @@
...
+  disable-all: true
+  enable:
+    - gofmt
+    - golint
+    - govet
+    - lll
+
+issues:
+  # Show only new issues created after specific git revision. Needed for now
+  # since many line lenght checks would fail.
```
Roasbeef: lenght -> length

.golangci.yml
```diff
@@ -0,0 +1,31 @@
...
+    # tab width in spaces. Default to 1.
+    tab-width: 8
+
+linters:
+  disable-all: true
+  enable:
+    - gofmt
+    - golint
+    - govet
+    - lll
```
Roasbeef: Linters care about lattices these days?

cfromknecht: lolol

halseth:
> As is, this fails travis due to line limit errors. IMO, we can just leave that off, since even after the target commit, the very next PR will need to fix _all_ these errors (many of them legacy), before the build can be green again.

It will only affect lines that are touched by the next PR. But yeah, line length linting hasn't been working up until now, so we don't have to enable it.

halseth:
Removed line length check.

Issue #1253, opened 2018-05-16
routing+channeldb: create new sub-system to keep track of "fitness" of various nodes


Roasbeef:
Throughout the operation of `lnd`, we indirectly and directly interact with various nodes within the network. It may be that we see a new `NodeAnnouncement` for a node, successfully route a payment where the node was a vertex traversed, receive an HTLC routing failure _from_ that node, or need to force close a channel due to a node being inactive. This is all very valuable information that currently isn't being persistently tracked at all. If we start to collate this information persistently, we can begin to use it in several sub-systems within the daemon. For example, within the `autopilot` subsystem, we can use this information to weed out a node as it's been known to be unreliable historically. Another example for usage lies in the interaction with the `missionControl` sub-system: we can preemptively black list any path that includes known historically faulty vertexes. 

# Steps To Completion 
- [ ] Create a new `NodeHistory` sub-bucket within the current `channeldb` package. Starting points for desirable metrics to track include: 
   * first time a node was seen
   * last time a node was seen
   * number of times they were included within an _attempted_ route
   * number of times a route they were involved w/ was successfully settled 
   * number of times a route they were involved w/ was cancelled back (can be more granular: sent error or was source of error) 
   * number of times a route they were involved w/ had to timeout 
   * number of times we had to force close a channel w/ the node due to inactivity 
   * number of times the node incurred a protocol violation (invalid commit or the like)
   * largest number of channels closed in single or consecutive blocks by the node
   * number of attempted breaches
   * number of successful breaches 
   * frequency of channel turn over (are channels opened, then closed only a few blocks after wards?)

- [ ] Due the that nature of each of the fields, we'll likely want to adopt a sort of "columnar" schema within the database. This can be implemented by using `key || nodeID` prefixes. This schema will allow for efficient queries/scans as all the data will be laid out contiguously. Metrics for a particular node can still be obtained by fully populating the two-tuple key. 

- [ ] Start to populate/integrate incrementing counters for the various metrics in the `ChannelRouter` as well as within `missionControl`. For example, each successful/failed payment should trigger an update (or several updates) to the `NodeHistory` database. 

[...]

PR #3332, opened 2019-07-22
Chanfitness: Track peer uptime

carlaKC:
This PR introduces a new package `chanfitness` which is used to track scores for a node's channels. 
Scores are currently kept in memory, with the intention of persisting them once the set of metrics needed to score nodes is more clearly defined.  It is related to #1253, although the issue is quite old.

This change implements tracking for peer uptime, by maintaining a log of channel events for each channel. Peer online/offline events are monitored by a goroutine on a per peer basis, and channel creation/close events are monitored by a single goroutine.

This change also includes adding a channel fitness rpc subserver, but that can be split out into another PR if the change is too big. 

#### PR Checklist
- [x] All changes are Go version 1.12 compliant (used go1.12.6)
- [x] Go fmt, commented and wrapped at 80 lines
- [x] Make check, lint and go vet ok, builds properly!
- [x] Code accompanied by tests


server.go
```diff
@@ -1235,6 +1253,10 @@ func (s *server) Start() error {
 			return
 		}
 
+		if err := s.scoreStore.Start(); err != nil {
+			startErr = err
```
halseth: return?

lnrpc/chanfitrpc/driver.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// createNewSubServer is a helper method that will create the new channel
+// fitness notifier sub server given the main config dispatcher method.
+// If we're unable to find the config that is meant for us in the config
+// dispatcher, then we'll exit with an error.
+func createNewSubServer(configRegistry lnrpc.SubServerConfigDispatcher) (
+	lnrpc.SubServer, lnrpc.MacaroonPerms, error) {
+	//
+	//	// We'll attempt to look up the config that we expect, according to our
```
halseth: double comment

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,226 @@
...
+	"sync"
+	"time"
+
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/channelnotifier"
+	"github.com/lightningnetwork/lnd/lnpeer"
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// ScoreStore maintains a set of scores for the node's channels. It is intended
```
halseth: Instead of the notion of this storing a "score", should we change it to store "events". The score can later be calculated on the basis of the events. We can name this `ChannelEventStore`?

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,226 @@
...
+			quit:      make(chan struct{}, 1),
+			firstSeen: time.Now(),
+		}
+
+		// Since this is the first time this channel has been seen, a
+		// channel opened event is recorded.
+		score.RecordEvent(channelOpenedEvent)
+	}
+
+	if !score.monitoring {
```
halseth: Seems like the goroutine can just be moved into the `!ok` scope, and `monitoring` removed?

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,226 @@
...
+	defer c.Unlock()
+
+	score, ok := c.channels[channelID]
+	if !ok {
+		// The channel is not already known to the scorer,
+		// so must be initialized.
+		score = &ChannelScore{
+			id:        channelID,
+			peer:      peer,
+			quit:      make(chan struct{}, 1),
```
halseth: take a look at how quit channels are defined and used other places in the codebase :)

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,226 @@
...
+	for _, ch := range channels {
+		var peerKey [33]byte
+		copy(peerKey[:], ch.IdentityPub.SerializeCompressed())
+
+		c.OpenChannel(ch.ShortChanID().ToUint64(), peerKey)
+	}
+
+	// channelUpdates provides a subscription to channel events which is used to
+	// listen for the creation of new channels and closing of existing channels.
+	channelUpdates, err := c.cfg.ChannelEvents()
```
halseth: Can this be called inside `MonitorChannelEvents` instead?

chanfitness/chanscore.go
```diff
@@ -0,0 +1,272 @@
...
+	// entirely from memory (as is removing scores would delete all historical
+	// data, which would be wasteful).
+	monitoring bool
+}
+
+// MonitorPeer watches for events when the channel's remote peer goes offline
+// or online and records them in the channel score's event log. This function
+// is terminated when the channel is closed because we are only concerned with
+// peer behaviour across the lifetime of the channel.
+func (s *ChannelScore) MonitorPeer(online notifyOnline, offline notifyOffline) {
```
halseth: Monitoring of peer uptime can be moved to the same level as channel close/opens. Each channel can have just a flat list of events: `[open, online, offline, online, offline, closed]` etc

halseth: Actually, a bit of the complexity now seems to come from the fact that we are only monitoring peers we have channels with. I see no real reason we couldn't just monitor all our peers, since it is a superset of our channel peers. It would simplify the monitoring, since we could track the two independently.

To track all peers (regardless of whether they are a channel peer or not), we could add a new API similar to ChannelEvents, that would notify about peers going online/offline. This API would be useful in other places too.

carlaKC: Not sure if I'm following what you mean by "flat list" of events? 

I'm going to go with storing offline/online on a per peer level as you suggested, so channels don't actually need to store any peer level events, and the EventStore can just get the range of peer events relevant to a channel based on its open/close time. 

In terms of storing an ordered set of peer events, is a []eventStruct{timestamp / type} like I currently have it ok?

chanfitness/chanfitness.go
carlaKC: I split it up to make testing easier, because I can't set the items in a `subscribe.Client` updates queue because it's not exported. 

I originally thought that there might be a case for putting client behind and interface but I see the subscribe server is aware of the update queue structure so it's not not that simple.

carlaKC:
Thanks for the review @halseth 🙏 
Agreed on monitoring all peers rather than peers on a per channel basis, makes a lot more sense. 

What do you think about adding a `peernotifier` package that provides a `subscribe.Client` like `channotifier` currently does? That way, `chanfitness` can subscribe to peer events in the same way that it does channels? 

With that approach, I'd break this (entirely too big - sorry 😅) PR down into:
1. Add `peernotifier` package in new a PR
2. Reduce this PR to just consume channel events for now (+ remove grpc)
3. Add peer monitoring to `chanfitness`
4. Add grpc subserver for querying `chanfitnes`
Sound ok?

chanfitness/chanscore.go
carlaKC: Actually, feels a little weird to have a set of channel events on the `chanEvent` level and then peer level events on the `EventStore` level which need to be combined when we need uptime/flat counts. 

I think I'm going to stick with everything on the channel event level, since there won't be much duplication (ie not expecting so many channels with the same peer that we'd be wasting tons of space).

halseth:
> What do you think about adding a `peernotifier` package that provides a `subscribe.Client` like `channotifier` currently does? That way, `chanfitness` can subscribe to peer events in the same way that it does channels?

👍 

> With that approach, I'd break this (entirely too big - sorry 😅) PR down into:
> 
> 1. Add `peernotifier` package in new a PR
> 2. Reduce this PR to just consume channel events for now (+ remove grpc)
> 3. Add peer monitoring to `chanfitness`
> 4. Add grpc subserver for querying `chanfitnes`
>    Sound ok?

I would start by doing 1 and 2, then we can take a look at how big 2 becomes and what we could reasonably include there 😄

chanfitness/chanscore.go
halseth: > In terms of storing an ordered set of peer events, is a []eventStruct{timestamp / type} like I currently have it ok?

Sounds good.

halseth:
Needs rebase!

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,199 @@
...
+		// been tracking events for it anyway.
+		return
+	}
+
+	eventLog.closeChannel()
+}
+
+// MonitorChannelEvents consumes channel opened and closed events and adds and
+// updates the channel's event log accordingly.
+func (c *ChannelEventStore) MonitorChannelEvents(updates <-chan interface{},
```
halseth: does this method need to be exported?

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,199 @@
...
+			case channelnotifier.OpenChannelEvent:
+				var peerKey [33]byte
+				copy(
+					peerKey[:],
+					event.Channel.IdentityPub.SerializeCompressed(),
+				)
+
+				c.openChannel(
+					event.Channel.ShortChanID().ToUint64(),
+					peerKey,
```
halseth: nit: looks like  `peerKey` can fit on the previous line

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,199 @@
...
+			// store and record a channel closed event.
+			case channelnotifier.ClosedChannelEvent:
+				c.closeChannel(event.CloseSummary.ShortChanID.ToUint64())
+			}
+
+		// receiving on the event store's quit channel indicates that monitoring
+		// should be terminated. To do so, cancel is called to terminate the
+		// subscription to channel updates.
+		case <-c.quit:
+			cancel()
```
halseth: can use `defer cancel()` at the beginning of the method instead.

chanfitness/chanfitness_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/channelnotifier"
+	"github.com/lightningnetwork/lnd/lntest"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+func TestMonitorChannelEvents(t *testing.T) {
```
halseth: needs godoc

chanfitness/chanfitness_test.go
```diff
@@ -0,0 +1,132 @@
...
+			for i, e := range test.expectedEvents {
+				if eventlog.events[i].eventType != e.eventType {
+					t.Fatalf("Expected: event type: %v, got: %v",
+						eventlog.events[i].eventType,
+						e.eventType,
+					)
+				}
+			}
+
+			store.Stop()
```
halseth: Is this needed? Doesn't look like the store is started anywhere.

halseth: Aah, it is to stop the goroutine. An alternative way to test this would be to start the store directly, but mock the `ChannelEvents` by changing its signature to
```go
ChannelEvents func() (<-chan interface{}, func(), error)
```

and then initiating it when creating the store:
```go
	eventStore, err := chanfitness.NewChannelEventStore(&chanfitness.Config{
		ChannelEvents:   func() (<-chan interface{}, func(), error) {
                                                      client, err := s.channelNotifier.SubscribeChannelEvents()
                                                       if err != nil {
                                                                return nil, nil, err
                                                      }
                                                      return client.Updates(), client.Cancel, nil
                                               },
		GetOpenChannels: s.chanDB.FetchAllOpenChannels,
	})
```

When starting the store in the test,  you would just return the `eventQueue` channel from this method. Would that work?

carlaKC:
Thanks for taking a look @joostjager  

> In general I think it is best if it also makes the changes end to end, so that one can verify (run the code) that the goal of the pr is met.

>  Imo that bears the risk that code get merged that needs to be revised later when doing the work in the higher level layers. It may take some creativity to define an end to end pr that is not too big, but it seems that in this new domain there must be some good opportunities.

I definitely agree on having a more horizontal PR in terms of being able to test the changes and figuring out how it will be used by higher levels. My original pass at this PR had a rpcsubserver for getting scores included, but that ended up being a pretty huge PR so I removed it. My thoughts on expanding this PR to be more than just kicking off a logical loop:
1. Re-add a RPC server so the endpoint can be called, although I don't feel like this properly covers what you mean by figuring out how it will be used on higher levels?
2. Add some opt-in channel closing behaviour to autopilot. Since we're only going to have uptime metrics available (see below), I'd say that this would have to be done with some minimum channel lifetime param (eg must be at least x days old) and threshold flap rate which is not acceptable so that we don't start maniacally closing channels. I'll give that a go and see how it looks, might end up being too much for one change. 

> Also, if the goal is to track channel open/closes on chain: it may be that that info is already available and doesn't need to be tracked. The channel point and the closing tx id are persisted in the database. Maybe the peer (tcp) up time metric would have been more suitable to kick off the fitness system.

This PR has been waiting on #3354 (which was originally included in my way too big first attempt) so that uptime can be added to the events store - that is the main goal for this PR. When peer events were split out, I updated this to just have open/close events to get an idea of what the structure would look like. I'm going to add online/offline events now that #3354 is merged - I should have made that clearer, sorry! 

In terms of having open/ close events at all, I think it still makes sense to have them while events are being kept in memory? That way the package doesn't have to touch the DB at all for now (and does not need to query it every time a score is requested). If/when we start looking at persisting score data, we can probably just get channel open/close events from chanDB.

joostjager:
I actually meant creating a vertical PR, cutting through several layers. Ok, horizontal, vertical, it isn't very self-describing, but I think we mean the same thing.

I would at least choose either channel open/close or peer online/offline, not both, for this pr. If you choose peer online/offline, maybe you can just expose a single value "average peer uptime %" over the rpc. Or even more minimalistic, just log the average peer uptime value periodically. Then at least, some business value is created and users can see that it works.

For channel open/close: I don't know if it is a requirement to not touch the database for now. If you think you'll query chandb later, it may not be worth to write the code to send channel events now and remove that again later. In general, with storing redundant information things may get out of sync.

halseth:
I think only doing peer uptime for now sounds like a good middle ground. That was what this PR originally set out to do, and would provide information we don't currently have anywhere.

The question is whether we can get away with channel events completely. Since a peer's uptime might not be very interesting if we don't have an open channel with it, it makes sense to somehow relate it to channel events. This information can _possibly_ be fetched from the DB, but I'm not sure if we have it all (we only have block heights, not timestamps, even though we can fetch these from block headers), and I can imagine this could get more complicated than the current approach, as it is quite self contained. Storing open/close events is also not that big of a deal IMO, as it is not in any way critical if it gets out of sync.

For using this information, we could start with something as simple as returning peer uptime % during a `listchannels` call :)

chanfitness/chanfitness_test.go
carlaKC: This will get a bit messy if we're also going to listen on `client.Quit` as well, signature would have to be:
```
ChannelEvents func() (<-chan interface{}, <-chan struct{}, func(), error)
```

I guess the trade off is between this ugliness and the neatness of just being able to call `store.Start()` on tests. I'm more on the side of a little ugliness for better looking tests, what would you say?

Another option would be to run actual peer/chan notifiers for tests but that may be overkill for tests?  Would need to connect to a channel DB and all sorts.

halseth: I see, yeah that can get messy. Can you just start a `subscribe.Server` for the test where you send the event queue to?

joostjager:
One other question: how does the channel fitness subsystem relate to `lndmon`?

From `lndmon`:

>while lnd provides some information, lndmon by far does the heavy lifting with regards to metrics. With lndmon's data, you can track routing fees over time [..]

This seems to overlap with what is started in this pr. For example iterating over time series to extract useful values.

carlaKC:
> One other question: how does the channel fitness subsystem relate to `lndmon`?

I haven't looked at `lndmon` in detail, but I think that `chanfitness` is a reduced, inhouse version of `lndmon` that we need for querying a specific subset of data? There are some things (like peer uptime) that could be exported to `lndmon` - which actually makes a case for tracking peer online status independent of channels in peernotifier so that we can query uptime over a period (which could be exported to `lndmon` and `chanfitness` . I think that may be overkill for now, but moving the logic in chanfitness to `peerNotifier` at a later time will be pretty trivial. 

> This seems to overlap with what is started in this pr. For example iterating over time series to extract useful values.

In terms of metrics that may be duplicated on both sides, I don't think we need to specifically track an event log for the payment level data (fees, failed/succeeded amounts) so that would minimise the amount of time series processing we do. Uptime is the only "stateful" series that made more sense to track as time series (I did implement a running counter for uptime but it was very messy). If it comes to a point where there's straight up duplication between the two, then we can look at chanfitness exporting metrics for `lndmon`?

joostjager:
We had a bit more discussion on the topic. One thing that was brought up, is the case where tracking of data can trigger certain actions like auto-closing of channels. With `lndmon`, a loop would be created where `lnd` exports to `lndmon`, `lndmon` does time series/analysis/etc and then pushes the result back into `lnd` to take action. This may not be the architecture that we want.

I think the main thing is for us to be aware that the fitness subsystem and `lndmon` are related and make a deliberate decision on where to implement new functionality. 

I am a bit worried though to which extent we can actually foresee what we need. But yes, you are right, the fitness subsystem could in that case export to `lnd` to prevent duplication.

With regard to useful things to track as time series in the fitness subsystem (in the future), the metric 'channel profitability' always comes to my mind. To be able to auto-close non-profitable channels. For channel profitability, an analysis over time of local balance and fwding fees is required.

joostjager:
Tagging @valentinewallace as she's been developing `lndmon`.

carlaKC:
PR updated to add a few changes:
1. Include uptime in `lnrpc/Channel` which is returned by `ListChannels`
2. Record online/offline events on startup for each channel and on open channel events
- When we restart, we need to add a peer online/offline event for existing channels otherwise we will have to wait until they go online/offline to start tracking
- A channel open event is not a guarantee that a peer is online, because they can go offline while the funding tx confirms, so we need to know their starting state

There is an edge case for existing channels where peers are recorded as offline because we have not connected to them at restart time, and then online when we establish a connection, which means existing channels will have 99% uptime rather than 100%. This can be addressed by adding a wait before recording initial peer state for channels (eg sleep one minute then check status) if necessary.

valentinewallace:
>With lndmon, a loop would be created where lnd exports to lndmon, lndmon does time series/analysis/etc and then pushes the result back into lnd to take action. This may not be the architecture that we want.

I agree that sounds like an architecture to avoid :smile: 

>There are some things (like peer uptime) that could be exported to lndmon

That sounds useful. Prometheus may be able to track peer uptime already since it does know when the set of peers changes :thinking: Let me know if I can contribute more lndmon insight.

chanfitness/chanevent.go
```diff
@@ -0,0 +1,179 @@
...
+	// If a close event was recorded before an open event, just return zero to
+	// prevent returning a negative lifespan.
+	if closeTime.Before(openTime) {
+		return 0
+	}
+
+	return closeTime.Sub(openTime)
+}
+
+// getUptime returns total uptime of a channel. Online periods are calculated
```
joostjager: Question from the user perspective: how would a user interpret the up time of an relatively old channel? Suppose it is 5 months old. The first 4 months the up time was 100%, the last month it was offline all the time. This would be reported as 80%, which may sound better than what the fitness of the channel really is.

I think the scope of the pr as it is is fine, but it might be good to give this a thought nevertheless.

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,300 @@
...
+// activity. All elements of the config must be non-nil for the event store to
+// operate.
+type Config struct {
+	// Channel events provides a subscription client which provides a
+	// stream of channel events.
+	ChannelEvents func() (*subscribe.Client, error)
+
+	// PeerEvents provides a subscription client which provides a stream of
+	// peer online/offline events.
+	PeerEvents func() (*subscribe.Client, error)
```
joostjager: If these are untyped anyway, and especially if this is extended to a large list in the future, can't the store not just have a single channel over which it receives all events? And then have all event sources drop their typed events into that channel.

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// PeerOnline return true if a peer is online, and false if it is offline
+	// or unknown to the server.
+	PeerOnline func(peer [33]byte) bool
```
joostjager: If every peer is assumed offline on startup and the peers would all push an event into the channel when they go online on startup, is this second pathway still necessary? Tracking started already when the open channels were listed?

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,300 @@
...
+
+	log.Infof("Adding %v channels to event store", len(channels))
+
+	for _, ch := range channels {
+		var peerKey [33]byte
+		copy(peerKey[:], ch.IdentityPub.SerializeCompressed())
+
+		// Add existing channels to the channel store with an initial peer online
+		// or offline event. Channel opened events are not added for these
+		// channels because they are already open.
```
joostjager: doesn't the open time need to be recorded?

lnrpc/rpc.proto
```diff
@@ -1169,6 +1169,12 @@ message Channel {
     The minimum satoshis the other node is required to reserve in its balance.
     */
     int64 remote_chan_reserve_sat = 21 [json_name = "remote_chan_reserve_sat"];
+
+    /**
+    The percentage of the channel's lifetime that the remote peer has been
+    online.
+    */
+    int64 uptime_percentage = 22 [json_name = "uptime_percentage"];
```
joostjager: Maybe we can also add the channel life time here, as we have it available anyway. And I think we should also mark the fields as 'experimental' in the comment, so that we can make breaking changes later if we want to.

chanfitness/chanfitness.go
```diff
@@ -0,0 +1,300 @@
...
+
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/channelnotifier"
+	"github.com/lightningnetwork/lnd/peernotifier"
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// ChannelEventStore maintains a set of event logs for the node's channels.
+// It is intended to provide insight into the performance of channels.
+type ChannelEventStore struct {
```
joostjager: Main type in this file doesn't match file name

chanfitness/chanevent.go
```diff
@@ -0,0 +1,179 @@
+package chanfitness
+
+import (
+	"time"
+)
+
+type eventType int
+
+const (
```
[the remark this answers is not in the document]
carlaKC: I've also been thinking about that, I think some kind of `stopMonitoring` event that needs to be added on shutdown. A bit annoying to have to record for each channel, so we could look at having some on-shutdown ChannelEventStore level event but that discussion can wait.

chanfitness/chanevent.go
```diff
@@ -0,0 +1,179 @@
...
+// chanEventLog stores all events that have occurred over a channel's lifetime.
+type chanEventLog struct {
+	// id is the uint64 of the short channel ID.
+	id uint64
+
+	// peer is the compressed public key of the peer being monitored.
+	peer [33]byte
+
+	// events is a log of timestamped events observed for the channel.
+	events []*channelEvent
```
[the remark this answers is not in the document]
carlaKC: That makes sense, I went for open/close events because I wanted to distinguish between channels that already exist at startup and newly opened ones but I don't think that actually matters? We have an observed lifetime for a channel and observed uptime, open/already existing is external to that.

chanfitness/chanevent.go
carlaKC: That could be addressed by adding a `startTime`/`endTime` param to the uptime calculation? 
For total uptime you just pass the open time/now and for more specific queries you can provide time spans within the channel's lifetime.

The user would have to get lifetime, then query specific uptime 'buckets' which is a bit roundabout, but a specialised endpoint which provides uptime per some duration chunk could be added? Eg give me uptime % per hour/day.

chanfitness/chanevent.go
```diff
@@ -0,0 +1,179 @@
...
+		return 0
+	}
+
+	return closeTime.Sub(openTime)
+}
+
+// getUptime returns total uptime of a channel. Online periods are calculated
+// as the time between the node reporting as online, then subsequently
+// being reported offline. Any out of state events, such as two adjacent online
+// events are ignored, and the earlier timestamp is used for uptime calculation.
```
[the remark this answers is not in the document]
carlaKC: They shouldn't be added to the log at all, so this really only covers the case where the peer subscription has a bug and sends two online events. I can add a warning log? Not sure what the go-to is for when lnd gets something unexpected like this?

chanfitness/chanfitness.go
carlaKC: So we'd construct a function which pipes all of the subscriptions into a channel which we create in `server.go` and then pass that to `Config`? I think that constructing that kind of aggregation from the caller bubbles complexity up rather than down sometimes, because if we have a lot of event sources it's simpler to just provide a list of functions than have some long aggregations of them all in the middle of the `newServer` call? 

Having had my philosophical ramble, I think this may address some of the concerns you have about the unit tests so will give this a try :)

chanfitness/chanfitness.go
carlaKC: This is necessary for new channels that are opened after startup. We don't know whether the peer is online when the channel confirms, and would have ignored the most recent peer event before we had a channel open with them so need some way of figuring out their starting state. We could technically infer a state if they have a single online/offline event while we're monitoring, but if they're online or offline for our whole monitoring period we don't know which state they're in.

chanfitness/chanfitness.go
carlaKC: Open time is currently just taken as the lowest timestamp in the event log, but I'm going to save open/close in the eventlog itself as you suggested so this will change.

lnrpc/rpc.proto
carlaKC: Sounds good, it also makes it less likely that the uptime is misinterpreted by users as applying to the whole lifetime of the channel. 

How about:
`observed_lifetime`: total seconds we have been monitoring the channel
`observed_uptime`: total seconds we have observed the peer online during monitoring

Then users can do their own calculations, and there's no rounding to be dealt with. There's probably a better var name than `observed` because that makes is sound like we didn't know about the channel before that, but I'll scour synonyms.com later :)

chanfitness/chanevent.go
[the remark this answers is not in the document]
joostjager: I don't think it matters, because it doesn't show in the reported value in any way.

chanfitness/chanevent.go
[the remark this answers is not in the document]
joostjager: That is actually a more general discussion that in a way also applies to this pr's scope. If there are multiple channels to the same peer, the peer activity is logged in each of those while there could also be one log per peer. Those logs would then be recombined when fitness data is requested. But I think we can start by storing it all on a per channel basis as you do now.

A `stopMonitoring` event may not be enough, because `lnd` can also be killed. When it then comes back up, there is no way to find out how long it was offline.

chanfitness/chanevent.go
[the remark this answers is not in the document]
joostjager: If it can only be caused by a bug, normally we'd return an error to the caller. To handle it at the origin. But in this case we are building a reporting system which doesn't directly influence the rest of the system. And also there is no return path available yet. I think logging it as error (in case this really should never happen) is the best option.

chanfitness/chanevent.go
joostjager: Yes, `startTime`/`endTime` sounds good.

Maybe eventually a single call can develop that returns the channel fitness within a given time frame. One of the fitness parameters that is returned, is the channel open duration within the time frame, so that the percentage calculation can still be made correctly.

And then indeed `listChannels` could show three sets of fitness values for let's say 'last day', 'last week' and 'all'.

chanfitness/chanfitness.go
joostjager: I was actually thinking something simpler. Just make the fitness system a dependencies of the other subsystems. It exposes a single function `SendEvent` to those subsystems. In the implementation of `SendEvent`, the only thing that happens is that the (untyped) event is dropped into a single buffered channel or queue. The main loop of the fitness system is receiving on that channel and processing the events via a type switch.

But I could be missing something here of course.

chanfitness/chanfitness.go
joostjager: If you want to keep the interface with the fitness system as narrow as possible (which I think is good), you could just have an in-memory map on the event store level that stores the peers that are online. Do the look up on that locally inside the fitness system. If the peer is not in the map, it is offline. If the peer is online for the whole monitoring period, the peer would still have send an online event on startup to the fitness system?

lnrpc/rpc.proto
joostjager: Sounds good. Sounds better. I don't think the `observed` prefix is needed, is it?

chanfitness/chanfitness.go
carlaKC: Ok will do. I was thinking that current peer state (and perhaps a history of online/offline events to get around potential duplication from per channel tracking with the same peer) could be added to `PeerNotifier` if there's demand for something like that in more than one place. But keeping it in `chanfitness` in the meantime works.

chanfitness/chanfitness.go
carlaKC: My main concern with a `SendEvent` call is that it feels a bit counterintuitive to be sending events when we have two subscriptions specifically intended for receiving events. 

In the case where we don't have subscriptions, a `SendEvent` makes sense. And I suspect this will be the case for the remainder of the events we add to channel scores (ie htlc level events). 

Since it probably (?) doesn't make sense to make subscriptions for htlc level events the options would be:
1. Use `SendEvent` everywhere, accepting that it's a bit of an off construction to send events in the cases where you have subscriptions available
2. Use `SendEvent` for non-subscription events and still pass the Peer/Chan Subscriptions in, but pipe those into the SendEvent channel so that monitoring happens in one central place.

lnrpc/rpc.proto
carlaKC: How should I mark fields as experimental? I can't see any examples in the code. 
Thinking of something along these lines in the field comments?
```
    [experimental]: this field is experimental and may be subject to change.
```

chanfitness/chanfitness.go
joostjager: Isn't it that these subscriptions only exist to serve the event store? In `server.go`, the call `s.peerNotifier.NotifyPeerOnline(pubKey)` is made, which could be replaced by a call directly into the event store `SendEvent`?

I am looking for the simplest thing that works, but may miss something because I didn't follow the design decision around peer notifier.

carlaKC: `PeerNotifier` does exist to serve the event store, so that could be replaced. but `ChannelNotifier` is more general. I thought that it would be better to have a central subscription queuing up events, rather than having calls out to `chanFitness` in the middle of other functions, but it would really only be out of place in `channelNotifier.NotifyOpen/ClosedChannelEvent` which I guess is ok.

joostjager: In the implementation of that function `chanFitness.SendEvent`, the only thing that would happen is queue the event in a non-blocking way. 

I don't want to give the impression that I know what the best design is. Just wanted to bring up this alternative, because I was wondering how the current approach would scale. If for many types of events a separate connecting component is required that basically just inverts the control flow, it may create too much boilerplate code compared to sending the event to `chanFitness` directly.

carlaKC: Of course, I'm glad you brought it up because I hadn't really thought what this will look like for events which don't have subscriptions available. You're right about not wanting to add large amounts of code just to subscribe to different kinds of events events. And unit testing with subscriptions has been tricky (which often means the design isn't quite right, I find) so I'm going to try with a SendEvent structure and see how that looks :)

Roasbeef: The amount of code needed to subscribe an event is pretty minor. We created the entire `subscribe` package in the first place in order to consolidate all the pub/sub (and infinite queue) duplication we had strewn around the codebase (chain notifier, etc). It also takes care of a number of details around client tracking, event cancellation, etc. All that would need to be replicated again vs just using an existing package that is meant to serve as a building block for sub-systems like this one. The model is essentially one of a messaging hub where any sub-system can listen on an even stream with minimal overhead.

joostjager: So the current design (taking peer notification as an example) is:
* `server` calls `peernotifier.NotifyPeerOnline`
* `peernotifier` pushes update into a channel to `subscribe.Server`
* `subscribe.Server` pushes update into a channel to `chanfitness`

What is the advantage of this over:
* `server` calls `chanfitness.NotifyPeerOnline`
* `chanfitness` internally manages a queue to not block reporting sub-systems. There only needs to be a single queue for all events.

I only see an advantage of having separate `*-notifier` objects if there are multiple listeners. Those can be added when a need arises. From the sub-system perspective the only required change would then be to call the `notifier` instead of `chanfitness`, which is just a replacement without any other design implications for that sub-system.

I think we should be careful not to over-engineer this.

lnrpc/rpc.proto
joostjager: We indeed don't have that yet. All our experimental fields are in sub-servers atm. Yes, something along those lines.

chanfitness/chanfitness.go
carlaKC: In my mind there are three sources that the scoring system will need events from channels, peers and HTLCs. And justifying using a subscribe server requires that we have more than one listener per subscription, otherwise we're just writing boilerplate as you say. 

1. `ChanNotifier`: already used elsewhere, so adding `SendEvent` call would be a duplication
2. `PeerNotifier`: created for this PR, only other use at present is potentially a rpc subscription
3. HTLC events: no subscription at present, although @Roasbeef has recently opened #3420

I am reluctant to start off with an event queue if it is inevitably going to be replaced by subscriptions, so I think the questions that need answering are:
1. Are there use cases for `PeerNotifier` and a potential `HTLCNotifier` and is a user exposed RPC call a good enough justification for the existence of a subscription? 
2. How likely is it that we extend the scoring system beyond these three sources? Because if there's anything else would reasonably want to add that wouldn't suit a subscription that's an argument for an events queue 

I'm not familiar enough with the repo to make a call about whether subscriptions for these things are required. I would like to move forward with this PR, so if I can get some clarity on those questions I think it will help determine what direction to move in?

carlaKC: Although an events chan is looking a lot cleaner for tests. The only ugliness is that it has to be called from the channel open/close notification functions which feels like doing the same thing twice. So it really is dependent on whether subscriptions actually have other use cases. which I'm unsure of.

Roasbeef: > chanfitness internally manages a queue to not block reporting sub-systems. There only needs to be a single queue for all events.


It would more or less need to re-implement the current queue handling in the `subscribe` package to do this. Therefore we should re-use what we have rather than attempt to over optimize for this use case which isn't performance critical and still nascent.

Roasbeef: I don't think we need to get caught up in this relatively minor difference. The original primary reviewer (johan) was OK with the current approach. We shouldn't undo that prior lineage due to a new reviewer stepping in. For the sake of making progress, I don't think there's a fundamental principle that tends us in one direction or the other. Therefore we can continue down the current path which had already received review by the former primary reviewer. 

If this was more critical code such as the channel state machine or revocation handling, then I would be willing to spend more time here, but it's a relatively independent sub-system that is non-critical.

joostjager: Even though Johan was ok with the current approach, it could still be that he likes the alternative more. I don't think we've discussed that. But yes, both approaches lead to functioning, correct code. This is only about preventing boiler plate. I can put aside my concerns about that and see how it works out.

chanfitness/chanfitness.go
halseth: +1 on renaming the file

lnrpc/rpc.proto
halseth: IMO we don't have to mark them experimental. We can rather deprecate and add new fields later if we need to.

chanfitness/chanfitness.go
halseth: I think the current approach is more general and would scale better with more subscription types, and more subscribers to the different types. 

We had to take a similar decision when we initially added the `NotifyPeerOnline` API to the `fundingmanager`. We could have made the server call into the funding manager directly, but rather we made it a general subscription API that any subsystem could call. This has turned out useful to have access to also in other packages.

I also don't see the boiler plate we would avoid, since it is mostly contained in the `subscribe`  package?

chanfitness/chanfitness.go
halseth: I would keep the package name, only change the file name :)

halseth:
> Thinking ahead, I'm wondering whether it would be better to have a `channelEvent` interface in `chanevent.go`

That indeed sounds like a good idea IMO, but that change can possibly be done when needed? :)

Roasbeef:
OP ping!

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,308 @@
...
+		defer c.wg.Done()
+
+		// Cancel subscriptions on exit.
+		defer channelClient.Cancel()
+		defer peerClient.Cancel()
+
+		for {
+			select {
+			case e := <-channelClient.Updates():
+				events <- e
```
halseth: should select on quit also

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,308 @@
...
+
+			case e := <-peerClient.Updates():
+				events <- e
+
+			// Close the events channel and exit if one of the subscriptions
+			// fails, or the store receives the instruction to quit.
+			case <-c.quit:
```
halseth: must `fallthrough` here

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,308 @@
...
+
+	log.Infof("Adding %v channels to event store", len(channels))
+
+	for _, ch := range channels {
+		var peerKey [33]byte
+		copy(peerKey[:], ch.IdentityPub.SerializeCompressed())
+
+		// Add existing channels to the channel store with an initial peer
+		// online or offline event.
+		c.addChannel(ch.ShortChanID().ToUint64(), peerKey)
```
halseth: Should this be done before the goroutines are started?

chanfitness/chanevent.go
```diff
@@ -0,0 +1,150 @@
...
+	openedAt time.Time
+
+	// closedAt is the time that the channel was closed.
+	closedAt time.Time
+}
+
+// add appends an event with the given type and current time to the event log.
+// If the channel is already closed, it returns early without adding an event.
+func (e *chanEventLog) add(eventType eventType) {
+	if !e.closedAt.IsZero() {
```
halseth: can add the `// If the channel is already closed, it returns early without adding an event.` also here

chanfitness/chanevent.go
```diff
@@ -0,0 +1,150 @@
...
+
+// uptime returns total uptime of a channel over the period of time specified
+// by startTime and endTime. Uptime is calculated as the period between an online
+// and offline event. Event logs which end on an online event additionally have
+// the period between the online event and the channel close or end of period
+// added to account for the fact that the peer remained online until that point.
+// If a peer is online at the time a chanEventLog is created, an online event is
+// expected to be added to the log to determine start state.
+func (e *chanEventLog) uptime(startTime, endTime time.Time) time.Duration {
+	var uptime time.Duration
```
halseth: nit: can move this down with the other `var`s

chanfitness/chanevent_test.go
```diff
@@ -0,0 +1,189 @@
...
+		// expectedUptime is the amount of uptime we expect to be calculated
+		// over the period specified by startTime and endTime.
+		expectedUptime time.Duration
+	}{
+		{
+			name:      "No events",
+			openedAt:  fourHoursAgo,
+			closedAt:  now,
+			startTime: fourHoursAgo,
+			endTime:   now,
```
halseth: set expected time=0?

chanfitness/chanevent_test.go
```diff
@@ -0,0 +1,189 @@
...
+			openedAt: fourHoursAgo,
+			events: []*channelEvent{
+				{
+					timestamp: twoHoursAgo,
+					eventType: peerOnlineEvent,
+				},
+			},
+			startTime:      threeHoursAgo,
+			endTime:        oneHourAgo,
+			expectedUptime: time.Hour,
```
halseth: adda test online->offline->online->offline a few times?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,308 @@
...
+func (c *ChannelEventStore) peerEvent(peer [33]byte, event eventType) {
+	c.Lock()
+	defer c.Unlock()
+
+	// Track current online status of peers in the channelEventStore.
+	peerIsOnline := true
+	if event == peerOfflineEvent {
+		peerIsOnline = false
+	}
+	c.peers[peer] = peerIsOnline
```
halseth: is the assumption that we will start listening for these peer events before lnd connects to any peers?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,308 @@
...
+
+// GetUptime returns the uptime of a channel over a period and a boolean to
+// indicate whether it was found in the channel event store. If a zero start or
+// end time is given, the function returns zero and false to indicate that
+// uptime was not calculated.
+func (c *ChannelEventStore) GetUptime(chanID uint64, startTime,
+	endTime time.Time) (time.Duration, bool) {
+
+	// Return early if the period specified has a zero start or end time.
+	if startTime.IsZero() || endTime.IsZero() {
```
halseth: shouldn't `startTime` be allowed set to zero to indicate start as early as possible?

chanfitness/chanevent.go
```diff
@@ -0,0 +1,150 @@
...
+			lastOnline = event.timestamp
+			wasOnline = true
+
+		case peerOfflineEvent:
+			wasOnline = false
+		}
+
+		// If our previous online time is before the period we're calculating
+		// uptime for, it must be progressed to the start time of the period.
+		if lastOnline.Before(startTime) {
```
halseth: To make it easier to follow this logic, I think this can be moved to the `peerOnlineEvent` case?

chanfitness/chanevent.go
```diff
@@ -0,0 +1,150 @@
...
+			lastOnline = startTime
+		}
+
+		// If event is before the period we are calculating uptime for, or we
+		// are currently in an online state, we do not need to increment uptime.
+		if event.timestamp.Before(startTime) || wasOnline {
```
halseth: similarly, can remove `wasOnline`, and move the continue to the `peerOnlineEvent` case

chanfitness/chanevent.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+
+		// The eventLog has recorded an offline event, so we add the previous
+		// online period to uptime.
+		uptime += event.timestamp.Sub(lastOnline)
```
halseth: same, Since this will be only in the offline case, I think it might be easier to follow if we move it to the offline case statement

chanfitness/chaneventstore.go
carlaKC: Which quit, got both subscriptions and the store itself below?

chanfitness/chaneventstore.go
carlaKC: T(I think)IL that you can't `fallthrough` a select?
Will defer close and just return for each case :)

chanfitness/chaneventstore.go
carlaKC: There's a chance that we miss a channel if this happens before we start monitoring
1. Add all channels from DB
2. Channel is opened and notification triggers
3. We start subscribing to events and have missed the channel open

Any reason for having it above that I'm missing? I also had it above to start because it looked better until Joost pointed this ^ out to me

chanfitness/chaneventstore.go
carlaKC: Yes. If I'm following properly, all subservers are kicked off in `server.go` before we connect to peers `server.go#1333`.

This does create an issue where channels that we have open on startup will record their peers being offline for the period that it takes to connect. I thought about adding some kind of sleep but I think the various impls have quite different connection times?

chanfitness/chaneventstore.go
carlaKC: It should! 

I'm still going to err on a 0 `endtime` though? It's easy enough to pass `now()` in for endtime.

chanfitness/chaneventstore.go
halseth: `c.quit` should be enough.

chanfitness/chaneventstore.go
halseth: > T(I think)IL that you can't `fallthrough` a select?

True! 😅 

> Will defer close and just return for each case :)

👍

chanfitness/chaneventstore.go
halseth: As long as the subscriptions is made before adding channels from the DB i think it should be fine. So order will be:
1. c.cfg.ChannelEvents() / PeerEvents
2. Add channels from db
3. start go routines

I can think of two good reasons to doing it in this order:
1. if `GetOpenChannels` returns an error, then we don't have to quit the goroutines before returning.
2. By adding the db channels before starting the go routines, we ensure we know about them before handling events.

chanfitness/chaneventstore.go
halseth: Sounds good, I think this is good enough 👍

chanfitness/chaneventstore.go
halseth: Do we even need to support custom `endtime` (and `startime`?)? Would it simplify things if `Uptime` just returns the uptime between `channel open` and `min(channelclose, now)` ?

chanfitness/chaneventstore.go
carlaKC: ah got it :) that's a better order

chanfitness/chaneventstore.go
carlaKC: I think that it may be nice to have for longer running channels, because we may care more about recent uptime periods?

Could update this func to not take params so it's simpler for callers (which I like) but leave the functionality there in case we want it?

halseth: Ok good point, can keep `startTime` and `endtime` options I think :)

chanfitness/chaneventstore_test.go
```diff
@@ -0,0 +1,421 @@
...
+			// Create an events channel with sufficient buffer and add all the
+			// test's events.
+			events := make(chan interface{}, len(test.events))
+			for _, e := range test.events {
+				events <- e
+			}
+			close(events)
+
+			store.wg.Add(1)
+			store.monitorChannelEvents(events)
```
joostjager: This looks like it is using a lot if implementation knowledge of the store, making this test expensive to maintain. It needs to change if the internals of the store change. It is preferable to test it more as a black box.

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,317 @@
...
+// indicate whether it was found in the channel event store. A zero start time
+// will be interpreted as wanting the earliest timestamp applicable to the
+// channel. A zero end time will return zero and false to indicate that uptime
+// was not calculated.
+func (c *ChannelEventStore) GetUptime(chanID uint64, startTime,
+	endTime time.Time) (time.Duration, bool) {
+
+	// Return early if the period specified has a zero end time.
+	if endTime.IsZero() {
+		return 0, false
```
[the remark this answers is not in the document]
carlaKC: It should, since we're just using it for listchannels, will update 👍

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,317 @@
...
+	ChannelEvents func() (*subscribe.Client, error)
+
+	// PeerEvents provides a subscription client which provides a stream of
+	// peer online/offline events.
+	PeerEvents func() (*subscribe.Client, error)
+
+	// Get open channels provides a list of existing open channels
+	// which is used to populate the ChannelEventStore with a set of channels
+	// on startup.
+	GetOpenChannels func() ([]*channeldb.OpenChannel, error)
```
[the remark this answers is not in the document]
carlaKC: So get rid of having to pass a list of open channels to the store at all?
Or find a way to pipe them through the same logic without necessarily having to trigger an event somehow?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+	}
+}
+
+// GetLifespan returns the opening and closing time observed for a channel and
+// a boolean to indicate whether the channel is known the the event store. If the
+// channel is still open, a zero close time is returned.
+func (c *ChannelEventStore) GetLifespan(chanID uint64) (time.Time, time.Time, bool) {
+	c.Lock()
+	defer c.Unlock()
```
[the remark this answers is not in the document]
carlaKC: Not sure if I'm following, do you mean something of this sort?
```
monitorChanEvents
    case req <- uptimeRequest
           {calculate uptime}
           req.ResponseChannel <- uptime
```
Feels a bit weird to be serving requests in the processing loop, but I'm not clear on the idea. Could you point me to an example?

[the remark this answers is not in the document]
joostjager: Yes, that's it indeed. An example is in `UtxoSweeper.collector` when we get the pending sweeps.

chanfitness/chaneventstore.go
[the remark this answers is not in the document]
halseth: It could be done, but that would break the 1:1 relationship to `SubscribeChanEvents` from `channelnotifier`, so I think the current way it is done is alright.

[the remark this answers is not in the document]
joostjager: Discussed offline. Conclusion is that the `ChannelNotifier` interface isn't optimal. You can subscribe to events, but there is no way to pass in a reference point from where you want to get those events. For example, in `RegisterBlockEpoch`, you can pass in a best block. This makes it clear what to expect. If `ChannelNotifier` would have had such a point (for example blockheight from which you receive channel events), the channel fitness subsystem could just pass in height 0 and get an initial sync via the same subscription client.

You could consider adding a TODO to `ChannelNotifier.SubscribeChannelEvents` to document the shortcoming. At least to prevent future devs from, without being fully aware, adding another `ListChannels` and event de-duplication in the next subsystem that uses channel notifier.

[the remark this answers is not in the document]
halseth: +1 to keeping the current design with an additional TODO 👍

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// GetUptime returns the uptime of a channel over a period and an error if the
+// channel cannot be found or the uptime calculation fails.
+func (c *ChannelEventStore) GetUptime(chanID uint64, startTime,
+	endTime time.Time) (time.Duration, error) {
+
+	response := make(chan uptimeResponse)
+
+	c.uptimeRequests <- uptimeRequest{
```
halseth: must select on quit

chanfitness/chanevent.go
```diff
@@ -0,0 +1,149 @@
...
+		uptime     time.Duration
+		lastOnline time.Time
+		wasOnline  bool
+	)
+
+	// Iterate through all the events in the log until we hit one that is after
+	// the end of the period we are calculating uptime for. Events before the
+	// period startTime need to be processed so we can determine the online/
+	// offline state of the peer at the start of the relevant period.
+	for _, event := range e.events {
```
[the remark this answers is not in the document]
carlaKC: I think that separating out has increased readability a bit, but unfortunately needing to track the state before the relevant period and after (in the case of ending on an online event) still makes it a bit edge-casey.

chanfitness/chaneventstore.go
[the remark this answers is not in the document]
carlaKC: TODO added to `channotifier`.

chanfitness/chanevent.go
```diff
@@ -0,0 +1,122 @@
...
+	timestamp time.Time
+	eventType eventType
+}
+
+// chanEventLog stores all events that have occurred over a channel's lifetime.
+type chanEventLog struct {
+	// id is the uInt64 of the short channel ID.
+	id uint64
+
+	peer [33]byte
```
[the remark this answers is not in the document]
joostjager: No `route.Vertex`?

chanfitness/chanevent.go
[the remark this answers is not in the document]
joostjager: For me as a non-author, it is much easier to verify that the logic is correct. I think it is an improvement.

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+// closeChannel records a closed time for a channel, and returns early is the
+// channel is not known to the event store.
+func (c *ChannelEventStore) closeChannel(channelID uint64) {
+	eventLog, ok := c.channels[channelID]
+	if !ok {
+		// Return early the channel is not known, since we have not been
+		// tracking events for it anyway.
+		return
```
joostjager: Log error, can it ever happen?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+
+// addChannel adds a new channel to the ChannelEventStore's map of channels with
+// an initial peer online state (if the peer is online). If the channel is
+// already present in the map, the function returns early. This function should
+// be called to add existing channels on startup and when open channel events
+// are observed.
+func (c *ChannelEventStore) addChannel(channelID uint64, peer [33]byte) {
+	_, ok := c.channels[channelID]
+	if ok {
+		return
```
joostjager: Can this happen? Log something, or explain how it can happen in a comment?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	eventLog.close()
+}
+
+// peerEvent adds a peer online or offline event to all channels we currently
+// have open with a peer.
+func (c *ChannelEventStore) peerEvent(peer [33]byte, event eventType) {
+	// Track current online status of peers in the channelEventStore.
+	peerIsOnline := true
```
joostjager: `peerIsOnline := event == peerOfflineEvent`?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+	wg sync.WaitGroup
+}
+
+// Config provides the event store with functions required to monitor channel
+// activity. All elements of the config must be non-nil for the event store to
+// operate.
+type Config struct {
+	// Channel events provides a subscription client which provides a
+	// stream of channel events.
+	ChannelEvents func() (*subscribe.Client, error)
```
joostjager: Rename this to `SubscribeChannelEvent`, to make it clear that every invocation creates a new client?

chanfitness/chaneventstore.go
joostjager: I think the subscription to channel/peer events is processed asynchronously. The call to subscribe returning, doesn't seem to guarantuee that it has happened.

rpcserver.go
```diff
@@ -2663,12 +2665,35 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
...
+	// zero values will be returned (which wil yield a zero lifetime).
+	startTime, endTime, err := r.server.chanEventStore.GetLifespan(chanID)
+	if err != nil {
+		log.Warnf("Could not get channel lifespan: %v", err)
+	}
+
+	// If the channel has not been closed yet, and it was found in the channel
+	// store set its endTime to now to calculate lifetime and uptime until the
+	// present.
+	if endTime.IsZero() && err != nil {
```
joostjager: Match on a specific error here?

rpcserver.go
```diff
@@ -2663,12 +2665,35 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
 	}
 	externalCommitFee := dbChannel.Capacity - sumOutputs
 
+	chanID := dbChannel.ShortChannelID.ToUint64()
+
+	// Get the lifespan observed by the channel event store. It it is unknown,
+	// zero values will be returned (which wil yield a zero lifetime).
```
joostjager: If the channel is unknown, shouldn't we then stop making additional calls?

channelnotifier/channelnotifier.go
```diff
@@ -73,7 +73,10 @@ func (c *ChannelNotifier) Stop() {
 }
 
 // SubscribeChannelEvents returns a subscribe.Client that will receive updates
-// any time the Server is made aware of a new event.
+// any time the Server is made aware of a new event. The subscription provides
+// channel events from the point of subscription onwards. TODO: update  to allow
+// subscriptions to specify a block height from which we would like to subscribe
+// to events.
```
joostjager: I think, but not sure if we really do this always, that we make the TODO stand out a bit more by adding an empty comment line above it.

chanfitness/chaneventstore.go
joostjager: Ok, my mistake. It is happening synchronously. Nvm

chanfitness/chaneventstore_test.go
joostjager: I think the solution to prevent peeking inside the event lists (implementation specific), would be to create a mock event log. But I am ok leaving it as is.

chanfitness/chanevent.go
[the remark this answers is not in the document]
halseth: I don't think we need to add a dependency on `routing`, so would prefer keeping as is.

chanfitness/chanevent.go
```diff
@@ -0,0 +1,214 @@
...
+// close sets the closing time for an event log.
+func (e *chanEventLog) close() {
+	e.closedAt = e.now()
+}
+
+// add appends an event with the given type and current time to the event log.
+// The open time for the eventLog will be set to the event's timestamp if it is
+// not set yet.
+func (e *chanEventLog) add(eventType eventType) {
+	// If the channel is already closed, it returns early without adding an event.
```
halseth: nit: `s/it returns early/return early`

chanfitness/chanevent.go
```diff
@@ -0,0 +1,214 @@
...
+type onlinePeriod struct {
+	start, end time.Time
+}
+
+// getOnlinePeriods returns a list of all the periods that the event log has
+// recorded the remote peer as being online. In the unexpected case where there
+// are no events, the function returns early. Online periods are defined as a
+// peer online event which is terminated by a peer offline event. This function
+// expects the event log to be ordered by ascending timestamp.
+func (e *chanEventLog) getOnlinePeriods() []*onlinePeriod {
```
halseth: Nice separation of responsibility into this method!

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+
+	// PeerEvents provides a subscription client which provides a stream of
+	// peer online/offline events.
+	PeerEvents func() (*subscribe.Client, error)
```
halseth: same: `SubscribePeerEvents`?

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+	// Create a subscription to channel events.
+	channelClient, err := c.cfg.ChannelEvents()
+	if err != nil {
+		return err
+	}
+
+	// Create a subscription to peer events.
+	peerClient, err := c.cfg.PeerEvents()
+	if err != nil {
+		return err
```
halseth: should cancel `channelClient` before returning here. Same with peerClient in error case below.

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+		// Exit if the store receives the signal to shutdown.
+		case <-c.quit:
+			return
+		}
+	}
+}
+
+// errShuttingDown is returned when the store cannot respond to a query because
+// it has received the shutdown signal.
+var errShuttingDown = errors.New("channel event store shutting down")
```
halseth: style nit: usually we declare errors at the top

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+		channelID:    chanID,
+		responseChan: make(chan lifespanResponse),
+	}
+
+	// Send a request for the channel's lifespan to the main event loop, or
+	// return early with an error if the store has already received a shutdown
+	// signal.
+	select {
+	case c.lifespanRequests <- request:
+
```
halseth: style nit: remove newline

chanfitness/chaneventstore.go
```diff
@@ -0,0 +1,381 @@
...
+		endTime:      endTime,
+		responseChan: make(chan uptimeResponse),
+	}
+
+	// Send a request for the channel's uptime to the main event loop, or
+	// return early with an error if the store has already received a shutdown
+	// signal.
+	select {
+	case c.uptimeRequests <- request:
+
```
halseth: style nit: remove newline

chanfitness/chaneventstore_test.go
```diff
@@ -0,0 +1,460 @@
...
+			expectStartErr: true,
+		},
+		{
+			name:           "Peer events fail",
+			ChannelEvents:  okSubscribeFunc,
+			PeerEvents:     errSubscribeFunc,
+			expectStartErr: true,
+		},
+		{
+			name:          "Get oepn channels fails",
```
halseth: nit: oepn

chanfitness/chaneventstore_test.go
```diff
@@ -0,0 +1,460 @@
...
+			ChannelEvents: okSubscribeFunc,
+			PeerEvents:    okSubscribeFunc,
+			GetChannels: func() (channels []*channeldb.OpenChannel, e error) {
+				return nil, errors.New("intentional test err")
+			},
+			expectStartErr: true,
```
halseth: add a test case for no error?

chanfitness/chaneventstore_test.go
```diff
@@ -0,0 +1,460 @@
...
+			if open != test.opened {
+				t.Errorf("Expected: %v, got %v", test.opened, open)
+			}
+
+			if close != test.closed {
+				t.Errorf("Expected: %v, got %v", test.closed, close)
+			}
+
+			// Stop the store's go routine.
+			store.Stop()
```
halseth: can defer stop this?

channelnotifier/channelnotifier.go
halseth: Yes, add it on a separate line, and with your handle, like this:
`TODO(carlaKC): update to....`

rpcserver.go
halseth: I think it will be handled correctly as it currently is?

rpcserver.go
```diff
@@ -2685,6 +2710,8 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
 		LocalChanReserveSat:   int64(dbChannel.LocalChanCfg.ChanReserve),
 		RemoteChanReserveSat:  int64(dbChannel.RemoteChanCfg.ChanReserve),
 		StaticRemoteKey:       dbChannel.ChanType.IsTweakless(),
+		Lifetime:              int64(endTime.Sub(startTime)),
```
halseth: Won't this result in `Nanoseconds`?

chanfitness/chanevent.go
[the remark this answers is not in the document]
carlaKC: Can we leave it as is for now and then I'll update `PeerNotifier` and this code in a separate PR?

chanfitness/chaneventstore.go
carlaKC: This would only happen in the case of a bug, but I thought it was worth adding the check in case. 
I'll add a log and a comment to explain? It can also be removed, since the duplicate would just be overwritten but I don't think there's much harm leaving the check in?

chanfitness/chaneventstore.go
carlaKC: As above, this should never happen.

chanfitness/chaneventstore_test.go
carlaKC: If we don't error, the start func will try to touch the subscribe clients `updates.ChanOut`, since it's unexported it can't be mocked. Will remove that bool since and update the comment to note that.

chanfitness/chaneventstore.go
joostjager: Just `return` while continuing would work too doesn't sound very useful. I can imagine that this sub-system is sensitive to timing issues / race conditions. In that case, it may be nice to at least log it. Unless we are very certain this will really never happen. Don't need to log here necessarily, can also bubble up the error a bit. What works best.

joostjager: Oh, seeing now that you added the log already. Should be `Errorf` if it is a bug

rpcserver.go
```diff
@@ -2702,12 +2701,43 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
...
+		// If the channel cannot be found, log an error and do not perform
+		// further calculations for uptime and lifespan.
+		rpcsLog.Warnf("GetLifespan %v error: %v", chanID, err)
+	} else {
+		// If endTime is zero, the channel is still open, progress endTime to
+		// the present so we can calculate lifespan.
+		if endTime.IsZero() {
+			endTime = time.Now()
+		}
+		lifespan = startTime.Sub(endTime)
```
halseth: reminder to reverse this

rpcserver.go
```diff
@@ -2702,12 +2701,43 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
...
+
+		uptime, err = r.server.chanEventStore.GetUptime(
+			chanID,
```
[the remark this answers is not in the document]
carlaKC: they don't quiiite fit :)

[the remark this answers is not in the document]
carlaKC: ah, I've been conflating functions and structs style wise :o

PR #3337, opened 2019-07-22
lnd: Add regtest exception for wait to sync on startup

### LINKED ISSUE #3336: Regtest hangs on startup when blocks are old
### Background

I run LND with a bitcoind regtest backend and have noticed that it hangs on `Waiting for chain backend to finish sync` on start up. Now that LND is considering a node unsynced if block timestamp > 2 hours old, it hangs quite often. 

It's a really minor issue but I found this exception for simnet in `lnd.go` and was wondering if there's any reason not to include `cfg.bitcoin.RegTest` as well? I would like to add it if there are no objections. 
```
	// If we're not in simnet mode, We'll wait until we're fully synced to
	// continue the start up of the remainder of the daemon. This ensures
	// that we don't accept any possibly invalid state transitions, or
	// accept channels with spent funds.
	if !(cfg.Bitcoin.SimNet || cfg.Litecoin.SimNet) {
```
### Your environment
* LND 0.7.0-beta
* Mojave 10.14.5
* Bitcoind

### Steps to reproduce
1. Run LND with Bitcoind backend
2. Stop LND and do not mine any blocks on regtest for two hours
3. Restart LND and it should hang on waiting to sync because 

### Expected behaviour

Regtest should also proceed with startup despite not being synced to the blockchain, as is the case with Bitcoin/Litecoin.

### Actual behaviour
The regtest LND node waits for you to mine more blocks before it proceeds with startup.

carlaKC:
Allow startup to continue without syncing to chain when in regtest mode, as is already the case for Bitcoin/Litecoin simnet. 

Fixes #3336.

### DIFF
--- a/lnd.go
+++ b/lnd.go
@@ -469,11 +469,13 @@ func Main() error {
 	}
 	defer rpcServer.Stop()
 
-	// If we're not in simnet mode, We'll wait until we're fully synced to
-	// continue the start up of the remainder of the daemon. This ensures
-	// that we don't accept any possibly invalid state transitions, or
+	// If we're not in regtest or simnet mode, We'll wait until we're fully
+	// synced to continue the start up of the remainder of the daemon. This
+	// ensures that we don't accept any possibly invalid state transitions, or
 	// accept channels with spent funds.
-	if !(cfg.Bitcoin.SimNet || cfg.Litecoin.SimNet) {
+	if !(cfg.Bitcoin.RegTest || cfg.Bitcoin.SimNet ||
+		cfg.Litecoin.RegTest || cfg.Litecoin.SimNet) {
+
 		_, bestHeight, err := activeChainControl.chainIO.GetBestBlock()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


lnd.go
```diff
@@ -469,11 +469,11 @@ func Main() error {
...
 
-	// If we're not in simnet mode, We'll wait until we're fully synced to
-	// continue the start up of the remainder of the daemon. This ensures
-	// that we don't accept any possibly invalid state transitions, or
+	// If we're not in regtest or simnet mode, We'll wait until we're fully
+	// synced to continue the start up of the remainder of the daemon. This
+	// ensures that we don't accept any possibly invalid state transitions, or
 	// accept channels with spent funds.
-	if !(cfg.Bitcoin.SimNet || cfg.Litecoin.SimNet) {
+	if !(cfg.Bitcoin.RegTest || cfg.Bitcoin.SimNet || cfg.Litecoin.SimNet) {
```
wpaulino: We should also include litecoin's regtest.

PR #3347, opened 2019-07-26
ticker+queue: run go mod tidy

guggero:
Apparently my IDE runs `go mod tidy` on all subfolders that have a `go.mod` in them.
This is quite annoying since it always changes my git working tree to dirty.
So the easiest fix would be to commit the changes, which I hereby suggest.
I know, I should probably use a "real" IDE that doesn't do stuff like this but I really like IntelliJ...


guggero:
The file `queue/go.sum` is not removed, just emptied.
I tried importing `github.com/lightningnetwork/lnd/queue` in an example project and it works as expected. It still pulls in the correct version of `ticker` as a transitive dependency.

Issue #3286, opened 2019-07-10
htlcswitch: allow nodes to express max acceptable CLTV value 


Roasbeef:
As is today, [we have a hard coded limit on the maximum outgoing relative CLTV delta that we'll accept for an outgoing payment](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L33). This limits our _worst case_ delay, in the case that we need to go on-chain to resolve on HTLC. In retrospect, this limit is rather high (nearly 30 days), and recently deployed node level features such as hodl invoices may make long held HTLCs more frequent. Therefore, it should be possible for nodes to express a lower max block delta value. If an HTLC is extended that has an outgoing CLTV delta _beyond_ this value, then it should be rejected. We may also want to consider lowering out default value, given that we now use 40 block CLTV details for all newly created channels.

PR #3348, opened 2019-07-27
htlcswitch: make max cltv expiry configurable and lower default

wpaulino:
The current value was based on the previous default CLTV delta of 144 blocks. This has been lowered to 40 since lnd v0.6.0-beta, making the current value of 5000 blocks a bit high. Lowering it to one week should be more than enough to account for the other major lightning implementations. Eclair currently has a default CLTV delta of 144, while c-lightning's is 14.

The value can be configured through the `--max-cltv-expiry` flag.

Fixes https://github.com/lightningnetwork/lnd/issues/3286.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -30,16 +30,16 @@ func init() {
 }
 
 const (
-	// maxCltvExpiry is the maximum outgoing time lock that the node accepts
-	// for forwarded payments. The value is relative to the current block
-	// height. The reason to have a maximum is to prevent funds getting
-	// locked up unreasonably long. Otherwise, an attacker willing to lock
-	// its own funds too, could force the funds of this node to be locked up
-	// for an indefinite (max int32) number of blocks.
+	// DefaultMaxOutgoingCltvExpiry is the maximum outgoing time lock that
+	// the node accepts for forwarded payments. The value is relative to the
+	// current block height. The reason to have a maximum is to prevent
+	// funds getting locked up unreasonably long. Otherwise, an attacker
+	// willing to lock its own funds too, could force the funds of this node
+	// to be locked up for an indefinite (max int32) number of blocks.
 	//
 	// The value 5000 is based on the maximum number of hops (20), the
 	// default cltv delta (144) and some extra margin.
-	maxCltvExpiry = 5000
+	DefaultMaxOutgoingCltvExpiry = 5000
 
 	// DefaultMinLinkFeeUpdateTimeout represents the minimum interval in
 	// which a link should propose to update its commitment fee rate.
@@ -249,6 +249,11 @@ type ChannelLinkConfig struct {
 	// encrypting, and uploading of justice transactions to the daemon's
 	// configured set of watchtowers.
 	TowerClient TowerClient
+
+	// MaxCltvExpiry is the maximum outgoing timelock that the link should
+	// accept for a forwarded HTLC. The value is relative to the current
+	// block height.
+	MaxOutgoingCltvExpiry uint32
 }
 
 // channelLink is the service which drives a channel's commitment update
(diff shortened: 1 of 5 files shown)


htlcswitch/link.go
```diff
@@ -37,9 +37,11 @@ const (
...
 	// to be locked up for an indefinite (max int32) number of blocks.
 	//
-	// The value 5000 is based on the maximum number of hops (20), the
-	// default cltv delta (144) and some extra margin.
-	DefaultMaxOutgoingCltvExpiry = 5000
+	// The value 1008 corresponds to on average one week worth of blocks and
+	// is based on the maximum number of hops (20), the default cltv delta
+	// (40) and some extra margin to account for the other lightning
+	// implementations.
+	DefaultMaxOutgoingCltvExpiry = 1008
```
cfromknecht: IMO cutting this value by 80% is far too drastic, i'd prefer to slowly reduce this value over time to prevent any unforeseen interactions with pathfinding and other implementations.

fwiw, after the variable-size onion payload is deployed, the max pathlength will be closer to 26. the extra margin included in the original value was 5000 - 144 * 20 = 2120 blocks. Adding that to say 40*26 would give a value of 3160. Somewhere around 3k I think a more reasonable reduction. We can drop this again in the future if we don't see any regressions

Roasbeef: 3k would still result in a _20 day_ timelock for HTLCs that will be auto accepted. That's even longer than our max CSV delay (where most of the funds ideally are in a channel). Those values are based on the max path length, which afaik based on the current graph diameter, is nowhere close to being hit. If there're nodes today that are routinely extending paths of length 20, then that likely indicates they're attempting to launch nuisance attacks on the network.

cfromknecht: all of that makes sense. i'm only expressing concern over how dramatic the drop si, not that we shouldn't shoot to end at this lower value.

wpaulino: One could also argue that the previous value didn't need to be so high.

joostjager:
@wpaulino, when we receive a `FailExpiryTooSoon` in mission control, we black list _all_ of the channels of that node. By lowering the default limit, your node is at risk of this happening. Is this intended?

A nicer, much more involved, solution would be to have nodes announce their max cltv (node or channel level), so we can take it into account during path finding.

joostjager:
Current distribution of time_lock_delta on mainnet:

time_lock_delta | occurences
--- | ---
144 | 28612
40 | 16650
30 | 7004
14 | 2985
4 | 2033
64 | 607
20 | 551
432 | 297
1000 | 295
10 | 253
72 | 180
39 | 156
300 | 141
34 | 137
6 | 135
7 | 131
142 | 104
140 | 69
100 | 59
12 | 49
43 | 44
24 | 31
60 | 19
288 | 14
420 | 14
48 | 13
80 | 10
50 | 9
143 | 7
600 | 6
576 | 5
101 | 4
120 | 4
2016 | 2
32 | 2
36 | 2
1440 | 1
145 | 1
16 | 1
21 | 1
250 | 1
44 | 1

wpaulino:
> @wpaulino, when we receive a FailExpiryTooSoon in mission control, we black list all of the channels of that node. By lowering the default limit, your node is at risk of this happening. Is this intended?

I wasn't aware we had this sort of interpretation on a `FailExpiryTooSoon` error.

> A nicer, much more involved, solution would be to have nodes announce their max cltv (node or channel level), so we can take it into account during path finding.

I don't think there's an existing way to do this in the protocol?

> Current distribution of time_lock_delta on mainnet:

Looking at this, bumping the default to 2016 should be more than enough.

joostjager:
>Looking at this, bumping the default to 2016 should be more than enough.

We keep running into the error, even with 2016 and long routes? Why do you think it is more than enough?

There is already a max cltv limit restriction. We could also just change the default from no limit to 1000 or so

PR #3282, opened 2019-07-09
[mobile] Mobile RPCs

halseth:
This PR adds the necessary APIs and build scripts in order to compile `lnd` for mobile platforms.

#### gomobile
Follow [gomobile](https://github.com/golang/go/wiki/Mobile) in order to intall `gomobile` and dependencies.

Remember to run `gomobile init`! (otherwise the `lnd` build might just hang).

#### falafel
Install [`falafel`](https://github.com/halseth/falafel):
```
go get -u -v github.com/halseth/falafel
``` 

### Building `lnd` for iOS
```
make ios
```

### Building `lnd` for Android
```
make android
```

`make mobile` will build both iOS and Android libs.

### Libraries
After the build has succeeded, the libraries will be found in `mobile/build/ios/Lndmobile.framework` and `mobile/build/android/Lndmobile.aar`. Reference your platforms' SDK documentation for how to add the library to your project.

### API docs
TODO

PR #3349, opened 2019-07-27
Export function allowing ChainControl access from LND package

jacohend:
This PR creates a new exported LND method to allow full initialization of LND while returning a ChainControl pointer. This is to allow other go applications (which may wish to integrate LND) access to the underlying chain. The alternative is to run LND with neutrino, then run a separate instance of neutrino for the go application.

### DIFF
--- a/chainregistry.go
+++ b/chainregistry.go
@@ -110,25 +110,25 @@ func (c chainCode) String() string {
 // particular chain together. A single chainControl instance will exist for all
 // the chains lnd is currently active on.
 type chainControl struct {
-	chainIO lnwallet.BlockChainIO
+	ChainIO lnwallet.BlockChainIO
 
-	feeEstimator lnwallet.FeeEstimator
+	FeeEstimator lnwallet.FeeEstimator
 
-	signer input.Signer
+	Signer input.Signer
 
-	keyRing keychain.SecretKeyRing
+	KeyRing keychain.SecretKeyRing
 
-	wc lnwallet.WalletController
+	Wc lnwallet.WalletController
 
-	msgSigner lnwallet.MessageSigner
+	MsgSigner lnwallet.MessageSigner
 
-	chainNotifier chainntnfs.ChainNotifier
+	ChainNotifier chainntnfs.ChainNotifier
 
-	chainView chainview.FilteredChainView
+	ChainView chainview.FilteredChainView
 
-	wallet *lnwallet.LightningWallet
+	Wallet *lnwallet.LightningWallet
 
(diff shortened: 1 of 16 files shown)


lnd.go
```diff
@@ -535,21 +531,30 @@ func Main() error {
...
 		defer tower.Stop()
 	}
+	return activeChainControl, nil
+}
 
+// Main is the true entry point for lnd. This function is required since defers
+// created in the top-level scope of a main method aren't executed if os.Exit()
+// is called.
+func Main() error {
+	if _, err := InitDaemon(); err != nil {
```
cfromknecht: not sure this will actually work in practice. all of the defers in `InitDaemon` will be executed upon it returning, meaning all databases will be closed and all servers will be shutdown. Those defers need to happen inside the scope of `Main`

utxonursery.go
```diff
@@ -1069,7 +1069,7 @@ type contractMaturityReport struct {
 	limboBalance btcutil.Amount
 
 	// recoveredBalance is the total value that has been successfully swept
-	// back to the user's wallet.
+	// back to the user's Wallet.
```
cfromknecht: IMO none of the Wallet capitalization comments in this commit need to be touched. this commit should only touch the changes that affect compilation

chainregistry.go
```diff
@@ -110,25 +110,25 @@ func (c chainCode) String() string {
 // particular chain together. A single chainControl instance will exist for all
 // the chains lnd is currently active on.
 type chainControl struct {
-	chainIO lnwallet.BlockChainIO
+	ChainIO lnwallet.BlockChainIO
```
cfromknecht: now that these fields are exported, they will need to have proper godocs

lnd.go
```diff
@@ -93,15 +93,12 @@ var (
 	}
 )
 
-// Main is the true entry point for lnd. This function is required since defers
-// created in the top-level scope of a main method aren't executed if os.Exit()
-// is called.
-func Main() error {
+func InitDaemon() (*chainControl, error) {
```
cfromknecht: `chainControl` would need to be exported too
```
lnd.go:99:20:warning: exported func InitDaemon returns unexported type *lnd.chainControl, which can be annoying to use (golint)
```

Roasbeef:
What use case does this cover that #3282 doesn't?

Roasbeef:
> access to the underlying chain

By chain do you mean the primary interfaces within `lnd`?

jacohend:
@Roasbeef #3282  works for me

PR #1734, opened 2018-08-16
 routing: prune based on channel sets instead of channels

joostjager:
**UPDATE 2019-08-02**

Things have moved on since this PR was opened. I pushed an updated version of node pair pruning in mission control.

Advantages of node pair pruning:

* When two nodes have multiple public channels between them, mission control is only going to try one of those channels. With non-strict forwarding (which all current implementations support), this should be enough. This does assume that the same policy is set for all channels between those nodes (as is recommended by the LN spec).

* When a node closes a channel and opens a new channel to the same peer, it won't need to rebuild its reputation with senders for the new channel.

**ORIGINAL PR DESCRIPTION**

In the light of the discussion in https://github.com/lightningnetwork/lnd/pull/1527, it could be useful to also acknowledge in the retry behaviour that the actual channel being used isn't known. It may be better to associate failures not with specific channels, but with the two endpoints (pubkeys) of the channel. So instead of pruning channel 123, we prune the set of channels from node A to node B. (In case of a single channel, the result is the same)

@cfromknecht:
>most certainly, that's something we've discussed before as it's very related to this PR.
>
>Since we don't have certainty about what channel the other hops try, the feedback to the router is lossy (and potentially malicious/incorrect). With sufficient fees, multiple channels between the same peers can be effectively treated as one channel from the source's perspective, so the suggestion to prune all channels connecting the pubkeys makes sense to me.
>
>It also shouldn't be an issue if we assume that a node will always try the "best" (or at least sufficient) channel first, as then there'd no point retrying if the others are worse. However, this change could still cause us to miss a possible route if they maintain a strict forwarding policy.
>
>Atm we will try to find the hop with the most bandwidth. I think c-lightning has a max of one channel per peer, so it wouldn't be an issue there. Eclair looks like it abides by the channel requested in the payload, though as a disclaimer, my scala is a little rusty. Just based on that knowledge, it may produce less-than-optimal results when routing through an eclair node that has multiple outgoing channels with the next peer.

So for Eclair what would happen:
- Eclair node E does strict forwarding over channel 123 to destination F, finds that its policy isn't satisfied and sends back a channel update for channel 123.
- Source node S processes the channel update and adds all channels from E to F to its failed once set.
- S will calculate a second route using the updated graph. It is not yet going to avoid E-F channels, because a channel of that set has only failed once. If the rerouting again chooses the E-F connection, three things can happen:
  - Channel 123 is selected again, but this time with the right fee/delta timelock. The forwarding from E to F is expected to succeed.
[...]

PR #3353, opened 2019-07-29
routing: undirected node pair based mission control [wip] [don't review]

joostjager:
Builds onto #1734

Previously mission control tracked failures on a directed node pair basis. This PR changes this to tracking on the level of _undirected_ node pairs. The goal of moving to this coarser-grained level is to reduce the number of required payment attempts without compromising payment reliability.

Deeper analysis of channel-level failures and the way they should be attributed to nodes suggests that there always remains an uncertainty about which of the two endpoints is responsible for the failure. It is therefore appropriate to penalize the pair of nodes as a whole. This penalizes the bad node, while the good node only sees its connection to the bad node penalized. This prevents bad node pairs from being tried in both directions.

An exception is insufficient channel balance (`FailTemporaryChannelFailure`). Undirected penalization may seem too aggressive in this case, but:
* It is likely that channels are used for sending payments in a single direction only, away from the sender. Theoretically it is possible to connect to the graph in such a way that a channel may get used both ways, but this seems rare. (Note: this is just about sending. Receiving obviously utilizes channels in the other direction, but for that, path finding is the other party's responsibility.)
* One could argue that a channel that isn't balanced in one direction, should be avoided in the opposite direction too. It could be a sign that one of the nodes is not properly taking care of the channel (and possibly its other channels).
* If we are more tolerant to temp. chan. failures, nodes may use that to downgrade their penalty. Example: a node advertises an unreliable public channel. The next peer is offline. By sending back a temp. chan. failure instead of an unknown next peer failure, the peer that is offline isn't penalized. This may be good for the failure-sending node because it may generate incoming traffic in the future.

This PR also prepares for #3256 where a correct, gaming-resistent, set of failure attribution rules is implemented.


joostjager:
Closing this pr for now. I still think we will eventually be forced to move towards undirected pair-based penalization to be fully gaming resistent, but this will probably not be on the short term. Will keep this pr around for when that moment comes.

PR #3354, opened 2019-07-29
Peernotifier: Add Peer Notifier package

carlaKC:
This PR adds a peer notifier package which supplies clients with peer online/offline subscriptions and a corresponding rpc endpoint. 

It is intended for use in work related to #3332 to incorporate peer online/offline events in channel scoring. 

#### Pull Request Checklist
- [x] All changes are Go version 1.12 compliant, commented and formatted
  logging level - peer logger was used, it felt like the most appropriate?
- [x] Commits build properly, tests pass
- [x] Make check, go vet and make lint ok

### DIFF
--- a/log.go
+++ b/log.go
@@ -31,6 +31,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/monitoring"
 	"github.com/lightningnetwork/lnd/netann"
+	"github.com/lightningnetwork/lnd/peernotifier"
 	"github.com/lightningnetwork/lnd/routing"
 	"github.com/lightningnetwork/lnd/signal"
 	"github.com/lightningnetwork/lnd/sweep"
@@ -90,6 +91,7 @@ var (
 	chbuLog = build.NewSubLogger("CHBU", backendLog.Logger)
 	promLog = build.NewSubLogger("PROM", backendLog.Logger)
 	wtclLog = build.NewSubLogger("WTCL", backendLog.Logger)
+	prnfLog = build.NewSubLogger("PRNF", backendLog.Logger)
 )
 
 // Initialize package-global logger variables.
@@ -119,6 +121,7 @@ func init() {
 	chanbackup.UseLogger(chbuLog)
 	monitoring.UseLogger(promLog)
 	wtclient.UseLogger(wtclLog)
+	peernotifier.UseLogger(prnfLog)
 
 	addSubLogger(routerrpc.Subsystem, routerrpc.UseLogger)
 	addSubLogger(wtclientrpc.Subsystem, wtclientrpc.UseLogger)
@@ -165,6 +168,7 @@ var subsystemLoggers = map[string]btclog.Logger{
 	"CHBU": chbuLog,
 	"PROM": promLog,
 	"WTCL": wtclLog,
+	"PRNF": prnfLog,
 }
 
 // initLogRotator initializes the logging rotator to write logs to logFile and
--- a/peernotifier/log.go
+++ b/peernotifier/log.go
@@ -1,0 +1,29 @@
+package peernotifier
+
+import (
+	"github.com/btcsuite/btclog"
+	"github.com/lightningnetwork/lnd/build"
+)
+
+// log is a logger that is initialized with no output filters.  This
+// means the package will not perform any logging by default until the caller
+// requests it.
+var log btclog.Logger
+
+// The default amount of logging is none.
+func init() {
+	UseLogger(build.NewSubLogger("PRNF", nil))
+}
+
+// DisableLog disables all library log output.  Logging output is disabled
+// by default until UseLogger is called.
+func DisableLog() {
+	UseLogger(btclog.Disabled)
+}
+
+// UseLogger uses a specified Logger to output package logging info.
+// This should be used in preference to SetLogWriter if the caller is also
+// using btclog.
+func UseLogger(logger btclog.Logger) {
+	log = logger
+}
--- a/peernotifier/peernotifier.go
+++ b/peernotifier/peernotifier.go
@@ -1,0 +1,87 @@
+package peernotifier
+
+import (
+	"sync"
+
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// PeerNotifier is a subsystem which observes peer offline and online events.
+// It takes subscriptions for its events, and whenever it observes a new event
+// it notifies its subscribers over the proper channel.
+type PeerNotifier struct {
+	started sync.Once
+	stopped sync.Once
+
+	ntfnServer *subscribe.Server
+}
+
+// PeerOnlineEvent represents a new event where a peer comes online.
+type PeerOnlineEvent struct {
+	// PubKey is the peer's compressed public key.
+	PubKey [33]byte
+}
+
+// PeerOfflineEvent represents a new event where a peer goes offline.
+type PeerOfflineEvent struct {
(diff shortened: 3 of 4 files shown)


Roasbeef:
If it's for #3332, then why do we need the RPC end point as well?

carlaKC:
> If it's for #3332, then why do we need the RPC end point as well?

I thought that it would be a useful endpoint to have for people writing their own channel management logic? Eg. somebody who's not using autopilot, and is deciding which one of their peers to open a channel to. If they subscribe to peer events and store them, they'd have an idea of how reliably online each peer has been.

peernotifier/log.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// log is a logger that is initialized with no output filters.  This
+// means the package will not perform any logging by default until the caller
+// requests it.
+var log btclog.Logger
+
+// The default amount of logging is none.
+func init() {
+	UseLogger(build.NewSubLogger("PEER", nil))
```
halseth: I think `PEER` is most likely already in use?

peernotifier/peernotifier.go
```diff
@@ -0,0 +1,94 @@
...
+func (p *PeerNotifier) SubscribePeerEvents() (*subscribe.Client, error) {
+	return p.ntfnServer.Subscribe()
+}
+
+// NotifyPeerOnline sends a peer online event to all clients subscribed to the
+// peer notifier.
+func (p *PeerNotifier) NotifyPeerOnline(pubKey *btcec.PublicKey) {
+	event := PeerOnlineEvent{PubKey: pubKey}
+
+	log.Debugf(
```
halseth: style nit: For logging and other string formatting calls we usually do
```go
log.Debugf("PeerNotifier notifying peer: %x online",
		string(pubKey.SerializeCompressed()[:]))
```

carlaKC: cool, so generally for string formatting, keep the string inline and then newlines for all the vars like you do for functions? 
```
log.Debugf("Long sentence taking up space: %x and two values: %v", 
                   firstVar, 
                   secondVar,
)
```

Or just wrap at 80
```
log.Debugf("Long sentence taking up space: %x and two values: %v", 
                   firstVar, secondVar)
```

peernotifier/log.go
carlaKC: It is, but from what I can tell it won't be used at all except in unit tests because the `genSubLogger` param is nil? 

More generally, I wasn't sure whether to make a new sublogger in `lnd/log.go` for peer notifier (it felt a bit much since it's only used a handful of times) so I reused the peer logger. The comments in `lnd/log.go` say new logger "per subsystem"?

carlaKC:
Ok great, updated to fix style nit and remove rpc code (in favour of a separate PR). 

I've changed the logger to still use `peerLog`, just with a different tag, although I'm not sure whether `peerNotifier` should use `peerLog`, `ntfnLog` or a new logger?

peernotifier/peernotifier.go
```diff
@@ -0,0 +1,92 @@
...
+	return p.ntfnServer.Subscribe()
+}
+
+// NotifyPeerOnline sends a peer online event to all clients subscribed to the
+// peer notifier.
+func (p *PeerNotifier) NotifyPeerOnline(pubKey *btcec.PublicKey) {
+	event := PeerOnlineEvent{PubKey: pubKey}
+
+	log.Debugf("PeerNotifier notifying peer: %x online",
+		string(pubKey.SerializeCompressed()[:]),
```
cfromknecht: the argument here i believe only needs to be `pubKey.SerializeCompressed()`. also one of the few exceptions to our line wrapping is for log statements, so this would be acceptable:
```golang
        log.Debugf("PeerNotifier notifying peer: %x online",
                        pubKey.SerializeCompressed())
```

peernotifier/peernotifier.go
```diff
@@ -0,0 +1,92 @@
...
+	started sync.Once
+	stopped sync.Once
+
+	ntfnServer *subscribe.Server
+}
+
+// PeerOnlineEvent represents a new event where a peer comes online.
+type PeerOnlineEvent struct {
+	// PubKey is the ecdsa.PubKey of the peer that came online.
+	PubKey *btcec.PublicKey
```
cfromknecht: we may want to consider making these just `[33]byte` arrays, since for the most i think we only need to use this as an identifier. not a block change though, it's something we can consider in the future

peernotifier/peernotifier.go
cfromknecht: @carlaKC typically we shoot for the latter wrt to logging statements

peernotifier/log.go
cfromknecht: i'm okay with using the new `PRNF` tag for now, we can consolidate later if need be

peernotifier/peernotifier.go
carlaKC: Agreed that it's more likely to be used as an identifier, so I've changed to `[33]byte`. Extra info can always be added if it's needed, but I don't see much of a case for it as is.

carlaKC:
Rebased and updated to fix formatting nits & switch pubkeys to `[33] byte` since they'll most likely just be used as identifiers. 
PTAL @halseth 🙏

log.go
```diff
@@ -119,6 +120,7 @@ func init() {
 	chanbackup.UseLogger(chbuLog)
 	monitoring.UseLogger(promLog)
 	wtclient.UseLogger(wtclLog)
+	peernotifier.UseLogger(peerLog)
```
halseth: You need to create a new subsystem logger above, with tag "PRNF". If you look at the `subsystemsLoggers` map below, there should be a 1:1 mapping between tags and loggers :)

peernotifier/peernotifier.go
```diff
@@ -0,0 +1,87 @@
...
+	started sync.Once
+	stopped sync.Once
+
+	ntfnServer *subscribe.Server
+}
+
+// PeerOnlineEvent represents a new event where a peer comes online.
+type PeerOnlineEvent struct {
+	// PubKey is the peer's compressed public key.
+	PubKey [33]byte
```
halseth: +1 on this change! 👍

server.go
```diff
@@ -2722,6 +2734,13 @@ func (s *server) addPeer(p *peer) {
 	} else {
 		s.outboundPeers[pubStr] = p
 	}
+
+	// Inform the peer notifier of a peer online event so that it can be reported
+	// to clients listening for peer events.
+	var pubKey [33]byte
+	copy(pubKey[:], p.addr.IdentityKey.SerializeCompressed())
```
halseth: optimization: do this before creating the `pubStr` above, so don't we have to serialize twice :)

server.go
```diff
@@ -2972,6 +2991,13 @@ func (s *server) removePeer(p *peer) {
 	} else {
 		delete(s.outboundPeers, pubStr)
 	}
+
+	// Inform the peer notifier of a peer offline event so that it can be
+	// reported to clients listening for peer events.
+	var pubKey [33]byte
+	copy(pubKey[:], p.addr.IdentityKey.SerializeCompressed())
```
halseth: same here

log.go
carlaKC: Ok cool, I was wondering about that - `ntfnLog` threw me off a bit because it's used twice for `NTFN` and `NTFR`? Typo maybe?

halseth: Ooh, yeah that's a typo. I put up a fix: https://github.com/lightningnetwork/lnd/pull/3374

Good catch! 😅

Issue #3289, opened 2019-07-10
discovery+rpc: expose rough graph sync progress externally 


Roasbeef:
When a node is joining the network, atm, before it can being to send payments, it needs to synchronize the graph from other peers on the network. This is relevant to wallets, as they can grow to _disable_ sending payments all together if they know they don't have all/most of the graph at the ready. If we can expose a notion of the amount of the total known graph that we've synced, then this can be used by wallets to show an indicator or just disable sending all together. 

 At the time of writing of this issue, the [`GossipSyncer`](https://github.com/lightningnetwork/lnd/blob/master/discovery/syncer.go#L240) carries out this task. Several `GossipSyncer` instances are managed by a `SyncManager` instance. The `SyncManager` knows the state of all the `GossipSyncer` instances, so it should be possible to it to report externally if all the active historical sync instances are finished or still underway. 

# Steps To Completion 
- [ ] Modify the `SyncManager` to expose a new method that reports some sort of progress indicator for how much of the known graph we've ingested. 

- [ ] Either extend the `GetInfo` call, or perhaps `GetNetworkInfo` to display this new graph sync progress indicator.

PR #3355, opened 2019-07-29
discovery+rpc: expose graph synced status within GetInfo

wpaulino:
Fixes https://github.com/lightningnetwork/lnd/issues/3289.

~~One thing to note is that the current implementation will lead to GetNetworkInfo blocking if it's called while lnd is waiting for the backend to be considered synced. This is because we don't initialize lnd until that's done first. This can be addressed, but I've decided to leave it out unless we think it's something definitely worth addressing.~~


Roasbeef:
> One thing to note is that the current implementation will lead to GetNetworkInfo blocking if it's called while lnd is waiting for the backend to be considered synced.

This'll have an affect on tools such as `lndmon`, as they won't be able to plot the increase in chans known over time. Still need to dive into the diff, but can't we just return false in the scenario that would cause us to block?

discovery/sync_manager.go
```diff
@@ -660,3 +675,21 @@ func (m *SyncManager) gossipSyncers() map[route.Vertex]*GossipSyncer {
...
+// The initial historical sync is done to ensure we've ingested as much of the
+// public graph as possible.
+func (m *SyncManager) IsGraphSynced() bool {
+	resultChan := make(chan bool, 1)
+	select {
+	case m.isGraphSyncedReqs <- &isGraphSyncedReq{resultChan: resultChan}:
+	case <-m.quit:
+		return false
+	}
+	select {
```
Roasbeef: Style nit: could use a new line between the two `select` statements.

rpcserver.go
```diff
@@ -4029,6 +4029,7 @@ func (r *rpcServer) GetNetworkInfo(ctx context.Context,
 		MaxChannelSize:       int64(maxChannelSize),
 		MedianChannelSizeSat: int64(medianChanSize),
 		NumZombieChans:       numZombies,
+		SyncedToGraph:        r.server.authGossiper.SyncManager().IsGraphSynced(),
```
Roasbeef: If you make the return value channel, then we don't need to block here and can fall through to `false` in that case.

wpaulino: Went with a different approach instead.

cfromknecht:
for some reason i was under the impression we wanted to expose a percentage rather than just a boolean, was that not the case? other the code changes look pretty good

wpaulino:
@cfromknecht I went for the simplest approach possible as exposing a percentage seemed a bit more involved, though I'm not opposed to doing that.

sergioabril:
@cfromknecht IMO, a percentage would be much more useful. The boolean is ok, but feels incomplete to me.

wpaulino:
@sergioabril could you provide more insight for your use case of the percentage? Would it just be to expose the percentage to users in a higher level application? With the boolean, it should quickly flip to true unless it's your first time syncing the graph or you've been offline for a while.

sergioabril:
Yes, exactly for that; but I'm thinking of mobile wallets, which can be offline for a few days/week, and where a more precise state description is needed for the user to wait. Desktop instances are probably fine with a boolean.

cfromknecht:
@wpaulino 

> With the boolean, it should quickly flip to true unless it's your first time syncing the graph or you've been offline for a while.

If we connect to a peer that doesn't respond, will this continue to show `false` indefinitely?

wpaulino:
> If we connect to a peer that doesn't respond, will this continue to show false indefinitely?

@cfromknecht wasn't the case previously, but should be fixed in the latest version!

cfromknecht:
> @cfromknecht wasn't the case previously, but should be fixed in the latest version!

just looked at the latest diff, how is this solved? if we send a `query_channel_range` and the initial peer we connect to is slow to reply, then `is_synced` will continue to show false. in the case of the app, that means a peer not responding can prevent you from making payments

wpaulino:
See https://github.com/lightningnetwork/lnd/commit/fcb6a8b829072bae90f39d9b45a6f1a561bda939. We’ll pick up on that status of the next historical sync if our initial one hasn’t completed by then. Did you have another solution in mind?

cfromknecht:
> See fcb6a8b. We’ll pick up on the status of the next historical sync if our initial one hasn’t completed by then. 

i don't think we're talking about the same issue. the `is_synced` field can only become true if a peer responds to our `query_channel_range` message. if none of our peers respond, then it will always be false. if we gate payments in the app based on this field, then you will never be able to make a payment if you connect to poor/malicious peers.

> Did you have another solution in mind?

one option is to add a deadline after which it will flip to true regardless, that way if a peer doesn't respond we're still covered.

in my mind, there's really three states:
 1. `UNKNOWN`
 2. `SYNCING`
 3. `SYNCED`

when we come up, we start at `UNKNOWN`. after receiving any reply from our historical syncer[s] we then transition to:
 * `SYNCING`, if they have unknown channels, transitioning to `SYNCED` once a syncer completes
 * `SYNCED`, if they don't

my proposal is to add a timeout where if we stay in `UNKNOWN` for too long, we'll transition to `SYNCED` anyway and `is_synced` becomes true.

alternatively, these enums could be exposed over rpc to allow higher-level applications to act on. for example, if an app sees that we're still in `UNKNOWN` after some timeout, it can make the judgement call as to whether to allow payments or not

wpaulino:
> if we gate payments in the app based on this field, then you will never be able to make a payment if you connect to poor/malicious peers.

Why not just alert users rather than preventing them from completely making payments? Higher level applications can choose to interpret this field as they wish. They can decide to only gate payments up to some timeout as you suggested, gate payments regardless of how long it takes, alert users, or something else entirely.

> when we come up, we start at UNKNOWN. after receiving any reply from our historical syncer[s] we then transition to:
>
> * SYNCING, if they have unknown channels, transitioning to SYNCED once a syncer completes
> * SYNCED, if they don't
>
> my proposal is to add a timeout where if we stay in UNKNOWN for too long, we'll transition to > SYNCED anyway and is_synced becomes true.

Transitioning to true even though we haven't synced the graph seems to defeat the purpose of the flag, which is just to give an indication that we have done our part to sync the graph with at least one peer fully and will continue to do so through the consequent historical syncs. Just because they replied once doesn't mean we can guarantee they'll continue to do so, so we can still be in the `SYNCING` state. If they don't reply, do we flip back to `UNKNOWN`? Do we try another peer after a timeout? The additional complexity doesn't seem worth it.

> alternatively, these enums could be exposed over rpc to allow higher-level applications to act on. for example, if an app sees that we're still in UNKNOWN after some timeout, it can make the judgement call as to whether to allow payments or not

They can already do this with how things are at the moment, i.e. `synced_to_chain` is false after some timeout.

discovery/sync_manager.go
```diff
@@ -375,7 +375,21 @@ func (m *SyncManager) syncerHandler() {
...
-			m.forceHistoricalSync()
+			s := m.forceHistoricalSync()
+
+			// If we've already performed our initial historical
+			// sync, then we have nothing left to do.
+			if m.IsGraphSynced() {
+				continue
+			}
+
+			// Otherwise, we'll track the peer we've performed a
```
halseth: interesting. So this is really an existing bug, but it wouldn't really matter that the `initialHistoricalSyncCompleted` was never set?

wpaulino: Exactly. It would matter as it could prevent us from transitioning to active syncers.

lnrpc/rpc.proto
```diff
@@ -1878,6 +1878,9 @@ message NetworkInfo {
     // The number of edges marked as zombies.
     uint64 num_zombie_chans = 11 [json_name = "num_zombie_chans"];
 
+    // Whether we consider ourselves synced with the public channel graph.
+    bool synced_to_graph = 12 [json_name = "synced_to_graph"];
```
cfromknecht: been thinking about it, I think it’d be best if this were on getinfo. this call is way more expensive, and would severely limit the ability to long poll. no reason to make two rpc calls to figure out if we’re synced

discovery/sync_manager.go
```diff
@@ -373,7 +375,21 @@ func (m *SyncManager) syncerHandler() {
 		// Our HistoricalSyncTicker has ticked, so we'll randomly select
 		// a peer and force a historical sync with them.
 		case <-m.cfg.HistoricalSyncTicker.Ticks():
-			m.forceHistoricalSync()
+			s := m.forceHistoricalSync()
+
+			// If we've already performed our initial historical
+			// sync, then we have nothing left to do.
+			if m.IsGraphSynced() {
+				continue
```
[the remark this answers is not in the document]
wpaulino: Don't think so as we still want the historical syncs to continue to happen with other peers.

lnrpc/rpc.proto
wpaulino: Done.

discovery/sync_manager.go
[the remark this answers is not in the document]
cfromknecht: ah okay, got it. was thinking this ticker was only for the initial one

Roasbeef:
On the question of if we need to care about a peer stalling: that seems to be more of an issue for our peer selection, and eclipse attack mitigations, no? If a peer isn't responding, then we can move to time them out, and rotate to another peer. 

Our initial use case for this is also pretty simple, in just greying things out, or displaying this as an active indicator. Also practically, in the case of the application, they're connected to mostly known "golden" peers as well.

githorray:
I get ` "synced_to_graph": true` via the getinfo rpc call, but not via `lncli getinfo`.

wpaulino:
Thanks @githorray, fixed with https://github.com/lightningnetwork/lnd/pull/3400.

PR #3356, opened 2019-07-30
lnrpc: remove lncli subscribechannelevents

openoms:
There is no shell command for:
`$ lncli subscribechannelevents`
```
No help topic for 'subscribechannelevents'
```
This leads to an error when regenerating the docs with https://github.com/lightninglabs/lightning-api.
The current documentation does not contain the shell command for lncli subscribechannelevents either for the same reason: https://api.lightning.community/?shell#subscribechannelevents

Fixed when removed from rpc.proto.

### DIFF
--- a/lnrpc/rpc.pb.go
+++ b/lnrpc/rpc.pb.go
@@ -9482,7 +9482,7 @@ type LightningClient interface {
 	//ListChannels returns a description of all the open channels that this node
 	//is a participant in.
 	ListChannels(ctx context.Context, in *ListChannelsRequest, opts ...grpc.CallOption) (*ListChannelsResponse, error)
-	//* lncli: `subscribechannelevents`
+	//*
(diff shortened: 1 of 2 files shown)


openoms:
Installed the dependencies with the specific versions defined in https://github.com/lightningnetwork/lnd/tree/master/lnrpc#generate-protobuf-definitions and recompiled the .pb.go files with `make rpc` (cd ./lnrpc; ./gen_protos.sh)
The resulting changes are committed.

wpaulino:
We should only see changes in `rpc.pb.go` and `rpc.swagger.json`, so I'd double check you have the correct versions.

wpaulino:
Thanks @openoms, should be good to go after squashing the commits into one.

openoms:
Thank you for the hints @wpaulino! (I had an other protoc package hanging around.)

PR #3359, opened 2019-07-30
discovery: flag flip no historical gossip

cfromknecht:
This PR adds the `ignore-historical-filters` CLI option, initially
defaulting to false. Users may use this option to prevent lnd from doing
historical gossip dumps to peers that set their `gossip_timestamp_range`
in the past. Enabling this option will result in lower bandwidth and
memory consumption. Down the road the plan is to make this default to
true.


cfromknecht:
@Roasbeef test added, stripped the existing test from #3315 so should make the code removal at that point simpler as well

Issue #3357, opened 2019-07-30
Explicitly specifying fwdinghistory start_time of 0 causes lnd to use start_time of -24h


tyzbit:
### Background

Caused by https://github.com/lightningnetwork/lnd/pull/3246 😊

When explicitly specifying `--start_time=0`, instead of looking for forwarding events since the Unix epoch, LND sets the start time to -24h:

https://github.com/lightningnetwork/lnd/blob/master/rpcserver.go#L4641-L4643

Some options:
A. Make 0 always mean the unix epoch, whether intentionally set or not.
B. Make no change, inform users not to use 0 for start_times, instead use 1.

I like option A, but open to discussion

### Your environment

* version of `lnd` - v0.7.1-beta
* which operating system (`uname -a` on *Nix) Linux
* version of `btcd`, `bitcoind`, or other backend bitcoind 0.18.0
* any other relevant environment details

### Steps to reproduce

`lncli fwdinghistory --start_time=0`

### Expected behaviour

Shows forwarding history since the unix epoch.

### Actual behaviour

Shows forwarding history since 24 hours ago.

tyzbit:
I'll start on the fix for this; I'll start with option A as I outlined above.

PR #3360, opened 2019-07-31
rpcserver: startTime defaults to the unix epoch

tyzbit:
Closes https://github.com/lightningnetwork/lnd/issues/3357.  When
start_time isn't specified, its default value is 0.  This meant when
users explicitly specified a start_time of 0, we would incorrectly set
start_time to 24 hours in the past.  Now, n0 means n0.

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
@@ -3532,7 +3533,7 @@ var forwardingHistoryCommand = cli.Command{
 	payment circuits (HTLCs) over a particular time range (--start_time and
 	--end_time). The start and end times are meant to be expressed in
 	seconds since the Unix epoch. If --start_time isn't provided,
-	then the Unix epoch (01-01-1970) is used.  If --end_time isn't provided,
+	then 24 hours ago is used.  If --end_time isn't provided,
 	then the current time is used.
 
 	The max number of events returned is 50k. The default number is 100,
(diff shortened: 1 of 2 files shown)


tyzbit:
If we push this out to the CLI, does that mean we make some or all parameters mandatory, since we won't be applying any default-setting logic to them?  I agree with the premise behind validating user input at the CLI level, I just want to consider other consumers of the RPC that might be expecting some fields to have default values if left unset.

halseth:
I would say interpret `0` as unix epoch at RPC level, then set the default to 24 hours in the past on the CLI level.

halseth:
0 will be default for protos.

cmd/lncli/commands.go
```diff
@@ -3532,7 +3532,7 @@ var forwardingHistoryCommand = cli.Command{
 	payment circuits (HTLCs) over a particular time range (--start_time and
 	--end_time). The start and end times are meant to be expressed in
 	seconds since the Unix epoch. If --start_time isn't provided,
-	then 24 hours ago is used.  If --end_time isn't provided,
+	then the Unix epoch (01-01-1970) is used.  If --end_time isn't provided,
```
halseth: this change can be removed

tyzbit: That text change was reverted, do you want me to squash commits maybe?

rpcserver.go
```diff
@@ -4637,13 +4637,8 @@ func (r *rpcServer) ForwardingHistory(ctx context.Context,
...
 	)
 
-	// If the start time wasn't specified, we'll default to 24 hours ago.
-	if req.StartTime == 0 {
-		now := time.Now()
-		startTime = now.Add(-time.Hour * 24)
-	} else {
-		startTime = time.Unix(int64(req.StartTime), 0)
-	}
+	// startTime defaults to the Unix epoch (0 unixtime, or midnight 01-01-1970)
```
Roasbeef: Missing a period.

cmd/lncli/commands.go
halseth: Just make one commit for `rpcserver` and one for `lncli` :)

rpcserver.go
tyzbit: Fixed.

PR #3361, opened 2019-07-31
htlcswitch/link: increase channel reestablish timeout to 3m

cfromknecht:
This commit increases our channel reestablish timeout from 30s to 3m.
In debugging with rusty we were able to repro a scenario where sending
reestablish takes much longer than the current timeout. Under a typical
restart, contention can cause this timeout to be exceeded and result in
a sync error. Bumping this allows us to more gracefully tolerate these
cases before giving up on the channel.

tHiS sHoUlD bE hIgH eNoUgH

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -48,6 +48,12 @@ const (
 	// DefaultMaxLinkFeeUpdateTimeout represents the maximum interval in
 	// which a link should propose to update its commitment fee rate.
 	DefaultMaxLinkFeeUpdateTimeout = 60 * time.Minute
+
+	// chanReetablishTimeout represents the maximum duration we will wait
+	// for the remote peer to send us a channel reestablish message. This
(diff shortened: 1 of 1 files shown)


cfromknecht:
tbh 3m was chose a little arbitray. cl has a fix in master that prevents this, but this promotes compact with older versions. if there are any suggestions as to what this value should be i’m all ears. down the road I think we should consider resurrecting a channel if we’ve already given up on it and a peer sends us a reestablish, but that’s a bigger change

molxyz:
@cfromknecht  But why does LND have a timeout limit on `channel_reestablish` when the protocol doesn't say there should or should not be one?
https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#message-retransmission
Is it to prevent dossing?

Roasbeef:
Superseded by #3369.

PR #3365, opened 2019-08-01
multi: implement new safu commitment format 

Roasbeef:
In this PR, we implement the current working draft of the new safu commitment format (has no tweak for the remote party's output in your commitment transaction). In short, once we roll this out, once a channel is closed by the remote party, the funds will go _directly_ into an output that you control, so there's no need for any additional action. This essentially replaces the existing DLP protocol (we still need to understand it for old nodes though), as we no longer need the remote party's last unrevoked commitment point. 

This is still labeled as **WIP**, [as the current spec draft is missing some necessary signalling cut outs](https://github.com/lightningnetwork/lightning-rfc/pull/642/files#r309489446). Beyond that, I still need to update the integration tests to cover the new format (and still the old), and possibly a debug build flag guarded RPC to force the old format. 

On the topic of how this affects our current SCB scheme, the short answer is that it only _kinda_ does. With the change in this PR, the existing SCB files in the wild can be used with new channels, with the caveat that we'll still sweep the funds into our wallet, although we don't need to. In order to cut down on this extra transaction, we can instead add the commitment key (as an address) into a sort of _look aside_ while we undergo our regular scan for keys on chain during the recovery process. I say "look aside" as we don't generate keys from the key family we use to generate commitment keys in the normal process, and it isn't expected that they'll be used in order. Therefore, we'll provide these addresses as hints, distinct from the normal look ahead process. This allows us to pick up these funds during the normal recovery process.

At a later time, we can then update the regular on-chain recovery process to also scan for these commitment keys as normal. However, it's likely that it won't locate all the funds since we can't really expect them to be used in order (you don't FIFO close your channels, it's more arbitrary).

**NOTE**: the PR as a whole mostly calls this new format "tweakless" commitments, but upon reflection I find that it isn't very descriptive since it requires you to know low level details w.r.t how the scripts are constructed. IMO "safu commitments" is gets the point across a bit better, and it's more meme-y to boot, which may promote users to update sooner to this new version: "dude, you're not using safu commitments?! wot??".


Roasbeef:
Also the commits were crafted knowing that some of them don't build in isolation, as otherwise I would've had to commit the bulk of the PR in a single commit. IMO this is undesirable, as there're portions that need to be reviewed in isolation, and having them be distinct commits creates a guide for the reviewer.

lnwallet/channel.go
```diff
@@ -1973,8 +1991,11 @@ func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
 
 	// With the commitment point generated, we can now generate the four
 	// keys we'll need to reconstruct the commitment state,
-	keyRing := deriveCommitmentKeys(commitmentPoint, false,
-		&chanState.LocalChanCfg, &chanState.RemoteChanCfg)
+	tweaklessCommit := chanState.ChanType == channeldb.SingleFunderTweakless
+	keyRing := deriveCommitmentKeys(
+		commitmentPoint, false, tweaklessCommit,
```
joostjager: To prevent all those `tweaklessCommit := chanState.ChanType == channeldb.SingleFunderTweakless` statements, the `ChanType` could be passed into `deriveCommitmentKeys`. Maybe a slightly different abstraction, but deriving keys for a specific channel type doesn't sound too bad.

lnwallet/channel.go
```diff
@@ -5173,6 +5205,12 @@ func NewUnilateralCloseSummary(chanState *channeldb.OpenChannel, signer input.Si
 			},
 			MaturityDelay: 0,
 		}
+
+		// If this is a tweakless commitment, then we can safely blank
+		// out the SingleTweak value as it isn't needed.
+		if tweaklessCommit {
```
joostjager: I think not setting it above and only setting the tweak in case of `!tweaklessCommit` is clearer. Then the legacy flow is the exception.

lnwallet/channel.go
```diff
@@ -849,6 +849,12 @@ func (lc *LightningChannel) diskCommitToMemCommit(isLocal bool,
 	diskCommit *channeldb.ChannelCommitment, localCommitPoint,
 	remoteCommitPoint *btcec.PublicKey) (*commitment, error) {
 
+	// If this commit is tweakless, then it'll affect the way we derive our
+	// keys, which will affect the commitment transaction reconstruction.
+	// So we'll determine this first, before we do anything else.
+	tweaklessCommit := (lc.channelState.ChanType ==
```
cfromknecht: parentheses can be elided, here and elsewhere

otoh we could define a similar `IsTweakless()` helper a la `IsSingleFunder()` to avoid code duplication

lnwallet/channel.go
```diff
@@ -3513,7 +3537,11 @@ func (lc *LightningChannel) ProcessChanSyncMsg(
 		commitPoint = lc.channelState.RemoteNextRevocation
 	}
 
-	if commitPoint != nil &&
+	// We only need to check the unrevoked commitment point value if this
+	// isn't a tweakless commit, as otherwise, it'll be zero.
```
cfromknecht: we're still setting it in `ChanSyncMsg()`

cfromknecht:
we may also need a way of signaling to the client whether the channel is tweakless or not, since we assume that it's tweaked here https://github.com/lightningnetwork/lnd/blob/a53323205ca95c2ffa8af6ada41a45798f0d22a0/watchtower/wtclient/backup_task.go#L90

input/script_utils.go
```diff
@@ -923,14 +925,28 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+	case false:
+		witness[1] = TweakPubKeyWithTweak(
+			signDesc.KeyDesc.PubKey, signDesc.SingleTweak,
+		).SerializeCompressed()
+
+	// Otherwise, we can just use the raw pubkey, since there's no random
+	// value to be combined.
+	case true:
+		witness[1] = signDesc.KeyDesc.PubKey.SerializeCompressed()
+	}
```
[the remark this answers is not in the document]
Roasbeef: 😂

I prefer this over if/else due to the way that `gofmt` handles indenting comments that are above the else case. Usually, it'll indent them to the right, when I prefer for them to read linearly. As a personal preference, I think `switch` statements are more visually appealing, since they're more linear, and let you pick up the information quickly at a glance.

lnwallet/channel.go
Roasbeef: Opted to make a `IsTweakless` method.

breacharbiter_test.go
```diff
@@ -1754,9 +1754,10 @@ func createInitChannels(revocationWindow int) (*lnwallet.LightningChannel, *lnwa
 	}
 	aliceCommitPoint := input.ComputeCommitmentPoint(aliceFirstRevoke[:])
 
-	aliceCommitTx, bobCommitTx, err := lnwallet.CreateCommitmentTxns(channelBal,
-		channelBal, &aliceCfg, &bobCfg, aliceCommitPoint, bobCommitPoint,
-		*fundingTxIn)
+	aliceCommitTx, bobCommitTx, err := lnwallet.CreateCommitmentTxns(
+		channelBal, channelBal, &aliceCfg, &bobCfg, aliceCommitPoint,
+		bobCommitPoint, *fundingTxIn, true,
+	)
```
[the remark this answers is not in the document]
Roasbeef: There's no actual signature validation here, no are the keys checked. However, I've gone and updated the breach arb to be aware of the new commitment format, and added a new test case for serialization (for the new format, the `SingleTweak` is no longer needed). State transitions (of the channel), happen in the test, but this is tested more comprehensively in `lnwallet`. 

The integration tests will cover all these cases in detail, as along the way it verifies the script change, as if all areas aren't updated, then funds won't be able to be swept, as the commitment will be invalid.

lnwallet/channel.go
Roasbeef: Ended up making a `IsTweakless` method to eliminate the code duplication.

lnwallet/channel.go
Roasbeef: Ah yeah you're right, will wait until we fix the issue in the current spec draft re _removing_ it all together. Also FWIW, there's no harm in still setting it (other than the extra CPU cycles to compute the point from the secret).

Roasbeef:
Responded to the initial set of comments, and added commits to update the brar logic, as well as the tower logic. Still need to update the itests (I have a format I used in #3362 that I think will work well), and then also add a new funding flag to allow negotiation at funding flow time.

lnwallet/channel_test.go
```diff
@@ -3593,7 +3605,7 @@ func TestChanSyncFailure(t *testing.T) {
 	// Create a test channel which will be used for the duration of this
 	// unittest. The channel will be funded evenly with Alice having 5 BTC,
 	// and Bob having 5 BTC.
-	aliceChannel, bobChannel, cleanUp, err := CreateTestChannels()
+	aliceChannel, bobChannel, cleanUp, err := CreateTestChannels(false)
```
halseth: What's the rationale behind doing some tweakless here, some not?

rpcserver.go
```diff
@@ -2664,6 +2664,7 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
 		ChanStatusFlags:       dbChannel.ChanStatus().String(),
 		LocalChanReserveSat:   int64(dbChannel.LocalChanCfg.ChanReserve),
 		RemoteChanReserveSat:  int64(dbChannel.RemoteChanCfg.ChanReserve),
+		Tweakless:             dbChannel.ChanType.IsTweakless(),
```
halseth: should be exposed as "safu" on rpc?

lnwire/features.go
```diff
@@ -55,6 +55,16 @@ const (
...
 
+	// StaticRemoteKeyOptional is an optional feature bit that signals that
+	// within one's commitment transaction, the key used for the remote
+	// party's non-delay output should not be tweaked.
+	StaticRemoteKeyOptional FeatureBit = 10
+
+	// StaticRemoteKeyRequired is an optional feature bit that signals that
+	// within one's commitment transaction, the key used for the remote
+	// party's non-delay output should not be tweaked.
+	StaticRemoteKeyRequired FeatureBit = 11
```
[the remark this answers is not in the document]
halseth: Ok. Fixed in later commit.

input/script_utils.go
```diff
@@ -923,14 +925,28 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+
+	// TODO(roasbeef): optionally can make this implicit, by checking if
+	// the SingleTweak is populated or not.
+
+	switch tweakless {
+	// If we're tweaking the key, then we use the tweaked public key as the
+	// last item in the witness stack which was originally used to created
+	// the pkScript we're spending.
+	case false:
+		witness[1] = TweakPubKeyWithTweak(
```
halseth: I wonder is this PR would be simpler if we just changed `TweakPubKeyWithTweak` to be a no-op when given a empty `SingleTweak`? 

That way most of the structs and logic would stay the same, we would only set a zero/empty tweak for these channels.

channeldb/channel.go
```diff
@@ -132,8 +132,19 @@ const (
 	// funds towards the total capacity of the channel. The channel may be
 	// funded symmetrically or asymmetrically.
 	DualFunder = 1
+
+	// SingleFunderTweakless is similar to the based SingleFunder channel
```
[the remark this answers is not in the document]
Roasbeef: Not a typo, giving thanks to the Based God.

lnwallet/channel_test.go
Roasbeef: I only do tweakless where it would materially affect the test. Alternatively, everything can just be wrapped into one giant for loop and always test both behavior. I'm not sure if this is worth it though since we test the basic channel open an update functionality for both, while many tests just repeat that logic then asserts some db transitions that are unrelated to the commitment format.

rpcserver.go
Roasbeef: heh, maybe? 

Or to be more descriptive, perhaps `CommitTweakless` here?

input/script_utils.go
Roasbeef: I considered that, but instead opted for a more explicit API design to force the caller to recognize which they of witness they're requesting.

fundingmanager.go
```diff
@@ -1073,6 +1071,18 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	// reservation attempt may be rejected. Note that since we're on the
 	// responding side of a single funder workflow, we don't commit any
 	// funds to the channel ourselves.
+	//
+	// Before we init the channel, we'll also check to see if we've
+	// negotiated the new tweakless commitment format. This is only the
+	// case if *both* us and the remote peer are signalling the proper
```
alexbosworth:
```suggestion
	// case if *both* us and the remote peer are signaling the proper
```

fundingmanager.go
```diff
@@ -2784,6 +2795,18 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	// Initialize a funding reservation with the local wallet. If the
 	// wallet doesn't have enough funds to commit to this channel, then the
 	// request will fail, and be aborted.
+	//
+	// Before we init the channel, we'll also check to see if we've
+	// negotiated the new tweakless commitment format. This is only the
+	// case if *both* us and the remote peer are signalling the proper
```
alexbosworth:
```suggestion
	// case if *both* us and the remote peer are signaling the proper
```

lncfg/protocol_legacy_on.go
```diff
@@ -10,6 +10,12 @@ type LegacyProtocol struct {
 	// As a result, nodes that include us in the route won't use the new
 	// modern onion framing.
 	Onion bool `long:"onion" description:"force node to not advertise the new modern TLV onion format"`
+
+	// CommitmentTweak guards if we should use the old legacy commitment
+	// protocol, or the newer variant that doesn't have a tweak for the
+	// remote paryt's output in the commitment. If set to true, then we
```
alexbosworth:
```suggestion
	// remote party's output in the commitment. If set to true, then we
```

lnpeer/peer.go
```diff
@@ -46,4 +46,16 @@ type Peer interface {
 	// using the interface to cancel any processing in the event the backing
 	// implementation exits.
 	QuitSignal() <-chan struct{}
+
+	// LocalGloabFeatures returns the set of global features that has been
```
alexbosworth:
```suggestion
	// LocalGlobalFeatures returns the set of global features that has been
```

lnpeer/peer.go
```diff
@@ -46,4 +46,16 @@ type Peer interface {
...
+	// advertised by the local peer. This allows sub-systems that use this
+	// interface to gate their behavior off the set of negotiated feature
+	// bits.
+	LocalGlobalFeatures() *lnwire.FeatureVector
+
+	// RemoteGloabFeatures returns the set of global features that has been
```
alexbosworth:
```suggestion
	// RemoteGlobalFeatures returns the set of global features that has been
```

lnwallet/channel_test.go
halseth: Interesting. It would be useful to do it at least once do, to make sure it all passes.

I'm not opposed to wrapping them all in a for loop, it seems like free test coverage of the tweakless channels without much effort.

halseth: Would also let use smoothly transition to remove the tweaked format later, as all tests would be running for the new format.

fundingmanager.go
```diff
@@ -2785,7 +2796,17 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	// wallet doesn't have enough funds to commit to this channel, then the
 	// request will fail, and be aborted.
 	//
-	// TODO(roasbeef): signal tweakless or not
+	// Before we init the channel, we'll also check to see if we've
+	// negotiated the new tweakless commitment format. This is only the
```
halseth: So there is really no negotiation here? Assumption is that as long as both supports it, it will be used.

fundingmanager.go
```diff
@@ -1072,7 +1072,17 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
...
+	// case if *both* us and the remote peer are signaling the proper
+	// feature bit.
+	localTweakless := fmsg.peer.LocalGlobalFeatures().HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	remoteTweakless := fmsg.peer.RemoteGlobalFeatures().HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	tweaklessCommitment := localTweakless && remoteTweakless
```
halseth: would be nice to add the tweakless state to one of the logs below.

fundingmanager.go
```diff
@@ -2785,7 +2796,17 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
...
+	// negotiated the new tweakless commitment format. This is only the
+	// case if *both* us and the remote peer are signaling the proper
+	// feature bit.
+	localTweakless := msg.peer.LocalGlobalFeatures().HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	remoteTweakless := msg.peer.RemoteGlobalFeatures().HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	tweaklessCommitment := localTweakless && remoteTweakless
```
halseth: same, add to logs

rpcserver.go
halseth: PR description has rationale 
> IMO "safu commitments" is gets the point across a bit better, 

so would assume that would the same for information expressed on the RPC

lntest/itest/lnd_test.go
```diff
@@ -944,52 +944,171 @@ func testBasicChannelFunding(net *lntest.NetworkHarness, t *harnessTest) {
...
+				carolTweakless, daveTweakless)
+
+			ht := t
+			success := t.t.Run(testName, func(t *testing.T) {
+				carolChannel, daveChannel, closeChan, err := basicChannelFundingTest(
+					ht, net, carol, dave,
+				)
+				if err != nil {
+					t.Fatalf("failed funding flow: %v", err)
+				}
```
halseth: `defer closeChan()` here

lntest/itest/lnd_test.go
```diff
@@ -944,52 +944,171 @@ func testBasicChannelFunding(net *lntest.NetworkHarness, t *harnessTest) {
...
+test:
+	// We'll test all possible combinations of the feature bit presence
+	// that both nodes can signal for this new channel type. We'll make a
+	// new Carol+Dave for each test instance as well.
+	for _, carolTweakless := range []bool{true, false} {
+		for _, daveTweakless := range []bool{true, false} {
+			// Based on the current tweak variable for Carol, we'll
+			// preferentially signal the legacy commitment format.
+			// We do the same for Dave shortly below.
+			var carolArgs []string
```
halseth: should all this setup be done inside `Run()`? Could potentially let us run them in parallel.

.golangci.yml
```diff
@@ -41,6 +41,12 @@ linters:
   # trigger funlen problems that we may not want to solve at that time.
   - funlen
 
+  - gocyclo
```
halseth: Why is this disabled?

chanbackup/single.go
```diff
@@ -139,6 +143,14 @@ func NewSingle(channel *channeldb.OpenChannel,
 			},
 		},
 	}
+
+	if !channel.ChanType.IsTweakless() {
```
halseth: style nit: easter to read as
```go
if channel.ChanType.IsTweakless() {
     single.Version = TweaklessCommitVersion
} else {
     single.Version = DefaultSingleVersion
}
```

chanbackup/single_test.go
```diff
@@ -124,8 +124,14 @@ func genRandomOpenChannelShell() (*channeldb.OpenChannel, error) {
 		isInitiator = true
 	}
 
+	var chanType channeldb.ChannelType
```
halseth: should explicitly set chan type for tweak case

rpcserver.go
joostjager: Why not stick to the spec and call it `static_remote_key`?

contractcourt/chain_watcher.go
```diff
@@ -423,6 +423,11 @@ func (c *chainWatcher) closeObserver(spendNtfn *chainntnfs.SpendEvent) {
 		// revoked state...!!!
 		commitTxBroadcast := commitSpend.SpendingTx
 
+		// An additional piece of information we need to properly
+		// dispatch a close event if is this channel was using the
+		// tweakless remove key format or not.
+		tweaklessCommit := c.cfg.chanState.ChanType.IsTweakless()
```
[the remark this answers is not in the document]
Roasbeef: Only declared in a single place in the final diff. At one point, it's moved up so we can use the value when we go to derive the set of expected commitment keys for that state.

fundingmanager.go
```diff
@@ -1073,6 +1071,8 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	// reservation attempt may be rejected. Note that since we're on the
 	// responding side of a single funder workflow, we don't commit any
 	// funds to the channel ourselves.
+	//
+	// TODO(roasbeef): signal tweakless or not
```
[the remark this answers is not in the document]
Roasbeef: Yep in the latest diff at the very end, this comment (the TODO) should no longer be in this set of commitments though.

lnwallet/channel.go
```diff
@@ -2052,6 +2052,12 @@ func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
 			},
 			HashType: txscript.SigHashAll,
 		}
+
+		// If this is a tweakless commitment, then we can safely blank
+		// out the SingleTweak value as it isn't needed.
+		if tweaklessCommit {
+			localSignDesc.SingleTweak = nil
```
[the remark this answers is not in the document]
Roasbeef: If we don't create a new channel type, then we're force to rely on this implicit value to determine if what type of script we should use for validating, spending, etc. Alternatively, as joost suggested in a earlier diff, we can not set it by default, then control if we set it or not based on the channel type.

fundingmanager.go
Roasbeef: Yeah we do a logical AND w.r.t the existence of the feature bit in the init message. I had suggested in the past that we also let them signal it by using a new funding flag, but we have a single unified way by just using the feature bits. Put simply: only if both nodes support it is it used. Arguably, being able to set that value on the fly would make it easier for our integration tests, as we could use the same nodes (no additional flag) and just use the RPC system to modulate if we're doing tweakless or not.

lntest/itest/lnd_test.go
Roasbeef: Does it matter? After the first failure, we'll exit everything (halt all other test execution).

lntest/itest/lnd_test.go
Roasbeef: Since we rely on blocks being mined at a certain time, don't think we can really do them in parallel unless we have a unique chain for each instance. I left it out since it's set up code and iirc the timing shown only applies to the code being executed in the `Run` block.

.golangci.yml
Roasbeef: It failed the build due to a code movement. We also haven't calibrated it at all for Go and our codebase in general. I can see new contributors being discourage when they add a small change (say typo fix or minor code movement), but then the build fails due to existing code exceeding this value. I'm down to enable it if we can calibrate it properly, but that seems outside the scope of this PR.

chanbackup/single_test.go
Roasbeef: I think you mean non-teak, but done.

rpcserver.go
Roasbeef: I guess next in-line will be "super safu" channels? ;) 

Moved to be static key, but just because it's clear outside the context of this PR. In a few other areas we deviate from the spec to use subjectively more descriptive names (see the chan reest messages in `lnwire` as an example). We can still use "safu" colloquially when talking about or describing the feature.

lnwallet/channel.go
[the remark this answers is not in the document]
Roasbeef: For example, the SCB version bump relies in the existence of a new channel type.

lnrpc/rpc.proto
```diff
@@ -1176,6 +1176,14 @@ message Channel {
...
     */
     int64 remote_chan_reserve_sat = 21 [json_name = "remote_chan_reserve_sat"];
+
+    /**
+    If true, then this channel uses the modern commitment format where the key
+    in the output of the remote party does not change each state. This makes
+    back up and recovery easier as when the channel is closed, the funds go
+    directly to that key. 
+    */
+    bool static_remote_key = 22 [json_name = "tweakless"];
```
halseth: change json_name also

lntest/itest/lnd_test.go
halseth: Does not matter.

.golangci.yml
halseth: Add rationale as a comment perhaps?

lnwallet/channel.go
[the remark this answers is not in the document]
halseth: Yeah, maybe not set it by default, as that is where we will move eventually anyway.

lnwallet/channel.go
```diff
@@ -3508,8 +3672,11 @@ func (lc *LightningChannel) ProcessChanSyncMsg(
 		commitPoint = lc.channelState.RemoteNextRevocation
 	}
 
-	if commitPoint != nil &&
-		!commitPoint.IsEqual(msg.LocalUnrevokedCommitPoint) {
+	// If we have a valid commitment point to check, then we'll verify it
+	// here. We do this unconditionally, as even those with the new
+	// tweakless format need a way to signal that they need the channel
+	// closed at state zero.
+	if commitPoint != nil && !commitPoint.IsEqual(msg.LocalUnrevokedCommitPoint) {
```
halseth: Can we actually be checking this? The spec says 
```
MUST set `my_current_per_commitment_point` to a valid point
```
Not that it must be set to the latest unrevoked one.

Roasbeef: We did it unconditionally before. A prior version wouldn't check if it was tweakless, but this is wrong as we still need to signal to the upgraded party that we've lost state when we're at state zero. If we don't check this, then it looks to the up to date party, that the other party lost no data since there're checks earlier that we skip for state zero. 

For state zero against `lnd`, we purposefully set the wrong value so we can trigger a force close. many of the other implementations will auto force close based on the error we send out so I think we have that base covered.

input/script_utils.go
```diff
@@ -923,14 +925,28 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+	// exact same as a regular p2wkh witness, depending on the value of the
+	// tweakless bool.
 	witness := make([][]byte, 2)
 	witness[0] = append(sweepSig, byte(signDesc.HashType))
-	witness[1] = TweakPubKeyWithTweak(
-		signDesc.KeyDesc.PubKey, signDesc.SingleTweak,
-	).SerializeCompressed()
+
+	// TODO(roasbeef): optionally can make this implicit, by checking if
+	// the SingleTweak is populated or not.
```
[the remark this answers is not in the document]
Roasbeef: Is it safer? Explicit means the caller _must_ know what type they want. Things were slightly more simplicity before, but then I reversed it as there was a bug that would've been caught if the caller was forced to know what they wanted. It's also usually the case that the sign desc is passed through several call stacks, so it may be difficult to ensure things are properly set vs having the caller know what type of commitment they really want.

Roasbeef:
> This is a big pr and therefore costly to review and re-review. A split would have been great, but that it is probably impossible. 

Yeah it's difficult given that how a commitment is constructed touches so many other areas :/ 

FWIW, most of the diff is tests, and I even left out a few additional modifications that _could_ have been made within the tests, just to make the diff smaller. I plan to make these additional modifications either during the rc phase of 0.8 or afterwards, just to make sure that we're testing as many scenarios as is possible given the current testing infrastructure in the project. 

> and ensuring every commit builds and ideally also passes tests

This can be difficult as a simple change like adding an extra arg to some spending utils, or updating the funding flow for awareness touch so many related sub-systems, that the entire PR would end up just being a handful of much larger commits. There's a tradeoff between build-ability and test alway executing (if the OP desires that) and commit size. 

> Besides that, my main question is whether more groundwork should be laid down in this PR for the next commitment format change (possibly anchor outputs). In this PR we move from a single format to two formats that are often distinguished in if/else constructs or using a tweak boolean.

I usually use switch statements ;)

As for the next planned update, we split things off so we could ship things more quickly to cover our users, ideally in this release. As far as future ground work, we have the current `lncfg` based build flags for testing infrastructure. We'll also add a new channel type (which already exists), which similar to this PR will be checked to ensure we construct the commitment properly. Rather than modify an _existing_ output, the new change will add _new_ outputs, so much of the code will be touching weight estimates, and commitment transaction construction/verification. This also won't affect breach logic, DLP logic, or tower logic like this PR did.

fundingmanager.go
```diff
@@ -1073,6 +1071,18 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
...
+	// negotiated the new tweakless commitment format. This is only the
+	// case if *both* us and the remote peer are signaling the proper
+	// feature bit.
+	localTweakless := fmsg.peer.LocalGlobalFeatures().HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	remoteTweakless := fmsg.peer.RemoteGlobalFeatures().HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	tweaklessCommitment := localTweakless && remoteTweakless
```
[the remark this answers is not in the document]
Roasbeef: If we have required and they don't recognize the bit at all, [then we'll fail the _connection_ much earlier in the lifecycle, in the peer logic](https://github.com/lightningnetwork/lnd/blob/master/peer.go#L2367). If for some reason things passed through, then the channel they created will have an invalid commitment.

lnwallet/channel.go
```diff
@@ -857,14 +863,14 @@ func (lc *LightningChannel) diskCommitToMemCommit(isLocal bool,
 	var localCommitKeys, remoteCommitKeys *CommitmentKeyRing
 	if localCommitPoint != nil {
 		localCommitKeys = deriveCommitmentKeys(
```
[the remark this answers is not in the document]
Roasbeef: I don't think it needs to be a method. It doesn't critically depend on any of the attributes of `LightningChannel`. Forcing it to be a method would also mean that we're forced to create an partially populate a `LightningChannel` instance each time we want to derive or verify commitment keys. [An an example, here we create a commitment transaction during the funding phase, but the funding isn't fully completed, so we don't have a full `LightningChannel` instance yet](https://github.com/lightningnetwork/lnd/blob/master/lnwallet/wallet.go#L830).

lnwallet/channel_test.go
```diff
@@ -348,6 +339,26 @@ func TestSimpleAddSettleWorkflow(t *testing.T) {
...
+//
+// TODO(roasbeef): write higher level framework to exercise various states of
+// the state machine
+//  * DSL language perhaps?
+//  * constructed via input/output files
+func TestSimpleAddSettleWorkflow(t *testing.T) {
+	t.Parallel()
+
+	for _, tweakless := range []bool{true, false} {
+		tweakless := tweakless
```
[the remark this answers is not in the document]
Roasbeef: It's actually a valid warning though IMO, albeit only when all the internal tests are _also_ being run in parallel.

Roasbeef:
>  In this PR we move from a single format to two formats that are often distinguished in if/else constructs or using a tweak boolean. Many of those will probably need to be revised when the third format is added, resulting in another big diff. 

Conditional logic based on the existence of this new output type will touch mostly distinct codepaths from what's modified here. The bigger areas of impact will be fee estimation, broadcasting, and also likely some sweeper components. In short I expect that PR to be much larger than this core PR.

input/script_utils.go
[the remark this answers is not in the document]
joostjager: There are two sides to it. Yes, adding redundant parameters can catch bugs. But it can also camouflage bugs. In this case it could perhaps be that `SingleTweak` is set somewhere and we never notice because the `tweakless` bool makes us ignore it.

I think in a case like this where the pros and cons are not very clear (also signaled by the TODO), I'd probably choose the minimum code / parameters option.

lnwallet/channel_test.go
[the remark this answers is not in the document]
joostjager: Did a quick test, because I wasn't sure about `t.Parallel` in combination with `t.Run`. Indeed, also sub tests can be ran in parallel, but only if `t.Parallel` is called in the sub test. Calling it in the root test doesn't make the sub tests run concurrently. In this case, afaik the sub test doesn't call `t.Parallel`, but it is a good precaution.

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: Yes, I know it is also used without a `LightningChannel`. That is also why I suggested to also keep a global version of the function. 

The advantage of (also) having a method is that is reduce the potential for mistakes in passing the wrong value and deduplication of code in general.

fundingmanager.go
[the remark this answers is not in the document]
joostjager: Ah that is right. It is not a parameter for open channel.

joostjager:
> > and ensuring every commit builds and ideally also passes tests
> 
> This can be difficult as a simple change like adding an extra arg to some spending utils, or updating the funding flow for awareness touch so many related sub-systems, that the entire PR would end up just being a handful of much larger commits. There's a tradeoff between build-ability and test properly executing (if the OP desires that) and commit size.

I don't think that it requires much to get the stack of commits in this PR building and I also don't think you need to end up with fewer commits. I looked at a few failing ones and it is mainly just a missing parameter. Having all the commits passing build allows to do `git bisect run`. I quite often use this to efficiently reduce the change set in which to look for a problem. If you'd reorganize the commits so that opening and accepting staticremotekey channels is added in one of the last, every commit would even still be a fully functioning version on its own. First all the preparatory steps are taken to commit by commit support the new tx format in all sub systems and finally it is enabled.
 
> As for the next planned update, we split things off so we could ship things more quickly to cover our users, ideally in this release. As far as future ground work, we have the current `lncfg` based build flags for testing infrastructure. We'll also add a new channel type (which already exists), which similar to this PR will be checked to ensure we construct the commitment properly. Rather than modify an _existing_ output, the new change will add _new_ outputs, so much of the code will be touching weight estimates, and commitment transaction construction/verification. This also won't affect breach logic, DLP logic, or tower logic like this PR did.

I had this idea where there would be an object that contains everything that is specific to a particular commitment tx type. This object would have an interface through which it is used by the rest of the system. Adding new commitment formats would require the creation of a new type that implements this interface.

breacharbiter.go
```diff
@@ -944,17 +944,24 @@ func newRetributionInfo(chanPoint *wire.OutPoint,
...
 	// First, record the breach information for the local channel point if
 	// it is not considered dust, which is signaled by a non-nil sign
-	// descriptor. Here we use CommitmentNoDelay since this output belongs
-	// to us and has no time-based constraints on spending.
+	// descriptor. Here we use CommitmentNoDelay (or
+	// CommitmentNoDelayTweakless for newer commitments) since this output
+	// belongs to us and has no time-based constraints on spending.
 	if breachInfo.LocalOutputSignDesc != nil {
+		witnessType := input.CommitmentNoDelay
+		if breachInfo.LocalOutputSignDesc.SingleTweak == nil {
```
halseth: Seems a bit arbitrary to use the tweak value to signal tweakless channel here, but the channel type other places. Should make a decision which approach to go with and use that everywhere? (would possibly require chan type to be added to the breach info).

htlcswitch/link.go
```diff
@@ -952,7 +952,7 @@ func (l *channelLink) htlcManager() {
 
 			l.fail(
 				LinkFailureError{
-					code:       ErrSyncError,
+					code:       ErrRecoveryError,
```
halseth: Safe change? I know some implementations string match on the previous error message...

lnwallet/channel.go
halseth: The case I'm still not sure about is where we are at state `n` and everything above checks out. If other implementations send a static commit point we cannot compare it to our `RemoteNextRevocation` since they won't match. Similarly, if other implementations send a non-static point, we cannot compare to a static point.

From the spec:
> `option_static_remotekey` removes the changing `to_remote` key,
> so the `my_current_per_commitment_point` is unnecessary and thus
> ignored (for parsing simplicity, it remains and must be a valid point,
> however), but the disclosure of previous secret still allows
> fall-behind detection.  An implementation can offer both, however, and
> fall back to the `option_data_loss_protect` behavior if
> `option_static_remotekey` is not negotiated.

Roasbeef: Resolved by having us advertise an incorrect next height. Also realized that the change to send the new error will enable the current/new recovery protocol to work better with CL as well.

input/script_utils.go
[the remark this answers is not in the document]
Roasbeef: The TODO there was more or less just me recording thoughts as I was going, but I'm on the side of having the explicit parameter since an earlier version that didn't resulted in a bug that's now fixed. The TOOD has been removed in the latest diff.

lnwallet/channel.go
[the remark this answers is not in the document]
Roasbeef: I don't see how this change would clean up the diff, it seems marginal tbh. Also in the one case where this is now used as a public function, it wouldn't be able to use a method that say had no params (as it uses all the internal state), since it needs to possibly derive the keys for a _prior_ state (a revoked one possibly), requiring it to pass in a new custom commitment point for that state.

lnwire/features.go
```diff
@@ -55,6 +55,16 @@ const (
 	// packet.
 	TLVOnionPayloadOptional FeatureBit = 9
 
+	// StaticRemoteKeyRequired is a required feature bit that signals that
```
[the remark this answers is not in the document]
Roasbeef: Off by one error in my scripts, in the end once things land it's enough to just sign the final commit as that commits to all the others.

breacharbiter.go
Roasbeef: This doesn't have access to the channel type nor the channel state. Doing so would require more changes in the chain watcher and also channel state machine to thread through all the state.

htlcswitch/link.go
Roasbeef: In this case it is, as this version won't timeout on chan reest anymore. Also see my comment above w.r.t how that change by other implementations actually caused them to ignore our DLP recovery at times, since they wouldn't close based on that error.

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: I don't suggest to use all the internal state, just the chan configs and the tweakless bool. I think that a little more encapsulation could be beneficial for the maintainability of the code over time. It would also move a little bit towards abstraction of commitment formats.

lnwallet/channel.go
```diff
@@ -3591,6 +3760,14 @@ func ChanSyncMsg(c *channeldb.OpenChannel,
 	// allowing us to sweep our funds.
 	if isRestoredChan {
 		currentCommitSecret[0] ^= 1
+
+		// If this is a tweakless channel, then we'll purposefully send
+		// a next local height thta's invalid to trigger a force close
```
halseth: typo: thta

lnwallet/channel.go
```diff
@@ -3591,6 +3760,14 @@ func ChanSyncMsg(c *channeldb.OpenChannel,
...
+		// on their end. We do this as tweakless channels don't require
+		// that the commitment point is valid, only that it's present.
+		if c.ChanType.IsTweakless() {
+			nextLocalCommitHeight--
```
halseth: This probably works, but I find it a bit hard to analyze. Won't this result in a local commit height equal to the situation where the last ACK we sent got lost (and hence we increased the height, but the remote did not)?

I'm also not sure what happens in the zero-state case.

Could a more bulletproof solution be to set it to `0` always? That would still leave the zero-state case an edge case.

The best solution would be to change the spec to always send the correct commitpoint (as before), but that train has probably left...

Roasbeef: Modified to instead always set to zero.

Roasbeef:
Pushed up new version w/ slightly diff commitment structure.

PR #3366, opened 2019-08-01
discovery: prevent broadcast of anns received during initial graph sync

wpaulino:
There's no need to broadcast these as we assume that online nodes have already received them. For nodes that were offline, they should receive them as part of their initial graph sync.

Depends on #3355.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -913,6 +913,12 @@ func (d *AuthenticatedGossiper) networkHandler() {
 			policyUpdate.errResp <- nil
 
 		case announcement := <-d.networkMsgs:
+			// We should only broadcast this message forward if it
+			// originated from us or it wasn't received as part of
+			// our initial historical sync.
+			shouldBroadcast := !announcement.isRemote ||
+				d.syncMgr.IsGraphSynced()
+
 			switch announcement.msg.(type) {
 			// Channel announcement signatures are amongst the only
 			// messages that we'll process serially.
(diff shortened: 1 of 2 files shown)


discovery/gossiper.go
```diff
@@ -906,18 +906,21 @@ func (d *AuthenticatedGossiper) networkHandler() {
...
 			case *lnwire.AnnounceSignatures:
 				emittedAnnouncements := d.processNetworkAnnouncement(
 					announcement,
 				)
-				if emittedAnnouncements != nil {
-					announcements.AddMsgs(
-						emittedAnnouncements...,
-					)
-				}
+				announcements.AddMsgs(
```
Roasbeef: Why omit the check? It lets us avoid grabbing a mutex FWIW.

wpaulino: Good point, fixed.

Roasbeef:
Can be rebased now!

wpaulino:
Rebased

discovery/gossiper.go
```diff
@@ -981,12 +986,16 @@ func (d *AuthenticatedGossiper) networkHandler() {
 				// the emitted announcements to our announce
 				// batch to be broadcast once the trickle timer
 				// ticks gain.
-				if emittedAnnouncements != nil {
+				if emittedAnnouncements != nil && shouldBroadcast {
```
cfromknecht: will this skip broadcasting both local and remote anns? shouldn't local ones be allowed through?

wpaulino: Good catch, didn't think of this case. Fixed.

PR #3367, opened 2019-08-02
Batched graph updates

### LINKED ISSUE #3288: routing+channeldb: implement batched channel announcement+update writing
As implemented today, [only a single channel can be written (synchronization wise) within the daemon at any given time](https://github.com/lightningnetwork/lnd/blob/master/channeldb/graph.go#L498). As a result, one of the current bottle necks to spinning up a new `lnd` node is actually just _ingesting_ the graph. Due to the recently added edge/node caching within `channeldb`, using [the `Batch` call from `bbolt`](https://godoc.org/github.com/etcd-io/bbolt#DB.Batch) won't result in any increased concurrency due to the required mutex synchronization. Instead, we can _coalesce_ several channels at the router level, which are then periodically written to disk in a _single_ database transaction. This change should _significantly_ reduce the amount of time it takes to ingest the graph given that atm, we need a _new_ database transaction for each channel, which can take some time (in the realm of computers). 

# Steps To Completion 

- [ ] Add a new `AddBatchedChannelEdge` (or diff name) method that's able to take a _variadic_ `*channeldb.ChannelEdgeInfo` to be written in a single database transaction. Alternatively, we can simply extend the prior version (`AddChannelEdge`) with a variadic parameters which will allow us to keep all existing callers in place. 

- [ ] A similar method should be created the permits the caller to write several `*channeldb.ChannelEdgeInfo` instances in a single database transaction. 

- [ ] Rather than write channels [one at a time](https://github.com/lightningnetwork/lnd/blob/master/routing/router.go#L1226), the router should return immediately to the caller, but _cache_ that channel in memory until the next batch flush criteria is met. This can either be time based, or based on a maximum/minimum batch size. 

- [ ] Add a new coalescer component to the router that will "periodically" write the latest batch of channels to disk. 

- [ ] Determine how to maintain external consistency w.r.t channels in the next scheduled batch that haven't yet been written to disk.

cfromknecht:
This PR introduces batched operations for channel graph ingestion in order to speed up IGD. Currently a single db transaction is used to process each `channel_announcement`, `channel_update`, and `node_announcement`, which with the current graph size results in around 160k-200k transactions.

With this PR we bring this down to linear in the duration it takes the sync the graph, as we'll no make at most 2 db transactions/sec. OMM I was able to get a 4 minute graph sync using this, implying only about 500 db transactions were made in total.

So far I'm seeing a 2x speedup in IGD. The CPU and memory usage should be much lower, but I'm still in the process of tuning and gathering exact performance metrics.

## `batch` package

To accomplish this, we introduce a new `batch` package, which exposes a simple API for constructing batched operations on a bbolt database. The primary reason we require this, and can't use bbolt's native `Batch` operation, is that we maintain two caches external to bbolt, the channel cache and the reject cache.

Since both are write through caches, it is critical that we maintain external consistency whenever a write succeeds. Updating the cache should atomically add the whole batch of updates, or none of them. To achieve this, we need a mechanism for doing locking at a level above bbolt, since we don't have the luxury of reentrant locks.

Most of the batching logic is [borrowed from bbolt itself](https://github.com/etcd-io/bbolt/blob/master/db.go#L752), but we also add some slight modifications to handle any updates to the external cache under a single mutex.

There are two primary components:

```golang
// Request defines an operation that can be batched into a single bbolt
// transaction.
type Request struct {
	// Reset is called before each invocation of Update and is used to clear
	// any possible modifications to local state as a result of previous
	// calls to Update that were not comitted due to a concurrent batch
	// failure.
        //
        // NOTE: This field is optional.
	Reset func()

	// Update is applied alongside other operations in the batch.
	//
	// NOTE: This method MUST NOT acquire any mutexes.
	Update func(tx kvdb.RwTx) error

	// OnCommit is called if the batch or a subset of the batch including
	// this request all succeeded without failure. The passed error should
	// contain the result of the transaction commit, as that can still fail
	// even if none of the closures returned an error.
	//
	// NOTE: This field is optional.
	OnCommit func(commitErr error) error
}

 // Scheduler abstracts a generic batching engine that accumulates an incoming
// set of Requests, executes them, and returns the error from the operation.
type Scheduler interface {
	// Execute schedules a Request for execution with the next available
	// batch. The resulting error is returned to the caller.
	Execute(req *Request) error
}
```

Different `Scheduler`s may implement different policies for when a batch gets executed. The current one implemented, `TimeScheduler`, uses a fixed horizon after which it executes any concurrent requests added. Future schedulers may take into account metrics such as number of pending items in the batch or overall request throughput before deciding when to trigger, but that is left as an area for future research.

Using this package, it becomes simple to express our batched operations, here's an example of `BatchedAddChannelEdge`:
```golang
func (c *ChannelGraph) BatchedAddChannelEdge(edge *ChannelEdgeInfo) error {
	var alreadyExists bool
	return c.chanScheduler.Execute(&batch.Request{
		Reset: func() {
			alreadyExists = false
		},
		Update: func(tx kvdb.RwTx) error {
			err := c.addChannelEdge(tx, edge)

			// Silence ErrEdgeAlreadyExist so that the batch can
			// succeed, but propagate the error via local state.
			if err == ErrEdgeAlreadyExist {
				alreadyExists = true
				return nil
			}

			alreadyExists = false
			return err
		},
		OnCommit: func(err error) error {
			switch {
			case err != nil:
				return err
			case alreadyExists:
				return ErrEdgeAlreadyExist
			default:
				c.rejectCache.remove(edge.ChannelID)
				c.chanCache.remove(edge.ChannelID)
				return nil
			}
		},
	})
}
```

whose logic largely coincides with what we'd expect, given the unbatched variant:

```golang
	c.cacheMu.Lock()
	defer c.cacheMu.Unlock()	

 	err := kvdb.Update(c.db, func(tx *bbolt.Tx) error {	
		return c.addChannelEdge(tx, edge)	
	}, func() {})	
	if err != nil {	
		return err	
	}	

 	c.rejectCache.remove(edge.ChannelID)	
	c.chanCache.remove(edge.ChannelID)	

 	return nil
}
```

Note that in the batched variant, `c.cacheMu` is passed into the creation of the `chanScheduler` and is automatically acquired by the scheduler, so it is safe to modify the cache directly within the `OnCommit` closure.

There are a few other areas in the code base where we may wish to adopt, e.g. the `CircuitMap`, which should be a straightforward change using this primitive. I suspect this may be of interest to other projects using bbolt, we may want to consider making the `batch` package its own versioned go module.

## testing

Before proceeding in writing some more tests, I'm putting this up on travis with an temporary commit that causes the unbatched APIs to use the batched ones under the hood. This should give us a good idea that the semantics are unchanged before proceeding in writing more extensive tests to handle edge cases in the batched API.

Closes #3288

### DIFF
--- a/batch/batch.go
+++ b/batch/batch.go
@@ -1,0 +1,102 @@
+package batch
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/lightningnetwork/lnd/channeldb/kvdb"
+)
+
+// errSolo is a sentinel error indicating that the requester should re-run the
+// operation in isolation.
+var errSolo = errors.New(
+	"batch function returned an error and should be re-run solo",
+)
+
+type request struct {
+	*Request
+	errChan chan error
+}
+
+type batch struct {
+	db     kvdb.Backend
+	start  sync.Once
+	reqs   []*request
+	clear  func(b *batch)
+	locker sync.Locker
+}
+
+// trigger is the entry point for the batch and ensures that run is started at
+// most once.
+func (b *batch) trigger() {
+	b.start.Do(b.run)
+}
+
+// run executes the current batch of requests. If any individual requests fail
+// alongside others they will be retried by the caller.
+func (b *batch) run() {
+	// Clear the batch from its scheduler, ensuring that no new requests are
+	// added to this batch.
+	b.clear(b)
+
+	// If a cache lock was provided, hold it until the this method returns.
+	// This is critical for ensuring external consistency of the operation,
+	// so that caches don't get out of sync with the on disk state.
+	if b.locker != nil {
+		b.locker.Lock()
+		defer b.locker.Unlock()
+	}
+
+	// Apply the batch until a subset succeeds or all of them fail. Requests
+	// that fail will be retried individually.
+	for len(b.reqs) > 0 {
+		var failIdx = -1
+		err := kvdb.Update(b.db, func(tx kvdb.RwTx) error {
+			for i, req := range b.reqs {
+				err := req.Update(tx)
+				if err != nil {
+					failIdx = i
+					return err
+				}
+			}
+			return nil
+		}, func() {
+			for _, req := range b.reqs {
+				if req.Reset != nil {
+					req.Reset()
+				}
+			}
+		})
+
+		// If a request's Update failed, extract it and re-run the
+		// batch. The removed request will be retried individually by
+		// the caller.
+		if failIdx >= 0 {
+			req := b.reqs[failIdx]
+
+			// It's safe to shorten b.reqs here because the
+			// scheduler's batch no longer points to us.
+			b.reqs[failIdx] = b.reqs[len(b.reqs)-1]
+			b.reqs = b.reqs[:len(b.reqs)-1]
+
+			// Tell the submitter re-run it solo, continue with the
+			// rest of the batch.
+			req.errChan <- errSolo
+			continue
+		}
+
+		// None of the remaining requests failed, process the errors
+		// using each request's OnCommit closure and return the error
+		// to the requester. If no OnCommit closure is provided, simply
+		// return the error directly.
+		for _, req := range b.reqs {
+			if req.OnCommit != nil {
+				req.errChan <- req.OnCommit(err)
+			} else {
+				req.errChan <- err
+			}
+		}
+
+		return
+	}
+}
--- a/batch/interface.go
+++ b/batch/interface.go
@@ -1,0 +1,38 @@
+package batch
+
+import "github.com/lightningnetwork/lnd/channeldb/kvdb"
+
+// Request defines an operation that can be batched into a single bbolt
+// transaction.
+type Request struct {
+	// Reset is called before each invocation of Update and is used to clear
+	// any possible modifications to local state as a result of previous
+	// calls to Update that were not committed due to a concurrent batch
+	// failure.
+	//
+	// NOTE: This field is optional.
+	Reset func()
+
+	// Update is applied alongside other operations in the batch.
+	//
+	// NOTE: This method MUST NOT acquire any mutexes.
+	Update func(tx kvdb.RwTx) error
+
+	// OnCommit is called if the batch or a subset of the batch including
+	// this request all succeeded without failure. The passed error should
+	// contain the result of the transaction commit, as that can still fail
+	// even if none of the closures returned an error.
+	//
+	// NOTE: This field is optional.
+	OnCommit func(commitErr error) error
+}
+
+// Scheduler abstracts a generic batching engine that accumulates an incoming
+// set of Requests, executes them, and returns the error from the operation.
+type Scheduler interface {
+	// Execute schedules a Request for execution with the next available
+	// batch. This method blocks until the the underlying closure has been
+	// run against the databse. The resulting error is returned to the
+	// caller.
+	Execute(req *Request) error
+}
--- a/batch/scheduler.go
+++ b/batch/scheduler.go
@@ -1,0 +1,103 @@
+package batch
+
+import (
+	"sync"
+	"time"
+
+	"github.com/lightningnetwork/lnd/channeldb/kvdb"
+)
+
+// TimeScheduler is a batching engine that executes requests within a fixed
+// horizon. When the first request is received, a TimeScheduler waits a
+// configurable duration for other concurrent requests to join the batch. Once
+// this time has elapsed, the batch is closed and executed. Subsequent requests
+// are then added to a new batch which undergoes the same process.
+type TimeScheduler struct {
+	db       kvdb.Backend
+	locker   sync.Locker
+	duration time.Duration
+
+	mu sync.Mutex
+	b  *batch
+}
+
+// NewTimeScheduler initializes a new TimeScheduler with a fixed duration at
+// which to schedule batches. If the operation needs to modify a higher-level
+// cache, the cache's lock should be provided to so that external consistency
+// can be maintained, as successful db operations will cause a request's
+// OnCommit method to be executed while holding this lock.
+func NewTimeScheduler(db kvdb.Backend, locker sync.Locker,
+	duration time.Duration) *TimeScheduler {
+
+	return &TimeScheduler{
+		db:       db,
+		locker:   locker,
+		duration: duration,
+	}
+}
+
+// Execute schedules the provided request for batch execution along with other
+// concurrent requests. The request will be executed within a fixed horizon,
+// parameterizeed by the duration of the scheduler. The error from the
+// underlying operation is returned to the caller.
+//
+// NOTE: Part of the Scheduler interface.
+func (s *TimeScheduler) Execute(r *Request) error {
+	req := request{
+		Request: r,
+		errChan: make(chan error, 1),
+	}
+
+	// Add the request to the current batch. If the batch has been cleared
+	// or no batch exists, create a new one.
+	s.mu.Lock()
+	if s.b == nil {
+		s.b = &batch{
+			db:     s.db,
+			clear:  s.clear,
+			locker: s.locker,
+		}
+		time.AfterFunc(s.duration, s.b.trigger)
+	}
+	s.b.reqs = append(s.b.reqs, &req)
+	s.mu.Unlock()
+
+	// Wait for the batch to process the request. If the batch didn't
+	// ask us to execute the request individually, simply return the error.
+	err := <-req.errChan
+	if err != errSolo {
+		return err
+	}
+
+	// Obtain exclusive access to the cache if this scheduler needs to
+	// modify the cache in OnCommit.
+	if s.locker != nil {
+		s.locker.Lock()
+		defer s.locker.Unlock()
+	}
+
+	// Otherwise, run the request on its own.
+	commitErr := kvdb.Update(s.db, req.Update, func() {
+		if req.Reset != nil {
+			req.Reset()
+		}
+	})
+
+	// Finally, return the commit error directly or execute the OnCommit
+	// closure with the commit error if present.
+	if req.OnCommit != nil {
+		return req.OnCommit(commitErr)
+	}
+
+	return commitErr
+}
+
+// clear resets the scheduler's batch to nil so that no more requests can be
+// added.
+func (s *TimeScheduler) clear(b *batch) {
+	s.mu.Lock()
+	if s.b == b {
+		s.b = nil
+	}
+	s.mu.Unlock()
+}
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -1947,33 +1947,37 @@ func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
 		return err
 	}
 
+	c.updateEdgeCache(edge, isUpdate1)
+
+	return nil
+}
+
+func (c *ChannelGraph) updateEdgeCache(e *ChannelEdgePolicy, isUpdate1 bool) {
 	// If an entry for this channel is found in reject cache, we'll modify
 	// the entry with the updated timestamp for the direction that was just
 	// written. If the edge doesn't exist, we'll load the cache entry lazily
 	// during the next query for this edge.
-	if entry, ok := c.rejectCache.get(edge.ChannelID); ok {
+	if entry, ok := c.rejectCache.get(e.ChannelID); ok {
 		if isUpdate1 {
-			entry.upd1Time = edge.LastUpdate.Unix()
+			entry.upd1Time = e.LastUpdate.Unix()
 		} else {
-			entry.upd2Time = edge.LastUpdate.Unix()
+			entry.upd2Time = e.LastUpdate.Unix()
 		}
-		c.rejectCache.insert(edge.ChannelID, entry)
+		c.rejectCache.insert(e.ChannelID, entry)
 	}
 
 	// If an entry for this channel is found in channel cache, we'll modify
 	// the entry with the updated policy for the direction that was just
 	// written. If the edge doesn't exist, we'll defer loading the info and
 	// policies and lazily read from disk during the next query.
-	if channel, ok := c.chanCache.get(edge.ChannelID); ok {
+	if channel, ok := c.chanCache.get(e.ChannelID); ok {
 		if isUpdate1 {
-			channel.Policy1 = edge
+			channel.Policy1 = e
 		} else {
-			channel.Policy2 = edge
+			channel.Policy2 = e
 		}
-		c.chanCache.insert(edge.ChannelID, channel)
+		c.chanCache.insert(e.ChannelID, channel)
 	}
-
-	return nil
 }
 
 // updateEdgePolicy attempts to update an edge's policy within the relevant
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -185,17 +185,18 @@ type ChannelGraph struct {
 
 // newChannelGraph allocates a new ChannelGraph backed by a DB instance. The
 // returned instance has its own unique reject cache and channel cache.
-func newChannelGraph(db *DB, rejectCacheSize, chanCacheSize int) *ChannelGraph {
+func newChannelGraph(db *DB, rejectCacheSize, chanCacheSize int,
+	batchCommitInterval time.Duration) *ChannelGraph {
 	g := &ChannelGraph{
 		db:          db,
 		rejectCache: newRejectCache(rejectCacheSize),
 		chanCache:   newChannelCache(chanCacheSize),
 	}
 	g.chanScheduler = batch.NewTimeScheduler(
-		db.Backend, &g.cacheMu, 500*time.Millisecond,
+		db.Backend, &g.cacheMu, batchCommitInterval,
 	)
 	g.nodeScheduler = batch.NewTimeScheduler(
-		db.Backend, nil, 500*time.Millisecond,
+		db.Backend, nil, batchCommitInterval,
 	)
 	return g
 }
--- a/channeldb/graph.go
+++ b/channeldb/graph.go
@@ -18,6 +18,7 @@ import (
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/batch"
 	"github.com/lightningnetwork/lnd/channeldb/kvdb"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing/route"
@@ -177,16 +178,26 @@ type ChannelGraph struct {
 	cacheMu     sync.RWMutex
 	rejectCache *rejectCache
 	chanCache   *channelCache
+
+	chanScheduler batch.Scheduler
+	nodeScheduler batch.Scheduler
 }
 
 // newChannelGraph allocates a new ChannelGraph backed by a DB instance. The
 // returned instance has its own unique reject cache and channel cache.
 func newChannelGraph(db *DB, rejectCacheSize, chanCacheSize int) *ChannelGraph {
-	return &ChannelGraph{
+	g := &ChannelGraph{
 		db:          db,
 		rejectCache: newRejectCache(rejectCacheSize),
 		chanCache:   newChannelCache(chanCacheSize),
 	}
+	g.chanScheduler = batch.NewTimeScheduler(
+		db.Backend, &g.cacheMu, 500*time.Millisecond,
+	)
+	g.nodeScheduler = batch.NewTimeScheduler(
+		db.Backend, nil, 500*time.Millisecond,
+	)
+	return g
 }
 
 // Database returns a pointer to the underlying database.
@@ -440,15 +451,17 @@ func (c *ChannelGraph) SetSourceNode(node *LightningNode) error {
 // AddLightningNode adds a vertex/node to the graph database. If the node is not
 // in the database from before, this will add a new, unconnected one to the
 // graph. If it is present from before, this will update that node's
-// information. Note that this method is expected to only be called to update
-// an already present node from a node announcement, or to insert a node found
-// in a channel update.
+// information. Note that this method is expected to only be called to update an
+// already present node from a node announcement, or to insert a node found in a
+// channel update.
 //
 // TODO(roasbeef): also need sig of announcement
 func (c *ChannelGraph) AddLightningNode(node *LightningNode) error {
-	return kvdb.Update(c.db, func(tx kvdb.RwTx) error {
-		return addLightningNode(tx, node)
-	}, func() {})
+	return c.nodeScheduler.Execute(&batch.Request{
+		Update: func(tx kvdb.RwTx) error {
+			return addLightningNode(tx, node)
+		},
+	})
 }
 
 func addLightningNode(tx kvdb.RwTx, node *LightningNode) error {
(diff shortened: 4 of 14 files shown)


channeldb/graph.go
```diff
@@ -173,16 +174,22 @@ type ChannelGraph struct {
...
 // returned instance has its own unique reject cache and channel cache.
 func newChannelGraph(db *DB, rejectCacheSize, chanCacheSize int) *ChannelGraph {
-	return &ChannelGraph{
+	g := &ChannelGraph{
 		db:          db,
 		rejectCache: newRejectCache(rejectCacheSize),
 		chanCache:   newChannelCache(chanCacheSize),
 	}
+	g.chanScheduler = batch.NewTimeScheduler(db.DB, &g.cacheMu, 10*time.Millisecond)
+	g.nodeScheduler = batch.NewTimeScheduler(db.DB, nil, 10*time.Millisecond)
```
Crypt-iQ: I was able to set this slightly higher and still achieve the same sync time of about 2 minutes (granted, the peers I synced with on different runs were very fast).  I used as high as 8 seconds, less db txn's.  What do you think?  I think perhaps the two schedulers could be staggered at different intervals so as to not cause db contention?

Crypt-iQ: I do see some memory usage with `dereference` calls when growing the mmap since these are large writes if 8 second batch time is used (~10MB) so a case could be made for `InitialMmapSize` as discussed in #3278 , but this is nothing compared to the `GetBlock` calls (6 GB +) made in the router.

Crypt-iQ: I also see a reduction from the 10ms scheduler at 845MB:
```
ROUTINE ======================== github.com/lightningnetwork/lnd/batch.(*batch).trigger in /Users/nsa/go/src/github.com/lightningnetwork/lnd/batch/batch.go
         0   845.71MB (flat, cum)  7.09% of Total
         .          .     27:}
         .          .     28:
         .          .     29:// trigger is the entry point for the batch and ensures that run is started at
         .          .     30:// most once.
         .          .     31:func (b *batch) trigger() {
         .   845.71MB     32:	b.start.Do(b.run)
         .          .     33:}
         .          .     34:
         .          .     35:// run executes the current batch of requests. If any individual requests fail
         .          .     36:// alongside others they will be retried by the caller.
         .          .     37:func (b *batch) run() {
```

to 269MB with the 8s scheduler:
```
ROUTINE ======================== github.com/lightningnetwork/lnd/batch.(*batch).trigger in /Users/nsa/go/src/github.com/lightningnetwork/lnd/batch/batch.go
         0   269.59MB (flat, cum)  1.80% of Total
         .          .     27:}
         .          .     28:
         .          .     29:// trigger is the entry point for the batch and ensures that run is started at
         .          .     30:// most once.
         .          .     31:func (b *batch) trigger() {
         .   269.59MB     32:	b.start.Do(b.run)
         .          .     33:}
         .          .     34:
         .          .     35:// run executes the current batch of requests. If any individual requests fail
         .          .     36:// alongside others they will be retried by the caller.
         .          .     37:func (b *batch) run() {
```

cfromknecht: @Crypt-iQ yes the 10ms time is just a temporary commit so that i could drop in the batched version into the regular API and verify it has the same semantics w/o having to modify any timeouts.

In practice we should be using a timeout on the order of seconds. I did my testing with a duration of 1 second and it performed quite well. We could probably do 8 seconds as well, though this also delays the time till first write pretty significantly. Most of the writes were taking 10-20ms, so only about 2-4% of one core would be dedicated to the db txn itself if we go with 1 second, and even lower with 8.

The main thing left to do is benchmark this on mobile and get a sense for how these parameters behave on different architectures, just haven't gotten around to doing so :)

batch/batch.go
```diff
@@ -0,0 +1,96 @@
...
+	start  sync.Once
+	reqs   []*request
+	clear  func(b *batch)
+	locker sync.Locker
+}
+
+// trigger is the entry point for the batch and ensures that run is started at
+// most once.
+func (b *batch) trigger() {
+	b.start.Do(b.run)
```
joostjager: Can it happen that `trigger` is called more than once? If not, does this `sync.Once` camouflage potential bugs?

batch/batch.go
```diff
@@ -0,0 +1,96 @@
...
+type request struct {
+	*Request
+	errChan chan error
+}
+
+type batch struct {
+	db     *bbolt.DB
+	start  sync.Once
+	reqs   []*request
+	clear  func(b *batch)
```
joostjager: Would it be better to make this `clear func()`? The scheduler has the instance already and could use it in a lambda.

batch/scheduler.go
```diff
@@ -0,0 +1,82 @@
...
+	// Add the request to the current batch. If the batch has been cleared
+	// or no batch exists, create a new one.
+	s.mu.Lock()
+	if s.b == nil {
+		s.b = &batch{
+			db:     s.db,
+			clear:  s.clear,
+			locker: s.locker,
+		}
+		time.AfterFunc(s.duration, s.b.trigger)
```
joostjager: bbolt also has a max batch size. Is that needed here too?

channeldb/graph.go
```diff
@@ -173,16 +174,22 @@ type ChannelGraph struct {
...
 // returned instance has its own unique reject cache and channel cache.
 func newChannelGraph(db *DB, rejectCacheSize, chanCacheSize int) *ChannelGraph {
-	return &ChannelGraph{
+	g := &ChannelGraph{
 		db:          db,
 		rejectCache: newRejectCache(rejectCacheSize),
 		chanCache:   newChannelCache(chanCacheSize),
 	}
+	g.chanScheduler = batch.NewTimeScheduler(db.DB, &g.cacheMu, time.Second)
+	g.nodeScheduler = batch.NewTimeScheduler(db.DB, nil, time.Second)
```
joostjager: If there is no node cache, couldn't this just use the normal bbolt batching?

channeldb/graph.go
```diff
@@ -439,12 +446,29 @@ func (c *ChannelGraph) SetSourceNode(node *LightningNode) error {
...
+// BatchedAddLightningNode adds a vertex/node to the graph database. If the node
+// is not in the database from before, this will add a new, unconnected one to
+// the graph. If it is present from before, this will update that node's
+// information. Note that this method is expected to only be called to update an
+// already present node from a node announcement, or to insert a node found in a
+// channel update.
+func (c *ChannelGraph) BatchedAddLightningNode(node *LightningNode) error {
+	return c.nodeScheduler.Execute(&batch.Request{
+		Update: func(tx *bbolt.Tx) error {
+			return addLightningNode(tx, node)
```
joostjager: Should some errors be suppressed to let the batch go through?

channeldb/graph.go
```diff
@@ -1905,13 +1969,57 @@ func delChannelEdge(edges, edgeIndex, chanIndex, zombieIndex,
...
+// updated, otherwise it's the second node's information. The node ordering is
+// determined by the lexicographical ordering of the identity public keys of the
+// nodes on either side of the channel.
+func (c *ChannelGraph) BatchedUpdateEdgePolicy(edge *ChannelEdgePolicy) error {
+	var (
+		isUpdate1    bool
+		edgeNotFound bool
+	)
+	return c.chanScheduler.Execute(&batch.Request{
+		Update: func(tx *bbolt.Tx) error {
```
joostjager: Is it an option to add two `error` return parameters, to prevent this pattern with the local variable?

discovery/gossiper.go
```diff
@@ -875,9 +874,7 @@ func (d *AuthenticatedGossiper) networkHandler() {
 
 	// We'll use this validation to ensure that we process jobs in their
 	// dependency order during parallel validation.
-	validationBarrier := routing.NewValidationBarrier(
-		runtime.NumCPU()*4, d.quit,
-	)
+	validationBarrier := routing.NewValidationBarrier(1000, d.quit)
```
joostjager: Could use some explanation in the commit message?

batch/scheduler.go
Crypt-iQ: This implementation uses bbolt's `Update` function to emulate batching, so the actual `Batch` call isn't called here.

batch/interface.go
```diff
@@ -0,0 +1,32 @@
...
+
+	// Solo is applied if this request's Update failed, and was asked to
+	// retry outside of the batch.
+	Solo func() error
+}
+
+// Scheduler abstracts a generic batching engine that accumulates an incoming
+// set of Requests, executes them, and returns the error from the operation.
+type Scheduler interface {
+	// Execute schedules a Request for execution with the next available
```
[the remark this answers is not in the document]
valentinewallace: Wondering in case the time-based scheduler increases its duration, in which case the callers of this function may not want to block until it completes :thinking:

nicolasburtey:
(what does IGD stand for?)

Roasbeef:
> (what does IGD stand for?)

Initial graph download. This PR is intended to speed up syncing the graph from the p2p network.

halseth:
@cfromknecht LMK when the rebased version is ready!

batch/batch.go
```diff
@@ -0,0 +1,96 @@
...
+			continue
+		}
+
+		// None of the remaining requests failed, process the errors
+		// using each request's OnSuccess closure and return the error
+		// to the requester. If no OnSuccess closure is provided, simply
+		// return the error directly.
+		for _, req := range b.reqs {
+			if req.OnSuccess != nil {
+				req.errChan <- req.OnSuccess(err)
```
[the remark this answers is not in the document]
cfromknecht: this is a little subtle, but this value can still be non-nil. it's true that at this point, none of the requests fail _inside_ the db txn. however, it's still possible for the txn itself to fail. that error is passed into `OnSuccess` so that any errors in persisting the changes can be returned at the call sites

batch/interface.go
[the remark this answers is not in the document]
cfromknecht: good call, i've updated the godoc. yes, the latency on the commit could be higher than before. for the initial use case though we already execute all of our concurrent requests within goroutines (during channel graph validation)

batch/scheduler.go
cfromknecht: that sounds like a worthy optimization, to initiate the batch as soon as we have X requests. i'd vote for keeping the simpler approach in the initial version, and then circle back if we find it to be a bottleneck

batch/batch.go
```diff
@@ -0,0 +1,96 @@
...
+	"sync"
+
+	"github.com/coreos/bbolt"
+)
+
+// errSolo is a sentinel error indicating that the requester should re-run the
+// operation in isolation.
+var errSolo = errors.New(
+	"batch function returned an error and should be re-run solo",
+)
```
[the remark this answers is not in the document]
cfromknecht: yeah a more descriptive name might be `errRunSolo`, or `errRequestFailedInBatch`, but i went for keeping it inline with bbolt 🤷‍♂️

channeldb/graph.go
cfromknecht: `addLightningNode` only produces critical errors, e.g. serialization, bucket structure, etc, so we don't need to catch anything afaict

channeldb/graph.go
```diff
@@ -439,18 +446,33 @@ func (c *ChannelGraph) SetSourceNode(node *LightningNode) error {
...
 // graph. If it is present from before, this will update that node's
-// information. Note that this method is expected to only be called to update
-// an already present node from a node announcement, or to insert a node found
-// in a channel update.
+// information. Note that this method is expected to only be called to update an
+// already present node from a node announcement, or to insert a node found in a
+// channel update.
 //
 // TODO(roasbeef): also need sig of announcement
 func (c *ChannelGraph) AddLightningNode(node *LightningNode) error {
```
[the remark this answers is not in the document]
cfromknecht: at some point, yes possibly. currently they are still useful esp. in unit tests, since we wouldn't have to set custom timeouts and stuff for tests. it would be a pretty big investment to do so tho because most of the test infra is already written with the one-off methods

cfromknecht:
@Crypt-iQ @halseth PTAL!

cfromknecht:
> My main question related to this PR is whether this is the right optimization path to take.

> How would this PR compare to a different, quite extreme, approach: The graph only exists in memory. It is updated in memory and queried from memory (which we want at some point anyway to speed up path finding). Existing caches can be removed / merged with the in-memory graph.

IMO yes something like that would be the end goal. However, it is quite a large departure from how the existing architecture works, this is only meant to be an incremental improvement over what we have today.

channeldb/graph.go
```diff
@@ -177,16 +178,22 @@ type ChannelGraph struct {
...
 // newChannelGraph allocates a new ChannelGraph backed by a DB instance. The
 // returned instance has its own unique reject cache and channel cache.
 func newChannelGraph(db *DB, rejectCacheSize, chanCacheSize int) *ChannelGraph {
-	return &ChannelGraph{
+	g := &ChannelGraph{
 		db:          db,
 		rejectCache: newRejectCache(rejectCacheSize),
 		chanCache:   newChannelCache(chanCacheSize),
 	}
+	g.chanScheduler = batch.NewTimeScheduler(db.Backend, &g.cacheMu, time.Second)
```
halseth: usage of the cache mutex here should have some comments IMO

batch/batch.go
```diff
@@ -0,0 +1,96 @@
...
+func (b *batch) trigger() {
+	b.start.Do(b.run)
+}
+
+// run executes the current batch of requests. If any individual requests fail
+// alongside others they will be retried by the caller.
+func (b *batch) run() {
+	// Clear the batch from its scheduler, ensuring that no new requests are
+	// added to this batch.
+	b.clear(b)
```
halseth, line 40: I find this interleaving of batch and scheduler a bit confusing. Could we instead let the scheduler accumulate requests, when the timer ticks it will assemble the requests into a batch and trigger it?

batch/interface.go
```diff
@@ -0,0 +1,32 @@
...
+	// Update is applied alongside other operations in the batch.
+	//
+	// NOTE: This method MUST NOT acquire any mutexes.
+	Update func(tx kvdb.RwTx) error
+
+	// OnSuccess is applied if the batch or a subset of the batch including
+	// this request all succeeded without failure.
+	//
+	// NOTE: This field is optional.
+	OnSuccess func(err error) error
```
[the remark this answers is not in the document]
halseth: `OnResult` perhaps?

halseth (changes requested):

chpio:
would this stop lnd from trashing ssds with >300gb of writes per day?

Crypt-iQ:
> 
> 
> would this stop lnd from trashing ssds with >300gb of writes per day?

More information would be necessary (you could open an issue). This should cut down on a lot of write contention though.

channeldb/graph.go
```diff
@@ -177,16 +178,22 @@ type ChannelGraph struct {
...
+	g.nodeScheduler = batch.NewTimeScheduler(db.Backend, nil, time.Second)
```
halseth: Would be useful to have this being configurable (could make it a dev only flag)

halseth:
Ran some benchmarks using Neutrino and `assumechanvalid`, this is a significant speedup! 

A full graph sync on mainnet (about 31k channels as we speak) went down from about 32 minutes to 5m 45s on my setup! 😮 

However, I experimented a bit with different batch intervals, and setting it to 500ms instead of 1s actually gave me a 4m 35s run! So I think it would make sense to expose this as a configuration option, at least until we know more about what is the "right" number here.

Crypt-iQ:
> Ran some benchmarks using Neutrino and `assumechanvalid`, this is a significant speedup!
> 
> A full graph sync on mainnet (about 31k channels as we speak) went down from about 32 minutes to 5m 45s on my setup! 😮
> 
> However, I experimented a bit with different batch intervals, and setting it to 500ms instead of 1s actually gave me a 4m 35s run! So I think it would make sense to expose this as a configuration option, at least until we know more about what is the "right" number here.

Could you get some pprof output comparisons? There is also a memory gain to be had and that would be especially beneficial to neutrino users.

channeldb/graph.go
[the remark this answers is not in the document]
cfromknecht: can't bc it references `&g.cacheMu`

discovery/gossiper.go
```diff
@@ -904,9 +903,7 @@ func (d *AuthenticatedGossiper) networkHandler() {
...
```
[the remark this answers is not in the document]
cfromknecht: it's just meant to high enough to facilitate processing of many concurrent requests. the actual number a node may be able to do probably depends more on hardware than anything. we can't make it unlimited since it's a semaphore, and it will be a DOS vector. we'd end up reading all of the incoming gossip messages at once and performing all of the dependent operations/requests at once.

batch/batch.go
Roasbeef: Still getting through this, but possibly couldn't a new batch instance be created that already removes the scheduled items from the scheduler? Seems like an odd "up call" in this context.

batch/interface.go
```diff
@@ -0,0 +1,34 @@
...
+	OnCommit func(commitErr error) error
+
+	// Solo is applied if this request's Update failed, and was asked to
+	// retry outside of the batch.
+	Solo func() error
+}
+
+// Scheduler abstracts a generic batching engine that accumulates an incoming
+// set of Requests, executes them, and returns the error from the operation.
+type Scheduler interface {
```
Roasbeef, line 28: Solid abstraction, could see us using this in many other areas as well possibly.

batch/scheduler.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+
+	// Otherwise, run the request on its own.
+	return req.Solo()
+}
+
+// clear resets the scheduler's batch to nil so that no more requests can be
+// added.
+func (s *TimeScheduler) clear(b *batch) {
```
Roasbeef, line 78: Ah ok, I see the interaction here now with the underlying batch...

channeldb/graph.go
```diff
@@ -432,12 +452,29 @@ func (c *ChannelGraph) SetSourceNode(node *LightningNode) error {
 	})
 }
 
+// BatchedAddLightningNode adds a vertex/node to the graph database. If the node
+// is not in the database from before, this will add a new, unconnected one to
+// the graph. If it is present from before, this will update that node's
+// information. Note that this method is expected to only be called to update an
+// already present node from a node announcement, or to insert a node found in a
+// channel update.
+func (c *ChannelGraph) BatchedAddLightningNode(node *LightningNode) error {
```
Roasbeef, line 461: Nice approach of adding wrapped batch variants o each of the calls!

channeldb/graph.go
```diff
@@ -30,7 +30,7 @@ const (
 	// This presents a tradeoff between the number concurrent operations
 	// that can to be applied by a single database transaction and the
 	// overall latency of the changes being applied.
-	GraphBatchCommitInterval = time.Second
+	GraphBatchCommitInterval = 10 * time.Millisecond
```
Roasbeef, line 33: This commit (the one that uses a 10 ms commit interval) should be scrapped right? 

(based on johan's testing thus far)

Roasbeef (commented):
Solid PR! Also a relatively compact diff. Completed my first "new" pass since this PR was resurrected. Will do at least one additional pass. Still need to wrap my head around the way the batch itself interacts with the scheduler in a concurrent safe manner...

batch/batch.go
cfromknecht: fwiw this design is modeled off bbolt's batch/scheduler. the notation looks a little strange, but the `clear` method is a reference to a member function of the scheduler. the batch `b` is passed in so that the scheduler can determine whether this batch matches the scheduler's current batch, which we do because the execution order of two batches is undefined

channeldb/graph.go
```diff
@@ -560,12 +584,52 @@ func (c *ChannelGraph) deleteLightningNode(nodes kvdb.RwBucket,
...
+				return err
+			case alreadyExists:
+				return ErrEdgeAlreadyExist
+			default:
+				c.rejectCache.remove(edge.ChannelID)
+				c.chanCache.remove(edge.ChannelID)
+				return nil
+			}
+		},
+		Solo: func() error {
```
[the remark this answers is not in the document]
cfromknecht: Good suggestion, removed!

channeldb/graph.go
cfromknecht: yes, this is just used so that we can test that the batched versions pass all of our unit tests that are currently written with the non-batched versions. the main question we have to decide is whether or not to keep the unbatched versions at all, though it could slow down the unit tests somewhat and possibly lead to flakes since the current timing assumptions may be invalidated due to the additional latency.

another option is to pass in this lower commit latency when the channeldb is instantiated, so that we can use a lower value in testing. @halseth also suggested elsewhere we make this configurable, tho i'm still not sure it needs to be exposed as a CLI/conf option

batch/interface.go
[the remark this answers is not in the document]
cfromknecht: went with `OnCommit`

Roasbeef (approved):
LGTM 🥮

(pending resolution of the last commit) 

I think we can either use a build tag to force the lower interval during the tests, or make it a config option and have all the tests specify their own value. No strong feelings in either direction.

channeldb/graph.go
halseth: Can we expose it as a conf option for dev or expermental builds only? And default it to 500 ms or whatever

channeldb/graph.go
```diff
@@ -567,12 +601,52 @@ func (c *ChannelGraph) deleteLightningNode(nodes kvdb.RwBucket,
...
+			err := c.addChannelEdge(tx, edge)
+
+			// Silence ErrEdgeAlreadyExist so that the batch can
+			// succeed, but propagate the error via local state.
+			if err == ErrEdgeAlreadyExist {
+				alreadyExists = true
+				return nil
+			}
+
+			alreadyExists = false
```
halseth, line 624: nit: not needed?

channeldb/graph.go
```diff
@@ -1924,13 +1998,58 @@ func delChannelEdge(edges, edgeIndex, chanIndex, zombieIndex,
...
+			isUpdate1, err = updateEdgePolicy(tx, edge)
+
+			// Silence ErrEdgeNotFound so that the batch can
+			// succeed, but propagate the error via local state.
+			if err == ErrEdgeNotFound {
+				edgeNotFound = true
+				return nil
+			}
+
+			edgeNotFound = false
```
halseth, line 2014: nit: not needed

channeldb/graph.go
```diff
@@ -648,20 +646,7 @@ func (c *ChannelGraph) BatchedAddChannelEdge(edge *ChannelEdgeInfo) error {
...
-	}, func() {})
-	if err != nil {
-		return err
-	}
-
-	c.rejectCache.remove(edge.ChannelID)
-	c.chanCache.remove(edge.ChannelID)
-
-	return nil
+	return c.BatchedAddChannelEdge(edge)
```
halseth, line 649: is this change to be reverted or no? If makes sense to _only_ have `AddChannelEdge` and let that be the batched version?

batch/scheduler.go
```diff
@@ -0,0 +1,95 @@
...
+	s.mu.Unlock()
+
+	// Wait for the batch to process the request. If the batch didn't
+	// ask us to execute the request individually, simply return the error.
+	err := <-req.errChan
+	if err != errSolo {
+		return err
+	}
+
+	if s.locker != nil {
```
halseth, line 72: add a comment here

batch/batch.go
halseth: Not sure if we should copy bbolt if it is not to our standards 😛  Just looks like this could be made much simpler, but also not a blocker for me.

halseth (changes requested):
Looks pretty good, still think we should make the interval configurable, should also help setting it low during tests!

batch/batch.go
cfromknecht: not seeing why it doesn't meet our standards, how would you suggest making it simpler?

Roasbeef:
Ping on this @cfromknecht, just needs the config flag now AFAIU.

batch/batch.go
halseth: See my first comment. I don't think it's a blocker, just saying that we could make this code more readable by not model the design off bbolt.

Roasbeef:
Config check is failing: 
```
Command line flag --db.batch-commit-interval not added to sample-lnd.conf
```

lncfg/db.go
```diff
@@ -3,20 +3,24 @@ package lncfg
...
+	BoltBackend                = "bolt"
+	EtcdBackend                = "etcd"
+	DefaultBatchCommitInterval = 500 * time.Millisecond
 )
 
 // DB holds database configuration for LND.
 type DB struct {
 	Backend string `long:"backend" description:"The selected database backend."`
 
+	BatchCommitInterval time.Duration `long:"batch-commit-interval" description:"The maximum duration the channel graph batch schedulers will wait before attempting to commit a batch of pending updates. This can be tradeoff database contenion for commit latency."`
```
halseth, line 22: alternatively we could add this behind a build flag to avoid exposing it to "normal" users (would also not require to update the sample config)

channeldb/graph.go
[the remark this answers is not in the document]
cfromknecht: ended up making the timeout configurable from the config file, so everything passes through the original methods!

PR #3369, opened 2019-08-02
htlcswitch/link: remove channel_reestablish timeout

cfromknecht:
An alternative approach to #3361. Instead of increasing the timeout this removes it all together. This eliminates the potential for escalating the timeout error into a protocol error.

@rustyrussell you requested that we don't modify the error string, is it compatible with the fix if we just don't send the error at all?


halseth:
Interesting turn of events.

cfromknecht:
> If for some reason that ChanSync message is never coming, is there no downside to keeping the connection open indefinitely?

with these changes, where `EligibleToForward` is conditioned on channel reestablish, no there shouldn't be. prior i believe this logic was added to handle the case where the remote party force closed, in which case they wouldn't send channel reestablish and only learn about it after the txn confirms

> Suppose EligibleToForward is checked and returns true, then the link flaps and gets stuck in sync and then the switch forwards the htlc to that link. Does that gap exist and is it worth looking at?

yes it is, tho it is also a gap that exists today, the solution is to have the mailboxes fail htlcs as their cltv expires. this pr slightly mitigates that by making sure channel reestablish is received before forwarding, but doesn't do anything for the case where a subsequent connection fails to reestablish

htlcswitch/test_utils.go
```diff
@@ -1335,3 +1377,49 @@ func timeout(t *testing.T) func() {
 		close(done)
 	}
 }
+
+// WaitPredicate is a helper test function that will wait for a timeout period
+// of time until the passed predicate returns true. This function is helpful as
+// timing doesn't always line up well when running integration tests with
+// several running lnd nodes. This function gives callers a way to assert that
+// some property is upheld within a particular time frame.
+func WaitPredicate(pred func() bool, timeout time.Duration) error {
```
halseth: `lntest/testutils`? 🤔

cfromknecht: import cycle because both packages import `routing`. yes this really should go in it's own utils package with no deps

cfromknecht: done

htlcswitch/mock.go
```diff
@@ -175,7 +175,6 @@ func initSwitchWithDB(startingHeight uint32, db *channeldb.DB) (*Switch, error)
 		FwdEventTicker:        ticker.NewForce(DefaultFwdEventInterval),
 		LogEventTicker:        ticker.NewForce(DefaultLogInterval),
 		AckEventTicker:        ticker.NewForce(DefaultAckInterval),
-		NotifyActiveChannel:   func(wire.OutPoint) {},
 		NotifyInactiveChannel: func(wire.OutPoint) {},
```
joostjager: Should inactive notification be moved to `Link.Stop` for consistency?

htlcswitch/test_utils.go
```diff
@@ -1229,6 +1252,25 @@ func (n *twoHopNetwork) start() error {
...
+		err := utils.WaitNoError(func() error {
+			if l.EligibleToForward() {
+				return nil
+			}
+			return fmt.Errorf("%s channel link not eligible", n)
+		}, 3*time.Second)
+		if err != nil {
+			return err
+		}
+	}
```
joostjager: Extract to method?

htlcswitch/link.go
```diff
@@ -870,6 +874,11 @@ func (l *channelLink) htlcManager() {
 	log.Infof("HTLC manager for ChannelPoint(%v) started, "+
 		"bandwidth=%v", l.channel.ChannelPoint(), l.Bandwidth())
 
+	// Funding locked has already been received, so we'll go ahead and
+	// deliver the active channel notification since EligibleToForward
+	// returns true now that the link has been added to the switch.
+	l.cfg.NotifyActiveChannel(*l.ChannelPoint())
```
halseth: I don't see how this changes anything. Isn't this more or less equivalent to this being called after `AddLink` has been called?

htlcswitch/link.go
```diff
@@ -954,6 +964,14 @@ func (l *channelLink) htlcManager() {
 		}
 	}
 
+	// We've successfully reestablished the channel, mark it as such to
+	// allow the switch to forward HTLCs in the outbound direction.
+	l.markReestablished()
+
+	// Now that we've received both funding locked and channel reestablish,
+	// we can go ahead and send the active channel notification.
+	l.cfg.NotifyActiveChannel(*l.ChannelPoint())
```
halseth: Ah, I see. It is moved here.

halseth: Could maybe `defer NotifyInactive()` here?

htlcswitch/test_utils.go
cfromknecht: done

htlcswitch/link.go
cfromknecht: good call, w/o doing so it could have caused us to send an inactive ntfn when no corresponding active ntfn had been sent. how that it's deferred, they should be symmetric.

lntest/utils/wait.go
```diff
@@ -0,0 +1,78 @@
+package utils
+
+import (
+	"fmt"
+	"time"
+)
+
+// WaitPredicate is a helper test function that will wait for a timeout period
```
Roasbeef: I don't think we really need a utils package here. What's the downside of them just living in the main `lntest` package? Or is this an attempt to resolve an import cycle/ 

The module import path is also much clearer as `lntest` vs `utils`. `utils` as a package name is ambiguous (what type of utils, just tests, everything?). In general `utils` is a bad package name, as there's no clear stance of exactly _what_ goes into the package.

Roasbeef: If it's an import cycle thing, then there's no harm in just copying over the exact function that's needed elsewhere to address the import cycle. Each of these functions are no more than 20 lines long, so there's no extreme duplication or multiplication of maintenance costs by copying the function needed.

cfromknecht: yes there's an import cycle because `lntest` depends on `routing`. the functions were originally copied here, but decided it best to resolve the import cycle now so that we don't end up an unbounded number copies as we continue to use them throughout the code base.

> In general utils is a bad package name, as there's no clear stance of exactly what goes into the package.

The package is `lntest/utils`, not just `utils`. idt it's unclear what kind of utils would go there given the scoping. If an import collision arrises it's trivial to rename it within the scope of the file. 

`testutils` is imo too long for something used so prevalently and is ideally an invisible package; it's also redundant in the scope of `lntest/testutils`. i don't really have a better suggestion, but i think _some_ package should be created here

cfromknecht: perhaps a package called `wait`?

Roasbeef: Ah I see, since it refs `routing.BlockPadding`.... 


>  idt it's unclear what kind of utils would go there given the scoping

I'd agree if devs were forced to type out `lntest/util.Func` each time, but instead one sees the `utils` package being referenced and is left to wonder exactly what else that package provides or where it's being import from until they scroll back up to the main import list. 

> perhaps a package called `wait`?

I'm cool with `wait`.

halseth: `lntest/wait` sgtm

joostjager: My opinion about this:
* I don't like copy/paste
* A package for wait functions will prompt a rename when a non-wait shared test function is added. Or another package.
* I can live with `testutils` as a package name, even though a bit long

How about putting the generic wait utils in `lntest` and moving things that are currently in `lntest` to a more descriptive package like `lntest/harness`?

Btw, I don't see the dependency of `lntest` on `routing`, just that `lntest/itest` depends on `routing`.

cfromknecht: okay, went with `lntest/wait`, can always change it later if we have better ideas.

> Btw, I don't see the dependency of lntest on routing, just that lntest/itest depends on routing.

`lntest` depends on `routerrpc` which depends on `routing`

> A package for wait functions will prompt a rename when a non-wait shared test function is added. Or another package.

I think this is what we want (adding a new package), keeping them small will keep the deps small and minimize import cycles going forward.

cfromknecht: > I'd agree if devs were forced to type out lntest/util.Func each time, but instead one sees the utils package being referenced and is left to wonder exactly what else that package provides or where it's being import from until they scroll back up to the main import list.

I presume devs would only have to do this once, and then remember what the package is. for me scrolling to the top of the file once is an acceptable tradeoff

PR #3372, opened 2019-08-05
routing: process payment successes in mission control

joostjager:
This PR modifies `paymentLifecycle` so that it not only feeds failures into mission control, but successes as well. This allows for more accurate probability estimates. Previously, the success probability for a successful pair and a pair with no history was equal. There was no force that pushed towards previously successful routes.

A fixed success probability is introduced for node pairs that proved to be successful on the previous payment attempt. This value could be exposed as a parameter in the future. 

Furthermore the default a priori success probability (defines the probability for node pairs with no history) is lowered from 95% to a more realistic 60%. This wasn't possible before, because it would discourage longer routes too much. With the newly introduced "success success probability", only long routes consisting of unknown node pairs are avoided.

No database migration is needed, because we persist only the raw payment results and not the interpretation.


routing/missioncontrol.go
```diff
@@ -383,6 +392,31 @@ func (m *MissionControl) ReportPaymentFail(paymentID uint64, rt *route.Route,
...
+func (m *MissionControl) ReportPaymentSuccess(paymentID uint64,
+	rt *route.Route) error {
+
+	timestamp := m.now()
+
+	result := &paymentResult{
+		timeFwd:   timestamp,
+		timeReply: timestamp,
+		id:        paymentID,
+		success:   true,
```
wpaulino: Could we instead use the existing failure fields of `paymentResult` to determine if the payment succeeded?

routing/result_interpretation_test.go
```diff
@@ -45,18 +45,44 @@ var (
 	}
 )
 
+// TestResultInterpretationSuccess asserts that a successful payment result is
+// interpreted correctly.
+func TestResultInterpretationSuccess(t *testing.T) {
```
wpaulino: We should also add a test case (if there isn't one already) that ensures pairs are marked as successful up to the error source for a failed payment.

routing/missioncontrol.go
```diff
@@ -56,7 +60,7 @@ const (
 // into the path finding process for subsequent payment attempts.
 type MissionControl struct {
 	// lastPairFailure tracks the last payment failure per node pair.
-	lastPairFailure map[DirectedNodePair]pairFailure
+	lastPairFailure map[DirectedNodePair]timedPairResult
```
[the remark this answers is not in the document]
joostjager: yes, renamed

routing/missioncontrol.go
```diff
@@ -383,6 +392,31 @@ func (m *MissionControl) ReportPaymentFail(paymentID uint64, rt *route.Route,
...
+		id:        paymentID,
+		success:   true,
+		route:     rt,
+	}
+
+	_, err := m.processPaymentResult(result)
+	return err
+}
+
+func (m *MissionControl) processPaymentResult(result *paymentResult) (
```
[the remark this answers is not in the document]
joostjager: added

routing/result_interpretation.go
```diff
@@ -290,7 +347,23 @@ func (i *interpretedResult) failPairBalance(
 
 	pair, amt := getPair(rt, channelIdx)
 
-	i.pairResults[pair] = amt
+	i.pairResults[pair] = pairResult{
+		minPenalizeAmt: amt,
+	}
+}
+
+// failPairRange marks the node pairs from node fromIdx to node toIdx as
```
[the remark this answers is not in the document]
joostjager: fixed

routing/missioncontrol.go
joostjager: I don't think there is an existing field that can serve this purpose. `failureSourceIdx` can be nil for failure with an unknown source and `failure` can be nil for unknown failure messages.

routing/result_interpretation_test.go
joostjager: Existing fail test extended

routing/missioncontrol_test.go
```diff
@@ -112,6 +112,16 @@ func (ctx *mcTestContext) reportFailure(t time.Time,
 	)
 }
 
+// reportSuccess reports a success by using a test route.
+func (ctx *mcTestContext) reportSuccess(t time.Time) {
```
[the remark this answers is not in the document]
joostjager: it is a left-over, removed.

routing/missioncontrol.go
```diff
@@ -138,15 +140,19 @@ type MissionControlPairSnapshot struct {
...
 	// MinPenalizeAmt is the minimum amount for which the channel will be
 	// penalized.
 	MinPenalizeAmt lnwire.MilliSatoshi
 
 	// SuccessProb is the success probability estimation for this channel.
 	SuccessProb float64
+
+	// LastAttemptSuccessful indicates whether the last payment attempt
+	// through this pair was successful.
+	LastAttemptSuccessful bool
```
wpaulino: I don't think this is being set when we retrieve the snapshot?

joostjager: Good catch, fixed

joostjager:
Converted to parameterized tests and added one for incorrect payment details.

wpaulino:
Currently failing travis though:
> --- FAIL: TestMissionControl (0.01s)
    missioncontrol_test.go:175: unexpected number of channels

routing/result_interpretation_test.go
```diff
@@ -70,13 +86,67 @@ var resultTestCases = []resultTestCase{
...
 
 		expectedResult: &interpretedResult{
-			pairResults: map[DirectedNodePair]lnwire.MilliSatoshi{
-				getTestPair(0, 1): 0,
-				getTestPair(1, 0): 0,
-				getTestPair(1, 2): 0,
-				getTestPair(2, 1): 0,
-				getTestPair(2, 3): 0,
-				getTestPair(3, 2): 0,
+			pairResults: map[DirectedNodePair]pairResult{
```
Roasbeef: Shouldn't all nodes but the final have been marked as a success since they successfully forwarded?

routing/missioncontrol.go
```diff
@@ -271,12 +277,16 @@ func (m *MissionControl) getPairProbability(fromNode,
...
 	// Retrieve the last pair outcome.
 	pair := NewDirectedNodePair(fromNode, toNode)
-	lastPairResult, ok := m.lastPairFailure[pair]
+	lastPairResult, ok := m.lastPairResult[pair]
 
 	// Only look at the last pair outcome if it happened after the last node
 	// level failure. Otherwise the node level failure is the most recent
 	// and used as the basis for calculation of the probability.
 	if ok && lastPairResult.timestamp.After(lastFail) {
+		if lastPairResult.success {
```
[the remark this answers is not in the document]
joostjager: I understand the scenario, but I think we should be careful with those kind of assumptions. Nodes may have shadow channels that make it hard to do these kind of predictions.

One other difference between successes and failures is that successes don't decay.

joostjager:
> Currently failing travis though:
> 
> > --- FAIL: TestMissionControl (0.01s)
> > missioncontrol_test.go:175: unexpected number of channels

Fixed

routing/result_interpretation.go
```diff
@@ -12,17 +14,32 @@ var (
 	reasonIncorrectDetails = channeldb.FailureReasonIncorrectPaymentDetails
 )
 
-// interpretedResult contains the result of the interpretation of a payment
-// outcome.
+type pairResult struct {
```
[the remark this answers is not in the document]
joostjager: fixed

routing/result_interpretation_test.go
joostjager: In case of `ExpiryTooSoon`, any of the nodes could have caused a delay triggering this failure.

joostjager:
Rebased after merge of #3256

joostjager:
Yes, valid concern. Those constants are set to something that looked reasonable to me, but there is no science behind it.

I spend quite some time to produce a test result that would show the benefit of the changes in this pr.

Approach:
* Try to pay to random (testnet) nodes with an unknown hash. The ones that reply with `invalid_payment_details` are logged in a file. I stopped when this file had 40 nodes in it. So those are nodes that I could pay to: the test set.
* Reset mission control
* Run a test where a unknown hash payment payment is made to each of those test set nodes. Payments are executed concurrently, but no more than 5 at the same time. This should really exercise mission control and provide path finding input.
* Count the number of payments attempts made by grepping on the `lnd` log. It took 66 attempts to complete the 40 payments.
* Repeat the same test on master. It needed 68 attempts.

No significant difference unfortunately. Intuitively I'd expect that accounting for successes should show up in the number of attempts needed, but with this test setup it doesn't. I tried with different amounts, but that didn't make much of a difference either. 

My current theory is that there are not enough failures to make any improvement in mission control stand out. However, in other circumstances (mainnet, node connected to a bad cluster) it could make a difference. It isn't ideal though that I cannot demonstrate this.

One thing that I did get out of this test is that there apparently isn't a regression with the changes in this pr. Payment performance is about the same as before.

What I didn't test is the difference in performance over time. Because failures decay (and we need to decay them otherwise we may end up with an empty channel set), a mission control based on failures only will slowly forget everything it has seen. The decay speed can be tuned, but it remains a difficult parameter to pick. Success observations as they are introduced in this pr, do not decay. They keep the success probability of the channel high as long as no failure is observed. That means that this part of mission control's memory isn't forgetting. Therefore payment performance after a day (with default mission control half life time) could very well be noticably better for a successes-aware mission control.

joostjager:
To validate that last hypothesis (performance over time), I did the following:
* Run the test as described above
* Re-running this test on master after a day would return the same score, because all mission control data would have decayed.
* With this pr, it should perform better because successes don't decay. To test this, I restarted `lnd` and set `routerrpc.penaltyhalflife` to 1 second. After `lnd` was started, I verified by running `querymc` that the failures were indeed decayed back to 60%. The success were still at 95%.
* I reran the test with mission control history from the first run of which the failures had decayed artificially fast.
* 42 payment attempts were needed, meaning every payment but two succeeded the first time.

So possibly the behavior over time is the real win of this PR.

1043465747:
اريد بيانات وحسابات في البنوك المحليه والدولي والعالميه


في السبت، ٢٤ أغسطس، ٢٠١٩ ٢:٣٣ ص Olaoluwa Osuntokun <notifications@github.com>
كتب:

PR #3373, opened 2019-08-05
routing: track in-flight htlcs in mission control [wip, don't review]

joostjager:
This PR is a first step towards avoiding paths that have stuck htlcs on them. Mission control is modified to be aware of initiated but not yet completed payments. Based on that information, the probability estimation can be modified to return lower probabilities for those paths.


joostjager:
This PR needs to be rebased. Most of the code has already been merged.

Roasbeef:
What of the fate of this PR?

joostjager:
This is a component for prepay probing. If a (min amount) probe get stuck, that route will be avoided for further probes/payments.

Its fate depends on how we prioritize this.

joostjager:
Examined this PR again. The only commit that isn't merged yet is `routing: report payment initiation`, which isn't much more than passing a signal to mission control when an attempt is initiated. The code to actually decrease the probability estimate based on which htlcs are still in-flight never left the planning stage.

The priority of this improvement is not very high at the moment. The 'black hole' event doesn't seem to be a pressing problem, at least not on mainnet.

One can also argue whether the approach is right. Especially with MPP (and previously already with hold invoice), an in-flight htlc isn't necessarily bad. Avoiding these channels may actually result in worse routing performance.

Closing this PR until there is a better plan and/or more urgency to solve the issue.

PR #3380, opened 2019-08-07
rpc: add unary+streaming interceptors to log all errors

Roasbeef:
Not all errors that occur when serving client requests in the gRPC
server are logged. As a result, at times, we can be lacking critic
information that can be used to debug issues that pop up. With this PR,
we create a basic streaming+unary interceptor that will log all errors
that occur when servicing calls.

The current format looks something like this in the logs when an error
occurs:

```
[ERR] RPCS: [/lnrpc.Lightning/SendCoins]: decoded address is of unknown format
```

### DIFF
--- a/log.go
+++ b/log.go
@@ -1,6 +1,7 @@
 package lnd
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


log.go
```diff
@@ -231,3 +233,37 @@ func (c logClosure) String() string {
...
+
+		return resp, err
+	}
+}
+
+// errorLogStreamServerInterceptor is a simple StreamServerInterceptor that
+// will log any errors that occur while processing a client or server streaming
+// RPC.
+func errorLogStreamServerInterceptor(logger btclog.Logger) grpc.StreamServerInterceptor {
+
```
cfromknecht: nit: newline

PR #3196, opened 2019-06-12
[lntest] remove rpctest build flag

halseth:
Now that the itests are in a separate package, we can remove the rpctest
build flag. This makes code completion and syntax checking work since
the files are no longer "hidden" behind the flag.

To make the tools able to properly compile the whole package without any
build flag, we also change the btcd backend compile when !neutrino is
active, instead of requiring the btcd flag.

PR #3382, opened 2019-08-07
lntest: allow the main test files to be buildable w/o the rpctest build tag

Roasbeef:
In this PR, we modify our build tag set up to allow the main test
files to be buildable w/o the current rpctest tag. We do this so that
those of us that use extensions which will compile live files like
vim-go can once again fix compile errors as we go in our editors.

In order to do this, we now make an external `testsCases` variable, and
have two variants: one that's empty (no build tag), and one that's fully
populated with all our tests (build tag active). As a result, the main
file will now always build regardless of if the build tag is active or
not, but we'll only actually execute tests if the `testCases` variable
has been populated.

As sample run w/ the tag off:
```
=== RUN   TestLightningNetworkDaemon
--- PASS: TestLightningNetworkDaemon (0.00s)
PASS
ok  	github.com/lightningnetwork/lnd/lntest/itest	0.051s
```


cfromknecht:
linter is unhappy:
```
...
lntest/itest/lnd_multi-hop_htlc_remote_chain_claim_test.go:293:9:error: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (vet)
```

Roasbeef:
Hmm, yeah all those were there before, but will fix. So then it would seem that the linter isn't currently run over anything that has a build tag?

cfromknecht:
Correct, metalinter doesn’t support build tags

halseth:
Nice, this is def a better approach than what I tried: https://github.com/lightningnetwork/lnd/pull/3196

https://github.com/lightningnetwork/lnd/pull/3196 has fixes for the linter errors btw, feel free to cherry-pick them.

Roasbeef:
Linter still seems very angry: 
> lntest/itest/macaroons.go:23:6: `testMacaroonAuthentication` is unused (deadcode)
func testMacaroonAuthentication(net *lntest.NetworkHarness, t *harnessTest) { // nolint:unused

Am I not using the `nolint` directive correctly? Btw, this _is_ indeed used, but in a file that's protected by a build-flag. Perhaps this is what's making it go crazy?

lntest/btcd.go
```diff
@@ -1,5 +1,3 @@
-// +build btcd
```
halseth: we might need to include build tags to exclude this if bitcoind or neutrino is active

lntest/itest/lnd_multi-hop_htlc_local_chain_claim_test.go
```diff
@@ -1,5 +1,3 @@
-// +build rpctest
```
halseth: commit title says "w/" instead of "w/o"

lntest/itest/lnd_test.go
```diff
@@ -14621,270 +14501,14 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
-		test: testMacaroonAuthentication,
-	},
-}
-
 // TestLightningNetworkDaemon performs a series of integration tests amongst a
 // programmatically driven network of lnd nodes.
 func TestLightningNetworkDaemon(t *testing.T) {
+	// If no tests are regsitered, then we can exit early.
+	if len(testsCases) == 0 {
+		return
```
halseth: should rather fail the test here, to avoid false negatives by forgetting the build tag

Roasbeef: Ah good point, I was still operating under a model where we still ran the itests from the top-level `lnd` directory.

cfromknecht: `t.Skip()`?

Roasbeef:
Updated with @guggero's latest changes.

Roasbeef:
Looks like theres an error in the commit that defer cancels all the contexts created in the itests, causing them to fail atm.

halseth:
Still want? If yes needs rabase?

guggero:
I'm taking another shot at this with https://github.com/lightningnetwork/lnd/pull/4593. I hope you don't mind me closing this one.

PR #3387, opened 2019-08-08
tlv: add tests for all primitive encodings

cfromknecht:


tlv/primitive_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+type fieldDecoder struct {
+	val     interface{}
+	decoder tlv.Decoder
+	size    uint64
+}
+
+// TestPrimitiveEncodings tests that we are able to serialize all known
+// primitive field types. After successfully encoding, we check that we are able
+// to decode the output and arrive at the same fields.
```
joostjager: Could you explain why it is better to test like this rather than in isolation?

cfromknecht: the primary reason is just that there's a lot of boiler plate that can be reused. another reason is because the `buf` is reused across all encoding/decodings, which helps detect if there are any side effects of fields on each other.

mainly the point of this pr is just to verify that the encodings work properly, tho i agree individual tests be a useful alternative

joostjager: Ok. A comment about this would be nice in the code, but not strictly necessary.

PR #3389, opened 2019-08-09
 invoices: fix synchronization issue with single invoice subscribers

joostjager:
Fixes a synchronization issue where a single invoice subscriber could receive duplicate and/or out of order invoice updates.

Modification required in master to reliably reproduce the test `TestHoldInvoice` failing:
```
case newClient := <-i.newSingleSubscriptions:
   // DEBUG: To trigger test fail, remove.
   time.Sleep(time.Second)

   err := i.deliverSingleBacklogEvents(newClient)
```

### DIFF
--- a/invoices/invoiceregistry.go
+++ b/invoices/invoiceregistry.go
@@ -226,13 +226,7 @@ func (i *InvoiceRegistry) dispatchToSingleClients(event *invoiceEvent) {
 			continue
 		}
 
-		select {
-		case client.ntfnQueue.ChanIn() <- &invoiceEvent{
-			invoice: event.invoice,
-		}:
-		case <-i.quit:
-			return
-		}
+		client.notify(event)
 	}
 }
 
--- a/invoices/invoiceregistry.go
+++ b/invoices/invoiceregistry.go
@@ -139,7 +139,6 @@ func (i *InvoiceRegistry) Stop() {
 // Only two event types are currently supported: newly created invoices, and
 // instance where invoices are settled.
 type invoiceEvent struct {
-	state   channeldb.ContractState
 	hash    lntypes.Hash
 	invoice *channeldb.Invoice
 }
@@ -204,8 +203,9 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
 		case event := <-i.invoiceEvents:
 			// For backwards compatibility, do not notify all
 			// invoice subscribers of cancel and accept events.
-			if event.state != channeldb.ContractCanceled &&
-				event.state != channeldb.ContractAccepted {
+			state := event.invoice.Terms.State
+			if state != channeldb.ContractCanceled &&
+				state != channeldb.ContractAccepted {
 
 				i.dispatchToClients(event)
 			}
@@ -228,7 +228,6 @@ func (i *InvoiceRegistry) dispatchToSingleClients(event *invoiceEvent) {
 
 		select {
 		case client.ntfnQueue.ChanIn() <- &invoiceEvent{
-			state:   event.state,
 			invoice: event.invoice,
 		}:
 		case <-i.quit:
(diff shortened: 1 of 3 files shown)


invoices/invoiceregistry.go
```diff
@@ -997,12 +997,23 @@ func (i *InvoiceRegistry) SubscribeSingleInvoice(
 		}
 	}()
 
+	// Within the lock, we both query the invoice state and pass the client
+	// subscription to the invoiceEvents channel. This is to make sure that
+	// the client receives a consistent stream of events.
+	i.Lock()
+	defer i.Unlock()
+
+	err := i.deliverSingleBacklogEvents(client)
```
wpaulino: It seems like this is the only change needed to address the issue? So no need to merge new subscription client events into the `invoiceEvents` channel.

Also, as someone new to this part of the codebase, I find the locking patterns hard to reason about. I noticed there are cases where we acquire the registry's lock while performing database operations and other cases where we don't.

joostjager: I merged the two types of events in one channel to prevent the following scenario:
1. suppose invoice is `open`
2. subscribesingleinvoice is called. the `open` state gets delivered immediately and subscriber is posted to the (separate) new subscribe channel
3. invoice is settled and invoice update is posted to (separate) invoice update channel
4. main loop picks up event from (3) first. subscriber is not yet in the list and doesn't receive a notification of this event
5. main loop picks up new subscriber event from (2) and adds it to the list.
-> event is missed

wpaulino: It doesn't seem like step 4 can actually happen if we acquire the lock though? An invoice update requires acquiring the lock, so since it's already acquired by `SubscribeSingleInvoice` when dispatching the new client to the main event loop, this ensures the new client gets handled first before the invoice update can get sent to the main event loop.

joostjager: Yes, the lock is acquired. So `subscribe` first posts to the subscriber channel, releases the lock and then settleinvoice posts to the update channel. If at that moment, the main loop enters the select again and both channels are readable (because it was doing slow processing of a previous event for example), it is undetermined which event is read first. But we do require them to be in the order subscribe then update.

wpaulino: > If at that moment, the main loop enters the select again and both channels are readable (because it was doing slow processing of a previous event for example), it is undetermined which event is read first.

Ah I didn't realize the `invoiceEvents` chan is buffered, which usually isn't the norm within event loops in other parts of the codebase.

joostjager: I didn't introduce the buffer, but I think the original reason is to not block the critical payment path when invoices are updated.

joostjager: > I noticed there are cases where we acquire the registry's lock while performing database operations and other cases where we don't.

Do you have a particular location where you saw something you think isn't right?

The general idea is to use the lock to make sure a db update and the event notification are synchronized. If we just read from the database such as in `LookupInvoice`, the lock is not required. The database operations themselves are already atomic because of the bbolt tx.

invoices/invoiceregistry.go
```diff
@@ -402,7 +409,9 @@ func (i *InvoiceRegistry) deliverSingleBacklogEvents(
 
 	// It is possible that the invoice does not exist yet, but the client is
 	// already watching it in anticipation.
-	if err == channeldb.ErrInvoiceNotFound {
+	if err == channeldb.ErrInvoiceNotFound ||
+		err == channeldb.ErrNoInvoicesCreated {
```
cfromknecht: nice catch, is this to handle the case of when your node is brand new?

invoices/invoiceregistry.go
```diff
@@ -237,18 +221,41 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
...
+
+			// A new single invoice subscription has arrived. Add it
+			// to the set of clients. It is important to do this in
+			// sequence with any other invoice events, because an
+			// initial invoice update has already been sent out to
+			// the subscriber.
+			case *SingleInvoiceSubscription:
+				log.Infof("New single invoice subscription "+
+					"client: id=%v, hash=%v",
+					e.id, e.hash,
```
cfromknecht: nit: trailing paren can be pulled into previous line

invoices/invoiceregistry.go
```diff
@@ -988,12 +997,23 @@ func (i *InvoiceRegistry) SubscribeSingleInvoice(
...
+	}
+
 	select {
-	case i.newSingleSubscriptions <- client:
+	case i.invoiceEvents <- client:
 	case <-i.quit:
 	}
 
-	return client
+	return client, nil
```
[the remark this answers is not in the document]
joostjager: Good point. Now that we have a return value, we should return an error. Fixed.

invoices/invoiceregistry.go
joostjager: Yes. But it actually surfaced in a unit test. I am not sure if a separate error for no bucket is justified though, but didn't want to go into that in this pr.

invoices/invoiceregistry.go

Roasbeef:
This passes the build, but fails on Travis atm due to an issue with coveralls.

PR #3391, opened 2019-08-09
htlcswitch+invoices: always return incorrect_or_unknown_payment_details

joostjager:
In order to prevent information leaks by nodes probing with a payment hash, this commit changes exit hop processing so that it always returns incorrect_or_unknown_payment_details and leaves the prober in the dark about whether an invoice actually exists.

This implements part of https://github.com/lightningnetwork/lightning-rfc/pull/608 by removing the deprecated failure message `FinalExpiryTooSoon`. This won't affect payment reliability, because this error was considered final just as `IncorrectOrUnknownPaymentDetails` is.

The follow up of this PR is #3186 in which the accepted height is reported, allowing senders to distinguish between incorrect details and delays on the forward path.


htlcswitch/link_test.go
```diff
@@ -1025,9 +1020,8 @@ func TestChannelLinkMultiHopInsufficientPayment(t *testing.T) {
 	).Wait(30 * time.Second)
 	if err == nil {
 		t.Fatal("error haven't been received")
-	} else if !strings.Contains(err.Error(), "insufficient capacity") {
```
Roasbeef: It may still be useful to report the error string we give out, for the sake of the RPC client. So we'd also extend this new `assertFailureCode` method to still assert our internal string.

htlcswitch/link.go
```diff
@@ -2856,9 +2856,7 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 			"time-lock: expected %v, got %v",
 			pd.RHash[:], pd.Timeout, fwdInfo.OutgoingCTLV)
 
-		failure := lnwire.NewFinalIncorrectCltvExpiry(
-			fwdInfo.OutgoingCTLV,
-		)
+		failure := lnwire.NewFinalIncorrectCltvExpiry(pd.Timeout)
```
Roasbeef: 👍

invoices/invoiceregistry.go
```diff
@@ -28,41 +28,6 @@ var (
 	DebugHash = DebugPre.Hash()
 )
 
-// HtlcCancelReason defines reasons for which htlcs can be canceled.
-type HtlcCancelReason uint8
-
-const (
```
Roasbeef: It may still be useful to log this  information internally for debugging purposes (dev using two nodes on simnet for example).

htlcswitch/link_test.go
joostjager: This sounds like a test that does not belong here. Is it right that you want tests that ensure that we return a specific string for an error over the rpc interface? In that case I'd think we should add those to `lnwire`. But how relevant is this really? We move to structured errors for `SendToRoute`, so for that call these strings won't appear anymore. Also `SendPayment` is moving to higher level structured messages.

invoices/invoiceregistry.go
joostjager: We do, see the `debugLog` statements in `NotifyExitHopHtlc`

PR #3396, opened 2019-08-12
multi: store fwdpkg locked-in time [wip] [no review]

joostjager:
This PR stores the block height and wall clock time when a forwarding package was locked in. The locked-in time of an htlc add is useful for at least the following fixes/improvements:
* #3186 where the accepted block height needs to be reported for the extended `incorrect_or_unknown_payment_details` failure message. The accepted height needs to be consistent across restarts.
* Multi-path payments where a maximum (wall clock) hold duration will be implemented.

### DIFF
--- a/breacharbiter_test.go
+++ b/breacharbiter_test.go
@@ -1963,7 +1963,9 @@ func forceStateTransition(chanA, chanB *lnwallet.LightningChannel) error {
 		return err
 	}
 
-	_, _, _, _, err = chanA.ReceiveRevocation(bobRevocation)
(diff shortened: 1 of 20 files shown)


joostjager:
Unsolved problem: the incoming resolver doesn't have access to the fwd package and can't pass the accepted height to the invoice registry.

This is a problem because it may cause the invoice registry to reject a previously accepted htlc if it goes on-chain.

joostjager:
Alternative approach #3390 where heights are stored in invoice registry looks more promising, closing

PR #3397, opened 2019-08-12
lnrpc: Add subscribe peer events rpc

carlaKC:
This PR adds a subscribe peer events rpc stream, following on from #3354. The motivation for adding this stream would be to provide information for people who want to write their own autopilot logic external to LND.

#### Pull Request Checklist
- [x] Code is commented, has been formatted with `go fmt` and lines are wrapped at 80
- [x] Running `make check`, `go vet` and`make lint` ok
- [x] All commits are logically structured, build properly and pass tests


rpcserver.go
```diff
@@ -17,12 +17,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/lightningnetwork/lnd/chanacceptor"
-	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
-	"github.com/lightningnetwork/lnd/routing/route"
-	"github.com/lightningnetwork/lnd/tlv"
-	"github.com/lightningnetwork/lnd/watchtower"
-
```
carlaKC: Are these separated for a reason? I can re-separate if necessary :)

halseth: No, I think un-separating is correct 👍

rpcserver.go
```diff
@@ -2155,6 +2155,55 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+			}
+
+			if err := eventStream.Send(event); err != nil {
+				return err
+			}
+		case <-r.quit:
+			return nil
+		}
+	}
+
```
halseth: extra newline

rpcserver.go
```diff
@@ -2155,6 +2155,55 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+	eventStream lnrpc.Lightning_SubscribePeerEventsServer) error {
+
+	peerEventSub, err := r.server.peerNotifier.SubscribePeerEvents()
+	if err != nil {
+		return err
+	}
+
+	// Ensure that the resources for the client is cleaned up once either
+	// the server, or client exits.
+	defer peerEventSub.Cancel()
```
halseth: style nit: do 
```go
	peerEventSub, err := r.server.peerNotifier.SubscribePeerEvents()
	if err != nil {
		return err
	}
        defer peerEventSub.Cancel()
```

rpcserver.go
```diff
@@ -2155,6 +2155,55 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+
+	for {
+		select {
+		// A new update has been sent by the peer notifier, we'll
+		// marshal it into the form expected by the gRPC client, then
+		// send it off to the client(s).
```
halseth: will only be one client per stream :)

carlaKC:
> Code LGTM! Question more becomes whether this RPC is useful to expose

I think that it's useful in the context of people building their own channel closing logic? If we're going to expose the final channel insights metrics (rather than time series), be it an all time value of with start/end params, I think that it makes sense to expose the individual elements over RPC so that people can easily build from scratch if they want to.

carlaKC:
> LGTM. As far as the RPC, are there any plans to extend it with other notification types that could make it more useful to expose?

None at present, but I can look into it if online/offline aren't enough to justify adding a new endpoint?

halseth:
Also useful for wallets that want to show the number of connected peers without having to poll listpeers I think :)

carlaKC:
This one has been sitting around for a while now. Should I rebase for merge or close because we don't want to add this rpc endpoint?

halseth:
I think it is still useful, but we can let it sit around a bit longer until someone has a specific use case for it in mind.

carlaKC:
Ok cool 😀 didn’t want to leave it hanging around without following up.

fusion44:
I have a usecase :)

In my Wallet, I display a list of currently connected peers. Wherever possible, I use subscriptions to make the app feel more responsive and dynamic. Subscribing to this means I don't have to implement pull to refresh and save some bytes of data transfered.

https://github.com/fusion44/sendmany/blob/master/docs/screenshots/screenshot_3.png

carlaKC:
> I have a usecase :)

Awesome! 

Rebased on master just in case we want to merge this :D

halseth:
Needs another rebase! 😅

carlaKC:
rerebased 😂 going to add it to the 0.9 milestone so it gets some attention

PR #3402, opened 2019-08-14
multi: remove debug invoices

joostjager:
Debug invoices are rarely used nowadays, but keep asking for maintenance every time refactoring in primarily the invoice registry occurs. We have passed the cost/benefit tipping point, so therefore the debug invoice concept is removed in this commit.

Previously the `debughtlc` flag also controlled whether hodl masks were active. It is safe to remove that additional condition because the hodl masks are still guarded by the `dev` build tag.


config.go
```diff
@@ -269,7 +269,6 @@ type config struct {
 
 	Profile string `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65535"`
 
-	DebugHTLC          bool   `long:"debughtlc" description:"Activate the debug htlc mode. With the debug HTLC mode, all payments sent use a pre-determined R-Hash. Additionally, all HTLCs sent to a node with the debug HTLC R-Hash are immediately settled in the next available state transition."`
```
wpaulino: Just something to note -- I've seen `lnd.conf`s with the option in it, so removing it will prevent some people from starting, in which case they'll see an error along the lines of `unknown option: debughtlc`.

rpcserver.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/btcsuite/btcwallet/wallet/txauthor"
 	"github.com/coreos/bbolt"
 	"github.com/davecgh/go-spew/spew"
-	"github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
wpaulino: Was this done automatically by `goimports`?

config.go
joostjager: I didn't think about that. But we have removed or renamed flags before. I don't know what the procedure was there, but it seems to me that a note in the release notes should be enough. Especially for this flag that has little use for real users.

rpcserver.go
joostjager: Yes

config.go
Roasbeef: We've renamed config options in the past (`noencryptwallet` -> `noseedbackup`) without much issue. I also would wager that there aren't manny nodes in production that're running with this flag in the wild, and if they are, then this update will be a wake up call for t hem.

joostjager:
I'd say a benchmark isn't a real benchmark if an alternative code path is executed. So in that regard, `debughtlc` wasn't ideal. If we want to do a benchmark, I don't think it is really a problem to generate a set of regular invoices in a script. And you say, amp is another option. Or even mpp and paying to the same invoice over and over again. That will at least benchmark the link logic.

config.go
cfromknecht: we disabled the use of `debughtlc` on mainnet [over a year ago](https://github.com/lightningnetwork/lnd/pull/1475). if people still have this enabled, they already can't start

wpaulino: They could've just flipped it from true to false. Nothing major though, just wanted to point it out.

cfromknecht: good point, yes that's certainly possible too :)

Issue #3398, opened 2019-08-12
LND 100% CPU Won't gracefully shutdown


MrRGnome:
### Background

Been running a node with stable config for ~ 9months. After some moderate usage of ln-service calls to LND the node has become unresponsive to shutdown commands and is consuming 100% of the available CPU. Attempting to close the node results in it spamming this message: "2019-08-12 17:04:40.013 [INF] UTXN: NurseryReport: building nursery report for channel 398678968111a6ec1d8517f93b0d9ff229ca43e77677b530dcacc1c6ad6512d2:0"

### Your environment

* lnd 0.7.1-beta`
* Ubuntu 18.04
* bitcoind 18.0

### Steps to reproduce

I'm not entirely sure how to reproduce the issue. I am using a very normal lnd.conf (see attached) and simple grpc commands through ln-service. The ln-service calls I am using include: 
decodePaymentRequest
pay
sendToChainAddress
subscribeToInvoices
createInvoice
createChainAddress
subscribeToChainAddress

### Expected behaviour

LND should shutdown when ctrl+C is hit

### Actual behaviour

LND spams the output: 
"2019-08-12 17:04:40.013 [INF] UTXN: NurseryReport: building nursery report for channel 398678968111a6ec1d8517f93b0d9ff229ca43e77677b530dcacc1c6ad6512d2:0"

[lnd.conf.txt](https://github.com/lightningnetwork/lnd/files/3493701/lnd.conf.txt)

wpaulino:
> Attempting to close the node results in it spamming this message: "2019-08-12 17:04:40.013 [INF] UTXN: NurseryReport: building nursery report for channel 398678968111a6ec1d8517f93b0d9ff229ca43e77677b530dcacc1c6ad6512d2:0"

This log only gets shown when calling the `pendingchannels` RPC.

Would you be able to obtain a CPU profile? You'll have to enable profiling in lnd with the --profile=PORT flag and restart. Once done, you can run `go tool pprof -png http://localhost:PORT/debug/pprof/profile`. This will produce a PNG file which you can then include here.

MrRGnome:
Thanks for the quick and helpful response. It should be noted the issue has yet to recur and I expect these CPU profiles to be within normal ranges.

[...]

PR #3405, opened 2019-08-15
chainrpc: default to one confirmation and require height hints

wpaulino:
A height hint not being set would cause lnd to scan for the confirmation/spend of a txid/outpoint/address from genesis.

The number of confirmations not being set within a confirmation request would cause the internal TxNotifier to deadlock when dispatching updates. To prevent this, we default to dispatching a notification upon one confirmation.

Fixes https://github.com/lightningnetwork/lnd/issues/3398.


lnrpc/chainrpc/chainnotifer_server.go
```diff
@@ -196,14 +196,26 @@ func (s *Server) RegisterWithRootServer(grpcServer *grpc.Server) error {
 func (s *Server) RegisterConfirmationsNtfn(in *ConfRequest,
 	confStream ChainNotifier_RegisterConfirmationsNtfnServer) error {
 
-	// We'll start by reconstructing the RPC request into what the
-	// underlying ChainNotifier expects.
+	// A height hint is required to prevent scanning the chain from genesis.
+	if in.HeightHint == 0 {
+		return errors.New("a height hint must be provided")
+	}
```
joostjager: Is this now true for all back ends? If it isn't, then we probably still want to enforce it here.

lnrpc/chainrpc/chainnotifer_server.go
```diff
@@ -196,14 +196,26 @@ func (s *Server) RegisterWithRootServer(grpcServer *grpc.Server) error {
...
+
+	// If the number of confirmations isn't specified, we'll use a sane
+	// default of 1.
+	numConfs := in.NumConfs
+	if numConfs == 0 {
+		numConfs = 1
+	}
```
joostjager: I think we should return an error here too. Keep the api as strict as we can. We need to educate the callers, not camouflage potential bugs.

lnrpc/chainrpc/chainnotifer_server.go
```diff
@@ -196,14 +196,26 @@ func (s *Server) RegisterWithRootServer(grpcServer *grpc.Server) error {
...
+
+	// We'll then reconstruct the RPC request into what the underlying
+	// ChainNotifier expects.
 	var txid chainhash.Hash
 	copy(txid[:], in.Txid)
 
 	// We'll then register for the spend notification of the request.
 	confEvent, err := s.cfg.ChainNotifier.RegisterConfirmationsNtfn(
-		&txid, in.Script, in.NumConfs, in.HeightHint,
+		&txid, in.Script, numConfs, in.HeightHint,
```
joostjager: Shouldn't these checks actually happen a level deeper inside `ChainNotifier`?

chainntnfs/txnotifier.go
```diff
@@ -1749,11 +1749,11 @@ func (n *TxNotifier) dispatchSpendReorg(ntfn *SpendNtfn) error {
 // closes the event channels of all registered notifications that have not been
 // dispatched yet.
 func (n *TxNotifier) TearDown() {
+	close(n.quit)
+
```
joostjager: Separate commit with explanation in commit msg? Looks unrelated to other changes.

lnrpc/chainrpc/chainnotifer_server.go
wpaulino: Done.

lnrpc/chainrpc/chainnotifer_server.go
wpaulino: Sure, done.

chainntnfs/txnotifier.go
wpaulino: It already is.

joostjager: Sorry, was viewing the all changes diff

chainntnfs/txnotifier.go
```diff
@@ -272,8 +285,17 @@ type SpendRequest struct {
...
+	heightHint uint32) (SpendRequest, error) {
+
 	var r SpendRequest
+	if len(pkScript) == 0 {
+		return r, errors.New("an output script must be provided")
+	}
+	if heightHint == 0 {
+		return r, errors.New("a height hint greater than 0 must be " +
+			"provided")
+	}
```
joostjager: Something looks as if it is not right here. This function takes two parameters that is does nothing with other than checking. Shouldn't those parameters be part of the `SpendRequest` struct too? Or is there a different design that doesn't show this friction? Same for function above.

wpaulino: I just decided to group the validation checks within these functions to prevent having another function as these are already doing some checks as well. We don't want to store these in the structs because they're used as keys for maps and we wouldn't be able to coalesce different requests for the same txid/outpoint/script. I'm not opposed to creating a separate function for them though if you think it's a cleaner approach.

wpaulino:
As a result of having the validation checks being done at the chain notifier level, a bug was discovered with the BumpFee RPC that would cause us to scan the chain from genesis due to a 0 height hint. See the relevant commit for more details.

chainntnfs/txnotifier.go
joostjager: It looked into this part of the code a bit more. It seems that the code in the three `RegisterConfirmationsNtfn` functions up to the `txNotifier.RegisterConf` call is duplicated. What would you think about moving that part into `txNotifier` and doing the parameter validation there as well?

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -516,7 +516,13 @@ func (w *WalletKit) BumpFee(ctx context.Context,
...
 	}
 
-	input := input.NewBaseInput(op, witnessType, signDesc, 0)
+	_, currentHeight, err := w.cfg.Chain.GetBestBlock()
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve current height: %v",
+			err)
+	}
+
+	input := input.NewBaseInput(op, witnessType, signDesc, uint32(currentHeight))
```
joostjager: I think usually the height hint that we pass to the sweeper is the height at which the input was created. So that in case the input is swept by something else (maybe an old tx of ourselves or the remote party), we don't miss the spend ntfn. For an external input, I think we need to pass that as a parameter with the rpc call?

wpaulino: The output is unconfirmed so we can assume the current height as the height hint. Only thing is that this RPC currently doesn't validate the output is indeed unconfirmed.

joostjager: Ah yes, that is right. And what will happen if you `bumpfee` a confirmed output? (suppose it confirms just when you are typing the command). The sweeper won't detect the spend and try to sweep `maxAttempts` times? Does it need some padding here, like `currentHeight-x`?

joostjager: If you'd want to add a check for confirmedness in the rpc, you'd need the height hint right? But if you have a proper height hint, you can just as well pass that into the sweeper. The sweeper will then get the spend ntfn immediately and report the outcome. So no checking on the rpc level would be required? Just thinking about your comment, not saying that this needs to be added now.

joostjager: Ok, I am mixing up conf and spent. In any case, if we want to do checking, we may want to do it in `SweepInput` and not on the rpc level.

wpaulino: > And what will happen if you bumpfee a confirmed output? 

You're essentially just sending to yourself at that point for no gain.

I think all that's needed is a check at the RPC level to detect if the output is unconfirmed. If it's confirmed then we can return an error.

wpaulino: Unconfirmed check added.

chainntnfs/txnotifier.go
wpaulino: Done.

chainntnfs/txnotifier.go
```diff
@@ -484,34 +506,62 @@ func NewTxNotifier(startHeight uint32, reorgSafetyLimit uint32,
 	}
 }
 
+// validateConfRegistration validates all of the parameters required to
+// successfully register a confirmation notification.
+func (n *TxNotifier) validateConfRegistration(txid *chainhash.Hash,
```
joostjager: Does the function name describe the code inside enough? It indeed validates parameters, but is also creates the ntfn object.

chainntnfs/bitcoindnotify/bitcoind.go
```diff
@@ -640,17 +628,18 @@ func (b *BitcoindNotifier) RegisterSpendNtfn(outpoint *wire.OutPoint,
 	// outpoint/output script as spent.
 	//
 	// TODO(wilmer): use LoadFilter API instead.
-	if spendRequest.OutPoint == chainntnfs.ZeroOutPoint {
-		addr, err := spendRequest.PkScript.Address(b.chainParams)
+	if outpoint == nil || *outpoint == chainntnfs.ZeroOutPoint {
```
joostjager: Maybe convert the parameter to a value (no pointer) to prevent needing to deal with two values that have the same meaning? (non blocking)

chainntnfs/bitcoindnotify/bitcoind.go
```diff
@@ -686,16 +675,16 @@ func (b *BitcoindNotifier) RegisterSpendNtfn(outpoint *wire.OutPoint,
 	// We'll start by checking the backend's UTXO set to determine whether
 	// the outpoint has been spent. If it hasn't, we can return to the
 	// caller as well.
-	txOut, err := b.chainConn.GetTxOut(
-		&spendRequest.OutPoint.Hash, spendRequest.OutPoint.Index, true,
-	)
+	txOut, err := b.chainConn.GetTxOut(&outpoint.Hash, outpoint.Index, true)
```
joostjager: It is kind of implicit that `outpoint` can no longer be nil here

chainntnfs/btcdnotify/btcd.go
```diff
@@ -705,26 +694,29 @@ func (b *BtcdNotifier) RegisterSpendNtfn(outpoint *wire.OutPoint,
 	// We'll short-circuit the path when dispatching the spend of a script,
 	// rather than an outpoint, as there aren't any additional checks we can
 	// make for scripts.
-	if spendRequest.OutPoint == chainntnfs.ZeroOutPoint {
+	if outpoint == nil || *outpoint == chainntnfs.ZeroOutPoint {
```
joostjager: Same comment about supporting two values

chainntnfs/neutrinonotify/neutrino.go
```diff
@@ -690,9 +677,12 @@ func (n *NeutrinoNotifier) RegisterSpendNtfn(outpoint *wire.OutPoint,
 	//
 	// We'll update our filter first to ensure we can immediately detect the
 	// spend at tip.
+	if outpoint == nil {
+		outpoint = &chainntnfs.ZeroOutPoint
+	}
```
joostjager: Do this as first thing in this function, so that `txNotifier.RegisterSpend` can accept it as a value parameter?

chainntnfs/txnotifier.go
```diff
@@ -887,29 +907,50 @@ func (n *TxNotifier) dispatchConfDetails(
 	return nil
 }
 
+// validateSpendRegistration validates all of the parameters required to
+// successfully register a spend notification.
+func (n *TxNotifier) validateSpendRegistration(outpoint *wire.OutPoint,
+	pkScript []byte, heightHint uint32) (*SpendNtfn, error) {
```
joostjager: Same comment about function name. Maybe `createSpendNtfn` describes the main purpose better, even if we later add some validation in here.

chainntnfs/txnotifier.go
```diff
@@ -46,10 +46,22 @@ var (
...
+	// registration is attempted without providing an accompanying height
+	// hint.
+	ErrNoHeightHint = errors.New("a height hint greater than 0 must be " +
+		"provided")
+
+	// ErrNumConfsOutOfRange is an error returned when a confirmation/spend
+	// registration is attempted and the number of confirmations provided is
+	// out of range.
+	ErrNumConfsOutOfRange = fmt.Errorf("number of confirmations must be "+
+		"between %d and %d", 0, MaxNumConfs)
```
joostjager: Between 0 an max means `>0 && <=max`. If 'between' is inclusive, it should be between 1 and max?

lnwallet/btcwallet/signer.go
```diff
@@ -26,17 +26,7 @@ func (b *BtcWallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.TxOut, error)
...
-	// First check to see if the output is already within the utxo cache.
-	// If so we can return directly saving a disk access.
-	b.cacheMtx.RLock()
-	if output, ok := b.utxoCache[*prevOut]; ok {
-		b.cacheMtx.RUnlock()
-		return output, nil
-	}
-	b.cacheMtx.RUnlock()
-
-	// Otherwise, we manually look up the output within the tx store.
```
joostjager: Could use a commit message explaining why this needs to be removed and doesn't have a negative effect?

lnwallet/btcwallet/signer.go
```diff
@@ -39,13 +37,40 @@ func (b *BtcWallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.TxOut, error)
...
+	}
+
+	// Determine the number of confirmations the output currently has.
+	_, currentHeight, err := b.GetBestBlock()
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve current height: %v",
+			err)
+	}
+	confs := int64(0)
+	if txDetail.Block.Height != -1 {
```
joostjager: `-1` means unconfirmed?

sweep/walletsweep.go
```diff
@@ -232,7 +232,10 @@ func CraftSweepAllTx(feeRate lnwallet.SatPerKWeight, blockHeight uint32,
 		// The rest of the items will be populated internally within
 		// the sweeper via the witness generation function.
 		signDesc := &input.SignDescriptor{
-			Output:   outputInfo,
+			Output: &wire.TxOut{
+				PkScript: outputInfo.PkScript,
+				Value:    int64(output.Value),
```
joostjager: Is there a reason not to use `Value` from `outputInfo`?

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -494,13 +495,27 @@ func (w *WalletKit) BumpFee(ctx context.Context,
...
 	// (CPFP).
 	//
+	// We'll use the current height as the height hint since we're dealing
+	// with an unconfirmed transaction.
+	_, currentHeight, err := w.cfg.Chain.GetBestBlock()
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve current height: %v",
+			err)
+	}
+
```
joostjager: This call can be moved down.

chainntnfs/bitcoindnotify/bitcoind.go
wpaulino: Sure, but I don't want to trigger even more changes since the ChainNotifier interface is used quite heavily throughout the codebase.

chainntnfs/txnotifier.go

chainntnfs/bitcoindnotify/bitcoind.go
wpaulino: Correct.

chainntnfs/txnotifier.go
wpaulino: Done.

chainntnfs/txnotifier.go
wpaulino: Yes, fixed.

lnwallet/btcwallet/signer.go
wpaulino: Done.

lnwallet/btcwallet/signer.go
wpaulino: Yeah that's what's returned by the underlying call for unconfirmed transactions.

sweep/walletsweep.go
wpaulino: We actually don't even need to make the `FetchInputInfo` call here because we already have all the information required from `ListUnspentWitness`.

lnrpc/walletrpc/walletkit_server.go
wpaulino: Done.

chainntnfs/bitcoindnotify/bitcoind.go
joostjager: Ok, next round :)

lnwallet/btcwallet/signer.go
joostjager: `const unconfirmed = -1` would be nice

sweep/walletsweep.go
joostjager: Ok, even better

alexbosworth:
I see an issue here where I get this error when trying to restore an SCB

`2 UNKNOWN: unable to unpack chan backup: a height hint greater than 0 must be provided`

PR #3406, opened 2019-08-15
pilot+discovery: die spew

cfromknecht:
removes prevalent info spews

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -2010,8 +2010,8 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 			prefix = "remote"
 		}
 
-		log.Infof("Received new %v channel announcement: %v", prefix,
-			spew.Sdump(msg))
+		log.Infof("Received new %v channel announcement for %v", prefix,
+			msg.ShortChannelID)
 
 		// By the specification, channel announcement proofs should be
 		// sent after some number of confirmations after channel was
--- a/pilot.go
+++ b/pilot.go
@@ -8,7 +8,6 @@ import (
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
-	"github.com/davecgh/go-spew/spew"
 	"github.com/lightningnetwork/lnd/autopilot"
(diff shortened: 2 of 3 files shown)


pilot.go
```diff
@@ -142,7 +141,10 @@ var _ autopilot.ChannelController = (*chanController)(nil)
 // and all interfaces needed to drive it won't be launched before the Manager's
 // StartAgent method is called.
 func initAutoPilot(svr *server, cfg *autoPilotConfig) (*autopilot.ManagerCfg, error) {
-	atplLog.Infof("Instantiating autopilot with cfg: %v", spew.Sdump(cfg))
+	atplLog.Infof("Instantiating autopilot with max_channels=%d, "+
```
wpaulino: Nit: there's a trailing comma here while the other params don't have one. Also I think we have a preference for using camel case for stuff like this, though it's not strictly enforced.

discovery/gossiper.go
```diff
@@ -2010,8 +2010,8 @@ func (d *AuthenticatedGossiper) processNetworkAnnouncement(
 			prefix = "remote"
 		}
 
-		log.Infof("Received new %v channel announcement: %v", prefix,
-			spew.Sdump(msg))
+		log.Infof("Received new %v channel announcement for %v", prefix,
+			msg.ShortChannelID)
```
Roasbeef: Nice, this definitely shouldn't have been on the `info` logging level.

pilot.go
Roasbeef: What in the config spews a curve?

utxonursery.go
```diff
@@ -498,7 +498,7 @@ func (u *utxoNursery) NurseryReport(
 	u.mu.Lock()
 	defer u.mu.Unlock()
 
-	utxnLog.Infof("NurseryReport: building nursery report for channel %v",
+	utxnLog.Debugf("NurseryReport: building nursery report for channel %v",
```
Roasbeef: 👍 

This can be spammy for those using the application that have a force closed channel.

pilot.go
cfromknecht: commas added.

> What in the config spews a curve?

There isn't one, this is just to get rid of any spews in info logs since it breaks parsing in the app, and generally make everything more consistent

PR #3414, opened 2019-08-19
lnwire+htlcswitch: report htlc accept height

joostjager:
Implements lightningnetwork/lightning-rfc#608 by extending the incorrect_or_unknown_payment_details message with the height at which the failed htlc was accepted.


lnrpc/routerrpc/router.proto
```diff
@@ -239,6 +239,9 @@ message Failure {
     the failure message. Position zero is the sender node.
     **/
     uint32 failure_source_index = 8;
+
+    /// A failure type-dependent block height.
+    uint32 height = 9;
```
cfromknecht: perhaps something like `accept_height`, to mimic what the invoice htlc's will show?

joostjager: I thought about that, but wasn't sure. The idea of the `message Failure` is that it isn't a `oneof` to keep it simple. Fields are reused in different messages. This `height` is meant as a generic height field associated with a failure. But yes, I could also make it more specific.

invoices/invoiceregistry.go
```diff
@@ -613,10 +614,16 @@ func (i *InvoiceRegistry) SettleHodlInvoice(preimage lntypes.Preimage) error {
 	log.Debugf("Invoice(%v): settled with preimage %v", hash,
 		invoice.Terms.PaymentPreimage)
 
-	i.notifyHodlSubscribers(HodlEvent{
-		Hash:     hash,
-		Preimage: &preimage,
-	})
+	for key, htlc := range invoice.Htlcs {
+		if htlc.State == channeldb.HtlcStateCancelled {
```
halseth: check that state is settled instead?

lnwire/onion_error_test.go
```diff
@@ -189,8 +187,32 @@ func TestFailIncorrectDetailsOptionalAmount(t *testing.T) {
...
+		t.Fatalf("expected FailIncorrectDetails, but got %T",
+			onionError2)
 	}
+
+	if invalidDetailsErr.amount != 0 {
+		t.Fatalf("expected amount to be zero")
+	}
+}
+
+type legacyFailIncorrectDetailsNoAmt struct {
```
halseth: `mockFailIncorrectDetailsNoAmte`?

lnwire/onion_error.go
```diff
@@ -381,14 +393,23 @@ func (f *FailIncorrectDetails) Decode(r io.Reader, pver uint32) error {
 		return err
 	}
 
+	err = ReadElement(r, &f.height)
```
halseth: should have a comment, as above.

invoices/invoiceregistry.go
joostjager: Why settled? The invoice is settled and we need to notify all links that hold htlcs paying to the invoice. Htlcs that were already canceled before, don't need to receive anything and are skipped.

lnwire/onion_error_test.go
joostjager: It really is the legacy failure that no longer exists, but we can still receive it.

lnwire/onion_error.go
joostjager: added

invoices/invoiceregistry.go
halseth: Seems more correct to notify that it is settled _only if_ the state is settled. The only remaining state looks like it is `HtlcStateAccepted`, is it expected that we would find the htlc in this state here?

lnwire/onion_error_test.go
halseth: Since methods are empty, and this is only in test code, looks more like a mock to me than an actual legacy thing :P

invoices/invoiceregistry.go
joostjager: Changed to checking for settled only and added comment

lnwire/onion_error_test.go
joostjager: Renamed

cfromknecht:
```
invoices/invoiceregistry_test.go:127:29: undefined: testInvoiceCltvDelta
```

joostjager:
Test failure fixed

invoices/invoiceregistry.go
```diff
@@ -664,9 +677,16 @@ func (i *InvoiceRegistry) CancelInvoice(payHash lntypes.Hash) error {
 	}
 
 	log.Debugf("Invoice(%v): canceled", payHash)
-	i.notifyHodlSubscribers(HodlEvent{
-		Hash: payHash,
-	})
+
+	for key, htlc := range invoice.Htlcs {
```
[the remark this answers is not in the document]
joostjager: Comment added. While writing the comment, I also realized that the invoice update could be a bit stricter. See commit `channeldb+invoices: make htlc cancelation stricter `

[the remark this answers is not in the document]
joostjager: Comment added. While writing the comment, I also realized that the invoice update could be a bit stricter. See commit `channeldb+invoices: make htlc cancelation stricter `

joostjager:
Requires #3503 to be merged so that the linter is tamed first.

invoices/invoiceregistry.go
```diff
@@ -640,7 +640,10 @@ func (i *InvoiceRegistry) CancelInvoice(payHash lntypes.Hash) error {
 		canceledHtlcs := make(
 			map[channeldb.CircuitKey]*channeldb.HtlcAcceptDesc,
 		)
-		for key := range invoice.Htlcs {
+		for key, htlc := range invoice.Htlcs {
+			if htlc.State == channeldb.HtlcStateCancelled {
```
halseth: Does this mean that we can cancel a settled one?

joostjager: Discussed offline. No htlc should be in the `settled` state here, because then the invoice state would have been `settled`. Added extra check to return an error if this happens.

PR #3415, opened 2019-08-19
htlcswitch+invoices: allow settling invoices via multi-path payments

joostjager:
Before this PR, MPP fields in the payload were parsed but not processed. This PR adds processing of those fields in the invoice registry. This means that it is now possible to settle an invoice using multiple partial payments.


cmd/lncli/commands.go
```diff
@@ -2505,41 +2507,36 @@ func sendToRoute(ctx *cli.Context) error {
...
 
-	paymentStream, err := client.SendToRoute(context.Background())
+	resp, err := client.SendToRoute(context.Background(), req)
 	if err != nil {
 		return err
 	}
 
-	if err := paymentStream.Send(req); err != nil {
-		return err
+	marshaller := jsonpb.Marshaler{
```
cfromknecht: is this different from `printRespJSON`?

cmd/lncli/commands.go
```diff
@@ -2505,41 +2507,36 @@ func sendToRoute(ctx *cli.Context) error {
 		route = routes.Route
 	}
 
-	req := &lnrpc.SendToRouteRequest{
+	req := &routerrpc.SendToRouteRequest{
```
cfromknecht: is this necessary? does it compile if users don't have `routerrpc` enabled?

cmd/lncli/commands.go
```diff
@@ -18,14 +18,13 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
```
cfromknecht: stray deletion

lnrpc/rpc.proto
```diff
@@ -2232,6 +2232,9 @@ message InvoiceHTLC {
 
     /// Current state the htlc is in.
     InvoiceHTLCState state = 8 [json_name = "state"];
+
+    /// The total amount of the mpp payment in msat.
+    uint64 total_amt_msat = 9 [json_name = "total_amt_msat"];
```
cfromknecht: also add `payment_addr`? could also add the `mpp_` prefix to match the sender side

joostjager:
Todo: handle payment_addr

invoices/invoiceregistry.go
```diff
@@ -465,19 +469,67 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+		if mpp == nil {
+			if inv.Terms.Value > 0 && amtPaid < inv.Terms.Value {
+				debugLog("amount too low and no mpp")
+				return nil, errNoUpdate
+			}
+			canSettle = true
+		} else {
+			// Check whether total amt matches other htlcs in the
+			// set.
+			for _, htlc := range inv.Htlcs {
```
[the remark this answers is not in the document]
joostjager: I think that is only needed for recurring payments? I now just added a flat check for payment address

invoices/invoiceregistry.go
```diff
@@ -416,6 +427,98 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+func (i *InvoiceRegistry) startHtlcTimer(hash lntypes.Hash,
+	key channeldb.CircuitKey, acceptTime time.Time) bool {
+
+	now := i.now()
+	releaseTime := acceptTime.Add(htlcHoldDuration)
+	if releaseTime.Before(now) {
+		return true
+	}
+
+	i.wg.Add(1)
```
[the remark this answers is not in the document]
joostjager: not sure what you mean. this is another goroutine that needs to be tracked in the wait group?

cmd/lncli/commands.go
joostjager: I don't remember why I did that. Replaced by `printRespJSON`. But this is a debug commit anyway.

lnrpc/rpc.proto
joostjager: Good question. Payment address is already a property on the invoice level. Maybe we want to expose everything that we got from the htlc here. But on the other hand, htlcs with a payment address that doesn't match aren't added here. We are only recording failed htlcs after they got accepted initially.

I added the `mpp_` prefix

invoices/invoiceregistry.go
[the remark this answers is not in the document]
cfromknecht: i'm talking about the (albeit rare) waitgroup panic that can happen when goroutines aren't spawned from inside other waitgrouped goroutines. idt it's a big deal for now

invoices/invoiceregistry.go
```diff
@@ -465,19 +469,67 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+		acceptDesc := &channeldb.HtlcAcceptDesc{
+			Amt:          amtPaid,
+			Expiry:       expiry,
+			AcceptHeight: currentHeight,
+		}
+
+		// If an invoice amount is specified and this is not an mpp,
+		// check that the exact amount is paid.
+		var canSettle bool
+		if mpp == nil {
```
cfromknecht: do we also need to check if `len(inv.Htlcs) == 0`? my reading atm is that this will also settle any partial htlcs if a regular payment comes thru that settles the invoice.

haven't thought about it too much, but seems we should reject regular payments if we already have some partial mpp payments. if we have an mpp invoice feature bit, this path should also fail if `option_mpp` is required on the invoice

invoices/invoiceregistry.go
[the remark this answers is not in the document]
cfromknecht: yes you're right

invoices/invoiceregistry.go
```diff
@@ -465,19 +469,67 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+		} else {
+			// Check whether total amt matches other htlcs in the
+			// set.
+			for _, htlc := range inv.Htlcs {
+				// Only consider accepted mpp htlcs.
+				if htlc.State != channeldb.HtlcStateAccepted {
+					continue
+				}
+
+				if htlc.TotalAmt == 0 {
```
cfromknecht: when would this happen? why would we have accepted it?

invoices/invoiceregistry.go
```diff
@@ -465,19 +469,67 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+				if mpp.TotalMsat < inv.Terms.Value {
+					debugLog("set total too low")
+					return nil, errNoUpdate
+				}
+			}
+
+			// Make sure the communicated set total isn't overpaid.
+			if inv.AmtPaid+amtPaid > mpp.TotalMsat {
+				debugLog("mpp is overpaying set total")
+				return nil, errNoUpdate
```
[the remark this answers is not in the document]
cfromknecht: got it, thanks!

invoices/invoiceregistry.go
joostjager: good point. added check and todo

invoices/invoiceregistry.go
joostjager: I don't remember exactly, but I think this was for non-mpp htlcs. But if they would be present for the hodl invoice case, we can never get here. Removed.

invoices/invoiceregistry.go
```diff
@@ -199,6 +220,27 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
 				i.singleNotificationClients[e.id] = e
 			}
 
+		case event := <-i.htlcAutoReleaseChan:
```
joostjager: Tried out launching the goroutine in the main registry event loop. What do you think? I am a bit hesitant to optimize this already now with a priority queue.

invoices/invoiceregistry.go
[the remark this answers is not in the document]
joostjager: see comment below about event loop

invoices/invoiceregistry.go
joostjager: Trying out prio queue now

joostjager: Pushed

invoices/invoiceregistry.go
```diff
@@ -465,19 +598,95 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+				}
+			}
+
+			// Make sure the communicated set total isn't overpaid.
+			if inv.AmtPaid+amtPaid > mpp.TotalMsat() {
+				debugLog("mpp is overpaying set total")
+				return nil, errNoUpdate
+			}
+
+			canSettle = inv.AmtPaid+amtPaid == mpp.TotalMsat()
```
cfromknecht: is the `inv.AmtPaid` updated as each htlc is accepted? should it not pull this from the htlc set directly, and only update after accept, so that the payment appears atomic externally?

invoices/invoiceregistry.go
```diff
@@ -465,19 +598,95 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+
+			// If an invoice amount is specified, check that enough
+			// is paid. Also check this for duplicate payments if
+			// the invoice is already settled or accepted.
+			if inv.Terms.Value > 0 && amtPaid < inv.Terms.Value {
+				debugLog("amount too low and no mpp")
+				return nil, errNoUpdate
+			}
+
+			// TODO(joostjager): Check invoice mpp required feature
```
cfromknecht: note to self, we won't need to assert the feature bit until we begin to make them mandatory in the invoice, so this can be deferred until we allow that

invoices/invoiceregistry.go
```diff
@@ -127,13 +143,26 @@ type invoiceEvent struct {
...
 
+	// Set up a heap for htlc auto-releases.
+	autoReleaseHeap := &releaseHeap{}
+	heap.Init(autoReleaseHeap)
+
 	for {
+		// If there is something to release, set up a release tick
+		// channel.
+		var nextReleaseTick <-chan time.Time
+		if autoReleaseHeap.Len() > 0 {
```
cfromknecht: this would make a great addition to the `ticker` package 🤓 `ticker.Heap`?

lnrpc/rpc.proto
cfromknecht: good point, yes the payment addr should be referenced from the invoice itself 👍

invoices/invoiceregistry.go
joostjager: Ok, updated comment.

invoices/invoiceregistry.go
joostjager: Yes it would, but I prefer to do this when the second release heap is introduced to see how it is best abstracted.

joostjager:
Just saw comment about updating `amtPaid`, which triggers more refactoring. Hold off review

invoices/invoiceregistry.go
joostjager: Good point. This triggered quite a refactor that is also good for the overall code quality of invoice registry.

invoices/invoiceregistry.go
```diff
@@ -486,13 +632,38 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
 		}, nil
 
 	case channeldb.HtlcStateSettled:
+		// Also settle any previously accepted htlcs.
```
halseth: s/accepted/settled ?

invoices/invoiceregistry.go
```diff
@@ -486,13 +632,38 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
 			CircuitKey:   circuitKey,
 			Preimage:     &invoice.Terms.PaymentPreimage,
 			AcceptHeight: acceptHeight,
 		}, nil
 
 	case channeldb.HtlcStateAccepted:
+		// (Re)start the htlc timer if the invoice is
+		// still open.
+		if invoice.State == channeldb.ContractOpen {
+			err := i.startHtlcTimer(
```
halseth: should timer only apply to MPP?

invoices/invoiceregistry.go
```diff
@@ -486,13 +632,38 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+		}
+
 		return &HodlEvent{
 			CircuitKey:   circuitKey,
 			Preimage:     &invoice.Terms.PaymentPreimage,
 			AcceptHeight: acceptHeight,
 		}, nil
 
 	case channeldb.HtlcStateAccepted:
+		// (Re)start the htlc timer if the invoice is
```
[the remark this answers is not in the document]
joostjager: Restart happens after a restart of `lnd`

channeldb/invoices.go
```diff
@@ -1187,6 +1199,7 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
 		htlc.ResolveTime = time.Unix(0, int64(resolveTime))
 		htlc.State = HtlcState(state)
 		htlc.Amt = lnwire.MilliSatoshi(amt)
+		htlc.MppTotalAmt = lnwire.MilliSatoshi(mppTotalAmt)
```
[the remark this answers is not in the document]
joostjager: It should be ok. If it is a non-mpp htlc, the way to signal that is with a zero total amount. Only in case we would have accepted an mpp htlc which isn't resolved yet, we may discard the mpp fields here. But previously we only held htlcs for hodl invoices and once an invoice is in the accepted state, the mpp logic is no longer exercised.

We probably should have been stricter when we started parsing mpp fields in a previous pr (without the implementation being complete) by rejecting those htlcs.

invoices/invoiceregistry.go
```diff
@@ -29,6 +33,12 @@ var (
 	errNoUpdate = errors.New("no update needed")
 )
 
+const (
+	// htlcHoldDuration defines how long mpp htlcs are held while waiting
+	// for the other set members to arrive.
+	htlcHoldDuration = 30 * time.Second
```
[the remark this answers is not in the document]
joostjager: What value do you suggest? Can make it configurable, but will people actually configure it? I am more in favor of picking a good value here. 30 secs seems too short indeed.

invoices/invoiceregistry.go
```diff
@@ -127,13 +149,26 @@ type invoiceEvent struct {
...
+	// Set up a heap for htlc auto-releases.
+	autoReleaseHeap := &releaseHeap{}
+	heap.Init(autoReleaseHeap)
+
 	for {
+		// If there is something to release, set up a release tick
+		// channel.
+		var nextReleaseTick <-chan time.Time
+		if autoReleaseHeap.Len() > 0 {
+			head := (*autoReleaseHeap)[0]
```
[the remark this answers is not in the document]
joostjager: added

invoices/invoiceregistry.go
```diff
@@ -127,13 +149,26 @@ type invoiceEvent struct {
...
+			now := i.now()
+			nextReleaseTick = i.getTick(head.releaseTime.Sub(now))
```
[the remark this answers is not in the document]
joostjager: added

invoices/invoiceregistry.go
```diff
@@ -415,6 +474,90 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+		return err
+	}
+
+	htlc, ok := invoice.Htlcs[key]
+	if !ok {
+		return fmt.Errorf("htlc %v not found", key)
+	}
+	if htlc.State == channeldb.HtlcStateCanceled {
+		i.notifyHodlSubscribers(HodlEvent{
+			CircuitKey:   key,
```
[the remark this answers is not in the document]
joostjager: added

invoices/invoiceregistry.go
joostjager: No, it is correct. An an htlc is now marked as settled, it may be the case that it was previously accepted.

invoices/invoiceregistry.go
joostjager: It does, comment added

lnrpc/rpc.proto
```diff
@@ -2377,6 +2377,9 @@ message InvoiceHTLC {
 
     /// Current state the htlc is in.
     InvoiceHTLCState state = 8 [json_name = "state"];
+
+    /// The total amount of the mpp payment in msat.
+    uint64 mpp_total_amt_msat = 9 [json_name = "mpp_total_amt_msat"];
```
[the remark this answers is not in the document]
joostjager: It is just two lines of non-generated code. For cohesion I kept it in this commit.

invoices/releaseheap.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (h *releaseHeap) Push(x interface{}) {
+	*h = append(*h, x.(*releaseEvent))
+}
+
+func (h *releaseHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
```
cfromknecht: needs `gofmt`?

channeldb/invoices.go
```diff
@@ -1243,78 +1248,73 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
 
 	// Process add actions from update descriptor.
 	for key, htlcUpdate := range update.AddHtlcs {
-		htlc, ok := invoice.Htlcs[key]
-
-		// Add new htlc paying to the invoice.
-		if ok {
-			return nil, fmt.Errorf("htlc %v already exists", key)
+		if _, exists := invoice.Htlcs[key]; exists {
+			return nil, errors.New("duplicate add")
```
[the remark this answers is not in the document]
joostjager: Restored

channeldb/invoices.go
```diff
@@ -1243,78 +1248,73 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
-		for _, htlc := range invoice.Htlcs {
-			if htlc.State != HtlcStateAccepted {
-				continue
+	// Align htlc states with invoice state and recalculate amount paid.
+	var amtPaid lnwire.MilliSatoshi
+	for key, htlc := range invoice.Htlcs {
+		_, cancel := update.CancelHtlcs[key]
+		if cancel {
+			if _, added := update.AddHtlcs[key]; added {
+				return nil, errors.New("added htlc canceled")
```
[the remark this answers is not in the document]
joostjager: Added

channeldb/invoices.go
```diff
@@ -1327,7 +1327,112 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
 	return &invoice, nil
 }
 
-func setSettleFields(settleIndex *bbolt.Bucket, invoiceNum []byte,
+// updateInvoiceState validates and processes an invoice state update.
+func updateInvoiceState(invoice *Invoice, hash lntypes.Hash,
+	update InvoiceStateUpdateDesc) error {
+
+	switch update.NewState {
```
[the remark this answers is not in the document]
joostjager: That is nicer indeed. Copied and pasted.

channeldb/invoices.go
```diff
@@ -1327,7 +1327,112 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+		case HtlcStateAccepted:
+			htlc.State = HtlcStateCanceled
+			htlc.ResolveTime = now
+
+		case HtlcStateSettled:
+			return fmt.Errorf("cannot have a settled htlc with " +
+				"invoice in state canceled")
+		}
+
+	case ContractOpen:
```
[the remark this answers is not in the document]
joostjager: yes indeed, fixed

channeldb/invoices.go
```diff
@@ -1327,7 +1327,112 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+	case ContractAccepted:
+		if htlc.State == HtlcStateSettled {
+			return fmt.Errorf("cannot have a settled htlc with "+
+				"invoice in state %v", invState)
+		}
+	}
```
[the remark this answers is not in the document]
joostjager: added

invoices/releaseheap.go
joostjager: This is what my `gofmt` makes of it and I also don't think travis fails. Will check out the latest CI

invoices/invoiceregistry.go
```diff
@@ -1043,3 +1141,309 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
 
 	delete(i.hodlReverseSubscriptions, subscriber)
 }
+
+// updateResult is the result of the invoice update call.
+type updateResult uint8
+
+const (
+	resultReplayToCanceled updateResult = iota
```
cfromknecht: perhaps add a `resultInvalid` state for 0, to prevent accidental initialization with `resultReplayToCanceled`?

invoices/invoiceregistry.go
```diff
@@ -1043,3 +1141,309 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+
+	// Start building the accept descriptor.
+	acceptDesc := &channeldb.HtlcAcceptDesc{
+		Amt:          ctx.amtPaid,
+		Expiry:       ctx.expiry,
+		AcceptHeight: ctx.currentHeight,
+		MppTotalAmt:  ctx.mpp.TotalMsat(),
+	}
+
+	// Only accept payments to open invoices. This behaviour
```
cfromknecht: comments in this section look < 80 chars

invoices/invoiceregistry.go
```diff
@@ -1043,3 +1141,309 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+
+	return &update, resultSettled, nil
+}
+
+// updateLegacy is a callback for DB.UpdateInvoice that contains the invoice
+// settlement logic for legacy payments.
+func updateLegacy(ctx *invoiceUpdateCtx, inv *channeldb.Invoice) (
+	*channeldb.InvoiceUpdateDesc, updateResult, error) {
+
+	// If the invoice is already canceled, there is no further
```
cfromknecht: some comments here also look < 80 chars

invoices/invoiceregistry.go
```diff
@@ -1043,3 +1141,309 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+
+	// Check the payment address that authorizes the
+	// payment.
+	if ctx.mpp.PaymentAddr() != inv.Terms.PaymentAddr {
+		return nil, resultAddressMismatch, nil
+	}
+
+	// If there is a set invoice amount, check that the
+	// total amt of the htlc set is high enough.
+	if inv.Terms.Value > 0 {
```
cfromknecht: is this needed? if `inv.Terms.Value == 0`, then it should still procude the same result if `ctx.mpp.TotalMsat < inv.Terms.Value` is checked directly. also handles cases where `TotalMsat()` is interpreted as negative

invoices/invoiceregistry.go
```diff
@@ -415,135 +478,168 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+			AcceptHeight: int32(htlc.AcceptHeight),
+			Preimage:     nil,
+		})
+	}
+	return nil
+}
 
-		return &update, nil
+// NotifyExitHopHtlc attempts to mark an invoice as settled. If the invoice is a
+// debug invoice, then this method is a noop as debug invoices are never fully
```
cfromknecht: comment re: debug invoice is outdated?

invoices/invoiceregistry.go
```diff
@@ -1043,3 +1141,309 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
 
 	delete(i.hodlReverseSubscriptions, subscriber)
 }
+
+// updateResult is the result of the invoice update call.
```
cfromknecht: this whole section could probably be moved to a separate file. perhaps `update.go`? `updateinvoice.go` would also work, but maybe redundant with the package name.

invoices/invoiceregistry.go
joostjager: added

invoices/invoiceregistry.go
joostjager: done

invoices/invoiceregistry.go

invoices/invoiceregistry.go
joostjager: Yes, clause removed. Also added an additional check for set_amt==0

channeldb/invoices.go
```diff
@@ -1326,7 +1359,119 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+		return errors.New("unknown state transition")
+	}
+
+	invoice.State = update.NewState
+
+	return nil
+}
+
+// cancelSingleHtlc validates cancelation of a single htlc and update its state.
+func cancelSingleHtlc(now time.Time, htlc *InvoiceHTLC,
```
bhandras: May I suggest renaming `now` to `resolveTime` for clarity.

cmd/lncli/commands.go
```diff
@@ -18,6 +18,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
```
halseth: is this commit supposed to be included, or removed?

channeldb/invoices.go
```diff
@@ -1326,7 +1359,119 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+	}
+
+	htlc.State = HtlcStateCanceled
+	htlc.ResolveTime = now
+
+	return nil
+}
+
+// updateHtlc aligns the state of an htlc with the given invoice state.
+func updateHtlc(now time.Time, htlc *InvoiceHTLC,
```
bhandras: Similarly to my previous comment, I'd suggest to rename `now` to `resolveTime`

go.mod
```diff
@@ -44,8 +44,8 @@ require (
 	github.com/rogpeppe/fastuuid v1.2.0 // indirect
 	github.com/tv42/zbase32 v0.0.0-20160707012821-501572607d02
 	github.com/urfave/cli v1.18.0
-	golang.org/x/crypto v0.0.0-20190211182817-74369b46fc67
-	golang.org/x/net v0.0.0-20190206173232-65e2d4e15006
+	golang.org/x/crypto v0.0.0-20191202143827-86a70503ff7e
```
halseth: unintended changes to `go.*` ?

invoices/invoiceregistry.go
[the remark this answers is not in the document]
halseth: not sure what a good value would be. Maybe on the order of 1-2 min?

invoices/update.go
```diff
@@ -0,0 +1,319 @@
+package invoices
```
bhandras: invoices/update.go needs unit tests

invoices/invoiceregistry.go
halseth: Then why skip the HTLC if `htlc.State != channeldb.HtlcStateSettled `? 🤔

queue/min_time_heap.go
```diff
@@ -0,0 +1,50 @@
+package queue
```
bhandras: Talked about this offline,but just leaving this here for others to give opinion: I'd suggest to use the `queue.PriorityQueue` type intorduced in #3694 as that does not require `container/heap` for usage and more general by allowing arbitrary values which have `Less` defined (and by this not forcing a particular order).

invoices/invoiceregistry_test.go
```diff
@@ -3,17 +3,21 @@ package invoices
 import (
```
bhandras: There's quiet a bit of refactor in #3694 to move all test data generation to a separate compulation unit. I'd suggest building on that. Not sure how to do this properly without merging that first though.

invoices/invoiceregistry.go
```diff
@@ -412,6 +489,101 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+		}, nil
+	}
+
+	var updated bool
+	invoice, err := i.cdb.UpdateInvoice(hash,
+		func(invoice *channeldb.Invoice) (
+			*channeldb.InvoiceUpdateDesc, error) {
+
+			update, err := updateInvoice(invoice)
+			updated = err == nil && update != nil
```
halseth: this is not very readable IMO.  Can't we do a normal `if err != nil { return err }` here?

halseth: Sam in the other PR I guess

channeldb/db.go
```diff
@@ -137,7 +137,8 @@ type DB struct {
 	*bbolt.DB
 	dbPath string
 	graph  *ChannelGraph
-	now    func() time.Time
+
+	Now func() time.Time
```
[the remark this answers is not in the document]
joostjager: We already do it like this in two other locations. I didn't notice much problems with a closure really. As you say, if it is extended with additional functions like `time.After` it probably makes sense to have an interface

channeldb/db.go
```diff
@@ -171,7 +172,7 @@ func Open(dbPath string, modifiers ...OptionModifier) (*DB, error) {
 	chanDB := &DB{
 		DB:     bdb,
 		dbPath: dbPath,
-		now:    time.Now,
+		Now:    time.Now,
```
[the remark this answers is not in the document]
joostjager: Yes, this is a shortcut to prevent a test-only dependency from becoming explicit everywhere. Maybe we should indeed make it explicit, but speaking purely functional, this works too?

channeldb/invoice_test.go
```diff
@@ -337,7 +402,7 @@ func TestDuplicateSettleInvoice(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to make test db: %v", err)
 	}
-	db.now = func() time.Time { return time.Unix(1, 0) }
+	db.Now = func() time.Time { return time.Unix(1, 0) }
```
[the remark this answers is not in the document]
joostjager: It would be nice to make that better, but it is also pre-existing.

invoices/update.go
joostjager: It is covered by `invoiceregistry_test.go`. There is indeed an opportunity to now test this code in isolation, but this PR also didn't introduce this code, just moved it.

channeldb/invoices.go
joostjager: Renamed

channeldb/invoices.go
joostjager: Renamed

channeldb/invoice_test.go
```diff
@@ -184,6 +185,70 @@ func TestInvoiceWorkflow(t *testing.T) {
...
+	t.Parallel()
+
+	db, cleanUp, err := makeTestDB()
+	defer cleanUp()
+	if err != nil {
+		t.Fatalf("unable to make test db: %v", err)
+	}
+
+	testInvoice := &Invoice{
+		CreationDate: time.Unix(time.Now().Unix(), 0),
```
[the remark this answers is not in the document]
joostjager: `CreationDate` removed completely, as it plays no role in this test.

cmd/lncli/commands.go
joostjager: Removed, eventually. I keep it here for easy testing (possibly by reviewers :stuck_out_tongue: )

invoices/invoiceregistry.go
joostjager: Fixed

channeldb/db.go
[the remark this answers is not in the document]
bhandras: An other option to make it explicit for tests only by adding a `TestOpen(..)` function with an appropriate comment. but in general all dependencies should be injected.

queue/min_time_heap.go
joostjager: `queue.PriorityQueue` cherry picked

invoices/invoiceregistry_test.go
joostjager: It is mostly rename/move it seems. I think it can be done independently?

go.mod
joostjager: removed

invoices/invoiceregistry.go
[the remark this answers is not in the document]
joostjager: Updated to 2 mins

invoices/invoiceregistry.go
joostjager: Cherry picked prio queue from #3694

invoices/invoiceregistry.go
joostjager: The invoice state is leading. If an htlc is marked as settled, we should follow now and settle the htlc with our peer. But only if it is indeed marked as settled. Comment added

channeldb/db.go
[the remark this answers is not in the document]
joostjager: I checked out the changes required: https://github.com/lightningnetwork/lnd/compare/master...joostjager:explicit-now-dependency-channeldb

It is an okay change I think, but I don't really want to do it for #3770 now that it is close to merge.

channeldb/db.go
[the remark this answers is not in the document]
bhandras: One nice thing about using the interface that once it's spread, refactor will be simpler as well as it may lead to better code hygene (common test clock implementation maybe?)

channeldb/invoice_test.go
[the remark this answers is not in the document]
bhandras: Or other option if you choose to stay with the `Now` closure is to still inject it in `makeTestDB()` or create a default `Now()` function for all tests and modfity `makeTestDB()` instead.

invoices/invoiceregistry_test.go
bhandras: I'm fine merging this if this PR is merged to master earlier. Optionally we could somehow make it part of this PR too (it's in commit: f663b33).

invoices/update.go
bhandras: Sorry, missed that this is a dependent PR. Discussed this offline too, but to restate I'm on the opinion that separate units should be tested separately if possible.

joostjager:
Itest fail, problem is that `PaymentAddr` doesn't match

invoices/invoiceregistry.go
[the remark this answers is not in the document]
joostjager: I also pass it in from `server.go` now.

invoices/releaseheap.go
joostjager: Removed `0:` and now it looks ok.

joostjager:
Rebased on top of @cfromknecht's commits that create the payment address.

invoices/invoiceregistry.go
joostjager: Has been merged as part of #3770

queue/priority_queue.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (pq *PriorityQueue) Push(item PriorityQueueItem) {
+	heap.Push(&pq.queue, item)
+}
+
+func (pq *PriorityQueue) Pop() PriorityQueueItem {
+	return heap.Pop(&pq.queue).(PriorityQueueItem)
+}
+
+func (pq *PriorityQueue) Top() PriorityQueueItem {
```
halseth: `Peek`?

invoices/invoiceregistry.go
halseth: Still not fixed?

invoices/invoiceregistry.go
```diff
@@ -419,6 +497,101 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+
+			return update, err
+		},
+	)
+	if err != nil {
+		return err
+	}
+	if !updated {
```
halseth: no comments on the various cases in this methods makes it a bit dense and hard to follow.

invoices/invoiceregistry.go
```diff
@@ -419,6 +497,101 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+	if htlc.State == channeldb.HtlcStateCanceled {
+		i.notifyHodlSubscribers(HodlEvent{
+			CircuitKey:   key,
+			AcceptHeight: int32(htlc.AcceptHeight),
+			Preimage:     nil,
+		})
+	}
+	return nil
+}
+
```
halseth: could maybe update the godoc of `NotifyExitHopHtlc` now?

invoices/invoiceregistry_test.go
```diff
@@ -678,3 +691,125 @@ type mockPayload struct {
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if event != nil {
+		t.Fatal("expected no direct resolution")
+	}
+
+	// Wait for timer to be set up.
+	time.Sleep(100 * time.Millisecond)
```
halseth: could have a channel signal it is ready instead

invoices/update.go
```diff
@@ -99,8 +129,122 @@ func updateInvoice(ctx *invoiceUpdateCtx, inv *channeldb.Invoice) (
...
+	// is a zero-valued invoice, it will always be enough.
+	if ctx.mpp.TotalMsat() < inv.Terms.Value {
+		return nil, resultHtlcSetTotalTooLow, nil
+	}
+
+	// Check whether total amt matches other htlcs in the set.
+	var newSetTotal lnwire.MilliSatoshi
+	for _, htlc := range inv.Htlcs {
+		// Only consider accepted mpp htlcs.
+		if htlc.State != channeldb.HtlcStateAccepted {
```
halseth: could there ever be any other htlc states in the set at this point?

queue/priority_queue.go
```diff
@@ -0,0 +1,66 @@
...
+	return pq[i].Less(pq[j])
+}
+
+// Swap swaps two items in the ExpiryQueue. Swap is used by heap.Interface.
+func (pq priorityQueue) Swap(i, j int) {
+	pq[i], pq[j] = pq[j], pq[i]
+}
+
+// Push adds a new item the the ExpiryQueue.
+func (pq *priorityQueue) Push(x interface{}) {
```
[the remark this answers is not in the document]
joostjager: It tried to change it, but it needs dereferences. The "official" go example also mixes values and pointers.

queue/priority_queue.go
```diff
@@ -0,0 +1,66 @@
...
+	"container/heap"
+)
+
+type PriorityQueueItem interface {
+	Less(other PriorityQueueItem) bool
+}
+
+type priorityQueue []PriorityQueueItem
+
+// Len returns the length of the ExpiryQueue.
```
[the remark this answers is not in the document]
joostjager: fixed

queue/priority_queue.go
```diff
@@ -0,0 +1,66 @@
...
+func (pq *priorityQueue) Pop() interface{} {
+	old := *pq
+	n := len(old)
+	item := old[n-1]
+	old[n-1] = nil
+	*pq = old[0 : n-1]
+	return item
+}
+
+type PriorityQueue struct {
```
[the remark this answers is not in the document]
joostjager: Added

queue/priority_queue.go
joostjager: I went with @bhandras naming

invoices/invoiceregistry.go
```diff
@@ -50,6 +58,21 @@ type InvoiceRegistryConfig struct {
 	// expiry of any invoice we create and the code effectuating this should
 	// not be hit.
 	FinalCltvRejectDelta int32
+
+	// HtlcHoldDuration defines for how long mpp htlcs are held while
+	// waiting for the other set members to arrive.
+	HtlcHoldDuration time.Duration
+}
+
+// releaseEvent describes an htlc auto-release event.
```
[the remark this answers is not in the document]
joostjager: added

invoices/invoiceregistry.go
```diff
@@ -50,6 +58,21 @@ type InvoiceRegistryConfig struct {
...
+type releaseEvent struct {
+	hash        lntypes.Hash
+	key         channeldb.CircuitKey
+	releaseTime time.Time
+}
+
+func (r *releaseEvent) Less(other queue.PriorityQueueItem) bool {
```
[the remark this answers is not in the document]
joostjager: added

invoices/invoiceregistry.go
joostjager: Needed a rebase, fixed now.

joostjager: Ah, this is the same thing in a different method

joostjager: Really fixed now

invoices/invoiceregistry.go
joostjager: added

invoices/invoiceregistry.go
joostjager: Lines added about mpp

invoices/invoiceregistry_test.go
joostjager: modified so that tick channels also tick immediately when the trigger time is already there.

invoices/update.go
joostjager: Yes, htlcs that were canceled previously because they timed out.

joostjager: Comment added

joostjager:
Note to self: don't forget to remove debug commits

queue/priority_queue.go
bhandras: I think the usual naming is top, at least many ADT level definitions refer to it as top. (my choice was based on C++ https://en.cppreference.com/w/cpp/container/priority_queue)

queue/priority_queue.go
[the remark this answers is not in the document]
bhandras: for reference: https://golang.org/pkg/container/heap/

invoices/invoiceregistry.go
```diff
@@ -41,6 +41,17 @@ type HodlEvent struct {
 	AcceptHeight int32
 }
 
+// InvoiceRegistryConfig contains the configuration parameters for invoice
+// registry.
+type InvoiceRegistryConfig struct {
```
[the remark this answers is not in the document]
joostjager: done

joostjager:
debug commits removed

queue/priority_queue.go
```diff
@@ -0,0 +1,70 @@
+package queue
+
+import (
+	"container/heap"
+)
+
+type PriorityQueueItem interface {
```
halseth: nit: exported type should have godoc

joostjager:
@cfromknecht brought up the [mpp timeout failure](https://github.com/lightningnetwork/lightning-rfc/pull/643/files#diff-9198bb316a3387cc67fd543b03339b35R984). Will address that in a follow up in coordination with @carlaKC.

Without this error returned, senders that actually support sending multiple shards _and_ interpret this new failure will not function optimally. But seems acceptable for where we are right now.

cmd/lncli/commands.go
joostjager: Removed now

invoices/invoiceregistry_test.go
```diff
@@ -683,3 +693,83 @@ func (p *mockPayload) MultiPath() *record.MPP {
...
+		t.Fatal(err)
+	}
+	if event != nil {
+		t.Fatal("expected no direct resolution")
+	}
+
+	// Simulate mpp timeout releasing htlc 1.
+	ctx.clock.setTime(testTime.Add(30 * time.Second))
+
+	hodlEvent := (<-hodlChan1).(HodlEvent)
```
halseth: nit: should have a timeout?

joostjager: my favorite `defer time(t)()` added :)

invoices/invoiceregistry.go
```diff
@@ -41,6 +49,43 @@ type HodlEvent struct {
 	AcceptHeight int32
```
bhandras: It's a bit difficult to see what this type is suppsed to do without looking at the comment. May I suggest to rename it to something more descriptive? eg. `htlcAutoReleaseEvent`. And maybe also for members. What does `hash` and `key` point to?

invoices/invoiceregistry.go
```diff
@@ -196,6 +259,29 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
 				i.singleNotificationClients[e.id] = e
```
bhandras: Can this (htlc not found in the invoice) ever happen though?

invoices/invoiceregistry.go
```diff
@@ -196,6 +259,29 @@ func (i *InvoiceRegistry) invoiceEventNotifier() {
...
 			}
 
+		// A new htlc came in for auto-release.
+		case event := <-i.htlcAutoReleaseChan:
+			log.Debugf("Scheduling auto-release for htlc: "+
```
bhandras: Shouldn't we use either signed or unsigned types for block height everywhere?

invoices/clock_test.go
```diff
@@ -0,0 +1,75 @@
+package invoices
+
+import (
+	"sync"
+	"time"
+)
+
+// testClock can be used in tests to mock time.
```
bhandras: Really nice! May I suggest you to extend `clock.Clock` interface and use that as a common ground instead? Maybe also move this test clock in that package such that others can use it?

invoices/invoiceregistry_test.go
```diff
@@ -59,19 +61,29 @@ var (
 
 type testContext struct {
 	registry *InvoiceRegistry
+	clock    *testClock
 
 	cleanup func()
 	t       *testing.T
 }
 
 func newTestContext(t *testing.T) *testContext {
```
bhandras: I think it's important to test all cases where settlement may fail as well.

invoices/invoiceregistry_test.go
```diff
@@ -674,3 +693,85 @@ func (p *mockPayload) MultiPath() *record.MPP {
 func (p *mockPayload) CustomRecords() hop.CustomRecordSet {
 	return make(hop.CustomRecordSet)
 }
+
+// TestSettleMpp tests settling of an invoice with multiple partial payments.
+func TestSettleMpp(t *testing.T) {
```
bhandras: Looks like the file changed during review. The above comment meant to be here: "I think it's important to test cases where settlement may fail".

queue/priority_queue.go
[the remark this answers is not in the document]
joostjager: Go doc copied

invoices/invoiceregistry.go
```diff
@@ -41,6 +49,43 @@ type HodlEvent struct {
...
+}
+
+// Less is used to order PriorityQueueItem's by their release time such that
+// items with the older release time are at the top of the queue.
+//
+// NOTE: Part of the queue.PriorityQueueItem interface.
+func (r *releaseEvent) Less(other queue.PriorityQueueItem) bool {
+	return r.releaseTime.Before(other.(*releaseEvent).releaseTime)
+}
+
```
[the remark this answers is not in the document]
joostjager: moved

invoices/clock_test.go
```diff
@@ -0,0 +1,75 @@
+package invoices
```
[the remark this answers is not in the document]
joostjager: I swapped around the words, because `_test.go` files are not compiled into non-test builds.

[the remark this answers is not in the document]
joostjager: Testing the test code I'll leave for another time.

invoices/clock_test.go
joostjager: I don't want to delay this PR with a discussion on where to put that package/file. Given the time those took in past prs I'd rather avoid it (reference class forecasting).

invoices/clock_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+// testClock can be used in tests to mock time.
+type testClock struct {
+	currentTime time.Time
+	chans       map[time.Time][]chan time.Time
```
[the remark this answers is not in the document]
joostjager: fixed

invoices/clock_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// setTime sets the (test) time and triggers tick channels when they expire.
+func (c *testClock) setTime(now time.Time) {
+	c.timeLock.Lock()
+	defer c.timeLock.Unlock()
+
+	c.currentTime = now
+	remainingChans := make(map[time.Time][]chan time.Time)
+	for t, chans := range c.chans {
```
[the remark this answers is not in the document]
joostjager: done

invoices/clock_test.go
```diff
@@ -0,0 +1,75 @@
...
+		if t.After(now) {
```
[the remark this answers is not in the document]
joostjager: Comment added to clarify.

invoices/invoiceregistry.go
joostjager: Yes, that is inconsistent at the moment. But if you start changing, you will expand into many different areas and also replace one case by the other sometimes.

invoices/invoiceregistry.go
joostjager: This comment seems to be misplaced by github. I think this is about `cancelSingleInvoice`? Yes, if we scheduled a timer, the htlc should have been record on the invoice.

bhandras: Hmm, looks like the PR was changed during my review. I'll place this comment to where it belongs.

invoices/invoiceregistry_test.go
joostjager: There is a case where the partial payment times out. Which other cases do you mean? Of course the total number of possible test cases is large, but I tried to cover at least the ones that usually occur.

invoices/invoiceregistry.go
```diff
@@ -412,6 +498,114 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+	}
+	if !updated {
+		return nil
+	}
+
+	// The invoice has been updated. Notify subscribers of the htlc
+	// resolution.
+	htlc, ok := invoice.Htlcs[key]
+	if !ok {
+		return fmt.Errorf("htlc %v not found", key)
```
bhandras: Can this ever happen though?

joostjager: I don't think so. Perhaps we will do invoice cleanup in the future?

invoices/invoiceregistry.go
joostjager: comments added

PR #3418, opened 2019-08-20
routing, channeldb: several optimizations for path finding

champo:
This PR brings down path finding time by using the channel cache and some other small optimizations. The goal was to bring the time down to ~10ms range. Using a benchmark, described below, it started at 220ms/72mb per path finding operation and with these changes it comes down to 9ms/1.1mb.

## Benchmarking

I've written [a benchmark](https://github.com/lightningnetwork/lnd/commit/fd34d90b89d500118ab10763c408cc55c2b969cf) based on a dump of the main net graph (its a few months old now) in order to test these changes. The bench is roughly as follows:
 - The first time only, a DB is built using the graph dump JSON
 - The DB is used for a simple path finding run to prime the channel cache
 - Path finding runs in a nested benchmark as many times as `go test` sees fit

A second bench, almost equal to the first one runs the nested benchmark in parallel to stress concurrency.

If you wish to run it, first uncomment the DB generation code and run the test with a ~20m timeout to let it generate the full DB. Then comment that again and run your benchmarks.

## Findings

Bench @ master
```
BenchmarkBigGraphCachePerformance/findPath-4         	     100	 203460951 ns/op	72351182 B/op	 1459914 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	      50	 204061673 ns/op	72480387 B/op	 1460464 allocs/op
```

At first glance path finding spent a lot of its time in three things:
 - Reading from the DB
 - Parsing the contents read
 - GC'ing

Adding the channel cache brings down the time and allocations considerably:
```
BenchmarkBigGraphCachePerformance/findPath-4         	     500	  27721522 ns/op	 9830663 B/op	  155751 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	     200	  65236737 ns/op	 9910162 B/op	  156245 allocs/op
```

This is mostly because it avoids hitting disk and allocating new copies of channels. The wire format parsing code seems to be heavy on small allocations, so avoiding that is also a huge win.

The next find was that the hot path has a log on trace level. That log creates a lot of small allocations regardless of the current log level. Adding a log level check around it avoids that for most configs around and removes quite a bit of allocations, also bringing down the runtime by 18%:
```
BenchmarkBigGraphCachePerformance/findPath-4         	    1000	  22749674 ns/op	 7599558 B/op	   97047 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	     200	  59610474 ns/op	 7676371 B/op	   97534 allocs/op
```

The `distance` map was being pre-populated with all nodes, consuming about 20% of the path finding time. Turns out the code used this fact in only one place to check if the candidate distance was better than the existing one. Changing that code to check for existance first let us remove the pre-population:
```
BenchmarkBigGraphCachePerformance/findPath-4         	    1000	  13426902 ns/op	 4516310 B/op	   19026 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	     200	  62240894 ns/op	 4592095 B/op	   19509 allocs/op
```

Path finding uses several structures that grow to hold references to a big part of the graph. Tweaking the capacity for those resulted in lower memory usage and some modest speed bump:
```
BenchmarkBigGraphCachePerformance/findPath-4         	    1000	  12426815 ns/op	 3432225 B/op	   18771 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	     300	  55189725 ns/op	 3507752 B/op	   19248 allocs/op
```

There were 3 structures used: `distance`, `next` and `distanceHeap`. `distance` and `next` could be merged by adding a `nextHop` pointer to `nodeWithDist` this is updated whenever a cheaper path is found. Also, `nodeWithDist` is a big struct now and `distanceHeap` held it by value, changing it to use a pointer removed a lot of copying work. Both these changes reduced memory usage by about half and runtime by 20%.
```
BenchmarkBigGraphCachePerformance/findPath-4         	    2000	  10203189 ns/op	 1639554 B/op	   18759 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	     300	  55169048 ns/op	 1716604 B/op	   19259 allocs/op
```

The last insight was that a function introduced to access the channel cache was returning a `*ChannelEdge`. That struct is small and the pointer was introducing an allocation that added up since its part of the hot path. Changing that function to return by value reduced time by 10% and memory by 39%.
```
BenchmarkBigGraphCachePerformance/findPath-4         	    2000	   9136682 ns/op	 1171369 B/op	    4083 allocs/op
BenchmarkBigGraphCacheConcurrency/findPath-4         	     300	  50998898 ns/op	 1247789 B/op	    4579 allocs/op
```

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/routing/heap.go
+++ b/routing/heap.go
@@ -50,9 +50,10 @@ type distanceHeap struct {
 
 // newDistanceHeap initializes a new distance heap. This is required because
 // we must initialize the pubkeyIndices map for path-finding optimizations.
-func newDistanceHeap() distanceHeap {
+func newDistanceHeap(numNodes int) distanceHeap {
 	distHeap := distanceHeap{
-		pubkeyIndices: make(map[route.Vertex]int),
+		pubkeyIndices: make(map[route.Vertex]int, numNodes),
+		nodes:         make([]nodeWithDist, 0, numNodes),
 	}
 
 	return distHeap
--- a/routing/heap.go
+++ b/routing/heap.go
@@ -3,6 +3,7 @@ package routing
 import (
 	"container/heap"
 
+	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing/route"
 )
@@ -35,12 +36,15 @@ type nodeWithDist struct {
 	// Includes the routing fees and a virtual cost factor to account for
 	// time locks.
 	weight int64
+
+	// nextHop is the edge this route comes from.
+	nextHop *channeldb.ChannelEdgePolicy
 }
 
 // distanceHeap is a min-distance heap that's used within our path finding
 // algorithm to keep track of the "closest" node to our source node.
 type distanceHeap struct {
-	nodes []nodeWithDist
+	nodes []*nodeWithDist
 
 	// pubkeyIndices maps public keys of nodes to their respective index in
 	// the heap. This is used as a way to avoid db lookups by using heap.Fix
@@ -53,7 +57,7 @@ type distanceHeap struct {
 func newDistanceHeap(numNodes int) distanceHeap {
 	distHeap := distanceHeap{
 		pubkeyIndices: make(map[route.Vertex]int, numNodes),
-		nodes:         make([]nodeWithDist, 0, numNodes),
+		nodes:         make([]*nodeWithDist, 0, numNodes),
 	}
 
 	return distHeap
@@ -85,7 +89,7 @@ func (d *distanceHeap) Swap(i, j int) {
 //
 // NOTE: This is part of the heap.Interface implementation.
 func (d *distanceHeap) Push(x interface{}) {
-	n := x.(nodeWithDist)
+	n := x.(*nodeWithDist)
 	d.nodes = append(d.nodes, n)
 	d.pubkeyIndices[n.node] = len(d.nodes) - 1
 }
@@ -97,6 +101,7 @@ func (d *distanceHeap) Push(x interface{}) {
 func (d *distanceHeap) Pop() interface{} {
 	n := len(d.nodes)
 	x := d.nodes[n-1]
+	d.nodes[n-1] = nil
 	d.nodes = d.nodes[0 : n-1]
 	delete(d.pubkeyIndices, x.node)
 	return x
@@ -107,7 +112,7 @@ func (d *distanceHeap) Pop() interface{} {
 // modify its position and reorder the heap. If the vertex does not already
 // exist in the heap, then it is pushed onto the heap. Otherwise, we will end
 // up performing more db lookups on the same node in the pathfinding algorithm.
-func (d *distanceHeap) PushOrFix(dist nodeWithDist) {
+func (d *distanceHeap) PushOrFix(dist *nodeWithDist) {
 	index, ok := d.pubkeyIndices[dist.node]
 	if !ok {
 		heap.Push(d, dist)
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -552,7 +552,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// better than the current best known distance to this node.
 		// The new better distance is recorded, and also our "next hop"
 		// map is populated with this edge.
-		distance[fromVertex] = nodeWithDist{
+		withDist := nodeWithDist{
 			dist:            tempDist,
 			weight:          tempWeight,
 			node:            fromVertex,
@@ -560,13 +560,14 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 			incomingCltv:    incomingCltv,
 			probability:     probability,
 		}
+		distance[fromVertex] = withDist
 
 		next[fromVertex] = edge
 
-		// Either push distance[fromVertex] onto the heap if the node
+		// Either push withDist onto the heap if the node
 		// represented by fromVertex is not already on the heap OR adjust
 		// its position within the heap via heap.Fix.
-		nodeHeap.PushOrFix(distance[fromVertex])
+		nodeHeap.PushOrFix(withDist)
 	}
 
 	// TODO(roasbeef): also add path caching
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -355,12 +355,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	nodeHeap := newDistanceHeap(estimatedNodeCount)
 
 	// Holds the current best distance for a given node.
-	distance := make(map[route.Vertex]nodeWithDist, estimatedNodeCount)
-
-	// We'll use this map as a series of "next" hop pointers. So to get
-	// from `Vertex` to the target node, we'll take the edge that it's
-	// mapped to within `next`.
-	next := make(map[route.Vertex]*channeldb.ChannelEdgePolicy, estimatedNodeCount)
+	distance := make(map[route.Vertex]*nodeWithDist, estimatedNodeCount)
 
 	additionalEdgesWithSrc := make(map[route.Vertex][]*edgePolicyWithSource)
 	for vertex, outgoingEdgePolicies := range g.additionalEdges {
@@ -386,7 +381,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	// is the starting point of the graph traversal. We are searching
 	// backwards to get the fees first time right and correctly match
 	// channel bandwidth.
-	distance[target] = nodeWithDist{
+	distance[target] = &nodeWithDist{
 		dist:            0,
 		weight:          0,
 		node:            target,
@@ -552,17 +547,16 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// better than the current best known distance to this node.
 		// The new better distance is recorded, and also our "next hop"
 		// map is populated with this edge.
-		withDist := nodeWithDist{
+		withDist := &nodeWithDist{
 			dist:            tempDist,
 			weight:          tempWeight,
 			node:            fromVertex,
 			amountToReceive: amountToReceive,
 			incomingCltv:    incomingCltv,
 			probability:     probability,
+			nextHop:         edge,
 		}
 		distance[fromVertex] = withDist
-
-		next[fromVertex] = edge
 
 		// Either push withDist onto the heap if the node
 		// represented by fromVertex is not already on the heap OR adjust
@@ -582,7 +576,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 
 		// Fetch the node within the smallest distance from our source
 		// from the heap.
-		partialPath := heap.Pop(&nodeHeap).(nodeWithDist)
+		partialPath := heap.Pop(&nodeHeap).(*nodeWithDist)
 		pivot := partialPath.node
 
 		// If we've reached our source (or we don't have any incoming
@@ -632,7 +626,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 
 			// Check if this candidate node is better than what we
 			// already have.
-			processEdge(route.Vertex(chanSource), edgeBandwidth, inEdge, pivot)
+			processEdge(chanSource, edgeBandwidth, inEdge, pivot)
 			return nil
 		}
 
@@ -656,26 +650,24 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		}
 	}
 
-	// If the source node isn't found in the next hop map, then a path
-	// doesn't exist, so we terminate in an error.
-	if _, ok := next[source]; !ok {
-		return nil, newErrf(ErrNoPathFound, "unable to find a path to "+
-			"destination")
-	}
-
-	// Use the nextHop map to unravel the forward path from source to
+	// Use the distance map to unravel the forward path from source to
 	// target.
 	var pathEdges []*channeldb.ChannelEdgePolicy
 	currentNode := source
 	for currentNode != target { // TODO(roasbeef): assumes no cycles
 		// Determine the next hop forward using the next map.
-		nextNode := next[currentNode]
+		currentNodeWithDist, ok := distance[currentNode]
+		if !ok {
+			// If the node doesnt have a next hop it means we didn't find a path.
+			return nil, newErrf(ErrNoPathFound, "unable to find a "+
+				"path to destination")
+		}
 
 		// Add the next hop to the list of path edges.
-		pathEdges = append(pathEdges, nextNode)
+		pathEdges = append(pathEdges, currentNodeWithDist.nextHop)
 
 		// Advance current node.
-		currentNode = route.Vertex(nextNode.Node.PubKeyBytes)
+		currentNode = currentNodeWithDist.nextHop.Node.PubKeyBytes
 	}
 
 	// The route is invalid if it spans more than 20 hops. The current
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec"
 	"github.com/coreos/bbolt"
 
 	"github.com/lightningnetwork/lnd/channeldb"
@@ -324,56 +325,35 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	// traversal.
 	nodeHeap := newDistanceHeap()
 
-	// For each node in the graph, we create an entry in the distance map
-	// for the node set with a distance of "infinity". graph.ForEachNode
-	// also returns the source node, so there is no need to add the source
-	// node explicitly.
+	// Holds the current best distance for a given node.
 	distance := make(map[route.Vertex]nodeWithDist)
-	if err := g.graph.ForEachNode(tx, func(_ *bbolt.Tx,
-		node *channeldb.LightningNode) error {
-		// TODO(roasbeef): with larger graph can just use disk seeks
-		// with a visited map
-		vertex := route.Vertex(node.PubKeyBytes)
-		distance[vertex] = nodeWithDist{
-			dist: infinity,
-			node: route.Vertex(node.PubKeyBytes),
+
+	if r.DestPayloadTLV {
+		// Check if the target has TLV enabled
+
+		targetKey, err := btcec.ParsePubKey(target[:], btcec.S256())
+		if err != nil {
+			return nil, err
 		}
 
-		// If we don't have any features for this node, then we can
-		// stop here.
-		if node.Features == nil || !r.DestPayloadTLV {
-			return nil
+		targetNode, err := g.graph.FetchLightningNode(targetKey)
+		if err != nil {
+			return nil, err
 		}
 
-		// We only need to perform this check for the final node, so we
-		// can exit here if this isn't them.
-		if vertex != target {
-			return nil
+		if targetNode.Features != nil {
+			supportsTLV := targetNode.Features.HasFeature(
+				lnwire.TLVOnionPayloadOptional,
+			)
+			if !supportsTLV {
+				return nil, fmt.Errorf("destination hop doesn't " +
+					"understand new TLV paylods")
+			}
 		}
-
-		// If we have any records for the final hop, then we'll check
-		// not to ensure that they are actually able to interpret them.
-		supportsTLV := node.Features.HasFeature(
-			lnwire.TLVOnionPayloadOptional,
-		)
-		if !supportsTLV {
-			return fmt.Errorf("destination hop doesn't " +
-				"understand new TLV paylods")
-		}
-
-		return nil
-	}); err != nil {
-		return nil, err
 	}
 
 	additionalEdgesWithSrc := make(map[route.Vertex][]*edgePolicyWithSource)
 	for vertex, outgoingEdgePolicies := range g.additionalEdges {
-		// We'll also include all the nodes found within the additional
-		// edges that are not known to us yet in the distance map.
-		distance[vertex] = nodeWithDist{
-			dist: infinity,
-			node: vertex,
-		}
 
 		// Build reverse lookup to find incoming edges. Needed because
 		// search is taken place from target to source.
@@ -391,11 +371,11 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	}
 
 	// We can't always assume that the end destination is publicly
-	// advertised to the network and included in the graph.ForEachNode call
-	// above, so we'll manually include the target node. The target node
-	// charges no fee. Distance is set to 0, because this is the starting
-	// point of the graph traversal. We are searching backwards to get the
-	// fees first time right and correctly match channel bandwidth.
+	// advertised to the network so we'll manually include the target node.
+	// The target node charges no fee. Distance is set to 0, because this
+	// is the starting point of the graph traversal. We are searching
+	// backwards to get the fees first time right and correctly match
+	// channel bandwidth.
 	distance[target] = nodeWithDist{
 		dist:            0,
 		weight:          0,
@@ -551,7 +531,8 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// route, return. It is important to also return if the distance
 		// is equal, because otherwise the algorithm could run into an
 		// endless loop.
-		if tempDist >= distance[fromVertex].dist {
+		current, ok := distance[fromVertex]
+		if ok && tempDist >= current.dist {
 			return
 		}
 
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -448,8 +448,11 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 			fromVertex, toNode, amountToSend,
 		)
 
-		log.Tracef("path finding probability: fromnode=%v, tonode=%v, "+
-			"probability=%v", fromVertex, toNode, edgeProbability)
+		log.Trace(newLogClosure(func() string {
+			return fmt.Sprintf("path finding probability: fromnode=%v,"+
+				" tonode=%v, probability=%v", fromVertex, toNode,
+				edgeProbability)
+		}))
 
 		// If the probability is zero, there is no point in trying.
 		if edgeProbability == 0 {
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -37,6 +37,11 @@ const (
 	// some effect with smaller time lock values. The value may need
 	// tweaking and/or be made configurable in the future.
 	RiskFactorBillionths = 15
+
+	// estimatedNodeCount is used to preallocate the path finding structures
+	// to avoid resizing and copies. It should be number on the same order as
+	// the number of active nodes in the network.
+	estimatedNodeCount = 10000
 )
 
 // pathFinder defines the interface of a path finding algorithm.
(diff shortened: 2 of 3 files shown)


channeldb/graph.go
```diff
@@ -2292,6 +2292,49 @@ func nodeTraversal(tx *bbolt.Tx, nodePub []byte, db *DB,
 		copy(nodeStart[:], nodePub)
 		copy(nodeStart[33:], chanStart[:])
 
+		fetchChannel := func(chanID []byte) (*ChannelEdge, error) {
+
+			intChanID := byteOrder.Uint64(chanID)
+
+			// db isn't assured to be present sadly
```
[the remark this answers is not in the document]
champo: I think a recently merged PR makes this removable. I had these changes wip for a while 😬

routing/pathfind.go
```diff
@@ -405,8 +406,10 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 			fromVertex, toNode, amountToSend,
 		)
 
-		log.Tracef("path finding probability: fromnode=%v, tonode=%v, "+
-			"probability=%v", fromVertex, toNode, edgeProbability)
+		if log.Level() <= btclog.LevelTrace {
+			log.Tracef("path finding probability: fromnode=%v, tonode=%v, "+
```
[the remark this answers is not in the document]
champo: I haven't got an exact answer. The working theory is that it generates some allocations to hold the var args that are too small to care most of the time but that it adds up because it's called a lot of times per `findPath`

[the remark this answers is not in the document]
joostjager: That would be easy to test by adding a test function locally in `pathfind.go` with var args and call that instead of `log.Tracef`. You could then also modify it to take four (non var) args and see the difference.

champo:
The overall memory usage should remain similar. The cache is limited to a number of channels set by a config value. If the number of channels exceeds that value, this code will still fallback to reading from the DB. If the graph is big in relation to the cache, the performance gains will be lower.

I'll address the coments in the next few days 🙂

channeldb/graph.go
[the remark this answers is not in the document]
champo: This function is called by `LightningNode.ForEachChannel`, which is in turn called from a quite a few places. Running this code in production I had a few seg faults due to `db` here. I couldn't quite figure out where the problem was coming from. There's quite a few places where `LightningNode` structs are created without setting `db` that could be causing those crashes.

routing/pathfind.go
```diff
@@ -308,23 +308,8 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
-			dist: infinity,
-			node: route.Vertex(node.PubKeyBytes),
-		}
-		return nil
-	}); err != nil {
-		return nil, err
-	}
 
 	additionalEdgesWithSrc := make(map[route.Vertex][]*edgePolicyWithSource)
 	for vertex, outgoingEdgePolicies := range g.additionalEdges {
```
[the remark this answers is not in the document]
champo: Yes! Removed :D

routing/pathfind.go
```diff
@@ -312,7 +312,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	// for the node set with a distance of "infinity". graph.ForEachNode
 	// also returns the source node, so there is no need to add the source
 	// node explicitly.
-	distance := make(map[route.Vertex]nodeWithDist)
+	distance := make(map[route.Vertex]nodeWithDist, 10000)
```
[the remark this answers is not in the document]
champo: Updated to read from the graph, it had almost no impact in perfomance to calculate it.

channeldb/graph.go
```diff
@@ -2292,6 +2292,49 @@ func nodeTraversal(tx *bbolt.Tx, nodePub []byte, db *DB,
...
+			channel := &ChannelEdge{
+				Info:    &edgeInfo,
+				Policy1: edge1,
+				Policy2: edge2,
+			}
+
+			if db != nil {
+				// give back to the cache
+				db.graph.chanCache.insert(intChanID, *channel)
+			}
```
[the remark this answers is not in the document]
champo: Yeah, there was also several other places with similar patterns that now use the cache too.

channeldb/graph.go
```diff
@@ -2303,33 +2346,31 @@ func nodeTraversal(tx *bbolt.Tx, nodePub []byte, db *DB,
...
+			channel, err := fetchChannel(chanID)
 			if err != nil {
 				return err
 			}
 
-			otherNode, err := edgeInfo.OtherNodeKeyBytes(nodePub)
-			if err != nil {
-				return err
+			var incomingPolicy, outgoingPolicy *ChannelEdgePolicy
+			if channel.Policy1 != nil && channel.Policy1.Node != nil {
```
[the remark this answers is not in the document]
champo: This used to happen because the cache was populated with an incomplete policy struct. While refactoring the common function with `ChanUpdatesInHorizon` I found the cause and made so that `.Node` is always present. 

Also, the code now uses `channel.Info.NodeKey1Bytes` and is much simpler!

champo:
@joostjager updated with the feedback and rebasing against latest master. I've fixed a bug along the way relating to edge policies with missing nodes. Lastly, a few smaller changes to bring memory usage down by about 1mb.

routing/pathfind.go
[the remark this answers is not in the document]
champo: It's definitely some overhead due to var args. The fixed arguments version had the same performance as the if. However, the var arg version had the same performance as calling `log.Tracef` directly.

champo:
Added 4 new changes bringing down the time to 9ms and memory usage to 1.1mb. This is the last of the updated I'll send, I've run out of ideas I can easily implement.

routing/pathfind.go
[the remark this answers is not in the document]
joostjager: Ok interesting. Maybe you could use a `logClosure` here to prevent the if statement.

routing/pathfind.go
```diff
@@ -667,33 +653,32 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
-	// Use the nextHop map to unravel the forward path from source to
+	// Use the distance map to unravel the forward path from source to
 	// target.
-	pathEdges := make([]*channeldb.ChannelEdgePolicy, 0, len(next))
+	pathEdges := make([]*channeldb.ChannelEdgePolicy, 0)
 	currentNode := source
 	for currentNode != target { // TODO(roasbeef): assumes no cycles
 		// Determine the next hop forward using the next map.
-		nextNode := next[currentNode]
+		nextNode, ok := distance[currentNode]
```
joostjager: change var name `nextNode`?

channeldb/graph.go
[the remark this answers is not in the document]
joostjager: So it cannot be removed despite the recently merged pr that you mentioned? This also means that `LightningNode.ForEachChannel` is not always querying the cache? Is it possible to move this `if db == nil` exception further up the call tree? I don't think you want your pr to sprawl too far, but ideally it shouldn't add any new technical debt in the form of these if statements. `db` is a singleton object that shouldn't be too difficult to obtain anywhere.

To found out where the seg fault comes from, you could just log a stack trace in case the function is entered with `db == nil`.

joostjager:
Good changes again :+1: 

Main question is what to do with the added `if node == nil` and `if db == nil` code. I think that with the right design this should be avoidable, but it probably requires more changes. On the other hand, building on top of something that shouldn't be there in the first place defers paying those costs to the future.

Final comment is about commit structure. In the end, these commits should be reorganized so that it becomes a logical stack. For example (some of these are already a separate commit now):
* all the cache changes in `graph.go` can be done first in one or multiple commits first. 
* add `NumNodes` to `graph.go` and pre-allocate map
* merging nextNode into the path finding heap
* change to nodeWithDist pointers
* remove unused k-shortest code
* remove distance map infinity initialization
* fix log performance

What you generally want to avoid in one pr is to make changes in a commit and change them again later. For example, initializing the map first with 10000 and later replacing it by `NumNodes`. I find it easier if those two are just squashed together and there is one commit that presents the final changes in a particular area.

You could also consider making two prs. One with the internal path finding improvements that are pretty straight-forward to review and get approval on. Then a second one to carefully do the cache.

routing/pathfind.go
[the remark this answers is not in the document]
joostjager: The idea of the closure is that it is only called when the log level is high enough and only then the performance hit is incurred.

```
log.Debug(newLogClosure(func() string {
		return fmt.Sprintf("%v %v %v", a, b, c)
	}))
```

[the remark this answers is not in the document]
joostjager: And perhaps add a comment here explaining for findings around the var args. So that that experiment is captured for future reference.

routing/pathfind.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 
+	"github.com/btcsuite/btcd/btcec"
```
joostjager: Commit message could be rephrased ("if we don't need to"). Nit.

channeldb/graph.go
```diff
@@ -190,6 +190,36 @@ func (c *ChannelGraph) Database() *DB {
 	return c.db
 }
 
+// NumNodes returns the number of nodes currently in the graph
+func (c *ChannelGraph) NumNodes(tx *bbolt.Tx) (int, error) {
+
```
joostjager: remove newline

channeldb/graph.go
```diff
@@ -190,6 +190,36 @@ func (c *ChannelGraph) Database() *DB {
...
+	count := func(tx *bbolt.Tx) error {
+		nodes := tx.Bucket(nodeBucket)
+		if nodes == nil {
+			return ErrGraphNotFound
+		}
+
+		return nodes.ForEach(func(pubKey, _ []byte) error {
+			if len(pubKey) == 33 {
+				result += 1
+			}
```
joostjager: Is there also something with a different length in this bucket?

joostjager: `result++`

channeldb/graph.go
```diff
@@ -190,6 +190,36 @@ func (c *ChannelGraph) Database() *DB {
...
+		})
+	}
+
+	if tx != nil {
+		err = count(tx)
+	} else {
+		err = c.db.View(count)
+	}
+
+	return result, err
```
joostjager: You could consider always returning `0, err` in case of an error. To prevent a more or less undefined result value.

routing/heap.go
```diff
@@ -51,9 +51,9 @@ type distanceHeap struct {
 
 // newDistanceHeap initializes a new distance heap. This is required because
 // we must initialize the pubkeyIndices map for path-finding optimizations.
-func newDistanceHeap() distanceHeap {
+func newDistanceHeap(numNodes int) distanceHeap {
 	distHeap := distanceHeap{
-		pubkeyIndices: make(map[route.Vertex]int),
+		pubkeyIndices: make(map[route.Vertex]int, numNodes),
```
joostjager: Do you also want to set the capacity of the `distanceHeap.nodes` slice?

joostjager: `numNodes` is really an upper bound for the size of the heap. It contains only the active set, the frontier of what path finding is exploring. Worst case the frontier can be all nodes (one center node with all other nodes connected to it), but it will be less typically. I don't think it is problem and also wouldn't know how to get a better estimated of the required size for the heap. But maybe this is something that can be explained (if the reasoning is correct) in a comment.

routing/pathfind.go
```diff
@@ -584,12 +578,15 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 
 	// To start, our target node will the sole item within our distance
 	// heap.
-	heap.Push(&nodeHeap, distance[target])
+	heap.Push(&nodeHeap, &nodeWithDist{
+		node: target,
+		dist: 0,
+	})
```
joostjager: If you don't want the map lookup here, it may be better to store the target struct in a variable. To prevent initializing the same `nodeWithDist` twice.

routing/pathfind.go
```diff
@@ -656,33 +653,32 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// of the private channel, we'll assume it was selected as a
 		// routing hint due to having enough capacity for the payment
 		// and use the payment amount as its capacity.
-		bandWidth := partialPath.amountToReceive
+		bandWidth := distance[pivot].amountToReceive
```
joostjager: Hmmm... is this line changed because we now have pointers and need an up to date version of `nodeWithDist`? In that case, are we still putting the right data structure on the heap? Or should it be more minimal with just the pubkey on the heap?

joostjager: I am wondering whether there is a pre-existing bug here?

routing/pathfind.go
joostjager: I mean change to `currentNodeDist` or something

routing/pathfind.go
```diff
@@ -588,6 +587,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// from the heap.
 		partialPath := heap.Pop(&nodeHeap).(*nodeWithDist)
 		pivot := partialPath.node
+		pivotWithDist := distance[pivot]
```
joostjager: This is again pointing to something not right with what we store in the heap.

routing/pathfind.go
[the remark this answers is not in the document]
halseth: I would also prefer using a `logClosure`, if that works

routing/pathfind.go
```diff
@@ -337,56 +338,35 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	// traversal.
 	nodeHeap := newDistanceHeap()
 
-	// For each node in the graph, we create an entry in the distance map
-	// for the node set with a distance of "infinity". graph.ForEachNode
-	// also returns the source node, so there is no need to add the source
-	// node explicitly.
+	// Holds the current best distance for a given node
```
halseth: nit: punctation

channeldb/graph.go
```diff
@@ -190,6 +190,36 @@ func (c *ChannelGraph) Database() *DB {
 	return c.db
 }
 
+// NumNodes returns the number of nodes currently in the graph
+func (c *ChannelGraph) NumNodes(tx *bbolt.Tx) (int, error) {
```
halseth: I would add to each commit message the resulting performance impact of the change.

routing/pathfind.go
```diff
@@ -365,6 +357,24 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
 
+	numNodes, err := g.graph.NumNodes(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	// First we'll initialize an empty heap which'll help us to quickly
+	// locate the next edge we should visit next during our graph
+	// traversal.
+	nodeHeap := newDistanceHeap(numNodes)
```
halseth: instead of fetching the exact number of nodes in the db, could we make an educated constant guess on this number? Avoids the extra db code, and I don't think it matters since we don't know if it will be filled anyway.

routing/pathfind.go
```diff
@@ -666,7 +671,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 
 	// Use the nextHop map to unravel the forward path from source to
 	// target.
-	pathEdges := make([]*channeldb.ChannelEdgePolicy, 0, len(next))
+	pathEdges := make([]*channeldb.ChannelEdgePolicy, 0)
```
halseth: shy this change?

channeldb/graph.go
halseth: Or at least why the change has a positive impact.

routing/pathfind.go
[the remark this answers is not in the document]
champo: Just tested log closure and it adds about 1MB of extra allocations and a couple ms of extra runtime. If it is ok with you, I'd keep the if with an explicative comment.

routing/pathfind.go
joostjager: what way of making that guess do you have in mind?

routing/pathfind.go
joostjager: I took another look and indeed, `distance[pivot]` isn't changing as you say.

routing/pathfind.go
champo: For what it's worth, counting from the DB caused little to no impact in the benchmarks. Since it's only using the index it's most likely already cached in memory. I tried to use some boltd guesstimates but they were really far from the real number (like number of leafs).

channeldb/graph.go
joostjager: Strange. Is this an old db where there are left-overs from something?

champo: I think there shouldn't be? The DB was generated using the path finding test harness code a few months ago.

joostjager: Can you find out what causes those non-33 byte keys to be written? I'd be hesitant to add the extra condition without fully understanding why.

routing/pathfind.go
```diff
@@ -563,7 +543,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// route, return. It is important to also return if the distance
 		// is equal, because otherwise the algorithm could run into an
 		// endless loop.
-		if tempDist >= distance[fromVertex].dist {
+		if current, ok := distance[fromVertex]; ok && tempDist >= current.dist {
```
halseth: nit: split over two lines

channeldb/graph.go
```diff
@@ -190,6 +190,39 @@ func (c *ChannelGraph) Database() *DB {
...
+			}
+
+			return nil
+		})
+	}
+
+	if tx != nil {
+		err = count(tx)
+	} else {
+		err = c.db.View(count)
```
halseth: is this codepath ever taken?

routing/pathfind.go
halseth: The previous commit removed the graph traversal, and this reeintroduces it. This is probably less expensive, since we only count, but I wonder if we would get the same effect from just setting `numNodes = 1000` for instance.

routing/pathfind.go
halseth: *why

I was wondering what removing the cap size achieves.

routing/heap.go
```diff
@@ -36,12 +36,15 @@ type nodeWithDist struct {
...
+
+	// nextHop is the edge this route comes from.
+	nextHop *channeldb.ChannelEdgePolicy
 }
 
 // distanceHeap is a min-distance heap that's used within our path finding
 // algorithm to keep track of the "closest" node to our source node.
 type distanceHeap struct {
-	nodes []nodeWithDist
+	nodes []*nodeWithDist
```
halseth: Now that this is holding a pointer, I think we should remember to nil the popped element.

routing/pathfind.go
```diff
@@ -399,7 +394,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	// charges no fee. Distance is set to 0, because this is the starting
 	// point of the graph traversal. We are searching backwards to get the
 	// fees first time right and correctly match channel bandwidth.
-	distance[target] = nodeWithDist{
+	distance[target] = &nodeWithDist{
```
halseth: comment above needs to be slightly updated in the commit removing `ForEachNode` :)

routing/pathfind.go
```diff
@@ -665,26 +659,25 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+	// Use the distance map to unravel the forward path from source to
 	// target.
 	pathEdges := make([]*channeldb.ChannelEdgePolicy, 0)
 	currentNode := source
 	for currentNode != target { // TODO(roasbeef): assumes no cycles
 		// Determine the next hop forward using the next map.
-		nextNode := next[currentNode]
+		currentNodeWithDist, ok := distance[currentNode]
+
+		if !ok || currentNodeWithDist.nextHop == nil {
```
halseth: is the `currentNodeWithDist.nextHop == nil` case expected to ever happen? Might be worth returning a different error in that case.

routing/pathfind.go
```diff
@@ -587,6 +586,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		// from the heap.
 		partialPath := heap.Pop(&nodeHeap).(*nodeWithDist)
 		pivot := partialPath.node
+		pivotWithDist := distance[pivot]
```
halseth: how is this different from `partialPath`?

routing/pathfind.go
halseth: Can just do `var pathedges []*channeldb.ChannelEdgePolicy` perhaps?

halseth:
> Also, would it make sense to add the performance benchmark to the codebase (as a new PR that we could merge before this change) so that we could make sure we don't have regressions in the future?

I wouldn't consider this blocking.

routing/pathfind.go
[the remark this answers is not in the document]
joostjager: Discussed with @halseth offline. For consistency, we prefer the log closure. We can re-evaluate the hot path once the cache changes are in, see how much more significant this becomes.

channeldb/graph.go
```diff
@@ -1475,44 +1500,16 @@ func (c *ChannelGraph) ChanUpdatesInHorizon(startTime, endTime time.Time) ([]Cha
...
-			if channel, ok := c.chanCache.get(chanIDInt); ok {
-				hits++
-				edgesSeen[chanIDInt] = struct{}{}
-				edgesInHorizon = append(edgesInHorizon, channel)
-				continue
-			}
-
-			// First, we'll fetch the static edge information.
-			edgeInfo, err := fetchChanEdgeInfo(edgeIndex, chanID)
+			channel, err := fetchChannel(c.db, nodes, edges, edgeIndex, chanID)
```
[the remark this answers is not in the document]
joostjager: Continue discussion in follow up pr.

channeldb/graph.go
```diff
@@ -1949,6 +1922,25 @@ func (c *ChannelGraph) UpdateEdgePolicy(edge *ChannelEdgePolicy) error {
 		} else {
 			channel.Policy2 = edge
 		}
+
+		// Updates arrive without a node set and make the cache incomplete
+		if edge.Node == nil {
```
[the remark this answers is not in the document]
joostjager: Continues in follow up pr #3435

routing/pathfind.go
joostjager: Discussed with @halseth offline. Doubts remain whether the perf. improvement is worth the extra db scan through all nodes. Maybe a good middle ground is to add a reasonable constant now. For example 10000 nodes. This is little extra memory, because it is only a list of pointers, and it should do the job too. Then, when the cache is being used, we can possibly take a count from the cache. That way, we'll have a better estimate after a few runs.

channeldb/graph.go
champo: no, I follow the pattern in other methods in this file. In any case, when I update to use a hardcoded node count this code will be removed.

routing/pathfind.go
champo: Oh, the thing here is that `next` doesnt exist anymore, and it's not easy to find out the length of the path with the current structures. Considering a path will probably be a between 1-10 hops, pre-setting the capacity didn't make much sense. I'll update changing the style.

routing/pathfind.go
champo: Well actually, no! It was expected before removing the `distance` prepopulation. What are you thinking in terms of different error? It's probably a programming error.

routing/pathfind.go
champo: it's not, removed!

routing/pathfind.go
champo: Sure, I'll add a constant for this. The only real gain with having this call is a bit of "future proofing".

channeldb/graph.go
champo: The only point I found is `SetSourceNode` uses a hardcoded `"source"` as key in this bucket. I can't tell if there's any other place.

routing/heap.go
champo: `Pop` is calling `delete` on this map already, if that's what you mean

halseth: Should nil it to avoid gc leak, as here: https://golang.org/pkg/container/heap/#example__priorityQueue

routing/pathfind.go
halseth: If it is always sent when adding nodes to the distance map, we can just assume it is there I think

halseth:
@champo This is close! Do you want to make the last fixups? :)

champo:
@halseth I'll work on it this weekend :D

champo:
@halseth done!

routing/pathfind.go
```diff
@@ -36,6 +36,11 @@ const (
 	// some effect with smaller time lock values. The value may need
 	// tweaking and/or be made configurable in the future.
 	RiskFactorBillionths = 15
+
+	// estimatedNodeCount is used to preallocate the path finding structures
+	// to avoid resizing and copies. It should be number on the same order as
+	// the number of active nodes in the network.
+	estimatedNodeCount = 0
```
halseth: Set this to an actual number. Worst case it will be the number of reachable nodes on the network, so maybe `10k` or something should be good?

halseth: Also  name it `heapSize`?

halseth:
Also, linter is failing (run `make lint`).

routing/pathfind.go
joostjager: I did a quick experiment with this again on testnet. Set the source node in the queryroutes call to something that doesn't exist, so that it will explore a relatively large portion of the graph. Result:
```
2019-10-23 12:03:04.616 [DBG] CRTR: Pathfinding perf metrics debug: distance_size=690, heap_key_map_size=372, heap_slice_size=372
2019-10-23 12:03:04.616 [DBG] CRTR: Pathfinding perf metrics: nodes=690, edges=4841, time=86.268991ms
```
It seems that the two slices and map don't really grow that large in comparison to the number of edges that are explored. Also, I didn't see much difference in execution time.

We've been discussing setting this upper bound for a while now. There isn't a natural default for it and querying from the db is slow. We could make it adaptive based on the previous run, but in the end I doubt whether it all matters much. There is also still that other slice in the heap that isn't pre-sized in the current version of this PR.

I wouldn't mind dropping this commit really, considering what we know now.

halseth: Given that mainnet is much larger, do you get the same results there?

champo: FWIW, I tested all changes using a (old) snapshot of mainnet with ~6000 nodes and these changes had measurable impact. Though I tested it with the channel cache changeset applied so smaller changes were more noticeable.

joostjager: Ok, then also preallocated the slice in the heap and set the constant to 10k?

champo:
@halseth @joostjager updated with the estimated node count changes and linter fixes :)

routing/pathfind.go
```diff
@@ -664,26 +658,25 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
-	// Use the nextHop map to unravel the forward path from source to
+	// Use the distance map to unravel the forward path from source to
 	// target.
 	var pathEdges []*channeldb.ChannelEdgePolicy
 	currentNode := source
 	for currentNode != target { // TODO(roasbeef): assumes no cycles
 		// Determine the next hop forward using the next map.
-		nextNode := next[currentNode]
+		currentNodeWithDist, ok := distance[currentNode]
+
```
halseth: nit: remove newline

routing/pathfind.go
```diff
@@ -664,26 +658,25 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+		if !ok {
+			// If the node doesnt have a next hop it means we didn't find a path
```
halseth: nit: missing period.

routing/pathfind.go
```diff
@@ -664,26 +658,25 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+			return nil, newErrf(ErrNoPathFound, "unable to find a path to "+
```
halseth: line over 80 chars.

champo:
@halseth nits taken care of!

@joostjager I've rebased against master and fixed that error. It seems the commit that removed the old heap struct was already in master too and that brought in a silent conflict

champo:
🚀🚀 thanks!

PR #3419, opened 2019-08-21
lnwallet: when signing create account if not found

Roasbeef:
In this commit, we address an edge case that can happen a user rescans
w/ their seed, while retaining their existing `channel.db`. Once they
rescan, if they go to sign for a channel sweep for example, the
commitment key family (actually an account) may not yet have been
created, causing the signing attempt to fail.

We remedy this always creating the account if we go to sign, and the
account isn't found. The change has been structured to make this the
exception, so we'll avoid always needing to do 2 DB hits (check if
account exists, sign), each time we sign.

A new test has been added to exercise this behavior. If the diff from
the `signer.go` file is removed, then the test will fail.

### DIFF
--- a/lnwallet/btcwallet/signer.go
+++ b/lnwallet/btcwallet/signer.go
@@ -83,6 +83,25 @@ func (b *BtcWallet) fetchOutputAddr(script []byte) (waddrmgr.ManagedAddress, err
 	return nil, lnwallet.ErrNotMine
 }
 
+// deriveFromKeyLoc attempts to derive a private key using a fully specified
+// KeyLocator.
+func deriveFromKeyLoc(scopedMgr *waddrmgr.ScopedKeyManager,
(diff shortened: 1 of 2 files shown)


lnwallet/btcwallet/signer.go
```diff
@@ -95,20 +114,31 @@ func (b *BtcWallet) deriveKeyByLocator(keyLoc keychain.KeyLocator) (*btcec.Priva
...
+			if acctErr != nil {
+				return acctErr
+			}
+
+			// Now that we know the account exists, we'll attempt
+			// to re-derive the private key.
+			key, err = deriveFromKeyLoc(
+				scopedMgr, addrmgrNs, keyLoc,
+			)
+			if err != nil {
```
cfromknecht: nit: this check isn't necessary, can fall through to `return` below

Issue #3290, opened 2019-07-10
watchtower: create a new `cmd` directory to enable a standalone watch tower


Roasbeef:
As implemented today, a watch tower can only run in a "monolith" mode beside `lnd`. It should be possible to run a binary that _only_ runs a watch tower, to allow for more flexible deployment configurations. Most of the scaffolding is already in place for towers, since they're able to run as a part of `lnd`. The remaining work is to "daemonize" the existing scaffolding such that the tower can run on its own. This stand-lone tower binary (`towerd`?) should come equipped with the current [watch tower sub-server](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/watchtowerrpc/watchtower.proto), and should also be able to interact with using the `lncli tower` command.

orbitalturtle:
Going to take a stab at this.

seth586:
Instead of a separate binary, removing the requirement for an unlocked wallet would be sufficient to deploy `lnd` as a watchtower-only daemon.

Currently, with `lnd` running configured for a watchtower, the command `lncli tower info` results in:

```
[lncli] Wallet is encrypted. Please unlock using 'lncli unlock', or set password using 'lncli create' if this is the first time starting lnd.
```

Wallet unlock prior to watchtower service running shouldn't be necessary, especially if the watchtower loses power and reboots. Perhaps a `wallet.enable=0` option for `lnd.conf`?

example watchtower daemon configuration for `lnd.conf`:
```
[Application Options]
nolisten=1
nobootstrap=1

[Bitcoin]
bitcoin.active=1
bitcoin.mainnet=1
bitcoin.node=neutrino

[neutrino]
neutrino.connect=btcd-mainnet.lightning.computer

[tor]
tor.active=1
tor.socks=localhost:9050

[watchtower]
watchtower.active=1
watchtower.listen=localhost:9911
watchtower.externalip=qwertyabc123onionaddress.onion
```

saubyk:
Converting this to a discussion. Can be reverted to the issue or a new one opened up if any development is planned for this.

PR #3425, opened 2019-08-22
watchtower: create cmd for standalone tower [WIP!]

orbitalturtle:
Takes a stab at implementing the towerd command described in issue #3290 .

Still a WIP! Looking for feedback to see if the code so far makes sense or not before ironing out some kinks.

### DIFF
--- a/chainregistry.go
+++ b/chainregistry.go
@@ -12,6 +12,9 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc"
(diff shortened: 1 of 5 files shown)


Crypt-iQ:
Needs rebase since the introduction of a new chainreg package that changes call patterns with activeNetParams for example.

Crypt-iQ:
Going to take this over for 0.12 and give credit for commits where possible - thanks for starting this!

PR #3427, opened 2019-08-23
rpcserver: order updates in getnodeinfo by increasing pubkey

### LINKED ISSUE #3426: GetNodeInfo returns inaccurate policies for channels
### Background

When using getNodeInfo to look up node channels, the set of channels with policies is returned, however some of the policies are reversed.

This error seems to be due to a confusion between returned values, `FetchChannelEdgesByID` returns edges as in their public key lexicographical order whereas `ForEachChannel` returns edges as in order of their relationship between outgoing and incoming vs the node queried.

A workaround is to use getchaninfo on the channel ids returned by getNodeInfo, getchaninfo will use `FetchChannelEdgesByID` which returns the lexicographical order 

### Your environment

* lnd 0.7.1-beta on Ubuntu  - through LND 0.11.1-beta
* Bitcoin Core 0.18.0

### Steps to reproduce

Call getNodeInfo on a large set of channels, where there will be a conflict between the outgoing vs ingoing and node1 vs node2.

Some of the policies will be reversed

### Expected behaviour

Policies should reflect their author's set policies

### Actual behaviour

Semi-random policies are returned due to potential conflicts

cfromknecht:
fixes #3426

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3770,6 +3770,13 @@ func (r *rpcServer) DescribeGraph(ctx context.Context,
 func marshalDbEdge(edgeInfo *channeldb.ChannelEdgeInfo,
 	c1, c2 *channeldb.ChannelEdgePolicy) *lnrpc.ChannelEdge {
 
+	// Order the edges by increasing pubkey.
+	if bytes.Compare(edgeInfo.NodeKey2Bytes[:],
+		edgeInfo.NodeKey1Bytes[:]) < 0 {
+
+		c2, c1 = c1, c2
+	}
+
 	var (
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -3890,6 +3890,13 @@ func (r *rpcServer) GetNodeInfo(ctx context.Context,
 				return nil
 			}
 
+			// Order the edges by increasing pubkey.
+			if bytes.Compare(c2.Node.PubKeyBytes[:],
```
halseth: shouldn't this more generally be done within `marshalDbEdge`?

PR #3434, opened 2019-08-27
lntest/itest: resolve mempool flake in multi-hop htlc local timeout test

wpaulino:
The test assumed that transactions would be broadcast and confirmed at incorrect heights. Due to timing issues, it was possible for the test to still succeed, resulting in a flake.

The test assumes that Bob will sweep a pending outgoing HTLC and commit output back to their wallet. This commit ensures that these operations are done when expected, i.e.:

1. Bob force closes the channel due to the HTLC timing out.
2. Once the channel is confirmed, Bob broadcasts their HTLC timeout transaction.
3. Bob broadcasts their commit output sweep transaction once its CSV expires.
4. Bob broadcasts their second layer sweep transaction for the timed out HTLC once its CSV expires.

### DIFF
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -10153,6 +10153,10 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
 	if err != nil {
 		t.Fatalf("unable to get txid: %v", err)
 	}
+	closeTxid, err := waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
+	if err != nil {
+		t.Fatalf("unable to find closing txid: %v", err)
+	}
 	assertSpendingTxInMempool(
 		t, net.Miner.Node, minerMempoolTimeout, wire.OutPoint{
 			Hash:  *bobFundingTxid,
@@ -10179,24 +10183,22 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("htlc mismatch: %v", predErr)
 	}
 
-	// We'll mine defaultCSV blocks in order to generate the sweep
-	// transaction of Bob's funding output. This will also bring us to the
-	// maturity height of the htlc tx output.
-	if _, err := net.Miner.Node.Generate(defaultCSV); err != nil {
-		t.Fatalf("unable to generate blocks: %v", err)
+	// With the closing transaction confirmed, we should expect Bob's HTLC
+	// timeout transaction to be broadcast due to the expiry being reached.
+	htlcTimeout, err := waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
+	if err != nil {
+		t.Fatalf("unable to find bob's htlc timeout tx: %v", err)
 	}
 
-	_, err = waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
-	if err != nil {
-		t.Fatalf("unable to find bob's funding output sweep tx: %v", err)
-	}
-
-	// The second layer HTLC timeout transaction should now have been
-	// broadcast on-chain.
-	secondLayerHash, err := waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
-	if err != nil {
-		t.Fatalf("unable to find bob's second layer transaction")
-	}
+	// We'll mine the remaining blocks in order to generate the sweep
+	// transaction of Bob's commitment output.
+	mineBlocks(t, net, defaultCSV, 1)
+	assertSpendingTxInMempool(
+		t, net.Miner.Node, minerMempoolTimeout, wire.OutPoint{
+			Hash:  *closeTxid,
+			Index: 1,
+		},
+	)
 
 	// Bob's pending channel report should show that he has a commitment
 	// output awaiting sweeping, and also that there's an outgoing HTLC
@@ -10220,14 +10222,20 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("bob should have pending htlc but doesn't")
 	}
 
-	// Now we'll mine an additional block, which should include the second
-	// layer sweep tx.
-	block := mineBlocks(t, net, 1, 1)[0]
+	// Now we'll mine an additional block, which should confirm Bob's commit
+	// sweep. This block should also prompt Bob to broadcast their second
+	// layer sweep due to the CSV on the HTLC timeout output.
+	mineBlocks(t, net, 1, 1)
+	assertSpendingTxInMempool(
+		t, net.Miner.Node, minerMempoolTimeout, wire.OutPoint{
+			Hash:  *htlcTimeout,
+			Index: 0,
+		},
+	)
 
-	// The block should have confirmed Bob's second layer sweeping
-	// transaction. Therefore, at this point, there should be no active
-	// HTLC's on the commitment transaction from Alice -> Bob.
-	assertTxInBlock(t, block, secondLayerHash)
+	// The block should have confirmed Bob's HTLC timeout transaction.
+	// Therefore, at this point, there should be no active HTLC's on the
+	// commitment transaction from Alice -> Bob.
 	nodes = []*lntest.HarnessNode{net.Alice}
 	err = lntest.WaitPredicate(func() bool {
 		predErr = assertNumActiveHtlcs(nodes, 0)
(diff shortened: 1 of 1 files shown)


lntest/itest/lnd_test.go
```diff
@@ -10172,23 +10172,19 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
...
 
-	// We'll mine defaultCSV blocks in order to generate the sweep
-	// transaction of Bob's funding output. This will also bring us to the
-	// maturity height of the htlc tx output.
-	if _, err := net.Miner.Node.Generate(defaultCSV); err != nil {
-		t.Fatalf("unable to generate blocks: %v", err)
-	}
-
+	// With the closing transaction confirmed, we should expect our HTLC
+	// timeout transaction to be broadcast.
```
joostjager: our -> Bob's. Add here that because the outgoing broadcast delta is zero, there is no cltv wait required?

lntest/itest/lnd_test.go
```diff
@@ -10172,23 +10172,19 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
...
 	_, err = waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
```
joostjager: Restore matching on `secondLayerHash`?

lntest/itest/lnd_test.go
```diff
@@ -10213,14 +10209,18 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("bob should have pending htlc but doesn't")
 	}
 
-	// Now we'll mine an additional block, which should include the second
-	// layer sweep tx.
-	block := mineBlocks(t, net, 1, 1)[0]
+	// Now we'll mine an additional block, which should confirm Bob's commit
+	// sweep. This block should also prompt Bob to broadcast their second
+	// layer sweep.
```
joostjager: I feel that the test would have been slightly better if we didn't use a zero outgoing broadcast delta. But I think out of scope for this pr.

lntest/itest/lnd_test.go
halseth: +1

lntest/itest/lnd_test.go
```diff
@@ -10213,14 +10209,18 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("bob should have pending htlc but doesn't")
 	}
 
-	// Now we'll mine an additional block, which should include the second
-	// layer sweep tx.
-	block := mineBlocks(t, net, 1, 1)[0]
+	// Now we'll mine an additional block, which should confirm Bob's commit
+	// sweep. This block should also prompt Bob to broadcast their second
```
halseth: Why is Bob broadcasting the second layer sweep here? Is the second-layer sweep also waiting for exactly `defaultCSV`? Add a comment.

lntest/itest/lnd_test.go
```diff
@@ -10172,23 +10172,19 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
...
 	if err != nil {
-		t.Fatalf("unable to find bob's funding output sweep tx: %v", err)
+		t.Fatalf("unable to find bob's htlc timeout tx: %v", err)
 	}
 
-	// The second layer HTLC timeout transaction should now have been
-	// broadcast on-chain.
-	secondLayerHash, err := waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
+	// We'll mine the remaining blocks in order to generate the sweep
+	// transaction of Bob's funding output.
```
halseth: commitment output?

lntest/itest/lnd_test.go
```diff
@@ -10213,14 +10209,18 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
...
 	}
 
-	// Now we'll mine an additional block, which should include the second
-	// layer sweep tx.
-	block := mineBlocks(t, net, 1, 1)[0]
+	// Now we'll mine an additional block, which should confirm Bob's commit
+	// sweep. This block should also prompt Bob to broadcast their second
+	// layer sweep.
+	mineBlocks(t, net, 1, 1)
+	_, err = waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
```
halseth: check that it is spending second layer tx :)

lntest/itest/lnd_test.go
joostjager: I mean that it could be that after the closing tx confirms, the resolver still needs to wait for the cltv lock. But not in this case, because the delta is zero. So channel arbitrator waits with publishing the commit tx until the exact block the htlc expires?

lntest/itest/lnd_test.go
joostjager: Actually, `DefaultOutgoingBroadcastDelta` is also zero, so we are testing what happens in production, which is good.

lntest/itest/lnd_test.go
```diff
@@ -10172,23 +10172,19 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
...
+		t.Fatalf("unable to find bob's htlc timeout tx: %v", err)
 	}
 
-	// The second layer HTLC timeout transaction should now have been
-	// broadcast on-chain.
-	secondLayerHash, err := waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
+	// We'll mine the remaining blocks in order to generate the sweep
+	// transaction of Bob's commitment output.
+	mineBlocks(t, net, defaultCSV, 1)
+	_, err = waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
```
halseth: nit: assert that this spends the commitment output?

lntest/itest/lnd_test.go
halseth: 👍

joostjager: @halseth I think this has to do with the outgoingbroadcastdelta being 0.

lntest/itest/lnd_test.go
```diff
@@ -10245,16 +10247,6 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("bob's htlc should have advanced to the second stage: %v", err)
```
halseth: This should not be true anymore? If the htlc timeout tx is not confirmed yet, the HTLC shouldn't be in stage 2.

lntest/itest/lnd_test.go
```diff
@@ -10245,16 +10247,6 @@ func testMultiHopHtlcLocalTimeout(net *lntest.NetworkHarness, t *harnessTest) {
...
 	}
 
-	// We'll now mine four more blocks. After the 4th block, a transaction
-	// sweeping the HTLC output should be broadcast.
-	if _, err := net.Miner.Node.Generate(4); err != nil {
```
halseth: I wonder where this 4 block constant came from 🤔

lntest/itest/lnd_test.go
wpaulino: The HTLC timeout transaction has already confirmed by the time we're here.

lntest/itest/lnd_test.go
wpaulino: 4 is the default CSV used throughout our tests.

lntest/itest/lnd_test.go
wpaulino: Done.

lntest/itest/lnd_test.go
wpaulino: Correct. What I meant is that the expiry is at height 100, so we decide to close then due to a 0 outgoing broadcast delta, but the timeout is actually being broadcast at height 101 because we wait until the confirmation of the closing transaction.

joostjager: Ok

lntest/itest/lnd_test.go
halseth: You're right (I think). I guess the comments were just wrong before, saying that the htlc sweep was already confirmed at this point

wpaulino:
We can't do exact height values because all the tests are being run on the same chain.

joostjager:
> We can't do exact height values because all the tests are being run on the same chain.

Ok. Or write the comments so they include heights relatively to the start height of the test. "We are now at height +50 and the commentment tx should be in the mempool. The htlc expires at +50 also, but we only publish the 2nd level tx after the commit tx confirms at +51. There is a 40 block csv wait, so this 2nd level tx can be swept at +91."

But as said before, it isn't blocking. Just brainstorming about how to clarify the comments, so that not as many questions need to be asked.

PR #3436, opened 2019-08-28
[wip] lnrpc/routerrpc/router.proto: mpp rpc changes

cfromknecht:
cc @joostjager @halseth


lnrpc/routerrpc/router_server.go
```diff
@@ -280,7 +280,18 @@ func (s *Server) SendToRoute(ctx context.Context,
 		return nil, err
 	}
 
-	preimage, err := s.cfg.Router.SendToRoute(hash, route)
+	total := lnwire.MilliSatoshi(req.TotalAmtMsat)
+
+	var addr lntypes.Hash
+	if len(req.PaymentAddr) > 0 {
```
joostjager: `len` check is also performed inside `lntypes.MakeHash`

lnrpc/routerrpc/router.proto
```diff
@@ -125,8 +124,27 @@ message PaymentStatus {
...
+
+    /**
+    List of all shard statuses to this payment_hash [EXPERIMENTAL].
+    */
+    repeated ShardStatus shards = 4 [json_name = "shards"];
 }
 
+message ShardStatus {
+    /// Current state the shard is in.
+    PaymentState state = 1;
```
joostjager: I don't think this `PaymentState` can be the same as the payment-level state. For example `FAILED_NO_ROUTE` and `FAILED_TIMEOUT` do not apply to a shard.

lnrpc/routerrpc/router.proto
```diff
@@ -125,8 +124,27 @@ message PaymentStatus {
...
+
+    /**
+    The preimage of the shard when state is SUCCEEDED.
+    */
+    bytes preimage = 2;
```
joostjager: For mpp, this is a payment level property. Do we report it here as a prep for amp? Maybe it is better to create a payment-level property 'payment proof' and only populate it for mpp payments.

lnrpc/routerrpc/router.proto
```diff
@@ -125,8 +124,27 @@ message PaymentStatus {
...
+
+    /**
+    The taken route when state is SUCCEEDED.
```
joostjager: This is actually no longer correct. Also for failed payments, we report the route. This is useful for probe payments to find out the final route taken.

routing/payment_session.go
```diff
@@ -60,6 +60,7 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
 	// If we have a pre-built route, use that directly.
 	case p.preBuiltRoute != nil && !p.preBuiltRouteTried:
 		p.preBuiltRouteTried = true
+		p.preBuiltRoute.AddFinalHopRecords(payment.FinalDestRecords)
```
joostjager: This brings up the discussion about how to represent the mpp fields internally. In my understanding, `FinalDestRecords` are untyped and only meant for application-specific fields from the custom range.

I can understand that this may be easy for now, but don't we eventually want to add proper fields for mpp and amp to the hop structure, just as `amt` is a proper field? And then convert that into the tlv stream only in PackHopPayload. This is then also where the tlv id is known for those fields.

routing/router.go
```diff
@@ -1728,12 +1729,47 @@ func (r *ChannelRouter) preparePayment(payment *LightningPayment) (
...
+			"for mpp")
+	}
+
+	// If we're using mpp, set the amount in the payment creation info to
+	// the total amount. The route will contain the amount paid in the
+	// shard.
+	if useMPP {
+		amt = total
+	}
+
```
joostjager: All of these changes check `useMPP` first. It could all happen in one `if` clause.

routing/router.go
```diff
@@ -1748,6 +1784,13 @@ func (r *ChannelRouter) SendToRoute(hash lntypes.Hash, route *route.Route) (
 		return [32]byte{}, err
 	}
 
+	// If the sender requested mpp, create a record for the final hop that
+	// will encode the total_msat and payment_addr fields.
+	var recs []tlv.Record
+	if useMPP {
+		recs = append(recs, record.NewMPP(uint64(total), addr).TLV())
```
joostjager: Same comment about layering violation here.

lnrpc/routerrpc/router.proto
```diff
@@ -125,8 +124,27 @@ message PaymentStatus {
     The taken route when state is SUCCEEDED.
     */
     lnrpc.Route route = 3;
+
+    /**
+    List of all shard statuses to this payment_hash [EXPERIMENTAL].
+    */
+    repeated ShardStatus shards = 4 [json_name = "shards"];
```
joostjager: I wouldn't mind leaving reporting of shards out of scope initially and just make sure that the payment-level reporting is mpp aware.

lnrpc/routerrpc/router.proto
```diff
@@ -125,8 +124,27 @@ message PaymentStatus {
     The taken route when state is SUCCEEDED.
     */
     lnrpc.Route route = 3;
```
joostjager: `route` is probably no longer applicable on the payment level.

lnrpc/routerrpc/router_server.go
joostjager: ah, ok

lnrpc/routerrpc/router.proto
```diff
@@ -111,23 +111,23 @@ enum PaymentState {
...
 message PaymentStatus {
-    /// Current state the payment is in.
+    /**
+    Current state the payment is in.
+    */
     PaymentState state = 1;
 
     /**
-    The pre-image of the payment when state is SUCCEEDED.
+    The pre-image of the payment when state is not IN_FLIGHT.
```
joostjager: not IN_FLIGHT and not FAILED -> SUCCEEDED

lnrpc/routerrpc/router.proto
```diff
@@ -111,23 +111,23 @@ enum PaymentState {
...
     */
     bytes preimage = 2;
 
     /**
-    The taken route when state is SUCCEEDED.
+    The taken route when state is not IN_FLIGHT.
     */
     lnrpc.Route route = 3;
```
joostjager: `route` is not possible to populate on the payment level with multiple shards.

routing/route/route.go
```diff
@@ -76,6 +77,10 @@ type Hop struct {
 	// carries as a fee will be subtracted by the hop.
 	AmtToForward lnwire.MilliSatoshi
 
+	// MPP encapsulates the data required for option_mpp. This field should
+	// only be set for the final hop.
+	MPP *record.MPP
```
joostjager: Does it need to be nested, or can those fields be top level fields? Probably dependent on what will happen when the set is extended to amp.

record/mpp.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"io"
+
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+type MPP struct {
+	TotalMsat   uint64
+	PaymentAddr [32]byte
```
joostjager: Yes, definitely better with concrete types.

routing/payment_session.go
Roasbeef: `FinalDestRecords` is indeed untyped at this stage, but the intention was that a higher layer would have already serialized the complete tlv payload at that point, using richer types at an earlier stage of the process. 

As far as adding additional fields to the `Hop` payload, if they're all in-line, rather than an optional pointer to a richer struct, we'll end up with fields that are mostly not used (depending on which one becomes the dominant payment type) sitting there in memory FWIW.

lnrpc/routerrpc/router_server.go
```diff
@@ -280,7 +280,18 @@ func (s *Server) SendToRoute(ctx context.Context,
...
+		addr, err = lntypes.MakeHash(req.PaymentAddr)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse "+
+				"payment_addr: %v", err)
+		}
+	}
+
+	preimage, err := s.cfg.Router.SendToRouteMPP(hash, route, total, addr)
```
Roasbeef: So then this assumes the legacy (current) payment type won't be used via this RPC method?

routing/route/route.go
Roasbeef: IMO it's better nested (with a pointer) as otherwise it isn't clear what the required vs optional fields are in the `Hop` struct itself.

PR #3439, opened 2019-08-30
discovery+switch: apply zero forwarding policy updates in-memory as w…

valentinewallace:
…ell as on disk

In this commit, we fix a bug where if a user updates a forwarding policy to be
zero, the update will be applied to the policy correctly on-disk, but not
in-memory.

We solve this issue by having the gossiper return the list of on-disk updated
policies and passing these policies to the switch for updates, so the switch can assume
that zero-valued fields are intentional and not just uninitialized.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -93,7 +93,7 @@ type chanPolicyUpdateRequest struct {
 	targetChans []wire.OutPoint
 	newSchema   routing.ChannelPolicy
 
-	errResp chan error
+	chanPolicies chan updatedChanPolicies
 }
 
 // Config defines the configuration for the service. ALL elements within the
@@ -338,27 +338,40 @@ func New(cfg Config, selfKey *btcec.PublicKey) *AuthenticatedGossiper {
 	return gossiper
 }
 
+// updatedChanPolicies is a set of channel policies that have been successfully
+// updated and written to disk, or an error if the policy update failed. This
+// struct's map field is intended to be used for updating channel policies on
+// the link layer.
+type updatedChanPolicies struct {
+	chanPolicies map[wire.OutPoint]*channeldb.ChannelEdgePolicy
+	err          error
+}
+
 // PropagateChanPolicyUpdate signals the AuthenticatedGossiper to update the
 // channel forwarding policies for the specified channels. If no channels are
 // specified, then the update will be applied to all outgoing channels from the
 // source node. Policy updates are done in two stages: first, the
 // AuthenticatedGossiper ensures the update has been committed by dependent
-// sub-systems, then it signs and broadcasts new updates to the network.
+// sub-systems, then it signs and broadcasts new updates to the network. A
+// mapping between outpoints and updated channel policies is returned, which is
+// used to update the forwarding policies of the underlying links.
 func (d *AuthenticatedGossiper) PropagateChanPolicyUpdate(
-	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) error {
+	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) (
+	map[wire.OutPoint]*channeldb.ChannelEdgePolicy, error) {
 
-	errChan := make(chan error, 1)
+	chanPolicyChan := make(chan updatedChanPolicies, 1)
 	policyUpdate := &chanPolicyUpdateRequest{
-		targetChans: chanPoints,
-		newSchema:   newSchema,
-		errResp:     errChan,
+		targetChans:  chanPoints,
+		newSchema:    newSchema,
+		chanPolicies: chanPolicyChan,
 	}
 
 	select {
 	case d.chanPolicyUpdates <- policyUpdate:
-		return <-errChan
+		updatedPolicies := <-chanPolicyChan
+		return updatedPolicies.chanPolicies, updatedPolicies.err
 	case <-d.quit:
-		return fmt.Errorf("AuthenticatedGossiper shutting down")
+		return nil, fmt.Errorf("AuthenticatedGossiper shutting down")
 	}
 }
 
@@ -895,13 +908,17 @@ func (d *AuthenticatedGossiper) networkHandler() {
 			// First, we'll now create new fully signed updates for
 			// the affected channels and also update the underlying
 			// graph with the new state.
-			newChanUpdates, err := d.processChanPolicyUpdate(
+			chanPolicies, newChanUpdates, err := d.processChanPolicyUpdate(
 				policyUpdate,
 			)
+			update := updatedChanPolicies{
+				chanPolicies,
+				err,
+			}
+			policyUpdate.chanPolicies <- update
 			if err != nil {
 				log.Errorf("Unable to craft policy updates: %v",
 					err)
-				policyUpdate.errResp <- err
 				continue
 			}
 
@@ -909,8 +926,6 @@ func (d *AuthenticatedGossiper) networkHandler() {
 			// them to the announcement batch to be flushed at the
 			// start of the next epoch.
 			announcements.AddMsgs(newChanUpdates...)
-
-			policyUpdate.errResp <- nil
 
 		case announcement := <-d.networkMsgs:
 			// We should only broadcast this message forward if it
@@ -1244,13 +1259,19 @@ func (d *AuthenticatedGossiper) retransmitStaleChannels() error {
 //
 // TODO(roasbeef): generalize into generic for any channel update
 func (d *AuthenticatedGossiper) processChanPolicyUpdate(
-	policyUpdate *chanPolicyUpdateRequest) ([]networkMsg, error) {
+	policyUpdate *chanPolicyUpdateRequest) (
+	map[wire.OutPoint]*channeldb.ChannelEdgePolicy, []networkMsg, error) {
+
 	// First, we'll construct a set of all the channels that need to be
 	// updated.
 	chansToUpdate := make(map[wire.OutPoint]struct{})
 	for _, chanPoint := range policyUpdate.targetChans {
 		chansToUpdate[chanPoint] = struct{}{}
 	}
+
+	// Next, we'll create a mapping from outpoint to edge policy that will
+	// be used by each edge's underlying link to update its policy.
+	chanPolicies := make(map[wire.OutPoint]*channeldb.ChannelEdgePolicy)
 
 	haveChanFilter := len(chansToUpdate) != 0
 	if haveChanFilter {
@@ -1295,7 +1316,7 @@ func (d *AuthenticatedGossiper) processChanPolicyUpdate(
 		return nil
 	})
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	// With the set of edges we need to update retrieved, we'll now re-sign
@@ -1309,8 +1330,12 @@ func (d *AuthenticatedGossiper) processChanPolicyUpdate(
 			edgeInfo.info, edgeInfo.edge,
 		)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
+
+		// Since the update succeeded, add the edge to our policy
+		// mapping.
+		chanPolicies[edgeInfo.info.ChannelPoint] = edgeInfo.edge
 
 		// We'll avoid broadcasting any updates for private channels to
 		// avoid directly giving away their existence. Instead, we'll
@@ -1340,7 +1365,7 @@ func (d *AuthenticatedGossiper) processChanPolicyUpdate(
 		})
 	}
 
-	return chanUpdates, nil
+	return chanPolicies, chanUpdates, nil
 }
 
 // processRejectedEdge examines a rejected edge to see if we can extract any
--- a/discovery/gossiper_test.go
+++ b/discovery/gossiper_test.go
@@ -3391,8 +3391,14 @@ func TestPropagateChanPolicyUpdate(t *testing.T) {
 	// the channel ann proof from the first channel in order to have it be
 	// marked as private channel.
 	firstChanID := channelsToAnnounce[0].localChanAnn.ShortChannelID
-	for _, batch := range channelsToAnnounce {
-		sendLocalMsg(t, ctx, batch.localChanAnn, localKey)
+	for i, batch := range channelsToAnnounce {
+		// channelPoint ensures that each channel policy in the map
+		// returned by PropagateChanPolicyUpdate has a unique key. Since
+		// the map is keyed by wire.OutPoint, we want to ensure that
+		// each channel has a unique channel point.
+		channelPoint := ChannelPoint(wire.OutPoint{Index: uint32(i)})
+
+		sendLocalMsg(t, ctx, batch.localChanAnn, localKey, channelPoint)
 		sendLocalMsg(t, ctx, batch.chanUpdAnn1, localKey)
 		sendLocalMsg(t, ctx, batch.nodeAnn1, localKey)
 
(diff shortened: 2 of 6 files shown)


discovery/gossiper.go
```diff
@@ -345,20 +346,23 @@ func New(cfg Config, selfKey *btcec.PublicKey) *AuthenticatedGossiper {
 // AuthenticatedGossiper ensures the update has been committed by dependent
 // sub-systems, then it signs and broadcasts new updates to the network.
 func (d *AuthenticatedGossiper) PropagateChanPolicyUpdate(
-	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) error {
+	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) (
+	map[wire.OutPoint]*channeldb.ChannelEdgePolicy, error) {
```
wpaulino: Nit: should document return parameters.

discovery/gossiper.go
```diff
@@ -1292,10 +1303,13 @@ func (d *AuthenticatedGossiper) processChanPolicyUpdate(
 			edge: edge,
 		})
 
+		// Finally, add the edge to our policy mapping.
+		chanPolicies[info.ChannelPoint] = edge
```
wpaulino: We should only be setting these for channels that have actually been updated (further below).

discovery/gossiper_test.go
```diff
@@ -3353,13 +3353,23 @@ func TestPropagateChanPolicyUpdate(t *testing.T) {
 
 	const numChannels = 3
 	channelsToAnnounce := make([]*annBatch, 0, numChannels)
+	channelPoints := make([]OptionalMsgField, 0, numChannels)
```
wpaulino: Alternatively, you could make unique channel points within `createAnnouncements` the same way without having to use the functional options.

htlcswitch/link.go
```diff
@@ -2139,31 +2139,19 @@ func (l *channelLink) AttachMailBox(mailbox MailBox) {
...
-	if newPolicy.TimeLockDelta != 0 {
-		l.cfg.FwrdingPolicy.TimeLockDelta = newPolicy.TimeLockDelta
-	}
-	if newPolicy.MinHTLC != 0 {
-		l.cfg.FwrdingPolicy.MinHTLC = newPolicy.MinHTLC
-	}
+	l.cfg.FwrdingPolicy.BaseFee = newPolicy.BaseFee
+	l.cfg.FwrdingPolicy.FeeRate = newPolicy.FeeRate
+	l.cfg.FwrdingPolicy.TimeLockDelta = newPolicy.TimeLockDelta
+	l.cfg.FwrdingPolicy.MinHTLC = newPolicy.MinHTLC
```
wpaulino: Looks like we can just do `l.cfg.FwrdingPolicy = newPolicy`?

htlcswitch/link_test.go
```diff
@@ -5508,9 +5508,9 @@ func TestForwardingAsymmetricTimeLockPolicies(t *testing.T) {
 	// Now that each of the links are up, we'll modify the link from Alice
 	// -> Bob to have a greater time lock delta than that of the link of
 	// Bob -> Carol.
-	n.firstBobChannelLink.UpdateForwardingPolicy(ForwardingPolicy{
-		TimeLockDelta: 7,
-	})
+	newPolicy := n.firstBobChannelLink.cfg.FwrdingPolicy
+	newPolicy.TimeLockDelta = 7
+	n.firstBobChannelLink.UpdateForwardingPolicy(newPolicy)
```
wpaulino: Why is this change needed?

htlcswitch/switch.go
```diff
@@ -445,55 +445,63 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
 	}
 
 	return nil
 }
 
+// dbPolicyToFwdingPolicy is a helper function that converts a channeldb
+// ChannelEdgePolicy into a ForwardingPolicy struct for the purpose of updating
+// the forwarding policy of a link.
+func dbPolicyToFwdingPolicy(policy *channeldb.ChannelEdgePolicy) ForwardingPolicy {
+	return ForwardingPolicy{
```
wpaulino: Missing `MaxHTLC` field?

htlcswitch/switch.go
```diff
@@ -445,55 +445,63 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
 // forwarding policies for all links have been updated, or the switch shuts
 // down.
-func (s *Switch) UpdateForwardingPolicies(newPolicy ForwardingPolicy,
-	targetChans ...wire.OutPoint) error {
-
-	log.Debugf("Updating link policies: %v", newLogClosure(func() string {
-		return spew.Sdump(newPolicy)
-	}))
+func (s *Switch) UpdateForwardingPolicies(
+	chanPolicies map[wire.OutPoint]*channeldb.ChannelEdgePolicy) error {
```
wpaulino: This method no longer needs to return an error.

htlcswitch/switch.go
```diff
@@ -445,55 +445,63 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
-			}
+	for targetLink := range chanPolicies {
+		cid := lnwire.NewChanIDFromOutPoint(&targetLink)
 
-			linksToUpdate = append(linksToUpdate, link)
+		// If we can't locate a link by its converted channel
+		// ID, then we'll log an error.
+		link, ok := s.linkIndex[cid]
+		if !ok {
+			log.Errorf("unable to find ChannelPoint(%v) to update "+
```
wpaulino: Can just log as debug as it's not a fatal error since the link will have the right policy once it becomes active again. Log should also start capitalized.

htlcswitch/switch.go
```diff
@@ -445,55 +445,63 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
 	// With all the links we need to update collected, we can release the
 	// mutex then update each link directly.
 	for _, link := range linksToUpdate {
+		newPolicy := dbPolicyToFwdingPolicy(
+			chanPolicies[*link.ChannelPoint()],
+		)
+
+		log.Debugf("Updating link policies: %v",
+			newLogClosure(func() string {
+				return spew.Sdump(newPolicy)
```
wpaulino: Any logs containing `spew`s should be trace.

wpaulino: This is also going to be logged for every link in its current placement.

htlcswitch/link_test.go
valentinewallace: Otherwise, it'll set the max HTLC to zero and generating hops will fail.

htlcswitch/switch.go
valentinewallace: thanks, fixed

htlcswitch/link.go
valentinewallace: nice lol, fixed

rpcserver.go
```diff
@@ -4612,30 +4612,16 @@ func (r *rpcServer) UpdateChannelPolicy(ctx context.Context,
...
-	err = r.server.htlcSwitch.UpdateForwardingPolicies(p, targetChans...)
-	if err != nil {
-		// If we're unable update the fees due to the links not being
-		// online, then we don't need to fail the call. We'll simply
-		// log the failure.
-		rpcsLog.Warnf("Unable to update link fees: %v", err)
-	}
+	// Finally, we'll apply the set of channel policies to the target
+	// channels' links.
+	r.server.htlcSwitch.UpdateForwardingPolicies(chanPolicies)
```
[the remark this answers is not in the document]
wpaulino: I'm not opposed to it, but going this route seems like a much bigger change for a bug that's causing routing failures at the moment. This change is pretty small in that it's just exposing the state the gossiper is already acting on. I understand that doing so will just make the separation of concerns worse though.

discovery/gossiper.go
```diff
@@ -1309,9 +1322,13 @@ func (d *AuthenticatedGossiper) processChanPolicyUpdate(
 			edgeInfo.info, edgeInfo.edge,
 		)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
```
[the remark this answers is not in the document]
wpaulino: This seems unrelated to this PR? In any case, I believe the check is there to automatically upgrade channels to include the field. I don't think we can remove it as we can't ensure all older nodes have gone through this flow.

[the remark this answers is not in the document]
joostjager: Ok, yes I was thinking of the follow up of this pr.

rpcserver.go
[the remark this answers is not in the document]
joostjager: Ok. I also realize now that links are not always up, so there must be an independent data store somewhere that stores the local policy. 

The only thing I am still thinking about: if we'd just make all policy parameters mandatory for the rpc call (including minhtlc and maxhtlc), would this feedback path of final policies from gossiper still be necessary? If the policy is complete, the links (the ones that are up) and gossiper can be updated independently.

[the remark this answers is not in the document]
valentinewallace: That's true, but I think it would be best if none of them were mandatory, meaning a user could only specify the ones they want to be updated. That makes me want to make max htlc non-mandatory and lay groundwork for making the other fields non-mandatory as well.

htlcswitch/switch.go
```diff
@@ -445,53 +445,58 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
+		cid := lnwire.NewChanIDFromOutPoint(&targetLink)
 
-				return fmt.Errorf("unable to find "+
-					"ChannelPoint(%v) to update link "+
-					"policy", targetLink)
-			}
-
-			linksToUpdate = append(linksToUpdate, link)
+		// If we can't locate a link by its converted channel
+		// ID, then we'll log an error.
```
wpaulino: Nit: not an error 😛

htlcswitch/switch.go
```diff
@@ -445,53 +445,58 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
-		// policies for the set of targeted links.
-		for _, targetLink := range targetChans {
-			cid := lnwire.NewChanIDFromOutPoint(&targetLink)
-
-			// If we can't locate a link by its converted channel
-			// ID, then we'll return an error back to the caller.
-			link, ok := s.linkIndex[cid]
-			if !ok {
-				s.indexMtx.RUnlock()
+	for targetLink := range chanPolicies {
```
wpaulino: Nit: comment above this and the godoc for the method are now outdated.

discovery/gossiper.go
```diff
@@ -343,22 +344,27 @@ func New(cfg Config, selfKey *btcec.PublicKey) *AuthenticatedGossiper {
...
+// sub-systems, then it signs and broadcasts new updates to the network. A
+// mapping between outpoints and updated channel policies is returned, which is
+// used to update the forwarding policies of the underlying links.
 func (d *AuthenticatedGossiper) PropagateChanPolicyUpdate(
-	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) error {
+	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) (
+	map[wire.OutPoint]*channeldb.ChannelEdgePolicy, error) {
 
 	errChan := make(chan error, 1)
+	chanPolicyChan := make(chan map[wire.OutPoint]*channeldb.ChannelEdgePolicy, 1)
```
joostjager: I'd make a struct containing the map and the error and use only a single channel.

discovery/gossiper.go
```diff
@@ -343,22 +344,27 @@ func New(cfg Config, selfKey *btcec.PublicKey) *AuthenticatedGossiper {
...
+		targetChans:  chanPoints,
+		newSchema:    newSchema,
+		chanPolicies: chanPolicyChan,
+		errResp:      errChan,
 	}
 
 	select {
 	case d.chanPolicyUpdates <- policyUpdate:
-		return <-errChan
+		return <-chanPolicyChan, <-errChan
```
joostjager: Can these chan reads block forever because there is no select on `d.quit`? I think in the error case, nothing is sent on `chanPolicyChan`.

discovery/gossiper_test.go
```diff
@@ -3353,13 +3353,23 @@ func TestPropagateChanPolicyUpdate(t *testing.T) {
...
 		channelsToAnnounce = append(channelsToAnnounce, newChan)
+
+		// channelPoints ensures that each channel policy in the map
+		// returned by PropagateChanPolicyUpdate has a unique key. Since
+		// the map is keyed by wire.OutPoint, we want to ensure that
+		// each channel has a unique channel point.
+		channelPoints = append(
+			channelPoints,
+			ChannelPoint(wire.OutPoint{Index: uint32(i)}),
+		)
```
joostjager: You could also not build the slice and just create the outpoint inside the `channelsToAnnounce` loop below.

htlcswitch/switch.go
```diff
@@ -445,53 +445,58 @@ func (s *Switch) SendHTLC(firstHop lnwire.ShortChannelID, paymentID uint64,
...
+	log.Tracef("Updating link policies: %v", newLogClosure(func() string {
+		return spew.Sdump(chanPolicies)
+	}))
+
 	for _, link := range linksToUpdate {
+		newPolicy := dbPolicyToFwdingPolicy(
+			chanPolicies[*link.ChannelPoint()],
+		)
+
 		link.UpdateForwardingPolicy(newPolicy)
```
joostjager: Why not update in the loop above? As we don't error out anymore, it can be done in a single pass?

rpcserver.go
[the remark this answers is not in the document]
joostjager: Discussed offline. Still not convinced about the business case for adding this granularity on the rpc level. Because the `UpdateChannelPolicy` is currently buggy with zero values, I think there is an opportunity to make the fields mandatory now without it being considered as really breaking (it was broken already). But as discussed, leaving further development out of this pr's scope.

htlcswitch/switch.go
valentinewallace: good one, fixed

rpcserver.go
```diff
@@ -4615,30 +4615,16 @@ func (r *rpcServer) UpdateChannelPolicy(ctx context.Context,
 	// With the scope resolved, we'll now send this to the
 	// AuthenticatedGossiper so it can propagate the new policy for our
 	// target channel(s).
-	err := r.server.authGossiper.PropagateChanPolicyUpdate(
+	chanPolicies, err := r.server.authGossiper.PropagateChanPolicyUpdate(
 		chanPolicy, targetChans...,
 	)
 	if err != nil {
 		return nil, err
 	}
```
joostjager: What happens if there are concurrent fee update calls, say call A and call B, for the same channel? I think it is possible that A updates the db, then B updates the db, then B applies policy B to the link and finally A applies policy A to the link. At that point the db and link are no longer in sync.

wpaulino: Looks like a pre-existing issue, but it seems unlikely given that the gossiper can only handle one `PropagateChanPolicyUpdate` call at a time? B would start after A is done, at which point A is likely to have reached the switch already.

joostjager: Yes, indeed looks unlikely, but it still is a small gap. And pre-existing indeed. Maybe we can just throw a lock around it all? Or at least add a TODO describing this. Or do nothing at all, given that it is pre-existing.

valentinewallace: Added a lock around the db+link update code!

rpcserver.go
```diff
@@ -4617,30 +4623,21 @@ func (r *rpcServer) UpdateChannelPolicy(ctx context.Context,
 	// With the scope resolved, we'll now send this to the
 	// AuthenticatedGossiper so it can propagate the new policy for our
 	// target channel(s).
-	err := r.server.authGossiper.PropagateChanPolicyUpdate(
+	policyUpdateLock.Lock()
+	chanPolicies, err := r.server.authGossiper.PropagateChanPolicyUpdate(
 		chanPolicy, targetChans...,
 	)
 	if err != nil {
 		return nil, err
```
joostjager: lock isn't released

valentinewallace: Released it [here](https://github.com/lightningnetwork/lnd/pull/3439/files#diff-01ada5680aaa18fc79cbf78b2d55a37bR4640), otherwise isn't there a chance the aforementioned concurrency problem could still happen?

valentinewallace: Released it [here](https://github.com/lightningnetwork/lnd/pull/3439/files#diff-01ada5680aaa18fc79cbf78b2d55a37bR4640), otherwise isn't there a chance the aforementioned concurrency problem could still happen?

joostjager: I mean in the error case

valentinewallace: Ahh. Thanks, fixed

PR #3440, opened 2019-08-30
routing: add build route functionality

joostjager:
This PR adds a new rpc for advanced users that allows them to build a fully specified route from a list of pubkeys. The call looks up the relevant channel policies from the graph and calculates the required fees and time lock values.

Route building is exposed via `lncli`. Example command to build a two hop route:

`lncli buildroute --hops <pubkey1>,<pubkey2> --amt 500`

It is also possible to leave out the `--amt` parameter. In that case, a route is built for the minimum amount that it can carry. This option is useful to execute probe payments for which the primary objective is to pick up black holes (stuck htlc). If the payment gets stuck, the time-locked amount is minimal.


routing/router.go
```diff
@@ -2293,3 +2293,265 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	errDisconnectedSequence = errors.New("disconnected channel sequence")
+
+	errUnknownPolicy = errors.New("unknown channel policy")
+
+	errIncompatiblePolicies = errors.New("incompatible channel policies")
+)
+
+// BuildRoute returns a fully specified route based on a list of channels.
+func (r *ChannelRouter) BuildRoute(source route.Vertex, amt *lnwire.MilliSatoshi,
+	addresses []interface{}, finalCltvDelta int32, useMinAmt bool) (
```
wpaulino: Nit: include the expected types of `addresses` in the godoc.

routing/router.go
```diff
@@ -2293,3 +2293,265 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+				return nil
+			}
+
+			if err := toNode.ForEachChannel(nil, cb); err != nil {
+				return nil, err
+			}
+
+			if chanAddr == nil {
+				return nil, errors.New("no channels available")
```
wpaulino: Nit: error could use some more context by including the public key.

routing/router.go
```diff
@@ -2293,3 +2293,265 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+				if inEdge == nil {
+					return nil
+				}
+
+				// Before we can process the edge, we'll need to
+				// fetch the node on the _other_ end of this
+				// channel as we may later need to iterate over
+				// the incoming edges of this node if we explore
+				// it further.
+				chanFromNode, err := edgeInfo.FetchOtherNode(
```
wpaulino: It doesn't seem necessary to deserialize the whole `LightningNode` here.

routing/router.go
```diff
@@ -2293,3 +2293,265 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+				return errors.New("channel htlc size " +
+					"constraints violated")
+			}
+		}
+
+		// Adapt the minimum and maximum amount to send to what this
+		// channel allows.
+		//
+		// TODO(joostager): can MaxHTLC be zero?
```
wpaulino: It can be 0 if the max HTLC bit isn't set. Also, the spec doesn't mention whether this value can be 0, but it should be less than min HTLC. It wouldn't make sense to have a min HTLC 0 so we can assume the lowest allowed is 1.

cmd/lncli/routerrpc_active.go
```diff
@@ -6,5 +6,6 @@ import "github.com/urfave/cli"
 
 // routerCommands will return nil for non-routerrpc builds.
 func routerCommands() []cli.Command {
-	return []cli.Command{queryMissionControlCommand, resetMissionControlCommand}
+	return []cli.Command{queryMissionControlCommand,
```
wpaulino: Nit: can place each one of these on its own line to help readability.

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,116 @@
...
+	defer conn.Close()
+
+	client := routerrpc.NewRouterClient(conn)
+
+	if !ctx.IsSet("hops") {
+		return errors.New("hops required")
+	}
+
+	// Build list of hop addresses for the rpc.
+	hops := strings.Split(ctx.String("hops"), ",")
```
wpaulino: Check hops >= 2.

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,116 @@
...
+		if err == nil {
+			rpcHop = &routerrpc.HopAddress{
+				Type: &routerrpc.HopAddress_Pubkey{
+					Pubkey: pubkey[:],
+				},
+			}
+		} else {
+			chanID, err := strconv.ParseUint(k, 10, 64)
+			if err != nil {
+				return errors.New("unrecognized hop address")
```
wpaulino: Nit: include invalid channel id in the error.

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,116 @@
...
+		mode = routerrpc.BuildRouteMode_SUBSTITUTE_MIN_AMT
+	case !amtSet && useMinAmt:
+		mode = routerrpc.BuildRouteMode_MIN_AMT
+	default:
+		return errors.New("must specify either amt or set use_min_amt")
+	}
+
+	// Call BuildRoute rpc.
+	req := &routerrpc.BuildRouteRequest{
+		AmtMsat:        ctx.Int64("amt") * 1000,
```
wpaulino: Any reason to allow msat over the RPC interface but not the CLI?

cmd/lncli/commands.go
```diff
@@ -2474,25 +2475,37 @@ func sendToRoute(ctx *cli.Context) error {
...
-		return fmt.Errorf("no routes provided")
-	}
+		if len(routes.Routes) != 1 {
+			return fmt.Errorf("expected a single route, but got %v",
+				len(routes.Routes))
+		}
+
+		route = routes.Routes[0]
+	} else {
+		routes := &routerrpc.BuildRouteResponse{}
```
wpaulino: We should update the description of the `sendtoroute` command with support for this.

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+		cli.Int64Flag{
+			Name: "final_cltv_delta",
+			Usage: "number of blocks the last hop has to reveal " +
+				"the preimage",
+			Value: lnd.DefaultBitcoinTimeLockDelta,
+		},
+		cli.StringFlag{
+			Name:  "hops",
+			Usage: "comma separated channel ids and/or hex pubkeys",
```
halseth: Shouldn't we just stick to pubkeys only? As chanids won't be guaranteed.

lnrpc/routerrpc/router.proto
```diff
@@ -377,6 +377,73 @@ message PairHistory {
...
+    AMT = 0;
+
+    /**
+    Build the route for the minimum amount that the route can carry. Ignore the
+    amt parameter
+    */
+    MIN_AMT = 1;
+
+    /**
+    Select and verify channels for the given amount, but build the final route
```
halseth: Can't this be simulated by calling with AMT first, then follow up with MIN_AMT?

routing/router.go
```diff
@@ -2293,3 +2293,265 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	// In the second pass, select channels for hops that are specified by
+	// pubkey and verify channel policies.
+	edges := make([]*channeldb.ChannelEdgePolicy, len(addresses))
+
+	// Keep running amounts for the payment amount and the min and max for
+	// this route.
+	var (
+		amtToSend    = amt
+		minAmtToSend = lnwire.MilliSatoshi(0)
+		maxAmtToSend = lnwire.MilliSatoshi(^uint64(0))
```
halseth: I'm wondering whether it could be useful to also expose the max amt on the RPC?

lnrpc/routerrpc/router.proto
```diff
@@ -377,6 +377,73 @@ message PairHistory {
     bool last_attempt_successful = 6 [json_name = "last_attempt_successful"];
 }
 
+
+/// HopAddress contains either a channel or a pubkey to address the next hop.
+message HopAddress {
```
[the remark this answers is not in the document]
joostjager: Removed oneof construct and this message.

routing/router.go
```diff
@@ -2293,3 +2293,265 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
 
 	return bandwidthHints, nil
 }
+
+var (
+	errDisconnectedSequence = errors.New("disconnected channel sequence")
```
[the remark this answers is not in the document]
joostjager: Errors replaced by a different one. Added comments to that one.

routing/router.go
joostjager: Addresses removed, but go doc expanded on new params.

routing/router.go
joostjager: Converted to `ForEachNodeChannel`

routing/router.go
joostjager: extended error added

routing/router.go
joostjager: I am not sure what you suggest here. To compensate for malformed channel updates? I understand the MaxHTLC=0 case, but anything else, I am not sure if we should 'try to make the best of it'.

cmd/lncli/routerrpc_active.go
joostjager: yes, nicer. fixed

cmd/lncli/cmd_build_route.go
joostjager: It also supports a one hop route. Source is implicitly the self node.

cmd/lncli/cmd_build_route.go
joostjager: fixed

cmd/lncli/cmd_build_route.go
joostjager: That stems from cli commands like `sendpayment` that also only accept sats.

cmd/lncli/commands.go
joostjager: updated

cmd/lncli/cmd_build_route.go
joostjager: Discussed offline and decided to stick to pubkeys only with an optional outgoing channel restriction for the first hop.

lnrpc/routerrpc/router.proto
joostjager: Looking at it again, I found this mode over-complicated. It would only have a function when there would be distinct policies within the set of channels between two nodes. This is discouraged anyway in the spec.

routing/router.go
joostjager: What use case do you have in mind? I'd think that the max mostly fails unless all channels are fully unbalanced.

halseth: True, it would be mostly for curiosity and can be estimated with just min(capacities) anyway.

routing/router.go
```diff
@@ -2293,3 +2293,278 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+			if chanFromNode.PubKeyBytes != fromNode {
+				return nil
+			}
+
+			// Validate whether this channel's policy is satisfied
+			// and obtain the new running amounts if this channel
+			// was to be selected.
+			newAmts, err := r.checkOrUpdateAmt(
+				inEdge, edgeInfo.Capacity, amts, localChan,
+				amt == nil,
```
halseth: could maybe extract this into `bool` to make it clear what it is

routing/router.go
```diff
@@ -2293,3 +2293,278 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+		// There is no matching channel. Stop building the route here.
+		if bestEdge == nil {
+			return nil, ErrNoChannel{
+				fromNode: fromNode,
+				position: i,
+			}
+		}
+
+		log.Debugf("Select channel %v at position %v", bestEdge.ChannelID, i)
```
halseth: trace?

routing/router.go
```diff
@@ -2293,3 +2293,278 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+		// along the way. Iterate over all edges to calculate the
+		// receiver amount.
+		receiverAmt = amts.amt
+		for _, edge := range edges[1:] {
+			receiverAmt -= edge.ComputeFeeFromIncoming(receiverAmt)
+		}
+	}
+
+	// Build and return the final route.
+	return newRoute(
```
halseth: sanity check that the first hop amt is equal to our `amts.amt`?

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,98 @@
...
+
+var buildRouteCommand = cli.Command{
+	Name:     "buildroute",
+	Category: "Payments",
+	Usage:    "Build a route from a list hop pubkeys.",
+	Action:   actionDecorator(buildRoute),
+	Flags: []cli.Flag{
+		cli.Int64Flag{
+			Name: "amt",
+			Usage: "the amount to send expressed in satoshis. if" +
```
halseth: s/if/If

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,98 @@
...
+				"not set, the minimum routable amount is used",
+		},
+		cli.Int64Flag{
+			Name: "final_cltv_delta",
+			Usage: "number of blocks the last hop has to reveal " +
```
halseth: This is the best explanation of `final_cltv_delta` I've seen 😅

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,98 @@
...
+			Usage: "comma separated hex pubkeys",
+		},
+		cli.Uint64Flag{
+			Name: "outgoing_chan_id",
+			Usage: "short channel id of the outgoing channel to " +
+				"use for the first hop of the payment",
+			Value: 0,
+		},
+		cli.BoolFlag{
+			Name: "use_min_amt",
```
halseth: redundant

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,98 @@
...
+		if err != nil {
+			return fmt.Errorf("error parsing %v: %v", k, err)
+		}
+		rpcHops = append(rpcHops, pubkey[:])
+	}
+
+	var amt int64
+	amtSet := ctx.IsSet("amt")
+	if amtSet {
+		amt = ctx.Int64("amt") * 1000
```
halseth: what if amt is 0?

routing/router.go
```diff
@@ -2293,3 +2293,278 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+
+func (e ErrNoChannel) Error() string {
+	return fmt.Sprintf("no matching outgoing channel available for "+
+		"node %v (%v)", e.position, e.fromNode)
+}
+
+// BuildRoute returns a fully specified route based on a list of pubkeys. If
+// amount is nil, the minimum routable amount is used. To force a specific
+// outgoing channel, use the outgoingChan parameter.
+func (r *ChannelRouter) BuildRoute(amt *lnwire.MilliSatoshi,
```
halseth: is method strictly dependent on the router, or could it be defined in a smaller file? 😛

channeldb/graph.go
```diff
@@ -2832,6 +2832,22 @@ func (c *ChannelEdgePolicy) ComputeFee(
...
+	return (dividend + factor - 1) / factor
+}
+
+// ComputeFeeFromIncoming computes the fee to forward an HTLC given the incoming
+// amount.
+func (c *ChannelEdgePolicy) ComputeFeeFromIncoming(
+	incomingAmt lnwire.MilliSatoshi) lnwire.MilliSatoshi {
+
+	return incomingAmt - divideCeil(
+		1e6*(incomingAmt-c.FeeBaseMSat),
```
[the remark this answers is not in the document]
joostjager: fixed

routing/router.go
joostjager: I don't think it gets very pretty if this method would be moved to a new struct. If we want to do something about the size of `ChannelRouter`, it probably makes more sense to separate the graph update part from the routing part.

routing/router.go
joostjager: extracted

routing/router.go
joostjager: I think this is covered in the unit test where we assert on `route.TotalAmount`. I am no fan of sanity checks.

cmd/lncli/cmd_build_route.go
joostjager: wasn't sure, because other 'sentences' explaining flags are not capitalized either. Changed to capital I.

cmd/lncli/cmd_build_route.go
joostjager: Copy/paste magic...

cmd/lncli/cmd_build_route.go
joostjager: It will try to build a route for 0 sats. This will require all minhtlc settings set to 0.

routing/pathfind.go
```diff
@@ -482,7 +482,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		var fee lnwire.MilliSatoshi
 		var timeLockDelta uint16
 		if fromVertex != source {
-			fee = computeFee(amountToSend, edge)
```
wpaulino: Can `computeFee` be removed now?

routing/router_test.go
```diff
@@ -3330,3 +3330,98 @@ func TestSendToRouteStructuredError(t *testing.T) {
...
+
+	// Test a route that contains incompatible channel htlc constraints.
+	// There is no amount that can pass through both channel 5 and 4.
+	hops = []route.Vertex{
+		ctx.aliases["e"], ctx.aliases["c"],
+	}
+	_, err = ctx.router.BuildRoute(
+		nil, hops, nil, 40,
+	)
+	if _, ok := err.(ErrNoChannel); !ok {
```
wpaulino: Asserts fields of error as well?

lnrpc/routerrpc/router_server.go
```diff
@@ -611,3 +615,49 @@ func marshallFailureReason(reason channeldb.FailureReason) (
...
+		if err != nil {
+			return nil, err
+		}
+		hops[i] = pubkey
+	}
+
+	// Prepare BuildRoute call parameters from rpc request.
+	var amt *lnwire.MilliSatoshi
+	if req.AmtMsat != -1 {
+		rpcAmt := lnwire.MilliSatoshi(req.AmtMsat)
```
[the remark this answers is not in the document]
joostjager: I need to get a reference to a value of the right type.

cmd/lncli/cmd_build_route.go
joostjager: I had a vague idea of some nodes supporting zero amount payments just to signal something. But thinking about it again, it sounds not very useful. Added check.

routing/pathfind.go
joostjager: Yes. Wondering why the linter didn't catch this. I have been experimenting with various settings of `golangci-lint`. It seems that some combinations suppress certain results.

routing/router_test.go
joostjager: Added

lnrpc/routerrpc/router.proto
```diff
@@ -380,6 +380,38 @@ message PairHistory {
     bool last_attempt_successful = 6 [json_name = "last_attempt_successful"];
 }
 
+message BuildRouteRequest {
+    /**
+    The amount to send expressed in msat. If set to -1, the minimum routable
+    amount is used.
```
wpaulino: Since 0 can't be used now, can we use that as the default instead? So this would also cover the case of the field not being set.

routing/router.go
```diff
@@ -2249,3 +2249,288 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	// amt is the intended amount to send via the route.
+	amt lnwire.MilliSatoshi
+
+	// max is the running maximum that the route can carry.
+	max lnwire.MilliSatoshi
+}
+
+// checkOrUpdateAmt checks or updates the running amounts while
+// traversing the channels.
+func (r *ChannelRouter) checkOrUpdateAmt(policy *channeldb.ChannelEdgePolicy,
```
wpaulino: Nit: doesn't need to be a method of `ChannelRouter`.

routing/router.go
```diff
@@ -2249,3 +2249,288 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+		max: lnwire.MilliSatoshi(^uint64(0)),
+	}
+	if amt != nil {
+		amts.amt = *amt
+	} else {
+		// For minimum amount routes, aim to deliver at least 1 msat to
+		// the destination. There are nodes in the wild that have a
+		// min_htlc channel policy of zero, which could lead to a zero
+		// amount payment being made.
+		amts.amt = 1
```
wpaulino: :+1:

lnrpc/routerrpc/router.proto
joostjager: I am a little hesitant to use zero as a magic value for amount, because in other contexts amount can legitimately be zero. Remembering min/max htlc setting. But looking at this particular situation, zero has no meaning. So I changed it.

routing/router.go
joostjager: Made it a method of runningAmts.

routing/router.go
joostjager: Yes, that showed up during probing on mainnet.

joostjager:
Also extended the test so that it also exercises selecting the best channel when there are multiple channels between a pair of nodes.

routing/router_test.go
```diff
@@ -3330,3 +3330,137 @@ func TestSendToRouteStructuredError(t *testing.T) {
...
+
+	// Build the route for the given amount.
+	rt, err := ctx.router.BuildRoute(
+		&amt, hops, nil, 40,
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	checkHops(rt, []uint64{1, 2})
+	if rt.TotalAmount != 105000 {
```
halseth: where does this number come from?

routing/router_test.go
```diff
@@ -3330,3 +3330,137 @@ func TestSendToRouteStructuredError(t *testing.T) {
...
+
+	// Build the route for the minimum amount.
+	rt, err = ctx.router.BuildRoute(
+		nil, hops, nil, 40,
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	checkHops(rt, []uint64{1, 2})
+	if rt.TotalAmount != 21000 {
```
halseth: this number should also have a comment

routing/router.go
```diff
@@ -2249,3 +2249,290 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+// prepending the given channel to the route. If canIncreaseAmt is set, the
+// amount may be increased if it is too small to satisfy the channel's minimum
+// htlc amount.
+func (r *runningAmounts) stepBack(policy *channeldb.ChannelEdgePolicy,
+	capacity btcutil.Amount, localChan bool, canIncreaseAmt bool) (
+	runningAmounts, error) {
+
+	// Determine max htlc value.
+	maxHtlc := lnwire.NewMSatFromSatoshis(capacity)
+	if policy.MaxHTLC != 0 {
```
halseth: Should use `HasMaxHtlc`?

routing/router.go
```diff
@@ -2249,3 +2249,290 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+		if amt < policy.MinHTLC || amt > maxHtlc {
+			return runningAmounts{}, fmt.Errorf("channel htlc "+
+				"constraints [%v - %v] violated with amt %v",
+				policy.MinHTLC, maxHtlc, amt)
+		}
+
+		// Update running amount by adding the fee for non-local
+		// channels.
+		if !localChan {
+			amt += policy.ComputeFee(amt)
```
halseth: I wonder if min/max htlc checks can also be skipped for local channels? 🤔

routing/router.go
```diff
@@ -2249,3 +2249,290 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	if maxHtlc < r.max {
+		max = maxHtlc
+	}
+
+	// If we get in the situation that the minimum amount exceeds
+	// the maximum amount (enforced further down stream), we have
+	// incompatible channel policies.
+	//
+	// There is possibility with pubkey addressing that we should
+	// have selected a different channel downstream, but we don't
```
halseth: Why don't we fix that? Should this be a TODO, or is there another reason?

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,94 @@
...
```
halseth: to avoid the confusion here, could you name one var `amt_sat` and the other `amt_msat`? 😅

routing/router.go
joostjager: Changed. This is the confusion that I think we shouldn't need to have.

joostjager: Hmm, looking in the link, it also checks for `MaxHTLC == 0`

routing/router.go
joostjager: The switch enforces it, so we need to check for it here too.

routing/router.go
joostjager: It would complicate path finding while we expect this situation to be rare. The spec recommends to keep all policies towards a peer identical. If that is the case, there isn't a better channel that we should have selected.

Added this as a comment

cmd/lncli/cmd_build_route.go
joostjager: one of the vars is `amtSet`, not `amtSat` :smile: Renamed the other one to `amtMsat`. I didn't change the command line parameter, because our standard in most cases is implictly sats for `lncli`.

routing/router_test.go
joostjager: comment added

joostjager: comment added

lnrpc/routerrpc/router.proto
```diff
@@ -380,6 +380,38 @@ message PairHistory {
...
+    /**
+    The channel id of the channel that must be taken to the first hop. If zero,
+    any channel may be used.
+    */
+    uint64 outgoing_chan_id = 3;
+
+    /**
+    A list of hops that defines the route.
+    */
+    repeated bytes hop_pubkeys = 4;
```
halseth: should note whether the source is part of this list or not.

routing/router.go
```diff
@@ -2249,3 +2249,297 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+	if policy.MaxHTLC != 0 {
+		maxHtlc = policy.MaxHTLC
+	}
+
+	amt := r.amt
+
+	// If we have a specific amount for which we are building the route,
+	// validate it against the channel constraints and return the new
+	// running amount.
+	if !canIncreaseAmt {
```
halseth: maybe this had would been less confusing if the param was called `findMinAmt` or similar.

routing/router.go
```diff
@@ -2249,3 +2249,297 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+			// Continue searching if this channel doesn't
+			// connect with the previous hop.
+			if chanFromNode.PubKeyBytes != fromNode {
+				return nil
+			}
+
+			// Validate whether this channel's policy is satisfied
+			// and obtain the new running amounts if this channel
+			// was to be selected.
+			newAmts, err := amts.stepBack(
```
halseth: naming suggestion: `prependChannel`

routing/router.go
```diff
@@ -2249,3 +2249,297 @@ func generateBandwidthHints(sourceNode *channeldb.LightningNode,
...
+				inEdge, edgeInfo.Capacity, localChan,
+				useMinAmt,
+			)
+			if err != nil {
+				log.Tracef("Skipping chan %v: %v",
+					inEdge.ChannelID, err)
```
halseth: Is this an expected error? If not maybe log as Warn

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/lightningnetwork/lnd"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/urfave/cli"
+)
+
+var buildRouteCommand = cli.Command{
+	Name:     "buildroute",
+	Category: "Payments",
+	Usage:    "Build a route from a list hop pubkeys.",
```
halseth: *a list of

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, k := range hops {
+		pubkey, err := route.NewVertexFromStr(k)
+		if err != nil {
+			return fmt.Errorf("error parsing %v: %v", k, err)
+		}
+		rpcHops = append(rpcHops, pubkey[:])
+	}
+
+	var amtMsat int64
+	amtSet := ctx.IsSet("amt")
```
halseth: naming suggestion: `hasAmt`

routing/router.go
joostjager: I purposely renamed it to `canIncreaseAmt` because I realized that it is a more generic function here.

routing/router.go
joostjager: yes, better

routing/router.go
joostjager: Yes, it can happen if channels have distinct policies. If all of them do not match, we return a proper error. `Warn` is definitely too heavy in this situation.

routing/router.go
halseth: Would it work to not pass this bool to the method, and instead let the caller check the returned amount? So if the amount has increased, but we cannot increase it, it will take action.

It should be equivalent, I just find the current branching within the `prependChannel` a bit confusing.

routing/router.go
halseth: It's just that if this is expected to happen, and it leads us to not be able to build a route, the reason will likely never surface, as it's on trace.

routing/router.go
joostjager: I understand your suggestion, but I think it may be confusing in another way. The current code has already seen a few earlier iterations. If there isn't a way that very clearly improves the readability, I'd rather leave it as is.

routing/router.go
joostjager: The trace log is intended for use during development. The problem with reporting the exact reason to the caller is similar to the situation that we have in path finding. There may be multiple reasons. I am afraid that just exposing a random one of them will lead to the same incorrect assumptions that we currently have for path finding.

joostjager: To be clear, even if the call succeeds in building a route, some channels may have been skipped. We can't log it on Warn, because it may really not be a warning.

routing/router.go
wpaulino: That's because the link uses `ForwardingPolicy`, which doesn't contain information about whether the bit is set or not. I'm in favor of adding the check for the bit being set here.

joostjager: Ok, it is indeed a good idea, because this function is also used for non-local channels. Changed

cmd/lncli/cmd_build_route.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		rpcHops = append(rpcHops, pubkey[:])
+	}
+
+	var amtMsat int64
+	hasAmt := ctx.IsSet("amt")
+	if hasAmt {
+		amtMsat = ctx.Int64("amt") * 1000
+		if amtMsat == 0 {
+			return fmt.Errorf("non-zero amount required")
```
wpaulino: One last nit: do we still need all of this? Seems like `AmtMsat` can just be `ctx.Uint64("amt") * 1000`.

PR #3441, opened 2019-08-30
htlcswitch+routing: omit next_hop_id in final hop payload

cfromknecht:
Currently we include a `short_channel_id` for all hops in the path, though [BOLT 4 says that it should be omitted for the final hop](https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#requirements). This PR corrects this, and consolidates instantiation of the fields into a `record` package so to reduce code duplication.


routing/route/route.go
```diff
@@ -105,19 +106,9 @@ func (h *Hop) PackHopPayload(w io.Writer, nextChanID uint64) error {
...
-		tlv.MakeDynamicRecord(
-			tlv.LockTimeOnionType, &h.OutgoingTimeLock, func() uint64 {
-				return tlv.SizeTUint32(h.OutgoingTimeLock)
-			},
-			tlv.ETUint32, tlv.DTUint32,
-		),
-		tlv.MakePrimitiveRecord(tlv.NextHopOnionType, &nextChanID),
+		record.NewAmtToFwdRecord(&amt),
+		record.NewLockTimeRecord(&h.OutgoingTimeLock),
+		record.NewNextHopIDRecord(&nextChanID),
```
joostjager: With regards to layering: there is a difference in the specification of the route (`route.Route`) and how to serialize this on the wire. I think ideally those things should stay separate. Maybe `PackHopPayload` and other wire-specific functions could all be moved into `record`?

htlcswitch/iterator.go
```diff
@@ -175,15 +176,9 @@ func (r *sphinxHopIterator) ForwardingInstructions() (ForwardingInfo, error) {
...
-				tlv.ETUint64, tlv.DTUint64,
-			),
-			tlv.MakeDynamicRecord(
-				tlv.LockTimeOnionType, &cltv, nil,
-				tlv.ETUint32, tlv.DTUint32,
-			),
-			tlv.MakePrimitiveRecord(tlv.NextHopOnionType, &cid),
+			record.NewAmtToFwdRecord(&amt),
+			record.NewLockTimeRecord(&cltv),
+			record.NewNextHopIDRecord(&cid),
```
joostjager: Is this actually partitioning the network for people running master?

cfromknecht: technically yes, but only for sender and receiver where one is older master and one has this commit. none of the implementations have done a release yet tho

routing/route/route.go
cfromknecht: interesting point, probably not a bad idea. tho primary purpose of `record` atm is just to reduce code duplication

joostjager: I don't think it is a must for now. But if forces in that direction show up anyway, it may be good to keep in mind.

htlcswitch/iterator.go
joostjager: ok, probably not a problem then. people running master now will probably upgrade relatively often to later master commits.

routing/route/route.go
cfromknecht: Definitely, fortunately I think it’s the direction we’re headed 👍

htlcswitch/iterator.go
joostjager: Minor comment: as a reviewer, I'd prefer this and the previous commit to be squashed so that it makes a logical unit of work and I can compare that it is indeed just a code move. Now looking at the first commit raises the question what is going to happen next and I see myself going back and forth between these commits.

But it isn't important and subjected to personal preference too.

record/hop.go
```diff
@@ -43,5 +43,10 @@ func NewLockTimeRecord(lockTime *uint32) tlv.Record {
 // NewNextHopIDRecord creates a tlv.Record that encodes the short_channel_id
 // (type 6) for an onion payload.
 func NewNextHopIDRecord(cid *uint64) tlv.Record {
-	return tlv.MakePrimitiveRecord(NextHopOnionType, cid)
+	return tlv.MakeDynamicRecord(
+		NextHopOnionType, cid, func() uint64 {
+			return tlv.SizeTUint64(*cid)
+		},
+		tlv.ETUint64, tlv.DTUint64,
+	)
```
joostjager: Is a more elaborate commit message needed, as this commit is (in theory) partitioning the network?

cfromknecht: now that we simply omit the field, rather than change the encoding it's no longer a partition since it wasn't being enforced before. this code has never been deployed in a release so i think its less critical even if other impls are enforcing it

cfromknecht:
@joostjager reworked the pr to omit the next hop id rather than modify the encoding, ptal

routing/route/route.go
```diff
@@ -108,9 +108,18 @@ func (h *Hop) PackHopPayload(w io.Writer, nextChanID uint64) error {
 	combinedRecords := append(h.TLVRecords,
```
joostjager: Is something bad happening here because the underlying slice array is modified?

cfromknecht: fixed

record/hop.go
Roasbeef: Yeah this doesn't change any behavior, as if the cid wasn't present, then we would be passing around the default value, which in this case, is the same as `exitHop`.

PR #3442, opened 2019-08-30
single-shot, sender-side mpp via sendtoroute

cfromknecht:
This PR lays some more groundwork for our road to MPP payments in lnd. Here we add the ability to send MPP shards from the sender-side via `SendToRoute`, whose additional fields are properly encoded in the onion blob and delivered to the receiver.

Apart from verifying that mpp records are not received at intermediate hops, the receiver does not perform any additional validation of the actual fields contained within. The receiver side has been modified to deliver these values all the way down to the invoice registry, but they are not inspected. Thus, the receiver will process the payment as they would a traditional payment. The additional behavior to enforce the MPP fields will be added in #3415, at which point the behaviors will diverge.

For now, the integration tests have been expanded to cover the three send to route endpoints `Lightning.SendToRoute`, `Lightning.SendToRouteSync`, and `RouterRPC.SendToRoute`, as well as testing that each can properly pay with single MPP shards. Future changes will continue to expand the assertions as more pieces come into place.


htlcswitch/payload/hop.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/record"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// Hop encapsulates all information delivered in an onion payload. A Hop can
+// represent either a TLV or legacy payload. The primary forwarding instructsion
+// can be accesed via ForwardingInfo, and additional records can be accessed by
+// other member functions.
+type Hop struct {
```
joostjager: Isn't `type Payload` more appropriate?

htlcswitch/payload/hop.go
```diff
@@ -0,0 +1,84 @@
...
+// provided by leegacy onion payloads.
+func NewLegacyHop(amt uint64, cltv uint32, nextHop lnwire.ShortChannelID) *Hop {
+	return &Hop{
+		FwdInfo: ForwardingInfo{
+			Network:         BitcoinNetwork,
+			NextHop:         nextHop,
+			AmountToForward: lnwire.MilliSatoshi(amt),
+			OutgoingCTLV:    cltv,
+		},
+	}
```
joostjager: Shouldn't deserialization of the legacy hop bytes happen in here as well? (tlv is also deserialized in this object)

htlcswitch/payload/hop.go
cfromknecht: done

htlcswitch/payload/hop.go
cfromknecht: done

cfromknecht:
> I also experienced that the onion blob decryption code needs to be moved out of htlcswitch so that it can be used in cnct without a circular reference. Will we move that into htlcswitch/payload too?

Discussed offline. Plan is to move decryption (`HopIterator`) into the `hop` package and replace with `hop.Iterator`

lnrpc/routerrpc/router.proto
```diff
@@ -161,6 +161,22 @@ message SendToRouteRequest {
...
+    */
+    int64 total_amt_msat = 3;
+
+    /**
+    If total_amt_msat is non-zero, a unique, random identifier used to
+    authenticate the sender as the intended payer of a multi-path payment. The
+    payment_addr must be the same for all subpayments, and match the
+    payment_addr provided in the receiver's invoice.
+    */
+    bytes payment_addr = 4;
```
joostjager: Does it need to be added here or into the message `Hop`? Previously, `Route` was a fully specified route that only needed to be launched. Now there are external values required too. I am also thinking about returning a route, for example after a successful payment. Then those external values would be required too for a complete view on what was executed.

htlcswitch/hop/payload.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return &Payload{
+		FwdInfo: ForwardingInfo{
+			Network:         BitcoinNetwork,
+			NextHop:         lnwire.NewShortChanIDFromInt(cid),
+			AmountToForward: lnwire.MilliSatoshi(amt),
+			OutgoingCTLV:    cltv,
+		},
+		MPP: mpp,
```
joostjager: We talked offline about "option_mpp shouldn't be included for a non-final node", but also for the final node it would be useful to know whether mpp fields where included. Otherwise we get matching on default values in invoice registry.

joostjager: Ok, that is what you already said on chat.

htlcswitch/hop/payload.go
```diff
@@ -0,0 +1,88 @@
...
+		cltv uint32
+		mpp  = new(record.MPP)
+	)
+
+	tlvStream, err := tlv.NewStream(
+		record.NewAmtToFwdRecord(&amt),
+		record.NewLockTimeRecord(&cltv),
+		record.NewNextHopIDRecord(&cid),
+		mpp.TLV(),
+	)
```
joostjager: Is there a way to prevent this re-instantiation of the tlv stream for every payload?

record/mpp.go
```diff
@@ -0,0 +1,43 @@
+package record
+
+import (
+	"io"
+
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+type MPP struct {
+	TotalMsat   uint64
```
joostjager: `Total lnwire.MilliSatoshi`

htlcswitch/hop/payload.go
joostjager: For the inv registry changes that I am making, it would be nice to set `MPP` to nil already here when `TotalAmt` is zero.

htlcswitch/payload/hop.go
joostjager: I actually meant `HopData.Decode` that is now in `lightning-onion`

htlcswitch/hop/payload.go
cfromknecht: yes, we could make a singleton decoding payload, and then use a deep copy to move the parsed values into a different payload struct

record/mpp.go
cfromknecht: agreed, i have these changes locally just need to rebase everything

htlcswitch/payload/hop.go
cfromknecht: ah i see, maybe we can add a todo since it is a bit more involved

lnrpc/routerrpc/router.proto
```diff
@@ -161,6 +161,22 @@ message SendToRouteRequest {
 
     /// Route that should be used to attempt to complete the payment.
     lnrpc.Route route = 2;
+
+    /**
+    If non-zero, the total amount in satoshis being sent as part of a larger
+    multi-path payment. The caller is responsible for ensuring subpayments to
```
halseth: This comment is a bit confusing, should say something like "the total amount in satoshis of the payment if this is a sub payment of smaller amount"? I don't know what the best way to put this..

routing/route/route.go
```diff
@@ -169,6 +177,16 @@ type Route struct {
 	Hops []*Hop
 }
 
+// AddMPP extends existing tlv records destined for the final hop
+// with an option_mpp record use to construct base AMP payments.
+func (r *Route) AddMPP(total lnwire.MilliSatoshi, addr [32]byte) {
```
halseth: I think it makes more sense to set this at `Route` creation time if possible, to avoid mutation.

record/mpp.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// MPPEncoder writes the MPP record to the provided io.Writer.
+func MPPEncoder(w io.Writer, val interface{}, buf *[8]byte) error {
+	if v, ok := val.(*MPP); ok {
+		err := tlv.EUint64T(w, uint64(v.totalMsat), buf)
+		if err != nil {
+			return err
+		}
+		return tlv.EBytes32(w, &v.paymentAddr, buf)
```
joostjager: I guess this now needs to be updated for the swapped order that we discussed in the last spec meeting.

record/mpp.go
```diff
@@ -0,0 +1,73 @@
...
+// NewMPP generates a new MPP record with the given total and payment address.
+func NewMPP(total lnwire.MilliSatoshi, addr [32]byte) *MPP {
+	return &MPP{
+		totalMsat:   total,
+		paymentAddr: addr,
+	}
+}
+
+// TotalMsat returns the total value of an MPP payment in msats.
+func (r *MPP) TotalMsat() lnwire.MilliSatoshi {
```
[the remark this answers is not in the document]
cfromknecht: not necessarily, but it ensures that `MPP` objects are only mutable while decoding, and forces instantiation of both fields via the constructor. it makes the surrounding code less brittle and the compiler will inline the getters so there isn't a performance overhead.

record/mpp.go
```diff
@@ -0,0 +1,73 @@
...
+// MPP is a record that encodes the fields necessary for multi-path payments.
+type MPP struct {
+	// totalMsat is the total value of the payment, potentially spread
+	// across more than one HTLC.
+	totalMsat lnwire.MilliSatoshi
+
+	// paymentAddr is a random, receiver-generated value used to avoid
+	// collisions with concurrent payers.
+	paymentAddr [32]byte
+}
```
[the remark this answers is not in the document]
cfromknecht: your follow up does this, no?

lnrpc/rpc.proto
```diff
@@ -1811,6 +1811,23 @@ message Hop {
...
+    */
+    int64 mpp_total_amt_msat = 10 [json_name = "mpp_total_amt_msat"];
+
+    /**
+    If mpp_total_amt_msat is non-zero, a unique, random identifier used to
+    authenticate the sender as the intended payer of a multi-path payment. The
+    mpp_payment_addr must be the same for all subpayments, and match the
+    mpp_payment_addr provided in the receiver's invoice.
+    */
+    bytes mpp_payment_addr = 11 [json_name = "mpp_payment_addr"];
```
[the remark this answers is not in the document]
cfromknecht: hmm good point, yes i hadn't really considered using via `lncli`, only via rpc. i suppose this one downside to adding the fields to the hop level, as they are inherently exposed everywhere where routes are used.

> Should we hide those two fields in the display code of lncli

I think we want to display them no? If an mpp payment is completed?

> When parsing a route for SendToRoute in lncli, we need to add code to convert the hex string in json to bytes.

This indeed should be fixed

routing/route/route.go
```diff
@@ -94,6 +103,10 @@ type Hop struct {
 	// carries as a fee will be subtracted by the hop.
 	AmtToForward lnwire.MilliSatoshi
 
+	// MPP encapsulates the data required for option_mpp. This field should
+	// only be set for the final hop.
+	MPP *record.MPP
```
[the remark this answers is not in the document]
cfromknecht: it is in `PackHopPayload`, you mean for channeldb?

lnrpc/rpc.proto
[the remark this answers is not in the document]
joostjager: Yes, let's just display them. I assume that there will be something mpp in 0.9, so then it is also ok to have these fields.

record/mpp.go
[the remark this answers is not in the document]
joostjager: oh, that is right indeed. https://github.com/lightningnetwork/lnd/pull/3415/commits/cca3a98367a7397c6b5bf34b6c3ec70e95c3aa5d

you can cherry-pick it. or not :)

record/mpp.go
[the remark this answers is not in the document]
joostjager: What can I say... approve.

routing/route/route.go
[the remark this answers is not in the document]
halseth: Yeah, when the route is stored to the DB in order to handle resuming a payment after a restart.

record/mpp.go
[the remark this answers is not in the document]
joostjager: Sure, I can do the GUI :stuck_out_tongue:

routing/route/route.go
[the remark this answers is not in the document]
cfromknecht: MPP fields are now serialized in the individual hops!

htlcswitch/hop/payload.go
```diff
@@ -99,13 +105,18 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
 		return nil, err
 	}
 
+	if _, ok := parsedTypes[record.MPPOnionType]; !ok {
```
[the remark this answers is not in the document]
cfromknecht: fixed

htlcswitch/hop/payload.go
```diff
@@ -74,12 +78,14 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
 		cid  uint64
 		amt  uint64
 		cltv uint32
+		mpp  = new(record.MPP)
```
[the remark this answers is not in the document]
cfromknecht: imo its cleaner when not populating fields, but fixed :)

lnrpc/rpc.proto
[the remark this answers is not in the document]
cfromknecht: ended up pulling in copies of `encoding/json` and `github.com/golang/protobuf/jsonpb` to print all byte slices as hex going forward. 

one caveat is that this is a breaking change for the listinvoices rpc which currently prints payment hashes and preimages in base64

record/mpp.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"io"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// MPPOnionType is the type used in the onion to reference the MPP fields:
+// total_amt and payment_addr.
+const MPPOnionType tlv.Type = 14890322186762188277 // temporary, optional field
```
Roasbeef: Such number, very large, wow

record/mpp.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		return tlv.ETUint64T(w, uint64(v.totalMsat), buf)
+	}
+	return tlv.NewTypeForEncodingErr(val, "MPP")
+}
+
+// MPPDecoder reads the MPP record to the provided io.Reader.
+func MPPDecoder(r io.Reader, val interface{}, buf *[8]byte, l uint64) error {
+	if v, ok := val.(*MPP); ok && l >= 32 && l <= 40 {
```
Roasbeef: Should now remove the first length check correct? As we always expect _both_ to be present.

htlcswitch/hop/payload.go
```diff
@@ -74,12 +78,14 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
 		cid  uint64
 		amt  uint64
 		cltv uint32
+		mpp  = &record.MPP{}
```
Roasbeef: Seems this is mostly there to just pass in the proper record below? Alternatively, we can create a new function in the `record` package, just as we do for the current default routing types.

lntest/itest/lnd_test.go
```diff
@@ -4406,6 +4408,225 @@ func testMultiHopPayments(net *lntest.NetworkHarness, t *harnessTest) {
...
+
+			ctxt, _ = context.WithTimeout(ctxb, defaultTimeout)
+			err = node.WaitForNetworkChannelOpen(ctxt, chanPoint)
+			if err != nil {
+				t.Fatalf("%s(%d): timeout waiting for "+
+					"channel(%s) open: %v", nodeNames[i],
+					node.NodeID, point, err)
+			}
+		}
+	}
```
Roasbeef: Everything above this is boiler plate code duplicated amongst several tests that could likely be extracted into a common set up function.

lntest/itest/lnd_test.go
```diff
@@ -4406,6 +4408,225 @@ func testMultiHopPayments(net *lntest.NetworkHarness, t *harnessTest) {
...
+		paymentsResp, err := node.ListPayments(ctxt, req)
+		if err != nil {
+			t.Fatalf("error when obtaining %s payments: %v",
+				node.Name(), err)
+		}
+		if len(paymentsResp.Payments) != numPayments {
+			t.Fatalf("incorrect number of payments, got %v, want %v",
+				len(paymentsResp.Payments), numPayments)
+		}
+
```
Roasbeef: Should also assert the correctness of the new fields.

lntest/itest/lnd_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/btcsuite/btcutil"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/go-errors/errors"
+	"github.com/golang/protobuf/jsonpb"
```
Roasbeef: Wrong import path being used if we're going with the direction of forking the proto package.

channeldb/payments.go
```diff
@@ -604,6 +615,24 @@ func deserializeHop(r io.Reader) (*route.Hop, error) {
 		tlvMap[tlvType] = rawRecordBytes
 	}
 
+	// If the MPP type is present, remove it from the generic TLV map and
```
Roasbeef: 👍

record/mpp.go
cfromknecht: first length check is needed to ensure the payload contains a full 32-byte payment secret

record/mpp.go
cfromknecht: ahh yeah i forgot to push up the change that now use the proper type from the spec

lnrpc/rpc.proto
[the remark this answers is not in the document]
cfromknecht: defering hex/base64 discussion to #3647

htlcswitch/hop/payload.go
cfromknecht: not sure i follow, the struct is defined in this scope so that it can be set on the payload later (and nilled if we don't need it)

lntest/itest/lnd_test.go
cfromknecht: jsonpb change is now in a separate pr, this was only used to print debug statements

lntest/itest/lnd_test.go
cfromknecht: the itest will be added in a later pr, can do that then

channeldb/payments.go
```diff
@@ -508,7 +509,9 @@ func deserializePaymentAttemptInfo(r io.Reader) (*PaymentAttemptInfo, error) {
 
 func serializeHop(w io.Writer, h *route.Hop) error {
 	if err := WriteElements(w,
-		h.PubKeyBytes[:], h.ChannelID, h.OutgoingTimeLock,
+		h.PubKeyBytes[:],
```
halseth: commit should be squashed with commit when field is introduced.

halseth:
This looks pretty good to me now, only a question

channeldb/payments.go
```diff
@@ -525,10 +528,18 @@ func serializeHop(w io.Writer, h *route.Hop) error {
 		return WriteElements(w, uint32(0))
 	}
 
+	// Gather all non-primitive TLV records so that they can be serialized
+	// as a single blob.
+	var records []tlv.Record
+	if h.MPP != nil {
+		records = append(records, h.MPP.Record())
+	}
+	records = append(records, h.TLVRecords...)
```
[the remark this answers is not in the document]
cfromknecht: `TLVRecords` is for custom fields offered via rpc. Since this is a known field we extract it. Unfortunately we will have to do this for every known field going forward...

channeldb/payments.go
cfromknecht: done!

lntest/itest/lnd_test.go
cfromknecht: fixed, and expanded the test suite!!

lntest/itest/lnd_test.go
cfromknecht: the new fields aren't exposed in listpayments or listinvoices yet

lntest/itest/lnd_test.go
```diff
@@ -4488,16 +4488,6 @@ func testSingleHopSendToRoute(net *lntest.NetworkHarness, t *harnessTest) {
...
 
-	// We'll wait for all parties to recognize the new channels within the
-	// network.
-	ctxt, _ = context.WithTimeout(ctxb, defaultTimeout)
-	err = net.Bob.WaitForNetworkChannelOpen(ctxt, chanPointAlice)
-	if err != nil {
-		t.Fatalf("alice didn't advertise her channel in time: %v", err)
-	}
-
-	time.Sleep(time.Millisecond * 50)
```
joostjager: Add comment to commit msg explaining why it is not need. Are the channels already recognized at this point?

lntest/itest/lnd_test.go
```diff
@@ -4456,101 +4547,238 @@ func testSingleHopSendToRoute(net *lntest.NetworkHarness, t *harnessTest) {
...
+				"want: \"\", got: %s",
+				i, p.PaymentRequest)
+		}
+
+		// Assert the payment ammount is correct.
+		if p.ValueSat != paymentAmtSat {
+			t.Fatalf("incorrect payment amt for payment %d, "+
+				"want: %d, got: %d",
+				i, paymentAmtSat, p.ValueSat)
+		}
```
joostjager: I assume that because there is no route returned yet, we cannot assert on the mpp fields being saved properly.

channeldb/payments.go
```diff
@@ -525,10 +528,18 @@ func serializeHop(w io.Writer, h *route.Hop) error {
...
+
 	// Otherwise, we'll transform our slice of records into a map of the
 	// raw bytes, then serialize them in-line with a length (number of
 	// elements) prefix.
-	mapRecords, err := tlv.RecordsToMap(h.TLVRecords)
+	mapRecords, err := tlv.RecordsToMap(records)
```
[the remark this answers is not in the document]
cfromknecht: indeed I agree, I have raised this concern several times in the past but so far the concerns have been dismissed every time. this is one of the reasons I pushed for disabling custom records in 0.8 until this had some resolution. 

my preferred solution is to store everything in a single blob, then the known fields can be extracted directly into the struct, and unknown ones can be extracted into their own map. 

the downside to writing the fields in the traditional way is that we need to add migrations to place the new fields in their logical spot, e.g after existing fields but before the custom records. this is why I prefer serializing everything in one blob, but never we never reached agreement that this was the way to proceed. it will also require a migration at this point since the current split has already been deployed..

lnrpc/routerrpc/router_backend.go
```diff
@@ -712,3 +739,36 @@ func ValidateCLTVLimit(val, max uint32) (uint32, error) {
...
+// if the payment address is not 0 or 32 bytes. If the total amount and payment
+// address are zero-value, the return value will be nil signaling there is no
+// MPP record to attach to this hop. Otherwise, a non-nil reocrd will be
+// contained combining the provided values.
+func UnmarshalMPP(reqTotal int64, reqAddr []byte) (*record.MPP, error) {
+	switch {
+
+	// No MPP fields were provided.
+	case reqTotal == 0 && len(reqAddr) == 0:
+		return nil, nil
```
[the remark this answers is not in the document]
cfromknecht: considered this as well, I’m not opposed. however it doesnt simplify these checks as we still need to ensure the parameters are sane..

channeldb/payments.go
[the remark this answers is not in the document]
joostjager: Is it a problem if they new fields aren't in a logical spot? We can also keep adding all new known fields at the end.

lntest/itest/lnd_test.go
cfromknecht: correct, see this comment https://github.com/lightningnetwork/lnd/pull/3442#discussion_r340993685

lnrpc/routerrpc/router_backend.go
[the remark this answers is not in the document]
cfromknecht: the last three commits (a2b66ac692f52e2dc3e53714906898f1672e3630, 8438444f6bafcdfc65f899aeb5d0af9a09eb8bf7, f608675016ed4fbd8a0076b3c04decc361401407) have this change. i kinda like it, lmk what you think

record/mpp.go
```diff
@@ -0,0 +1,87 @@
...
+func MPPEncoder(w io.Writer, val interface{}, buf *[8]byte) error {
+	if v, ok := val.(*MPP); ok {
+		err := tlv.EBytes32(w, &v.paymentAddr, buf)
+		if err != nil {
+			return err
+		}
+
+		return tlv.ETUint64T(w, uint64(v.totalMsat), buf)
+	}
+	return tlv.NewTypeForEncodingErr(val, "MPP")
```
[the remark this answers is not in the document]
cfromknecht: a little overkill imo, we don't do this anywhere the tlv package, it's just for debugging if people pass in the wrong type

tlv/truncated.go
```diff
@@ -40,6 +40,15 @@ func ETUint16(w io.Writer, val interface{}, buf *[8]byte) error {
...
 }
 
+// ETUint16T is an Encoder for truncated uint16 values, where leading zeros will
+// be omitted. An error is returned if val is not a *uint16.
+func ETUint16T(w io.Writer, val uint16, buf *[8]byte) error {
+	binary.BigEndian.PutUint16(buf[:2], val)
+	numZeros := numLeadingZeroBytes16(val)
+	_, err := w.Write(buf[numZeros:2])
+	return err
+}
```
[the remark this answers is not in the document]
cfromknecht: not necessarily, the prior unified record used the non-generic encoder, so this keeps it on par with the prior version. it felt like the proper time to add it for completeness of the library since we will surely use it again in the future, and we have already have rigorous testing on the truncated encodings that we were able to reuse

lntest/itest/lnd_test.go
cfromknecht: done

record/mpp.go
cfromknecht: constants added

tlv/truncated.go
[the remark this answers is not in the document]
joostjager: ok

lntest/itest/lnd_test.go
joostjager: ok, duplicate. missed that

lnrpc/routerrpc/router_backend.go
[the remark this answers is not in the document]
joostjager: I think it is better because it doesn't suggest that you could use the payment addr without mpp

channeldb/payments.go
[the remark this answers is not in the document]
cfromknecht: squashing down

cfromknecht:
> Also, PR desc is still empty.

@joostjager added

halseth:
ACK on using single blob in follow-up. I still have opinions about not having the optional MPP record be its own field on the struct, but that we can discuss in  the follow-up.

htlcswitch/hop/payload.go
halseth: Agree that it's a bit confusing. I don't know the underlying details of the TLV parsing, but what makes it confusing is that this empty `mpp` variable is used to create the `mpp.Record()` below.  Seems like we could just call a static `mpp.DeafaultRecord()` method, but I guess the variable is passed in by reference (as the other vars here) and modified by the underlying decoder?

lntest/itest/lnd_test.go
```diff
@@ -4406,43 +4405,135 @@ func testMultiHopPayments(net *lntest.NetworkHarness, t *harnessTest) {
...
-// through a provided route with a single hop. We'll create the
-// following network topology:
-//      Alice --100k--> Bob
-// We'll query the daemon for routes from Alice to Bob and then
-// send payments through the route.
+type singleHopSendToRouteCase struct {
+	name string
+
+	// streaming tests streaming SendToRoute if true, otherwise tests
+	// synchronous SenToRoute.
```
halseth: Hm, when is the time to get rid of one of these RPCs?

lntest/itest/lnd_test.go
```diff
@@ -4406,43 +4405,135 @@ func testMultiHopPayments(net *lntest.NetworkHarness, t *harnessTest) {
...
+
+// testSingleHopSendToRoute tests that payments are properly processed through a
+// provided route with a single hop. We'll create the following network
+// topology:
+//      Carol --100k--> Dave
+// We'll query the daemon for routes from Carol to Dave and then send payments
+// by feeding the route back into the various SendToRoute RPC methods. Here we
+// test all three SendToRoute endpoints, forcing each to perform both a regular
+// payment and an MPP payment.
 func testSingleHopSendToRoute(net *lntest.NetworkHarness, t *harnessTest) {
```
halseth: Nice test setup! 👍

Roasbeef:
Needs a rebase! Then I think this is ready to land.

cfromknecht:
@Roasbeef rebased!

lntest/itest/lnd_test.go
cfromknecht: i wouldn't be opposed to deprecating the streaming version and telling people to switch to `SendToRouteSync`, or `RouterRPC.SendToRoute` if they don't mind dealing with breaking changes.

htlcswitch/hop/payload.go
cfromknecht: yeah doing something like `mpp.DefaultRecord()` means we never capture the decoded object and we'd just be throwing away the data. the `mpp.Record()` contains pointers to the fields in `mpp` so that they can be set by the decoder

is the difference just a matter of `NewMPPRecord(&mpp)` vs `mpp.Record()`? fwiw i went with the latter so that they can be easily unit tested, since there's now a single interface new records have to implement and we get testing for free.

Issue #3453, opened 2019-09-03
Cannot use gRPC when server is sitting in Docker container


torkelrogstad:
### Background

I'm having trouble connecting over gRPC when LND is running in a Docker container. 

### Your environment

* version of `lnd`: I've tried both master and 0.7.1
* which operating system: Linux
* version of `btcd`, `bitcoind`, or other backend: `btcd` 0.12.0

### Steps to reproduce

1) Add two lines to the official Docker [guide](https://github.com/lightningnetwork/lnd/tree/master/docker): 
    ``` 
    ports: 
      -  "10009:10009"
    ```
    under `lnd_btc` 
1) Follow the guide, ending up with two LND nodes and one `btcd`
2) Ensure `lnd_btc` has 10009 exposed to the outside world: 
    ```
    $ docker port lnd_btc
   10009/tcp -> 0.0.0.0:10009
    ```
3) Observe that the gRPC server is listening on 10009, here's an excerpt from my log: 
    ```
    $ docker logs lnd_btc
    ...
    2019-09-03 16:55:48.568 [INF] RPCS: RPC server listening on 127.0.0.1:10009
    2019-09-03 16:55:48.568 [INF] RPCS: gRPC proxy started at 127.0.0.1:8080
    ```
4) Verify that it's possible to call `lncli` from within the container: `docker exec -it lnd_btc lncli --network simnet getinfo`
5) Verify that it's possible to `curl` the port, and get a "proper" response: 
    ```
    $ docker exec -it lnd_btc apk add curl
    ...
    $ docker exec pit lnd_btc curl localhost:10009
    curl: (52) Empty reply from server
    ```
6) Observe that this does not happen when `curl`ing from the outside world: 
    ```
    $ curl 127.0.0.1:10009
    curl: (56) Recv failure: Connection reset by peer
    ```

### Expected behaviour

I expect to get the same behavior from my host machine as from the container itself. Note that the end goal here is not sending a `curl` request, but to use gRPC/`lncli`. I get the exact same behavior when using `lncli`.

The end goal is to be able to send gRPC through Go code to a LND instance running in a Docker container. If there's any end to end tutorials on that, that'd be much appreciated.

tyzbit:
Ensure that LND is listening on all interfaces in the conf.  This can be accomplished with `rpclisten=0.0.0.0:10009`.

torkelrogstad:
That worked! Thank you so much.

PR #3461, opened 2019-09-04
Add rpclisten to start-lnd.sh

torkelrogstad:
To make it possible to communicate over gRPC with an instance running in Docker we need to expose the correct RPC interface.

See #3453

### DIFF
--- a/docker/lnd/start-lnd.sh
+++ b/docker/lnd/start-lnd.sh
@@ -59,5 +59,6 @@ exec lnd \
     "--$BACKEND.rpchost"="blockchain" \
     "--$BACKEND.rpcuser"="$RPCUSER" \
     "--$BACKEND.rpcpass"="$RPCPASS" \
+    "--rpclisten=0.0.0.0:10009" \
     --debuglevel="$DEBUG" \
     "$@"


tyzbit:
Depending on your Docker setup, this could be a security concern (for example, if you ran the container with `--network host`, it could expose your RPC port to the Internet).  Generally, listening on all interfaces is something you should do with caution.  With that out of the way, I think that if you're wielding `--network host` you're aware of the inherent dangers of such an action and will know to modify the Docker image accordingly if you want to prevent your RPC from being available on every interface.

utACK.

docker/lnd/start-lnd.sh
```diff
@@ -59,5 +59,6 @@ exec lnd \
     "--$BACKEND.rpchost"="blockchain" \
     "--$BACKEND.rpcuser"="$RPCUSER" \
     "--$BACKEND.rpcpass"="$RPCPASS" \
+    "--rpclisten=0.0.0.0:10009" \
```
halseth: Add "Do not do this!" command perhaps? 😛  Same for `noseedbackup` above

torkelrogstad: > Add "Do not do this!" command perhaps? stuck_out_tongue Same for `noseedbackup` above

As in add a comment above, saying that this is a very bad idea in prod?

guggero:
Taken over and replaced by #4790, original commit attributed to this PR's author.

PR #3462, opened 2019-09-04
routing+routerrpc: improve prob. estimation for untried connections

joostjager:
This PR changes mission control to partially base the estimated probability for untried connections on historical results obtained in previous payment attempts. This prevents the payment loop from getting stuck in an area of the network that is not of good quality. It also incentivizes routing nodes to keep all
of their channels in good shape.

The configuration parameter `--routerrpc.aprioriweight` is a value in the range [0, 1] that defines to what extent historical results should be extrapolated to untried connections. Setting it to 1 will completely ignore historical results and always assume the configured a priori probability for untried connections. A value of 0 will ignore the a priori probability completely and only base the probability on historical results, unless there are none available.

Using this parameter, senders can configure how aggressive nodes should be penalized. A low a priori weight will make mission control switch quickly to using a different node if a tried node doesn't perform well. The downside of this is that the sender may miss out on low fee channels that could possibly complete the payment successfully.

Indirectly, the collective of senders also shapes the network with their settings of this parameter. When senders start to demand a higher performance from routing nodes, routing nodes need to meet that expectation in order to keep generating revenue.


joostjager:
Did a bit of performance testing with and without these changes. With a clean mission control history, there is no noticeable difference. This is expected, because there are hardly iterations in the loop inside `GetNodeProbability`. For the test call to `QueryRoutes`, 330 nodes were visited and 7661 edges expanded.

To see a performance effect, it is probably needed to have a fully loaded mission control history. It is a question how likely this situation is, especially on low-powered devices where pathfinding performance becomes important. There is also a default limit to the mission control history of 1000 past payments, although this can be changed.

I tried a simulation of a full mission control by inserting 10 fake results for every node visited, but didn't see an effect either. Db access during path finding is likely still where time is mainly spent.

I'd propose to not prematurely add optimizations like caching of the node probability in this pr and first see what feedback we get on it from users. I've added another commit that logs useful performance metrics.

routing/missioncontrol.go
```diff
@@ -246,22 +254,51 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
...
+	// and historical observations. This is the part that incentives nodes
+	// to make sure that all (not just some) of their channels are in good
+	// shape. Senders will steer around nodes that have shown a few
+	// failures, even though there may be many channels still untried.
+	total := m.cfg.AprioriHopProbability * aprioriFactor
+	totalWeight := aprioriFactor
 
-	return probability
+	now := m.now()
+	for _, result := range results {
```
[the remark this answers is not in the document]
halseth: In particular I'm interested in the scenarios we talked about earlier:
A node having 100s of untried channels, and 1 tried channel (success/failure)
A node having 100s of untried channels, and X tried channels (success/failure), where X is about 2-5
A node having few channels, and some result (say 3 channels, 2 failures/successes).

routing/missioncontrol.go
```diff
@@ -246,22 +254,51 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
...
+		if result.success {
+			weight = 1
+			total += weight * 1.0
+		} else {
+			weight = m.getWeight(age)
+		}
+		totalWeight += weight
+	}
+
+	return total / totalWeight
```
joostjager: Note to self: handle `totalWeight == 0` case

joostjager: Note to self: explain single observation case

routing/missioncontrol.go
```diff
@@ -60,7 +60,7 @@ const (
 // into the path finding process for subsequent payment attempts.
 type MissionControl struct {
 	// lastPairResult tracks the last payment result per node pair.
-	lastPairResult map[DirectedNodePair]timedPairResult
+	lastPairResult map[route.Vertex]map[route.Vertex]timedPairResult
```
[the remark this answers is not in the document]
joostjager: commit message added

routing/missioncontrol.go
```diff
@@ -254,16 +254,25 @@ func (m *MissionControl) getProbAfterFail(lastFailure time.Time) float64 {
...
+	// probability.
+	weight := m.getWeight(timeSinceLastFailure)
+	probability := m.cfg.AprioriHopProbability * (1 - weight)
 
 	return probability
 }
 
+// getWeight calculates a weight in the range [0, 1] that should be assigned to
+// a payment result.Weight follows an exponential curve that starts at 1 when
+// the result is freshed and asymptotically approaches zero over time. The rate
```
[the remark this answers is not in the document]
joostjager: yes, you know, freshing. fixed :smile:

routing/missioncontrol.go
```diff
@@ -246,22 +254,51 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
...
+	results, ok := m.lastPairResult[node]
+	if !ok {
 		return m.cfg.AprioriHopProbability
 	}
 
-	timeSinceLastFailure := m.now().Sub(lastFailure)
+	// The value of the apriori weight is in the range [0, 1]. Convert it to
+	// a factor that properly expresses the intention of the weight in the
+	// following weight average calculation.
+	aprioriFactor := 1/(1-m.cfg.AprioriWeight) - 1
```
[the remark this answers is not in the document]
joostjager: explanation added

routing/missioncontrol.go
```diff
@@ -246,22 +254,51 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
...
-	// probability 0. Over time the probability recovers to the a priori
-	// probability.
-	weight := m.getWeight(timeSinceLastFailure)
-	probability := m.cfg.AprioriHopProbability * (1 - weight)
+	// Calculate a weighted average consisting of the apriori probability
+	// and historical observations. This is the part that incentives nodes
+	// to make sure that all (not just some) of their channels are in good
+	// shape. Senders will steer around nodes that have shown a few
+	// failures, even though there may be many channels still untried.
+	total := m.cfg.AprioriHopProbability * aprioriFactor
```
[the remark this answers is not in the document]
joostjager: renamed to `probabilitiesTotal`

routing/missioncontrol.go
```diff
@@ -303,40 +340,62 @@ func (m *MissionControl) setLastPairResult(fromNode,
 	nodePairs[toNode] = *result
 }
 
+// setAllPairResult stores a result for all known connection of the given node.
+func (m *MissionControl) setAllPairResult(fromNode route.Vertex,
+	result *timedPairResult) {
```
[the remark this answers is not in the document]
joostjager: This is how we do it for the other functions too. Do you think it will be faster as a value?

routing/missioncontrol.go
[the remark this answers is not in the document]
joostjager: Extracted probability estimation from mission control and added more unit tests.

The number of untried channels isn't a factor in the calculation at all, so I can't write a unit test for that. Probability estimates are only based on previous payment results.

routing/result_interpretation.go
```diff
@@ -367,7 +367,19 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
 // failNode marks the node indicated by idx in the route as failed. This
 // function intentionally panics when the self node is failed.
 func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	// Mark the node as failing.
```
Roasbeef: `idx-1` can be lifted to a new local variable, say `targetNode` perhaps.

routing/result_interpretation.go
```diff
@@ -367,7 +367,19 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
...
 	i.nodeFailure = &rt.Hops[idx-1].PubKeyBytes
+
+	// Mark the incoming connection as failed for the node.
+	inPair, _ := getPair(rt, idx-1)
+	i.pairResults[inPair.Reverse()] = pairResult{}
+
+	// If not the ultimate node, mark the outgoing connection as failed for
+	// the node.
```
Roasbeef: If we have `A <-> B <-> C` and `B` failed, the incoming is ` A -> B` and outgoing is `B -> C`?

routing/missioncontrol.go
```diff
@@ -377,28 +333,21 @@ func (m *MissionControl) GetHistorySnapshot() *MissionControlSnapshot {
...
-	for v, h := range m.lastNodeFailure {
-		otherProb := m.getPairProbability(v, route.Vertex{}, 0)
-
-		nodes = append(nodes, MissionControlNodeSnapshot{
-			Node:             v,
-			LastFail:         h,
-			OtherSuccessProb: otherProb,
-		})
-	}
+	nodes := make([]MissionControlNodeSnapshot, 0)
```
Roasbeef: Style nit: can instead just declare it (guessing you hit the "should pre allocate" directive in the liter):
```go
var nodes []MissionControlSnapshot
```

Roasbeef: If we _really_ want to pre-allocate though, we can add a method or function to compute the total size of the multi-dimensional map.

routing/probability_estimator.go
```diff
@@ -0,0 +1,154 @@
...
+	totalWeight := aprioriFactor
+
+	for _, result := range results {
+		age := now.Sub(result.timestamp)
+
+		if result.success {
+			// Weigh success with a constant high weight of 1. There
+			// is no decay.
+			totalWeight++
+			probabilitiesTotal += prevSuccessProbability
```
Roasbeef: Should be `p.prevSuccessProbability`.

joostjager:
> Completed an initial surface-level pass. Will do another one focusing on the core estimation changes to ensure I fully grok them. Overall I think this is a step in the right direction, but wonder how we can test these assumptions in the wild, and also if it's too early to make a change like this given that as we know it, node operators aren't very sophisticated _yet_.

We can always set the default a priori weight quite high, so the effect of this will be light initially. Give some time for users and operators to become aware of the new scoring.

routing/missioncontrol.go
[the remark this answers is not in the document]
halseth: Was just so you dont have to dereference below.

routing/missioncontrol.go
```diff
@@ -505,11 +460,31 @@ func (m *MissionControl) applyPaymentResult(
...
+	// that case, a node-level failure would not be applied to untried
+	// channels.
 	if i.nodeFailure != nil {
 		log.Debugf("Reporting node failure to Mission Control: "+
 			"node=%v", *i.nodeFailure)
 
-		m.lastNodeFailure[*i.nodeFailure] = result.timeReply
+		m.setAllPairResult(*i.nodeFailure, &timedPairResult{
+			timestamp:  result.timeReply,
+			pairResult: pairResult{},
```
halseth: same comment about setting `success=false`

routing/probability_estimator.go
```diff
@@ -0,0 +1,154 @@
...
+	// Retrieve the node probability that is used for untried channels.
+	nodeProbability := p.getNodeProbability(now, results)
+
+	// Retrieve the last pair outcome.
+	lastPairResult, ok := results[toNode]
+
+	// If there is no history for this pair, return the node probability
+	// that is a probability estimate for untried channel.
+	if !ok {
+		return nodeProbability
```
halseth: maybe defer gettting `nodeProbsbility` to where it is actually returned, to avoid doing result iteration unnecessarily

routing/probability_estimator_test.go
```diff
@@ -0,0 +1,162 @@
...
+			expectedProb, toNode, p)
+	}
+}
+
+// TestProbabilityEstimatorNoResults tests the probability estimation when no
+// results are available.
+func TestProbabilityEstimatorNoResults(t *testing.T) {
+	ctx := newEstimatorTestContext(t)
+
+	ctx.assertPairProbability(testTime, 0, 0, 0.6)
```
halseth: use `aprioriHopProb`?

routing/probability_estimator_test.go
```diff
@@ -0,0 +1,162 @@
...
+	ctx := newEstimatorTestContext(t)
+
+	ctx.results = []timedPairResult{
+		{
+			timestamp:  testTime.Add(-time.Hour),
+			pairResult: pairResult{},
+		},
+	}
+
+	// For an untried node, we expected the node probability. The weight for
```
halseth: Good explanations in these test cases 👍

routing/probability_estimator_test.go
```diff
@@ -0,0 +1,162 @@
...
+	// remain high.
+	ctx.assertPairProbability(testTime, 0, 100, prevSuccessProbability)
+
+	// For an untried node, we expected the node probability to be returned.
+	// This is a weighted average of the results above and the a priori
+	// probability: 0.62.
+	expectedNodeProb := (3*aprioriHopProb + 1*prevSuccessProbability) /
+		(3 + 1 + 0.25 + 0.125)
+
+	ctx.assertPairProbability(testTime, untriedNode, 100, expectedNodeProb)
```
halseth: it is a bit confusing that `untriedNode` and node `1` get different results here, since neither are referenced. I think the index in `ctx.result` is used to denote the `toNode` of the result. Could we instead set this explicitly (by making `results ` a map), and extract `node0 = 0`, `node1 = 1` into constants for readability?

routing/pathfind.go
```diff
@@ -296,6 +297,18 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	source, target route.Vertex, amt lnwire.MilliSatoshi) (
 	[]*channeldb.ChannelEdgePolicy, error) {
 
+	// Pathfinding can be a significant portion of the total payment
+	// latency, especially on low-powered devices. Log several metrics to
+	// aid in the analysis performance problems in this area.
+	start := time.Now()
```
halseth: nice, we should consider exposing these metrics in the future.

routing/result_interpretation.go
```diff
@@ -367,7 +367,19 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
 // failNode marks the node indicated by idx in the route as failed. This
 // function intentionally panics when the self node is failed.
 func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	// Mark the node as failing.
 	i.nodeFailure = &rt.Hops[idx-1].PubKeyBytes
+
+	// Mark the incoming connection as failed for the node.
+	inPair, _ := getPair(rt, idx-1)
+	i.pairResults[inPair.Reverse()] = pairResult{}
```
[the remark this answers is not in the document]
joostjager: If I remember correctly, there were comments in other prs about not adding redundant initializers. Not sure what you standard is here.

routing/missioncontrol.go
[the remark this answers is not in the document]
joostjager: Fixed

routing/result_interpretation.go
joostjager: Technically it could, both `idx-1` expression don't have the exact same meaning. It is an index into the route hops list and a route channel index.

routing/result_interpretation.go
joostjager: Yes. Added a helper variables to make this clearer.

routing/missioncontrol.go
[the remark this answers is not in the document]
joostjager: Fixed

routing/missioncontrol.go
joostjager: Converted to var initializer. No, I don't necessarily want to pre-allocate :smile:

routing/probability_estimator_test.go
joostjager: fixed

routing/probability_estimator_test.go
joostjager: Yes it is a bit obscure. Followed your suggestion

routing/pathfind.go
joostjager: Where would you want to expose it in the future?

joostjager:
I realized that the calculation of the node probability should also be amount independent. Especially if we make the amount more important in the follow up pr #3493. Updated `GetNodeProbability`. I also removed those precalculated probabilities from the mc state dump and provided a new rpc to query a specific probability.

routing/result_interpretation.go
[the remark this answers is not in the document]
halseth: Personally I like adding them (exceptions being `nil` initializers perhaps), and in this case I feel it makes sense, since an empty struct is not obvious what it means.

joostjager:
Move new `querymcprob` call to #3556

routing/result_interpretation.go
```diff
@@ -367,7 +367,21 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
 // failNode marks the node indicated by idx in the route as failed. This
 // function intentionally panics when the self node is failed.
 func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	// Mark the node as failing.
 	i.nodeFailure = &rt.Hops[idx-1].PubKeyBytes
+
+	// Mark the incoming connection as failed for the node.
+	incomingChannelIdx := idx - 1
```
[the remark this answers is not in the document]
joostjager: See https://github.com/lightningnetwork/lnd/pull/3462/commits/ae91eb79e3959a08f2a2c1699194142144208b68#r326447064

lnrpc/routerrpc/router.proto
```diff
@@ -373,8 +355,7 @@ message PairHistory {
     /// Minimum penalization amount (only applies to failed attempts).
     int64 min_penalize_amt_sat = 4 [json_name = "min_penalize_amt_sat"];
 
-    /// Estimation of success probability for this pair.
-    float success_prob = 5 [json_name = "success_prob"];
+    reserved 5;
```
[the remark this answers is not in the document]
joostjager: Discussed offline. We don't want probabilities for an implicit amount that are prone to misinterpretation.

routing/result_interpretation.go
```diff
@@ -367,7 +367,21 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
 // failNode marks the node indicated by idx in the route as failed. This
 // function intentionally panics when the self node is failed.
 func (i *interpretedResult) failNode(rt *route.Route, idx int) {
```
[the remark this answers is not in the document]
joostjager: Sentence added

routing/missioncontrol.go
```diff
@@ -446,8 +452,7 @@ func (m *MissionControl) applyPaymentResult(
 			"node=%v", *i.nodeFailure)
 
 		m.setAllPairResult(*i.nodeFailure, timedPairResult{
-			timestamp:  result.timeReply,
-			pairResult: pairResult{},
+			timestamp: result.timeReply,
```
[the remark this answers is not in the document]
joostjager: I've added constructors to make it more obvious without adding redundant field initializers.

lnrpc/routerrpc/router.proto
```diff
@@ -348,17 +348,21 @@ message PairHistory {
     
     /// The destination node pubkey of the pair.
     bytes node_to = 2 [json_name="node_to"];
-    
+
+    reserved 3, 4, 5, 6;
+
+    PairData data = 7 [json_name="data"];
```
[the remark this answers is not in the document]
joostjager: Moved to #3556

routing/pathfind.go
joostjager: I am not sure if it is very informative, because it depends mainly on payment details.

routing/result_interpretation.go
```diff
@@ -26,6 +26,20 @@ type pairResult struct {
 	success bool
 }
 
+// newFailPairResult creates a new result struct for a failure.
+func newFailPairResult(minPenalizeAmt lnwire.MilliSatoshi) pairResult {
```
halseth: nit: maybe `new` is redundant? Could do `failPairResult(99)` and `successPairResult` only

routing/missioncontrol.go
```diff
@@ -226,100 +245,37 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
...
-	lastPairResult := m.getLastPairResult(fromNode, toNode)
-
-	// Only look at the last pair outcome if it happened after the last node
-	// level failure. Otherwise the node level failure is the most recent
-	// and used as the basis for calculation of the probability.
-	if lastPairResult != nil && lastPairResult.timestamp.After(lastFail) {
-		if lastPairResult.success {
-			return prevSuccessProbability
-		}
+// setAllPairResult stores a result for all known connection of the given node.
```
halseth: should be `setAllFail`, since only used for failure?

routing/pathfind.go
halseth: It depends on uniform payments being made regularly indeed, but I can imagine it being useful in some cases. But was just an idea for future work.

routing/missioncontrol.go
joostjager: It describes what the function does

routing/result_interpretation.go
joostjager: renamed

routing/probability_estimator.go
```diff
@@ -0,0 +1,161 @@
...
+	// estimates for all other channels being 0 too. The probability for the
+	// channel that was tried will not even recover, because it is
+	// recovering to the node probability (which is zero). So one failure
+	// effectively prunes all channels of the node forever. This is the most
+	// aggressive way in which we can penalize nodes and unlikely to yield
+	// good results in a real network.
+	probabilitiesTotal := p.aprioriHopProbability * aprioriFactor
+	totalWeight := aprioriFactor
+
+	for _, result := range results {
```
Roasbeef: Latest version of this loop is much easier to follow 👍

routing/probability_estimator.go
```diff
@@ -0,0 +1,161 @@
...
+// getPairProbability estimates the probability of successfully traversing to
+// toNode based on historical payment outcomes for the from node. Those outcomes
+// are passed in via the results parameter.
+func (p *probabilityEstimator) getPairProbability(
+	now time.Time, results map[route.Vertex]timedPairResult,
+	toNode route.Vertex, amt lnwire.MilliSatoshi) float64 {
+
+	// Getter for the node probability that is used for untried channels.
+	getNodeProbability := func() float64 {
+		return p.getNodeProbability(now, results, amt)
```
Roasbeef: I don't think this function closure really adds much, other than introducing 3 extra lines in this diff.

server.go
```diff
@@ -660,6 +660,7 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 			AprioriHopProbability: routingConfig.AprioriHopProbability,
 			PenaltyHalfLife:       routingConfig.PenaltyHalfLife,
 			MaxMcHistory:          routingConfig.MaxMcHistory,
+			AprioriWeight:         routing.DefaultAprioriWeight,
```
Roasbeef: We should add this to the routing config so we can experiment with different a prior weight values without having to recompile the target `lnd` binary.

routing/missioncontrol.go
```diff
@@ -60,7 +63,7 @@ const (
 // into the path finding process for subsequent payment attempts.
 type MissionControl struct {
 	// lastPairResult tracks the last payment result per node pair.
-	lastPairResult map[DirectedNodePair]timedPairResult
+	lastPairResult map[route.Vertex]NodeResults
```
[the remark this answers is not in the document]
joostjager: excellent

routing/result_interpretation.go
```diff
@@ -378,10 +378,25 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
...
+	incomingChannelIdx := idx - 1
+	inPair, _ := getPair(rt, incomingChannelIdx)
+	i.pairResults[inPair.Reverse()] = failPairResult(0)
+
+	// If not the ultimate node, mark the outgoing connection as failed for
+	// the node.
+	if idx < len(rt.Hops) {
+		outgoingChannelIdx := idx
+		outPair, _ := getPair(rt, outgoingChannelIdx)
+		i.pairResults[outPair] = failPairResult(0)
```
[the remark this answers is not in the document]
joostjager: We don't need it for this PR. Here it is intentionally set to zero, because it should be a penalty to any amount, not just `>= amt`.

routing/result_interpretation.go
```diff
@@ -378,10 +378,25 @@ func (i *interpretedResult) processPaymentOutcomeUnknown(route *route.Route) {
...
+// marks the incoming and outgoing channels of the node as failed. This function
+// intentionally panics when the self node is failed.
 func (i *interpretedResult) failNode(rt *route.Route, idx int) {
+	// Mark the node as failing.
 	i.nodeFailure = &rt.Hops[idx-1].PubKeyBytes
+
+	// Mark the incoming connection as failed for the node.
+	incomingChannelIdx := idx - 1
+	inPair, _ := getPair(rt, incomingChannelIdx)
+	i.pairResults[inPair.Reverse()] = failPairResult(0)
```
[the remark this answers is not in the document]
joostjager: We intent to penalize as much as we can for a node level failure, including future outgoing traffic for this connection. The pair as it is returned by getPair is directed towards the failed node. Therefore we first reverse the pair.

Added this as comment.

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
joostjager: Yes, a follow up pr adds this info again, but in a separate rpc call that allows the caller the specify an amount.

The "assume" part was indeed light. The only assumption is that the probability applies to any amount greater than or equal to `min_penalize_amt`. This could also have been clarified with a comment. 

For the previously existing `other_success_prob` it was less clear to what amount it actually applied.

But bottom line is that with the changes in this and follow up prs, those assumptions don't hold anymore.

routing/probability_estimator.go
joostjager: It was an attempt to deduplicate code. But was in doubt already myself about it. Unrolled.

routing/missioncontrol.go
halseth: I think an earlier version of the PR changed this to `setAllFail`, since we never use it to set successes?

joostjager: I don't remember exactly, but I think in that version the function could also only set a failure.

halseth: question is why we need one that can set success if that is never used

server.go
joostjager: Definitely needs to be configurable. This is _the_ parameter that is going to shrink the graph, hopefully :) Fixed

routing/missioncontrol.go
joostjager: yes, yagni, fixed.

routing/missioncontrol_test.go
```diff
@@ -174,7 +179,7 @@ func TestMissionControl(t *testing.T) {
 	history := ctx.mc.GetHistorySnapshot()
 
 	if len(history.Pairs) != 3 {
-		t.Fatalf("expected 2 pairs, but got %v", len(history.Pairs))
+		t.Fatalf("expected 3 pairs, but got %v", len(history.Pairs))
```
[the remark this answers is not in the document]
joostjager: done

routing/probability_estimator.go
```diff
@@ -0,0 +1,156 @@
...
+	// prevSuccessProbability is the assumed probability for node pairs that
+	// successfully relayed the previous attempt.
+	prevSuccessProbability float64
+}
+
+// getNodeProbability calculates the probability for connections from a node
+// that have not been tried before. The results parameter is a list of last
+// payment results for that node.
+func (p *probabilityEstimator) getNodeProbability(now time.Time,
+	results map[route.Vertex]timedPairResult,
```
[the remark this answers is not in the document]
joostjager: fixed

routing/probability_estimator.go
```diff
@@ -0,0 +1,156 @@
...
+func (p *probabilityEstimator) getWeight(age time.Duration) float64 {
+	exp := -age.Hours() / p.penaltyHalfLife.Hours()
+	return math.Pow(2, exp)
+}
+
+// getPairProbability estimates the probability of successfully traversing to
+// toNode based on historical payment outcomes for the from node. Those outcomes
+// are passed in via the results parameter.
+func (p *probabilityEstimator) getPairProbability(
+	now time.Time, results map[route.Vertex]timedPairResult,
```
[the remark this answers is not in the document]
joostjager: fixed

routing/probability_estimator_test.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// assertPairProbability asserts that the calculated success probability is
+// correct.
+func (c *estimatorTestContext) assertPairProbability(now time.Time,
+	toNode byte, amt lnwire.MilliSatoshi, expectedProb float64) {
+
+	c.t.Helper()
+
+	results := make(map[route.Vertex]timedPairResult)
```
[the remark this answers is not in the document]
joostjager: fixed

PR #3465, opened 2019-09-04
tlv+htlcswitch: validate presence/omission of parsed onion payload types

cfromknecht:
Builds on #3441 

This commit adds an additional return value to `Stream.Decode`, which
returns the set of types that were encountered during decoding. The set
will contain all known types that were decoded, as well as unknown odd
types that were ignored.

The rationale for the return value (rather than an internal member) is
so that the stream remains stateless.

This return value can be used by callers during decoding to make
assertions as to whether specific types were included in the stream.
This is need, for example, when parsing onion payloads where certain
fields must be included/omitted depending on the hop type.

### DIFF
--- a/htlcswitch/hop/payload.go
+++ b/htlcswitch/hop/payload.go
@@ -53,7 +53,7 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
 		return nil, err
 	}
 
-	err = tlvStream.Decode(r)
+	_, err = tlvStream.DecodeWithParsedTypes(r)
 	if err != nil {
 		return nil, err
 	}
--- a/htlcswitch/hop/payload.go
+++ b/htlcswitch/hop/payload.go
@@ -2,6 +2,7 @@ package hop
 
 import (
 	"encoding/binary"
+	"fmt"
 	"io"
 
 	"github.com/lightningnetwork/lightning-onion"
@@ -9,6 +10,37 @@ import (
 	"github.com/lightningnetwork/lnd/record"
 	"github.com/lightningnetwork/lnd/tlv"
 )
+
+// ErrInvalidPayload is an error returned when a parsed onion payload either
+// included or omitted incorrect records for a particular hop type.
+type ErrInvalidPayload struct {
+	// Type the record's type that cause the violation.
+	Type tlv.Type
+
+	// Ommitted if true, signals that the sender did not include the record.
+	// Otherwise, the sender included the record when it shouldn't have.
+	Omitted bool
+
+	// FinalHop if true, indicates that the violation is for the final hop
+	// in the route (identified by next hop id), otherwise the violation is
+	// for an intermediate hop.
+	FinalHop bool
+}
+
+// Error returns a human-readable description of the invalid payload error.
+func (e ErrInvalidPayload) Error() string {
+	hopType := "intermediate"
+	if e.FinalHop {
+		hopType = "final"
+	}
+	violation := "included"
+	if e.Omitted {
+		violation = "omitted"
+	}
+
+	return fmt.Sprintf("onion payload for %s hop %s record with type %d",
+		hopType, violation, e.Type)
+}
 
 // Payload encapsulates all information delivered to a hop in an onion payload.
 // A Hop can represent either a TLV or legacy payload. The primary forwarding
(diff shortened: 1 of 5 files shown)


tlv/stream.go
```diff
@@ -239,14 +240,17 @@ func (s *Stream) Decode(r io.Reader) error {
...
 
 			// Other unexpected errors.
 			case err != nil:
-				return err
+				return nil, err
 			}
 		}
 
+		// Record the successfully decoded or ignored type.
+		parsedTypes[typ] = struct{}{}
```
joostjager: I don't know how concerned we are with performance of this. An idea could be to let the caller pass in an empty `parsedTypes` map. Here we could skip the assignment if the map is `nil`.

htlcswitch/iterator.go
```diff
@@ -193,6 +193,36 @@ func (r *sphinxHopIterator) ForwardingInstructions() (ForwardingInfo, error) {
 
 		nextHop = lnwire.NewShortChanIDFromInt(cid)
```
joostjager: Is this still needed? I would think that checking `hasNextHop` is enough now?

cfromknecht: we still need to check that if it `hasNextHop`, that they didn't provide an all zeroes sid

tlv/stream.go
cfromknecht: good idea, maybe it should be a separate method, e.g. `DecodeWithParsedTypes` that can call into the inner method and pass nil or non-nil map

joostjager: That is better indeed, to prevent mistakes with passing a reused map.

htlcswitch/iterator.go
joostjager: Ok. I guess that is the consequence of us using a magic value internally..

cfromknecht:
> I dig the stricter parsing, only comment is that it would be nice to have a switch level test on this front to ensure that if we receive a payload as the exit hop that does have the chan ID, then we reject it. This may also be a good opportunity to incorporate the new invalid tlv payload onion error.

Does it make sense to do this in a separate PR? It is not a strict requirement for the mpp scope, but can proceed in parallel and unblock the mpp rebase.

I've added a unit test to assert that `NewPayloadFromReader` fails with the appropriate errors, which should be propagate back as any existing tlv decoding failure. I would propose that we integrate the switch-level test in the same PR that the `InvalidOnionPayload` failure is added.

htlcswitch/iterator.go
cfromknecht: Indeed 😂

tlv/stream.go
```diff
@@ -239,14 +254,20 @@ func (s *Stream) Decode(r io.Reader) error {
...
 
 			// Other unexpected errors.
 			case err != nil:
-				return err
+				return nil, err
 			}
 		}
 
+		// Record the successfully decoded or ignored type if the
+		// calleer provided an initialized TypeSet.
```
joostjager: check sentence

htlcswitch/hop/payload.go
```diff
@@ -73,3 +117,47 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
+	_, hasLockTime := parsedTypes[record.LockTimeOnionType]
+	_, hasNextHop := parsedTypes[record.NextHopOnionType]
+
+	switch {
+
+	// All hops must include an amount to forward.
+	case !hasAmt:
+		return ErrInvalidPayload{
+			Type:     record.AmtOnionType,
+			Name:     "amount to forward",
```
joostjager: Not sure about the string here. Would an enum be overkill? The string here and in the unit test must be kept equal  manually and maybe more matching on the string will happen.

htlcswitch/hop/payload.go
```diff
@@ -53,15 +88,24 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
+	parsedTypes, err := tlvStream.DecodeWithParsedTypes(r)
+	if err != nil {
+		return nil, err
+	}
+
+	nextHop := lnwire.NewShortChanIDFromInt(cid)
+
+	// Validate whether the sender properly included or omitted tlv records
+	// in accordance with BOLT 04.
+	err = ValidateParsedPayloadTypes(parsedTypes, nextHop == Exit)
```
joostjager: Still there is something not ideal here. The function does the validation, but part of the validation (`nextHop == Exit`) happens outside of the function.

htlcswitch/hop/payload.go
```diff
@@ -53,15 +88,24 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
 		return nil, err
 	}
 
-	_, err = tlvStream.DecodeWithParsedTypes(r)
+	parsedTypes, err := tlvStream.DecodeWithParsedTypes(r)
+	if err != nil {
+		return nil, err
```
joostjager: Should the required error also be mapped onto the fancy error type?

htlcswitch/hop/payload.go
cfromknecht: `nextHop` is now passed directly

htlcswitch/hop/payload.go
cfromknecht: done, removed string and only print type in log

htlcswitch/hop/payload.go
cfromknecht: very likely yes, was thinking we might do that when adding the invalid onion payload failure in the follow up

PR #3466, opened 2019-09-04
htlcswitch/hop: new package for hop payload parsing/encapsulation

cfromknecht:
Builds on #3441 

This PR begins the process of moving our hop payload processing within lnd into its own subpackage, `htlcswitch/hop`. This will allow the processing to be imported into the `htlcswitch` and `contractcourt`, and also have a single `hop.Payload` object that can be shared with the invoice registry. The last four commits are predominately a code move, but it will be nice to make progress on the more trivial things so we can do less rebasing on #3442 

This will also allow @joostjager to add his follow up in moving the `HopIterator` components into this package, and establish a stable base to continue mpp/amp development

### DIFF
--- a/htlcswitch/hop/payload.go
+++ b/htlcswitch/hop/payload.go
@@ -1,0 +1,75 @@
+package hop
+
+import (
+	"encoding/binary"
+	"io"
+
+	"github.com/lightningnetwork/lightning-onion"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/record"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// Payload encapsulates all information delivered to a hop in an onion payload.
+// A Hop can represent either a TLV or legacy payload. The primary forwarding
+// instruction can be accessed via ForwardingInfo, and additional records can be
+// accessed by other member functions.
+type Payload struct {
+	// FwdInfo holds the basic parameters required for HTLC forwarding, e.g.
+	// amount, cltv, and next hop.
+	FwdInfo ForwardingInfo
+}
+
+// NewLegacyPayload builds a Payload from the amount, cltv, and next hop
+// parameters provided by leegacy onion payloads.
+func NewLegacyPayload(f *sphinx.HopData) *Payload {
+	nextHop := binary.BigEndian.Uint64(f.NextAddress[:])
+
+	return &Payload{
+		FwdInfo: ForwardingInfo{
+			Network:         BitcoinNetwork,
+			NextHop:         lnwire.NewShortChanIDFromInt(nextHop),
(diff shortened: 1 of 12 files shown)


htlcswitch/hop/payload.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/lightningnetwork/lightning-onion"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/record"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// Payload encapsulates all information delivered to a hop in an onion payload.
+// A Hop can represent either a TLV or legacy payload. The primary forwarding
+// instructsion can be accesed via ForwardingInfo, and additional records can be
```
joostjager: instructsion

htlcswitch/iterator.go
```diff
@@ -85,64 +82,32 @@ func (r *sphinxHopIterator) EncodeNextHop(w io.Writer) error {
...
 		fwdInst := r.processedPacket.ForwardingInstructions
+		p := hop.NewLegacyPayload(fwdInst)
 
-		switch r.processedPacket.Action {
-		case sphinx.ExitNode:
-			nextHop = hop.Exit
-		case sphinx.MoreHops:
-			s := binary.BigEndian.Uint64(fwdInst.NextAddress[:])
-			nextHop = lnwire.NewShortChanIDFromInt(s)
-		}
```
joostjager: Isn't this mapping needed anymore? Is the `Action` field required at all then anymore?

htlcswitch/hop/payload.go
```diff
@@ -0,0 +1,75 @@
...
+
+	err = tlvStream.Decode(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Payload{
+		FwdInfo: ForwardingInfo{
+			Network:         BitcoinNetwork,
+			NextHop:         lnwire.NewShortChanIDFromInt(cid),
```
joostjager: Will this be the location where we query `IsDecoded` for the next hop id, to detect that we didn't get it for the final hop?

htlcswitch/iterator.go
cfromknecht: nope, the action field is no longer required, it can be removed from lightning-onion

htlcswitch/hop/payload.go
cfromknecht: yes, essentially this block of code will be moved into this constructor https://github.com/lightningnetwork/lnd/pull/3465/files#diff-93706aed560df47864bbf09673f8f0b2R196-R225

joostjager: ok looking good

htlcswitch/iterator.go
joostjager: i think we need to talk again about embedding lightning-onion in the lnd repo

PR #3470, opened 2019-09-05
htlcswitch+lnwire: invalid onion payload

cfromknecht:
Builds on #3465 

Modifies the link to return an `InvalidOnionPayload` failure when it is unable to process a TLV onion payload. In the process, we unify the `hop.ErrInvalidPayload` failure to include required type failures, and propagate the type back to the caller in the `InvalidOnionPayload`.


joostjager:
Do you want to make the change to `result_interpretation.go` too?

cfromknecht:
> Do you want to make the change to result_interpretation.go too?

Done

htlcswitch/hop/payload.go
```diff
@@ -11,15 +11,50 @@ import (
...
+	OmittedViolation PayloadViolation = iota
+
+	// IncludedViolation indicates that a type was expected to be omitted
+	// from the payload but was present.
+	IncludedViolation
+
+	// RequiredViolation indicates that an unknown even type was found in
+	// the payload that we could not process.
+	RequiredViolation
+)
```
joostjager: How is this going to work if we require exactly one of two fields? Return A omitted, B required or A required and B omitted?

htlcswitch/hop/payload.go
```diff
@@ -87,6 +118,18 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
+		// Promote any required type failures into ErrInvalidPayload.
+		if e, required := err.(tlv.ErrUnknownRequiredType); required {
+			// NOTE: FinalHop will be incorrect if the unknown
+			// required was type 0. Otherwise, the failure must have
+			// occurred after type 6 and cid should contain an
+			// accurate value.
+			return nil, ErrInvalidPayload{
+				Type:      tlv.Type(e),
+				Violation: RequiredViolation,
+				FinalHop:  cid == 0,
```
joostjager: Is there a use for the `FinalHop` field?

lnwire/onion_error.go
```diff
@@ -78,6 +79,7 @@ const (
 	CodeFinalIncorrectCltvExpiry         FailCode = 18
 	CodeFinalIncorrectHtlcAmount         FailCode = 19
 	CodeExpiryTooFar                     FailCode = 21
+	CodeInvalidOnionPayload                       = FlagPerm | 22
```
joostjager: I still don't understand the `Perm` idea. To me it is similar to the odd/even situation: info that you don't need.

htlcswitch/link.go
```diff
@@ -2648,12 +2648,17 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
...
+			// can be cancelled.
+			var failedType uint64
+			if e, ok := err.(hop.ErrInvalidPayload); ok {
+				failedType = uint64(e.Type)
+			}
+
 			l.sendHTLCError(
 				pd.HtlcIndex,
-				lnwire.NewInvalidOnionVersion(onionBlob[:]),
+				lnwire.NewInvalidOnionPayload(failedType, 0),
```
joostjager: Maybe document offset zero in the code too?

joostjager: Add the TODO in the code as well? A todo in a commit message, is it ever going to be seen again?

routing/result_interpretation.go
```diff
@@ -288,6 +288,14 @@ func (i *interpretedResult) processPaymentOutcomeIntermediate(
 
 		reportOutgoing()
 
+	// If InvalidOnionPayload is received, we penalize only the reporting
+	// node. We know the preceding hop didn't corrupt the onion, since the
+	// reporting node is able to send the failure. We assume that we
+	// constructed a valid onion payload and that the failure is most likely
+	// an unknown required type or a bug in their implementation.
+	case *lnwire.InvalidOnionPayload:
+		i.failNode(route, errorSourceIdx)
```
joostjager: Clean :+1:

Nice to see the overhauled result interpretation being extended

routing/result_interpretation_test.go
```diff
@@ -176,6 +176,61 @@ var resultTestCases = []resultTestCase{
 			nodeFailure:        &hops[1],
 		},
 	},
+
+	// Tests an invalid onion payload from a final hop. The final hop should
+	// be failed while the proceeding hops are reproed as successes. The
+	// failure is terminal since the receiver can't process our onions.
```
joostjager: singular onion?

routing/result_interpretation_test.go
```diff
@@ -176,6 +176,61 @@ var resultTestCases = []resultTestCase{
...
+	// can still try other paths.
+	{
+		name:          "fail invalid onion payload intermediate",
+		route:         &routeFourHop,
+		failureSrcIdx: 2,
+		failure:       lnwire.NewInvalidOnionPayload(0, 0),
+
+		expectedResult: &interpretedResult{
+			nodeFailure: &hops[2],
+		},
```
joostjager: No successes for preceding hops?

tlv/stream.go
```diff
@@ -241,10 +245,30 @@ func (s *Stream) decode(r io.Reader, parsedTypes TypeSet) (TypeSet, error) {
...
+		case typ%2 == 0 && parsedTypes == nil:
 			return nil, ErrUnknownRequiredType(typ)
 
+		// This record type is unknown to the stream and is even, but
+		// since we are tracking the set of parsed types, we'll track
+		// the first such failure and allow parsing to continue. If no
+		// other types of errors are encoutnered, the first failure will
+		// be returned as an ErrUnknownRequiredType so that the full set
+		// of included types can be returned.
+		case typ%2 == 0 && parsedTypes != nil:
```
joostjager: Or just one `case` and start with an early return for `parsedTypes == nil`.

routing/result_interpretation_test.go
cfromknecht: fixed, added `failNode` closure for intermediate hops that also reports successes for preceding channels

htlcswitch/hop/payload.go
cfromknecht: good question, perhaps add an `XorViolation`?

htlcswitch/hop/payload.go
cfromknecht: for now it is only used to print `final hop` or `intermediate hop` in the failure message

htlcswitch/link.go
cfromknecht: moved todo into file

routing/result_interpretation_test.go
cfromknecht: one q: why does `reportIncoming` test that `errorSourceId` is greater than 2, and then only subtract 2?

lnwire/onion_error.go
cfromknecht: not sure, it predates my involvement :)

routing/result_interpretation_test.go
joostjager: I think it is a bug. For `errorSourceIdx` 2, we seem to be skipping a success marking

cfromknecht: makes sense, i tried changing the code and our existing unit tests still passed so we should add that as a test case

cfromknecht: bug fixed and test case added

htlcswitch/hop/payload.go
```diff
@@ -11,15 +11,50 @@ import (
...
+	// payload but was absent.
+	OmittedViolation PayloadViolation = iota
+
+	// IncludedViolation indicates that a type was expected to be omitted
+	// from the payload but was present.
+	IncludedViolation
+
+	// RequiredViolation indicates that an unknown even type was found in
+	// the payload that we could not process.
+	RequiredViolation
```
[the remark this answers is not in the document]
cfromknecht: no they are different, required means that the _sender_ required us to understand a record, but we don't. omitted means that the receiver expected a field that the sender didn't include.

htlcswitch/link.go
```diff
@@ -2649,12 +2649,23 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 		fwdInfo, err := chanIterator.ForwardingInstructions()
 		if err != nil {
 			// If we're unable to process the onion payload, or we
-			// we received malformed TLV stream, then we should
-			// send an error back to the caller so the HTLC can be
-			// canceled.
+			// we received invalid onion payload failure, then we
```
guggero: nit: redundant "we"

htlcswitch/link.go
```diff
@@ -2649,12 +2649,23 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
...
+			// should send an error back to the caller so the HTLC
+			// can be cancelled.
```
guggero: nit: we use "canceled" everywhere now :joy:

htlcswitch/hop/payload.go
```diff
@@ -120,21 +120,20 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
-			//
-			// As a result, this may need to be revisted if it is
-			// decided that the checks below overrule an unknown
-			// required type failure, in which case an
-			// IncludedViolation should be returned instead of the
-			// RequiredViolation.
+			// If the parser returned an unknown required type
+			// failure, we'll first check that the payload is
+			// properly formed according to our known set of
+			// contraints. If an error is discovered, this overrides
```
guggero: nit: typos in "constraints" and "the"

htlcswitch/hop/payload.go
```diff
@@ -120,21 +120,20 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
+			// thre required type failure.
+			nextHop := lnwire.NewShortChanIDFromInt(cid)
```
guggero: nit: maybe move the `nextHop` line to before `DecodeWithParsedTypes` to reduce the amount of duplicated lines?

cfromknecht: the assignment has to be done after `DecodeWithParsedTypes` otherwise `cid` will always be zero since it hasn't been parsed yet. thought about doing it between `DecodeWithParsedTypes` and the error check but wasn't happy with the aesthetics :P

cfromknecht:
thanks for the review @guggero!

htlcswitch/hop/payload.go
```diff
@@ -120,21 +120,20 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
+			err = ValidateParsedPayloadTypes(parsedTypes, nextHop)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
cfromknecht: it is in some sense overkill, but i think an accurate log message will help debug if things go wrong. perhaps another solution is to not log the final/intermediate info when encountering a required violation?

PR #3476, opened 2019-09-06
channeldb: return errors instead of nil 

chokoboko:
This PR fixes a bug where errors are checked, but not returned.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -1870,7 +1870,7 @@ func (c *OpenChannel) CommitmentHeight() (uint64, error) {
 		return nil
 	})
 	if err != nil {
-		return 0, nil
+		return 0, err
 	}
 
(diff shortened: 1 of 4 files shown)


joostjager:
Nice finds. Inspired by this, I did a regex search:
`err != nil {\n.+?return .+?, nil\n` and `err != nil {\n.+?return nil\n`

It yielded a few more:
`cipherseed.go`: `cipherTextToMnemonic`, `ToMnemonic`
`chanbackup/single.go`: `readRemoteKeyDesc` (2x)
`decayedLog.go`: `gcExpiredHashes`
`rpcServer.go`: `VerifyMessage`, `FeeReport`
`briefcase.go`: `encodeOutgoingResolution`
`zpay32.go`: `writeTaggedFields`

If you want, you can check those out too

PR #3478, opened 2019-09-06
lntest: use filepath.Abs for lnd binary path

Roasbeef:
In this commit, we fix a bug that meant the integration tests weren't
able to be run on windows platforms. Before this commit, we manually
constructed a file path, using the regular unix based file path
delimiters. As a result, this wasn't done in a portable way , meaning
that the tests would only run on unix based systems. We resolve this by
using `filepath.Join` which will produce the proper path depending on
the host operating system.

### DIFF
--- a/lntest/node.go
+++ b/lntest/node.go
@@ -345,7 +345,11 @@ func (hn *HarnessNode) start(lndError chan<- error) error {
 	hn.quit = make(chan struct{})
 
 	args := hn.cfg.genArgs()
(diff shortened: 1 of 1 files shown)


Roasbeef:
Got a report from a windows user that this version still doesn't work.

cfromknecht:
test after #4220 is merged

guggero:
Will be fixed by #4475.

PR #10, opened 2016-02-25
README: a couple more improvements

PaulCapestany:
A handful of commits with a few different potential README improvements, one of which aims to address issue #9.

PR #3479, opened 2019-09-07
channeldb: fix migration bug due to interplay between migration #9 an…

Roasbeef:
…d #10

In this commit, we fix an issue that was recently introduced as a result
of migration #10. The new TLV format ended up modifying the
serialization functions called in `serializePaymentAttemptInfo`.
Migration #9, also used this `serializePaymentAttemptInfo` method to
serialize the _new_ (pre TLV, but new payment attempt structure) routes
into the database during its migration. However, migration #10 failed to
copy over the existing unmodified `serializePaymentAttemptInfo` method
into the legacy serialization for migration #9. As a result, once
migration #9 was run, the routes/payments were serialized using the
_new_ format, rather than the format used for v0.7.1. This then lead to
de-serialization either failing, or causing partial payment corruption
as migration #10 was expecting the "legacy" format (no TLV info).

We fix this issue by adding a new fully enclosed
`serializePaymentAttemptInfoMigration9`method that will be used for
migration #9. Note that our tests didn't catch this, as they test the
migration in isolation, rather than in series which is how users will
encounter the migrations.

### DIFF
--- a/channeldb/migration_09_legacy_serialization.go
+++ b/channeldb/migration_09_legacy_serialization.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/coreos/bbolt"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


channeldb/migration_09_legacy_serialization.go
```diff
@@ -253,3 +254,46 @@ func deserializeOutgoingPayment(r io.Reader) (*outgoingPayment, error) {
 
 	return p, nil
 }
+
+func serializePaymentAttemptInfoMigration9(w io.Writer, a *PaymentAttemptInfo) error {
```
halseth: Should have a godoc explaining what these are used for?

PR #3480, opened 2019-09-07
contractcourt: supplement resolvers with confirmed commit set HTLCs

Roasbeef:
In this commit, we fix an existing bug in the package, causing
resolutions to be restarted without their required supplementary
information. This can happen if a distinct HTLC set gets confirmed
compared to the HTLCs that we may have had our commitment at time of
close. Due to this bug, on restart certain HTLCS would be rejected as
they would present their state to the invoice registry, but be rejected
due to checks such as amount value.

To fix this, we'll now pass in the set of confirmed HTLCs into the
resolvers when we re-launch them, giving us access to all the
information we need to supplement the HTLCS.

### DIFF
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -415,7 +415,19 @@ func (c *ChannelArbitrator) Start() error {
 	if startingState == StateWaitingFullResolution &&
 		nextState == StateWaitingFullResolution {
 
-		if err := c.relaunchResolvers(); err != nil {
+		// In order to relaunch the resolvers, we'll need to fetch the
+		// set of HTLCs that were present in the commitment transaction
+		// at the time it was confirmed. commitSet.ConfCommitKey can't
+		// be nil at this point since we're in
+		// StateWaitingFullResolution. We can only be in
+		// StateWaitingFullResolution after we've transitioned from
+		// StateContractClosed which can only be triggered by the local
+		// or remote close trigger. This trigger is only fired when we
+		// receive a chain event from the chain watcher than the
+		// commitment has been confirmed on chain, and before we
+		// advance our state step, we call InsertConfirmedCommitSet.
+		confCommitSet := commitSet.HtlcSets[*commitSet.ConfCommitKey]
+		if err := c.relaunchResolvers(confCommitSet); err != nil {
 			c.cfg.BlockEpochs.Cancel()
 			return err
 		}
@@ -431,7 +443,7 @@ func (c *ChannelArbitrator) Start() error {
 // starting the ChannelArbitrator. This information should ideally be stored in
 // the database, so this only serves as a intermediate work-around to prevent a
 // migration.
-func (c *ChannelArbitrator) relaunchResolvers() error {
+func (c *ChannelArbitrator) relaunchResolvers(confirmedHTLCs []channeldb.HTLC) error {
 	// We'll now query our log to see if there are any active
 	// unresolved contracts. If this is the case, then we'll
 	// relaunch all contract resolvers.
@@ -456,31 +468,22 @@ func (c *ChannelArbitrator) relaunchResolvers() error {
 	// to prevent a db migration. We use all available htlc sets here in
 	// order to ensure we have complete coverage.
 	htlcMap := make(map[wire.OutPoint]*channeldb.HTLC)
-	for _, htlcs := range c.activeHTLCs {
-		for _, htlc := range htlcs.incomingHTLCs {
-			htlc := htlc
-			outpoint := wire.OutPoint{
-				Hash:  commitHash,
-				Index: uint32(htlc.OutputIndex),
-			}
-			htlcMap[outpoint] = &htlc
+	for _, htlc := range confirmedHTLCs {
+		htlc := htlc
+		outpoint := wire.OutPoint{
+			Hash:  commitHash,
+			Index: uint32(htlc.OutputIndex),
 		}
-
-		for _, htlc := range htlcs.outgoingHTLCs {
-			htlc := htlc
-			outpoint := wire.OutPoint{
-				Hash:  commitHash,
-				Index: uint32(htlc.OutputIndex),
-			}
-			htlcMap[outpoint] = &htlc
-		}
+		htlcMap[outpoint] = &htlc
 	}
 
 	log.Infof("ChannelArbitrator(%v): relaunching %v contract "+
 		"resolvers", c.cfg.ChanPoint, len(unresolvedContracts))
 
 	for _, resolver := range unresolvedContracts {
-		c.supplementResolver(resolver, htlcMap)
+		if err := c.supplementResolver(resolver, htlcMap); err != nil {
+			return err
+		}
 	}
 
 	c.launchResolvers(unresolvedContracts)
--- a/contractcourt/channel_arbitrator_test.go
+++ b/contractcourt/channel_arbitrator_test.go
@@ -3,12 +3,16 @@ package contractcourt
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/coreos/bbolt"
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/input"
@@ -127,6 +131,25 @@ func (b *mockArbitratorLog) WipeHistory() error {
 	return nil
 }
 
+// testArbLog is a wrapper around an existing (ideally fully concrete
+// ArbitratorLog) that lets us intercept certain calls like transitioning to a
+// new state.
+type testArbLog struct {
+	ArbitratorLog
+
+	newStates chan ArbitratorState
+}
+
+func (t *testArbLog) CommitState(s ArbitratorState) error {
+	if err := t.ArbitratorLog.CommitState(s); err != nil {
+		return err
+	}
+
+	t.newStates <- s
+
+	return nil
+}
+
 type mockChainIO struct{}
 
 var _ lnwallet.BlockChainIO = (*mockChainIO)(nil)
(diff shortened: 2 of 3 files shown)


contractcourt/channel_arbitrator.go
```diff
@@ -412,7 +412,8 @@ func (c *ChannelArbitrator) Start() error {
 	if startingState == StateWaitingFullResolution &&
 		nextState == StateWaitingFullResolution {
 
-		if err := c.relaunchResolvers(); err != nil {
+		confCommitSet := commitSet.HtlcSets[*commitSet.ConfCommitKey]
```
joostjager: Can you add a comment here explaining why this line can never be hit with a `nil` commit set? (`errNoCommitSet`)

joostjager: When we decide which resolvers to create, we also look at other versions of the commit tx. Could it somehow be that a resolver was created that needs info from a different commit tx version? I seems unlikely, because we mostly create `HtlcFailNowAction` in those cases, but checking just in case.

contractcourt/channel_arbitrator.go
```diff
@@ -455,31 +456,22 @@ func (c *ChannelArbitrator) relaunchResolvers() error {
...
-
-		for _, htlc := range htlcs.outgoingHTLCs {
-			htlc := htlc
-			outpoint := wire.OutPoint{
-				Hash:  commitHash,
-				Index: uint32(htlc.OutputIndex),
-			}
-			htlcMap[outpoint] = &htlc
+	for _, htlc := range confirmedHTLCs {
+		htlc := htlc
```
joostjager: I would expect this not to be possible, because the loop variable was already defined in this scope. But apparently it works. Was already like that I see in diff.

contractcourt/channel_arbitrator.go
Roasbeef: > Could it somehow be that a resolver was created that needs info from a different commit tx version?

Off the top, I can't think of any such scenario. Resolvers are created based on what's in the `htlcActions` map. The `htlcActions` map[populated here](https://github.com/lightningnetwork/lnd/blob/master/contractcourt/channel_arbitrator.go#L1605) by the `constructChainActions` method which only uses the `confCommitSet`. The `confCommitSet` is written to disk as soon as we receive an event of the channel closing on chain. 

If we want to play it safe, we can very well just include _every_ HTLC outpoint in the map to supplement all resolvers.

contractcourt/channel_arbitrator.go
Roasbeef: What wouldn't be possible? Shadowing the loop variable? It's another way to solve the issue of binding to the loop variable when it's a pointer/slice.

joostjager: I was just surprised that there are two scopes inside a for loop, that's all.

contractcourt/channel_arbitrator.go
joostjager: For me, I'd rather not play safe if we don't think it will happen. Can only confuse devs looking at it later.

I was thinking about `checkLocalChainActions` called in `StateDefault` using `activeHtlcs` coming from the link which may contain htlcs that are not available in the finally confirmed commit set. But I don't think that can happen, because the chain actions are only used to decide whether to go on chain at that point and discarded.

contractcourt/channel_arbitrator.go
```diff
@@ -455,31 +467,22 @@ func (c *ChannelArbitrator) relaunchResolvers() error {
 	// to prevent a db migration. We use all available htlc sets here in
 	// order to ensure we have complete coverage.
 	htlcMap := make(map[wire.OutPoint]*channeldb.HTLC)
-	for _, htlcs := range c.activeHTLCs {
```
cfromknecht: is `activeHTLCs` still used?

joostjager: yes, when the channel_arbitrator is in `StateDefault` and watching for expiring htlcs, `activeHTLCs` tracks what is happening in the link.

contractcourt/channel_arbitrator_test.go
```diff
@@ -58,6 +58,34 @@ func (b *mockArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver,
 	for resolver := range b.resolvers {
 		v[idx] = resolver
 		idx++
+
+		// In order to permit restarts, for all possible resolvers,
+		// we'll give them a fresh quit channel. This is required as
+		// otherwise, if a restart is simulated within a test (stopping
+		// the existing channel arb along the way), then the resolvers
+		// will immediately exit, as their quit channel is still
+		// closed.
```
joostjager: Especially because we want to test the contract resolution executes properly after a restart, I don't think we should reuse the resolver instances and their internal state. This also relies on knowledge of the internal implementation of the resolvers, which makes for brittle tests.

I think that creating a temp db and using the real arbitrator log is a better option.

cfromknecht: setting the quit channels in the below cases also produces a race condition in the unit tests, perhaps new resolvers are better as joost suggested

Roasbeef: Modified to use a real DB, added the _start_ of a test context along the way. There's still some weirdness since the arb log inherits part of the config of the channel arb that initially creates it, but this'll require a bigger overhual to rectify.

contractcourt/channel_arbitrator_test.go
```diff
@@ -62,6 +62,34 @@ func (b *mockArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver,
...
+			res.Quit = make(chan struct{})
+			res.htlcAmt = 0
+
+		case *htlcTimeoutResolver:
+			res.Quit = make(chan struct{})
+			res.htlcAmt = 0
+
+		case *commitSweepResolver:
+			res.Quit = make(chan struct{})
+		}
```
joostjager: This block can be removed now?

contractcourt/channel_arbitrator_test.go
```diff
@@ -621,30 +649,15 @@ func TestChannelArbitratorBreachClose(t *testing.T) {
...
-		newStates: make(chan ArbitratorState, 5),
-		resolvers: make(map[ContractResolver]struct{}),
-	}
-
-	chanArbCtx, err := createTestChannelArbitrator(
-		arbLog,
-	)
+	// We create a new test context for this channel arb, notice that we
+	// pass in a nil ArbitratorLog which means that a default one backed by
+	// a real DB will be created. We need this for out test as we want to
```
joostjager: s/out/our

contractcourt/channel_arbitrator_test.go
```diff
@@ -801,29 +816,80 @@ func TestChannelArbitratorLocalForceClosePendingHtlc(t *testing.T) {
...
+	// restart the channel arb below.
+	oldNotifier := chanArb.cfg.Notifier.(*mockNotifier)
+
+	// At this point, in order to simulate a restart, we'll re-create the
+	// channel arbitrator. We do this to ensure that all information
+	// required to properly resolve this HTLC are populated.
+	if err := chanArb.Stop(); err != nil {
+		t.Fatalf("unable to stop chan arb: %v", err)
+	}
+
```
joostjager: call `chanArb.CleanUp()`?

contractcourt/channel_arbitrator_test.go
```diff
@@ -801,29 +816,80 @@ func TestChannelArbitratorLocalForceClosePendingHtlc(t *testing.T) {
 		t.Fatalf("resolution msgs not sent")
 	}
 
+	// We'll grab the old notifier here as our resolvers are still holding
+	// a reference to this instance, and a new one will be created when we
+	// restart the channel arb below.
+	oldNotifier := chanArb.cfg.Notifier.(*mockNotifier)
```
joostjager: Is this still necessary now that the resolvers are reinstantiated?

contractcourt/channel_arbitrator_test.go
```diff
@@ -801,29 +816,80 @@ func TestChannelArbitratorLocalForceClosePendingHtlc(t *testing.T) {
...
+	if err := chanArb.Start(); err != nil {
+		t.Fatalf("unable to start ChannelArbitrator: %v", err)
+	}
+	defer func() {
+		if err := chanArb.Stop(); err != nil {
+			t.Fatalf("unable to stop chan arb: %v", err)
+		}
+
+		chanArbCtx.CleanUp()
+	}()
```
joostjager: Restart can be a clean method on the context.

contractcourt/channel_arbitrator_test.go
```diff
@@ -148,8 +171,41 @@ func (*mockChainIO) GetBlock(blockHash *chainhash.Hash) (*wire.MsgBlock, error)
...
+
+	resolutions chan []ResolutionMsg
+}
+
+func (c *chanArbTestCtx) CleanUp() {
+	c.cleanUp()
+}
+
+func (c *chanArbTestCtx) ResolvedContracts() chan struct{} {
+	return c.resolvedChan
```
joostjager: Why the accessor functions and now using the chan directly?

contractcourt/channel_arbitrator_test.go
```diff
@@ -704,6 +718,7 @@ func TestChannelArbitratorLocalForceClosePendingHtlc(t *testing.T) {
 
 	// The force close request should trigger broadcast of the commitment
 	// transaction.
+	arbLog := chanArb.log.(*testArbLog)
 	assertStateTransitions(
 		t, arbLog.newStates, StateBroadcastCommit,
 		StateCommitmentBroadcasted,
```
joostjager: These are nice to implement as methods on the context, especially if the context knows `t` as well.

contractcourt/channel_arbitrator_test.go
Roasbeef: Started to go in a slightly more elaborate direction, but stopped short. Will modify to just use the channels directly.

contractcourt/channel_arbitrator_test.go
Roasbeef: Unfortunately yes, since the arb log isn't re-created fully, so it passes a `ResolverKit` to the newly deserialized resolvers that still points to the arb config of the _old_ channel arb. The existing set also relied on this fact, as otherwise the channel arb would also be consuming the block epoch notifications, potentially robbing the resolvers of them.

contractcourt/channel_arbitrator_test.go
Roasbeef: Hmm, must've slipped in a rebase, removed!

contractcourt/channel_arbitrator_test.go
Roasbeef: We don't want to call clean up yet, as that'll close the DB and also remove the temp files on disk.

contractcourt/channel_arbitrator_test.go
Roasbeef: Ported them over along with `assertState`.

PR #3481, opened 2019-09-07
Fixed dockerfile by adding version numbers to base containers

legacycode:
The latest golang docker images is based on golang 13. lnd does not compile with golang 13 docker image. It is a good practice to specify version information in the Dockerfile. This ensures, that the correct docker image is used. I also removed the CMD, because in this case it is redundant to ENTRYPOINT.

If possible it would be great, to have the latest images available at docker hub as well. ATM there is a very old image at Docker hub, that did not worked for me.

/legacycode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:alpine as builder
+FROM golang:1.13-alpine as builder
 
 # Force Go to use the cgo based DNS resolver. This is required to ensure DNS
 # queries required to connect to linked containers succeed.
@@ -34,4 +34,3 @@ EXPOSE 9735 10009
 
 # Specify the start command and entrypoint as the lnd daemon.
 ENTRYPOINT ["lnd"]
-CMD ["lnd"]


halseth:
@legacycode lnd works with go 1.13 now, can this be closed?

legacycode:
@halseth i upgraded the build image to 1.13 and it works. It is a good practice to specify the docker tags in the build container. This makes sure the build environment is clearly defined and stays the same. Feel free to merge or close this PR :-)

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:alpine as builder
+FROM golang:1.13-alpine3.10 as builder
```
guggero: Please change this to `1.13-alpine` only.

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apk add --no-cache --update alpine-sdk \
 &&  make install tags="signrpc walletrpc chainrpc invoicesrpc routerrpc"
 
 # Start a new, final image.
-FROM alpine as final
+FROM alpine:3.10 as final
```
guggero: Please remove this.

Dockerfile
legacycode: done

Dockerfile
legacycode: done

guggero:
Thanks, @legacycode! If you could squash the commits into a single one, then I think we'd be good to go.

legacycode:
@guggero done :-)

PR #3485, opened 2019-09-10
channeldb: remove older migrations

joostjager:
To reduce maintenance on migration code, we move to only supporting upgrades from previous major releases.


channeldb/error.go
```diff
@@ -14,6 +14,10 @@ var (
 	// prior database version.
 	ErrDBReversion = fmt.Errorf("channel db cannot revert to prior version")
 
+	// ErrDBVersionTooLow is returned when detecting an attempt to upgrade a
+	// version for which migration is no longer supported.
+	ErrDBVersionTooLow = fmt.Errorf("channel db version to old to upgrade")
```
halseth: s/to/too

channeldb/db.go
```diff
@@ -1125,6 +1127,13 @@ func (d *DB) syncVersions(versions []version) error {
 			latestVersion)
 		return ErrDBReversion
 
+	case meta.DbVersionNumber < minUpgradeVersion:
+		log.Errorf("Refusing to upgrade from db_version=%d to "+
+			"lower version=%d. Upgrade via intermediate major "+
```
halseth: it is not really lower version though? Should be something like "Refusing to upgrade from too old db version=..."?

channeldb/db.go
```diff
@@ -37,72 +37,6 @@ var (
 	// for retrieving all migration function that are need to apply to the
 	// current db.
 	dbVersions = []version{
-		{
```
halseth: carnage! 😮

channeldb/db.go
```diff
@@ -1168,6 +1176,21 @@ func getLatestDBVersion(versions []version) uint32 {
 	return versions[len(versions)-1].number
 }
 
+// getMinUpgradeVersion returns the minimum version required to upgrade the
+// database.
+func getMinUpgradeVersion(versions []version) uint32 {
```
cfromknecht: not a blocker, but now that we're removing migrations, this logic should probably be able to handle the case where there are no migrations in a given release. would need a similar change to `getLatestDBVersion` as well

PR #3486, opened 2019-09-10
multi: remove dead code

joostjager:


contractcourt/briefcase.go
```diff
@@ -176,11 +176,6 @@ func (a ArbitratorState) String() string {
 	}
 }
 
-// resolverType is an enum that enumerates the various types of resolvers. When
-// writing resolvers to disk, we prepend this to the raw bytes stored. This
-// allows us to properly decode the resolver into the proper type.
-type resolverType uint8
```
wpaulino: I think the intent of this was to mark the constants below with this type.

lnwallet/channel.go
```diff
@@ -109,7 +109,7 @@ type channelState uint8
 const (
 	// channelPending indicates this channel is still going through the
 	// funding workflow, and isn't yet open.
-	channelPending channelState = iota
+	channelPending channelState = iota // nolint: unused
```
wpaulino: I don't think this enum is actually serving a purpose at all, perhaps we can remove it completely?

joostjager: What do you mean exactly? `channelState` is used. This particular value not, but I didn't want to remove it to change the numbering (serialization in mind perhaps)

contractcourt/briefcase.go
joostjager: Ok, left in place and added to the values

lnwallet/channel.go
wpaulino: I don't think `LightningChannel` is serialized to disk. I was referring more to if you look at the uses of this enum (not just `channelPending`) within `LightningChannel`, it doesn't seem to be doing much.

joostjager: I gave it a shot, but the closed state is still set and checked. I'd say, "not in this pr" ?

wpaulino: Sure can be left for a follow-up. I think it's just outdated code.

PR #3493, opened 2019-09-12
routing: track amt ranges in mission control

joostjager:
Previously we tracked the last result for a node connection only. This could lead to the situation where a meaningful observation would be overwritten by a far less meaningful one. For example: a failure for 100.000 sats is overwritten by a success for 100 sats.

Besides better route selection in general, this also helps with building the bos node score list.

Probing with minimal amounts in particular will likely wipe out useful history for more representative amounts.

This PR changes mission control to track both that the success and the last failure. This creates three amount ranges for which probabilities are calculated differently:
- `<=successAmt`: return a high probability (default 0.95)
- `successAmt+1` to `failAmt-1`: in this range, we don't have much information. Return the a priori (default 0.6)
- `>=failAmt`: penalize the connection and decay over time back to the a priori


lnrpc/routerrpc/router.proto
```diff
@@ -355,14 +355,22 @@ message PairHistory {
...
+    /**
+    Lowest amount that failed to forward. This may be set to zero if the failure
+    is independent of amount.
+    */
+    int64 fail_amt_sat = 2 [json_name = "fail_amt_sat"];
+
+    reserved 3;
+
+    /// Time of last success.
+    int64 success_time = 4 [json_name = "success_time"];
```
halseth: 👍  This makes a lot of sense from the RPC perspective:)

lnrpc/routerrpc/router.proto
```diff
@@ -355,14 +355,22 @@ message PairHistory {
...
+    /// Time of last failure.
+    int64 fail_time = 1 [json_name = "fail_time"];
 
-    /// Last amount that was (attempted to be) forwarded.
-    int64 amt_sat = 2 [json_name = "min_penalize_amt_sat"];
+    /**
+    Lowest amount that failed to forward. This may be set to zero if the failure
+    is independent of amount.
+    */
+    int64 fail_amt_sat = 2 [json_name = "fail_amt_sat"];
```
halseth: Could also extract this two fields into a new type `Result`.

lnrpc/routerrpc/router_server.go
```diff
@@ -484,12 +484,16 @@ func (s *Server) QueryMissionControl(ctx context.Context,
...
-		AmtSat:                int64(data.Amt.ToSatoshis()),
-		LastAttemptSuccessful: data.Success,
+		FailAmtSat:    int64(data.FailAmt.ToSatoshis()),
+		SuccessAmtSat: int64(data.SuccessAmt.ToSatoshis()),
 	}
 
-	if !data.Timestamp.IsZero() {
-		rpcData.Timestamp = data.Timestamp.Unix()
+	if !data.FailTime.IsZero() {
+		rpcData.FailTime = data.FailTime.Unix()
```
halseth: Won't calling `Unix()` on zero time just be zero anyway?

routing/probability_estimator.go
```diff
@@ -81,19 +81,18 @@ func (p *probabilityEstimator) getNodeProbability(now time.Time,
...
 	for _, result := range results {
-		age := now.Sub(result.Timestamp)
-
-		switch {
-		// Weigh success with a constant high weight of 1. There is no
-		// decay.
-		case result.Success:
+		// Weigh success with a constant high weight of 1. There
+		// is no decay.
+		if result.SuccessAmt != 0 && amt <= result.SuccessAmt {
```
halseth: Shouldn't `SuccessTIme` be used here,, similar to how `FailTIme.IsZero` is checked below?

routing/missioncontrol.go
```diff
@@ -260,11 +261,58 @@ func (m *MissionControl) setLastPairResult(fromNode,
...
-		Success:   result.success,
+	current := nodePairs[toNode]
+
+	// Apply the new result to the existing data for this pair. If there is
+	// no existing data, apply it to the default values for TimedPairResult.
+	if result.success {
+		current.SuccessTime = timestamp
+		successAmt := result.amt
+
+		// Only update the success amount if this amount is higher. This
```
halseth: Won't this be wrong if say you succeeded in sending 100k sats a week ago, and now you successfully send 1 sat? This will be recorded as you successfully just sent 100k sats.

routing/missioncontrol.go
```diff
@@ -260,11 +261,58 @@ func (m *MissionControl) setLastPairResult(fromNode,
...
+		// If the success amount goes into the failure range, move the
+		// failure range up. Future attempts up to the success amount
+		// are likely to succeed. We don't want to clear the failure
+		// completely, because we haven't learnt much for amounts above
+		// the current success amount.
+		if !current.FailTime.IsZero() && successAmt >= current.FailAmt {
+			current.FailAmt = successAmt + 1
+		}
+	} else {
+		// For failures we always want to update both the amount and the
```
halseth: I guess we have something similar here, with a failure to send 100k sats overwrites the information we had about failure to send 1 sat earlier. But this information is most likely hard to keep without actually storing all previous results (which might not be such a bad idea to do eventually?

lnrpc/routerrpc/router.proto
joostjager: I considered that, but didn't want to add the nesting for only two fields.

lnrpc/routerrpc/router_server.go
joostjager: No, zero time is somewhere far before 1970.

routing/probability_estimator.go
joostjager: The condition can actually be removed. If there is no success recorded (or pushed down by an amt independent failure), `successAmt` will be zero and amt is always greater than zero.

routing/missioncontrol.go
joostjager: The reason to update the timestamp is mainly to get an idea of freshness for (future) prepay probe functionality. It isn't used for the probability calculation.

routing/missioncontrol.go
joostjager: It may overwrite it, but it is unlikely to happen soon after the 1 sat failure. Because then the success probability estimate for that edge is still low.

I hope we will never have to store the full list of results and come up with some complicated formula to translate all of that into a probability. If we are lucky, the network quality improves enough to actually make this all less fine-grained.

Roasbeef:
Can be rebased now that the dependent PR is merged.

routing/missioncontrol.go
```diff
@@ -275,11 +265,58 @@ func (m *MissionControl) setLastPairResult(fromNode,
...
+	// no existing data, apply it to the default values for TimedPairResult.
+	if result.success {
+		current.SuccessTime = timestamp
+		successAmt := result.amt
+
+		// Only update the success amount if this amount is higher. This
+		// prevents the success range from shrinking when there is no
+		// reason to do so. For example: small amount probes shouldn't
+		// affect a previous success for a much larger amount.
+		if successAmt > current.SuccessAmt {
```
carlaKC: Is there an argument that scaling down the success amount makes sense, since larger success amounts may deplete the outgoing capacity on a channel? As is, a single large payment "vouches" for all amounts below it succeeding, even though we may not necessarily be seeing payments of this magnitude.

routing/missioncontrol.go
```diff
@@ -125,9 +125,9 @@ type TimedPairResult struct {
 	// timestamp is the time when this result was obtained.
 	Timestamp time.Time
 
-	// minPenalizeAmt is the minimum amount for which a penalty should be
-	// applied based on this result. Only applies to fail results.
-	MinPenalizeAmt lnwire.MilliSatoshi
+	// Amt is the amount that was (attempted to be) forwarded in the last
+	// payment attempt.
```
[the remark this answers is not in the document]
joostjager: Not sure what you mean with 'attempted to be forwarded= failed'. But this is (usually) the amount of the last failure.

Comments fixed to explain this better.

routing/missioncontrol.go
```diff
@@ -494,11 +501,7 @@ func (m *MissionControl) applyPaymentResult(
 		}
 
 		m.setLastPairResult(
-			pair.From, pair.To,
-			newTimedPairResult(
-				result.timeReply,
-				pairResult,
-			),
+			pair.From, pair.To, result.timeReply, &pairResult,
```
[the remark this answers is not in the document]
joostjager: which newline exactly do you mean?

routing/missioncontrol.go
joostjager: Imo we should stay away from that kind of inference. There may be private channels that shadow a single public channel, which interferes with assumptions of that kind.

routing/missioncontrol.go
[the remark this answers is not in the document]
joostjager: Ah, now seeing that this refers to the first commit. Removed the "(attempted to be)" part. It may not be that helpful, because there is also a success boolean in this struct.

routing/missioncontrol.go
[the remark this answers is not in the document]
carlaKC: All the params can fit on one line within 80 char limit
`m.setLastPairResult(pair.From, pair.To, result.timeReply, &pairResult)`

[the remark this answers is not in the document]
joostjager: Also with our standard tab size 8?

routing/missioncontrol_test.go
```diff
@@ -204,12 +204,12 @@ func TestMissionControlChannelUpdate(t *testing.T) {
 	ctx.reportFailure(
 		0, lnwire.NewFeeInsufficient(0, lnwire.ChannelUpdate{}),
 	)
-	ctx.expectP(0, testAprioriHopProbability)
+	ctx.expectP(100, testAprioriHopProbability)
```
carlaKC: Why has this value changed?

joostjager: Amount `0` was never supported, but with the changes in this commit it returns a different result.

I have been thinking about adding an `error` return value to catch this case, but not sure if it is worth it.

carlaKC: Should be fine as is.

lnrpc/routerrpc/router.proto
```diff
@@ -389,8 +389,8 @@ message PairData {
     /// Time stamp of last result.
     int64 timestamp = 1 [json_name = "timestamp"];
 
-    /// Minimum penalization amount (only applies to failed attempts).
-    int64 min_penalize_amt_sat = 2 [json_name = "min_penalize_amt_sat"];
+    /// Last amount that was forwarded.
+    int64 amt_sat = 2 [json_name = "min_penalize_amt_sat"];
```
halseth: should we make this msat? Since msat is used internally, it could lead to some confusion in the rounding case.

routing/result_interpretation.go
```diff
@@ -29,24 +29,28 @@ type pairResult struct {
...
+func successPairResult(successAmt lnwire.MilliSatoshi) pairResult {
 	return pairResult{
 		success: true,
+		amt:     successAmt,
 	}
 }
 
 // String returns the human-readable representation of a pair result.
 func (p pairResult) String() string {
+	var tp string
```
halseth: tp? 😛

routing/missioncontrol.go
```diff
@@ -273,15 +273,20 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
 
 // setLastPairResult stores a result for a node pair.
 func (m *MissionControl) setLastPairResult(fromNode,
-	toNode route.Vertex, result TimedPairResult) {
+	toNode route.Vertex, timestamp time.Time,
```
halseth: these lines actually look too short...

routing/missioncontrol.go
```diff
@@ -273,15 +273,20 @@ func (m *MissionControl) GetProbability(fromNode, toNode route.Vertex,
...
+	result *pairResult) {
 
 	nodePairs, ok := m.lastPairResult[fromNode]
 	if !ok {
 		nodePairs = make(NodeResults)
 		m.lastPairResult[fromNode] = nodePairs
 	}
 
-	nodePairs[toNode] = result
+	nodePairs[toNode] = TimedPairResult{
```
halseth: why this change? Looks like the creation of `TimedPairResult` is just moved.

routing/missioncontrol.go
```diff
@@ -501,11 +508,7 @@ func (m *MissionControl) applyPaymentResult(
 		}
 
 		m.setLastPairResult(
-			pair.From, pair.To,
-			newTimedPairResult(
```
halseth: can `newTimedPairResult` method be deleted?

lnrpc/routerrpc/router.proto
```diff
@@ -386,14 +386,22 @@ message PairHistory {
 }
 
 message PairData {
-    /// Time stamp of last result.
-    int64 timestamp = 1 [json_name = "timestamp"];
+    /// Time of last failure.
```
halseth: could RPC changes have been made its own commit?

routing/probability_estimator.go
```diff
@@ -80,19 +80,20 @@ func (p *probabilityEstimator) getNodeProbability(now time.Time,
...
+
 		// Weigh success with a constant high weight of 1. There is no
-		// decay.
-		case result.Success:
+		// decay. Amt is never zero, so this clause is never executed
+		// when result.SuccessAmt is zero.
+		case amt <= result.SuccessAmt:
 			totalWeight++
-			probabilitiesTotal += p.prevSuccessProbability
+			probabilitiesTotal += prevSuccessProbability
```
halseth: how come this is not `p.prevSuccessProbability`? I don't see where this value is coming from.

routing/probability_estimator.go
```diff
@@ -155,22 +156,23 @@ func (p *probabilityEstimator) calculateProbability(
...
 
-	// For successes, we have a fixed (high) probability. Those pairs
-	// will be assumed good until proven otherwise.
-	if lastPairResult.Success {
-		return p.prevSuccessProbability
+	// For successes, we have a fixed (high) probability. Those pairs will
+	// be assumed good until proven otherwise. Amt is never zero, so this
+	// clause is never executed when lastPairResult.SuccessAmt is zero.
+	if amt <= lastPairResult.SuccessAmt {
+		return prevSuccessProbability
```
halseth: if this is a global constant now, maybe delete the field on the `probabilityEstimator`

routing/missioncontrol.go
halseth: Ok, I see it is deleted in the next commit 👍

lnrpc/routerrpc/router.proto
joostjager: added msat and kept sat

routing/result_interpretation.go
joostjager: Ok, this is not C code... fixed

routing/missioncontrol.go
joostjager: fixed

routing/missioncontrol.go
joostjager: squashed with next commit

lnrpc/routerrpc/router.proto
joostjager: difficult to do, because the internal data model changes and won't match the not-yet-updated rpc protos

routing/probability_estimator.go
joostjager: I think a rebase mistake, fixed

PR #3495, opened 2019-09-12
chainntnfs: start concurrent queues prior to connection

halseth:
We would establish the connection to the chain backend and start getting
block notifications before we had started the concurrent queues, which
would lead to the OnBlockConnected call being blocked, and a deadlock
(since GetBestBlock would never return).

Instead we make sure to start the queues before establishing the
connection, consuming the notifications right away.

This should fix the "router not started" error seen during integration tests.

### DIFF
--- a/chainntnfs/btcdnotify/btcd.go
+++ b/chainntnfs/btcdnotify/btcd.go
@@ -139,17 +139,24 @@ func (b *BtcdNotifier) Start() error {
 		return nil
 	}
 
+	// Start our concurrent queues before starting the chain connection, to
+	// ensure onBlockConnected and onRedeemingTx callbacks won't be
+	// blocked.
+	b.chainUpdates.Start()
+	b.txUpdates.Start()
+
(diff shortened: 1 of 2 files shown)


chainntnfs/btcdnotify/btcd.go
```diff
@@ -162,9 +169,11 @@ func (b *BtcdNotifier) Start() error {
 		Height: currentHeight,
 		Hash:   currentHash,
 	}
-
-	b.chainUpdates.Start()
-	b.txUpdates.Start()
+	if err := b.chainConn.NotifyBlocks(); err != nil {
```
wpaulino: Nit: add newline before.

wpaulino:
Needs a rebase on master with 1.13 support.

chainntnfs/btcdnotify/btcd.go
halseth: done

PR #3499, opened 2019-09-12
channeldb+rpcserver: expose legacy payments as multi-path payments

cfromknecht:
Builds on #3442 

This PR extends the `ListPayments`, `RouterRPC.SendPayment`, and `RouterRPC.TrackPayment` to include an experimental field for displaying HTLC attempts for a given payment. This will us to display multiple HTLCs when MPP is fully integrated. Legacy payments will have at most one attempt, and we simply convert them into a more generic `MPPayment` struct after loading from disk.

In the future we will perform a migration to convert the legacy on-disk format into real MPP payments on disk, but this allows us to fleshing out the API before the internal work is completed.


rpcserver.go
```diff
@@ -4248,6 +4248,40 @@ func (r *rpcServer) ListPayments(ctx context.Context,
...
+
+			default:
+				status = lnrpc.HTLCAttempt_IN_FLIGHT
+			}
+
+			htlcPath := make([]string, len(htlc.Route.Hops))
+			for i, hop := range htlc.Route.Hops {
+				htlcPath[i] = hex.EncodeToString(
+					hop.PubKeyBytes[:],
+				)
```
joostjager: Why not `repeated bytes` and format it in `lncli`?

channeldb/mp_payment.go
```diff
@@ -0,0 +1,137 @@
...
+	PaymentHash lntypes.Hash
+
+	// PaymentAddr is a receiver-generated nonce used to prevent concurrent
+	// payments from colliding.
+	PaymentAddr lntypes.Hash
+
+	// Value is the amount we are paying.
+	Value lnwire.MilliSatoshi
+
+	// CreatingDate is the at which this payment was started.
```
joostjager: typo: the at

channeldb/mp_payment.go
```diff
@@ -0,0 +1,137 @@
...
+	CreationDate time.Time
```
joostjager: `CreationTime` ?

channeldb/mp_payment.go
```diff
@@ -0,0 +1,137 @@
...
+func (p *MPPayment) PaymentPreimage() lntypes.Preimage {
+	for _, htlc := range p.HTLCs {
+		// Return the preimage if we find a successful HTLC.
+		if htlc.Settle != nil {
+			return htlc.Settle.Preimage
+		}
+	}
+
+	return lntypes.Preimage{}
+}
```
joostjager: For review, I'd like it more to also have code using this struct in this commit, prevent flipping back and forth between two commits. Personal pref. though

lnrpc/rpc.proto
```diff
@@ -2295,6 +2295,25 @@ message Payment {
...
+                FAILED = 2;
+        }
+
+        HTLCStatus status = 1 [json_name = "status"];
+
+        repeated string path = 2 [json_name = "path"];
+
+        int64 attempt_time = 3 [json_name = "attempt_time"];
+
+        int64 resolve_time = 4 [json_name = "resolve_time"];
```
joostjager: Same comment as before, for review I'd prefer having the code using this proto in this commit too.

channeldb/mp_payment.go
```diff
@@ -0,0 +1,137 @@
...
+		if htlc.Settle != nil {
+			return htlc.Route
+		}
+
+		// Otherwise stash this route which will be returned if their
+		// are no successful htlcs.
+		someRoute = htlc.Route
+	}
+	return someRoute
+}
```
joostjager: I think this code only exists to satisfy the payment-level route field for the rpc response. Maybe we should move this code in the rpc server, because we wouldn't want it to be called from anywhere else. The migrations changes will be rebased out when #3485 lands.

channeldb/payments.go
```diff
@@ -251,9 +251,84 @@ type Payment struct {
...
+func (p *Payment) PaymentPreimage() lntypes.Preimage {
+	var preimage lntypes.Preimage
+	if p.Preimage != nil {
+		preimage = *p.Preimage
+	}
+
+	return preimage
+}
+
+// ToMPPayment converts a legacy payment into an MPPayment.
```
joostjager: Confusion around the word legacy. In this context, it isn't a 'duplicate' payment is it?

lnrpc/rpc.proto
```diff
@@ -2295,6 +2295,25 @@ message Payment {
 
     ///  The fee paid for this payment in milli-satoshis
     int64 fee_msat = 12 [json_name = "fee_msat"];
+
+    /// The HTLCs made in attempt to settle the payment [EXPERIMENTAL].
+    repeated HTLCAttempt htlcs = 13[json_name = "htlcs"];
```
joostjager: Should we deprecate the payment level route field now? There are other fields like `value` that can be derived from the htlc list. That is just redundant. But the payment level route is not very well defined. Users may keep using it and then run into unpleasant surprises when they start doing mpps.

lnrpc/rpc.proto
```diff
@@ -2295,6 +2295,25 @@ message Payment {
...
+        HTLCStatus status = 1 [json_name = "status"];
+
+        repeated string path = 2 [json_name = "path"];
+
+        int64 attempt_time = 3 [json_name = "attempt_time"];
+
+        int64 resolve_time = 4 [json_name = "resolve_time"];
 }
 
 message ListPaymentsRequest {
```
joostjager: Maybe `routerrpc.PaymentStatus` should also be updated to include the htlc list. For users that use the reliable payments api in `routerrpc`, this may be the primary way to look up payments.

channeldb/payments.go
```diff
@@ -251,6 +251,29 @@ type Payment struct {
...
 
+// Route returns the route used to complete the payment, or the final route
+// attempted if the payment failed or is in-flight.
+func (p *Payment) Route() route.Route {
+	var rt route.Route
+	if p.Attempt != nil {
+		rt = p.Attempt.Route
+	}
+
+	return rt
```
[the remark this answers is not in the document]
cfromknecht: we could, tho this is meant to match the current behavior, where an empty route with no hops renders as an empty `"path"` via the rpc

channeldb/payments.go
cfromknecht: it's used for non-duplicate payments in this pr as well, but it's meant to refer to the legacy bucket structure. in the future, the non-duplicates will be migrated to the new mpp format, and then only the duplicates will continue using this conversion

lnrpc/rpc.proto
cfromknecht: I think it may be premature to deprecate now considering that the field is still tagged EXPERIMENTAL. It is somewhat ill-defined, but it is no worse than it exists today. The route field right now will show the route from the last attempt, which is also what is shown for an MPP payment

rpcserver.go
cfromknecht: now uses bytes and formatted in `lncli`. the changes spread quite a lot tho as you'll see. maybe it's time we fork `jsonpb` to do this for us instead of needing to manually do this workaround

channeldb/mp_payment.go
cfromknecht: Also added a `creation_time_ns` to the rpc response and deprecated `creation_date`

lnrpc/rpc.proto
cfromknecht: We can do that, tho I don't know if it's critical to this PR. It would require some significant refactoring of the way `PaymentResult` are propagated up, though that is part of the longer term plan

channeldb/payments.go
```diff
@@ -274,9 +274,61 @@ func (p *Payment) PaymentPreimage() lntypes.Preimage {
...
+		// NOTE: SettleTime is not set for legacy payments.
+		settle = &HTLCSettleInfo{
+			Preimage: *p.Preimage,
+		}
+	}
+
+	// Promote the payment failure to a proper fail struct, if it exists.
+	if p.Failure != nil {
+		// NOTE: FailTime is not set for legacy payments.
+		failure = &HTLCFailInfo{
```
halseth: panic/error if both failure and settle exists?

lnrpc/rpc.proto
```diff
@@ -2295,6 +2295,35 @@ message Payment {
...
+                IN_FLIGHT = 0;
+                SUCCEEDED = 1;
+                FAILED = 2;
+        }
+
+        /// The status of the HTLC.
+        HTLCStatus status = 1 [json_name = "status"];
+
+        /// The route taken by this HTLC.
+        repeated bytes path = 2 [json_name = "path"];
```
halseth: A question is whether we should take the opportunity to expand the information that is possible to get from the path field. If we instead exposed the full "route" here, a caller could get useful information such as fees etc.

cfromknecht: not a bad idea, we could reuse the existing route serialization?

channeldb/mp_payment.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Fail is a failure reason code indicating the reason the payment
+	// failed. It is only non-nil for failed payments.
+	//
+	// NOTE: Can be nil if payment is not failed.
+	Failure *HTLCFailInfo
+}
+
+// HTLCSettleInfo encapsulates the information that augments an HTLCAttempt int
+// he event that the HTLC is successful.
```
alexbosworth:
```suggestion
// the event that the HTLC is successful.
```

channeldb/mp_payment.go
```diff
@@ -0,0 +1,137 @@
...
+	Settle *HTLCSettleInfo
+
+	// Fail is a failure reason code indicating the reason the payment
+	// failed. It is only non-nil for failed payments.
+	//
+	// NOTE: Can be nil if payment is not failed.
+	Failure *HTLCFailInfo
+}
+
+// HTLCSettleInfo encapsulates the information that augments an HTLCAttempt int
```
alexbosworth:
```suggestion
// HTLCSettleInfo encapsulates the information that augments an HTLCAttempt in
```

lnrpc/rpc.proto
```diff
@@ -2295,6 +2295,35 @@ message Payment {
...
+
+        /// The time in UNIX nanonseconds at which this HTLC was sent.
+        int64 attempt_time_ns = 3 [json_name = "attempt_time_ns"];
+
+        /**
+        The time in UNIX nanonseconds at which this HTLC was settled or failed.
+        This value will / not be set if the HTLC is still IN_FLIGHT.
```
Roasbeef: Extraneous `/` here?

lnrpc/rpc.proto
Roasbeef: SGTM

cmd/lncli/types.go
```diff
@@ -66,3 +87,124 @@ func NewUtxoFromProto(utxo *lnrpc.Utxo) *Utxo {
...
+
+// ListPaymentsResponse contains a list of payments made by this node.
+type ListPaymentsResponse struct {
+	// Payments is the list of payments made.
+	Payments []Payment `json:"payments"`
+}
+
+// Payment encapsulates all information about a particular payment including all
+// HTLCs made in attempt to pay the recipient.
+type Payment struct {
```
Roasbeef: What more beyond this would've needed to have been added that inclined you towards the direction of forking the proto lib?

cfromknecht: Minimally `QueryRoutes` and build `BulidRoute` would have needed wrappers. Also the `Path` field on the `HTLCAttempt` will be replaced with `Route`, which itself will need a wrapper and then of course the internal `Hop` message, so four layers deep. Turned out to be way more code than it was worth for something that can be totally automated. Also helps to keep our `_types.go` files from expanding..

channeldb/payments.go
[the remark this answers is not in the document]
cfromknecht: done

lnrpc/rpc.proto
cfromknecht: done

lnrpc/rpc.proto
cfromknecht: extended `PaymentStatus` with htlc attempts

channeldb/payments.go
cfromknecht: done

cfromknecht:
> What about the PaymentStatus call in the router sub-server?

Last few commits now add the HTLC attempts to `PaymentStatus`

I rebased everything over #3436, which also allowed me test that the MPP record is being set properly in the new sendtoroute itests. This actually pointed out a bug in the test where we would send before receiving the node ann, causing us to downgrade to legacy payments for some of the htlcs. That bug has been fixed in #3436, and here we extend the itest with additional checks on the listpayments rpc.

.golangci.yml
```diff
@@ -30,6 +30,8 @@ linters-settings:
   gofmt:
     # simplify code: gofmt with `-s` option, true by default
     simplify: true
+  dupl:
+    threshold: 250
```
joostjager: Shouldn't a local `nolint` exception be added instead?

channeldb/mp_payment.go
```diff
@@ -0,0 +1,106 @@
...
+	// SettleTime is the time at which this HTLC was settled.
+	SettleTime time.Time
+}
+
+// HTLCFailInfo encapsulates the information that augments an HTLCAttempt in the
+// event that the HTLC fails.
+type HTLCFailInfo struct {
+	// Reason is the failure reason code indicating the reason the HTLC
+	// failed.
+	Reason FailureReason
```
joostjager: This is a payment-level failure enum, which differs from the htlc-level failures.

channeldb/payments.go
```diff
@@ -252,9 +252,74 @@ type Payment struct {
...
+			{
+				PaymentID:  p.Attempt.PaymentID,
+				SessionKey: p.Attempt.SessionKey,
+				Route:      p.Attempt.Route,
+				Settle:     settle,
+				Failure:    failure,
+			},
+		}
+	} else {
+		htlcs = []HTLCAttempt{
```
joostjager: Isn't it a possibility that the payment was never even attempted and that `p.Attempt` is therefore `nil`? In that case it doesn't seem right to add an `HTLCAttempt`

joostjager: I prefer `HtlcAttempt` because it doesn't stand out so much. To me "Htlc" is a word :)

rpcserver.go
```diff
@@ -4324,24 +4324,26 @@ func (r *rpcServer) ListPayments(ctx context.Context,
...
-			route = payment.Attempt.Route
+		// Fetch the payment's route and preimage. If no HTLC was
+		// successful, an empty route and preimage will be used.
+		var (
+			route    route.Route
+			preimage lntypes.Preimage
+		)
+		for _, htlc := range payment.HTLCs {
+			if htlc.Settle != nil {
+				route = htlc.Route
```
joostjager: Didn't we previously also expose the path for failed payments? Or was that only in `routerrpc`?

lnrpc/rpc.proto
```diff
@@ -2407,6 +2407,35 @@ message Payment {
...
+        Route route = 2 [json_name = "route"];
+
+        /// The time in UNIX nanonseconds at which this HTLC was sent.
+        int64 attempt_time_ns = 3 [json_name = "attempt_time_ns"];
+
+        /**
+        The time in UNIX nanonseconds at which this HTLC was settled or failed.
+        This value will not be set if the HTLC is still IN_FLIGHT.
+        */
+        int64 resolve_time_ns = 4 [json_name = "resolve_time_ns"];
```
joostjager: Should we update the main rpc at all? We encourage switching to `routerrpc` because it is safe. Only adding the fields there creates another reason to switch over.

routing/control_tower.go
```diff
@@ -107,43 +111,67 @@ func (p *controlTower) RegisterAttempt(paymentHash lntypes.Hash,
 func (p *controlTower) Success(paymentHash lntypes.Hash,
 	preimage lntypes.Preimage) error {
 
-	route, err := p.db.Success(paymentHash, preimage)
+	_, err := p.db.Success(paymentHash, preimage)
+	if err != nil {
+		return err
+	}
+
+	payment, err := p.db.FetchPayment(paymentHash)
```
joostjager: Can something happen in between these two calls?

lnrpc/rpc.proto
cfromknecht: what would they switch to? we only have one `listpayments` rpc

.golangci.yml
cfromknecht: didn't realize you could nolint a whole function, fixed

channeldb/mp_payment.go
cfromknecht: good call, moved to top-level `MPPayment`

channeldb/payments.go
cfromknecht: good catch, also modified the test to support `Fail` after registering so that we could test the case where the failure and attempt are both populated.

rpcserver.go
cfromknecht: fixed, we now show a route for failed+successful

routing/control_tower.go
cfromknecht: modified `Success` and `Fail` to return the `MPPayment` so that they are updated and read atomically

lnrpc/rpc.proto
joostjager: Maybe this is the time to copy it to the `routerrpc` ?

channeldb/payments.go
```diff
@@ -252,9 +252,70 @@ type Payment struct {
...
+	// Either a settle or a failure may be set, but not both.
+	if settle != nil && failure != nil {
+		panic("htlc attempt has both settle and failure info")
+	}
+
+	// Populate a single HTLC on the MPPayment if an attempt exists on the
+	// legacy payment. If none exists we will leave the attempt info blank
+	// since we cannot recover it.
+	if p.Attempt != nil {
+		// NOTE: AttemptTime is not set for legacy payments.
```
joostjager: The checks for `p.Preimage`  and `p.Failure`, can they be moved inside this if?

channeldb/mp_payment.go
```diff
@@ -0,0 +1,109 @@
...
+// Each populated HTLCAttempt represents an attempted HTLC, each of which may
+// have the associated Settle or Fail struct populated if the HTLC is no longer
+// in-flight.
+type MPPayment struct {
+	// sequenceNum is a unique identifier used to sort the payments in
+	// order of creation.
+	sequenceNum uint64
+
+	// Status is the current PaymentStatus of this payment.
+	Status PaymentStatus
```
joostjager: This `PaymentStatus` is mostly a product of the raw outcomes of the attempted htlcs for this payment. I am not sure how to deal with that. `FailureReasonTimeout` is not related to any of the htlcs, because it is a failure at the level of the payment loop itself. `FailureReasonError` can be several things. 

Do we also need to store a different type of state on the htlc level? If a new htlc is added for an mp payment, we need to rederive the top payment-level status. I am not sure, just feel that there is something here.

I don't mean reverting the change you just made. I am sure we need a payment-level status. But it may be useful to think through some multi-path scenarios and see how we'd update that payment-level status.

lnrpc/routerrpc/router_backend.go
```diff
@@ -780,3 +781,48 @@ func UnmarshalMPP(reqMPP *lnrpc.MPPRecord) (*record.MPP, error) {
...
+
+	var (
+		status      lnrpc.HTLCAttempt_HTLCStatus
+		resolveTime int64
+	)
+
+	switch {
+	case htlc.Settle != nil:
+		status = lnrpc.HTLCAttempt_SUCCEEDED
+		resolveTime = MarshalTimeNano(htlc.Settle.SettleTime)
```
joostjager: Also use a single `ResolveTime` field on `htlc`?

lnrpc/rpc.proto
cfromknecht: and have two rpcs?

joostjager: We need two at some point. The old one deprecated and the new one already available. Identical to the current SendPayment/SendToRoute situation.

lnrpc/routerrpc/router_backend.go
cfromknecht: i'm not sure what benefit it offers, we would still have two optional fields: `ResolveTime` and `Preimage`. keeping them isolated i think makes the logic simpler, o/w settle is `ResolveTime != nil && Preimage != nil`, and failed is `ResolveTime != nil && Preimage == nil`

channeldb/mp_payment.go
cfromknecht: yes at the moment this just moves the existing field to the top-level payment, and since we only have one htlc there shouldn't be any gaps.

as we move to real MPP, there are some additional states we would need to consider. do you think they should be addressed in this PR?

> we need to rederive the top payment-level status.

Yes my plan is that this field will be populated as a pure function of the htlc attempts and their individual states

joostjager: No, doesn't need to be in this PR. But won't there by a migration problem when you will do that pure function, but don't have the htlc outcomes?

lnrpc/routerrpc/router_backend.go
joostjager: Makes sense

channeldb/mp_payment.go
halseth: Currently we assume the payment is being actively tried if settle/fail is not present, and there's an associated attempt known by the switch. I think you are right that this will be a bit different, as there are both HTLC-level failures and payment level failures possible.  I don't think it needs to be it's own field though, as we can derive it from the HTLCs and FailureReason.

channeldb/payments.go
```diff
@@ -252,9 +252,70 @@ type Payment struct {
...
+		},
+		HTLCs: htlcs,
+	}
+}
+
 // FetchPayments returns all sent payments found in the DB.
-func (db *DB) FetchPayments() ([]*Payment, error) {
-	var payments []*Payment
+//
+// nolint: dupl
```
halseth: is this because of the migration code?

lnrpc/rpc.proto
```diff
@@ -2371,11 +2371,11 @@ message Payment {
...
     int64 value = 2 [json_name = "value", deprecated = true];
 
-    /// The date of this payment
-    int64 creation_date = 3 [json_name = "creation_date"];
+    /// Deprecated, use creation_time_ns
+    int64 creation_date = 3 [json_name = "creation_date", deprecated = true];
 
-    /// The path this payment took
-    repeated string path = 4 [ json_name = "path" ];
+    /// Deprecated, use path on individual htlcs
```
halseth: should we deprecate if the alternative is experimental?

rpcserver.go
```diff
@@ -4355,20 +4355,32 @@ func (r *rpcServer) ListPayments(ctx context.Context,
...
+		}
+
 		paymentHash := payment.Info.PaymentHash
 		paymentsResp.Payments = append(paymentsResp.Payments, &lnrpc.Payment{
 			PaymentHash:     hex.EncodeToString(paymentHash[:]),
 			Value:           satValue,
 			ValueMsat:       msatValue,
 			ValueSat:        satValue,
 			CreationDate:    payment.Info.CreationTime.Unix(),
+			CreationTimeNs:  payment.Info.CreationTime.UnixNano(),
```
halseth: marshalNano?

routing/control_tower.go
```diff
@@ -63,6 +63,10 @@ type PaymentResult struct {
 	// FailureReason is a failure reason code indicating the reason the
 	// payment failed. It is only set for failed payments.
 	FailureReason channeldb.FailureReason
+
+	// HTLCs is a list of HTLCs that have been attempted in order to settle
```
halseth: remove above Route field?

routing/control_tower.go
```diff
@@ -107,46 +111,61 @@ func (p *controlTower) RegisterAttempt(paymentHash lntypes.Hash,
...
 }
 
 // createSuccessResult creates a success result to send to subscribers.
-func createSuccessResult(rt *route.Route,
-	preimage lntypes.Preimage) *PaymentResult {
+func createSuccessResult(htlcs []channeldb.HTLCAttempt) *PaymentResult {
+	// Extract any preimage from the list of HTLCs.
+	var (
+		preimage lntypes.Preimage
+		route    *route.Route
```
halseth: could rather have this single route being extracted by the consumer of the PaymentResult

channeldb/mp_payment.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// SettleTime is the time at which this HTLC was settled.
+	SettleTime time.Time
+}
+
+// HTLCFailInfo encapsulates the information that augments an HTLCAttempt in the
+// event that the HTLC fails.
+type HTLCFailInfo struct {
+	// FailTime is the time at which this HTLC was failed.
+	FailTime time.Time
```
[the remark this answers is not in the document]
cfromknecht: we don't have any to include atm, there is only the payment level failures. for now this is only in memory so we have until we do the migration to decide if we want to embelish `HTLCFailInfo`

channeldb/payments.go
cfromknecht: yeah it picks up that there's a duplicated function in the migration package

lnrpc/rpc.proto
cfromknecht: touché, undeprecating

channeldb/payments.go
cfromknecht: turns out the preimage can, but failure needs to be outside. we have have a failure reason w/o an attempt

halseth: I think timeout can have happened even without an attempt. That being said, we must double check that legacy-legacy payment was given an attempt during migration. In any case I don't think we should assume it will always be present even if we have success/failure

cfromknecht: @halseth not sure I follow, does that change anything about my last comment?

routing/control_tower.go
cfromknecht: fixed in last commit!

rpcserver.go
cfromknecht: fixed

channeldb/payments.go
halseth: No, I was just attempting to agree with you 🤣  I guess maybe I was confused by "have have"

cfromknecht: ah okay sorry, fixed the have have 😂

lnrpc/rpc.proto
cfromknecht: what specifically about the current one can't be changed or fixed?

channeldb/mp_payment.go
joostjager: Discussed offline. If we want to make `SendToRoute` reliable too (left over from the reliable payment project), we need a way to retrieve the raw htlc failure with `TrackPayment`. It seems that we need to add it before we do the great payment db migration

lnrpc/rpc.proto
joostjager: Discussed offline. As there is virtual no cost to adding it to the main rpc, we can just do it. Need to think about when to mainline `routerrpc`. 0.9 seems too early

channeldb/mp_payment.go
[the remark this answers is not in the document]
halseth: Seems like we definitely have attempt level failures? Each HTLC is attempted along a given route, and if it fails, there is a reason to store?

channeldb/payments.go
```diff
@@ -252,9 +252,71 @@ type Payment struct {
...
+	}
+
+	// Populate a single HTLC on the MPPayment if an attempt exists on the
+	// legacy payment. If none exists we will leave the attempt info blank
+	// since we cannot recover it.
+	if p.Attempt != nil {
+		// Promote the payment preimage to proper settle struct, if it
+		// exists.
+		var settle *HTLCSettleInfo
+		if p.Preimage != nil {
```
halseth: I see no reason to have this inside the if though. Why not just handle it the same way as failure? (imagine sending a payment, but before you are able to make an attempt, somebody gives you the preimage and you can settle)

channeldb/payments.go
```diff
@@ -101,9 +101,9 @@ const (
 	// payment.
 	FailureReasonError FailureReason = 2
 
-	// FailureReasonIncorrectPaymentDetails indicates that either the hash
-	// is unknown or the final cltv delta or amount is incorrect.
-	FailureReasonIncorrectPaymentDetails FailureReason = 3
+	// FailureReasonPaymentDetails indicates that either the hash is unknown
```
halseth: too long for what? :P

channeldb/payment_control.go
```diff
@@ -192,16 +191,17 @@ func (p *PaymentControl) RegisterAttempt(paymentHash lntypes.Hash,
 // duplicate payments to the same payment hash. The provided preimage is
 // atomically saved to the DB for record keeping.
 func (p *PaymentControl) Success(paymentHash lntypes.Hash,
-	preimage lntypes.Preimage) (*route.Route, error) {
+	preimage lntypes.Preimage) (*MPPayment, error) {
```
halseth: nit: would be nice if these `channeldb` changes could get their own commit

lnrpc/routerrpc/router_server.go
```diff
@@ -605,15 +598,48 @@ func (s *Server) trackPayment(paymentHash lntypes.Hash,
...
-			if result.Route != nil {
-				status.Route, err = router.MarshallRoute(
-					result.Route,
-				)
-				if err != nil {
-					return err
-				}
+		}
+
+		// Extact the last route from the given list of HTLCs. This will
```
halseth: typo: extact

lnrpc/routerrpc/router_server.go
```diff
@@ -605,15 +598,48 @@ func (s *Server) trackPayment(paymentHash lntypes.Hash,
...
+		// be used the legacy route field for backwards compatibility.
```
halseth: "this will use"?

channeldb/mp_payment.go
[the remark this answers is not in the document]
halseth: Ok, I guess it would need us to start storing these. Sounds like work for a future PR indeed.

channeldb/payments.go
cfromknecht: for everything 😂but primarily as being an effective identifier, considering our mile wide indentation policy

lnrpc/rpc.proto
```diff
@@ -2465,6 +2465,35 @@ message Payment {
...
+    /// The HTLCs made in attempt to settle the payment [EXPERIMENTAL].
+    repeated HTLCAttempt htlcs = 14 [json_name = "htlcs"];
+}
+
+message HTLCAttempt {
+        enum HTLCStatus {
+                IN_FLIGHT = 0;
+                SUCCEEDED = 1;
+                FAILED = 2;
+        }
```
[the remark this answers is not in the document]
cfromknecht: it shouldn't affect the enum, i think we'd just add some new field types to hold the failure info

channeldb/payments.go
cfromknecht: reverted to prior structure

channeldb/mp_payment.go
```diff
@@ -0,0 +1,109 @@
...
+
+// MPPaymentCreationInfo is the information necessary to have ready when
+// initiating a payment, moving it into state InFlight.
+type MPPaymentCreationInfo struct {
+	// PaymentHash is the hash this payment is paying to.
+	PaymentHash lntypes.Hash
+
+	// PaymentAddr is a receiver-generated nonce used to prevent concurrent
+	// payments from colliding.
+	PaymentAddr lntypes.Hash
```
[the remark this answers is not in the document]
cfromknecht: done

routing/control_tower_test.go
```diff
@@ -223,9 +246,27 @@ func TestPaymentControlSubscribeFail(t *testing.T) {
...
 		}
-		if result.Route != nil {
-			t.Fatal("expected no route")
+
+		// The number of htlcs will depend on whether or not we
+		// registered this payment attempt.
+		var expHtlcs int
+		if registerAttempt {
+			expHtlcs = 1
+		}
```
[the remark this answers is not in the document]
cfromknecht: done

channeldb/payment_control.go
cfromknecht: i attempted to do so originally, but it ended up causing more code shuffling than it was worth. the extraction code ends up being duplicated and move around, the tests end up double testing various aspects, etc over 4 commits. so i squashed it all into one since it's i found it easier to just look at the whole change

channeldb/mp_payment.go
```diff
@@ -0,0 +1,105 @@
...
+	PaymentID uint64
+
+	// SessionKey is the ephemeral key used for this payment attempt.
+	SessionKey *btcec.PrivateKey
+
+	// Route is the route attempted to send the HTLC.
+	Route route.Route
+
+	// AttemptTime is the time at which this HTLC was attempted.
+	AttemptTime time.Time
```
joostjager: This field doesn't seem to be set at all in this pr. Remove and introduce in the follow up?

PR #3505, opened 2019-09-14
zpay32: check route+hop hints while decoding

Crypt-iQ:
This commit checks that there are not greater than 20 route hints while decoding or more than 20 hop hints for a single route hint. This mitigates a potential DoS vector where an attacker could craft a very large bech32 invoice string containing an absurd amount of route and/or hop hints. If sent to an application that processes payment requests, this would allocate a burdensome amount of memory due to the public key parsing for each route/hop hint.

For a 1.7MB payment request, this yielded about 38MB in allocations from just parsing public keys (`bech32VerifyChecksum` allocated another 25MB):
```
   45.51MB  7.31% 92.07%    45.51MB  7.31%  math/big.nat.make
   25.50MB  4.09% 96.16%    25.50MB  4.09%  github.com/lightningnetwork/lnd/zpay32.bech32VerifyChecksum
       1MB  0.16% 96.32%    39.50MB  6.34%  github.com/lightningnetwork/lnd/zpay32.parseRouteHint
       1MB  0.16% 96.48%    33.50MB  5.38%  github.com/btcsuite/btcd/btcec.decompressPoint
    0.50MB  0.08% 96.56%     7.50MB  1.20%  crypto/elliptic.(*CurveParams).doubleJacobian
    0.50MB  0.08% 96.64%       38MB  6.10%  github.com/btcsuite/btcd/btcec.ParsePubKey
         0     0% 96.64%       12MB  1.93%  crypto/ecdsa.Verify
         0     0% 96.64%        8MB  1.28%  crypto/elliptic.(*CurveParams).ScalarBaseMult
         0     0% 96.64%       12MB  1.93%  crypto/elliptic.(*CurveParams).ScalarMult
```
With this change, memory usage will be far lower as decoding will exit early with an error if too many route/hop hints are found.

EDIT:
With https://github.com/btcsuite/btcd/pull/1429, the memory allocation from parsing pubkeys occupies about ~6.5MB which is significant improvement.  The `bech32VerifyChecksum` still allocates a ridiculous amount of memory.  Another fix would be to limit the size of invoices before attempting any decoding.  In conjunction with the commits in this PR, it should limit any potential sources of memory blow-up.


zpay32/invoice.go
```diff
@@ -668,6 +668,12 @@ func parseTaggedFields(invoice *Invoice, fields []byte, net *chaincfg.Params) er
 
 			invoice.FallbackAddr, err = parseFallbackAddr(base32Data, net)
 		case fieldTypeR:
+			// Check that we don't already have 20 route hints. Otherwise this
+			// represents a potential DoS vector because of pubkey parsing.
+			if len(invoice.RouteHints) == 20 {
```
wpaulino: We use this constant in a couple places now, so we should extract it into a constant in this package.

wpaulino: Would be nice to add a test for these cases as well.

Crypt-iQ:
I'd also like to include a length limit on the invoices that'll be checked before starting the decoding process.  Ideas @wpaulino ?

I'm by no means in expert in QR codes or anything, but from here https://en.wikipedia.org/wiki/QR_code#Storage it seems that `len(invoice) < 7089` for it to fit into a QR code.  I think this might even be implementation depending on the app used.  However, if each field can have a maximum of `1024` bytes and there can be 20 route hints, this would give 20480 bytes, not counting the other data in the invoice.  With the current code, it's legal for an invoice to not fit into a QR code which I think was one of the points of bolt 11?  

Another reason to limit the length besides the QR code argument is that you can also make invoices huge by including unknown fields that'll be skipped over while decoding.  So limiting route hints doesn't solve the memory blow up issues entirely.

zpay32/invoice.go
Crypt-iQ: Added

zpay32/invoice.go
```diff
@@ -71,12 +72,31 @@ const (
 	// fieldType9 contains one or more bytes for signaling features
 	// supported or required by the receiver.
 	fieldType9 = 5
+
+	// maxRouteHints is the maximum number of route hints allowed in a single
+	// invoice.
+	maxRouteHints = 20
```
halseth: 20 seems like a reasonable choice. Any reason in particular? :)

Crypt-iQ: Was the value previously used.  Currently the codebase only generates one hop per route hint so any invoice that lnd generates should fit into a qr code with 20.  However, if we stick with 20 here, it's possible that invoices could be larger-than-fits-in-qr-code while decoding.  Could lower `maxHopHints` since 12 hops for a route hint might be overkill or could lower `maxRouteHints` if 20 route hints is overkill.  Not sure

halseth: > Was the value previously used. 
When LND creates invoices you mean? Could be useful share this constant with that code.

halseth: I think as long as we keep a max total invoice length, maybe the other restrictions aren't even needed?

Crypt-iQ: Yup agree

> > Was the value previously used.
> 
> When LND creates invoices you mean? Could be useful share this constant with that code.

Yeah both when creating invoices + decoding invoices, but can just check the length now

Crypt-iQ: Done, now check that invoice is not greater than 7092 bytes.  Removed `maxRouteHints` but kept `maxHopHints` as it may be useful for callers of `Encode` to know why a specific invoice can't be serialized.

zpay32/invoice.go
```diff
@@ -71,12 +72,29 @@ const (
...
 	fieldType9 = 5
+
+	// maxHopHints is the maximum number of hop hints allowed in a route hint.
+	// This number is derived by noting that the length of the hop hints must
+	// fit into a base32 encoding of length 2:
+	//
+	// ceil(maxHopHints * hopHintLen * 8/5) < 2^10
+	maxHopHints = 12
+
+	// maxInvoiceLength is the maximum total length an invoice can have.
```
halseth: should add to comment why this constant is chosen :)

zpay32/invoice_test.go
```diff
@@ -811,6 +811,84 @@ func TestNewInvoice(t *testing.T) {
...
+		},
+		{
+			// Invalid since it is greater than maxInvoiceLength. Also not a valid bech32 string.
```
halseth: why not a valid bech32 string? The proper test would be to attempt to decode a valid bech32 string that's too large I guess.

Crypt-iQ: I figured that it would fail the length check anyway since that is before the bech32 correctness checks, but you're right.  Added correct bech32 string.

zpay32/invoice.go
wpaulino: Ah I see, every tagged field has a limit of 10 bits, but we don't seem to perform this check for any of them except `r` as of this PR.

halseth: I think we check that all data will fit within 10 bits when encoding the invoice, so that should be enough for those cases?

I'm kinda wondering if we can just skip the `maxHopHints` check as well since we'll get the error when trying to fit within 10 bits 😛 What do you think @Crypt-iQ?

Crypt-iQ: Yeah I'm cool w that

PR #3508, opened 2019-09-15
peer: buffer the outgoingQueue channel

Crypt-iQ:
Small fix, the `outgoingQueue` chan wasn't buffered.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -247,7 +247,7 @@ func newPeer(conn net.Conn, connReq *connmgr.ConnReq, server *server,
 		outgoingCltvRejectDelta: outgoingCltvRejectDelta,
 
 		sendQueue:     make(chan outgoingMsg),
-		outgoingQueue: make(chan outgoingMsg),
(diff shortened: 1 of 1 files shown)


peer.go
```diff
@@ -247,7 +247,7 @@ func newPeer(conn net.Conn, connReq *connmgr.ConnReq, server *server,
 		outgoingCltvRejectDelta: outgoingCltvRejectDelta,
 
 		sendQueue:     make(chan outgoingMsg),
-		outgoingQueue: make(chan outgoingMsg),
+		outgoingQueue: make(chan outgoingMsg, outgoingQueueLen),
```
cfromknecht: does this address an existing issue? atm the `queue` method blocks until the message is actually being assigned it's final outgoing order, which i think we may want to keep. if there isn't an issue, then maybe remove `outgoingQueueLen` instead?

Crypt-iQ: My rationale was that this would give a speedup when asynchronously sending messages which happens in a couple places in the code.  I guess the question is if blocking until order assignment is desirable?

PR #3509, opened 2019-09-16
lnwallet: fix goroutines in tests

alrs:
This PR fixes a test in `lnwallet` where some of the error conditions could never be recorded, because they were happening inside a goroutine. 

`T.Fatal()` is a convenience function that combines `T.Log()` and `T.FailNow()`. 

From the godoc of `T.FailNow()`: 

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.

To see this in action, take a look at `main_test.go` in my [example repo](https://github.com/alrs/fatalfail/).

### DIFF
--- a/lnwallet/interface_test.go
+++ b/lnwallet/interface_test.go
@@ -1269,7 +1269,7 @@ func testTransactionSubscriptions(miner *rpctest.Harness,
 		outputAmt = btcutil.SatoshiPerBitcoin
 		numTxns   = 3
 	)
-	unconfirmedNtfns := make(chan struct{})
+	errCh1 := make(chan error, 1)
 	switch alice.BackEnd() {
 	case "neutrino":
 		// Neutrino doesn't listen for unconfirmed transactions.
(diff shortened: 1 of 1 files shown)


lnwallet/interface_test.go
```diff
@@ -1333,24 +1334,32 @@ func testTransactionSubscriptions(miner *rpctest.Harness,
...
+			if err != nil {
+				t.Fatal(err)
+			}
 		}
 	}
 
-	confirmedNtfns := make(chan struct{})
+	errCh2 := make(chan error)
 	go func() {
+		defer close(errCh2)
```
halseth: don't have to close channel when sending error below.

lnwallet/interface_test.go
```diff
@@ -1269,32 +1269,33 @@ func testTransactionSubscriptions(miner *rpctest.Harness,
...
 		numTxns   = 3
 	)
-	unconfirmedNtfns := make(chan struct{})
+	errCh1 := make(chan error)
 	switch alice.BackEnd() {
 	case "neutrino":
 		// Neutrino doesn't listen for unconfirmed transactions.
 	default:
 		go func() {
+			defer close(errCh1)
```
halseth: should use the same pattern both places. I would not close the channel, but instead explicitly send `nil` to indicate success.

lnwallet/interface_test.go
```diff
@@ -1269,32 +1269,33 @@ func testTransactionSubscriptions(miner *rpctest.Harness,
 		outputAmt = btcutil.SatoshiPerBitcoin
 		numTxns   = 3
 	)
-	unconfirmedNtfns := make(chan struct{})
+	errCh1 := make(chan error)
```
halseth: buffer with 1

lnwallet/interface_test.go
```diff
@@ -1333,24 +1334,32 @@ func testTransactionSubscriptions(miner *rpctest.Harness,
...
-		case <-unconfirmedNtfns: // Fall through on success
+		case err := <-errCh1:
+			if err != nil {
+				t.Fatal(err)
+			}
 		}
 	}
 
-	confirmedNtfns := make(chan struct{})
+	errCh2 := make(chan error)
```
halseth: buffer with 1

lnwallet/interface_test.go
```diff
@@ -1333,24 +1334,32 @@ func testTransactionSubscriptions(miner *rpctest.Harness,
...
 	go func() {
+		defer close(errCh2)
 		for i := 0; i < numTxns; i++ {
 			txDetail := <-txClient.ConfirmedTransactions()
 			if txDetail.NumConfirmations != 1 {
-				t.Fatalf("incorrect number of confs for %s, expected %v got %v",
+				err := fmt.Errorf("incorrect number of confs for %s, expected %v got %v",
```
halseth: nit: just send error directly on channel without creating variable

lnwallet/interface_test.go
alrs: OK.

lnwallet/interface_test.go
alrs: OK.

lnwallet/interface_test.go
alrs: OK.

PR #3510, opened 2019-09-16
Handle RBF signaling publication failures in wallet

halseth:
Earlier we would not return `ErrDoubleSpend` if a transaction failed to publish because there was an existing, RBF-signaling, transaction in the mempool. This could happen in cases where we attempted to force close the channel, but a different closing transaction was already broadcasted.

Now we'll properly now that this error is not critical, such that we can continue resolving the closing channel.

Depends on https://github.com/btcsuite/btcwallet/pull/647

Needed for #3016

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/btcsuite/btcd v0.0.0-20190824003749-130ea5bddde3
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
-	github.com/btcsuite/btcwallet v0.0.0-20190911065739-d5cdeb4b91b0
+	github.com/btcsuite/btcwallet v0.0.0-20190925005052-95d7aa0b4953
 	github.com/btcsuite/btcwallet/wallet/txauthor v1.0.0
 	github.com/btcsuite/btcwallet/wallet/txrules v1.0.0
 	github.com/btcsuite/btcwallet/walletdb v1.0.0
--- a/lnwallet/btcwallet/btcwallet.go
+++ b/lnwallet/btcwallet/btcwallet.go
@@ -5,7 +5,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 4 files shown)


lnwallet/interface_test.go
```diff
@@ -1484,13 +1484,20 @@ func txFromOutput(tx *wire.MsgTx, signer input.Signer, fromPubKey,
 	// referenced output.
 	tx1 := wire.NewMsgTx(2)
 
+	// If we want to create a tx that signals replacement, set its
+	// sequence number to the max one that signals replacement.
+	// Otherwise we just use the standard max sequence.
+	sequence := wire.MaxTxInSequenceNum
+	if rbf {
+		sequence = wire.MaxTxInSequenceNum - 2
```
wpaulino: Nit: can use `mempool.MaxRBFSequence` instead.

lnwallet/btcwallet/btcwallet.go
```diff
@@ -431,42 +431,61 @@ func (b *BtcWallet) PublishTransaction(tx *wire.MsgTx) error {
...
 
+			// RBF enabled transaction was rejected for whatever
+			// reason.
+			if strings.Contains(err.Error(), "replacement transaction") {
+				return lnwallet.ErrDoubleSpend
+			}
+
+			// TODO(halseth): neutrino errors depend on connected
+			// peers. Makes more sense to make it match on a
+			// combination of the btcd and bitcoind errors?
```
wpaulino: Yep, this is what's done in the btcwallet and neutrino repos. I would add them as part of this PR.

lnwallet/btcwallet/btcwallet.go
```diff
@@ -431,42 +431,61 @@ func (b *BtcWallet) PublishTransaction(tx *wire.MsgTx) error {
...
+			// missing or already spent.
 			if strings.Contains(err.Error(), "orphan transaction") {
-				// Transaction is spending either output that
-				// is missing or already spent.
+				return lnwallet.ErrDoubleSpend
+			}
+
+			// RBF enabled transaction was rejected for whatever
+			// reason.
+			if strings.Contains(err.Error(), "replacement transaction") {
```
wpaulino: Heh, good thing the errors returned always include this, makes string matching a bit simpler.

lnwallet/btcwallet/btcwallet.go
```diff
@@ -431,42 +431,61 @@ func (b *BtcWallet) PublishTransaction(tx *wire.MsgTx) error {
...
+			// Output was already spent.
 			if strings.Contains(err.Error(), "already spent") {
-				// Output was already spent.
 				return lnwallet.ErrDoubleSpend
 			}
 
+			// RBF enabled transaction was rejected for whatever
+			// reason.
+			if strings.Contains(err.Error(), "replacement transaction") {
+				return lnwallet.ErrDoubleSpend
```
wpaulino: We might want to end up returning a different error in the future for the purpose of detecting an insufficient fee bump, but can likely be deferred.

lnwallet/btcwallet/btcwallet.go
halseth: Yeah, it isn't very scalable to keep maintaining this error handling both places. Should we instead define a set of errors in btcwallet (like ErrDoubleSpend), then move all error matching there?

Roasbeef: Yeah for a while we've been wanting to shove all the error string matching into `btcwallet` itself, then define a set of errors on the `WalletController` interface level. This way, in `lnd` we just use `Is/As` to check for errors and we can maintain all our error matching in a single place (as there're a few areas that we duplicate this type of code between `lnd` and `btcwallet`).

lnwallet/btcwallet/btcwallet.go
Roasbeef: Will this error actually contain `replacement`? Shouldn't we instead check for a reject code here?

go.mod
```diff
@@ -58,6 +58,8 @@ replace github.com/lightningnetwork/lnd/ticker => ./ticker
 
 replace github.com/lightningnetwork/lnd/queue => ./queue
 
+replace github.com/btcsuite/btcwallet => github.com/halseth/btcwallet v0.0.0-20190919123305-ca06abc72b9c
```
halseth: Note to self: remove when btcwallet PR is merged.

lnwallet/interface_test.go
halseth: Done

lnwallet/btcwallet/btcwallet.go
halseth: Made btcwallet PR: https://github.com/btcsuite/btcwallet/pull/647

And first use of `errors.As` 😬

lnwallet/btcwallet/btcwallet.go
halseth: Hard to check for error code, as both `RejectInvalid`, `RejectInsufficientFee` and `RejectNonstandard` are used for these.

See https://github.com/btcsuite/btcwallet/pull/647/commits/ca06abc72b9cb7ceb1d4531585cf26aeac1560b7

lnwallet/btcwallet/btcwallet.go
```diff
@@ -429,47 +429,29 @@ func (b *BtcWallet) ListUnspentWitness(minConfs, maxConfs int32) (
...
 
-		case *chain.NeutrinoClient:
-			if strings.Contains(err.Error(), "already spent") {
-				// Output was already spent.
-				return lnwallet.ErrDoubleSpend
-			}
+		switch {
+		// If the wallet reports a double spend, convert it to our
+		// internal ErrDoubleSpend and return.
+		case errors.As(err, &doubleSpendErr):
```
Roasbeef: I think we can just get by with using `Is` here since we're not actually using any of the fields in the populated error (the double (?) pointer receiver). However, we may need to do a type assertion if we want to allow things to be built with go 1.12. Alternatively, I _think_ we can still use the `xerrors` package.

Roasbeef: Very nice clean up here though, something we've been wanting to do for some time now.

halseth: Changed to regular type switch, to support go 1.12: https://github.com/lightningnetwork/lnd/pull/3510/commits/f1b5809b0a589ee9888fb194603275bafe50233e

halseth:
Reveiw addressed. PTAL @Roasbeef @wpaulino

halseth:
`btcwallet` PR merged, and dependency updated.

PR #3513, opened 2019-09-17
[trivial] lint: skip generated files

halseth:
Would fail on files created by mobile build tools, since they redeclare certain method names.


.golangci.yml
```diff
@@ -6,6 +6,9 @@ run:
   # worker.
   concurrency: 1
 
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  deadline: 5m
```
wpaulino: What's the rationale for this change? Would be nice to include it in the commit message.

halseth: just to try to make travis pass

halseth: Removed

halseth:
Travis should be fixed now. Also added verbose mode to the linter to get more insight into what it is doing.

PR #3514, opened 2019-09-17
[trivial] make: change dir for btcd

halseth:
Not changing dir leads to the go.* files being modified when calling
make btcd (and every other command that requires btcd).

In addition, seems like we must change dir and install in the same
command for the dir change to actually be effective (macOS).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,10 @@ GOBUILD := GO111MODULE=on go build -v -trimpath
 GOINSTALL := GO111MODULE=on go install -v -trimpath
 GOTEST := GO111MODULE=on go test -v
 
+# Switch to tmp directory to prevent lnd go.mod from being modified.
+# Hopefully someday a flag for go get will be added for this.
+GOGET := cd /tmp; GO111MODULE=on go get -u -v
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -73,8 +73,7 @@ $(LINT_BIN):
 	@$(call print, "Fetching linter")
 	# Switch to tmp directory to prevent lnd go.mod from being modified.
 	# Hopefully someday a flag for go get will be added for this.
-	cd /tmp
-	GO111MODULE=on go get $(LINT_PKG)
+	cd /tmp && GO111MODULE=on go get $(LINT_PKG)
```
wpaulino: Perhaps use `pushd` and `popd` instead so that we don't stay within `/tmp`?

cfromknecht: should probably make a command for this since we could use it in a number of places

halseth: Added a command to do this for all deps. Note that each makefile line is done in a new shell, so no need to change back.

cfromknecht:
heh i didn't realize this ended up looking so close to #3521

halseth:
Replaced by https://github.com/lightningnetwork/lnd/pull/3521

PR #3518, opened 2019-09-18
[trivial] make: default backend variable to btcd

halseth:
This fixes the logging of used backend in the default (btcd) case.

Otherwise the `${backend}` variable would be left unset.

### DIFF
--- a/make/testing_flags.mk
+++ b/make/testing_flags.mk
@@ -55,14 +55,12 @@ UNIT_RACE := $(UNIT) -race
 endif
 
 
-# Construct the integration test command with the added build flags.
-ITEST_TAGS := $(DEV_TAGS) rpctest chainrpc walletrpc signrpc invoicesrpc autopilotrpc routerrpc watchtowerrpc wtclientrpc
-
 # Default to btcd backend if not set.
-ifneq ($(backend),)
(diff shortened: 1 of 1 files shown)


make/testing_flags.mk
```diff
@@ -55,14 +55,12 @@ UNIT_RACE := $(UNIT) -race
...
-# Construct the integration test command with the added build flags.
-ITEST_TAGS := $(DEV_TAGS) rpctest chainrpc walletrpc signrpc invoicesrpc autopilotrpc routerrpc watchtowerrpc wtclientrpc
-
 # Default to btcd backend if not set.
-ifneq ($(backend),)
-ITEST_TAGS += ${backend}
-else
-ITEST_TAGS += btcd
+ifeq ($(backend),)
+backend = btcd
```
cfromknecht: so far we've been using all capital letters to indicate that the variables are defined within the makefile, and lower case to represent arguments parsed from the command line. i think we want to keep the same pattern here, so that we know command line args aren't being modified elsewhere in the makefile

halseth: This is actually an argument passed from the command line, that we just set in case it hasn't been specified.

PR #3520, opened 2019-09-18
[autopilot] Trigger channel opening on external score update 

halseth:
This PR adds a new signal to the autopilot agent, meant to signal
when any of the available heuristics has gotten an update.

We currently use this to trigger a new channel opening after the
external scores have been updated.

### DIFF
--- a/autopilot/agent.go
+++ b/autopilot/agent.go
@@ -541,10 +541,28 @@ func (a *Agent) openChans(availableFunds btcutil.Amount, numChans uint32,
 	connectedNodes := a.chanState.ConnectedNodes()
 	a.chanStateMtx.Unlock()
 
+	for nID := range connectedNodes {
+		log.Tracef("Skipping node %x with open channel", nID[:])
+	}
+
 	a.pendingMtx.Lock()
+
+	for nID := range a.pendingOpens {
+		log.Tracef("Skipping node %x with pending channel open", nID[:])
+	}
+
+	for nID := range a.pendingConns {
+		log.Tracef("Skipping node %x with pending connection", nID[:])
+	}
+
+	for nID := range a.failedNodes {
+		log.Tracef("Skipping failed node %v", nID[:])
+	}
+
 	nodesToSkip := mergeNodeMaps(a.pendingOpens,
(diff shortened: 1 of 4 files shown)


autopilot/manager.go
```diff
@@ -319,8 +319,12 @@ func (m *Manager) queryHeuristics(nodes map[NodeID]struct{}, localState bool) (
 
 	// We'll start by getting the scores from each available sub-heuristic,
 	// in addition the current agent heuristic.
+	var heuristics []AttachmentHeuristic
```
Roasbeef: In the commit message: instaed -> instead

autopilot/agent.go
```diff
@@ -256,6 +261,13 @@ type chanPendingOpenUpdate struct{}
 // a previous channel open failed, and that it might be possible to try again.
 type chanOpenFailureUpdate struct{}
 
+// heuristicUpdate is an update sent when one of the autopilot heuristics has
+// changed, and prompts the agent to make a new attempt at opening more
+// channels.
+type heuristicUpdate struct {
+	heuristic AttachmentHeuristic
+}
```
carlaKC: Are we expecting more fields to be added to this struct other than a `AttachmentHeuristic`? 
I'm not clear on the purpose of this struct if it just wraps a single heuristic?

halseth: Good question! It could definitely work without the extra struct, it's just there to mimic the other event types we have here, so felt like we should use the same pattern.

carlaKC: Ok cool, LGTM otherwise 😄

autopilot/manager.go
halseth: Fixed.

autopilot/manager.go
```diff
@@ -372,5 +379,11 @@ func (m *Manager) SetNodeScores(name string, scores map[NodeID]float64) error {
 		return fmt.Errorf("heuristic with name %v not found", name)
 	}
 
+	// If the autopilot agent is active, notify about the updated
+	// heuristic.
+	if m.pilot != nil {
```
carlaKC: `m.IsActive()` here?

PR #3523, opened 2019-09-19
multi: enable max htlc update

joostjager:
In this PR, we enable users to update the max HTLC for a given channel or channels.

Rewrite of #2438


routing/localchans/manager.go
```diff
@@ -59,6 +60,13 @@ func (r *Manager) UpdatePolicy(newSchema routing.ChannelPolicy,
 		info *channeldb.ChannelEdgeInfo,
 		edge *channeldb.ChannelEdgePolicy) error {
 
+		// Because gossiper is migrating policies without max htlc on
+		// startup, the MaxHtlc flag should always be set at this point.
+		if !edge.MessageFlags.HasMaxHtlc() {
+			return fmt.Errorf("channel %v has no max htlc",
+				info.ChannelPoint)
+		}
```
joostjager: Not sure about this sanity check. I actually don't like sanity checks.

discovery/gossiper.go
```diff
@@ -1320,90 +1335,35 @@ func (d *AuthenticatedGossiper) retransmitStaleAnns(now time.Time) error {
 //
 // TODO(roasbeef): generalize into generic for any channel update
 func (d *AuthenticatedGossiper) processChanPolicyUpdate(
-	policyUpdate *chanPolicyUpdateRequest) (
+	edgesToUpdate []EdgeWithInfo) (
```
halseth: comment must be updated

routing/localchans/manager.go
halseth: This assumes an order that the gossiper always runs before this is called, so I think it is safest to keep the check?

discovery/gossiper.go
```diff
@@ -1211,6 +1218,14 @@ func (d *AuthenticatedGossiper) retransmitStaleAnns(now time.Time) error {
 		// introduction of the MaxHTLC field, then we'll update this
 		// edge to propagate this information in the network.
 		if !edge.MessageFlags.HasMaxHtlc() {
+			// We'll make sure we support the new max_htlc field if
```
halseth: maybe we should move this check/update to the local manager instead? Making gossiper as unaware of local changes as possible,

joostjager: I wanted to keep the change set small. Moving retransmission of stale channels to the local chan mgr is a much larger change. Also not sure if that is the place where it belongs.

discovery/gossiper.go
```diff
@@ -1218,6 +1218,14 @@ func (d *AuthenticatedGossiper) retransmitStaleAnns(now time.Time) error {
 		// introduction of the MaxHTLC field, then we'll update this
 		// edge to propagate this information in the network.
 		if !edge.MessageFlags.HasMaxHtlc() {
```
joostjager: It never gets here for private channels, so the migration isn't always performed at this point.

discovery/gossiper.go
joostjager: updated

discovery/gossiper.go
joostjager: Removed migration changes from the pr

routing/localchans/manager.go
joostjager: Removed those changes from the pr. It is more complicated, because the gossiper migrates the updates in the background.

discovery/gossiper.go
```diff
@@ -922,9 +926,10 @@ func (d *AuthenticatedGossiper) networkHandler() {
 			// First, we'll now create new fully signed updates for
 			// the affected channels and also update the underlying
 			// graph with the new state.
-			chanPolicies, newChanUpdates, err := d.processChanPolicyUpdate(
-				policyUpdate,
-			)
+			chanPolicies, newChanUpdates, err :=
```
Roasbeef: Style nit: imo overflowing on one line is better than making a break and created a "dangling" function invocation.

discovery/gossiper.go
```diff
@@ -361,23 +359,29 @@ type updatedChanPolicies struct {
...
+
+// PropagateChanPolicyUpdate signals the AuthenticatedGossiper to perform the
+// specified edge updates. Updates are done in two stages: first, the
 // AuthenticatedGossiper ensures the update has been committed by dependent
 // sub-systems, then it signs and broadcasts new updates to the network. A
 // mapping between outpoints and updated channel policies is returned, which is
 // used to update the forwarding policies of the underlying links.
 func (d *AuthenticatedGossiper) PropagateChanPolicyUpdate(
-	newSchema routing.ChannelPolicy, chanPoints ...wire.OutPoint) (
+	edgesToUpdate []EdgeWithInfo) (
```
Roasbeef: Is this return value still needed if we now pass in the channel policy to be broadcast? Whereas before we used this to update the switch with what was returned by the gossiper.

routing/localchans/manager.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/discovery"
+	"github.com/lightningnetwork/lnd/htlcswitch"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing"
+)
+
+// Manager manages the node's local channels. The only operation that is
+// currently implemented is updating forwarding policies.
+type Manager struct {
```
Roasbeef: Nice set up for future changes in this area which may have automated updates triggered due to fees on chain, off-chain fees, time locks, etc.

routing/localchans/manager.go
```diff
@@ -0,0 +1,100 @@
...
+	// HtlcSwitch is used by the manager to update active links with a new
+	// policy.
+	HtlcSwitch *htlcswitch.Switch
+
+	// AuthGossiper is called to persist a new policy to disk and broadcast
+	// it to the network.
+	AuthGossiper *discovery.AuthenticatedGossiper
+
+	// Router is required to iterate over all our local channels.
+	Router *routing.ChannelRouter
```
Roasbeef: Not a blocker, but we could use either a slimmer interface here, or a function closure to avoid depending on the entire router struct.

routing/localchans/manager.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// policyUpdateLock ensures that the database and the link do not fall
+	// out of sync if there are concurrent fee update calls. Without it,
+	// there is a chance that policy A updates the database, then policy B
+	// updates the database, then policy B updates the link, then policy A
+	// updates the link.
+	policyUpdateLock sync.Mutex
```
Roasbeef: 👍

discovery/gossiper_test.go
```diff
@@ -3602,12 +3602,15 @@ out:
 	// Now that all of our channels are loaded, we'll attempt to update the
 	// policy of all of them.
 	const newTimeLockDelta = 100
+	const newMaxHtlc = (1 << 63) - 1
```
Roasbeef: Shouldn't this be rejected by our logic since it's greater than the size of the channels themselves?

routing/localchans/manager.go
```diff
@@ -74,6 +74,12 @@ func (r *Manager) UpdatePolicy(newSchema routing.ChannelPolicy,
 		)
 		edge.TimeLockDelta = uint16(newSchema.TimeLockDelta)
 
+		// MaxHTLC is optional in channel policy update requests.
+		if newSchema.MaxHTLC != 0 {
```
Roasbeef: Shouldn't this optionality logic be applied for all the other fields? Otherwise, we'll modify the behavior for users which are using the existing API as is.

routing/localchans/manager.go
```diff
@@ -14,6 +15,9 @@ import (
 // Manager manages the node's local channels. The only operation that is
 // currently implemented is updating forwarding policies.
 type Manager struct {
+	// Db is used to query local channel parameters
+	Db *channeldb.DB
```
Roasbeef: Similar comment here re slimmer interface or a function closure as it'll make this logic easier to test in the future.

routing/localchans/manager.go
```diff
@@ -104,3 +115,34 @@ func (r *Manager) UpdatePolicy(newSchema routing.ChannelPolicy,
...
+	}
+
+	if maxHtlcRequested < ch.LocalChanCfg.MinHTLC {
+		return fmt.Errorf("max htlc size of %v msat is below "+
+			"min HTLC policy of %v msat for channel %v",
+			maxHtlcRequested, ch.LocalChanCfg.MinHTLC,
+			ch.FundingOutpoint)
+	}
+
+	// The max HTLC policy field must be less than or equal to the
```
Roasbeef: Could combine this and the if statement above into a single switch statement.

discovery/gossiper.go
joostjager: Good point. The only reason to keep it, is to properly deal with partial updates in the gossiper. But if we validate all policy updates in `localchans.Manager`, validation errors shouldn't occur. Only when there is a bug, the link could get desynchronized. I think that is reasonable. Carried out the simplification.

discovery/gossiper.go
joostjager: Wasn't sure how hard our line limit is. But overflow looks better indeed.

routing/localchans/manager.go
joostjager: Didn't think about that yet. But indeed, we are seeing node operators starting to use dynamic policies based on the state of a channel.

routing/localchans/manager.go
joostjager: Picked this change from val's pr

routing/localchans/manager.go
joostjager: I tried to bring this up in this pr's predecessor #3439. Imo we should either make all fields mandatory (my preference, I think this is good enough) or use flags to signal which fields to update. However, it was then decided that we only wanted to special case max htlc if I didn't misunderstood. 

I don't think the optionality logic as it is presented in this pr breaks the existing api, because the max htlc field didn't yet exist on the api.

routing/localchans/manager.go
joostjager: Code rewritten

discovery/gossiper_test.go
joostjager: It isn't rejected because the capacity of the test channels is set to zero. This appears to be an exception for light clients. But I removed this code completely, as the gossiper isn't involved anymore with specifics of the policy update.

routing/localchans/manager.go
joostjager: Converted to function closures and added a unit test.

routing/localchans/manager.go
```diff
@@ -0,0 +1,201 @@
...
+
+		// If we have a channel filter, and this channel isn't a part
+		// of it, then we'll skip it.
+		_, ok := chansToUpdate[info.ChannelPoint]
+		if !ok && haveChanFilter {
+			return nil
+		}
+
+		// Apply the new policy to the edge.
+		err := r.updateEdge(info.ChannelPoint, edge, newSchema)
```
halseth: This is a bit strange, since the updated edge policy is invalid after this call (it hasn't been re-signed). A more natural split would be for the localChanManager to fully update and re-sign the policies, and then for the gossiper to just forward them around as any other network update. but maybe that's a change too big for this PR.

joostjager: Summary from offline discussion:
* Ideally gossiper would just gossip and not sign.
* In an `lnd` mode of operation where there is no gossip (all is private), there wouldn't be a need for signatures on local updates. Signing in `localchans.Manager` wouldn't be necessary either.
* An interconnecting component that takes policies from `localchans.Manager` and resigns them when they change or become stale could be a solution.

To keep the scope of this pr under control, I only added a call to clear the signature in `localchans.Manager`.

PR #3531, opened 2019-09-23
Add BlockHeight param to SubscribeTransactions when tx is confirmed

vapopov:
This PR adds a `BlockHeight` parameter to the `SubscribeTransactions` stream, when
transaction is confirmed, otherwise we should call `GetTransaction` method
right after getting event of tx.


guggero:
LGTM.
The information is kind of redundant since we already have the block hash in there. But since the field is already present in the gRPC message, I don't see any harm of setting its value and saving a `GetTransaction` lookup later.

PR #3532, opened 2019-09-23
build+lnd: refactor reusable log file rotator code into build package

guggero:
Since we want to use persistent logging with log file rotation in other projects like loop as well, it would be nice if we could reuse the code directly instead of copy/pasting it.

This PR moves most of the log file rotation code and the code used for parsing and setting the log level into the `build` package.

### DIFF
--- a/build/log.go
+++ b/build/log.go
@@ -1,7 +1,9 @@
 package build
 
 import (
+	"fmt"
 	"io"
+	"strings"
 
 	"github.com/btcsuite/btclog"
 )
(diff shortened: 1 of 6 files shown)


lnd.go
```diff
@@ -124,9 +124,10 @@ func Main(lisCfg ListenerCfg) error {
 	}
 	cfg = loadedConfig
 	defer func() {
-		if logRotator != nil {
-			ltndLog.Info("Shutdown complete")
-			logRotator.Close()
+		ltndLog.Info("Shutdown complete")
+		err := logWriter.Close()
+		if err != nil {
+			fmt.Printf("Could not close log rotator: %v", err)
```
cfromknecht: is it possible to still use `logWriter.Errorf` if it failed to close? or is it assumed that it's already closed and that is the reason for `Printf`?

build/logrotator.go
```diff
@@ -0,0 +1,154 @@
...
+// startup and possible and must be closed on shutdown by calling `Close`.
+func (r *RotatingLogWriter) InitLogRotator(logFile string, maxLogFileSize int,
+	maxLogFiles int) {
+
+	logDir, _ := filepath.Split(logFile)
+	err := os.MkdirAll(logDir, 0700)
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr,
+			"failed to create log directory: %v\n", err)
+		os.Exit(1)
```
cfromknecht: should we return the errors in this method instead of calling `os.Exit`? as is this would prevent any deferred methods in the call stack from being executed since the program will terminate immediately.

build/log.go
```diff
@@ -93,3 +95,100 @@ func NewSubLogger(subsystem string,
...
+	// issues and update the log levels accordingly.
+	for _, logLevelPair := range strings.Split(level, ",") {
+		if !strings.Contains(logLevelPair, "=") {
+			str := "the specified debug level contains an " +
+				"invalid subsystem/level pair [%v]"
+			return fmt.Errorf(str, logLevelPair)
+		}
+
+		// Extract the specified subsystem and log level.
+		fields := strings.Split(logLevelPair, "=")
```
cfromknecht: assert that the number of fields is 2? given the above check it should be at least 2, but we can be stricter i think

lnd.go
guggero: it was my initial assumption that we probably couldn't use the logger any more after closing failed. but I looked into it and we should still be able to write to stout through the logger at least. so I fixed this.

build/logrotator.go
guggero: yes, good catch! I just copied this from the original code and didn't question it. but it makes a lot more sense to not forcefully exit here.

build/log.go
guggero: yes, agree. fixed.

rpcserver.go
```diff
@@ -3643,6 +3643,7 @@ func (r *rpcServer) SubscribeTransactions(req *lnrpc.GetTransactionsRequest,
 				Amount:           int64(tx.Value),
 				NumConfirmations: tx.NumConfirmations,
 				BlockHash:        tx.BlockHash.String(),
+				BlockHeight:      tx.BlockHeight,
```
halseth: commit should disappear on rebase?

config.go
```diff
@@ -1147,92 +1146,6 @@ func cleanAndExpandPath(path string) string {
 	return filepath.Clean(os.ExpandEnv(path))
 }
 
-// parseAndSetDebugLevels attempts to parse the specified debug level and set
```
halseth: nit: maybe hard while still having every commit compile, but nice if the code can be deleted in the same commit as it is moved. Easier to see what is new vs moved code then :)

build/logrotator.go
```diff
@@ -0,0 +1,151 @@
...
+var _ LeveledSubLogger = (*RotatingLogWriter)(nil)
+
+// NewRotatingLogWriter creates a new file rotating log writer.
+//
+// NOTE: `InitLogRotator` must be called to set up log rotation after creating
+// the writer.
+func NewRotatingLogWriter() *RotatingLogWriter {
+	logWriter := &LogWriter{}
+	backendLog := btclog.NewBackend(logWriter)
+	r := &RotatingLogWriter{
```
halseth: style nit: return struct directly

log.go
```diff
@@ -48,52 +43,79 @@ import (
...
+	chnfLog = build.NewSubLogger("CHNF", logWriter.GenSubLogger)
+	chbuLog = build.NewSubLogger("CHBU", logWriter.GenSubLogger)
+	promLog = build.NewSubLogger("PROM", logWriter.GenSubLogger)
+	wtclLog = build.NewSubLogger("WTCL", logWriter.GenSubLogger)
+	prnfLog = build.NewSubLogger("PRNF", logWriter.GenSubLogger)
+
+	// subsystemLoggers maps each subsystem identifier to its associated
+	// logger.
+	subsystemLoggers = build.SubLoggers{
+		"LTND": ltndLog,
```
halseth: can we build this map by doing something like:

```go
build.SubLoggers{
    ltndlog.Abbr(): ltndLog,
    ...
}
```

rpcserver.go
guggero: Oops, yes. It's gone.

config.go
guggero: Makes sense. I commented out some stuff with interim TODO comments that are removed later but I got it to compile on each commit.

log.go
guggero: I did some more refactoring and now it's even more compact.

config.go
halseth: If code must be commented out for it to compile, then maybe it's better to do move and delete in separates commit (I can just use the total file diff to see what was moved), but ok for now :)

Rationale is for each commit to be self-contained and "working", such that `git bisect` can function properly.

PR #3533, opened 2019-09-23
macaroons: Test Improvements

alrs:
macaroons: defer service.Close() after error check in tests
macaroons: explicitly drop unneeded variables in tests

### DIFF
--- a/macaroons/service_test.go
+++ b/macaroons/service_test.go
@@ -61,18 +61,19 @@ func TestNewService(t *testing.T) {
 	// Second, create the new service instance, unlock it and pass in a
 	// checker that we expect it to add to the bakery.
 	service, err := macaroons.NewService(tempDir, macaroons.IPLockChecker)
-	defer service.Close()
 	if err != nil {
 		t.Fatalf("Error creating new service: %v", err)
 	}
+	defer service.Close()
 	err = service.CreateUnlock(&defaultPw)
 	if err != nil {
 		t.Fatalf("Error unlocking root key storage: %v", err)
 	}
 
 	// Third, check if the created service can bake macaroons.
-	macaroon, err := service.Oven.NewMacaroon(nil, bakery.LatestVersion,
-		nil, testOperation)
+	macaroon, err := service.Oven.NewMacaroon(
+		context.TODO(), bakery.LatestVersion, nil, testOperation,
+	)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


macaroons/service_test.go
```diff
@@ -61,17 +61,17 @@ func TestNewService(t *testing.T) {
...
 	}
+	defer service.Close()
 	err = service.CreateUnlock(&defaultPw)
 	if err != nil {
 		t.Fatalf("Error unlocking root key storage: %v", err)
 	}
 
 	// Third, check if the created service can bake macaroons.
-	macaroon, err := service.Oven.NewMacaroon(nil, bakery.LatestVersion,
+	macaroon, err := service.Oven.NewMacaroon(context.TODO(), bakery.LatestVersion,
```
guggero: nit: line gets too long with the change.

macaroons/service_test.go
```diff
@@ -104,17 +104,18 @@ func TestValidateMacaroon(t *testing.T) {
...
+	defer service.Close()
+
 	err = service.CreateUnlock(&defaultPw)
 	if err != nil {
 		t.Fatalf("Error unlocking root key storage: %v", err)
 	}
 
 	// Then, create a new macaroon that we can serialize.
-	macaroon, err := service.Oven.NewMacaroon(nil, bakery.LatestVersion,
+	macaroon, err := service.Oven.NewMacaroon(context.TODO(), bakery.LatestVersion,
```
guggero: nit: line length

macaroons/service_test.go
alrs: @guggero Done.

macaroons/service_test.go
alrs: @Guggero Done.

macaroons/service_test.go
guggero: @alrs thanks!
We should describe this in the contribution guidelines, but we have a specific way of formatting argument lists of long method calls.
The code you changed is pretty old and therefore wasn't yet properly formatted to the new agreed upon style. So it would be nice if you could fix just these two calls to `NewMacaroon` to the new style. Sorry I missed this before.

What I mean is:
```
	macaroon, err := service.Oven.NewMacaroon(
		context.TODO(), bakery.LatestVersion, nil, testOperation,
	)
```

alrs: Done.

PR #3535, opened 2019-09-24
peer: link updater refactor

cfromknecht:
This PR fixes a long standing TODO, creating a minimal interface for messages that are destined for links. This allows a much simpler method of extracting the channel id rather than type switching every message individually. Several PRs in the future will build on this as we begin to refactor the peer and make it easier to test.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -1042,8 +1042,8 @@ out:
 		}
 
 		var (
-			isLinkUpdate bool
 			targetChan   lnwire.ChannelID
+			isLinkUpdate bool
 		)
 
 		switch msg := nextMsg.(type) {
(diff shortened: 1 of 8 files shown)


peer.go
```diff
@@ -1081,64 +1085,12 @@ out:
...
-					chanStream.AddMsg(nextMsg)
-
-					// Also marked this channel as failed,
-					// so we won't try to restart it on
-					// reconnect with this peer.
-					p.failedChannels[chanID] = struct{}{}
-				}
-
-			// If the channel ID for the error message corresponds
-			// to a pending channel, then the funding manager will
```
Roasbeef: Nice! Have been wanting to do this for a while, likely one of the oldest lingering TOODs in the codebase.

peer.go
```diff
@@ -1089,8 +1063,23 @@ out:
...
+			// message received was a channel sync message, then
+			// this might be a peer trying to resync closed channel.
+			// In this case we'll try to resend our last channel
+			// sync message, such that the peer can recover funds
+			// from the closed channel.
+			if !isLinkUpdate {
+				err := p.resendChanSyncMsg(targetChan)
+				if err != nil {
+					// TODO(halseth): send error to peer?
+					peerLog.Errorf("resend failed: %v",
```
wpaulino: Nit: I know this was pre-existing, but would be nice to add additional context to this log. Should also start capitalized.

PR #3545, opened 2019-09-26
build/release: create new release package for reproducible builds

Roasbeef:
In this commit, we create a new `build/release` package which houses the
container and scripts that we need to do deterministic builds across
Windows, Linux, and MacOS. As a result, all builds of `lnd` will now
take place from within this container environment. The container will
mount to the local `lnd` directory, resulting in all the build artifacts
being present in the host's cwd.

With this new system, it's now possible for all `lnd` developers, as
well as users to verify the posted build binaries. This wasn't possible
in prior release as only in Go 1.13 did deterministic builds becomes
easier/possible. See the new `README.md` for further details.

### DIFF
--- a/build/release/README.md
+++ b/build/release/README.md
@@ -1,0 +1,34 @@
+# lnd's Reproducible Build System
+
+This package contains the buidl script that the `lnd` project uses in order to
+build binaries for each new release. As of Go 1.13, with some new build flags,
+binaries are now reproducible, allowing developers to build the binary on
+distinct machines, and end up with a byte-for-byte identical binary.  However,
+this wasn't _fully_ solved in Go 1.13, as the build system still includes the
+directory the binary is built into the binary itself. As a result, our scripts
+utilize a work around needed until Go 1.13.2.  
+
+## Building a New Release
+
+### MacOS/Linux/Windows (WSL)
+
+No prior set up is needed on Linux or MacOS is required in order to build the
+release binaries.  However, on Windows, the only way to build the binaries atm
+is using the Windows Subsystem Linux. One can build the release binaries from
+one's `lnd` directory, no matter where it's located.  One can download `lnd` to
+
+```
+git clone https://github.com/lightningnetwork/lnd.git
+```
+
+Afterwards, the release manager/verifier simply needs to run:
+`./build/release/release.sh <TAG>` (from the top-level `lnd` directory) , where
+`<TAG>` is the name of the next release/tag.
+
+## Verifying a Release
+
+With Go 1.13, it's now possible for third parties to verify a release binary.
+Before this release, one had to trust that release manager to build the proper
+binary. With this new system, third parties can now _independently_ run the
+release process, and verify that all the hashes in the final `manifest.txt`
+match up.
--- a/build/release/release.sh
+++ b/build/release/release.sh
@@ -7,6 +7,8 @@
 # Use of this source code is governed by the ISC
 # license.
 
+set -e
+
 # If no tag specified, use date + version otherwise use tag.
 if [[ $1x = x ]]; then
     DATE=`date +%Y%m%d`
@@ -60,8 +62,8 @@ for i in $SYS; do
     cd $PACKAGE-$i-$TAG
 
     echo "Building:" $OS $ARCH $ARM
-    env GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags "$COMMITFLAGS" -tags="autopilotrpc signrpc walletrpc chainrpc invoicesrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lnd
-    env GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags "$COMMITFLAGS" -tags="autopilotrpc invoicesrpc walletrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lncli
+    env CGO_ENABLED=0 GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags="-s -w -buildid= $COMMITFLAGS" -tags="autopilotrpc signrpc walletrpc chainrpc invoicesrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lnd
+    env CGO_ENABLED=0 GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags="-s -w -buildid= $COMMITFLAGS" -tags="autopilotrpc invoicesrpc walletrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lncli
     cd ..
 
     if [[ $OS = "windows" ]]; then


build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+needs to be built from the _same directory_.
+
+## Building a New Release
+
+### MacOS/Windows
+
+The first requirement is to have [`docker`](https://www.docker.com/)
+installed locally and running.
+
+Next, from the top-level `lnd` directory, run the following command:
```
guggero: Needs to say:
> Next, change into the `build/release` directory and run the following command:

Otherwise it's not going to work. The `docker run` command shown later also expects you to be in `build/release`.

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+container is set up, we'll mount the present `lnd` directory _into_ the
+container at `/tmp/build/lnd`. This way we can use the existing code checked
+in, but ensure the build will be deterministic from the container's
+perspective.  After the container build is done, a new file should be
+available in the directory the container was run in, which will include the
+final release directory.
+
+To build a release, run the following command:
+```
+docker run --it -v $(pwd)/../..:/tmp/build/lnd lnd-release <TAG>
```
guggero: Needs to be `-it` not `--it`

build/release/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.13.1 as builder
+
+MAINTAINER Olaoluwa Osuntokun <laolu@lightning.engineering>
+
+ENV GO111MODULE=auto
+
+RUN apt-get install git
```
guggero: `git` and `tar` are already installed in the base image. But `zip` is missing, we use that too.
We need to change these two lines into:
`RUN apt-get update && apt-get install -y zip`

Roasbeef:
So far the script itself works on plain Linux (comparison between two
machines) all the way down to the cross compiled binaries. However, I
haven't been able to get the container to produce matching binaries when
run on MacOS.

Roasbeef:
Alright, push up a new version, that _should_ be working now!

Roasbeef:
My build for `linux-amd64` on this commit: `59162ece3287061bcdc2a81ce637b9a3f4bd9e986440325b9a4782c2adc48df7`

Built using:
```
LNDBUILDSYS=linux-amd64 ./build/release/release.sh v0.8.0-rc2
```

Roasbeef:
It also should be the case that on linux the directory no longer matters, so it can be built from _anywhere_.

Roasbeef:
Confirmed I can do the same release on MacOS and it matches that `lnd` binary hash above.

wpaulino:
This will require a clean git working tree and `go1.13.1`.

build/release/README.md
```diff
@@ -0,0 +1,62 @@
+# lnd's Reproducible Build System
+
+This package contains the container that the `lnd` project uses in order to
+build binaries for each new release. As go Go 1.13, with some new build
```
wpaulino:
```suggestion
build binaries for each new release. As of Go 1.13, with some new build
```

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+flags, binaries are now reproducible, meaning to developers can build the
```
wpaulino:
```suggestion
flags, binaries are now reproducible, allowing developers to build the
```

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+binary on distinct machines, and end up with a byte-for-byte identical
+binary. However, this wasn't _fully_ solved in Go 1.13, as the build system
```
wpaulino: Do we still need to include this note with the addition of the `-s -w` flags?

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+docker run -it -v $(pwd)/../..:/tmp/build/lnd lnd-release <TAG>
+```
+
+Where `<TAG>` is the name of the next release of `lnd`.
+
+### Linux
+
+No prior set up is needed on Linux if one doesn't wish to use the container
+packaged in this library. Instead, the only requirement is that `lnd` be
```
wpaulino: Update to remove no longer needed requirement.

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+cloned into the following directory: `/tmp/build/lnd`.
+
+This can be done with the following command: 
+```
+git clone https://github.com/lightningnetwork/lnd.git /tmp/build/
```
wpaulino: Can remove `/tmp/build` path.

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Afterwards, the release manager/verified simply need to run: `./release.sh
```
wpaulino:
```suggestion
Afterwards, the release manager/verifier simply needs to run: `./release.sh
```

build/release/README.md
```diff
@@ -0,0 +1,62 @@
...
+<TAG>`, where `<TAG>` is the name of this next release/tag.
```
wpaulino:
```suggestion
<TAG>`, where `<TAG>` is the name of the next release/tag.
```

build/release/release.sh
```diff
@@ -60,8 +62,8 @@ for i in $SYS; do
     cd $PACKAGE-$i-$TAG
 
     echo "Building:" $OS $ARCH $ARM
-    env GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags "$COMMITFLAGS" -tags="autopilotrpc signrpc walletrpc chainrpc invoicesrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lnd
-    env GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags "$COMMITFLAGS" -tags="autopilotrpc invoicesrpc walletrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lncli
+    env CGO_ENABLED=0 GOOS=$OS GOARCH=$ARCH GOARM=$ARM go build -v -trimpath -ldflags="-s -w -buildid= $COMMITFLAGS" -tags="autopilotrpc signrpc walletrpc chainrpc invoicesrpc routerrpc watchtowerrpc" github.com/lightningnetwork/lnd/cmd/lnd
```
wpaulino: Nit: extract common parts into var?

guggero:
Can confirm that with linux and go `1.13.1` I can build inside my home directory and the hash matches.

build/release/README.md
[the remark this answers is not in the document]
Roasbeef: No 1.13 is correct here. 1.13.2 will let us drop the `-buildid=` requirement.

build/release/README.md
```diff
@@ -0,0 +1,62 @@
+# lnd's Reproducible Build System
+
+This package contains the container that the `lnd` project uses in order to
+build binaries for each new release. As go Go 1.13, with some new build
+flags, binaries are now reproducible, meaning to developers can build the
+binary on distinct machines, and end up with a byte-for-byte identical
+binary. However, this wasn't _fully_ solved in Go 1.13, as the build system
+still includes the directory the binary is built into the binary itself,.
+This means that in order for developers/users to produce this binary, it
```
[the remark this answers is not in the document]
Roasbeef: Don't think so given that we can't get the same binary with Go 1.13 vs 1.13.1.

build/release/release.sh
Roasbeef: Just a bash script. I eventually want to lift this into the make file, or go port it all over to Go instead so we can use it for the rest of our Go projects.

guggero:
Did some tests on Windows 10 today:
- `git bash` (cygwin), from source directly: did *NOT* create correct hash on binaries :x: 
- `WSL` (Ubuntu on Windows), from source directly: :heavy_check_mark: 
- `docker` on Windows: 
  - source must be checked out with `git config --global core.autocrlf false`, otherwise `release.sh` won't start in the container
  - had some problems with the vendor tar, got error message `tar: file changed as we read it`. probably related to the way docker uses Windows file system and maybe even virus scanner.
  - commented out everything but build instructions and finally got matching hashes for the binaries :heavy_check_mark:

build/release/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get update && apt-get install -y zip
+
+# This will be run at container execution time, so the container should be
+# executed from to top-level directory of lnd.
+ADD release.sh .
+
+# Set our work directory to /tmp/build/lnd until Go 1.13.1 is out in order for
+# regular Linux users to reproduce these binaries, they'll also need to build
+# lnd in the /tmp/build directory, cloning lnd into that directory.
```
wpaulino: This can be removed now.

build/release/README.md
```diff
@@ -0,0 +1,60 @@
...
+build binaries for each new release. As of Go 1.13, with some new build flags,
+binaries are now reproducible, allowing developers to build the binary on
+distinct machines, and end up with a byte-for-byte identical binary.  However,
+this wasn't _fully_ solved in Go 1.13, as the build system still includes the
+directory the binary is built into the binary itself. As a result, our scripts
+utilize a work around needed until Go 1.13.2.  
+
+## Building a New Release
+
+### MacOS/Windows
```
wpaulino: Windows (Git Bash) is the only one that requires this now.

build/release/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+No prior set up is needed on Linux if one doesn't wish to use the container
+packaged in this library. Instead, you can build the release binaries from your
+`lnd` directory, no matter where it's located. One can download `lnd` to one's
+CWD with the following command:
+```
+git clone https://github.com/lightningnetwork/lnd.git
+```
+
+Afterwards, the release manager/verified simply needs to run: `./release.sh
```
wpaulino:
```suggestion
Afterwards, the release manager/verifier simply needs to run: `./release.sh
```

build/release/README.md
```diff
@@ -0,0 +1,60 @@
...
+release directory.
+
+To build a release, run the following command (from 'build/release`):
+```
+docker run -it -v $(pwd)/../..:/tmp/build/lnd lnd-release <TAG>
+```
+
+Where `<TAG>` is the name of the next release of `lnd`.
+
+### Linux
```
wpaulino:
```suggestion
### Linux/macOS/Windows (WSL)
```

wpaulino: I would also move this section above the docker container since most users will fall under this category.

build/release/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+No prior set up is needed on Linux if one doesn't wish to use the container
```
wpaulino:
```suggestion
No prior set up is needed on these systems if one doesn't wish to use the container
```

Roasbeef:
Container removed, instructions updated.

build/release/README.md
```diff
@@ -0,0 +1,34 @@
...
+utilize a work around needed until Go 1.13.2.  
+
+## Building a New Release
+
+### MacOS/Linux/Windows (WSL)
+
+No prior set up is needed on Linux or MacOS is required in order to build the
+release binaries.  However, on Windows, the only way to build the binaries atm
+is using the Windows Subsystem Linux. You can build the release binaries from
+your `lnd` directory, no matter where it's located.  One can download `lnd` to
```
wpaulino: Nit: inconsistent use of "one" and "you" 😛

build/release/README.md
```diff
@@ -0,0 +1,34 @@
+# lnd's Reproducible Build System
+
+This package contains the container that the `lnd` project uses in order to
```
wpaulino: Can remove the container reference.

build/release/README.md
Roasbeef: Must've been written at distinct times my two of my distinct selves.

build/release/README.md
```diff
@@ -0,0 +1,34 @@
...
+`./build/release/release.sh <TAG>` (from the top-level `lnd` directory) , where
+`<TAG>` is the name of the next release/tag.
+
+## Verifying a Release
+
+With Go 1.13, it's now possible for third parties to verify a release binary.
+Before this release, one had to trust that release manager to build the proper
+binary. With this new system, third parties can now _independently_ run the
+release process, and verify that all the hashes in the final `manifest.txt`
+match up.
```
cfromknecht: should we also note that verification must be done using the same go version?

build/release/README.md
```diff
@@ -0,0 +1,34 @@
+# lnd's Reproducible Build System
+
+This package contains the container that the `lnd` project uses in order to
+build binaries for each new release. As of Go 1.13, with some new build flags,
+binaries are now reproducible, allowing developers to build the binary on
+distinct machines, and end up with a byte-for-byte identical binary.  However,
+this wasn't _fully_ solved in Go 1.13, as the build system still includes the
+directory the binary is built into the binary itself. As a result, our scripts
```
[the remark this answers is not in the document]
Roasbeef: Well we'll only do our "official" builds w/ the latest one at the time likely, so after Go 1.13.2 re can remove the buildid stuff, but they'll still need to use that latest version.

build/release/README.md
Roasbeef: Was thinking to leave it to the release notes to dictate which version is used so we don't need to update this README each time a new version is released.

cfromknecht: gotcha, yeah i was thinking here we'd just say something like "when verifying, you must use the same version of go as the releaser. check the release notes for the exact version of go to use for a particular set of binaries."

molxyz:
@guggero 
I'm not sure if WSL can do a good job handling lnd, but WSL2 might be able to since it uses a real Linux kernel (https://devblogs.microsoft.com/commandline/announcing-wsl-2/).

But for those who want to run this build on native Windows command line it's also possible by installing either Cygwin or "Git for Windows" which gives you Linux commands that can run on windows. From what I understand it installs the command `sh.exe` which makes it possible for windows to build linux scripts . After the install, to make these commands run anywhere you just need to set paths for them in System Environment Variables, eg. `C:\Program Files\Git\usr\bin`

You can use either Windows PowerShell or Command Prompt to build. This is how I did on Command Prompt: 
```
cd %GOPATH%\src\github.com\lightningnetwork
git clone https://github.com/lightningnetwork/lnd
cd lnd
set LNDBUILDSYS=windows-amd64
build\release\release.sh v0.8.0-rc2
```
The build creates `lnd.exe` and `lncli.exe` in directory `%GOPATH%\src\github.com\lightningnetwork\lnd\lnd-v0.8.0-rc2\lnd-windows-amd64-v0.8.0-rc2`
Then I copy them over to the other directory `cp lncli.exe lnd.exe %GOPATH%\bin` (Not sure if there's a better way like "make install" or something?)

PR #3547, opened 2019-09-27
htlcswitch: fix non-strict forwarding failures

joostjager:
This PR fixes inconsistencies in the returned failure messages when forwarding non-strict.

### DIFF
--- a/htlcswitch/interfaces.go
+++ b/htlcswitch/interfaces.go
@@ -100,22 +100,21 @@ type ChannelLink interface {
 	// policy to govern if it an incoming HTLC should be forwarded or not.
 	UpdateForwardingPolicy(ForwardingPolicy)
 
-	// HtlcSatifiesPolicy should return a nil error if the passed HTLC
-	// details satisfy the current forwarding policy fo the target link.
-	// Otherwise, a valid protocol failure message should be returned in
-	// order to signal to the source of the HTLC, the policy consistency
-	// issue.
-	HtlcSatifiesPolicy(payHash [32]byte, incomingAmt lnwire.MilliSatoshi,
+	// CheckHtlcForward should return a nil error if the passed HTLC details
+	// satisfy the current forwarding policy fo the target link. Otherwise,
+	// a valid protocol failure message should be returned in order to
+	// signal to the source of the HTLC, the policy consistency issue.
+	CheckHtlcForward(payHash [32]byte, incomingAmt lnwire.MilliSatoshi,
 		amtToForward lnwire.MilliSatoshi,
 		incomingTimeout, outgoingTimeout uint32,
 		heightNow uint32) lnwire.FailureMessage
 
-	// HtlcSatifiesPolicyLocal should return a nil error if the passed HTLC
-	// details satisfy the current channel policy.  Otherwise, a valid
-	// protocol failure message should be returned in order to signal the
-	// violation. This call is intended to be used for locally initiated
-	// payments for which there is no corresponding incoming htlc.
-	HtlcSatifiesPolicyLocal(payHash [32]byte, amt lnwire.MilliSatoshi,
+	// CheckHtlcTransit should return a nil error if the passed HTLC details
+	// satisfy the current channel policy.  Otherwise, a valid protocol
+	// failure message should be returned in order to signal the violation.
+	// This call is intended to be used for locally initiated payments for
+	// which there is no corresponding incoming htlc.
+	CheckHtlcTransit(payHash [32]byte, amt lnwire.MilliSatoshi,
 		timeout uint32, heightNow uint32) lnwire.FailureMessage
 
 	// Bandwidth returns the amount of milli-satoshis which current link
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -2338,7 +2338,7 @@ func (l *channelLink) htlcSatifiesPolicyOutgoing(policy ForwardingPolicy,
 			l.ShortChanID(),
 		)
 		if err != nil {
-			failure = lnwire.NewTemporaryChannelFailure(update)
+			failure = &lnwire.FailTemporaryNodeFailure{}
 		} else {
 			failure = lnwire.NewExpiryTooSoon(*update)
 		}
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -2341,6 +2341,15 @@ func (l *channelLink) canSendHtlc(policy ForwardingPolicy,
 		return &lnwire.FailExpiryTooFar{}
 	}
 
+	// Check to see if there is enough balance in this channel.
+	if amt > l.Bandwidth() {
+		return l.createFailureWithUpdate(
+			func(upd *lnwire.ChannelUpdate) lnwire.FailureMessage {
+				return lnwire.NewTemporaryChannelFailure(upd)
+			},
+		)
+	}
+
 	return nil
 }
 
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -605,6 +605,19 @@ func shouldAdjustCommitFee(netFee, chanFee lnwallet.SatPerKWeight) bool {
 	}
 }
 
+// createFailureWithUpdate retrieves this link's last channel update message and
+// passes it into the callback. It expects a fully populated failure message.
+func (l *channelLink) createFailureWithUpdate(
+	cb func(update *lnwire.ChannelUpdate) lnwire.FailureMessage) lnwire.FailureMessage {
+
+	update, err := l.cfg.FetchLastChannelUpdate(l.ShortChanID())
+	if err != nil {
+		return &lnwire.FailTemporaryNodeFailure{}
+	}
+
+	return cb(update)
+}
+
 // syncChanState attempts to synchronize channel states with the remote party.
 // This method is to be called upon reconnection after the initial funding
 // flow. We'll compare out commitment chains with the remote party, and re-send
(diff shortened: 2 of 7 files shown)


htlcswitch/link.go
```diff
@@ -2271,6 +2258,19 @@ func (l *channelLink) HtlcSatifiesPolicyLocal(payHash [32]byte,
 	)
 }
 
+// createFailureWithUpdate retrieves this link's last channel update message and
+// passes it into the callback. It expects a fully populated failure message.
+func (l *channelLink) createFailureWithUpdate(
```
cfromknecht: nice simplification

htlcswitch/link.go
```diff
@@ -2327,17 +2319,11 @@ func (l *channelLink) htlcSatifiesPolicyOutgoing(policy ForwardingPolicy,
 			"outgoing_expiry=%v, best_height=%v", payHash[:],
 			timeout, heightNow)
 
-		var failure lnwire.FailureMessage
-		update, err := l.cfg.FetchLastChannelUpdate(
-			l.ShortChanID(),
+		return l.createFailureWithUpdate(
+			func(upd *lnwire.ChannelUpdate) lnwire.FailureMessage {
+				return lnwire.NewExpiryTooSoon(*upd)
```
cfromknecht: not related to this pr, but i notice some of failure constructors take a pointer while others take a value. are some of them truly optional or it just an inconsistency in the API?

htlcswitch/interfaces.go
```diff
@@ -105,17 +105,17 @@ type ChannelLink interface {
 	// Otherwise, a valid protocol failure message should be returned in
 	// order to signal to the source of the HTLC, the policy consistency
 	// issue.
-	HtlcSatifiesPolicy(payHash [32]byte, incomingAmt lnwire.MilliSatoshi,
+	HtlcCanForward(payHash [32]byte, incomingAmt lnwire.MilliSatoshi,
```
cfromknecht: perhaps `CanForwardHTLC`? godoc needs updating

htlcswitch/link.go
```diff
@@ -2240,20 +2240,20 @@ func (l *channelLink) HtlcSatifiesPolicy(payHash [32]byte,
...
-// message should be returned in order to signal the violation. This call is
-// intended to be used for locally initiated payments for which there is no
-// corresponding incoming htlc.
-func (l *channelLink) HtlcSatifiesPolicyLocal(payHash [32]byte,
+// HtlcCanSend should return a nil error if the passed HTLC details satisfy the
+// current channel policy.  Otherwise, a valid protocol failure message should
+// be returned in order to signal the violation. This call is intended to be
+// used for locally initiated payments for which there is no corresponding
+// incoming htlc.
+func (l *channelLink) HtlcCanSend(payHash [32]byte,
```
cfromknecht: similarly, `CanSendHTLC`?

htlcswitch/interfaces.go
```diff
@@ -105,17 +105,17 @@ type ChannelLink interface {
...
 		amtToForward lnwire.MilliSatoshi,
 		incomingTimeout, outgoingTimeout uint32,
 		heightNow uint32) lnwire.FailureMessage
 
-	// HtlcSatifiesPolicyLocal should return a nil error if the passed HTLC
+	// HtlcCanSend should return a nil error if the passed HTLC
```
cfromknecht: remaineder of the godoc needs to be realigned

htlcswitch/switch.go
```diff
@@ -1036,13 +1036,7 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 		for _, link := range interfaceLinks {
 			// We'll skip any links that aren't yet eligible for
 			// forwarding.
-			switch {
-			case !link.EligibleToForward():
-				continue
-
-			// If the link doesn't yet have a source chan ID, then
-			// we'll skip it as well.
-			case link.ShortChanID() == hop.Source:
```
cfromknecht: nice, it didn't used to be and looks like we missed it in the cleanup

htlcswitch/link.go
```diff
@@ -2349,6 +2335,15 @@ func (l *channelLink) htlcSatifiesPolicyOutgoing(policy ForwardingPolicy,
 		return &lnwire.FailExpiryTooFar{}
 	}
 
+	// Check to see if there is enough balance in this channel.
+	if amt > l.Bandwidth() {
```
cfromknecht: previous comparison used `>=`

htlcswitch/link.go
joostjager: I think it is correct. The only one that is optional that I saw is `FailTemporaryChannelFailure` and indeed in the decode we recognize a missing update.

htlcswitch/interfaces.go
joostjager: Renamed

htlcswitch/link.go
joostjager: If was `bandwidth < amt` and now it is `amt > bandwidth`. Identical?

cfromknecht: got it i see now, prior one used non-strict inequality for success case, and here we use the strict inequality for the failure case

htlcswitch/link.go
```diff
@@ -2271,6 +2258,19 @@ func (l *channelLink) HtlcSatifiesPolicyLocal(payHash [32]byte,
 	)
 }
 
+// createFailureWithUpdate retrieves this link's last channel update message and
```
Roasbeef: Should be moved in the file to lie before it's used.

htlcswitch/interfaces.go
```diff
@@ -100,22 +100,21 @@ type ChannelLink interface {
...
-	// protocol failure message should be returned in order to signal the
-	// violation. This call is intended to be used for locally initiated
-	// payments for which there is no corresponding incoming htlc.
-	HtlcSatifiesPolicyLocal(payHash [32]byte, amt lnwire.MilliSatoshi,
+	// CanSendHtlc should return a nil error if the passed HTLC details
+	// satisfy the current channel policy.  Otherwise, a valid protocol
+	// failure message should be returned in order to signal the violation.
+	// This call is intended to be used for locally initiated payments for
+	// which there is no corresponding incoming htlc.
+	CanSendHtlc(payHash [32]byte, amt lnwire.MilliSatoshi,
```
Roasbeef: Send -> Carry? As it's asking if this link can serve as transit for the outgoing HTLC (it's "hop zero").

htlcswitch/link_test.go
```diff
@@ -5623,47 +5623,47 @@ func TestHtlcSatisfyPolicy(t *testing.T) {
 	var hash [32]byte
 
 	t.Run("satisfied", func(t *testing.T) {
-		result := link.HtlcSatifiesPolicy(hash, 1500, 1000,
+		result := link.CanForwardHtlc(hash, 1500, 1000,
```
Roasbeef: Test name should also be changed if the method is.

htlcswitch/link.go
```diff
@@ -2335,6 +2335,15 @@ func (l *channelLink) canSendHtlc(policy ForwardingPolicy,
 		return &lnwire.FailExpiryTooFar{}
 	}
 
+	// Check to see if there is enough balance in this channel.
```
Roasbeef: Commit message leaves some to be desired, a bug is actually being addressed here, but the commit message instead conveys some generic "clean up". Which cases were previously incorrectly handled as a result of the existing logic?

htlcswitch/switch.go
```diff
@@ -790,22 +790,6 @@ func (s *Switch) handleLocalDispatch(pkt *htlcPacket) error {
...
-		if link.Bandwidth() < htlc.Amount {
-			err := fmt.Errorf("Link %v has insufficient capacity: "+
-				"need %v, has %v", pkt.outgoingChanID,
-				htlc.Amount, link.Bandwidth())
-			log.Error(err)
-
-			// The update does not need to be populated as the error
-			// will be returned back to the router.
-			htlcErr := lnwire.NewTemporaryChannelFailure(nil)
-			return &ForwardingError{
```
Roasbeef: The behavior here has changed: as is we send a `ForwardingError` back to the router, while this changes it to be a naked onion error.

htlcswitch/link.go
Roasbeef: Continuing, if a bug is indeed being fixed here, then the fix should be accompanied by a test. IIUC, the prior logic would cause us to send `UnknownNextPeer` in the scenario that there're multiple links to pick from with one of them having insufficient bandwidth and the other triggering a forwarding policy fail error (diff time lock that the other channel for example).

htlcswitch/interfaces.go
```diff
@@ -100,22 +100,21 @@ type ChannelLink interface {
...
-	// details satisfy the current forwarding policy fo the target link.
-	// Otherwise, a valid protocol failure message should be returned in
-	// order to signal to the source of the HTLC, the policy consistency
-	// issue.
-	HtlcSatifiesPolicy(payHash [32]byte, incomingAmt lnwire.MilliSatoshi,
+	// CanForwardHtlc should return a nil error if the passed HTLC details
+	// satisfy the current forwarding policy fo the target link. Otherwise,
+	// a valid protocol failure message should be returned in order to
+	// signal to the source of the HTLC, the policy consistency issue.
+	CanForwardHtlc(payHash [32]byte, incomingAmt lnwire.MilliSatoshi,
```
Roasbeef: I don't think these new names really fit, one would expect this method to return a _bool_ based on its name.

htlcswitch/link.go
joostjager: moved up

htlcswitch/interfaces.go
joostjager: Good point. I made this one `CheckHtlcForward`

htlcswitch/interfaces.go
joostjager: And this one CheckHtlcTransit

htlcswitch/link_test.go
joostjager: changed

htlcswitch/link.go
joostjager: Commit message fixed, tests added.

htlcswitch/switch.go
joostjager: This case is now checked in `CheckHtlcTransit` and that error is also wrapped in a `ForwardingError` above.

htlcswitch/test_utils.go
```diff
@@ -92,27 +92,28 @@ var (
...
-	chanID2 := lnwire.NewChanIDFromOutPoint(chanPoint2)
-
 	aliceChanID := lnwire.NewShortChanIDFromInt(id)
-	bobChanID := lnwire.NewShortChanIDFromInt(id + 1)
+
+	return chanID1, aliceChanID
+}
+
+// genIDs generates ids for two test channels.
+func genIDs() (lnwire.ChannelID, lnwire.ChannelID, lnwire.ShortChannelID,
```
Roasbeef: 👍

PR #3556, opened 2019-09-30
routerrpc: add queryprob rpc

joostjager:
In #3462, we removed probability reporting for `querymc`. This PR brings it back as a separate rpc for which a specific amount needs to be given.

In addition to the probability, it also returns the raw data for the requested pair. An example use case for this is to get a sense for the freshness of a pair and use this to decide whether to do a prepay probe for a route.


cmd/lncli/cmd_query_mc_probability.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/urfave/cli"
+)
+
+var queryMcProbCommand = cli.Command{
+	Name:      "querymcprob",
```
carlaKC: Wondering is it's necessary to include `mc` in the command name?
Mission control is an internal concept that users might not be familiar with, `queryPaymentProbability` or something more user suited may be more informative?

cmd/lncli/cmd_query_mc_probability.go
```diff
@@ -0,0 +1,70 @@
...
+		btcutil.Amount(amtSat),
+	)
+
+	conn := getClientConn(ctx, false)
+	defer conn.Close()
+
+	client := routerrpc.NewRouterClient(conn)
+
+	req := &routerrpc.QueryMcProbabilityRequest{
+		NodeFrom: fromNode[:],
```
carlaKC: nit: RPC uses `Node(to/from)` and lncli uses `(To/From)Node`, possibly unify to one arrangement for both?

lnrpc/routerrpc/router_server.go
```diff
@@ -491,6 +487,22 @@ func (s *Server) QueryMissionControl(ctx context.Context,
...
+func toRPCPairData(data *routing.TimedPairResult) *PairData {
+	rpcData := PairData{
+		MinPenalizeAmtSat: int64(
+			data.MinPenalizeAmt.ToSatoshis(),
+		),
+		LastAttemptSuccessful: data.Success,
+	}
+
+	if !data.Timestamp.IsZero() {
+		rpcData.Timestamp = data.Timestamp.Unix()
```
halseth: why don't do `Unix()` also when zero?

cmd/lncli/cmd_query_mc_probability.go
halseth: just `queryprob`?

lnrpc/routerrpc/router.proto
```diff
@@ -367,6 +367,25 @@ message PairData {
...
+    /// The amount for which to calculate a probability.
+    int64 amt_msat = 3 [json_name = "amt_msat"];
+}
+
+message QueryMcProbabilityResponse{
+    /// The success probability for the requested pair.
+    double probability = 1 [json_name = "probability"];
+
+    /// The historical data for the requested pair.
+    PairData history = 2 [json_name = "history"];
```
halseth: is the reason for `PairData` to reuse it here?

lnrpc/routerrpc/router_server.go
joostjager: It is a very negative number. The beginning of `time.Time` is far before 1970.

lnrpc/routerrpc/router.proto
joostjager: yes

cmd/lncli/cmd_query_mc_probability.go
joostjager: changed to `queryprob`

cmd/lncli/cmd_query_mc_probability.go
joostjager: swapped

lnrpc/routerrpc/router_server.go
halseth: ouch, TIL

PR #3557, opened 2019-09-30
input+sweep: rework witness type into interface to be used externally

guggero:
This PR has the goal to make the sweeper module usable outside of `lnd`.

This is achieved by reworking the `WitnessType` into an interface.
There is a list of `StandardWitnessType`s that represent all internally known witness types of lnd.
In internal functionalities like `breacharbiter` or `utxonursery` only the standard witness types are used.

Only the `input` package's `MakeBaseInput` accepts generic witness types so the sweeper can be used to sweep custom witness outputs outside of lnd.

### DIFF
--- a/breacharbiter.go
+++ b/breacharbiter.go
@@ -821,7 +821,7 @@ func (b *breachArbiter) handleBreachHandoff(breachEvent *ContractBreachEvent) {
 type breachedOutput struct {
 	amt         btcutil.Amount
 	outpoint    wire.OutPoint
-	witnessType input.WitnessType
+	witnessType input.StandardWitnessType
 	signDesc    input.SignDescriptor
 	confHeight  uint32
 
@@ -833,7 +833,7 @@ type breachedOutput struct {
 // makeBreachedOutput assembles a new breachedOutput that can be used by the
 // breach arbiter to construct a justice or sweep transaction.
 func makeBreachedOutput(outpoint *wire.OutPoint,
-	witnessType input.WitnessType,
+	witnessType input.StandardWitnessType,
 	secondLevelScript []byte,
 	signDescriptor *input.SignDescriptor,
 	confHeight uint32) breachedOutput {
@@ -883,9 +883,7 @@ func (bo *breachedOutput) CraftInputScript(signer input.Signer, txn *wire.MsgTx,
 
 	// First, we ensure that the witness generation function has been
 	// initialized for this breached output.
-	bo.witnessFunc = bo.witnessType.GenWitnessFunc(
-		signer, bo.SignDesc(),
-	)
+	bo.witnessFunc = bo.witnessType.WitnessGenerator(signer, bo.SignDesc())
 
 	// Now that we have ensured that the witness generation function has
 	// been initialized, we can proceed to execute it and generate the
@@ -993,7 +991,7 @@ func newRetributionInfo(chanPoint *wire.OutPoint,
 		// Using the breachedHtlc's incoming flag, determine the
 		// appropriate witness type that needs to be generated in order
 		// to sweep the HTLC output.
-		var htlcWitnessType input.WitnessType
+		var htlcWitnessType input.StandardWitnessType
 		if breachedHtlc.IsIncoming {
 			htlcWitnessType = input.HtlcAcceptedRevoke
 		} else {
@@ -1051,32 +1049,15 @@ func (b *breachArbiter) createJusticeTx(
 		// Grab locally scoped reference to breached output.
 		inp := &r.breachedOutputs[i]
 
-		// First, select the appropriate estimated witness weight for
+		// First, determine the appropriate estimated witness weight for
 		// the give witness type of this breached output. If the witness
-		// type is unrecognized, we will omit it from the transaction.
-		var witnessWeight int
-		switch inp.WitnessType() {
-		case input.CommitSpendNoDelayTweakless:
-			fallthrough
-		case input.CommitmentNoDelay:
-			witnessWeight = input.P2WKHWitnessSize
-
-		case input.CommitmentRevoke:
-			witnessWeight = input.ToLocalPenaltyWitnessSize
-
-		case input.HtlcOfferedRevoke:
-			witnessWeight = input.OfferedHtlcPenaltyWitnessSize
-
-		case input.HtlcAcceptedRevoke:
-			witnessWeight = input.AcceptedHtlcPenaltyWitnessSize
-
-		case input.HtlcSecondLevelRevoke:
-			witnessWeight = input.ToLocalPenaltyWitnessSize
-
-		default:
-			brarLog.Warnf("breached output in retribution info "+
-				"contains unexpected witness type: %v",
-				inp.WitnessType())
+		// weight cannot be estimated, we will omit it from the
+		// transaction.
+		witnessWeight, _, err := inp.WitnessType().SizeUpperBound()
+		if err != nil {
+			brarLog.Warnf("could not determine witness weight "+
+				"for breached output in retribution info: %v",
+				err)
 			continue
 		}
 		weightEstimate.AddWitnessInput(witnessWeight)
@@ -1555,7 +1536,7 @@ func (bo *breachedOutput) Decode(r io.Reader) error {
 	if _, err := io.ReadFull(r, scratch[:2]); err != nil {
 		return err
 	}
-	bo.witnessType = input.WitnessType(
+	bo.witnessType = input.StandardWitnessType(
 		binary.BigEndian.Uint16(scratch[:2]),
 	)
 
--- a/input/input.go
+++ b/input/input.go
@@ -114,9 +114,7 @@ func NewBaseInput(outpoint *wire.OutPoint, witnessType WitnessType,
 func (bi *BaseInput) CraftInputScript(signer Signer, txn *wire.MsgTx,
 	hashCache *txscript.TxSigHashes, txinIdx int) (*Script, error) {
 
-	witnessFunc := bi.witnessType.GenWitnessFunc(
-		signer, bi.SignDesc(),
-	)
+	witnessFunc := bi.witnessType.WitnessGenerator(signer, bi.SignDesc())
 
 	return witnessFunc(txn, hashCache, txinIdx)
 }
--- a/input/witnessgen.go
+++ b/input/witnessgen.go
@@ -7,87 +7,125 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
-// WitnessType determines how an output's witness will be generated. The
-// default commitmentTimeLock type will generate a witness that will allow
-// spending of a time-locked transaction enforced by CheckSequenceVerify.
-type WitnessType uint16
+// WitnessGenerator represents a function that is able to generate the final
+// witness for a particular public key script. Additionally, if required, this
+// function will also return the sigScript for spending nested P2SH witness
+// outputs. This function acts as an abstraction layer, hiding the details of
+// the underlying script.
+type WitnessGenerator func(tx *wire.MsgTx, hc *txscript.TxSigHashes,
+	inputIndex int) (*Script, error)
+
+// WitnessType determines how an output's witness will be generated. This
+// interface can be implemented to be used for custom sweep scripts if the
+// pre-defined StandardWitnessType list doesn't provide a suitable one.
+type WitnessType interface {
+	// String returns a human readable version of the WitnessType.
+	String() string
+
+	// WitnessGenerator will return a WitnessGenerator function that an
+	// output uses to generate the witness and optionally the sigScript for
+	// a sweep transaction.
+	WitnessGenerator(signer Signer,
+		descriptor *SignDescriptor) WitnessGenerator
+
+	// SizeUpperBound returns the maximum length of the witness of this
+	// WitnessType if it would be included in a tx. It also returns if the
+	// output itself is a nested p2sh output, if so then we need to take
+	// into account the extra sigScript data size.
+	SizeUpperBound() (int, bool, error)
+
+	// AddWeightEstimation adds the estimated size of the witness in bytes
+	// to the given weight estimator and returns the number of
+	// CSVs/CLTVs used by the script.
+	AddWeightEstimation(estimator *TxWeightEstimator) (int, int, error)
+}
+
+// StandardWitnessType is a numeric representation of standard pre-defined types
+// of witness configurations.
+type StandardWitnessType uint16
+
+// A compile time check to ensure StandardWitnessType implements the
+// WitnessType interface.
+var _ WitnessType = (StandardWitnessType)(0)
 
 const (
 	// CommitmentTimeLock is a witness that allows us to spend the output of
 	// a commitment transaction after a relative lock-time lockout.
-	CommitmentTimeLock WitnessType = 0
+	CommitmentTimeLock StandardWitnessType = 0
 
 	// CommitmentNoDelay is a witness that allows us to spend a settled
 	// no-delay output immediately on a counterparty's commitment
 	// transaction.
-	CommitmentNoDelay WitnessType = 1
+	CommitmentNoDelay StandardWitnessType = 1
 
 	// CommitmentRevoke is a witness that allows us to sweep the settled
 	// output of a malicious counterparty's who broadcasts a revoked
 	// commitment transaction.
-	CommitmentRevoke WitnessType = 2
+	CommitmentRevoke StandardWitnessType = 2
 
 	// HtlcOfferedRevoke is a witness that allows us to sweep an HTLC which
 	// we offered to the remote party in the case that they broadcast a
 	// revoked commitment state.
-	HtlcOfferedRevoke WitnessType = 3
+	HtlcOfferedRevoke StandardWitnessType = 3
 
 	// HtlcAcceptedRevoke is a witness that allows us to sweep an HTLC
 	// output sent to us in the case that the remote party broadcasts a
 	// revoked commitment state.
-	HtlcAcceptedRevoke WitnessType = 4
+	HtlcAcceptedRevoke StandardWitnessType = 4
 
 	// HtlcOfferedTimeoutSecondLevel is a witness that allows us to sweep
 	// an HTLC output that we extended to a party, but was never fulfilled.
 	// This HTLC output isn't directly on the commitment transaction, but
 	// is the result of a confirmed second-level HTLC transaction. As a
 	// result, we can only spend this after a CSV delay.
-	HtlcOfferedTimeoutSecondLevel WitnessType = 5
+	HtlcOfferedTimeoutSecondLevel StandardWitnessType = 5
 
 	// HtlcAcceptedSuccessSecondLevel is a witness that allows us to sweep
 	// an HTLC output that was offered to us, and for which we have a
 	// payment preimage. This HTLC output isn't directly on our commitment
 	// transaction, but is the result of confirmed second-level HTLC
 	// transaction. As a result, we can only spend this after a CSV delay.
-	HtlcAcceptedSuccessSecondLevel WitnessType = 6
+	HtlcAcceptedSuccessSecondLevel StandardWitnessType = 6
 
 	// HtlcOfferedRemoteTimeout is a witness that allows us to sweep an
 	// HTLC that we offered to the remote party which lies in the
 	// commitment transaction of the remote party. We can spend this output
 	// after the absolute CLTV timeout of the HTLC as passed.
-	HtlcOfferedRemoteTimeout WitnessType = 7
+	HtlcOfferedRemoteTimeout StandardWitnessType = 7
 
 	// HtlcAcceptedRemoteSuccess is a witness that allows us to sweep an
 	// HTLC that was offered to us by the remote party. We use this witness
 	// in the case that the remote party goes to chain, and we know the
 	// pre-image to the HTLC. We can sweep this without any additional
 	// timeout.
-	HtlcAcceptedRemoteSuccess WitnessType = 8
+	HtlcAcceptedRemoteSuccess StandardWitnessType = 8
 
 	// HtlcSecondLevelRevoke is a witness that allows us to sweep an HTLC
 	// from the remote party's commitment transaction in the case that the
 	// broadcast a revoked commitment, but then also immediately attempt to
 	// go to the second level to claim the HTLC.
-	HtlcSecondLevelRevoke WitnessType = 9
+	HtlcSecondLevelRevoke StandardWitnessType = 9
 
 	// WitnessKeyHash is a witness type that allows us to spend a regular
 	// p2wkh output that's sent to an output which is under complete
 	// control of the backing wallet.
-	WitnessKeyHash WitnessType = 10
+	WitnessKeyHash StandardWitnessType = 10
 
 	// NestedWitnessKeyHash is a witness type that allows us to sweep an
 	// output that sends to a nested P2SH script that pays to a key solely
 	// under our control. The witness generated needs to include the
-	NestedWitnessKeyHash WitnessType = 11
+	NestedWitnessKeyHash StandardWitnessType = 11
 
 	// CommitSpendNoDelayTweakless is similar to the CommitSpendNoDelay
 	// type, but it omits the tweak that randomizes the key we need to
 	// spend with a channel peer supplied set of randomness.
-	CommitSpendNoDelayTweakless = 12
+	CommitSpendNoDelayTweakless StandardWitnessType = 12
 )
 
-// Stirng returns a human readable version of the target WitnessType.
-func (wt WitnessType) String() string {
+// String returns a human readable version of the target WitnessType.
+//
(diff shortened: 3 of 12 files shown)


input/witnessgen.go
```diff
@@ -262,5 +266,103 @@ func (wt WitnessType) GenWitnessFunc(signer Signer,
...
+
+// GetWitnessSizeUpperBound returns the maximum length of the witness for
+// the given type if it would be included in a tx. We also return if the
+// output itself is a nested p2sh output, if so then we need to take into
+// account the extra sigScript data size.
+func GetWitnessSizeUpperBound(wt WitnessType) (int, bool, error) {
+	switch wt {
+
+	// Outputs on a remote commitment transaction that pay directly to us.
+	case CommitSpendNoDelayTweakless:
```
joostjager: Isn't this switch an indication that `GetWitnessSizeUpperBound` should really be a method of `WitnessType`? Or having a `WitnessType` interface, so that witness types can be implemented independently?

input/witnessgen.go
```diff
@@ -2,92 +2,124 @@ package input
...
+	inputIndex int) (*Script, error)
+
 // WitnessType determines how an output's witness will be generated. The
-// default commitmentTimeLock type will generate a witness that will allow
-// spending of a time-locked transaction enforced by CheckSequenceVerify.
-type WitnessType uint16
+//// default commitmentTimeLock type will generate a witness that will allow
+//// spending of a time-locked transaction enforced by CheckSequenceVerify.
+type WitnessType interface {
+	Type() uint16
```
joostjager: return type `StandardWitnessType`?

input/witnessgen.go
```diff
@@ -2,92 +2,124 @@ package input
...
+	String() string
+	GenWitnessFunc(signer Signer,
+		descriptor *SignDescriptor) WitnessGenerator
+	GetWitnessSizeUpperBound() (int, bool, error)
+	GetWitnessWeightEstimation() (int, int, int, error)
```
joostjager: Remove `Witness` prefix in this method names?

input/witnessgen.go
```diff
@@ -262,5 +286,103 @@ func (wt WitnessType) GenWitnessFunc(signer Signer,
...
+	}
+
+	switch wt {
+	case CommitmentTimeLock,
+		HtlcOfferedTimeoutSecondLevel,
+		HtlcAcceptedSuccessSecondLevel:
+		csvCount++
+	case HtlcOfferedRemoteTimeout:
+		cltvCount++
+	}
```
joostjager: I was even thinking about making all these individual witness type separate objects. To avoid the kind of if/switch logic above.

input/witnessgen.go
```diff
@@ -2,92 +2,124 @@ package input
...
 	// CommitSpendNoDelayTweakless is similar to the CommitSpendNoDelay
 	// type, but it omits the tweak that randomizes the key we need to
 	// spend with a channel peer supplied set of randomness.
-	CommitSpendNoDelayTweakless = 12
+	CommitSpendNoDelayTweakless StandardWitnessType = 12
+	
+	// Custom is a special witness type that should be used if the
+	// WitnessType interface is implemented and none of the pre-defined
+	// internal witness types are used.
+	Custom StandardWitnessType = 13
```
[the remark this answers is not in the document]
guggero: I've implemented it as we discussed and with the proposed value of `999`.

input/witnessgen.go
guggero: Makes sense, done.

input/witnessgen.go
guggero: As discussed it makes sense to not return the stricter type here since an implementing interface would return the value of `CustomType` here.

input/witnessgen.go
guggero: We should try to do this as well but that is probably content for another PR.

input/witnessgen.go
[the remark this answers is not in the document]
guggero: I was even able to get rid of the custom type after all!
Only `breacharbiter` and `utxonursery` actually use the `Type()` function and they are only ever handling standard witness types. So I changed them to only accept standard types.
Now the sweeper is the only part of lnd that accepts generic `WitnessType`s.

sweep/txgenerator.go
```diff
@@ -56,7 +56,7 @@ func generateInputPartitionings(sweepableInputs []input.Input,
 	// on the signature length, which is not known yet at this point.
 	yields := make(map[wire.OutPoint]int64)
 	for _, input := range sweepableInputs {
-		size, _, err := getInputWitnessSizeUpperBound(input)
+		size, _, err := input.WitnessType().GetWitnessSizeUpperBound()
```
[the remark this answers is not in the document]
guggero: The input doesn't know the witness size, it would only delegate the call to the witness type.

But I was able to eliminate the `Custom` type by removing the `Type()` method from the `WitnessType`. The numeric type is only used in `utxonursery` and `breacharbiter` and they work fine by only using the standard witness types.

server.go
```diff
@@ -388,10 +388,10 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 		ignorePeerTermination:   make(map[*peer]struct{}),
 		scheduledPeerConnection: make(map[string]func()),
 
-		peersByPub:                make(map[string]*peer),
-		inboundPeers:              make(map[string]*peer),
-		outboundPeers:             make(map[string]*peer),
-		peerConnectedListeners:    make(map[string][]chan<- lnpeer.Peer),
+		peersByPub:             make(map[string]*peer),
```
halseth: what changed here?

sweep/sweeper.go
```diff
@@ -323,11 +321,26 @@ func (s *UtxoSweeper) Start() error {
...
+		bestHash   *chainhash.Hash
+	)
+	bestBlockEpochs, err := s.cfg.Notifier.RegisterBlockEpochNtfn(nil)
 	if err != nil {
-		return fmt.Errorf("get best block: %v", err)
+		return fmt.Errorf("register best block ntfn: %v", err)
 	}
+	
+	// We need to block and wait for the current best block to arrive.
+	epoch := <-bestBlockEpochs.Epochs
```
halseth: this is a bit dangerous...

sweep/sweeper.go
```diff
@@ -323,11 +321,26 @@ func (s *UtxoSweeper) Start() error {
...
-	// Register for block epochs to retry sweeping every block.
-	bestHash, bestHeight, err := s.cfg.ChainIO.GetBestBlock()
+	// We need to get the current best block so we can register for block
+	// epochs and retry sweeping every block. We should get a notification
+	// immediately if we don't provide any epoch.
+	var (
+		bestHeight int32
+		bestHash   *chainhash.Hash
+	)
+	bestBlockEpochs, err := s.cfg.Notifier.RegisterBlockEpochNtfn(nil)
```
halseth: since we re-register for block epochs below, is this even needed? Could make the `collector` below always wait for first block to arrive, as it seems like its the only one using it

input/witnessgen.go
```diff
@@ -7,87 +7,118 @@ import (
...
+// the underlying script.
+type WitnessGenerator func(tx *wire.MsgTx, hc *txscript.TxSigHashes,
+	inputIndex int) (*Script, error)
+
+// WitnessType determines how an output's witness will be generated. This
+// interface can be implemented to use for custom sweep scripts if the
+// pre-defined StandardWitnessType list doesn't provide a suitable one.
+//
+// NOTE: Custom implementations of this interface should return the value
+// CustomType (9999) in the Type() function.
```
halseth: outdated comment?

input/witnessgen.go
```diff
@@ -7,87 +7,118 @@ import (
...
+type WitnessType interface {
+	String() string
```
halseth: these should have godoc

server.go
guggero: Was unintended formatting change. Fixed it.

sweep/sweeper.go
guggero: I moved the wait for the first block into the `collector` as you suggested. I wasn't sure how the notifier would behave after the first nil request. But I looked into it and it seems like we get exactly what we want with only registering once.
So I removed the extra registration.

sweep/sweeper.go
guggero: I agree. See comment below.

input/witnessgen.go
guggero: Fixed

input/witnessgen.go
guggero: Fixed

input/witnessgen.go
```diff
@@ -7,87 +7,118 @@ import (
...
 	// type, but it omits the tweak that randomizes the key we need to
 	// spend with a channel peer supplied set of randomness.
-	CommitSpendNoDelayTweakless = 12
+	CommitSpendNoDelayTweakless StandardWitnessType = 12
 )
 
-// Stirng returns a human readable version of the target WitnessType.
-func (wt WitnessType) String() string {
+// Type returns the numerical type of the StandardWitnessType.
+func (wt StandardWitnessType) Type() StandardWitnessType {
```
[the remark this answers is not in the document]
guggero: No, you're right, it's not even used anywhere any more. Removed.

sweep/txgenerator.go
```diff
@@ -328,38 +276,20 @@ func getWeightEstimate(inputs []input.Input) ([]input.Input, int64, int, int) {
 	for i := range inputs {
 		inp := inputs[i]
 
-		// For fee estimation purposes, we'll now attempt to obtain an
-		// upper bound on the weight this input will add when fully
-		// populated.
-		size, isNestedP2SH, err := getInputWitnessSizeUpperBound(inp)
+		weight, csvCnt, cltvCnt, err := 
+			inp.WitnessType().GetWeightEstimation()
```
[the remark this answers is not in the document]
guggero: Yes, looks cleaner that way, fixed.

sweep/sweeper.go
```diff
@@ -323,11 +321,26 @@ func (s *UtxoSweeper) Start() error {
 	// not change from here on.
 	s.relayFeeRate = s.cfg.FeeEstimator.RelayFeePerKW()
 
-	// Register for block epochs to retry sweeping every block.
-	bestHash, bestHeight, err := s.cfg.ChainIO.GetBestBlock()
+	// We need to get the current best block so we can register for block
+	// epochs and retry sweeping every block. We should get a notification
+	// immediately if we don't provide any epoch.
```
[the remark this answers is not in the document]
guggero: The rationale is that we don't want to use `GetBestBlock` any more because it is not exposed through any RPC at the moment. So by using the chain notifier to get the current best block we have everything provided through RPCs that the sweeper needs to work outside of lnd.
I updated the commit message to make this clearer.

breacharbiter.go
```diff
@@ -1051,32 +1049,16 @@ func (b *breachArbiter) createJusticeTx(
...
-			brarLog.Warnf("breached output in retribution info "+
-				"contains unexpected witness type: %v",
-				inp.WitnessType())
+		// weight cannot be estimated, we will omit it from the
+		// transaction.
+		witnessWeight, _, err := inp.WitnessType().GetSizeUpperBound()
+		if err != nil {
+			brarLog.Warnf("could not determine witness weight "+
+				"for breached output in retribution info: %v",
+				err,
```
halseth: style nit: we usually don't add newline before closing bracket for formatting style method calls

input/witnessgen.go
```diff
@@ -7,87 +7,123 @@ import (
...
+// interface can be implemented to be used for custom sweep scripts if the
+// pre-defined StandardWitnessType list doesn't provide a suitable one.
+type WitnessType interface {
+	// String returns a human readable version of the WitnessType.
+	String() string
+	
+	// GenFunc will return a WitnessGenerator function that an output uses
+	// to generate the witness and optionally the sigScript for a sweep
+	// transaction.
+	GenFunc(signer Signer, descriptor *SignDescriptor) WitnessGenerator
```
halseth: I feel `GenFunc` is not a very descriptive name. Could we call it just `(Get)WitnessGenerator`?

input/witnessgen.go
```diff
@@ -7,87 +7,123 @@ import (
...
+	
+	// GetSizeUpperBound returns the maximum length of the witness of this
+	// WitnessType if it would be included in a tx. It also returns if the
+	// output itself is a nested p2sh output, if so then we need to take
+	// into account the extra sigScript data size.
+	GetSizeUpperBound() (int, bool, error)
```
halseth: I think idiomatic golang is to just name getters like`SizeUpperBound`.

input/input.go
```diff
@@ -74,6 +74,25 @@ func (i *inputKit) HeightHint() uint32 {
...
+// txIndex within the passed transaction. The input scripts generated by this
+// method support spending p2wkh, p2wsh, and also nested p2sh outputs.
+func (i *inputKit) CraftInputScript(signer Signer, txn *wire.MsgTx,
+	hashCache *txscript.TxSigHashes, txinIdx int) (*Script, error) {
+
+	witnessFunc := i.witnessType.GenWitnessFunc(signer, i.SignDesc())
+
+	return witnessFunc(txn, hashCache, txinIdx)
+}
+
```
joostjager: Add commit message with rationale for this? I think the original idea was to allow `Input` implementations to provide their own implementations of these functions. But maybe those don't exist, other than `BaseInput`. Or is there some hiding going on now with this change?

sweep/test_utils.go
```diff
@@ -155,12 +155,24 @@ func (m *MockNotifier) RegisterBlockEpochNtfn(
...
+	case bestBlock == nil:
+		epochChan = make(chan *chainntnfs.BlockEpoch, 1)
+		epochChan <- &chainntnfs.BlockEpoch{
+			Hash:   mockChainHash,
+			Height: mockChainHeight,
+		}
+		m.epochChan[epochChan] = mockChainHeight
+	default:
+		epochChan = make(chan *chainntnfs.BlockEpoch)
+		m.epochChan[epochChan] = bestBlock.Height
```
joostjager: Not sure if the two different channel buffer sizes are going to create surprises for future devs? Maybe this initial send can be done in a goroutine. Or will that create the risk that the first block that is manually pushed in comes before that initial send? Could be prevented by having a notifyBlock method that first checks whether the initial send goroutine has completed.

input/witnessgen.go
```diff
@@ -262,5 +292,109 @@ func (wt WitnessType) GenWitnessFunc(signer Signer,
...
+	switch wt {
+	case CommitmentTimeLock,
+		HtlcOfferedTimeoutSecondLevel,
+		HtlcAcceptedSuccessSecondLevel:
+		csvCount++
+	case HtlcOfferedRemoteTimeout:
+		cltvCount++
+	}
 
+	return weightEstimate.Weight(), csvCount, cltvCount, nil
```
joostjager: These counts are either 0 or 1. Could be bools?

sweep/txgenerator.go
```diff
@@ -317,6 +264,7 @@ func getWeightEstimate(inputs []input.Input) ([]input.Input, int64, int, int) {
 	// Our sweep transaction will pay to a single segwit p2wkh address,
 	// ensure it contributes to our weight estimate.
 	weightEstimate.AddP2WKHOutput()
+	txWeight := int64(weightEstimate.Weight())
```
joostjager: The inputs and outputs are now all calculating their weights in isolation. I think things are being counted multiple times now, see implementation of `Weight()`. For example the BaseTxSize.

sweep/sweeper.go
```diff
@@ -445,9 +432,15 @@ func (s *UtxoSweeper) feeRateForPreference(
...
-func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
-	bestHeight int32) error {
-
+func (s *UtxoSweeper) collector(
+	blockEpochs <-chan *chainntnfs.BlockEpoch) error {
+
+	// We registered for the block epochs with a nil request. The notifier
+	// should send us the current best block immediately. So we need to wait
+	// for it here because we need to know the current best height.
+	bestBlock := <- blockEpochs
```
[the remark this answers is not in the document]
guggero: Yes, good catch!
The linter also urged me to remove the return value of `collector` since we ever only return nil.

input/witnessgen.go
guggero: Renamed it to `WitnessGenerator` and removed the `Get` prefixes of the other methods.

input/input.go
guggero: This was some leftover refactoring from my early trials. But you are right, it's not very clear why I did this. So I only moved the `BlocksToMaturity` to the `inputKit` to avoid some code duplication. Also described that in the commit message.

sweep/test_utils.go
guggero: It seems to work fine with a goroutine. Just didn't think about doing it that way when implementing it.
But makes a lot of sense to do it that way to not create surprises for future devs.

input/witnessgen.go
guggero: For the current types they can either be 0 or 1. But in theory a script could contain multiple of those.
The real issue is that these values are only ever used for a log statement. I did think about removing that information from the log statement but thought it might have been put there for a reason.

sweep/txgenerator.go
guggero: Very good catch! I did not expect the `WeightEstimator` to behave this way. Fixed this and also added a test to lock down the fee estimation.
Ran the test against master and got the same results (apart from four witness types that didn't have a size implemented in master before the refactor).

input/witnessgen.go
```diff
@@ -7,87 +7,125 @@ import (
...
+	// SizeUpperBound returns the maximum length of the witness of this
+	// WitnessType if it would be included in a tx. It also returns if the
+	// output itself is a nested p2sh output, if so then we need to take
+	// into account the extra sigScript data size.
+	SizeUpperBound() (int, bool, error)
+
+	// AddWeightEstimation adds the estimated size of the witness in bytes
+	// to the given weight estimator and returns the number of
+	// CSVs/CLTVs used by the script.
+	AddWeightEstimation(estimator *TxWeightEstimator) (int, int, error)
```
joostjager: Yes, better approach :+1:

sweep/txgenerator.go
```diff
@@ -328,34 +275,17 @@ func getWeightEstimate(inputs []input.Input) ([]input.Input, int64, int, int) {
 	for i := range inputs {
 		inp := inputs[i]
 
-		// For fee estimation purposes, we'll now attempt to obtain an
-		// upper bound on the weight this input will add when fully
-		// populated.
-		size, isNestedP2SH, err := getInputWitnessSizeUpperBound(inp)
+		wt := inp.WitnessType()
+		csvCnt, cltvCnt, err := wt.AddWeightEstimation(&weightEstimate)
```
joostjager: I think these counts are now overwritten every time. Need to make a `var err error` and use `=` instead of `:=`

sweep/txgenerator_test.go
```diff
@@ -0,0 +1,89 @@
...
+			),
+		})
+		if weight != test.expectedWeight {
+			t.Fatalf("unexpected weight for witness type %s. "+
+				"expected %d but got %d",
+				test.witnessType.String(), test.expectedWeight,
+				weight)
+		}
+	}
+}
```
joostjager: What is the purpose of this test? It seems like it is just testing the constants in the code. A different setup is needed to catch the bug that was introduced in the previous revision of this pr. It would need to have multiple components added to the same estimator.

input/input.go
```diff
@@ -177,13 +177,6 @@ func (h *HtlcSucceedInput) CraftInputScript(signer Signer, txn *wire.MsgTx,
 	}, nil
 }
 
-// BlocksToMaturity returns the relative timelock, as a number of blocks, that
-// must be built on top of the confirmation height before the output can be
-// spent.
-func (h *HtlcSucceedInput) BlocksToMaturity() uint32 {
-	return 0
-}
```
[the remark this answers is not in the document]
guggero: Ok, I dropped the commit.

sweep/txgenerator.go
guggero: Yes, these variables are overwritten every time. But at the end of the loop they are added to the total counts (variables that are outside of the loop and not shown in the diff).
I renamed the vars inside the loop to make it more clear.

sweep/txgenerator_test.go
guggero: Correct, this test is only useful for more than one input. I fixed it and again verified it to be correct against master too.

sweep/txgenerator.go
joostjager: Ah, they were indeed almost identically named

sweep/txgenerator_test.go
joostjager: Ok. Only I don't think it is necessary to test all these witness types. What more is it than asserting on constants? Having a single `getWeightEstimate` test to assert that we only use one weight estimator internally seems like enough to me.

sweep/sweeper.go
```diff
@@ -445,8 +429,19 @@ func (s *UtxoSweeper) feeRateForPreference(
 
 // collector is the sweeper main loop. It processes new inputs, spend
 // notifications and counts down to publication of the sweep tx.
-func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch,
-	bestHeight int32) error {
+func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
+
```
halseth: style nit: no newline needed when method signature is a single line

sweep/test_utils.go
```diff
@@ -155,12 +155,26 @@ func (m *MockNotifier) RegisterBlockEpochNtfn(
...
+	case bestBlock == nil:
+		// Send out the notification immediately but do it in a
+		// goroutine to not block execution of this method.
+		go func() {
+			epochChan <- &chainntnfs.BlockEpoch{
+				Hash:   mockChainHash,
+				Height: mockChainHeight,
+			}
+		}()
+		m.epochChan[epochChan] = mockChainHeight
```
halseth: @joostjager How would buffering the channel with 1 and sending the block immediately lead to surprises to future devs?  This is how we usually add elements to a channel before returning afaik, and I would prefer that approach to a goroutine  also here.

input/witnessgen.go
halseth: Oh, yeah pretty nice 😀

sweep/txgenerator_test.go
halseth: A simple testcase with mulitple inputs that asserts final weight and CSV/CLTV count is maybe sufficient?

sweep/test_utils.go
joostjager: Buffering with 1 is ok. But in the previous iteration it would make only a buffered channel in case of a `bestBlock` being present. Otherwise it would be an unbuffered channel. That was what I meant by surprise - it is kind of unclear what kind of channel you get and how it behaves in the test. If you can always buffer it with one element (but didn't check if that works with the test), I also prefer that over the goroutine.

Question about the code as it currently is, is whether there can be a race condition with the goroutine and a line in the test also passing in a block. This would make the ordering non-deterministic.

halseth: Ah, I see.  I think we can just buffer it with 1 always then, or just keep the goroutine. I'm fine with either.

sweep/sweeper.go
guggero: Fixed.

sweep/test_utils.go
guggero: I was able to change it back to buffer size 1 and remove the goroutine without any impact. Apparently the mock notifier hasn't really been used for a test before or didn't expect a specific buffer size.

sweep/txgenerator_test.go
guggero: Yes, that makes sense. Testing every single type doesn't add much value. So I picked a few that use CSV and CLTV and moved the commit with the test to the beginning. That way we can be assert that the estimation isn't changed by the refactor.

joostjager: Looking good

input/witnessgen.go
```diff
@@ -262,5 +294,116 @@ func (wt WitnessType) GenWitnessFunc(signer Signer,
...
+		estimator.AddNestedP2WSHInput(size)
+	} else {
+		estimator.AddWitnessInput(size)
+	}
+
+	switch wt {
+	case CommitmentTimeLock,
+		HtlcOfferedTimeoutSecondLevel,
+		HtlcAcceptedSuccessSecondLevel:
+		csvCount++
```
joostjager: Instead of the `csvCount++`, this could also just be `return 0, 1, nil`. But really, this is still not what you want probably. What is actually needed is a  `TimeLockType()` method returning `Cltv`, `Csv` or `None`. But this is pre-existing and optional to fix.

cfromknecht:
discussed offline, @guggero to follow up on @joostjager's suggestion for refactoring the timelocked output counting

PR #3558, opened 2019-09-30
routing: non-strict path finding

joostjager:
This pr makes path finding (and build route) aware of non-strict forwarding. Before deciding on a path edge, it calculates a single unified policy for the complete set of channels between two nodes.

### DIFF
--- a/lnwire/msat.go
+++ b/lnwire/msat.go
@@ -6,9 +6,15 @@ import (
 	"github.com/btcsuite/btcutil"
 )
 
-// mSatScale is a value that's used to scale satoshis to milli-satoshis, and
-// the other way around.
-const mSatScale uint64 = 1000
+const (
+	// mSatScale is a value that's used to scale satoshis to milli-satoshis, and
+	// the other way around.
+	mSatScale uint64 = 1000
+
+	// MaxMilliSatoshi is the maximum number of msats that can be expressed
+	// in this data type.
+	MaxMilliSatoshi = ^MilliSatoshi(0)
+)
 
 // MilliSatoshi are the native unit of the Lightning Network. A milli-satoshi
 // is simply 1/1000th of a satoshi. There are 1000 milli-satoshis in a single
--- a/routing/unified_policies.go
+++ b/routing/unified_policies.go
@@ -267,3 +267,15 @@ func (u *unifiedPolicy) getPolicyNetwork(
 
 	return &modifiedPolicy
 }
+
+// minAmt returns the minimum amount that can be forwarded on this connection.
+func (u *unifiedPolicy) minAmt() lnwire.MilliSatoshi {
+	min := lnwire.MaxMilliSatoshi
+	for _, edge := range u.edges {
+		if edge.policy.MinHTLC < min {
+			min = edge.policy.MinHTLC
+		}
+	}
+
+	return min
+}
--- a/routing/unified_policies.go
+++ b/routing/unified_policies.go
@@ -1,0 +1,269 @@
+package routing
+
+import (
+	"github.com/btcsuite/btcutil"
+	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// unifiedPolicies holds all unified policies for connections towards a node.
+type unifiedPolicies struct {
+	// policies contains a unified policy for every from node.
+	policies map[route.Vertex]*unifiedPolicy
+
+	// sourceNode is the sender of a payment. The rules to pick the final
+	// policy are different for local channels.
+	sourceNode route.Vertex
+
+	// toNode is the node for which the unified policies are instantiated.
+	toNode route.Vertex
+
+	// outChanRestr is an optional outgoing channel restriction for the
+	// local channel to use.
+	outChanRestr *uint64
+}
+
+// newUnifiedPolicies instantiates a new unifiedPolicies object. Channel
+// policies can be added to this object.
+func newUnifiedPolicies(sourceNode, toNode route.Vertex,
+	outChanRestr *uint64) *unifiedPolicies {
+
+	return &unifiedPolicies{
+		policies:     make(map[route.Vertex]*unifiedPolicy),
+		toNode:       toNode,
+		sourceNode:   sourceNode,
+		outChanRestr: outChanRestr,
+	}
+}
+
+// addPolicy adds a single channel policy. Capacity may be zero if unknown
+// (light clients).
+func (u *unifiedPolicies) addPolicy(fromNode route.Vertex,
+	edge *channeldb.ChannelEdgePolicy, capacity btcutil.Amount) {
+
+	localChan := fromNode == u.sourceNode
+
+	// Skip channels if there is an outgoing channel restriction.
+	if localChan && u.outChanRestr != nil &&
+		*u.outChanRestr != edge.ChannelID {
+
+		return
+	}
+
+	// Update the policies map.
+	policy, ok := u.policies[fromNode]
+	if !ok {
+		policy = &unifiedPolicy{
+			localChan: localChan,
+		}
+		u.policies[fromNode] = policy
+	}
+
+	policy.edges = append(policy.edges, &unifiedPolicyEdge{
+		policy:   edge,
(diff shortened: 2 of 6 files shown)


routing/unified_policy.go
```diff
@@ -0,0 +1,85 @@
+package routing
+
+import (
+	"github.com/go-errors/errors"
```
cfromknecht: use `"errors"` from std lib, a few instances in this pr

routing/unified_policy.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	if policy.FeeProportionalMillionths > u.feeProportionalMillionths {
+		u.feeProportionalMillionths = policy.FeeProportionalMillionths
+	}
+
+	// For htlc amount constraints, take the widest range.
+	if policy.MinHTLC > u.minHTLC {
+		u.minHTLC = policy.MinHTLC
+	}
+	if maxHTLC > u.maxHTLC {
```
cfromknecht: the min and max both get bigger?

routing/unified_policy.go
```diff
@@ -0,0 +1,85 @@
...
+	// For fees, set a fee schedule that meets all distinct channel
+	// policies.
+	if policy.FeeBaseMSat > u.feeBaseMSat {
+		u.feeBaseMSat = policy.FeeBaseMSat
+	}
+	if policy.FeeProportionalMillionths > u.feeProportionalMillionths {
+		u.feeProportionalMillionths = policy.FeeProportionalMillionths
+	}
+
+	// For htlc amount constraints, take the widest range.
```
cfromknecht: are we not trying to take the narrowest range, as to be able to satisfy any of the links? maybe we are thinking about the widths differently

routing/unified_policy.go
```diff
@@ -0,0 +1,85 @@
...
+	if policy.MinHTLC > u.minHTLC {
+		u.minHTLC = policy.MinHTLC
+	}
+	if maxHTLC > u.maxHTLC {
+		u.maxHTLC = maxHTLC
```
cfromknecht: should it be an error if `maxHTLC` is lower than `minHTLC`?

if i have one big channel and one little channel, these regions may never overlap. i wonder, should these be defined over amt ranges?

routing/unified_policy.go
```diff
@@ -0,0 +1,85 @@
...
+
+	edgePolicy
+}
+
+// update adds a new channel policy to the unified policy. This method can be
+// called on a nil pointer to initialize a unified policy struct. The maxHTLC
+// parameter is the best known upper bound for bandwidth. This can be an
+// accurate estimate, the channel capacity or, when no information is available,
+// lnwire.MaxMilliSatoshi.
+func (u *unifiedPolicy) update(maxHTLC lnwire.MilliSatoshi,
```
cfromknecht: should this also take the base payment amount as an input? if the amount being paid is greater than the channel capacity, do we need to include it's parameters in the GCD? we won't know how much will be added from fees, but may help to prune early

lnwire/msat.go
```diff
@@ -6,9 +6,15 @@ import (
...
-// the other way around.
-const mSatScale uint64 = 1000
+const (
+	// mSatScale is a value that's used to scale satoshis to milli-satoshis, and
+	// the other way around.
+	mSatScale uint64 = 1000
+
+	// MaxMilliSatoshi is the maximum number of msats that can be expressed
+	// in this data type.
+	MaxMilliSatoshi = ^MilliSatoshi(0)
```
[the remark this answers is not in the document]
joostjager: ltc...?

routing/unified_policy.go
joostjager: fixed

joostjager:
Did a quick check how prevalent multiple (advertized) channels between a pair of nodes at the moment. About 7.5% of the node pairs have multiple channels. This doesn't say much though about how often multi-channel node connections are traversed during payments.

`lncli describegraph | jq '.edges[] | .node1_pub + " " + .node2_pub' -r | sort | uniq -c | awk '{print $1}' | sort | uniq -c`

nof channels | node pair count
--- | ---
1 | 29093
2 | 1834
3 | 367
4 | 102
5 | 23
6 | 10
7 | 3
8 | 2
9 | 5
10 | 1
12 | 1
15 | 1
17 | 1
18 | 1
29 | 1
35 | 1
105 | 1
150 | 1

joostjager:
Results when filtered for channels between bos list nodes is 20%.

nof channels | node pair count
--- | ---
1 | 3015
2 | 427
3 | 252
4 | 73
5 | 15
6 | 2
7 | 1
15 | 1
17 | 1

routing/unified_policy.go
joostjager: Went for a different approach where I don't squash all policies together, but return the best based on the actual amount.

routing/unified_policy.go
joostjager: I took a different approach where I don't squash all policies together. In stead, the optimal policy is returned based on the actual amount.

routing/unified_policy.go
joostjager: Outdated

routing/unified_policy.go
joostjager: In the new approach, the amt is used to prune.

routing/unified_policies.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil
+	}
+
+	// We have already picked the highest fee that could be required for
+	// non-strict forwarding. To also cover the case where a lower fee
+	// channel requires a longer time lock, we modify the policy by setting
+	// the maximum encountered time lock. Note that this results in a
+	// synthetic policy that is not actually present on the routing node.
+	modifiedPolicy := *bestPolicy
+	modifiedPolicy.TimeLockDelta = maxTimelock
```
wpaulino: If I understand correctly, this is done so that there isn't an additional round trip in the payment process in the event that the channel yielding the highest fee does not have sufficient bandwidth to carry the payment but the highest timelock one does?

joostjager: Yes indeed. We try to maximize the chance that we get forwarded. Because we penalize pair-wise, there won't be a second chance for this node pair. But this all really only exists for nodes that have distinct policies for channels to the same peer. It seems unlikely it is configured like that intentionally. Currently it does happen on the network after upgrading to a `lnd` version that changed the default time lock delta. The new delta will be applied to newly opened channels.

routing/unified_policies.go
```diff
@@ -0,0 +1,263 @@
...
+type unifiedPolicyEdge struct {
+	policy   *channeldb.ChannelEdgePolicy
+	capacity btcutil.Amount
+}
+
+// amtInRange checks whether an amount falls within the valid range for a
+// channel.
+func (u *unifiedPolicyEdge) amtInRange(amt lnwire.MilliSatoshi) bool {
+	// Skip channels that are too small.
+	if u.capacity > 0 &&
```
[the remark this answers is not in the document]
joostjager: Added

routing/unified_policies.go
wpaulino: Nice, I think including this insight in the comment would be helpful.

wpaulino:
Travis failed with:

```
--- FAIL: TestChannelLinkWaitForRevocation (10.99s)
    link_test.go:4615: expected RevokeAndAck, got *lnwire.CommitSig
```

Looks unrelated, but is it known?

joostjager:
> Travis failed with:
> 
> ```
> --- FAIL: TestChannelLinkWaitForRevocation (10.99s)
>     link_test.go:4615: expected RevokeAndAck, got *lnwire.CommitSig
> ```
> 
> Looks unrelated, but is it known?

Looks unrelated indeed. Maybe caused by the log commit ticker... Could be fixed by #2927

routing/unified_policies.go
joostjager: Included

routing/unified_policies.go
```diff
@@ -261,3 +261,15 @@ func (u *unifiedPolicy) getPolicyNetwork(
 
 	return &modifiedPolicy
 }
+
+// minAmt returns the minimum amount that can be forwarded on this connection.
+func (u *unifiedPolicy) minAmt() lnwire.MilliSatoshi {
+	min := ^lnwire.MilliSatoshi(0)
```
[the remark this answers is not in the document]
joostjager: fixed

PR #3560, opened 2019-09-30
WIP: Replace Deprecated gprc.WithTimeout()

alrs:
The option `grpc.WithTimeout()` is deprecated, and suggests replacement with `context.WithTimeout()`.


alrs:
This is worth coming back to, but this WIP need not clog up the PR queue.

PR #3561, opened 2019-10-01
chanvalidate: create new channel validation package 

Roasbeef:
In this commit, we create a new `chanvalidate` package which is to house
all logic required for 1st and 3rd party channel verification. 1st party
verification occurs when we find a channel in the chain that is
allegedly ours, while 3rd party verification will occur when a peer
sends us a channel proof of a new channel.

In the scope of the recent CVE, we actually fully verified 3rd party
channels, but failed to also include those checks in our 1st party
verification code. In order to unify this logic, and prevent future
issues, in this PR we move to concentrate all validation logic into a
single function. Both 1st and 3rd party validation will then use this
function. Additionally, having all the logic in a single place makes it
easier to audit, and also write tests against.

### DIFF
--- a/lnwallet/chanvalidate/validate.go
+++ b/lnwallet/chanvalidate/validate.go
@@ -143,7 +143,7 @@ type Context struct {
 	FundingTx *wire.MsgTx
 
 	// CommitCtx is an optional additional set of validation context
-	// required to validate a self-owned channel. If present, then fully
+	// required to validate a self-owned channel. If present, then a full
 	// Script VM validation will be performed.
 	CommitCtx *CommitmentContext
 }
@@ -184,7 +184,7 @@ func Validate(ctx *Context) (*wire.OutPoint, error) {
 	}
 
 	// If we reach this point, then all other checks have succeeded, so
-	// we'll now attempt fully Script VM execution to ensure that we're
+	// we'll now attempt a full Script VM execution to ensure that we're
 	// able to close the channel using this initial state.
 	vm, err := txscript.NewEngine(
 		ctx.MultiSigPkScript, ctx.CommitCtx.FullySignedCommitTx,
--- a/lnwallet/chanvalidate/validate.go
+++ b/lnwallet/chanvalidate/validate.go
@@ -1,0 +1,205 @@
+package chanvalidate
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+var (
+	// ErrInvalidOutPoint is returned when the ChanLocator is unable to
+	// find the target outpoint.
+	ErrInvalidOutPoint = fmt.Errorf("output meant to create channel cannot " +
+		"be found")
+
+	// ErrWrongPkScript is returned when the alleged funding transaction is
+	// found to have an incorrect pkSript.
+	ErrWrongPkScript = fmt.Errorf("wrong pk script")
+
+	// ErrInvalidSize is returned when the alleged funding transaction
+	// output has the wrong size (channel capacity).
+	ErrInvalidSize = fmt.Errorf("channel has wrong size")
+)
+
+// ErrScriptValidateError is returned when Script VM validation fails for an
+// alleged channel output.
+type ErrScriptValidateError struct {
+	err error
+}
+
+// Error returns a human readable string describing the error.
+func (e *ErrScriptValidateError) Error() string {
+	return fmt.Sprintf("script validation failed: %v", e.err)
+}
+
+// Unwrap returns the underlying wrapped VM execution failure error.
+func (e *ErrScriptValidateError) Unwrap() error {
+	return e.err
+}
+
+// ChanLocator abstracts away obtaining the output that created the channel, as
+// well as validating its existence given the funding transaction.  We need
+// this as there are several ways (outpoint, short chan ID) to identify the
+// output of a channel given the funding transaction.
+type ChanLocator interface {
+	// Locate attempts to locate the funding output within the funding
+	// transaction. It also returns the final out point of the channel
+	// which uniquely identifies the output which creates the channel. If
+	// the target output cannot be found, or cannot exist on the funding
+	// transaction, then an error is to be returned.
+	Locate(*wire.MsgTx) (*wire.TxOut, *wire.OutPoint, error)
+}
+
+// OutPointChanLocator is an implementation of the ChanLocator that can be used
+// when one already knows the expected chan point.
+type OutPointChanLocator struct {
+	// ChanPoint is the expected chan point.
+	ChanPoint wire.OutPoint
+}
+
+// Locate attempts to locate the funding output within the passed funding
+// transaction.
+//
+// NOTE: Part of the ChanLocator interface.
+func (o *OutPointChanLocator) Locate(fundingTx *wire.MsgTx) (
+	*wire.TxOut, *wire.OutPoint, error) {
+
+	// If the expected index is greater than the amount of output in the
+	// transaction, then we'll reject this channel as it's invalid.
+	if int(o.ChanPoint.Index) >= len(fundingTx.TxOut) {
+		return nil, nil, ErrInvalidOutPoint
+	}
+
+	// As an extra sanity check, we'll also ensure the txid hash matches.
+	fundingHash := fundingTx.TxHash()
+	if !bytes.Equal(fundingHash[:], o.ChanPoint.Hash[:]) {
+		return nil, nil, ErrInvalidOutPoint
+	}
+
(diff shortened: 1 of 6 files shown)


lnwallet/chanvalidate/validate.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// The scripts should match up exactly, otherwise the channel is
+	// invalid.
+	fundingScript := fundingOutput.PkScript
+	if !bytes.Equal(ctx.MultiSigPkScript, fundingScript) {
+		return nil, ErrWrongPkScript
+	}
+
+	// If there's no commitment context, then we're done here as this is a
```
halseth: Rather than having an optional commitment context, it would make sense to expose two methods:
`ValidateFundingOutput` and `ValidateCommitment`, where the latter would be the same as this current method, while the first would only consist of the parts up to the optional context (`ValidateCommitment` could just call `ValidateFundingOutput` for its first part, or the caller could just call each in succession).

That would make it less confusing to see that the two uses of this method is doing different validation, and avoids the awkward use of optional fields.

lnwallet/chanvalidate/validate.go
```diff
@@ -0,0 +1,199 @@
...
+func (e *ErrScriptValidateError) Error() string {
+	return fmt.Sprintf("script validation failed: %v", e.err)
+}
+
+// Unwrap returns the underlying wrapped VM execution failure error.
+func (e *ErrScriptValidateError) Unwrap() error {
+	return e.err
+}
+
+// ChanLocator abstracts away obtaining as well as validating output that
```
halseth: It doesn't actually validate anything, so I think that should be removed from the doc.

lnwallet/chanvalidate/validate.go
```diff
@@ -0,0 +1,199 @@
...
+func (o *OutPointChanLocator) Locate(fundingTx *wire.MsgTx) (
+	*wire.TxOut, *wire.OutPoint, error) {
+
+	// If the expected index is greater than the amount of output in the
+	// transaction, then we'll reject this channel as it's invalid.
+	if int(o.ChanPoint.Index) >= len(fundingTx.TxOut) {
+		return nil, nil, ErrInvalidOutPoint
+	}
+
+	return fundingTx.TxOut[o.ChanPoint.Index], &o.ChanPoint, nil
```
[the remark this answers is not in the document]
Roasbeef: For the outpoint based locator we know the `txid` and when it's specified, the commit context is as well. Added the check in any case as our environment in the future might change.

lnwallet/chanvalidate/validate.go
Roasbeef: I prefer a single method as this means there's a single code path in either scenario where we can focus all tests/review. I don't see how having the context is confusing: it simply gates two extra validation steps. If we want a "proper" optional field, it could very well use a set of functional arguments.

lnwallet/chanvalidate/validate.go
Roasbeef: It validates the existence of the outpoint.

lnwallet/chanvalidate/validate.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// Now that we know this is our channel, we'll verify the amount of the
+	// created output against our expected size of the channel.
+	fundingValue := fundingOutput.Value
+	if btcutil.Amount(fundingValue) != ctx.CommitCtx.Value {
+		return nil, ErrInvalidSize
+	}
+
+	// If we reach this point, then all other checks have succeeded, so
+	// we'll now attempt fully Script VM execution to ensure that we're
```
cfromknecht:
```suggestion
	// we'll now attempt a full Script VM execution to ensure that we're
```

lnwallet/chanvalidate/validate.go
```diff
@@ -0,0 +1,205 @@
...
+
+	return fundingTx.TxOut[o.ChanPoint.Index], &o.ChanPoint, nil
+}
+
+// ShortChanIDChanLocator is an implementation of the ChanLocator that can be
+// used when one only knows the short channel ID of a channel. This should be
+// used in contexts when one is verifying a 3rd party channel.
+type ShortChanIDChanLocator struct {
+	// Cid is the short channel ID of the target channel.
+	Cid lnwire.ShortChannelID
```
cfromknecht: naming suggestion: change field name to `ID` since the name is already `ShortChanIDChanLocator` and `Cid` makes me think of `lnwire.ChannelID`?

lnwallet/chanvalidate/validate.go
cfromknecht: I agree that in either case only one function should be called, but I think we can achieve that and still have one validation path. For example, we can expose two distinct methods for each case, that simply wrap the existing `Validate` method and populate the context:

```golang
func ValidateRemote(lnwire.ShortChannelID, []byte, *wire.MsgTx) (*wire.OutPoint, error) {
    return Validate(...)
} 

func ValidateLocal(wire.OutPoint, []byte, *wire.MsgTx, btcutil.Amount, *wire.MsgTx) error {
    _, err := Validate(...)
    return err
} 
```
Would be less verbose, less prone to errors, and more readable at the call site imo. From purely an API/library perspective (if others go on to use this package), I think it's be preferable to avoid the optional field and enforce its presence/absence via a function signature. That being said, we only call this in two locations (+1 for tests) so it wouldn't be saving ourselves very much.

Roasbeef: Yeah there're two call sites right now, only the wallet call site requires more set up as we need to obtain a signed commitment transaction. I'm not sure the local/remote function really achieves much since the main `Validate` method would still be there. We more or less already have `ValidateLocal` as the funding manager calls `wallet.ValidateChannel` and passes in information similar to your draft function signature above. Also just to put this in perspective, without comments, we're talking about literally 30 lines of code here (the core validate method). We can make a remote validation method for the router, but then that's just more code, making the diff bigger than it needs to be.

routing/router.go
```diff
@@ -1339,47 +1349,37 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
...
 
 	// As a sanity check, ensure that the advertised transaction index is
 	// within the bounds of the total number of transactions within a
 	// block.
 	numTxns := uint32(len(fundingBlock.Transactions))
 	if chanID.TxIndex > numTxns-1 {
-		return nil, nil, fmt.Errorf("tx_index=#%v is out of range "+
-			"(max_index=%v), network_chan_id=%v\n", chanID.TxIndex,
+		return nil, fmt.Errorf("tx_index=#%v is out of range "+
+			"(max_index=%v), network_chan_id=%v", chanID.TxIndex,
```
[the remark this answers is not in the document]
Roasbeef: Removed.

lnwallet/chanvalidate/validate.go
Roasbeef: Yehh I agree, done.

lnwallet/chanvalidate/validate.go
halseth: I agree in this case it's not that big a deal, since we only call this in two places, and the validation logic is fairly simple. But as a general principle I prefer having the fields clearly enforced using a method signature instead of optional fields on a struct, since they tend to be forgotten about and harder to reason about as the struct grows larger.

lnwallet/chanvalidate/validate.go
halseth: Change to 
```go
// ChanLocator abstracts away obtaining the output that created the channel, as well as validating its existence given the funding transaction.
```
perhaps?

lnwallet/chanvalidate/validate.go
```diff
@@ -88,7 +88,7 @@ func (o *OutPointChanLocator) Locate(fundingTx *wire.MsgTx) (
 // used in contexts when one is verifying a 3rd party channel.
 type ShortChanIDChanLocator struct {
 	// Cid is the short channel ID of the target channel.
-	Cid lnwire.ShortChannelID
+	ID lnwire.ShortChannelID
```
halseth: should be squashed with the OG commit.

lnwallet/chanvalidate/validate_test.go
```diff
@@ -179,7 +179,7 @@ func TestValidate(t *testing.T) {
 		{
 			expectedErr: ErrInvalidOutPoint,
 			locator: &ShortChanIDChanLocator{
-				Cid: lnwire.NewShortChanIDFromInt(9),
+				ID: lnwire.NewShortChanIDFromInt(9),
```
halseth: same, squash

routing/router.go
```diff
@@ -1189,14 +1190,33 @@ func (r *ChannelRouter) processUpdate(msg interface{}) error {
 		if err != nil {
 			return err
 		}
-		fundingPkScript, err := input.WitnessScriptHash(witnessScript)
+		pkScript, err := input.WitnessScriptHash(witnessScript)
```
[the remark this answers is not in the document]
Roasbeef: The two fragments are a logical stanza IMO.

Issue #3553, opened 2019-09-28
autopilot/fundingmanager nil pointer dereference


qubenix:
### Background

After upgrading to v0.8.0-beta-rc1 my node will crash when using autopilot. Running without autopilot does not crash.

### Your environment

* version of `lnd`
0.8.0-beta commit=v0.8.0-beta-rc1
* which operating system (`uname -a` on *Nix)
Debian 10
* version of `btcd`, `bitcoind`, or other backend
v0.18.1.0-00ffe5aca
* any other relevant environment details

### Steps to reproduce

Run `lnd` with autopilot enabled.

### Expected behaviour

Doesn't crash.

### Actual behaviour

Does crash.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe19fc8]

goroutine 347 [running]:
github.com/lightningnetwork/lnd/lnwire.(*FeatureVector).HasFeature(...)
/home/user/gocode/src/github.com/lightningnetwork/lnd/lnwire/features.go:283
github.com/lightningnetwork/lnd.(*fundingManager).handleInitFundingMsg(0xc0004b5b00, 0xc002d56ee0)
/home/user/gocode/src/github.com/lightningnetwork/lnd/fundingmanager.go:2773 +0x418
github.com/lightningnetwork/lnd.(*fundingManager).reservationCoordinator(0xc0004b5b00)
/home/user/gocode/src/github.com/lightningnetwork/lnd/fundingmanager.go:753 +0x5ec
created by github.com/lightningnetwork/lnd.(*fundingManager).start
/home/user/gocode/src/github.com/lightningnetwork/lnd/fundingmanager.go:548 +0x562
```

vwoo:
I started a new lnd testnet node without autopilot and I'm running into the same problem.

lnd version 0.8.0-beta commit=v0.8.0-beta-rc1-dirty
MacOS 10.14.6
bitcoind v0.18.1

Open channel log is here: [funding_log.txt](https://github.com/lightningnetwork/lnd/files/3666572/funding_log.txt)

The tail of the log:

```
lncli openchannel --node_key 03d5e17a3c213fe490e1b0c389f8cfcfcea08a29717d50a9f453735e0ab2a7c003 --connect 3.16.119.191:9735 --local_amt 1000000 --push_amt 500000
{
[...]

PR #3566, opened 2019-10-01
peer+server: wait until peer is active to begin channel opening process

wpaulino:
Without waiting, we would proceed to retrieve the remote peer's supported features, which may have not been set due to not yet receiving their Init message.

Fixes https://github.com/lightningnetwork/lnd/issues/3553.

### DIFF
--- a/server.go
+++ b/server.go
@@ -3207,6 +3207,18 @@ func (s *server) OpenChannel(
 	}
 	s.mu.RUnlock()
 
+	// We'll wait until the peer is active before beginning the channel
+	// opening process.
+	select {
+	case <-peer.activeSignal:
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -3207,6 +3207,18 @@ func (s *server) OpenChannel(
 	}
 	s.mu.RUnlock()
 
+	// We'll wait until the peer is active before beginning the channel
+	// opening process.
+	select {
+	case <-peer.activeSignal:
```
cfromknecht: we should avoid adding more places where we access internal members of the peer, but this works for now :) we can defer that to when we move the peer into its own package

PR #3569, opened 2019-10-02
chainntnfs: fix nil pointer on address reuse

guggero:
When a confirmation notification for a script hash is registered, the notifier panics with a nil pointer dereference in case two transactions send funds to this script and are confirmed at different block heights.

The test that triggers the panic is added in the first commit with a fix for it in the second commit.

### DIFF
--- a/chainntnfs/txnotifier.go
+++ b/chainntnfs/txnotifier.go
@@ -103,6 +103,9 @@ type confNtfnSet struct {
 	// details serves as a cache of the confirmation details of a
 	// transaction that we'll use to determine if a transaction/output
 	// script has already confirmed at the time of registration.
+	// details is also used to make sure that in case of an address reuse
+	// (funds sent to a previously confirmed script) no additional
+	// notification is registered which would lead to an inconsistent state.
 	details *TxConfirmation
 }
 
@@ -1507,6 +1510,15 @@ func (n *TxNotifier) handleConfDetailsAtTip(confRequest ConfRequest,
 
 	// TODO(wilmer): cancel pending historical rescans if any?
 	confSet := n.confNotifications[confRequest]
+
+	// If we already have details for this request, we don't want to add it
+	// again since we have already dispatched notifications for it.
+	if confSet.details != nil {
+		Log.Warnf("Ignoring address reuse for %s at height %d.",
+			confRequest, details.BlockHeight)
+		return
+	}
+
 	confSet.rescanStatus = rescanComplete
 	confSet.details = details
 
--- a/chainntnfs/txnotifier_test.go
+++ b/chainntnfs/txnotifier_test.go
@@ -604,6 +604,166 @@ func TestTxNotifierFutureSpendDispatch(t *testing.T) {
 	}
 }
 
+// TestTxNotifierFutureConfDispatchReuseSafe tests that the notifier does not
+// misbehave even if two confirmation requests for the same script are issued
+// at different block heights (which means funds are being sent to the same
+// script multiple times).
+func TestTxNotifierFutureConfDispatchReuseSafe(t *testing.T) {
+	t.Parallel()
+
+	currentBlock := uint32(10)
+	hintCache := newMockHintCache()
+	n := chainntnfs.NewTxNotifier(
+		currentBlock, 2, hintCache, hintCache,
+	)
+
+	// We'll register a TX that sends to our test script and put it into a
+	// block. Additionally we register a notification request for just the
+	// script which should also be confirmed with that block.
+	tx1 := wire.MsgTx{Version: 1}
+	tx1.AddTxOut(&wire.TxOut{PkScript: testRawScript})
+	tx1Hash := tx1.TxHash()
+	ntfn1, err := n.RegisterConf(&tx1Hash, testRawScript, 1, 1)
+	if err != nil {
+		t.Fatalf("unable to register ntfn: %v", err)
+	}
+	scriptNtfn1, err := n.RegisterConf(nil, testRawScript, 1, 1)
+	if err != nil {
+		t.Fatalf("unable to register ntfn: %v", err)
+	}
+	block := btcutil.NewBlock(&wire.MsgBlock{
+		Transactions: []*wire.MsgTx{&tx1},
+	})
+	currentBlock++
+	err = n.ConnectTip(block.Hash(), currentBlock, block.Transactions())
+	if err != nil {
+		t.Fatalf("unable to connect block: %v", err)
+	}
+	if err := n.NotifyHeight(currentBlock); err != nil {
+		t.Fatalf("unable to dispatch notifications: %v", err)
+	}
+
+	// Expect an update and confirmation of TX 1 at this point. We save the
+	// confirmation details because we expect to receive the same details
+	// for all further registrations.
+	var confDetails *chainntnfs.TxConfirmation
+	select {
+	case <-ntfn1.Event.Updates:
+	default:
+		t.Fatal("expected update of TX 1")
+	}
+	select {
+	case confDetails = <-ntfn1.Event.Confirmed:
+		if confDetails.BlockHeight != currentBlock {
+			t.Fatalf("expected TX to be confirmed in latest block")
+		}
+	default:
+		t.Fatal("expected confirmation of TX 1")
+	}
+
+	// The notification for the script should also have received a
+	// confirmation.
+	select {
+	case <-scriptNtfn1.Event.Updates:
+	default:
+		t.Fatal("expected update of script ntfn")
+	}
+	select {
+	case details := <-scriptNtfn1.Event.Confirmed:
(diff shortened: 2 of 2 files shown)


chainntnfs/txnotifier_test.go
```diff
@@ -604,6 +605,112 @@ func TestTxNotifierFutureSpendDispatch(t *testing.T) {
 	}
 }
 
+// TestTxNotifierFutureConfDispatchReuseSafe tests that the notifier does not
```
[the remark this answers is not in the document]
guggero: Right, makes sense.

chainntnfs/txnotifier_test.go
```diff
@@ -604,6 +605,112 @@ func TestTxNotifierFutureSpendDispatch(t *testing.T) {
...
+			t.Fatalf("unable to connect block: %v", err)
+		}
+		if err := n.NotifyHeight(currentBlock); err != nil {
+			t.Fatalf("unable to dispatch notifications: %v", err)
+		}
+	}
+
+	// If we reach this point without panic, the bug this test was written
+	// for no longer exists.
+	if !status1.confirmed || !status2.confirmed {
```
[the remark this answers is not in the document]
guggero: Yes, I just saw the race test failure today.
We need to read multiple events from the channels because there will be multiple confirmations the way it is implemented at the moment. So I solved it with wait groups for now.

But maybe that's another reason to go with the other implementation where we clean out all scripts that have been confirmed and therefore would only send one confirmation event.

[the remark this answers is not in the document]
halseth: Yeah, I think that's the way it should be fixed.

guggero:
I now implemented the version that will not notify again on reuse.
Turns out, it's also less code.
Tested it with my regtest setup and was unable to reproduce it after the fix.

The unit test setup is still a bit peculiar with the event listener in the goroutine since we still get multiple messages on the `Event.Update` channel.
The reason for this is that because both `confRequest`s are identical, there are two `ntfns` created for the same request and they both get notified. And since the request is referenced from two block heights, we get a total of 4 `Update` messages (two per TX).
And because the chan buffer is only 1, the test would block if we didn't read in a goroutine.

Not sure if this can still cause problems on the receiver end?

wpaulino:
> The reason for this is that because both confRequests are identical, there are two ntfns created for the same request and they both get notified. And since the request is referenced from two block heights, we get a total of 4 Update messages (two per TX).

I think we can get around this by adding an additional check to ensure no details exist for the notification before notifying the transaction at:
https://github.com/lightningnetwork/lnd/blob/6765b8668a2a08f58b38480fff3534fcfa4c001e/chainntnfs/txnotifier.go#L1494

guggero:
I was able to get around the problem by not notifying `ntfn`s that have already been marked as `dispatched`. Now the test is simpler too.

guggero:
Ok, I think I was finally able to fix it the way you suggested, @wpaulino.

chainntnfs/txnotifier.go
```diff
@@ -633,7 +633,34 @@ func (n *TxNotifier) RegisterConf(txid *chainhash.Hash, pkScript []byte,
 		confSet = newConfNtfnSet()
 		n.confNotifications[ntfn.ConfRequest] = confSet
 	}
-	confSet.ntfns[ntfn.ConfID] = ntfn
+
+	// In case of an address reuse, we would get multiple confirmation
```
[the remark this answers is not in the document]
wpaulino: Yeah that's what I also had in mind, adding the small check in `handleConfDetailsAtTip`. I don't think any changes need to be done for registrations, the assumption there is that it'll dispatch on the first transaction that satisfies it starting from the initial height hint.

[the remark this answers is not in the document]
guggero: Ok, it would have been very simple if I wasn't looking in the wrong place the whole time... Sorry about that.
This indeed doesn't make it necessary to change anything with the registrations.

chainntnfs/txnotifier_test.go
```diff
@@ -604,6 +604,112 @@ func TestTxNotifierFutureSpendDispatch(t *testing.T) {
...
+	}
+	select {
+	case <-ntfn1.Event.Confirmed:
+	default:
+		t.Fatal("expected confirmation of TX 1")
+	}
+
+	// Now register a second TX that spends to the same output and get it
+	// confirmed. This TX needs to have a different TX hash, otherwise
+	// we'd get strange results. So we tweak one field in the TxIn.
```
wpaulino: Does this comment still apply? `tx2` looks the same as `tx1`. I think we do want different txids here.

chainntnfs/txnotifier.go
```diff
@@ -1507,6 +1507,13 @@ func (n *TxNotifier) handleConfDetailsAtTip(confRequest ConfRequest,
 
 	// TODO(wilmer): cancel pending historical rescans if any?
 	confSet := n.confNotifications[confRequest]
+	
+	// If we already have details for this request, we don't want to add it
+	// again since we have already dispatched notifications for it.
+	if confSet.details != nil {
```
wpaulino: Commit message needs to be rewritten for this change.

chainntnfs/txnotifier_test.go
guggero: Comment was wrong, we want the same TX. Fixed.

chainntnfs/txnotifier.go
guggero: Done.

chainntnfs/txnotifier.go
```diff
@@ -1507,6 +1507,13 @@ func (n *TxNotifier) handleConfDetailsAtTip(confRequest ConfRequest,
 
 	// TODO(wilmer): cancel pending historical rescans if any?
 	confSet := n.confNotifications[confRequest]
+
+	// If we already have details for this request, we don't want to add it
+	// again since we have already dispatched notifications for it.
+	if confSet.details != nil {
+		return
```
halseth: add a logging statement?

chainntnfs/txnotifier_test.go
```diff
@@ -604,6 +604,110 @@ func TestTxNotifierFutureSpendDispatch(t *testing.T) {
...
+	case <-ntfn1.Event.Confirmed:
+	default:
+		t.Fatal("expected confirmation of TX 1")
+	}
+
+	// Now register a second TX that spends to the same output.
+	tx2 := wire.MsgTx{Version: 1}
+	tx2.AddTxOut(&wire.TxOut{PkScript: testRawScript})
+	tx2Hash := tx2.TxHash()
+	ntfn2, err := n.RegisterConf(&tx2Hash, testRawScript, 1, 1)
```
halseth: We should also test registering with only the script (no txid). In that case we expect to receive confirmation only for the first tx I think, shich we can check by inspecting the `TxConfirmtion`

chainntnfs/txnotifier.go
```diff
@@ -1507,6 +1507,13 @@ func (n *TxNotifier) handleConfDetailsAtTip(confRequest ConfRequest,
 
 	// TODO(wilmer): cancel pending historical rescans if any?
 	confSet := n.confNotifications[confRequest]
+
+	// If we already have details for this request, we don't want to add it
+	// again since we have already dispatched notifications for it.
+	if confSet.details != nil {
```
halseth: Maybe also update the godoc for the `detail` field as it is becoming more than just a cache at this point

chainntnfs/txnotifier.go
guggero: Done. Hope the wording and log level is adequate.

chainntnfs/txnotifier_test.go
guggero: Yes, definitely. Added the additional checks.

chainntnfs/txnotifier_test.go
```diff
@@ -604,6 +604,157 @@ func TestTxNotifierFutureSpendDispatch(t *testing.T) {
...
+	case details := <-scriptNtfn1.Event.Confirmed:
+		assertConfDetails(t, details, confDetails)
+	default:
+		t.Fatal("expected update of script ntfn")
+	}
+
+	// Now register a second TX that spends to the same output and again
+	// register a confirmation for just the script.
+	tx2 := wire.MsgTx{Version: 1}
+	tx2.AddTxOut(&wire.TxOut{PkScript: testRawScript})
```
wpaulino: Nit: maybe add another output to the transaction so that it results in a different txid?

chainntnfs/txnotifier.go
```diff
@@ -1507,6 +1510,17 @@ func (n *TxNotifier) handleConfDetailsAtTip(confRequest ConfRequest,
...
+		Log.Infof("Not registering another notification for %v. "+
```
wpaulino: Nit: would just reword this to `Ignoring address reuse for <script> at height <height>`.

PR #3572, opened 2019-10-03
multi: fix canceled spelling

joostjager:
Besides consistency, there are two non-comment changes. See below.


cmd/lncli/commands.go
```diff
@@ -2025,7 +2025,7 @@ func closedChannels(ctx *cli.Context) error {
 		LocalForce:      ctx.Bool("local_force"),
 		RemoteForce:     ctx.Bool("remote_force"),
 		Breach:          ctx.Bool("breach"),
-		FundingCanceled: ctx.Bool("funding_cancelled"),
+		FundingCanceled: ctx.Bool("funding_canceled"),
```
joostjager: An actual bug

PR #3574, opened 2019-10-03
htlcswitch: enforce fee floor on max fee allocation

wpaulino:
Without this, it was possible for a combination of our balance and max fee allocation to result in a fee rate below the fee floor causing the remote party to reject the update and close the channel.

### DIFF
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -5,6 +5,7 @@ import (
 	"container/list"
 	"crypto/sha256"
 	"fmt"
+	"math"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


htlcswitch/link.go
```diff
@@ -1041,12 +1041,16 @@ out:
...
 			// based on our current set fee rate. We'll cap the new
-			// fee rate to our max fee allocation.
+			// fee rate to our max fee allocation and ensure it's
+			// above our fee floor to prevent a channel failure.
 			commitFee := l.channel.CommitFeeRate()
 			maxFee := l.channel.MaxFeeRate(l.cfg.MaxFeeAllocation)
 			newCommitFee := lnwallet.SatPerKWeight(
 				math.Min(float64(netFee), float64(maxFee)),
 			)
+			if newCommitFee < lnwallet.FeePerKwFloor {
```
cfromknecht: would a better place for this be inside `MaxFeeRate`? I can see keeping this here as a sanity check, but really `MaxFeeRate` shouldn't propose anything lower than `lnwallet.FeePerKwFloor`

cfromknecht: would also be easier to test

Roasbeef: Or above you can do:
```go
max.Max(
    math.Min(float64(netFee), float64(maxFee)),
    lnwallet.FeePerKwFloor,
)
```

wpaulino: Good point, done.

PR #3575, opened 2019-10-03
lnrpc+routerrpc: disable custom tlv records via rpc

cfromknecht:
In order to prevent future unforeseen issues, we are temporarily
disabling the ability to send custom tlv records to the receiver of a
payment. Currently the receiver does not process or expose these
additional fields via rpc or internally, so they are being disabled
until the end-to-end flow is finished and fully validated.


Roasbeef:
> Only thing is what we discussed before, we are still writing a zero here

I don't think that's an issue, as we can still project any tlv values written in the payload within the main `Hop` struct. Usage of the TLV format on disk lets us decouple the in-memory representation from the on-disk representation. The same goes for any of the proto structs as well.

PR #3576, opened 2019-10-04
contractcourt: use legacy chain actions of commitSet not found

### LINKED ISSUE #3549: ChannelArbitrator panic: runtime error: invalid memory address or nil pointer dereference
### Background

My lnd node will no longer startup. The last part of the logs say:

```
2019-09-27 16:56:25.684 [INF] CNCT: ChannelArbitrator(ae48c9acaae2fda06bee0581256af6f9f51f499e90103f5a449e390bdfab009a:0): starting state=StateDefault
2019-09-27 16:56:25.691 [INF] CNCT: ChannelArbitrator(de1b7caba631137e23c96990e9cf85f89cd67eba7a0d6b9fe40ebb284ccb8b34:1): starting state=StateDefault
2019-09-27 16:56:25.697 [INF] CNCT: ChannelArbitrator(ea119bebcec989757157287918387b510deb34d0c1446306be9a15dc537fd6ba:1): starting state=StateWaitingFullResolution
2019-09-27 16:56:25.701 [INF] CNCT: ChannelArbitrator(ea119bebcec989757157287918387b510deb34d0c1446306be9a15dc537fd6ba:1): still awaiting contract resolution
2019-09-27 16:56:25.701 [INF] RPCS: Stopping RPC Server
2019-09-27 16:56:25.701 [INF] RPCS: Stopping RouterRPC Sub-RPC Server
2019-09-27 16:56:25.701 [INF] RPCS: Stopping SignRPC Sub-RPC Server
2019-09-27 16:56:25.701 [INF] RPCS: Stopping ChainRPC Sub-RPC Server
2019-09-27 16:56:25.701 [INF] RPCS: Stopping InvoicesRPC Sub-RPC Server
2019-09-27 16:56:25.701 [INF] RPCS: Stopping WalletKitRPC Sub-RPC Server
2019-09-27 16:56:25.701 [INF] RPCS: Stopping WatchtowerClientRPC Sub-RPC Server
2019-09-27 16:56:25.705 [INF] LTND: Shutdown complete
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16bc19d]

goroutine 1 [running]:
github.com/lightningnetwork/lnd/contractcourt.(*ChannelArbitrator).Start(0xc001064000, 0x0, 0x0)
	/Users/vwoo/gocode/src/github.com/lightningnetwork/lnd/contractcourt/channel_arbitrator.go:429 +0x66d
github.com/lightningnetwork/lnd/contractcourt.(*ChainArbitrator).Start(0xc0000bf680, 0x0, 0x0)
	/Users/vwoo/gocode/src/github.com/lightningnetwork/lnd/contractcourt/chain_arbitrator.go:543 +0x12e2
github.com/lightningnetwork/lnd.(*server).Start.func1()
	/Users/vwoo/gocode/src/github.com/lightningnetwork/lnd/server.go:1238 +0x40b
sync.(*Once).doSlow(0xc000478008, 0xc0003375f0)
	/usr/local/Cellar/go/1.13/libexec/src/sync/once.go:66 +0xe3
sync.(*Once).Do(...)
	/usr/local/Cellar/go/1.13/libexec/src/sync/once.go:57
github.com/lightningnetwork/lnd.(*server).Start(0xc000478000, 0x1e3f217, 0x2e)
	/Users/vwoo/gocode/src/github.com/lightningnetwork/lnd/server.go:1174 +0x95
github.com/lightningnetwork/lnd.Main(0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/vwoo/gocode/src/github.com/lightningnetwork/lnd/lnd.go:620 +0x1b01
main.main()
	/Users/vwoo/gocode/src/github.com/lightningnetwork/lnd/cmd/lnd/main.go:14 +0x32
```

Full startup log is here: [log.txt](https://github.com/lightningnetwork/lnd/files/3664876/log.txt)

I had just recompiled lnd with the HEAD of master. I tried recompiling after checking out a commit from a couple of weeks ago to roughly where it last worked but it still won't startup.

### Your environment

* lnd @ git commit 7fcac54dffcc97964db22b0d8eaf31a9d751b76c, Sep 26, 2019
* macOS 10.14.6
* bitcoind v0.18.1
* bitcoin network: testnet

### Steps to reproduce

cd $GOPATH/src/github.com/lightningnetwork/lnd && git pull && make install tags="signrpc walletrpc chainrpc invoicesrpc routerrpc"
lnd
lncli unlock

Roasbeef:
In this commit, we fix a bug that would prevent users that had
unresolved contracts at the time of update from starting their nodes.
Before we added the conf commit set, the information needed to
supplement the resolvers was found in the chain action map. As a result,
if the conf commit set is nil, then we also need to check this legacy
data to ensure that we can supplement the resolvers to the best of our
ability based on the available data.

Fixes #3549.

### DIFF
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -426,8 +426,7 @@ func (c *ChannelArbitrator) Start() error {
 		// receive a chain event from the chain watcher than the
 		// commitment has been confirmed on chain, and before we
 		// advance our state step, we call InsertConfirmedCommitSet.
-		confCommitSet := commitSet.HtlcSets[*commitSet.ConfCommitKey]
-		if err := c.relaunchResolvers(confCommitSet); err != nil {
+		if err := c.relaunchResolvers(commitSet); err != nil {
 			c.cfg.BlockEpochs.Cancel()
 			return err
 		}
@@ -443,10 +442,10 @@ func (c *ChannelArbitrator) Start() error {
 // starting the ChannelArbitrator. This information should ideally be stored in
 // the database, so this only serves as a intermediate work-around to prevent a
 // migration.
-func (c *ChannelArbitrator) relaunchResolvers(confirmedHTLCs []channeldb.HTLC) error {
-	// We'll now query our log to see if there are any active
-	// unresolved contracts. If this is the case, then we'll
-	// relaunch all contract resolvers.
+func (c *ChannelArbitrator) relaunchResolvers(commitSet *CommitSet) error {
+	// We'll now query our log to see if there are any active unresolved
+	// contracts. If this is the case, then we'll relaunch all contract
+	// resolvers.
 	unresolvedContracts, err := c.log.FetchUnresolvedContracts()
 	if err != nil {
 		return err
@@ -460,6 +459,27 @@ func (c *ChannelArbitrator) relaunchResolvers(confirmedHTLCs []channeldb.HTLC) e
 		return err
 	}
 	commitHash := contractResolutions.CommitHash
+
+	// In prior versions of lnd, the information needed to supplement the
+	// resolvers (in most cases, the full amount of the HTLC) was found in
+	// the chain action map, which is now deprecated.  As a result, if the
+	// commitSet is nil (an older node with unresolved HTLCs at time of
+	// upgrade), then we'll use the chain action information in place. The
+	// chain actions may exclude some information, but we cannot recover it
+	// for these older nodes at the moment.
+	var confirmedHTLCs []channeldb.HTLC
+	if commitSet != nil {
+		confirmedHTLCs = commitSet.HtlcSets[*commitSet.ConfCommitKey]
+	} else {
+		chainActions, err := c.log.FetchChainActions()
+		if err != nil {
+			log.Errorf("unable to fetch chain actions: %v", err)
+			return err
+		}
+		for _, htlcs := range chainActions {
+			confirmedHTLCs = append(confirmedHTLCs, htlcs...)
+		}
+	}
 
 	// Reconstruct the htlc outpoints and data from the chain action log.
 	// The purpose of the constructed htlc map is to supplement to


cfromknecht:
small linter failure
```
contractcourt/channel_arbitrator.go:480:4: S1011: should replace loop with `confirmedHTLCs = append(confirmedHTLCs, htlcs...)` (gosimple)
			for _, htlc := range htlcs {
```

contractcourt/channel_arbitrator.go
```diff
@@ -461,6 +460,29 @@ func (c *ChannelArbitrator) relaunchResolvers(confirmedHTLCs []channeldb.HTLC) e
...
+	// deprecated.  As a result, if the commitSet is nil (an older node
+	// with unresolved HTLCs at time of upgrade), then we'll use the chain
+	// action information in place. The chain actions may exclude some
+	// information, but we cannot recover it for these older nodes at the
+	// moment.
+	var confirmedHTLCs []channeldb.HTLC
+	if commitSet != nil {
+		confirmedHTLCs = commitSet.HtlcSets[*commitSet.ConfCommitKey]
+	} else {
+		chainActions, err := c.log.FetchChainActions()
```
joostjager: What is the long term plan to get rid of chain actions? Can they be migrated to confirmed commit sets? 

One nice (or maybe not so nice) possibilty is to do an in-flight migration in the channel arbitrator, if it is difficult to write a migration without context. Because we now only do major-to-major upgrades, we can remove the in-flight migration code again for the next major release. Only we need to be sure that it actually ran successfully. Not sure how to do that.

contractcourt/channel_arbitrator.go
```diff
@@ -461,6 +460,29 @@ func (c *ChannelArbitrator) relaunchResolvers(confirmedHTLCs []channeldb.HTLC) e
 	}
 	commitHash := contractResolutions.CommitHash
 
+	// In prior versions of lnd, the information needed to supplement the
+	// resolvers was found in the chain action map, which is now
+	// deprecated.  As a result, if the commitSet is nil (an older node
+	// with unresolved HTLCs at time of upgrade), then we'll use the chain
+	// action information in place. The chain actions may exclude some
+	// information, but we cannot recover it for these older nodes at the
+	// moment.
```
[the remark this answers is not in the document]
Roasbeef: It isn't known which information is missing, as we didn't store the chain actions for all the valid commitments at play. The information needed to supplement is the HTLC itself.

contractcourt/channel_arbitrator.go
Roasbeef: Chain actions are deprecated, only those that upgraded with unresolved contracts need to deal with them. We can't directly port them over, as with the on-disk state, it isn't clear _which_ commitment the chain actions were created from (as the trigger that initiated the state transition isn't persisted anywhere). At this stage, the resolvers have already been created as well. 

AFAIK, the success resolver is the only one that can be potentially cancelled back by the invoice registry due to having a zero value amount for example.

joostjager: Ok, so best effort here. It would be nice to at some point be able to remove the chain actions bucket completely though.

Roasbeef: Agreed, we can look at it for 0.9 perhaps? Or even make our update policy even more aggressive: there's a min version, but also you can't update if you have pending contract resolutions. Ofc, this makes a pretty big trade off as far as agility if we need to roll out critical fixes in the future.

PR #3577, opened 2019-10-04
build/release: update release targets

cfromknecht:
Notably, solaris/amd64 and illumos/amd64 were added in go1.13. The
others were have been supported previously but were not build targets

### DIFF
--- a/build/release/release.sh
+++ b/build/release/release.sh
@@ -37,7 +37,28 @@ cd $MAINDIR
 
 # If LNDBUILDSYS is set the default list is ignored. Useful to release
 # for a subset of systems/architectures.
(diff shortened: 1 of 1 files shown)


build/release/release.sh
```diff
@@ -44,18 +44,26 @@ SYS=${LNDBUILDSYS:-"
...
+        linux-mipsle
         linux-mips64
         linux-mips64le
+        linux-s390x
         netbsd-386
         netbsd-amd64
+        netbsd-arm
         openbsd-386
         openbsd-amd64
+        openbsd-arm
```
wpaulino: Any reason for adding some of these and not others? Looking at `go tool dist list`, there are still some missing.

cfromknecht: > `go tool dist list`

TIL! Added in some missing `arm` targets. `plan9` is still experimental and `nacl` will be sunset after go1.13, so those are omitted. Feel free to suggest removing any of these, was just trying to lay out all the options.

PR #3582, opened 2019-10-05
rpc: Add PendingOpenChannel to SubscribeChannelEvents

hsjoberg:
In this PR `PendingOpenChannel` is added to the `SubscribeChannelEvents` stream.

It is already possible to keep track of pending channel openings as the `OpenChannel` command opens a stream where you will get an update when the channel enters pending open state.  
The problem though is that this does not apply to Autopilot, so there's no good way of knowing if Autopilot opened a channel without polling `PendingOpenChannels`, which is not a great solution.  

It can also be useful for services and programs that want to track channel openings to them ("non-initiator"). Right now they can only see the channel if it has entered open state.
For pending, I suspect they poll `PendingOpenChannels` right now or have made a custom version of lnd. They can likely use `SubscribeChannelEvents`  in the future for their needs.

Currently this PR mirrors what `OpenChannel` sends, which is the ChannelPoint. If you need more information, you need to call `PendingOpenChannel` and match the right txid.
I could do so `SubscribeChannelEvents` sends all data about the channel as `PendingChannels` does, but I need some help figuring out how to do that. 
The current method works fine though.

This PR does _not_ include other pending states (i.e pending close etc).


channelnotifier/channelnotifier.go
```diff
@@ -45,6 +46,13 @@ type ClosedChannelEvent struct {
 	CloseSummary *channeldb.ChannelCloseSummary
 }
 
+// PendingOpenChannelEvent represents a new event where a new channel has
+// entered a pending open state
```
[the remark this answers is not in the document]
hsjoberg: Fixed.

channelnotifier/channelnotifier.go
```diff
@@ -128,3 +136,12 @@ func (c *ChannelNotifier) NotifyInactiveChannelEvent(chanPoint wire.OutPoint) {
 		log.Warnf("Unable to send inactive channel update: %v", err)
 	}
 }
+
+// NotifyPendingOpenChannelEvent notifies the channelEventNotifier goroutine that a
+// new channel is pending.
+func (c *ChannelNotifier) NotifyPendingOpenChannelEvent(chanPending lnrpc.PendingUpdate) {
+	event := PendingOpenChannelEvent{ChanPending: &chanPending}
+	if err := c.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send op channel update: %v", err)
```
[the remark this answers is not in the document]
hsjoberg: Definitely! Fixed.

fundingmanager.go
```diff
@@ -350,6 +350,10 @@ type fundingConfig struct {
 	// and on the requesting node's public key that returns a bool which tells
 	// the funding manager whether or not to accept the channel.
 	OpenChannelPredicate chanacceptor.ChannelAcceptor
+
+	// NotifyPendingOpenChannelEvent informs the ChannelNotifier when channels
+	// enter a pending state.
+	NotifyPendingOpenChannelEvent func(lnrpc.PendingUpdate)
```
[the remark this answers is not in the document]
hsjoberg: The reason is that it's the same struct that the `OpenChannel` stream uses.  
If you think it makes more sense to use `wire.OutPoint` instead, I could use that.

Alos see my reply on lnrpc/rpc.proto below as these two a related to each other.

lnrpc/rpc.proto
```diff
@@ -1660,13 +1660,15 @@ message ChannelEventUpdate {
         ChannelCloseSummary closed_channel = 2 [ json_name = "closed_channel" ];
         ChannelPoint active_channel = 3 [ json_name = "active_channel" ];
         ChannelPoint inactive_channel = 4 [ json_name = "inactive_channel" ];
+        ChannelOpenUpdate pending_open_channel = 6 [json_name = "pending_open_channel"];
```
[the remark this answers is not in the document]
hsjoberg: The reason why I used `ChannelOpenUpdate` is because it is the same type that is used for the `OpenChannel` stream. Both use the data and gets triggered in the same place (https://github.com/lightningnetwork/lnd/pull/3582/files#diff-691ca64693e704fd2e5080998f3e640eR1785-R1800).
I skipped `lnrpc.OpenStatusUpdate` though.

If we want to send more data from the `SubscribeChannelEvents` stream in the future, I think it would be a good idea to use `ChannelOpenUpdate` instead of `ChannelPoint`.

If you think it makes more sense to `ChannelPoint` instead, I could change to that.

hsjoberg:
Thank you for the review @carlaKC!☺️
I've rebased the branch and addressed some of the points.

fundingmanager.go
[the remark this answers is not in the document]
halseth: Agreed that it makes more sense to use `wire.Outpoint` here, and then we can convert it to the `lnrpc` type in `rpcserver`.

lnrpc/rpc.proto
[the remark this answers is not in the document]
halseth: Doesnt the stream use `PendingUpdate` here?

channelnotifier/channelnotifier.go
```diff
@@ -45,6 +46,13 @@ type ClosedChannelEvent struct {
 	CloseSummary *channeldb.ChannelCloseSummary
 }
 
+// PendingOpenChannelEvent represents a new event where a new channel has
+// entered a pending open state.
+type PendingOpenChannelEvent struct {
```
halseth: nit: move up such that the channel states are presented in the natural order

hsjoberg:
Thanks for the feedback @halseth!
I've been busy recently but hope to look into this (very) soon. :smiley:

lnrpc/rpc.proto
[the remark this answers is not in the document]
hsjoberg: @halseth You're right, `PendingUpdate` is what `OpenChannel` uses in this case.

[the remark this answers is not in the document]
hsjoberg: Fixed.

channelnotifier/channelnotifier.go
hsjoberg: Makes sense, fixed.

fundingmanager.go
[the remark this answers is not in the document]
hsjoberg: Changed to wire.Outpoint.

hsjoberg:
Hi @carlaKC and @halseth, sorry for the long delay on this one.
I've rebased against master and addressed the suggestions you provided. ☺️

* NotifyPendingOpenChannelEvent now uses wire.OutPoint
* Protobuf: ChannelEventUpdate pending_open_channel now uses the correct type, PendingUpdate
* Miscellaneous things such as reordering code to be more logical and updating the SubscribeChannelEvents function inside rpcserver.go to work with the changes made to the channelnotifier (event uses outpoint)

fundingmanager.go
```diff
@@ -1833,6 +1841,10 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 		PendingChanId: pendingChanID[:],
 	}
 
+	// Inform the ChannelNotifier that the channel has entered
+	// pending open state.
+	f.cfg.NotifyPendingOpenChannelEvent(*fundingPoint)
```
carlaKC: Should we move this into the select below? The `resCtx.updates` channel is where we report the upstream client, so it would make sense to notify the pending channel on send into that channel?

hsjoberg:
@carlaKC 
> Thanks for sticking with this one @hsjoberg, it's going to be nice to have a complete channel subscription api. Just one small comment from me, and an itest fix, this is getting really close!

No worries, I appreciate the reviewing.

> Itests are failing on SubscribeChannelBackups at the moment, because it subscribes to channel updates and then skips over events that aren't opens/closes. You just need to skip over pending channels in rpcserver.go#SubscribeChannelBackups in the switch on line 5553 and that should set it right :)

Right! My bad, I only tested against `basic_channel_creation_and_updates`.
`make check` should go through now.

fundingmanager.go
carlaKC: Something like this:
```
	select {
	case resCtx.updates <- upd:
		// Inform the ChannelNotifier that the channel has entered
		// pending open state.
		f.cfg.NotifyPendingOpenChannelEvent(*fundingPoint)
	case <-f.quit:
		return
	}
```

hsjoberg: @carlaKC Thank you.
Fixed!

lnrpc/rpc.proto
```diff
@@ -1930,13 +1930,15 @@ message ChannelEventUpdate {
         ChannelCloseSummary closed_channel = 2 [ json_name = "closed_channel" ];
         ChannelPoint active_channel = 3 [ json_name = "active_channel" ];
         ChannelPoint inactive_channel = 4 [ json_name = "inactive_channel" ];
+        PendingUpdate pending_open_channel = 6 [json_name = "pending_open_channel"];
```
carlaKC: Last thing, sorry that I missed this! This should be 5 rather than 6?
You also need to run `make rpc` after this most recent change, some things change when I run it.

hsjoberg: > Last thing, sorry that I missed this! This should be 5 rather than 6?

This was deliberate from my side -- maybe misunderstanding how protobufs works, but `UpdateType type` is number 5.

https://github.com/lightningnetwork/lnd/blob/ab04a8b3c0566a9321f5f672c81c6db719efbb15/lnrpc/rpc.proto#L1944 

>  You also need to run make rpc after this most recent change, some things change when I run it.

Weird, I will do that.

hsjoberg: @carlaKC You're right, changes to `rpc.swagger.json` is missing.
Quite a lot of other changes seem to happen as well to the file, so I suspect I have the wrong version of some tool. I'll work on fixing that.

hsjoberg: Okay `rpc.swagger.json` has been updated.

carlaKC: > This was deliberate from my side -- maybe misunderstanding how protobufs works, but UpdateType type is number 5.

Yes that's correct, missed that field when I was scanning it.

fundingmanager.go
```diff
@@ -1835,6 +1843,9 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 
 	select {
 	case resCtx.updates <- upd:
+		// Inform the ChannelNotifier that the channel has entered
```
halseth: nit: just move this code after the select statement

hsjoberg:
Rebased from master and fixed the last nit. :)

hsjoberg:
Sorry, the review requests are probably not needed. I made an erroneous rebase so review requests got triggered because of CODEOWNERS.

PR #3584, opened 2019-10-07
docker: fix ltcd ports

bjarnemagnussen:
# Fixes `ltcd` ports inside the Dockerfile

The ports exposed in the Dockerfile for Litecoin `ltcd` are wrong. It is therefore not possible to connect to the `ltcd` container from another container without this fix.

The Litecoin ports are as follows:

* Mainnet: Server `9333` and RPC `9334`
* Testnet: Server `19335` and RPC `19334`
* Simnet: Server `18555` and RPC `18556`

Those ports can easily be verified by starting the docker-container for `ltcd` (`NETWORK="testnet" docker-up -d ltcd`) and check the log (`docker logs ltcd`). 

Further, the Litecoin "segnet" has apparently been merged into the "testnet" and the network options has been changed accordingly inside the start scripts `start-ltcd.sh` and `start-ltcctl.sh` (follows the start script for `btcd`). There is not option to set network to "segnet" for `ltcd` anymore.

### DIFF
--- a/docker/ltcd/start-ltcctl.sh
+++ b/docker/ltcd/start-ltcctl.sh
@@ -43,10 +43,17 @@ RPCUSER=$(set_default "$RPCUSER" "devuser")
 RPCPASS=$(set_default "$RPCPASS" "devpass")
 NETWORK=$(set_default "$NETWORK" "simnet")
 
-exec ltcctl \
-    "--$NETWORK" \
-    --rpccert="/rpc/rpc.cert" \
-    --rpcuser="$RPCUSER" \
-    --rpcpass="$RPCPASS" \
-    --rpcserver="rpcserver" \
-    "$@"
+PARAMS=""
+if [ "$NETWORK" != "mainnet" ]; then
+    PARAMS="--$NETWORK"
+fi
+
+PARAMS=$(echo $PARAMS \
+    "--rpccert=/rpc/rpc.cert" \
+    "--rpcuser=$RPCUSER" \
+    "--rpcpass=$RPCPASS" \
+    "--rpcserver=rpcserver" \
+)
+
+PARAMS="$PARAMS $@"
+exec ltcctl $PARAMS
--- a/docker/ltcd/start-ltcd.sh
+++ b/docker/ltcd/start-ltcd.sh
@@ -44,8 +44,12 @@ RPCPASS=$(set_default "$RPCPASS" "devpass")
 DEBUG=$(set_default "$DEBUG" "info")
 NETWORK=$(set_default "$NETWORK" "simnet")
 
-PARAMS=$(echo \
-    "--$NETWORK" \
+PARAMS=""
+if [ "$NETWORK" != "mainnet" ]; then
+   PARAMS="--$NETWORK"
+fi
+
+PARAMS=$(echo $PARAMS \
     "--debuglevel=$DEBUG" \
     "--rpcuser=$RPCUSER" \
     "--rpcpass=$RPCPASS" \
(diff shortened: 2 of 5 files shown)


docker/ltcd/start-ltcctl.sh
```diff
@@ -43,8 +43,12 @@ RPCUSER=$(set_default "$RPCUSER" "devuser")
 RPCPASS=$(set_default "$RPCPASS" "devpass")
 NETWORK=$(set_default "$NETWORK" "simnet")
 
-exec ltcctl \
-    "--$NETWORK" \
+PARAMS=""
+if [ "$NETWORK" != "mainnet" ]; then
+    PARAMS=$(echo --$NETWORK)
```
guggero: No need to use `echo`. This line can just be:
`PARAMS="--$NETWORK"`

docker/ltcd/start-ltcctl.sh
```diff
@@ -43,8 +43,12 @@ RPCUSER=$(set_default "$RPCUSER" "devuser")
...
+fi
+
+PARAMS=$(echo $PARAMS \
```
guggero: This breaks the whole command.
1. The `exec ltcctl` part is gone
2. Missing closing parens
3. Why assigning to `PARAMS` again?

docker/ltcd/start-ltcd.sh
```diff
@@ -44,8 +44,12 @@ RPCPASS=$(set_default "$RPCPASS" "devpass")
 DEBUG=$(set_default "$DEBUG" "info")
 NETWORK=$(set_default "$NETWORK" "simnet")
 
-PARAMS=$(echo \
-    "--$NETWORK" \
+PARAMS=""
+if [ "$NETWORK" != "mainnet" ]; then
+   PARAMS=$(echo --$NETWORK)
```
guggero: Same as above, for a one-liner you don't need echo.

bjarnemagnussen: Will also change that for `btcd/start-btcd.sh?

docker/ltcd/start-ltcctl.sh
bjarnemagnussen: Ups, thanks for finding that!

docker/ltcd/start-ltcctl.sh
bjarnemagnussen: OK! I will also change that for `btcd/start-btcctl.sh`.

bjarnemagnussen:
I am not very familiar with Docker, but as far as I understand the `EXPOSE` instructions does however in combination with `docker-compose` allow for inter-container communication on the exposed port(s). 

This is therefore important as otherwise the inter-container communication between the `lnd` and `blockchain` container cannot happen for Litecoin.

docker/ltcd/start-ltcctl.sh
```diff
@@ -43,10 +43,17 @@ RPCUSER=$(set_default "$RPCUSER" "devuser")
...
+    "--rpccert=/rpc/rpc.cert" \
+    "--rpcuser=$RPCUSER" \
+    "--rpcpass=$RPCPASS" \
+    "--rpcserver=rpcserver" \
+)
+
+PARAMS="$PARAMS $@"
+exec btcctl $PARAMS
```
guggero: This needs to say `exec ltcctl $PARAMS`.

bjarnemagnussen: Oh my! Fixed!

guggero:
Thanks for the quick fix. Could you please also squash the 3 commits into a single one? Then I think we're good to go!

PR #3591, opened 2019-10-10
Documentation only: Update recovery.md

robclark56:
The original text says that the SCB file is only updated on open/close of a channel.

But observed behavior is that it is also updated when a channel 'changes' - typically when funds move from one end to the other as an HTLC passes through without closing the channel.

Am I correct?

#### Pull Request Checklist

  logging level
  per default)
  already exist

### DIFF
--- a/docs/recovery.md
+++ b/docs/recovery.md
@@ -240,7 +240,7 @@ process.
 There are multiple ways of obtaining SCBs from `lnd`. The most commonly used
 method will likely be via the `channels.backup` file that's stored on-disk
 alongside the rest of the chain data. This is a special file that contains SCB
-entries for _all_ currently open channels. Each time a channel is opened or
+entries for _all_ currently open channels. Each time a channel is opened, changed, or
 closed, this file is updated on disk in a safe manner (atomic file rename). As
 a result, unlike the `channel.db` file, it's _always_ safe to copy this file
 for backup at ones desired location. The default location on Linux is: 


guggero:
AFAIK this is not correct. The 'S' in SCB stands for static. So it would defeat the whole purpose if the file would need to be updated for every routed payment.
Can you maybe go into a bit more detail with what you mean by "observed behavior"?

robclark56:
1. SCB is updated every time LND is restarted.
This occurs regardless of whether a channel is opened or closed (i.e. no onchain transactions). So there is one data point where the SCB file is updated with no channel open/close.

2. As for SCB file being updated if just an HTLC flows through ... I will have to do more observations. I have plenty of timestamped pairs of channel.backup & 'lnci listchannels' outputs. But what I don't have is the information if the SCB file was updated because of (1) lnd restart, or (2) HTLC passing through, or (3) some other state change on a channel (but not open/close).  I get an email when these pairs are written ... and from memory the emails arrived even when channels were not opened/closed, nor lnd restarted. 

Maybe a dev can confirm all the events where channels.backup is updated:

* Restart (Stop? or Start?) = Yes
* Channel Open = Yes
* Channel Close = Yes
* Channel changes offline/online?
* Local/Remote mix changes? 
* others?

robclark56:
Closing this PR as my original comment that SCB files change on HTLCs flowing through is wrong.

Leave it to the powers that be to add ", on LND start, ..." instead.

Thanks

Rob

PR #3593, opened 2019-10-11
[neutrino] Handle GetCFilter fetch errors

halseth:
Earlier GetCFilter could return a nil-error along with a nil-filter in
cases where we failed to fetch the filter. We would just ignore the
block causing us to miss potential crucial information.

Now that GetCFilter will return an error in cases where the fetch
failed, the filter will never be nil (even for blocks with only coinbase
transactions), and  we'll instead return an error.

Depends on https://github.com/lightninglabs/neutrino/pull/187


chainntnfs/neutrinonotify/neutrino.go
```diff
@@ -529,15 +529,8 @@ func (n *NeutrinoNotifier) historicalConfDetails(confRequest chainntnfs.ConfRequ
...
-			return nil, fmt.Errorf("unable to retrieve regular filter for "+
-				"height=%v: %v", scanHeight, err)
-		}
-
-		// If the block has no transactions other than the Coinbase
-		// transaction, then the filter may be nil, so we'll continue
-		// forward int that case.
-		if regFilter == nil {
-			continue
+			return nil, fmt.Errorf("unable to retrieve regular "+
```
wpaulino: Do we also want to include some simple polling logic here until the planned fix is in? Otherwise the historical rescan will fail and won't be retried, requiring a restart of lnd for it to happen again.

halseth:
> `btcwallet` will also require the dependency update for wallet recovery.

My assumption was that `btcwallet` will also use the version specified in `lnd`'s `go.mod` when it's an indirect dependency? But I'm not sure, so created https://github.com/btcsuite/btcwallet/pull/649

chainntnfs/neutrinonotify/neutrino.go
halseth: Yeah, for now wanted to do the minimal change possible to avoid the disaster scenario. Do you think we should add that mitigation here? (the real fix as you say, is a little more advanced)

wpaulino: I think so, since this is critical for channel confirmation. Just copying the `pollCFilter` method that's used in `btcwallet` should help a bit, I'd increase the poll interval to something like a second though. We can then remove it once the proper fix is in.

wpaulino:
> My assumption was that btcwallet will also use the version specified in lnd's go.mod when it's an indirect dependency? But I'm not sure, so created btcsuite/btcwallet#649

I'd think `btcwallet` would just use whatever it has pinned, since it isn't aware of `lnd`.

Roasbeef:
> I'd think btcwallet would just use whatever it has pinned, since it isn't aware of lnd.

The opposite dependnacy is what matters: btcwallet's dep will be committed to in the `go.sum` but what `lnd` specifies will be used to build.

Roasbeef:
If you build the `btcwallet` binary and use _that_ directly, then it'll pull in what `btcwallet` specifies.

chainntnfs/neutrinonotify/neutrino.go
Roasbeef: Can't we just pass in the required optional arguments for it to _continually_ poll itself inside neutrino? This was the default in the past, but the re-org issues were discovered causing us to remove retries at the query level, so higher level logic could handle awareness of non-existent blocks, etc.

Roasbeef: Alternatively, can't we just _re use_ the rescan logic here? This is basically a bounded rescan from a known point in the past to now. This would allow us to leverage all the retry and re-org awareness already in that core routine.

Roasbeef: This is likely a larger change though, as we'll need to construct all its interface dependancies here in the router.

halseth: Yeah, these are bigger changes. A simple poll loop would be ok to add this late in the game I would say.

halseth: Retry loop: https://github.com/lightningnetwork/lnd/pull/3593/commits/4c1d50e6782f0cda68e7f0dafa0e7ef78054cf0a

wpaulino: > Can't we just pass in the required optional arguments for it to continually poll itself inside neutrino?

We can and it seems to me like the cleaner approach as well.

wpaulino: Would also add a TODO to remove the retrying once the proper fix lands.

halseth: Done.

halseth:
Rebased and updated with neutrino+btcd+btcwallet

go.mod
```diff
@@ -58,6 +58,8 @@ replace github.com/lightningnetwork/lnd/ticker => ./ticker
 
 replace github.com/lightningnetwork/lnd/queue => ./queue
 
+replace github.com/lightninglabs/neutrino => github.com/halseth/neutrino v0.0.0-20191011081321-f8fae25c0ccc
```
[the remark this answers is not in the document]
halseth: Dang, removed.

PR #3594, opened 2019-10-11
cnct: decode onion for incoming contest resolver

joostjager:
Also for on-chain resolution, we may require access to the onion payload. An example of this is the resolution of an mpp shard. We need the mpp fields in order to properly offer the htlc to the invoice registry.

### DIFF
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -501,96 +501,24 @@ func (c *ChannelArbitrator) relaunchResolvers(commitSet *CommitSet) error {
 		"resolvers", c.cfg.ChanPoint, len(unresolvedContracts))
 
 	for _, resolver := range unresolvedContracts {
-		if err := c.supplementResolver(resolver, htlcMap); err != nil {
-			return err
+		htlcResolver, ok := resolver.(htlcContractResolver)
+		if !ok {
+			continue
 		}
+
+		htlcPoint := htlcResolver.HtlcPoint()
+		htlc, ok := htlcMap[htlcPoint]
+		if !ok {
+			return fmt.Errorf(
+				"htlc resolver %T unavailable", resolver,
+			)
+		}
+
+		htlcResolver.Supplement(*htlc)
 	}
 
 	c.launchResolvers(unresolvedContracts)
 
-	return nil
-}
-
-// supplementResolver takes a resolver as it is restored from the log and fills
-// in missing data from the htlcMap.
-func (c *ChannelArbitrator) supplementResolver(resolver ContractResolver,
-	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
-
-	switch r := resolver.(type) {
-
-	case *htlcSuccessResolver:
-		return c.supplementSuccessResolver(r, htlcMap)
-
-	case *htlcIncomingContestResolver:
-		return c.supplementIncomingContestResolver(r, htlcMap)
-
-	case *htlcTimeoutResolver:
-		return c.supplementTimeoutResolver(r, htlcMap)
-
-	case *htlcOutgoingContestResolver:
-		return c.supplementTimeoutResolver(
-			&r.htlcTimeoutResolver, htlcMap,
-		)
-	}
-
-	return nil
-}
-
-// supplementSuccessResolver takes a htlcIncomingContestResolver as it is
-// restored from the log and fills in missing data from the htlcMap.
-func (c *ChannelArbitrator) supplementIncomingContestResolver(
-	r *htlcIncomingContestResolver,
-	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
-
-	res := r.htlcResolution
-	htlcPoint := res.HtlcPoint()
-	htlc, ok := htlcMap[htlcPoint]
-	if !ok {
-		return errors.New(
-			"htlc for incoming contest resolver unavailable",
-		)
-	}
-
-	r.htlcAmt = htlc.Amt
-	r.circuitKey = channeldb.CircuitKey{
-		ChanID: c.cfg.ShortChanID,
-		HtlcID: htlc.HtlcIndex,
-	}
-
-	return nil
-}
-
-// supplementSuccessResolver takes a htlcSuccessResolver as it is restored from
-// the log and fills in missing data from the htlcMap.
-func (c *ChannelArbitrator) supplementSuccessResolver(r *htlcSuccessResolver,
-	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
-
-	res := r.htlcResolution
-	htlcPoint := res.HtlcPoint()
-	htlc, ok := htlcMap[htlcPoint]
-	if !ok {
-		return errors.New(
-			"htlc for success resolver unavailable",
-		)
-	}
-	r.htlcAmt = htlc.Amt
-	return nil
-}
-
-// supplementTimeoutResolver takes a htlcSuccessResolver as it is restored from
-// the log and fills in missing data from the htlcMap.
-func (c *ChannelArbitrator) supplementTimeoutResolver(r *htlcTimeoutResolver,
-	htlcMap map[wire.OutPoint]*channeldb.HTLC) error {
-
-	res := r.htlcResolution
-	htlcPoint := res.HtlcPoint()
-	htlc, ok := htlcMap[htlcPoint]
-	if !ok {
-		return errors.New(
-			"htlc for timeout resolver unavailable",
-		)
-	}
-	r.htlcAmt = htlc.Amt
 	return nil
 }
 
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -1647,6 +1647,8 @@ func (c *ChannelArbitrator) prepContractResolutions(
 		// claim the HTLC (second-level or directly), then add the pre
 		case HtlcClaimAction:
 			for _, htlc := range htlcs {
+				htlc := htlc
+
 				htlcOp := wire.OutPoint{
 					Hash:  commitHash,
 					Index: uint32(htlc.OutputIndex),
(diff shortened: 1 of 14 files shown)


KarolStrzelec:
hello pls check that;
https://drive.google.com/file/d/1H0SLt2ntS2yqjJ7lkeisFhyHLIFH070D/view?usp=sharing

contractcourt/channel_arbitrator.go
```diff
@@ -551,6 +567,18 @@ func (c *ChannelArbitrator) supplementIncomingContestResolver(
...
 	}
 
+	payload, err := c.decodePayload(htlc)
+	if err != nil {
+		log.Debugf("ChannelArbitrator(%v): cannot decode payload of "+
+			"htlc %v", c.cfg.ChanPoint, htlcPoint)
+
+		// If we've locked in an htlc with an invalid payload on our
+		// commitment tx, we don't need to resolve it.
+		r.resolved = true
```
cfromknecht: do we need to fail back the htlc in this case? that's what we would have done if we processed the htlc entirely off chain

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -29,6 +29,10 @@ type htlcIncomingContestResolver struct {
 	// circuitKey describes the incoming htlc that is being resolved.
 	circuitKey channeldb.CircuitKey
 
+	// payload is the hop payload which can contain additional data that is
+	// required for resolution.
+	payload *hop.Payload
```
cfromknecht: make `invoices.Payload`?

contractcourt/channel_arbitrator.go
```diff
@@ -1224,8 +1224,9 @@ func (c *ChannelArbitrator) checkCommitChainActions(height uint32,
 	//  * race condition if adding and we broadcast, etc
 	//  * or would make each instance sync?
 
-	log.Debugf("ChannelArbitrator(%v): checking chain actions at "+
-		"height=%v", c.cfg.ChanPoint, height)
+	log.Debugf("ChannelArbitrator(%v): checking commit chain actions at "+
+		"height=%v, inHtlcCount=%v, outHtlcCount=%v", c.cfg.ChanPoint,
```
cfromknecht: nit: use snake case for logs

contractcourt/htlc_incoming_contest_resolver.go
joostjager: do you suggest renaming the whole package?

contractcourt/channel_arbitrator.go
joostjager: fixed

contractcourt/channel_arbitrator.go
joostjager: but we are on-chain now, so what does fail back look like? onion parsing is only relevant for incoming htlcs, so there is also not another side of the circuit to signal

contractcourt/htlc_incoming_contest_resolver.go
cfromknecht: I’m suggesting the type be the interface `invoices.Payload` instead of the concrete type

joostjager:
Changed the design slightly to make sure we always parse the onion blob. Also if we didn't just restart.

contractcourt/channel_arbitrator.go
cfromknecht: yes of course, you're right. since this is always an incoming channel there is nothing the fail, the remote node should propagate the failure from w/e happens on chain

contractcourt/contract_resolvers.go
```diff
@@ -61,6 +64,18 @@ type ContractResolver interface {
...
+// htlcContractResolver is the required interface for htlc resolvers.
+type htlcContractResolver interface {
+	ContractResolver
+
+	// HtlcPoint returns the htlc's outpoint on the commitment tx.
+	HtlcPoint() wire.OutPoint
+
+	// Supplement adds additional information to the resolver that is
+	// required before Resolve() is called.
+	Supplement(htlc *channeldb.HTLC) error
```
cfromknecht: atm none of these can fail (other than panicking), does it need a return value? are there cases in the future where we might that would justify leaving it?

contractcourt/htlc_success_resolver.go
```diff
@@ -39,19 +37,15 @@ type htlcSuccessResolver struct {
...
 	// from the commitment transaction of the remote party.
 	//
 	// TODO(roasbeef): send off to utxobundler
 	sweepTx *wire.MsgTx
 
-	// htlcAmt is the original amount of the htlc, not taking into
-	// account any fees that may have to be paid if it goes on chain.
-	htlcAmt lnwire.MilliSatoshi
+	// htlc contains information on the htlc that we are resolving on-chain.
+	htlc *channeldb.HTLC
```
cfromknecht: should we copy the htlc instead, i.e. no pointer, so that it can't be mutated externally?

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -50,6 +51,19 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
 
+	// First try to parse the payload. If that fails, we can stop resolution
+	// now.
+	payload, err := h.decodePayload(h.htlc)
+	if err != nil {
+		log.Debugf("ChannelArbitrator(%v): cannot decode payload of "+
+			"htlc %v", h.ChanPoint, h.HtlcPoint())
+
+		// If we've locked in an htlc with an invalid payload on our
+		// commitment tx, we don't need to resolve it.
```
cfromknecht: perhaps update the comment to explain that the other party will time out the htlc if they want their funds back?

cfromknecht:
would be nice to get this in before #3644

contractcourt/contract_resolvers.go
joostjager: I don't know any future cases, so removed it for now.

contractcourt/htlc_success_resolver.go
joostjager: I don't know. It is a safety precaution at the cost of a very minimal performance degradation, if any. But just removing the pointer still doesn't make it a deep copy. Also, `channeldb.HTLC` isn't a struct that is typically modified afaik. The most likely bug here would be binding in a range. That would be disastrous, but it is also caught by the linter. I tend towards not defensively dereferencing, but interested you hear ideas

contractcourt/htlc_incoming_contest_resolver.go
joostjager: added

contractcourt/htlc_success_resolver.go
halseth: It would at least let us avoid the awkward loop binding hack I think

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -50,6 +51,20 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 		return nil, nil
 	}
 
+	// First try to parse the payload. If that fails, we can stop resolution
+	// now.
+	payload, err := h.decodePayload(h.htlc)
```
halseth: just name this `payload`? Since you won't have to pass the htlc as a param

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -50,6 +51,20 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
+	if err != nil {
+		log.Debugf("ChannelArbitrator(%v): cannot decode payload of "+
+			"htlc %v", h.ChanPoint, h.HtlcPoint())
+
+		// If we've locked in an htlc with an invalid payload on our
```
halseth: how could we lock in such an HTLC?

joostjager: crash before we reached `processRemoteAdds` in the link

halseth: So we will lock in the HTLC, crash, restart and forget about it? Shouldn't we either forward or cancel it in that case?

Just seems a bit dangerous to ignore a locked in HTLC without returning any error here...

joostjager: We can forget about it, because the onion payload is invalid. In the off-chain flow nothing would have happened either.

contractcourt/htlc_success_resolver.go
cfromknecht: > But just removing the pointer still doesn't make it a deep copy

Everything on `channeldb.HTLC` is shallow besides the `Sig` field, which idt we use in the cnct atm. Only brought it up bc as you say, a range binding bug could be pretty disastrous in this area, and seemed like an easy way to avoid the risk altogether.

joostjager: Ok, will remove the pointer then. Cheap insurance

joostjager: Done

joostjager:
@cfromknecht re-requesting review, as I got a rebase that was a bit more involved. I think it is better to be a bit more cautious with this sensitive part of the code.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -344,6 +362,20 @@ func (h *htlcIncomingContestResolver) Supplement(htlc channeldb.HTLC) {
 	h.htlc = htlc
 }
 
+// decodePayload (re)decodes the hop payload of a received htlc.
+func (h *htlcIncomingContestResolver) decodePayload() (*hop.Payload, error) {
+
```
cfromknecht: nit: extra new line

contractcourt/htlc_incoming_contest_resolver.go
cfromknecht: if the payload is invalid we would have canceled the htlc back offchain. now that we are on chain we can't cancel (or decode the payload and figure out what needs to be done) so we just let the incoming party timeout the htlc

Issue #3590, opened 2019-10-10
QueryRoutes request is missing argument for maximum CLTV value


alexbosworth:
When sending a payment normally using `SendRequest` the caller can set a maximum `cltv_limit` for their payment. In `QueryRoutesRequest` there is no such argument and this means that routes may be returned that have too-high CLTV locktime values.

The QueryRoutes method should match the SendPayment method in allowing the caller to specify a max CLTV timeout.

PR #3595, opened 2019-10-11
routing+routerrpc: take max cltv limit into account within path finding

wpaulino:
With the introduction of the max CLTV limit parameter, nodes are able to reject HTLCs that exceed it. This should also be applied to path finding, otherwise HTLCs crafted by the same node that exceed it never left the switch. This wasn't a big deal since the previous max CLTV limit was ~5000 blocks. Once it was lowered to 1008, the issue became more apparent. Therefore, all of our path finding attempts now have a restriction of said limit in in order to properly carry out HTLCs to the network.

Along the way, we also increase the default limit to 2016 blocks as the previous one of 1008 blocks proved to be low, given that almost 50% of the network still advertises CLTV deltas of 144 blocks, possibly resulting in routes with many hops failing.

Fixes https://github.com/lightningnetwork/lnd/issues/3590.


lnrpc/routerrpc/router_backend.go
```diff
@@ -169,6 +169,11 @@ func (r *RouterBackend) QueryRoutes(ctx context.Context,
 		ignoredPairs[pair] = struct{}{}
 	}
 
+	cltvLimit, err := ValidateCLTVLimit(in.CltvLimit, r.MaxTotalTimelock)
```
joostjager: I am in doubt about this now. It is possible to call `QueryRoutes` with a different source pubkey. This allows the construction of routes not starting from ourselves. Maybe we should allow higher values, but if none is given, default to the max?

lnrpc/rpc.proto
```diff
@@ -1741,6 +1742,13 @@ message QueryRoutesRequest {
     A list of directed node pairs that will be ignored during path finding.
     */
     repeated NodePair ignored_pairs = 10;
+
+    /** 
+    An optional maximum total time lock for the route. This should not exceed
+    lnd's `--max-cltv-expiry` setting. If zero, then the value of
```
joostjager: So maybe for queryroutes, we should allow exceeding the max? Or only if the source pubkey is set and not equal to our own pubkey?

routing/pathfind_test.go
```diff
@@ -1999,9 +2002,9 @@ func testCltvLimit(t *testing.T, limit uint32, expectedChannel uint64) {
 	target := testGraphInstance.aliasMap["target"]
 
 	// Find the best path given the cltv limit.
-	var cltvLimit *uint32
+	cltvLimit := uint32(2016)
 	if limit != 0 {
-		cltvLimit = &limit
+		cltvLimit = limit
 	}
```
joostjager: Just set 2016 in the testcase above?

lnrpc/routerrpc/router_backend.go
```diff
@@ -192,6 +201,7 @@ func (r *RouterBackend) QueryRoutes(ctx context.Context,
 			)
 		},
 		DestPayloadTLV: len(destTLV) != 0,
+		CltvLimit:      cltvLimit,
```
joostjager: Should the final cltv delta be substracted from this? I think it may be a pre-existing bug, but now that the limit is enforced hard by the outgoing link, it may be more important to fix.

wpaulino: Nice catch, fixed.

routing/pathfind_test.go
wpaulino: Fixed.

lnrpc/rpc.proto

lnrpc/routerrpc/router_backend.go
wpaulino: Good point, fixed.

lnrpc/rpc.proto
```diff
@@ -1742,6 +1742,13 @@ message QueryRoutesRequest {
     A list of directed node pairs that will be ignored during path finding.
     */
     repeated NodePair ignored_pairs = 10;
+
+    /** 
+    An optional maximum total time lock for the route. If the source is empty or
+    ourselves, this should not exceed lnd's `--max-cltv-expiry` setting. If
+    zero, then the value of `--max-cltv-expiry` is used as the limit.
+    */
+    uint32 cltv_limit = 11;
```
Roasbeef: This is actually required for the base bug fix? I can see how it's a nice to have to ensure users don't obtain a route than we can't send to however.

Roasbeef: On the other hand, nice that it unifies some code around the handling of the final CLTV delta.

lnrpc/routerrpc/router_backend.go
```diff
@@ -169,6 +170,26 @@ func (r *RouterBackend) QueryRoutes(ctx context.Context,
...
+	}
+
+	// We need to subtract the final delta before passing it into path
+	// finding. The optimal path is independent of the final cltv delta and
+	// the path finding algorithm is unaware of this value.
+	finalCLTVDelta := uint16(zpay32.DefaultFinalCLTVDelta)
+	if in.FinalCltvDelta != 0 {
+		finalCLTVDelta = uint16(in.FinalCltvDelta)
+	}
+	cltvLimit -= uint32(finalCLTVDelta)
```
joostjager: My thought was to do this subtracting in `ChannelRouter.FindRoute`. To me it sounds more logical to accept a `cltvLimit` for the whole route as a parameter to `FindRoute`.

lnrpc/rpc.proto
joostjager: We could have left it out and always queryroutes with the max cltv that is configured (if routing from self). I don't mind it being added in the pr.

lnrpc/routerrpc/router_backend.go
wpaulino: The godoc of the field in the `RestrictParams` struct indicates that it should not include the final CLTV delta. Are you suggesting we change this as well?

joostjager: Oh.. yes, that is indeed the interface towards `findPath`, because `findPath` knows nothing about final cltv delta. Maybe it would have been better if `FindRoute` had a different interface that doesn't expose this implementation detail. 

Probably leave it as it is for now.

lnrpc/routerrpc/router_backend.go
```diff
@@ -205,24 +231,12 @@ func (r *RouterBackend) QueryRoutes(ctx context.Context,
 	// Query the channel router for a possible path to the destination that
 	// can carry `in.Amt` satoshis _including_ the total fee required on
 	// the route.
-	var (
-		route   *route.Route
-		findErr error
+	route, err := r.FindRoute(
+		sourcePubKey, targetPubKey, amtMSat, restrictions,
+		destTlvRecords, finalCLTVDelta,
```
joostjager: Can the `FindRoute` without cltv delta parameter now be removed?

wpaulino: That's already done in the diff.

wpaulino: Did you mean to ask why? If so, because the router's `FindRoute` already applies the default when the value is 0.

joostjager: I mean the `finalExpiry ...uint16` (three dots)

lnrpc/rpc.proto
Roasbeef: Yep, fine with me as well.

PR #3597, opened 2019-10-12
channeldb: revert to prior migration policy 

Roasbeef:
In this commit, we revert #3485, thereby reverting to our existing migration policy. The prior stricter policy created a large burden on applications that package `lnd` as they're forced to deploy special code to handle certain upgrade paths. In addition, it was realized that the policy was most damaging to mobile deployments as it's customary for users to skip versions, making the stricter upgrade policy difficult to manage without dramatically affecting the end user.

### DIFF
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -1111,10 +1111,8 @@ func (d *DB) syncVersions(versions []version) error {
 	}
 
 	latestVersion := getLatestDBVersion(versions)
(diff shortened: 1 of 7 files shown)


lockhartheavyindustries:
Bless you sir.

Roasbeef:
🙏🏾 thanks to all that gave feedback on this topic!

Issue #3287, opened 2019-07-10
lnd: implement "safe mode" node stand up 


Roasbeef:
Although we now have the proper base set of tools in place (SCB) to enable nodes to safely reclaim their channels in the event of data loss, it's still possible that a node boots up with stale data. If this is the case, then the node is at risk of breaching the other channel peer inadvertently. Due to systems like the `contractcourt` which will automatically force close channels that have expired HTLCs, this can happen in an automated fashion on restarts. 

Rather than resuming normal operation if one _knows_ they may be restoring with an out dated state, we can instead implement a "safe mode" of sorts. When users boot up in this mode, **all commitment broadcasts are forbidden**.  Once `lnd` has booted up, the user can then examine the set of channel states to see if they become borked once we connect to peers (indicative of local data loss). 

# Steps To Completion 

- [ ] Add new `--safemode` config parameter to the `lnd` binary. 

- [ ] If safe mode is enabled, reject all RPC level force close requests. 

- [ ] If safe mode is enabled, reject all automated force close requests by channel arbitrators.

mlerner:
I am going to take a shot at working on this.

alexbosworth:
It would be nice to have a flag on force close to do a "dangerous" force close

mlerner:
@alexbosworth: Is the case in which that would be helpful when you want to make sure that no other channels are force-closed besides ones you specifically choose? I could see that being useful. 

What do you think about adding a confirmation message in the case of a "dangerous force close" command received over RPC (I'm not sure if there is a precedent for that type of user interaction)? Also, I'm assuming that automated force closing requests would still be prohibited.

Alternatively, one could argue that "safe mode" should prevent a user from performing dangerous operations, and that the user should restart `lnd` without "safe mode" on if they want to do dangerous things like force close channels with a node in an outdated state - part of the goal of this feature is to allow a user to start `lnd` when they "know it is in an outdated state".

alexbosworth:
The use case I am specifically thinking about is one where you have an out of date backup and you want to use it to recover funds

I'm not sure if this is handled in this PR, but blanket banning force closes seems risky to me in the event of race conditions relating to HTLC resolution

So where I would see using this is:

1. User has an out of date backup
2. They load their out of date backup into safe mode
3. They recover as much as they can in safe mode, knowing that they are protected from breaching
4. When that is finished, they decide for themselves the risk of force closing with unresponsive peers, hopefully after a long period of time of no connectivity

[...]

PR #3601, opened 2019-10-15
lnd: implement "safe mode" node stand up

mlerner:
This PR implements `safe mode` for `lnd` by disabling all RPC and automated force close requests, as described in the original issue here: https://github.com/lightningnetwork/lnd/issues/3287

- [x] Add new --safemode config parameter to the lnd binary.

- [x] If safe mode is enabled, reject all RPC level force close requests.

- [x] If safe mode is enabled, reject all automated force close requests by channel arbitrators.

### DIFF
--- a/config.go
+++ b/config.go
@@ -273,6 +273,7 @@ type config struct {
 
 	UnsafeDisconnect   bool   `long:"unsafe-disconnect" description:"Allows the rpcserver to intentionally disconnect from peers with open channels. USED FOR TESTING ONLY."`
 	UnsafeReplay       bool   `long:"unsafe-replay" description:"Causes a link to replay the adds on its commitment txn after starting up, this enables testing of the sphinx replay logic."`
+	SafeMode           bool   `long:"safemode" description:"Start in safe mode that forbids broadcasting of all commitment transactions"`
 	MaxPendingChannels int    `long:"maxpendingchannels" description:"The maximum number of incoming pending channels permitted per peer."`
 	BackupFilePath     string `long:"backupfilepath" description:"The target location of the channel backup file"`
 
--- a/contractcourt/chain_arbitrator.go
+++ b/contractcourt/chain_arbitrator.go
@@ -150,6 +150,10 @@ type ChainArbitratorConfig struct {
 	// NotifyClosedChannel is a function closure that the ChainArbitrator
 	// will use to notify the ChannelNotifier about a newly closed channel.
 	NotifyClosedChannel func(wire.OutPoint)
+
+	// SafeMode is a boolean value that determines whether the ChainArbitrator
+	// should be able to force close channels or not
+	SafeMode bool
 }
 
 // ChainArbitrator is a sub-system that oversees the on-chain resolution of all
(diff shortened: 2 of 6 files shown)


contractcourt/chain_arbitrator.go
```diff
@@ -232,6 +236,9 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 		ShortChanID: channel.ShortChanID(),
 		BlockEpochs: blockEpoch,
 		ForceCloseChan: func() (*lnwallet.LocalForceCloseSummary, error) {
+			if c.cfg.SafeMode {
```
halseth: this will beed to be checked when the force close request is first sent. Otherwise I think we'll get into a state where we will attempt to force close the channel on every restart.

config.go
```diff
@@ -273,6 +273,7 @@ type config struct {
 
 	UnsafeDisconnect   bool   `long:"unsafe-disconnect" description:"Allows the rpcserver to intentionally disconnect from peers with open channels. USED FOR TESTING ONLY."`
 	UnsafeReplay       bool   `long:"unsafe-replay" description:"Causes a link to replay the adds on its commitment txn after starting up, this enables testing of the sphinx replay logic."`
+	SafeMode           bool   `long:"safemode" description:"Start in safe mode that forbids broadcasting of all commitment transactions"`
```
cfromknecht: perhaps we should call this something more descriptive than safe mode, e.g. `ForbidForceClose`?

contractcourt/chain_arbitrator.go
cfromknecht: indeed, and we should mark it before we mark the channel borked and save the force close, o/w we will broadcast if the node is restarted and we're not in safe mode

lnd.go
```diff
@@ -130,6 +130,11 @@ func Main(lisCfg ListenerCfg) error {
 		}
 	}()
 
+	if cfg.SafeMode {
+		ltndLog.Infof("Starting LND in safe mode." +
+			"All commitment transactions are forbidden in this state.")
```
carlaKC: logging nit" "broadcast of commitment transactions is forbidden"?

config.go
carlaKC: +1, I think the term safe in lightning is quite a strong word 😅

contractcourt/chain_arbitrator.go
halseth: I don't think we should mark the channel at all in safe mode though, just to not alter any state?

mlerner:
Thanks for the feedback @carlaKC @cfromknecht @halseth. I'm going to start updating the PR with your suggested changes.

contractcourt/chain_arbitrator.go
cfromknecht: sorry, check* it before we mark

yes we want to make sure no state is altered

lntest/itest/lnd_test.go
```diff
@@ -3546,6 +3546,52 @@ func testChannelForceClosure(net *lntest.NetworkHarness, t *harnessTest) {
...
+			Amt:     chanAmt,
+			PushAmt: pushAmt,
+		},
+	)
+
+	// Now that the channel is open, we try to force close the channel. Sending a
+	// force close message to a node in safe mode should fail.
+	ctxt, _ = context.WithTimeout(ctxb, channelCloseTimeout)
+	_, _, err = net.CloseChannel(ctxt, carol, chanPoint, true)
+	if err == nil {
```
Crypt-iQ: You could check whether or not the error matches the "can not force close channels" text

Crypt-iQ:
Might want to add a unit test to test the new behavior

halseth:
Friendly ping @mlerner

Issue #3302, opened 2019-07-11
watchtower: use tor controller to configure hidden service automatically


MDrollette:
The watchtower can currently use Tor for outbound connections, and a hidden service can be manually configured to point at a watchtower listener. But it should be possible for the watchtower to communicate with the Tor ControlPort to create its own hidden service in the same way that `lnd` does.

PR #3613, opened 2019-10-18
watchtower: automatically create watchtower onion service

orbitalturtle:
As described in #3302, this PR allows a watchtower to be configured to automatically create an onion service.

### DIFF
--- a/config.go
+++ b/config.go
@@ -215,15 +215,16 @@ type autoPilotConfig struct {
 }
 
 type torConfig struct {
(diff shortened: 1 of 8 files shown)


orbitalturtle:
Thanks for reviewing @Crypt-iQ , I'll look over your comments soon

Crypt-iQ:
Thanks for the PR @orbitalturtle ! We'll take this over and it will be included in the 0.10 release. Will give credit in commits where appropriate.

PR #3621, opened 2019-10-22
cnct: make stateless [PoC]

joostjager:
For anchor outputs, we'll need changes to the on-chain handling of commitment. The anchor outputs need to be swept and their resolution will likely show up in reports and/or accounting info (which we may make more accurate in the future). Also, the sweeping of anchor outputs starts already before the commit tx confirms (that is the whole point of anchoring down the commit tx).

The required changes in `ChannelArbitrator` are complicated because various bits of data are persisted. This pr explores the possibility to make it all stateless:

* Postpone deleting open channel data from the db until the channel is fully closed
* Always instantiate a chain watcher for channels (also pending closes) - enabled by having open channel data still available
* Remove briefcase and keep channel arbitrator and resolver state in memory.

Additional benefit is that this would also make it easier to deal with reorgs.

Note: this pr deletes all unit tests, which is definitely not the intention for a final pr


Roasbeef:
Closing due to inactivity.

PR #3627, opened 2019-10-23
input+sweep: extract CSV/CLTV count into TimeLockType method

guggero:
This is a follow-up PR for #3557 where we decided to also extract the CSV/CLTV count of the standard witness types into a method of the `WitnessType` interface.


input/witnessgen.go
```diff
@@ -35,11 +35,32 @@ type WitnessType interface {
...
 	// AddWeightEstimation adds the estimated size of the witness in bytes
-	// to the given weight estimator and returns the number of
-	// CSVs/CLTVs used by the script.
-	AddWeightEstimation(estimator *TxWeightEstimator) (int, int, error)
+	// to the given weight estimator.
+	AddWeightEstimation(estimator *TxWeightEstimator) error
+
+	// TimeLockType returns the type of time lock used by a script of this
+	// witness type (if any).
+	TimeLockType() TimeLockType
```
halseth: Does this mean that the interface cannot be extended to types having both CSV and CLTV?

joostjager: Oh, these are coming for anchors, isn't it?

guggero: At the moment, yes. I think we can extend this if we need it. And since the witness only spends one path of an output it is unlikely that there would be a CSV and CLTV in the same script path. At least for the use cases of lnd and loop this should be sufficient. But I can add a type `Both` if you think that could be used in the future.

halseth: Looking at the usage of `getWeightEstimate` it seems that the num,ber of csv/cltv is only used for logging. Should we do ourselves a favor and just remove it entirely ?

guggero: Yeah, that's the other possibility... I'm not sure how important that log statement there really is. So I didn't just want to remove stuff. But it would remove quite a bit of code.

joostjager: Or just make a nice helper method that shows how many of each witness type we have in the sweep tx? I would like that. Also good for anchor out dev.

halseth: sgtm

guggero: Ok, I changed it according to @joostjager's suggestion.

sweep/txgenerator.go
```diff
@@ -268,28 +269,51 @@ func getWeightEstimate(inputs []input.Input) ([]input.Input, int64, int, int) {
...
+	sort.Strings(keys)
+
+	// Return a nice string representation of the counts by comma joining a
+	// slice.
+	var parts []string
+	for _, witnessType := range keys {
+		part := fmt.Sprintf("%d %s", counts[witnessType], witnessType)
+		parts = append(parts, part)
+	}
+	return strings.Join(parts, ", ")
```
halseth: 👍

PR #3632, opened 2019-10-23
rpcserver: ensure server has started before CloseChannel

wpaulino:
We do this as a general sanity check to ensure channels cannot be closed while either the backend is still syncing or the server is still starting.


joostjager:
Are there any other operations that should be blocked before startup is finished?

wpaulino:
Went through the RPCs in the main RPC server and I don't think so.

PR #3633, opened 2019-10-24
channeldb: isolate migrations

### LINKED ISSUE #3628: Isolate migration logic
In the past, we've had instances of migration code unintentionally breaking by changes to `channeldb`. A migration typically reads database version `n-1` and writes it as version `n`. The code that does the read/write should remain stable and must not be updated. Otherwise an older migration may break and this will only surface when users running an old version upgrade.

The usual way to deal with this is to duplicate all dependencies of the migration. This however is something that must not be forgotten. Also, there are currently migrations that still rely on the main `channeldb` code. Those are at risk of being broken at some point in the future.

The plan to deal with this is to put every migration in its own sub-package of `channeldb`. Migration package names could be something along the lines of `migration_01_edgepolicies`. Because circular references are not allowed and `channeldb` references all migrations, it won't be possible anymore for a migration to use code from `channeldb`. It is allowed for migrations to be dependent on previous migrations. This is useful, because the write code of migration `n` can be used as read code for migration `n+1`. The dependencies will be visible as imports of a migration package and it will be simple to check for any violations.

Ideally we'd go back in time to every commit that introduces a migration and move that code into its own package together with its dependencies at that point.

The minimum thing that we should do is to take all currently existing migrations and move those in a separate package `migration_01_to_11`. That will at least seal what we have.

joostjager:
Fixes #3628


channeldb/migration_01_to_11/channel_cache_test.go
```diff
@@ -0,0 +1,105 @@
+package migration_01_to_11
+
+import (
+	"reflect"
+	"testing"
+)
+
+// TestChannelCache checks the behavior of the channelCache with respect to
+// insertion, eviction, and removal of cache entries.
+func TestChannelCache(t *testing.T) {
```
Roasbeef: Do we really need to copy over all the tests as well? Can't we just copy the call graph of any encode/decode functions used in the migration.

joostjager: Previously I found that I needed some test code because it was reused. Left all of it in, but I think that also had a lot of deps that prevents more removal of code. Will look at the diff again.

joostjager: Ok, removed many more lines. Down to 5k lines now.

I used `staticcheck` from https://github.com/dominikh/go-tools with the `-unused.whole-program` flag to find more unused code.

joostjager: Ok, removed many more lines. Down to 5k lines now.

I used `staticcheck` from https://github.com/dominikh/go-tools with the `-unused.whole-program` flag to find more unused code.

joostjager:
Unit tests failing. Even by only removing code, I seem to have broken something. Will investigate.

joostjager:
Test fixed. Problem was only an uninitialized logger. Not something magical triggered by code removal. Ready for review again.

channeldb/migration_01_to_11/db.go
```diff
@@ -318,7 +227,7 @@ func createChannelDB(dbPath string) error {
 		}
 
 		meta := &Meta{
-			DbVersionNumber: getLatestDBVersion(dbVersions),
```
halseth: hm, can't this code be removed?

channeldb/migration_01_to_11/codec.go
```diff
@@ -48,12 +48,6 @@ type UnknownElementType struct {
 	element interface{}
 }
 
-// NewUnknownElementType creates a new UnknownElementType error from the passed
-// method name and element.
-func NewUnknownElementType(method string, el interface{}) UnknownElementType {
```
halseth: ok, guess the code is removed here.

channeldb/migration_01_to_11/db.go
joostjager: Always set to 0 now. It is just for migration testing

channeldb/db.go
```diff
@@ -266,10 +267,6 @@ func createChannelDB(dbPath string) error {
 			return err
 		}
 
-		if _, err := tx.CreateBucket(paymentBucket); err != nil {
```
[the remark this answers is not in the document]
joostjager: Seems like a mistake. Good catch.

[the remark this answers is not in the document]
joostjager: Ok, not a mistake. We just don't use that payment bucket anymore and if we instantiate a fresh db, that bucket does not longer need to be created.

channeldb/migration_01_to_11/meta.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Meta structure holds the database meta information.
+type Meta struct {
+	// DbVersionNumber is the current schema version of the database.
+	DbVersionNumber uint32
+}
+
+// FetchMeta fetches the meta data from boltdb and returns filled meta
+// structure.
+func (d *DB) FetchMeta(tx *bbolt.Tx) (*Meta, error) {
```
halseth: Is this code used by anything now?

Also I wouldn't mind removing tests, since they could depend on code that otherwise could be deleted.

joostjager: The code is used to do `syncVersion` for migration tests. But that isn't really necessary to test every time. Added a commit that removes it.

Non-migration tests are all removed (afaik), the `_test.go` files that remain just contain some vars and helpers that are used in the migration tests.

PR #3636, opened 2019-10-24
discovery: skip chanann validation for assumechanvalid

halseth:
AssumeChannelValid toggles whether or not the router will verify
signatures for incoming channel announcements. We skip this to increase
performance on low powered devices, and since we are not checking
unspentness of the channel in this mode, they are easily spoofed anyway.

### DIFF
--- a/discovery/gossiper.go
+++ b/discovery/gossiper.go
@@ -229,6 +229,13 @@ type Config struct {
 	// This prevents ranges with old start times from causing us to dump the
 	// graph on connect.
 	IgnoreHistoricalFilters bool
+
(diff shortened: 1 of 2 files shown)


halseth:
Discussed this one with @cfromknecht a bit earlier, and the question is whether we want to go down this route, improving `assumechanvalid`. I think it is worth doing it for now, since it _drastically_ speeds up graph sync on low powered devices, with no real downside. When we have an alternative to `assumechanvalid` then this will go away ofc.

halseth:
Actually meant to close this after offline discussion, sorry.

Issue #3420, opened 2019-08-21
htlcswitch+rpcserver: add new streaming ForwardingEvent API


Roasbeef:
Today users are able to poll the `ForwardingHistory` API in to obtain the past history of successful HTLC forwards. This is useful to be able to gauge if your node is forwarding payments at all, but only exposing successful payment settles doesn't provide enough information to be able to make channel management/allocation decisions. The original API didn't also move to _persist_ HTLC failures as they can be much more numerous than failures since they have low cost today in the network. However, we can still surface this information based on our in-memory state, and rely on projects such as `lndmon` to persist this information for analysis purposes. 

In order to provide node operators with additional information that they can use to optimize the channel balances and placement, we should expose a _streaming_ variant of the `ForwardingHistory` API that also includes HTLCs accepts _and_ fails. The final API should be able to export events pertaining to: successful HTLC forwards, multi-hop failures, failed forwarding attempts and successful payment settlement.

# Steps To Completion

- [ ] Extend the [`htlcswitch`](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/switch.go#L176) to expose a subscription API to allow clients to be notified of relevant HTLC forwarding events. 

- [ ] A sample draft of relevant forwarding events may resemble something like the following:
```go
type ForwardingEvent struct {
    IncomingChannel, OutgoingChannel lnwire.ShortChannelID

    IncomingTimeLock, OutgoingTimeLock uint32
    
    IncomingAmt, OutgoingAmt lnwire.Milisatoshi

    PayHash [32]byte
}

type LinkFailEvent struct { // only for HTLCS that come in, but fail outgoing link requirements (capacity, fees, etc)
    FailReason FailReason // new enum exposing various failure reasons
 
    ForwardingEvent
}

type ForwardingFailEvent  struct { // no error as multi-hop errors are opaque
    ElapsedTime time.Duration // amount of time it took for us to recv the multi-hop fail

    ForwardingEvent 
}

type SettleEvent struct {
    ElapsedTime time.Duration 

    ForwardingEvent
}
```

- [ ] Create a new `SubscribeForwardingEvents` streaming API that maps something similar to the structs above to actual protos. Note that since we don't persist the information in this form, we won't support any sort of backlog filling.

carlaKC:
I've started having a look at implementing this, starting off looking at the code in `switch.go#handlePacketForward`. 

A few questions that have come up:
[...]

PR #3639, opened 2019-10-25
[WIP]: HTLC Forwarder Events

carlaKC:
First shot at a HTLC Forwarder events (as specified in #3420). PR is not ready for review at present, but I'd like to get some early feedback on this design. 

Specifically:
1. Bubbling errors up from `canSendHTLC` with `linkFailure`
2. Use of an interface to mock HTLCNotifier in switch tests
3. Placement of `FailureReason` and `ForwardFailType` - both feel like they belong in `link.go` but that creates a circular dep; perhaps a new package just with the errors?

To dos:
1. Figure out circuits & set payment hash for settle/fail events
2. Add timestamps to events (and elapsed time to settled/forward fail)
3. Add subscribe RPC
4. Tests can be refined, I just wanted to check functionality


htlcnotifier/htlcnotifier.go
```diff
@@ -0,0 +1,300 @@
...
+
+// SubscribeHTLCEvents returns a subscribe.Client that will receive updates
+// any time the Server is made aware of a new event.
+func (h *HTLCNotifier) SubscribeHTLCEvents() (*subscribe.Client, error) {
+	return h.ntfnServer.Subscribe()
+}
+
+// ForwardingEvent represents a HTLC that was forwarded through our node.
+type ForwardingEvent struct {
+	// IncomingChannel is the channel that the HTLC arrived at our node on.
```
joostjager: I'd attach a timestamp as close to the origin as possible.

htlcnotifier/htlcnotifier.go
```diff
@@ -0,0 +1,300 @@
...
+
+// SettleEvent represents a HTLC that was forwarded by our node and successfully
+// settled.
+type SettleEvent struct {
+	//TODO(carla): Add elapsed time field.
+
+	// ForwardingEvent provides the details of the HTLC forward on our node
+	// which was settled.
+	ForwardingEvent
+}
```
joostjager: Nodes forward htlcs, but - on the same channels - may also send or receive payments for which they are the endpoint. For profitability calculation that takes into account balance, those events are important too. Will they be communicated over this same notifier channel, or do they go a different way?

htlcnotifier/interface.go
```diff
@@ -0,0 +1,21 @@
...
+	// NotifyForwardEvent is called when we successfully forward a HTLC.
+	NotifyForwardingEvent(f ForwardingEvent)
+
+	// NotifyLinkFailEvent is called when we fail to forward a HTLC over our
+	// outgoing link.
+	NotifyLinkFailEvent(reason FailReason, f ForwardingEvent)
+
+	// NotifySettleEvent is called when a HTLC we previously successfully
+	// forwarded is settled.
+	NotifySettleEvent(f ForwardingEvent)
```
joostjager: Probably need some way to correlate the forwarding event and the outcome (fail/settle). Payment hash is not enough, because it may be reused.

Roasbeef:
>  Is there current ForwardingLog completely safe atm?

What does safe mean here? It's possible that a panic or crash fault causes us to not write 100% of the data in the in-memory buffer. However, on a clean shutdown, we'll attempt to flush all items on to disk.

htlcnotifier/htlcnotifier.go
carlaKC: From the insights side of things, I see our node's HTLCs as a separate concern to profitability. They capture "personal utility" gained from a channel, which could be another aspect to factor into scoring; routing nodes would prioritize profitability, non-routing would be more concerned with utility. 

But from a `htlcNotifier`, I think it makes sense to include them to keep the notifier generic. I'd probably go with a `our_htlc` bool in the settle/fail event types (since these HTLCs can't be forwarded/have link failures, but could also go with another event type entirely?

joostjager: I was mainly thinking that we also need "our htlcs" to be able to reconstruct the channel balance at the time of an event.

htlcnotifier/interface.go
carlaKC: Looking at using circuit `inKey`/`outKey` to identify HTLCs?
Forward events can have the `outKey` of the outgoing HTLC and they can be matched to settle/fail events.

joostjager:
> What does safe mean here? It's possible that a panic or crash fault causes us to not write 100% of the data in the in-memory buffer. However, on a clean shutdown, we'll attempt to flush all items on to disk.

I'd then call that not safe, because we can miss events. I was hoping that the replay on startup would re-offer the events to the forwarding log, in case it crashed. But with the flushing, the switch has probably already fully completed the htlc lifecycle before we write. Does this mean that we have an accounting gap there?

htlcnotifier/htlcnotifier.go
carlaKC: > I think the main challenge is to decide how consistent the stream of data coming from the notifier is going to be and carefully document that in the code. 

If we want restart safety and assurances that we'll receive every HTLC that's almost a separate project making a persisted subscription of some sort. I see the fear of doing a lot of work then having to redo it, but I think that if we choose to go with a safer option we can just rip the subscribe backend out, which won't be too much work since we're not spending time coding that logic up now?

> The channel fitness subsystem needs to be able to deal with any inconsistencies that may occur because of for example restarts (replays), delays and out of order delivery (not sure if that last thing happens).
 
If the HTLCs are timestamped and uuid'ed with some combination of `inKey`/`outKey` (soundness of that approach pending further investigation) and the subscription is appropriately documented, would you say that's sufficient?

carlaKC:
> I think the main challenge is to decide how consistent the stream of data coming from the notifier is going to be and carefully document that in the code. 

If we want restart safety and assurances that we'll receive every HTLC that's almost a separate project making a persisted subscription of some sort. I see the fear of doing a lot of work then having to redo it, but I think that if we choose to go with a safer option we can just rip the subscribe backend out, which won't be too much work since we're not spending time coding that logic up now?

> The channel fitness subsystem needs to be able to deal with any inconsistencies that may occur because of for example restarts (replays), delays and out of order delivery (not sure if that last thing happens).
 
If the HTLCs are timestamped and uuid'ed with some combination of `inKey`/`outKey` (soundness of that approach pending further investigation) and the subscription is appropriately documented, would you say that's sufficient?

joostjager:
Can you make it so that the tlv payload is exposed for htlc (fail) events over rpc? It is needed for Whatsat :grimacing:

carlaKC:
> Can you make it so that the tlv payload is exposed for htlc (fail) events over rpc? It is needed for Whatsat 😬

For a small stake in this hot new business, anything is possible 😉 

Assuming the use case of custom TLV types (unknown to lnd, that is), would TLV just be exposed as a `[]byte` which end users can parse if they understand them?

joostjager:
Yes, just `map<uint64, bytes>` in the proto. We are also planning to do that for the settled invoices: https://github.com/lightningnetwork/lnd/issues/3670

joostjager:
Also, it is only needed for htlcs that fail at the endpoint, not necessarily for intermediate nodes. To be clear, this is totally a nice to have. A use case to keep in the back of your mind only.

carlaKC:
> Also, it is only needed for htlcs that fail at the endpoint. It is not needed for intermediate nodes. To be clear, this is totally a nice to have. A use case to keep in the back of your mind only.

Of course, but I'll keep it in mind nonetheless 😄
It'll be an easy addition to a solution which flags HTLCs that terminate at our node.

carlaKC:
Not ready for review yet, just backing up what I have so far. 
Going to have to look at moving successful forward events onto switch level because `outKey` is set on link level.

halseth:
High level look at this PR I think it is going in the right direction 👍

htlcnotifier/interface.go
```diff
@@ -0,0 +1,21 @@
...
+
+	// NotifyForwardingFailEvent is called when a HTLC we previously
+	// successfully forwarded fails down the line.
+	NotifyForwardingFailEvent(failType ForwardFailType, f ForwardingEvent)
```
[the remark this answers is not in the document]
carlaKC: 👍 updated in the more recent WIP draft, cleaned these up into individual event types with a common identifier struct (ID, to be renamed to something better) rather than trying to tack failures onto `ForwardingEvent` when needed.

htlcnotifier/interface.go
```diff
@@ -0,0 +1,21 @@
+package htlcnotifier
```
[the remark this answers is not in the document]
carlaKC: Yeah makes sense, also helps with naming redundancy `switch.HTLCNotifier` reads a lot better than `htlcnotifier.Notifier`

htlcswitch/link.go
```diff
@@ -2272,16 +2279,27 @@ func (l *channelLink) CheckHtlcTransit(payHash [32]byte,
...
+
+	return err.message
 }
 
-// htlcSatifiesPolicyOutgoing checks whether the given htlc parameters satisfy
-// the channel's amount and time lock constraints.
+// linkFailure represents failure to forward a HTLC along the link. It contains
+// an appropriate lnwire message and the specific reason for failure.
+type linkFailure struct {
+	message    lnwire.FailureMessage
```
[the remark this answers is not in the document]
carlaKC: Using `ForwardingError` and using existing `lnwire` errors would be good, since there are a lot of duplicates. What about defining the extra errors in the `htlcnotifier` (rather than `lnwire`) package which implement `FailureMessage` and then mapping custom errors back to their broadcast-friendly error? Eg, insufficient balance -> temp channel failure.

htlcnotifier/htlcnotifier.go
```diff
@@ -0,0 +1,300 @@
...
+	// the current block height.
+	FailReasonExpiryTooSoon
+
+	// FailReasonExpiryTooFar is used when the expiry on a HTLC is too far from
+	// the current block height.
+	FailReasonExpiryTooFar
+
+	// FailReasonInsufficientBandwidth is used when our available balance is
+	// less than the HTLC amount.
+	FailReasonInsufficientBandwidth
```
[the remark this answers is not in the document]
halseth: If the error reasons in lnwire are not suffiecient, we could define a custom error type that can act as a oneof between lnwire and custom codes

htlcswitch/switch.go
```diff
@@ -485,14 +485,18 @@ func (s *Switch) forward(packet *htlcPacket) error {
 				return err
 			}
 
-			var failure lnwire.FailureMessage
+			var failure *ForwardingError
```
halseth: what `bool` does the commit msg refer to?

htlcswitch/htlcnotifier/htlcnotifier.go
```diff
@@ -117,6 +117,20 @@ type ForwardingEvent struct {
 // provide detailed information about failures.
 type FailureDetail int
 
+const (
+	// FailureDetailNone is used when there are no further failure details
+	// beyond lnwire's failure message.
+	FailureDetailNone FailureDetail = iota
```
halseth: Nice solution 👍

MrManPew:
@carlaKC Is this still in Draft or should it be moved to Needs Review?

carlaKC:
> @carlaKC Is this still in Draft or should it be moved to Needs Review?

Still a WIP, working on some itests for this and then it will be good to go. Should be up early this week @MrManPew

carlaKC:
Draft PR replaced with #3781.

Issue #3587, opened 2019-10-09
Add anchor output to the commitment transaction


joostjager:
In the current commitment format, the miner fee is negotiated through the `update_fee` message. This creates several complications:
* It is currently not possible to reject a fee rate. The only option at that point is to force close the channel. Nodes may have different sources for fee estimation and may unintentionally disagree about the fee to use.
* Even if a satisfactory fee has been negotiated, it may turn out not to be enough to get it confirmed when one of the parties actually goes to chain. The party that publishes the commitment tx has its `to_local` output timelocked and can't use that to cpfp the transaction.

A possible solution (originally proposed by @halseth on the [lightning-dev mailing list](https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-January/000943.html)) is to add a so called anchor output to the commitment transaction. This is a non-timelocked output that can be used to cpfp the transaction in order to get it confirmed. 

With the anchor output present, the fee on the commitment transaction itself can be minimal as both parties have the opportunity to attach additional fees.

halseth:
Ideally we would have a no-fee commitment transaction, and add all fees using CPFP, but that would probably require something like package relay in order to ensure propagation: 
https://github.com/bitcoin/bitcoin/issues/14895
https://github.com/bitcoin/bitcoin/pull/16401

Having a "minimal but sufficient fee" for propagation on the commitment should therefore be the goal. We might have to keep fee negotiation using `update_fee` for now, to ensure that we have this sufficient fee during fee spikes. This will still leave the first issue in the list above though.

Roasbeef:
Fixed by #3829.

PR #3644, opened 2019-10-29
cnct: commit sweep without nursery

joostjager:
This PR (finally) removes the dependency on the legacy utxo nursery from the commit sweep resolver.

The reason for doing this now, is that the new anchor output commitment format will also csv-lock the `to_remote` output. This requires changes and testing in this area and it seems like a good opportunity to clean up this technical debt.

Subtask of #3587


joostjager:
Note to sef: reporting on rpc eve

joostjager:
Unit test added, ready for review

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	ctx := newCommitSweepResolverTestContext(t, &res)
+	ctx.resolve()
+
+	ctx.notifier.confChan <- &chainntnfs.TxConfirmation{
+		BlockHeight: testInitialBlockHeight - 1,
+	}
+
+	// Notify initial block height.
+	ctx.notifyEpoch(testInitialBlockHeight)
```
halseth: should check that sweep is not attempted before the  maturity height

contractcourt/channel_arbitrator.go
```diff
@@ -129,13 +129,31 @@ type ChannelArbitratorConfig struct {
...
+
 // ContractReport provides a summary of a commitment tx output.
 type ContractReport struct {
 	// Outpoint is the final output that will be swept back to the wallet.
 	Outpoint wire.OutPoint
 
-	// Incoming indicates whether the htlc was incoming to this channel.
-	Incoming bool
+	// Type indicates the type of the reported output.
+	Type ContractReportOutputType
```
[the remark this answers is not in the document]
joostjager: `OutputType` is a bit too bare for me. Went for `ReportOutputType`

rpcserver.go
```diff
@@ -2451,25 +2451,51 @@ func (r *rpcServer) arbitratorPopulateForceCloseResp(chanPoint *wire.OutPoint,
 	reports := arbitrator.Report()
 
 	for _, report := range reports {
-		incoming := report.Type == contractcourt.ReportOutputIncomingHtlc
+		switch report.Type {
```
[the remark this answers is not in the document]
joostjager: Nursery is still the first to populate it. Could possibly switch the order and have nursery only modify it, but I'd rather leave it as is now and do the real fix when the nursery is gone.

rpcserver.go
```diff
@@ -2500,15 +2526,6 @@ func (r *rpcServer) nurseryPopulateForceCloseResp(chanPoint *wire.OutPoint,
 	// wallet.
 	forceClose.LimboBalance = int64(nurseryInfo.limboBalance)
 	forceClose.RecoveredBalance = int64(nurseryInfo.recoveredBalance)
-	forceClose.MaturityHeight = nurseryInfo.maturityHeight
```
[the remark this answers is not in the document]
joostjager: No, because it still reports on some htlcs when they are not yet tracked fully in their resolvers

contractcourt/commit_sweep_resolver_test.go
joostjager: added

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -0,0 +1,178 @@
...
+	// A new block arrives. The commit tx confirmed at height -1 and the csv
+	// is 3, so a spend will be valid in the first block after height +1.
+	ctx.notifyEpoch(testInitialBlockHeight + 1)
+
+	<-ctx.sweeper.sweptInputs
+
+	ctx.waitForResult()
+}
+
+type commitSweepResolverTestContext struct {
```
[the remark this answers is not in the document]
joostjager: moved

contractcourt/htlc_incoming_resolver_test.go
```diff
@@ -251,7 +251,7 @@ func (i *incomingResolverTestContext) waitForResult(expectSuccessRes bool) {
 	}
 
 	if !expectSuccessRes {
-		if err != nil {
+		if i.nextResolver != nil {
```
cfromknecht: does seem like this case is executed, should it be?

input/input.go
```diff
@@ -107,6 +115,23 @@ func NewBaseInput(outpoint *wire.OutPoint, witnessType WitnessType,
 	return &input
 }
 
+// NewCsvBaseInput assembles a new csv-locked input that can be used to
+// construct a sweep transaction.
+func NewCsvBaseInput(outpoint *wire.OutPoint, witnessType WitnessType,
```
cfromknecht: nit: naming suggestion `NewCSVInput`, idt we need base in the name

contractcourt/channel_arbitrator.go
```diff
@@ -129,13 +129,31 @@ type ChannelArbitratorConfig struct {
 	ChainArbitratorConfig
 }
 
+// ReportOutputType describes the type of output that is being reported
+// on.
+type ReportOutputType uint8
```
cfromknecht: 👍

contractcourt/chain_arbitrator.go
```diff
@@ -45,6 +45,19 @@ type ResolutionMsg struct {
 	PreImage *[32]byte
 }
 
+// UtxoSweeper defines the sweep functions that contract court requires.
+type UtxoSweeper interface {
```
cfromknecht: add to `interfaces.go`?

rpcserver.go
```diff
@@ -2450,23 +2451,51 @@ func (r *rpcServer) arbitratorPopulateForceCloseResp(chanPoint *wire.OutPoint,
...
+			}
+
+			if htlc.MaturityHeight != 0 {
+				htlc.BlocksTilMaturity =
+					int32(htlc.MaturityHeight) - currentHeight
+			}
+
+			forceClose.PendingHtlcs = append(forceClose.PendingHtlcs, htlc)
+
+		default:
```
cfromknecht: what about the third type?

contractcourt/chain_arbitrator.go
```diff
@@ -112,8 +125,7 @@ type ChainArbitratorConfig struct {
 	// the process of incubation. This is used when a resolver wishes to
 	// pass off the output to the nursery as we're only waiting on an
 	// absolute/relative item block.
-	IncubateOutputs func(wire.OutPoint, *lnwallet.CommitOutputResolution,
-		*lnwallet.OutgoingHtlcResolution,
+	IncubateOutputs func(wire.OutPoint, *lnwallet.OutgoingHtlcResolution,
```
cfromknecht: one down, two to go!

contractcourt/htlc_incoming_resolver_test.go
joostjager: Not sure if I get what you mean, but in some test cases we don't expect the incoming contest resolver to transform into another resolver. This asserts that.

input/input.go
joostjager: renamed

rpcserver.go
joostjager: There are three types listed above?

contractcourt/chain_arbitrator.go
joostjager: eventually we'll get there

contractcourt/chain_arbitrator.go
joostjager: added

contractcourt/commit_sweep_resolver.go
```diff
@@ -59,155 +122,109 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
-	}
+	amt := btcutil.Amount(
+		c.commitResolution.SelfOutputSignDesc.Output.Value,
+	)
 
-	// We're dealing with our commitment transaction if the delay on the
-	// resolution isn't zero.
-	isLocalCommitTx := c.commitResolution.MaturityDelay != 0
+	c.reportLock.Lock()
+	c.currentReport = &ContractReport{
```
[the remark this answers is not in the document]
joostjager: Constructor pr: https://github.com/lightningnetwork/lnd/pull/3665

rpcserver.go
cfromknecht: ah i missed that of the cases had two conditions.

contractcourt/channel_arbitrator.go
```diff
@@ -129,13 +129,31 @@ type ChannelArbitratorConfig struct {
...
+	// the commitment tx.
+	ReportOutputIncomingHtlc ReportOutputType = iota
+
+	// ReportOutputOutgoingHtlc is an outgoing hash time locked contract on
+	// the commitment tx.
+	ReportOutputOutgoingHtlc
+
+	// ReportOutputDirect is an uncontested output on the commitment
+	// transaction paying to us directly.
+	ReportOutputDirect
```
halseth: `ReportOutputToUs`?

halseth: nit: add type in the commit where it is first used

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -0,0 +1,186 @@
...
+		t:        t,
+	}
+}
+
+func (i *commitSweepResolverTestContext) resolve() {
+	// Start resolver.
+	i.resolveErr = make(chan error, 1)
+	go func() {
+		var err error
+		i.nextResolver, err = i.resolver.Resolve()
```
halseth: needs mutex to access `nextResolver`? Or can send it on channel

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// Notify initial block height. The csv lock is still in effect, so we
+	// don't expect any sweep to happen yet.
+	ctx.notifyEpoch(testInitialBlockHeight)
+
+	select {
+	case <-ctx.sweeper.sweptInputs:
+		t.Fatal("no sweep expected")
+	case <-time.After(time.Second):
```
halseth: can probably lower this to the order of milliseconds

contractcourt/channel_arbitrator.go
joostjager: Everything that we report is (potentially) to us. But agreed that `Direct` is also pretty vague.

contractcourt/channel_arbitrator.go
```diff
@@ -1850,14 +1850,15 @@ func (c *ChannelArbitrator) prepContractResolutions(
...
-		resKit.Quit = make(chan struct{})
 		resolver := &commitSweepResolver{
 			commitResolution: *contractResolutions.CommitResolution,
 			broadcastHeight:  height,
 			chanPoint:        c.cfg.ChanPoint,
-			ResolverKit:      resKit,
 		}
 
+		resKit.Quit = make(chan struct{})
+		resolver.AttachResolverKit(resKit)
```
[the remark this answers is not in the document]
joostjager: I'll revisit this (gh) comment later. Possibly #3665 goes in first and this disappears.

contractcourt/channel_arbitrator.go
joostjager: It is used in this commit in `arbitratorPopulateForceCloseResp` ?

contractcourt/commit_sweep_resolver_test.go
joostjager: it seems safe to me, because `waitForResult` waits for the `resolveErr` first

contractcourt/commit_sweep_resolver_test.go
joostjager: isn't that possibly causing travis flakes? I thought that because everything is executed in parallel, it is ok to take a more generous timeout

contractcourt/htlc_incoming_resolver_test.go
cfromknecht: my comment was re: that the assertion changed, but none of the code did. is this a stricter check now that still passes?

cfromknecht: it appears to be the latter, as if `err != nil` the test would have already failed

contractcourt/commit_sweep_resolver_test.go
halseth: Yeah, but it is not a very nice assumption that that will always be done. Just send on a shared channel instead (I think this was done in a different PR?)?

contractcourt/commit_sweep_resolver_test.go
halseth: But this is the happy flow, so if the timeout is "too low" it will just silently succeed. In the event a bug is introduced such that inputs get swept, I think it will be caught in most cases also with a lower timeot.

contractcourt/channel_arbitrator.go
halseth: only `ReportOutputIncomingHtlc` is used there.

halseth: but this is no blocker ofc

halseth: Maybe `unencumbered` is what we use other places in the codebase?

joostjager: Renamed

joostjager: Ah, was confused by 'type' vs 'value'. Moved to next commit

contractcourt/commit_sweep_resolver_test.go
joostjager: Fixed

contractcourt/commit_sweep_resolver_test.go
joostjager: Sounds that lowering it increases the risk of something slipping through. With parallel execution, I think there isn't much to gain? Lowered to 100 ms nevertheless.

contractcourt/channel_arbitrator.go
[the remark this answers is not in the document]
joostjager: It went in and the problem disappeared

joostjager:
Rebased on top of resolver constructors

contractcourt/channel_arbitrator.go
```diff
@@ -141,6 +141,10 @@ const (
 	// ReportOutputOutgoingHtlc is an outgoing hash time locked contract on
 	// the commitment tx.
 	ReportOutputOutgoingHtlc
+
+	// ReportOutputUnencumbered is an uncontested output on the commitment
+	// transaction paying to us directly.
+	ReportOutputUnencumbered
```
halseth: nice name :)

contractcourt/commit_sweep_resolver.go
```diff
@@ -36,6 +39,9 @@ type commitSweepResolver struct {
 	// chanPoint is the channel point of the original contract.
 	chanPoint wire.OutPoint
 
+	currentReport *ContractReport
```
halseth: should have a godoc perhaps?

halseth: also needs to be initialized in the constructor

contractcourt/commit_sweep_resolver.go
```diff
@@ -76,155 +139,109 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
+	amt := btcutil.Amount(
+		c.commitResolution.SelfOutputSignDesc.Output.Value,
+	)
 
-	// We're dealing with our commitment transaction if the delay on the
-	// resolution isn't zero.
-	isLocalCommitTx := c.commitResolution.MaturityDelay != 0
+	c.reportLock.Lock()
+	c.currentReport = &ContractReport{
+		Outpoint:         c.commitResolution.SelfOutPoint,
```
halseth: most of these values are available at constructor time, we can update only what is known first here. (perhaps with a method that acquired the mutex?)

contractcourt/commit_sweep_resolver.go
```diff
@@ -313,6 +330,19 @@ func newCommitSweepResolverFromReader(r io.Reader, resCfg ResolverConfig) (
 	return c, nil
 }
 
+// report returns a report on the resolution state of the contract.
+func (c *commitSweepResolver) report() *ContractReport {
+	c.reportLock.Lock()
+	defer c.reportLock.Unlock()
+
+	if c.currentReport == nil {
```
halseth: having this be non-pointer we could just return it directly? Assuming it is always populated at constructor time

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -0,0 +1,187 @@
...
+	res := lnwallet.CommitOutputResolution{
+		SelfOutputSignDesc: input.SignDescriptor{
+			Output: &wire.TxOut{
+				Value: 100,
+			},
+		},
+		MaturityDelay: 3,
+	}
+
+	ctx := newCommitSweepResolverTestContext(t, &res)
```
halseth: check the report at various times during the test?

contractcourt/commit_sweep_resolver.go
joostjager: added

joostjager: `initReport` added

contractcourt/commit_sweep_resolver.go
joostjager: report initialized with what we know

contractcourt/commit_sweep_resolver.go
joostjager: removed pointer, but don't want to change the `report` return value to non-ptr, because for other resolvers the report is still optional. that can change once nursery is deleted completely

contractcourt/commit_sweep_resolver_test.go
joostjager: check added

PR #3647, opened 2019-10-29
build+lncli: default to hex encoding for byte slices

cfromknecht:
This commit swaps out golang/protobuf/jsonpb for a custom variant https://github.com/lightninglabs/protobuf-hex-display that
by default prints byte slices as hex, which is more useful for our
setting. Some existing wrapper structs are removed as they can now be
printed directly with the new jsonpb.

!!! NOTE !!!

This commit introduces a breaking change to lncli listinvoices since
payment hashes and preimages will now be printed in hex instead of
base64.


joostjager:
I am not worried about the breaking change to `lncli`. We try to keep our grpc api reasonably stable, but I think the cli is intended for use by humans and it shouldn't be a problem to change it without preceding deprecation.

Aren't there more display structs that can now be removed?

cfromknecht:
> I am not worried about the breaking change to lncli. We try to keep our grpc api reasonably stable, but I think the cli is intended for use by humans and it shouldn't be a problem to change it without preceding deprecation.

I have seen people use lncli in production-ish settings, so we should still try to announce as widely as possible to avoid breaking. I agree tho we can probably proceed faster given that it is mostly used as a dev tool.

> Aren't there more display structs that can now be removed?

Indeed, I found many more :)

halseth:
Users can stay on the old `lncli` if they need the old format.

halseth:
Needs update and rebase.

cfromknecht:
rebased

cmd/lncli/cmd_build_route.go
```diff
@@ -88,7 +88,7 @@ func buildRoute(ctx *cli.Context) error {
 		return err
 	}
 
-	printJSON(route)
```
halseth: can remove the `printJSON` method now?

halseth:
Since it's a breaking change, would perhaps be nice to get in before 0.9?

cmd/lncli/cmd_build_route.go
cfromknecht: no there are still a few others that use it, mostly for things that encode outpoints

cfromknecht:
@joostjager @halseth added a commit to reorder `getinfo` via the proto directly, so we no longer need the display struct. lmk what you think

PR #1658, opened 2018-07-31
[server]: Start Peers Asynchronously

cfromknecht:
This PR adds asynchronous starting of peers,
in order to avoid potential DOS vectors. Currently,
we block with the server's mutex while peers exchange
Init messages and perform other setup. Thus, a remote
peer that does not reply with an init message will
cause server to block for 15s per attempt.

We also modify the startup behavior to spawn
peerTerminationWatchers before starting the
peer itself, ensuring that a peer is properly
cleaned up if the initialization fails. Currently,
failing to start a peer does not execute the bulk
of the teardown logic, since it is not spawned
until after a successful Start occurs.

The final commit is purely just a code move to place
the relevant methods in closer proximity, and
organize them roughly in the expected execution
order.

Prerequesities:
 - [x] #1668 
 - [x] #1551

PR #1551, opened 2018-07-13
[htlcswitch]: revert replace link, ensure removed links are stopped

cfromknecht:
This PR reverts the switch's current `AddLink` behavior, such that it rejects duplicate links with the same channel id. At the moment, we will hot swap in another link without fully shutting down the first, which is dangerous as the two can share a mailbox.

In a related commit, we changed `RemoveLink` such that it called `go link.Stop()`, and processed in the background to avoid deadlocking on the forwarding index mutex. The result of this is that it is possible for a prior `Stop()` to still be executing by the time a new link is added with the same channel id.

In order to recreate the behavior of replacing the old link with a new one, `RemoveLink` has been altered to return the removed link if it was found, and to not call/spawn `link.Stop()` at all. Instead, the caller is now responsible  for stopping the link, which gets around the deadlocking w/in the switch.

Finally, though `AddLink` rejects duplicate links, we've modified the peer to first try and remove any links going by the same channel id. If one is found, we ensure that `link.Stop()` finishes [before moving on to adding the link to the switch](https://github.com/lightningnetwork/lnd/compare/master...cfromknecht:switch-revert-replace-link?expand=1#diff-32e1ab53c69bf3272bd9e4b51b9bb105R562).

PR #1166, opened 2018-05-02
Isolate Pending Channels in Switch

cfromknecht:
In this PR, we add additional protections and internal safety precautions against forwarding on pending channels. The bulk of the fix is to add an additional `pendingLinkIndex`, that allows pending links to be added. This allows them to receive their FundingLocked msg, but nothing else.

We also refactor much of the logic behind `UpdateShortChanID`, and fixes some potential causes of inconsistency in the in-memory state. We introduce a `RefreshShortChanID` method on `OpenChannel`, such that we can sync the `ShortChanID` with the current persistent state, which may have been modified by a different `OpenChannel` instance in the funding manager. This ensures that the `LightningChannel` used in the link has the latest `ShortChanID`, before it begins to perform state transtions and create `AddRefs` and/or `SettleFailRefs`, which critically depend on this accuracy.

In order to reconcile this new status of pending links, we've introduced a `mailOrchestrator`, responsible for managing the lifecyle of mailboxes, and facilitates post-facto delivery of messages in the event that the link has not been added already.

PR #982, opened 2018-03-30
server: cleanup persistent connection retries

cfromknecht:
Complementary to #981.

This commits changes the behavior of our connection
reestablishment, and resolves some minor issues that
could lead to uncancelled requests or an infinite
connection loop.

 - Will not attempt to Remove connection requests with
   an ID of 0. This can happen for reconnect attempts
   that get scheduled, but have not started at the
   time the server cancels the connection requests.

 - Adds a per-peer cancellation channel, that is
   closed upon a successful inbound or outbound
   connection. The goroutine spwaned to handle the
   reconnect by the peerTerminationWatch now
   selects on this channel, and skips reconnecting
   if it is closed before the backoff matures.

 - Properly computes the backoff when no entry in
   persistentPeersBackoff is found. Previously, a
   value of 0 would be returned, cause all subsequent
   backoff attempts to use a backoff of 0.

 - Cancels a peer's retries and remove connections
   immediately after receiving an inbound connection,
   to mimic the structure of OutboundPeerConnected.

Fixes #935, #941, #979, #982.

PR #3648, opened 2019-10-29
config+itest: allow unsafe disconnect by default

cfromknecht:
This commit beings the process of deprecating unsafe-disconnect. Many
moons ago this was disallowed to prevent concurency bugs surrounding
reconnect. Despite the name, it has been safe to enable this feature for
well over a year, as several PRs have been merged that addressed the
possible issues that existed when the feature was added.


config.go
```diff
@@ -271,7 +271,7 @@ type config struct {
 
 	Profile string `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65535"`
 
-	UnsafeDisconnect   bool   `long:"unsafe-disconnect" description:"Allows the rpcserver to intentionally disconnect from peers with open channels. USED FOR TESTING ONLY."`
+	UnsafeDisconnect   bool   `long:"unsafe-disconnect" description:"Allows the rpcserver to intentionally disconnect from peers with open channels."`
```
wpaulino: Log that it'll be deprecated in the next major release (0.10?).

lntest/itest/lnd_test.go
```diff
@@ -14412,14 +14410,16 @@ var testsCases = []*testCase{
...
-	},
-	{
-		name: "async bidirectional payments",
-		test: testBidirectionalAsyncPayments,
-	},
+	/*
+		{
+			name: "async payments benchmark",
+			test: testAsyncPayments,
+		},
```
wpaulino: Unintentional?

config.go
cfromknecht: can't really log since the default value is now set to true, so we can't tell if it was set manually or not.

lntest/itest/lnd_test.go
cfromknecht: yes, restored

config.go
wpaulino: Good point. We'll just need to make a note in the release notes.

config.go
```diff
@@ -389,6 +389,7 @@ func loadConfig() (*config, error) {
 			Dir:     defaultLitecoindDir,
 			RPCHost: defaultRPCHost,
 		},
+		UnsafeDisconnect:   true,
```
joostjager: Can't the field be completely removed now?

cfromknecht: that would break existing conf files or CLI commands, it will be fully removed in a later PR after more people have had enough time to adjust for the deprecation

cfromknecht:
> I find it hard to form an opinion about this without references to the original concurrency issues and PRs that fixed those. Perhaps getting a review from someone who reviewed the original fix PRs is more efficient.

good point, i hadn't thought of that 😂

config.go
joostjager: Oh yes, ofc, it can still be flagged

lntest/itest/lnd_test.go
```diff
@@ -2173,44 +2173,70 @@ func testOpenChannelAfterReorg(net *lntest.NetworkHarness, t *harnessTest) {
...
 }
 
-// testDisconnectingTargetPeer performs a test which
-// disconnects Alice-peer from Bob-peer and then re-connects them again
+// testDisconnectingTargetPeer performs a test which disconnects Alice-peer from
+// Bob-peer and then re-connects them again
 func testDisconnectingTargetPeer(net *lntest.NetworkHarness, t *harnessTest) {
 	ctxb := context.Background()
 
+	alice, err := net.NewNode("Alice", nil)
```
halseth: I don't think we usually name other nodes Alice and Bob? Not sure if it's a problem, but could possibly be confusing.

cfromknecht:
> Why do we want to allow this by default? IIRC it was only added to allow disconnecting nodes during testing, I don't think it had to do with any concurrency issues.

@halseth It was both. there were concurrency issues in the peer lifecycle and possibility of duplicate spawning duplicate mailboxes at the time we needed to test with this flag, which have since been resolved via #1658 #1166 #1551 #982 #821 and probably others i'm missing. prior to those changes the link's mailbox could have desynced and caused a channel closure, and preventing disconnects made it less likely to happen as a result of user behavior.

lntest/itest/lnd_test.go
cfromknecht: idt it's more confusing than have 40-50 carol and daves :P

halseth:
> > Why do we want to allow this by default?
> 
> There are subtle issues that can be resolved by simply disconnecting and reconnecting to your peer. For most who don't know about this option, they need to restart just to be able to disconnect from a channel partner. [Our own node operator guide](https://blog.lightning.engineering/posts/2019/08/15/routing-quide-1.html) even suggests using this option.

Ok, cool. I think that should be the argument for enabling this by default now, so maybe add that to the PR description? :)

lntest/itest/lnd_test.go
```diff
@@ -2173,44 +2173,70 @@ func testOpenChannelAfterReorg(net *lntest.NetworkHarness, t *harnessTest) {
 	closeReorgedChannelAndAssert(ctxt, t, net, net.Alice, chanPoint, false)
 }
 
-// testDisconnectingTargetPeer performs a test which
-// disconnects Alice-peer from Bob-peer and then re-connects them again
+// testDisconnectingTargetPeer performs a test which disconnects Alice-peer from
```
Roasbeef: In commit message: arguents -> arguments

lntest/harness.go
```diff
@@ -1195,28 +1195,25 @@ func (n *NetworkHarness) AssertChannelExists(ctx context.Context,
 
 	req := &lnrpc.ListChannelsRequest{}
 
-	var predErr error
-	pred := func() bool {
+	return wait.NoError(func() error {
```
Roasbeef: wiat -> wait 

in the commit message

PR #592, opened 2018-01-11
lnrpc: make ChannelPoint.funding_txid a protobuf oneof and change CloseChannel endpoint

wilmerpaulino:
Fixes #552.

This PR introduces the following changes:

* Makes `ChannelPoint`'s `funding_txid` a oneof between its `bytes` and `string` representation.
* Updates the rest of `lnd` and `lncli` to handle the new proto update from above.
* Changes the endpoint of `CloseChannel` to use the string representation `funding_txid_str` instead.

Issue #552, opened 2018-01-03
Force-Closing a channel over Rest fails


Richard87:
Hi, I cannot close a channel with either the rest API or `lncli`...

Rest API:
```
Client error: `DELETE https://lightningsale:***@lnd:8080/v1/channels/1198623c07ceb98ccfe4674e74d1f7b22840cc805e1a9491e9d17b88c32d98ec/0` resulted in a `400 Bad Request` response:
{"error":"type mismatch, parameter: channel_point.funding_txid, error: unsupported field type uint8","code":3}
```

Rest API (output index 1):
```
Client error: `DELETE https://lightningsale:***@lnd:8080/v1/channels/1198623c07ceb98ccfe4674e74d1f7b22840cc805e1a9491e9d17b88c32d98ec/1` resulted in a `400 Bad Request` response:
{"error":"type mismatch, parameter: channel_point.funding_txid, error: unsupported field type uint8","code":3}
```

Please advice :)

Also how do I force close a channel over Rest?

halseth:
What is the result of the `lncli` command you are using?

Richard87:
It works nicely :)

for example:
```
docker-compose exec lnd lncli closechannel 093a53310dffb963da5887abcc397227a6f70a355c86d75a6256719bb340869b        
{
        "closing_txid": "c076966d9069ffd70583e3f0165d302e0c1dafc35dd273b3cd23e347cd61d9e1"
}
```

MDrollette:
I think the cause of this problem is that the type of the [channel_point.funding_txid](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/rpc.proto#L526) is `bytes`, but there is a [known bug](https://github.com/grpc-ecosystem/grpc-gateway/pull/489) with the grpc-gateway when using [PopulateFieldFromPath](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/rpc.pb.gw.go#L223).

Fixing grpc-gateway seems ideal, but it seems that there is already a duplicate [string](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/rpc.proto#L529) field that could maybe be used instead - though I'm not sure how.

wilmerpaulino:
[...]

PR #3650, opened 2019-10-30
proto+lnrpc: update gRPC-REST gateway to v1.8.6

### LINKED ISSUE #3168: Wallet recovery with passphrase is not working
initwallet REST API is working without passphrase but it fails with passphrase, with 'invalid passphrase' error message.

I opened first issue about this API #3128 (https://github.com/lightningnetwork/lnd/issues/3128) almost 10 days back also and i am testing these scenarios from almost last 15 days. I found that initwallet REST API is behaving strangly.

If I generate the genseed with passphrase through lncli and use this cipher_seed_mnemonic array to recover the wallet through initwallet REST API, then it works perfectly. But, if i will generate the genseed with passphrase through REST and use that cipher_seed_mnemonic array to recover with initwallet REST then i am getting 'invalid passphrase' error.

Your environment
version of lnd: 0.6.0-beta
which operating system: Windows
version bitcoind: v0.17.1

Steps to reproduce
1) Call GET /v1/genseed with aezeed_passphrase value.
2) Call POST /v1/initwallet with wallet_password. it will initialize the wallet successfully.
3) Delete the wallet manually. 
4) Call POST /v1/initwallet with wallet_password, aezeed_passphrase and cipher_seed_mnemonic array generated from step 1.

Initialize wallet with below details:
Method: POST,
url: https://192.168.1.8:8080/v1/initwallet,
payload: {
"wallet_password":"c3VoZWIxMjM=",
"cipher_seed_mnemonic":["absent","stuff","layer","holiday","slam","pelican","drift","heart","prepare","casual","treat","liar","small","venue","fragile","wise","frown","safe","brief","cherry","pigeon","pole","sound","rocket"],
"aezeed_passphrase":"dGhpc2lzbXlwYXNzcGhyYXNl"
}

Expected behaviour
It should successfully recover the wallet.

Actual behaviour
Failed response with { "error": "invalid passphrase", "code": 2 }

I have tested initwallet with multiple scenarios and all the scenarios are working with GRPC but not with REST.

guggero:
Fixes #3168

When using the REST gateway, fields that have the type `bytes` in `rpc.proto` cannot be used in query or path parameters (see #552 which lead to workarounds like #592).

The root cause of this is that the version of `grpc-gateway` we use doesn't support mapping to `[]byte` yet.
Trying to use `bytes` parameters results in `{"error":"unsupported field type uint8","code":3}`.
This has been fixed in https://github.com/grpc-ecosystem/grpc-gateway/pull/489.

Unfortunately, we can't use the newest version of `grpc-gateway` (which would be `1.11.3` at the moment) because that has a bug that messes up our `rpc.swagger.json` (https://github.com/grpc-ecosystem/grpc-gateway/issues/746).

After some trial and error, I found that version `1.8.6` is the newest version that works for our use case and produces the smallest diff in our generated code.

### DIFF
--- a/lnrpc/README.md
+++ b/lnrpc/README.md
@@ -170,11 +170,11 @@ $ cd $GOPATH/src/google.golang.org/genproto
 $ git reset --hard a8101f21cf983e773d0c1133ebc5424792003214
 ```
 
-4. Install `grpc-ecosystem/grpc-gateway` at commit `f2862b476edcef83412c7af8687c9cd8e4097c0f`.
+4. Install `grpc-ecosystem/grpc-gateway` at version `v1.8.6`.
 ```bash
 $ git clone https://github.com/grpc-ecosystem/grpc-gateway $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
 $ cd $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
-$ git reset --hard f2862b476edcef83412c7af8687c9cd8e4097c0f
+$ git reset --hard v1.8.6
 $ go install ./protoc-gen-grpc-gateway ./protoc-gen-swagger
 ```
 
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/golang/protobuf v1.3.1
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
-	github.com/grpc-ecosystem/grpc-gateway v0.0.0-20170724004829-f2862b476edc
+	github.com/grpc-ecosystem/grpc-gateway v1.8.6
 	github.com/jackpal/gateway v1.0.5
 	github.com/jackpal/go-nat-pmp v0.0.0-20170405195558-28a68d0c24ad
 	github.com/jessevdk/go-flags v1.4.0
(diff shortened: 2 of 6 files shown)


lnrpc/README.md
```diff
@@ -170,11 +170,12 @@ $ cd $GOPATH/src/google.golang.org/genproto
...
 ```
 
-4. Install `grpc-ecosystem/grpc-gateway` at commit `f2862b476edcef83412c7af8687c9cd8e4097c0f`.
+4. Install `grpc-ecosystem/grpc-gateway` at version `v1.8.6`.
 ```bash
 $ git clone https://github.com/grpc-ecosystem/grpc-gateway $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
 $ cd $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
-$ git reset --hard f2862b476edcef83412c7af8687c9cd8e4097c0f
+$ git reset --hard v1.8.6
+$ go mod init
```
wpaulino: Why is this required? This version seems to support go modules.

guggero: It's not required. That was left over from a trial with an older version. Fixed.

joostjager:
it would be nice if we can also mark all the string fields as deprecated now and remove them in 0.10

wpaulino:
> it would be nice if we can also mark all the string fields as deprecated now and remove them in 0.10

@joostjager did you mean to leave this comment on https://github.com/lightningnetwork/lnd/pull/3647 instead?

guggero:
@wpaulino no, he meant the fields introduced in #592 that worked around the bug that this PR fixes.

halseth:
> > it would be nice if we can also mark all the string fields as deprecated now and remove them in 0.10
> 
> Do we want to do this? I think it would be nice to be able to remove the string fields at some point

Yep, can be done in separate PR though.

joostjager:
I am getting a new error:

```
joost@joost-desktop $ make rpc
 Compiling protos.
cd ./lnrpc; ./gen_protos.sh
Generating root gRPC server protos
F1102 13:10:55.891383    8655 main.go:73] Cannot set flag paths=source_relative
--grpc-gateway_out: protoc-gen-grpc-gateway: Plugin failed with status code 255.
```

joostjager:
Nvm, needed to update dep

PR #3654, opened 2019-10-31
multi: convert the existing channeldb.ChannelType uint8 into a bit field

Roasbeef:
In this commit, we convert the existing `channeldb.ChannelType` type
into a _bit field_. This doesn't require us to change the current
serialization or interpretation or the type as it is, since all the
current defined values us a distinct bit. This PR lays the ground work
for any future changes that may introduce new channel types (like anchor
outputs), and also any changes that may modify the existing invariants
around channels (if we're the initiator, we always have the funding
transaction).


channeldb/channel.go
```diff
@@ -125,39 +125,46 @@ var (
 // ChannelType is an enum-like type that describes one of several possible
 // channel types. Each open channel is associated with a particular type as the
 // channel type may determine how higher level operations are conducted such as
-// fee negotiation, channel closing, the format of HTLCs, etc.
-// TODO(roasbeef): split up per-chain?
+// fee negotiation, channel closing, the format of HTLCs, etc. Structure-wise,
+// a ChannelType is a bit field, with each bit denoting a modification from the
+// base channel type of single funder.
```
cfromknecht: this last bit isn't quite accurate since dual and single funder each have their own bit

channeldb/channel.go
```diff
@@ -125,39 +125,46 @@ var (
...
 type ChannelType uint8
 
 const (
 	// NOTE: iota isn't used here for this enum needs to be stable
 	// long-term as it will be persisted to the database.
 
-	// SingleFunder represents a channel wherein one party solely funds the
-	// entire capacity of the channel.
-	SingleFunder ChannelType = 0
+	// SingleFunderBit represents a channel wherein one party solely funds
```
cfromknecht: would personally refer to these as `Flags` rather than `Bits`, but not a blocker

channeldb/channel.go
```diff
@@ -125,39 +125,46 @@ var (
...
 
 // IsTweakless returns true if the target channel uses a commitment that
 // doesn't tweak the key for the remote party.
 func (c ChannelType) IsTweakless() bool {
-	return c == SingleFunderTweakless
+	return c&SingleFunderTweaklessBit == SingleFunderTweaklessBit
+}
+
+// IsDualFunder returns true if the ChannelType has the DualFunderBit set.
+func (c ChannelType) IsDualFunder() bool {
```
cfromknecht: order helper methods in the order the flags are defined?

guggero:
Travis fails for the dual funder flow:

```
--- FAIL: TestLightningWallet (356.83s)
    --- PASS: TestLightningWallet/btcwallet/bitcoind:change_output_spend_confirmation (6.29s)
    --- PASS: TestLightningWallet/btcwallet/bitcoind:insane_fee_reject (0.01s)
    --- PASS: TestLightningWallet/btcwallet/bitcoind:single_funding_workflow (0.72s)
    --- PASS: TestLightningWallet/btcwallet/bitcoind:single_funding_workflow_tweakless (2.46s)
    --- FAIL: TestLightningWallet/btcwallet/bitcoind:dual_funder_workflow (0.33s)
        interface_test.go:489: unable to consume alice's sigs: counterparty's commitment signature is invalid
```

channeldb/channel.go
```diff
@@ -125,39 +125,46 @@ var (
...
 const (
 	// NOTE: iota isn't used here for this enum needs to be stable
 	// long-term as it will be persisted to the database.
 
-	// SingleFunder represents a channel wherein one party solely funds the
-	// entire capacity of the channel.
-	SingleFunder ChannelType = 0
+	// SingleFunderBit represents a channel wherein one party solely funds
+	// the entire capacity of the channel.
+	SingleFunderBit ChannelType = 0
```
cfromknecht: heh, now there is no single funder bit

Roasbeef: Yeah the old one was incorrect, in that it started things off at zero instead of 1, so it wasn't backwards compatible.

Roasbeef:
Fixed the test failure. An earlier version was incorrect in that it re-defined the base channel type (single funder) to one instead of zero. I fixed that, but then introduced another bug that would cause it to not recognize dual funder channels properly. This has been fixed now, by defining `IsSingleFunder` using the dual funder bit.

PR #449, opened 2017-12-03
lnwire: Support upfront shutdown script feature

stevenroose:
I noticed that this was part of the RFC, but not yet implemented here.

PR #3655, opened 2019-10-31
Add Option Upfront Shutdown

carlaKC:
This PR adds support for `option_upfront_shutdown_commitment`, which allows nodes to specify a script to which funds should be paid out in the event of a cooperative close during channel negotiation. 

Upfront shutdown addresses are set by default for channels that we open or accept if the remote peer supports this feature. 

Replaces https://github.com/lightningnetwork/lnd/pull/449


channeldb/channel.go
```diff
@@ -532,6 +532,16 @@ type OpenChannel struct {
 	// for which we are the initiator.
 	FundingTxn *wire.MsgTx
 
+	// LocalShutdownScript is set to a pre-set script if the channel was opened
+	// by the local node with option_upfront_shutdown_script set. If the option
```
halseth: this change should be moved to the commit where they are stored in the database :)

fundingmanager.go
```diff
@@ -1212,18 +1221,33 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	)
 	tweaklessCommitment := localTweakless && remoteTweakless
 	chainHash := chainhash.Hash(msg.ChainHash)
+
+	// If the peer sent an upfront shutdown script, but we do not support the
+	// feature, log a warning. Enforcement of this script occurs entirely on
+	// our side, so we can ignore it.
+	localUpfrontShutdown := fmsg.peer.LocalLocalFeatures().HasFeature(
```
halseth: is this configurable now? If its not I guess we don't have to do this check. 

It should probably be configurable though :)

discovery/mock_test.go
```diff
@@ -56,12 +56,22 @@ func (p *mockPeer) Address() net.Addr { return nil }
 func (p *mockPeer) QuitSignal() <-chan struct{} {
 	return p.quit
 }
+func (p *mockPeer) GenDeliveryScript() ([]byte, error) {
```
halseth: use the "embed interface" trick to avoid all these empty methods?  😄

peer.go
```diff
@@ -1728,9 +1728,11 @@ func (p *peer) ChannelSnapshots() []*channeldb.ChannelSnapshot {
 	return snapshots
 }
 
-// genDeliveryScript returns a new script to be used to send our funds to in
+// GenDeliveryScript returns a new script to be used to send our funds to in
 // the case of a cooperative channel close negotiation.
-func (p *peer) genDeliveryScript() ([]byte, error) {
+//
+// NOTE: Part of the lnpeer.Peer interface.
+func (p *peer) GenDeliveryScript() ([]byte, error) {
```
halseth: hm, not sure this should be part of the `Peer` interface.

server.go
```diff
@@ -2735,6 +2735,9 @@ func (s *server) peerConnected(conn net.Conn, connReq *connmgr.ConnReq,
 	localFeatures.Set(lnwire.DataLossProtectRequired)
 	localFeatures.Set(lnwire.GossipQueriesOptional)
 
+	// Signal that we understand option upfront shutdown script.
+	localFeatures.Set(lnwire.UpfrontShutdownScript)
```
halseth: should be made optional/configurable?

lnwire/features.go
```diff
@@ -35,6 +35,11 @@ const (
 	// connection is established.
 	InitialRoutingSync FeatureBit = 3
 
+	// UpfrontShutdownScript indicates that the node supports committing upfront
+	// to the shutdown script when opening channels.  This prevents a node that
+	// has been compromised from having all the channel funds stolen.
+	UpfrontShutdownScript FeatureBit = 5
```
halseth: we should have required and a optional bit, similar to the other fields.

lnwire/accept_channel.go
```diff
@@ -86,6 +86,11 @@ type AcceptChannel struct {
 	// base point in order to derive the revocation keys that are placed
 	// within the commitment transaction of the sender.
 	FirstCommitmentPoint *btcec.PublicKey
+
+	// UpfrontShutdownScript is the script to which the channel funds should
+	// be paid when mutually closing the channel.  This is to be enforced by
+	// the remote node even if this node is compromised later.
+	UpfrontShutdownScript DeliveryAddress
```
cfromknecht: reuse the existing `PkScript` type? https://github.com/lightningnetwork/lnd/blob/master/lnwire/lnwire.go#L27

lnwire/shutdown.go
```diff
@@ -71,11 +80,8 @@ func (s *Shutdown) MaxPayloadLength(pver uint32) uint32 {
 	// Len - 2 bytes
 	length += 2
 
-	// ScriptPubKey - 34 bytes for pay to witness script hash
-	length += 34
-
-	// NOTE: pay to pubkey hash is 25 bytes, pay to script hash is 22
-	// bytes, and pay to witness pubkey hash is 22 bytes in length.
+	// ScriptPubKey
+	length += deliveryAddressMaxSize
```
cfromknecht: missing length prefix?

lnwire/features.go
```diff
@@ -89,6 +94,7 @@ var LocalFeatures = map[FeatureBit]string{
 	InitialRoutingSync:      "initial-routing-sync",
 	GossipQueriesRequired:   "gossip-queries",
 	GossipQueriesOptional:   "gossip-queries",
+	UpfrontShutdownScript:   "option-upfront-shutdown-script",
```
cfromknecht: can remove `option-` prefix

lnwire/features.go
cfromknecht: indeed, they need to be introduced in pairs otherwise we can't tell if the sender is _requiring_ this option. we made a similar mistake in adding the gossip queries back in the day

lnwire/accept_channel.go
```diff
@@ -113,6 +118,7 @@ func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 		a.DelayedPaymentPoint,
 		a.HtlcPoint,
 		a.FirstCommitmentPoint,
+		a.UpfrontShutdownScript,
```
cfromknecht: add conditional logic to encode here, rather than in the codec?

channeldb/channel.go
```diff
@@ -2680,6 +2721,39 @@ func fetchChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
...
+	addr *lnwire.DeliveryAddress) error {
+
+	// Return early if the bucket does not exit, a shutdown script was not set.
+	bs := chanBucket.Get(key)
+	if bs == nil {
+		return nil
+	}
+
+	r := bytes.NewReader(bs)
+	if err := ReadElements(r, addr); err != nil {
```
cfromknecht: error value can be returned directly

channeldb/channel.go
```diff
@@ -2680,6 +2721,39 @@ func fetchChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
...
 	channel.Packager = NewChannelPackager(channel.ShortChannelID)
 
+	// Finally, read the optional shutdown scripts.
+	if err := readOptionalShutdownScript(
+		chanBucket, localUpfrontShutdownKey, &channel.LocalShutdownScript,
+	); err != nil {
+		return err
+	}
+
+	if err := readOptionalShutdownScript(
```
cfromknecht: same here re: returning error directly

fundingmanager.go
```diff
@@ -1212,18 +1221,33 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
...
+		NodeAddr:              fmsg.peer.Address(),
+		LocalFundingAmt:       0,
+		RemoteFundingAmt:      amt,
+		CommitFeePerKw:        lnwallet.SatPerKWeight(msg.FeePerKiloWeight),
+		FundingFeePerKw:       0,
+		PushMSat:              msg.PushAmount,
+		Flags:                 msg.ChannelFlags,
+		MinConfs:              1,
+		Tweakless:             tweaklessCommitment,
+		LocalUpfrontShutdown:  shutdown,
```
cfromknecht: how do we handle the case where these are nil?

lnwire/accept_channel.go
cfromknecht: scratch this, i thought `DeliveryAddress` was added in this pr

discovery/mock_test.go
carlaKC: Yeah will do! Just wanted to check in lnd-dev before I started using embedded interfaces.

fundingmanager.go
carlaKC: It should be, I'll add a `enable_upfront_script` option?
I was thinking that it would be cool if people could specify an upfront shutdown addr when they call `openChannel` from lncli/rpc but that requires knowledge of peer feature bits etc so I think I'll leave it simple for now.

fundingmanager.go
carlaKC: A nil `DeliveryAddress` is just an empty byte array, and there are checks for `len!=0` when we do use it.

lnwire/accept_channel.go
carlaKC: I'll separate it out as in `Decode`, but if by codec you mean `channeldb/codec` (I'm not sure what codec is :/) then won't we still need conditional logic because `WriteElement` isn't being passed a wire message? It will be passed another struct which has the field with a nil value.

lnwire/features.go
carlaKC: I'll add the even bit as well. 

In terms of mandatory bits, if a peer has the mandatory upfront shutdown bit set it just means that we have to support that feature for them, right? Not that we **have to** set a shutdown addr ourselves.

lnwire/shutdown.go
carlaKC: Isn't the +2 on [line 81](https://github.com/lightningnetwork/lnd/pull/3655/files#diff-b928d202d95ccbba631a8cd4fe621844R81) the length prefix?

lnwire/accept_channel.go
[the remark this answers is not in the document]
carlaKC: By optional, do you mean anything other than the way we encode/decode it?
Re encoding, I'm going to address conner's comment below.

lnwire/lnwire.go
```diff
@@ -401,6 +401,11 @@ func WriteElement(w io.Writer, element interface{}) error {
 		}
 
 	case DeliveryAddress:
+		// If there is no delivery address, do not write anything.
```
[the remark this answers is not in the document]
carlaKC: I don't think so? That would make messages without the optional delivery address fields fail, wouldn't it?

peer.go
carlaKC: yeah, it's a bit of a weird one. I'll see if I can thread it down another way :)

channeldb/channel.go
carlaKC: Should I update  `LightningChannel.Local(/Remote)UpfrontShutdownScript` to just return empty addresses for this commit? Because I do like being able to spit this code from the DB code.

lnwire/features.go
halseth: yep, I think so. If they require it, they want to know that we can support honouring their wish.

carlaKC:
Pushed some small fixes + the ability to enable upfront shutdown a with `enable_upfront_shutdown` flag.

lnwire/features.go
carlaKC: Ok cool, that makes sense :)

lnwire/lnwire.go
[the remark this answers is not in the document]
halseth: since we don't attempt to write it in case it is zero-length, it should be okay to return an error here.

However see my other comment about zero-length address.

lnwire/open_channel.go
```diff
@@ -122,6 +122,11 @@ type OpenChannel struct {
 	// Currently, the least significant bit of this bit field indicates the
 	// initiator of the channel wishes to advertise this channel publicly.
 	ChannelFlags FundingFlag
+
+	// UpfrontShutdownScriptOptional is the script to which the channel funds should
```
halseth: godoc doesn't match field na,e

halseth: godoc doesn't match field name

lnwire/accept_channel.go
```diff
@@ -113,7 +118,15 @@ func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 		a.DelayedPaymentPoint,
 		a.HtlcPoint,
 		a.FirstCommitmentPoint,
-	)
+	); err != nil {
+		return err
+	}
+
+	if len(a.UpfrontShutdownScript) == 0 {
```
halseth: The spec says 
```
if both nodes advertised the option_upfront_shutdown_script feature:
MUST include either a valid shutdown_scriptpubkey as required by shutdown scriptpubkey, or a zero-length shutdown_scriptpubkey.`
```

I read this as we must write a zero length field in this case?

lnwire/open_channel.go
```diff
@@ -153,6 +158,7 @@ func (o *OpenChannel) Encode(w io.Writer, pver uint32) error {
 		o.HtlcPoint,
 		o.FirstCommitmentPoint,
 		o.ChannelFlags,
+		o.UpfrontShutdownScript,
```
halseth: why doesn't this match what we do in `AcceptChannel`?

server.go
```diff
@@ -2736,6 +2743,11 @@ func (s *server) peerConnected(conn net.Conn, connReq *connmgr.ConnReq,
 	localFeatures.Set(lnwire.DataLossProtectRequired)
 	localFeatures.Set(lnwire.GossipQueriesOptional)
 
+	// Add the set of local features which have been enabled by the user.
+	for _, f := range s.localFeatures {
```
halseth: can we just check the `config` here instead, and set the feature if enabled?

fundingmanager.go
```diff
@@ -1388,6 +1418,15 @@ func (f *fundingManager) handleFundingAccept(fmsg *fundingAcceptMsg) {
 		return
 	}
 
+	// Set the remote shutdown script on the reservation. If the upfront shutdown
+	// script is empty, this call will have no effect.
+	if err := resCtx.reservation.ProcessUpfrontShutdown(
```
halseth: name `SetUpfrontShutdown` instead? TO indicate we are not processing a new remote message.

fundingmanager.go
```diff
@@ -1190,6 +1191,19 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 		return
 	}
 
+	shutdown, err := getUpfrontShutdownScript(fmsg.peer,
```
halseth: Could we do this on a per-channel basis? Meaning that we accept a upfront address when we open a channel. That way we can actually expose it on the RPC (in a later PR).

carlaKC: The problem with requiring user input for upfront shutdown is that they aren't around to signal whether we should use upfront shutdown addresses in our `acceptChannel` messages when peers initiate channel opens.

Re getting addresses from users, I was thinking that we have this flag (which will enable for accepts and opens) and then allow users to specify an upfront address via lncli/rpc when opening channels. This would provide the ability to set a specific address when `enable_upfront_shutdown` is true, and to signal upfront shutdown for the specific channel that they're opening if `enable_upfront_shutdown` is false.

This does require some knowledge of peer feature bits, but I thinks something of the sort may be in the pipeline?

lnwallet/reservation.go
```diff
@@ -131,8 +131,8 @@ type ChannelReservation struct {
 func NewChannelReservation(capacity, localFundingAmt btcutil.Amount,
 	commitFeePerKw chainfee.SatPerKWeight, wallet *LightningWallet,
 	id uint64, pushMSat lnwire.MilliSatoshi, chainHash *chainhash.Hash,
-	flags lnwire.FundingFlag,
-	tweaklessCommit bool) (*ChannelReservation, error) {
+	flags lnwire.FundingFlag, tweaklessCommit bool, localShutdown,
+	remoteShutdown lnwire.DeliveryAddress) (*ChannelReservation, error) {
```
Roasbeef: Following from my other comment, I think this can be handled as a contribution.

lnwallet/wallet.go
```diff
@@ -960,6 +982,26 @@ func (l *LightningWallet) handleSingleContribution(req *addSingleContributionMsg
 	return
 }
 
+// handleUpfrontShutdown sets the remote party's upfront shutdown script.
+func (l *LightningWallet) handleUpfrontShutdown(req *addUpfrontShutdown) {
```
Roasbeef: See comment about moving this to a contribution level datum.

lnwire/accept_channel.go
carlaKC: Checked this out and you're right about the 0 value 👍 

So proceeding with the following:
- Encode a 0 for empty shutdown address (fine for non upfront shutdown understanding lnd, will have to double check interop)
- Encode `len`/`addr` when given a non-0 delivery address
- Always try to decode with the EOF trick

lnwire/lnwire.go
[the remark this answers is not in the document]
carlaKC: Updated to write a zero for empty addresses, as per comment above

discovery/mock_test.go
```diff
@@ -15,6 +15,11 @@ import (
 // mockPeer implements the lnpeer.Peer interface and is used to test the
 // gossiper's interaction with peers.
 type mockPeer struct {
+	// embed the peer interface so that mockPeer will implement lnpeer.Peer by
+	// default. Methods that need to be called in tests still need to be
+	// implemented.
+	lnpeer.Peer
```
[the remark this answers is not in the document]
carlaKC: My argument for using this is that it results in tests failing loudly. If you extend an interface and a test uses the new function call it will panic, which you'll pick up and have to go see exactly what your change is affecting. If the tests need the new function, you will have to update the mock in your PR (as before), this just saves on some boilerplate :)

config.go
```diff
@@ -675,6 +679,12 @@ func loadConfig() (*config, error) {
 			"listening is disabled")
 	}
 
+	if cfg.EnableUpfrontShutdown {
```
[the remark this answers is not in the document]
carlaKC: Going to go with the following:
- Always signal that we support the feature for peers
- Flag which enables using fresh wallet addresses by default
- (follow up PR): Ability to specify your own upfront addr for openChannel calls (going to need to expose peer feature bits to users etc so I'm going keep this separate)

fundingmanager_test.go
```diff
@@ -2948,3 +2959,97 @@ func TestFundingManagerFundAll(t *testing.T) {
...
 	}
 }
+
+type mockPeer struct {
+	lnpeer.Peer
+}
+
+func (p *mockPeer) RemoteLocalFeatures() *lnwire.FeatureVector {
+	return lnwire.NewFeatureVector(&lnwire.RawFeatureVector{}, nil)
+}
```
carlaKC: Rebase mishap, this will be removed.

lnwire/accept_channel.go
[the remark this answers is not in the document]
halseth: I was thinking as a pointer that can be nil, but since this is just a byte slice that is already the case 👍

.golangci.yml
```diff
@@ -64,3 +64,9 @@ linters:
 issues:
   # Only show newly introduced problems.
   new-from-rev: 01f696afce2f9c0d4ed854edefa3846891d01d8a
+
+  exclude-rules:
+    # Exclude gosec from running for tests.
```
halseth: Can you add to the commen what `gosec` is checking?

lnwire/features.go
```diff
@@ -84,11 +94,13 @@ const (
 // not advertised to the entire network. A full description of these feature
 // bits is provided in the BOLT-09 specification.
 var LocalFeatures = map[FeatureBit]string{
-	DataLossProtectRequired: "data-loss-protect",
-	DataLossProtectOptional: "data-loss-protect",
-	InitialRoutingSync:      "initial-routing-sync",
-	GossipQueriesRequired:   "gossip-queries",
-	GossipQueriesOptional:   "gossip-queries",
+	DataLossProtectRequired:       "data-loss-protect",
```
halseth: must be rebased on master, this map changed slightly in #3685

discovery/mock_test.go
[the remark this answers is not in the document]
halseth: I like this boilerplate-free world :)

htlcswitch/link_test.go
```diff
@@ -1610,6 +1610,11 @@ func TestChannelLinkSingleHopMessageOrdering(t *testing.T) {
 }
 
 type mockPeer struct {
+	// embed the peer interface so that mockPeer will implement lnpeer.Peer by
```
halseth: nit: probably don't need this comment everywhere

peer.go
```diff
@@ -2437,6 +2437,24 @@ func (p *peer) RemoteGlobalFeatures() *lnwire.FeatureVector {
 	return p.remoteGlobalFeatures
 }
 
+// RemoteGlobalFeatures returns the set of local features that has been
+// advertised by the remote node. This allows sub-systems that use this
+// interface to gate their behavior off the set of negotiated feature bits.
+//
+// NOTE: Part of the lnpeer.Peer interface.
+func (p *peer) RemoteLocalFeatures() *lnwire.FeatureVector {
+	return p.remoteLocalFeatures
```
halseth: will also sligthly change with rebase

server.go
```diff
@@ -2737,6 +2737,9 @@ func (s *server) peerConnected(conn net.Conn, connReq *connmgr.ConnReq,
 	localFeatures.Set(lnwire.DataLossProtectRequired)
 	localFeatures.Set(lnwire.GossipQueriesOptional)
 
+	// Signal that we support peers providing an upfront shutdown address.
+	localFeatures.Set(lnwire.UpfrontShutdownScriptOptional)
```
halseth: 👍

config.go
```diff
@@ -321,6 +321,8 @@ type config struct {
 
 	MaxChannelFeeAllocation float64 `long:"max-channel-fee-allocation" description:"The maximum percentage of total funds that can be allocated to a channel's commitment fee. This only applies for the initiator of the channel. Valid values are within [0.1, 1]."`
 
+	EnableUpfrontShutdown bool `long:"enable-upfront-shutdown" description:"If true, option upfront shutdown script will be enabled. If peers that we open channels with support this feature, we will set the script to which cooperative closes should be paid out to upon channel opening, and allow peers to do the same. This offers the partial protection of a channel peer diconnecting if cooperative close is attempted with a different script, but does not protect against force closes."`
```
halseth: not used?

peer.go
```diff
@@ -2037,6 +2037,20 @@ func (p *peer) reenableActiveChannels() {
 	}
 }
 
+// getLocalDeliveryScript returns an script to close the channel out to. If
+// an upfront shutdown script was set, it is returned. Otherwise, a random
```
halseth: random is probably not the right wording here... Perhaps "a new delivery script is returned"?

fundingmanager.go
```diff
@@ -1259,10 +1261,34 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 		return
 	}
 
+	// Check whether the peer supports upfront shutdown, and get a new wallet
+	// address if our node is configured to set shutdown addresses by default.
+	// A nil address is set in place of user input, because this channel open
+	// was not initiated by the user.
+	shutdown, err := getUpfrontShutdown(
+		fmsg.peer,
+		nil,
```
halseth: nit: can fit on prev line

channeldb/channel.go
```diff
@@ -538,6 +548,16 @@ type OpenChannel struct {
 	// for which we are the initiator.
 	FundingTxn *wire.MsgTx
 
+	// LocalShutdownScript is set to a pre-set script if the channel was opened/
+	// by the local node with option_upfront_shutdown_script set. If the option/
+	// was not set, the field is empty.
+	LocalShutdownScript lnwire.DeliveryAddress
```
halseth: What do you think about making the methods `LocalShutdownScript()` fetch directly from the database instead of this pre-populated field? Would it have any downsides?

config.go
carlaKC: Called in `getUpfrontShutdown` to check whether we should set new addresses?

channeldb/channel.go
carlaKC: The fields in `OpenChannel` are also used for writing to the DB, and threading the values through the funding flow (as part of chanState). Since we need the fields in `OpenChannel` to write it makes sense to read them in/accessin memory?

To keep them out of channel state completely, they'd need to be in `ChannelReservation` which leaks into `lnwallet`.

peer.go
```diff
@@ -2134,10 +2151,11 @@ func (p *peer) handleLocalCloseReq(req *htlcswitch.ChanClose) {
 	// out this channel on-chain, so we execute the cooperative channel
 	// closure workflow.
 	case htlcswitch.CloseRegular:
-		// First, we'll fetch a fresh delivery address that we'll use
-		// to send the funds to in the case of a successful
-		// negotiation.
-		deliveryAddr, err := p.genDeliveryScript()
+		// First, we'll fetch a  delivery address that we'll use to send the
+		// funds to in the case of a successful negotiation.
+		deliveryAddr, err := p.getLocalDeliveryScript(
```
halseth: Not sure a method is needed here? Maybe it reads better just as 
```go
// First, we'll fetch a  delivery address that we'll use to send the
// funds to in the case of a successful negotiation.
deliveryAddr, err := channel.LocalUpfrontShutdownScript()

// If no upfromnt script was set, get anew address from the wallet.
if err == lnwallet.ErrNoLocalUpfrontShutdownScript:
        deliveryAddr, err = p.genDeliveryScript()
}
if err != nil {
...
```
alternatively match on the zero length script here, and fetch a new one in that case.

channeldb/channel.go
```diff
@@ -538,6 +548,16 @@ type OpenChannel struct {
...
+
+	// RemoteShutdownScript is set to a pre-set script if the channel was opened/
```
halseth: trailing slashes?

channeldb/channel.go
```diff
@@ -2676,7 +2727,32 @@ func fetchChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
...
+
+	return readOptionalShutdownScript(
+		chanBucket, remoteUpfrontShutdownKey, &channel.RemoteShutdownScript,
+	)
+}
+
+// readOptionalShutdownScript reads the shutdown script stored under the key
+// provided if it is present. Upfront shutdown scripts are optional, so the
+// function returns with no error if the key is not present.
+func readOptionalShutdownScript(chanBucket *bbolt.Bucket, key []byte,
```
halseth: naming suggestion: `get....`

Also keep put and get methods close in the file?

channeldb/codec.go
```diff
@@ -213,6 +213,9 @@ func WriteElement(w io.Writer, element interface{}) error {
 			}
 		}
 
+	case lnwire.DeliveryAddress:
```
halseth: I don't think we need to add this here? we can just pass the bytes directly from the put method.

fundingmanager.go
```diff
@@ -2728,6 +2754,48 @@ func (f *fundingManager) initFundingWorkflow(peer lnpeer.Peer, req *openChanReq)
 	}
 }
 
+// getUpfrontShutdown takes a user provided shutdown address and checks that
+// errors if a non-nil address is provided for a peer that does not support the
```
halseth: returns an error?

lnwallet/reservation.go
```diff
@@ -131,8 +135,7 @@ type ChannelReservation struct {
 func NewChannelReservation(capacity, localFundingAmt btcutil.Amount,
 	commitFeePerKw chainfee.SatPerKWeight, wallet *LightningWallet,
 	id uint64, pushMSat lnwire.MilliSatoshi, chainHash *chainhash.Hash,
-	flags lnwire.FundingFlag,
-	tweaklessCommit bool) (*ChannelReservation, error) {
+	flags lnwire.FundingFlag, tweaklessCommit bool) (*ChannelReservation, error) {
```
halseth: nit: unrelated change?

server.go
```diff
@@ -3088,6 +3088,8 @@ type openChanReq struct {
 	// output selected to fund the channel should satisfy.
 	minConfs int32
 
+	shutdownAddress lnwire.DeliveryAddress
```
halseth: add comment indication this might be nil

lnwallet/reservation.go
```diff
@@ -521,6 +524,16 @@ func (r *ChannelReservation) FundingOutpoint() *wire.OutPoint {
...
 }
 
+// SetOurUpfrontShutdown sets the upfront shutdown address on our contribution
+// and partial state in the reservation.
+func (r *ChannelReservation) SetOurUpfrontShutdown(shutdown lnwire.DeliveryAddress) {
+	r.Lock()
+	defer r.Unlock()
+
+	r.ourContribution.UpfrontShutdown = shutdown
+	r.partialState.LocalShutdownScript = shutdown
```
halseth: is this necessary to set here?

fundingmanager.go
```diff
@@ -1259,10 +1260,32 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
...
+		},
+	)
+	if err != nil {
+		f.failFundingFlow(
+			fmsg.peer, fmsg.msg.PendingChannelID,
+			fmt.Errorf("getUpfrontShutdownScript error: %v", err),
+		)
+		return
+	}
+	reservation.SetOurUpfrontShutdown(shutdown)
```
halseth: looks like the shutdown script can be moved to being part of the `InitChannelReservation` call?

lnwire/accept_channel.go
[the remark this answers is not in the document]
carlaKC: 👍 it will write a 0 length, test added.

chancloser.go
```diff
@@ -283,9 +305,21 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
...
 
-		// Next, we'll note the other party's preference for their
-		// delivery address. We'll use this when we craft the closure
-		// transaction.
+		// If the remote node opened the channel with option upfront shutdown
+		// script, check that the address they provided matches.
+		err := checkRemoteDeliveryAddress(
+			c.cfg.channel.RemoteUpfrontShutdownScript(), shutDownMsg.Address,
+		)
+		if err != nil {
```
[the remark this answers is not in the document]
carlaKC: sounds good, also makes testing easier :)
I'm going to leave it unexported for now though?

channeldb/codec.go
carlaKC: We don't need this for put, but do need it to read into a `DeliveryAddress` type on the channel struct.
Is it ok to have `ReadElement` but not `WriteElement` for a type?

Can also get around if for read by reading into a `[]byte` then casting to `DeliveryAddress` but that feels messier?

I'm going to remove for write for now, and can do this ^ workaround for read if it's necessary to have `ReadElement`/`WriteElement` implemented in pairs.

chancloser.go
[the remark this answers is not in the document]
halseth: yep, as long as it is only used here, we don't have to export it.

channeldb/codec.go
halseth: Yeah, we should definitely have pairs. I was thinking just doing the casting.

fundingmanager.go
carlaKC: Previously got some [comments](https://github.com/lightningnetwork/lnd/pull/3655#discussion_r342907070) from @Roasbeef about having this data in contributions, which is why it's not in `InitChannelReservation`. 

Having these addresses in `ChannelReservation` itself requires a method similar to this because we still need to lock the reservation and set the value. I think where we put this (contributions vs reservation itself) may depend on if/how we have plans to decouple funding manager/ wallet, which I don't have the best context for.

lnwallet/reservation.go
carlaKC: We  use the values on `partialState` when we call `SyncPending` in `handleFundingCounterPartySigs` or `handleSingleFunderSigs`, which both use the `ChannelReservation.partialState`. 

I'm going to update to only set the values on `partialState` when we write to disk, but I do think it's a bit strange to have these two out of sync with each other. Fine for now, but future changes are going to have to know that this value is not set on `partialState`.

channeldb/codec.go
carlaKC: Any specific reason we don't want to add delivery address to `Read/WriteElement`?
Casting works fine for writes, but it's not as smooth for reads because our existing pattern is to pass in a pointer to the destination. 

Could have something like `getOptionalUpfrontShutdownScript` return the value (rather than setting the pointer's value in the function) but I don't think that's better than just having delivery addr is codec?

halseth: Just to avoid cluttering the codec with types, as it is really just an alias for `[]byte`

lnwire/lnwire_test.go
```diff
@@ -353,6 +365,17 @@ func TestLightningWireProtocol(t *testing.T) {
 				return
 			}
 
+			// 1/2 chance empty upfront shutdown script
```
carlaKC: nit to self, full stop

feature/default_sets.go
```diff
@@ -29,4 +29,8 @@ var defaultSetDesc = setDesc{
 		SetNodeAnn:      {}, // N
 		SetLegacyGlobal: {},
 	},
```
carlaKC: just spotted that the commit message on this needs updating

chancloser.go
```diff
@@ -283,9 +318,23 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
...
-		// transaction.
+		// If the remote node opened the channel with option upfront shutdown
+		// script, check that the script they provided matches.
+		err := maybeMatchScript(
+			c.cfg.channel.RemoteUpfrontShutdownScript(), shutDownMsg.Address,
+		)
+		if err == errUpfrontShutdownScriptMismatch {
+			// The remote peer has provided a script that does not match the
+			// one set on channel open. Disconnect from the peer, as per BOLT 2.
+			c.cfg.disconnect(err)
```
halseth: less error prone: 
```go
if err == errUpf... {
        disconnect()
        return nil, false, err
} else if err != nil {
      return nil, false, err
}
```
alternatively:
```go
 if err != nil {
        if err == errUpf... {
              disconnect()
        }
        return nil, false, err
}
```

peer.go
```diff
@@ -2068,9 +2068,16 @@ func (p *peer) fetchActiveChanCloser(chanID lnwire.ChannelID) (*channelCloser, e
 				"channel w/ active htlcs")
 		}
 
-		// We'll create a valid closing state machine in order to
-		// respond to the initiated cooperative channel closure.
-		deliveryAddr, err := p.genDeliveryScript()
+		// We'll create a valid closing state machine in order to respond to the
+		// initiated cooperative channel closure. First, we set the delivery
+		// script that our funds will be paid out to. If an upfront shutdown script
+		// was set, we will use it. Otherwise, Otherwise, we get a fresh delivery
```
halseth: otherwise otherwise 😅

peer.go
```diff
@@ -2068,9 +2068,16 @@ func (p *peer) fetchActiveChanCloser(chanID lnwire.ChannelID) (*channelCloser, e
...
+		// script.
+		var err error
+		deliveryScript := channel.LocalUpfrontShutdownScript()
+		if len(deliveryScript) == 0 {
+			deliveryScript, err = p.genDeliveryScript()
```
halseth: looks like err check can be move inside the brackets here

lnwire/accept_channel_test.go
```diff
@@ -0,0 +1,69 @@
...
+				t.Fatalf("cannot write message: %v", err)
+			}
+
+			msg, err := ReadMessage(buf, 0)
+			if err != nil {
+				t.Fatalf("cannot read message: %v", err)
+			}
+
+			decoded := msg.(*AcceptChannel)
+			if !bytes.Equal(decoded.UpfrontShutdownScript, encoded.UpfrontShutdownScript) {
```
cfromknecht: nit: lines look over 80 chars

peer.go
```diff
@@ -2134,10 +2142,15 @@ func (p *peer) handleLocalCloseReq(req *htlcswitch.ChanClose) {
...
-		deliveryAddr, err := p.genDeliveryScript()
+		// First, we'll fetch a  delivery script that we'll use to send the
+		// funds to in the case of a successful negotiation. If an upfront
+		// shutdown script was set, we will use it. Otherwise, we get a fresh
+		// delivery script.
+		var err error
+		deliveryScript := channel.LocalUpfrontShutdownScript()
+		if len(deliveryScript) == 0 {
+			deliveryScript, err = p.genDeliveryScript()
+		}
```
cfromknecht: similar to johan's comment, below err check can be in this bracket

chancloser.go
```diff
@@ -333,7 +382,21 @@ func (c *channelCloser) ProcessCloseMsg(msg lnwire.Message) ([]lnwire.Message, b
 				"instead have %v", spew.Sdump(msg))
 		}
 
-		// Now that we know this is a valid shutdown message, we'll
+		// If the remote node opened the channel with option upfront shutdown
+		// script, check that the script they provided matches.
+		err := maybeMatchScript(
```
cfromknecht: is it possible to pull this into a helper so that isn't duplicated?

chancloser_test.go
```diff
@@ -0,0 +1,68 @@
...
+	addr2 := randDeliveryAddress(t)
+
+	tests := []struct {
+		name           string
+		shutdownScript lnwire.DeliveryAddress
+		upfrontScript  lnwire.DeliveryAddress
+		expectedErr    error
+	}{
+		{
+			name:           "No upfront shutdown set, script ok",
```
cfromknecht: also test that `nil + []byte{}` is okay?

channeldb/channel_test.go
```diff
@@ -345,6 +345,101 @@ func TestOpenChannelPutGetDelete(t *testing.T) {
...
+	}
+
+	tests := []struct {
+		name           string
+		modifyChannel  func(channel *OpenChannel)
+		expectedLocal  lnwire.DeliveryAddress
+		expectedRemote lnwire.DeliveryAddress
+	}{
+		{
+			name:          "No shutdown scripts",
```
cfromknecht: typically we use all lower cases for test names, not a blocker tho

fundingmanager.go
```diff
@@ -2808,6 +2877,28 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		return
 	}
 
+	// Check whether the peer supports upfront shutdown, and get an address which
+	// should be used (either a user specified address or a new address from the
+	// wallet if our node is configured to set shutdown address by default).
+	shutdown, err := getUpfrontShutdownScript(
+		msg.peer,
+		msg.openChanReq.shutdownScript,
+		func() (lnwire.DeliveryAddress, error) {
```
cfromknecht: perhaps extract this closure as a member of the funding manager? can this pass it to both as, e.g., `f.genDeliveryAddr`

carlaKC: Peer already has a `genDeliveryScript`, so this (as is, and if added to funding manager) results in some duplication. An alternative would be to expose `genDeliveryScript` on the peer interface, or move all of the logic in `getUpfrontShutdownScript` into `f.genUpfrontShutdown`?

carlaKC:
Done some testing to make sure these changes work with other impls/ lnd as is.

lnd:
Branch -> Master: open and close channel ok, does not set upfront for peers that don't support it
Branch -> Branch: local and remote upfront preserved through restart, shutdown addresss checked when appropriate

c-lightning:
Supports upfront shutdown
Open channel from lndl with and without upfront shutdown ok.
Open channel from c-lightning with and without upfront shutdown ok.

C-Lightning does disconnect from us on channel close, but this occurs on master as well, so I'm assuming that's because they only support single channels so do cleanup on close. Will double check.

eclair: 
TBD

Just want to clarify the meaning of "MUST fail the connection" in BOLT 2 before requesting review again. Specifically whether we should error out of the close channel interaction or disconnect from the peer entirely.

carlaKC:
Ecalir interop done, they don't support upfront shutdown so I just checked that channel open/close work as before.

Updated disconnection function call to permanently disconnect from the violating peer by removing them from our persistent peers. We do still have a channel open with the peer, and they will most likely just reconnect to us anyway (since we don't have the ability to ban).

fundingmanager.go
halseth: I think it is just to avoid some clutter and indentation here, instead having the method be

```go
func(f *fundingmanager) genDeliveryAddr() (lnwire.DeliveryAddress, error) {
			addr, err := f.cfg.Wallet.NewAddress(lnwallet.WitnessPubKey, false)
			if err != nil {
				return nil, err
			}
			return txscript.PayToAddrScript(addr)
}
```

GlenCooper:
:eyes:

PR #3659, opened 2019-11-01
chanfunding: create new package to abstract over funding workflows

Roasbeef:
In this commit, we create a new package `chanfunding`, that introduces a number of abstractions to our funding workflow. The base implementation of the core new abstraction (`chanfunding.Assembler`), is`chanfunding.WalletAssembler`. Construction of this concrete implementation of the new interfaces involved heavily refactoring much of the code within `wallet.go`. As a result, much of the channel funding code now lives outside of the `lnwallet` package. 

The second instance of a new channel funder introduced in this PR is the `chanfunding.CannedAssembler`. This channel funder is used in the scenario wherein the "shape" of the funding output is agreed upon _outside_ the normal protocol, while the construction of the commitment transaction and so forth are carried out as normal. Instances where this might be the case include variants of channel factory leaf channel construction, and external hardware wallet based channel funding. 

This PR is the 3rd in a series of PRs that revamps the way channel funding works within `lnd` to allow funding transactions to be constructed externally, with the grand goal of a PSBT (safety first!!!) based channel funding workflow.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -147,6 +147,11 @@ const (
 	// type, but it omits the tweak for one's key in the commitment
 	// transaction of the remote party.
 	SingleFunderTweaklessBit ChannelType = 1 << 1
+
+	// NoFundingTxBit denotes if we have the funding transaction locally on
+	// disk. This bit may be on if the funding transaction was crafted by a
+	// wallet external to the primary daemon.
+	NoFundingTxBit ChannelType = 1 << 2
 )
 
 // IsSingleFunder returns true if the channel type if one of the known single
@@ -164,6 +169,12 @@ func (c ChannelType) IsDualFunder() bool {
 // doesn't tweak the key for the remote party.
 func (c ChannelType) IsTweakless() bool {
 	return c&SingleFunderTweaklessBit == SingleFunderTweaklessBit
+}
+
+// HasFundingTx returns true if this channel type is one that has a funding
+// transaction stored locally.
+func (c ChannelType) HasFundingTx() bool {
+	return c&NoFundingTxBit == 0
 }
 
 // ChannelConstraints represents a set of constraints meant to allow a node to
@@ -535,7 +546,9 @@ type OpenChannel struct {
 	// is found to be pending.
 	//
 	// NOTE: This value will only be populated for single-funder channels
-	// for which we are the initiator.
+	// for which we are the initiator, and that we also have the funding
+	// transaction for. One can check this by using the HasFundingTx()
+	// method on the ChanType field.
 	FundingTxn *wire.MsgTx
 
 	// TODO(roasbeef): eww
@@ -2522,6 +2535,16 @@ func writeChanConfig(b io.Writer, c *ChannelConfig) error {
 	)
 }
 
+// fundingTxPresent returns true if expect the funding transcation to be found
+// on disk or already populated within the passed oen chanel struct.
+func fundingTxPresent(channel *OpenChannel) bool {
+	chanType := channel.ChanType
+
+	return chanType.IsSingleFunder() && chanType.HasFundingTx() &&
+		channel.IsInitiator &&
+		!channel.hasChanStatus(ChanStatusRestored)
+}
+
 func putChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
 	var w bytes.Buffer
 	if err := WriteElements(&w,
@@ -2535,10 +2558,9 @@ func putChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
 		return err
 	}
 
-	// For single funder channels that we initiated, write the funding txn.
-	if channel.ChanType.IsSingleFunder() && channel.IsInitiator &&
-		!channel.hasChanStatus(ChanStatusRestored) {
-
+	// For single funder channels that we initiated, and we have the
+	// funding transaction, then write the funding txn.
+	if fundingTxPresent(channel) {
 		if err := WriteElement(&w, channel.FundingTxn); err != nil {
 			return err
 		}
@@ -2657,10 +2679,9 @@ func fetchChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
 		return err
 	}
 
-	// For single funder channels that we initiated, read the funding txn.
-	if channel.ChanType.IsSingleFunder() && channel.IsInitiator &&
-		!channel.hasChanStatus(ChanStatusRestored) {
-
+	// For single funder channels that we initiated and have the funding
+	// transaction to, read the funding txn.
+	if fundingTxPresent(channel) {
 		if err := ReadElement(r, &channel.FundingTxn); err != nil {
 			return err
 		}
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -521,8 +521,9 @@ func (f *fundingManager) start() error {
 
 			// Rebroadcast the funding transaction for any pending
 			// channel that we initiated. No error will be returned
-			// if the transaction already has been broadcasted.
-			if channel.ChanType.IsSingleFunder() &&
+			// if the transaction already has been broadcast.
+			chanType := channel.ChanType
+			if chanType.IsSingleFunder() && chanType.HasFundingTx() &&
 				channel.IsInitiator {
 
 				err := f.cfg.PublishTransaction(
@@ -1739,21 +1740,28 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 	// delete it from our set of active reservations.
 	f.deleteReservationCtx(peerKey, pendingChanID)
 
-	// Broadcast the finalized funding transaction to the network.
-	fundingTx := completeChan.FundingTxn
-	fndgLog.Infof("Broadcasting funding tx for ChannelPoint(%v): %v",
-		completeChan.FundingOutpoint, spew.Sdump(fundingTx))
+	// Broadcast the finalized funding transaction to the network, but only
+	// if we actually have the funding transaction.
+	if completeChan.ChanType.HasFundingTx() {
+		fundingTx := completeChan.FundingTxn
 
-	err = f.cfg.PublishTransaction(fundingTx)
-	if err != nil {
-		fndgLog.Errorf("Unable to broadcast funding tx for "+
-			"ChannelPoint(%v): %v", completeChan.FundingOutpoint,
-			err)
-		// We failed to broadcast the funding transaction, but watch
-		// the channel regardless, in case the transaction made it to
-		// the network. We will retry broadcast at startup.
-		// TODO(halseth): retry more often? Handle with CPFP? Just
-		// delete from the DB?
+		fndgLog.Infof("Broadcasting funding tx for ChannelPoint(%v): %v",
+			completeChan.FundingOutpoint, spew.Sdump(fundingTx))
+
+		err = f.cfg.PublishTransaction(fundingTx)
+		if err != nil {
+			fndgLog.Errorf("Unable to broadcast funding tx for "+
+				"ChannelPoint(%v): %v",
+				completeChan.FundingOutpoint, err)
+
+			// We failed to broadcast the funding transaction, but
+			// watch the channel regardless, in case the
+			// transaction made it to the network. We will retry
+			// broadcast at startup.
+			//
+			// TODO(halseth): retry more often? Handle with CPFP?
+			// Just delete from the DB?
+		}
 	}
 
 	// Now that we have a finalized reservation for this funding flow,
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -1216,6 +1216,7 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	chainHash := chainhash.Hash(msg.ChainHash)
 	req := &lnwallet.InitFundingReserveMsg{
 		ChainHash:        &chainHash,
+		PendingChanID:    msg.PendingChannelID,
 		NodeID:           fmsg.peer.IdentityKey(),
 		NodeAddr:         fmsg.peer.Address(),
 		LocalFundingAmt:  0,
@@ -2781,6 +2782,10 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		channelFlags = lnwire.FFAnnounceChannel
 	}
 
+	// Obtain a new pending channel ID which is used to track this
+	// reservation throughout its lifetime.
+	chanID := f.nextPendingChanID()
+
 	// Initialize a funding reservation with the local wallet. If the
 	// wallet doesn't have enough funds to commit to this channel, then the
 	// request will fail, and be aborted.
@@ -2798,6 +2803,7 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	tweaklessCommitment := localTweakless && remoteTweakless
 	req := &lnwallet.InitFundingReserveMsg{
 		ChainHash:        &msg.chainHash,
+		PendingChanID:    chanID,
 		NodeID:           peerKey,
 		NodeAddr:         msg.peer.Address(),
 		SubtractFees:     msg.subtractFees,
@@ -2823,11 +2829,7 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	// SubtractFees=true.
 	capacity := reservation.Capacity()
 
-	// Obtain a new pending channel ID which is used to track this
-	// reservation throughout its lifetime.
-	chanID := f.nextPendingChanID()
-
-	fndgLog.Infof("Target commit tx sat/kw for pending_id(%x): %v", chanID,
+	fndgLog.Infof("Target commit tx sat/kw for pendingID(%x): %v", chanID,
 		int64(commitFeePerKw))
 
 	// If the remote CSV delay was not set in the open channel request,
--- a/lnwallet/chanfunding/assembler.go
+++ b/lnwallet/chanfunding/assembler.go
@@ -1,0 +1,137 @@
+package chanfunding
+
+import (
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// CoinSource is an interface that allows a caller to access a source of UTXOs
+// to use when attempting to fund a new channel.
+type CoinSource interface {
+	// ListCoins returns all UTXOs from the source that have between
+	// minConfs and maxConfs number of confirmations.
+	ListCoins(minConfs, maxConfs int32) ([]Coin, error)
+
+	// CoinFromOutPoint attempts to locate details pertaining to a coin
+	// based on its outpoint. If the coin isn't under the control of the
+	// backing CoinSource, then an error should be returned.
+	CoinFromOutPoint(wire.OutPoint) (*Coin, error)
+}
+
+// CoinSelectionLocker is an interface that allows the caller to perform an
+// operation, which is synchronized with all coin selection attempts. This can
+// be used when an operation requires that all coin selection operations cease
+// forward progress. Think of this as an exclusive lock on coin selection
+// operations.
+type CoinSelectionLocker interface {
+	// WithCoinSelectLock will execute the passed function closure in a
+	// synchronized manner preventing any coin selection operations from
+	// proceeding while the closure if executing. This can be seen as the
+	// ability to execute a function closure under an exclusive coin
+	// selection lock.
+	WithCoinSelectLock(func() error) error
+}
+
+// OutpointLocker allows a caller to lock/unlock an outpoint. When locked, the
+// outpoints shouldn't be used for any sort of channel funding of coin
+// selection. Locked outpoints are not expected to be persisted between
+// restarts.
+type OutpointLocker interface {
+	// LockOutpoint locks a target outpoint, rendering it unusable for coin
+	// selection.
+	LockOutpoint(o wire.OutPoint)
+
+	// UnlockOutpoint unlocks a target outpoint, allowing it to be used for
+	// coin selection once again.
+	UnlockOutpoint(o wire.OutPoint)
+}
+
+// Request is a new request for funding a channel. The items in the struct
+// governs how the final channel point will be provisioned by the target
+// Assembler.
+type Request struct {
+	// LocalAmt is the amount of coins we're placing into the funding
+	// output.
+	LocalAmt btcutil.Amount
+
+	// RemoteAmt is the amount of coins the remote party is contributing to
+	// the funding output.
+	RemoteAmt btcutil.Amount
+
+	// MinConfs controls how many confirmations a coin need to be eligible
+	// to be used as an input to the funding transaction. If this value is
+	// set to zero, then zero conf outputs may be spent.
+	MinConfs int32
+
+	// SubtractFees should be set if we intend to spend exactly LocalAmt
+	// when opening the channel, subtracting the fees from the funding
+	// output. This can be used for instance to use all our remaining funds
+	// to open the channel, since it will take fees into
+	// account.
+	SubtractFees bool
+
+	// FeeRate is the fee rate in sat/kw that the funding transaction
+	// should carry.
+	FeeRate chainfee.SatPerKWeight
+
+	// ChangeAddr is a closure that will provide the Assembler with a
+	// change address for the funding transaction if needed.
+	ChangeAddr func() (btcutil.Address, error)
+}
+
+// Intent is returned by an Assembler and represents the base functionality the
+// caller needs to proceed with channel funding on a higher level. If the
+// Cancel method is called, then all resources assembled to fund the channel
+// will be released back to the eligible pool.
+type Intent interface {
+	// FundingOutput returns the witness script, and the output that
+	// creates the funding output.
+	FundingOutput() ([]byte, *wire.TxOut, error)
+
+	// ChanPoint returns the final outpoint that will create the funding
+	// output described above.
+	ChanPoint() (*wire.OutPoint, error)
+
+	// RemoteFundingAmt is the amount the remote party put into the
+	// channel.
+	RemoteFundingAmt() btcutil.Amount
+
+	// LocalFundingAmt is the amount we put into the channel. This may
+	// differ from the local amount requested, as depending on coin
+	// selection, we may bleed from of that LocalAmt into fees to minimize
+	// change.
+	LocalFundingAmt() btcutil.Amount
+
+	// Cancel allows the caller to cancel a funding Intent at any time.
+	// This will return any resources such as coins back to the eligible
+	// pool to be used in order channel fundings.
+	Cancel()
+}
+
+// Assembler is an abstract object that is capable of assembling everything
+// needed to create a new funding output. As an example, this assembler may be
+// our core backing wallet, an interactive PSBT based assembler, an assembler
+// than can aggregate multiple intents into a single funding transaction, or an
+// external protocol that creates a funding output out-of-band such as channel
+// factories.
+type Assembler interface {
+	// ProvisionChannel returns a populated Intent that can be used to
+	// further the channel funding workflow. Depending on the
+	// implementation of Assembler, additional state machine (Intent)
+	// actions may be required before the FundingOutput and ChanPoint are
+	// made available to the caller.
+	ProvisionChannel(*Request) (Intent, error)
+}
+
+// FundingTxAssembler is a super-set of the regular Assembler interface that's
+// also able to provide a fully populated funding transaction via the intents
+// that it produuces.
+type FundingTxAssembler interface {
+	Assembler
+
+	// FundingTxAvailable is an empty method that an assembler can
+	// implement to signal to callers that its able to provide the funding
+	// transaction for the channel via the intent it returns.
+	FundingTxAvailable()
+}
--- a/lnwallet/chanfunding/canned_assembler.go
+++ b/lnwallet/chanfunding/canned_assembler.go
@@ -1,0 +1,187 @@
+package chanfunding
+
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+// ShimIntent is an intent created by the CannedAssembler which represents a
+// funding output to be created that was constructed outside the wallet. This
+// might be used when a hardware wallet, or a channel factory is the entity
+// crafting the funding transaction, and not lnd.
+type ShimIntent struct {
+	// localFundingAmt is the final amount we put into the funding output.
+	localFundingAmt btcutil.Amount
+
+	// remoteFundingAmt is the final amount the remote party put into the
+	// funding output.
+	remoteFundingAmt btcutil.Amount
+
+	// localKey is our multi-sig key.
+	localKey *keychain.KeyDescriptor
+
+	// remoteKey is the remote party's multi-sig key.
+	remoteKey *btcec.PublicKey
+
+	// chanPoint is the final channel point for the to be created channel.
+	chanPoint *wire.OutPoint
+}
+
+// FundingOutput returns the witness script, and the output that creates the
+// funding output.
+//
+// NOTE: This method satisfies the chanfunding.Intent interface.
+func (s *ShimIntent) FundingOutput() ([]byte, *wire.TxOut, error) {
+	if s.localKey == nil || s.remoteKey == nil {
+		return nil, nil, fmt.Errorf("unable to create witness " +
+			"script, no funding keys")
+	}
+
+	totalAmt := s.localFundingAmt + s.remoteFundingAmt
+	return input.GenFundingPkScript(
+		s.localKey.PubKey.SerializeCompressed(),
+		s.remoteKey.SerializeCompressed(),
+		int64(totalAmt),
+	)
+}
+
+// Cancel allows the caller to cancel a funding Intent at any time.  This will
+// return any resources such as coins back to the eligible pool to be used in
+// order channel fundings.
+//
+// NOTE: This method satisfies the chanfunding.Intent interface.
+func (s *ShimIntent) Cancel() {
+}
+
+// RemoteFundingAmt is the amount the remote party put into the channel.
+//
+// NOTE: This method satisfies the chanfunding.Intent interface.
+func (s *ShimIntent) LocalFundingAmt() btcutil.Amount {
+	return s.localFundingAmt
+}
+
+// LocalFundingAmt is the amount we put into the channel. This may differ from
+// the local amount requested, as depending on coin selection, we may bleed
+// from of that LocalAmt into fees to minimize change.
+//
+// NOTE: This method satisfies the chanfunding.Intent interface.
+func (s *ShimIntent) RemoteFundingAmt() btcutil.Amount {
+	return s.remoteFundingAmt
+}
+
+// ChanPoint returns the final outpoint that will create the funding output
+// described above.
+//
+// NOTE: This method satisfies the chanfunding.Intent interface.
+func (s *ShimIntent) ChanPoint() (*wire.OutPoint, error) {
+	if s.chanPoint == nil {
+		return nil, fmt.Errorf("chan point unknown, funding output " +
+			"not constructed")
+	}
+
+	return s.chanPoint, nil
+}
+
+// FundingKeys couples our multi-sig key along with the remote party's key.
+type FundingKeys struct {
+	// LocalKey is our multi-sig key.
+	LocalKey *keychain.KeyDescriptor
+
+	// RemoteKey is the multi-sig key of the remote party.
+	RemoteKey *btcec.PublicKey
+}
+
+// MultiSigKeys returns the committed multi-sig keys, but only if they've been
+// specified/provided.
+func (s *ShimIntent) MultiSigKeys() (*FundingKeys, error) {
+	if s.localKey == nil || s.remoteKey == nil {
+		return nil, fmt.Errorf("unknown funding keys")
+	}
+
+	return &FundingKeys{
+		LocalKey:  s.localKey,
+		RemoteKey: s.remoteKey,
+	}, nil
+}
+
+// A compile-time check to ensure ShimIntent adheres to the Intent interface.
+var _ Intent = (*ShimIntent)(nil)
+
+// CannedAssembler is a type of chanfunding.Assembler wherein the funding
+// transaction is constructed outside of lnd, and may already exist. This
+// Assembler serves as a shim which gives the funding flow the only thing it
+// actually needs to proceed: the channel point.
+type CannedAssembler struct {
+	// fundingAmt is the total amount of coins in the funding output.
+	fundingAmt btcutil.Amount
+
+	// localKey is our multi-sig key.
+	localKey *keychain.KeyDescriptor
+
+	// remoteKey is the remote party's multi-sig key.
+	remoteKey *btcec.PublicKey
+
+	// chanPoint is the final channel point for the to be created channel.
+	chanPoint wire.OutPoint
+
+	// initiator indicates if we're the initiator or the channel or not.
+	initiator bool
+}
+
+// NewCannedAssembler creates a new CannedAssembler from the material required
+// to construct a funding output and channel point.
+func NewCannedAssembler(chanPoint wire.OutPoint, fundingAmt btcutil.Amount,
+	localKey *keychain.KeyDescriptor,
+	remoteKey *btcec.PublicKey, initiator bool) *CannedAssembler {
+
+	return &CannedAssembler{
+		initiator:  initiator,
+		localKey:   localKey,
+		remoteKey:  remoteKey,
+		fundingAmt: fundingAmt,
+		chanPoint:  chanPoint,
+	}
+}
+
+// ProvisionChannel creates a new ShimIntent given the passed funding Request.
+// The returned intent is immediately able to provide the channel point and
+// funding output as they've already been created outside lnd.
+//
+// NOTE: This method satisfies the chanfunding.Assembler interface.
+func (c *CannedAssembler) ProvisionChannel(req *Request) (Intent, error) {
+	switch {
+	// A simple sanity check to ensure the provision request matches the
+	// re-made shim intent.
+	case req.LocalAmt != c.fundingAmt:
+		return nil, fmt.Errorf("intent doesn't match canned assembler")
+
+	// We'll exit out if this field is set as the funding transaction has
+	// already been assembled, so we don't influence coin selection..
+	case req.SubtractFees:
+		return nil, fmt.Errorf("SubtractFees ignored, funding " +
+			"transaction is frozen")
+	}
+
+	intent := &ShimIntent{
+		localKey:  c.localKey,
+		remoteKey: c.remoteKey,
+		chanPoint: &c.chanPoint,
+	}
+
+	if c.initiator {
+		intent.localFundingAmt = c.fundingAmt
+	} else {
+		intent.remoteFundingAmt = c.fundingAmt
+	}
+
+	return intent, nil
+}
+
+// A compile-time assertion to ensure CannedAssembler meets the Assembler
+// interface.
+var _ Assembler = (*CannedAssembler)(nil)
--- a/lnwallet/chanfunding/coin_select.go
+++ b/lnwallet/chanfunding/coin_select.go
@@ -1,0 +1,216 @@
+package chanfunding
+
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ErrInsufficientFunds is a type matching the error interface which is
+// returned when coin selection for a new funding transaction fails to due
+// having an insufficient amount of confirmed funds.
+type ErrInsufficientFunds struct {
+	amountAvailable btcutil.Amount
+	amountSelected  btcutil.Amount
+}
+
+// Error returns a human readable string describing the error.
+func (e *ErrInsufficientFunds) Error() string {
+	return fmt.Sprintf("not enough witness outputs to create funding "+
+		"transaction, need %v only have %v  available",
+		e.amountAvailable, e.amountSelected)
+}
+
+// Coin represents a spendable UTXO which is available for channel funding.
+// This UTXO need not reside in our internal wallet as an example, and instead
+// may be derived from an existing watch-only wallet. It wraps both the output
+// present within the UTXO set, and also the outpoint that generates this coin.
+type Coin struct {
+	wire.TxOut
+
+	wire.OutPoint
+}
+
+// selectInputs selects a slice of inputs necessary to meet the specified
+// selection amount. If input selection is unable to succeed due to insufficient
+// funds, a non-nil error is returned. Additionally, the total amount of the
+// selected coins are returned in order for the caller to properly handle
+// change+fees.
+func selectInputs(amt btcutil.Amount, coins []Coin) (btcutil.Amount, []Coin, error) {
+	satSelected := btcutil.Amount(0)
+	for i, coin := range coins {
+		satSelected += btcutil.Amount(coin.Value)
+		if satSelected >= amt {
+			return satSelected, coins[:i+1], nil
+		}
+	}
+
+	return 0, nil, &ErrInsufficientFunds{amt, satSelected}
+}
+
+// CoinSelect attempts to select a sufficient amount of coins, including a
+// change output to fund amt satoshis, adhering to the specified fee rate. The
+// specified fee rate should be expressed in sat/kw for coin selection to
+// function properly.
+func CoinSelect(feeRate chainfee.SatPerKWeight, amt btcutil.Amount,
+	coins []Coin) ([]Coin, btcutil.Amount, error) {
+
+	amtNeeded := amt
+	for {
+		// First perform an initial round of coin selection to estimate
+		// the required fee.
+		totalSat, selectedUtxos, err := selectInputs(amtNeeded, coins)
+		if err != nil {
+			return nil, 0, err
+		}
+
+		var weightEstimate input.TxWeightEstimator
+
+		for _, utxo := range selectedUtxos {
+			switch {
+
+			case txscript.IsPayToWitnessPubKeyHash(utxo.PkScript):
+				weightEstimate.AddP2WKHInput()
+
+			case txscript.IsPayToScriptHash(utxo.PkScript):
+				weightEstimate.AddNestedP2WKHInput()
+
+			default:
+				return nil, 0, fmt.Errorf("unsupported address type: %x",
+					utxo.PkScript)
+			}
+		}
+
+		// Channel funding multisig output is P2WSH.
+		weightEstimate.AddP2WSHOutput()
+
+		// Assume that change output is a P2WKH output.
+		//
+		// TODO: Handle wallets that generate non-witness change
+		// addresses.
+		// TODO(halseth): make coinSelect not estimate change output
+		// for dust change.
+		weightEstimate.AddP2WKHOutput()
+
+		// The difference between the selected amount and the amount
+		// requested will be used to pay fees, and generate a change
+		// output with the remaining.
+		overShootAmt := totalSat - amt
+
+		// Based on the estimated size and fee rate, if the excess
+		// amount isn't enough to pay fees, then increase the requested
+		// coin amount by the estimate required fee, performing another
+		// round of coin selection.
+		totalWeight := int64(weightEstimate.Weight())
+		requiredFee := feeRate.FeeForWeight(totalWeight)
+		if overShootAmt < requiredFee {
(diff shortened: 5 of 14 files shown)


Roasbeef:
The final diff looks larger than it actually is due to the code movement that has taken place.

wpaulino:
Haven't started review yet, but there are a few linter failures causing Travis to fail.

lnwallet/wallet.go
```diff
@@ -446,11 +452,25 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 		return
 	}
 
+	// If no chanFunder was provided, then we'll assume the default
+	// assembler, which is backed by the wallet's internal coin selection.
+	if req.ChanFunder == nil {
```
halseth: Isn't a better separation to let the  creator of `InitFundingReserveMsg` set this? That way we can (slowly) move towards separating this channel funding code from the wallet.

lnwallet/wallet.go
```diff
@@ -461,41 +481,61 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
+			FeeRate:      req.FundingFeePerKw,
+			ChangeAddr: func() (btcutil.Address, error) {
+				return l.NewAddress(WitnessPubKey, true)
+			},
+		}
+		fundingIntent, err = req.ChanFunder.ProvisionChannel(
+			fundingReq,
 		)
 		if err != nil {
+			if fundingIntent != nil {
```
halseth: assumption  should be that `fundingIntent` is always nil whenever an error is returned.

lnwallet/wallet.go
```diff
@@ -518,52 +558,55 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 
 	// Grab the mutex on the ChannelReservation to ensure thread-safety
 	reservation.Lock()
 	defer reservation.Unlock()
 
-	if selected != nil {
-		reservation.ourContribution.Inputs = selected.coins
-		reservation.ourContribution.ChangeOutputs = selected.change
+	if fundingIntent != nil {
+		if intent, ok := fundingIntent.(*chanfunding.FullIntent); ok {
```
halseth: add an explaining comment

lnwallet/wallet.go
```diff
@@ -707,65 +750,20 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
...
-	for i, txIn := range fundingTx.TxIn {
-		info, err := l.FetchInputInfo(&txIn.PreviousOutPoint)
-		if err == ErrNotMine {
-			continue
-		} else if err != nil {
-			req.err <- err
+	// At this point, we can now construct our channel point. Depenindg on
+	// which type of intent we obtained from our chanfunding.Assembler,
+	// we'll carry out a distinct set of steps.
+	switch fundingIntent := pendingReservation.fundingIntent.(type) {
```
halseth: uh, what's going on here? Does this compile?

halseth: Looks like `case` is missing.

lnwallet/chanfunding/external_assembler.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// RemoteFundingAmt is the amount the remote party put into the channel.
+func (s *ShimIntent) LocalFundingAmt() btcutil.Amount {
+	return s.localFundingAmt
+}
+
+// LocalFundingAmt is the amount we put into the channel. This may differ from
+// the local amount requested, as depending on coin selection, we may bleed
+// from of that LocalAmt into fees to minimize change.
```
halseth: Add a note which methods are satisfying the interface.

lnwallet/chanfunding/external_assembler.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// A compile-time check to ensure ShimIntent adheres to the Intent interface.
+var _ Intent = (*ShimIntent)(nil)
+
+// CannedAssembler is a type of chanfunding.Assembler wherein the funding
+// transaction is constructed outside of lnd, and the may already exist. This
+// Assembler serves as a shim which gives the funding flow the only thing it
+// actually needs to proceed: the channel point.
+type CannedAssembler struct {
```
halseth: name: ExternalAssembler? Where is the can coming from? 😛

lnwallet/wallet.go
```diff
@@ -510,7 +530,16 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 
 		localFundingAmt = fundingIntent.LocalFundingAmt()
 		remoteFundingAmt = fundingIntent.RemoteFundingAmt()
+
+	} else if shimIntent, ok := l.fundingIntents[req.PendingChanID]; ok {
+		// Alternatively, if we've just received an inbound funding
+		// request that we have a registered him intent to, then we'll
```
halseth: s/him/shim

lnwallet/wallet.go
```diff
@@ -510,7 +530,16 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
+		// obtain the backing intent now. In this case, we're doing a
+		// special funding workflow that allows more advanced
+		// constructions such as channel factories to be instantiated.
+		fundingIntent = shimIntent
 	}
+	// TODO(roasbeef): need mutex above
```
halseth: reminder to use mutex, or pass intent via a `handleRegisterFundingIntent` to avoid mutex?

log.go
```diff
@@ -96,6 +96,7 @@ func init() {
 	addSubLogger("PROM", monitoring.UseLogger)
 	addSubLogger("WTCL", wtclient.UseLogger)
 	addSubLogger("PRNF", peernotifier.UseLogger)
+	addSubLogger("CHFD", chanfitness.UseLogger)
```
halseth: squash with commit where logger was added

lnwallet/chanfunding/assembler.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// CoinSource is an interface that allows a caller to access a source of UTXOs
+// to use when attempting to fund a new channel.
+type CoinSource interface {
+	// ListCoins returns all UTXOs from the source that have between
+	// minConfs and maxConfs number of confirmations.
+	ListCoins(minConfs, maxConfs int32) ([]Coin, error)
```
halseth: Use  `*Coin` as return value, similar to below method?

lnwallet/chanfunding/assembler.go
```diff
@@ -0,0 +1,124 @@
...
+// Cancel method is called, then all resources assembled to fund the channel
+// will be released back to the eligible pool.
+type Intent interface {
+	// FundingOutput returns the witness script, and the output that
+	// creates the funding output.
+	FundingOutput() ([]byte, *wire.TxOut, error)
+
+	// ChanPoint returns the final outpoint that will create the funding
+	// output described above.
+	ChanPoint() (*wire.OutPoint, error)
```
halseth: Is there a reason we cannot also (always, even for external intents) access the final funding tx from the `Intent`?

lnwallet/chanfunding/external_assembler.go
halseth: oh, I see the file name is actually `external_assembler`. Why was the name changed to `canned`?

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,320 @@
...
+// return any resources such as coins back to the eligible pool to be used in
+// order channel fundings.
+//
+// NOTE: Part of the chanfunding.Intent interface.
+func (f *FullIntent) Cancel() error {
+	for _, coin := range f.InputCoins {
+		f.coinLocker.UnlockOutpoint(coin.OutPoint)
+	}
+
+	return nil
```
halseth: nit: should also call the `Cancel` of the underlying `ShimIntent`? (even though it is empty as of now)

lnwallet/wallet.go
```diff
@@ -720,100 +800,82 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
...
-			PkScript: info.PkScript,
-			Value:    int64(info.Value),
-		}
-		signDesc.InputIndex = i
+		pendingReservation.partialState.FundingOutpoint = *chanPoint
+
+	case *chanfunding.FullIntent:
+		// Now that we know their public key, we can bind theirs as
+		// well as ours to the funding intent.
+		fundingIntent.BindKeys(
```
halseth: a safer way of doing this could be to provide the keys as parameters to `CompileFundingTx()`

lnwallet/chanfunding/external_assembler.go
```diff
@@ -0,0 +1,168 @@
...
+	// localFundingAmt is the final amount we put into the funding output.
+	localFundingAmt btcutil.Amount
+
+	// remoteFundingAmt is the final amount the remote party put into the
+	// funding output.
+	remoteFundingAmt btcutil.Amount
+
+	// Request is the original request that led to the creation of this
+	// funding intent.
+	Request *Request
```
halseth: is this field used anywhere?

lnwallet/chanfunding/external_assembler.go
```diff
@@ -0,0 +1,168 @@
...
+	chanPoint wire.OutPoint
+
+	// initiator indicates if we're the initiator or the channel or not.
+	initiator bool
+}
+
+// NewCannedAssembler creates a new CannedAssembler from the material required
+// to construct a funding output and channel point.
+func NewCannedAssembler(chanPoint wire.OutPoint, fundingAmt btcutil.Amount,
+	localKey *keychain.KeyDescriptor,
```
halseth: seems like there's an overlap between these params and the `Request`

lnwallet/reservation.go
```diff
@@ -220,6 +224,14 @@ func NewChannelReservation(capacity, localFundingAmt btcutil.Amount,
 		} else {
 			chanType |= channeldb.SingleFunderBit
 		}
+
+		// If this isn't the regular internal channel funder, then this
+		// means that we won't have the funding transaction regardless
+		// of if we're the initiator or not.
+		if _, ok := fundingAssembler.(*chanfunding.WalletAssembler); !ok {
```
halseth: hm, could the `Assembler` instead expose whether the funding tx is available? Doing it this way adding more assemblers in the future, they will all default to `NoFundingTx`

lnwallet/wallet.go
```diff
@@ -1370,7 +1368,7 @@ func (l *LightningWallet) selectCoinsAndChange(feeRate chainfee.SatPerKWeight,
 	// amount.
 	default:
 		fundingAmt = amt
-		selectedCoins, changeAmt, err = coinSelect(feeRate, amt, coins)
+		selectedCoins, changeAmt, err = chanfunding.CoinSelect(feeRate, amt, coins)
```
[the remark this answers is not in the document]
Roasbeef: No present in the final diff.

lnwallet/wallet.go
Roasbeef: As is it does, but the idea here it to provide a default so the caller doesn't have to. In this case, the caller is the funding manager, which isn't _yet_ aware of the new channel funding abstractions. Having the default here also meshes well into the existing funding workflow tests in `interface_test.go` as well. 

In the end, we call this file `wallet.go`, but it's almost purely implementing the funding workflow.

lnwallet/wallet.go
Roasbeef: Hmm, yeah that's a safe assumption, removed.

lnwallet/wallet.go
Roasbeef: Fixed.

lnwallet/chanfunding/external_assembler.go
Roasbeef: lol, canned as in "pre-recorded", not spontaneous: https://en.wiktionary.org/wiki/canned

lnwallet/wallet.go
Roasbeef: To grab mutex, TODO resolved.

lnwallet/chanfunding/assembler.go
Roasbeef: Pointer is only used there so I can return `nil` and not have to create a dummy struct.

lnwallet/chanfunding/assembler.go
Roasbeef: Not all intents can provide a funding transaction (if it was constructed externally as an example).

lnwallet/wallet.go
Roasbeef: I considered that, but then it wouldn't be possible to obtain the funding output in isolation without also constructing+signing the funding transaction.

lnwallet/chanfunding/external_assembler.go
Roasbeef: Not atm, but imo makes sense to also package the request as well, not strongly attached to it though.

lnwallet/chanfunding/external_assembler.go
Roasbeef: Yeah there's some overlap, but also newer arguments like the `initiator` bool.

lnwallet/reservation.go
Roasbeef: >  Doing it this way adding more assemblers in the future, they will all default to NoFundingTx

Don't see anything wrong with this, as they may also start to define _new_ type bits themselves. As is, only the core wallet assembler has a funding transaction.

lnwallet/chanfunding/assembler.go
halseth: The only case I see is that the signatures can be added later, but the tx itself must be assembled for the txid to be known.

lnwallet/reservation.go
```diff
@@ -114,11 +114,6 @@ type ChannelReservation struct {
 	// commitment state.
 	pushMSat lnwire.MilliSatoshi
 
-	// chanOpen houses a struct containing the channel and additional
-	// confirmation details will be sent on once the channel is considered
-	// 'open'. A channel is open once the funding transaction has reached a
-	// sufficient number of confirmations.
-	chanOpen    chan *openChanDetails
```
halseth: commit doesn't compile
```
lnwallet/reservation.go:257:3: unknown field 'chanOpen' in struct literal of type ChannelReservation
```

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,322 @@
...
+// before CompileFundingTx can be called.
+//
+// Steps to final channel provisioning:
+//  1. Call BindKeys to notify the intent which keys to use when constructing
+//  the multi-sig output.
+//  2. Call CompileFundingTx afterwrds to obtain the funding transaction.
+//
+// If either of these steps fail, then the Cancel method MUST be called.
+type FullIntent struct {
+	ShimIntent
```
halseth:
```
lnwallet/chanfunding/wallet_assembler.go:28:2: undefined: ShimIntent
```

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// signer is the Assembler's instance of the Singer interface.
+	signer input.Signer
+}
+
+// BindKeys is a method unique to the FullIntent variant. This allows the
+// caller to decide precisely which keys are used in the final funding
+// transaction. This is kept out of the main Assembler as these may may not
+// necessarily be under full control of the wallet. Only after this method has
+// been executed is will CompileFundingTx succeed.
```
halseth: ~~is~~

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+
+		intent = &FullIntent{
+			ShimIntent: ShimIntent{
+				Request:          r,
+				localFundingAmt:  localContributionAmt,
+				remoteFundingAmt: r.RemoteAmt,
+			},
+			InputCoins:    selectedCoins,
+			ChangeOutputs: []*wire.TxOut{changeOutput},
```
halseth: change output can be nil here

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,322 @@
...
+			coinLocker:    w.cfg.CoinLocker,
+			coinSource:    w.cfg.CoinSource,
+			signer:        w.cfg.Signer,
```
halseth: why do we need to set these unexported fields?

log.go
halseth: should be `chanfunding`?

lnwallet/reservation.go
halseth: I mean this way the reservation logic must be aware of the actual type of assembler. This won't allow me to define a new wallet assembler in an external package that satisfies the interface and also provide the funding tx. Defining it on the interface would let me do that.

Could even define a new interface 
```go
type FundingTxProvider interface {
   FundingTx func() *wire.MsgTx
}
```
and match on that here
```go
if _, ok := fundingAssembler.(chanfunding.FundingTxProvider); !ok {
    chanType |= channeldb.NoFundingTxBit
}
```

lnwallet/reservation.go
```diff
@@ -554,3 +566,7 @@ type OpenChannelDetails struct {
 	// transaction resides.
 	TransactionIndex uint32
 }
+
+// A compile-time assertion to ensure coinSource meets the
+// chanFunder.CoinSelectLocker interface.
```
halseth: type: should be `CoinSource`

lnwallet/reservation.go
```diff
@@ -116,7 +116,10 @@ type ChannelReservation struct {
 
 	chanOpenErr chan error
 
-	wallet *LightningWallet
+	wallet     *LightningWallet
+	chanFunder chanfunding.Assembler
```
halseth: chanfunding not imported

lnwallet/wallet.go
```diff
@@ -518,52 +553,60 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
+					reservation.ourContribution.Inputs,
+					&wire.TxIn{
+						PreviousOutPoint: coin.OutPoint,
+					},
+				)
+			}
+			reservation.ourContribution.ChangeOutputs = intent.ChangeOutputs
+		}
+
+		reservation.fundingIntent = fundingIntent
```
halseth: would it make sense to define a type of "remote funding intent"? Such that if you're the the responder in the workflow, then you also have an intent to refer to.

lnwallet/wallet.go
```diff
@@ -707,100 +750,73 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
...
-	// Sort the transaction. Since both side agree to a canonical ordering,
-	// by sorting we no longer need to send the entire transaction. Only
-	// signatures will be exchanged.
-	fundingTx.AddTxOut(multiSigOut)
-	txsort.InPlaceSort(pendingReservation.fundingTx)
+	// At this point, we can now construct our channel point. Depenindg on
+	// which type of intent we obtained from our chanfunding.Assembler,
+	// we'll carry out a distinct set of steps.
+	switch fundingIntent := pendingReservation.fundingIntent.(type) {
+	case *chanfunding.FullIntent:
```
halseth: Doing it this way, you will need to add to the switch case every time one adds a new intent type. Would be nice to have this part of the code work only with the interface, but maybe not that easy.

lnwallet/wallet.go
```diff
@@ -990,44 +994,78 @@ func (l *LightningWallet) handleFundingCounterPartySigs(msg *addCounterPartySigs
...
+	defer res.Unlock()
+
+	// Now we can complete the funding transaction by adding their
+	// signatures to their inputs.
+	res.theirFundingInputScripts = msg.theirFundingInputScripts
+	inputScripts := msg.theirFundingInputScripts
+
+	// Only if we have the final funding transaction do we need to verify
+	// the final set of inputs. Otherwise, it may be the case that the
+	// channel was funded via n external wallet.
```
halseth: typo: n

lnwallet/chanfunding/external_assembler.go
halseth: Not a very descriptive name IMO, but maybe because I'm not a native speaker 😛

lnwallet/chanfunding/external_assembler.go
halseth: Probably make sense to remove, and add if we need it at some point later

lnwallet/chanfunding/assembler.go
```diff
@@ -0,0 +1,124 @@
...
+	// synchronized manner preventing any coin selection operations from
+	// proceeding while the closure if executing. This can be seen as the
+	// ability to execute a function closure under an exclusive coin
+	// selection lock.
+	WithCoinSelectLock(func() error) error
+}
+
+// OutpointLocker allows a caller to lock/unlock an outpoint. When locked, the
+// outpoints shouldn't be used for any sort of channel funding of coin
+// selection. Locked outpoints are not expect to be persisted between restarts.
```
guggero: nit: s/expect/expected/

lnwallet/chanfunding/wallet_assembler.go
guggero: They are used in `CompileFundingTx`.

lnwallet/chanfunding/canned_assembler.go
```diff
@@ -0,0 +1,179 @@
...
+		LocalKey:  s.localKey,
+		RemoteKey: s.remoteKey,
+	}, nil
+}
+
+// A compile-time check to ensure ShimIntent adheres to the Intent interface.
+var _ Intent = (*ShimIntent)(nil)
+
+// CannedAssembler is a type of chanfunding.Assembler wherein the funding
+// transaction is constructed outside of lnd, and the may already exist. This
```
guggero: nit: word missing?

lnwallet/chanfunding/canned_assembler.go
```diff
@@ -0,0 +1,179 @@
...
+// funding output to be created that was constructed outside the wallet. This
+// might be used when a hardware wallet, or a channel factory is the entity
+// crafting the funding transaction, and not lnd.
+type ShimIntent struct {
+	// localFundingAmt is the final amount we put into the funding output.
+	localFundingAmt btcutil.Amount
+
+	// remoteFundingAmt is the final amount the remote party put into the
+	// funding output.
+	remoteFundingAmt btcutil.Amount
```
wpaulino: Would be nice to include the rationale for these fields in their `godoc` as they already exist within `Request`.

lnwallet/wallet.go
```diff
@@ -505,6 +526,18 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 
 		localFundingAmt = fundingIntent.LocalFundingAmt()
 		remoteFundingAmt = fundingIntent.RemoteFundingAmt()
+
+	} else {
+		l.intentMtx.Lock()
+		// Alternatively, if we've just received an inbound funding
+		// request that we have a regisstered him intent to, then we'll
```
wpaulino: Nit: s/regisstered him intent to/registered intent for/

lnwallet/wallet.go
```diff
@@ -698,6 +731,10 @@ func (l *LightningWallet) handleFundingCancelRequest(req *fundingReserveCancelMs
 
 	delete(l.fundingLimbo, req.pendingFundingID)
 
+	l.intentMtx.Lock()
+	delete(l.fundingIntents, pendingReservation.pendingChanID)
```
wpaulino: Should this also cancel the intent as well?

lnwallet/wallet.go
```diff
@@ -474,41 +505,82 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 	reservation, err := NewChannelReservation(
 		capacity, localFundingAmt, req.CommitFeePerKw, l, id,
 		req.PushMSat, l.Cfg.NetParams.GenesisHash, req.Flags,
-		req.Tweakless,
+		req.Tweakless, req.ChanFunder, req.PendingChanID,
 	)
 	if err != nil {
-		selected.unlockCoins()
+		if fundingIntent != nil {
+			fundingIntent.Cancel()
```
wpaulino: Shouldn't we also do these for the other steps, e.g., `handleContributionMsg`?

lnwallet/chanfunding/assembler.go
```diff
@@ -0,0 +1,124 @@
...
+type OutpointLocker interface {
```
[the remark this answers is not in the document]
Roasbeef: They were purposefully copied over as they're pretty small interfaces, so would rather not create additional dependancies. It's also the case that if the interfaces in the sweeper are used, then a circular dependency cycle is created.

lnwallet/chanfunding/assembler.go
Roasbeef: No. I can pass you a channel point and you can make a channel, all without _you_ ever knowing the funding transaction until it has been confirmed.

lnwallet/wallet.go
Roasbeef: Not sure if you've seen later, but we extend things to allow a caller to inject an intent. This is meant for the case when you're the responder to an extra-protocol initiated funding flow.

lnwallet/wallet.go
```diff
@@ -707,100 +750,73 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
...
+			}
 
-	// Locate the index of the multi-sig outpoint in order to record it
-	// since the outputs are canonically sorted. If this is a single funder
-	// workflow, then we'll also need to send this to the remote node.
-	fundingTxID := fundingTx.TxHash()
-	_, multiSigIndex := input.FindScriptOutputIndex(fundingTx, multiSigOut.PkScript)
-	fundingOutpoint := wire.NewOutPoint(&fundingTxID, multiSigIndex)
-	pendingReservation.partialState.FundingOutpoint = *fundingOutpoint
+			pendingReservation.ourFundingInputScripts = append(
```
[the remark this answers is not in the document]
Roasbeef: The reservations? Yes, there's an overlap, but I kept them both in place to minimize the size of the PR. As is now, the reservations are basically a "funding builder" in that they're a bridge between the internal wallet flow/data-format and what we use on the wire. The new intents are solely concerned with assembling what's needed in order to sign commitment transactions and progress in the workflow. 

I tried to make the smallest cutout possible (abstract over how we obtain the chan point and funding transactions), to allow the intended use case.

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,322 @@
...
+type FullIntent struct {
+	ShimIntent
+
+	// InputCoins are the set of coins selected as inputs to this funding
+	// transaction.
+	InputCoins []Coin
+
+	// ChangeOutputs are the set of outputs that the Assembler will use as
+	// change from the main funding transaction.
+	ChangeOutputs []*wire.TxOut
```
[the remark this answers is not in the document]
Roasbeef: Left over from our dual funding workflow that allows each side to have multiple change outputs so they can piggy back on the funding transaction to produce additional possibly unrelated outputs.

lnwallet/chanfunding/canned_assembler.go
Roasbeef: The request is just riding along at this point. The actual amounts are passed in via the `NewCannedAssembler` method as all the parameters for the channel are already known at this point.

lnwallet/wallet.go
Roasbeef: To free up the coins? If a reservation is failed, and doesn't proceed, then they'll be reclaimed via the zombie reservation sweeper within the `fundingManager`.

lnwallet/reservation.go
Roasbeef: Gotcha, implemented something like this for the current assembler. However it can't return the funding transaction at that time as it doesn't know everything it needs to do so. Instead, I've added what's more or less just a signalling method for outside callers.

lnwallet/chanfunding/assembler.go
```diff
@@ -0,0 +1,124 @@
...
+// be used when an operation requires that all coin selection operations cease
+// forward progress. Think of this as an exclusive lock on coin selection
+// operations.
+type CoinSelectionLocker interface {
+	// WithCoinSelectLock will execute the passed function closure in a
+	// synchronized manner preventing any coin selection operations from
+	// proceeding while the closure if executing. This can be seen as the
+	// ability to execute a function closure under an exclusive coin
+	// selection lock.
+	WithCoinSelectLock(func() error) error
```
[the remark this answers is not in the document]
Roasbeef: The same could also be said for the `OutpointLocker` interface as well (which is basically how things are on the mega `WalletController` interface that we're trying to break down over time). I chose to keep them distinct (when they were originally added) as they're independently useful on their own, and one may be able to lock/unlock coins without also being able to return the full set of coins. There's an implicit requirement that since they've been passed in to the config, they're all pointing to the same coin source.

lnwallet/chanfunding/assembler.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// OutpointLocker allows a caller to lock/unlock an outpoint. When locked, the
+// outpoints shouldn't be used for any sort of channel funding of coin
+// selection. Locked outpoints are not expect to be persisted between restarts.
+type OutpointLocker interface {
+	// LockOutpoint locks a target outpoint, rendering it unusable for coin
+	// selection.
+	LockOutpoint(o wire.OutPoint)
```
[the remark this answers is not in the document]
Roasbeef: That's currently undefined. The main `WalletController` implementation uses locking that is only persisted in memory, so the core call cannot fail. We don't have any implementations that persist the coins as is. Only when we _do_ gain one should these methods gain an error return type IMO. Not adding this also saves us from adding yet another wrapper struct to satisfy a _slightly_ different interface.

lnwallet/wallet.go
Roasbeef: I think after other funding intents are added, a set of shared capabilities will emerge, leading us to create a new set of interfaces that communicates those shared capabilities. The next main `chanfunding.Intent` I have planned is one that exposes PSBT functionality. I've thought about a few ways to add this to the current package, and one of them is to cause `CompileFundingTx` to block until signing has been completed, with a new string of messages being passed from the RPC client, to the funding manager, then finally into the backing intent.

fundingmanager.go
```diff
@@ -521,8 +521,9 @@ func (f *fundingManager) start() error {
 
 			// Rebroadcast the funding transaction for any pending
 			// channel that we initiated. No error will be returned
-			// if the transaction already has been broadcasted.
-			if channel.ChanType.IsSingleFunder() &&
+			// if the transaction already has been broadcast.
+			chanType := channel.ChanType
+			if chanType.IsSingleFunder() && chanType.HasFundingTx() &&
```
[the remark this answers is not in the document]
Roasbeef: by 2 characters 😅

lnwallet/chanfunding/coin_select.go
```diff
@@ -0,0 +1,216 @@
...
+type ErrInsufficientFunds struct {
+	amountAvailable btcutil.Amount
+	amountSelected  btcutil.Amount
+}
+
+// Error returns a human readable string describing the error.
+func (e *ErrInsufficientFunds) Error() string {
+	return fmt.Sprintf("not enough witness outputs to create funding "+
+		"transaction, need %v only have %v  available",
+		e.amountAvailable, e.amountSelected)
```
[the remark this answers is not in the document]
Roasbeef: It's actually correct, though the naming is a bit off here. `amountAvailable` is the amount that we _want_, while `amountSelected` is what we've been able to gather so far.

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+	for i, txIn := range fundingTx.TxIn {
+		// We can only sign this input if it's ours, so we'll ask the
+		// coin source if it can map this outpoint into a coin we own.
+		// If not, then we'll continue as it isn't our input.
+		info, err := f.coinSource.CoinFromOutPoint(
+			txIn.PreviousOutPoint,
+		)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
Roasbeef: The interface states to return an error when the coin isn't found. This interface doesn't define such an error as is. If the funding transaction isn't complete then it'll fail full verification later in the funding workflow.

fundingmanager.go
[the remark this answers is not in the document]
guggero: I wasn't sure if the linter enforced this, but apparently not. Never mind then :)

lnwallet/wallet.go
```diff
@@ -463,41 +482,57 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
+		}
+
 		req.err <- err
 		req.resp <- nil
 		return
 	}
 
 	err = l.initOurContribution(
-		reservation, selected, req.NodeAddr, req.NodeID,
+		reservation, fundingIntent, req.NodeAddr, req.NodeID, keyRing,
```
guggero: Commit doesn't compile:
```
lnwallet/wallet.go:529:57: undefined: keyRing
```

lnwallet/wallet.go
```diff
@@ -463,41 +482,57 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 	// created plus the remote contribution.
-	capacity := localFundingAmt + req.RemoteFundingAmt
+	capacity := localFundingAmt + remoteFundingAmt
 
 	id := atomic.AddUint64(&l.nextFundingID, 1)
 	reservation, err := NewChannelReservation(
 		capacity, localFundingAmt, req.CommitFeePerKw, l, id,
 		req.PushMSat, l.Cfg.NetParams.GenesisHash, req.Flags,
-		req.Tweakless,
+		req.Tweakless, req.ChanFunder, req.PendingChanID,
```
guggero: Commit doesn't compile:
```
lnwallet/wallet.go:516:37: req.PendingChanID undefined (type *InitFundingReserveMsg has no field or method PendingChanID)
```

lnwallet/chanfunding/canned_assembler.go
```diff
@@ -0,0 +1,179 @@
...
+		chanPoint:  chanPoint,
+	}
+}
+
+// ProvisionChannel creates a new ShimIntent given the passed funding Request.
+// The returned intent is immediately able to provide the channel point and
+// funding output as they've already been created outside lnd.
+//
+// NOTE: This method satisfies the chanfunding.Assembler interface.
+func (c *CannedAssembler) ProvisionChannel(req *Request) (Intent, error) {
```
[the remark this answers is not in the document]
halseth: It should at least provide some basic validation IMO to make sure the returned `Intent` is actually usable with the request. Or is that added in the follow-up PR?

lnwallet/wallet.go
Roasbeef: The method as is signs the funding transaction. For the upcoming PSBT integration, we'll need just the output itself to insert into the partial transaction for signing. As far as the current state, the docs are clear that one it to interact with the object in a builder-like format. A similar format will be used for the PSBT integration (get unsigned, sign our portion, combine with remote party's sig, etc).

lnwallet/chanfunding/assembler.go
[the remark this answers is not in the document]
Roasbeef: The `godoc` states: 
> should be synchronized with all coin selection attempts

These are the very same interfaces we use today in `sweep/walletsweep.go`.

lnwallet/reservation.go
```diff
@@ -550,3 +564,7 @@ type OpenChannelDetails struct {
 	// transaction resides.
 	TransactionIndex uint32
 }
+
+// A compile-time assertion to ensure coinSource meets the
+// chanfunding.CoinSource interface.
+var _ chanfunding.CoinSource = (*CoinSource)(nil)
```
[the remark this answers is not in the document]
Roasbeef: Updated!

channeldb/channel.go
```diff
@@ -166,6 +171,12 @@ func (c ChannelType) IsTweakless() bool {
 	return c&SingleFunderTweaklessBit == SingleFunderTweaklessBit
 }
 
+// HasFundingTx returns true if this channel type is one that has a funding
+// transaction stored locally.
+func (c ChannelType) HasFundingTx() bool {
+	return c&NoFundingTxBit == 0
```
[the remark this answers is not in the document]
Roasbeef: Serialization logic has been update to check this value. The channel type is written first, so after that has been read, we need to know if we should attempt to read the funding transaction or not.

fundingmanager.go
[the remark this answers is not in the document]
Roasbeef: Initiator is still needed as things are written as is since `HasFundingTx` isn't dependent on who the initiator is.

channeldb/channel.go
```diff
@@ -166,6 +171,12 @@ func (c ChannelType) IsTweakless() bool {
 	return c&SingleFunderTweaklessBit == SingleFunderTweaklessBit
 }
 
+// HasFundingTx returns true if this channel type is one that has a funding
+// transaction stored locally.
+func (c ChannelType) HasFundingTx() bool {
```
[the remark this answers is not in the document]
Roasbeef: If the bit _isn't_ set, then that means we _have_ the funding transaction. If it's set, then it means we don't. As written, if we're about to attempt to handle the funding transaction, we also check if we're the initiator or not.

lnwallet/chanfunding/canned_assembler.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// A compile-time check to ensure ShimIntent adheres to the Intent interface.
+var _ Intent = (*ShimIntent)(nil)
+
+// CannedAssembler is a type of chanfunding.Assembler wherein the funding
+// transaction is constructed outside of lnd, and may already exist. This
+// Assembler serves as a shim which gives the funding flow the only thing it
+// actually needs to proceed: the channel point.
+type CannedAssembler struct {
```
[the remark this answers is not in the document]
Roasbeef: It means that it's pre-made (could come in a can).

lnwallet/chanfunding/canned_assembler.go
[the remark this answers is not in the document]
Roasbeef: It's a facade, it allows the wallet to always call `ProvisionChannel` no matter what assembler is passed to it. By definition, all the parameters are already known (it's a canned assembler). Have you checked out how it's actually used at the call site? 

What type of validation would be added? The assumption is that validation has taken place externally: both sides already know all the parameters. As a result, we pass along this shim to the wallet, it does biz as normal but doesn't need to know that the funding has actually been externally derived.

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,338 @@
...
+
+	return intent, nil
+}
+
+// FundingTx is an empty method that an assembler can implement to
+// signal to callers that its able to provide the funding transaction
+// for the channel via the intent it returns.
+//
+// NOTE: This method is a part of the FundingTxAssembler interface.
+func (w *WalletAssembler) FundingTx() {}
```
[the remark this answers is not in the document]
Roasbeef: It isn't available until it has all the necessary information. This is just an empty method to allow callers to assert against the interface rather than the concrete type. I've renamed the method to better reflect that.

lnwallet/wallet.go
```diff
@@ -525,6 +525,19 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 		return
 	}
 
+	var keyRing keychain.KeyRing = l.SecretKeyRing
+
+	// If this is a shim intent, then it may be attempting to use an
+	// existing set of keys for the funding workflow. In this case, we'll
+	// make a simple wrapper keychain.KeyRing that will proxy certain
+	// derivation calls to future callers.
+	if shimIntent, ok := fundingIntent.(*chanfunding.ShimIntent); ok {
```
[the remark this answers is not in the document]
Roasbeef: That would just be another layer of indirection which isn't even strictly needed in this case. It isn't clear what you're referring to, not what this would add (but another empty method interface). The `shimKeyRing` requires a `ShimIntent` in order to work as laid out: it passes the already agreed upon key to the wallet in order to further the funding flow. 

We can attempt to reason about all the possible ways that this may be used in the future, but they're all speculative. The only thing we need this for as is, is to enable external channel funding. The way things are set up, it also eventually enables callers to also specify their own _commitment key_ as well (since that can be entirely offline until we need to sweep our output on chain).

Roasbeef:
The issue in the tests should be fixed now, see the last commit w.r.t the exact situation (`GetUtxo` for neutrino treats the height hint as a _precise_ birth height rather than a lower bound on the birth height of the UTXO).

lnwallet/chanfunding/wallet_assembler.go
```diff
@@ -0,0 +1,338 @@
...
+				Value:    int64(changeAmt),
+				PkScript: changeScript,
+			}
+		}
+
+		// Lock the selected coins. These coins are now "reserved",
+		// this prevents concurrent funding requests from referring to
+		// and this double-spending the same set of coins.
+		for _, coin := range selectedCoins {
+			outpoint := &coin.OutPoint
```
wpaulino: Nit: this seems unnecessary? Method below does not require a pointer.

lnwallet/wallet.go
wpaulino: Following that logic, why bother doing it here then? Seems inconsistent.

lnwallet/interface_test.go
```diff
@@ -3130,9 +3130,18 @@ func runTests(t *testing.T, walletDriver *lnwallet.WalletDriver,
...
 	// wallet state after each step.
 	for _, walletTest := range walletTests {
+
+		walletTest := walletTest
+
 		testName := fmt.Sprintf("%v/%v:%v", walletType, backEnd,
 			walletTest.name)
 		success := t.Run(testName, func(t *testing.T) {
+			if backEnd == "neutrino" &&
+				strings.Contains(walletTest.name, "dual funder") {
```
wpaulino: Nit: just add backend check in the test case itself?

channeldb/channel.go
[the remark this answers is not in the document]
halseth: > ... we also check if we're the initiator or not.

Yeah, this needs to be done which is why I think `HasFundingTx` is a dangerous name, as there is no guarantee we have it even though it returns true.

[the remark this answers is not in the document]
halseth: I guess it makes sense since it is the _channel type_ that has the funding tx, not the _channel_. I'm okay with keeping the name, but adding to the godoc that it is only the initiator that has the tx strored locally, but I agree with @wpaulino that a more descriptive name could lessen the confusion.

lnwallet/wallet.go
```diff
@@ -992,44 +996,78 @@ func (l *LightningWallet) handleFundingCounterPartySigs(msg *addCounterPartySigs
...
+
+	// Now we can complete the funding transaction by adding their
+	// signatures to their inputs.
+	res.theirFundingInputScripts = msg.theirFundingInputScripts
+	inputScripts := msg.theirFundingInputScripts
+
+	// Only if we have the final funding transaction do we need to verify
+	// the final set of inputs. Otherwise, it may be the case that the
+	// channel was funded via an external wallet.
+	fundingTx := res.fundingTx
```
halseth: can move this inside the `if` since it will be `nil` otherwise

halseth: Can move the below `res.partialState.FundingTxn = fundingTx` inside the if as well.

lnwallet/wallet.go
```diff
@@ -507,6 +528,18 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 		remoteFundingAmt = fundingIntent.RemoteFundingAmt()
+
+	} else {
+		l.intentMtx.Lock()
+		// Alternatively, if we've just received an inbound funding
+		// request that we have a registered shim intent to, then we'll
+		// obtain the backing intent now. In this case, we're doing a
+		// special funding workflow that allows more advanced
+		// constructions such as channel factories to be instantiated.
+		if shimIntent, ok := l.fundingIntents[req.PendingChanID]; ok {
```
halseth: I don't get this. Isn't the point of the canned assembler that you can use it to fetch such `pre-created` intents? Now there are two paths to achieve this: 1) pre-register an intent 2) create a canned assembler that when called will return the pre-created intent.

It should be enough with one of them?

lnwallet/wallet.go
```diff
@@ -1061,16 +1061,16 @@ func (l *LightningWallet) verifyFundingInputs(fundingTx *wire.MsgTx,
...
+				txin.SignatureScript, txin.Witness,
 			)
 			if err != nil {
 				return fmt.Errorf("cannot create script: %v", err)
 			}
 
 			output, err := l.Cfg.ChainIO.GetUtxo(
 				&txin.PreviousOutPoint,
-				pkScript, 0, l.quit,
+				pkScript.Script(), 0, l.quit,
```
halseth: add TODO to fix this for neutrino

Roasbeef:
> However, I still think there are still significant cleanup that can be made, by removing reservation logic from the wallet that is now all handled by the ChanFunder

IMO this is out of scope for this PR, I purposefully kept that logic in place in order to not have the PR be too large. The reservation logic is still the bridge between our internal representatio+provisioning of the various channel parmaters (keys, csv delay, sigs, etc) and how things are manifested on the p2p layer (openchannel, acceptchannel, etc). This PR is only about abstracting away how the funding transaction and channel point are obtained by the wallet, and nothing more.

lnwallet/chanfunding/coin_select.go
```diff
@@ -0,0 +1,216 @@
...
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// ErrInsufficientFunds is a type matching the error interface which is
```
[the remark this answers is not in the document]
Roasbeef: Indeed, it just moves code.

lnwallet/wallet.go
Roasbeef: No we need both. Please see how this is exposed on the RPC interface in the PR that builds upon this one. In the context of your example: #1 is done by the _responder_ to an external funding event, while your #2 is done by the _initiator_ of the external funding event.

lnwallet/wallet.go
Roasbeef: Kept as is since we want to ensure external consistency (the full funding transaction is only available on the reservation struct after all other checks have passed). Only in the second instance is the variable actually accessed as is.

lnwallet/wallet.go
```diff
@@ -709,100 +752,73 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
...
-		txIn.Witness = inputScript.Witness
-		pendingReservation.ourFundingInputScripts = append(
-			pendingReservation.ourFundingInputScripts,
-			inputScript,
+		// Finally, we'll populate the relevant information in our
+		// pendingReservation so the rest of the funding flow can
+		// continue as normal.
+		pendingReservation.fundingTx = fundingTx
+		pendingReservation.partialState.FundingOutpoint = *chanPoint
+		pendingReservation.ourFundingInputScripts = make(
```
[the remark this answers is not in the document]
Roasbeef: It does sign, these fields are left in place as it's how the tests assert that populated the signatures for our inputs. For dual funding, we check that these are even populated which helps catch bugs early, as debugging why a script isn't satisfied (usually invalid signatures) can be rather painful.

lnwallet/wallet.go
```diff
@@ -709,100 +752,73 @@ func (l *LightningWallet) handleContributionMsg(req *addContributionMsg) {
...
-	txsort.InPlaceSort(pendingReservation.fundingTx)
+	// At this point, we can now construct our channel point. Depenindg on
+	// which type of intent we obtained from our chanfunding.Assembler,
+	// we'll carry out a distinct set of steps.
+	switch fundingIntent := pendingReservation.fundingIntent.(type) {
+	case *chanfunding.FullIntent:
+		// Now that we know their public key, we can bind theirs as
+		// well as ours to the funding intent.
+		fundingIntent.BindKeys(
+			&theirContribution.MultiSigKey,
```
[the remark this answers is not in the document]
Roasbeef: Nice catch! Though it doesn't matter in practice as we only use this to construct the funding script, and the function that does so always sorts the keys before inserting them into the script template. Fixed nevertheless.

lnwallet/wallet.go
```diff
@@ -520,26 +555,34 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 
-	if selected != nil {
-		reservation.ourContribution.Inputs = selected.coins
-		reservation.ourContribution.ChangeOutputs = selected.change
+	// At this point, if we have a funding intent, we'll use it to populate
+	// the existing reservation state entries for our coin selection.
+	if fundingIntent != nil {
+		if intent, ok := fundingIntent.(*chanfunding.FullIntent); ok {
+			for _, coin := range intent.InputCoins {
+				reservation.ourContribution.Inputs = append(
```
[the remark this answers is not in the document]
Roasbeef: I left them in place simply to keep the existing test assertions in place. The reservation struct is still the primary bridge between the wallet and the outside world as far as advancing the internal state machine.

lnwallet/wallet.go
```diff
@@ -463,41 +482,57 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 		// Coin selection is done on the basis of sat/kw, so we'll use
 		// the fee rate passed in to perform coin selection.
 		var err error
-		selected, err = l.selectCoinsAndChange(
-			req.FundingFeePerKw, req.LocalFundingAmt, req.MinConfs,
-			req.SubtractFees,
+		fundingReq := &chanfunding.Request{
```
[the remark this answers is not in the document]
Roasbeef: Not sure I follow, we still need to handle the case of an incoming funding request that we already have a shim intent registered for. Also if we don't have any coins to contribute to a funding reservation, we don't need to provision any inputs/outputs or a funding transaction at all, meaning we don't need to call `ProvisionChannel`.

lnwallet/wallet.go
```diff
@@ -463,41 +482,57 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
+			RemoteAmt:    req.RemoteFundingAmt,
+			LocalAmt:     req.LocalFundingAmt,
+			MinConfs:     req.MinConfs,
+			SubtractFees: req.SubtractFees,
+			FeeRate:      req.FundingFeePerKw,
+			ChangeAddr: func() (btcutil.Address, error) {
```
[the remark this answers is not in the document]
Roasbeef: The funder is the one that controls the change address still. This is only used for non-external funding. For the external funding requests, it isn't checked at all (as the funding transaction has already been constructed). This closure lives on the `Request` struct in order to allow a caller to specify a custom change address which may not actually belong to the wallet. As this is a request to control how the funding transaction is constructed, and the change address is a part of the funding transaction, IMO the `Request` struct is an apt place for this parameter to live.

lnwallet/chanfunding/canned_assembler.go
```diff
@@ -0,0 +1,179 @@
...
+			"not constructed")
+	}
+
+	return s.chanPoint, nil
+}
+
+// FundingKeys couples our multi-sig key along with the remote party's key.
+type FundingKeys struct {
+	// LocalKey is our multi-sig key.
+	LocalKey *keychain.KeyDescriptor
```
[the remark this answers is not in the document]
Roasbeef: No, we need this to be a key descriptor so we can properly store it as such in the channel database. The key desc is used to derive the proper key for signing, and also is stored in the SCB so we can re-derive the key from scratch without needing to scan the keyspace.

lnwallet/interface_test.go
Roasbeef: AFAICT, the test doesn't know which backend is being used. The config for `btcwallet` is a private variable so the test can't access that to perform a type assertion, I tried that route before this one.

fundingmanager.go
```diff
@@ -1739,21 +1740,28 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 	// delete it from our set of active reservations.
 	f.deleteReservationCtx(peerKey, pendingChanID)
 
-	// Broadcast the finalized funding transaction to the network.
-	fundingTx := completeChan.FundingTxn
-	fndgLog.Infof("Broadcasting funding tx for ChannelPoint(%v): %v",
-		completeChan.FundingOutpoint, spew.Sdump(fundingTx))
+	// Broadcast the finalized funding transaction to the network, but only
+	// if we actually have the funding transaction.
+	if completeChan.ChanType.HasFundingTx() {
```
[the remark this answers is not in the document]
Roasbeef: Yep.

lnwallet/wallet.go
Roasbeef: It's done here optimistically to immediately clean up any coins for a funding request that hasn't even advanced passed the reservation phase. `failFundingFlow` already does this (in response to another error, or an invalid funding workflow message by the remote peer) since it calls `Cancel` on the reservation object, which will cancel and remove the intent.

lnwallet/wallet.go
Roasbeef: The `initiator` field is used to make the local/remote funding amount attributes consistent. See `lnwallet/wallet.go` for how the attributes/methods are used within the wallet. They're a part of the main intent interface.

lnwallet/wallet.go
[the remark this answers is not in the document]
Roasbeef: Flipped!

AbelLykens:
Since this is merged, it seems #3772 should now be possible. Looking forward to that.

PR #3664, opened 2019-11-01
cnct/test: add outgoing contest resolver test

joostjager:
This PR adds test coverage for the outgoing contest resolver. Currently this resolver is also hit via one of the channel arbitrator tests (`TestChannelArbitratorLocalForceClosePendingHtlc`), but covering it in an isolated unit test has the following benefits:
* Makes it easier to test multiple scenarios for the outgoing contest resolver.
* Because the resolver itself is covered, it allows the channel arbitrator test to be refactored with mock resolvers. This makes it easier to also cover multiple scenarios there. For anchor outputs, refactoring of the channel arbitrator is expected (#3587).

### DIFF
--- a/contractcourt/htlc_outgoing_contest_resolver.go
+++ b/contractcourt/htlc_outgoing_contest_resolver.go
@@ -76,36 +76,6 @@ func (h *htlcOutgoingContestResolver) Resolve() (ContractResolver, error) {
 	default:
 	}
 
-	// We'll check the current height, if the HTLC has already expired,
-	// then we'll morph immediately into a resolver that can sweep the
-	// HTLC.
-	//
-	// TODO(roasbeef): use grace period instead?
-	_, currentHeight, err := h.ChainIO.GetBestBlock()
-	if err != nil {
-		return nil, err
-	}
-
-	// If the current height is >= expiry-1, then a spend will be valid to
-	// be included in the next block, and we can immediately return the
-	// resolver.
-	//
-	// TODO(joostjager): Statement above may not be valid. For CLTV locks,
-	// the expiry value is the last _invalid_ block. The likely reason that
-	// this does not create a problem, is that utxonursery is checking the
-	// expiry again (in the proper way). Same holds for minus one operation
-	// below.
-	//
-	// Source:
-	// https://github.com/btcsuite/btcd/blob/991d32e72fe84d5fbf9c47cd604d793a0cd3a072/blockchain/validate.go#L154
-	if uint32(currentHeight) >= h.htlcResolution.Expiry-1 {
-		log.Infof("%T(%v): HTLC has expired (height=%v, expiry=%v), "+
-			"transforming into timeout resolver", h,
-			h.htlcResolution.ClaimOutpoint, currentHeight,
-			h.htlcResolution.Expiry)
-		return &h.htlcTimeoutResolver, nil
-	}
-
 	// If we reach this point, then we can't fully act yet, so we'll await
 	// either of our signals triggering: the HTLC expires, or we learn of
 	// the preimage.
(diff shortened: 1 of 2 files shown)


contractcourt/htlc_outgoing_contest_resolver.go
```diff
@@ -76,36 +76,6 @@ func (h *htlcOutgoingContestResolver) Resolve() (ContractResolver, error) {
 	default:
 	}
 
-	// We'll check the current height, if the HTLC has already expired,
```
halseth: nice! 👍

contractcourt/htlc_outgoing_contest_resolver_test.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+// TestHtlcOutgoingResolverTimeout tests resolution of an offered htlc that
+// timed out.
+func TestHtlcOutgoingResolverTimeout(t *testing.T) {
+	t.Parallel()
+	defer timeout(t)()
+
+	ctx := newOutgoingResolverTestContext(t)
+	ctx.resolve()
```
halseth: needs comments to explain the steps

contractcourt/htlc_outgoing_contest_resolver_test.go
```diff
@@ -0,0 +1,158 @@
...
+// TestHtlcOutgoingResolverRemoteClaim tests resolution of an offered htlc that
+// is claimed by the remote party.
+func TestHtlcOutgoingResolverRemoteClaim(t *testing.T) {
+	t.Parallel()
+	defer timeout(t)()
+
+	ctx := newOutgoingResolverTestContext(t)
+	ctx.resolve()
+	preimage := lntypes.Preimage{}
+	ctx.notifier.spendChan <- &chainntnfs.SpendDetail{
```
halseth: same, needs comments, otherwise very hard to know what's going on

contractcourt/htlc_outgoing_contest_resolver_test.go
```diff
@@ -0,0 +1,158 @@
...
+		t:          t,
+	}
+}
+
+func (i *outgoingResolverTestContext) resolve() {
+	// Start resolver.
+	i.resolveErr = make(chan error, 1)
+	go func() {
+		var err error
+		i.nextResolver, err = i.resolver.Resolve()
```
halseth: seems like it would be better to define a channel where both the resolver and error could be sent, to avoid this awkward struct field.

contractcourt/htlc_outgoing_contest_resolver.go
```diff
@@ -76,36 +76,6 @@ func (h *htlcOutgoingContestResolver) Resolve() (ContractResolver, error) {
...
-	// then we'll morph immediately into a resolver that can sweep the
-	// HTLC.
-	//
-	// TODO(roasbeef): use grace period instead?
-	_, currentHeight, err := h.ChainIO.GetBestBlock()
```
carlaKC: More a question for my own understanding than review. 
Is the effect of removing this code that we no longer get the current block and check the expiry? And instead go straight into the for/select because that will receive the next block anyway (as per commit message)?

contractcourt/htlc_outgoing_contest_resolver_test.go
```diff
@@ -0,0 +1,158 @@
...
+			TxIn: []*wire.TxIn{
+				{
+					Witness: [][]byte{
+						{0}, {1}, {2}, preimage[:],
+					},
+				},
+			},
+		},
+	}
+	<-ctx.preimageDB.newPreimages
```
carlaKC: It's not immediately apparent to me why we need to block on this? I see the `ChainArbitratorConfig` shares a `preimageDB` with `ctx`, could you add a comment which explains the interaction here so it's clearer why we need to unblock it?

contractcourt/htlc_outgoing_contest_resolver.go
joostjager: Yes, the block notifier immediately sends the current block (even before the next) without delay. So that is effectively the same as `GetBestBlock` call that is removed here.

contractcourt/htlc_outgoing_contest_resolver_test.go
joostjager: added

contractcourt/htlc_outgoing_contest_resolver_test.go
joostjager: yes fair point. added

contractcourt/htlc_outgoing_contest_resolver_test.go
joostjager: comment added. we are asserting the the resolver is saving the preimage that it extracted from the chain tx

contractcourt/htlc_outgoing_contest_resolver_test.go
joostjager: infinitely better, done.

PR #3665, opened 2019-11-01
cnct: add resolver constructors

joostjager:
This PR makes the instantiation of resolvers stricter and thereby opens up the path to also add initialization logic to a resolver.

Spin off of #3644

### DIFF
--- a/contractcourt/briefcase.go
+++ b/contractcourt/briefcase.go
@@ -473,7 +473,6 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
 				if err := timeoutRes.Decode(resReader); err != nil {
 					return err
 				}
-				timeoutRes.AttachResolverKit(resKit)
 
 				res = timeoutRes
 
--- a/contractcourt/briefcase.go
+++ b/contractcourt/briefcase.go
@@ -440,7 +440,7 @@ func (b *boltArbitratorLog) CommitState(s ArbitratorState) error {
 //
 // NOTE: Part of the ContractResolver interface.
 func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, error) {
-	resKit := ResolverKit{
+	resolverCfg := ResolverConfig{
 		ChannelArbitratorConfig: b.cfg,
 		Checkpoint:              b.checkpointContract,
 	}
@@ -516,8 +516,7 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
 				return fmt.Errorf("unknown resolver type: %v", resType)
 			}
 
-			resKit.quit = make(chan struct{})
-			res.AttachResolverKit(resKit)
+			res.AttachConfig(resolverCfg)
 			contracts = append(contracts, res)
 			return nil
 		})
--- a/contractcourt/briefcase.go
+++ b/contractcourt/briefcase.go
@@ -516,7 +516,7 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
 				return fmt.Errorf("unknown resolver type: %v", resType)
 			}
 
-			resKit.Quit = make(chan struct{})
+			resKit.quit = make(chan struct{})
 			res.AttachResolverKit(resKit)
 			contracts = append(contracts, res)
 			return nil
--- a/contractcourt/briefcase.go
+++ b/contractcourt/briefcase.go
@@ -469,54 +469,38 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
 
 			switch resType {
 			case resolverTimeout:
-				timeoutRes := &htlcTimeoutResolver{}
-				if err := timeoutRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = timeoutRes
+				res, err = newTimeoutResolverFromReader(
+					resReader, resolverCfg,
+				)
 
 			case resolverSuccess:
-				successRes := &htlcSuccessResolver{}
-				if err := successRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = successRes
+				res, err = newSuccessResolverFromReader(
+					resReader, resolverCfg,
+				)
 
 			case resolverOutgoingContest:
-				outContestRes := &htlcOutgoingContestResolver{
-					htlcTimeoutResolver: htlcTimeoutResolver{},
-				}
-				if err := outContestRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = outContestRes
+				res, err = newOutgoingContestResolverFromReader(
+					resReader, resolverCfg,
+				)
 
 			case resolverIncomingContest:
-				inContestRes := &htlcIncomingContestResolver{
-					htlcSuccessResolver: htlcSuccessResolver{},
-				}
-				if err := inContestRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = inContestRes
+				res, err = newIncomingContestResolverFromReader(
+					resReader, resolverCfg,
+				)
 
 			case resolverUnilateralSweep:
-				sweepRes := &commitSweepResolver{}
-				if err := sweepRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = sweepRes
+				res, err = newCommitSweepResolverFromReader(
+					resReader, resolverCfg,
+				)
 
 			default:
 				return fmt.Errorf("unknown resolver type: %v", resType)
 			}
 
-			res.AttachConfig(resolverCfg)
+			if err != nil {
+				return err
+			}
+
 			contracts = append(contracts, res)
 			return nil
 		})
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -1733,14 +1733,10 @@ func (c *ChannelArbitrator) prepContractResolutions(
 					continue
 				}
 
-				resKit := newContractResolverKit(resolverCfg)
-				resolver := &htlcSuccessResolver{
-					htlcResolution:      resolution,
-					broadcastHeight:     height,
-					payHash:             htlc.RHash,
-					htlcAmt:             htlc.Amt,
-					contractResolverKit: *resKit,
-				}
+				resolver := newSuccessResolver(
+					resolution, height,
+					htlc.RHash, htlc.Amt, resolverCfg,
+				)
 				htlcResolvers = append(htlcResolvers, resolver)
 			}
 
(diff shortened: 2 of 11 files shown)


contractcourt/contract_resolvers.go
```diff
@@ -46,11 +46,6 @@ type ContractResolver interface {
 	// passed Writer.
 	Encode(w io.Writer) error
 
-	// Decode attempts to decode an encoded ContractResolver from the
-	// passed Reader instance, returning an active ContractResolver
-	// instance.
-	Decode(r io.Reader) error
```
halseth: it is a bit weird to have an `Encode` method, but no `Decode`. I think we should keep it, if you need to make it used you can call it outside the switch statement where the resolver type is checked.

contractcourt/briefcase.go
```diff
@@ -473,7 +473,6 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
 				if err := timeoutRes.Decode(resReader); err != nil {
 					return err
 				}
-				timeoutRes.AttachResolverKit(resKit)
```
halseth: here we can instead call `res.Decode` after the switch.

contractcourt/channel_arbitrator.go
```diff
@@ -1734,12 +1734,10 @@ func (c *ChannelArbitrator) prepContractResolutions(
 				}
 
 				resKit.Quit = make(chan struct{})
-				resolver := &htlcSuccessResolver{
-					htlcResolution:  resolution,
-					broadcastHeight: height,
-					payHash:         htlc.RHash,
-					htlcAmt:         htlc.Amt,
-				}
+				resolver := newSuccessResolver(
```
halseth: why not make `resKit` a param as well?

contractcourt/briefcase.go
```diff
@@ -469,50 +469,35 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
...
 				}
 
-				res = outContestRes
-
 			case resolverIncomingContest:
-				inContestRes := &htlcIncomingContestResolver{
-					htlcSuccessResolver: htlcSuccessResolver{},
-				}
-				if err := inContestRes.Decode(resReader); err != nil {
+				res, err = decodeIncomingContestResolver(resReader)
```
halseth: What do these methods really achieve, that `Decode` doesn't?

contractcourt/briefcase.go
joostjager: that is still possible in this commit, but not anymore when `decode` becomes a constructor

contractcourt/briefcase.go
joostjager: The end result is the same, but it encapsulated the instantiation of the resolver inside the resolver. If we want to follow up with additional initialization of resolvers (for example set up the initial report), it is clear where to do it. Otherwise that init code will need to be called separately from chan arb which carries more risk of making a mistake.

contractcourt/channel_arbitrator.go
joostjager: yes, make it definitely better! :+1:

joostjager: this also exposed some dirtyness around the quit channel. added clean up commit for that too

contractcourt/contract_resolvers.go
joostjager: See comment about encapsulation

joostjager:
Now the resolvers are really ready to get their reports initialized in their (new or decode) constructors

contractcourt/contract_resolvers.go
joostjager: Also, this method was never called on the `ContractResolver` interface. This commit builds

contractcourt/contract_resolvers.go
```diff
@@ -81,8 +70,15 @@ type ResolverKit struct {
...
+}
+
+// contractResolverKit is meant to be used as a mix-in struct to be embedded within a
+// given ContractResolver implementation. It contains all the common items that
+// a resolver requires to carry out its duties.
+type contractResolverKit struct {
+	ResolverKit
 
-	Quit chan struct{}
+	quit chan struct{}
```
halseth: make quit channel on the contract resolver instead?

halseth: or maybe the code duplication gets worse... Just a bit weird with a wrapper just to add a quit channel

contractcourt/channel_arbitrator.go
```diff
@@ -1767,8 +1767,8 @@ func (c *ChannelArbitrator) prepContractResolutions(
 					broadcastHeight: height,
 					htlcIndex:       htlc.HtlcIndex,
 					htlcAmt:         htlc.Amt,
-					ResolverKit:     resKit,
 				}
+				resolver.AttachResolverKit(resKit)
```
halseth: hm, seems like this commit is reverted later, so might as well remove

contractcourt/htlc_success_resolver.go
```diff
@@ -52,7 +52,7 @@ type htlcSuccessResolver struct {
 	// account any fees that may have to be paid if it goes on chain.
 	htlcAmt lnwire.MilliSatoshi
 
-	ResolverKit
+	contractResolverKit
```
halseth: Is the pure `ResolverKit` used anywhere now after this change?

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -34,6 +35,24 @@ type htlcIncomingContestResolver struct {
 	htlcSuccessResolver
 }
 
+// newIncomingContestResolver instantiates a new incoming htlc contest resolver.
+func newIncomingContestResolver(htlcExpiry uint32,
```
halseth: should decode methods just call the constructors also now?

contractcourt/contract_resolvers.go
joostjager: i felt the same, couldn't come up with something better

contractcourt/channel_arbitrator.go
joostjager: removed

contractcourt/contract_resolvers.go
joostjager: maybe this will be useful later if we get another shared field

contractcourt/htlc_incoming_contest_resolver.go
joostjager: hmm idk. it would add boiler plate to first decode everything in local vars. I am already happy that everything is in the resolver file now

contractcourt/htlc_success_resolver.go
joostjager: yes, it is passed as an argument to all constructors

contractcourt/htlc_timeout_resolver.go
```diff
@@ -48,7 +48,24 @@ type htlcTimeoutResolver struct {
...
+	contractResolverKit
+}
+
+// newTimeoutResolver instantiates a new timeout htlc resolver.
+func newTimeoutResolver(res lnwallet.OutgoingHtlcResolution,
+	broadcastHeight uint32, htlcIndex uint64,
+	htlcAmt lnwire.MilliSatoshi, resKit ResolverKit) *htlcTimeoutResolver {
+
+	return &htlcTimeoutResolver{
+		contractResolverKit: contractResolverKit{
```
carlaKC: would it make sense to have a `newContractResolver` func which takes the resolver kit and make the channel there? Nit, but we do have this a repeated a few times.

contractcourt/htlc_outgoing_contest_resolver.go
```diff
@@ -18,6 +20,23 @@ type htlcOutgoingContestResolver struct {
 	htlcTimeoutResolver
 }
 
+// newOutgoingContestResolver instantiates a new outgoing contested htlc
+// resolver.
+func newOutgoingContestResolver(
```
carlaKC: formatting nit (/question), do we usually wrap function params like this?

joostjager: there was an enter too many, removed.

contractcourt/htlc_timeout_resolver.go
joostjager: Good suggestion, reworked commits to do this. Also renamed `ResolverKit` to `ResolverConfig` which is really what it is now

contractcourt/htlc_success_resolver.go
joostjager: has now been renamed to `ResolverConfig` to make its purpose clear

contractcourt/briefcase.go
```diff
@@ -469,54 +469,38 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
...
+				)
 
 			case resolverUnilateralSweep:
-				sweepRes := &commitSweepResolver{}
-				if err := sweepRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = sweepRes
+				res, err = decodeCommitSweepResolver(
```
halseth: I don't really see the point of splitting these into different methods instead of using the interface `Decode` method, now that they all have the same method signature.

joostjager: Renamed to `newResolverFromReader`. Resolver structs aren't dumb data structs that can just have a `Serializable` interfaces. There is more logic required, which is probably also why the `Attach` and `Supplement` method appeared.

contractcourt/briefcase.go
```diff
@@ -469,54 +469,38 @@ func (b *boltArbitratorLog) FetchUnresolvedContracts() ([]ContractResolver, erro
...
 			case resolverOutgoingContest:
-				outContestRes := &htlcOutgoingContestResolver{
-					htlcTimeoutResolver: htlcTimeoutResolver{},
-				}
-				if err := outContestRes.Decode(resReader); err != nil {
-					return err
-				}
-
-				res = outContestRes
+				res, err = newOutgoingContestResolverFromReader(
```
halseth: should be named `newOutgoingContestResolverFromReaderAndResolverConfig`? 😝

contractcourt/commit_sweep_resolver.go
```diff
@@ -298,15 +311,6 @@ func newCommitSweepResolverFromReader(r io.Reader, resCfg ResolverConfig) (
 	return c, nil
 }
 
-// AttachConfig should be called once a resolved is successfully decoded from
```
halseth: nit: delete in the same commit as we remove the last reference to this method.

contractcourt/briefcase.go
joostjager: Not every parameter can be equally important :stuck_out_tongue:

contractcourt/commit_sweep_resolver.go
joostjager: This is the last commit that contains the last ref?

PR #3667, opened 2019-11-02
fundingmanager+routing log touch ups

cfromknecht:


fundingmanager.go
```diff
@@ -812,7 +812,7 @@ func (f *fundingManager) advanceFundingState(channel *channeldb.OpenChannel,
 			// network.
 			// TODO(halseth): could do graph consistency check
 			// here, and re-add the edge if missing.
-			fndgLog.Debugf("ChannlPoint(%v) with chanID=%v not "+
+			fndgLog.Debugf("ChannlPoint(%v) with chan_id=%x not "+
```
halseth: also fix `ChannlPoint`?

cfromknecht: lol i didn't see that, also fixed a bunch of other camelcase logs

cfromknecht:
> It would have been even better if this pr would quickly add a PrefixLogger to funding manager. Then the "ChannlPoint" typo wouldn't need fixing :P

Which channel point would we use as the prefix? Or make a new logger per channel?

joostjager:
> > It would have been even better if this pr would quickly add a PrefixLogger to funding manager. Then the "ChannlPoint" typo wouldn't need fixing :P
> 
> Which channel point would we use as the prefix? Or make a new logger per channel?

Right, good point. It demonstrates how little I know about the funding manager. Not sure if anything is practical, just saw the prefix repeated several times in the file.

cfromknecht:
i also ended up running the entire itest suite and then used 
```
grep -r "[a-f0-9]\{128,132\}" lntest/itest/ | grep -v "reason=[a-f0-9]\{133,\}"
```
to find any other cases of double-hex encodings of 32 and 33 byte values. the prominent ones should be fixed now

cfromknecht:
@joostjager definitely agree, there is a lot to be desired there :)

halseth:
Why do we want to use snake_case? Seems like camelCase is being used overwhelmingly in the fundingmanager, so why not stick to that?

cfromknecht:
@halseth because snake case is most widely used in lnd, so we should converge on that

Issue #3660, opened 2019-11-01
Private channel can't receive despite being notified as active.


roeierez:
### Background

Private channel that is opened by a remote node is not ready for payment and not listed in ListChannels RPC output despite a "CHANNEL_ACTIVE" notification dispatched.

### Your environment

* lnd 0.8.0-beta with neutrino backend running on mobile.

### Steps to reproduce

* Open a private channel from a remote node (B) to a local node (A)
* Track channel activeness on node A via `SubscribeChannelEvents`
* When active:
** Generate invoice - no routing hints exist in the invoice.
** Call GetChanInfo RPC for this specific channel - error returned `[ERR] RPCS: [/lnrpc.Lightning/GetChanInfo]: edge not found`
** The call:
 ```
client.ListChannels(context.Background(), &lnrpc.ListChannelsRequest{		
		ActiveOnly:  true,
	})
```
returns empty list.

### Expected behaviour

Should have routing hints on the invoice so the channel can receive payment, should be listed in `ListChannels` and get no error from `GetChanInfo`

I think the reason is that the `CHANNEL_ACTIVE` event is dispatched immediately after channel reestablishment https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L992
And the `channel_update` message which is required for the channel to be usable only received later.

roeierez:
**Update**: When generating invoices right after channel active event is dispatched the routing hints are not added since the remote node announcement haven't been received yet.
I deduce that from the logs running in trace:
`2019-11-01 15:21:14.901 [INF] IRPC: Skipping channel 6064b72e280933a652e85efae0e16e5e0d0fa3404a85cf69017a16ba11463f3d due to counterparty 031015a7839468a3c266d662d5bb21ea4cea24226936e2864a7ca4f2c3939836e0 being unadvertised`

wpaulino:
> 2019-11-01 15:21:14.901 [INF] IRPC: Skipping channel 6064b72e280933a652e85efae0e16e5e0d0fa3404a85cf69017a16ba11463f3d due to counterparty 031015a7839468a3c266d662d5bb21ea4cea24226936e2864a7ca4f2c3939836e0 being unadvertised

Is the node that the channel is being opened to aware of the other channels of the node opening the channel? Hints are only added for public nodes.

roeierez:
[...]

PR #3672, opened 2019-11-04
Allow user specify routing hints in private invoice.

roeierez:
This PR changes the `AddInvoice/AddHoldInvoice `RPC endpoints to be able to get routing hints and populate them in the resulted invoice.
In fact the `RoutingHints` already existed in both `lnrpc.Invoice` and `lnrpc.AddHoldInvoiceRequest` but were ignored and this change only adds them to the invoice.

This is useful in the following cases:
1. When ones want to expose specific private channels, or to get payment via them.
2. When the default routing hints logic skips some channels due to incomplete graph state (see issue https://github.com/lightningnetwork/lnd/issues/3660). This is specifically important to enable receiving funds on mobile wallets that just bootstrapped with a new channel.

### DIFF
--- a/lnrpc/invoicesrpc/addinvoice.go
+++ b/lnrpc/invoicesrpc/addinvoice.go
@@ -94,6 +94,10 @@ type AddInvoiceData struct {
 	// HodlInvoice signals that this invoice shouldn't be settled
 	// immediately upon receiving the payment.
 	HodlInvoice bool
+
+	// RouteHints are optional route hints that can each be individually used
+	// to assist in reaching the invoice's destination.
+	RouteHints [][]zpay32.HopHint
 }
 
 // AddInvoice attempts to add a new invoice to the invoice database. Any
@@ -246,6 +250,27 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		options = append(options, zpay32.CLTVExpiry(uint64(defaultDelta)))
 	}
 
+	// We make sure that the given invoice routing hints number is within the
+	// valid range
+	if len(invoice.RouteHints) > 20 {
+		return nil, nil, fmt.Errorf("number of routing hints must not exceed " +
+			"maximum of 20")
+	}
+
+	// We continue by populating the requested routing hints indexing their
+	// corresponding channels so we won't duplicate them.
+	forcedHints := make(map[uint64]struct{})
+	for _, h := range invoice.RouteHints {
+		if len(h) == 0 {
+			return nil, nil, fmt.Errorf("number of hop hint within a route must " +
+				"be positive")
+		}
+		options = append(options, zpay32.RouteHint(h))
+
+		// Only this first hop is our direct channel.
+		forcedHints[h[0].ChannelID] = struct{}{}
+	}
+
 	// If we were requested to include routing hints in the invoice, then
 	// we'll fetch all of our available private channels and create routing
 	// hints for them.
@@ -256,11 +281,21 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		}
 
 		if len(openChannels) > 0 {
+			// We filter the channels by excluding the ones that were specified by
+			// the caller and were already added.
+			var filteredChannels []*channeldb.OpenChannel
+			for _, c := range openChannels {
+				if _, ok := forcedHints[c.ShortChanID().ToUint64()]; ok {
+					continue
+				}
+				filteredChannels = append(filteredChannels, c)
+			}
+
 			// We'll restrict the number of individual route hints
 			// to 20 to avoid creating overly large invoices.
-			const numMaxHophints = 20
+			numMaxHophints := 20 - len(forcedHints)
 			hopHints := selectHopHints(
-				amtMSat, cfg, openChannels, numMaxHophints,
+				amtMSat, cfg, filteredChannels, numMaxHophints,
 			)
 
 			options = append(options, hopHints...)
--- a/lntest/itest/lnd_single_hop_invoice_test.go
+++ b/lntest/itest/lnd_single_hop_invoice_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lightningnetwork/lnd/lntest"
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/record"
 )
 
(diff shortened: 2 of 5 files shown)


halseth:
Can be consolidated with #3022? Or is it different?

roeierez:
> Can be consolidated with #3022? Or is it different?

Yes, needs to. I didn't see the other PR when opened this. I Actually think the approach there is better (user needs to only pass channel ids as hops).
I believe what made me walk this path is that I saw `routing_hints` field in `AddHoldInvoiceRequest`, worth removing it as part of https://github.com/lightningnetwork/lnd/pull/3022.

roeierez:
@halseth Looking again at https://github.com/lightningnetwork/lnd/pull/3022 it won't be usable for the use case in this PR. In https://github.com/lightningnetwork/lnd/pull/3022 the private channels passed only serve as filter while at this PR they are given by the user explicitly.

halseth:
in what cases will it be useful to add hints for channels not known by the receiver (and hence the #3022 approach won't work)?

roeierez:
> EDIT: Just read the latest response on #3660, and see that it could be useful in the case where the remote node info is not yet known.

Yes that's the case.

roeierez:
> @roeierez what are your thoughts on limiting the hints used for invoices to the route hints provided by the user? This would cover the use case of #3022 because the user can just provide hints for the channels they want to include.

I am fine with this approach.


> There is a UX consideration of having to construct hints yourself wrt work for the user + potential for error. Generally, it would be nice to be able to check that the user provided hints are ok, but not in the case of #3660. We could potentially address this with a `getHopHint` endpoint which will produce a hop hint for a given channel and perform validation on the channel? Most users can query this endpoint to get hints for the channels which they want to use, and in special use cases you can just roll your own?

I actually like the approach of https://github.com/lightningnetwork/lnd/pull/3022 where the user is only pass channel ids as input instead of RouteHint structures and your arguments explains that well.
I would suggest to go with the https://github.com/lightningnetwork/lnd/pull/3022 with some small changes to solve this use case as well.

I can think of two approaches for such adaptation:
1. If user passes explicit channel ids, then skip validation (public channel) in LND and force populating the route hints.
2. Add additional flag to force populating the route hints.

Since this is a bit "advanced" use case I tend to go with #1.

Crypt-iQ:
I think either method (3022 or 3672) would work. My motivation for opening 3022 was for privacy reasons since creating an invoice populates *all* private channels by default with no option for specificity. As long as that is addressed (which both method would), I'm fine :)

halseth:
@roeierez Do you want to pick this up again? I think we can get this merged addressing the following:

1. If `private == false` and the user specifies route hints, add only these route hints when creating the invoice (this should address the privacy concern from #3022).
2. If `private == true`, add additional route hints for private channels to the invoice.

This should also let the user specify public channels as route hints, which can be useful in certain cases.

We should also update the RPC documentation to reflect the behaviour.

roeierez:
> @roeierez Do you want to pick this up again?

Sounds good. I'll address the feedback.

hsjoberg:
Needs rebase.

Unless I'm missing something, but this PR seems more powerful than #3022 as it allows for fake channels and the Pay to Open use-case explained here: #4018. But perhaps these two features could be combined?

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -92,6 +92,9 @@ type AddInvoiceData struct {
 	// HodlInvoice signals that this invoice shouldn't be settled
 	// immediately upon receiving the payment.
 	HodlInvoice bool
+	// Optional route hints that can each be individually used to assist in
```
halseth, line 95: add newline above

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -236,6 +239,17 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		options = append(options, zpay32.CLTVExpiry(uint64(defaultDelta)))
 	}
 
+	// We start by populating the requested routing hints marking their
+	// channels as used so we won't duplicate them.
+	usedChannels := make(map[uint64]struct{})
+	for i, h := range invoice.RouteHints {
+		if i >= 20 {
+			break
```
halseth, line 247: return error instead?

halseth (changes requested):
A few nits otherwise LGTM 👍

halseth:
Also needs rebase as pointed out by @hsjoberg

Crypt-iQ:
Closed the other one in favor of this one @roeierez

hsjoberg:
Note that this one currently does not support providing route hints via the CLI, only RPC

lnrpc/rpc.proto
```diff
@@ -2659,7 +2659,8 @@ message Invoice {
 
     /*
     Route hints that can each be individually used to assist in reaching the
-    invoice's destination.
+    invoice's destination. The specified routing hints will be included in
+    the invoice.
```
Crypt-iQ, line 2663: Comment changes need to re-generate protos as the comments are replicated in other `lnrpc` files

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -236,6 +239,17 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
...
+		}
+		options = append(options, zpay32.RouteHint(h))
+		usedChannels[h[0].ChannelID] = struct{}{}
```
Crypt-iQ, line 250: I don't think we should index like this - what if `h` is empty? Why is only the first channel added to `usedChannels`?

Crypt-iQ (commented):
I think we'll want `lncli` to be updated with the same capabilities.

Roasbeef:
Is this PR still active? The 0.12 window is closing in a week or so.

halseth:
ping @roeierez

roeierez:
> ping @roeierez

Yes, I am here :). Going to pick this up again.

lnrpc/invoicesrpc/addinvoice.go
roeierez: @Crypt-iQ 
Added validation for h is empty, missed that so thanks for catching it.
Regarding indexing, I am thinking only the first hop is our direct channel, am I missing something?

roeierez:
@halseth rebased and adapted to the code changes that have been introduced since the start of this PR.
Also added an integration test.

lnrpc/rpc.proto
roeierez: Removed the comment change, not needed after all.

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -246,6 +250,27 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
...
+	// corresponding channels so we won't duplicate them.
+	forcedHints := make(map[uint64]struct{})
+	for _, h := range invoice.RouteHints {
+		if len(h) == 0 {
+			return nil, nil, fmt.Errorf("number of hop hint within a route must " +
+				"be positive")
+		}
+		options = append(options, zpay32.RouteHint(h))
+
+		// only this first hop is our direct chanel.
```
halseth, line 270: typo: start sentence with capital O
typo: chanel

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -256,11 +281,19 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 		}
 
 		if len(openChannels) > 0 {
+			// We filter the channels by excluding the ones that were specified by
+			// the caller and were already added.
+			var filteredChannels []*channeldb.OpenChannel
+			for _, c := range openChannels {
+				if _, exists := forcedHints[c.ShortChanID().ToUint64()]; !exists {
```
halseth, line 288: style suggestion: 

```go
if _, ok := forcedHints[c.ShortChanID().ToUint64()]; ok {
   continue
}

filteredChannels = append(filteredChannels, c)
```

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -256,11 +281,19 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
...
+					filteredChannels = append(filteredChannels, c)
+				}
+			}
```
halseth, line 291: nit: newline after closing bracket

lntest/itest/lnd_single_hop_invoice_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"encoding/hex"
 	"time"
 
+	"github.com/lightningnetwork/lnd/lnwire"
```
halseth, line 9: this should be moved down with the other imports

halseth (approved):
Apart from a few nits this LGTM 👯‍♂️

lnrpc/invoicesrpc/addinvoice.go
roeierez: Suggestion accepted.

lnrpc/invoicesrpc/addinvoice.go
roeierez: Fixed

lnrpc/invoicesrpc/addinvoice.go
roeierez: Fixed.

lntest/itest/lnd_single_hop_invoice_test.go
roeierez: Fixed

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -94,6 +94,10 @@ type AddInvoiceData struct {
 	// HodlInvoice signals that this invoice shouldn't be settled
 	// immediately upon receiving the payment.
 	HodlInvoice bool
+
+	// Optional route hints that can each be individually used to assist in
+	// reaching the invoice's destination.
+	RouteHints [][]zpay32.HopHint
```
carlaKC, line 100: nittynit: godocs shoudl start with the var name `RouteHints...`

lntest/itest/lnd_single_hop_invoice_test.go
```diff
@@ -161,6 +162,68 @@ func testSingleHopInvoice(net *lntest.NetworkHarness, t *harnessTest) {
...
+		t.Fatalf("unable to add invoice: %v", err)
+	}
+	payreq, err := net.Bob.DecodePayReq(ctxt, &lnrpc.PayReqString{PayReq: invoiceResp.PaymentRequest})
+	if err != nil {
+		t.Fatalf("failed to decode payment request %v", err)
+	}
+	if len(payreq.RouteHints) != 1 {
+		t.Fatalf("expected one routing hint")
+	}
+	routingHint := payreq.RouteHints[0]
```
carlaKC, line 201: could we use require.Equal(payreq.RouteHints, hints)?

carlaKC (approved):
Great to get this in!

lntest/itest/lnd_single_hop_invoice_test.go
roeierez: @carlaKC  would love to but comparing protobuf generated structures by value in't really working since they contain "hidden" generated fields which are assigned with different values for different instances (such as `XXX_sizecache`)

hsjoberg:
utACK @ 4463a4ce178301611751c5439ced44b446410ac5, although I've tested previous iterations.

lntest/itest/lnd_single_hop_invoice_test.go
carlaKC: ah sneaky, no worries then!

Issue #3675, opened 2019-11-05
Channel transitions back to pending after it is opened.


roeierez:
### Background

When a channel is opened from B to A and then a payment is received on A the channel is transitioned back to "pending" state while only restart brings it back to "open" state.

### Your environment

*lnd 0.8.0 using neutrino.

### Steps to reproduce

1. Start a local LND node, let a remote node open a channel to the local node.
2. When the channel is confirmed on-chain and marked open in the local node, create an invoice and receive a payment.

### Expected behaviour

Channel should stay in "open" state.

### Actual behaviour

Channel is transitioned to "pending" state as can be seen by `lncli pendingchannels` & `lncli listchannels`

**Analysis**:
The problem resides in updating old state (pending) to the database when updating the commitment transaction as a result of a payment.
When the channel is confirmed on-chain the fundingManager updates correctly the state to `IsPending=false` which is committed to the db here: https://github.com/lightningnetwork/lnd/blob/4592f870334d80d705d79febfac69e07ef653106/fundingmanager.go#L2059
A few lines later it reports back to the switch to load the new short channel id here: ReportShortChanID
The link indeed refreshes the short channel id: https://github.com/lightningnetwork/lnd/blob/cc159decf5ab967af4168ddb977d2a324bcc4b6c/channeldb/channel.go#L599 and transition the link to live link but the `IsPending` of the internal channel is not refreshed and stays `true`.
Later on when a payment is received the commitment is updated: https://github.com/lightningnetwork/lnd/blob/7ecb7136cfb1a62e6367a0cc1e4962c00c58477f/htlcswitch/link.go#L1774 saving the channel with the old state to the db making it pending again.

It looks like that at the time we refresh the short channel id (due to a funding tx confirmation) we should refresh the `IsPending` value as well.

Crypt-iQ:
With the steps you described, the issue doesn't happen.  Only when you restart the local (or remote) node before confirmation.  Then the link is added to the switch and has an old channel state.

roeierez:
> With the steps you described, the issue doesn't happen. Only when you restart the local (or remote) node before confirmation. Then the link is added to the switch and has an old channel state.

@Crypt-iQ The channel that is opened should be configured to require 1 confirmations. With this setting I consistently reproduce without restart.

Crypt-iQ:
[...]

PR #3676, opened 2019-11-05
channeldb: refresh pending state.

roeierez:
This PR fixes an issue where after channel is moved to "open" state it transitioned back to "pending" due to old state committed to the db.
The issue is well explained here: https://github.com/lightningnetwork/lnd/issues/3675
The idea is that when we refresh the short channel ID, it probably happened due to on-chain confirmation (re-org maybe) and in that case it makes sense to refresh the `IsPending` value as well to cover the case where the channel has just been confirmed and by that ensure we are well synced with the new state.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -594,13 +594,15 @@ func (c *OpenChannel) hasChanStatus(status ChannelStatus) bool {
 	return c.chanStatus&status == status
 }
 
-// RefreshShortChanID updates the in-memory short channel ID using the latest
+// RefreshShortChanID updates the in-memory channel state using the latest
 // value observed on disk.
+// TODO: the name of this function should be changed to reflect the fact that
+// it is not only refreshing the short channel id but all the channel state.
+// maybe Refresh/Reload?
 func (c *OpenChannel) RefreshShortChanID() error {
 	c.Lock()
 	defer c.Unlock()
 
-	var sid lnwire.ShortChannelID
 	err := c.Db.View(func(tx *bbolt.Tx) error {
 		chanBucket, err := fetchChanBucket(
 			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
@@ -609,21 +611,16 @@ func (c *OpenChannel) RefreshShortChanID() error {
 			return err
 		}
 
-		channel, err := fetchOpenChannel(chanBucket, &c.FundingOutpoint)
-		if err != nil {
-			return err
+		// populating the in-memory channel with the info fetched from disk.
+		if err := fetchChanInfo(chanBucket, c); err != nil {
+			return fmt.Errorf("unable to fetch chan info: %v", err)
 		}
-
-		sid = channel.ShortChannelID
 
 		return nil
 	})
 	if err != nil {
 		return err
 	}
-
-	c.ShortChannelID = sid
-	c.Packager = NewChannelPackager(sid)
 
 	return nil
 }
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -1219,6 +1219,90 @@ func testUnconfirmedChannelFunding(net *lntest.NetworkHarness, t *harnessTest) {
 	closeChannelAndAssert(ctxt, t, net, carol, chanPoint, false)
 }
 
+// testPaymentFollowingChannelOpen tests that the channel transition from
+// 'pending' to 'open' state does not cause any inconsistencies within other
+// subsystems trying to udpate the channel state in the db. We follow
+// this transition with a payment that updates the commitment state and
+// verify that the pending state is up to date.
+func testPaymentFollowingChannelOpen(net *lntest.NetworkHarness, t *harnessTest) {
+	ctxb := context.Background()
+
+	// We first establish a channel between Alice and Bob.
+	const paymentAmt = btcutil.Amount(100)
+	ctxt, _ := context.WithTimeout(ctxb, channelOpenTimeout)
+	channelCapacity := btcutil.Amount(paymentAmt * 1000)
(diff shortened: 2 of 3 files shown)


Crypt-iQ:
Looks good to me

roeierez:
@Roasbeef I changed a bit the `RefreshShortChannelID` to be more robust in my opinion. It now reloads the channel info from the db and no longer select specific fields. I think it will prevent similar bugs in the future.
I also updated an existing test to check also my changes.

roeierez:
Also Added integration test.

channeldb/channel.go
```diff
@@ -589,36 +589,30 @@ func (c *OpenChannel) hasChanStatus(status ChannelStatus) bool {
...
 			tx, c.IdentityPub, &c.FundingOutpoint, c.ChainHash,
 		)
 		if err != nil {
 			return err
 		}
-
-		channel, err := fetchOpenChannel(chanBucket, &c.FundingOutpoint)
-		if err != nil {
-			return err
+		if err := fetchChanInfo(chanBucket, c); err != nil {
```
bhandras: Althought I have no familiarity with this part of the code yet, I noticed that this change is way broader than what the function name implies. Users of `RefreshShortChanID` function would expect that it only refreshes `c.ShortChannelID` and `c.IsPending` as documented, yet many other members are changed as well.

roeierez: @bhandras  You are right, I was asked by the lightning lab team not to change the function name as major refactoring is scheduled for this part.

bhandras: Could you please add a TODO to the doc so that it's clear that this is intentional?

Roasbeef: >  I was asked by the lightning lab team not to change the function name as major refactoring is scheduled for this part.

The prior diff for this PR was more minimal: it only also refreshed the short channel ID and pending state of the channel. Why the departure?

channeldb/channel.go
```diff
@@ -596,34 +596,23 @@ func (c *OpenChannel) RefreshShortChanID() error {
...
 		}
-
-		channel, err := fetchOpenChannel(chanBucket, &c.FundingOutpoint)
-		if err != nil {
-			return err
-		}
-
-		sid = channel.ShortChannelID
-		isPending = channel.IsPending
+		if err := fetchChanInfo(chanBucket, c); err != nil {
```
halseth: this should have a comment, and an updated godoc on the method. Also looks like it can be squashed with the previous commit.

channeldb/channel.go
```diff
@@ -605,7 +605,7 @@ func (c *OpenChannel) RefreshShortChanID() error {
 		}
 		if err := fetchChanInfo(chanBucket, c); err != nil {
 			return fmt.Errorf("unable to fetch chan info: %v", err)
-		}			
+		}
```
halseth: squash with prev commit

lntest/itest/lnd_test.go
```diff
@@ -1220,6 +1220,92 @@ func testUnconfirmedChannelFunding(net *lntest.NetworkHarness, t *harnessTest) {
...
+// testPaymentFollowingChannelOpen tests that the channel transition from
+// 'pending' to 'open' state does not cause any inconsistencies within other
+// subsystems trying to udpate the channel state in the db. We follow
+// this transition with a payment that updates the commitment state and
+// verify that the pending state is up to date.
+func testPaymentFollowingChannelOpen(net *lntest.NetworkHarness, t *harnessTest) {
+	ctxb := context.Background()
+
+	const (
+		paymentAmt = 100
```
halseth: nit: do `const paymentAmt = btcutil.Amount(100)`

lntest/itest/lnd_test.go
```diff
@@ -1220,6 +1220,92 @@ func testUnconfirmedChannelFunding(net *lntest.NetworkHarness, t *harnessTest) {
...
+	bobPayReqs, _, _, err := createPayReqs(
+		net.Bob, paymentAmt, 1,
+	)
+	if err != nil {
+		t.Fatalf("unable to create pay reqs: %v", err)
+	}
+
+	// Send payment to Bob so there a chanel update to disk will be executed.
+	ctxt, _ = context.WithTimeout(ctxb, defaultTimeout)
+	_, err = net.Alice.SendPaymentSync(ctxt, &lnrpc.SendRequest{PaymentRequest: bobPayReqs[0]})
```
halseth: nit: long line

channeldb/channel.go
roeierez: > The prior diff for this PR was more minimal: it only also refreshed the short channel ID and pending state of the channel. Why the departure?

Although not minimal as the prior diff I think it is a more correct approach then selecting specific fields.
I also think it is not risky as this exact loading logic is executed on startup anyway and it seems ok to consider the db as the single source of truth regarding the channel state.

lntest/itest/lnd_test.go
roeierez: fixed.

lntest/itest/lnd_test.go
roeierez: fixed

channeldb/channel.go
roeierez: squashed.

channeldb/channel.go
roeierez: fixed.

channeldb/channel.go
roeierez: > Could you please add a TODO to the doc so that it's clear that this is intentional?

Added.

bhandras: I'm fine with this after adding the comment. Not entirely sure if the refactor is affected though.

Roasbeef:
Merged as 8b3dd9415ef6b5f83505068231557d275b65db53, tacked on some code style fixes in an extra commit.

PR #3679, opened 2019-11-06
bolt 11 payment addr + single shot mpp for SendPayment

cfromknecht:
In this PR, we make single-shot MPP payments the default payment type in lnd, and marks a major milestone in our long series of PRs preparing us for multi-path payments. 

## Notable changes

 - Invoices generated by lnd will:
   - Advertise optional support for TLV onions (`var_onion_optin`) and payment address (`option_payment_addr`).
   - Include a random 32-byte payment address.
 - Paying such an invoice via `Lightning.SendPayment` or `RouterRPC.SendPayment` will:
    - allow the router to determine support for MPP records using threaded feature vector state.
    - include the new `MPP` record in the final hop's payload if both the above options are supported, which includes the payment address from the invoice.
 - Fixes a bug that would prevent us from not being able to send custom records to unadvertised nodes

By upgrading to this payment type, lnd senders and receivers are offered increased protection from payment sniping and against probing attacks on the receiver. In the future both features will be made mandatory, entirely disabling support for the legacy payment style.

Depends on:
 - [x] #3685 
 - [x] #3693 
 - [x] #3802
 - [x] #3825 
 - [x] #3834


zpay32/invoice.go
```diff
@@ -118,6 +123,10 @@ type Invoice struct {
 	// invoice.
 	PaymentHash *[32]byte
 
+	// PaymentAddr is the payment address to be used by payments to prevent
+	// probing of the destination.
+	PaymentAddr *[32]byte
```
joostjager: In another PR, I experimented with a strong type for this. What do you think of that?

zpay32/invoice.go
```diff
@@ -715,12 +732,12 @@ func parseFieldDataLength(data []byte) (uint16, error) {
 	return uint16(data[0])<<5 | uint16(data[1]), nil
 }
 
-// parsePaymentHash converts a 256-bit payment hash (encoded in base32)
-// to *[32]byte.
-func parsePaymentHash(data []byte) (*[32]byte, error) {
+// parse32Bytes converts a 256-bit value (encoded in base32) to *[32]byte. This
+// can be used for payment hashes, description hashes, payment addresses, etc.
+func parse32Bytes(data []byte) (*[32]byte, error) {
```
joostjager: Refactor could be separated from adding the new field

zpay32/invoice.go
```diff
@@ -87,7 +87,10 @@ const (
 var (
 	// InvoiceFeatures holds the set of all known feature bits that are
 	// exposed as BOLT 11 features.
-	InvoiceFeatures = map[lnwire.FeatureBit]string{}
+	InvoiceFeatures = map[lnwire.FeatureBit]string{
+		lnwire.TLVOnionPayloadRequired: "tlv-onion",
+		lnwire.TLVOnionPayloadOptional: "tlv-onion",
+	}
```
joostjager: `tlv-onion` is implictly signaling mpp support right?

routing/pathfind.go
```diff
@@ -73,17 +73,26 @@ type edgePolicyWithSource struct {
...
 
+// finalHopParams encapsulates various parameters for route construction that
+// apply to the final hop in a route. These features include basic payment data
+// such as amounts and cltvs, as well as more complex features like destination
+// TLV records.
+type finalHopParams struct {
+	amt       lnwire.MilliSatoshi
+	cltvDelta uint16
+	records   []tlv.Record
+}
```
joostjager: nice

routing/pathfind.go
```diff
@@ -115,18 +115,34 @@ func newRoute(sourceVertex route.Vertex,
...
+		// in the route. The base case is the final hop which includes
+		// their amount and timelocks. These values will accumulate
+		// contributions from the preceding hops back to the sender as
+		// we compute the route in reverse.
+		var (
+			amtToForward     lnwire.MilliSatoshi
+			fee              lnwire.MilliSatoshi
+			outgoingTimeLock uint32
+		)
+		if i == len(pathEdges)-1 {
```
joostjager: also nice

routing/pathfind.go
```diff
@@ -142,6 +156,17 @@ func newRoute(sourceVertex route.Vertex,
...
+			// If we have a feature vector for the final hop, we'll
+			// determine if they support TLV onion payloads by
+			// checking if a var_onion_optin feature is set. If we
+			// already know the hop supports TLV from their node
+			// announcement, we can skip this step.
+			if !tlvPayload && finalHop.features != nil {
+				tlvPayload = finalHop.features.HasFeature(
+					lnwire.TLVOnionPayloadOptional,
+				)
+			}
```
joostjager: What if the graph says yes, but the invoice says no?

lnrpc/routerrpc/router_backend.go
```diff
@@ -620,6 +620,7 @@ func (r *RouterBackend) extractIntentFromSendRequest(
 		payIntent.RouteHints = append(
 			payIntent.RouteHints, payReq.RouteHints...,
 		)
+		payIntent.InvoiceFeatures = payReq.Features
```
joostjager: Isn't there an intermediate step before this PR where we still only offer mpp through sendtoroute, but where queryroutes is updated so that it decides whether to mpp or not (probably via an rpc field)? 

Somewhat related: What is the state of payment tracking when the payment is fired through one (or multiple in the case of manual mpp) sendtoroute calls?

joostjager: If not using a pay req, can mpp be specified in a separate field?

routing/pathfind.go
```diff
@@ -167,6 +170,16 @@ func newRoute(sourceVertex route.Vertex,
 					lnwire.TLVOnionPayloadOptional,
 				)
 			}
+
+			// If the receiver supports TLV onion payloads and we
+			// have a payment address from the invoice, we'll go
```
joostjager: If the payment address signals mpp, what signals moon amp then in the invoice?

rpcserver.go
```diff
@@ -3519,6 +3519,11 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 		defaultDelta = cfg.Litecoin.TimeLockDelta
 	}
 
+	invoiceFeatures := lnwire.NewFeatureVector(
+		lnwire.NewRawFeatureVector(lnwire.TLVOnionPayloadOptional),
+		zpay32.InvoiceFeatures,
+	)
```
joostjager: Set this in `AddInvoice` to prevent duplication?

zpay32/invoice.go
joostjager: Ok, found out that `PaymentAddress` is the actual signal

cfromknecht: for now, yes. i think there's still some spec discussion as to whether we want a feature bit. i think we will end up adding one so that the receiver can require the parser to understand payment addresses, and fail during parsing rather than just getting a payment failure

routing/pathfind.go
cfromknecht: good point, i think you're right, the invoice should always override the graph.

routing/pathfind.go
cfromknecht: related to my other comment re: feature bits, but we will likely have another feature that signal moon amp

lnrpc/routerrpc/router_backend.go
cfromknecht: > Isn't there an intermediate step before this PR where we still only offer mpp through sendtoroute, but where queryroutes is updated so that it decides whether to mpp or not (probably via an rpc field)?

for the moment i'm treating the sendpayment and sendtoroute paths as orthogonal, as the purpose of the PR is have the negotiation of mpp fields be automatic so regular payments can take advantage of the security benefits of mpp. i see pr more as step one towards deprecating legacy payments altogether

i can open another pr that adds this parameter to query routes as well

zpay32/invoice.go
cfromknecht: hmm if we go with a strong type it would hinder the consolidation of logic in this pr. my thought was that a type here doesn't offer that much as it's just a static byte array with no hashes/preimages, but it could be cast into a proper type elsewhere in the codebase if necessary. thoughts?

lnrpc/routerrpc/router_backend.go
cfromknecht: yes, tho i was thinking of leaving that to a different pr. my main goal here is to add safety to the 99% of use cases

zpay32/invoice.go
joostjager: Maybe you could do `([32]byte, error)` as the return type of `parse32Bytes` and directly cast it into the `Invoice` field? I am fine either way, but noticed that I do appreciate the stricter types.

lnrpc/routerrpc/router_backend.go
joostjager: I thought that maybe the queryroutes-only scope would be a subset of this pr. Slowing building up from the totally manual amp case.

>the payment tracking should be complete for single shards in #3499

I realize I asked the wrong question. The tracking is there, but it is not possible to get the raw failure message with `TrackPayment`. It is only possible to obtain that if you keep the sync call to `SendToRoute` open. But if we start exposing the routes and raw failure codes on `TrackPayment`, that is sorted.

joostjager: I am not sure if wallets actually pass in the payment request or decode it first and pass in separate components. I don't know how much extra code it is to make it consistent? I think it should be a single `bool` field, or is there more that is required? We also have this gap possibly for signaling the tlv capability of the recipient.

joostjager: Another way to learn the flat features?

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -371,6 +371,22 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
...
+	options = append(options, zpay32.Features(invoiceFeatures))
+
+	// Generate and set a random payment address for this invoice. If the
+	// sender understands payment addresses, this can be used to avoid
+	// intermediaries probing the receiver.
+	var paymentAddr [32]byte
+	if _, err := rand.Read(paymentAddr[:]); err != nil {
+		return nil, nil, err
+	}
+	options = append(options, zpay32.PaymentAddr(&paymentAddr))
```
joostjager: I think we also need to set it on `channeldb.Invoice` and serialize separately.

lnrpc/routerrpc/router_backend.go
cfromknecht: > I think it should be a single bool field, or is there more that is required?

there are three cases for feature bits: no support, optional, or required. i think there is also a larger discussion to be had about what level of control we offer over feature bits. do we validate and sanity check for invalid or unsafe combinations? so it would be a little more than a bool at the rpc level to do it properly. 

> We also have this gap possibly for signaling the tlv capability of the recipient.

what do you mean?

joostjager: Suppose your graph is not up to date and you want to sendpayment to a node that requires tlv. 

Ideally it should be possible to signal all the same feature bits in a dedicated sendpayment field that we also have in the invoice.

lnrpc/invoicesrpc/addinvoice.go
cfromknecht: done

lnrpc/routerrpc/router_backend.go
cfromknecht: i see the use case, and agree that it both _is_ possible and _should_ be feature we support. my point tho is that i think there is some design questions as to what that interface might look like via rpc that would make it out of scope for this pr which is just meant to handle the most widely used method of sending payments, in which the negotiation is done via invoices

zpay32/invoice.go
cfromknecht: removed `InvoiceFeatures` entirely since flat features was merged

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -370,6 +370,15 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
...
 	options = append(options, zpay32.Features(invoiceFeatures))
 
+	// Generate and set a random payment address for this invoice. If the
+	// sender understands payment addresses, this can be used to avoid
+	// intermediaries probing the receiver.
+	var paymentAddr [32]byte
+	if _, err := rand.Read(paymentAddr[:]); err != nil {
+		return nil, nil, err
+	}
+	options = append(options, zpay32.PaymentAddr(&paymentAddr))
```
Roasbeef: Don't we also need to write this to disk, so we can compare it against what we receive from the extended HTLC?

routing/pathfind.go
Roasbeef: > The graph is telling me noooo, but the invoice, the invoice...is telling me...to deliver a single hop mpp payment with the adequate set of records contained in the TLV payload

lnrpc/invoicesrpc/addinvoice.go
Roasbeef: Scratch that, see it below in the expanded diff.

rpcserver.go
```diff
@@ -4599,6 +4599,20 @@ func (r *rpcServer) DecodePayReq(ctx context.Context,
...
+	// for the rpc.
+	invFeatures := payReq.Features.Features()
+	features := make([]*lnrpc.Feature, 0, len(invFeatures))
+	for bit := range invFeatures {
+		name := payReq.Features.Name(bit)
+		features = append(features, &lnrpc.Feature{
+			Bit:        uint32(bit),
+			Name:       name,
+			IsRequired: bit.IsRequired(),
+			IsUnknown:  name == "unknown",
```
halseth: Better to use `IsKnown` here?

feature/deps.go
```diff
@@ -0,0 +1,109 @@
...
+// ValidateDeps asserts that a feature vector sets all features and their
+// transitive dependencies properly. It assumes that the dependencies between
+// optional and required features are identical, e.g. if a feature is required
+// but its depdency is optional, that is sufficient.
+func ValidateDeps(fv *lnwire.FeatureVector) error {
+	features := fv.Features()
+	checked := initChecked(features)
+
+	return validateDeps(features, checked)
+}
```
joostjager: Can't we just do two if statements for now instead of this more advanced recursive bit flipping solution?

zpay32/invoice.go
```diff
@@ -606,6 +607,11 @@ func validateInvoice(invoice *Invoice) error {
 		return fmt.Errorf("missing feature vector")
 	}
 
+	// Validate any transitive feature depdepencies.
+	if err := feature.ValidateDeps(invoice.Features); err != nil {
+		return err
+	}
```
joostjager: Do we want to be so strict when parsing payment requests? Probably better to be strict, to force implementations to implement it right.

routing/pathfind.go
```diff
@@ -90,17 +90,26 @@ type edgePolicyWithSource struct {
...
 }
 
+// finalHopParams encapsulates various parameters for route construction that
+// apply to the final hop in a route. These features include basic payment data
+// such as amounts and cltvs, as well as more complex features like destination
+// TLV records.
+type finalHopParams struct {
+	amt       lnwire.MilliSatoshi
+	cltvDelta uint16
+	records   []tlv.Record
```
joostjager: I am converting this into `hop.CustomRecordSet` in https://github.com/lightningnetwork/lnd/pull/3819. But will rebase if this pr goes first.

routing/pathfind.go
```diff
@@ -141,7 +143,40 @@ func newRoute(sourceVertex route.Vertex,
...
+		supports := func(fv *lnwire.FeatureVector,
+			bit lnwire.FeatureBit) bool {
+
+			if fv == nil {
+				log.Debugf("Node %v has no features",
+					edge.Node.PubKeyBytes)
+				return false
+			}
+
+			err := feature.ValidateDep(fv, bit)
```
joostjager: Is it still necessary to validate at this point? I had the impression we were now checking all these features 'at the gate'?

routing/pathfind.go
```diff
@@ -159,6 +194,16 @@ func newRoute(sourceVertex route.Vertex,
...
+
+			// If we have a feature vector for the final hop, we'll
+			// determine if they support TLV onion payloads by
+			// checking if a var_onion_optin feature is set.
+			// Regardless of what the graph suggests, if the invoice
+			// has feature bits we will assume those override
+			// anything on a node announcement.
+			tlvPayload = supports(
+				finalHop.features, lnwire.TLVOnionPayloadOptional,
+			)
```
joostjager: Separation of concerns doesn't feel completely right here. Is `pathEdges []*channeldb.ChannelEdgePolicy` still the right thing to pass into this function? A different structure that contains exactly what is needed could maybe remove the need to do this interpretation here.

routing/pathfind.go
```diff
@@ -204,6 +207,19 @@ func newRoute(sourceVertex route.Vertex,
...
+			// ahead and include an MPP record to gain the
+			// additional protections offered by MPP.
+			supportsPayAddr := supports(
+				finalHop.features, lnwire.PaymentAddrOptional,
+			)
+			if supportsPayAddr && finalHop.paymentAddr != nil {
+				mpp = record.NewMPP(
+					finalHop.amt, *finalHop.paymentAddr,
+				)
+			}
```
joostjager: Yes, we may want to rethink what happens where. If path finding becomes payload size aware, it will also need to do something similar.

lnwire/features.go
```diff
@@ -114,6 +134,10 @@ var Features = map[FeatureBit]string{
 	TLVOnionPayloadOptional:       "tlv-onion",
 	StaticRemoteKeyOptional:       "static-remote-key",
 	StaticRemoteKeyRequired:       "static-remote-key",
+	PaymentAddrOptional:           "payment-addr",
+	PaymentAddrRequired:           "payment-addr",
+	MPPOptional:                   "multi-path-payments",
+	MPPRequired:                   "multi-path-payments",
```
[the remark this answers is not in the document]
cfromknecht: atm `HasFeature` relies on them being the same: https://github.com/lightningnetwork/lnd/blob/master/lnwire/features.go#L388 doesn't necessarily have to be this way tho

routing/pathfind.go
cfromknecht: currently only the invoice feature bits are checked, but not the ones on the node announcement. we could move to only do this validation there instead of for the final hop as well

zpay32/invoice.go
cfromknecht: yes i think it's also good to catch this before doing any sort of path finding

routing/pathfind.go
cfromknecht: ideally we'd get #3819 in first, then i can rebase over it

routing/pathfind.go
cfromknecht: indeed i think there is a more minimal struct that would suffice, ideally tied to our final design for the router cache

routing/pathfind.go
```diff
@@ -502,29 +507,41 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
 			return nil, err
 		}
 
 		targetNode, err := g.graph.FetchLightningNode(targetKey)
-		if err != nil {
+		switch {
+
+		// If the node exists and has features, use them directly.
+		case err == nil && targetNode.Features != nil:
+			features = targetNode.Features
```
joostjager: Also for this, I thought this was done already in `finalHopParameters`?

routing/pathfind.go
```diff
@@ -502,29 +507,41 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+		// node announcement, populate a blank feature vector.
+		default:
+			features = lnwire.EmptyFeatureVector()
 		}
 	}
 
+	// If the caller needs to send custom records, check that our
+	// destination feature vector supports TLV.
+	if len(r.DestCustomRecords) > 0 &&
+		!features.HasFeature(lnwire.TLVOnionPayloadOptional) {
```
joostjager: Looks familiar from `finalHopParameters` too

routing/payment_session.go
```diff
@@ -98,6 +98,7 @@ func (p *paymentSession) RequestRoute(payment *LightningPayment,
 		LastHop:           payment.LastHop,
 		CltvLimit:         cltvLimit,
 		DestCustomRecords: payment.DestCustomRecords,
+		DestFeatures:      payment.InvoiceFeatures,
```
joostjager: Shouldn't we just pass in the `finalHopParameters` now instead?

lnrpc/routerrpc/router.proto
```diff
@@ -105,6 +105,13 @@ message SendPaymentRequest {
 
     /// If set, circular payments to self are permitted.
     bool allow_self_payment = 15;
+
+    /**
+    Features assumed to be supported by the final node. All transitive feature
+    depdencies must also be set properly. For a given feature bit pair, either
+    optional or remote may be set, but not both.
+    */
+    repeated uint32 dest_features = 16;
```
joostjager: Nice, the plan is unfolding

lnrpc/routerrpc/router.proto
```diff
@@ -105,6 +105,13 @@ message SendPaymentRequest {
 
     /// If set, circular payments to self are permitted.
     bool allow_self_payment = 15;
+
+    /**
+    Features assumed to be supported by the final node. All transitive feature
+    depdencies must also be set properly. For a given feature bit pair, either
+    optional or remote may be set, but not both.
```
joostjager: Is it possible to define an enum in the proto for the feature bits, to prevent mistakes?

lnrpc/routerrpc/router_backend.go
```diff
@@ -603,6 +603,7 @@ func (r *RouterBackend) extractIntentFromSendRequest(
 			payIntent.RouteHints, payReq.RouteHints...,
 		)
 		payIntent.InvoiceFeatures = payReq.Features
+		payIntent.PaymentAddr = payReq.PaymentAddr
 	} else {
 		// Otherwise, If the payment request field was not specified
```
joostjager: We need to take `payment_addr` from the rpc request. Otherwise we can't do key send in combination with single shot mpp as the default payment method.

lnrpc/routerrpc/router.proto
joostjager: I am thinking, maybe we also need to introduce these fields on the main payment rpc. Otherwise `lncli` in the keysend pr will need to switch to `routerrpc` already, to be able to pass in the payment addr.

feature/deps.go
```diff
@@ -56,6 +56,17 @@ func ValidateDeps(fv *lnwire.FeatureVector) error {
...
 
+// ValidateDep asserts that a feature vector sets all features and their
+// transitive dependencies properly for a specific feature bit. It assumes that
+// the dependencies between optional and required features are identical, e.g.
+// if a feature is required but its depdency is optional, that is sufficient.
+func ValidateDep(fv *lnwire.FeatureVector, bit lnwire.FeatureBit) error {
+	features := featureSet{bit: {}}
+	checked := initChecked(fv.Features())
+
+	return validateDeps(features, checked)
```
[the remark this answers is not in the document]
cfromknecht: i looked into this, but it's not quite as clean imo (more arguments being passed in). i think it's clearer that `initChecked` happens only once before entering the recursion.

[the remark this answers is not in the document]
joostjager: Ok, I forgot about the recursion.

routing/pathfind.go
```diff
@@ -141,8 +162,51 @@ func newRoute(sourceVertex route.Vertex,
...
+			tlvPayload       bool
 		)
+
+		// Define a closure that determines whether or not the edge
+		// supports a particular feature. This will be used to test
+		// intermediate hops for support of a particular feature.
+		nodeSupports := func(bit lnwire.FeatureBit) bool {
+			// If the node has no feature vector, we can assume it
+			// doesn't support the feature.
+			if edge.Node.Features == nil {
```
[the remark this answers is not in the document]
cfromknecht: atm it's possible for them to be nil, but i will fix this by always setting non-nil feature bits when reading from the db

lnrpc/routerrpc/router_backend.go
cfromknecht: we don't need to use mpp for keysend because intermediaries can't snipe them (since the preimage is in the onion). we could if we want, but idt it adds anything security wise

routing/pathfind.go
cfromknecht: atm that's done when creating a route, this is before pathfinding is attempted

routing/pathfind.go
```diff
@@ -502,29 +507,41 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		defer tx.Rollback()
 	}
 
-	if len(r.DestCustomRecords) > 0 {
-		// Check if the target has TLV enabled
-
+	// If no destination features are provided, we will load what features
+	// we have for the target node from our graph.
+	features := r.DestFeatures
+	if features == nil {
```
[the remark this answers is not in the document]
cfromknecht: i think it makes sense to keep this one in particular, otherwise we will have no fallback option. how can we distinguish between an (intentionally) empty feature vector and not having one at all? similar to my comment below, this check happens before starting pathfinding, so we haven't yet used `finalHopParams`

routing/pathfind.go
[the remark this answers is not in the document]
joostjager: sounds good, get rid of multiple `nil` checks and replace it by one master check

routing/pathfind.go
joostjager: Yes, I mean instantiating `finalHopParameters` before pathfinding already, so that this logic doesnt need to be executed twice.

Or otherwise return the path from pathfinding so that it includes everything `newRoute` needs. Something doesn't feel right in the split between pathfinding and `newRoute`.

lnrpc/routerrpc/router_backend.go
joostjager: But how do we communicate that we don't want payment address over the rpc if we do a `SendPayment` with the custom key send record? We will default to mpp for 0.9?

joostjager: Discussed offline. We will force old-style payment key send by passing in manual features to sendpayment.

Reminder: if key send gets merged first, the manual feature set needs to be added to `lncli` in this pr.

routing/pathfind.go
```diff
@@ -141,8 +162,51 @@ func newRoute(sourceVertex route.Vertex,
 			amtToForward     lnwire.MilliSatoshi
 			fee              lnwire.MilliSatoshi
 			outgoingTimeLock uint32
+			tlvPayload       bool
 		)
+
+		// Define a closure that determines whether or not the edge
+		// supports a particular feature. This will be used to test
+		// intermediate hops for support of a particular feature.
+		nodeSupports := func(bit lnwire.FeatureBit) bool {
```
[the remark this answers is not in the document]
cfromknecht: no longer needed

routing/pathfind.go
```diff
@@ -233,6 +235,21 @@ func newRoute(sourceVertex route.Vertex,
...
+			// If the receiver supports TLV onion payloads and we
+			// have a payment address from the invoice, we'll go
+			// ahead and include an MPP record to gain the
+			// additional protections offered by MPP. If the feature
+			// is present, we know we can dereference the payment
+			// address due to our final hop validation above.
+			payAddr := finalHop.features.HasFeature(
+				lnwire.PaymentAddrOptional,
+			)
+			if payAddr && finalHop.paymentAddr != nil {
```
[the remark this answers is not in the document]
cfromknecht: done

routing/pathfind.go
cfromknecht: fixed by moving checks to pathfinding, and only having lighter checks in route construction

lnrpc/routerrpc/router.proto
cfromknecht: i was thinking i made sense to add an experimental feature to routerrpc, but either way is okay with me

joostjager: I am ok with that too, but how is that going to work in `lncli`? Call the same sendpayment logic from a different `lncli` command that is behind a build flag?

routing/pathfind.go
```diff
@@ -99,11 +99,12 @@ type edgePolicyWithSource struct {
 // finalHopParams encapsulates various parameters for route construction that
 // apply to the final hop in a route. These features include basic payment data
 // such as amounts and cltvs, as well as more complex features like destination
-// TLV records.
+// TLV records and payment address.
```
joostjager: 'destination custom records' (pre-existing)

routing/pathfind.go
```diff
@@ -189,6 +191,23 @@ func newRoute(sourceVertex route.Vertex,
 					"custom records")
 			}
 			customRecords = finalHop.records
+
+			// If we're attaching a payment addr but the receiver
+			// doesn't support both TLV and payment addrs, fail.
+			payAddr := supports(lnwire.PaymentAddrOptional)
+			if (!tlvPayload || !payAddr) &&
```
joostjager: Isn't `tlvPayload` implicitly supported after we've validated the feature bits?

routing/pathfind.go
```diff
@@ -189,6 +191,23 @@ func newRoute(sourceVertex route.Vertex,
...
+				return nil, errors.New("cannot attach " +
+					"payment addr")
+			}
+
+			// Otherwise attach the mpp record if it exists.
+			if finalHop.paymentAddr != nil {
+				mpp = record.NewMPP(
+					finalHop.amt, *finalHop.paymentAddr,
+				)
+			}
```
joostjager: Should we do this here, or create `mpp` and `requiredFeatures` fields on `finalHopParams`? It is a question about how intelligent we want `newRoute` to be.

routing/pathfind.go
```diff
@@ -99,11 +99,12 @@ type edgePolicyWithSource struct {
...
-// TLV records.
+// TLV records and payment address.
 type finalHopParams struct {
-	amt       lnwire.MilliSatoshi
-	cltvDelta uint16
-	records   record.CustomSet
+	amt         lnwire.MilliSatoshi
+	cltvDelta   uint16
+	records     record.CustomSet
+	paymentAddr *[32]byte
```
joostjager: Can this struct `finalHopParams` be reused to pass into findPath?

joostjager:
To show you what I am up to with checking the max routing info size: https://github.com/lightningnetwork/lnd/pull/3841

The idea of constructing `route.Hop` objects while searching looks good to me and could also allow us to get rid of the somewhat artificial split between pathfinding and route construction.

routing/pathfind.go
cfromknecht: good point, fixed

routing/pathfind.go
```diff
@@ -427,10 +428,22 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+	// checking that it supports the features we need, given our
+	// restrictions on the final hop.
+	switch {
+
 	// If the caller needs to send custom records, check that our
 	// destination feature vector supports TLV.
-	if len(r.DestCustomRecords) > 0 &&
-		!features.HasFeature(lnwire.TLVOnionPayloadOptional) {
+	case len(r.DestCustomRecords) > 0 &&
+		!features.HasFeature(lnwire.TLVOnionPayloadOptional):
```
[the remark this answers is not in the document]
cfromknecht: fixed

routing/pathfind.go
```diff
@@ -446,6 +456,13 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		!features.HasFeature(lnwire.TLVOnionPayloadOptional):
 
 		return nil, errNoTlvPayload
+
+	// If the caller has a payment address to attach, check that our
+	// destination feature vector supports them.
+	case r.PaymentAddr != nil &&
```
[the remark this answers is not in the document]
cfromknecht: yes good point, fixed

routing/testdata/basic_graph.json
```diff
@@ -45,7 +45,8 @@
         {
             "source": false, 
             "pubkey": "03c19f0027ffbb0ae0e14a4d958788793f9d74e107462473ec0c3891e4feb12e99", 
-            "alias": "satoshi" 
+            "alias": "satoshi",
+            "features": [9]
```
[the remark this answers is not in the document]
cfromknecht: JSON doesn't support comments, i could add a `_comment` field but that seems hacky

routing/pathfind_test.go
```diff
@@ -952,22 +963,57 @@ func TestPathFindingWithAdditionalEdges(t *testing.T) {
 		graph.aliasMap["songoku"]: {songokuToDoge},
 	}
 
+	find := func(
+		r *RestrictParams) ([]*channeldb.ChannelEdgePolicy, error) {
```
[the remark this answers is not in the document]
cfromknecht: done

routing/pathfind.go
```diff
@@ -147,7 +147,22 @@ func newRoute(sourceVertex route.Vertex,
...
+		// point that the feature vectors transitive dependencies have
+		// been validated.
+		supports := edge.Node.Features.HasFeature
+
+		// We start by assuming the node doesn't support TLV. We'll now
+		// inspect the node's feature vector to see if we can promote
+		// the hop. We assume already that the feature vector's
+		// transitive dependencies have already been validated by path
+		// finding or some other means.
+		tlvPayload = supports(lnwire.TLVOnionPayloadOptional)
```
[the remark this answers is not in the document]
cfromknecht: my reading is no because it loads the policies from the graph

routing/pathfind.go
cfromknecht: do we gain anything functionally by doing so? in this pr, we would only be using 2 of the 4 fields, perhaps it is  better left to when `findPath` and `newRoute` are fully unified? given how long this pr has been open, i would prefer to defer pure refactorings so that we can unblock other changes

cfromknecht:
I'll investigate how much effort it would be to decouple the new tests from the test graph

lnrpc/routerrpc/router.proto
cfromknecht: done. it should also be possible for people to send custom features by casting using `lnrpc.FeatureBit(x)`

routing/pathfind_test.go
```diff
@@ -1280,6 +1326,85 @@ func TestPathNotAvailable(t *testing.T) {
...
+		target route.Vertex) ([]*channeldb.ChannelEdgePolicy, error) {
+
+		return findPath(
+			&graphParams{
+				graph: graph.graph,
+			},
+			r, testPathFindingConfig,
+			sourceNode.PubKeyBytes, target, 100,
+		)
+	}
```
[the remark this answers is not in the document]
cfromknecht: done

routing/pathfind.go
```diff
@@ -292,6 +292,11 @@ type RestrictParams struct {
 	// DestCustomRecords contains the custom records to drop off at the
 	// final hop, if any.
 	DestCustomRecords record.CustomSet
+
```
joostjager: Left over last sentence in commit msg

routing/pathfind_test.go
```diff
@@ -1296,6 +1342,110 @@ func TestPathNotAvailable(t *testing.T) {
...
+	ctx := newPathFindingTestContext(t, testChannels, "roasbeef")
+	defer ctx.cleanup()
+
+	sourceNode, err := ctx.graphParams.graph.SourceNode()
+	if err != nil {
+		t.Fatalf("unable to fetch source node: %v", err)
+
+	}
+
+	find := func(r *RestrictParams,
```
joostjager: can't this be `pathFindingTestContext.findPath`?

routing/payment_lifecycle.go
```diff
@@ -190,7 +190,11 @@ func (p *paymentLifecycle) resumePayment() ([32]byte, *route.Route, error) {
...
 	switch err {
-	case errNoTlvPayload, errNoPathFound, errMaxHopsExceeded,
+	case
+		errNoTlvPayload,
+		errNoPaymentAddr,
+		errNoPathFound,
+		errMaxHopsExceeded,
 		errPrebuiltRouteTried:
 
 		return channeldb.FailureReasonNoRoute
```
joostjager: Maybe later we want to reconsider the mapping of all of this to `NoRoute`

routing/pathfind.go
```diff
@@ -644,6 +644,34 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+		// route construction within findPath, and avoid using
+		// ChannelEdgePolicy altogether.
+		case edge.Node.PubKeyBytes == target:
+			edge.Node.Features = features
+
+		// Otherwise, this is some other intermediary node. Verify the
+		// transitive feature dependencies for this node, and skip the
+		// channel if they are invalid.
+		default:
+			err := feature.ValidateDeps(edge.Node.Features)
```
joostjager: This is potentially executed for every incoming edge of a node, while it really only need to be done once. Probably not an issue for now. Or would you want to keep a separate map for this, mapping `Node` to `Features` with a 'seen' marker?

routing/pathfind.go
joostjager: Discussed offline, future pr.

lnrpc/rpc.proto
```diff
@@ -2647,6 +2654,26 @@ message PayReq {
...
+        TLV_ONION_OPT = 9;
+        EXT_GOSSIP_QUERIES_REQ = 10;
+        EXT_GOSSIP_QUERIES_OPT = 11;
+        STATIC_REMOTE_KEY_REQ = 12;
+        STATIC_REMOTE_KEY_OPT = 13;
+        PAYMENT_ADDR_REQ = 14;
+        PAYMENT_ADDR_OPT = 15;
+        MPP_REQ = 16;
+        MPP_OPT = 17;
+}
```
joostjager: We also add the non-payment related ones for if people query node info for example?

routing/payment_lifecycle.go
joostjager: NoTlvPayload for the final node isn't really a `NoRoute` result. `InvalidDetails` would cover the situation better probably. But doesn't need to be discussed/changed now.

routing/pathfind.go
joostjager: I also need to have those features available at the right time for the running payload size calculation. What do you think of this one? https://github.com/lightningnetwork/lnd/pull/3851 (last commit)

routing/pathfind.go
```diff
@@ -188,6 +190,21 @@ func newRoute(sourceVertex route.Vertex,
...
+			// doesn't support both TLV and payment addrs, fail.
+			payAddr := supports(lnwire.PaymentAddrOptional)
+			if !payAddr && finalHop.paymentAddr != nil {
+				return nil, errors.New("cannot attach " +
+					"payment addr")
+			}
+
+			// Otherwise attach the mpp record if it exists.
+			if finalHop.paymentAddr != nil {
+				mpp = record.NewMPP(
```
halseth: this logic should really have been in `newRoute` creation, but I guess we are stuck with doing it here now?

lnrpc/routerrpc/router_backend.go
```diff
@@ -699,6 +708,27 @@ func unmarshallHopHint(rpcHint *lnrpc.HopHint) (zpay32.HopHint, error) {
...
+	raw := lnwire.NewRawFeatureVector()
+	for _, bit := range rpcFeatures {
+		err := raw.SafeSet(lnwire.FeatureBit(bit))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	fv := lnwire.NewFeatureVector(raw, lnwire.Features)
+	err := feature.ValidateDeps(fv)
```
halseth: I'm not sure if invalid deps should stand in the way of invoice parsing.

halseth:
Great commit structure, makes even a massive PR like this manageable to review 👍

routing/pathfind.go
cfromknecht: this is in `newRoute`?

routing/pathfind_test.go
```diff
@@ -447,7 +454,7 @@ func createTestGraphFromChannels(testChannels []*testChannel, source string) (
 	}
 
 	// Add the source node.
-	dbNode, err := addNodeWithAlias(source)
+	dbNode, err := addNodeWithAlias(source, nil)
```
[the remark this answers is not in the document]
cfromknecht: done

lnrpc/routerrpc/router_backend.go
cfromknecht: good point, we check this later on during pathfinding anyway. removed

routing/pathfind.go
```diff
@@ -292,6 +292,11 @@ type RestrictParams struct {
 	// DestCustomRecords contains the custom records to drop off at the
 	// final hop, if any.
 	DestCustomRecords record.CustomSet
+
+	// DestFeatures is a feature vector describing what the final hop
+	// supports. If none are provided, pathfinding will try to inspect any
```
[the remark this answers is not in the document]
cfromknecht: we could do that, for now we just override on the assumption that the invoice is more critical than the node announcement (since it may advertise a distinct combination from the graph). we could consider changing this in the future tho

routing/pathfind.go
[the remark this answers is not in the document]
cfromknecht: for now actually, we only care about TLV support which has no dependencies. in general tho i think it is better to be strict and ignore nodes that don't generate well formed announcemements

joostjager: Discussed offline, will be dealt with in https://github.com/lightningnetwork/lnd/pull/3841

routing/pathfind.go
halseth: lol

PR #3685, opened 2019-11-07
flat features

cfromknecht:
An implementation of https://github.com/lightningnetwork/lightning-rfc/pull/666


peer.go
```diff
@@ -2388,10 +2388,10 @@ func (p *peer) WipeChannel(chanPoint *wire.OutPoint) error {
 // local features vectors. If feature vectors are incompatible then disconnect.
 func (p *peer) handleInitMsg(msg *lnwire.Init) error {
 	p.remoteLocalFeatures = lnwire.NewFeatureVector(
-		msg.LocalFeatures, lnwire.LocalFeatures,
+		msg.LocalFeatures, lnwire.Features,
 	)
 	p.remoteGlobalFeatures = lnwire.NewFeatureVector(
-		msg.GlobalFeatures, lnwire.GlobalFeatures,
+		msg.GlobalFeatures, lnwire.Features,
```
joostjager: Can something go wrong here with communicating to older nodes previously required local-only features as global and vv? Maybe fixed in a later commit

feature/manager_test.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, test := range managerTests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			testManager(t, test)
+		})
+	}
+}
+
+func testManager(t *testing.T, test managerTest) {
+	m, err := feature.NewManager(test.cfg)
```
joostjager: Should `defaultSets` be a parameter of `NewManager`, to isolate the test from the current feature set?

feature/default_sets.go
```diff
@@ -0,0 +1,28 @@
+package feature
+
+import "github.com/lightningnetwork/lnd/lnwire"
+
+// defaultSets are the default set descriptors for generating feature vectors.
+// Each set is annotated with the corresponding identifier from BOLT 9
+// indicating where it should be advertised.
+var defaultSets = map[lnwire.FeatureBit]map[Set]struct{}{
```
joostjager: Should this commit be the first, to build on top of it with the selection of features in different contexts?

channeldb/migration_01_to_11/graph.go
```diff
@@ -692,6 +692,8 @@ func fetchLightningNode(nodeBucket *bbolt.Bucket,
 	return deserializeLightningNode(nodeReader)
 }
 
+var globalFeatures = map[lnwire.FeatureBit]string{}
```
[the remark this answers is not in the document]
cfromknecht: this was the global features map at the time this migration happened. either way, it is only inspected if `HasFeature` is called, which doesn't happen during the migration. it could equally be nil and wouldn't affect serialization, perhaps that is better?

peer.go
cfromknecht: it is less restrictive, so we might accept a required global feature in local and vice versa, but that only applies if someone lands on an intermediate commit.

feature/manager_test.go
cfromknecht: done

channeldb/migration_01_to_11/graph.go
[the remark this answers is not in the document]
joostjager: Ok. With nil the migration would at least fail (?) if we would use it. Seems better.

feature/default_sets.go
cfromknecht: done

channeldb/migration_01_to_11/graph.go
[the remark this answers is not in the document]
cfromknecht: done.

peer.go
cfromknecht: reworked so that we send ff first, then merge the remote's features

lnwire/init_message.go
```diff
@@ -7,20 +7,26 @@ import "io"
...
 
-	// LocalFeatures is feature vector which only affect the protocol
-	// between two nodes.
-	LocalFeatures *RawFeatureVector
+	// Features is a feature vector containing a the features supported by
+	// the remote node.
+	//
+	// NOTE: Older nodes may place some features in GlobalFeatures, but all
+	// new features are to be added in Features. When handling an Init
+	// message, any GlobalFeatures should be mereged into the unified
```
joostjager: mereged

cfromknecht: fixed

peer.go
```diff
@@ -2387,62 +2391,64 @@ func (p *peer) WipeChannel(chanPoint *wire.OutPoint) error {
...
 func (p *peer) handleInitMsg(msg *lnwire.Init) error {
-	p.remoteLocalFeatures = lnwire.NewFeatureVector(
-		msg.LocalFeatures, lnwire.LocalFeatures,
-	)
-	p.remoteGlobalFeatures = lnwire.NewFeatureVector(
-		msg.GlobalFeatures, lnwire.GlobalFeatures,
+	// First, merge any features from the legacy global features field into
+	// those presented in the local features fields.
+	err := msg.Features.Merge(msg.GlobalFeatures)
+	if err != nil {
```
halseth: Just making sure this'll work: If an old node is sending overlapping global and local features (which is allowed), this will fail. I'm guessing we should be good since no such overlapping features (seemingly) was defined in the spec before. 

But the new spec specifies we should just take the "logical OR" of the vectors, which is not what we do here.

channeldb/migration_01_to_11/graph.go
```diff
@@ -744,7 +744,7 @@ func deserializeLightningNode(r io.Reader) (LightningNode, error) {
 		return LightningNode{}, err
 	}
 
-	fv := lnwire.NewFeatureVector(nil, lnwire.GlobalFeatures)
+	fv := lnwire.NewFeatureVector(nil, lnwire.Features)
```
halseth: (maybe this was already discussed but the diff is gone) Shouldn't this be isolated?

lnwire/features.go
```diff
@@ -102,6 +102,21 @@ var GlobalFeatures = map[FeatureBit]string{
 	StaticRemoteKeyRequired: "static-remote-key",
 }
 
+// Features is a mapping of known feature bits to a descriptive name. All known
+// feature bits must be assigned a name in this mapping, and feature bit pairs
+// must be assigned together for correct behavior.
+var Features = map[FeatureBit]string{
```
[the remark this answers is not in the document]
cfromknecht: the commits are structured this way so that each commit won't break existing behavior, so we first send flat features and then read them. the prior iteration did as you suggest but joost pointed out that the intermediate commits could break compatibility. in order to do accommodate, we have to first add `Features`, the later remove the local/global features

peer.go
cfromknecht: it will only fail if they set optional and required for the same feature bit, which is a little stricter than the spec requires, but i think it's better to prevent unintended behavior. we still take the OR otherwise

peer.go
```diff
@@ -2395,54 +2391,56 @@ func (p *peer) WipeChannel(chanPoint *wire.OutPoint) error {
...
+		err := fmt.Errorf("unable to merge legacy global featues: %v",
```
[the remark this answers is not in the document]
cfromknecht: fixed

lnpeer/peer.go
```diff
@@ -47,15 +47,13 @@ type Peer interface {
...
-	// bits.
-	LocalGlobalFeatures() *lnwire.FeatureVector
-
-	// RemoteGlobalFeatures returns the set of global features that has
-	// been advertised by the remote peer. This allows sub-systems that use
-	// this interface to gate their behavior off the set of negotiated
-	// feature bits.
-	RemoteGlobalFeatures() *lnwire.FeatureVector
+	// LocalFeatures returns the set of features that has been advertised by
+	// the local peer. This allows sub-systems that use this interface to
```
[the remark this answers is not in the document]
cfromknecht: done

server.go
```diff
@@ -232,9 +233,9 @@ type server struct {
 
 	readPool *pool.Read
 
-	// globalFeatures feature vector which affects HTLCs and thus are also
-	// advertised to other nodes.
-	globalFeatures *lnwire.FeatureVector
+	// featureMgr dispatches feature vectors for various contexts within the
+	// daemon.
+	featureMgr *feature.Manager
```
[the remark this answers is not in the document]
cfromknecht: we will need to make one for each new context, but that is my plan for invoices specifically since it will be used outside of the server/main pkg

server.go
```diff
@@ -594,7 +601,7 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 		LastUpdate:           time.Now(),
 		Addresses:            selfAddrs,
 		Alias:                nodeAlias.String(),
-		Features:             s.globalFeatures,
+		Features:             s.featureMgr.Get(feature.SetNodeAnn),
```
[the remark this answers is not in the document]
cfromknecht: it is quite nice to have it all abstracted :)

feature/manager.go
```diff
@@ -0,0 +1,97 @@
...
+// feature sets.
+type Manager struct {
+	// fsets is a static map of feature set to raw feature vectors. Requests
+	// are fulfilled by cloning these interal feature vectors.
+	fsets map[Set]*lnwire.RawFeatureVector
+}
+
+// NewManager creates a new feature Manager, applying any custom modifications
+// to its feature sets before returning.
+func NewManager(cfg Config) (*Manager, error) {
```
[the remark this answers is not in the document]
cfromknecht: ehh too long imo :)

lnwire/features.go
```diff
@@ -146,6 +163,17 @@ func (fv *RawFeatureVector) Set(feature FeatureBit) {
 	fv.features[feature] = true
 }
 
+// SafeSet sets the chosen feature bit in the feature vector, but returns an
+// error if the opposing feature bit is already set.
```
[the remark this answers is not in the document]
cfromknecht: done!

lnwire/features.go
```diff
@@ -4,6 +4,14 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+
+	"github.com/go-errors/errors"
```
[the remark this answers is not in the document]
cfromknecht: fixed

channeldb/migration_01_to_11/graph.go
cfromknecht: also fixed, replaced w/ nil

peer.go
cfromknecht: it will only fail if they set optional and required for the same feature bit, which is a little stricter than the spec requires, but i think it's better to prevent unintended behavior. we still take the OR otherwise

lnwire/features.go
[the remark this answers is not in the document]
joostjager: yes, I found the pr better like this. But maybe if there isn't a perfect order, there are too many commits.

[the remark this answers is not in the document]
halseth: Makes sense, agree on compatibility over reviewability for a small diff as this.

server.go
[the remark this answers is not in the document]
halseth: I think it would work here though, as the `get` method is all that is used from this manager after it is created. As we thread this into other packages, then we wouldn't need to depend on the manager struct. But no blocker, just an observation.

peer.go
halseth: hm, okay, I guess it couldn't hurt.

Issue #3539, opened 2019-09-24
Unable to start server: a height hint greater than 0 must be provided


cyclopes2011:
### Background

Attempting to restore LND from wallet seed and channel.backup file.

### Your environment

* lnd version 0.7.1-beta commit=v0.7.1-beta-352-9b1ecbd3fa3fd93a6d95a5f71921c73777340760
* 16.04.1-Ubuntu
* Bitcoin Core Daemon version v0.16.3.0-g49e34e288 (installed on dedicated server)

### Steps to reproduce

Performed the steps to restore LND from wallet seed and channel.backup file via `lncli create` command.

### Expected behaviour

Existing wallet restored via `lncli create` and funds returned to wallet after channel restore and closeall per the SCB recovery process.

### Actual behaviour

LND fails to start due to the following error:

`2019-09-24 03:55:09.062 [ERR] LTND: Unable to start server: a height hint greater than 0 must be provided`

Complete startup log attached below:
[20190924 1026 steropes.log](https://github.com/lightningnetwork/lnd/files/3647692/20190924.1026.steropes.log)

cyclopes2011:
Similar to #3444

wpaulino:
The following patch should allow you to start your node again:

```diff --git a/chanrestore.go b/chanrestore.go
index f771f91b..34f74f69 100644
--- a/chanrestore.go
+++ b/chanrestore.go
@@ -138,7 +138,8 @@ func (c *chanDBRestorer) RestoreChansFromSingles(backups ...chanbackup.Single) e
 	for _, restoredChannel := range channelShells {
 		err := c.chainArb.WatchNewChannel(restoredChannel.Chan)
 		if err != nil {
-			return err
+			continue
+			//return err
 		}
 	}
 
```

Closing the issue as a duplicate of #3444.

cyclopes2011:
Here is what I did but the error remains the same.

Sorry, I am not a developer so I assume the @@ is a location for the code function and not part of the edit. I have edited the "chanrestore.go" file per your instruction and have included the comments prior to the section to be edited for clarity:

        // Finally, we'll need to inform the chain arbitrator of these new
        // channels so we'll properly watch for their ultimate closure on chain
        // and sweep them via the DLP.
        for _, restoredChannel := range channelShells {
                err := c.chainArb.WatchNewChannel(restoredChannel.Chan)
                if err != nil {
                        continue
                        // return err
                }
        }

Thank you for your response and assistance. :)

[...]

Issue #3444, opened 2019-08-31
Static channel backups made in pending state return error when used to recover


alexbosworth:
### Background

When restoring funds using a static channel backup, in some scenarios the backup recovery method fails.

The static channel backup verification also shows this failing backup as valid

### Your environment

* lnd commit `17b6205f3a7b06b2db5f7807420629211ecb86ad`
* regtest btcd

### Steps to reproduce

1. Create a new channel between A and B
2. While the channel is still pending, grab the backup via `ExportAllChannelBackups`
3. Kill the node and wipe state
4. Attempt to restore the channel using the backup file `RestoreChannelBackups`

### Expected behavior

No error should be returned and the funds should attempt to be restored using the backup file

### Actual behavior

Restoration fails with error:

```
2 UNKNOWN: unable to unpack chan backup: a height hint greater than 0 must be provided
```

cyclopes2011:
Not sure if this is related but I am getting a similar error after attempting to do a restore. I am currently stuck and unable to get LND running to salvage funds.

> 2019-09-19 09:56:40.301 [ERR] LTND: Unable to start server: a height hint greater than 0 must be provided
2019-09-19 09:56:40.302 [INF] RPCS: Stopping RPC Server
2019-09-19 09:56:40.310 [INF] LTND: Shutdown complete Unable to start server: a height hint greater than 0 must be provided

How do I specify a "height hint" and get back online? I have a mess of a history with a server move, return to an old server, and now a restore of a new channel.backup so I think I've messed up permanently. The new server was never successfully online, both servers were never online at the same time, etc. so I may have a slim chance of things getting back to normal.

Sorry I haven't been keeping up with the release notes so I am currently reading up on what I can find. Thank you in advance for your assistance. I can also provide more details if needed--I'm not a developer so please be specific. :)

Roasbeef:
Fixed by #3690.

PR #3690, opened 2019-11-07
chanrestore: assume height hint for unconfirmed channels in SCBs

guggero:
This is an attempt at fixing #3444 and #3539. This issue also came up in #2468.

The `chainWatcher` will take the height hint to register spend notifications from `ShortChanID`, which will be 0 for unconfirmed channels. As a fall back, it then takes the `FundingBroadcastHeight` of the channel. But this value is also 0 in case of a channel shell restored from an SCB.

So for unconfirmed channels in SCBs we have no idea what height hint to use. We can't use 0 because of a restriction in `chainntnfs` introduced in #3405.

This PR tries to use the block height of the earliest confirmed channel in the backup passed to it as a height hint. If there is no unconfirmed channel, it falls back to a static block height which corresponds to the launch date of SCBs in lnd.

Maybe there is a more elegant way to solve this, suggestions are very welcome.

### DIFF
--- a/chanrestore.go
+++ b/chanrestore.go
@@ -2,9 +2,11 @@ package lnd
 
 import (
 	"fmt"
+	"math"
 	"net"
 
 	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/lightningnetwork/lnd/chanbackup"
 	"github.com/lightningnetwork/lnd/channeldb"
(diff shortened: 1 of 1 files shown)


guggero:
Yes, I think approach 2 should also be implemented. I'll create that PR next.

But I still think we need this workaround for people that need to rely on an SCB that they created before 0.9 or whenever we release that fix. Would you agree?

Roasbeef:
Revisiting this, what's wrong with the first option? The SCBs stay the same, we just populate them slightly differently. Once everyone upgrades, then they're now covered for those _new_ SCBs. Even if we go with the second option, those that have SCBs (older ones) that _don't_ set the broadcast height properly will still have rescanning issues.

guggero:
Setting the broadcast height in the `ShortChannelID` would certainly be the easiest fix since no version change in the SCBs would be needed. I just wasn't sure about the impact of that change since I don't know where the `ShortChannelID` is used for identifying purposes. I thought the change would perhaps break things in other places so option 2 would be "safer" from that point of view.
I can implement option 1 as a separate PR and then we can discuss the impact there.

But you're saying, we still need and want this PR, right?

wpaulino:
> Even if we go with the second option, those that have SCBs (older ones) that don't set the broadcast height properly will still have rescanning issues.

We can detect that an old version is being used on startup and upgrade it, but that still won't work for nodes that haven't upgraded.

> But you're saying, we still need and want this PR, right?

Seems fine to me to proceed with it for now as neither of the options we've discussed will handle older SCBs.

Roasbeef:
> But I still think we need this workaround for people that need to rely on an SCB that they created before 0.9 or whenever we release that fix. Would you agree?

Yes, which seems to be #3719. We could even package this in an earlier minor release since it doesn't create any new SCBs, but instead interprets the existing ones in a slightly different way during recovery.

guggero:
Going to merge this into #3719 and consolidate.

PR #3691, opened 2019-11-07
lnwallet: Make available balance HTLC fee aware

### LINKED ISSUE #3787: Take channel reserve into account within pathfinding
### Background

Describe your issue here.
Paying a destination until it becomes one sided and dips into channel reserve, LND will continue to try this route even after it errors.
### Your environment

* version of `lnd` 0.8.0-beta commit=v0.8.0-beta-274-g59c4db7f33e557dcfec7dae8f204294b5a627670-dirty"

### Steps to reproduce

Tell us how to reproduce this issue. Please provide stacktraces and links to code in question.

### Expected behaviour

Tell us what should happen
LND should avoid the channel when attempting future payments if it dips into channel reserve and attempt other routes.

### Actual behaviour
I get hammered with these logs making sending unusable to the destination:

2019-12-04 15:47:28.755 [WRN] HSWC: ChannelLink(579497:1927:1): unable to handle downstream add HTLC: commitment transaction dips peer below chan reserve
2019-12-04 15:47:28.774 [ERR] CRTR: Attempt to send payment 2506cab6ef89726de7342971a56b5f78f986481f534b8b8abd235f5f3f42c299 failed: TemporaryChannelFailure(update=(*lnwire.ChannelUpdate)(0xc01032f600)({
 Signature: (lnwire.Sig) (len=64 cap=64) {
  00000000  f7 77 3b ac 51 fd 65 85  20 56 73 71 91 11 d9 52  |.w;.Q.e. Vsq...R|
  00000010  eb ee 9f 46 6c c8 d0 7a  a0 b9 63 de 45 5a 42 cd  |...Fl..z..c.EZB.|
  00000020  08 40 a9 c6 8e 47 d3 cb  11 f2 70 2b f8 03 f8 22  |.@...G....p+..."|
  00000030  40 4a 7e 3e c5 e1 11 9e  cf 45 9b a7 7f 51 b6 c4  |@J~>.....E...Q..|
 },
 ChainHash: (chainhash.Hash) (len=32 cap=32) 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f,
 ShortChannelID: (lnwire.ShortChannelID) 579497:1927:1,
 Timestamp: (uint32) 1575428703,
 MessageFlags: (lnwire.ChanUpdateMsgFlags) 00000001,
 ChannelFlags: (lnwire.ChanUpdateChanFlags) 00000000,
 TimeLockDelta: (uint16) 40,
 HtlcMinimumMsat: (lnwire.MilliSatoshi) 1000 mSAT,
 BaseFee: (uint32) 4000,
 FeeRate: (uint32) 1000,
 HtlcMaximumMsat: (lnwire.MilliSatoshi) 3960000000 mSAT,
 ExtraOpaqueData: ([]uint8) <nil>
})
)@0
2019-12-04 15:47:28.775 [ERR] CRTR: Channel update of ourselves received
2019-12-04 15:47:28.783 [WRN] CRTR: Routing failure for local channel 637163689887596545 occurred
2019-12-04 15:47:29.520 [WRN] HSWC: ChannelLink(579497:1927:1): unable to handle downstream add HTLC: commitment transaction dips peer below chan reserve
2019-12-04 15:47:29.539 [ERR] CRTR: Attempt to send payment 2506cab6ef89726de7342971a56b5f78f986481f534b8b8abd235f5f3f42c299 failed: TemporaryChannelFailure(update=(*lnwire.ChannelUpdate)(0xc01032f760)({
 Signature: (lnwire.Sig) (len=64 cap=64) {
  00000000  f7 77 3b ac 51 fd 65 85  20 56 73 71 91 11 d9 52  |.w;.Q.e. Vsq...R|
  00000010  eb ee 9f 46 6c c8 d0 7a  a0 b9 63 de 45 5a 42 cd  |...Fl..z..c.EZB.|
  00000020  08 40 a9 c6 8e 47 d3 cb  11 f2 70 2b f8 03 f8 22  |.@...G....p+..."|
  00000030  40 4a 7e 3e c5 e1 11 9e  cf 45 9b a7 7f 51 b6 c4  |@J~>.....E...Q..|
 },
 ChainHash: (chainhash.Hash) (len=32 cap=32) 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f,
 ShortChannelID: (lnwire.ShortChannelID) 579497:1927:1,
 Timestamp: (uint32) 1575428703,
 MessageFlags: (lnwire.ChanUpdateMsgFlags) 00000001,
 ChannelFlags: (lnwire.ChanUpdateChanFlags) 00000000,
 TimeLockDelta: (uint16) 40,
 HtlcMinimumMsat: (lnwire.MilliSatoshi) 1000 mSAT,
 BaseFee: (uint32) 4000,
 FeeRate: (uint32) 1000,
 HtlcMaximumMsat: (lnwire.MilliSatoshi) 3960000000 mSAT,
 ExtraOpaqueData: ([]uint8) <nil>
})
)@0
2019-12-04 15:47:29.539 [ERR] CRTR: Channel update of ourselves received
2019-12-04 15:47:29.547 [WRN] CRTR: Routing failure for local channel 637163689887596545 occurred

I believe @alexbosworth also has this issue.

Closing the channel helps, but not ideal.

halseth:
This PR makes the balance calculation take into account the extra fee needed to be paid by the initiator when sending non-dust HTLCs. This is a case that could previously be hit both if we as an initiator couldn't pay the HTLC fee, and when the remote (as initiator) couldn't pay to add our HTLC to the commitment.

This PR mitigates this by in those cases reporting a lower number, ensuring the pathfinding algorithm won't attempt to use the channel in such cases. 

Fixes #3787.

### DIFF
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -2127,26 +2127,21 @@ func (l *channelLink) ChanID() lnwire.ChannelID {
 //
 // NOTE: Part of the ChannelLink interface.
 func (l *channelLink) Bandwidth() lnwire.MilliSatoshi {
+	// Get the balance available on the channel for new HTLCs. This takes
+	// the channel reserve into account so HTLCs up to this value won't
+	// violate it.
 	channelBandwidth := l.channel.AvailableBalance()
-	overflowBandwidth := l.overflowQueue.TotalHtlcAmount()
 
 	// To compute the total bandwidth, we'll take the current available
 	// bandwidth, then subtract the overflow bandwidth as we'll eventually
 	// also need to evaluate those HTLC's once space on the commitment
 	// transaction is free.
-	linkBandwidth := channelBandwidth - overflowBandwidth
-
-	// If the channel reserve is greater than the total available balance
-	// of the link, just return 0.
-	reserve := lnwire.NewMSatFromSatoshis(l.channel.LocalChanReserve())
-	if linkBandwidth < reserve {
+	overflowBandwidth := l.overflowQueue.TotalHtlcAmount()
+	if channelBandwidth < overflowBandwidth {
 		return 0
 	}
 
-	// Else the amount that is available to flow through the link at this
-	// point is the available balance minus the reserve amount we are
-	// required to keep as collateral.
-	return linkBandwidth - reserve
+	return channelBandwidth - overflowBandwidth
 }
 
 // AttachMailBox updates the current mailbox used by this link, and hooks up
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -6068,20 +6068,30 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 		ourBalance = 0
 	}
 
-	// Given the commitment weight, find the commitment fee in case of no
-	// added HTLC output.
+	// Calculate the commitment fee in the case where we would add another
+	// HTLC to the commitment, as only the balance remaining after this fee
+	// has been paid is actually available for sending.
 	feePerKw := filteredView.feePerKw
-	baseCommitFee := lnwire.NewMSatFromSatoshis(
-		feePerKw.FeeForWeight(commitWeight),
+	htlcCommitFee := lnwire.NewMSatFromSatoshis(
+		feePerKw.FeeForWeight(commitWeight + input.HTLCWeight),
 	)
 
-	// If we are the channel initiator, we must to subtract the commitment
-	// fee from our available balance.
+	// If we are the channel initiator, we must to subtract this commitment
+	// fee from our available balance in order to ensure we can afford both
+	// the value of the HTLC and the additional commitment fee from adding
+	// the HTLC.
 	if lc.channelState.IsInitiator {
-		if ourBalance < baseCommitFee {
+		// There is an edge case where our non-zero balance is lower
+		// than the htlcCommitFee, where we could still be sending dust
+		// HTLCs, but we return 0 in this case. This is to avoid
+		// lowering our balance even further, as this takes us into a
+		// bad state wehere neither we nor our channel counterparty can
+		// add HTLCs.
+		if ourBalance < htlcCommitFee {
 			return 0, commitWeight
 		}
-		ourBalance -= baseCommitFee
+
+		ourBalance -= htlcCommitFee
 	}
 
 	return ourBalance, commitWeight
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -4642,10 +4642,26 @@ func (lc *LightningChannel) AddHTLC(htlc *lnwire.UpdateAddHTLC,
 	}
 
 	// Make sure adding this HTLC won't violate any of the constraints we
-	// must keep on our commitment transaction.
+	// must keep on the commitment transactions.
 	remoteACKedIndex := lc.localCommitChain.tail().theirMessageIndex
+
+	// First we'll check whether this HTLC can be added to the remote
+	// commitment transaction without violation any of the constraints.
 	err := lc.validateCommitmentSanity(
 		remoteACKedIndex, lc.localUpdateLog.logIndex, true, pd, nil,
+	)
+	if err != nil {
+		return 0, err
+	}
+
+	// We must also check whether it can be added to our own commitment
+	// transaction, or the remote node will refuse to sign. This is not
+	// totally bullet proof, as the remote might be adding updates
+	// concurrently, but if we fail this check there is for sure not
+	// possible for us to add the HTLC.
+	err = lc.validateCommitmentSanity(
+		lc.remoteUpdateLog.logIndex, lc.localUpdateLog.logIndex,
+		false, pd, nil,
 	)
 	if err != nil {
 		return 0, err
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -5997,11 +5997,14 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
 	return closeTx, ourBalance, nil
 }
 
-// AvailableBalance returns the current available balance within the channel.
-// By available balance, we mean that if at this very instance s new commitment
-// were to be created which evals all the log entries, what would our available
-// balance me. This method is useful when deciding if a given channel can
-// accept an HTLC in the multi-hop forwarding scenario.
+// AvailableBalance returns the current balance available for sending within
+// the channel. By available balance, we mean that if at this very instance a
+// new commitment were to be created which evals all the log entries, what
+// would our available balance for adding an additional HTLC be. It takes into
+// account the fee that must be paid for adding this HTLC (if we're the
+// initiator), and that we cannot spend from the channel reserve. This method
+// is useful when deciding if a given channel can accept an HTLC in the
+// multi-hop forwarding scenario.
 func (lc *LightningChannel) AvailableBalance() lnwire.MilliSatoshi {
 	lc.RLock()
 	defer lc.RUnlock()
@@ -6021,19 +6024,50 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 	htlcView := lc.fetchHTLCView(remoteACKedIndex,
 		lc.localUpdateLog.logIndex)
 
-	// Then compute our current balance for that view.
-	ourBalance, _, commitWeight, filteredView, err :=
-		lc.computeView(htlcView, false, false)
+	// Calculate our available balance from our local commitment.
+	//
+	// NOTE: This is not always accurate, since the remote node can always
+	// add updates concurrently, causing our balance to go down if we're
+	// the initiator, but this is a problem on the protocol level.
+	ourLocalCommitBalance, commitWeight := lc.availableCommitmentBalance(
+		htlcView,
+	)
+
+	return ourLocalCommitBalance, commitWeight
+}
+
+// availableCommitmentBalance attempts to calculate the balance we have
+// available for HTLCs on the local/remote commitment given the htlcView. To
+// account for sending HTLCs of different sizes, it will report the balance
+// available for sending non-dust HTLCs, which will be manifested on the
+// commitment, increasing the commitment fee we must pay as an initiator,
+// eating into our balance. It will make sure we won't violate the channel
+// reserve constraints for this amount.
+func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
+	lnwire.MilliSatoshi, int64) {
+
+	// Compute the current balances for this commitment. This will take
+	// into account HTLCs to determine the commit weight, which the
+	// initiator must pay the fee for.
+	ourBalance, _, commitWeight, filteredView, err := lc.computeView(
+		view, false, false,
+	)
 	if err != nil {
 		lc.log.Errorf("Unable to fetch available balance: %v", err)
 		return 0, 0
 	}
 
-	// If we are the channel initiator, we must remember to subtract the
-	// commitment fee from our available balance.
-	commitFee := filteredView.feePerKw.FeeForWeight(commitWeight)
+	// Given the commitment weight, find the commitment fee in case of no
+	// added HTLC output.
+	feePerKw := filteredView.feePerKw
+	baseCommitFee := lnwire.NewMSatFromSatoshis(
+		feePerKw.FeeForWeight(commitWeight),
+	)
+
+	// If we are the channel initiator, we must to subtract the commitment
+	// fee from our available balance.
 	if lc.channelState.IsInitiator {
-		ourBalance -= lnwire.NewMSatFromSatoshis(commitFee)
+		ourBalance -= baseCommitFee
 	}
 
 	return ourBalance, commitWeight
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -6030,8 +6030,18 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 	// add updates concurrently, causing our balance to go down if we're
 	// the initiator, but this is a problem on the protocol level.
 	ourLocalCommitBalance, commitWeight := lc.availableCommitmentBalance(
-		htlcView,
+		htlcView, false,
 	)
+
+	// Do the same calculation from the remote commitment point of view.
+	ourRemoteCommitBalance, _ := lc.availableCommitmentBalance(
+		htlcView, true,
+	)
+
+	// Return which ever balance is lowest.
+	if ourRemoteCommitBalance < ourLocalCommitBalance {
+		return ourRemoteCommitBalance, commitWeight
+	}
 
 	return ourLocalCommitBalance, commitWeight
 }
@@ -6043,14 +6053,14 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 // commitment, increasing the commitment fee we must pay as an initiator,
 // eating into our balance. It will make sure we won't violate the channel
 // reserve constraints for this amount.
-func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
-	lnwire.MilliSatoshi, int64) {
+func (lc *LightningChannel) availableCommitmentBalance(view *htlcView,
+	remoteChain bool) (lnwire.MilliSatoshi, int64) {
 
 	// Compute the current balances for this commitment. This will take
 	// into account HTLCs to determine the commit weight, which the
 	// initiator must pay the fee for.
 	ourBalance, theirBalance, commitWeight, filteredView, err := lc.computeView(
-		view, false, false,
+		view, remoteChain, false,
 	)
 	if err != nil {
 		lc.log.Errorf("Unable to fetch available balance: %v", err)
@@ -6117,6 +6127,17 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 	htlcFee := lnwire.NewMSatFromSatoshis(
 		htlcTimeoutFee(feePerKw),
 	)
+
+	// If we are looking at the remote commitment, we must use the remote
+	// dust limit and the fee for adding an HTLC success transaction.
+	if remoteChain {
+		dustlimit = lnwire.NewMSatFromSatoshis(
+			lc.channelState.RemoteChanCfg.DustLimit,
+		)
+		htlcFee = lnwire.NewMSatFromSatoshis(
+			htlcSuccessFee(feePerKw),
+		)
+	}
 
 	// The HTLC output will be manifested on the commitment if it
 	// is non-dust after paying the HTLC fee.
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -6057,6 +6057,17 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 		return 0, 0
 	}
 
+	// We can never spend from the channel reserve, so we'll subtract it
+	// from our available balance.
+	ourReserve := lnwire.NewMSatFromSatoshis(
+		lc.channelState.LocalChanCfg.ChanReserve,
+	)
+	if ourReserve <= ourBalance {
+		ourBalance -= ourReserve
+	} else {
+		ourBalance = 0
+	}
+
 	// Given the commitment weight, find the commitment fee in case of no
 	// added HTLC output.
 	feePerKw := filteredView.feePerKw
@@ -6067,6 +6078,9 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 	// If we are the channel initiator, we must to subtract the commitment
 	// fee from our available balance.
 	if lc.channelState.IsInitiator {
+		if ourBalance < baseCommitFee {
+			return 0, commitWeight
+		}
 		ourBalance -= baseCommitFee
 	}
 
@@ -6278,13 +6292,14 @@ func (lc *LightningChannel) MaxFeeRate(maxAllocation float64) chainfee.SatPerKWe
 
 	// The maximum fee depends of the available balance that can be
 	// committed towards fees.
-	balance, weight := lc.availableBalance()
+	commit := lc.channelState.LocalCommitment
 	feeBalance := float64(
-		balance.ToSatoshis() + lc.channelState.LocalCommitment.CommitFee,
+		commit.LocalBalance.ToSatoshis() + commit.CommitFee,
 	)
 	maxFee := feeBalance * maxAllocation
 
 	// Ensure the fee rate doesn't dip below the fee floor.
+	_, weight := lc.availableBalance()
 	maxFeeRate := maxFee / (float64(weight) / 1000)
 	return chainfee.SatPerKWeight(
 		math.Max(maxFeeRate, float64(chainfee.FeePerKwFloor)),
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -4623,6 +4623,12 @@ func (lc *LightningChannel) InitNextRevocation(revKey *btcec.PublicKey) error {
 // The additional openKey argument corresponds to the incoming CircuitKey of the
 // committed circuit for this HTLC. This value should never be nil.
 //
+// Note that AddHTLC doesn't reserve the HTLC fee for future payment (like
+// AvailableBalance does), so one could get into the "stuck channel" state by
+// sending dust HTLCs.
+// TODO(halseth): fix this either by using additional reserve, or better commit
+// format. See https://github.com/lightningnetwork/lightning-rfc/issues/728
+//
 // NOTE: It is okay for sourceRef to be nil when unit testing the wallet.
 func (lc *LightningChannel) AddHTLC(htlc *lnwire.UpdateAddHTLC,
 	openKey *channeldb.CircuitKey) (uint64, error) {
@@ -6025,6 +6031,8 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 		lc.localUpdateLog.logIndex)
 
 	// Calculate our available balance from our local commitment.
+	// TODO(halseth): could reuse parts validateCommitmentSanity to do this
+	// balance calculation, as most of the logic is the same.
 	//
 	// NOTE: This is not always accurate, since the remote node can always
 	// add updates concurrently, causing our balance to go down if we're
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -6049,7 +6049,7 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 	// Compute the current balances for this commitment. This will take
 	// into account HTLCs to determine the commit weight, which the
 	// initiator must pay the fee for.
-	ourBalance, _, commitWeight, filteredView, err := lc.computeView(
+	ourBalance, theirBalance, commitWeight, filteredView, err := lc.computeView(
 		view, false, false,
 	)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


halseth:
Unit tests not yet pushed

htlcswitch/interfaces.go
```diff
@@ -117,12 +117,15 @@ type ChannelLink interface {
...
-	Bandwidth() lnwire.MilliSatoshi
+	// Bandwidth returns the total amount that can flow through the channel
+	// link at this given instance. The amt parameter should be set
+	// non-zero to evaluate whether the link can carry this amt as part of
+	// its bandwidth. The value returned is expressed in millisatoshi and
+	// can be used by callers when making forwarding decisions to determine
+	// if a link can accept an HTLC of this size. This takes into account
+	// any forwarded but un-cleared HTLCs, and any HTLCs which have been
+	// set to the over flow queue.
+	Bandwidth(lnwire.MilliSatoshi) lnwire.MilliSatoshi
```
joostjager: Does it need this parameter, or should the calculation just return the highest amt that we can send without getting into trouble?

halseth: That sounds like a much better way of solving it indeed. Will rework the PR to try that.

halseth:
This is ready for review. (ignore failing travis for now, need to amend unit tests that have expectations about available balance)

lnwallet/channel_test.go
```diff
@@ -5801,6 +5801,58 @@ func TestChanReserve(t *testing.T) {
...
+	// Bob should refuse to add this HTLC, since he realizes it will create
+	// an invalid commitment.
+	_, err = bobChannel.AddHTLC(htlc, nil)
+	if err != ErrBelowChanReserve {
+		t.Fatalf("expected ErrBelowChanReserve, instead received: %v", err)
+	}
+
+	// Same for Alice, she should refuse to accept a state transtion with
+	// this second HTLC.
+	if _, err := aliceChannel.ReceiveHTLC(htlc); err != nil {
```
Roasbeef: Perhaps we should also check here when Alice attempts to add the HTLC? An earlier check means the link can be notified of this failure condition, and cancel back the incoming HTLC.

lnwallet/channel.go
```diff
@@ -4604,15 +4604,28 @@ func (lc *LightningChannel) AddHTLC(htlc *lnwire.UpdateAddHTLC,
...
 	err := lc.validateCommitmentSanity(
 		remoteACKedIndex, lc.localUpdateLog.logIndex, true, pd,
 	)
 	if err != nil {
 		return 0, err
 	}
 
+	// We must also check whether it can be added to our own commitment
+	// transaction, or the remote node will refuse to sign.
+	err = lc.validateCommitmentSanity(
```
Roasbeef: 👍

lnwallet/channel_test.go
```diff
@@ -5851,6 +5851,41 @@ func TestChanReserveRemoteInitiator(t *testing.T) {
...
+		true,
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cleanUp()
+
+	// The amount of the HTLC should be above Alice's dust limit and below
+	// Bob's dust limit.
+	htlcSat := (btcutil.Amount(500) + htlcTimeoutFee(
```
Roasbeef: Where's this 500 coming from? Can we use an existing constant for p2wkh dust here instead?

lnwallet/channel.go
```diff
@@ -5969,15 +5969,43 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 	htlcView := lc.fetchHTLCView(remoteACKedIndex,
 		lc.localUpdateLog.logIndex)
 
-	// Then compute our current balance for that view.
-	ourBalance, _, commitWeight, filteredView :=
-		lc.computeView(htlcView, false, false)
+	// Calculate our available balance from our local commitment.
+	// NOTE: this is not always accurate, since the remote node can always
```
Roasbeef: Style nit: should have a newline before the note section, and should start with a capital letter.

lnwallet/channel.go
```diff
@@ -5996,16 +5996,49 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 	)
 	feePerKw := filteredView.feePerKw
 
+	// Set the dustlimit based on who's commitment we are looking at.
+	dustlimit := lnwire.NewMSatFromSatoshis(
```
Roasbeef: This always uses the dust value of the local channel rather than selecting the proper value based on the commitment we're looking at.

lnwallet/channel.go
```diff
@@ -5996,16 +5996,49 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
...
+
 	// If we are the channel initiator, we must to subtract the commitment
 	// fee from our available balance.
 	if lc.channelState.IsInitiator {
-		ourBalance -= baseCommitFee
+		switch {
+		// If our balance is high enough to pay for a non-dust HTLC and
+		// the resulting commitment fee, subtract this commitment fee
+		// to reflect what we have left for such payments.
+		case ourBalance >= htlcCommitFee+nonDustHtlcAmt:
```
Roasbeef: Can we use `htlcIsDust` here directly?

lnwallet/channel.go
```diff
@@ -5996,16 +5996,49 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
...
+	// Also calculate the commitment fee in the case where we would add
+	// another HTLC large enough to create a new output on the commitment.
+	htlcCommitFee := lnwire.NewMSatFromSatoshis(
+		feePerKw.FeeForWeight(commitWeight + input.HTLCWeight),
+	)
+
+	// For an extra HTLC fee to be paid on our commitment, the HTLC must be
+	// large enough to make a non-dust HTLC timeout transaction.
+	htlcFee := lnwire.NewMSatFromSatoshis(
+		htlcTimeoutFee(feePerKw),
```
Roasbeef: When we compute for the remote party's commitment, we need to use the `htlcSuccessFee` here. I think we can actually just use `htlcIsDust` directly which accounts for the relative change in fee depending on the commitment we're factoring in.

lnwallet/channel_test.go
```diff
@@ -4693,6 +4704,55 @@ func TestChanAvailableBandwidth(t *testing.T) {
...
+	)
+	htlcFee := lnwire.NewMSatFromSatoshis(
+		feeRate.FeeForWeight(input.HTLCWeight),
+	)
+	commitFee := lnwire.NewMSatFromSatoshis(
+		aliceChannel.channelState.LocalCommitment.CommitFee,
+	)
+
+	// Helper method to check the current reported balance.
+	checkBalance := func(t *testing.T, expBalance lnwire.MilliSatoshi) {
```
Roasbeef: Do we really need this function given that it's used in only a single location below?

lnwallet/channel_test.go
```diff
@@ -4745,12 +4754,87 @@ func TestChanAvailableBalanceNearHtlcFee(t *testing.T) {
...
+	// nonDustHtlc left after paying the commit fee and htlc fee.
+	htlcAmt := aliceBalance - (commitFee + htlcFee + nonDustHtlc)
+	sendHtlc(htlcAmt)
+
+	// Now the real balance left will be nonDustHtlc+commitfee+htlcfee. But
+	// since any htlc larger than the dustlimit will have to reserve
+	// htlcfee to be added (and the commitfee must always be reserved), the
+	// available balance will just be nonDustHtlc.
+	expBalance = nonDustHtlc
+	checkBalance(t, expBalance)
```
Roasbeef: Scratch my earlier comment about this function ;)

lnwallet/channel.go
```diff
@@ -6033,6 +6033,13 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 		case ourBalance >= htlcCommitFee+nonDustHtlcAmt:
 			ourBalance -= htlcCommitFee
 
+		// We have enough to pay the non-dust HTLC, but not enough to
+		// cover the resulting fee. Therefore we set our available
+		// balance slightly below the non-dust amount to allow HTLCs up
+		// to that size.
+		case ourBalance >= baseCommitFee+nonDustHtlcAmt:
+			ourBalance = nonDustHtlcAmt - 1
```
Roasbeef: What's the rationale for the minus one here? Also why aren't we _subtracting_ the base fee _plus_ the amount of funds that need to enter limbo for the HTLC?

lnwallet/channel.go
```diff
@@ -5974,9 +5974,19 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
...
+		htlcView, false,
 	)
 
+	// Do the same caluclation from the remote commitment point of view.
+	ourRemoteCommitBalance, _ := lc.availableCommitmentBalance(
+		htlcView, true,
+	)
+
+	// Return which ever balance is lowest.
+	if ourRemoteCommitBalance < ourLocalCommitBalance {
```
Roasbeef: Why return the lowest instead of the proper balance depending on the context?

lnwallet/channel.go
```diff
@@ -6068,6 +6068,15 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView,
...
 			ourBalance -= baseCommitFee
 		}
+	} else {
+		// If we're not the initiator, we must check whether the remote
+		// has enough balance to pay for the fee of our HTLC. If the
+		// they cannot pay the fee if we add another non-dust HTLC, set
+		// our available balance just below the non-dust amount, to
+		// avoid attempting such large HTLCs.
+		if theirBalance < htlcCommitFee && ourBalance >= nonDustHtlcAmt {
+			ourBalance = nonDustHtlcAmt - 1
```
Roasbeef: Same question re subtracting vs replacing applies here.

lnwallet/channel.go
halseth: If we send `>= nonDustHtlcAmt` we will need to pay the extra htlc fee, which we cannot afford.

`nonDustHtlcAmt-1` is the largest amount we can send without having to pay this fee.

lnwallet/channel.go
halseth: Because the HTLC will need to be considered from both sides, as it will be added on both coimmitments.

lnwallet/channel.go
halseth: Changes in later commit

lnwallet/channel.go
halseth: Changes in later commit. 

Will check if I can use `htlcIsDust`

lnwallet/channel_test.go
halseth: You mean canceling the HTLC  immediately instead of waiting for the state transition? Not sure what we would gain from that, IIUC if Bob sends such an HTLC that is a protocol violation so we should fail the channel regardless. 

Would be a simple addition though, just calling `validateCommitmentSanity` in `ReceiveHTLC` I believe. let me know what you think would be best.

lnwallet/channel.go
halseth: Looks like we cannot use `htlcIsDust` since it expects the `htlcAmount` as an argument, while we want the reverse: get the minimum sized HTLC that is non-dust.

lnwallet/channel.go
halseth: See above comment.

lnwallet/channel_test.go
halseth: Added an explanation to the comment: https://github.com/lightningnetwork/lnd/pull/3691/commits/76455942017c29c03d8f3b1f2071643c93e1a39f

lnwallet/channel.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3691/commits/6389d2968d636c82d630e7f38e8ef602002acd49

lnwallet/channel.go
```diff
@@ -5985,21 +5995,26 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 // account for sending HTLCs of different sizes, it will take into account
 // whether the HTLCs will be manifested on the commitment, increasing the
 // commitment fee we must pay as an initiator, eating into our balance.
-func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
-	lnwire.MilliSatoshi, int64) {
+func (lc *LightningChannel) availableCommitmentBalance(view *htlcView,
+	remoteChain bool) (lnwire.MilliSatoshi, int64) {
```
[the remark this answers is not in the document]
joostjager: I am wondering if it all would be more readable if this function was split into two, even if there would be some duplication. One for the remote and one for the local commit. It is again hard to interpret the terms from the right perspective in a single function.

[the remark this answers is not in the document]
joostjager: Maybe it is ok, because `computeView` uses fixed perspectives?

lnwallet/channel.go
```diff
@@ -6019,26 +6034,50 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
...
 			ourBalance -= baseCommitFee
 		}
+
+	// If we're not the initiator, we must check whether the remote has
+	// enough balance to pay for the fee of our HTLC. If the they cannot
+	// pay the fee if we add another non-dust HTLC, set our available
+	// balance just below the non-dust amount, to avoid attempting such
+	// large HTLCs.
+	case theirBalance < htlcCommitFee && ourBalance >= nonDustHtlcAmt:
+		ourBalance = nonDustHtlcAmt - 1
```
joostjager: I know it is 'correct', but won't it be weird that your balance can go down from a high amount to almost zero if the _other_ party sends you money and brings their balance close to the limit?

Do we need to distinguish between balance (the money that you actually have, including the reserve) and amount sendable?

lnwallet/channel_test.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3691/commits/1617dc769c17f7f489aa7fdd5fe408caccb45b8e

lnwallet/channel.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3691/commits/d47c707964d0f0688a11f67a84e07494687117de

lnwallet/channel_test.go
```diff
@@ -5851,6 +5851,41 @@ func TestChanReserveRemoteInitiator(t *testing.T) {
 		t.Fatalf("expected ErrBelowChanReserve, instead received: %v", err)
 	}
 
+	// Now we'll test the case where Alice is adding an HTLC which Bob
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3691/commits/94dfc9d9bd304057e38ba10f8788fbcd3ed30c2b

lnwallet/channel.go
halseth: I did split them with the intention that the added case goes together with the testcase and commit message to explain what it achieves.

lnwallet/channel.go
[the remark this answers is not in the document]
halseth: There are really only two conditionals, the dustlimit and the HTLC fee. Would it be worth duplicatiing the rest to get rid of that?

lnwallet/channel.go
halseth: So this method is really used as "amount sendable", as it is called from the link when finding the available bandwidth. It is not used to expose the balance to the user in any way. Maybe changing the method name would make it more clear?

joostjager: Ah yes, that is true. I like `amountSendable` as a name, to make things a bit more clear. But non-blocking.

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: I think it is also ok to leave it as is. Maybe some function can be extracted that takes the `remote` boolean, but not sure how that would work out.

joostjager:
Tested PR and can confirm that I now get a single `insufficient local balance` failure instead of a never-ending loop.

lnwallet/channel.go
```diff
@@ -5956,15 +5972,116 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
...
+	switch {
+
+	// If we are the channel initiator, we must to subtract the commitment
+	// fee from our available balance.
+	case lc.channelState.IsInitiator:
+		switch {
+		// If our balance is high enough to pay for a non-dust HTLC and
+		// the resulting commitment fee, subtract this commitment fee
+		// to reflect what we have left for such payments.
+		case ourBalance >= htlcCommitFee+nonDustHtlcAmt:
```
joostjager: During testing I found that `ourBalance` still includes the reserve here. So our balance is always high enough to pay for a dust htlc.

joostjager: If in this case and the default case below, we'd just return `0` as available balance, could we prevent users from sending out micropayments that get their channel into an unusable state?

halseth: Good catch! Yeah, this was wrong all along... Now moved reserve calculation into `availableBalance`

halseth: Goo idea, done for the initiator!

I still keep this calculation around for the non-initiator, since we letting them send dust HTLCs moves us away from the unusable state.

halseth:
Pushed new update where channel reserve calculation is moved into `availableBalance` and we now report a 0 balance for the initiator  if only the HTLC fee is left, to avoid getting into the unusable channel situation. PTAL.

lnwallet/channel_test.go
```diff
@@ -4714,6 +4721,129 @@ func TestChanAvailableBandwidth(t *testing.T) {
...
+	// nonDustHtlc left after paying the commit fee and htlc fee.
+	htlcAmt := aliceBalance - (commitFee + aliceReserve + htlcFee + aliceNonDustHtlc)
+	sendHtlc(htlcAmt)
+
+	// Now the real balance left will be
+	// nonDustHtlc+commitfee+aliceReserve+htlcfee. The available balance
+	// reported will just be nonDustHtlc, since the rest of the balance is
+	// reserved.
+	expBalance = aliceNonDustHtlc
+	checkBalance(t, expBalance)
```
joostjager: don't use `expBalance` variable but pass it into `checkBalance` directly?

lnwallet/channel.go
```diff
@@ -6039,6 +6070,14 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
 		}
 
 		ourBalance -= htlcCommitFee
+
+	// If we're not the initiator, we must check whether the remote has
+	// enough balance to pay for the fee of our HTLC. If the they cannot
+	// pay the fee if we add another non-dust HTLC, set our available
+	// balance just below the non-dust amount, to avoid attempting HTLCs
+	// larger than this size.
+	case theirBalance < htlcCommitFee && ourBalance >= nonDustHtlcAmt:
```
joostjager: I think the readability of this function isn't optimal. There are two relatively independent cases: initiator and non-initiator. But several variables are set in the parent scope, making it less clear what is used where. Maybe converting this `switch` to a classic `if initiator then .. else ..` and defining the specific vars in the if blocks is better.

lnwallet/channel_test.go
```diff
@@ -4802,20 +4818,26 @@ func TestChanAvailableBalanceNearHtlcFee(t *testing.T) {
...
 	}
 
 	// Balance should start out equal to half the channel capacity minus
 	// the commitment fee Alice must pay and the channel reserve. In
 	// addition the HTLC fee will be subtracted fromt the balance to
 	// reflect that this value must be reserved for any payment above the
 	// dust limit.
-	expBalance := aliceBalance - commitFee - aliceReserve - htlcFee
-	checkBalance(t, expBalance)
+	expAliceBalance := aliceBalance - commitFee - aliceReserve - htlcFee
```
joostjager: name this `expAliceBalance` already in the earlier commit, to make the delta smaller?

lnwallet/channel_test.go
```diff
@@ -4824,24 +4846,53 @@ func TestChanAvailableBalanceNearHtlcFee(t *testing.T) {
 	// nonDustHtlc+commitfee+aliceReserve+htlcfee. The available balance
 	// reported will just be nonDustHtlc, since the rest of the balance is
 	// reserved.
-	expBalance = aliceNonDustHtlc
-	checkBalance(t, expBalance)
+	expAliceBalance = aliceNonDustHtlc
+	expBobBalance = bobBalance - bobReserve
+	checkBalance(t, expAliceBalance, expBobBalance)
```
joostjager: These expectations refer to variables that are also updated themselves. I think it is more robust to assert against literal values here, to be sure that no mistake in the test code itself can cause the wrong thing being asserted.

lnwallet/channel_test.go
```diff
@@ -4824,24 +4846,53 @@ func TestChanAvailableBalanceNearHtlcFee(t *testing.T) {
...
+
+	// Even though Alice has a reported balance of 0, this is because we
+	// try to avoid getting into the position where she cannot pay the fee
+	// for Bob adding another HTLC. This means she actually _has_ some
+	// balance left, and we now force the channel into this situation by
+	// sending yet another HTLC. In practice this can also happen if a fee
+	// update eats into Alice's balance.
+	htlcAmt = 1
+	sendHtlc(htlcAmt)
```
joostjager: Hmm is it good that this is still possible if there is no amount sendable anymore?

lnwallet/channel.go
```diff
@@ -5981,9 +5981,19 @@ func (lc *LightningChannel) availableBalance() (lnwire.MilliSatoshi, int64) {
 	// add updates concurrently, causing our balance to go down if we're
 	// the initiator, but this is a problem on the protocol level.
 	ourLocalCommitBalance, commitWeight := lc.availableCommitmentBalance(
-		htlcView,
+		htlcView, false,
 	)
 
+	// Do the same calculation from the remote commitment point of view.
+	ourRemoteCommitBalance, _ := lc.availableCommitmentBalance(
```
joostjager: It looks like `availableBalance` does many of the things that `validateCommitmentSanity` does too. Is it possible to unify both calls by just having `validateCommitmentSanity` return the amount sendable too? That way we can also be more sure that `availableBalance` really returns the maximum that `AddHtlc` would accept.

htlcswitch/link.go
```diff
@@ -2127,26 +2127,21 @@ func (l *channelLink) ChanID() lnwire.ChannelID {
...
+	overflowBandwidth := l.overflowQueue.TotalHtlcAmount()
+	if channelBandwidth < overflowBandwidth {
 		return 0
 	}
 
-	// Else the amount that is available to flow through the link at this
-	// point is the available balance minus the reserve amount we are
-	// required to keep as collateral.
-	return linkBandwidth - reserve
+	return channelBandwidth - overflowBandwidth
```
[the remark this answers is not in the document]
halseth: `AvailableBalance` will return the balance available after subtracting `htlcCommitFee` (regular commitment fee + fee for adding one extra htlc) and the channel reserve. This means that we already have subtracted the required `1*htlcFee` buffer. So think we should be good?

But you are right, it gets a bit iffy with the overflow queue. The problem is that the overflow queue by definition contains HTLCs of some value, but doesn't account for the fee it would take to add these to the commitment. This calculation wouldn't be an exact science anyway, since for HTLCs from the overflow queue to be added, we first would have to remove some of the other HTLCs.

I think the best way to estimate how much balance we have available at that point would be to subtract the value in the overflow queue plus `len(overflowQueue)*htlcFee`. This will not always be correct, but conservative. wdyt?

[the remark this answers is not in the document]
joostjager: Maybe you don't need to subtract `len * htlcFee`, because the htlcs can also all be send and settled one after the other. The `htlcFee` will be returned every time. I was thinking to first subtract the sum of the overflow queue and only then see that is left and compare that with `htlcFee`.

halseth:
Won't that be equivalent to what we do here? (Only diff in which order we
subtract)

htlcswitch/link.go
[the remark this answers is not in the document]
joostjager: Discussed offline. It is indeed equivalent

lnwallet/channel_test.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3691/commits/1354a838e213c89b73cdc288414d87c3b02843e9

lnwallet/channel.go
halseth: Is this better? https://github.com/lightningnetwork/lnd/pull/3691/commits/ae2a41b3464133cd80d38021c48c172796f4c3a1

lnwallet/channel_test.go
halseth: It would be changing `AddHTLC` to not allow sending HTLCs even though there is balance left. IMO it is a nice upper layer heuristic to not attempt this by reporting this balance not being available. We can revisit if this becomes a hard protocol requirement,  re https://github.com/lightningnetwork/lightning-rfc/issues/728

lnwallet/channel_test.go
```diff
@@ -5815,6 +5815,66 @@ func TestChanReserve(t *testing.T) {
...
+	// high is unrealistic, a channel can easiliy get into a situation
+	// where the initiator cannot pay for the fee of any more HTLCs.
+	commitFee := aliceChannel.channelState.LocalCommitment.CommitFee
+	aliceMinReserve := 5*btcutil.SatoshiPerBitcoin - commitFee
+
+	aliceChannel.channelState.LocalChanCfg.ChanReserve = aliceMinReserve
+	bobChannel.channelState.RemoteChanCfg.ChanReserve = aliceMinReserve
+
+	// Now let Bob attempt to add an HTLC of 0.1 BTC. He has plenty of
+	// money, but Alice, which is the initiator, cannot afford any more
```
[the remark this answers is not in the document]
halseth: He has 5 BTC. Don't know about you, but I think that's plenty of money :)

lnwallet/channel_test.go
halseth: With next commit it is nice to easily distinguish alice and bob imo

lnwallet/channel_test.go
halseth: Don't see how that would be better

lnwallet/channel.go
halseth: Added TODO.

lnwallet/channel_test.go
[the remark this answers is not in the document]
joostjager: Alice also has plenty of money with that definition. I mean, it is about amount spendable?

lnwallet/channel_test.go
joostjager: It is more difficult to see that this test asserts the right behavior.

lnwallet/channel_test.go
joostjager: It could lead to getting caught in the danger zone if there are concurrent payments. Both payments used the same reported `availableBalance`, which was enough at that point, but when executed the second payment pushes us below the threshold. If `AddHTLC` would fail, this wouldn't happen and a new path for the second payment could be searched for.

But this is arguably a rare condition, especially because we plan to do non-concurrent pathfinding for mpp. Maybe it is worth a comment near the `AddHTLC` implementation.

lnwallet/channel.go
joostjager: Yes

lnwallet/channel_test.go
joostjager: Btw, does this mean that you can still go below the threshold with SendToRoute?

htlcswitch/link.go
```diff
@@ -2127,26 +2127,21 @@ func (l *channelLink) ChanID() lnwire.ChannelID {
 //
 // NOTE: Part of the ChannelLink interface.
 func (l *channelLink) Bandwidth() lnwire.MilliSatoshi {
+	// Get the balance available on the channel for new HTLCs. This takes
```
Roasbeef: In commit message: subratcion -> subtraction

htlcswitch/link.go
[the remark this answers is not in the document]
Roasbeef: One related item is that we don't attempt to manage this overflow queue at all. We assume that things will eventually drain out, so we don't require any sort of active queue management. However, one could argue that if we emptied out this queue (by cancelling the HTLCs backwards) during times of network congestion, then we free up the sending clients (they can retry another link as they assume it's either congestion or a lack of available bandwidth) and also tend to an empty queue. 

There's some active research in this area with papers like the Spider network we could draw from in an attempt to address this when we decide to tackle it. This may also partially mitigate channel jamming attacks as we'll cancel back those HTLCs sooner, which frees up HTLC slots in the _incoming_ channel.

lnwallet/channel.go
```diff
@@ -6055,20 +6055,30 @@ func (lc *LightningChannel) availableCommitmentBalance(view *htlcView) (
...
 
-	// Given the commitment weight, find the commitment fee in case of no
-	// added HTLC output.
+	// Calculate the commitment fee in the case where we would add another
+	// HTLC to the commitment, as only the balance remaining after this fee
+	// has been paid is actually available for sending.
 	feePerKw := filteredView.feePerKw
-	baseCommitFee := lnwire.NewMSatFromSatoshis(
-		feePerKw.FeeForWeight(commitWeight),
+	htlcCommitFee := lnwire.NewMSatFromSatoshis(
```
Roasbeef: Similar to the `predictAdded` field in `validateCommitmentSanity`, shouldn't we only conditionally add this weight if we're attempting to look ahead by one HTLC to see the expected balance _after_ this HTLC is added.

Roasbeef: Or is the assumption that this method is only used atm by the switch when it _is_ attempting to add a new HTLC to the commitment transaction?

Roasbeef: Non blocking, more of a side point, but if we _do_ add this param, then we can use this to display the current balance of a channel in `ListChannels` and `WalletBalance`, if we opt to just mask the concept of a channel reserver all together. If we go this route, then at least the user's wallet will show zero satoshis, rather than showing a balance which can't actually be sent/received (though there're other implications).

lnwallet/channel_test.go
Roasbeef: `SendToRoute` will still eventually send an HTLC through the link, so none of these checks are bypassed.

lnwallet/channel.go
halseth: > Or is the assumption that this method is only used atm by the switch when it _is_ attempting to add a new HTLC to the commitment transaction?

Yeah, that is the assumption, that it returns the "balance available for HTLCs". And that has to include the HTLC fee if we are the initiator.

lnwallet/channel_test.go
halseth: `SendToRoute` will also be forwarded by the switch, and `Bandwidth` is called here: https://github.com/lightningnetwork/lnd/blob/92b79f6b6abc7e027ee1d7253b7c4e67c0090476/htlcswitch/link.go#L2339

lnwallet/channel_test.go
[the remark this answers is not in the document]
halseth: Expanded the comment. Hopefully more clear now.

PR #3693, opened 2019-11-08
migrate to tlv for invoice body

cfromknecht:
Builds on #3685 

This PR migrates our invoices to use TLV for their bodies. In the process we add feature vectors and payment addrs to make way for our mpp series. The unused receipt field is ultimately dropped from the db and the rpc calls. Finally, invoices are now populated with an empty set of features, which will not appear on payment requests until a non-zero number of bits are initially set.


lnrpc/rpc.proto
```diff
@@ -2515,6 +2515,13 @@ message PayReq {
     string fallback_addr = 8 [json_name = "fallback_addr"];
     int64 cltv_expiry = 9 [json_name = "cltv_expiry"];
     repeated RouteHint route_hints = 10 [json_name = "route_hints"];
+    repeated Feature features = 11 [json_name = "features"];
+}
+
+message Feature {
+        uint32 bit = 1 [json_name = "bit"];
+        string name = 2 [json_name = "name"];
+        bool is_required = 3 [json_name = "is_required"];
```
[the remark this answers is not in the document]
joostjager: Looks like feature bits can be its own pr

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,475 @@
...
+
+	return htlcs, nil
+}
+
+// serializeInvoice serializes an invoice to a writer.
+//
+// Note: this function is in use for a migration. Before making changes that
+// would modify the on disk format, make a copy of the original code and store
+// it with the migration.
+func serializeInvoice(w io.Writer, i *Invoice) error {
```
halseth: This code does look new. It should be added to the next commit to make it clear what code has just been copied and what is new (and hence needs a more close review)

channeldb/migration12/migration.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+var emptyFeatures = lnwire.NewFeatureVector(nil, nil)
+
+// MigrateInvoiceTLV migrates all existing invoice bodies over to be serialized
+// in a single TLV stream. In the process, we drop the Receipt field and add
+// PaymentAddr and Features to the invoice Terms.
+func MigrateInvoiceTLV(tx *bbolt.Tx) error {
```
halseth: this should have a test.

cmd/lncli/commands.go
```diff
@@ -2561,10 +2561,6 @@ var addInvoiceCommand = cli.Command{
 			Usage: "a description of the payment to attach along " +
 				"with the invoice (default=\"\")",
 		},
-		cli.StringFlag{
-			Name:  "receipt",
-			Usage: "an optional cryptographic receipt of payment",
-		},
```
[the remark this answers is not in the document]
cfromknecht: done

rpcserver.go
```diff
@@ -3525,7 +3525,6 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 
 	addInvoiceData := &invoicesrpc.AddInvoiceData{
 		Memo:            invoice.Memo,
-		Receipt:         invoice.Receipt,
```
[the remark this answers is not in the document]
cfromknecht: done

channeldb/invoices.go
```diff
@@ -154,8 +164,19 @@ type ContractTerm struct {
...
+
+	// Expiry defines how long after creation this invoice should expire.
+	Expiry time.Duration
+
+	// PaymentAddr is a randomly generated value include in the MPP record
+	// by the sender to prevent probing of the receiver.
+	PaymentAddr [32]byte
+
+	// Features is the feature vectors advertised on the payment request.
+	Features *lnwire.FeatureVector
```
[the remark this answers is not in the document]
cfromknecht: field restructuring is now is now in a separate commit

channeldb/invoices.go
```diff
@@ -154,8 +164,19 @@ type ContractTerm struct {
...
+	// FinalCltvDelta is the minimum required number of blocks before htlc
+	// expiry when the invoice is accepted.
+	FinalCltvDelta int32
+
+	// Expiry defines how long after creation this invoice should expire.
+	Expiry time.Duration
+
+	// PaymentAddr is a randomly generated value include in the MPP record
+	// by the sender to prevent probing of the receiver.
+	PaymentAddr [32]byte
```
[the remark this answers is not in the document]
cfromknecht: i'd rather add these fields during the migration so that users can't accidentally revert to a commit where we don't process them. the downgrade logic in opening channeldb will prevent this if they are added here

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,475 @@
...
+		addIndexType    tlv.Type = 9
+		settleIndexType tlv.Type = 11
+		preimageType    tlv.Type = 13
+		valueType       tlv.Type = 15
+		cltvDeltaType   tlv.Type = 17
+		expiryType      tlv.Type = 19
+		paymentAddrType tlv.Type = 21
+		featuresType    tlv.Type = 23
+		stateType       tlv.Type = 25
+		amtPaidType     tlv.Type = 27
```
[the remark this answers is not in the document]
cfromknecht: it's purely cosmetic, tho likely that #3701 will be merged before at which point we can reevaluate the constants

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,475 @@
...
+	// A set of tlv type definitions used to serialize invoice htlcs to the
+	// database.
+	chanIDType       tlv.Type = 1
+	htlcIDType       tlv.Type = 3
+	amtType          tlv.Type = 5
+	acceptHeightType tlv.Type = 7
+	acceptTimeType   tlv.Type = 9
+	resolveTimeType  tlv.Type = 11
+	expiryHeightType tlv.Type = 13
+	stateType        tlv.Type = 15
```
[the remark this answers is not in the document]
cfromknecht: given that we do have additional data, i don't see how we could intermingle them without running into issues. 

> Maybe passing through every custom tlv fields in the 2^16+ range works and then doing a custom mapping for the spec fields that we know of?

are you referring to the tlvs presented in the onion payload? for now those are kept in the onion, but if we wanted to persist them separately i imagine adding another field for all the experimental types

.golangci.yml
```diff
@@ -12,6 +12,7 @@ run:
 
   skip-dirs:
     - channeldb/migration_01_to_11
+    - channeldb/migration12
```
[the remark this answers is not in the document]
cfromknecht: ended up adding this to the list of linted packages, fixed any other linter errors and added the `nolint: dupl` directives for the methods in this package

lnwire/features.go
```diff
@@ -349,10 +348,10 @@ func (fv *FeatureVector) UnknownRequiredFeatures() []FeatureBit {
 // as much.
 func (fv *FeatureVector) Name(bit FeatureBit) string {
 	name, known := fv.featureNames[bit]
-	if !known {
-		name = "unknown"
+	if known {
+		return name
 	}
-	return fmt.Sprintf("%s(%d)", name, bit)
+	return "unknown"
```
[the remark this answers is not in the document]
cfromknecht: will update in separate pr

lnrpc/rpc.proto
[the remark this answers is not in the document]
cfromknecht: i kept the changes to populate an empty feature vector, but will move the rest of it to another pr

lnrpc/invoicesrpc/invoices_server.go
```diff
@@ -247,13 +247,14 @@ func (s *Server) AddHoldInvoice(ctx context.Context,
...
-		DefaultCLTVExpiry: s.cfg.DefaultCLTVExpiry,
-		ChanDB:            s.cfg.ChanDB,
+		AddInvoice:         s.cfg.InvoiceRegistry.AddInvoice,
+		IsChannelActive:    s.cfg.IsChannelActive,
+		ChainParams:        s.cfg.ChainParams,
+		NodeSigner:         s.cfg.NodeSigner,
+		MaxPaymentMSat:     s.cfg.MaxPaymentMSat,
+		DefaultCLTVExpiry:  s.cfg.DefaultCLTVExpiry,
+		ChanDB:             s.cfg.ChanDB,
+		GenInvoiceFeatures: s.cfg.GenInvoiceFeatures,
```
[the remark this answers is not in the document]
cfromknecht: they are constant, but i went with the approach of generating a new vector for each invoice to avoid any possibility of mutations

channeldb/migration12/invoices.go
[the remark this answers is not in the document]
joostjager: so you mean a single tlv field that contains another tlv array?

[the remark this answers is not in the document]
joostjager: Yes, indeed referring to the custom fields in the onion payload

channeldb/invoices.go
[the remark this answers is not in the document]
joostjager: Does this mean that we are going to increase the db version with every tlv field that we add, a no-op migration>

channeldb/migration12/invoices.go
```diff
@@ -105,12 +112,6 @@ type Invoice struct {
 	// or any other message which fits within the size constraints.
 	Memo []byte
 
-	// Receipt is an optional field dedicated for storing a
-	// cryptographically binding receipt of payment.
-	//
-	// TODO(roasbeef): document scheme.
-	Receipt []byte
-
```
joostjager: Wasn't `Receipt` already gone?

channeldb/migration12/invoices.go
```diff
@@ -345,71 +346,90 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
...
-	if err := wire.WriteVarBytes(w, 0, i.PaymentRequest[:]); err != nil {
-		return err
-	}
-
-	if err := binary.Write(w, byteOrder, i.FinalCltvDelta); err != nil {
-		return err
-	}
-
-	if err := binary.Write(w, byteOrder, int64(i.Expiry)); err != nil {
-		return err
```
joostjager: Maybe this shouldn't have been copied in the previous commit?

channeldb/migration12/invoices.go
```diff
@@ -345,71 +346,90 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
...
+		tlv.MakePrimitiveRecord(preimageType, &preimage),
+		tlv.MakePrimitiveRecord(valueType, &value),
+		tlv.MakePrimitiveRecord(cltvDeltaType, &cltvDelta),
+		tlv.MakePrimitiveRecord(expiryType, &expiry),
+		tlv.MakePrimitiveRecord(paymentAddrType, &i.Terms.PaymentAddr),
+		tlv.MakePrimitiveRecord(featuresType, &featureBytes),
+
+		// Invoice state.
+		tlv.MakePrimitiveRecord(stateType, &state),
+		tlv.MakePrimitiveRecord(amtPaidType, &amtPaid),
```
joostjager: Difficult to see whether this is exactly what will be added in a later commit in `channeldb`

channeldb/invoices.go
```diff
@@ -174,12 +191,6 @@ type Invoice struct {
 	// or any other message which fits within the size constraints.
 	Memo []byte
 
-	// Receipt is an optional field dedicated for storing a
-	// cryptographically binding receipt of payment.
-	//
-	// TODO(roasbeef): document scheme.
-	Receipt []byte
-
```
joostjager: Also not removed in first commit?

channeldb/invoices.go
```diff
@@ -312,16 +323,14 @@ func validateInvoice(i *Invoice) error {
...
-			len(i.Receipt))
-	}
 	if len(i.PaymentRequest) > MaxPaymentRequestSize {
 		return fmt.Errorf("max length of payment request is %v, length "+
 			"provided was %v", MaxPaymentRequestSize,
 			len(i.PaymentRequest))
 	}
+	if i.Terms.Features == nil {
+		return errors.New("invoice must have a feature vector")
+	}
```
joostjager: Can the addition of features be separated from the db changes?

channeldb/invoices.go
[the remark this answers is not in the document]
cfromknecht: i think it's the safest path

channeldb/migration12/invoices.go
cfromknecht: not from the invoice, it was still present on the data structure itself since the migration hadn't been performed. prior it was removed from the rpc and everywhere outside of channeldb

channeldb/invoices.go
cfromknecht: we could remove it sooner and instead write a blank receipt, but it is done here because here this is the commit that actually modifies the on-disk format. if we do it sooner then the existing serialization code can't reference any (in theory) existing receipts, and means that there would side effects if an invoice is decoded and deserialized before running the migration.

channeldb/invoices.go
cfromknecht: in what sense? another pr?

channeldb/migration12/invoices.go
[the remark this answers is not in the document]
cfromknecht: > so you mean a single tlv field that contains another tlv array?

possibly yes, or appending it after all the htlcs

channeldb/migration12/migration.go
cfromknecht: done

channeldb/migration12/invoices.go
cfromknecht: done

channeldb/invoices.go
joostjager: i thought separate commit. isolate addition of the feature from moving to tlv. but nvm, it doesn't really matter

channeldb/migration12/invoices.go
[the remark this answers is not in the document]
joostjager: ok, something to think about later.

channeldb/migration12/invoices.go
joostjager: But probably nothing we can do about it. Hopefully this is the last invoice migration

channeldb/migration12/migration_test.go
```diff
@@ -0,0 +1,306 @@
...
+		0x01, 0x03, 0x1b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x01, 0xa4, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x41,
+		0x01, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x08,
+		0x03, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x09,
+		0x05, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x64,
+		0x07, 0x04, 0x00, 0x00, 0x00, 0x58, 0x09, 0x08, 0x00, 0x13,
+		0xbc, 0xbf, 0x72, 0x4e, 0x1e, 0x00, 0x0b, 0x08, 0x00, 0x17,
+		0xaf, 0x4c, 0x22, 0xc4, 0x24, 0x00, 0x0d, 0x04, 0x00, 0x00,
+		0x23, 0x1d, 0x0f, 0x01, 0x02,
+	}
```
joostjager: This is indeed an idea too, to prevent even more serialize code.

channeldb/migtest/migtest.go
```diff
@@ -0,0 +1,94 @@
...
+		cdb.Close()
+		os.RemoveAll(tempDirName)
+	}
+
+	return cdb, cleanUp, nil
+}
+
+// ApplyMigration is a helper test function that encapsulates the general steps
+// which are needed to properly check the result of applying migration function.
+func ApplyMigration(t *testing.T,
```
joostjager: Ref this in `migration_01_to_11`? Maybe not...

channeldb/invoices.go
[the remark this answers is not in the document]
joostjager: Ok. It is easy to forget though. Should we add a NOTE comment to every list of db tlv type constants in the code base to remind future devs about this? Pre-existing, but maybe not worth the overhead of a separate pr

channeldb/invoices.go
joostjager: ok

channeldb/invoices.go
```diff
@@ -1121,7 +1121,7 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
 
 	for {
 		// Read the length of the tlv stream for this htlc.
-		var streamLen uint64
+		var streamLen int64
```
joostjager: typo in commit msg

joostjager:
Build generates a diff in `go.mod`

channeldb/migration12/invoices.go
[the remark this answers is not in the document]
joostjager: It has been merged, so the option is available to do cosmetics

channeldb/migtest/migtest.go
cfromknecht: in that package the after and before functions operate on `channeldb.DB`, where as here they operate on `bbolt.Tx`. some of those functions rely on db-level methods, so it is a bigger refactor. i made the decision here to only offer `bbolt.Tx` because it makes the closures more isolated, and hopefully less code to copy in future ones.

channeldb/migration12/migration_test.go
cfromknecht: yes this prevented me from having to copy in the old serialization code :)

channeldb/invoices.go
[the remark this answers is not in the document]
cfromknecht: done

channeldb/migration12/invoices.go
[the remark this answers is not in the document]
cfromknecht: done

cfromknecht:
> This is the ListInvoices diff.

Looked into this, the `channeldb.Invoice` struct was still using `invoice.Terms.State` for encode decode. Fixed in latest commit by pointing it to `i.State` and removing the old field.

> Build generates a diff in go.mod

also fixed, a `log.Error` in `migtest` caused it to import the prometheus log. it has not been replaced with `t.Fatalf`

subrpcserver_config.go
```diff
@@ -210,6 +212,9 @@ func (s *subRPCServerConfigs) PopulateDependencies(cc *chainControl,
 			subCfgValue.FieldByName("ChanDB").Set(
 				reflect.ValueOf(chanDB),
 			)
+			subCfgValue.FieldByName("GenInvoiceFeatures").Set(
+				reflect.ValueOf(genInvoiceFeatures),
+			)
```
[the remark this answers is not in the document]
cfromknecht: yes that sounds like a good idea, since we need it for the main rpc

channeldb/migtest/migtest.go
joostjager: sounds good

joostjager:
```
--- FAIL: TestSettleInvoice (0.01s)
    invoiceregistry_test.go:94: invoice must have a feature vector
```

channeldb/migration12/invoices.go
```diff
@@ -36,7 +36,22 @@ const (
...
+	addIndexType    tlv.Type = 4
+	settleIndexType tlv.Type = 5
+	preimageType    tlv.Type = 6
+	valueType       tlv.Type = 7
+	cltvDeltaType   tlv.Type = 8
+	expiryType      tlv.Type = 9
+	paymentAddrType tlv.Type = 10
+	featuresType    tlv.Type = 11
+	invStateType    tlv.Type = 12
+	amtPaidType     tlv.Type = 13
```
joostjager: Beautiful

channeldb/migration12/invoices.go
```diff
@@ -36,7 +36,22 @@ const (
 	acceptTimeType   tlv.Type = 9
 	resolveTimeType  tlv.Type = 11
 	expiryHeightType tlv.Type = 13
-	stateType        tlv.Type = 15
+	htlcStateType    tlv.Type = 15
```
joostjager: I don't think we want to use this migration where we write out all htlcs again to beautify these constant too?

channeldb/migration12/invoices.go
```diff
@@ -339,7 +361,73 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
...
+
+		// Invoice state.
+		tlv.MakePrimitiveRecord(invStateType, &state),
+		tlv.MakePrimitiveRecord(amtPaidType, &amtPaid),
+	)
+	if err != nil {
+		return err
+	}
+
+	b.Reset()
```
joostjager: Hmm, isn't this going to overwrite the underlying array of `featureBytes`? If not, maybe add a comment stating this

channeldb/invoices.go
```diff
@@ -96,14 +91,40 @@ const (
...
 	expiryHeightType tlv.Type = 13
-	stateType        tlv.Type = 15
+	htlcStateType    tlv.Type = 15
+
+	// A set of tlv type definitions used to serialize invoice bodiees.
+	//
+	// NOTE: A migration should be added whenever this list changes. This
+	// prevents against the database being rolled back to an older
+	// format where the surrounding logic might assume a different set of
+	// fields are known.
```
joostjager: :+1:
It is a no-op migration, hopefully that is clear.

channeldb/invoice_test.go
```diff
@@ -114,7 +114,7 @@ func TestInvoiceWorkflow(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to fetch invoice: %v", err)
 	}
-	if dbInvoice2.Terms.State != ContractSettled {
+	if dbInvoice2.State != ContractSettled {
```
joostjager: Commit message describes move of final delta and expiry, but that doesn't happen in this commit anymore

channeldb/invoices.go
cfromknecht: it is only a noop migration if the field is truly optional, in most cases that won't be the case i presume

channeldb/migration12/invoices.go
cfromknecht: we definitely can, but it adds nothing functionally. i'd rather save this for the next migration where we do for htlcs since it is purely cosmetic and at this point can only introduce more bugs

channeldb/invoices.go
joostjager: hmm, so in order to minimize migrations, what would we do with for example a new required int field? write a migration or define non-present as 0? i was hoping we could have fewer migrations now, but maybe not

channeldb/migration12/invoices.go
joostjager: agrred

channeldb/invoice_test.go
cfromknecht: good catch, restored

channeldb/migration12/invoices.go
cfromknecht: TIL! yes it does, tho fortunately the empty features serialize to zero bytes and so don't affect the migration. however, this was copied into the main serialization method, so that could have been an issue but i've modified both to use a distinct buffer.

cfromknecht: https://play.golang.org/p/kcNLH-34qH4

channeldb/invoices.go
cfromknecht: if it is required, idt there's a way around it. some things may be less critical and it would be okay to do an in-flight migration. in this case tho, these parameters affect invoice settling and i think qualify as required. i think it will have to be a case by case basis

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,392 @@
...
+
+// serializeInvoice serializes an invoice to a writer.
+//
+// Note: this function is in use for a migration. Before making changes that
+// would modify the on disk format, make a copy of the original code and store
+// it with the migration.
+//
+// nolint: dupl
+func serializeInvoice(w io.Writer, i *Invoice) error {
+	panic("unimplemented")
```
halseth: I think the method can just be removed, it is not used ion this commit.

channeldb/invoices.go
```diff
@@ -1141,11 +1141,9 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
...
 		}
 
-		streamBytes := make([]byte, streamLen)
-		if _, err := r.Read(streamBytes); err != nil {
-			return nil, err
-		}
-		streamReader := bytes.NewReader(streamBytes)
+		// Limit the reader so that it stops at the end of this htlc's
+		// stream.
+		htlcReader := io.LimitReader(r, streamLen)
```
halseth: is this in any hot-path?

channeldb/invoices.go
```diff
@@ -166,6 +166,13 @@ func (c ContractState) String() string {
 // the necessary conditions required before the invoice can be considered fully
 // settled by the payee.
 type ContractTerm struct {
+	// FinalCltvDelta is the minimum required number of blocks before htlc
+	// expiry when the invoice is accepted.
+	FinalCltvDelta int32
```
halseth: Should we use the opportunity to remove the `ContractTerm` struct? Doesn't seem it add much, it only adds to the confusion where to put the fields. 

As it is not serialized separately either, I would prefer having all these fields on the invoice instead.

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,572 @@
...
+	"time"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+const (
+	// MaxMemoSize is maximum size of the memo field within invoices stored
```
halseth: what have we gotten ourself into 😅

channeldb/invoices.go
```diff
@@ -892,71 +912,73 @@ func putInvoice(invoices, invoiceIndex, addIndex *bbolt.Bucket,
...
-	if err := binary.Write(w, byteOrder, i.FinalCltvDelta); err != nil {
-		return err
-	}
-
-	if err := binary.Write(w, byteOrder, int64(i.Expiry)); err != nil {
-		return err
-	}
-
-	birthBytes, err := i.CreationDate.MarshalBinary()
+	creationDateBytes, err := i.CreationDate.MarshalBinary()
```
halseth: didn't we want to import this method from the v12 migration to avoid the duplication?

channeldb/migration12/invoices.go
joostjager: Can we use this migration to also add the payment hash as a proper field here? Especially for hodl invoices, there is some awkwardness to get to that value.

joostjager: Idea came up in conversation with @bhandras

channeldb/migration12/invoices.go
cfromknecht: none of the code in this commit is used

channeldb/invoices.go
cfromknecht: not that i know of, but it will be as invoices start to have multiple htlcs or payments. might improve `listinvoices` performance slightly

channeldb/invoices.go
```diff
@@ -96,14 +91,40 @@ const (
...
 	acceptHeightType tlv.Type = 7
 	acceptTimeType   tlv.Type = 9
 	resolveTimeType  tlv.Type = 11
 	expiryHeightType tlv.Type = 13
-	stateType        tlv.Type = 15
+	htlcStateType    tlv.Type = 15
+
+	// A set of tlv type definitions used to serialize invoice bodiees.
+	//
+	// NOTE: A migration should be added whenever this list changes. This
```
Roasbeef: The migration would only just update the DB version to prevent a downgrade? Seems like we could also implement a run-time sanity check based on if the DB includes a type not in the set of known types, or the other way around.

channeldb/invoices.go
Roasbeef: I'd be down with removing it, it was created back in a time when the realm of possibilities was a bit more unbounded...🌈

channeldb/invoices.go
```diff
@@ -174,23 +206,10 @@ type Invoice struct {
 	// or any other message which fits within the size constraints.
 	Memo []byte
 
-	// Receipt is an optional field dedicated for storing a
```
Roasbeef: R.I.P 😭

channeldb/invoices.go
```diff
@@ -174,23 +206,10 @@ type Invoice struct {
...
 	// PaymentRequest is an optional field where a payment request created
 	// for this invoice can be stored.
 	PaymentRequest []byte
 
-	// FinalCltvDelta is the minimum required number of blocks before htlc
-	// expiry when the invoice is accepted.
-	FinalCltvDelta int32
-
-	// Expiry defines how long after creation this invoice should expire.
-	Expiry time.Duration
```
Roasbeef: Don't we want to maintain these fields, as otherwise, the invoice needs to be decoded just to access them.

channeldb/invoices.go
```diff
@@ -223,6 +242,9 @@ type Invoice struct {
 	// NOTE: This index starts at 1.
 	SettleIndex uint64
 
+	// State describes the state the invoice is in.
+	State ContractState
```
Roasbeef: If the `ContractTerm` field is staying, then why move this _outside_ of it?

channeldb/invoices.go
```diff
@@ -1018,79 +1040,89 @@ func fetchInvoice(invoiceNum []byte, invoices *bbolt.Bucket) (Invoice, error) {
...
+	var bodyLen int64
+	err = binary.Read(r, byteOrder, &bodyLen)
 	if err != nil {
-		return invoice, err
+		return i, err
 	}
 
-	if err := binary.Read(r, byteOrder, &invoice.FinalCltvDelta); err != nil {
-		return invoice, err
+	lr := io.LimitReader(r, bodyLen)
```
Roasbeef: Nice usage of `LimitReader` 👍

channeldb/migration12/invoices.go
Roasbeef: ?

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,572 @@
...
+	}
+
+	return invoice, nil
+}
+
+// deserializeHtlcs reads a list of invoice htlcs from a reader and returns it
+// as a map.
+//
+// nolint: dupl
+func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
```
Roasbeef: Since this didn't change, can't we use the version from `migration_1_to_11`?

joostjager: or just don't deserialize/serialize the htlc list at all. just pass through the bytes

channeldb/migration12/invoices.go
```diff
@@ -339,7 +361,73 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
...
+	err = binary.Write(w, byteOrder, uint64(b.Len()))
+	if err != nil {
+		return err
+	}
+
+	if _, err = w.Write(b.Bytes()); err != nil {
+		return err
+	}
+
+	return serializeHtlcs(w, i.Htlcs)
```
joostjager: Why don't we actually make this a tlv field?

channeldb/migration12/invoices.go
```diff
@@ -339,7 +361,73 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
...
+		tlv.MakePrimitiveRecord(payReqType, &i.PaymentRequest),
+
+		// Add/settle metadata.
+		tlv.MakePrimitiveRecord(createTimeType, &creationDateBytes),
+		tlv.MakePrimitiveRecord(settleTimeType, &settleDateBytes),
+		tlv.MakePrimitiveRecord(addIndexType, &i.AddIndex),
+		tlv.MakePrimitiveRecord(settleIndexType, &i.SettleIndex),
+
+		// Terms.
+		tlv.MakePrimitiveRecord(preimageType, &preimage),
```
joostjager: I start to realize that this migration offers more opportunities to clean up some things. The situation with "preimage unknown" could be dealt with much more elegant by just not writing out the preimage tlv record.

For the add/settle metadata similar a similar thing could be possible.

I know we want to move on with the rest of the stack of mpp prs, but the effort required to fix these things now seems to be much less than when it would need to be a separate migration later on. What do you think?

channeldb/migration12/invoices.go
joostjager: Then it also wouldn't be necessary to start the serialized htlcs blob with a length.

channeldb/migration12/invoices.go
cfromknecht: it does change w/ the `LimitReader` optimization

channeldb/invoices.go
cfromknecht: imo like having a separation of invoice parameters and invoice metadata. the latest restructuring moves the immutable constraints into `ContractTerm`, which are ultimately embedded in the payment request. things like `State`, which is a) mutable, and b) not present in the payreq, is now contained on the main body with the other metadata.

if we don't think this adds any value tho we can flatten entirely.

channeldb/invoices.go
cfromknecht: done

channeldb/invoices.go
cfromknecht: > The migration would only just update the DB version to prevent a downgrade? 

There will also be code added to read and write the new field

> Seems like we could also implement a run-time sanity check based on if the DB includes a type not in the set of known types, or the other way around.

Does that mean we'd read the entire db on every open? or only do it when the data is accessed? i think it ultimately depends on what the field is, in certain cases we could lose money if a downstream action fails that we are expecting to succeed. the migration is def the safest option because it doesn't create the illusion that things are okay

channeldb/invoices.go
cfromknecht: we still have the fields, they've just been moved

channeldb/invoices.go
cfromknecht: see the other inline comment re: `ContractTerm`, but the main rationale is that `State` is more metadata than it is a constraint on the sender.

channeldb/migration12/invoices.go
cfromknecht: > Then it also wouldn't be necessary to start the serialized htlcs blob with a length.

i think each one would still need a length, no? perhaps you mean the body wouldn't need to be length prefixed?

imo it'd be preferable to serialize each htlc under it's own circuit key instead, then we eliminate length-prefixed tlv streams altogether. do we think that's worth doing now?

channeldb/migration12/invoices.go
cfromknecht: refactored and now `SerializeInvoice` is always used

channeldb/migration12/invoices.go
cfromknecht: the payment hash is already serialized in the invoice index, so preferably we would just extract it from there instead of serializing again. unfortunately we only track `hash => invoice number`, so it's not so simple to recover the hash from the invoice on it's own.

> Especially for hodl invoices, there is some awkwardness to get to that value.

How is it extracted today?

joostjager: In `CreateRPCInvoice`, the hash is decoded again from the payment request. But maybe we should only add `PaymentHash` to the in-memory invoice and make sure we always populate it in `LookupInvoice`, `FetchAllInvocies`, `QueryInvoices`, etc?

channeldb/migration12/invoices.go
joostjager: Hmm, there are indeed multiple variations. But to add least add the htlc tlv blob as is as a tlv record on the invoice level looks more consistent to me.

channeldb/migration12/invoices.go
cfromknecht: yeah i think that'd better than serializing it again on the invoice, though it may be simpler to do that

channeldb/migration12/invoices.go
cfromknecht: i would vote to revisit these things if they become an issue, since those optimizations don't provide any tangible benefits in the grand scheme of things

channeldb/migration12/invoices.go
cfromknecht: i think for now we should leave them where they are, since for the time being that would _introduce_ a performance penalty in needed to allocate a backing slice for all htlcs. i'd rather not move them preemptively until we know what the workload/bottlenecks are surrounding it.

channeldb/channel.go
```diff
@@ -101,8 +102,7 @@ var (
 
 	// ErrInvalidCircuitKeyLen signals that a circuit key could not be
 	// decoded because the byte slice is of an invalid length.
-	ErrInvalidCircuitKeyLen = fmt.Errorf(
-		"length of serialized circuit key must be 16 bytes")
+	ErrInvalidCircuitKeyLen = migration12.ErrInvalidCircuitKeyLen
```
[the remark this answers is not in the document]
cfromknecht: reverted to prior version of the pr

channeldb/migration12/invoices.go
cfromknecht: i think that can be done without a migration, so it should be okay to defer for now

cfromknecht:
@joostjager @Roasbeef @halseth reverted to prior pr and touched up for another review. ended up using joost's suggestion of passing through the htlc bytes directly, which shaved 200+ lines off the diff

channeldb/invoices.go
cfromknecht: ultimately tho this doesn't affect the migration at all, since we have a working copy of the serialization in `channeldb` and are free to change the in-memory structure until the next one

channeldb/migration12/invoices.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Htlcs records all htlcs that paid to this invoice. Some of these
+	// htlcs may have been marked as canceled.
+	Htlcs []byte
+}
+
+// LegacyDeserializeInvoice decodes an invoice from the passed io.Reader using
+// the pre-TLV serialization.
+//
+// nolint: dupl
```
halseth: nolint not needed anymore?

channeldb/migration12/invoices.go
```diff
@@ -225,6 +232,95 @@ func LegacyDeserializeInvoice(r io.Reader) (Invoice, error) {
 	return invoice, nil
 }
 
+// DeserializeInvoice deserializes an invoice from an io.Reader.
+//
+// nolint: dupl
+func DeserializeInvoice(r io.Reader) (Invoice, error) {
```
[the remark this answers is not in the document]
halseth: is it needed for anything other than tests? Maybe we could just check the final bytes in the tests and just remove this method.

channeldb/invoices.go
halseth: If we are to keep the `ContractTerms`, I think there are a few more immutable things to move there? pay req, memo perhaps. I would still prefer just flatten it, but not a blocker  :)

channeldb/migration12/invoices.go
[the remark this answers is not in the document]
cfromknecht: done

Issue #3370, opened 2019-08-03
High CPU usage after a few hours


LLeny:
### Background

High CPU usage since 0.7.1.
The node runs perfectly for a few hours and at one point the CPU usage of lnd reaches 100%, I've let it for a couple of days at 100% but it never goes down. Restarting lnd fixes the issue for a few hours. This behavior wasn't seen with 0.6.
bitcoind CPU usage stays low.

### Your environment

* version of `lnd`
lnd version 0.7.1-beta commit=v0.7.1-beta
* which operating system (`uname -a` on *Nix)
Linux 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5 (2019-06-19) x86_64 GNU/Linux   
* version of `btcd`, `bitcoind`, or other backend
Bitcoin Core Daemon version v0.18.0
* any other relevant environment details

CPU profile when it happens :
![profile001](https://user-images.githubusercontent.com/5269958/62411199-340e3380-b622-11e9-8dd0-b3aa089c572a.png)

halseth:
Thanks for the report! 

cc @wpaulino

wpaulino:
Do the logs show anything relevant while this is happening?

LLeny:
I couldn't see anything unusual in the logs, the node appears to still be working as expected, peers are connected, channels are active.

wpaulino:
Are you experiencing this reliably? Do you see bitcoind falling behind at any point and then catching up?

LLeny:
Got it after a few day this time.
Bitcoind blocks are up to date, lnd is also in sync with bitcoind.
Bitcoind is not take any CPU.

wpaulino:
Could you also obtain a heap profile while experiencing the issue?

LLeny:
![profile004](https://user-images.githubusercontent.com/5269958/62819687-00d02500-bb8b-11e9-8a17-609a0b0c6c2d.png)

cfromknecht:
@LLeny  curiously, what kind of device is this running on?

i'd guess by your description that you have a peer that requests a lot of graph data, which can a lot of allocations/deserializations as the objects are created and destroyed and gc pressure. we do keep an in-memory cache of these responses in order to avoid deserializing them every time a peer requests them, but depending on your peers or hardware, it may not be enough.

[...]

PR #3695, opened 2019-11-08
build: update to btcwallet dependency with ZMQ message buffer reuse

wpaulino:
`bitcoind` notifies transactions once they're accepted into the mempool and once they're confirmed in a block. Previously, reading a message from ZMQ would allocate a buffer with the size of the message. This can cause nodes to perform a large number of allocations within a small amount periodically (3000 300B allocations every 10 mins on average), which can cause a lot of GC pressure on lower resourced nodes. To remedy this, we introduce two static buffers, one for blocks and another for transactions, that will be reused for every message read. Each is constrained by its maximum expected size.

Fixes https://github.com/lightningnetwork/lnd/issues/3370.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,3 +61,7 @@ replace github.com/lightningnetwork/lnd/queue => ./queue
 replace git.schwanenlied.me/yawning/bsaes.git => github.com/Yawning/bsaes v0.0.0-20180720073208-c0276d75487e
 
 go 1.12
+
+replace github.com/btcsuite/btcwallet => github.com/wpaulino/btcwallet v0.0.0-20191111190735-c56c957c98c3
+
+replace github.com/lightninglabs/gozmq => github.com/wpaulino/gozmq v0.0.0-20191109022438-b2d46362b57b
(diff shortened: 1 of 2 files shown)


sangaman:
I am testing this on a system that was reliably experiencing ~100% cpu usage at nearly all times. So far so good, I will update this thread again if cpu usage is still low after some time. Thanks a lot!

sangaman:
Unfortunately I just checked cpu utilization again and it's back over 100%. This is consistent with previous behavior, where cpu utilization started off low (0-15%) for at least half an hour or so, and then gets pegged at 100%+ after some hours.

wpaulino:
@sangaman could you re-apply the branch (new changes pushed) and try again? You should confirm that `lnd --version` shows the commit https://github.com/lightningnetwork/lnd/pull/3695/commits/b5827e3079fb59ac9d0c0bc563305f4d43764c09. You should also start `lnd` with the `--profile=PORT` flag so that you can obtain a CPU profile with `go tool pprof http://localhost:PORT/debug/pprof/profile` if you happen to experience the high CPU issue again.

sangaman:
Done. I'm using the latest from this branch, restarted lnd with the `--profile` option, and `getinfo` prints `"version": "0.8.0-beta commit=v0.8.0-beta-130-g39e3a45d999f59f39ad3f33c939680e2e77d3fa2"`. I will obtain a CPU profile if the 100% cpu usage returns.

sangaman:
So far so good... cpu usage still low.

sangaman:
Cpu usage still low after more than 24 hours, I'd say the issue is fixed for me with the latest commit.

sangaman:
Never mind, cpu usage back up over 100%. Profiling is on, however. What command should I use to obtain a cpu profile, `png`?

go.mod
```diff
@@ -61,3 +61,7 @@ replace github.com/lightningnetwork/lnd/queue => ./queue
 replace git.schwanenlied.me/yawning/bsaes.git => github.com/Yawning/bsaes v0.0.0-20180720073208-c0276d75487e
 
 go 1.12
+
+replace github.com/btcsuite/btcwallet => github.com/wpaulino/btcwallet v0.0.0-20191111190735-c56c957c98c3
```
Roasbeef: Can be removed now.

wpaulino:
@sangaman the issue seems to not be resolved yet. Will close this and continue investigating.

PR #3697, opened 2019-11-09
config: improve management of htlc amount lower limit

joostjager:
The minimum htlc amount is set when a channel is opened and thereafter it isn't possible anymore to change it. The other end of the channel will never be able to advertize a htlc_minimum_msat forwarding policy below this lower bound.

Until now, the default value was 1000 msat. Users who are not aware of the importance of this minimum value when opening the channel, may inadvertently contribute to a Lightning Network that is not able to carry sub-satoshi payments. Also when accepting a channel open request, this default value is used.

In this commit, the minimum htlc amount is lowered to zero. The reason to pick zero is to prepare for a possible future where zero amount payments become valid. A use case for this is to deliver a message and only pay intermediate nodes for the transport.

Because zero amount payments are currently considered invalid, this change effectively sets the minimum to 1 msat.

Changing the default will not force nodes to forward lower amount payments. If the forward doesn't satisfy the node's htlc_minimum_msat policy of the outgoing channel, the htlc is still canceled back. It
remains possible to set htlc_minimum_msat to any desired value and will still default to 1000 msat.

This commit improves the capability of the network to facilitate sub-satoshi (streaming) micro payments without reducing the level of control that node operators have over the htlcs they forward.

C-lightning also sets a minimum htlc amount value of zero in the channel open flow.

Furthermore the rpc `UpdateChanPolicy` is extended with the ability set the min htlc value.


joostjager:
The default effect seems real:

image

joostjager:
> Should we consider allowing this value to be modified through `UpdateChannelPolicy` as a follow-up?

So this value isn't the regular `min_htlc` that is advertized on the network. This is the absolute minimum that a channel will ever see in its life. After opening, it cannot be changed anymore.

The benefit of setting this to anything other than 0 (or 1) msat seems very limited and it does reduce the flexibility.

Also, if node A sets the minimum on channel open, this has consequences for node B. node B won't be able to forward anything below A's min_htlc channel parameters. It won't affect A's policy towards B.

For that reason, LND nodes connected to CL advertize a min_htlc channel policy of 0. CL has no minimum by default (or maybe it isn't even possible to set it).

Adding `min_htlc` setting to `UpdateChannelPolicy` is useful, but it isn't linked to this PR as much as one might expect.

joostjager:
@wpaulino re-requesting review because I changed significantly more to separate fwd policy from the min htlc channel parameter.

fundingmanager.go
```diff
@@ -2337,12 +2354,8 @@ func (f *fundingManager) annAfterSixConfs(completeChan *channeldb.OpenChannel,
...
 			&fundingPoint, shortChanID)
 
-		// We'll obtain the min HTLC value we can forward in our
-		// direction, as we'll use this value within our ChannelUpdate.
-		// This constraint is originally set by the remote node, as it
-		// will be the one that will need to determine the smallest
-		// HTLC it deems economically relevant.
-		fwdMinHTLC := completeChan.LocalChanCfg.MinHTLC
+		// Determine min htlc amount that we can/want to forward.
+		fwdMinHTLC := f.getMinHtlcPolicy(&completeChan.LocalChanCfg)
```
wpaulino: Nit: not caused by your change, but doesn't seem like we'd even need this parameter here since it's only for the channel update, which is not required to announce a channel. We are reusing `newChanAnnouncement` in `announceChannel` though.

lnrpc/rpc.proto
```diff
@@ -2617,6 +2617,12 @@ message PolicyUpdateRequest {
 
     /// If set, the maximum HTLC size in milli-satoshis. If unset, the maximum HTLC will be unchanged.
     uint64 max_htlc_msat = 6 [json_name = "max_htlc_msat"];
+
+    /// The minimum HTLC size in milli-satoshis. Only applied if min_htlc_msat_specified is true.
+    uint64 min_htlc_msat = 7 [json_name = "min_htlc_msat"];
+
+    /// If true, min_htlc_msat is applied.
+    bool min_htlc_msat_specified = 8 [json_name = "set_min_htlc_msat"];
```
wpaulino: Would be nice if protobuf supported setting default values...

routing/localchans/manager.go
```diff
@@ -160,6 +160,11 @@ func (r *Manager) updateEdge(chanPoint wire.OutPoint,
 		edge.MaxHTLC = amtMax
 	}
 
+	// If a new min htlc is specified, update the edge.
+	if newSchema.MinHTLC != nil {
+		edge.MinHTLC = *newSchema.MinHTLC
```
wpaulino: Check it's not higher than `MaxHTLC`?

lnrpc/rpc.proto
joostjager: You mean we could have used something like -1 to signal 'not specified'?

htlcswitch/link.go
```diff
@@ -74,7 +74,7 @@ type ForwardingPolicy struct {
 	// MinHTLC is the smallest HTLC that is to be forwarded. This is
 	// set when a channel is first opened, and will be static for the
 	// lifetime of the channel.
-	MinHTLC lnwire.MilliSatoshi
+	MinHTLCOut lnwire.MilliSatoshi
```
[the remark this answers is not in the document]
joostjager: Yes indeed, it no longer needs to with this commit. Fixed.

pilot.go
```diff
@@ -92,7 +92,7 @@ func (c *chanController) OpenChannel(target *btcec.PublicKey,
 	}
 
 	// TODO(halseth): make configurable?
-	minHtlc := lnwire.NewMSatFromSatoshis(1)
+	minHtlcIn := lnwire.NewMSatFromSatoshis(1)
```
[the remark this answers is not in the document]
joostjager: That is what it looks like. Also sub-sat ready

fundingmanager.go
joostjager: Added a commit to improve it a little. I don't want to start refactoring further in this pr.

routing/localchans/manager.go
joostjager: That check is there, a few lines down.

pilot.go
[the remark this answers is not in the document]
wpaulino: It's set to 1 sat, why not use the default here as well?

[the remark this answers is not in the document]
joostjager: Done

fundingmanager.go
```diff
@@ -2209,6 +2209,12 @@ func (f *fundingManager) addToRouterGraph(completeChan *channeldb.OpenChannel,
 	// need to determine the smallest HTLC it deems economically relevant.
 	fwdMinHTLC := completeChan.LocalChanCfg.MinHTLC
 
+	// We don't necessarily want to go as low as the remote party
```
Roasbeef: Not clear to me why we need to clamp this...

chainregistry.go
```diff
@@ -38,12 +38,14 @@ import (
...
-	// and cannot be changed during the life of the channel.
+	// and cannot be changed during the life of the channel. It is zero by
+	// default to allow maximum flexibility in deciding what size payments
+	// to forward.
 	//
 	// All forwarded payments are subjected to the min htlc constraint of
 	// the routing policy of the outgoing channel. This implicitly controls
 	// the minimum htlc value on the incoming channel too.
-	defaultBitcoinMinHTLCInMSat = lnwire.MilliSatoshi(1000)
+	defaultBitcoinMinHTLCInMSat = lnwire.MilliSatoshi(0)
```
Roasbeef: Why would accepting a zero amount HTLC ever be useful? I'm not even sure we have any tests to exercise if it'll even flow as normal for a point-to-point (direct, no hops) payment.

lnrpc/rpc.proto
Roasbeef: Why did we never do something like this for the other fields? Using the cli to determine if they were present or not? We could've even added a new message like: 
```proto
message MaybeUint64 {
    uint64 value = 1; 
    bool present = 2; 
}

fundingmanager.go
joostjager: This is the change that you requested. Not to just use the remote's min htlc in our forwarding policy, but apply our default instead. I'd rather not make this change though and stay closer to what people are used to. Let me know your final opinion on it.

chainregistry.go
joostjager: The reason for this is pretty far-fetched. It would prepare for a future where 0 sat payments are possible. It seemed safe to me, as CL is also setting this value to zero. But the direct payment case I didn't consider. Changed to 1 msat.

Roasbeef:
Needs rebase.

joostjager:
rebased

PR #3698, opened 2019-11-09
Warn user before doing SCB restore

aantonop:
SCB restore closes all open channels, which may result in significant on-chain fees. 

The current implementation of lncli create does not warn users.

We need to warn users that this is what will happen if they try to do an SCB restore, and get their affirmative consent.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -1401,6 +1401,86 @@ func create(ctx *cli.Context) error {
 	client, cleanUp := getWalletUnlockerClient(ctx)
 	defer cleanUp()
 
+	var (
+		chanBackups *lnrpc.ChanBackupSnapshot
+
+		// We use var restoreSCB to track if we will be including an SCB
+		// recovery in the init wallet request.
+		restoreSCB = false
+	)
+
+	backups, err := parseChanBackups(ctx)
+
+	// We'll check to see if the user provided any static channel backups (SCB),
+	// if so, we will warn the user that SCB recovery closes all open channels
+	// and ask them to confirm their intention.
+	// If the user agrees, we'll add the SCB recovery onto the final init wallet
+	// request.
+	switch {
+	// parseChanBackups returns an errMissingBackup error (which we ignore) if
+	// the user did not request a SCB recovery.
+	case err == errMissingChanBackup:
+
+	// Passed an invalid channel backup file.
+	case err != nil:
+		return fmt.Errorf("unable to parse chan backups: %v", err)
+
+	// We have an SCB recovery option with a valid backup file.
+	default:
+
+	warningLoop:
+		for {
+
+			fmt.Println()
+			fmt.Printf("WARNING: You are attempting to restore from a " +
+				"static channel backup (SCB) file.\nThis action will CLOSE " +
+				"all currently open channels, and you will pay on-chain fees." +
+				"\n\nAre you sure you want to recover funds from a" +
+				" static channel backup? (Enter y/n): ")
+
+			reader := bufio.NewReader(os.Stdin)
+			answer, err := reader.ReadString('\n')
+			if err != nil {
+				return err
+			}
+
+			answer = strings.TrimSpace(answer)
+			answer = strings.ToLower(answer)
+
+			switch answer {
+			case "y":
+				restoreSCB = true
+				break warningLoop
+			case "n":
+				restoreSCB = false
+				break warningLoop
+			}
+		}
+	}
+
+	// Proceed with SCB recovery.
+	if restoreSCB {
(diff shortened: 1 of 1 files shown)


aantonop:
Please be gentle. This is my first ever golang work. 
I tested it on the commandline (lncli create --multi_file ...) and it works with both "y" and "n" on the warning response

aantonop:
Took 7 commits to conform to the go linter expectations. "Better late than never". I don't yet know how to replicate the travis-ci linter rules on my local system, so that I can preview the warnings, so...

flack:
would be nice if the warning would only show if there actually are channels (i.e. a warning about closing channels when none exist could also be confusing)

aantonop:
> would be nice if the warning would only show if there actually are channels (i.e. a warning about closing channels when none exist could also be confusing)

I agree, we should suppress this warning if there are no open channels. Furthermore, we should show the number of open channels in the warning. 

I am not sure if at that stage of lnd's boot process, lncli can access the current open channels. I'll have to see if that is possible. Let me dig around. I'm new to this codebase

aantonop:
From what I can tell, the channel arbiter is not running during the "create" phase of lnd's boot cycle and therefore I can't check open channels. Let me know if I missed something

cmd/lncli/commands.go
```diff
@@ -1400,6 +1400,102 @@ func create(ctx *cli.Context) error {
...
+	// and ask them to confirm their intention.
+	//
+	// If the user agrees, we'll add the SCB recovery onto
+	// the final init wallet request.
+	//
+
+	var (
+		chanBackups *lnrpc.ChanBackupSnapshot
+
+		// We use var restoreSCB to track if we will be including
```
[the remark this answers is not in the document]
aantonop: Fixed everywhere, thanks

cmd/lncli/commands.go
```diff
@@ -1400,6 +1400,76 @@ func create(ctx *cli.Context) error {
...
+				break warningLoop
+			case "n":
+				restoreSCB = false
+				break warningLoop
+			}
+		}
+	}
+
+	if restoreSCB == true {
+		if backups != nil {
```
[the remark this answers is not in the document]
aantonop: It's more of a defensive guard, in case someone changes the logic flow above.

aantonop:
Fixed comment formatting, addressed all your review issues, squashed and rebased!

cmd/lncli/commands.go
```diff
@@ -1401,6 +1401,101 @@ func create(ctx *cli.Context) error {
 	client, cleanUp := getWalletUnlockerClient(ctx)
 	defer cleanUp()
 
+	// We'll check to see if the user provided any static channel backups (SCB),
```
guggero: The comment looks a bit lost up here. I know it is intended to describe the next few blocks.
I think it would better fit as the comment block for the switch statement below.

cmd/lncli/commands.go
```diff
@@ -1401,6 +1401,101 @@ func create(ctx *cli.Context) error {
...
+	// We can therefore ignore the errMissingChanBackup error as it's an
+	// optional field.
+	//
+	// Try to load an SCB file.
+	backups, err := parseChanBackups(ctx)
+
+	// Evaluate result (err) from parseChanBackups.
+	switch {
+	case err == errMissingChanBackup:
+		// SCB recovery is an optional parameter of the create command.
```
halseth: style nit: we usually put switch comments above the `case` instead of below. This avoids some indentation  :)

aantonop: Fixing

cmd/lncli/commands.go
aantonop: Moving

aantonop:
Cleaned up comments some more (still getting used to the style conventions of Go and the lnd team). Squashed, rebased and pushed anew.

cmd/lncli/commands.go
```diff
@@ -1401,6 +1401,86 @@ func create(ctx *cli.Context) error {
...
+
+			answer = strings.TrimSpace(answer)
+			answer = strings.ToLower(answer)
+
+			switch answer {
+			case "y":
+				restoreSCB = true
+				break warningLoop
+			case "n":
+				restoreSCB = false
```
halseth: If the user provided channel backups, but enters `n` here, maybe it is more natural to return instead of continuing unlocking the wallet?

aantonop:
Great suggestion. Please hold while I revise it to exit the "create"
command.. As you noticed, it doesn't make sense to continue with the wallet
recovery/creation if the user has said "no"

Roasbeef:
Merging as is so this can make it into `0.8.2`, will add on a commit addressing Johan's final comment.

PR #3701, opened 2019-11-10
tlv+hop: contain odd/even logic in payload parsing

joostjager:
Tlv is used more widely in lnd than just for the onion payload. This commit isolates the protocol-specific odd/even logic, so that tlv can be used freely elsewhere. An example of this use is db serialization.


tlv/record.go
```diff
@@ -13,7 +13,7 @@ import (
 type Type uint64
 
 // TypeSet is an unordered set of Types.
-type TypeSet map[Type]struct{}
+type TypeSet map[Type]bool
```
cfromknecht: godoc should explain the boolean value

htlcswitch/hop/payload.go
```diff
@@ -124,31 +130,52 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
+		requiredViolation = true
+	}
+
+	if requiredViolation {
+		// If the parser encountered an unknown required type failure,
+		// we'll first check that the payload is properly formed
+		// according to our known set of constraints. If an error is
+		// discovered, this overrides the required type failure.
+		nextHop := lnwire.NewShortChanIDFromInt(cid)
+		err = ValidateParsedPayloadTypes(parsedTypes, nextHop)
```
cfromknecht: Is it possible to remove this duplicate call? currently we call this in two different paths, but I think now we can do the required checks after a single `ValudateParsedPayloadTypes`

htlcswitch/hop/payload.go
```diff
@@ -124,31 +130,52 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
 		return nil, err
 	}
 
+	// Check for unrecognized mandatory (even) fields and save the lowest
+	// required type to return a deterministic failure.
+	var (
+		requiredViolation        bool
+		minRequiredViolationType tlv.Type
+	)
+	for t, known := range parsedTypes {
```
cfromknecht: Perhaps this code should live in the TLV package? We will need it in other places besides the onion payload, something like a global `CheckUnknownRequiredTypes` method would allow us to reuse

htlcswitch/hop/payload.go
joostjager: Yes indeed, fixed

htlcswitch/hop/payload.go
joostjager: Where else do we need it? I did extract the logic to a separate function. We can also lift it when needed.

cfromknecht: we'll need it for parsing wire messages as they're extended, but we don't use it anywhere else atm

joostjager: do you prefer to move this func into the `tlv` package now?

cfromknecht: either is fine, for now at least it is already tested by existing tests

PR #1462, opened 2018-06-28
Add optional delivery address to cooperative channel close

Bluetegu:
Addresses (part of) #478 

rpc+peer+htlcswitch: Add delivery_addr parameter to closechannel
command. When specified funds are sent to this address upon
cooperative channel close. Add DeliveryAddr parameter to htlcswitch
CloseChan struct to propagate the address to the channel closer.

Add a test function that checks that channel closer uses the provided 
new DeliveryAddr parameter.

The following test was done on simnet with Alice/Bob/Charlie nodes.

Regression (Charlie):
1. Open a channel from Charlie to Bob
2. Issue a closechannel command without delivery_addr specified. 
3. Check funds return to wallet and transaction output is a new wallet address

Delivery (Charlie):
1. Open a channel from Charlie to Bob
2. Generate new wallet address 
3. Issue a closechannel command with delivery_addr set to new wallet
address of (2). 
4. Check funds return to wallet and transaction output is to specified 
delivery address.

Extreme:
1. Open a channel from Charlie to Bob
2. Generate new wallet address on Alice's wallet 
2. Issue a closechannel command with delivery_addr set to Alice's new 
wallet address. 
3. Check funds added to Alice's wallet.

Here are cli outputs of the last test:

ubuntu@btc:~/go/dev/charlie$ lncli-c openchannel --node_key=026a52eadd42eabb70e9b0481af7001d299f7d7b41030977bb37a2faac7bbe0012 --private --connect=localhost:10012 --local_amt=150000
{
        "funding_txid": "44b16c12f84c6acac9f8c98483406ad0237cf17fdffa9cfc58b08beb3af8cf36"
}

ubuntu@btc:~/go/dev/charlie$ btcc generate 4

ubuntu@btc:~/go/dev/charlie$ lncli-c listchannels
{
    "channels": [
        {
            "active": true,
            "remote_pubkey": "026a52eadd42eabb70e9b0481af7001d299f7d7b41030977bb37a2faac7bbe0012",
            "channel_point": "44b16c12f84c6acac9f8c98483406ad0237cf17fdffa9cfc58b08beb3af8cf36:0",
            "chan_id": "2011006767267840",
            "capacity": "150000",
            "local_balance": "140950",
            "remote_balance": "0",
            "commit_fee": "9050",
            "commit_weight": "600",
            "fee_per_kw": "12500",
            "unsettled_balance": "0",
            "total_satoshis_sent": "0",
            "total_satoshis_received": "0",
            "num_updates": "0",
[...]

PR #3702, opened 2019-11-11
Add optional close address

carlaKC:
This PR adds a `delivery_address` field to lnrpc and lncli's `CloseChannel` request which allows users to specify an address to close out to on cooperative closure of a channel. If an upfront shutdown address was set on channel open, requesting to close the channel out to a specific address will fail. 

This PR replaces #1462.

This PR is rebased on the open upfront shutdown PR #3655, as much of the logic is shared. 
**Only the last 3 commits in this PR are relevant for review.**


joostjager:
Needs rebase

lnrpc/rpc.swagger.json
```diff
@@ -441,59 +429,25 @@
...
         "parameters": [
           {
             "name": "channel_point.funding_txid_str",
-            "description": "*\nHex-encoded string representing the byte-reversed hash of the funding\ntransaction.",
             "in": "path",
             "required": true,
             "type": "string"
           },
           {
             "name": "channel_point.output_index",
```
bhandras: There are a few desciption fields that are now gone. I suppose this is generated but should these fields remain?

lnwallet/reservation.go
```diff
@@ -521,6 +524,16 @@ func (r *ChannelReservation) FundingOutpoint() *wire.OutPoint {
 	return &r.partialState.FundingOutpoint
 }
 
+// SetOurUpfrontShutdown sets the upfront shutdown address on our contribution
+// and partial state in the reservation.
+func (r *ChannelReservation) SetOurUpfrontShutdown(shutdown lnwire.DeliveryAddress) {
+	r.Lock()
+	defer r.Unlock()
+
+	r.ourContribution.UpfrontShutdown = shutdown
```
bhandras: nit: As a reader it's really confusing to me that sometimes it's Shutdown, sometimes ShutdownScript. I guess this is because I'm not familiar with the code base yet.

lnwire/accept_channel_test.go
```diff
@@ -0,0 +1,70 @@
...
+			if _, err := WriteMessage(buf, encoded, 0); err != nil {
+				t.Fatalf("cannot write message: %v", err)
+			}
+
+			msg, err := ReadMessage(buf, 0)
+			if err != nil {
+				t.Fatalf("cannot read message: %v", err)
+			}
+
+			switch e := msg.(type) {
```
bhandras: nit: I'd probably call `e` as `decoded`.

lnwire/shutdown.go
```diff
@@ -71,11 +80,8 @@ func (s *Shutdown) MaxPayloadLength(pver uint32) uint32 {
 	// Len - 2 bytes
 	length += 2
 
-	// ScriptPubKey - 34 bytes for pay to witness script hash
-	length += 34
-
-	// NOTE: pay to pubkey hash is 25 bytes, pay to script hash is 22
-	// bytes, and pay to witness pubkey hash is 22 bytes in length.
+	// ScriptPubKey
```
bhandras: Maybe add a comment here that it's 34 bytes. (to not break the commenting pattern)

peer_test.go
```diff
@@ -588,3 +594,152 @@ func TestPeerChannelClosureFeeNegotiationsInitiator(t *testing.T) {
...
+
+			// Send the close command for the correct channel and check that a
+			// shutdown message is sent.
+			initiator.localCloseChanReqs <- &closeCommand
+
+			var msg lnwire.Message
+			select {
+			case outMsg := <-initiator.outgoingQueue:
+				msg = outMsg.msg
+			case <-time.After(time.Second * 5):
```
bhandras: Why 5 sec? Should this be a timeout const?

chancloser.go
```diff
@@ -83,6 +84,10 @@ type chanCloseCfg struct {
 	// forward payments.
 	disableChannel func(wire.OutPoint) error
 
+	// disconnect should be called to disconnect from the remote peer in this
```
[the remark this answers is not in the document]
carlaKC: np, does "disconnect is called to disconnect from a peer" clearer?
I was trying to avoid a comment along the lines on "disconnect disconnects from peer", which isn't very helpful.

chancloser.go
```diff
@@ -263,6 +268,23 @@ func (c *channelCloser) CloseRequest() *htlcswitch.ChanClose {
 	return c.closeReq
 }
 
+// checkRemoteDeliveryAddress checks that the address provided by the remote
+// peer matches the upfront shutdown address, if it is set. It does not perform
+// any validation on the address provided.
+func checkRemoteDeliveryAddress(upfront, address lnwire.DeliveryAddress) error {
```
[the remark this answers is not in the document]
carlaKC: +1 for the unit test, see comment below for context on this check :)

chancloser.go
```diff
@@ -263,6 +268,23 @@ func (c *channelCloser) CloseRequest() *htlcswitch.ChanClose {
...
+	// If an upfront shutdown address is empty, return early.
+	if len(upfront) == 0 {
```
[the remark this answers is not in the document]
carlaKC: So here we are checking whether the address our peer has provided matches the pre-negotiated upfront address. There are two cases:
1. They did not negotiate an upfront address (it's an optional feature), so we do not need to enforce close out to a specific address
2. They did negotiate an upfront address, and we need to check that they're sticking to it; if not, we need to disconnect as per the spec's instruction because they may be comporomised

chancloser_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			err := checkRemoteDeliveryAddress(
+				test.upfrontAddr, test.shutdownAddr,
+			)
+			if (err != nil) != (test.expectErr) {
```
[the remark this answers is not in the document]
carlaKC: much prefer this way, thanks! thoughts on `hadErr` instead of `matches` for the var name? Since we're not matching the err, we're checking for its presence

channeldb/channel.go
```diff
@@ -2552,7 +2572,38 @@ func putChanInfo(chanBucket *bbolt.Bucket, channel *OpenChannel) error {
...
 	}
 
-	return chanBucket.Put(chanInfoKey, w.Bytes())
+	if err := chanBucket.Put(chanInfoKey, w.Bytes()); err != nil {
+		return err
+	}
+
+	// Finally, add optional shutdown scripts for the local and remote peer if
+	// they are present.
+	if err := putOptionalUpfrontShutdownScript(
```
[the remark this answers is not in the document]
carlaKC: This PR is built on 3655, so the first few commits belong to that PR, sorry for the confusion!
But if you have a look at `putOpenChannel` which calls `putChanInfo`, is already not atomic? 
We did have a discussion about migration vs new key in lnd/friends, but maybe raise it on the original PR? I'd be interested to see what other people think.

fundingmanager.go
```diff
@@ -1259,10 +1260,32 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
...
+	// address if our node is configured to set shutdown addresses by default.
+	// A nil address is set in place of user input, because this channel open
+	// was not initiated by the user.
+	shutdown, err := getUpfrontShutdown(fmsg.peer, nil,
+		func() (lnwire.DeliveryAddress, error) {
+			addr, err := f.cfg.Wallet.NewAddress(lnwallet.WitnessPubKey, false)
+			if err != nil {
+				return nil, err
+			}
+			return txscript.PayToAddrScript(addr)
```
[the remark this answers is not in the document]
carlaKC: This is a go thing :) returning a function call returns all of its return values, so `txscript.PayToAddrScript(addr)` is returning a delivery address and an error. Adding ,nil would mean we were returning three items.

fundingmanager.go
```diff
@@ -2728,6 +2754,48 @@ func (f *fundingManager) initFundingWorkflow(peer lnpeer.Peer, req *openChanReq)
...
+		return address, nil
+	}
+
+	// If we do not have setting of upfront shutdown addresses enabled, return
+	// an empty address.
+	if !cfg.EnableUpfrontShutdown {
+		return nil, nil
+	}
+
+	return getAddr()
```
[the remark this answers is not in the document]
carlaKC: As above, a go thing. Takes a while to get used to the way it looks 😄

fundingmanager.go
```diff
@@ -2728,6 +2754,48 @@ func (f *fundingManager) initFundingWorkflow(peer lnpeer.Peer, req *openChanReq)
 	}
 }
 
+// getUpfrontShutdown takes a user provided shutdown address and checks that
```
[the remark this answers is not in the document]
carlaKC: Me neither :') I'll fix it.

lnwallet/reservation.go
carlaKC: That's an inconsistency on my side, trying to keep var names brief. I'll settle on one or the other,

carlaKC:
Thanks for the review @bhandras! 
Since this PR is rebased on #3655, only the last two commits are relevant for review in this PR, sorry for not making that clearer! I'm not going to reply to the comments on the original PR commits, because that's going to add noise to this PR, but I will address them on 3655.

peer_test.go
carlaKC: tbh, this PR was originally a contributor PR which I took over, but I think they just followed the existing pattern in this file. I'll make a timeout var though, nice idea 👍

htlcswitch/switch.go
```diff
@@ -1366,11 +1369,12 @@ func (s *Switch) teardownCircuit(pkt *htlcPacket) error {
 
 // CloseLink creates and sends the close channel command to the target link
 // directing the specified closure type. If the closure type if CloseRegular,
-// then the last parameter should be the ideal fee-per-kw that will be used as
-// a starting point for close negotiation.
-func (s *Switch) CloseLink(chanPoint *wire.OutPoint, closeType ChannelCloseType,
-	targetFeePerKw chainfee.SatPerKWeight) (chan interface{},
-	chan error) {
+// targetFeePerKw parameter should be the ideal fee-per-kw that will be used as
```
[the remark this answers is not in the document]
carlaKC: gotta catch em all ✊

lnrpc/rpc.swagger.json
carlaKC: Not sure what's happening here, I'm just running `make rpc`, which should work out. Will investigate.

rpcserver.go
```diff
@@ -1877,8 +1877,36 @@ func (r *rpcServer) CloseChannel(in *lnrpc.CloseChannelRequest,
 		// cooperative channel closure. So we'll forward the request to
 		// the htlc switch which will handle the negotiation and
 		// broadcast details.
+
+		var deliveryScript lnwire.DeliveryAddress
+
+		// If a delivery address to close out to was specified and option
+		// upfront shutdown is not set, decode the address.
+		if len(in.DeliveryAddr) > 0 {
+			// Error if an upfront shutdown address has already been set.
```
carlaKC: nit to self: replace address with script

carlaKC:
Updated to address comments and rebased on #3655.

chancloser.go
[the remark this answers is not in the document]
bhandras: Yeah, diconnect disconnects would be weird :) I'm fine with both "is called" or "can be called". Should sounds like one should do this, otherwise something may happen.

fundingmanager.go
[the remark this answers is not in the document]
bhandras: Duh :) Sorry, didn't pay attention to what txscropt.PayToAddrScript(addr) returns :)

carlaKC:
Rebased, but going to hold off on review for this one until parent PR is in.

lnrpc/rpc.swagger.json
carlaKC: Needed to reset my proto back to the right version, not sure how it got out of sync since I use `make rpc`, but all sorted after some good old fashioned manual intervention 😬

lnrpc/rpc.proto
```diff
@@ -1557,6 +1557,14 @@ message CloseChannelRequest {
...
     /// A manual fee rate set in sat/byte that should be used when crafting the closure transaction.
     int64 sat_per_byte = 4;
+
+    /*
+    An optional address to send funds to in the case of a cooperative close.
+    If the channel was opened with an upfront shutdown script and this field
+    is set, the request to close will fail because the channel must pay out
+    to the upfront shutdown addresss.
+    */
+    string delivery_addr = 5 [json_name = "delivery_address"];
```
bhandras: Is there any reason to have different naming in proto vs json?

chancloser.go
```diff
@@ -28,10 +28,10 @@ var (
...
 	ErrInvalidState = fmt.Errorf("invalid state")
 
-	// errUpfrontShutdownScriptMismatch is returned when our peer sends us a
-	// shutdown message with a script that does not match the upfront shutdown
-	// script previously set.
-	errUpfrontShutdownScriptMismatch = fmt.Errorf("peer's shutdown " +
+	// ErrUpfrontShutdownScriptMismatch is returned when a peer or end user
+	// provides a script to cooperatively close out to which does not match
+	// the upfront shutdown script previously set for that party.
+	errUpfrontShutdownScriptMismatch = fmt.Errorf("shutdown " +
```
bhandras: nit: comment is `ErrUpfrontShutdownScriptMismatch` while the var name is `errUpfrontShutdownScriptMismatch` (capital vs non capital).

rpcserver.go
```diff
@@ -1881,8 +1881,36 @@ func (r *rpcServer) CloseChannel(in *lnrpc.CloseChannelRequest,
...
+					"shudown is enabled")
+			}
+
+			// Decode the address provided.
+			addr, err := btcutil.DecodeAddress(
+				in.DeliveryAddr, activeNetParams.Params,
+			)
+			if err != nil {
+				return fmt.Errorf("invalid delivery address: %v",
+					err)
```
bhandras: nit: no need for the line break.

server.go
```diff
@@ -820,11 +820,12 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
 	})
 
 	// Construct a closure that wraps the htlcswitch's CloseLink method.
+	// Provide no close out delivery address, as user input is not available.
 	closeLink := func(chanPoint *wire.OutPoint,
 		closureType htlcswitch.ChannelCloseType) {
 		// TODO(conner): Properly respect the update and error channels
 		// returned by CloseLink.
-		s.htlcSwitch.CloseLink(chanPoint, closureType, 0)
+		s.htlcSwitch.CloseLink(chanPoint, closureType, 0, nil)
```
bhandras: readability nit: consider adding a variable eg. `deliveryAddress := nil` and the top comment above that, so that it's even more clear what `nil` refers to.

htlcswitch/switch.go
```diff
@@ -105,6 +105,9 @@ type ChanClose struct {
 	// process for the cooperative closure transaction kicks off.
 	TargetFeePerKw chainfee.SatPerKWeight
 
+	// Optional delivery script to pay funds out to.
```
wpaulino: Nit: godoc is usually of the form "DeliveryScript ...`, weird the linter didn't catch this.

peer.go
```diff
@@ -2148,13 +2179,26 @@ func (p *peer) handleLocalCloseReq(req *htlcswitch.ChanClose) {
...
+		// An upfront shutdown and user provided script are both optional,
+		// but must be equal if both set  (because we cannot serve a request
+		// to close out to a script which violates upfront shutdown). Get the
+		// appropriate address to close out to (which may be nil if neither
+		// are set) and error if they are both set and do not match.
+		deliveryScript, err := chooseDeliveryScript(
+			channel.LocalUpfrontShutdownScript(), req.DeliveryScript,
+		)
+		if err != nil {
+			peerLog.Errorf(err.Error())
```
wpaulino: Nit: add some context to this log message, e.g., which channel it's for.

peer_test.go
```diff
@@ -584,7 +602,233 @@ func TestPeerChannelClosureFeeNegotiationsInitiator(t *testing.T) {
...
+	case <-time.After(timeout):
 		t.Fatalf("closing tx not broadcast")
 	}
 }
+
+// TestChooseDiscoveryScript tests that chooseDeliveryScript correctly errors
+// when upfront and user set scripts that do not match are provided, allows
+// matching values and returns appropriate values in the case where one or none
+// are set.
+func TestChooseDiscoveryScript(t *testing.T) {
```
wpaulino: Nit: TestChooseDeliveryScript*

lnrpc/rpc.proto
carlaKC: No, slipped through from original PR. Will update!

server.go
carlaKC: Declaring a var just so we can pass in a named nil variable is a bit atypical for go, going to leave this one if that's ok? Rather going to move the comment down so it's closer to the nil value.

rpcserver.go
```diff
@@ -1881,8 +1881,35 @@ func (r *rpcServer) CloseChannel(in *lnrpc.CloseChannelRequest,
...
 		// the htlc switch which will handle the negotiation and
 		// broadcast details.
+
+		var deliveryScript lnwire.DeliveryAddress
+
+		// If a delivery address to close out to was specified and option
+		// upfront shutdown is not set, decode the address.
+		if len(in.DeliveryAddress) > 0 {
+			// Error if an upfront shutdown script has already been set.
+			if len(channel.LocalUpfrontShutdownScript()) > 0 {
```
halseth: isn't this also checked by the switch?

server.go
halseth: Agreed that it is a bit hard to read just by passing nil, but a comment seem sufficient  in this case :)

rpcserver.go
carlaKC: it is, will remove!

carlaKC:
Rebased to clear proto conflicts :)

PR #3705, opened 2019-11-11
routing+htlcswitch: improve failed payment logging

joostjager:
This pr contains several changes to improve the information that is logged when a payment attempt fails, which will hopefully help us in debugging routing/mission control related issues.

### DIFF
--- a/htlcswitch/failure.go
+++ b/htlcswitch/failure.go
@@ -30,10 +30,14 @@ type ForwardingError struct {
 // returned.
 func (f *ForwardingError) Error() string {
 	if f.ExtraMsg == "" {
-		return fmt.Sprintf("%v", f.FailureMessage)
+		return fmt.Sprintf(
+			"%v@%v", f.FailureMessage, f.FailureSourceIdx,
+		)
 	}
 
-	return fmt.Sprintf("%v: %v", f.FailureMessage, f.ExtraMsg)
+	return fmt.Sprintf(
+		"%v@%v: %v", f.FailureMessage, f.FailureSourceIdx, f.ExtraMsg,
+	)
 }
 
 // ErrorDecrypter is an interface that is used to decrypt the onion encrypted
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -1248,8 +1248,14 @@ func TestChannelLinkMultiHopUnknownNextHop(t *testing.T) {
 		totalTimelock).Wait(30 * time.Second)
 	if err == nil {
 		t.Fatal("error haven't been received")
-	} else if err.Error() != lnwire.CodeUnknownNextPeer.String() {
-		t.Fatalf("wrong error have been received: %v", err)
+	}
+	fErr, ok := err.(*ForwardingError)
+	if !ok {
+		t.Fatalf("expected ForwardingError")
+	}
+	if _, ok = fErr.FailureMessage.(*lnwire.FailUnknownNextPeer); !ok {
+		t.Fatalf("wrong error has been received: %T",
+			fErr.FailureMessage)
 	}
 
 	// Wait for Alice to receive the revocation.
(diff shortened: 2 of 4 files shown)


htlcswitch/failure.go
```diff
@@ -30,10 +30,14 @@ type ForwardingError struct {
 // returned.
 func (f *ForwardingError) Error() string {
```
carlaKC: Had a quick look at the places where we string match errors in the codebase, and it doesn't look like we do it for ForwardingErrs except the test you've already picked up. May be worth triple checking before merge juust in case.

htlcswitch/link_test.go
```diff
@@ -1248,8 +1248,14 @@ func TestChannelLinkMultiHopUnknownNextHop(t *testing.T) {
...
 		t.Fatal("error haven't been received")
-	} else if err.Error() != lnwire.CodeUnknownNextPeer.String() {
-		t.Fatalf("wrong error have been received: %v", err)
+	}
+	fErr, ok := err.(*ForwardingError)
+	if !ok {
+		t.Fatalf("expected ForwardingError")
+	}
+	if _, ok = fErr.FailureMessage.(*lnwire.FailUnknownNextPeer); !ok {
+		t.Fatalf("wrong error have been received: %T",
```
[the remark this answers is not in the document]

htlcswitch/failure.go
joostjager: I assume you mean string matching in non-test code? (In test code, the test would fail?) I think that may be quite difficult to triple check, because they may be passed anywhere as type `error`. At this point, I think I trust that everything that really matters doesn't do string matching in `lnd` :grimacing:

carlaKC: haha cool, I got nervous just seeing it 😂

routing/route/route.go
```diff
@@ -340,15 +340,19 @@ func (r *Route) ToSphinxPath() (*sphinx.PaymentPath, error) {
 func (r *Route) String() string {
 	var b strings.Builder
 
+	amt := r.TotalAmount
 	for i, hop := range r.Hops {
 		if i > 0 {
-			b.WriteString(",")
+			b.WriteString(" -> ")
```
halseth: can you add an example of the old vs new output? Looks like fees aren't explicitly printed now?

routing/payment_lifecycle.go
```diff
@@ -341,8 +341,8 @@ func (p *paymentLifecycle) sendPaymentAttempt(firstHop lnwire.ShortChannelID,
 		return err
 	}
 
-	log.Debugf("Payment %x (pid=%v) successfully sent to switch",
-		p.payment.PaymentHash, p.attempt.PaymentID)
+	log.Debugf("Payment %x (pid=%v) successfully sent to switch, route: %v",
```
halseth: hm, won't this be quite spammy? Maybe it is okay, since it is useful information indeed.

routing/route/route.go
joostjager: new output:
`1727264597517467648 (5001005 mSAT) -> 1727817651866959873 (5000000 mSAT), cltv 1608341`

The fees aren't explictly printed anymore, but can still be calculated by subtracting the last amount from the first amount. I am seeking for minimal log spam but still providing everything we need for debugging

routing/payment_lifecycle.go
joostjager: It is still a single log line

routing/route/route.go
```diff
@@ -340,15 +340,19 @@ func (r *Route) ToSphinxPath() (*sphinx.PaymentPath, error) {
...
+			strconv.FormatUint(hop.ChannelID, 10),
+			amt,
+		))
+		amt = hop.AmtToForward
 	}
 
-	return fmt.Sprintf("amt=%v, fees=%v, tl=%v, chans=%v",
-		r.TotalAmount-r.TotalFees(), r.TotalFees(), r.TotalTimeLock,
-		b.String(),
+	return fmt.Sprintf("%v, cltv %v",
```
halseth: nit: looks like can fit on single line

routing/route/route.go
halseth: sgtm

routing/payment_lifecycle.go
halseth: Yeah true, just  a potentially long line 😂

PR #3706, opened 2019-11-11
lnrpc+routerrpc+lncli: add msat fields

### LINKED ISSUE #3689: QueryRoutes API - amt ->amt_msat changing ?
I am developing a rebalancer and it seems to me that `QueryRoutes` method should have a amount value in millisatoshis not satoshis. For example:

I want to get a path for payment:

A -> B -> C -> D

Here are A & D - my nodes. May be first and last nodes are same (A === D)

If i will try to find path for this payment by `QueryRoutes` - the LND will not find because he cannot pay to hirself. Therefore rebalancers use a tricky trick: they use this API method for finding of path A -> B -> C but the C node in request gets amount + fee of hop (C->D). When results are gotten from LND they add last hop (C->D) to path. But there is big problem - quantity of amounts in satoshis. I can’t subtly set the amount for the node C so that then I can easily recalculate the fees of the last node (C) and the forwarded amount (I have to recount (and other rebalancers too - i saw sources) the entire chain of hops, since initially the LND server calculated them for the amount received by the C node of a multiple of 1000 millisatoshis (because `amt` satoshis), but I need to quantize in millisatoshis).

In addition, there is another argument for such changes: if you analyze your trends in API changes, then you move away from satoshis and move towards millisatoshis units (gRPC messages: Route, Hop - there all units in satoshis marked as depricated).

In other words the value of the results from the `QueryRoutes` method drops when satoshis units of measure are fed to it because these results cannot be easy used in rebalancing methods.

What do you think?

joostjager:
Lightning supports payments at the msat granularity, but `lnd` does not expose this on the rpc interface. This PR adds msat fields for payment amounts and fee limits.

Fixes #3689

NOTE: This renames some of the existing rpc fields which is a breaking change for users building against our proto files. The change doesn't change anything on the wire level.


lnrpc/rpc.proto
```diff
@@ -1766,8 +1788,19 @@ message QueryRoutesRequest {
...
     string pub_key = 1;
 
-    /// The amount to send expressed in satoshis
-    int64 amt = 2;
+    /**
+    The amount to send expressed in satoshis.
+
+    The fields amt_sat and amt_msat are mutually exclusive.
+    */
+    int64 amt_sat = 2;
```
wpaulino: Any reason to not deprecate `amt_sat`?

joostjager: I think it was brought up that msat causes confusion. I'd rather deprecated sat completely though.

joostjager: @Roasbeef did you have an opinion about the necessity of sat fields?

lnrpc/marshall_utils.go
```diff
@@ -0,0 +1,52 @@
...
+	// amount are set.
+	ErrSatMsatMutualExclusive = errors.New(
+		"sat and msat arguments are mutually exclusive",
+	)
+)
+
+// CalculateFeeLimit returns the fee limit in millisatoshis. If a percentage
+// based fee limit has been requested, we'll factor in the ratio provided with
+// the amount of the payment.
+func CalculateFeeLimit(feeLimit *FeeLimit,
```
guggero: Nice to see these two functions extracted!

lnrpc/marshall_utils.go
```diff
@@ -0,0 +1,52 @@
...
+	amount lnwire.MilliSatoshi) lnwire.MilliSatoshi {
+
+	switch feeLimit.GetLimit().(type) {
+	case *FeeLimit_FixedSat:
+		return lnwire.NewMSatFromSatoshis(
+			btcutil.Amount(feeLimit.GetFixedSat()),
+		)
```
guggero: style nit: don't we usually add a newline after the body of a case block?

lnrpc/rpc.proto
```diff
@@ -873,8 +884,19 @@ message SendRequest {
...
     string dest_string = 2 [deprecated = true];
 
-    /// Number of satoshis to send.
-    int64 amt = 3;
+    /**
+    The amount to send expressed in satoshis.
+
+    The fields amt_sat and amt_msat are mutually exclusive.
+    */
+    int64 amt_sat = 3;
```
guggero: This is a breaking change for users of the API. Should we start adding a "Breaking Changes" section to the release notes?

rpcserver.go
```diff
@@ -3100,21 +3079,27 @@ func extractPaymentIntent(rpcPayReq *rpcPaymentRequest) (rpcPaymentIntent, error
...
+			}
+			if amt == 0 {
 				return payIntent, errors.New("amount must be " +
 					"specified when paying a zero amount " +
 					"invoice")
 			}
 
 			payIntent.msat = lnwire.NewMSatFromSatoshis(
-				btcutil.Amount(rpcPayReq.Amt),
+				btcutil.Amount(amt),
```
guggero: Can replace these three lines with just `payIntent.msat = amt`

lnrpc/rpc.proto
joostjager: It is a breaking change, but not on the wire level. I will add a comment to the PR description so that it can be picked up when we write the release notes.

lnrpc/marshall_utils.go
joostjager: fixed

rpcserver.go
joostjager: ok, forgot that. crazy casts

joostjager:
>If we introduce a braking change in the API anyway, wouldn't this be the moment to also just get rid of the sat amounts and only deal with msat?

Get rid of it is also a breaking change on the wire level. So that needs to be preceded with deprecation first.

Outstanding question on that above

lnrpc/rpc.proto
guggero: Ah, right, because the field names aren't sent over the wire (in gRPC). And in REST it's an unnamed parameter in the URL.
My comment about removing all sat fields doesn't make sense then.

lnrpc/rpc.proto
joostjager: Discussed offine. No deprecation yet.

joostjager:
Left existing field unchanged, to prevent cascading changes in projects that depend on it.

PR #3710, opened 2019-11-12
lncli: do not hang on ctrl-c

joostjager:
If the lnd package is referenced, the interrupt signal handler will be initialized causing lncli to not properly exit on ctrl-c.

This commit prevents this by making the initialization of the interrupt interceptor an explicit action.

### DIFF
--- a/signal/signal.go
+++ b/signal/signal.go
@@ -26,7 +26,8 @@ var (
 	shutdownChannel = make(chan struct{})
 )
 
-func init() {
+// Intercept starts the interception of interrupt signals.
+func Intercept() {
 	signalsToCatch := []os.Signal{
 		os.Interrupt,
 		os.Kill,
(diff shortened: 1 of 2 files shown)


joostjager:
Yes, I described those alternatives in the commit message. Copied them over to the PR description for visibility.

joostjager:
Switched to explicit init approach

joostjager:
Add did a hodl invoice on the other side

signal/signal.go
```diff
@@ -26,7 +26,8 @@ var (
 	shutdownChannel = make(chan struct{})
 )
 
-func init() {
+// Start starts the interception of interrupt signals.
+func Start() {
```
wpaulino: Nit: rename to `Intercept`?

joostjager: Alright, `signal.Intercept()`

Crypt-iQ:
Yay, late to the party, but this PR fixed the signal race that happened on shutdown. :)

PR #3713, opened 2019-11-12
cert: extract common certificate related functions into own package

guggero:
We use the certificate generation part from `lnd.go` in other projects, at the moment by copy/pasting the code.
This PR extracts some commonly used code around certificates and TLS into the new package called `cert`.

The autocert functionality of #2428 could also be extracted to this package as this feature is also used in other code bases.


guggero:
Yeah, it definitely makes sense to extract `cert` into its own module. But to decouple it completely from the rest of lnd, I had to make `build` into its own module too since the `cert` package depends on it for the logging.
We'll also need a tag in git for the `cert` module to get rid of the `v0.0.0-00010101000000-000000000000` version number in the root `go.mod`.

cert/selfsigned.go
```diff
@@ -0,0 +1,172 @@
...
+
+// GenCertPair generates a key/cert pair to the paths provided. The
+// auto-generated certificates should *not* be used in production for public
+// access as they're self-signed and don't necessarily contain all of the
+// desired hostnames for the service. For production/public use, consider a
+// real PKI.
+//
+// This function is adapted from https://github.com/btcsuite/btcd and
+// https://github.com/btcsuite/btcutil
+func GenCertPair(org, certFile, keyFile string, tlsExtraIPs,
```
joostjager: This is a clean move that only exports the funcs? Diff doesn't show diffs. An alternative way that we sometimes use is to do a commit first that is only an unchanged 1:1 copy. Then in the next commit, make the required changes (export, remove old code, reference `cert`)

lnd.go
```diff
@@ -689,12 +668,7 @@ func getTLSConfig(tlsCertPath string, tlsKeyPath string, tlsExtraIPs,
 
 	}
 
-	tlsCfg := &tls.Config{
-		Certificates: []tls.Certificate{certData},
-		CipherSuites: tlsCipherSuites,
-		MinVersion:   tls.VersionTLS12,
-	}
-
+	tlsCfg := cert.TLSConfFromCert(certData)
```
joostjager: Also here it may be easier to review if the refactor is separate from the move. But this is a small commit, so not really a problem.

cert/selfsigned.go
guggero: Changed commit structure to address this.

cert/selfsigned.go
```diff
@@ -0,0 +1,172 @@
...
+	tlsExtraDomains []string) error {
+
+	log.Infof("Generating TLS certificates...")
```
wpaulino: Any reason we need these log messages to happen here? We could have them one layer above this (the caller of this function) and get rid of the `build` dependency on the module.

cert/selfsigned.go
joostjager: Yes, much more confident about the changes with this new structure

cert/tls.go
```diff
@@ -0,0 +1,53 @@
...
+		tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+		tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+	}
+)
+
+// LoadCert loads a certificate and its corresponding private key from the PEM
+// files indicated and returns the certificate in the two formats it is most
+// commonly used.
+func LoadCert(tlsCertPath, tlsKeyPath string) (tls.Certificate,
```
joostjager: Hmm, if this is intended as a 1:1 copy commit, this func isn't a copy?

cert/tls.go
```diff
@@ -26,20 +26,26 @@ var (
...
-func LoadCert(tlsCertPath, tlsKeyPath string) (tls.Certificate,
-	*x509.Certificate, error) {
+func LoadCert(certPath, keyPath string) (tls.Certificate, *x509.Certificate,
+	error) {
 
-	certData, err := tls.LoadX509KeyPair(tlsCertPath, tlsKeyPath)
+	// The certData returned here is just a wrapper around the PEM blocks
+	// loaded from the file. The PEM is not yet fully parsed but a basic
+	// check is performed that the certificate and private key actually
+	// belong together.
```
joostjager: Why not add comments in commit where this code is introduced? (assuming it is indeed new)

guggero:
I cleaned up the commit structure. Now the refactor happens before the move.

cert/selfsigned.go
guggero: I moved the log statements outside of the function and removed the logger in the `cert` package completely.
The only log statement that is now missing is the following, when the resolving of the local hostname fails:
```
		rpcsLog.Errorf("Failed getting hostname, falling back to "+
			"localhost: %v", err)
```
But I think we can live without that.

lnd.go
```diff
@@ -695,21 +698,19 @@ func getTLSConfig(tlsCertPath string, tlsKeyPath string, tlsExtraIPs,
 			return nil, nil, "", err
 		}
 
+		rpcsLog.Infof("Generating TLS certificates...")
```
wpaulino: Nit: s/Generating/Renewing/ to distinguish log messages?

go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	github.com/kkdai/bstream v0.0.0-20181106074824-b3251f7901ec
 	github.com/lightninglabs/neutrino v0.10.1-0.20191111074637-d9b036bf675c
 	github.com/lightningnetwork/lightning-onion v0.0.0-20190909101754-850081b08b6a
+	github.com/lightningnetwork/lnd/cert v0.0.0-00010101000000-000000000000
```
wpaulino: You should be able to push a tag yourself for this and reference it here.

guggero: I didn't know I had the rights to do so. Pushed the tag `cert/v1.0.0` for 
b145378.

PR #3719, opened 2019-11-13
chanbackup: encode broadcast height in chan ID for unconfirmed channels

guggero:
Replacement for/extension of #3690.

If a channel is unconfirmed, the block height of the ShortChannelID in the SCB is zero. This will lead to problems when trying to restore that channel as the spend notifier would get a height hint of zero.
To work around that problem, we add the channel broadcast height to the channel ID so we can use that as height hint on restore.

I tested this on regtest and it seems to work as expected.

### DIFF
--- a/chanbackup/single.go
+++ b/chanbackup/single.go
@@ -66,6 +66,9 @@ type Single struct {
 	// ShortChannelID encodes the exact location in the chain in which the
 	// channel was initially confirmed. This includes: the block height,
 	// transaction index, and the output within the target transaction.
+	// Channels that were not confirmed at the time of backup creation will
+	// have the funding TX broadcast height set as their block height in
+	// the ShortChannelID.
 	ShortChannelID lnwire.ShortChannelID
 
 	// RemoteNodePub is the identity public key of the remote node this
@@ -126,11 +129,21 @@ func NewSingle(channel *channeldb.OpenChannel,
 	// key.
 	_, shaChainPoint := btcec.PrivKeyFromBytes(btcec.S256(), b.Bytes())
 
+	// If a channel is unconfirmed, the block height of the ShortChannelID
+	// is zero. This will lead to problems when trying to restore that
+	// channel as the spend notifier would get a height hint of zero.
+	// To work around that problem, we add the channel broadcast height
+	// to the channel ID so we can use that as height hint on restore.
+	chanID := channel.ShortChanID()
+	if chanID.BlockHeight == 0 {
+		chanID.BlockHeight = channel.FundingBroadcastHeight
+	}
+
 	single := Single{
 		IsInitiator:     channel.IsInitiator,
 		ChainHash:       channel.ChainHash,
 		FundingOutpoint: channel.FundingOutpoint,
-		ShortChannelID:  channel.ShortChannelID,
+		ShortChannelID:  chanID,
 		RemoteNodePub:   channel.IdentityPub,
 		Addresses:       nodeAddrs,
 		Capacity:        channel.Capacity,
--- a/chanbackup/single_test.go
+++ b/chanbackup/single_test.go
@@ -410,4 +410,44 @@ func TestSinglePackStaticChanBackups(t *testing.T) {
 	}
 }
 
+// TestSingleUnconfirmedChannel tests that unconfirmed channels get serialized
+// correctly by encoding the funding broadcast height as block height of the
+// short channel ID.
+func TestSingleUnconfirmedChannel(t *testing.T) {
+	t.Parallel()
+
+	var fundingBroadcastHeight = uint32(1234)
+
+	// Let's create an open channel shell that contains all the information
+	// we need to create a static channel backup but simulate an
+	// unconfirmed channel by setting the block height to 0.
+	channel, err := genRandomOpenChannelShell()
+	if err != nil {
+		t.Fatalf("unable to gen open channel: %v", err)
+	}
+	channel.ShortChannelID.BlockHeight = 0
+	channel.FundingBroadcastHeight = fundingBroadcastHeight
+
+	singleChanBackup := NewSingle(channel, []net.Addr{addr1, addr2})
+	keyRing := &mockKeyRing{}
+
+	// Pack it and then unpack it again to make sure everything is written
+	// correctly, then check that the block height of the unpacked
+	// is the funding broadcast height we set before.
+	var b bytes.Buffer
+	if err := singleChanBackup.PackToWriter(&b, keyRing); err != nil {
+		t.Fatalf("unable to pack single: %v", err)
+	}
+	var unpackedSingle Single
+	err = unpackedSingle.UnpackFromReader(&b, keyRing)
+	if err != nil {
+		t.Fatalf("unable to unpack single: %v", err)
+	}
+	if unpackedSingle.ShortChannelID.BlockHeight != fundingBroadcastHeight {
+		t.Fatalf("invalid block height. got %d expected %d.",
+			unpackedSingle.ShortChannelID.BlockHeight,
+			fundingBroadcastHeight)
+	}
+}
+
 // TODO(roasbsef): fuzz parsing
--- a/chanrestore.go
+++ b/chanrestore.go
@@ -2,9 +2,11 @@ package lnd
 
 import (
 	"fmt"
+	"math"
 	"net"
 
 	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/lightningnetwork/lnd/chanbackup"
 	"github.com/lightningnetwork/lnd/channeldb"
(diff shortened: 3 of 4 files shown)


chanbackup/single_test.go
```diff
@@ -410,4 +410,44 @@ func TestSinglePackStaticChanBackups(t *testing.T) {
...
+	var b bytes.Buffer
+	if err := singleChanBackup.PackToWriter(&b, keyRing); err != nil {
+		t.Fatalf("unable to pack single: %v", err)
+	}
+	var unpackedSingle Single
+	err = unpackedSingle.UnpackFromReader(&b, keyRing)
+	if err != nil {
+		t.Fatalf("unable to unpack single: %v", err)
+	}
+	if unpackedSingle.ShortChannelID.BlockHeight != fundingBroadcastHeight {
```
wpaulino: We should update the `godoc` of `chanbackup.Single.ShortChannelID` to note its new behavior.

chanbackup/single.go
```diff
@@ -125,12 +125,22 @@ func NewSingle(channel *channeldb.OpenChannel,
 	// go to recover, we'll present this in order to derive the private
 	// key.
 	_, shaChainPoint := btcec.PrivKeyFromBytes(btcec.S256(), b.Bytes())
+	
+	// If a channel is unconfirmed, the block height of the ShortChannelID
+	// is zero. This will lead to problems when trying to restore that
+	// channel as the spend notifier would get a height hint of zero.
+	// To work around that problem, we add the channel broadcast height
+	// to the channel ID so we can use that as height hint on restore.
+	chanId := lnwire.NewShortChanIDFromInt(channel.ShortChanID().ToUint64())
```
cfromknecht:
```suggestion
	chanId := channel.ShortChanID()
```

chanbackup/single.go
```diff
@@ -125,12 +125,22 @@ func NewSingle(channel *channeldb.OpenChannel,
...
+	if chanId.BlockHeight == 0 {
+		chanId.BlockHeight = channel.FundingBroadcastHeight
```
cfromknecht: should we do something else signal that this is a pending channel and not the real channel id? we could set the high order bit, or make the assumption that channels won't be in coinbase outputs and detect them via `TxIndex == 0`.

i'm not super familiar with the restore logic, but is it possible that this invalid short channel id gets used in incorrect places after restoring? i'd be concerned that certain subsystems make assumptions about the validity of the short channel ids they receive

chanbackup/single.go
guggero: Ah, since the return value of `channel.ShortChanID()` is not a pointer, we get a clone of the value automatically?

chanbackup/single_test.go
guggero: Done.

chanbackup/single.go
guggero: I'm not at all familiar with the restore logic. That's why I originally favored the second option mentioned in #3690. But the second option means adding a new version to SCBs or switch to TLV encoding entirely. Both seemed like larger changes than using the short channel ID as a workaround. But I'm still not sure which way to go.

What you are proposing is that on restore, we extract the height hint from the short channel ID but then set the block height back to 0 if `TxIndex == 0` before handing it to the channel manager? That sounds like a nice workaround too. How likely is it that there would ever be a channel opening in a coinbase TX? You wouldn't be able to close the channel for 100 blocks, but other than that it's possible in theory even if I don't see the use case yet...

chanbackup/single.go
cfromknecht: Exactly :)

Roasbeef:
So the main side effect of this route is that upon _recovery_ the graph will have the incorrect SID. However, I don't think that has much of an impact since the channels will be closing shortly anyway (ideally) after recovery.

chanbackup/single.go
Roasbeef: > i'm not super familiar with the restore logic, but is it possible that this invalid short channel id gets used in incorrect places after restoring

The impact will be the channel that's inserted into the channel graph after recovery, and also the final close summary for a channel. IMO those two areas are mostly cosmetic, so they shouldn't affect any operational logic within the system. Closed channels are also looked up according to their channel point, which is stable, so they'll still be able to be loacted.

Roasbeef:
We'll end up using the broadcast height as the SCID here: https://github.com/lightningnetwork/lnd/blob/38e313a869538ce4a6d8d85ae26e81c5d49d0e3c/channeldb/db.go#L970.

chanbackup/single.go
```diff
@@ -126,11 +129,21 @@ func NewSingle(channel *channeldb.OpenChannel,
 	// key.
 	_, shaChainPoint := btcec.PrivKeyFromBytes(btcec.S256(), b.Bytes())
 
+	// If a channel is unconfirmed, the block height of the ShortChannelID
+	// is zero. This will lead to problems when trying to restore that
+	// channel as the spend notifier would get a height hint of zero.
+	// To work around that problem, we add the channel broadcast height
+	// to the channel ID so we can use that as height hint on restore.
+	chanID := channel.ShortChanID()
+	if chanID.BlockHeight == 0 {
```
Roasbeef: Also occurs to me that this is needed for the case where a user has a _dated_ backup, and the channel has already confirmed, but was still pending at the time the backup was created. They should still be able to use this to close+restore the ultimate channel (for both the initiator and the responder).

guggero: Yes, exactly.

guggero:
I really liked @cfromknecht's idea. And since it is quite unlikely that there would be a coinbase funding TX at this point in time, I see this as the easier fix than not knowing exactly what happens with the "invalid" SCID.
I added 997fffc4 that restores the `BlockHeight` of the SCID before inserting it into the channel DB.

> So the plan is to defer the version upgrade to include the field until we decide to do TLV SCBs?

Yes, my proposal is that the next time we need to add any new feature to SCBs, we switch the format to TLV and encode the broadcast height properly. Although we probably would need to keep these workarounds for a while to still support people who have dated backups.

chanrestore.go
```diff
@@ -96,6 +96,18 @@ func (c *chanDBRestorer) openChannelShell(backup chanbackup.Single) (
...
+	// It is extremely unlikely at this point that a channel we are trying
+	// to restore has a coinbase funding TX. Therefore we can be quite
+	// certain that if the TxIndex is zero, it was an unconfirmed channel
+	// where we used the BlockHeight to encode the funding TX broadcast
+	// height. To not end up with an invalid short channel ID that looks
+	// valid, we restore the "original" unconfirmed one here.
+	fundingBroadcastHeight := uint32(0)
+	if backup.ShortChannelID.TxIndex == 0 {
+		fundingBroadcastHeight = backup.ShortChannelID.BlockHeight
+		backup.ShortChannelID.BlockHeight = 0
```
wpaulino: We may want to consolidate this logic with https://github.com/lightningnetwork/lnd/pull/3690 to update the field in a single place.

chanbackup/single.go
cfromknecht: what if someone has two pending channels with the same broadcast height? inserting into the graph will fail bc the edge already exists. my main concern is an invalid short chan id being used in the links

chanrestore.go
cfromknecht: nice work @guggero, i think this will greatly limit the potential for any unintended side effects :)

guggero: @wpaulino: right, makes sense. Closed #3690 and will add the changes to this PR.

guggero:
Merged the code from #3690 into this and cleaned it up a little.

chanbackup/single.go
wpaulino: @cfromknecht there shouldn't be any interaction in the link other than sending `ChannelReestablish`, and the links are indexed by their channel ID. Is there something else you had in mind?

guggero:
I added an itest case and confirmed that it fails without the fixes in this PR.

lntest/itest/lnd_test.go
```diff
@@ -13943,25 +13947,44 @@ func testChanRestoreScenario(t *harnessTest, net *lntest.NetworkHarness,
...
+	switch {
+	case testCase.unconfirmed:
+		ctxt, _ = context.WithTimeout(ctxb, channelOpenTimeout)
+		update, err := net.OpenPendingChannel(
+			ctxt, from, to, chanAmt, pushAmt,
+		)
+		if err != nil {
+			t.Fatalf("couldn't open pending channel: %v", err)
+		}
+		if len(update.Txid) == 0 {
```
wpaulino: Nit: is this check needed for anything? I'd assume that if err == nil then things should be fine.

lntest/itest/lnd_test.go
```diff
@@ -14320,6 +14343,56 @@ func testChannelBackupRestore(net *lntest.NetworkHarness, t *harnessTest) {
...
+				req := &lnrpc.ChanBackupExportRequest{}
+				chanBackup, err := oldNode.ExportAllChannelBackups(
+					ctxb, req,
+				)
+				if err != nil {
+					return nil, fmt.Errorf("unable to obtain "+
+						"channel backup: %v", err)
+				}
+				chanPoints := chanBackup.MultiChanBackup.ChanPoints
+				if len(chanPoints) == 0{
```
wpaulino: Looks like this wasn't formatted properly. Is the linter not being run on the itests?

lntest/itest/lnd_test.go
guggero: No, you're right, this is probably not needed.

lntest/itest/lnd_test.go
guggero: Oops! It seems the lntest package is excluded from the linter. Fixed it.

Roasbeef:
itest failure: 
> lntest/itest/lnd_test.go:13956:3: update declared and not used


Need to resurrect https://github.com/lightningnetwork/lnd/pull/3382 so we can catch these on save like we do the rest of our files.

guggero:
Fixed the itest.

PR #3721, opened 2019-11-14
rpc+cnct: update AbandonChannel to also remove cnct and channel graph state 

### LINKED ISSUE #3716: Extend AbandonChannel RPC to remove all channel-related state
Users that have previously used the `AbandonChannel` RPC may leave their node in an inconsistent state due to it not removing all channel-related state. This became more of an issue when users use the RPC on invalid channels that were exploited by the recent CVE. The RPC only removes the channel state, but still leaves behind its graph and contract court state.

Roasbeef:
In this commit, we update the `AbandonChannel` method to also remove the
state from the countract court as well as the channel graph. Abandoning
a channel is now a three step process: remove from the open hannel
state, remove from the graph, remove from the contract court. Between
any step it's possible that the users restarts the process all over
again. As a result, each of the steps below are intended to be
idempotent. The final step will always fail once the log of the channel
arbitrator has been clead up, giving the user a signal that all channel
state has been removed.

We also update the integration test to assert that no channel is found
in the graph any longer. Before this commit, this test would fail as the
channel was still found in the graph, which can cause other issues for
an operational daemon.

Fixes #3716.

### DIFF
--- a/contractcourt/chain_arbitrator.go
+++ b/contractcourt/chain_arbitrator.go
@@ -296,7 +296,7 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 	}
 
 	arbCfg.MarkChannelResolved = func() error {
-		return c.resolveContract(chanPoint, chanLog)
+		return c.ResolveContract(chanPoint)
 	}
 
 	// Finally, we'll need to construct a series of htlc Sets based on all
@@ -321,11 +321,10 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 	), nil
 }
 
-// resolveContract marks a contract as fully resolved within the database.
+// ResolveContract marks a contract as fully resolved within the database.
 // This is only to be done once all contracts which were live on the channel
 // before hitting the chain have been resolved.
-func (c *ChainArbitrator) resolveContract(chanPoint wire.OutPoint,
-	arbLog ArbitratorLog) error {
+func (c *ChainArbitrator) ResolveContract(chanPoint wire.OutPoint) error {
 
 	log.Infof("Marking ChannelPoint(%v) fully resolved", chanPoint)
 
@@ -338,26 +337,43 @@ func (c *ChainArbitrator) resolveContract(chanPoint wire.OutPoint,
 		return err
 	}
 
+	// Now that the channel has been marked as fully closed, we'll stop
+	// both the channel arbitrator and chain watcher for this channel if
+	// they're still active.
+	var arbLog ArbitratorLog
+	c.Lock()
+	chainArb := c.activeChannels[chanPoint]
+	delete(c.activeChannels, chanPoint)
+
+	chainWatcher := c.activeWatchers[chanPoint]
+	delete(c.activeWatchers, chanPoint)
+	c.Unlock()
+
+	if chainArb != nil {
+		arbLog = chainArb.log
+
+		if err := chainArb.Stop(); err != nil {
+			log.Warnf("unable to stop ChannelArbitrator(%v): %v",
+				chanPoint, err)
+		}
+	}
+	if chainWatcher != nil {
+		if err := chainWatcher.Stop(); err != nil {
+			log.Warnf("unable to stop ChainWatcher(%v): %v",
+				chanPoint, err)
+		}
+	}
+
+	// Once this has been marked as resolved, we'll wipe the log that the
+	// channel arbitrator was using to store its persistent state. We do
+	// this after marking the channel resolved, as otherwise, the
+	// arbitrator would be re-created, and think it was starting from the
+	// default state.
 	if arbLog != nil {
-		// Once this has been marked as resolved, we'll wipe the log
-		// that the channel arbitrator was using to store its
-		// persistent state. We do this after marking the channel
-		// resolved, as otherwise, the arbitrator would be re-created,
-		// and think it was starting from the default state.
 		if err := arbLog.WipeHistory(); err != nil {
 			return err
 		}
 	}
-
-	c.Lock()
-	delete(c.activeChannels, chanPoint)
-
-	chainWatcher, ok := c.activeWatchers[chanPoint]
-	if ok {
-		chainWatcher.Stop()
-	}
-	delete(c.activeWatchers, chanPoint)
-	c.Unlock()
 
 	return nil
 }
@@ -491,7 +507,7 @@ func (c *ChainArbitrator) Start() error {
 			return err
 		}
 		arbCfg.MarkChannelResolved = func() error {
-			return c.resolveContract(chanPoint, chanLog)
+			return c.ResolveContract(chanPoint)
 		}
 
 		// We can also leave off the set of HTLC's here as since the
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -13155,17 +13155,42 @@ func testAbandonChannel(net *lntest.NetworkHarness, t *harnessTest) {
 
 	ctxt, _ := context.WithTimeout(ctxb, channelOpenTimeout)
 	chanPoint := openChannelAndAssert(
-		ctxt, t, net, net.Alice, net.Bob, channelParam)
+		ctxt, t, net, net.Alice, net.Bob, channelParam,
+	)
+	txid, err := lnd.GetChanPointFundingTxid(chanPoint)
+	if err != nil {
+		t.Fatalf("unable to get txid: %v", err)
+	}
+	chanPointStr := fmt.Sprintf("%v:%v", txid, chanPoint.OutputIndex)
 
 	// Wait for channel to be confirmed open.
 	ctxt, _ = context.WithTimeout(ctxb, defaultTimeout)
-	err := net.Alice.WaitForNetworkChannelOpen(ctxt, chanPoint)
+	err = net.Alice.WaitForNetworkChannelOpen(ctxt, chanPoint)
 	if err != nil {
 		t.Fatalf("alice didn't report channel: %v", err)
 	}
 	err = net.Bob.WaitForNetworkChannelOpen(ctxt, chanPoint)
 	if err != nil {
 		t.Fatalf("bob didn't report channel: %v", err)
+	}
+
+	// Now that the channel is open, we'll obtain its channel ID real quick
+	// so we can use it to query the graph below.
+	listReq := &lnrpc.ListChannelsRequest{}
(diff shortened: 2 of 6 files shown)


cfromknecht:
Should this also clean up the nursery?

Roasbeef:
Hmm, just to be safe, I think so. As people typically use this RPC for when they have a channel whose commitment transaction cannot be broadcast, typically they channel won't have any outputs that have been handed off to the nursery.

rpcserver.go
```diff
@@ -1946,6 +1947,93 @@ func createRPCCloseUpdate(update interface{}) (
...
+		ShortChanID:             dbChan.ShortChanID(),
+		RemoteCurrentRevocation: dbChan.RemoteCurrentRevocation,
+		RemoteNextRevocation:    dbChan.RemoteNextRevocation,
+		LocalChanConfig:         dbChan.LocalChanCfg,
+	}
+
+	// Finally, we'll close the channel in the DB, and return back to the
+	// caller.
+	return dbChan.CloseChannel(summary)
+}
```
joostjager: Should this be a method on `channeldb.DB` where `bestHeight` is passed in? `rpcserver.go` is already so big.

rpcserver.go
```diff
@@ -1946,6 +1947,93 @@ func createRPCCloseUpdate(update interface{}) (
...
+	case err == channeldb.ErrEdgeNotFound:
+		return nil
+	case err != nil:
+		return err
+	}
+
+	// If the channel ID is still in the graph, then that means the channel
+	// is still open, so we'll now move to purge it from the graph.
+	return chanGraph.DeleteChannelEdges(chanID)
+}
```
joostjager: method on `ChannelGraph`?

rpcserver.go
```diff
@@ -1946,6 +1947,93 @@ func createRPCCloseUpdate(update interface{}) (
...
+// abandonChanFromArbitrator attempts to remove all active contractcourt state
+// concerning the passed channel point. This is the last abandon channel method
+// to be called as it will fail if the channel state has already been removed.
+// This serves to notify the user if they channel has already been _fully_
+// abandoned.
+func abandonChanFromArbitrator(chainArb *contractcourt.ChainArbitrator,
+	chanPoint *wire.OutPoint) error {
+
+	return chainArb.ResolveContract(*chanPoint)
+}
```
joostjager: Does this function add anything?

rpcserver.go
```diff
@@ -1946,6 +1947,93 @@ func createRPCCloseUpdate(update interface{}) (
...
+	// If the channel wasn't found, then it's possible that it was already
+	// abandoned from the database.
+	case err == channeldb.ErrChannelNotFound:
+		_, closedErr := db.FetchClosedChannel(chanPoint)
+		if closedErr != nil {
+			return closedErr
+		}
+
+		// If the channel was already closed, then we don't return an
+		// error as we'd like fro this step to be repeatable.
```
carlaKC: /s fro -> for

lntest/itest/lnd_test.go
```diff
@@ -13168,6 +13169,16 @@ func testAbandonChannel(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("bob didn't report channel: %v", err)
 	}
 
+	// Now that the chanenl is open, we'll obtain its channel ID real quick
```
carlaKC: /s chanenl -> channel

rpcserver.go
carlaKC: +1 for moving these functions out of `rpcserver`

bhandras: +1

rpcserver.go
bhandras: +1, feels unrelated to rpcserver and could be unit tested on its own.

rpcserver.go
bhandras: Agree with @joostjager here, as this is a one line function, unless called multiple times, you might consider inlinging it.

rpcserver.go
```diff
@@ -1969,39 +2057,26 @@ func (r *rpcServer) AbandonChannel(ctx context.Context,
...
 	if err != nil {
 		return nil, err
 	}
-
-	// Now that we've found the channel, we'll populate a close summary for
-	// the channel, so we can store as much information for this abounded
-	// channel as possible. We also ensure that we set Pending to false, to
-	// indicate that this channel has been "fully" closed.
-	_, bestHeight, err := r.server.cc.chainIO.GetBestBlock()
+	err = abandonChanFromGraph(
```
bhandras: Maybe dumb question due to my unfamiliarity with the codebase, but what happens if the abandonChanFromDatabase suceeds but abandonChanFromGraph fails?

rpcserver.go
```diff
@@ -1969,39 +2057,26 @@ func (r *rpcServer) AbandonChannel(ctx context.Context,
...
-		ShortChanID:             dbChan.ShortChanID(),
-		RemoteCurrentRevocation: dbChan.RemoteCurrentRevocation,
-		RemoteNextRevocation:    dbChan.RemoteNextRevocation,
-		LocalChanConfig:         dbChan.LocalChanCfg,
-	}
-
-	// Finally, we'll close the channel in the DB, and return back to the
-	// caller.
-	err = dbChan.CloseChannel(summary)
+	err = abandonChanFromArbitrator(
```
bhandras: Similarly to my previous comment, what happens if every "abandon" state change succeeds but abandonChanFromArbitrator fails? In what case can this happen and should this be tested?

contractcourt/chain_arbitrator.go
```diff
@@ -323,11 +323,10 @@ func newActiveChannelArbitrator(channel *channeldb.OpenChannel,
 	), nil
 }
 
-// resolveContract marks a contract as fully resolved within the database.
+// ResolveContract marks a contract as fully resolved within the database.
 // This is only to be done once all contracts which were live on the channel
 // before hitting the chain have been resolved.
-func (c *ChainArbitrator) resolveContract(chanPoint wire.OutPoint,
-	arbLog ArbitratorLog) error {
+func (c *ChainArbitrator) ResolveContract(chanPoint wire.OutPoint) error {
```
bhandras: Q: Should this public method have a unit test as well? (Not sure about the existing test coverage)

contractcourt/chain_arbitrator.go
```diff
@@ -340,18 +339,17 @@ func (c *ChainArbitrator) resolveContract(chanPoint wire.OutPoint,
...
-		// and think it was starting from the default state.
-		if err := arbLog.WipeHistory(); err != nil {
-			return err
-		}
-	}
-
+	// Now that the channel has been marked as fully closed, we'll stop
+	// both the channel arbitrator and chain watcher for this channel if
+	// they're still active.
+	var arbLog ArbitratorLog
```
bhandras: It seems that locking is unnecessary for chainArb.Stop() and chainWatcher.Stop(), and should only be used while removing the chainArb and chainWatcher from their respective maps. (Please discard my comment if this is untrue and the whole process needs locking.).

lntest/itest/lnd_test.go
```diff
@@ -13155,7 +13155,8 @@ func testAbandonChannel(net *lntest.NetworkHarness, t *harnessTest) {
 
 	ctxt, _ := context.WithTimeout(ctxb, channelOpenTimeout)
 	chanPoint := openChannelAndAssert(
-		ctxt, t, net, net.Alice, net.Bob, channelParam)
+		ctxt, t, net, net.Alice, net.Bob, channelParam,
```
bhandras: nit: maybe you wanted to add a param but deleted it? (sorry, this may be a stylistic thing I'm not familar with)

lntest/itest/lnd_test.go
```diff
@@ -13230,9 +13240,18 @@ func testAbandonChannel(net *lntest.NetworkHarness, t *harnessTest) {
...
 	}
 
-	// Now that we're done with the test, the channel can be closed. This is
-	// necessary to avoid unexpected outcomes of other tests that use Bob's
-	// lnd instance.
+	// Ensure that the channel can no longer be found in the channel graph.
+	_, err = net.Alice.GetChanInfo(ctxb, &lnrpc.ChanInfoRequest{
+		ChanId: chanID,
+	})
+	if err == nil {
```
bhandras: Consider having an assert for the specific error instead.

Roasbeef:
> Pretty trivial, but is abandonChanFromDatabase strictly idempotent if it will update the close height on channels that have already been abandoned?

Will it? The second time around, it'll see that they're already in the close channel bucket and return early.

Roasbeef:
> Should this also clean up the nursery?

State now removed in the latest diff.

lntest/itest/lnd_test.go
Roasbeef: Just updating older code to conform to the new style guidelines.

contractcourt/chain_arbitrator.go
Roasbeef: Updated to only hold the lock while reading from the map. All other operations directly on the objects are goroutine safe.

contractcourt/chain_arbitrator.go
Roasbeef: Test added!

rpcserver.go
Roasbeef: Same response: the user can pick up again where things left off. However if the user calls the RPC _again_ (after the channel has actually already been abandoned), they won't get any error as the channel has already been removed.

rpcserver.go
Roasbeef: Great question! If `abandonChanFromGraph` the RPC call will return with an error. At that point, depending on the nature of the failure, the caller can re-try and essentially have the logic pick up where it left off. 

For those that have already used the existing version of this call, the first abandonment phase will be a noop, and they'll move onto removing things from the channel graph.

rpcserver.go
Roasbeef: When I started it, I expected to have to do more in the body, but turns out `ResolveContract` did most of the heavy lifting for us. Ultimately, I still like the function though as its `godoc` gives information to the reader which would otherwise be more scrunched in the main `AbandonChannel` method.

rpcserver.go
Roasbeef: Both of these methods are already tested within the `channeldb` package. We simply compose two calls into a greater method for our use case. I don't think there're any other call sites (at the time of writing) that would utilize these back-to-back calls.

Roasbeef:
Updated to also remove the channel from the link/peer before modifying on-disk state (making sure to mark as borked before doing anything so we don't re-load the channel into memory).

PR #3722, opened 2019-11-14
lnrpc+lnwallet+funding: implement new RPCs to enable external channel funding

Roasbeef:
In this commit, we start to expose some of the new external funding
functionality over the RPC interface.

First, we add a new `funding_shim` field to the regular `OpenChannel`
method. This can be used by a caller to express that certain parameters
of the funding flow have already been negotiated outside the protocol,
and should be used instead. For example, a shim can be provided to use a
particular key for the commitment key (ideally cold) rather than use one
this is generated by the wallet as normal, or signal that signing will
be carried out in an interactive manner (PSBT based).

Next, we add a brand new method: `FundingStateStep`. FundingStateStep is
an advanced funding related call that allows the caller to either
execute some preparatory steps for a funding workflow, or manually
progress a funding workflow. The primary way a funding flow is
identified is via its pending channel ID. As an example, this method can
be used to specify that we're expecting a funding flow for a particular
pending channel ID, for which we need to use specific parameters.
Alternatively, this can be used to interactively drive PSBT signing for
funding for partially complete funding transactions.

The new transition methods (funding state machine modifiers) in this
commit allow a party to register a funding intent that should be used
for a specified incoming pending channel ID. The "responder" to the
external channel flow should use this to prep lnd to be able to handle
the channel flow properly.

**NOTE:** this is an early version of the PR without additional tests in order to get early feedback on the proposed cut outs to enable a class of external channel funding useful in channel factory like contexts.  

This builds on #3659.


lnrpc/rpc.proto
```diff
@@ -1579,14 +1595,82 @@ message OpenChannelRequest {
...
+    ChannelPoint chan_point = 2;
+
+    /// Our local key to use when creating the multi-sig output.
+    signrpc.KeyDescriptor local_key = 3;
+
+    /// The key of the remote party to use when creating the multi-sig output.
+    bytes remote_key = 4;
+
+    /**
+    If non-zero, then this will be used as the pending chanenl ID on the wire
```
alexbosworth:
```suggestion
    If non-zero, then this will be used as the pending channel ID on the wire
```

Roasbeef:
Oh also only the last two commits are new.

lnrpc/rpc.proto
```diff
@@ -1579,14 +1595,82 @@ message OpenChannelRequest {
 
     /// Whether unconfirmed outputs should be used as inputs for the funding transaction.
     bool spend_unconfirmed = 12 [json_name = "spend_unconfirmed"];
+
+    /**
+    Funding shims are an optional argument that allow the caller to intercept
+    certain funding functionality. For example, a shim can be provided to use a
+    particular key for the commitment key (ideally cold) rather than use one
+    this is generated by the wallet as normal, or signal that signing will be
```
guggero: nit: s/this/that/

lnrpc/rpc.proto
```diff
@@ -1579,14 +1595,82 @@ message OpenChannelRequest {
...
+    carried out in an interactive manner (PSBT based).
+    */
+    FundingShim funding_shim = 13;
```
guggero: What was the original reason for all the `[json_name = "..."]` in this file? Are they still needed? If yes, then they are missing throughout the changes for this PR.

lnrpc/rpc.proto
```diff
@@ -1579,14 +1595,82 @@ message OpenChannelRequest {
...
+    /**
+    The pending channel ID of the created channel. This value may be used to
+    further the funding flow manually via teh FundingStateStep method.
+    */
+    bytes pending_chan_id = 4 [json_name = "pending_chan_id"];
+}
+
+message ChanPointShim {
+    /**
+    The size of the pre-crafted output to used a the channel point for this
```
guggero: nit: s/to used a the/to be used as the/

lnrpc/rpc.proto
```diff
@@ -1579,14 +1595,82 @@ message OpenChannelRequest {
...
+}
+
+message FundingShim {
+    oneof shim {
+        ChanPointShim chan_point_shim = 1;
+    }
+}
+
+message FundingShimCancel {
+    /// The pending channel ID of the channel to cancel the fuding shim for.
```
guggero: nit: s/fuding/funding/

lnrpc/rpc.proto
```diff
@@ -430,10 +431,25 @@ service Lightning {
...
+    FundingStateStep is an advanced funding related call that allows the caller
+    to either execute some preparatory steps for a funding workflow, or
+    manually progress a funding workflow. The primary way a funding flow is
+    identified is via its pending channel ID. As an example, this method can be
+    used to specify that we're expecting a funding flow for a particular
+    pending channel ID, for which we need to use specific parameters.
+    Alternatively, this can be used to interactively drive PSBT signing for
+    funding for partially complete funding transactions.
+    */
+    rpc FundingStateStep(FundingTransitionMsg) returns (FundingStateStepResp);
```
guggero: Commit doesn't compile without implementation:
```
./rpcserver.go:495:5: cannot use (*rpcServer)(nil) (type *rpcServer) as type lnrpc.LightningServer in assignment:
        *rpcServer does not implement lnrpc.LightningServer (missing FundingStateStep method)
```

fundingmanager.go
```diff
@@ -2772,6 +2785,16 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		channelFlags = lnwire.FFAnnounceChannel
 	}
 
+	// If the caller secpfied their own channle ID, then we'll use that.
```
alexbosworth:
```suggestion
	// If the caller specified their own channel ID, then we'll use that.
```

lnrpc/rpc.proto
Roasbeef: They're needed for things that will be displayed by `lncli` on the command line. It's only for JSON encoding of the field names. By default, the proto library will default to using CamelCase, but we use snake_case for our JSON output on the command line, hence the need for specifying the name manually.

lnrpc/rpc.proto
Roasbeef: I don't think we've ever made that a requirement for proto changes, as IMO it's nice to have commits that only touch `lnrpc`, then do the actual implementation as a follow up.

halseth: It's preferable to just add a stub `FundingStateStep` method IMO, since it makes bisecting much more powerful if each commit compiles.

rpcserver.go
```diff
@@ -1424,6 +1426,53 @@ func extractOpenChannelMinConfs(in *lnrpc.OpenChannelRequest) (int32, error) {
...
+	remoteKey, err := btcec.ParsePubKey(
+		chanPointShim.RemoteKey, btcec.S256(),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	shimKeyDesc := chanPointShim.LocalKey
+	localKey, err := btcec.ParsePubKey(
+		shimKeyDesc.RawKeyBytes, btcec.S256(),
```
halseth: no gurantee the rawbytes will be set here

rpcserver.go
```diff
@@ -1424,6 +1426,53 @@ func extractOpenChannelMinConfs(in *lnrpc.OpenChannelRequest) (int32, error) {
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	localKeyDesc := keychain.KeyDescriptor{
+		PubKey: localKey,
+		KeyLocator: keychain.KeyLocator{
+			Family: keychain.KeyFamily(
```
halseth: same here. The local key is either bytes or locator, so i think we must explicitly check that before attempting to read it.

rpcserver.go
```diff
@@ -5445,3 +5445,75 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+		// Using the rpc shim as a template, we'll construct a new
+		// chanfunding.Assembler that is able to express proper
+		// formulation of this expected channel.
+		shimAssembler, err := newFundingShimAssembler(
+			rpcShimIntent, false,
+		)
+		if err != nil {
+			return nil, err
+		}
+		shimIntent, err := shimAssembler.ProvisionChannel(nil)
```
halseth: cannot pass nil here anymore

rpcserver.go
```diff
@@ -5373,3 +5445,75 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+			return nil, err
+		}
+
+		// Once we have the intent, we'll register it with the wallet.
+		// Once we receive an incoming funding request that uses this
+		// pending channel ID, then this shim will be dispatched in
+		// place of our regular funding workflow.
+		var pendingChanID [32]byte
+		copy(pendingChanID[:], rpcShimIntent.PendingChanId)
+		err = r.server.cc.wallet.RegisterFundingIntent(
```
halseth: as noted in the other review, since the canned assembler can also handle responder requests, I think we could register the `shimAssembler` with the wallet here instead of the intent, such that the logic in the reservation code would have one less code path to take.

Roasbeef: Which code path? This is only intended to mark an incoming request for interception. We'll need to track the set of requests to be intercepted _somewhere_, with the other alternative being in the funding manager itself. I prefer doing it in the wallet as it allows all the logic to be tested in a single package, with the funding manager being mostly unaware of it all (notice how it's hardly changed in this diff and the prior PR, that was intended).

Roasbeef:
Updated to be based off of master, won't build yet though until the next push.

rpcserver.go
halseth: Suggestion was to just register the `shimAssembler` instead, but now that the other PR is merged this is no longer an option.

rpcserver.go
Roasbeef: Updated, also modified the checks in `ProvisionChannel` to check that the local+remote amount sum u properly to the full amount.

rpcserver.go
Roasbeef: Added a switch statement before this method is called which performs initial validation.

Roasbeef:
Will add an itest, then this should be ready for final review.

rpcserver.go
```diff
@@ -5494,3 +5588,77 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+	// core protocol.
+	case in.GetShimRegister() != nil &&
+		in.GetShimRegister().GetChanPointShim() != nil:
+
+		rpcShimIntent := in.GetShimRegister().GetChanPointShim()
+
+		// Using the rpc shim as a template, we'll construct a new
+		// chanfunding.Assembler that is able to express proper
+		// formulation of this expected channel.
+		shimAssembler, err := newFundingShimAssembler(
```
guggero: Don't we need the same basic sanity switch statement here too? Or better move that check into `newFundingShimAssembler`.

fundingmanager.go
```diff
@@ -1821,6 +1824,7 @@ func (f *fundingManager) handleFundingSigned(fmsg *fundingSignedMsg) {
 				OutputIndex: fundingPoint.Index,
 			},
 		},
+		PendingChanId: pendingChanID[:],
```
guggero: Should we add the `PendingChanId` to `stateStep` (line 931 ff) also? So the caller knows it's their channel that is now fully open and they can clean up the ship on their side.

rpcserver.go
```diff
@@ -1429,6 +1431,53 @@ func extractOpenChannelMinConfs(in *lnrpc.OpenChannelRequest) (int32, error) {
...
+	}
+
+	shimKeyDesc := chanPointShim.LocalKey
+	localKey, err := btcec.ParsePubKey(
+		shimKeyDesc.RawKeyBytes, btcec.S256(),
+	)
+	if err != nil {
+		return nil, err
+	}
+	localKeyDesc := keychain.KeyDescriptor{
```
guggero: Would it make sense to also add a sanity check to make sure the index/family correspond to the public key provided? Otherwise we might run into strange errors during the funding process down the line. Not a blocker though.

rpcserver.go
Roasbeef: Good point! Moved the sanity checks into `newFundingShimAssembler`.

fundingmanager.go
Roasbeef: Done.

rpcserver.go
Roasbeef: Good idea, added.

Roasbeef:
New integration tests pushed! Along the way, I had to copy over the `KeyLocator` and `KeyDescriptor` structs in order to avoid an import cycle with the `signrpc` package. An alternative is just to move all the new RPC calls to a sub-server otherwise....

rpcserver.go
```diff
@@ -1445,6 +1447,88 @@ func extractOpenChannelMinConfs(in *lnrpc.OpenChannelRequest) (int32, error) {
...
+			Family: keychain.KeyFamily(
+				shimKeyDesc.KeyLoc.KeyFamily,
+			),
+			Index: uint32(shimKeyDesc.KeyLoc.KeyIndex),
+		},
+	}
+
+	// Verify that if we re-derive this key according to the passed
+	// KeyLocator, that we get the exact same key back. Otherwise, we may
+	// end up in a situating where we aren't able to actually sign for this
```
halseth: situation

lntest/itest/lnd_test.go
```diff
@@ -14900,6 +14904,168 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+	// confirm it now for simplicity.
+	const chanSize = lnd.MaxBtcFundingAmount
+	_, fundingOutput, err := input.GenFundingPkScript(
+		carolFundingKey.RawKeyBytes, daveFundingKey.RawKeyBytes,
+		int64(chanSize),
+	)
+	if err != nil {
+		t.Fatalf("unable to create funding script: %v", err)
+	}
+	txid, err := net.Miner.SendOutputsWithoutChange(
```
halseth: a more realistic test would be to publish the funding tx first after the commitment was properly signed

lntest/itest/lnd_test.go
```diff
@@ -14900,6 +14904,168 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+	)
+	if err != nil {
+		t.Fatalf("unable to make payments between Carol and Dave")
+	}
+
+	// To conclude, we'll close the newly created channel between Carol and
+	// Dave.
+	closeChans()
+
+	shutdownAndAssert(net, t, carol)
```
halseth: defer this method after creating the nodes instead.

lntest/itest/lnd_test.go
```diff
@@ -14900,6 +14904,168 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+	err = completePaymentRequests(
+		ctxt, carol, []string{resp.PaymentRequest}, true,
+	)
+	if err != nil {
+		t.Fatalf("unable to make payments between Carol and Dave")
+	}
+
+	// To conclude, we'll close the newly created channel between Carol and
+	// Dave.
+	closeChans()
```
halseth: defer close when declared

lnrpc/rpc.proto
```diff
@@ -1618,14 +1633,104 @@ message OpenChannelRequest {
...
     }
 }
 
+message FundingShimCancel {
+    /// The pending channel ID of the channel to cancel the funding shim for.
+    bytes pending_chan_id = 1;
+}
+
+message FundingTransitionMsg {
+    oneof trigger {
```
halseth: What's the rational behind having a oneof here, instead of just two RPC endpoints? Seems simpler not having to deal with the oneonf, and the switch in the implementation.

rpcserver.go
```diff
@@ -5572,5 +5678,5 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
 func (r *rpcServer) FundingStateStep(context.Context,
 	*lnrpc.FundingTransitionMsg) (*lnrpc.FundingStateStepResp, error) {
 
-	return nil, nil
+	return nil, fmt.Errorf("not implemented")
```
halseth: nit: should be part of prev commit

rpcserver.go
```diff
@@ -5675,8 +5675,68 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
 
-	return nil, fmt.Errorf("not implemented")
+	switch {
+
+	// If this is a message to register a new shim that is an external
+	// channel point, then we'll contact the wallet to register this new
+	// shim. A user will use this method to register a new channel funding
+	// workflow which has already been partially negotiated outside of the
+	// core protocol.
+	case in.GetShimRegister() != nil &&
```
halseth: this could just be two different RPCs instead.

lnrpc/rpc.proto
Roasbeef: That would mean a new RPC method for each new funding variation we think of in the future. This instead presents a set of distinct triggers similar to how we do things in the contract court.

lntest/itest/lnd_test.go
Roasbeef: Sure but we don't have the capabilities to do something like that easily using the current RPC system for both `lnd` and the miner. There's an internal `memWallet` in the main `btcd` harness we can expose for this, but this set gives us a full run through either way. This method only broadcasts the transaction, it's still mined as normal in the regular basic funding tests.

lntest/itest/lnd_test.go
Roasbeef: If the test fails earlier, we don't care about closing these out IMO, but added the defer for the shutdown of the nodes.

fundingmanager.go
```diff
@@ -2853,9 +2859,15 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
...
-	// reservation throughout its lifetime.
-	chanID := f.nextPendingChanID()
+	// If the caller specified their own channel ID, then we'll use that.
+	// Otherwise we'll generate a fresh one as normal.  This will be used
+	// to track this reservation throughout its lifetime.
+	var chanID [32]byte
+	if msg.pendingChanID == zeroID {
+		chanID = f.nextPendingChanID()
+	} else {
+		chanID = msg.pendingChanID
```
wpaulino: Should we perform some additional validation here to ensure there isn't a conflict with existing ones?

rpcserver.go
```diff
@@ -1445,6 +1447,88 @@ func extractOpenChannelMinConfs(in *lnrpc.OpenChannelRequest) (int32, error) {
...
+	// fields are populated.
+	switch {
+	case chanPointShim.RemoteKey == nil:
+		return nil, fmt.Errorf("remote key not set")
+
+	case chanPointShim.LocalKey == nil:
+		return nil, fmt.Errorf("local key desc not set")
+
+	case chanPointShim.LocalKey.RawKeyBytes == nil:
+		return nil, fmt.Errorf("local raw key bytes not set")
```
wpaulino: If a key locator is provided, then this still fails. The method states that either can be provided.

lnwallet/wallet.go
```diff
@@ -456,6 +456,21 @@ func (l *LightningWallet) RegisterFundingIntent(expectedID [32]byte,
...
 }
 
+// CancelFundingIntent allows a caller to cancel a previously registered
+// funding intent. If no intent was found, then an error will be returned.
+func (l *LightningWallet) CancelFundingIntent(pid [32]byte) error {
+	l.intentMtx.Lock()
+	defer l.intentMtx.Unlock()
+
+	if _, ok := l.fundingIntents[pid]; !ok {
+		return fmt.Errorf("no funding intent found")
```
wpaulino: Nit: include pending ID in error.

rpcserver.go
```diff
@@ -5660,8 +5660,68 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+		err := r.server.cc.wallet.CancelFundingIntent(pendingChanID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// TODO(roasbeef): extend PendingChannels to also show shims
+
+	// TODO(roasbeef): return resulting state? also add a method to query
+	// current state?
```
wpaulino: If we go down this route, I think it'd be nice to have a funding related subserver.

lnwallet/chanfunding/canned_assembler.go
```diff
@@ -176,7 +176,9 @@ func (c *CannedAssembler) ProvisionChannel(req *Request) (Intent, error) {
 	// A simple sanity check to ensure the provisioned request matches the
 	// re-made shim intent.
 	if req.LocalAmt+req.RemoteAmt != c.fundingAmt {
-		return nil, fmt.Errorf("intent doesn't match canned assembler")
+		return nil, fmt.Errorf("intent doesn't match canned "+
```
wpaulino: Nit: squash with commit that introduced this.

lntest/itest/lnd_test.go
```diff
@@ -956,7 +958,8 @@ func testOnchainFundRecovery(net *lntest.NetworkHarness, t *harnessTest) {
 // channel closure.
 func basicChannelFundingTest(t *harnessTest, net *lntest.NetworkHarness,
 	alice *lntest.HarnessNode,
-	bob *lntest.HarnessNode) (*lnrpc.Channel, *lnrpc.Channel, func(), error) {
+	bob *lntest.HarnessNode,
```
wpaulino: Nit: move this parameter to line above?

lntest/itest/lnd_test.go
```diff
@@ -14900,6 +14904,169 @@ func testHoldInvoicePersistence(net *lntest.NetworkHarness, t *harnessTest) {
...
+		RawKeyBytes: carolFundingKey.RawKeyBytes,
+		KeyLoc: &lnrpc.KeyLocator{
+			KeyFamily: carolFundingKey.KeyLoc.KeyFamily,
+			KeyIndex:  carolFundingKey.KeyLoc.KeyIndex,
+		},
+	}
+	fundingShim.GetChanPointShim().RemoteKey = daveFundingKey.RawKeyBytes
+
+	// At this point, we'll now carry out the normal basic channel funding
+	// test as everything should now [proceed as normal (a regular channel
```
wpaulino: Nit: `[`

fundingmanager.go
Roasbeef: Done, added here and also in the wallet where the intents are actually stored.

rpcserver.go
Roasbeef: Was left over from where it was copied from (signerrpc), we need both to be populated as is atm, which isn't a big burden as when it's obtained from the wallekitrpc, both are always present.

lnrpc/rpc.proto
halseth: SGTM

Issue #3712, opened 2019-11-12
LND crashed (panic) during QueryRoutes / SendToRouteSync


LNBIG-COM:
### Background

I had intense code executions `QueryRoutes` / `SendToRouteSync` with big lists of ignored_pairs & ignored_nodes. One from LNDs (lnd-01) was crashed

### Your environment

* version of `lnd` - 0.8.0-beta

### Steps to reproduce

May be intense rebalancing through QueryRoutes (with `ignored_pairs` &amp; `ignored_nodes`) -&gt; SendToRouteSync executions.

stacktrace:

```
2019-11-12 12:53:08.431 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:08.819 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:09.160 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:09.694 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:09.823 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:10.453 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:10.505 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:10.878 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:10.944 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:11.204 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:11.723 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:11.912 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
2019-11-12 12:53:12.307 [ERR] RPCS: [/lnrpc.Lightning/QueryRoutes]: unable to find a path to destination
[...]

PR #3724, opened 2019-11-14
sphinx: embed lightning-onion repo

joostjager:
There is no reason to have a separate repository for our sphinx-related code but it does add overhead to update two repositories when changes are required.

Users of just the former `lightning-onion` repository can reference the `lnd` repository to keep access to the package.

This PR moves the sphinx package into the `lnd` repository. It also removes a file that was copied over from the `tlv` package and eliminates the risk of those getting out of sync.

Embedding the repo also showed pre-existing linter errors. Those have been disabled for this PR and can be corrected later.

Issues that (may) touch sphinx and can be kept in a single repo with this PR:
* crash when exceeded max onion blob size: https://github.com/lightningnetwork/lnd/issues/3712
* tracking of the onion blob size during pathfinding https://github.com/lightningnetwork/lnd/issues/3725
* privacy leak: https://github.com/lightningnetwork/lightning-onion/pull/40


cfromknecht:
@joostjager what's the status of this? should i review?

joostjager:
No.

joostjager:
Closing again, need to collect more votes for this first.

PR #3727, opened 2019-11-14
Add StartBeforeSynced configuration flag

yaslama:
When StartBeforeSynced is set, let lnd daemon completely start even if the
chain is not fully synchronized.


wpaulino:
Could you explain the motivation behind this change? We currently wait for safety reasons.

yaslama:
The main motivation for this configuration flag is to improve the user experience when running lnd in a mobile device. If a user doesn't use the app for more than few days and then open it, she needs to wait a lot (more than several minutes depending on her network connection) in order to do any operation.

And this is the case, even if the user is willing to take the risk that the channel is closed and want to make a payment. In fact, there is a way for the user to check - in a less secure way - that the channel is not closed by querying external blockchain explorers before making the payment.

In fact, even without this configuration parameter, the same situation can arise when lnd is running and there is some connectivity problem for an important period of time. synced_to_chain will be false, but lnd will work.

PR #3737, opened 2019-11-18
chanbackup: continue recovery if channel already exists

guggero:
While recovering a multi backup file, if for some reason one of the channels in that backup already exists, the whole recovery is aborted.
This leaves the restore in a half-finished state and a re-try or resume is not possible.
With this small fix, the same multi-file can be added again and all channels that were not imported before can now be added.

### DIFF
--- a/chanbackup/recover.go
+++ b/chanbackup/recover.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/davecgh/go-spew/spew"
+	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/keychain"
 )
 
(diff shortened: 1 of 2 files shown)


guggero:
@Roasbeef bump as potential candidate for 0.8.2.

chanbackup/recover.go
```diff
@@ -47,6 +48,14 @@ func Recover(backups []Single, restorer ChannelRestorer,
 			backup.FundingOutpoint)
 
 		err := restorer.RestoreChansFromSingles(backup)
+
+		// If a channel is already present in the channel DB, we can
+		// just continue. No reason to fail a whole set of multi backups
+		// for example. This allows resume of a restore in case another
+		// error happens.
+		if err == channeldb.ErrChanAlreadyExists {
```
cfromknecht: do we know that this is the only error that we should silence?

Roasbeef: Yes, restoring channel shells happens in a single DB transaction. If it succeeds, then the second time around, it'll fail when trying to create the channel point bucket: https://github.com/lightningnetwork/lnd/blob/master/channeldb/channel.go#L711.

cfromknecht: nice, that's what i was looking for

guggero:
Added a test on the itest level to make sure an SCB file can be imported twice without running into an error.
Unit test would be quite a bigger diff since the `*channeldb.DB` and `*contractcourt.ChainArbitrator` of `chanDBRestorer` would need to be abstracted behind an interface to be able to mock them.

PR #3738, opened 2019-11-18
Chanfitness: Revenue Calculation

carlaKC:
Add a `GetRevenue` Query to channel fitness event store and an `ChannelInsights` rpc endpoint.


carlaKC:
Closing in favor of decoupling from event store.

PR #3739, opened 2019-11-18
routing: last hop restriction

joostjager:
This PR adds an optional last hop restriction to payments.


routing/pathfind_test.go
```diff
@@ -2217,3 +2116,71 @@ func testProbabilityRouting(t *testing.T, p10, p11, p20, minProbability float64,
...
+		source:            route.Vertex(sourceNode.PubKeyBytes),
+	}
+
+	ctx.pathFindingConfig = *testPathFindingConfig
+
+	ctx.graphParams.graph = testGraphInstance.graph
+
+	ctx.restrictParams.FeeLimit = noFeeLimit
+	ctx.restrictParams.ProbabilitySource = noProbabilitySource
+	ctx.restrictParams.CltvLimit = math.MaxUint32
```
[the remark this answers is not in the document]
joostjager: Good point, fixed

joostjager:
> Very nice and clean PR! This will allow us to craft rebalance payments without external tools, right?

In combination with #3736 it does.

Also we can use this to loop into a specific channel.

routing/router.go
```diff
@@ -1596,6 +1596,10 @@ type LightningPayment struct {
 	// hop. If nil, any channel may be used.
 	OutgoingChannelID *uint64
 
+	// LastHop is the pubkey of the last node before the final destination
+	// is reached. If nil, any node may be used.
+	LastHop *route.Vertex
```
wpaulino: Do we care about validating whether this is an existing node in the graph? Path finding would result in `unable to find a path to destination`, which can be ambiguous.

routing/pathfind_test.go
```diff
@@ -1874,6 +1874,51 @@ func TestRestrictOutgoingChannel(t *testing.T) {
 	}
 }
 
+// TestRestrictLastHop asserts that a last hop restriction is obeyed by the path
+// finding algorithm.
+func TestRestrictLastHop(t *testing.T) {
+	t.Parallel()
+
+	// Set up a test graph with three possible paths from roasbeef to
+	// target. The path through channel 2 is the highest cost path.
```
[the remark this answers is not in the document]
joostjager: Forgot to update comments. Fixed

routing/router.go
joostjager: We also don't do this for the outgoing channel restriction. As this is advanced functionality, I am ok with leaving it up to the caller to make sure this doesn't happen. What do you think?

Maybe if we start fixing https://github.com/lightningnetwork/lnd/issues/1680, we can add two new payment failures:
`no_outgoing_channel` (with enough balance and not filtered out) and `no_last_hop` (with enough capacity and not filtered out).

wpaulino: We can defer it until then.

Issue #3645, opened 2019-10-29
Track Channel Profit


carlaKC:
Routing nodes that need to allocate scarce capital will want to know how much profit each of their channel is making so that they can efficiently manage capital, have visibility on the amount their capital is earning and potentially close channels that are not profitable. 

Dependent on #3420, this can be achieved by tracking the HTLCs that are forwarded through a channel. HTLC forwards arrive on an incoming channel and depart on an outgoing channel, so profit from a forward must be attributed to both channels to prevent underestimating the influence of mostly-inbound channels. 

Two related measures that operators may be interested in:
1. Utility: The portion of funds moving along a channel which are for sends from the node or receives to it. This gives a measure of how much personal use the node gets from a channel, despite those HTLCs not earning fees.
2. Failed forwards: The portion of HTLCs which are forwarded though the channel and then later failed. This is not an actionable metric, but provides interesting insight into how often capital is being locked up (and how long it is locked up for) without earning a fee.

halseth:
To get historical data, can we use the open channel's revocation log to extract the information about which HTLCs got settled/failed and the profitability? The information should be there and would allow us to extract this information for all existing channels

carlaKC:
I've been having a look at this today and I think that querying the forwarding log, as in `rpcserver.go#ForwardingHistory`, would be a good place to start? Since we want to attribute profit across channels, it would be nice to have the information by channel pair? Our receives won't be included in this query, but that information is available elsewhere. 

Since I'd expect we'll be doing these calculations for channels that are open for a long time, I'm wondering whether to query periodically to keep running track of profit in memory (we can discard the events once processed and query DB just with an offset from then) or to just process on demand. 

**Rough Calcs**
Transaction volumes per day (as per LNBig's twitter reports):
Regular: 200 
Upper bound: 3500
and for completeness, a low tx count of 5 a day

Size of DB query for regular/upper tx volumes by length of query period, given a forwarding event size of [40 bytes](https://github.com/lightningnetwork/lnd/blob/master/rpcserver.go#L4728):

1 month:
- Low; 05 * 31 * 40 bytes =.0062 MB
- Regular: 200 * 31 * 40 bytes = 0.248 MB
- Upper: 3500 * 31 * 40 bytes = 4.34 MB

6 months:
- Low: 5 * 182 * 40 bytes = 0.0364
- Regular: 200 * 182 * 40 bytes = 1.456 MB
[...]

PR #3741, opened 2019-11-18
multi: Channel Revenue

carlaKC:
This PR adds a channel revenue calculation which attributes channel fees evenly across incoming and outgoing channels.  A `ChannelRevenue` endpint to query this value  is added to `lnrpc` and `lncli`.

Revenue is calculated by querying the forwarder log and adding fees for all the events where our target channel is the incoming or outgoing channel. The forwarder log cannot be queried on a per channel basis, only by time period, so effort is made to minimize the size of this query by bounding queries to the channel's open/close time. 

Part of: #3645


lnrpc/rpc.proto
```diff
@@ -793,6 +793,21 @@ service Lightning {
...
+    present, or channel close time if the channel is closed. Requests with
+    start or end time set will also be restricted to channel open/close time
+    because revenue cannot be generated before open or after close.
+*/
+    rpc ChannelRevenue(ChannelRevenueRequest) returns (ChannelRevenueResponse) {
+        option (google.api.http) = {
+            post: "/v1/revenue"
+            body: "*"
+        };
+    };
```
joostjager: Shouldn't this be integrated with the uptime query, make a start with a framework for querying channel fitness? Or at least mark these methods as experimental so they can be changed at any time without going through the deprecation process.

lnrpc/rpc.proto
```diff
@@ -2768,3 +2783,19 @@ message BakeMacaroonResponse {
     /// The hex encoded macaroon, serialized in binary format.
     string macaroon = 1 [json_name = "macaroon"];
 }
+
+message ChannelRevenueRequest{
+    ///  The outpoint (txid:index) of the funding transaction.
+    ChannelPoint channel_point = 1;
```
joostjager: I remember we wanted to move away from this `ChannelPoint` message in favor of the more generic `Outpoint`

lnrpc/rpc.proto
```diff
@@ -2768,3 +2783,19 @@ message BakeMacaroonResponse {
...
+
+    /// Start time is the beginning of the range that revenue should be calcualted over.
+    uint64 start_time = 2 [json_name = "start_time"];
```
joostjager: Add something to the field name to clarify the units. Is it unix time?

rpcserver.go
```diff
@@ -5397,3 +5404,149 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+	// Obtain paginated forwarder events by querying the forwarder log in the
+	// period provided.
+	query := func(offset, maxEvents uint32) (channeldb.ForwardingLogTimeSlice, error) {
+		return r.server.chanDB.ForwardingLog().Query(
+			channeldb.ForwardingEventQuery{
+				StartTime:    startTime,
+				EndTime:      endTime,
+				IndexOffset:  offset,
+				NumMaxEvents: maxEvents,
+			})
```
joostjager: If you query your oldest channel and if that channel is still open, it seems that you still retrieve the complete forwarding log. I can see that for some channels the start/end range may speed up the query, but is that enough to justify the extra complexity? And is it good enough then?

I think we should first address the lower layer, the forwarding log. Make that ready for accounting and other stats like this by for example adding another index. Otherwise we are spending time on building, reviewing and maintaining code that we already know is not ideal.

lnrpc/rpc.proto
```diff
@@ -2768,3 +2783,19 @@ message BakeMacaroonResponse {
...
+
+    /// End time is the end of the range that revenue should be calcualted over.
+    uint64 end_time = 3 [json_name = "end_time"];
+}
+
+message ChannelRevenueResponse{
+    /// The revenue of the channel in msat.
+    int64 revenue_msat = 1 [json_name = "revenue_msat"];
+}
```
joostjager: Question about the use case: isn't it more valuable to see the revenue per pair of channels?

lnwallet/interface.go
```diff
@@ -281,6 +281,9 @@ type BlockChainIO interface {
 	// GetBlock returns the block in the main chain identified by the given
 	// hash.
 	GetBlock(blockHash *chainhash.Hash) (*wire.MsgBlock, error)
+
+	// GetBlockHeader returns the header for the block with the given hash.
+	GetBlockHeader(blockHash *chainhash.Hash) (*wire.BlockHeader, error)
```
Roasbeef: After the latest round of offline discussions, I don't think we'll need this anymore? Since it's possible due to block time skew we start too early and miss a set of events for a channel.

chanfitness/revenue.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// attributeIncoming is the percentage of fees that are attributed to the
+// incoming channel in a forwarding event. The remaining 1- attributeIncoming
+// percent is attributed to the outgoing channel.
+var attributeIncoming = 0.5
```
Roasbeef: One lower level detail that's relevant here is that the fee policy if the _outgoing_ channel is the one that's applied during forwarding/settlement. We accept the incoming HTLC, but then our bandwidth is allocated on the outgoing HTLC, which is where we actually charge our fees: we send out less than we send in. However, the _incoming_ channel is actually where the funds are accrued since there's where the difference between the incoming+outgoing HTLC is stored.

chanfitness/revenue.go
```diff
@@ -0,0 +1,52 @@
...
+// calculateRevenue calculates the revenue for a channel for the set of events
+// provided. It distributes fees between incoming and outgoing channels
+// involved in htlc forwards according the the value set in attributeIncoming.
+func calculateRevenue(chanID lnwire.ShortChannelID,
+	events []channeldb.ForwardingEvent) lnwire.MilliSatoshi {
+
+	// Revenue is the total revenue earned by this channel in millisatoshis.
+	// This value is accumulated as a float then converted to millisatoshis so
+	// that we can minimize the amount of rounding from float -> int.
+	var revenue float64
```
Roasbeef: Alternatively, we can used fixed point arithmetic (working in `mSAT`) all the way throughout.

chanfitness/revenue.go
```diff
@@ -0,0 +1,52 @@
...
+
+			revenue += earned
+		}
+
+		// If our target channel was the outgoing channel for this event,
+		// attribute (1 - percentage for incoming) percent of the fees to it.
+		if event.OutgoingChanID == chanID {
+			earned := float64(fee) * (1 - attributeIncoming)
+
+			log.Debugf("Channel: %v is outgoing channel, incrementing "+
```
Roasbeef: Perhaps trace instead? Otherwise this could be rather spammy for a popular or long lived node.

chanfitness/revenue.go
Roasbeef: One other direction that may be worth exploring is exactly how we surface this information on the RPC interface. Rather than showing a single value (total revenue) for a channel, we could instead unpack the revenue into channel pairs. This would allow the end caller to reason about the relationship between the target channel and the set of open channels, essentially: typically where is the incoming HTLC being sourced from?

chanfitness/revenue.go
```diff
@@ -21,6 +21,21 @@ var (
 type eventsQuery func(
 	offset, maxEvents uint32) (channeldb.ForwardingLogTimeSlice, error)
 
+// GetRevenue calculates accumulates relevant forwarding logs and calcualtes
+// revenue for a target channel.
+func GetRevenue(channelID lnwire.ShortChannelID,
```
Roasbeef: `RevenueReport`? 😉

Roasbeef: Typically we like to accept channel points where possible as they're more immediately queryable (ask your node for example). The forwarding log is based on channel ID, but there're utility methods to obtain the channel ID that maps to a channel point. There're several possibilities, but in practice we can ignore most of them.

Roasbeef: Actually see this method: https://godoc.org/github.com/lightningnetwork/lnd/lnwire#ChannelID.IsChanPoint

lnrpc/rpc.proto
Roasbeef: I think maybe the OP made a new call rather than stuffing it into `ListChannels` as computing this (as is) is more expensive than obtaining the current uptime values.

lnrpc/rpc.proto
Roasbeef: Methinks the same, see this comment (copy-pasting here since still in draft): 
> One other direction that may be worth exploring is exactly how we surface this information on the RPC interface. Rather than showing a single value (total revenue) for a channel, we could instead unpack the revenue into channel pairs. This would allow the end caller to reason about the relationship between the target channel and the set of open channels, essentially: typically where is the incoming HTLC being sourced from?

rpcserver.go
```diff
@@ -5397,3 +5404,149 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
 	return resp, nil
 }
+
+// ChannelRevenue returns the revenue for a channel over a period of time.
+// To limit the size of query required to calculate revenue, channel open
+// an close times are looked up by block height, and the user provided period
+// is bounded by these values; no revenue generating events can occur before
+// channel open or after channel close.
+func (r *rpcServer) ChannelRevenue(ctx context.Context,
+	req *lnrpc.ChannelRevenueRequest) (*lnrpc.ChannelRevenueResponse, error) {
```
Roasbeef: Alternatively, we could expand the scope of this method into more of a "channel analysis" endpoint, so this would as is, also encompass uptime information (which can be used to compute the max effective utilization of a channel). Just a drive by thought.

rpcserver.go
```diff
@@ -5397,3 +5404,149 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+		shortChanID   lnwire.ShortChannelID
+	)
+
+	// Attempt to lookup the channel, first checking open channels and then
+	// closed if it is not found. We will save the short channel ID (containing
+	// open height) and close height (if the channel is closed) so that we can
+	// obtain channel open/close times from the block height. These times are
+	// needed to reduce to size of the forwarder log query that get revenue
+	// requires.
+
```
Roasbeef: Style nit: extra new line.

rpcserver.go
```diff
@@ -5397,3 +5404,149 @@ func (r *rpcServer) BakeMacaroon(ctx context.Context,
...
+	startTime := time.Unix(int64(req.StartTime), 0)
+
+	// Set end time from the user specified time. If no value was provided,
+	// end time will be set to the present.
+	endTime := time.Unix(int64(req.EndTime), 0)
+	if req.EndTime == 0 {
+		endTime = time.Now()
+	}
+
+	// Get open time for the channel from block height.
```
Roasbeef: Many items below this block can likely now be simplified after the latest set of offline discussions.

cmd/lncli/commands.go
```diff
@@ -1289,14 +1291,15 @@ func parseChannelPoint(ctx *cli.Context) (*lnrpc.ChannelPoint, error) {
 	case args.Present():
 		index, err := strconv.ParseUint(args.First(), 10, 32)
 		if err != nil {
-			return nil, fmt.Errorf("unable to decode output index: %v", err)
+			return nil, nil, fmt.Errorf("unable to decode output index: %v", err)
 		}
 		channelPoint.OutputIndex = uint32(index)
+		args = args.Tail()
```
Roasbeef: Have all the callers been updated to compensate for this API change? Not immediately clear why this change is required.

lnwallet/interface.go
carlaKC: Yep, will remove and go with a basic first iteration as discussed.

lnrpc/rpc.proto
carlaKC: In the comment or var name? `start_time_seconds`/ `start_time_unix` would be most appropriate in the name, but we don't seem to have that elsewhere in the rpc for times. Otherwise I'll just update the comment.

lnrpc/rpc.proto
carlaKC: > I think maybe the OP made a new call rather than stuffing it into `ListChannels` as computing this (as is) is more expensive than obtaining the current uptime values.

This was the original thought, not to overload listchannels with this call. 
Depending on how we go with exposing revenue info, I'll either add a call which will return uptime and revenue over a start-end period for a specific chan, or go with a `RevenueRepot` which does some pairwise calcs over a range (and leave uptime in `listchannels`).

rpcserver.go
carlaKC: I think that these two can proceed in parallel? Reliable accounting brainstorm is on the go, and the logic of this call is unlikely to change significantly - perhaps the query changes a bit if things are better indexed, but I don't forsee it being much of an issue.

cmd/lncli/commands.go
carlaKC: So our pattern here is to consume an arg then pop it off of args so that we can process what's left. In `parseChannelPoint` we were not popping the output index off of the list of args so code after `parseChanelPoint` would re-process the output index (in my case, try to read it as the start time). 

This wasn't a problem for the other places where it's called (`abandonChannel`/`closeChannel`) because they do not take any further args. They've been updated, but it's not a significant change. 

On second look, this would prob be nicer as `parseChannel(ctx cli.Context, args *cli.Args)` to avoid an extra return item. I was against passing `ctx` and `args` because cli has args but I think that way is better.

lnrpc/rpc.proto
carlaKC: Worth adding a comment on the `ChannelPoint` message to use `Outpoint` instead? Even if we're not deprecating it, maybe just a note to use `Outpoint` where possible?

lnrpc/rpc.proto
carlaKC: Something of this sort appeals to me?
Request channel report for channel A, from `start_time` -> `end_time`
```
ChannelReport{
  Uptime
  Lifespan
  Revenue{
     B: x msat,
     C: y msat, 
  }
}
```

lnrpc/rpc.proto
joostjager: That is true indeed. Recently we got `time_ns`, but that still doesn't express that it is unix time. Ok, comment then indeed.

lnrpc/rpc.proto
joostjager: Sounds good

chanfitness/revenue.go
```diff
@@ -0,0 +1,227 @@
...
+		// our events list.
+		for _, fwd := range forwarding.ForwardingEvents {
+			incoming, ok := channelIDs[fwd.IncomingChanID]
+			if !ok {
+				return nil, errUnknownChannelID
+			}
+
+			outgoing, ok := channelIDs[fwd.OutgoingChanID]
+			if !ok {
+				return nil, errUnknownChannelID
```
halseth: seems like this is likely to happen? If there are events for channels we are currently not querying for.

lnrpc/rpc.proto
```diff
@@ -2797,3 +2810,91 @@ message BakeMacaroonResponse {
...
+    */
+    repeated ChannelReport reports = 1 [json_name = "channel_report"];
+}
+
+message ChannelReport{
+    /*
+    Output is the funding outpoint of the channel this report has been generated
+    for.
+    */
+    OutPoint outpoint = 1 [json_name = "channel_id"];
```
halseth: json name `chan(nel)_point` perhaps

lnrpc/rpc.proto
halseth: Personally I like seeing the uptime in `listchannels`

chanfitness/revenue.go
```diff
@@ -0,0 +1,147 @@
...
+
+// setRevenue sets the revenue value for a target channel and its forwarding
+// pair. This function expects both maps to be initialized.
+func (r RevenueReport) setRevenue(targetChan, pairChan wire.OutPoint,
+	revenue Revenue) {
+
+	r.ChannelPairs[targetChan][pairChan] = revenue
+}
+
+// addUpstream gets the existing revenue record for the incoming channel with
```
halseth: name in comment not updated

chanfitness/revenue.go
```diff
@@ -0,0 +1,147 @@
...
+// channel outpoints rather than short channel ids.
+type revenueEvent struct {
+	incomingChannel wire.OutPoint
+	outgoingChannel wire.OutPoint
+	incomingAmt     lnwire.MilliSatoshi
+	outgoingAmt     lnwire.MilliSatoshi
+}
+
+// revenueReport creates a revenue report for the set of events provided. It takes
+// an attribute incoming float which determines the fee spit between incoming and
```
halseth: s/spit/split

chanfitness/revenue.go
```diff
@@ -1,10 +1,77 @@
...
+
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/channeldb"
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
+var (
+	// maxQueryEvents is the number of events to query the forward log for at
+	// a time.
+	maxQueryEvents uint32 = 500
```
halseth: can make it const

chanfitness/revenue.go
```diff
@@ -1,10 +1,77 @@
...
+
+	// errUnknownChannelID is returned when we cannot map a short channel ID to
+	// a channel outpoint.
+	errUnknownChannelID = errors.New("cannot find channel outpoint")
+)
+
+// eventsQuery is a function which returns paginated queries for forwarding
+// events.
+type eventsQuery func(
```
halseth: alternative method signature could be 

```go
query func(func(event ForwardingEvent) error) error
```
such that each event would be provided in the callback. Offset tracking would still need to be done on a higher level though.

chanfitness/revenue.go
```diff
@@ -0,0 +1,227 @@
...
+	record := &RevenueReport{
+		ChannelPairs: make(map[wire.OutPoint]map[wire.OutPoint]Revenue),
+	}
+
+	for _, event := range events {
+		// Calculate total fees earned for this event.
+		fee := event.incomingAmt - event.outgoingAmt
+
+		// Calculate fees earned by the incoming channel in this event.
+		incomingFees := lnwire.MilliSatoshi(float64(fee) * attributeIncoming)
```
halseth: I'm wondering if we need `attributeIncoming` at all. When querying a specific channel, you get a pairwise breakdown of the total fees this channel was involved in.  There will be an overlap, but if I just want to get the total fees earned I'll get the report for all my channels and divide by 2.

chanfitness/revenue.go
carlaKC: The map should have all of our channels in it. It's not very efficient, but we don't have a way of determining which channels we have open over a timestamped period, so I thought it would be ok like this?

chanfitness/revenue.go
carlaKC: I think it's more intuitive to have it already divided by 2? It doesn't make sense to double report by default imo.

lnrpc/rpc.proto
carlaKC: Rename both to `channel_outpoint`? Seems weird to have them be different.

lnrpc/rpc.proto
joostjager: I am ok with keeping it then. It is marked experimental anyway.

rpcserver.go
joostjager: Now that the optimization is removed, this comment does no longer apply.

chanfitness/revenue.go
joostjager: I am not sure about the `/2` either. It is kind of implicit that that is happening. The factor is even configurable, but would that be something that users need?

carlaKC:
Closing this in favor of moving logic out of lnd for now.

PR #3742, opened 2019-11-19
invoices: expose custom tlv records from the payload

joostjager:
This PR extracts the received custom records from the hop payload and stores them in the invoice database. Users can retrieve the custom records through the usual invoice queries / subscriptions.


tlv/record.go
```diff
@@ -12,9 +12,10 @@ import (
 // Type is an 64-bit identifier for a TLV Record.
 type Type uint64
 
-// TypeSet is an unordered set of Types. The map item boolean values indicate
-// whether the type that we parsed was known.
-type TypeSet map[Type]bool
+// TypeSet is an unordered set of Types. The map values are byte slices. If the
+// byte slice is nil, the type was successfully parsed. Otherwise the value is
+// byte slice containing the encoded data.
+type TypeSet map[Type][]byte
```
Roasbeef: This seems to be really overloading the type beyond its original intended usage. It's only meant to be used in order to determine violations w.r.t known/unknown types. We already return the full set of unparsed types after decoding, so not sure why this is needed.

Roasbeef: Also have you benched marked this? As is, it appears we'll end up copying all the values from the main stream _twice_.

htlcswitch/hop/payload.go
```diff
@@ -157,14 +164,22 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
 		mpp = nil
 	}
 
+	for t, parseResult := range parsedTypes {
```
Roasbeef: Feels like there's another way to do this with less copying.

tlv/record.go
joostjager: Yes it is overloading the type, but we can deviate from an original intention. 

I haven't benchmarked. If we want to process the raw bytes of unrecognized records, we need to copy anyway. Just in case our intention is to really ignore them, the copy is unnecessary. Afaik, for db tlv that isn't expected. And for the tlv payload it may happen with an unrecognized standard field (the custom ones we probably want to store somewhere). To me it seems premature to optimize for.

I am open to different ways of implementing this. What do you prefer?

htlcswitch/hop/payload.go
joostjager: In line with comment above, I don't think it matters much. If you want the tlv stream to return a map that we can use directly, we need to add more controls to the stream to specify what exactly we expect.

channeldb/invoices.go
```diff
@@ -982,6 +988,15 @@ func serializeHtlcs(w io.Writer, htlcs map[CircuitKey]*InvoiceHTLC) error {
...
 		)
+
+		customRecords, err := tlv.MapToRecords(htlc.CustomRecords)
+		if err != nil {
+			return err
+		}
+
+		records = append(records, customRecords...)
+
+		tlvStream, err := tlv.NewStream(records...)
```
joostjager: Question open as to whether we want to bump the db version, to prevent users running master from downgrading and loosing their custom records

htlcswitch/hop/payload.go
joostjager: suggestion from @cfromknecht: delete items from the map

channeldb/invoices.go
```diff
@@ -982,6 +988,15 @@ func serializeHtlcs(w io.Writer, htlcs map[CircuitKey]*InvoiceHTLC) error {
 			tlv.MakePrimitiveRecord(expiryHeightType, &htlc.Expiry),
 			tlv.MakePrimitiveRecord(stateType, &state),
 		)
+
+		customRecords, err := tlv.MapToRecords(htlc.CustomRecords)
+		if err != nil {
+			return err
+		}
+
+		records = append(records, customRecords...)
```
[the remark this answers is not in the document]
joostjager: Discussed this further. It may be better to create a single tlv record for which the value is a tlv stream with just the custom records. As a simple insurance against accidental modification of standard fields.

htlcswitch/hop/payload.go
joostjager: It seems it isn't possible to cast `map[Type][]byte` to `map[uint64][]byte` though

tlv/record.go
joostjager: Where do you see the copy twice btw? Afaik we only read bytes from the stream and store them in a new byte slice that is a value of the `TypeSet` map.

joostjager: Renamed to `TypeMap` and preallocated the buffer.

htlcswitch/hop/payload.go
joostjager: Talked about this offline with @roasbeef. We do want to stop `tlv.Type` from being used anywhere. Copying the map (but not the values) looks to be ok.

channeldb/invoices.go
joostjager: Discussed this in the context of another PR. We don't give this guarantee to users running master generally.

channeldb/invoices.go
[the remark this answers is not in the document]
joostjager: Discussed offline. There seems to be more support for a flat structure, so leaving it as is.

invoices/interface.go
```diff
@@ -1,11 +1,16 @@
...
+)
 
 // Payload abstracts access to any additional fields provided in the final hop's
 // TLV onion payload.
 type Payload interface {
 	// MultiPath returns the record corresponding the option_mpp parsed from
 	// the onion payload.
 	MultiPath() *record.MPP
+
+	CustomRecords() hop.CustomRecordSet
```
cfromknecht: needs godoc

channeldb/invoices.go
```diff
@@ -1290,12 +1318,20 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
 		if _, exists := invoice.Htlcs[key]; exists {
 			return nil, fmt.Errorf("duplicate add of htlc %v", key)
 		}
+
+		// Force caller to supply htlc without custom records in a
+		// consistent way.
+		if htlcUpdate.CustomRecords == nil {
```
cfromknecht: i guess this conflicts with what i said earlier about constructor returning nil, i'm fine if you prefer it to always be initialized.

channeldb/invoices.go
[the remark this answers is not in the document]
cfromknecht: I am still somewhat concerned about mingling minimally validated user data with known records. I think it is possible to do it safely, but i think it forces us to be stricter in how we handle migrations.

Consider the scenario where we add a field to the invoice body but do not bump the db version. If a user downgrades, this will now appear in the custom record set even tho its type is beneath the cutoff. I think it's possible that data gets written back, but it's difficult imo to ascertain whether this can lead to unforeseen bugs.

We could prevent this by filter the custom records when reading as we do in the hop, then the data is actually dropped when writing back. This would maintain the current behavior in that scenario.

tlv/stream.go
```diff
@@ -230,10 +231,25 @@ func (s *Stream) decode(r io.Reader, parsedTypes TypeSet) (TypeSet, error) {
...
 		// Otherwise, the record type is unknown and is odd, discard the
 		// number of bytes specified by length.
 		default:
-			_, err := io.CopyN(ioutil.Discard, r, int64(length))
+			// If the caller provided an initialized TypeMap, record
+			// the encoded bytes.
+			var b *bytes.Buffer
+			writer := ioutil.Discard
+			if parsedTypes != nil {
+				b = bytes.NewBuffer(make([]byte, 0, length))
```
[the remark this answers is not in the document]
joostjager: I don't get how we are double allocating now. We allocate a slice of the right length, wrap a buffer around that slice and write the bytes into it?

channeldb/invoices.go
joostjager: Yes, I wanted to prevent to case where a future dev would do a `nil` check and forget about the case where the map is empty.

htlcswitch/hop/payload.go
```diff
@@ -157,14 +166,22 @@ func NewPayloadFromReader(r io.Reader) (*Payload, error) {
...
```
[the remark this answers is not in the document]
joostjager: Good suggestion, func added

channeldb/invoices.go
```diff
@@ -1200,6 +1221,13 @@ func deserializeHtlcs(r io.Reader) (map[CircuitKey]*InvoiceHTLC, error) {
 		htlc.ResolveTime = time.Unix(0, int64(resolveTime))
 		htlc.State = HtlcState(state)
 		htlc.Amt = lnwire.MilliSatoshi(amt)
+		htlc.CustomRecords = make(hop.CustomRecordSet)
+
+		for t, parseResult := range parsedTypes {
```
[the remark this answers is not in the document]
joostjager: Added

htlcswitch/hop/payload_test.go
```diff
@@ -237,4 +241,14 @@ func testDecodeHopPayloadValidation(t *testing.T, test decodePayloadTest) {
 	} else if p.MPP != nil {
 		t.Fatalf("unexpected MPP payload")
 	}
+
+	// Convert expected nil map to empty map, because we always expect an
```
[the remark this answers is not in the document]
joostjager: discussion in comment below

invoices/interface.go
joostjager: added

joostjager:
>We could prevent this by filter the custom records when reading as we do in the hop, then the data is actually dropped when writing back. This would maintain the current behavior in that scenario.

Yes that is a nice safe guard that doesn't require a different db struct. It should be in place now

tlv/stream.go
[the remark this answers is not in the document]
cfromknecht: oh wait sorry, missed that the definition was a pointer and not just a zero-value `bytes.Buffer`

tlv/stream.go
```diff
@@ -230,10 +231,25 @@ func (s *Stream) decode(r io.Reader, parsedTypes TypeSet) (TypeSet, error) {
 				return nil, err
 			}
 
+			// Record the successfully decoded type if the caller
+			// provided an initialized TypeMap.
+			if parsedTypes != nil {
+				parsedTypes[typ] = nil
```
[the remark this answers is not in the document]
joostjager: In that case we would copy the data twice for the known types. And need another data structure to signal which ones were decoded. We can also rethink that when the need arises to access the raw bytes of known types, this is all in memory.

cfromknecht:
needs rebase

joostjager:
rebased

PR #3743, opened 2019-11-20
lnwire: assert sorted short channel ids

cfromknecht:
This PR modifies our decoding of short channel id arrays to also assert that the decoded ids are in order for the plain encodings. We already do so for the zlib encoding, though this isn't tested directly. Here we correct the decoding make this assertion for plain encodings, and add unit tests to both `QueryShortChanIDs` and `ReplyChannelRange` to ensure that they fail with this error when presented with duplicate or unsorted channel ids for either plain or zlib.

### DIFF
--- a/lnwire/query_short_chan_ids.go
+++ b/lnwire/query_short_chan_ids.go
@@ -35,6 +35,19 @@ const (
 	maxZlibBufSize = 67413630
 )
 
+// ErrUnsortedSIDs is returned when decoding a QueryShortChannelID request whose
+// items were not sorted.
+type ErrUnsortedSIDs struct {
+	prevSID ShortChannelID
+	curSID  ShortChannelID
+}
+
+// Error returns a human-readable description of the error.
+func (e ErrUnsortedSIDs) Error() string {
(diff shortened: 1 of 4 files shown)


lnwire/query_short_chan_ids.go
```diff
@@ -67,6 +80,12 @@ type QueryShortChanIDs struct {
 
 	// ShortChanIDs is a slice of decoded short channel ID's.
 	ShortChanIDs []ShortChannelID
+
+	// noSort indicates whether or not to sort the short channel ids before
+	// writing them out.
+	//
+	// NOTE: This should only be used during testing.
+	noSort bool
```
wpaulino: Any reason to add this? They should always be sorted so this is only used for testing unused behavior.

cfromknecht: we always sort, but this tests that we behave as expected if non-lnd peers don't sort. mainly this was the simplest way of passing the bool internally without modifying `Encode` or `Decode`, or having to duplicate the encoding logic w/o sorting just for testing

PR #3744, opened 2019-11-20
lnrpc: re-enable custom records

joostjager:
This PR re-enables custom record sending that was disabled in #3575.

It leaves the database serialization format as is, because there currently is no functional need to change it.


joostjager:
@cfromknecht, we spoke offline about the risk that we wouldn't be able to deserialize even custom records from our own database as long as odd/even checking was part of the `tlv` package. That has been removed in #3701, but looking at the hop serialization code again it also seems that we weren't using tlv there in the first place?

joostjager:
Added the custom range check. Hopefully this PR didn't miss any entry/exit point for custom records. Because of the duplicated send interface, it is quite a few places.

lnrpc/routerrpc/router_backend.go
```diff
@@ -504,9 +536,11 @@ func (r *RouterBackend) extractIntentFromSendRequest(
...
 	}
 
-	var destTLV map[uint64][]byte
-	if len(destTLV) != 0 {
-		payIntent.FinalDestRecords = tlv.MapToRecords(destTLV)
+	payIntent.FinalDestRecords, err = UnmarshallCustomRecords(
+		rpcPayReq.DestTlv,
+	)
+	if err != nil {
+		return payIntent, err
```
guggero: nit: returned value is ignored anyway, could also just be `nil, err`.

lnrpc/rpc.proto
```diff
@@ -1854,6 +1854,14 @@ message QueryRoutesRequest {
...
     */
     uint32 cltv_limit = 11;
+
+    /** 
+    An optional field that can be used to pass an arbitrary set of TLV records
+    to a peer which understands the new records. This can be used to pass
+    application specific data during the payment attempt. If the destination
+    does not support the specified recrods, and error will be returned.
+    */
+    map<uint64, bytes> dest_tlv = 13;
```
guggero: On every other field that has the type `bytes` involved, we have the following comment:
> When using REST, this field must be encoded as base64.

Do we want to add this here too?

lnrpc/routerrpc/router_backend.go
```diff
@@ -364,90 +364,66 @@ func (r *RouterBackend) MarshallRoute(route *route.Route) (*lnrpc.Route, error)
...
+			return nil, fmt.Errorf("cannot decode pubkey %s",
+				rpcHop.PubKey)
+		}
+		copy(pubKeyBytes[:], pubKey)
+	} else {
 		// If no pub key is given of the hop, the local channel
 		// graph needs to be queried to complete the information
 		// necessary for routing.
-		return r.UnmarshallHopByChannelLookup(hop, prevNodePubKey)
+
```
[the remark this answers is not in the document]
joostjager: merged

lnrpc/routerrpc/router_backend.go
joostjager: fixed

lnrpc/rpc.proto
joostjager: Done

lnrpc/routerrpc/router_backend.go
```diff
@@ -358,99 +363,97 @@ func (r *RouterBackend) MarshallRoute(route *route.Route) (*lnrpc.Route, error)
...
 
-	var tlvRecords []tlv.Record
+	tlvRecords := tlv.MapToRecords(rpcRecords)
 
-	mpp, err := UnmarshalMPP(hop.MppRecord)
-	if err != nil {
-		return nil, err
+	// tlvRecords is sorted, so we only need to check that the first
+	// element is within the custom range.
+	if uint64(tlvRecords[0].Type()) < hop.CustomTypeStart {
```
cfromknecht: do we know length > 0?

joostjager: Maybe a zero element map from grpc is always nil, but not fully sure. Replaced the nil check at the top by a len()==0 check

lnrpc/rpc.proto
cfromknecht: To be consistent with the hop message, I would say this should be renamed to something like `dest_custom_records`. Comment should also mention custom record range

PR #3745, opened 2019-11-20
lncli addholdinvoice: allow specifying msat with --amt_msat

akovalenko:
Make `lncli addholdinvoice` send value in millisatoshi (ValueMsat introduced by commit fcbf1bc07a032f1f95ecf64ba9a216dd229dfb59). 

New flag `--amt_msat` is introduced for this purpose, while `--amt` and unnamed command-line argument are still supposed to be in satoshi for backward compatibility.

### DIFF
--- a/cmd/lncli/invoicesrpc_active.go
+++ b/cmd/lncli/invoicesrpc_active.go
@@ -160,6 +160,10 @@ var addHoldInvoiceCommand = cli.Command{
 			Name:  "amt",
 			Usage: "the amt of satoshis in this invoice",
 		},
+		cli.Int64Flag{
+			Name:  "amt_msat",
+			Usage: "the amt of millisatoshis in this invoice",
+		},
 		cli.StringFlag{
 			Name: "description_hash",
 			Usage: "SHA-256 hash of the description of the payment. " +
@@ -192,7 +196,6 @@ var addHoldInvoiceCommand = cli.Command{
 func addHoldInvoice(ctx *cli.Context) error {
 	var (
 		descHash []byte
-		amt      int64
 		err      error
 	)
 
@@ -212,12 +215,11 @@ func addHoldInvoice(ctx *cli.Context) error {
 
 	args = args.Tail()
 
-	switch {
-	case ctx.IsSet("amt"):
-		amt = ctx.Int64("amt")
-	case args.Present():
-		amt, err = strconv.ParseInt(args.First(), 10, 64)
+	amt := ctx.Int64("amt")
+	amtMsat := ctx.Int64("amt_msat")
 
+	if !ctx.IsSet("amt") && !ctx.IsSet("amt_msat") && args.Present() {
+		amt, err = strconv.ParseInt(args.First(), 10, 64)
 		if err != nil {
 			return fmt.Errorf("unable to decode amt argument: %v", err)
 		}
@@ -236,6 +238,7 @@ func addHoldInvoice(ctx *cli.Context) error {
 		Memo:            ctx.String("memo"),
 		Hash:            hash,
 		Value:           amt,
+		ValueMsat:       amtMsat,
 		DescriptionHash: descHash,
 		FallbackAddr:    ctx.String("fallback_addr"),
 		Expiry:          ctx.Int64("expiry"),


cmd/lncli/invoicesrpc_active.go
```diff
@@ -214,9 +218,18 @@ func addHoldInvoice(ctx *cli.Context) error {
...
 	switch {
 	case ctx.IsSet("amt"):
-		amt = ctx.Int64("amt")
+		if ctx.IsSet("amt_msat") {
+			return fmt.Errorf("amt and amt_msat are mutually exclusive")
+		}
+		ms = ctx.Int64("amt") * 1000
+	case ctx.IsSet("amt_msat"):
+		if ctx.IsSet("amt") {
+			return fmt.Errorf("amt and amt_msat are mutually exclusive")
```
wpaulino: This is already checked at the RPC level, so we don't need to do so again here.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -192,7 +196,7 @@ var addHoldInvoiceCommand = cli.Command{
 func addHoldInvoice(ctx *cli.Context) error {
 	var (
 		descHash []byte
-		amt      int64
+		ms       int64
```
wpaulino: Nit: rename to `amtMsat`.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -214,9 +218,18 @@ func addHoldInvoice(ctx *cli.Context) error {
...
+		ms = ctx.Int64("amt") * 1000
+	case ctx.IsSet("amt_msat"):
+		if ctx.IsSet("amt") {
+			return fmt.Errorf("amt and amt_msat are mutually exclusive")
+		}
+		ms = ctx.Int64("amt_msat")
 	case args.Present():
-		amt, err = strconv.ParseInt(args.First(), 10, 64)
+		amt, err := strconv.ParseInt(args.First(), 10, 64)
+		ms = amt * 1000
```
wpaulino: Nit: move this below the error check. You can also use `lnwire.NewMSatFromSatoshis` instead of the multiplier.

cmd/lncli/invoicesrpc_active.go
akovalenko: Rewritten following the logic of this comment, so both amt and amt_msat are passed to rpc, and if both are present, the server errors out.

cmd/lncli/invoicesrpc_active.go
akovalenko: Thanks. Now as I'm passing both amt and amtMsat to RPC as is, no conversion is needed.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -212,12 +217,14 @@ func addHoldInvoice(ctx *cli.Context) error {
 
 	args = args.Tail()
 
-	switch {
-	case ctx.IsSet("amt"):
+	if ctx.IsSet("amt") {
 		amt = ctx.Int64("amt")
-	case args.Present():
+	}
+	if ctx.IsSet("amt_msat") {
```
wpaulino: Nit: the `IsSet` checks aren't needed here as if they're not set, then they'll just be 0 (the default).

cmd/lncli/invoicesrpc_active.go
```diff
@@ -212,12 +217,14 @@ func addHoldInvoice(ctx *cli.Context) error {
...
+		amtMsat = ctx.Int64("amt_msat")
+	}
+	if amt == 0 && amtMsat == 0 && args.Present() {
```
wpaulino: Any reason these checks were added? To override the amount used in the flag rather than the argument?

cmd/lncli/invoicesrpc_active.go
akovalenko: Will remove it.

cmd/lncli/invoicesrpc_active.go
akovalenko: In old code, there was a switch, taking --amt amount if it's present, and parsing an argument otherwise.
I decided to retain compatibility, making --amt, --amt_ms "prioritized", however, some edge cases will still behave differently (--amt 0, arg 182). To attain perfect compatibility, if !ctx.IsSet("amt") && !ctx.IsSet("amt_ms") should have been used.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -212,12 +217,11 @@ func addHoldInvoice(ctx *cli.Context) error {
 
 	args = args.Tail()
 
-	switch {
-	case ctx.IsSet("amt"):
-		amt = ctx.Int64("amt")
-	case args.Present():
-		amt, err = strconv.ParseInt(args.First(), 10, 64)
+	amt = ctx.Int64("amt")
+	amtMsat = ctx.Int64("amt_msat")
```
wpaulino: One final nit: use shorthand variable declaration `:=`.

akovalenko: Done.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -212,12 +215,11 @@ func addHoldInvoice(ctx *cli.Context) error {
 
 	args = args.Tail()
 
-	switch {
-	case ctx.IsSet("amt"):
-		amt = ctx.Int64("amt")
-	case args.Present():
-		amt, err = strconv.ParseInt(args.First(), 10, 64)
+	amt := ctx.Int64("amt")
+	amtMsat := ctx.Int64("amt_msat")
```
halseth: should fail if both are set

halseth: Ok, I saw @wpaulino's comment about this now. Ignore my comment 😛

akovalenko:
> should also be done for regular `addInvoice`?

I'll likely do it, but in a separate PR.
Started with addHoldInvoice because it's needed for my pet project.

cmd/lncli/invoicesrpc_active.go
```diff
@@ -212,12 +215,11 @@ func addHoldInvoice(ctx *cli.Context) error {
...
 
+	if amt == 0 && amtMsat == 0 && args.Present() {
```
halseth: Actually, we should use `IsSet` here, since it will distinguish from attempting to create a zero value invoice

akovalenko: Amended my patch to use IsSet.

joostjager:
Why do we do this only for hold invoices and leave the regular `lncli` commands without `--amt_msat`?

akovalenko:
> Why do we do this only for hold invoices and leave the regular `lncli` commands without `--amt_msat`?

I'll do it for addinvoice as well, but as separate PR. Started with a command actually needed for my project.

joostjager:
Ok. Out of curiosity, can you share your use case for msat hodl invoices with u

akovalenko:
> Ok. Out of curiosity, can you share your use case for msat hodl invoices with us?

My LN-to-RussianRUB [conversion bot](https://t.me/LnToRubBot) now has [LNURL-pay front-end](https://vds.sw4me.com/rulnurl/). Flexible-amount LNURL-pay receives user-specified "amount" in msat, and my service has to answer with payment request for exact same amount.

BLW actually sends non-integer-sat amount when a user gives amount in fiat currency. Unfortunately, response with payment request for rounded-down-to-sat amount is not accepted by BLW, hence my need for this patch.

I use hold invoices because fiat payout may fail for multiple reasons (insufficient reserves, incorrect destination, etc.). My service waits for invoice to become "accepted", attempts fiat payment and settles or cancels the invoice depending on fiat payment success.

joostjager:
Interesting. Thx for sharing

PR #3749, opened 2019-11-21
routing: local balance check before path finding

### LINKED ISSUE #1680: Incorrect payment error message

joostjager:
Fixes #1680


routing/pathfind.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/coreos/bbolt"
+	"github.com/go-errors/errors"
```
cfromknecht: wrong import

routing/payment_session.go
```diff
@@ -1,8 +1,7 @@
 package routing
 
 import (
-	"fmt"
-
+	"github.com/go-errors/errors"
```
cfromknecht: import

routing/payment_lifecycle.go
```diff
@@ -186,6 +186,19 @@ func (p *paymentLifecycle) resumePayment() ([32]byte, *route.Route, error) {
 
 }
 
+// errorToPaymentFailure takes a path finding error and converts it into a
+// payment-level failure.
+func errorToPaymentFailure(err error) channeldb.FailureReason {
+	switch err {
+	case errNoTlvPayload, errNoPathFound, errMaxHopsExceeded,
```
cfromknecht: `errNoTlvPayload` isn't necessarily a path issue, but maybe that should be addressed in a different pr?

joostjager: Yes, I left that part unchanged. Just made it more obvious that that is the current logic. But agree that we should differentiate more. Another failure that we could expose is not having any info on the node at all (not in graph, not in the hints).

channeldb/payments.go
```diff
@@ -105,6 +105,10 @@ const (
 	// or the final cltv delta or amount is incorrect.
 	FailureReasonPaymentDetails FailureReason = 3
 
+	// FailureReasonInsufficientBalance indicates that we didn't have enough
+	// balance to complete the payment.
+	FailureReasonInsufficientBalance = 4
```
joostjager: Do we need a db version bump? I am not completely convinced that we should add this protection against downgrading on nodes running master.

joostjager:
Todo: add test

channeldb/payments.go
joostjager: Discussed with @Roasbeef offline. Up to now we generally don't give this level of support to users on master.

guggero: Should add explicit `FailureReason` type.

routing/pathfind.go
```diff
@@ -305,6 +309,50 @@ type PathFindingConfig struct {
...
+		if outgoingChan != nil && chanID != *outgoingChan {
+			return nil
+		}
+
+		bandwidth, ok := g.bandwidthHints[chanID]
+
+		// If the bandwidth is not available for whatever reason, don't
+		// fail the pathfinding early.
+		if !ok {
+			max = lnwire.MaxMilliSatoshi
```
guggero: If there is no bandwidth hint for just one of the channels, this causes the `getMaxOutgoingAmt` to always return `lnwire.MaxMilliSatoshi` because no other channel can overwrite the max with a "real" value. Should this maybe just be ignored? I guess it depends on what the causes for the hints being empty could be and if it's likely that the hint for just one channel is missing or that the hints for all of them are missing.

routing/pathfind_test.go
```diff
@@ -2299,16 +2299,13 @@ func newPathFindingTestContext(t *testing.T, testChannels []*testChannel,
 		t:                 t,
 		testGraphInstance: testGraphInstance,
 		source:            route.Vertex(sourceNode.PubKeyBytes),
+		pathFindingConfig: *testPathFindingConfig,
```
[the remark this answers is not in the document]
joostjager: Yes, it somehow disappeared. I didn't find out what I did wrong, maybe a rebase mistake

routing/errors.go
```diff
@@ -11,16 +11,6 @@ const (
 	// does not exist in the graph.
```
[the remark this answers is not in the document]
joostjager: Removed

channeldb/payments.go
joostjager: Added

routing/pathfind.go
joostjager: There may be a race condition. A channel becoming available right after the bandwidth hints are queried. It is an edge case and it felt safer to do it like this. Not to fail the routing while there is a channel available that just doesn't have a bandwidth hint.

routing/errors.go
```diff
@@ -7,18 +7,10 @@ import "github.com/go-errors/errors"
...
-
-	// ErrMaxHopsExceeded is returned when a candidate path is found, but
-	// the length of that path exceeds HopLimit.
-	ErrMaxHopsExceeded
-
 	// ErrTargetNotInNetwork is returned when the target of a path-finding
 	// or payment attempt isn't known to be within the current version of
 	// the channel graph.
-	ErrTargetNotInNetwork
+	ErrTargetNotInNetwork errorCode = iota
```
cfromknecht: preexisting, but `errorCode` should be exported if `ErrTargetNotInNetwork` is also exported for godocs

cfromknecht: could be made into a trivial follow up

cfromknecht:
two of the async benchmarks failed, perhaps an indicator that this introduces a performance regression?

joostjager:
> two of the async benchmarks failed, perhaps an indicator that this introduces a performance regression?

Manually checked the timing of a queryroutes call on testnet, didn't see a difference with master. Restarted itests

joostjager:
Yes, does seem to be failing consistently. Will check it out.

joostjager:
Turned out to be a deadlock with getting the source node. Fixed now

Issue #3692, opened 2019-11-07
Balance "in limbo" on 101 pending channels 


aantonop:
### Background

I have 101 pendinchannels, with a big balance "in limbo".

2 pending force close, -8000 and -2500 blocks beyond maturity
99 waiting close channels

I have 78 channels operating without problems

Problem may have happened during downtime when I migrated lnd to another host and IP address.

### Your environment

LN Version: 0.8.0-beta
Ubuntu 18.04
Bitcoin Core Daemon version v0.18.0.0-g2472733a24a9364e4c6233ccd04166a26a68cc65

### Workaround and fixes already tried:

- Upgraded lnd to latest
- Restarted lnd (dozens of times over weeks)
- Checked that bitcoind and lnd are fully synced
- Removed and restored wallet from seed
- Re-scanned wallet addresses in lnd (by restoring) almost 90,000 blocks back

### Relevant info

Sample of pendingchannels:

```
"pending_force_closing_channels": [
        {
            "channel": {
                "remote_node_pub": "026bf2df2251e7ce8dde6cfd3949514d8c6b233a08c6a5f63c86062ace365bfcff",
                "channel_point": "44fd13584212f9b6f29f976724107f6191500b3a936d1398cad020276c9d9471:0",
                "capacity": "394774",
                "local_balance": "51001",
                "remote_balance": "0",
                "local_chan_reserve_sat": "0",
                "remote_chan_reserve_sat": "0"
            },
            "closing_txid": "80c945a5130611a0a948eecac268ad182c7694288c1c346a2dfa3f949494d755",
            "limbo_balance": "51001",
            "maturity_height": 594713,
            "blocks_til_maturity": -7392,
            "recovered_balance": "0",
            "pending_htlcs": [
            ]
        },
```
Sample of waiting_close_channel:

        {
            "channel": {
                "remote_node_pub": "0233502f6370758ede277d0ee5308a900feffeea29dbe9fc9593d1d0c30b2eb30e",
                "channel_point": "7cb442c165f461c086add987a0f2658c59274246523a508cb8eee5149d2bf379:0",
                "capacity": "2500000",
                "local_balance": "1408248",
                "remote_balance": "1088097",
                "local_chan_reserve_sat": "25000",
                "remote_chan_reserve_sat": "25000"
            },
            "limbo_balance": "1408248"
        },
        {
            "channel": {
[...]

PR #3752, opened 2019-11-21
[chainntnfs] add more logging for spend requests

halseth:
To help track down the root cause for #3692

### DIFF
--- a/chainntnfs/interface.go
+++ b/chainntnfs/interface.go
@@ -241,6 +241,12 @@ type SpendDetail struct {
 	SpendingHeight    int32
 }
 
+// String returns a string representation of SpendDetail.
+func (s *SpendDetail) String() string {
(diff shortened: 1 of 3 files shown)


chainntnfs/interface.go
```diff
@@ -241,6 +241,12 @@ type SpendDetail struct {
 	SpendingHeight    int32
 }
 
+// String returns a string representation of SpendDetail.
+func (s *SpendDetail) String() string {
+	return fmt.Sprintf("%v[%d] spending %v at height=%v", s.SpenderTxHash,
+		s.SpenderInputIndex, s.SpentOutPoint, s.SpendingHeight)
```
wpaulino: This will show a zero outpoint for script spends.

halseth: Looks like it will be set based on the `TxIn` when a spend is detected, and it therefore should be non-zero?

PR #3756, opened 2019-11-22
Makefile: non-verbose unit tests

cfromknecht:
Makes it easier to spot failures without scrolling through pages of
successful tests.


cfromknecht:
Indeed, and it's also more difficult to search for compilation failures because they won't have a `--- FAIL` to search for

cfromknecht:
As a final thought, another benefit is that the execution time per package is more concisely displayed, allowing us to pinpoint hotspots in our unit tests. ATM lnwallet is by far the most expensive package to test, followed by `htlcswitch` and `keychain`.

PR #3758, opened 2019-11-25
cnct: anchor output sweeping and fee bumping

joostjager:
LND supports an experimental commitment format that produces anchor outputs (#3829, https://github.com/lightningnetwork/lightning-rfc/pull/688). Anchor outputs allow users to increase the fee of their commitment txes after they've been published. It allows timely confirmation even if the miner fee of the commitment tx itself is insufficient.

This PR adds the functionality required to do so. When the commitment tx isn't confirming, the `BumpFee` rpc can be used to increase the fee of the total package. Anchor outpoints that are available can be discovered via the `PendingChannels` and `PendingSweeps` rpcs.

To make the experience more friendly towards users, a composite `lncli` command `bumpclosefee` is added.

For channels in the `waiting_close_channels` stage, this command can be used as follows:

` lncli wallet bumpclosefee 2b66cf489298b5b4a9305127413e4124e385588178e4b0fc4cec2d28c277b17d:0 --sat_per_byte 15`

Note that lnd needs to be built with the `walletrpc` subserver enabled.

Important implementation changes made:
* Keeping certain anchor inputs in separate sweep transactions always. When the commitment transaction hasn't confirmed yet, we don't know which of the three versions (local, remote, remote pending) will confirm. We attempt to sweep all three anchors, but batching those together would always create an invalid tx.
* A dedicated bucket for sweeping at the minimum fee rate. Anchor outputs are very small and if they would be batched in a higher fee rate bucket, they are unlikely to be economical.
* Try to sweep anchors before and after commit confirms
* Report status of anchor sweep


Roasbeef:
Needs rebase.

joostjager:
PR is ready for a first pass

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,170 @@
...
+			},
+			Force: true,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	select {
+	case sweepRes := <-resultChan:
```
joostjager: Comment from @Roasbeef: In a high fee situation, the sweep may never succeed. Do channels remain pending forever?

lnwallet/channel.go
```diff
@@ -5115,6 +5115,11 @@ type UnilateralCloseSummary struct {
 	// RemoteCommit is the exact commitment state that the remote party
 	// broadcast.
 	RemoteCommit channeldb.ChannelCommitment
+
+	// AnchorResolution contains the data required to sweep the anchor
```
cfromknecht:
```suggestion
	// AnchorResolution contains the data required to sweep our anchor
```

sweep/sweeper.go
```diff
@@ -714,7 +714,15 @@ func (s *UtxoSweeper) sweepCluster(cluster inputCluster,
 func (s *UtxoSweeper) bucketForFeeRate(
 	feeRate chainfee.SatPerKWeight) int {
 
-	return int(feeRate-s.relayFeeRate) / s.cfg.FeeRateBucketSize
+	// Create an isolated bucket for sweeps at the minimum fee rate. This is
+	// to prevent very small outputs (anchors) to become uneconomical if
```
cfromknecht:
```suggestion
	// to prevent very small outputs (anchors) from becoming uneconomical if
```

contractcourt/channel_arbitrator.go
```diff
@@ -354,6 +359,10 @@ func (c *ChannelArbitrator) Start() error {
 		}),
 	)
 
+	// Retrieve relay fee for dust limit calculation. Assume that this will
+	// not change from here on.
```
cfromknecht: is this a safe assumption?

contractcourt/channel_arbitrator.go
```diff
@@ -859,6 +865,25 @@ func (c *ChannelArbitrator) stateStep(
 		case chainTrigger, userTrigger:
 			log.Infof("ChannelArbitrator(%v): noop trigger %v",
 				c.cfg.ChanPoint, trigger)
+
+			// The commitment transaction has been broadcast, but it
+			// doesn't necessarily need to be the commitment
+			// transaction version that is going to be confirmed. To
+			// be sure that we any of those versions can be anchored
```
cfromknecht:
```suggestion
			// be sure that any of those versions can be anchored
```

contractcourt/channel_arbitrator.go
```diff
@@ -990,6 +1015,60 @@ func (c *ChannelArbitrator) stateStep(
...
+
+	if anchors.RemotePendingCommitAnchor != nil {
+		anchorList = append(
+			anchorList, anchors.RemotePendingCommitAnchor,
+		)
+	}
+
+	// Use the chan id as the exclusive group. This prevents any of these
+	// three anchors from being batched together.
+	exclusiveGroup := c.cfg.ShortChanID.ToUint64()
```
cfromknecht: very cool 😲

contractcourt/anchor_resolver.go
```diff
@@ -126,6 +152,15 @@ func (c *anchorResolver) IsResolved() bool {
 	return c.resolved
 }
 
+// report returns a report on the resolution state of the contract.
+func (c *anchorResolver) report() *ContractReport {
+	c.reportLock.Lock()
+	defer c.reportLock.Unlock()
+
+	copy := c.currentReport
```
cfromknecht: collides with builtin method

contractcourt/channel_arbitrator.go
Roasbeef: indeed 😎

contractcourt/channel_arbitrator.go
```diff
@@ -859,6 +865,25 @@ func (c *ChannelArbitrator) stateStep(
...
+			// be sure that we any of those versions can be anchored
+			// down, we now submit all anchor resolutions to the
+			// sweeper. The sweeper will keep trying to sweep all of
+			// them.
+			anchors, err := c.cfg.Channel.NewAnchorResolutions()
+			if err != nil {
+				return StateError, closeTx, err
+			}
+			if anchors != nil {
+				err = c.sweepAnchors(anchors, triggerHeight)
```
Roasbeef: Perhaps we should make this configurable? As we still attempt to set our commitment fee for a timely confirmation, this'll most likely just end up costing the user additional fees.

rpcserver.go
```diff
@@ -2790,6 +2790,12 @@ func (r *rpcServer) arbitratorPopulateForceCloseResp(chanPoint *wire.OutPoint,
 		case contractcourt.ReportOutputIncomingHtlc,
 			contractcourt.ReportOutputOutgoingHtlc:
 
+			// Don't report details on htlcs that are no longer in
```
Roasbeef: Is an accompanying extension to the existing integration tests planned?

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"errors"
+	"io"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/sweep"
+)
+
+// anchorResolver is a resolver that will attempt to sweep our anchor output.
```
Roasbeef: I think this may actually need to live independent of any given channel arb. Otherwise, they'll never transition from pending close to fully closed, since there'll be a lingering contract that's unresolved.

lnwallet/channel.go
```diff
@@ -5229,12 +5234,20 @@ func NewUnilateralCloseSummary(chanState *channeldb.OpenChannel, signer input.Si
 		closeSummary.LastChanSyncMsg = chanSync
 	}
 
+	anchorResolution, err := GetAnchorResolution(
```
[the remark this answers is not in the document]
joostjager: yes, more consistent. renamed

lnwallet/channel.go
```diff
@@ -5717,6 +5740,11 @@ type LocalForceCloseSummary struct {
 	// ChanSnapshot is a snapshot of the final state of the channel at the
 	// time the summary was created.
 	ChanSnapshot channeldb.ChannelSnapshot
+
+	// AnchorResolution contains the data required to sweep the anchor
+	// output. If the channel type doesn't include anchors, the value of
+	// this field will be nil.
+	AnchorResolution *AnchorResolution
```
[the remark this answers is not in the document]
joostjager: Do you mean anchoring down breaches? That is out of scope for this pr.

lnwallet/channel.go
```diff
@@ -6019,6 +6055,127 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
...
+
+	// RemoteCommitAnchor is the anchor outpoint on the acked remote
+	// commitment. This is nil if the current channel type doesn't have
+	// anchor outputs.
+	RemoteCommitAnchor *AnchorResolution
+
+	// RemotePendingCommitAnchor is the anchor outpoint on the pending ack
+	// remote commitment. This is nil if the current channel type doesn't
+	// have anchor outputs or there are no pending commit.
+	RemotePendingCommitAnchor *AnchorResolution
```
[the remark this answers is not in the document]
joostjager: I see what you mean. Local/Remote/RemotePending refers to the version of the commitment tx, not which one of the two anchors. I could prefix everything with `Our`, like `OurRemoteCommitAnchor`. But it gets a bit long and also we don't deal with anchors for the remote party anywhere. There isn't that confusion. I think the comments do make it sufficiently clear that the current prefix refers to the commitment version. Added a bit more explanation to the type comment.

lnwallet/channel.go
```diff
@@ -6019,6 +6055,127 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
...
+
+		resolutions.RemotePendingCommitAnchor = remotePendingRes
+	}
+
+	return resolutions, nil
+}
+
+// GetAnchorResolution returns the information that is required to sweep the
+// local anchor.
+func GetAnchorResolution(chanState *channeldb.OpenChannel,
```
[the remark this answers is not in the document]
joostjager: I like `NewAnchorResolution` the most.

lnwallet/channel.go
```diff
@@ -6019,6 +6055,127 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
...
+// NewAnchorResolutions returns the anchor resolutions for the channel state. It
+// takes into consideration all commitment txes that are currently valid.
+// Because we have no view on the mempool, we can only blindly anchor all of
+// these txes down.
+func NewAnchorResolutions(chanState *channeldb.OpenChannel) (*AnchorResolutions,
+	error) {
+
+	// Return nil resolution if the channel has no anchors.
+	if !chanState.ChanType.HasAnchors() {
+		return nil, nil
```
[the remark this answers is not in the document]
joostjager: Yes I had that before. Bringing that back now, reduces the number of states.

lnwallet/channel.go
```diff
@@ -6019,6 +6055,127 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
...
+
+	// Derive our local anchor script and look it up on the commitment tx.
+	localAnchor, _, err := CommitScriptAnchors(
+		&chanState.LocalChanCfg, &chanState.RemoteChanCfg,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	var outPoint *wire.OutPoint
```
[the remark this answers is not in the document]
joostjager: Nice, fixed

lnwallet/channel.go
```diff
@@ -6019,6 +6055,127 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
...
+		}
+	}
+
+	if outPoint == nil {
+		return nil, errors.New("anchor not found")
+	}
+
+	// Instantiate the sign descriptor that allows sweeping of the anchor.
+	signDesc := &input.SignDescriptor{
+		SingleTweak:   nil,
```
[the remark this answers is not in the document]
joostjager: removed

lnwallet/channel_test.go
```diff
@@ -594,6 +617,14 @@ func TestForceClose(t *testing.T) {
 			1, len(closeSummary.HtlcResolutions.IncomingHTLCs))
 	}
 
+	// Verify that anchor resolution is present for the anchor commitment
+	// format.
+	if testCase.chanType.HasAnchors() {
+		if closeSummary.AnchorResolution == nil {
```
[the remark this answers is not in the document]
joostjager: Assertions added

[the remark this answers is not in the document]
joostjager: What do you mean exactly by the pending force close case in this context? In a later commit, a test is added that asserts pre-confirmation sweeping.

sweep/sweeper.go

contractcourt/chain_arbitrator.go
```diff
@@ -47,6 +48,20 @@ type ResolutionMsg struct {
 	PreImage *[32]byte
 }
 
+// ArbSweeper is an abstraction that allows the contract court to interact with
```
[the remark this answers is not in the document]
joostjager: Really? Yes it is true indeed. Ok, dropped commit.

contractcourt/channel_arbitrator.go
joostjager: I guess you could restart the backend with a different fee rate. Sweeper already makes this assumption though. Fixing that doesn't seem high prio to me. But I reworked this commit to obtain the relay fee rate from the sweeper. At least make the assumption only in one place.

contractcourt/channel_arbitrator.go
joostjager: We try to sweep at the min fee rate here. Assuming 1 sat/b, the sweep will be net positive.

contractcourt/channel_arbitrator.go
joostjager: Glad you guys like it

rpcserver.go
joostjager: Do you mean specifically for this commit, or reporting on anchors in general? For anchors in general, my plan is to build that on top of #4001. Or possibly change a few of the asserted values in that pr when this pr lands first.

contractcourt/anchor_resolver.go
joostjager: renamed

contractcourt/anchor_resolver.go
joostjager: Add select clause to mark as resolved when the sweeper gives up. By that time, the 16 block csv delay has long expired, so if fees ever come down far enough, the anchor will probably be swept by one of the cleaning services that we envision.

joostjager:
Changed anchor resolver to reset force sweep flag. We don't want to sweep anchors post-confirmation that have a negative yield.

contractcourt/anchor_resolver.go
joostjager: Addressed via `sweep.ErrTooManyAttempts`

contractcourt/channel_arbitrator.go
```diff
@@ -850,6 +856,25 @@ func (c *ChannelArbitrator) stateStep(
...
+			// be sure that any of those versions can be anchored
+			// down, we now submit all anchor resolutions to the
+			// sweeper. The sweeper will keep trying to sweep all of
+			// them.
+			anchors, err := c.cfg.Channel.NewAnchorResolutions()
+			if err != nil {
+				return StateError, closeTx, err
+			}
+			if anchors != nil {
+				err = c.sweepAnchors(anchors, triggerHeight)
```
halseth: add log?

contractcourt/channel_arbitrator.go
```diff
@@ -850,6 +856,25 @@ func (c *ChannelArbitrator) stateStep(
 		case chainTrigger, userTrigger:
```
halseth: comment for case slightly outdated

contractcourt/channel_arbitrator.go
```diff
@@ -850,6 +856,25 @@ func (c *ChannelArbitrator) stateStep(
...
 				c.cfg.ChanPoint, trigger)
+
+			// The commitment transaction has been broadcast, but it
+			// doesn't necessarily need to be the commitment
+			// transaction version that is going to be confirmed. To
+			// be sure that any of those versions can be anchored
+			// down, we now submit all anchor resolutions to the
+			// sweeper. The sweeper will keep trying to sweep all of
+			// them.
+			anchors, err := c.cfg.Channel.NewAnchorResolutions()
```
halseth: This state will potentially be executed several times. Need to ensure this operation is idempotent, or we could add new state for anchor sweep publication (could also possibly be easier to test).

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,159 @@
...
+				sweepRes.Tx.TxHash())
+
+		// Anchor was swept by someone else. This is possible after the
+		// 16 block csv lock.
+		case sweep.ErrRemoteSpend:
+			c.log.Warnf("our anchor spent by someone else")
+
+		// The sweeper gave up on sweeping the anchor. We consider the
+		// anchor as being lost.
+		case sweep.ErrTooManyAttempts:
```
halseth: when is this being triggered?

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,159 @@
...
+			c.log.Errorf("unable to sweep anchor: %v", sweepRes.Err)
+
+			return nil, sweepRes.Err
+		}
+
+	case <-c.quit:
+		return nil, errResolverShuttingDown
+	}
+
+	c.resolved = true
```
halseth: How will this be communicated on the RPC in the case where we give up? Will the fact that we didn't sweep the anchor show somewhere in the close summary? (explicitly or implicitly )

input/size.go
```diff
@@ -406,6 +410,47 @@ const (
...
+			KeyDesc: keychain.KeyDescriptor{
+				PubKey: &dummyPubKey,
+			},
+			WitnessScript: dummyAnchorScript,
+		},
+		nil,
+	)
+
+	// AnchorWitnessSize 116 bytes
+	AnchorWitnessSize = dummyAnchorWitness.SerializeSize()
```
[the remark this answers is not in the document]
joostjager: yes! no more miscalculations!

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: Ended up removing the whole struct. There is no need to distinguish the resolutions, it only adds more code.

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: Converted to a list

contractcourt/channel_arbitrator.go
joostjager: log added inside `sweepAnchors`

contractcourt/channel_arbitrator.go
joostjager: fixed

contractcourt/channel_arbitrator.go
joostjager: Afaik there is no side-effect of this call. And I also believe it returns the same resolutions every time. Do you see a problem somewhere?

contractcourt/anchor_resolver.go
joostjager: comment added

contractcourt/anchor_resolver.go
joostjager: It is shown in the pending close state. The anchor value is part of the limbo or recovered balance and there is also the `anchor_recovered` boolean. However, once the channel is fully resolved that information is no longer accessible. But that is also the case for the outcome of other resolvers.

lnwallet/channel_test.go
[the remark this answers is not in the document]
joostjager: Also added a basic assertion for the pre-confirm case

cmd/lncli/walletrpc_active.go
```diff
@@ -168,3 +173,130 @@ func bumpFee(ctx *cli.Context) error {
...
+		// Bump fee of the anchor sweep.
+		fmt.Printf("Bumping fee of %v:%v\n",
+			sweepTxID, sweep.Outpoint.OutputIndex)
+
+		_, err = walletClient.BumpFee(ctxb, &walletrpc.BumpFeeRequest{
+			Outpoint:   sweep.Outpoint,
+			TargetConf: uint32(ctx.Uint64("conf_target")),
+			SatPerByte: uint32(ctx.Uint64("sat_per_byte")),
+			Force:      true,
+		})
```
joostjager: Problem to solve: this call clears the exclusive group.

joostjager: Fixed by making exclusive group static, see commit "sweep: leave exclusive group unchanged on parameter update"

contractcourt/channel_arbitrator.go
```diff
@@ -973,6 +1019,54 @@ func (c *ChannelArbitrator) stateStep(
...
+
+	// Use the chan id as the exclusive group. This prevents any of the
+	// anchors from being batched together.
+	exclusiveGroup := c.cfg.ShortChanID.ToUint64()
+
+	// Retrieve the current minimum fee rate from the sweeper.
+	minFeeRate := c.cfg.Sweeper.RelayFeePerKW()
+
+	for _, anchor := range anchors {
+		log.Tracef("ChannelArbitrator(%v): pre-confirmation sweep of "+
```
halseth: should at least be debug IMO

contractcourt/channel_arbitrator.go
```diff
@@ -837,11 +865,29 @@ func (c *ChannelArbitrator) stateStep(
 	// to be confirmed.
 	case StateCommitmentBroadcasted:
 		switch trigger {
-		// We are waiting for a commitment to be confirmed, so any
-		// other trigger will be ignored.
+
+		// We are waiting for a commitment to be confirmed.
```
halseth: should we remove this case now, instead move the anchor sweep out of the switch? (such that it is clear that it is executed for all triggers that don't take the state machine to a new state)

contractcourt/channel_arbitrator.go
```diff
@@ -837,11 +865,29 @@ func (c *ChannelArbitrator) stateStep(
...
 		case chainTrigger, userTrigger:
 			log.Infof("ChannelArbitrator(%v): noop trigger %v",
 				c.cfg.ChanPoint, trigger)
```
halseth: this log is now wrong

joostjager:
Fixed two issues:
* Updated `NewAnchorResolution` to not report a missing anchor as an error. If there is no output, an anchor may not be present on the commitment tx.
* Make `triggerHeight` in channel arbitrator to always be the closing height. Otherwise height hints may be too high and miss a spend.

joostjager:
Tried running #4001 on top of this. Needs a limited set of changes to account for the anchor sweep transactions, but other than that it passes.

contractcourt/channel_arbitrator.go
joostjager: added commit cnct: simplify channel arbitrator state logging

contractcourt/channel_arbitrator.go
joostjager: converted to debug

contractcourt/channel_arbitrator.go
joostjager: Comment added about idempotency

sweep/sweeper.go
```diff
@@ -1169,10 +1181,16 @@ func (s *UtxoSweeper) handleUpdateReq(req *updateReq, bestHeight int32) (
 		return nil, lnwallet.ErrNotMine
 	}
 
+	// Create the updated parameters struct. Leave the exclusive group
```
cfromknecht: 👍

lnrpc/rpc.proto
```diff
@@ -1938,6 +1938,23 @@ message PendingChannelsResponse {
...
+        /**
+        A list of valid commitment transactions. Any of these can confirm at
+        this point.
+        */
+        Commitments commitments = 3;
+    }
+
+    message Commitments {
+        /// Hash of the local version of the commitment tx.
+        string local_commit_hash = 1;
```
cfromknecht: nit: would `txid` be more accurate than `hash`? IMO the word `_commit_` is also redundant when the encapsulating field is `Commitments`

joostjager: Good suggestion, change made

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,167 @@
...
+	r.initLogger(r)
+
+	return r
+}
+
+// ResolverKey returns an identifier which should be globally unique for this
+// particular resolver within the chain the original contract resides within.
+func (c *anchorResolver) ResolverKey() []byte {
+	// The anchor resolver is stateless and doesn't need a database key.
+
```
halseth: nit: newline

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,167 @@
...
+	return nil
+}
+
+// Resolve offers the anchor output to the sweeper and waits for it to be swept.
+//
+// NOTE: This function MUST be run as a goroutine.
+func (c *anchorResolver) Resolve() (ContractResolver, error) {
+	// Attempt to update the sweep parameters to the post-confirmation
```
halseth: shuold have waitgroup?

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,167 @@
...
+	// situation. We don't want to force sweep anymore, because the anchor
+	// lost its special purpose to get the commitment confirmed. It is just
+	// an output that we want to sweep only if it is economical to do so.
+	relayFeeRate := c.Sweeper.RelayFeePerKW()
+
+	resultChan, err := c.Sweeper.UpdateParams(
+		c.anchor,
+		sweep.ParamsUpdate{
+			Fee: sweep.FeePreference{
+				FeeRate: relayFeeRate,
```
halseth: set `Force: false` for clarity

contractcourt/anchor_resolver.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+		},
+	)
+
+	// After a restart or when the remote force closes, the sweeper is not
+	// yet aware of the anchor. In that case, offer it as a new input to the
+	// sweeper. An exclusive group is not necessary anymore, because we know
+	// that this is the only anchor that can be swept.
+	if err == lnwallet.ErrNotMine {
+		anchorInput := input.MakeBaseInput(
```
halseth: can we just offer it like this in both cases, or won't that override what is already in the sweeper?

contractcourt/anchor_resolver.go
halseth: Ok for now, but in the future we probably want to stop trying exactly when 16 blocks deep, and start attempting  a no sig sweep instead.

contractcourt/anchor_resolver.go
halseth: For accounting purposes we should probably add a way of finding this information after it has been resolved (non blocking). I don't know if other resolvers are expected to be resolved in cases where we are unable to sweep them?

rpcserver.go
```diff
@@ -2812,6 +2812,11 @@ func (r *rpcServer) arbitratorPopulateForceCloseResp(chanPoint *wire.OutPoint,
 
 			forceClose.PendingHtlcs = append(forceClose.PendingHtlcs, htlc)
 
+		case contractcourt.ReportOutputAnchor:
+			if report.LimboBalance == 0 {
+				forceClose.AnchorRecovered = true
```
halseth: should have a comment

lnrpc/rpc.proto
```diff
@@ -1938,6 +1938,23 @@ message PendingChannelsResponse {
...
         int64 limbo_balance = 2;
+
+        /**
+        A list of valid commitment transactions. Any of these can confirm at
+        this point.
+        */
+        Commitments commitments = 3;
+    }
+
+    message Commitments {
```
halseth: Is it necessary to have this all be named? We could just make this aa list of commit txs

contractcourt/channel_arbitrator.go
```diff
@@ -391,12 +391,13 @@ func (c *ChannelArbitrator) Start() error {
...
 				trigger = remoteCloseTrigger
 			}
-			triggerHeight = c.cfg.ClosingHeight
 
 			log.Warnf("ChannelArbitrator(%v): detected stalled "+
 				"state=%v for closed channel, using "+
 				"trigger=%v", c.cfg.ChanPoint, c.state, trigger)
 		}
+
+		triggerHeight = c.cfg.ClosingHeight
```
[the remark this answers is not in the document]
joostjager: I think that if we come up in the state `ContractClosed`, we will create and persist resolvers with the best height as the height hint.

contractcourt/anchor_resolver.go
joostjager: I also get contrary comments about removing unnecessary initializers.

contractcourt/anchor_resolver.go
joostjager: Yes, won't override indeed.

lnrpc/rpc.proto
joostjager: I was in doubt about that, because I use a plain list internally too. But I thought it may be useful for debugging to have them named.

contractcourt/anchor_resolver.go
joostjager: Yes, @wpaulino had a case with a remote spend in which we are also forced to mark as resolved. Otherwise the channel would stay pending forever.

I consider fixing accounting to be out of scope. It is a whole another project, even just for the channel closes.

lnwallet/channel.go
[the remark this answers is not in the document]
Roasbeef: OK, we should make a follow up issue to tracks what's remaining to make anchors first-class supported like the rest of our commitment formats.

lnwallet/channel_test.go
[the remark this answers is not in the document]
Roasbeef: By pending force close I mean when we have a dangling commitment, so we need to be ready to play both commitments of the remote party.

rpcserver.go
Roasbeef: SGTM to build on top of #4001. Timing wise, it may not need to land, but it's important that we're at least able to _run_ them all to find any potentially issues.

contractcourt/anchor_resolver.go
Roasbeef: When will the sweeper "give up"? Will check out the diff for more details...

contractcourt/anchor_resolver.go
Roasbeef: We never set things like `started: 0`, elsewhere in the codebase when initializing sub-systems.

lnrpc/rpc.proto
Roasbeef: Yeah I can def see this being useful for debugging user tickets.

halseth: Mhm, not that we will have to add many more fields here in the future.

lnwallet/channel_test.go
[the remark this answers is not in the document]
joostjager: Ok. So the anchor resolution on the local force close summary is only used when our local commitment tx has confirmed. We can ignore the other versions at that point. Pre-confirm we indeed play every version of the commitment.

contractcourt/anchor_resolver.go
joostjager: There is a comment about this in the `Resolve` method

contractcourt/anchor_resolver.go
joostjager: removed

contractcourt/anchor_resolver.go
joostjager: There is a waitgroup for this goroutine, the add happens in `ChannelArbitrator.launchResolvers`

rpcserver.go
joostjager: comment added. Also changed condition to `RecoveredBalance != 0`, because an anchor may also be lost.

joostjager:
One more issue found. Recording commit https://github.com/lightningnetwork/lnd/pull/3758/commits/a9c08ab5161a310bbc0f3f90a83dde4415288850 as base for comparison.

contractcourt/anchor_resolver.go
halseth: Ok looking at the callsite I think the comment `NOTE: This function MUST be run as a goroutine.` is wrong, so I wouldn't mind removing it. But this is also pre-existing.

contractcourt/anchor_resolver.go
halseth: IMO this is a bit different, since we are overwriting the existing true value. It definitely doesn't hurt readability to add.

contractcourt/anchor_resolver.go
halseth: one day @carlaKC will fix accounting and the world will have `lnd v1.0`

rpcserver.go
halseth: shouldn't we attempt to report all the states it could be in?

rpcserver.go
```diff
@@ -2756,6 +2756,42 @@ func (r *rpcServer) PendingChannels(ctx context.Context,
...
+		if waitingClose.RemoteCommitment.CommitTx != nil {
+			commitments.RemoteTxid =
+				waitingClose.RemoteCommitment.CommitTx.TxHash().
+					String()
+		}
+
+		// Report the remote pending commit if any.
+		remoteCommitDiff, err := waitingClose.RemoteCommitChainTip()
+
+		switch err {
```
halseth: I think we usually do 
```go
switch {
case err == ErrNoPending:
case err != nil:
     return nil, err
default:
      hash := ....
} 
```
(non-blocking)

contractcourt/channel_arbitrator.go
```diff
@@ -391,12 +391,13 @@ func (c *ChannelArbitrator) Start() error {
 			case channeldb.RemoteForceClose:
 				trigger = remoteCloseTrigger
 			}
-			triggerHeight = c.cfg.ClosingHeight
 
 			log.Warnf("ChannelArbitrator(%v): detected stalled "+
 				"state=%v for closed channel, using "+
 				"trigger=%v", c.cfg.ChanPoint, c.state, trigger)
 		}
+
```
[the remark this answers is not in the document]
joostjager: Moved outside

input/size.go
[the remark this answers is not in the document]
joostjager: added

contractcourt/anchor_resolver.go
joostjager: Yes indeed. It is up to the caller to decide how to run this. And as you say, the call site already says it should be run as a goroutine

joostjager: Removed comment

contractcourt/anchor_resolver.go
joostjager: Added. I also prefer to add it, but tried to comply with the unwritten code standard for this

rpcserver.go
joostjager: all states exposed

rpcserver.go
joostjager: changed to suggested style

joostjager:
Need to fix `restore_from_backup_file_anchors` itest

joostjager:
Anchors itest failure fixed as part of 'cnct: pre-confirmation anchor sweep'

PR #3763, opened 2019-11-26
cnct: keep open channel data after channel commitment tx confirms

joostjager:
This PR is a preparation for the addition of the anchor resolver. By keeping the open channel data around, we can avoid persisting the anchor resolutions in the database. This simplifies the implementation and reduces the number of execution paths (especially after a restart of `lnd`).

It also opens up the possibility to convert other resolvers to exist in memory only, although this is a long term goal - if it is a goal at all.

In general, having the channel data available allows for more flexibility in implementations.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -37,6 +37,11 @@ var (
 	// TODO(roasbeef): flesh out comment
 	openChannelBucket = []byte("open-chan-bucket")
 
+	// historicalChannelBucket stores all channels that have seen their
+	// commitment tx confirm. All information from their previous open state
+	// is retained.
+	historicalChannelBucket = []byte("historical-chan-bucket")
+
 	// chanInfoKey can be accessed within the bucket for a channel
 	// (identified by its chanPoint). This key stores all the static
 	// information for a channel which is decided at the end of  the
@@ -2229,7 +2234,8 @@ func (c *OpenChannel) CloseChannel(summary *ChannelCloseSummary) error {
 		if err != nil {
 			return err
 		}
-		chanBucket := chainBucket.Bucket(chanPointBuf.Bytes())
+		chanKey := chanPointBuf.Bytes()
+		chanBucket := chainBucket.Bucket(chanKey)
 		if chanBucket == nil {
 			return ErrNoActiveChannels
 		}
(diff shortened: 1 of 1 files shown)


channeldb/db.go
```diff
@@ -549,6 +555,14 @@ func (d *DB) FetchChannel(chanPoint wire.OutPoint) (*OpenChannel, error) {
...
 				}
 
+				// Don't return the channel if the commitment tx
+				// has already confirmed.
+				if channel.HasChanStatus(
+					ChanStatusCommitConfirmed,
+				) {
+					return nil
+				}
+
```
joostjager: This is to keep `FetchChannel` backwards compatible. But maybe better to perform this check at the various call sites.

guggero:
Should we maybe also keep the data until the `to_remote` output is fully spent? Because if we delete the local revocation data (which is used for the local node's `per_commit_point`) then a client restoring from SCB will never get the `per_commit_point` from us after we delete that data.

joostjager:
Yes, that would require tracking also the remote resolution processes. Maybe even their second level sweep? That is a big change. I think the only alternative that we have is to not delete the pending close bucket data at all.

halseth:
The latest message used for dataloss protection is stored in the `ClosedChannelSummary`: https://github.com/lightningnetwork/lnd/blob/db21c394b2ac1bbda2d5ed5592f7e311e67a88e7/channeldb/channel.go#L2178

That includes the commitment point.

joostjager:
Offline discussion:
* Keep around forever

channeldb/channel.go
```diff
@@ -37,6 +37,11 @@ var (
 	// TODO(roasbeef): flesh out comment
 	openChannelBucket = []byte("open-chan-bucket")
 
+	// closedChannelBucket stores all channels that have seen their
+	// commitment tx confirm. All information from their previous open state
+	// is retained.
+	closedChannelBucket = []byte("confirmed-chan-bucket")
```
cfromknecht: nit: "confirmed" is ambiguous. to me it sounds like the transition from pending->open. perhaps historical-chan-bucket?

joostjager: renamed

PR #3767, opened 2019-11-26
zpay32: Fix broken last tagged field

matheusd:
This fixes an issue with bech32 decoding where inserting a specific character in the checksum might generate a valid signature albeit with a different destination node than the original invoice.

This fixes the issue by ensuring that the last tagged field has all the required data instead of silently discarding a partial field, such as a field with only a type element but no length elements.

I tested this behavior in other implementations. c-lightning and lightning-payencode (python) fail to decode invoices with a partial field. Bolt11 (node) accepts it.

This should probably be clarified in BOLT-11 so that the behavior is consistent across implementations.

For context: https://github.com/sipa/bech32/issues/51

### DIFF
--- a/zpay32/invoice_test.go
+++ b/zpay32/invoice_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


matheusd:
> We could start adding the tagged field for pubkey by default to mitigate

Other options (but these involves changing BOLT-11) would be to include the length of tagged data as a prefixed field (along with the timestamp) or as a new tagged field. These have the advantage of being smaller than the pubkey.

> if someone can modify an invoice like this to change the destination, they probably can also just resign the invoice.

Indeed.

zpay32/invoice_test.go
```diff
@@ -843,6 +844,56 @@ func TestMaxInvoiceLength(t *testing.T) {
...
+	// We thus modify the checksum here and verify the invoice gets broken
+	// enough that it fails to decode.
+	if !strings.HasSuffix(encoded, "p") {
+		t.Logf("Invoice: %s", encoded)
+		t.Fatalf("Generated invoice checksum does not end in 'p'")
+	}
+	encoded = encoded[:len(encoded)-1] + "qp"
+
+	_, err = Decode(encoded, chain)
+	if err == nil {
```
cfromknecht: should we test for the expected error, rather than passing for any error?

matheusd: While adding a single `q` will cause the "broken last tagged field error", adding other combinations (`qq`, `qqq`) will cause different errors, so that's why I didn't explicitly add the concrete error check.

I could add a comment explaining this, add the concrete error check or add test cases for the other situations, whichever you feel is more appropriate.

cfromknecht: ah i see, in that case i think it's fine as is. thanks for explaining!

PR #3770, opened 2019-11-27
channeldb+invoices: prepare invoice db for mpp

joostjager:
This PR refactors the invoice update logic to prepare it for new update flows that are required for mpp.

The main functional preparation is to make sure that the invoice `amtPaid` is recalculated with every update. For mpp we don't want to reflect accepted htlcs in the `amtPaid` until the invoice is no longer open.

Follow-up: #3415


invoices/invoiceregistry_test.go
```diff
@@ -137,10 +155,10 @@ func TestSettleInvoice(t *testing.T) {
 	hodlChan := make(chan interface{}, 1)
 
 	// Try to settle invoice with an htlc that expires too soon.
-	event, err := registry.NotifyExitHopHtlc(
+	event, err := ctx.registry.NotifyExitHopHtlc(
 		hash, testInvoice.Terms.Value,
 		uint32(testCurrentHeight)+testInvoiceCltvDelta-1,
-		testCurrentHeight, getCircuitKey(10), hodlChan, nil,
+		testCurrentHeight, getCircuitKey(10), hodlChan, testPayload,
```
halseth: why this change from nil to `testPayload`?

record/mpp.go
```diff
@@ -96,3 +97,7 @@ func (r *MPP) Record() tlv.Record {
 		MPPOnionType, r, size, MPPEncoder, MPPDecoder,
 	)
 }
+
+func (r *MPP) String() string {
```
halseth: exported methods need godoc

invoices/invoiceregistry.go
```diff
@@ -1043,3 +955,114 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
 
 	delete(i.hodlReverseSubscriptions, subscriber)
 }
+
+// invoiceUpdater is an object that is responsible for updating the invoice
+// state and storing actions that need to be carried out after the invoice is
+// updated.
+type invoiceUpdater struct {
+	// update input fields
```
halseth: wip?

invoices/invoiceregistry.go
```diff
@@ -1043,3 +955,114 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+// settlement logic.
+func (i *invoiceUpdater) update(inv *channeldb.Invoice) (
+	*channeldb.InvoiceUpdateDesc, error) {
+
+	// Don't update the invoice when this is a replayed htlc.
+	htlc, ok := inv.Htlcs[i.circuitKey]
+	if ok {
+		switch htlc.State {
+		case channeldb.HtlcStateCanceled:
+			i.result = "replayed htlc to canceled invoice"
```
halseth: this is a bit awkward. Could we just debug log here instead?

invoices/invoiceregistry.go
```diff
@@ -439,113 +439,25 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
 	// returned to the link or contract resolver.
-	invoice, err := i.cdb.UpdateInvoice(rHash, updateInvoice)
+	updater := invoiceUpdater{
+		circuitKey:           circuitKey,
+		amtPaid:              amtPaid,
+		expiry:               expiry,
+		currentHeight:        currentHeight,
+		finalCltvRejectDelta: i.finalCltvRejectDelta,
+	}
+	invoice, err := i.cdb.UpdateInvoice(rHash, updater.update)
```
halseth: nit: newline above

invoices/invoiceregistry.go
```diff
@@ -439,113 +439,25 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
-		updateSubscribers = true
-
-		return &update, nil
-	}
-
 	// We'll attempt to settle an invoice matching this rHash on disk (if
 	// one exists). The callback will set the resolution action that is
 	// returned to the link or contract resolver.
-	invoice, err := i.cdb.UpdateInvoice(rHash, updateInvoice)
+	updater := invoiceUpdater{
```
halseth: personal style preference: I think encapsulating these in a struct hides a little bit what is going on with the callback. What do you think about something like

```go
invoice, err := i.cdb.UpdateInvoice(
    rHash, InvoiceUpdateCallback  {
     return update( ... params ....)
})
```

channeldb/invoices.go
```diff
@@ -316,10 +316,12 @@ type InvoiceUpdateDesc struct {
 	// State is the new state that this invoice should progress to.
 	State ContractState
 
-	// Htlcs describes the changes that need to be made to the invoice htlcs
-	// in the database. Htlc map entries with their value set should be
-	// added. If the map value is nil, the htlc should be canceled.
-	Htlcs map[CircuitKey]*HtlcAcceptDesc
+	// CancelHtlcs describes the htlcs that need to be canceled.
```
halseth: add rationale for the change to commit msg?

channeldb/invoices.go
```diff
@@ -314,16 +314,22 @@ type HtlcAcceptDesc struct {
 // invoice.
 type InvoiceUpdateDesc struct {
 	// State is the new state that this invoice should progress to.
-	State ContractState
```
halseth: `NextState`?

channeldb/invoices.go
```diff
@@ -314,16 +314,22 @@ type HtlcAcceptDesc struct {
...
 
 	// AddHtlcs describes the newly accepted htlcs that need to be added to
 	// the invoice.
 	AddHtlcs map[CircuitKey]*HtlcAcceptDesc
+}
+
+// InvoiceStateUpdateDesc describes an invoice-level state transition.
+type InvoiceStateUpdateDesc struct {
+	// NewState is the new state that this invoice should progress to.
+	NewState ContractState
```
halseth: `NextState`

channeldb/invoices.go
```diff
@@ -314,16 +314,22 @@ type HtlcAcceptDesc struct {
...
 	// CancelHtlcs describes the htlcs that need to be canceled.
 	CancelHtlcs map[CircuitKey]struct{}
 
 	// AddHtlcs describes the newly accepted htlcs that need to be added to
 	// the invoice.
 	AddHtlcs map[CircuitKey]*HtlcAcceptDesc
+}
+
+// InvoiceStateUpdateDesc describes an invoice-level state transition.
+type InvoiceStateUpdateDesc struct {
```
halseth: I don't really see why these two fields have to go into a new struct

channeldb/invoices.go
```diff
@@ -1243,78 +1247,68 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
-	for key := range update.CancelHtlcs {
-		htlc, ok := invoice.Htlcs[key]
-		if !ok {
-			return nil, fmt.Errorf("unknown htlc %v", key)
-		}
-		if htlc.State != HtlcStateAccepted {
-			return nil, fmt.Errorf("can only cancel " +
-				"accepted htlcs")
+	// Update invoice state.
+	if update.State != nil {
```
halseth: when do we expect this to be nil? Should at least update the doc to explain.

channeldb/invoices.go
```diff
@@ -1243,78 +1247,68 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
 
 	// Process add actions from update descriptor.
 	for key, htlcUpdate := range update.AddHtlcs {
-		htlc, ok := invoice.Htlcs[key]
-
-		// Add new htlc paying to the invoice.
-		if ok {
-			return nil, fmt.Errorf("htlc %v already exists", key)
+		if _, exists := invoice.Htlcs[key]; exists {
+			return nil, errors.New("duplicate add")
```
halseth: keep error message as is?

channeldb/invoices.go
```diff
@@ -1327,6 +1321,109 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
 	return &invoice, nil
 }
 
+// updateInvoiceState validates and processes an invoice state update.
+func updateInvoiceState(invoice *Invoice, hash lntypes.Hash,
```
halseth: 👍

channeldb/invoices.go
```diff
@@ -1243,78 +1247,68 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+		err := updateInvoiceState(&invoice, hash, *update.State)
+		if err != nil {
+			return nil, err
 		}
 
-		htlc.State = HtlcStateCanceled
-		htlc.ResolveTime = now
-		invoice.AmtPaid -= htlc.Amt
+		if update.State.NewState == ContractSettled {
+			err := setSettleFields(
```
halseth: it is a bit confusing that both `updateInvoiceState` and `setSettleFields` is uptating the state. Could we make the first only validate the state update without altering the invoice?

channeldb/invoices.go
```diff
@@ -1327,6 +1321,109 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+		if update.Preimage.Hash() != hash {
+			return errors.New("preimage does not match")
+		}
+
+		switch invoice.State {
+		case ContractSettled:
+			return errors.New("invoice already settled")
+
+		case ContractCanceled:
+			return errors.New("invoice already canceled")
```
halseth: use errors like `channeldb.ErrInvoiceAlreadyCanceled`

invoices/invoiceregistry.go
```diff
@@ -574,32 +576,11 @@ func (i *InvoiceRegistry) CancelInvoice(payHash lntypes.Hash) error {
...
-			}
-
-			canceledHtlcs[key] = struct{}{}
-		}
-
 		// Move invoice to the canceled state.
 		return &channeldb.InvoiceUpdateDesc{
-			CancelHtlcs: canceledHtlcs,
-			State:       channeldb.ContractCanceled,
+			State: &channeldb.InvoiceStateUpdateDesc{
```
halseth: is the above state validation needed now?

channeldb/invoices.go
```diff
@@ -1243,78 +1247,68 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
-		invoice.Terms.PaymentPreimage = update.Preimage
-
-		// Settle all accepted htlcs.
-		for _, htlc := range invoice.Htlcs {
-			if htlc.State != HtlcStateAccepted {
-				continue
+	// Align htlc states with invoice state and recalculate amount paid.
+	var amtPaid lnwire.MilliSatoshi
+	for key, htlc := range invoice.Htlcs {
+		_, cancel := update.CancelHtlcs[key]
```
halseth: this reordering changes the behavior slightly, by now being able the cancel htlcs in the same update as adding them. Is that intentional?

invoices/invoiceregistry_test.go
joostjager: Added commit message

record/mpp.go
joostjager: added

invoices/invoiceregistry.go
joostjager: There are quite a few parameters (another one will be added for mpp later) and also two return values atm. It could be rewritten, but I favor this approach.

invoices/invoiceregistry.go
joostjager: no, not wip. it is a 'group comment'. I don't want to describe all these obvious fields again.

invoices/invoiceregistry.go
joostjager: Then I need to pass in another parameter (the payment hash), to keep the log line the same. I think it is no problem to leave it as is. @carlaKC is working on exposing these failure reasons and this also prepares for that.

channeldb/invoices.go
joostjager: I think the type name `UpdateDesc` already indicates that this is about the next.

channeldb/invoices.go
joostjager: added

channeldb/invoices.go
joostjager: To allow the `StateUpdate` to be nil, meaning no state update.

channeldb/invoices.go
joostjager: New, Next, does it matter?

channeldb/invoices.go
joostjager: Added explanation here and at the definition of the nillable `State` field.

channeldb/invoices.go
joostjager: Removed duplicate state updated and renamed function name.

channeldb/invoices.go
joostjager: I removed the hash, because it is already logged at the call site.

channeldb/invoices.go
joostjager: I didn't realize that. Check added.

channeldb/invoices.go
joostjager: done

invoices/invoiceregistry.go
joostjager: No not anymore, removed.

invoices/invoiceregistry_test.go
```diff
@@ -46,19 +50,27 @@ var (
...
-func newTestContext(t *testing.T) (*InvoiceRegistry, func()) {
+type testContext struct {
+	registry *InvoiceRegistry
+
+	cleanup func()
+	t       *testing.T
+}
+
+func newTestContext(t *testing.T) *testContext {
+
```
carlaKC: unnecessary newline?

invoices/invoiceregistry.go
carlaKC: Perhaps move the group comment to the comment on the struct then? It sticks out a bit relative to our regular comment style.

invoices/invoiceregistry.go
carlaKC: Could we go straight to an enum for `result`? I'd have to string match or replace with an enum as is. 
I'll also need `NotifyExitHopHtlc` to return the outcome of the update, but that's less relevant for this PR.

invoices/invoiceregistry.go
joostjager: Reconsidering this again. Hold off re-review

joostjager:
Ok, I need to restructure this PR quite heavily. Will let know when ready again for review

invoices/invoiceregistry.go
joostjager: Changed this part to a `invoiceUpdateCtx` and pass back the outcome as return values.

invoices/invoiceregistry.go
joostjager: Enum created

invoices/invoiceregistry.go
joostjager: I came around to this approach. Updated.

joostjager:
@carlaKC @halseth this pr is ready again for re-review

invoices/invoiceregistry.go
```diff
@@ -1043,3 +972,147 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+			return nil, resultReplayToCanceled
+
+		case channeldb.HtlcStateAccepted:
+			return nil, resultReplayToAccepted
+
+		case channeldb.HtlcStateSettled:
+			return nil, resultReplayToSettled
+		}
+
+		panic("unexpected htlc state")
```
carlaKC: Why is this being upgraded from an error to a panic?
More for my own understanding of when we panic vs error in lnd.

channeldb/invoices.go
```diff
@@ -1332,7 +1332,112 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+	case ContractAccepted:
+		if invoice.State != ContractOpen {
+			return errors.New(
+				"can only accept open invoices")
+		}
+
+	case ContractSettled:
+		// Validate preimage.
+		if update.Preimage.Hash() != hash {
+			return errors.New("preimage does not match")
```
carlaKC: Define these errors as done for `ErrInvoiceAlreadyAccepted` and update tests to check specific error while you're here?

invoices/invoiceregistry.go
joostjager: It is a good question. I am not too sure myself either. If you for example do an index out of bounds, there will be a panic. This is to me kind of a similar situation. But if we want to be more defensive, we can return an error.

invoices/invoiceregistry.go
```diff
@@ -1043,3 +972,147 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+// update to be carried out.
+type invoiceUpdateCtx struct {
+	circuitKey           channeldb.CircuitKey
+	amtPaid              lnwire.MilliSatoshi
+	expiry               uint32
+	currentHeight        int32
+	finalCltvRejectDelta int32
+}
+
+// update is a callback for DB.UpdateInvoice that contains the invoice
```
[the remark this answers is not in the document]

channeldb/invoices.go
joostjager: Errors defined. But where did you see these errors being checked on?

invoices/invoiceregistry.go
carlaKC: What I mean is, why is it changed from an error to a panic in this PR? 
Because in master it just returns an [error](https://github.com/lightningnetwork/lnd/blob/cb2b8b451326d79808697af0d7f1d8d56e97dc07/invoices/invoiceregistry.go#L462)? 

Unless this branch is based on something other than master? Or I'm comparing this to the wrong code?

channeldb/invoices.go
joostjager: @cfromknecht made the same comment. Looking at it again, this particular thing isn't logged at the call site. Restored error msg.

invoices/invoiceregistry.go
joostjager: There was no reason for the `error` return value anymore except for this. Then the question is, is it worth the extra code?

channeldb/invoices.go
joostjager: This is also a more fundamental issue. When to define an error var and when just return it in-line? In other places in this file, we still create in-line errors. Some of which also have parameters. That can all be structured too, but without that specific error being matched, it may be just boiler plate?

invoices/invoiceregistry.go
joostjager: Brought back the `error` and leaving this discussion for another time.

channeldb/invoice_test.go
```diff
@@ -686,7 +686,7 @@ func getUpdateInvoice(amt lnwire.MilliSatoshi) InvoiceUpdateCallback {
 		update := &InvoiceUpdateDesc{
 			Preimage: invoice.Terms.PaymentPreimage,
 			State:    ContractSettled,
-			Htlcs: map[CircuitKey]*HtlcAcceptDesc{
+			AddHtlcs: map[CircuitKey]*HtlcAcceptDesc{
```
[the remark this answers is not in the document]
joostjager: I don't think so. Test added

invoices/invoiceregistry_test.go
joostjager: hmm, removed again. wondering if there is something weird in my rebase aliases

invoices/invoiceregistry.go
```diff
@@ -439,111 +436,46 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+			*channeldb.InvoiceUpdateDesc, error) {
+
+			var (
+				updateDesc *channeldb.InvoiceUpdateDesc
+				err        error
+			)
+			updateDesc, result, err = updateInvoice(&updateCtx, inv)
+
+			// Only send an update if the invoice state was changed.
+			updateSubscribers = err == nil && updateDesc != nil &&
```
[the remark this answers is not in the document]
joostjager: Yes, I had the same thought.

invoices/invoiceregistry.go
```diff
@@ -1043,3 +975,152 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+	// If the invoice is already canceled, there is no further checking to
+	// do.
+	if inv.State == channeldb.ContractCanceled {
+		return nil, resultInvoiceAlreadyCanceled, nil
+	}
+
+	// If an invoice amount is specified, check that enough is paid. Also
+	// check this for duplicate payments if the invoice is already settled
+	// or accepted.
+	if inv.Terms.Value > 0 && ctx.amtPaid < inv.Terms.Value {
```
cfromknecht: similar to my comment on mpp, i'm not sure checking `inv.Terms.Value > 0` is necessary. i know this is preexisting but thought i'd bring it up now

channeldb/invoices.go
```diff
@@ -1332,7 +1345,119 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+	if update.NewState == ContractOpen {
+		return ErrInvoiceCannotOpen
+	}
+
+	switch invoice.State {
+
+	// Once a contract is accepted, we can only transition to settled or
+	// canceled. Forbid transitioning back into this state, otherwise this
+	// state is identical to ContractOpen, so we fallthrough to apply the
+	// same checks for accepted invoices.
```
cfromknecht: i know i typed this, but "accepted" should be "open" invoices 😂

invoices/invoiceregistry.go
joostjager: Fixed

channeldb/invoices.go
joostjager: fixed

channeldb/invoices.go
```diff
@@ -1241,79 +1260,79 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
-
-			continue
 		}
+	}
 
-		// Add new htlc paying to the invoice.
-		if ok {
-			return nil, fmt.Errorf("htlc %v already exists", key)
+	// Process add actions from update descriptor.
+	for key, htlcUpdate := range update.AddHtlcs {
```
halseth: what's the rationale behind doing the invoice state update before adding/canceling HTLCs? (which is a change of order)

channeldb/invoices.go
joostjager: Yes, it is a bit more strict like this. Also a matter of personal pref.

channeldb/invoices.go
joostjager: Another check added :)

invoices/invoiceregistry.go
```diff
@@ -1043,3 +975,152 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
...
+		update.State = channeldb.ContractAccepted
+		return &update, resultDuplicateToAccepted, nil
+
+	case channeldb.ContractSettled:
+		update.State = channeldb.ContractSettled
+		return &update, resultDuplicateToSettled, nil
+	}
+
+	holdInvoice := inv.Terms.PaymentPreimage == channeldb.UnknownPreimage
+	if holdInvoice {
```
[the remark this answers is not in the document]
joostjager: indeed. added again

channeldb/invoices.go
```diff
@@ -1241,79 +1260,79 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
 		return &invoice, err
 	}
 
-	// Update invoice state.
-	invoice.State = update.State
+	// If there is nothing to update, return early.
+	if update == nil {
```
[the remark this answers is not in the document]
joostjager: Everywhere we return `nil` in `InvoiceRegistry.updateInvoice`. For example an htlc that has the wrong amount. We execute the update logic and conclude we don't need to update the invoice.

invoices/invoiceregistry.go
```diff
@@ -439,111 +436,46 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+
+			return updateDesc, err
+		},
+	)
+	if err != nil {
 		debugLog(err.Error())
 
 		return nil, err
 	}
+	debugLog(result.String())
```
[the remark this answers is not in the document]
joostjager: Not really, because an error may also occur after the callback, when writing back the invoice.

channeldb/invoices.go
joostjager: This isn't changing is this commit, or is it? It only pulls apart the add and cancel actions?

joostjager: Or maybe you mean the next commit? But there we still update the state first, as we do in master?

channeldb/invoices.go
```diff
@@ -1248,78 +1265,74 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
...
+			continue
 		}
 
-		err := setSettleFields(settleIndex, invoiceNum, &invoice, now)
+		err := updateHtlc(now, htlc, invoice.State)
 		if err != nil {
 			return nil, err
 		}
+
+		if invoice.State != ContractOpen &&
```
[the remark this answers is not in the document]
joostjager: Comments added

invoices/invoiceregistry.go
```diff
@@ -587,39 +589,13 @@ func (i *InvoiceRegistry) CancelInvoice(payHash lntypes.Hash) error {
...
-
-			canceledHtlcs[key] = struct{}{}
-		}
-
-		// Move invoice to the canceled state.
+		// Move invoice to the canceled state. Rely on validation in
+		// channeldb to return an error if the invoice is already
+		// settled or canceled.
 		return &channeldb.InvoiceUpdateDesc{
-			CancelHtlcs: canceledHtlcs,
```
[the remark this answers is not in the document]
joostjager: `channeldb.updateHtlc` does it when aligning the htlc state with the invoice state

invoices/update.go
```diff
@@ -107,8 +107,9 @@ func updateInvoice(ctx *invoiceUpdateCtx, inv *channeldb.Invoice) (
 
 	// If an invoice amount is specified, check that enough is paid. Also
 	// check this for duplicate payments if the invoice is already settled
-	// or accepted.
-	if inv.Terms.Value > 0 && ctx.amtPaid < inv.Terms.Value {
+	// or accepted. In case this is a zero-valued invoice, it will always be
+	// enough.
+	if ctx.amtPaid < inv.Terms.Value {
```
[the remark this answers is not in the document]
joostjager: an htlc cannot have amount 0

invoices/invoiceregistry.go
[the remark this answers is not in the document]
halseth: Ok, I see. We make the implicit assumption that canceling an invoice will also cancel all the HTLCs.

So `CancelHtlcs` is not really being used anymore?

invoices/invoiceregistry.go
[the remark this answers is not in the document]
joostjager: Yes, previously we'd log the update result even if the invoice update failed. So strictly speaking it is a change indeed.

channeldb/invoices.go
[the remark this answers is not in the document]
joostjager: It is documented on the field description of `InvoiceUpdateDesc`

invoices/invoiceregistry.go
```diff
@@ -977,154 +977,3 @@ func (i *InvoiceRegistry) HodlUnsubscribeAll(subscriber chan<- interface{}) {
 
 	delete(i.hodlReverseSubscriptions, subscriber)
 }
-
```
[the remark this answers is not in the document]
joostjager: the idea is that this would make git smarter with rebases, but not sure if it actually is smarter.

channeldb/invoices.go
joostjager: Yes, the rationale is that we don't need to loop through the htlcs twice. So the invoice state is leading and that is set first. Then we go through the htlcs and update their states based on the invoice state and what we got from the update descriptor

invoices/invoiceregistry.go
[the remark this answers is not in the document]
joostjager: It is used to cancel individual htlcs in the mpp follow up.

channeldb/invoices.go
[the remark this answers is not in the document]
halseth: That is for the `State` field, not the `InvoiceStateUpdateDesc` itself, which is what we check here. I think it would make sense for the callback function documentation to note this.

PR #3773, opened 2019-11-28
[anchors] Pluggable commitments

halseth:
As a preparation for the anchor output commitment type, we abstract the current (tweak/tweakless) commitment types behind an interface to make it easier to work with in the channel state machine.


halseth:
Replaced by #3829

PR #3775, opened 2019-11-29
[mobile] make Ready signal for custom listeners

### LINKED ISSUE #3614: [mobile] Callback when grpc server is ready
Currently the callback from starting lnd on mobile is called immediately. Instead it should wait for lnd to be ready to receive commands.

halseth:
This allows the caller to know when lnd is ready to accept RPC calls,
which is important for mobile applications where everything happens in
process.

Fixes #3614

### DIFF
--- a/lnd.go
+++ b/lnd.go
@@ -59,16 +59,25 @@ var (
 	networkDir string
 )
 
+// ListnerWithSignal is a net.Listner that has an additional Ready channel that
+// will be closed when a server starts listening.
+type ListenerWithSignal struct {
+	net.Listener
+
+	// Ready will be closed by the server listening on Listener.
+	Ready chan struct{}
+}
+
 // ListenerCfg is a wrapper around custom listeners that can be passed to lnd
 // when calling its main method.
 type ListenerCfg struct {
 	// WalletUnlocker can be set to the listener to use for the wallet
 	// unlocker. If nil a regular network listener will be created.
-	WalletUnlocker net.Listener
+	WalletUnlocker *ListenerWithSignal
 
 	// RPCListener can be set to the listener to use for the RPC server. If
 	// nil a regular network listener will be created.
-	RPCListener net.Listener
+	RPCListener *ListenerWithSignal
 }
 
 // rpcListeners is a function type used for closures that fetches a set of RPC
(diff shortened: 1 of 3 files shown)


mobile/bindings.go
```diff
@@ -53,9 +70,15 @@ func Start(extraArgs string, callback Callback) {
...
 	}()
 
-	// TODO(halseth): callback when RPC server is actually running. Since
-	// the RPC server might take a while to start up, the client might
-	// assume it is ready to accept calls when this callback is sent, while
-	// it's not.
-	callback.OnResponse([]byte("started"))
+	// Finally we start two go routines that will call the provided
+	// callbacks when the RPC servers are ready to accept calls.
+	go func() {
```
bhandras: Maybe a simpler approach could be to modify `lnd.Main()` by adding a `started` channel as an incoming argument which then could be used to signal that all initialization is finished and we're waiting on	`<-signal.ShutdownChannel()`. WDYT?

lnd.go
```diff
@@ -59,24 +59,34 @@ var (
 	networkDir string
 )
 
+// ListnerWithSignal is a net.Listner that has an additional Ready channel that
+// will be closed when a server starts listening.
+type ListenerWithSignal struct {
+	net.Listener
+
+	// Ready will be closed by the server listening on Listener.
+	Ready chan struct{}
```
guggero: Could we make this `<-chan struct{}` to make sure we only ever read from it?

mobile/bindings.go
guggero: Because the unlocker is conditional (it's only started if the wallet is actually locked) you need to know what the state of the individual servers is to be able to react. So there would need to be two `started` channels which basically comes down to the implemented solution.
I think this is probably the most simple you can get with the current architecture.

bhandras: I see, thanks for the clarificaton @guggero .

lnd.go
halseth: We can't really, since we are reading from it one place, and closing it the other place.

mobile/bindings.go
halseth: 👍

PR #3777, opened 2019-11-29
signrpc: add DeriveSharedKey

guggero:
Allow ECDH based shared key derivation.

This would allow multiple use cases where a sender wants to encrypt a message for a target node without prior key exchange.


lnrpc/walletrpc/walletkit_server.go
```diff
@@ -542,3 +548,57 @@ func (w *WalletKit) BumpFee(ctx context.Context,
...
+
+// ecdh performs an ECDH operation between pub and priv. The returned value is
+// the sha256 of the compressed shared point.
+func ecdh(pub *btcec.PublicKey, priv *btcec.PrivateKey) []byte {
+	s := &btcec.PublicKey{}
+	x, y := btcec.S256().ScalarMult(pub.X, pub.Y, priv.D.Bytes())
+	s.X = x
+	s.Y = y
+
+	h := sha256.Sum256(s.SerializeCompressed())
```
joostjager: Would the API be more flexible if we would return `s` so that the caller can also do something else with it than sha256?

joostjager: https://en.wikipedia.org/wiki/Elliptic-curve_Diffie%E2%80%93Hellman

Seems a bad idea

lnrpc/walletrpc/walletkit.proto
```diff
@@ -220,46 +220,54 @@ message BumpFeeRequest {
 message BumpFeeResponse {
 }
 
+message SharedKeyRequest {
+    bytes ephemeral_pubkey = 1;
```
joostjager: Does it need to be ephemeral? I think it could be any pubkey? We wouldn't be able to verify that it is ephemeral anyway.

lnrpc/walletrpc/walletkit.proto
```diff
@@ -220,46 +220,54 @@ message BumpFeeRequest {
...
+}
+
+message SharedKeyResponse {
+    bytes shared_key = 1;
```
joostjager: Comments on proto fields?

lnrpc/walletrpc/walletkit.proto
```diff
@@ -220,46 +220,54 @@ message BumpFeeRequest {
...
+}
+
 service WalletKit {
     /**
     DeriveNextKey attempts to derive the *next* key within the key family
     (account in BIP43) specified. This method should return the next external
     child within this branch.
     */
-    rpc DeriveNextKey(KeyReq) returns (signrpc.KeyDescriptor);
+    rpc DeriveNextKey (KeyReq) returns (signrpc.KeyDescriptor);
```
joostjager: Do those spaces come from some kind of proto cleaner tool?

lnrpc/walletrpc/walletkit.proto
```diff
@@ -220,46 +220,54 @@ message BumpFeeRequest {
 message BumpFeeResponse {
 }
 
+message SharedKeyRequest {
+    bytes ephemeral_pubkey = 1;
+}
```
joostjager: Could it be interesting to supply a key deriviation path for the node key, similar to `SignMessage`? Maybe not, because nobody on the network knows anything other than our usual pubkey.

lnrpc/walletrpc/walletkit.proto
```diff
@@ -300,5 +308,53 @@ service WalletKit {
...
+    rpc BumpFee (BumpFeeRequest) returns (BumpFeeResponse);
+
+    /*
+    DeriveSharedKey returns a shared secret key by performing Diffie-Hellman key
+    derivation between the ephemeral public key in the request and the node's
+    identity public key. The resulting shared public key is hashed with sha256
+    and the final key length therefore is 256bit.
+    
+    If any sender wants to encrypt a message that only the receiving lnd node
+    can decrypt, the Diffie-Hellman key exchange is very useful. As the
```
joostjager: State this is just one use case of DH

lnrpc/walletrpc/walletkit.proto
```diff
@@ -300,5 +308,53 @@ service WalletKit {
...
+          P_shared = d_recipient * P_send
+          P_shared = d_recipient * (d_send * G)
+          P_shared = d_send * (d_recipient * G)
+          P_shared = d_send * P_recipient
+
+      2. Serialize the shared public key P_shared in the compressed format and
+         sha256 hash it to arrive at the shared secret key.
+      3. Use the derived key to decrypt the message using a symmetrical
+         decryption algorithm.
+    */
```
joostjager: I don't know if we should include this whole explanation of DH. I think it is, similar to `SignMessage` upon the caller to understand what they should do.

lnrpc/walletrpc/walletkit_server.go
```diff
@@ -88,6 +90,10 @@ var (
 			Entity: "onchain",
 			Action: "write",
 		}},
+		"/walletrpc.WalletKit/DeriveSharedKey": {{
+			Entity: "address",
```
joostjager: Or is this more `signer` permission? Maybe that is also the sub-server it needs to be in?

lnrpc/walletrpc/walletkit.proto
guggero: This is more of a usage recommendation to signal that ephemeral keys _should_ be used.

lnrpc/walletrpc/walletkit.proto
guggero: Yes, the formatter of my IDE has support for proto files. But we probably want this in a separate PR, most of the proto files have inconsistent formatting.

lnrpc/walletrpc/walletkit.proto
joostjager: Doesn't that depend on the use case? The other key isn't ephemeral either, while there are applications where both keys are ephemeral.

lnrpc/walletrpc/walletkit.proto
joostjager: I don't mind

lnrpc/walletrpc/walletkit.proto
guggero: I don't see any use case for this at the moment.

lnrpc/walletrpc/walletkit_server.go
guggero: Yes, definitely. Now that we have the secret key ring in the signer it makes much more sense to have this functionality there.

PR #3781, opened 2019-12-02
[htlcnotifier 3/4]: Add HTLC Notifier

carlaKC:
This PR is change 3 of 4 proposed to add a htlc event notifier to lnd.
1: #3843
2: #3844
3: This PR
4: #3848 

Since the PR is dependent on the preceding PRs, it has been rebased on #3844.
**Only the last 9 commits are eligible for review.**

--------------

This PR adds a `htlcnotifier` to the `htlcswitch` which provides a stream of notifications with the following types of events:
- Forwards: when we forward a HTLC or initiate a send from our node
- Forward Failures: when a HTLC we forwarded fails down the line
- Settles: when a HTLC we forwarded is settled, or we receive a payment
- Link Failures: when we fail a HTLC on our incoming or outgoing link

### DIFF
--- a/htlcswitch/htlcnotifier.go
+++ b/htlcswitch/htlcnotifier.go
@@ -289,6 +289,8 @@ type SettleEvent struct {
 
 // NotifyForwardingEvent notifies the HtlcNotifier than a htlc has been
 // forwarded.
+//
+// Note this is part of the htlcNotifier interface.
 func (h *HtlcNotifier) NotifyForwardingEvent(key HtlcKey, info HtlcInfo,
 	eventType HtlcEventType) {
 
@@ -309,6 +311,8 @@ func (h *HtlcNotifier) NotifyForwardingEvent(key HtlcKey, info HtlcInfo,
 
 // NotifyLinkFailEvent notifies that a htlc has failed on our incoming
 // or outgoing link.
+//
+// Note this is part of the htlcNotifier interface.
 func (h *HtlcNotifier) NotifyLinkFailEvent(key HtlcKey, info HtlcInfo,
 	eventType HtlcEventType, linkErr *LinkError, incoming bool) {
 
@@ -331,6 +335,8 @@ func (h *HtlcNotifier) NotifyLinkFailEvent(key HtlcKey, info HtlcInfo,
 
 // NotifyForwardingFailEvent notifies the HtlcNotifier that a htlc we
 // forwarded has failed down the line.
+//
+// Note this is part of the htlcNotifier interface.
 func (h *HtlcNotifier) NotifyForwardingFailEvent(key HtlcKey,
 	eventType HtlcEventType) {
 
@@ -350,6 +356,8 @@ func (h *HtlcNotifier) NotifyForwardingFailEvent(key HtlcKey,
 
 // NotifySettleEvent notifies the HtlcNotifier that a htlc that we committed
 // to as part of a forward or a receive to our node has been settled.
+//
+// Note this is part of the htlcNotifier interface.
 func (h *HtlcNotifier) NotifySettleEvent(key HtlcKey, eventType HtlcEventType) {
 	event := &SettleEvent{
 		HtlcKey:       key,
--- a/htlcswitch/htlcnotifier.go
+++ b/htlcswitch/htlcnotifier.go
@@ -371,3 +371,59 @@ func (h *HtlcNotifier) NotifySettleEvent(key HtlcKey, eventType HtlcEventType) {
 		log.Warnf("Unable to send settle event: %v", err)
 	}
 }
+
+// newHtlc key returns a htlc key for the packet provided. If the packet
+// has a zero incoming channel ID, the packet is for one of our own sends,
+// which has the payment id stashed in the incoming htlc id. If this is the
+// case, we replace the incoming htlc id with zero so that the notifier
+// consistently reports zero circuit keys for events that terminate or
+// originate at our node.
+func newHtlcKey(pkt *htlcPacket) HtlcKey {
+	htlcKey := HtlcKey{
+		IncomingCircuit: channeldb.CircuitKey{
+			ChanID: pkt.incomingChanID,
+			HtlcID: pkt.incomingHTLCID,
+		},
+		OutgoingCircuit: CircuitKey{
+			ChanID: pkt.outgoingChanID,
+			HtlcID: pkt.outgoingHTLCID,
+		},
+	}
+
+	// If the packet has a zero incoming channel ID, it is a send that was
+	// initiated at our node. If this is the case, our internal pid is in
+	// the incoming htlc ID, so we overwrite it with 0 for notification
+	// purposes.
+	if pkt.incomingChanID == hop.Source {
+		htlcKey.IncomingCircuit.HtlcID = 0
+	}
+
+	return htlcKey
+}
+
+// newHtlcInfo returns HtlcInfo for the packet provided.
+func newHtlcInfo(pkt *htlcPacket) HtlcInfo {
+	return HtlcInfo{
+		IncomingTimeLock: pkt.incomingTimeout,
+		OutgoingTimeLock: pkt.outgoingTimeout,
+		IncomingAmt:      pkt.incomingAmount,
+		OutgoingAmt:      pkt.amount,
+	}
+}
+
+// getEventType returns the htlc type based on the fields set in the htlc
+// packet. Sends that originate at our node have the source (zero) incoming
+// channel ID. Receives to our node have the exit (zero) outgoing channel ID
+// and forwards have both fields set.
+func getEventType(pkt *htlcPacket) HtlcEventType {
+	switch {
+	case pkt.incomingChanID == hop.Source:
+		return HtlcEventTypeSend
+
+	case pkt.outgoingChanID == hop.Exit:
+		return HtlcEventTypeReceive
+
+	default:
+		return HtlcEventTypeForward
+	}
+}
--- a/htlcswitch/htlcnotifier.go
+++ b/htlcswitch/htlcnotifier.go
@@ -1,0 +1,365 @@
+package htlcswitch
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/htlcswitch/hop"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// HtlcNotifier notifies clients of htlc forwards, failures and settles for
+// htlcs that the switch handles. It takes subscriptions for its events and
+// notifies them when htlc events occur. These are served on a best-effort
+// basis; events are not persisted, delivery is not guaranteed (in the event
+// of a crash in the switch, forward events may be lost) and some events may
+// be replayed upon restart. Events consumed from this package should be
+// de-duplicated by the htlc's unique combination of incoming+outgoing circuit
+// and not relied upon for critical operations.
+//
+// The htlc notifier sends the following kinds of events:
+// Forwarding Event:
+// - Represents a htlc which is forwarded onward from our node.
+// - Present for htlc forwards through our node and local sends.
+//
+// Link Failure Event:
+// - Indicates that a htlc has failed on our incoming or outgoing link,
+//   with an incoming boolean which indicates where the failure occurred.
+// - Incoming link failures are present for failed attempts to pay one of
+//   our invoices (insufficient amount or mpp timeout, for example) and for
+//   forwards that we cannot decode to forward onwards.
+// - Outgoing link failures are present for forwards or local payments that
+//   do not meet our outgoing link's policy (insufficient fees, for example)
+//   and when we fail to forward the payment on (insufficient outgoing
+//   capacity, or an unknown outgoing link).
+//
+// Forwarding Failure Event:
+// - Forwarding failures indicate that a htlc we forwarded has failed at
+//   another node down the route.
+// - Present for local sends and htlc forwards which fail after they left
+//   our node.
+//
+// Settle event:
+// - Settle events are present when a htlc which we added is settled through
+//   the release of a preimage.
+// - Present for local receives, and successful local sends or forwards.
+//
+// Each htlc is identified by its incoming and outgoing circuit key. Htlcs,
+// and their subsequent settles or fails, can be identified by the combination
+// of incoming and outgoing circuits. Note that receives to our node will
+// have a zero outgoing circuit key because the htlc terminates at our
+// node, and sends from our node will have a zero incoming circuit key because
+// the send originates at our node.
+type HtlcNotifier struct {
+	started sync.Once
+	stopped sync.Once
+
+	// now returns the current time, it is set in the htlcnotifier to allow
+	// for timestamp mocking in tests.
+	now func() time.Time
+
+	ntfnServer *subscribe.Server
+}
+
+// NewHtlcNotifier creates a new HtlcNotifier which gets htlc forwarded,
+// failed and settled events from links our node has established with peers
+// and sends notifications to subscribing clients.
+func NewHtlcNotifier(now func() time.Time) *HtlcNotifier {
+	return &HtlcNotifier{
+		now:        now,
+		ntfnServer: subscribe.NewServer(),
+	}
+}
+
+// Start starts the HtlcNotifier and all goroutines it needs to consume
+// events and provide subscriptions to clients.
+func (h *HtlcNotifier) Start() error {
+	var err error
+	h.started.Do(func() {
+		log.Trace("HtlcNotifier starting")
+		err = h.ntfnServer.Start()
+	})
+	return err
+}
+
+// Stop signals the notifier for a graceful shutdown.
+func (h *HtlcNotifier) Stop() {
+	h.stopped.Do(func() {
+		if err := h.ntfnServer.Stop(); err != nil {
+			log.Warnf("error stopping htlc notifier: %v", err)
+		}
+	})
+}
+
+// SubscribeHtlcEvents returns a subscribe.Client that will receive updates
+// any time the server is made aware of a new event.
+func (h *HtlcNotifier) SubscribeHtlcEvents() (*subscribe.Client, error) {
+	return h.ntfnServer.Subscribe()
+}
+
+// HtlcKey uniquely identifies the htlc.
+type HtlcKey struct {
+	// IncomingCircuit is the channel an htlc id of the incoming htlc.
+	IncomingCircuit channeldb.CircuitKey
+
+	// OutgoingCircuit is the channel and htlc id of the outgoing htlc.
+	OutgoingCircuit channeldb.CircuitKey
+}
+
+// String returns a string representation of a htlc key.
+func (k HtlcKey) String() string {
+	switch {
+	case k.IncomingCircuit.ChanID == hop.Source:
+		return k.OutgoingCircuit.String()
+
+	case k.OutgoingCircuit.ChanID == hop.Exit:
+		return k.IncomingCircuit.String()
+
+	default:
+		return fmt.Sprintf("%v -> %v", k.IncomingCircuit,
+			k.OutgoingCircuit)
+	}
+}
+
+// HtlcInfo provides the details of a htlc that our node has processed. For
+// forwards, incoming and outgoing values are set, whereas sends and receives
+// will only have outgoing or incoming details set.
+type HtlcInfo struct {
+	// IncomingTimelock is the time lock of the htlc on our incoming
+	// channel.
+	IncomingTimeLock uint32
+
+	// OutgoingTimelock is the time lock the htlc on our outgoing channel.
+	OutgoingTimeLock uint32
+
+	// IncomingAmt is the amount of the htlc on our incoming channel.
+	IncomingAmt lnwire.MilliSatoshi
+
+	// OutgoingAmt is the amount of the htlc on our outgoing channel.
+	OutgoingAmt lnwire.MilliSatoshi
+}
+
+// String returns a string representation of a htlc.
+func (h HtlcInfo) String() string {
+	var details []string
+
+	// If the incoming information is not zero, as is the case for a send,
+	// we include the incoming amount and timelock.
+	if h.IncomingAmt != 0 || h.IncomingTimeLock != 0 {
+		str := fmt.Sprintf("incoming amount: %v, "+
+			"incoming timelock: %v", h.IncomingAmt,
+			h.IncomingTimeLock)
+
+		details = append(details, str)
+	}
+
+	// If the outgoing information is not zero, as is the case for a
+	// receive, we include the outgoing amount and timelock.
+	if h.OutgoingAmt != 0 || h.OutgoingTimeLock != 0 {
+		str := fmt.Sprintf("outgoing amount: %v, "+
+			"outgoing timelock: %v", h.OutgoingAmt,
+			h.OutgoingTimeLock)
+
+		details = append(details, str)
+	}
+
+	return strings.Join(details, ", ")
+}
+
+// HtlcEventType represents the type of event that a htlc was part of.
+type HtlcEventType int
+
+const (
+	// HtlcEventTypeSend represents a htlc that was part of a send from
+	// our node.
+	HtlcEventTypeSend HtlcEventType = iota
+
+	// HtlcEventTypeReceive represents a htlc that was part of a receive
+	// to our node.
+	HtlcEventTypeReceive
+
+	// HtlcEventTypeForward represents a htlc that was forwarded through
+	// our node.
+	HtlcEventTypeForward
+)
+
+// String returns a string representation of a htlc event type.
+func (h HtlcEventType) String() string {
+	switch h {
+	case HtlcEventTypeSend:
+		return "send"
+
+	case HtlcEventTypeReceive:
+		return "receive"
+
+	case HtlcEventTypeForward:
+		return "forward"
+
+	default:
+		return "unknown"
+	}
+}
+
+// ForwardingEvent represents a htlc that was forwarded onwards from our node.
+// Sends which originate from our node will report forward events with zero
+// incoming circuits in their htlc key.
+type ForwardingEvent struct {
+	// HtlcKey uniquely identifies the htlc, and can be used to match the
+	// forwarding event with subsequent settle/fail events.
+	HtlcKey
+
+	// HtlcInfo contains details about the htlc.
+	HtlcInfo
+
+	// HtlcEventType classifies the event as part of a local send or
+	// receive, or as part of a forward.
+	HtlcEventType
+
+	// Timestamp is the time when this htlc was forwarded.
+	Timestamp time.Time
+}
+
+// LinkFailEvent describes a htlc that failed on our incoming or outgoing
+// link. The incoming bool is true for failures on incoming links, and false
+// for failures on outgoing links. The failure reason is provided by a lnwire
+// failure message which is enriched with a failure detail in the cases where
+// the wire failure message does not contain full information about the
+// failure.
+type LinkFailEvent struct {
+	// HtlcKey uniquely identifies the htlc.
+	HtlcKey
+
+	// HtlcInfo contains details about the htlc.
+	HtlcInfo
+
+	// HtlcEventType classifies the event as part of a local send or
+	// receive, or as part of a forward.
+	HtlcEventType
+
+	// LinkError is the reason that we failed the htlc.
+	LinkError *LinkError
+
+	// Incoming is true if the htlc was failed on an incoming link.
+	// If it failed on the outgoing link, it is false.
+	Incoming bool
+
+	// Timestamp is the time when the link failure occurred.
+	Timestamp time.Time
+}
+
+// ForwardingFailEvent represents a htlc failure which occurred down the line
+// after we forwarded a htlc onwards. An error is not included in this event
+// because errors returned down the route are encrypted. HtlcInfo is not
+// reliably available for forwarding failures, so it is omitted. These events
+// should be matched with their corresponding forward event to obtain this
+// information.
+type ForwardingFailEvent struct {
+	// HtlcKey uniquely identifies the htlc, and can be used to match the
+	// htlc with its corresponding forwarding event.
+	HtlcKey
+
+	// HtlcEventType classifies the event as part of a local send or
+	// receive, or as part of a forward.
+	HtlcEventType
+
+	// Timestamp is the time when the forwarding failure was received.
+	Timestamp time.Time
+}
+
+// SettleEvent represents a htlc that was settled. HtlcInfo is not reliably
+// available for forwarding failures, so it is omitted. These events should
+// be matched with corresponding forward events or invoices (for receives)
+// to obtain additional information about the htlc.
+type SettleEvent struct {
+	// HtlcKey uniquely identifies the htlc, and can be used to match
+	// forwards with their corresponding forwarding event.
+	HtlcKey
+
+	// HtlcEventType classifies the event as part of a local send or
+	// receive, or as part of a forward.
+	HtlcEventType
+
+	// Timestamp is the time when this htlc was settled.
+	Timestamp time.Time
+}
+
+// NotifyForwardingEvent notifies the HtlcNotifier than a htlc has been
+// forwarded.
+func (h *HtlcNotifier) NotifyForwardingEvent(key HtlcKey, info HtlcInfo,
+	eventType HtlcEventType) {
+
+	event := &ForwardingEvent{
+		HtlcKey:       key,
+		HtlcInfo:      info,
+		HtlcEventType: eventType,
+		Timestamp:     h.now(),
+	}
+
+	log.Tracef("Notifying forward event: %v over %v, %v", eventType, key,
+		info)
+
+	if err := h.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send forwarding event: %v", err)
+	}
+}
+
+// NotifyLinkFailEvent notifies that a htlc has failed on our incoming
+// or outgoing link.
+func (h *HtlcNotifier) NotifyLinkFailEvent(key HtlcKey, info HtlcInfo,
+	eventType HtlcEventType, linkErr *LinkError, incoming bool) {
+
+	event := &LinkFailEvent{
+		HtlcKey:       key,
+		HtlcInfo:      info,
+		HtlcEventType: eventType,
+		LinkError:     linkErr,
+		Incoming:      incoming,
+		Timestamp:     h.now(),
+	}
+
+	log.Tracef("Notifying link failure event: %v over %v, %v", eventType,
+		key, info)
+
+	if err := h.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send link fail event: %v", err)
+	}
+}
+
+// NotifyForwardingFailEvent notifies the HtlcNotifier that a htlc we
+// forwarded has failed down the line.
+func (h *HtlcNotifier) NotifyForwardingFailEvent(key HtlcKey,
+	eventType HtlcEventType) {
+
+	event := &ForwardingFailEvent{
+		HtlcKey:       key,
+		HtlcEventType: eventType,
+		Timestamp:     h.now(),
+	}
+
+	log.Tracef("Notifying forwarding failure event: %v over %v", eventType,
+		key)
+
+	if err := h.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send forwarding fail event: %v", err)
+	}
+}
+
+// NotifySettleEvent notifies the HtlcNotifier that a htlc that we committed
+// to as part of a forward or a receive to our node has been settled.
+func (h *HtlcNotifier) NotifySettleEvent(key HtlcKey, eventType HtlcEventType) {
+	event := &SettleEvent{
+		HtlcKey:       key,
+		HtlcEventType: eventType,
+		Timestamp:     h.now(),
+	}
+
+	log.Tracef("Notifying settle event: %v over %v", eventType, key)
+
+	if err := h.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send settle event: %v", err)
+	}
+}
--- a/htlcswitch/interfaces.go
+++ b/htlcswitch/interfaces.go
@@ -180,3 +180,29 @@ type TowerClient interface {
 	// isTweakless should be true.
 	BackupState(*lnwire.ChannelID, *lnwallet.BreachRetribution, bool) error
 }
+
+// htlcNotifier is an interface which represents the input side of the
+// HtlcNotifier which htlc events are piped through. This interface is intended
+// to allow for mocking of the htlcNotifier in tests, so is unexported because
+// it is not needed outside of the htlcSwitch package.
+type htlcNotifier interface {
+	// NotifyForwardingEvent notifies the HtlcNotifier than a htlc has been
+	// forwarded.
+	NotifyForwardingEvent(key HtlcKey, info HtlcInfo,
+		eventType HtlcEventType)
+
+	// NotifyIncomingLinkFailEvent notifies that a htlc has failed on our
+	// incoming link. It takes an isReceive bool to differentiate between
+	// our node's receives and forwards.
+	NotifyLinkFailEvent(key HtlcKey, info HtlcInfo,
+		eventType HtlcEventType, linkErr *LinkError, incoming bool)
+
+	// NotifyForwardingFailEvent notifies the HtlcNotifier that a htlc we
+	// forwarded has failed down the line.
+	NotifyForwardingFailEvent(key HtlcKey, eventType HtlcEventType)
+
+	// NotifySettleEvent notifies the HtlcNotifier that a htlc that we
+	// committed to as part of a forward or a receive to our node has been
+	// settled.
+	NotifySettleEvent(key HtlcKey, eventType HtlcEventType)
+}
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -1224,8 +1224,7 @@ func (l *channelLink) processHtlcResolution(resolution invoices.HtlcResolution,
 		failure := getResolutionFailure(res, htlc.pd.Amount)
 
 		l.sendHTLCError(
-			htlc.pd.HtlcIndex, failure,
-			htlc.obfuscator, htlc.pd.SourceRef,
+			htlc.pd, failure, htlc.obfuscator, true,
 		)
 		return nil
 
@@ -2696,9 +2695,7 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 			// later date
 			failure := lnwire.NewInvalidOnionPayload(failedType, 0)
 			l.sendHTLCError(
-				pd.HtlcIndex,
-				NewLinkError(failure),
-				obfuscator, pd.SourceRef,
+				pd, NewLinkError(failure), obfuscator, false,
 			)
 
 			l.log.Errorf("unable to decode forwarding "+
@@ -2811,10 +2808,7 @@ func (l *channelLink) processRemoteAdds(fwdPkg *channeldb.FwdPkg,
 				)
 
 				l.sendHTLCError(
-					pd.HtlcIndex,
-					NewLinkError(failure),
-					obfuscator,
-					pd.SourceRef,
+					pd, NewLinkError(failure), obfuscator, false,
 				)
 				continue
 			}
@@ -2900,7 +2894,7 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		failure := NewLinkError(
 			lnwire.NewFinalIncorrectHtlcAmount(pd.Amount),
 		)
-		l.sendHTLCError(pd.HtlcIndex, failure, obfuscator, pd.SourceRef)
+		l.sendHTLCError(pd, failure, obfuscator, true)
 
 		return nil
 	}
@@ -2915,7 +2909,7 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		failure := NewLinkError(
 			lnwire.NewFinalIncorrectCltvExpiry(pd.Timeout),
 		)
-		l.sendHTLCError(pd.HtlcIndex, failure, obfuscator, pd.SourceRef)
+		l.sendHTLCError(pd, failure, obfuscator, true)
 
 		return nil
 	}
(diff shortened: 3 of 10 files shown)


lntest/itest/lnd_test.go
```diff
@@ -9250,13 +9470,34 @@ out:
...
+
+	// Bob forwarded the failed payment, so we expect him to have a forward
+	// event and a forwarding failure.
+	if err := testHTLCEvents(bobCl, expectedHash, nil, nil, 1); err != nil {
+		t.Fatalf("bob did not have expected HTLC events: %v", err)
+	}
+
+	// Carol was the recipient of a payment with an incorrect amount, so
+	// we expect her to have a link failure event with reason incorrect amount.
+	// TODO(carla): replace with specific error once available.
```
carlaKC: This TODO is here intentionally, as is errors for invoice payment attempts which fail due to underpayment or expiry too soon are silenced and the `processExitHop` code falls through to `processHodlEvent` which treats these failures as a hodl cancel. 

This structure is being changed in #3770, which is currently in review. Once that change is in, reasons for invoice failure can be better surfaced and this test can be updated.

htlcswitch/htlcnotifier/htlcnotifier.go
```diff
@@ -0,0 +1,221 @@
...
+// as part of a forward or a receive to our node has been settled.
+func (h *HTLCNotifier) NotifySettleEvent(event SettleEvent) {
+	log.Debugf("Notifying settle event from %v(%v) to %v(%v)",
+		event.IncomingChannel, event.HTLCInIndex, event.OutgoingChannel,
+		event.HTLCOutIndex)
+
+	if err := h.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send settle event: %v", err)
+	}
+}
```
joostjager: Question that comes up is which event (sequences) to expect in the different scenarios:
* Incoming exit hop htlc rejected
* Incoming exit hop htlc settled
* Incoming exit hop htlc accepted but not yet resolved (hodl invoice), then settled
* Incoming exit hop htlc accepted but not yet resolved (hodl invoice), then rejected
* Outgoing origin htlc (our payment) not handed out (didn't even leave the link)
* Outgoing origin htlc (our payment) failed back by peer
* Outgoing origin htlc (our payment) settled
* Incoming htlc to forward not accepted
* Incoming htlc to forward accepted, but rejected by outgoing link
* Htlc forwarded, but failed back by downstream peer
* Htlc forwarded and settled

Are you sure that we are able to cover and distinguish all we need with the four events defined above? Are in any of the scenarios above multiple events sent?

htlcswitch/link.go
```diff
@@ -2158,13 +2244,15 @@ func (l *channelLink) CheckHtlcForward(payHash [32]byte,
...
-		)
+		return &ForwardingError{
+			FailureMessage: l.createFailureWithUpdate(
+				func(upd *lnwire.ChannelUpdate) lnwire.FailureMessage {
+					return lnwire.NewFeeInsufficient(
+						amtToForward, *upd,
+					)
+				},
+			),
+		}
```
joostjager: Why wrap in `ForwardingError` if there is no FailureDetail yet assigned? Maybe we should always set `FailureDetail` with our internal failure code and decouple it completely from the wire message.

htlcswitch/htlcnotifier/htlcnotifier.go
```diff
@@ -0,0 +1,262 @@
...
+// for HTLCs that our node handles. HTLCs which originated at our node can be
+// identified by a zero channel ID, on the incoming channel for sends and on the
+// outgoing channel for receives.
+//
+// It takes subscriptions for its events and notifies them when HTLC events
+// occur. These are served on a best-effort basis; events are not persisted,
+// delivery is not guaranteed (in the event of a crash in the switch, forward
+// events may be lost) and events will be replayed upon restart. Events consumed
+// from this package should be de-duplicated, and not relied upon for critical
+// operations.
```
joostjager: From a greater perspective, I think it is still unsatisfactory that we are not building a reliable log of all decisions that we made. We are saving huge amounts of historical channel data in the database (including the raw onion payloads), but it is just not enough and/or not in the right format to extract what we want from it.

I believe that eventually there will be no way around creating a airtight audit trail.

carlaKC:
Running through some different scenarios, as suggested by @joostjager, to make sure we can distinguish between different types of events.

✅ = Value set ; ❌ = Value not set

**Forward Events: Ok**

| Forward Event | Incoming Key | Outgoing Key | 
| ---------------|---------------|-------------- |
| Send | ❌ | ✅ | 
| Forward | ✅ | ✅ |

**Forward Failure: Ok**

| Forward Event | Incoming Key | Outgoing Key | 
| ---------------|---------------|-------------- |
| Send | ✅ | ❌ |
| Forward | ✅ | ✅ |

**Settle Event: Potential overlap**

| Settle Event | Incoming Key | Outgoing Key | 
| ---------------|---------------|-------------- |
| Own Receive |  ✅ | ❌ |
| Own Send |  ✅ | ❌ |
| Forward | ✅ | ✅ |

In the ideal case, we anything with a zero outgoing key belonging to us then check for matching forward events to determine whether the settle is for a send or receive. However, we will not have a forward event for sends which were forwarded before restart and settled afterwards.

**Link Failures: Overlap** 

| Link Failure| Incoming Key | Outgoing Key | IncomingFailed | Detail |
|------------ | ------------ | ------------ | ------------ | ------------ |
| Receive fails| ✅ | ❌ | true | Invoice specific |
| HODL Cancel| ✅ | ❌ | true | HODL Cancel |
| Send fails | ❌ | ✅ | false | failure specific |
| Fwd fails incoming | ✅ | ❌ | true | failure specific |
| Fwd Fails outgoing | ✅ | ✅ | false | failure specific |

For this set of events, there is an overlap issue between forwarding failures which occur on the incoming link and failures paying our own invoices. Receives have a specific set of failure details, which we could use to distinguish, but that is not forward thinking because we may have a failure which is common for receives and forwards.

**Solution:**
The overlap issues here are mostly due to the inclusion of local sends/receives. A naive solution would be to match on payment hashes from `listinvoices`/`payments`, but that is clunky and not suitable in the case of reused payment hashes. Given that we aren't guaranteed to have forward events for local sends (which differentiate them from receives), I would say that a type field with `forward`, `send` and `receive` would be the simplest way to go? This will solve all the overlap issues that the current event structs have.

htlcswitch/link.go
carlaKC: `FailureDetail` is intended to add further granularity to `lnwire.FailureMessage`, so will have no value when there is no further information to add. The first run of this PR had a big enum for all the failure reasons, but it ended up having a lot of duplication with the existing wire message. 

Do we have need for an internal error code beyond this use case? We're not tied to the numerical value of this enum since we map the (message:detail) to a different rpc enum, so we can always change this to have an internal failure code when we need it?

htlcswitch/switch.go
```diff
@@ -485,18 +485,19 @@ func (s *Switch) forward(packet *htlcPacket) error {
 				return err
 			}
 
-			var failure lnwire.FailureMessage
+			failure := &ForwardingError{
+				ExtraMsg: ErrIncompleteForward.Error(),
```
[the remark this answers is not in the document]
carlaKC: We don't use this error message for much except tests and logging. Could just add a string to `FailureDetail` which covers logging?

We do have some tests where we check that they have a specific amount/height specified, but there's probably a way around that.

htlcswitch/htlcnotifier/htlcnotifier.go
carlaKC: I don't think we are tying ourselves to anything that can't be fleshed out with a more reliable audit trail at a later stage?

I think that this best effort approach will provide node operators with value add, specifically if we add it to lndmon and get it as visible as possible. This is also a good opportunity for us to get some user feedback on the feature (specifically forward failures). Maybe people only care about actionable failures like `InsufficientBalance` (because they can potentially rebalance/loop), maybe they don't care about local events, hard to say till we get something usable out.

joostjager: Yes, I see that. I guess I am missing a bit the grand master plan of where we eventually want all of this to end up, but this PR could also be a good way to get a feel for requirements in that area.

htlcswitch/link.go
joostjager: I think an internal error code could be useful beyond this use case. Also in the interpretation of failures, there is some sketchiness around failures that we generated ourselves. We use the wire message (even though the htlc never left our node) to convey an internal error.

I see the problem with duplication though. Also not sure if there is a good solution other than just adding the optional `FailureDetail` as you do now. Maybe only using internal errors and mapping them last minute to wire messages could work long-term, but I think you tried something like that already?

htlcswitch/switch.go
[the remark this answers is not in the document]
joostjager: Instead of adding a string to `FailureDetail`, can't `FailureDetail` contain the failure metadata in a structured way?

joostjager:
With regards to the event sequences, would it make sense to have only a single event per htlc as it is handled by our system? So a forward will generate just a single even, even though there are an incoming and an outgoing link involved. Or is that what you suggest with the send/receive/fwd event type?

carlaKC:
> With regards to the event sequences, would it make sense to have only a single event per htlc as it is handled by our system? So a forward will generate just a single even, even though there are an incoming and an outgoing link involved. 

As is we have a single event for a forward, which has the information for the incoming and outgoing links in `HTLCInfo`. Running through scenarios is a great idea, this is how events will be generated for the scenarios you've provided:

**Incoming exit hop htlc rejected**
Link failure {Incoming link key, rejection reason, incomingLink = true}

**Incoming exit hop htlc settled**
Settle {Incoming link key, payment hash}

**Incoming exit hop htlc accepted but not yet resolved (hodl invoice), then settled**
No event on acceptance
Settle {Incoming link key, payment hash}

**Incoming exit hop htlc accepted but not yet resolved (hodl invoice), then rejected**
No event on acceptance
Link failure {Incoming link key, fail detail = hodlCancel, incomingLink = true}
 
**Outgoing origin htlc (our payment) not handed out (didn't even leave the link)**
Link failure {Incoming & outgoing link key, failure detail, incomingLink = false}

**Outgoing origin htlc (our payment) failed back by peer**
Forward {Incoming & outgoing key, payment hash}
Forward Failure {Incoming & outgoing key, payment hash}

**Outgoing origin htlc (our payment) settled**
Forward {Incoming & outgoing key, payment hash}
Settle {Incoming link key, payment hash} + TBD: would set type = `receive`

**Incoming htlc to forward not accepted (assuming this is on incoming link?):**
Link failure  {Incoming key, fail detail, incomingLink = true}

**Incoming htlc to forward accepted, but rejected by outgoing link**
Link failure  {Incoming key, fail detail, incomingLink = false}

**Htlc forwarded, but failed back by downstream peer**
Forward {Incoming & outgoing key, payment hash}
Forward Failure {Incoming & outgoing key, payment hash}

**Htlc forwarded and settled**
Forward {Incoming & outgoing key, payment hash}
Settle {Incoming link key, payment hash}


> Or is that what you suggest with the send/receive/fwd event type?

A type enum would go in the settle events to distinguish our own sends being settled from our receives (since they have the same data right now). Unrelated to the four main event types. I will update the PR with this today, which will hopefully clarify what I mean by it.

htlcswitch/htlcnotifier/htlcnotifier.go
carlaKC: I think this PR specifically is about exposing useful information to routing operators. And there could very well be no master plan specifically in this region. We expose this information, put it in lndMon so it's nice and visible and people can start to understand their routing activities more. Best effort (which is actually pretty good, barring a crash in the switch) is sufficient. 

To me, this PR is about surfacing failures, because the rest of the information is already available (in forwarder log and list invoices/payments). In terms of an audit log, I do not think we are at the stage where we need to worry about auditing HTLC fails (yet). We need to start with the basics, auditing channel balances against the successful events in the channel and hardening the forwarder log.

htlcswitch/link.go
carlaKC: > Maybe only using internal errors and mapping them last minute to wire messages could work long-term, but I think you tried something like that already?

This is a nice idea, I can see `switchErrors` being a thing :) I do worry about duplication though, specifically in the case where add channel updates to wire messages. Would have to duplicate the metadata and error one for one in a lot of cases (`FailExpiryTooSoon`/`FailIncorrectCltvExpiry` etc)

joostjager: Maybe the channel update can also be attached last minute. The duplication may be worth it, especially if we for some reason start to deviate with our wire messages. But ok, something for later if we want to do it at all.

joostjager:
Ok, nice overview. I see that there are multiple events fired for a single forward, one for the forward-going path and one for the backward path. This is also the case for an outgoing origin htlc. For incoming there is only one event. Maybe it is more consistent to always do two events and cover the hodl invoice accept too? So one event that describes what is requested from us by the htlc metadata and the second event what happened to the request.

Btw, don't forget the on-chain resolution flow where htlc decision are made too.

htlcswitch/link.go
halseth: What is the issue with keep returning `lnwire.FailureMessage` where we don't (yet) have a `failurDetail`, and only creating the `ForwardingError`  when that is available?

carlaKC: Can actually keep returning `lnwire.FailureMessage` and then just check for `ForwardingError` type when we need the extra field 👍

htlcswitch/link.go
```diff
@@ -1356,6 +1369,23 @@ func (l *channelLink) handleDownStreamPkt(pkt *htlcPacket, isReProcess bool) {
 		pkt.outgoingHTLCID = index
 		htlc.ID = index
 
+		l.cfg.HTLCNotifier.NotifyForwardingEvent(htlcnotifier.ForwardingEvent{
```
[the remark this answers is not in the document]
carlaKC: Think it will be ok to export `htlcPacket`? Otherwise `htlcnotifier` needs to be moved into `htlcSwitch`.  Nice idea though, these do get a bit lenghty!

carlaKC:
PR needs some work adding on chain events, not ready for review right now.

MrManPew:
related to this: https://github.com/lightningnetwork/lnd/issues/3420

MrManPew:
#3420

MrManPew:
as specified here: #3420

carlaKC:
> make setting `FailureDetail` mandatory when instantiating the `ForwardingError`, possibly by creating a constructor function that takes both `FailureMessage` and `FailureDetail`. Otherwise it remains a bit loose and bugs may sneak in.

This is a little tricky because the point of `FailureDetail` was for it to be optional (otherwise we end up with duplicate messages to those in lnwire), and there are some legacy cases where we don't even set a failure message (`SphinxErrorDecrypter.DecryptError`). 

One approach for this would be to implement an internal error interface `htlcSwitch.linkError` which will map to wire messages with some`getWireMessage` which does not implement `lnwire.FailureMessage` itself, so there are no accidental overlaps. You suggested something of this sort in an earlier comment?
 
>     Check carefully whether the notification interface is consistent.
Updated in this version. There is some ambiguity for certain events (eg an incoming link failure for a forward and an invoice look the same), so I have gone with a event type enum which indicates whether an event is linked to a send/receive/forward. This makes it clear to the user, and is easily implemented because forwards and our own sends/receives are mostly handled by different code.

carlaKC:
Requesting review on this, although I suspect there's more discussion to be had around `ForwardingError`, so after a conceptual pass at this stage rather than in depth review.

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,337 @@
...
+		"to %v(%v)", event.HTLCEventType, event.IncomingChannel,
+		event.HTLCInIndex, event.OutgoingChannel, event.HTLCOutIndex)
+
+	if err := h.ntfnServer.SendUpdate(event); err != nil {
+		log.Warnf("Unable to send forwarding fail event: %v", err)
+	}
+}
+
+// NotifySettleEvent notifies the HTLCNotifier that a HTLC that we committed to
+// as part of a forward or a receive to our node has been settled.
```
joostjager: To clarify, it is us who settled the htlc right? Not our peer settling our htlc?

htlcswitch/failure.go
```diff
@@ -43,6 +44,34 @@ func (f *ForwardingError) Error() string {
 	)
 }
 
+// Decode reads the bytes stream and converts it to the object. If the failure
+// message that the forwarding error wraps is serializable, its decode function
+// is used. Otherwise, the function is a nop.
+//
+// Note is is part of the lnwire.Serializable interface.
+func (f *ForwardingError) Decode(r io.Reader, pver uint32) error {
```
joostjager: Can't the caller just call `Decode` on `FailureMessage`? This suggests that all information is serialized, while in reality the detail gets lost.

htlcswitch/link.go
```diff
@@ -2830,31 +2830,46 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
...
+	// This allows for failures with a specific detail attached, rather than
+	// falling through to processHodlEvent and treating all failures like hodl
+	// cancels.
+	switch result {
+	case invoices.ResultAmountTooLow:
+		return sendFailure(&FailureDetailInvoiceUnderpaid{})
+	case invoices.ResultExpiryTooSoon:
+		return sendFailure(&FailureDetailInvoiceExpiryTooSoon{})
+	case invoices.ResultInvoiceAlreadyCanceled:
+		return sendFailure(&FailureDetailInvoiceAlreadyCancelled{})
```
joostjager: What about the other results?

It also make me wonder whether we should return `FailureDetail` objects instead of `UpdateResult`s from the registry.

htlcswitch/packet.go
```diff
@@ -54,6 +54,13 @@ type htlcPacket struct {
 	// encrypted with any shared secret.
 	localFailure bool
 
+	// linkFailure is non-nil if we have failed the HTLC on our incoming or
+	// outgoing link. The values contained in the linkFailure are used to notify
+	// detailed link failure events. If the wire failure reason is enriched with
+	// additional information which we do not expose to the network, linkFailure
+	// will be a ForwardingError which contains additional information.
+	linkFailure lnwire.FailureMessage
```
joostjager: Does this now have overlap with `localFailure`? (also failures occurring at our node)

htlcswitch/switch.go
```diff
@@ -754,66 +755,90 @@ func (s *Switch) handleLocalDispatch(pkt *htlcPacket) error {
 	// User have created the htlc update therefore we should find the
 	// appropriate channel link and send the payment over this link.
 	if htlc, ok := pkt.htlc.(*lnwire.UpdateAddHTLC); ok {
-		// Try to find links by node destination.
-		s.indexMtx.RLock()
-		link, err := s.getLinkByShortID(pkt.outgoingChanID)
-		s.indexMtx.RUnlock()
+		// Try to get the link that the HTLC should be forwarded outward on,
+		// and check that it satisfies the link's policy.
+		link, err := s.handleLocalAddHTLC(pkt, htlc)
```
joostjager: Isolating code move in its own commit makes review easier

contractcourt/interfaces.go
```diff
@@ -27,7 +27,7 @@ type Registry interface {
 	NotifyExitHopHtlc(payHash lntypes.Hash, paidAmount lnwire.MilliSatoshi,
 		expiry uint32, currentHeight int32,
 		circuitKey channeldb.CircuitKey, hodlChan chan<- interface{},
-		payload invoices.Payload) (*invoices.HodlEvent, error)
+		payload invoices.Payload) (*invoices.HodlEvent, invoices.UpdateResult, error)
```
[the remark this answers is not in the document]
joostjager: Although, I think for hodl invoices that are canceled eventually, there is still a failure detail needed?

htlcswitch/switch.go
carlaKC: Will do, I'm going to split in to multiple PRs as suggested somewhere else here and split up the notify commits as well (into sends/receives and forwards where possible)

htlcswitch/link.go
carlaKC: The other results don't need `htlcNotifier` events, replays will already have been notified and settles are handled by `processHodlEvent` when we settle the invoice (and needs to be there because we call it in multiple places).

htlcswitch/failure.go
carlaKC: This is where my concern with `ForwardingError` embedding `lnwire.FailureMessage` comes in. They are mostly being passed around as `lnwire.FailuresMessage` and then we check whether it is a `ForwardingError`.

This gets buggy with `Serialize` because you have to check whether your`lnwire.FailureMessage` is a forwarding error, and unwrap it if you want to use the serialize code (since forwarding error doesn't implement Serialize). It's fine in one place but I worry that we will forget to unwrap somewhere else and then end up with bugs. 

Starting to think that an internal error type interface with `GetWireMessage() lnwire.FailureMessage` may be the way to go. To prevent having to duplicate every wire message we just have one type that is a wrapped wire message. 
```
type WrappedWire{
  msg lnwire.FailureMessage
}

func (w * WrappedWire) GetWireMessage() { return w.msg }
```

And then messages that are specific can return their associated wire failure:
```
type InsufficientBalanceError struct{}
func (i * InsufficientBalanceError) GetWireMessage() {return TemporaryChanFailure }
```

I like that this puts a barrier between internal errors and the wire by having to specifically call `GetWireMessage` and provides more info about failures. Thoughts?

htlcswitch/packet.go
carlaKC: It does, I thought they were different because we set `linkFailure` for incoming failures as well but local failure is only used for local sends so can combine the two.

joostjager:
Comment about on-chain resolver notification is still outstanding also.

carlaKC:
> Do you have an idea how to add test coverage for the notifications?

The follow up PR  to this one (#3848, only last 2 commits) adds checks to existing itests to make sure that the correct notifications are being sent. It needs some updating (holding off until this PR has stabilized) so it might not cover every case, but the general idea is there. 

> Where to document what events can be expected? Is it worth adding something like the table that is now in the pr discussion to the code as a comment? And/or to the proto files in a later PR?

I like the idea of adding a table/explanation of all the different scenarios for the notifications you'd get and the fields that would be set. I think the package comment in `htlcnotifier` would be the best place for this?

> Comment about on-chain resolver notification is still outstanding also.

This was discussed this a while back and decided that on chain resolutions would be handled in a follow up PR.

joostjager:
> The follow up PR to this one (#3848, only last 2 commits) adds checks to existing itests to make sure that the correct notifications are being sent. It needs some updating (holding off until this PR has stabilized) so it might not cover every case, but the general idea is there.

Ah, okay. I was actually thinking about unit tests. We already test many scenarios for link and/or switch. Maybe some extra asserts can be added in there.

> > Where to document what events can be expected? Is it worth adding something like the table that is now in the pr discussion to the code as a comment? And/or to the proto files in a later PR?
> 
> I like the idea of adding a table/explanation of all the different scenarios for the notifications you'd get and the fields that would be set. I think the package comment in `htlcnotifier` would be the best place for this?

Sounds good.

> > Comment about on-chain resolver notification is still outstanding also.
> 
> This was discussed this a while back and decided that on chain resolutions would be handled in a follow up PR.

What was again the reason for postponing that? Isn't that a matter of just sprinkling a few more notifications in the resolvers? I do think that it can be useful to at least think through the required changes, in case a new architectural requirement comes up.

carlaKC:
> Ah, okay. I was actually thinking about unit tests. We already test many scenarios for link and/or switch. Maybe some extra asserts can be added in there.

Nice, I'll have a look at the unit tests that test a full flow. 

> What was again the reason for postponing that? Isn't that a matter of just sprinkling a few more notifications in the resolvers? 

iirc the reason was that the use case for the `htlcnotifier` is anticipated to be for node operators who are concerned with the activities of their open channels (#3420). While we care about htlcs which are swept on chain, there is less of a use case for a live subscription to these events. It is more code to write, test and review, which doesn't have much demand for the feature.

> I do think that it can be useful to at least think through the required changes, in case a new architectural requirement comes up.

I'll have another look at this during the week, but I think it's as simple as dropping an `OnChain` bool into the existing types. It's not a part of the codebase I've touched much, so would rather separate concerns and follow up with another PR if it really is trivial.

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,388 @@
...
+	h.notifyLinkFailure(pkt, hash, eventType)
+}
+
+// notifyLinkFailure sends a link failure notification with the event type
+// provided.
+func (h *HTLCNotifier) notifyLinkFailure(pkt *htlcPacket, hash lntypes.Hash,
+	eventType HTLCEventType) {
+
+	// Sanity check that the packet's link failure is provided.
+	if pkt.linkFailure == nil {
```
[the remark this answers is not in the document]
carlaKC: not expected to be, I'll remove it and it'll crash below if it's nil 🤠

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,388 @@
...
+	// HTLCInfo contains further information about the HTLC.
+	HTLCInfo
+
+	// Timestamp is the time when the forwarding failure was received.
+	Timestamp time.Time
+}
+
+// SettleEvent represents a HTLC that was forwarded by our node and
+// successfully settled.
+type SettleEvent struct {
```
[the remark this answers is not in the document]
carlaKC: I do like having the long comment on `HtlcNotifier` at the top of the file, could move the notifier functions up so everything is grouped better?

htlcswitch/interfaces.go
```diff
@@ -180,3 +180,31 @@ type TowerClient interface {
 	// isTweakless should be true.
 	BackupState(*lnwire.ChannelID, *lnwallet.BreachRetribution, bool) error
 }
+
+// htlcNotifier is an interface which represents the input side of the
+// HTLCNotifier which htlc events are piped through. This interface is intended
+// to allow for mocking of the htlcNotifier in tests, so is unexported because
+// it is not needed outside of the htlcSwitch package.
+type htlcNotifier interface {
```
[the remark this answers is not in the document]
carlaKC: Because the interface is actually only required to make tests in link/switch easier. The implementation is set in the server so needs to be exported.

htlcswitch/link.go
```diff
@@ -1413,6 +1413,9 @@ func (l *channelLink) handleDownStreamPkt(pkt *htlcPacket, isReProcess bool) {
 		pkt.outgoingHTLCID = index
 		htlc.ID = index
 
+		// Send a forward event notification to htlcNotifier.
```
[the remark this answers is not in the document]
carlaKC: lnd clock subscription to htlc events to trigger alarms? 😅 🔔

htlcswitch/htlcnotifier.go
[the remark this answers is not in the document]
halseth: #reckless

htlcswitch/switch.go
```diff
@@ -150,6 +150,10 @@ type Config struct {
 	// the switch when a new block has arrived.
 	Notifier chainntnfs.ChainNotifier
 
+	// HtlcNotifier is an instance of a htlcNotifier which we will pipe HTLC
+	// events through.
+	HtlcNotifier htlcNotifier
```
halseth: I'm actually a bit surprised that this works, that you can export a field of an unexported type 🤔

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,381 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/subscribe"
+)
+
+// Htlcnotifier notifies clients of htlc forwards, failures and settles for
```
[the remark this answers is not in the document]
carlaKC: typo being `Htlcnotifier` -> `HtlcNotifier`? 
or am I so bad at spelling I don't know what I can't spell 🤔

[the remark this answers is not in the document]
halseth: yes 😂

joostjager:
Needs rebase

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,381 @@
...
+//
+// Note this is part of the htlcNotifier interface.
+func (h *HtlcNotifier) NotifyForwardingEvent(pkt *htlcPacket,
+	hash lntypes.Hash) {
+
+	event := &ForwardingEvent{
+		HtlcKey:   newHtlcKey(pkt),
+		HtlcInfo:  newHtlcInfo(pkt, hash, getEventType(pkt)),
+		Timestamp: time.Now(),
+	}
```
joostjager: I am not totally fully comfortable with the function signature and passing in `htlcPacket`. Especially because sometimes an instance of `htlcPacket` is created only for the notification. Isn't it better to let the caller instantiate the event? Then the `HtlcNotifier` could maybe even disappear completely and replaced by just an instance of `subscribe.Server`.

htlcswitch/link.go
```diff
@@ -1525,10 +1537,22 @@ func (l *channelLink) handleDownStreamPkt(pkt *htlcPacket, isReProcess bool) {
...
 		l.cfg.Peer.SendMessage(false, htlc)
 		isSettle = true
+
+		// Finally, we notify the failure. If the packet does not have a
+		// link failure set, it failed further down the route so we
+		// notify a forwarding failure. Otherwise, we notify a link
+		// failure because it failed at our node.
+		if pkt.linkFailure != nil {
+			l.cfg.HtlcNotifier.NotifyOutgoingLinkFailEvent(
+				pkt, lntypes.Hash{},
```
joostjager: Empty hash?

htlcswitch/htlcnotifier.go
carlaKC: Decided on passing in raw packets because there's some `hltcNotifier` side logic that I don't want calling functions to have to do (eg, figuring out event type based on presence of channel IDs). Could do this with a constructor for each type, it will just have a ton of arguments?

idk about just a subscribe server, our current pattern is to wrap these servers in structs which strictly control what can be piped in which I like. Plus subscribe servers aren't great for unit tests, whereas htlcnotifier can unobtrusively mocked when we need.

htlcswitch/link.go
carlaKC: Needed to circle back to this, I wasn't sure if `pkt.circuit` would always be non-nil so that we could use `pkt.circuit.PaymentHash` here. Will investigate and update.

htlcswitch/link.go
```diff
@@ -1529,10 +1548,36 @@ func (l *channelLink) handleDownStreamPkt(pkt *htlcPacket, isReProcess bool) {
...
 		isSettle = true
+
+		// Finally, we notify the failure.
+		key := newHtlcKey(pkt)
+		info := HtlcInfo{
+			IncomingTimeLock: pkt.incomingTimeout,
+			OutgoingTimeLock: pkt.outgoingTimeout,
+			IncomingAmt:      pkt.incomingAmount,
+			OutgoingAmt:      pkt.amount,
+			HtlcEventType:    getEventType(pkt),
```
carlaKC: Payment hash is not immediately available here, unsure of how critical this is given that we identify htlcs by circuit key? 

The only place I can see this being a problem is when you subscribe to events after the forward and before the failure, which is the case where we would not know the hash this is associated with.

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,276 @@
...
+	// on.
+	IncomingChannel lnwire.ShortChannelID
+
+	// OutgoingChannel is the channel that the outgoing htlc left our
+	// node on.
+	OutgoingChannel lnwire.ShortChannelID
+
+	// HtlcInIndex is the index of the incoming htlc in the incoming
+	// channel.
+	HtlcInIndex uint64
```
joostjager: `IncomingHtlcIndex` or ` ChannelIn`, to make it consistent? Or just two `CircuitKey` instances.

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,276 @@
...
+
+	// HtlcOutIndex is the index of the outgoing htlc in the outgoing
+	// channel.
+	HtlcOutIndex uint64
+}
+
+// String returns a string representation of a htlc key.
+func (k HtlcKey) String() string {
+	return fmt.Sprintf("%v(%v) -> %v(%v)", k.IncomingChannel,
+		k.HtlcInIndex, k.OutgoingChannel, k.HtlcOutIndex)
```
joostjager: Special case local sends and receives (nice to have)

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,276 @@
...
+	Timestamp time.Time
+}
+
+// SettleEvent represents a htlc that we settled.
+type SettleEvent struct {
+	// HtlcKey contains the information which uniquely identifies the htlc.
+	HtlcKey
+
+	// HtlcInfo contains further information about the htlc.
+	HtlcInfo
```
joostjager: Merge `HtlcInfo` and `HtlcKey`, as they are always both present?

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,276 @@
...
+// and outgoing channel id and index. Htlcs, and their subsequent settles or
+// fails, can be identified by these four elements. Note that sends from our
+// node will have a zero incoming channel id/index, because they originated
+// locally, and receives will have a zero outgoing channel id/index because
+// they terminate at our node.
+type HtlcNotifier struct {
+	started sync.Once
+	stopped sync.Once
+
+	ntfnServer *subscribe.Server
```
joostjager: Can `subscribe.Server` be embedded, so that the `Start/Stop functions are inherited?

htlcswitch/link.go
joostjager: Shouldn't we even discourage users of the htlc notifier of using the hash to relate events? Maybe none of the 'outcome' events should include a hash.

htlcswitch/link.go
```diff
@@ -1414,6 +1418,14 @@ func (l *channelLink) handleDownStreamPkt(pkt *htlcPacket, isReProcess bool) {
 
 		l.cfg.Peer.SendMessage(false, htlc)
 
+		// Send a forward event notification to htlcNotifier.
+		l.cfg.HtlcNotifier.NotifyForwardingEvent(
+			NewForwardingEvent(
```
joostjager: It seems that you are using these constructors virtually everywhere. In that case, why not pass in the values to the notify methods directly? Also allows the timestamp to be set in the notifier itself.

htlcswitch/link.go
carlaKC: > Maybe none of the 'outcome' events should include a hash.

I think this is a  good idea. I will move hash into forwarding events, and take it out of settles/fails.

carlaKC:
Picked up that `pid` is stashed in our incoming `htlcID` for sends in a unit test, sneaky. 
Decided to separate it out so that sends/receives have consistent zero circuit keys.

htlcswitch/htlcnotifier.go
joostjager: Like, resolved, but no change? :)

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,317 @@
...
+// HtlcInfo provides the details of a htlc that our node has processed. For
+// forwards, incoming and outgoing values are set, whereas sends and receives
+// will only have outgoing or incoming details set. Htlcs can be uniquely
+// identified by their combined incoming and outgoing circuits.
+type HtlcInfo struct {
+	// IncomingCircuit is the channel an htlc id of the incoming htlc.
+	IncomingCircuit channeldb.CircuitKey
+
+	// OutgoingCircuit is the channel and htlc id of the outgoing htlc.
+	OutgoingCircuit CircuitKey
```
joostjager: reference type directly as for `IncomingCircuit`?

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,317 @@
...
+
+	// IncomingAmt is the amount of the htlc on our incoming channel.
+	IncomingAmt lnwire.MilliSatoshi
+
+	// OutgoingAmt is the amount of the htlc on our outgoing channel.
+	OutgoingAmt lnwire.MilliSatoshi
+
+	// HtlcEventType indicates whether the event is part of a send from
+	// our node receive to our node of forward through our node.
+	HtlcEventType
```
joostjager: Isn't event type redundant because it can be derived from the value of `IncomingCircuit` and `OutgoingCircuit` (nil checks)? If this is indeed true, then maybe `HtlcEventType` could become method.

htlcswitch/htlcnotifier.go
joostjager: I think it is much cleaner now that you keep the packet in the link/switch.

htlcswitch/htlcnotifier.go
```diff
@@ -323,3 +325,39 @@ func (h *HtlcNotifier) NotifySettleEvent(info HtlcInfo) {
...
+		OutgoingAmt:      pkt.amount,
+		HtlcEventType:    getEventType(pkt),
+	}
+}
+
+// getEventType returns the htlc type based on the fields set in the htlc
+// packet. Sends that originate at our node have the source (zero) incoming
+// channel ID. Receives to our node have the exit (zero) outgoing channel ID
+// and forwards have both fields set.
+func getEventType(pkt *htlcPacket) HtlcEventType {
```
joostjager: Oh, you add this method now. I'd just remove the `HtlcEventType` field completely and only keep the method.

htlcswitch/link.go
```diff
@@ -1418,6 +1418,12 @@ func (l *channelLink) handleDownStreamPkt(pkt *htlcPacket, isReProcess bool) {
 
 		l.cfg.Peer.SendMessage(false, htlc)
 
+		// Send a forward event notification to htlcNotifier.
+		info, pid := downstreamPacketInfo(pkt)
```
joostjager: Do we indeed want to expose the pid? Or do we stick to reporting the outgoing chan/htlc id and let users of the notifier associated htlcs and payments using the to-be-added circuit key in the payment db?

htlcswitch/htlcnotifier.go
carlaKC: ah sorry, was conflating it with [this](https://github.com/lightningnetwork/lnd/pull/3781#discussion_r378064851) so resolved to de-dup. 

I like having the log line for starting like we have for other notifiers (eg `ChannelNotifier`).

htlcswitch/switch_test.go
```diff
@@ -2690,3 +2692,407 @@ func TestInvalidFailure(t *testing.T) {
...
+	select {
+	case e := <-alice:
+		settle, ok := e.(*SettleEvent)
+		if !ok {
+			t.Fatalf("expected alice to have a settle "+
+				"event, got: %T", e)
+		}
+
+		assertCircuit(t, zeroCircuit, settle.IncomingCircuit)
+		assertCircuit(t, aliceToBob, settle.OutgoingCircuit)
```
joostjager: Some method extraction is possible to reduce duplication

htlcswitch/switch_test.go
```diff
@@ -2690,3 +2692,407 @@ func TestInvalidFailure(t *testing.T) {
...
+		HtlcID: htlcID,
+	}
+
+	return aliceToBob, bobFromAlice, bobToCarol, carolFromBob
+}
+
+// checkSuccessful is a function that checks that Alice, Bob and Carol
+// have the correct set of event for a payment that is successfully
+// routed over Alice -> Bob -> Carol.
+func checkSuccessful(t *testing.T, alice, bob, carol <-chan interface{},
```
joostjager: There is quite a bit of assertion code here. Can't this test be written by just having three (one for every node) lists of fully-specified events that are expected? The actual events are then just read from the channels and deep-equaled with the expectations.

htlcswitch/htlcnotifier.go
joostjager: Ok

htlcswitch/htlcnotifier.go
carlaKC: There's one case where we can't just use circuits. When a forward fails on our incoming link, we have not decoded the outgoing link information yet so it presents like a receive. The type field was added after walking through the info we have for each type [here](https://github.com/lightningnetwork/lnd/pull/3781#issuecomment-560361544). 

For the case where we fail an incoming htlc, the `getEventType` function is not used and we set the type manually.

htlcswitch/switch_test.go
carlaKC: Switched over to a set of results, still quite a lot of setup code, but I think it's better. Less duplication at least.

htlcswitch/link.go
carlaKC: Indeed, cut the pid and payment hash 👍

htlcswitch/htlcnotifier.go
joostjager: Ah, missed that.

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,356 @@
...
+	HtlcEventType
+
+	// Timestamp is the time when the forwarding failure was received.
+	Timestamp time.Time
+}
+
+// SettleEvent represents a htlc that was settled.
+type SettleEvent struct {
+	// HtlcKey uniquely identifies the htlc.
+	HtlcKey
```
joostjager: No `HtlcInfo` here?

htlcswitch/htlcnotifier.go
```diff
@@ -0,0 +1,356 @@
...
+	// Timestamp is the time when the link failure occurred.
+	Timestamp time.Time
+}
+
+// ForwardingFailEvent represents a htlc failure which occurred down the line
+// after we forwarded a htlc onwards. An error is not included in this event
+// because errors returned down the route are encrypted.
+type ForwardingFailEvent struct {
+	// HtlcKey uniquely identifies the htlc.
+	HtlcKey
```
joostjager: No `HtlcInfo` here?

htlcswitch/switch.go
```diff
@@ -898,6 +898,19 @@ func (s *Switch) handleLocalResponse(pkt *htlcPacket) {
...
+	eventType := getEventType(pkt)
+
+	switch pkt.htlc.(type) {
+	case *lnwire.UpdateFulfillHTLC:
+		s.cfg.HtlcNotifier.NotifySettleEvent(key, eventType)
+
+	case *lnwire.UpdateFailHTLC:
+		s.cfg.HtlcNotifier.NotifyForwardingFailEvent(key, eventType)
+	}
+
```
joostjager: newline

htlcswitch/switch_test.go
```diff
@@ -2697,3 +2700,349 @@ func TestInvalidFailure(t *testing.T) {
...
+				IncomingCircuit: channeldb.CircuitKey{
+					ChanID: channels.carolToBob.ShortChanID(),
+					HtlcID: htlcID,
+				},
+				OutgoingCircuit: zeroCircuit,
+			},
+			HtlcEventType: HtlcEventTypeReceive,
+			Timestamp:     ts,
+		},
+	}
```
joostjager: Nice set of expectations

htlcswitch/htlcnotifier.go
halseth: Wasnt't it also suggested earlier to merge `HlcJey` and `HtlcInfo`?

htlcswitch/switch.go
```diff
@@ -1256,12 +1256,21 @@ func (s *Switch) failAddPacket(packet *htlcPacket, failure *LinkError) error {
...
+	// Create a failure packet for this htlc. The the full set of
+	// information about the htlc failure is included so that they can
+	// be included in link failure notifications.
 	failPkt := &htlcPacket{
-		sourceRef:      packet.sourceRef,
-		incomingChanID: packet.incomingChanID,
-		incomingHTLCID: packet.incomingHTLCID,
-		circuit:        packet.circuit,
-		linkFailure:    failure,
+		sourceRef:       packet.sourceRef,
```
halseth: nit: this change could be made its own commit with an explaining commit message. Reason being this is touching existing code, so it would be nice (IMO) to see it in isolation to deem whether it is safe.

htlcswitch/switch.go
```diff
@@ -811,6 +782,47 @@ func (s *Switch) handleLocalDispatch(pkt *htlcPacket) error {
 	return nil
 }
 
+// handleLocalAddHTLC handles the addition of a htlc for a send that
```
halseth: nice to do code moves in separate commits 👍

htlcswitch/switch_test.go
```diff
@@ -2697,3 +2700,349 @@ func TestInvalidFailure(t *testing.T) {
...
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			// First, we'll create our traditional three hop
```
halseth: style nit: to save on indentation could move test case into separate method.

htlcswitch/test_utils.go
```diff
@@ -1045,6 +1053,34 @@ func newThreeHopNetwork(t testing.TB, aliceChannel, firstBobChannel,
...
+
+// serverOptionWithHtlcNotifier is a functional option for the creation of
+// three hop network servers which allows setting of htlc notifiers.
+// Note that these notifiers should be started and stopped by the calling
+// function.
+func serverOptionWithHtlcNotifier(alice, bob,
+	carol *HtlcNotifier) serverOption {
+
+	return func(aliceServer, bobServer, carolServer *mockServer) {
+		aliceServer.htlcSwitch.cfg.HtlcNotifier = alice
```
halseth: could we add the htlc notifier by default when creating the mockServers? It should not interfere with existing test cases,

carlaKC:
Some clarification on `HtlcKey`/`HtlcInfo`: we do not have the information we need for `HtlcInfo` reliably around for settles/fails, particularly on startup so I thought it would be better to split them out and exclude them entirely from settle and fail events. I see this being used in rate limiting/ performance metrics rather than exact accounting of htlc movement. If people need to look up additional information for events, they can look in forwarding log/payments/invoices.

htlcswitch/test_utils.go
carlaKC: This gets a bit more involved, because we then need to pass mocked time into `newThreeHopNetwork` (which would probably need to be with functional options like this anyway, so all the other calls don't need to be updated), thread the notifier down to `initSwithcWithDB` (more functional opts so we don't need to update every caller) and either expand the `htlcNotifier` interface to include `Start`/`Stop`/`SubscribeEvents` or do a type assertion in the test to get the event subscription. If it's not a blocker, I think I'd rather leave it like this?

htlcswitch/switch_test.go
carlaKC: nice

carlaKC:
Thank you for all the review @halseth and @joostjager! The saga is almost complete 🎢 

> Final question: did you try running a three hop regtest network and sending some payments back and forth manually with log on trace? As a rough sanity check to see that the log messages make sense.

Indeed, I have a very rich local forwarding node after all the iterations of this PR. 

Send from Alice -> Bob -> Carol looks like this in logs:

Alice:
```
Notifying forward event: send over (Chan ID=6160:1:1, HTLC ID=4), outgoing amount: 2000 mSAT, outgoing timelock: 6260

Notifying settle event: send over (Chan ID=6160:1:1, HTLC ID=4)
```

Bob:
```
Notifying forward event: forward over (Chan ID=6160:1:1, HTLC ID=4) -> (Chan ID=6172:1:1, HTLC ID=0), incoming amount: 2000 mSAT, incoming timelock: 6260, outgoing amount: 1000 mSAT, outgoing timelock: 6220

Notifying settle event: forward over (Chan ID=6160:1:1, HTLC ID=4) -> (Chan ID=6172:1:1, HTLC ID=0)
```

Carol:
```
Notifying settle event: receive over (Chan ID=6172:1:1, HTLC ID=0)
```

PR #3782, opened 2019-12-02
routing: improve equal cost route comparison

joostjager:
When the (virtual) payment attempt cost is set to zero, probabilities are no longer a factor in determining the best route. In case of routes with equal costs, we'd just go with the first one found. This commit refines this behavior by picking the route with the highest probability. So even though probability doesn't affect the route cost, it is still used as a tie breaker.

### DIFF
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -493,13 +493,25 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 			int64(cfg.PaymentAttemptPenalty),
 		)
 
-		// If the current best route is better than this candidate
-		// route, return. It is important to also return if the distance
-		// is equal, because otherwise the algorithm could run into an
-		// endless loop.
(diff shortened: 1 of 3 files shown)


routing/pathfind.go
```diff
@@ -493,13 +493,25 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
-		// route, return. It is important to also return if the distance
-		// is equal, because otherwise the algorithm could run into an
-		// endless loop.
+		// If there is already a best route stored, compare this
+		// candidate route with the best route so far.
 		current, ok := distance[fromVertex]
-		if ok && tempDist >= current.dist {
-			return
+		if ok {
+			// If this route is worse then what we already found,
```
guggero: nit: s/then/than/

joostjager:
Test added. Also ran into the heap comparison, which I updated as well.

PR #3785, opened 2019-12-03
discovery: properly set FirstBlockHeight and NumBlocks in responses

### LINKED ISSUE #3728: LND mis-implements gossip_queries
### Background

c-lightning has started sending errors to LND nodes because of their `reply_channel_range` answers for large queries.

Tested against @bitconner's node.

### Steps to reproduce

We send 01076fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d61900000000000007b2b400018402 which is:

```
QUERY_CHANNEL_RANGE:
chain_hash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
first_blocknum=504500
number_of_blocks=99330
```

LND replies with four messages (because of non-zlib):

```
REPLY_CHANNEL_RANGE:
chain_hash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
first_blocknum=504500
number_of_blocks=99330
complete=0
...
REPLY_CHANNEL_RANGE:
chain_hash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
first_blocknum=504500
number_of_blocks=99330
complete=0
...
REPLY_CHANNEL_RANGE:
chain_hash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
first_blocknum=504500
number_of_blocks=99330
complete=0
...
REPLY_CHANNEL_RANGE:
chain_hash=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
first_blocknum=504500
number_of_blocks=99330
complete=1
```

These responses are incorrect.

### Expected behaviour

There are two problems: first, the range of blocks is wrong, second, the complete flags is wrong.

From the spec:

```markdown
  - MUST respond with one or more `reply_channel_range` whose combined range
	cover the requested `first_blocknum` to `first_blocknum` plus
	`number_of_blocks` minus one.
  - For each `reply_channel_range`:
    - MUST set with `chain_hash` equal to that of `query_channel_range`,
    - MUST encode a `short_channel_id` for every open channel it knows in blocks `first_blocknum` to `first_blocknum` plus `number_of_blocks` minus one.
    - MUST limit `number_of_blocks` to the maximum number of blocks whose
      results could fit in `encoded_short_ids`
```

You do not limit `number_of_blocks` to the maximum number of blocks which fit in the `encoded_short_ids` field, but instead, reuse the same value.  The intention was that you return (non-overlapping, though that is not stated) `first_blocknum` + `number_of_blocks` ranges.  Instead you return the same number over and over, which is wrong.

Secondly:

```markdown
    - if does not maintain up-to-date channel information for `chain_hash`:
      - MUST set `complete` to 0.
    - otherwise:
      - SHOULD set `complete` to 1.
```

`complete` is NOT an end marker.  It is used to indicate a lite node (or node bootstrapping) which doesn't have full gossip information.

### Actual behaviour

Eclair and c-lightning now need to figure out how to detect and mitigate this for now, and make sure we test protocol compatibility properly in the future!

Roasbeef:
In this commit we fix in a bug in `lnd` that could cause other
implementations which implement a strict version of the spec to
disconnect when trying to sync their channel graph using the gossip
query feature. Before this commit, we would embed the request to a
`QueryChannelRange` in the response, causing some clients to reject the
response as the `FirstBlockHeight` and `NumBlocks` field would be
identical for each chunk of the response.

In order to remedy this, we now properly set these two fields with each
returned chunk. Note that even after this commit, we keep our existing
behavior surrounding the `Complete` field as is. Otherwise, current
`lnd` clients which rely on this field (rather than the two
aforementioned fields) wouldn't be able to properly detect when a set of
responses to their query was "complete".

Partially fixes #3728.

### DIFF
--- a/discovery/syncer.go
+++ b/discovery/syncer.go
@@ -824,6 +824,14 @@ func (g *GossipSyncer) replyChanRangeQuery(query *lnwire.QueryChannelRange) erro
 	// TODO(roasbeef): means can't send max uint above?
 	//  * or make internal 64
 
+	// In the base case (no actual response) the first block and the last
+	// block in the query will be the same. In the loop below, we'll update
+	// these two variables incrementally with each chunk to properly
+	// compute the starting block for each response and the number of
+	// blocks in a response.
+	firstBlockHeight := query.FirstBlockHeight
+	lastBlockHeight := query.FirstBlockHeight
+
 	numChannels := int32(len(channelRange))
 	numChansSent := int32(0)
 	for {
@@ -854,13 +862,31 @@ func (g *GossipSyncer) replyChanRangeQuery(query *lnwire.QueryChannelRange) erro
 				"size=%v", g.cfg.peerPub[:], len(channelChunk))
 		}
 
+		// If we have any channels at all to return, then we need to
+		// update our pointers to the first and last blocks for each
+		// response.
+		if len(channelChunk) > 0 {
+			firstBlockHeight = channelChunk[0].BlockHeight
+			lastBlockHeight = channelChunk[len(channelChunk)-1].BlockHeight
+		}
+
+		// The number of blocks contained in this response (the total
+		// span) is the difference between the last channel ID and the
+		// first in the range. We add one as even if all channels
+		// returned are in the same block, we need to count that.
+		numBlocksInResp := lastBlockHeight - firstBlockHeight + 1
+
 		// With our chunk assembled, we'll now send to the remote peer
 		// the current chunk.
 		replyChunk := lnwire.ReplyChannelRange{
-			QueryChannelRange: *query,
-			Complete:          0,
-			EncodingType:      g.cfg.encodingType,
-			ShortChanIDs:      channelChunk,
+			QueryChannelRange: lnwire.QueryChannelRange{
+				ChainHash:        query.ChainHash,
+				NumBlocks:        numBlocksInResp,
+				FirstBlockHeight: firstBlockHeight,
+			},
+			Complete:     0,
+			EncodingType: g.cfg.encodingType,
+			ShortChanIDs: channelChunk,
 		}
 		if isFinalChunk {
 			replyChunk.Complete = 1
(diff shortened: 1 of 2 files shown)


discovery/syncer_test.go
```diff
@@ -769,14 +780,45 @@ func TestGossipSyncerReplyChanRangeQuery(t *testing.T) {
 
 			// If this is not the last chunk, then Complete should
 			// be set to zero. Otherwise, it should be one.
+			offset := 0
```
cfromknecht: nit: comment should be moved down? also add rationale for why offset is set this way?

discovery/syncer.go
```diff
@@ -854,13 +862,31 @@ func (g *GossipSyncer) replyChanRangeQuery(query *lnwire.QueryChannelRange) erro
...
+		if len(channelChunk) > 0 {
+			firstBlockHeight = channelChunk[0].BlockHeight
+			lastBlockHeight = channelChunk[len(channelChunk)-1].BlockHeight
+		}
+
+		// The number of blocks contained in this response (the total
+		// span) is the difference between the last channel ID and the
+		// first in the range. We add one as even if all channels
+		// returned are in the same block, we need to count that.
+		numBlocksInResp := lastBlockHeight - firstBlockHeight + 1
```
bhandras: I might be totally off here, but to me it seems that the result of `ChannelDB.FetchChannelRange` is not ordered as it is simply a filter over a cursor which iterates a bucket (`edgeIndex`). If that is true then `lastBlockHeight` might not be larger than `firstBlockHeight` and the above will overflow as both are `uint32`.

wpaulino: IIRC the keys are sorted in ascending order in bolt, so this should be fine.

bhandras: Thanks @wpaulino, I missed that the Cursor iterates in order. (https://godoc.org/github.com/boltdb/bolt#Cursor)

cfromknecht:
@Roasbeef needs squash!

PR #3788, opened 2019-12-04
parse and generate bolt 11 invoices w/ payment addrs

cfromknecht:
Split off from #3679 

This PR:
 - adds decoding of invoice payment addresses in the `s` field
 - adds ability to set payment address on invoice via functional option
 - generates a random payment address for all newly created invoices
 - exposes payment addresses via `decodepayreq`

This will enable #3415 to properly test receiving crowd funded mpp payments via the `SendToRoute` rpcs.


lnrpc/rpc.proto
```diff
@@ -2615,6 +2615,7 @@ message PayReq {
     string fallback_addr = 8 [json_name = "fallback_addr"];
     int64 cltv_expiry = 9 [json_name = "cltv_expiry"];
     repeated RouteHint route_hints = 10 [json_name = "route_hints"];
+    bytes payment_addr = 11 [json_name = "payment_addr"];
```
joostjager: This is shown in `lncli decodepayreq` as base64. There we go again...

lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -397,6 +406,7 @@ func AddInvoice(ctx context.Context, cfg *AddInvoiceConfig,
 			Expiry:          payReq.Expiry(),
 			Value:           amtMSat,
 			PaymentPreimage: paymentPreimage,
+			PaymentAddr:     paymentAddr,
```
joostjager: The pay address isn't exposed on for example `ListInvoices`

joostjager: Features not either I noticed

zpay32/invoice_internal_test.go
```diff
@@ -458,56 +458,6 @@ func TestParseDestination(t *testing.T) {
...
-		}
-		if test.valid && !compareHashes(descriptionHash, test.result) {
-			t.Fatalf("test %d failed decoding description hash: "+
-				"expected %x, got %x",
-				i, *test.result, *descriptionHash)
-			return
-		}
-	}
-}
-
```
joostjager: :+1: useless copy paste material

zpay32/invoice_test.go
```diff
@@ -727,6 +756,20 @@ func TestNewInvoice(t *testing.T) {
...
+					testPaymentHash, time.Unix(1503429093, 0),
+					Amount(testMillisat24BTC),
+					Description(testEmptyString),
+					Destination(testPubKey),
+					PaymentAddr(testPaymentAddr),
+				)
+			},
+			valid:          true,
+			encodedInvoice: "lnbc241pveeq09pp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdqqnp4q0n326hr8v9zprg8gsvezcch06gfaqqhde2aj730yg0durunfhv66sp5qszsvpcgpyqsyps8pqysqqgzqvyqjqqpqgpsgpgqqypqxpq9qcrsmmw6t0kplachqfzd76tvuuln6k2zqmxf2u57xcla9lcxfcrsrd48q5ekp2xul50yc6x4wu5qaen3zm3wvuukf9fespkytwqzgya5cxgpeyed0p",
+		},
```
joostjager: Don't you want to use the same invoice as in the payment addr test above? So the encoded invoice is exactly the same and we can be more certain that it works?

halseth: The first test checks both encoding and decoding. This test checks the functional options I guess.

lnrpc/invoicesrpc/addinvoice.go
cfromknecht: both are exposed in the original pr

lnrpc/rpc.proto
cfromknecht: there’s a pr for that ;)

zpay32/invoice_test.go
cfromknecht: looking at it now, the entire second test suite looks redundant.

cfromknecht: removed the second test, since it tests both encode and decode

PR #3789, opened 2019-12-04
republish force and coop closes on startup

cfromknecht:
This PR adds granularity to our internal tracking of closed channels, allowing us to store closing transactions for both unilateral and cooperative closures. In the case that a channel is unilaterally closed and then later force closed, the contractcourt is now able to reliably publish both until either confirm. We also fix a race condition in the rpc interface to make the coop close status externally consistent with the coop close requests.

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -59,9 +59,13 @@ var (
 	// remote peer during a channel sync in case we have lost channel state.
 	dataLossCommitPointKey = []byte("data-loss-commit-point-key")
 
-	// closingTxKey points to a the closing tx that we broadcasted when
-	// moving the channel to state CommitBroadcasted.
-	closingTxKey = []byte("closing-tx-key")
+	// forceCloseTxKey points to a the unilateral closing tx that we
+	// broadcasted when moving the channel to state CommitBroadcasted.
+	forceCloseTxKey = []byte("closing-tx-key")
+
+	// coopCloseTxKey points to a the cooperative closing tx that we
+	// broadcasted when moving the channel to state CoopBroadcasted.
+	coopCloseTxKey = []byte("coop-closing-tx-key")
 
 	// commitDiffKey stores the current pending commitment state we've
 	// extended to the remote party (if any). Each time we propose a new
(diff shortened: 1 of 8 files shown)


channeldb/channel.go
```diff
@@ -919,6 +929,30 @@ func (c *OpenChannel) isBorked(chanBucket *bbolt.Bucket) (bool, error) {
...
 // handle the case where a different tx actually hits the chain.
 func (c *OpenChannel) MarkCommitmentBroadcasted(closeTx *wire.MsgTx) error {
+	return c.markBroadcasted(
+		ChanStatusCommitBroadcasted, forceCloseTxKey, closeTx,
+	)
+}
+
+// MarkCoopBroadcasted marks the channel to indicate that a cooperative close
+// transaction has been broadcast, either our own or the remote, and that we
+// should wach the chian for it to confirm before taking further action. It
```
wpaulino: Nit: watch the chain*

peer.go
```diff
@@ -463,6 +463,8 @@ func (p *peer) loadActiveChannels(chans []*channeldb.OpenChannel) (
 			fallthrough
 		case dbChan.HasChanStatus(channeldb.ChanStatusCommitBroadcasted):
 			fallthrough
+		case dbChan.HasChanStatus(channeldb.ChanStatusCoopBroadcasted):
```
wpaulino: If a peer shuts down during the coop close negotiation, then we won't be able to retry another coop close, which seems undesired?

cfromknecht: hmm that's true, hadn't thought about that. i have another follow up in this area where it'd be easy to change, for now tho it prevents the itests from flaking.

contractcourt/chain_arbitrator.go
```diff
@@ -570,6 +546,90 @@ func (c *ChainArbitrator) Start() error {
...
 
+// publishClosingTxs will load any stored cooperative or unilater closing
+// transactions and republish them. This helps ensure propagation of the
+// transactions in the event that prior publications failed.
+func (c *ChainArbitrator) publishClosingTxs(
+	channel *channeldb.OpenChannel) error {
+
+	// If the channel has had its unilateral close broadcasted already,
+	// republish it in case it didn't propagate.
+	if channel.HasChanStatus(channeldb.ChanStatusCommitBroadcasted) {
```
Roasbeef: Perhaps these two should be mutually exclusive? Or even maybe we should prefer the _coop_ close as it doesn't lead to coins being time locked.

cfromknecht: that's definitely possible, favoring the coop close sounds good to me since the force close would probably win out due to the higher fee rate. this is preexisting tho, and fixing it will sprawl a bit so maybe we do it in a followup? as is we fix the issue of possibly overwriting a one transaction with another

cfromknecht: discussed offline, leave for follow up

PR #3790, opened 2019-12-04
multi: make SendToRoute consistent with other payment RPC

Roasbeef:
In this PR, we make the `SendToRoute` RPC call consistent with
all the other payment RPCs which will properly adhere to the current max
payment sat limit. This is a prep commit for the future wumbo soft cap
that will eventually land in lnd. Along the we also add some additional consistency 
checks to ensure our externally advertised `MaxHTLC` routing policy is consistent 
with this value.

### DIFF
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -2359,6 +2359,9 @@ func (f *fundingManager) annAfterSixConfs(completeChan *channeldb.OpenChannel,
 		if fwdMaxHTLC > capacityMSat {
 			fwdMaxHTLC = capacityMSat
 		}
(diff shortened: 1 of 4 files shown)


juscamarena:
I can understand limiting the SendToRoute from paying wumbo invoices, but why kill routing? The network has allowed it without issue and no one has complained ever. Eclair also allows the same and I just asked them and they're not adding a routing htlc max limit.

Bitrefill already has users paying us over the old payment limit with some of them being routed wumbo payments. . . Releasing this as the default will just kill this unless most node operators change their defaults. Why not wait until MPP is out then a change like this isn't such a big deal.

PR #317, opened 2017-09-05
Use the BOLT-11 invoice format within lnd

halseth:
This follows #268, which added the BOLT-11 parsing package. This PR moves this package into the `zpay32` namespace and makes use of it within `lnd`. These changes mainly affect the `lncli` command line tool, and the `rpcserver`, in addition to the protos these use to do RPC.

#### channeldb
A new field `PaymentRequest` is added to the `Invoice` struct in `channeldb`, to allow easy listing of the payment request created when the invoice was added. Adding this field involved less changes than adding all the fields necessary to deterministically recreate the payment request.

#### nodesigner
The `nodesigner` gets a new utility method for signing hashed data. This is needed to sign a single SHA-256 hash when creating the payment request, which was not supported before.

#### lnrpc
New BOLT-11 fields such as `expiry`, `timestamp`, `fallback_addr`,  and `description_hash` added to the protos.

#### rpcserver and lncli
Support for the proto changes and extra fields added with the new format.

**Note:** BOLT-11 payment requests contain a mandatory `description` field. We reuse the `memo` argument already present in `lncli`/`lnrpc` for this purpose.

PR #3795, opened 2019-12-05
lncli+invoices: experimental key send mode

joostjager:
This PR implements the key send functionality that is described in #2455 in `lnd` as it is today. 

Throughout the year 2019, many PRs have been merged that together created a solid foundation to implement key send on. This is reflected in the size of this PR. The core changes to the invoice registry are relatively small and non-intrusive.

Key send isn't part of the Lightning spec yet. Therefore the preimage is sent as a custom record (in the type range >= 65536):

`KeySendRecord uint64 = 5482373484`

It is likely that this custom record will be deprecated eventually and replaced by either spontaneous sending as an [AMP payment](https://github.com/lightningnetwork/lightning-rfc/pull/658) or a an official key send payload record.

**What this means is that applications using the functionality as it is introduced in this PR may break when either the sender or receiver upgrades to a newer version of `lnd`.** 

As before, intermediate hops will support key send with any version of the software.

To enable receiving key send spontaneous payments, start `lnd` with the `--accept-key-send` flag. Without this flag set, spontaneous payments will be rejected.

Sending a spontaneous payment can be done via `lncli`:

`lncli sendpayment -d <dest> -a <amount> --final_cltv_delta 40 --key_send`

This PR is also the final step in enabling mainline `lnd` to send and receive chat messages (Whatsat).


lnrpc/invoicesrpc/utils.go
```diff
@@ -10,15 +11,42 @@ import (
...
+// decodePayReq decodes the invoice payment request if present. This is needed,
+// because not all information is stored in dedicated invoice fields. If there
+// is no payment request present, a dummy request will be returned. This can
+// happen with very old invoices and key send.
+func decodePayReq(invoice *channeldb.Invoice,
+	activeNetParams *chaincfg.Params) (*zpay32.Invoice, error) {
+
+	paymentRequest := string(invoice.PaymentRequest)
+	if paymentRequest == "" {
+		preimage := invoice.Terms.PaymentPreimage
```
Roasbeef: Do we also want to manually set the feature bit we'll use to signal this on the graph level (in the `NodeAnnouncement`)?

Roasbeef: This would be another way to differentiate this payment from other payments when looking at the set of received payments.

invoices/invoiceregistry.go
```diff
@@ -12,6 +14,11 @@ import (
 	"github.com/lightningnetwork/lnd/queue"
 )
 
+const (
+	// KeySendRecord is the custom record identifier for key send preimages.
+	KeySendRecord uint64 = 5482373484
```
Roasbeef: Perhaps this should instead live in the `record` package like all our "recognized" TLV types.

invoices/invoiceregistry.go
```diff
@@ -419,6 +430,55 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+// processKeySend just-in-time inserts an invoice if this htlc is a key send
+// htlc.
+func (i *InvoiceRegistry) processKeySend(ctx invoiceUpdateCtx,
+	hash lntypes.Hash) error {
+
+	preimageSlice, ok := ctx.customRecords[KeySendRecord]
+	if !ok {
+		return nil
+	}
+	// Cancel htlc is preimage is invalid.
```
Roasbeef: Style nit: missing new line above this.

invoices/invoiceregistry.go
```diff
@@ -419,6 +430,55 @@ func (i *InvoiceRegistry) LookupInvoice(rHash lntypes.Hash) (channeldb.Invoice,
...
+
+	// Don't accept zero preimages as those have a special meaning in our
+	// database for hodl invoices.
+	if preimage == channeldb.UnknownPreimage {
+		return errors.New("invalid key send preimage")
+	}
+
+	// Create placeholder invoice.
+	invoiceFeatures := lnwire.NewFeatureVector(
+		lnwire.NewRawFeatureVector(), lnwire.Features,
```
Roasbeef: See comment about setting a new feature bit here to indicate that this is actually a `keysend` payment. Alternatively, we could just add a new payment type field which stays consistent regardless of the underlying `keysend` implementation (AMP, or custom record, etc).

invoices/invoiceregistry.go
```diff
@@ -447,7 +509,22 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+
+	// Process key send if present. Do this outside of the lock, because
+	// AddInvoice obtains its own lock. This is no problem, because the
+	// operation is idempotent.
+	if i.cfg.AcceptKeySend {
+		err := i.processKeySend(updateCtx, rHash)
+		if err != nil {
+			debugLog(fmt.Sprintf("key send error: %v", err))
+
+			return &HodlEvent{
```
Roasbeef: So we'll just hold onto it if the pre-image doesn't match the payment hash? I think instead, we want to cancel it like any other HTLC.

invoices/invoiceregistry_test.go
```diff
@@ -672,14 +673,73 @@ func TestUnknownInvoice(t *testing.T) {
...
+	preimage := lntypes.Preimage{1, 2, 3}
+	hash := preimage.Hash()
+
+	keySendPayload := &mockPayload{
+		customRecords: map[uint64][]byte{
+			KeySendRecord: preimage[:],
+		},
+	}
+
+	// Try to settle invoice with an htlc that expires too soon.
```
Roasbeef: Left over comment from copy/paste?

cmd/lncli/commands.go
```diff
@@ -2285,14 +2292,32 @@ func sendPayment(ctx *cli.Context) error {
...
-	switch {
-	case ctx.IsSet("payment_hash"):
-		rHash, err = hex.DecodeString(ctx.String("payment_hash"))
-	case args.Present():
-		rHash, err = hex.DecodeString(args.First())
-		args = args.Tail()
-	default:
-		return fmt.Errorf("payment hash argument missing")
+	if ctx.Bool("key_send") {
+		if ctx.IsSet("payment_hash") {
```
Roasbeef: Similarly, error out of the dest isn't set.

cmd/lncli/commands.go
```diff
@@ -2285,14 +2292,32 @@ func sendPayment(ctx *cli.Context) error {
...
+			return errors.New("cannot set payment hash when using " +
+				"key send")
+		}
+		var preimage lntypes.Preimage
+		if _, err := rand.Read(preimage[:]); err != nil {
+			return err
+		}
+
+		req.DestCustomRecords = map[uint64][]byte{
+			invoices.KeySendRecord: preimage[:],
```
Roasbeef: Re payload what about also adding a set of optional 32 bytes? This kind of nudges users towards an intended use cases of specifying some sort of account/payment ID which is to be looked up on the backend to accept/deny the payments.

invoices/invoiceregistry.go
Roasbeef: Or is this meant to allow a caller using the HODL invoice calls to _manually_ settle this?

invoices/invoiceregistry.go
```diff
@@ -427,25 +494,42 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
-		finalCltvRejectDelta: i.finalCltvRejectDelta,
-		customRecords:        payload.CustomRecords(),
+		finalCltvRejectDelta: i.cfg.FinalCltvRejectDelta,
+		customRecords:        customRecords,
+	}
+
+	// Process key send if present. Do this outside of the lock, because
+	// AddInvoice obtains its own lock. This is no problem, because the
+	// operation is idempotent.
+	if i.cfg.AcceptKeySend {
```
Roasbeef: Not a blocker, but would be nice to have an RPC mode similar to our `ChannelAcceptance` call that lets the caller examine the full payload set and reject/accept based on its contents. An example would be depositing to an exchange, they check they're actually expecting a deposit before they accept/pull the payment to avoid having to deal with refunds.

invoices/invoiceregistry_test.go
```diff
@@ -663,14 +673,73 @@ func TestUnknownInvoice(t *testing.T) {
 	}
 }
 
+// TestKeySend tests receiving a spontaneous payment.
+func TestKeySend(t *testing.T) {
```
Roasbeef: Missing a case for rejecting incoming keysend payments when we have the flag off.

cmd/lncli/commands.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/rand"
```
Roasbeef: Commit message doesn't match at all? Super bare in general.

invoices/invoiceregistry.go
```diff
@@ -453,6 +450,9 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
 		customRecords:        payload.CustomRecords(),
 	}
 
+	i.Lock()
```
Roasbeef: Would be nice to have rationale for this in the commit message body.

lnrpc/invoicesrpc/utils.go
```diff
@@ -2,6 +2,7 @@ package invoicesrpc
 
 import (
 	"encoding/hex"
+	"errors"
```
Roasbeef: We commit message, we don't have a `utils` folder or package.

lnrpc/invoicesrpc/utils.go
joostjager: My idea was to keep it minimal as long as it is experimental. The node announcement isn't very critical. You can just try to key send and see if it gets pulled.

The payment can be differentiated because there is no pay req, but I think that can also happen with very old payments. It was already an optional field before. Would you want to add a new (experimental) field on the invoices returns by for example `ListInvoices`?

invoices/invoiceregistry.go
joostjager: Yes, @cfromknecht and I spoke about that too. But it is another project.

invoices/invoiceregistry.go
joostjager: No, returning a `HodlEvent` without a preimage is an immediate cancel. So it behaves like you expect.

cmd/lncli/commands.go
joostjager: People can do that easily through the rpc interface if they want. I am not sure if we should make assumptions like that they require 32 bytes in `lncli`. Also, if we always use the same record id for that, different uses may interfere with each other.

lnrpc/invoicesrpc/utils.go
joostjager: The can also be recognized by the custom record that is present and exposed on invoices.

invoices/invoiceregistry.go
joostjager: Done. Reads nice: `record.KeySend`

cmd/lncli/commands.go
joostjager: Is already checked above

invoices/invoiceregistry_test.go
joostjager: Added

lnrpc/invoicesrpc/utils.go
joostjager: fixed

invoices/invoiceregistry.go
joostjager: Added

cmd/lncli/commands.go
joostjager: Yes, very bare, not good. Added more info.

joostjager:
Main open question is at what level(s) we want to support key send. My philosophy for this pr was to keep it as non-intrusive as possible. Postpone adding any new rpc fields until we reconcile with amp. What is in now, is enough for people to start building off of it without requiring a fork of `lnd`. I think that is good enough given the experimental status of it.

cmd/lncli/commands.go
```diff
@@ -2298,14 +2305,32 @@ func sendPayment(ctx *cli.Context) error {
...
+		if ctx.IsSet("payment_hash") {
+			return errors.New("cannot set payment hash when using " +
+				"key send")
+		}
+		var preimage lntypes.Preimage
+		if _, err := rand.Read(preimage[:]); err != nil {
+			return err
+		}
+
+		req.DestCustomRecords = map[uint64][]byte{
```
cfromknecht: should we have an integration test to assert this behavior?

config.go
```diff
@@ -325,6 +325,8 @@ type config struct {
 
 	EnableUpfrontShutdown bool `long:"enable-upfront-shutdown" description:"If true, option upfront shutdown script will be enabled. If peers that we open channels with support this feature, we will automatically set the script to which cooperative closes should be paid out to on channel open. This offers the partial protection of a channel peer diconnecting from us if cooperative close is attempted with a different script."`
 
+	EnableKeySend bool `long:"enable-key-send" description:"If true, spontaneous payments through key send will be accepted. [experimental]"`
```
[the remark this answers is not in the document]
joostjager: I considered that. `Accept` is more accurate imo and therefore I also use it internally. But I called it `Enable` here for consistency with the other `enable-` options.

invoices/invoiceregistry.go
```diff
@@ -611,6 +616,57 @@ func (i *InvoiceRegistry) cancelSingleHtlc(hash lntypes.Hash,
 	return nil
 }
 
+// processKeySend just-in-time inserts an invoice if this htlc is a key send
+// htlc.
+func (i *InvoiceRegistry) processKeySend(ctx invoiceUpdateCtx,
+	hash lntypes.Hash) error {
+
+	// Retrieve key send record if present.
+	preimageSlice, ok := ctx.customRecords[record.KeySend]
```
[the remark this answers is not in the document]
joostjager: I considered that, but didn't want to do the look up twice. Would you, with the external check, pass in the preimage slice? Also I tried to keep all the key send functionality together in this function. But I have no problem with changing it either.

invoices/invoiceregistry.go
joostjager: Added a commit that adds this rpc field

invoices/invoiceregistry.go
```diff
@@ -611,6 +616,57 @@ func (i *InvoiceRegistry) cancelSingleHtlc(hash lntypes.Hash,
...
+	// Create placeholder invoice.
+	invoiceFeatures := lnwire.NewFeatureVector(
+		lnwire.NewRawFeatureVector(), lnwire.Features,
+	)
+
+	// Use the minimum block delta that we require for settling htlcs.
+	finalCltvDelta := i.cfg.FinalCltvRejectDelta
+
+	invoice := &channeldb.Invoice{
+		CreationDate: time.Now(),
```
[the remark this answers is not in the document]

cmd/lncli/commands.go
joostjager: Do you mean an itest for `lncli`? Or an itest that tests the sender-receiver combination?

My view on itests is to only use them to cover things that cannot be covered in a unit test. It seems to me that every aspect of key send can be tested in isolation.

joostjager: Is there a particular interaction that you'd like to see covered?

cfromknecht: i think it'd be useful to cover the exact scenario that `lncli` feeds into the rpc, e.g. the correct custom records, dest, amt, payhash, etc

invoices/invoiceregistry.go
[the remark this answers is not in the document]
cfromknecht: agreed, the double lookup is def undesired, though it can be avoided by passing the slice as you described. for now i think it's fine to keep it this way, esp considering that the locking is more complicated since we need to JIT insert an invoice. perhaps just something to note on the road to a unified keysend via AMP.

config.go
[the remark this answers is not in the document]
cfromknecht: in most cases, i would tend to agree and like the unification. in this case, i think the distinction is that the ability to do keysends is irrespective of this flag, and so accept is the more accurate term. just my two cents

invoices/invoiceregistry.go
cfromknecht: no need to instantiate the raw feature vector, nil will suffice.

lnrpc/rpc.proto
```diff
@@ -2371,6 +2371,12 @@ message Invoice {
 
     /// List of HTLCs paying to this invoice [EXPERIMENTAL].
     repeated InvoiceHTLC htlcs = 22 [json_name = "htlcs"];
+
+    /**
+    Indicates if this invoice was a spontaneous payment that arrived via keysend
+    [EXPERIMENTAL].
+    */
+    bool key_send = 24 [json_name = "key_send"];
```
cfromknecht: naming suggestion: `is_keysend` to indicate boolean, but not a blocker

config.go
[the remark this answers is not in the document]
joostjager: updated to `AcceptKeySend`

lnrpc/rpc.proto
joostjager: I was thinking about the go convention of leaving off `is`. But changed.

invoices/invoiceregistry.go
joostjager: instantiated with nil

cmd/lncli/commands.go
joostjager: Added

lnrpc/invoicesrpc/utils.go
joostjager: The differentation part is solved now with a `is_key_send` rpc bool

joostjager:
With regards to server-side preimage generation: @cfromknecht and I realized that this would create the need for a pathway to communicate the payment hash to the caller. Seems best to go with external generation for now and rethink this in the amp context. The control tower offers some protection against non-randomly picked preimages.

joostjager:
Added code to handle mpp key send payments. Still need to update `lncli --keysend` to generate a payment address. Waiting for #3679 which I expect will introduce a `payment_addr` rpc field for `SendPayment`.

joostjager:
Reverted changes for mpp key send. It seems better not to support this for this release. LND 0.9 will probably default to single shot mpp payments, but by passing in a feature vector with send payment we can force an old-style payment for key send.

cmd/lncli/commands.go
```diff
@@ -2196,10 +2190,6 @@ var sendPaymentCommand = cli.Command{
 			Name:  "payment_hash, r",
 			Usage: "the hash to use within the payment's HTLC",
 		},
-		cli.BoolFlag{
```
cfromknecht: 👍

lnrpc/invoicesrpc/utils.go
```diff
@@ -11,15 +12,42 @@ import (
 	"github.com/lightningnetwork/lnd/zpay32"
 )
 
+// decodePayReq decodes the invoice payment request if present. This is needed,
+// because not all information is stored in dedicated invoice fields. If there
+// is no payment request present, a dummy request will be returned. This can
+// happen with very old invoices and key send.
```
cfromknecht: what is it that can happen with old invoices? is it possible for an old invoice to have a zero length payment request, or something else?

record/experimental.go
```diff
@@ -0,0 +1,6 @@
+package record
+
+const (
+	// KeySend is the custom record identifier for key send preimages.
+	KeySend uint64 = 5482373484
```
cfromknecht: nit: should be `KeySendType` to match everything else

invoices/invoiceregistry.go
```diff
@@ -631,16 +687,15 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
-	defer i.Unlock()
-
 	mpp := payload.MultiPath()
 
 	debugLog := func(s string) {
 		log.Debugf("Invoice(%x): %v, amt=%v, expiry=%v, circuit=%v, "+
 			"mpp=%v", rHash[:], s, amtPaid, expiry, circuitKey, mpp)
 	}
 
+	customRecords := payload.CustomRecords()
```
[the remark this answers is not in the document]
joostjager: removed

record/experimental.go
joostjager: fixed

lnrpc/invoicesrpc/utils.go
joostjager: I went back in history and found the introduction of the invoice payment request field in PR https://github.com/lightningnetwork/lnd/pull/317. From the change set I get that migrations weren't done yet at that point. The subsequent release https://github.com/lightningnetwork/lnd/releases/tag/v0.4-beta also describes this.

My conclusion is therefore that we can assume that the payment request is always present.

Comment updated.

joostjager:
Should we bump the db version? The empty payment request isn't backwards compatible

joostjager:
Discussed version bump offline. Same line of thought as before: 0.9 will already include a version bump and we don't support master downgrades. So no changes needed.

joostjager:
Rebased but itest fails because we cannot set `nil` `dest_features` to indicate that the graph features should be used.

invoices/invoiceregistry.go
```diff
@@ -661,6 +666,68 @@ func (i *InvoiceRegistry) cancelSingleHtlc(hash lntypes.Hash,
...
+		return errors.New("invalid key send preimage")
+	}
+
+	// Only allow key send for non-mpp payments.
+	if ctx.mpp != nil {
+		return errors.New("no mpp key send supported")
+	}
+
+	// Create an invoice for the htlc amount.
+	amt := ctx.amtPaid
```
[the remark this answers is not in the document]
joostjager: Filling up the harddrive is possible anyway because we also store data for a failed htlc.

lntest/itest/lnd_single_hop_invoice_test.go
```diff
@@ -0,0 +1,177 @@
+// +build rpctest
```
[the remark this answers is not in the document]
joostjager: That is the scheme that was started. It comes from `lnd_test.go`, but perhaps it isn't so great. If we want to rename it, I would tack on a commit right before merge. To prevent rebases over moves. But as this PR is open for a while already and it isn't really in scope, I'd rather leave it for another time.

joostjager:
Nil `dest_features` pr is merged, key send is operational again

lntest/itest/lnd_single_hop_invoice_test.go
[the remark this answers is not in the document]
cfromknecht: sgtm

invoices/invoiceregistry.go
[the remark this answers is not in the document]
cfromknecht: also true, will defer to a later time perhaps

joostjager:
Itest was failing because I happened to choose the same preimage for the key send test that was in a later test for a hodl invoice. Fixed now.

PR #3796, opened 2019-12-05
Add Upfront Shutdown Address to OpenChannelRequest

### LINKED ISSUE #3786: Set Upfront Shutdown Address via lnrpc/lncli
Once #3655 is merged, lnd will support the `option_upfront_shutdown_address` feature which allows nodes to specify an address that a channel should be paid out to upon cooperative close. If the peer we open a channel with supports the feature and we choose to set this address, the peer will not cooperatively close out to an address other than the one provided on channel open. 
Further details are provided in [BOLT2](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#the-open_channel-message).

### Steps to Completion

- [ ] Add a `shutdown_address` to lnrpc/lncli's `OpenChanne` calls
- [ ] Decode the address using`btcutil.DecodeAddress(addr, activeNetParams.Params)`
- [ ] Create a pay to address script using `txscript.PayToAddrScript(addr)` and set the value on `openChanReq.ShutdownScript`
- [ ] [Optional] expose the feature bits that our peers support via rpc so that users know when they can set an upfront address (the OpenChannel call will fail if you try to set an upfront address for a channel with a peer that does not support the feature)

carlaKC:
This PR adds an upfront shutdown address to lnrpc and lncli. Cooperative close out to these addresses is enforced provided that the remote peer supports the feature bit. The request will fail if an address is set for a peer that does not support upfront shutdown.

It also exposes the features that peers support in a new rpc endpoint `GetFeatures` so that users can check whether the peer supports the feature before setting the field (and check which features their node supports, if desired), which has the following output:
```
{
    "features": [
        {
            "bit": 5,
            "name": "upfront-shutdown-script",
            "is_required": false,
            "is_known": true
        },
        {
            "bit": 7,
            "name": "gossip-queries",
            "is_required": false,
            "is_known": true
        },
        {
            "bit": 9,
            "name": "tlv-onion",
            "is_required": false,
            "is_known": true
        },
        {
            "bit": 13,
            "name": "static-remote-key",
            "is_required": false,
            "is_known": true
        },
        {
            "bit": 0,
            "name": "data-loss-protect",
            "is_required": true,
            "is_known": true
        }
    ]
}
```

Fixes #3786


carlaKC:
While working on a first iteration of this PR which had feature bits in `ListPeers` I picked up a race condition in our itests between connecting to a peer and `handleInitMessage` because `remoteFeatures` are only set on receipt of `init` so tests would panic on calls to `ListPeers` which had nil `remoteFeatures`. Not going to be a real world issue, just making a note of it here.

carlaKC:
Rebased on #3824 to use common method.

feature/manager.go
```diff
@@ -95,3 +95,14 @@ func (m *Manager) Get(set Set) *lnwire.FeatureVector {
 	raw := m.GetRaw(set)
 	return lnwire.NewFeatureVector(raw, lnwire.Features)
 }
+
+// List returns a list of the feature vectors for our node.
+func (m *Manager) List() []*lnwire.FeatureVector {
+	var vectors []*lnwire.FeatureVector
+
+	for set := range m.fsets {
+		vectors = append(vectors, m.Get(set))
```
cfromknecht: another option here is to add `feature.SetList`, and then add this in each feature's set descriptor. might simplify the rpc conversion for self node

rpcserver.go
```diff
@@ -2278,6 +2282,68 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+
+		return &lnrpc.GetFeaturesResponse{
+			Features: features,
+		}, nil
+	}
+
+	var featureVector *lnwire.FeatureVector
+
+	// Try to find the peer in our list of peers.
+	for _, serverPeer := range r.server.Peers() {
```
cfromknecht: would [server.FindPeer](https://github.com/lightningnetwork/lnd/blob/master/server.go#L2386) work here?

rpcserver.go
```diff
@@ -2278,6 +2282,68 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+		pubKey := serverPeer.PubKey()
+		if bytes.Equal(in.PubKey, pubKey[:]) {
+			featureVector = serverPeer.remoteFeatures
+			break
+		}
+	}
+
```
cfromknecht: we could also try to find the node in our graph, and Merge any features we find there

lnrpc/rpc.proto
```diff
@@ -1643,6 +1643,18 @@ message OpenChannelRequest {
...
+    Upfront address is an optional address which specifies the address to which
+    funds should be paid out to upon cooperative close. This field may only be
+    set if the peer supports the option upfront feature bit (call getfeatures
+    with the node pubkey to check). The remote peer will only accept cooperative
+    closes to this address if it is set.
+
+    Note: If this value is set on channel creation, you will *not* be able to
+    cooperatively close out to a different address.
+    */
+    string upfront_address = 13[json_name = "upfront_address"];
```
cfromknecht: while we're here, should we also display the shutdown addresses in `listchannels`?

rpcserver.go
```diff
@@ -2278,6 +2312,68 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+// GetFeatures returns the features that a peer supports.
+func (r *rpcServer) GetFeatures(ctx context.Context,
+	in *lnrpc.GetFeaturesRequest) (*lnrpc.GetFeaturesResponse, error) {
+
+	if len(in.PubKey) == 0 {
+		return nil, fmt.Errorf("pubkey must be provided")
+	}
+
+	// If the query is for our node, get our set of enabled features.
+	if bytes.Equal(in.PubKey, r.selfNode[:]) {
```
cfromknecht: we could also consider adding these features to `getinfo` so what we don't have to query with our own pubkey

cfromknecht:
also note that #3837 proposes changing the rpc features to a map

carlaKC:
> also note that #3837 proposes changing the rpc features to a map

Given the addition of features to `ListPeers` and `GetNode` in #3837, I'm just going to add our own features to `GetInfo`, since peer features are sufficiently exposed.

lnrpc/rpc.proto
```diff
@@ -1517,6 +1517,9 @@ message GetInfoResponse {
 
     // Whether we consider ourselves synced with the public channel graph.
     bool synced_to_graph = 18 [json_name = "synced_to_graph"];
+
+    /// Features we advertised in our init message.
+    map<uint32, Feature> features = 19 [ json_name = "features"];
```
wpaulino: Nit: extra space after `[`.

carlaKC:
Rebased to address merge conflict 👮‍♀

lnrpc/rpc.proto
```diff
@@ -1517,6 +1517,9 @@ message GetInfoResponse {
 
     // Whether we consider ourselves synced with the public channel graph.
     bool synced_to_graph = 18 [json_name = "synced_to_graph"];
+
+    /// Features we advertised in our init message.
```
cfromknecht: nit: this is the union of all features advertised in init messages, node announcments, and invoices

cfromknecht:
@carlaKC needs rebase

PR #3797, opened 2019-12-05
channeldb: inject clock into database

joostjager:
This PR fixes a leftover of the mpp invoice registry pr #3415. There we took a shortcut to expose `Now` as a public field on the database in order to unblock the stack of dependent prs. This is now converted to a proper `clock.Clock` dependency.

### DIFF
--- a/channeldb/db.go
+++ b/channeldb/db.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/lightningnetwork/lnd/channeldb/migration12"
 	"github.com/lightningnetwork/lnd/channeldb/migration_01_to_11"
+	"github.com/lightningnetwork/lnd/clock"
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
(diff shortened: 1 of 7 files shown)


channeldb/db.go
```diff
@@ -142,7 +142,9 @@ type DB struct {
 
 // Open opens an existing channeldb. Any necessary schemas migrations due to
 // updates will take place as necessary.
-func Open(dbPath string, modifiers ...OptionModifier) (*DB, error) {
+func Open(dbPath string, now func() time.Time, modifiers ...OptionModifier) (
```
cfromknecht: why not set `time.Now` as the default in `DefaultOptions()`, then add a functional modifier? much smaller diff

bhandras:
Just saw this. Now that we have clock.Clock you might as well use that instead of the now closure.

channeldb/db.go
joostjager: Great suggestion, done

PR #3799, opened 2019-12-05
Bugfix: Chanfitness open time for offline peers

carlaKC:
This PR fixes a bug in the `chanfitness` subsystem which would not set the open time for channels with peers that are offline at startup. This open time (renamed to monitored since in this PR) is important because it indicates how long we have been monitoring the remote peer's offline status. 

Previously, open time would be set on the first event for a channel. For peers that are online at startup, this event would be immediately added to reflect the current state of the peer. Channels with peers that remain offline and do not have an event will not have their opened time set. 

This PR fixes this by setting `monitoredSince` for all channels on startup. 
It also switches `chanfitness` over to querying by channel outpoint, as this is preferred in lnd. 

This bug was first picked up in #3741 but I'd like to get the fix in while the form that 3741 should take is decided on.

### DIFF
--- a/chanfitness/chanevent.go
+++ b/chanfitness/chanevent.go
@@ -60,14 +60,18 @@ type chanEventLog struct {
 	closedAt time.Time
 }
 
-func newEventLog(outpoint wire.OutPoint, peer route.Vertex,
+// newEventLog creates an event log for a channel with the openedAt time set.
+func newEventLog(channelPoint wire.OutPoint, peer route.Vertex,
 	now func() time.Time) *chanEventLog {
 
-	return &chanEventLog{
-		channelPoint: outpoint,
+	eventlog := &chanEventLog{
+		channelPoint: channelPoint,
 		peer:         peer,
 		now:          now,
+		openedAt:     now(),
 	}
+
+	return eventlog
 }
 
 // close sets the closing time for an event log.
@@ -90,13 +94,6 @@ func (e *chanEventLog) add(eventType eventType) {
 		eventType: eventType,
 	}
 	e.events = append(e.events, event)
-
-	// If the eventLog does not have an opened time set, set it to the timestamp
-	// of the event. This has the effect of setting the eventLog's open time to
-	// the timestamp of the first event added.
-	if e.openedAt.IsZero() {
-		e.openedAt = event.timestamp
-	}
 
 	log.Debugf("Channel %v recording event: %v", e.channelPoint, eventType)
 }
--- a/chanfitness/chanevent.go
+++ b/chanfitness/chanevent.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/btcsuite/btcd/wire"
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
@@ -36,8 +37,8 @@ type channelEvent struct {
 
 // chanEventLog stores all events that have occurred over a channel's lifetime.
 type chanEventLog struct {
-	// id is the uint64 of the short channel ID.
-	id uint64
+	// channelPoint is the outpoint for the channel's funding transaction.
+	channelPoint wire.OutPoint
 
 	// peer is the compressed public key of the peer being monitored.
 	peer route.Vertex
@@ -59,11 +60,13 @@ type chanEventLog struct {
 	closedAt time.Time
 }
 
-func newEventLog(id uint64, peer route.Vertex, now func() time.Time) *chanEventLog {
+func newEventLog(outpoint wire.OutPoint, peer route.Vertex,
+	now func() time.Time) *chanEventLog {
+
 	return &chanEventLog{
-		id:   id,
-		peer: peer,
-		now:  now,
+		channelPoint: outpoint,
+		peer:         peer,
+		now:          now,
 	}
 }
 
@@ -95,7 +98,7 @@ func (e *chanEventLog) add(eventType eventType) {
 		e.openedAt = event.timestamp
 	}
 
-	log.Debugf("Channel %v recording event: %v", e.id, eventType)
+	log.Debugf("Channel %v recording event: %v", e.channelPoint, eventType)
 }
 
 // onlinePeriod represents a period of time over which a peer was online.
--- a/chanfitness/chaneventstore_test.go
+++ b/chanfitness/chaneventstore_test.go
@@ -466,3 +466,26 @@ func TestGetUptime(t *testing.T) {
 		})
 	}
 }
+
+// TestAddChannel tests that channels are added to the event store with
+// appropriate timestamps. This test addresses a bug where offline channels
+// did not have an opened time set.
+func TestAddChannel(t *testing.T) {
+	_, vertex, chanPoint := getTestChannel(t)
+
+	store := NewChannelEventStore(&Config{})
+
+	// Add channel to the store.
+	store.addChannel(chanPoint, vertex)
+
+	// Check that the eventlog is successfully added.
+	eventlog, ok := store.channels[chanPoint]
+	if !ok {
+		t.Fatalf("channel should be in store")
+	}
+
+	// Ensure that open time is always set.
+	if eventlog.openedAt.IsZero() {
+		t.Fatalf("channel should have opened at set")
+	}
+}
--- a/chanfitness/chaneventstore_test.go
+++ b/chanfitness/chaneventstore_test.go
@@ -263,23 +263,23 @@ func TestGetLifetime(t *testing.T) {
 	now := time.Now()
 
 	tests := []struct {
-		name         string
-		channelFound bool
-		chanID       uint64
-		opened       time.Time
-		closed       time.Time
-		expectErr    bool
+		name          string
+		channelFound  bool
+		chanID        uint64
+		opened        time.Time
+		closed        time.Time
+		expectedError error
 	}{
 		{
-			name:         "Channel found",
-			channelFound: true,
-			opened:       now,
-			closed:       now.Add(time.Hour * -1),
-			expectErr:    false,
+			name:          "Channel found",
+			channelFound:  true,
+			opened:        now,
+			closed:        now.Add(time.Hour * -1),
+			expectedError: nil,
 		},
 		{
-			name:      "Channel not found",
-			expectErr: true,
+			name:          "Channel not found",
+			expectedError: ErrChannelNotFound,
 		},
 	}
 
@@ -311,11 +311,8 @@ func TestGetLifetime(t *testing.T) {
 			}
 
 			open, close, err := store.GetLifespan(test.chanID)
-			if test.expectErr && err == nil {
-				t.Fatal("Expected an error, got nil")
-			}
-			if !test.expectErr && err != nil {
-				t.Fatalf("Expected no error, got: %v", err)
+			if test.expectedError != err {
+				t.Fatalf("Expected: %v, got: %v", test.expectedError, err)
 			}
 
 			if open != test.opened {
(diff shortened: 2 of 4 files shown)


chanfitness/chanevent.go
```diff
@@ -60,14 +60,36 @@ type chanEventLog struct {
...
+	eventlog := &chanEventLog{
 		fundingTxID: fundingTxID,
 		peer:        peer,
 		now:         now,
+		openedAt:    openedAt,
+	}
+
+	// If the peer is online at startup, add a peer online event with the same
+	// timestamp as the event log opening time.
+	if peerOnline {
```
joostjager: I would think that just setting `openedAt` here is enough to fix the bug? Or is the logging of the first event right here also required?

chanfitness/chanevent.go
```diff
@@ -36,8 +37,8 @@ type channelEvent struct {
 
 // chanEventLog stores all events that have occurred over a channel's lifetime.
 type chanEventLog struct {
-	// id is the uint64 of the short channel ID.
-	id uint64
+	// fundingTxID is the funding transaction for the channel.
+	fundingTxID wire.OutPoint
```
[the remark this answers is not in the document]
halseth: Yeah, `fundingPoint`, `channelPoint` or `chanPoint` is more common.

chanfitness/chanevent.go
carlaKC: Adding an online event on startup for online peers is not new behaviour, it just used to be separated into creating the event log then adding the event. This is required to have a starting state for online calculations (ie, knowing whether we started with the peer online or offline). If we do not log this event at the same time as the channel opening, all channels that exit at startup will present as if they had a brief offline period.

I will update the comment to be more clear.

joostjager: >all channels that exit at startup will present as if they had a brief offline period

Is this worth adding the alternative code path for? Is 100% online even possible at all? When `lnd` starts, not all channels are up all of a sudden. Maybe it is even more realistic to keep it disconnected.

carlaKC: They aren't all up, but if the peer is online, they are being a good channel peer (which is what we are measuring), regardless of what our current state is. If/When we start to calculate peer flaps based on the number of offline periods the peer has, this will falsely report an extra flap for the peer as well. I think it's better to have some way of identifying that this is a different scenario. 

I could record an event regardless of state, if that makes it more consistent? Or add a bool on the eventLog, but I think that adds more code than this approach.

joostjager: Ok, so we are measuring uptime excluding our own down time? When this is moved to persistent storage, there will not be a off and on event when we shut down `lnd` ourselves?

I am fine with what you have here, it is not very important.

carlaKC: Yes, I think the goal here is to measure peer uptime, since our downtime is the same for all channels.

chanfitness/chaneventstore_test.go
```diff
@@ -100,10 +96,20 @@ func TestMonitorChannelEvents(t *testing.T) {
 		t.Fatalf("error creating txid: %v", err)
```
joostjager: Random is non-deterministic, ideally not used in unit tests

chanfitness/chaneventstore_test.go
```diff
@@ -100,10 +96,20 @@ func TestMonitorChannelEvents(t *testing.T) {
...
 	}
 
-	chanPoint := wire.OutPoint{
+	return privKey.PubKey(), pubKey, wire.OutPoint{
```
joostjager: Random is non-deterministic, ideally not used in unit tests.

chanfitness/chaneventstore_test.go
```diff
@@ -466,3 +472,59 @@ func TestGetUptime(t *testing.T) {
...
+				t.Fatalf("channel should have opened at set")
+			}
+
+			// Check that an initial event has been added, where appropriate.
+			for i, event := range test.expctedEvents {
+				if eventlog.events[i].eventType != event.eventType {
+					t.Fatalf("expected: %v, got: %v", event.eventType,
+						eventlog.events[i].eventType)
+				}
+			}
```
joostjager: No uptime assertion, which was what the bug was about?

chanfitness/chaneventstore_test.go
```diff
@@ -93,10 +94,15 @@ func TestMonitorChannelEvents(t *testing.T) {
...
 
-	shortID := lnwire.ShortChannelID{
-		BlockHeight: 1234,
-		TxIndex:     2,
-		TxPosition:  2,
+	// Create test outpoint.
+	var txid [32]byte
+	if _, err := rand.Read(txid[:]); err != nil {
+		t.Fatalf("error creating txid: %v", err)
+	}
```
joostjager: Would pull the later refactor to create a test channel to this commit, to prevent touch this code twice.

rpcserver.go
```diff
@@ -2798,12 +2799,10 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
...
 
 		uptime, err = r.server.chanEventStore.GetUptime(
-			chanID,
-			startTime,
-			endTime,
+			dbChannel.FundingOutpoint, startTime, endTime,
 		)
 		if err != nil {
-			rpcsLog.Warnf("GetUptime %v error: %v", chanID, err)
+			rpcsLog.Warnf("GetUptime %v error: %v", dbChannel.FundingOutpoint, err)
```
joostjager: Is every error only a warning? Or only the channel not found error?

carlaKC: This can probably be upgraded to an error, if we get channel not found then there's a bug somewhere because all the channels returned by `listchannels` should be tracked in chafitness. 

Likewise, the other errors would be from incorrect the start/end times, which is also a bug.

chanfitness/chaneventstore_test.go
carlaKC: The bug was that we had a zero open time for channels that started with offline peers, which is checked above. The effect of a zero open time was that lifespan would be calculated as `endtime -0` instead of `endtime - openedAt`.

chanfitness/chaneventstore_test.go
carlaKC: nice, will do!

chanfitness/chaneventstore_test.go
joostjager: Ah right. I was thinking about the tiny gap between init and the first event, but probably hardly measurable

rpcserver.go
```diff
@@ -2776,43 +2779,41 @@ func createRPCOpenChannel(r *rpcServer, graph *channeldb.ChannelGraph,
...
-		// further calculations for uptime and lifespan.
-		rpcsLog.Warnf("GetLifespan %v error: %v", chanID, err)
-	} else {
-		// If endTime is zero, the channel is still open, progress endTime to
-		// the present so we can calculate lifespan.
-		if endTime.IsZero() {
-			endTime = time.Now()
-		}
-		lifespan = endTime.Sub(startTime)
+		return nil, fmt.Errorf("GetLifespan %v error: %v",
```
joostjager: This is an independent change that could be an isolated commit. Not for me necessarily, but perhaps the second reviewer.

chanfitness/chanevent.go
```diff
@@ -59,12 +60,40 @@ type chanEventLog struct {
...
-	return &chanEventLog{
-		id:   id,
-		peer: peer,
-		now:  now,
+// newEventLog creates an event log for a channel. If the peer is online at
+// the time the event log is created, add an online event with the same
+// timestamp as the event log so that the event log does not have a gap
+// between open time and the online event in the case where the peer was
+// already online.
+func newEventLog(fundingTxID wire.OutPoint, peer route.Vertex,
```
joostjager: still `fundingTxID`

chanfitness/chaneventstore_test.go
```diff
@@ -93,11 +90,25 @@ func TestMonitorChannelEvents(t *testing.T) {
...
-	shortID := lnwire.ShortChannelID{
-		BlockHeight: 1234,
-		TxIndex:     2,
-		TxPosition:  2,
+	return privKey.PubKey(), pubKey, wire.OutPoint{
+		Hash: [chainhash.HashSize]byte{
+			0x51, 0xb6, 0x37, 0xd8, 0xfc, 0xd2, 0xc6, 0xda,
+			0x48, 0x59, 0xe6, 0x96, 0x31, 0x13, 0xa1, 0x17,
+			0x2d, 0xe7, 0x93, 0xe4, 0xb7, 0x25, 0xb8, 0x4d,
+			0x1f, 0xb, 0x4c, 0xf9, 0x9e, 0xc5, 0x8c, 0xe9,
```
joostjager: Could also be: `chainhash.Hash{1,2,3}`

carlaKC:
Nits done, error check moved into a separate commit.

joostjager:
Nice. Will hold off a final round until the second reviewer has taken a look.

chanfitness/chanevent.go
halseth: When we start the peer notifier, we'll get the peer online event again I guess. Does this mean we will have two consecutive `online` events for each peer in the beginning of the log?

carlaKC: It does. I thought we were already listening for online events at this stage so it would be missed, but that is not the case. I'm going to remove this online start event, since we will record it once we start consuming peer online events. There will be a small gap but it looks a lot cleaner doing it this way, so I think it's worth living with it.

halseth: I think that is fine for now. If it turns out we cannot live with the gap, we can address it later 😆

carlaKC:
> Please re-request review when @halseth has looked at this PR

@joostjager this is ready for another look :)

carlaKC:
> Please re-request review when @halseth has looked at this PR

@joostjager this is ready for another look :)

rpcserver.go
```diff
@@ -2938,8 +2939,12 @@ func (r *rpcServer) SubscribeChannelEvents(req *lnrpc.ChannelEventSubscription,
 			var update *lnrpc.ChannelEventUpdate
 			switch event := e.(type) {
 			case channelnotifier.OpenChannelEvent:
-				channel := createRPCOpenChannel(r, graph,
+				channel, err := createRPCOpenChannel(r, graph,
 					event.Channel, true)
+				if err != nil {
```
joostjager: There can't be a race condition here? That the fitness subsystem isn't ready yet / hasn't processed the open channel event and returns an err?

chanfitness/chanevent.go
```diff
@@ -60,14 +60,18 @@ type chanEventLog struct {
...
-	return &chanEventLog{
-		channelPoint: outpoint,
+	eventlog := &chanEventLog{
+		channelPoint: channelPoint,
 		peer:         peer,
 		now:          now,
+		openedAt:     now(),
 	}
+
+	return eventlog
```
joostjager: What happened to setting the initial event here with the same timestamp? It is still described in the commit message.

carlaKC: I've decided to remove it, rather have channels that exist at startup have a small gap between monitored time and online time than have the bug prone code path (it would also sometimes result in 2x consecutive online events).

rpcserver.go
carlaKC: I think it does, I suspect this is why travis is failing. 

I am going to export the channel not found error and only error in other cases.

joostjager:
Commits look out of order on github. Do you use that delay rewrite script to make sure they are all spaced at least 1 second apart? (it is crazy indeed...)

rpcserver.go
joostjager: Ok, sounds safe. Maybe in the future, this uptime stat will be moved somewhere else anyway.

halseth:
Did look through the changes, still LGTM :)

carlaKC:
> Commits look out of order on github. Do you use that delay rewrite script to make sure they are all spaced at least 1 second apart? (it is crazy indeed...)

🤦‍♀ cannot believe this is an open github issue. 

fixed up to have commits in order

PR #3802, opened 2019-12-05
rpc: expose invoice features in decodepayreq

cfromknecht:
Split off from #3679

Now that all invoices have been migrated to have features vectors (#3693),
we can expose this info via the rpc interface. Invoices generated prior to
this commit will show an empty list, but new invoices will set
`lnwire.TLVOnionPayloadOptional` as configured in the default
`feature.Manager`.

Note that for now this feature bit will be ignored by the router, but this will
change once #3679 is finalized.


lnrpc/invoicesrpc/addinvoice.go
```diff
@@ -47,6 +47,10 @@ type AddInvoiceConfig struct {
 	// ChanDB is a global boltdb instance which is needed to access the
 	// channel graph.
 	ChanDB *channeldb.DB
+
+	// GenInvoiceFeatures returns a feature containing feature bits that
+	// should be advertised on freshly generated invoices.
+	GenInvoiceFeatures func() *lnwire.FeatureVector
```
Roasbeef: Any reason this can't just be a field rather than a function closure?

lnwire/features.go
```diff
@@ -362,9 +361,9 @@ func (fv *FeatureVector) UnknownRequiredFeatures() []FeatureBit {
 func (fv *FeatureVector) Name(bit FeatureBit) string {
 	name, known := fv.featureNames[bit]
 	if !known {
-		name = "unknown"
+		return "unknown"
```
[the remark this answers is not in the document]
cfromknecht: the bit is exposed separately in the rpc, wouldn't it be redundant?

lnrpc/rpc.proto
```diff
@@ -2616,6 +2616,13 @@ message PayReq {
     int64 cltv_expiry = 9 [json_name = "cltv_expiry"];
     repeated RouteHint route_hints = 10 [json_name = "route_hints"];
     bytes payment_addr = 11 [json_name = "payment_addr"];
+    repeated Feature features = 12 [json_name = "features"];
+}
+
+message Feature {
+        uint32 bit = 1 [json_name = "bit"];
+        string name = 2 [json_name = "name"];
+        bool is_required = 3 [json_name = "is_required"];
```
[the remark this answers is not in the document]
cfromknecht: added

lnrpc/invoicesrpc/addinvoice.go
cfromknecht: i wanted to make sure that there was no possibility of mutation, so this allows each invoice to have a distinct feature vector. we could just used the `Clone()` method, but in the future it's likely that this set will be dynamic depending on the type of invoice being generated, so i thought it made sense to make it a function

rpcserver.go
```diff
@@ -4599,6 +4599,20 @@ func (r *rpcServer) DecodePayReq(ctx context.Context,
...
+	// for the rpc.
+	invFeatures := payReq.Features.Features()
+	features := make([]*lnrpc.Feature, 0, len(invFeatures))
+	for bit := range invFeatures {
+		name := payReq.Features.Name(bit)
+		features = append(features, &lnrpc.Feature{
+			Bit:        uint32(bit),
+			Name:       name,
+			IsRequired: bit.IsRequired(),
+			IsUnknown:  name == "unknown",
```
halseth: Better to use `IsKnown`?

cfromknecht: done

Roasbeef:
Needs rebase (just protos).

cfromknecht:
@Roasbeef rebased

PR #3806, opened 2019-12-06
lnrpc: add msats to return value of DecodePayReq

Roasbeef:
In this commit, we add `msats` to the return value of `DecodePayReq` to
ensure we always show full value information as we're moving to do
generally for all RPC calls that deal with off-chain amounts.


juscamarena:
wohoo, faster than I could create an issue.

PR #3809, opened 2019-12-09
sweep: allow force sweeping of negatively yielding inputs

joostjager:
This PR adds force sweep functionality to the sweeper. Force sweep signals that an input needs to be swept even if its yield is negative. This is useful in CPFP scenarios where the parent tx output is only small. This will be common in the upcoming new anchor outputs commitment format.

The force flag is exposed on the `BumpFee` rpc call and through `lncli`. In many cases it will be required to have a wallet utxo available to bring the tx output above the dust limit. PR #3814 extended the sweeper with the capability to attach wallet utxos if needed.


roeierez:
@joostjager Will this PR solve the use case where channel was closed with small local balance?
Right now I have ~2000 sat that can't be swept due to fee rate.
in https://github.com/lightningnetwork/lnd/pull/3814 there is an attempt to use additional input but I don't have any in my wallet.

joostjager:
With an input of that size, the idea is to modify the sweep fee rate to a lower value using the `bumpfee` rpc, so that it becomes economical to sweep. Without a wallet utxo, it still needs to produce a tx that has an output above the dust limit.

Note that the sweeper uses fee rate buckets to batch inputs together. The first bucket is 1-10 sat/byte. The actual sweep fee rate is the average of all inputs in a bucket. If your input is the only input in the 1-10 bucket and you've lowered the input fee rate preference to 1 sat/b, it should be swept at that rate.

This PR is not adding anything to that. It mainly allows negatively yielding inputs to be swept using the force flag, but for that you also need a wallet utxo.

joostjager:
Note to self: @cfromknecht suggested to require regular inputs to not just be positively yielding, but yield at least 2/3 (for example) of their value

joostjager:
Checked what is needed for that minimum yield requirement. The sweeper change is small, but there are quite a few changes in the test (setup) required. Leaving that change for a separate pr.

cmd/lncli/walletrpc_active.go
```diff
@@ -125,16 +132,27 @@ var bumpFeeCommand = cli.Command{
...
-	if ctx.NArg() != 1 || ctx.NumFlags() != 1 {
+	if ctx.NArg() != 1 {
 		return cli.ShowCommandHelp(ctx, "bumpfee")
 	}
+	if ctx.IsSet("conf_target") && ctx.IsSet("sat_per_byte") {
+		return errors.New("sat_per_byte and conf_target are " +
+			"mutually exclusive")
+	}
+	if !ctx.IsSet("conf_target") && !ctx.IsSet("sat_per_byte") {
+		return errors.New("must specify a fee preference")
```
wpaulino: Aren't these already checked at the RPC level?

sweep/sweeper.go
```diff
@@ -398,10 +402,12 @@ func (s *UtxoSweeper) SweepInput(input input.Input,
 	}
 
 	log.Infof("Sweep request received: out_point=%v, witness_type=%v, "+
-		"time_lock=%v, amount=%v, fee_preference=%v", input.OutPoint(),
+		"time_lock=%v, amount=%v, fee_preference=%v, "+
+		"force=%v",
+		input.OutPoint(),
```
wpaulino: Nit: looks like we can save on some whitespace by merging these lines with the above.

sweep/sweeper.go
```diff
@@ -67,6 +67,10 @@ type Params struct {
 	// swept. If a confirmation target is specified, then we'll map it into
 	// a fee rate whenever we attempt to cluster inputs for a sweep.
 	Fee FeePreference
+
+	// Force indicates whether the input should be swept regardless of
+	// whether it is economical to do so.
+	Force bool
```
wpaulino: Nit: include in `Params.String()`?

cmd/lncli/walletrpc_active.go
joostjager: Yes, that is true. They are. And if they weren't, they should. Removed all this validation code.

sweep/sweeper.go
joostjager: added

sweep/sweeper.go

joostjager:
Comments processed, ready for another pass

sweep/sweeper.go
```diff
@@ -390,10 +402,11 @@ func (s *UtxoSweeper) SweepInput(input input.Input,
 	}
 
 	log.Infof("Sweep request received: out_point=%v, witness_type=%v, "+
-		"time_lock=%v, amount=%v, fee_preference=%v", input.OutPoint(),
-		input.WitnessType(), input.BlocksToMaturity(),
+		"time_lock=%v, amount=%v, fee_preference=%v, "+
+		"force=%v",
```
wpaulino: Nit: the args can continue on this line.

sweep/sweeper.go
```diff
@@ -151,17 +156,17 @@ type PendingInput struct {
 	Params Params
 }
 
-// bumpFeeReq is an internal message we'll use to represent an external caller's
-// intent to bump the fee rate of a given input.
-type bumpFeeReq struct {
-	input         wire.OutPoint
-	feePreference FeePreference
-	responseChan  chan *bumpFeeResp
+// updateParametersReq is an internal message we'll use to represent an external
```
halseth: godoc name inconsistent

sweep/tx_input_set.go
```diff
@@ -108,19 +123,33 @@ func (t *txInputSet) add(input input.Input, fromWallet bool) bool {
...
 
-	// If adding this input makes the total output value of the set
-	// decrease, this is a negative yield input. We don't add the input to
-	// the set and return the outcome.
-	if newOutputValue <= t.outputValue {
-		return false
-	}
+	// Initialize new wallet total with the current wallet total. This is
+	// updated below if this input is a wallet input.
+	newWalletTotal := t.walletInputTotal
```
halseth: seems like the handling of this field can stay inside the switch case

cmd/lncli/walletrpc_active.go
```diff
@@ -142,24 +151,14 @@ func bumpFee(ctx *cli.Context) error {
 		return err
 	}
 
-	var confTarget, satPerByte uint32
```
halseth: why remove this validation?

sweep/tx_input_set.go
```diff
@@ -24,6 +24,10 @@ const (
 	// constraintsWallet is for wallet inputs that are only added to bring up the tx
 	// output value.
 	constraintsWallet
+
+	// constraintsForce is for inputs that should be swept even with a negative
+	// yield at the set fee rate.
+	constraintsForce
```
halseth: maybe not a use for it right now, but could be interesting to make this a bit field at some point st multiple constraints could be combined.

sweep/sweeper.go

sweep/tx_input_set.go
joostjager: I moved it outside, because otherwise you have to be very careful not to update `t.walletInputTotal` before you do any additional checks. In that case, the total would be updated without the add happening.

cmd/lncli/walletrpc_active.go
joostjager: The server already validates this, so this is just unnecessary maintenance. Change was suggested by @wpaulino and I liked it.

sweep/sweeper.go

sweep/tx_input_set.go
joostjager: maybe, but maybe wagni...

joostjager:
Holding off merge until 0.9 final is cut.

PR #3811, opened 2019-12-10
WIP: Surface Update Reason for Exit Hop HTLCs

carlaKC:
This PR updates `processExitHopHTLC` to take action based on the `updateResult` returned by `NotifyExitHopHTLC`. 

Reasons for the change:
1. More understandable handler code: as is, invoice errors such as invoice underpaid and expiry fall through to `processHODLEvent` and would present as if they were hodl invoice cancels (we log that we're cancelling a hodl event) and hodl acceptances are inferred by the presence of a nil hodl event, which is not clear on first read of the code.
2. Better unit tests: surfacing the update result allows for unit testing of specific outcomes
3. HTLCNotifier will be able to notify events with their specific failure reasons. 

Disadvantage of change:
1. Update to an interface which is widely used in lnd
2. Potential for bugs/behaviour change (further unit tests to follow after discussion of this change)

### DIFF
--- a/contractcourt/htlc_incoming_contest_resolver.go
+++ b/contractcourt/htlc_incoming_contest_resolver.go
@@ -201,7 +201,7 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 		HtlcID: h.htlc.HtlcIndex,
 	}
 
-	event, err := h.Registry.NotifyExitHopHtlc(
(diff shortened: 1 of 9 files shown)


carlaKC:
PR created to demonstrate an approach we're not going with - goal is to move invoice awareness out of the link entirely (this does the opposite).

PR #3812, opened 2019-12-10
signrpc: sign and verify messages with custom key 

guggero:
The existing `SignMessage` and `VerifyMessage` in the main `rpc.proto` file are very restrictive in that they only allow creating signatures over the node's identity private key and that a signature's recovered public key must correspond to a nodes' identity public key known to the network.
The new methods in the `signer.proto` don't have those restrictions and allow signing and verifying over custom wallet keys (limited to the `1017'` special purpose derivation branch).


lnrpc/signrpc/signer_server.go
```diff
@@ -383,3 +400,75 @@ func (s *Server) ComputeInputScript(ctx context.Context,
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("can't derive private key: %v", err)
+	}
+
+	// The signature is over the double-sha256 hash of the prefixed message.
+	in.Msg = append(signedMsgPrefix, in.Msg...)
+	digest := chainhash.DoubleHashB(in.Msg)
+
+	// Create the pubkey-recoverable signature using the compressed pubkey.
```
Roasbeef: I think for this, we just want a raw regular signature: so no compact encoding, and no `zbase32`, just a raw ECDSA signature.

lnrpc/signrpc/signer_server.go
```diff
@@ -383,3 +400,75 @@ func (s *Server) ComputeInputScript(ctx context.Context,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode signature: %v", err)
+	}
+
+	// The signature is over the double-sha256 hash of the prefixed message.
+	in.Msg = append(signedMsgPrefix, in.Msg...)
+	digest := chainhash.DoubleHashB(in.Msg)
+
+	// RecoverCompact both recovers the pubkey and validates the signature.
+	pubKey, valid, err := btcec.RecoverCompact(btcec.S256(), sig, digest)
```
Roasbeef: Similar message here, at least for the use case we have in mind.

guggero:
I switched to default ECDSA encoding but left the message prefixing in to avoid signing anything sensitive. This means that you can only verify signatures generated with the lnd signer which should be fine for our use case.

I've also given the `VerifyMessage` the `signer/read` permission instead of `signer/generate`.

joostjager:
I think `admin.macaroon` needs to be regenerated? In that case it may be good to add that to the release notes.

PR #3814, opened 2019-12-10
sweep: add wallet inputs to reach dust limit

joostjager:
This PR enables the sweeper to sweep inputs that are positively-yielding, but by itself wouldn't reach the dust limit for the sweep transaction output value at the chosen fee rate.

They way to deal with this is to attach additional wallet input(s) to bring up the output value of the transaction.

This is a preparation for sweeping anchor outputs that are too small to reach the dust limit of the sweep tx output.

There are also a few commits in this PR that aren't strictly necessary for the functionality that this PR adds, but are included anyway to prevent touching the code again in the follow-up.

Next step after this PR is to add forced sweeping of negatively yielding inputs.

### DIFF
--- a/sweep/interface.go
+++ b/sweep/interface.go
@@ -2,6 +2,7 @@ package sweep
 
 import (
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/lnwallet"
 )
 
 // Wallet contains all wallet related functionality required by sweeper.
@@ -9,4 +10,18 @@ type Wallet interface {
 	// PublishTransaction performs cursory validation (dust checks, etc) and
 	// broadcasts the passed transaction to the Bitcoin network.
 	PublishTransaction(tx *wire.MsgTx) error
+
+	// ListUnspentWitness returns all unspent outputs which are version 0
+	// witness programs. The 'minconfirms' and 'maxconfirms' parameters
+	// indicate the minimum and maximum number of confirmations an output
+	// needs in order to be returned by this method.
+	ListUnspentWitness(minconfirms, maxconfirms int32) ([]*lnwallet.Utxo,
+		error)
+
+	// WithCoinSelectLock will execute the passed function closure in a
+	// synchronized manner preventing any coin selection operations from
+	// proceeding while the closure if executing. This can be seen as the
+	// ability to execute a function closure under an exclusive coin
+	// selection lock.
+	WithCoinSelectLock(f func() error) error
 }
--- a/sweep/interface.go
+++ b/sweep/interface.go
@@ -1,0 +1,12 @@
+package sweep
+
+import (
+	"github.com/btcsuite/btcd/wire"
+)
+
+// Wallet contains all wallet related functionality required by sweeper.
+type Wallet interface {
+	// PublishTransaction performs cursory validation (dust checks, etc) and
+	// broadcasts the passed transaction to the Bitcoin network.
+	PublishTransaction(tx *wire.MsgTx) error
+}
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -599,27 +599,10 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 					inputClusters[j].sweepFeeRate
 			})
 			for _, cluster := range inputClusters {
-				// Examine pending inputs and try to construct
-				// lists of inputs.
-				inputLists, err := s.getInputLists(
-					cluster, bestHeight,
-				)
+				err := s.sweepCluster(cluster, bestHeight)
 				if err != nil {
-					log.Errorf("Unable to examine pending "+
-						"inputs: %v", err)
-					continue
-				}
-
-				// Sweep selected inputs.
-				for _, inputs := range inputLists {
-					err := s.sweep(
-						inputs, cluster.sweepFeeRate,
-						bestHeight,
-					)
-					if err != nil {
-						log.Errorf("Unable to sweep "+
-							"inputs: %v", err)
-					}
+					log.Errorf("input cluster sweep: %v",
+						err)
 				}
 			}
 
@@ -643,6 +626,27 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 			return
 		}
 	}
+}
+
+// sweepCluster tries to sweep the given input cluster.
+func (s *UtxoSweeper) sweepCluster(cluster inputCluster,
+	currentHeight int32) error {
+
+	// Examine pending inputs and try to construct lists of inputs.
+	inputLists, err := s.getInputLists(cluster, currentHeight)
+	if err != nil {
+		return fmt.Errorf("unable to examine pending inputs: %v", err)
+	}
+
+	// Sweep selected inputs.
+	for _, inputs := range inputLists {
+		err := s.sweep(inputs, cluster.sweepFeeRate, currentHeight)
+		if err != nil {
+			return fmt.Errorf("unable to sweep inputs: %v", err)
+		}
+	}
+
+	return nil
 }
 
 // bucketForFeeReate determines the proper bucket for a fee rate. This is done
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -638,12 +638,10 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 // bucketForFeeReate determines the proper bucket for a fee rate. This is done
 // in order to batch inputs with similar fee rates together.
 func (s *UtxoSweeper) bucketForFeeRate(
-	feeRate chainfee.SatPerKWeight) chainfee.SatPerKWeight {
+	feeRate chainfee.SatPerKWeight) int {
 
 	minBucket := s.relayFeeRate + chainfee.SatPerKWeight(s.cfg.FeeRateBucketSize)
-	return chainfee.SatPerKWeight(
-		math.Ceil(float64(feeRate) / float64(minBucket)),
-	)
+	return int(math.Ceil(float64(feeRate) / float64(minBucket)))
 }
 
 // clusterBySweepFeeRate takes the set of pending inputs within the UtxoSweeper
@@ -651,7 +649,7 @@ func (s *UtxoSweeper) bucketForFeeRate(
 // sweep fee rate, which is determined by calculating the average fee rate of
 // all inputs within that cluster.
 func (s *UtxoSweeper) clusterBySweepFeeRate() []inputCluster {
-	bucketInputs := make(map[chainfee.SatPerKWeight]pendingInputs)
+	bucketInputs := make(map[int]pendingInputs)
 	inputFeeRates := make(map[wire.OutPoint]chainfee.SatPerKWeight)
 
 	// First, we'll group together all inputs with similar fee rates. This
@@ -662,12 +660,12 @@ func (s *UtxoSweeper) clusterBySweepFeeRate() []inputCluster {
 			log.Warnf("Skipping input %v: %v", op, err)
 			continue
 		}
-		bucket := s.bucketForFeeRate(feeRate)
+		feeGroup := s.bucketForFeeRate(feeRate)
 
-		inputs, ok := bucketInputs[bucket]
+		inputs, ok := bucketInputs[feeGroup]
 		if !ok {
 			inputs = make(pendingInputs)
-			bucketInputs[bucket] = inputs
+			bucketInputs[feeGroup] = inputs
 		}
 
 		input.lastFeeRate = feeRate
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -210,9 +210,8 @@ type UtxoSweeperConfig struct {
 	// transaction.
 	FeeEstimator chainfee.Estimator
 
-	// PublishTransaction facilitates the process of broadcasting a signed
-	// transaction to the appropriate network.
-	PublishTransaction func(*wire.MsgTx) error
+	// Wallet contains the wallet functions that sweeper requires.
+	Wallet Wallet
 
 	// NewBatchTimer creates a channel that will be sent on when a certain
 	// time window has passed. During this time window, new inputs can still
@@ -321,7 +320,7 @@ func (s *UtxoSweeper) Start() error {
 
 		// Error can be ignored. Because we are starting up, there are
 		// no pending inputs to update based on the publish result.
-		err := s.cfg.PublishTransaction(lastTx)
+		err := s.cfg.Wallet.PublishTransaction(lastTx)
 		if err != nil && err != lnwallet.ErrDoubleSpend {
 			log.Errorf("last tx publish: %v", err)
 		}
@@ -886,7 +885,7 @@ func (s *UtxoSweeper) sweep(inputs inputSet, feeRate chainfee.SatPerKWeight,
 		}),
 	)
 
-	err = s.cfg.PublishTransaction(tx)
+	err = s.cfg.Wallet.PublishTransaction(tx)
 
 	// In case of an unexpected error, don't try to recover.
 	if err != nil && err != lnwallet.ErrDoubleSpend {
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -3,7 +3,6 @@ package sweep
 import (
 	"errors"
 	"fmt"
-	"math"
 	"math/rand"
 	"sort"
 	"sync"
@@ -246,8 +245,8 @@ type UtxoSweeperConfig struct {
 	// of 10 would result in the following fee rate buckets up to the
 	// maximum fee rate:
 	//
-	//   #1: min = 1 sat/vbyte, max = 10 sat/vbyte
-	//   #2: min = 11 sat/vbyte, max = 20 sat/vbyte...
+	//   #1: min = 1 sat/vbyte, max (exclusive) = 11 sat/vbyte
+	//   #2: min = 11 sat/vbyte, max (exclusive) = 21 sat/vbyte...
 	FeeRateBucketSize int
 }
 
@@ -640,8 +639,7 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 func (s *UtxoSweeper) bucketForFeeRate(
 	feeRate chainfee.SatPerKWeight) int {
 
-	minBucket := s.relayFeeRate + chainfee.SatPerKWeight(s.cfg.FeeRateBucketSize)
-	return int(math.Ceil(float64(feeRate) / float64(minBucket)))
+	return int(feeRate-s.relayFeeRate) / s.cfg.FeeRateBucketSize
 }
 
 // clusterBySweepFeeRate takes the set of pending inputs within the UtxoSweeper
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -70,15 +70,14 @@ type Params struct {
 }
 
 // pendingInput is created when an input reaches the main loop for the first
-// time. It tracks all relevant state that is needed for sweeping.
+// time. It wraps the input and tracks all relevant state that is needed for
+// sweeping.
 type pendingInput struct {
+	input.Input
+
 	// listeners is a list of channels over which the final outcome of the
 	// sweep needs to be broadcasted.
 	listeners []chan Result
-
-	// input is the original struct that contains the input and sign
-	// descriptor.
-	input input.Input
 
 	// ntfnRegCancel is populated with a function that cancels the chain
 	// notifier spend registration.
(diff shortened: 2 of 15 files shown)


sweep/sweeper.go
```diff
@@ -70,16 +70,15 @@ type Params struct {
 }
 
 // pendingInput is created when an input reaches the main loop for the first
-// time. It tracks all relevant state that is needed for sweeping.
+// time. It wraps the input and tracks all relevant state that is needed for
+// sweeping.
 type pendingInput struct {
+	input.Input
```
halseth: why this change exactly?

sweep/tx_input_set.go
```diff
@@ -0,0 +1,80 @@
...
+
+// txInputSet is an object that accumulates tx inputs and keeps running counters
+// on various properties of the tx.
+type txInputSet struct {
+	// weightEstimate is the (worst case) tx weight with the current set of
+	// inputs.
+	weightEstimate input.TxWeightEstimator
+
+	// inputTotal is the total value of all inputs.
+	inputTotal btcutil.Amount
```
halseth: can just loop over current inputs to get this instead of storing it as a var?

sweep/sweeper.go
```diff
@@ -789,7 +796,7 @@ func (s *UtxoSweeper) getInputLists(cluster inputCluster,
 	// contain inputs that failed before. Therefore we also add sets
 	// consisting of only new inputs to the list, to make sure that new
 	// inputs are given a good, isolated chance of being published.
-	var newInputs, retryInputs []input.Input
+	var newInputs, retryInputs []txInput
```
halseth: ~~any reason not using pendingInput directly?~~
got the answer in later commit

sweep/tx_input_set.go
```diff
@@ -0,0 +1,80 @@
...
+	weight := b.weightEstimate.Weight()
+	fee := b.feePerKW.FeeForWeight(int64(weight))
+
+	// Calculate the output value if the current input would be
+	// added to the set.
+	newOutputValue := newInputTotal - fee
+
+	// If adding this input makes the total output value of the set
+	// decrease, this is a negative yield input. It shouldn't be
+	// added to the set. We return the current index as the number
```
halseth: comment slightly out of date

sweep/txgenerator.go
```diff
@@ -75,96 +82,74 @@ func generateInputPartitionings(sweepableInputs []input.Input,
...
 // minimizing any negative externalities we cause for the Bitcoin system as a
 // whole.
-func getPositiveYieldInputs(sweepableInputs []input.Input, maxInputs int,
-	feePerKW chainfee.SatPerKWeight) (int, btcutil.Amount) {
-
-	var weightEstimate input.TxWeightEstimator
-
-	// Add the sweep tx output to the weight estimate.
-	weightEstimate.AddP2WKHOutput()
+func addPositiveYieldInputs(inputSet *txInputSet, sweepableInputs []txInput,
```
halseth: make this a method on `inputSet`intead?

sweep/txgenerator.go
```diff
@@ -75,96 +82,74 @@ func generateInputPartitionings(sweepableInputs []input.Input,
...
 		}
 
 		log.Infof("Candidate sweep set of size=%v, has yield=%v",
-			count, outputValue)
+			inputCount, txInputs.outputValue)
 
-		sets = append(sets, sweepableInputs[:count])
-		sweepableInputs = sweepableInputs[count:]
+		sweepSet := make([]input.Input, len(txInputs.inputs))
+		copy(sweepSet, txInputs.inputs)
```
halseth: why copy here rather than slicing as earlier?

sweep/interface.go
```diff
@@ -0,0 +1,12 @@
+package sweep
+
+import (
+	"github.com/btcsuite/btcd/wire"
+)
+
+// Wallet contains all wallet related functionality required by sweeper.
+type Wallet interface {
```
halseth: 👍

sweep/sweeper.go
```diff
@@ -632,22 +632,24 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
...
-	// lists of inputs.
-	inputLists, err := s.getInputLists(cluster, currentHeight)
-	if err != nil {
-		return fmt.Errorf("unable to examine pending inputs: %v", err)
-	}
-
-	// Sweep selected inputs.
-	for _, inputs := range inputLists {
-		err := s.sweep(inputs, cluster.sweepFeeRate, currentHeight)
+	return s.cfg.Wallet.WithCoinSelectLock(func() error {
```
halseth: should add a comment why coin select lock is needed here

sweep/txgenerator.go
```diff
@@ -105,13 +107,22 @@ func generateInputPartitionings(sweepableInputs []txInput,
 		// continuing with the remaining inputs will only lead to sets
 		// with a even lower output value.
 		if txInputs.outputValue < dustLimit {
-			log.Debugf("Set value %v below dust limit of %v",
-				txInputs.outputValue, dustLimit)
-			return sets, nil
+			dustLimitReached, err := addWalletInputs(
```
halseth: should have comment

sweep/txgenerator.go
```diff
@@ -123,6 +134,87 @@ func generateInputPartitionings(sweepableInputs []txInput,
...
 }
 
+// addWalletInputs retrieves utxos from the wallet and tries adding as many as
+// required to bring the tx output value above the given minimum.
+func addWalletInputs(inputSet *txInputSet, minOutputAmt btcutil.Amount,
+	wallet Wallet) (bool, error) {
+
+	// Retrieve wallet utxos. Only consider confirmed utxos to prevent
+	// problems around RBF rules for unconfirmed inputs.
+	utxos, err := wallet.ListUnspentWitness(1, math.MaxInt32)
```
halseth: according to the godoc minConfs can be set `=0` here? I'm not sure what is correct or if it's the same, worth checking out probably.

sweep/txgenerator.go
```diff
@@ -123,6 +134,87 @@ func generateInputPartitionings(sweepableInputs []txInput,
...
+// to the other inputs to sweep.
+func createWalletTxInput(utxo *lnwallet.Utxo) (txInput, error) {
+	var witnessType input.WitnessType
+	switch utxo.AddressType {
+	case lnwallet.WitnessPubKey:
+		witnessType = input.WitnessKeyHash
+	case lnwallet.NestedWitnessPubKey:
+		witnessType = input.NestedWitnessKeyHash
+	default:
+		return nil, fmt.Errorf("unknown input witness")
```
halseth: add the unkown type to error

sweep/tx_input_set.go
```diff
@@ -71,6 +74,20 @@ func (b *txInputSet) add(input txInput) bool {
 		return false
 	}
 
+	// If this input comes from the wallet, verify that we still gain
+	// something with this transaction.
+	_, fromWallet := input.(*walletTxInput)
+	if fromWallet {
+		newWalletTotal := b.walletInputTotal + value
```
halseth: taken in isolation, it looks like this can make the effective yield go down. Zooming out it should be okay, since we only add this when we are below the dustlimit, which obviously is what we want, even though it effectively costs us more because of the extra input. But this is not obvious from reading it in isolation.

Some comments would go a long way making it more clear I think.

sweep/txgenerator.go
```diff
@@ -123,6 +134,87 @@ func generateInputPartitionings(sweepableInputs []txInput,
...
+	)
+
+	return &walletTxInput{
+		Input: walletInput,
+	}, nil
+}
+
+// walletTxInput wraps a wallet input in the same way other sweeper inputs are
+// wrapped.
+type walletTxInput struct {
```
halseth: instead of having  a special `walletTxInput` type, what do you think about having the `txInput` define a `gain` or similar. It would be `0` for wallet inputs, and the actual  output value for other types we sweep.

When adding the wallet input, it would check that the total `gain-fee` is positive.

sweep/sweeper.go
joostjager: Two reasons:
1. Less code
2. Allows us to use `pendingInput` as `txInput` in next commits.

sweep/tx_input_set.go
joostjager: This struct already tracks several running counters. Why would we recalculate `inputTotal` specifically every time we need it?

sweep/tx_input_set.go
joostjager: fixed

sweep/txgenerator.go
joostjager: A suggestion from the linter. I like it because it is less code. Apparently this is the way to do covariance.

sweep/txgenerator.go
joostjager: good one, moved

sweep/sweeper.go
joostjager: the main reason is a bit of isolation from everything else that the sweeper tracks.

sweep/sweeper.go
joostjager: added

sweep/txgenerator.go
joostjager: added

sweep/txgenerator.go
joostjager: Where did you see this? I find in btcwallet the following:

```
// Outputs with fewer confirmations than the minimum or more
// confs than the maximum are excluded.
```

I'd think we need to set min to 1?

sweep/txgenerator.go
joostjager: fixed

sweep/txgenerator.go
joostjager: I still need to create an object that conforms to the `txInput` interface. I don't want to use `pendingInput` for this.

joostjager: Gain could also work, but it would then always be either zero or the input value. A bool would do the same thing. But adding a bool also seemed redundant if I can just type check here.

sweep/tx_input_set.go
joostjager: Comments added

sweep/txgenerator.go
joostjager: Decided to make a `fromWallet` boolean for now.

joostjager:
Unit test added

sweep/tx_input_set.go
```diff
@@ -0,0 +1,124 @@
...
+		// Try to add the input to the transaction. If that doesn't
+		// succeed because it wouldn't increase the output value,
+		// return. Assuming inputs are sorted by yield, any further
+		// inputs wouldn't increase the output value either.
+		if !b.add(input) {
+			return
+		}
+
+		// Stop if max inputs is reached.
+		if idx == maxInputs-1 {
```
wpaulino: Shouldn't `maxInputs` be enforced to `b.inputs` if `addPositiveYieldInputs` is called multiple times to the same set?

sweep/tx_input_set.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Add the sweep tx output to the weight estimate.
+	b.weightEstimate.AddP2WKHOutput()
+
+	return &b
+}
+
+// dustLimitReached returns true if we've accumulated enough inputs to meet the
+// dust limit.
+func (b *txInputSet) dustLimitReached() bool {
```
wpaulino: Nit: what does `b` stand for here?

sweep/tx_input_set.go
```diff
@@ -122,3 +171,71 @@ func (b *txInputSet) addPositiveYieldInputs(sweepableInputs []txInput,
...
+
+	// Retrieve wallet utxos. Only consider confirmed utxos to prevent
+	// problems around RBF rules for unconfirmed inputs.
+	utxos, err := b.wallet.ListUnspentWitness(1, math.MaxInt32)
+	if err != nil {
+		return err
+	}
+
+	for _, utxo := range utxos {
+		input, err := createWalletTxInput(utxo)
```
wpaulino: Don't we want to lock these if they get chosen?

sweep/sweeper.go
```diff
@@ -719,6 +724,8 @@ func (s *UtxoSweeper) scheduleSweep(currentHeight int32) error {
 	startTimer := false
 	for _, cluster := range s.clusterBySweepFeeRate() {
 		// Examine pending inputs and try to construct lists of inputs.
+		// We don't need to obtain the coin selection lock, because
+		// those coins may not be available when we sweep anyway.
```
wpaulino: May not be available because they were used in another transaction?

sweep/sweeper.go
```diff
@@ -640,8 +639,7 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 func (s *UtxoSweeper) bucketForFeeRate(
 	feeRate chainfee.SatPerKWeight) int {
 
-	minBucket := s.relayFeeRate + chainfee.SatPerKWeight(s.cfg.FeeRateBucketSize)
```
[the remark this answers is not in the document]
joostjager: The first bucket still starts at the min relay fee. But previously the bucket size was `relayFeeRate + bucketSize`. I think that isn't what one would expect when you set a bucket size?

sweep/tx_input_set.go
joostjager: converted to a nice `t`

sweep/tx_input_set.go
```diff
@@ -0,0 +1,124 @@
...
+// input was added to the set. An input is rejected if it decreases the tx
+// output value after paying fees.
+func (b *txInputSet) add(input input.Input) bool {
+	// Can ignore error, because it has already been checked when
+	// calculating the yields.
+	size, isNestedP2SH, _ := input.WitnessType().SizeUpperBound()
+
+	// Keep a running weight estimate of the input set.
+	if isNestedP2SH {
+		b.weightEstimate.AddNestedP2WSHInput(size)
```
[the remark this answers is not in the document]
joostjager: Very good catch. It wasn't a problem previously, but because we are now trying another add after a failed add, it has become important. Made a copy now.

sweep/tx_input_set.go
joostjager: That doesn't happen at the moment, but changed this just in case.

sweep/sweeper.go
joostjager: Rewritten comment. The basic idea is that we are just after an indication here

sweep/tx_input_set.go
joostjager: If we are actually sweeping, this is executed within a coin selection lock. So no other processes should be selecting coins then.

sweep/tx_input_set.go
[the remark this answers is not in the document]
joostjager: Unit test added for this

sweep/txgenerator.go
halseth: Looks like it is more code tho? Anyway, no big deal, linter is always right.

sweep/sweeper.go
```diff
@@ -645,6 +628,28 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 	}
 }
 
+// sweepCluster tries to sweep the given input cluster.
+func (s *UtxoSweeper) sweepCluster(cluster inputCluster,
+	currentHeight int32) error {
+
+	// Examine pending inputs and try to construct
```
halseth: nit: comment fits one one line.

sweep/tx_input_set.go
```diff
@@ -125,3 +174,71 @@ func (t *txInputSet) addPositiveYieldInputs(sweepableInputs []txInput,
...
+		}
+
+		// Return if we've reached the minimum output amount.
+		if t.dustLimitReached() {
+			return nil
+		}
+	}
+
+	// We were not able to reach the minimum output amount.
+	return nil
```
halseth: nit: can return error and remove the dustlimit check after calling this method.

sweep/tx_input_set.go
```diff
@@ -125,3 +174,71 @@ func (t *txInputSet) addPositiveYieldInputs(sweepableInputs []txInput,
...
+// many as required to bring the tx output value above the given minimum.
+func (t *txInputSet) tryAddWalletInputsIfNeeded() error {
+	// If we've already reached the dust limit, no action is needed.
+	if t.dustLimitReached() {
+		return nil
+	}
+
+	// Retrieve wallet utxos. Only consider confirmed utxos to prevent
+	// problems around RBF rules for unconfirmed inputs.
+	utxos, err := t.wallet.ListUnspentWitness(1, math.MaxInt32)
```
halseth: to avoid having the wallet as a field on the input set, the list of utxos can be passed into this method.

sweep/interface.go
```diff
@@ -0,0 +1,29 @@
...
+	// PublishTransaction performs cursory validation (dust checks, etc) and
+	// broadcasts the passed transaction to the Bitcoin network.
+	PublishTransaction(tx *wire.MsgTx) error
+
+	// ListUnspentWitness returns all unspent outputs which are version 0
+	// witness programs. The 'minconfirms' and 'maxconfirms' parameters
+	// indicate the minimum and maximum number of confirmations an output
+	// needs in order to be returned by this method. Passing -1 as
```
halseth: Here is the confusing minConf comment. Does passing 0 mean unconfirmed or confirmed outputs? 😛 

Probably best to pass 1, and just test that actually works.

sweep/txgenerator.go
halseth: See other comment. But yeah, setting to 1 is probably correct.

sweep/tx_input_set.go
halseth: Nice, easier to read now :)

sweep/tx_input_set.go
joostjager: Resolving this comment because it is the same as the one below

sweep/tx_input_set.go
```diff
@@ -0,0 +1,127 @@
...
+	// the set and return the outcome.
+	if newOutputValue <= t.outputValue {
+		return false
+	}
+
+	// Update running values.
+	t.inputTotal = newInputTotal
+	t.outputValue = newOutputValue
+	t.inputs = append(t.inputs, input)
+	t.weightEstimate = newWeightEstimate
```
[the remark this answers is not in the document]
joostjager: In a later commit, i would also need the `fromWallet` state. I am not too convinced of it. When you add 20 inputs, 200 (virtual) add operations will be executed. (1+2+3+...+20).

@wpaulino what do you think?

sweep/txgenerator.go
joostjager: Oh, you mean even earlier. There was an intermediate revision that had a for loop to make the copy instead of calling `copy`. That is what the linter suggested.

But after that some types changed and it is again possible to slice. Fixed

sweep/interface.go
joostjager: Ah ok. That was copy pasted. I removed the story about `-1`. It seems to be a left over from old code.

sweep/tx_input_set.go
joostjager: yes, but then we'd always get utxos, even if the dust limit is already reached. could also move that out then. but not sure if it is better

sweep/tx_input_set.go
joostjager: idk, it isn't really an error. i could return an additional bool, but to me it isn't all that different.

sweep/tx_input_set_test.go
```diff
@@ -0,0 +1,113 @@
...
+		t.Fatal(err)
+	}
+
+	if !set.dustLimitReached() {
+		t.Fatal("expected dust limit to be reached")
+	}
+}
+
+// testInput returns a test input with the specified amount.
+func testInput(amt btcutil.Amount) input.Input {
```
wpaulino: Why not use `createTestInput`?

sweep/tx_input_set.go
```diff
@@ -0,0 +1,127 @@
...
+// addPositiveYieldInputs adds sweepableInputs that have a positive yield to the
+// input set. This function assumes that the list of inputs is sorted descending
+// by yield.
+//
+// TODO(roasbeef): Consider including some negative yield inputs too to clean
+// up the utxo set even if it costs us some fees up front.  In the spirit of
+// minimizing any negative externalities we cause for the Bitcoin system as a
+// whole.
+func (t *txInputSet) addPositiveYieldInputs(sweepableInputs []txInput,
+	maxInputs int) {
```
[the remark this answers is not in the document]
joostjager: It is a method? Or what do you mean with 'member'?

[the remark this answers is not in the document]
joostjager: Ah, max inputs. Ok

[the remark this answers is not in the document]
joostjager: Added

sweep/sweeper.go
```diff
@@ -718,6 +724,10 @@ func (s *UtxoSweeper) scheduleSweep(currentHeight int32) error {
 	startTimer := false
 	for _, cluster := range s.clusterBySweepFeeRate() {
 		// Examine pending inputs and try to construct lists of inputs.
+		// We don't need to obtain the coin selection lock, because we
+		// just need an indication as to whether we can sweep. More
+		// inputs may be added until we publish the transaction and
+		// coins that we select now may be used in other transactions.
```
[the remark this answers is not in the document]
joostjager: We opportunistically schedule a sweep, I think that is how this should be looked at.

sweep/tx_input_set_test.go
joostjager: Ok, made a wrapper for the existing `createTestInput`

PR #3818, opened 2019-12-11
routing: send to route max hops check

joostjager:
Depends on https://github.com/lightningnetwork/lightning-onion/pull/45

A follow up pr https://github.com/lightningnetwork/lnd/pull/3841 will also limit path finding to the maximum payload size of 1300 bytes.

Fixes https://github.com/lightningnetwork/lnd/issues/3712

### DIFF
--- a/routing/payment_lifecycle.go
+++ b/routing/payment_lifecycle.go
@@ -239,7 +239,7 @@ func (p *paymentLifecycle) createNewPaymentAttempt() (lnwire.ShortChannelID,
 	}
 
 	// Create a new payment attempt from the given payment session.
-	route, err := p.paySession.RequestRoute(
+	rt, err := p.paySession.RequestRoute(
 		p.payment, uint32(p.currentHeight), p.finalCLTVDelta,
 	)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


Roasbeef:
Let's also update this to pin to the latest version of `lightning-onion` created here: https://github.com/lightningnetwork/lightning-onion/pull/45

joostjager:
Updated to f34e9dc1651d174d79cc7a9965ac26844eb4ec98

routing/payment_lifecycle.go
```diff
@@ -279,8 +281,28 @@ func (p *paymentLifecycle) createNewPaymentAttempt() (lnwire.ShortChannelID,
...
+		err == sphinx.ErrMaxRoutingInfoSizeExceeded {
+
+		log.Debugf("Invalid route provided for "+
+			"payment %x: %v", p.payment.PaymentHash,
+			err)
+
+		controlErr := p.router.cfg.Control.Fail(
+			p.payment.PaymentHash,
+			channeldb.FailureReasonError,
+		)
```
guggero: nit: those two statements are both wrapped a bit too early and could be more compact if some line breaks are removed.

routing/route/route.go
```diff
@@ -267,6 +271,11 @@ func NewRouteFromHops(amtToSend lnwire.MilliSatoshi, timeLock uint32,
 func (r *Route) ToSphinxPath() (*sphinx.PaymentPath, error) {
 	var path sphinx.PaymentPath
 
+	// Check maximum route length.
+	if len(r.Hops) > sphinx.NumMaxHops {
```
guggero: Could we move the `len(rt.Hops) == 0` check from `generateSphinxPacket` to here as well? It returns an error from the `route` package anyway.

joostjager: Nice suggestion, done

PR #3820, opened 2019-12-11
build: prevent loggers from overwriting each other.

joostjager:
Previously only the last registered logger would receive debug level changes.

### DIFF
--- a/log.go
+++ b/log.go
@@ -67,12 +67,12 @@ var (
 func init() {
 	setSubLogger("LTND", ltndLog, signal.UseLogger)
 	setSubLogger("ATPL", atplLog, autopilot.UseLogger)
-	setSubLogger("PEER", peerLog, nil)
-	setSubLogger("RPCS", rpcsLog, nil)
-	setSubLogger("SRVR", srvrLog, nil)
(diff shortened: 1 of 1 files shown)


log.go
```diff
@@ -99,18 +99,23 @@ func init() {
...
 // addSubLogger is a helper method to conveniently create and register the
-// logger of a sub system.
-func addSubLogger(subsystem string, useLogger func(btclog.Logger)) {
+// logger of one or more sub systems.
+func addSubLogger(subsystem string, useLoggers ...func(btclog.Logger)) {
+	// Create and register just a single logger to prevent them from
+	// overwriting each other internally.
 	logger := build.NewSubLogger(subsystem, logWriter.GenSubLogger)
-	setSubLogger(subsystem, logger, useLogger)
+	logWriter.RegisterSubLogger(subsystem, logger)
```
guggero: To avoid this code duplication, couldn't we also extend `setSubLogger` to take multiple loggers and then call that from here? The `range` can just replace this nil check which is also nice.

joostjager: Good suggestion, fixed

PR #3821, opened 2019-12-11
[anchor] pluggable anchor commitments

halseth:
Missing from end-to-end anchors working:
- [x] Commitment resolution for delayed `to_remote` output
- [x] Negotiating anchor type during funding flow

TODO:
- [ ] Unit tests for anchor specific changes
- [x] One itest case using anchors: #4001 

Long term:
- [ ] Automatically sweep anchors
- [ ] whole itest package run using anchors
- [ ] watchtower support
- [ ] channel backups

Builds on #3829

### DIFF
--- a/config.go
+++ b/config.go
@@ -345,7 +345,7 @@ type config struct {
 
 	Watchtower *lncfg.Watchtower `group:"watchtower" namespace:"watchtower"`
 
-	LegacyProtocol *lncfg.LegacyProtocol `group:"legacyprotocol" namespace:"legacyprotocol"`
+	ProtocolOptions *lncfg.ProtocolOptions `group:"protocol" namespace:"protocol"`
 
 	AllowCircularRoute bool `long:"allow-circular-route" description:"If true, our node will allow htlc forwards that arrive and depart on the same channel."`
 }
--- a/contractcourt/commit_sweep_resolver.go
+++ b/contractcourt/commit_sweep_resolver.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"sync"
 
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"github.com/lightningnetwork/lnd/input"
@@ -172,23 +173,44 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 		}
 	}
 
-	// We're dealing with our commitment transaction if the delay on the
-	// resolution isn't zero.
-	isLocalCommitTx := c.commitResolution.MaturityDelay != 0
+	// The output is on our local commitment if the script starts with
+	// OP_IF for the revocation clause. On the remote commitment it will
+	// either be a regular P2WKH or a simple sig spend with a CSV delay.
+	isLocalCommitTx := c.commitResolution.SelfOutputSignDesc.WitnessScript[0] == txscript.OP_IF
+	isDelayedOutput := c.commitResolution.MaturityDelay != 0
 
-	// There're two types of commitments, those that have tweaks
-	// for the remote key (us in this case), and those that don't.
-	// We'll rely on the presence of the commitment tweak to to
-	// discern which type of commitment this is.
+	c.log.Debugf("isDelayedOutput=%v, isLocalCommitTx=%v", isDelayedOutput,
+		isLocalCommitTx)
+
+	// There're three types of commitments, those that have tweaks
+	// for the remote key (us in this case), those that don't, and a third
+	// where there is no tweak and the output is delayed. On the local
+	// commitment our output will always be delayed. We'll rely on the
+	// presence of the commitment tweak to to discern which type of
+	// commitment this is.
 	var witnessType input.WitnessType
 	switch {
+
+	// Delayed output to us on our local commitment.
 	case isLocalCommitTx:
 		witnessType = input.CommitmentTimeLock
+
+	// A confirmed output to us on the remote commitment.
+	case isDelayedOutput:
+		witnessType = input.CommitmentToRemoteConfirmed
+
+	// A non-delayed output on the remote commitment where the key is
+	// tweakless.
 	case c.commitResolution.SelfOutputSignDesc.SingleTweak == nil:
 		witnessType = input.CommitSpendNoDelayTweakless
+
+	// A non-delayed output on the remote commitment where the key is
+	// tweaked.
 	default:
 		witnessType = input.CommitmentNoDelay
 	}
+
+	c.log.Infof("Sweeping with witness type: %v", witnessType)
 
 	// We'll craft an input with all the information required for
 	// the sweeper to create a fully valid sweeping transaction to
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -1106,6 +1106,28 @@ func (f *fundingManager) processFundingOpen(msg *lnwire.OpenChannel,
 	}
 }
 
+// commitmentType returns the commitment type to use for the channel, based on
+// the features the two peers have available.
+func commitmentType(localFeatures,
+	remoteFeatures *lnwire.FeatureVector) lnwallet.CommitmentType {
+
+	localTweakless := localFeatures.HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+	remoteTweakless := remoteFeatures.HasFeature(
+		lnwire.StaticRemoteKeyOptional,
+	)
+
+	// If both nodes are signaling the proper feature bit for tweakless
+	// copmmitments, we'll use that.
+	if localTweakless && remoteTweakless {
+		return lnwallet.CommitmentTypeTweakless
+	}
+
+	// Otherwise we'll fall back to the legacy type.
+	return lnwallet.CommitmentTypeLegacy
+}
+
 // handleFundingOpen creates an initial 'ChannelReservation' within the wallet,
 // then responds to the source peer with an accept channel message progressing
 // the funding workflow.
@@ -1228,13 +1250,9 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	// negotiated the new tweakless commitment format. This is only the
 	// case if *both* us and the remote peer are signaling the proper
 	// feature bit.
-	localTweakless := fmsg.peer.LocalFeatures().HasFeature(
-		lnwire.StaticRemoteKeyOptional,
+	commitType := commitmentType(
+		fmsg.peer.LocalFeatures(), fmsg.peer.RemoteFeatures(),
 	)
-	remoteTweakless := fmsg.peer.RemoteFeatures().HasFeature(
-		lnwire.StaticRemoteKeyOptional,
-	)
-	tweaklessCommitment := localTweakless && remoteTweakless
 	chainHash := chainhash.Hash(msg.ChainHash)
 	req := &lnwallet.InitFundingReserveMsg{
 		ChainHash:        &chainHash,
@@ -1248,7 +1266,7 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 		PushMSat:         msg.PushAmount,
 		Flags:            msg.ChannelFlags,
 		MinConfs:         1,
-		Tweakless:        tweaklessCommitment,
+		CommitType:       commitType,
 	}
 
 	reservation, err := f.cfg.Wallet.InitChannelReservation(req)
@@ -1307,9 +1325,9 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 	reservation.SetOurUpfrontShutdown(shutdown)
 
 	fndgLog.Infof("Requiring %v confirmations for pendingChan(%x): "+
-		"amt=%v, push_amt=%v, tweakless=%v, upfrontShutdown=%x", numConfsReq,
+		"amt=%v, push_amt=%v, committype=%v, upfrontShutdown=%x", numConfsReq,
 		fmsg.msg.PendingChannelID, amt, msg.PushAmount,
-		tweaklessCommitment, msg.UpfrontShutdownScript)
+		commitType, msg.UpfrontShutdownScript)
 
 	// Generate our required constraints for the remote party.
 	remoteCsvDelay := f.cfg.RequiredRemoteDelay(amt)
@@ -2904,13 +2922,9 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	// negotiated the new tweakless commitment format. This is only the
 	// case if *both* us and the remote peer are signaling the proper
 	// feature bit.
-	localTweakless := msg.peer.LocalFeatures().HasFeature(
-		lnwire.StaticRemoteKeyOptional,
+	commitType := commitmentType(
+		msg.peer.LocalFeatures(), msg.peer.RemoteFeatures(),
 	)
-	remoteTweakless := msg.peer.RemoteFeatures().HasFeature(
-		lnwire.StaticRemoteKeyOptional,
-	)
-	tweaklessCommitment := localTweakless && remoteTweakless
 	req := &lnwallet.InitFundingReserveMsg{
 		ChainHash:        &msg.chainHash,
 		PendingChanID:    chanID,
@@ -2924,7 +2938,7 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		PushMSat:         msg.pushAmt,
 		Flags:            channelFlags,
 		MinConfs:         msg.minConfs,
-		Tweakless:        tweaklessCommitment,
+		CommitType:       commitType,
 		ChanFunder:       msg.chanFunder,
 	}
 
@@ -3012,7 +3026,7 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 	maxHtlcs := f.cfg.RequiredRemoteMaxHTLCs(capacity)
 
 	fndgLog.Infof("Starting funding workflow with %v for pending_id(%x), "+
-		"tweakless=%v", msg.peer.Address(), chanID, tweaklessCommitment)
+		"committype=%v", msg.peer.Address(), chanID, commitType)
 
 	fundingOpen := lnwire.OpenChannel{
 		ChainHash:             *f.cfg.Wallet.Cfg.NetParams.GenesisHash,
--- a/fundingmanager.go
+++ b/fundingmanager.go
@@ -1111,6 +1111,20 @@ func (f *fundingManager) processFundingOpen(msg *lnwire.OpenChannel,
 func commitmentType(localFeatures,
 	remoteFeatures *lnwire.FeatureVector) lnwallet.CommitmentType {
 
+	// If both peers are signalling support for anchor commitments, this
+	// implicitly mean we'll create the channel of this type. Note that
+	// this also enables tweakless commitments, as anchor commitments are
+	// always tweakless.
+	localAnchors := localFeatures.HasFeature(
+		lnwire.AnchorsOptional,
+	)
+	remoteAnchors := remoteFeatures.HasFeature(
+		lnwire.AnchorsOptional,
+	)
+	if localAnchors && remoteAnchors {
+		return lnwallet.CommitmentTypeAnchors
+	}
+
 	localTweakless := localFeatures.HasFeature(
 		lnwire.StaticRemoteKeyOptional,
 	)
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -419,7 +419,12 @@ func (l *channelLink) Start() error {
 
 	// If the config supplied watchtower client, ensure the channel is
 	// registered before trying to use it during operation.
-	if l.cfg.TowerClient != nil {
+	// TODO(halseth): support anchor types for watchtower.
+	state := l.channel.State()
+	if l.cfg.TowerClient != nil && state.ChanType.HasAnchors() {
+		l.log.Warnf("Skipping tower registration for anchor " +
+			"channel type")
+	} else if l.cfg.TowerClient != nil && !state.ChanType.HasAnchors() {
 		err := l.cfg.TowerClient.RegisterChannel(l.ChanID())
 		if err != nil {
 			return err
@@ -1883,8 +1888,12 @@ func (l *channelLink) handleUpstreamMsg(msg lnwire.Message) {
 
 		// If we have a tower client, we'll proceed in backing up the
 		// state that was just revoked.
-		if l.cfg.TowerClient != nil {
-			state := l.channel.State()
+		// TODO(halseth): support anchor types for watchtower.
+		state := l.channel.State()
+		if l.cfg.TowerClient != nil && state.ChanType.HasAnchors() {
+			l.log.Warnf("Skipping tower backup for anchor " +
+				"channel type")
+		} else if l.cfg.TowerClient != nil && !state.ChanType.HasAnchors() {
 			breachInfo, err := lnwallet.NewBreachRetribution(
 				state, state.RemoteCommitment.CommitHeight-1, 0,
 			)
--- a/input/script_utils.go
+++ b/input/script_utils.go
@@ -853,18 +853,6 @@ func CommitScriptToSelf(csvTimeout uint32, selfKey, revokeKey *btcec.PublicKey) 
 	return builder.Script()
 }
 
-// CommitScriptUnencumbered constructs the public key script on the commitment
-// transaction paying to the "other" party. The constructed output is a normal
-// p2wkh output spendable immediately, requiring no contestation period.
-func CommitScriptUnencumbered(key *btcec.PublicKey) ([]byte, error) {
-	// This script goes to the "other" party, and is spendable immediately.
-	builder := txscript.NewScriptBuilder()
-	builder.AddOp(txscript.OP_0)
-	builder.AddData(btcutil.Hash160(key.SerializeCompressed()))
-
-	return builder.Script()
-}
-
 // CommitSpendTimeout constructs a valid witness allowing the owner of a
 // particular commitment transaction to spend the output returning settled
 // funds back to themselves after a relative block timeout.  In order to
@@ -971,6 +959,69 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
 	}
 
 	return witness, nil
+}
+
+// CommitScriptUnencumbered constructs the public key script on the commitment
+// transaction paying to the "other" party. The constructed output is a normal
+// p2wkh output spendable immediately, requiring no contestation period.
+func CommitScriptUnencumbered(key *btcec.PublicKey) ([]byte, error) {
+	// This script goes to the "other" party, and is spendable immediately.
+	builder := txscript.NewScriptBuilder()
+	builder.AddOp(txscript.OP_0)
+	builder.AddData(btcutil.Hash160(key.SerializeCompressed()))
+
+	return builder.Script()
+}
+
+// CommitScriptToRemoteConfirmed constructs the script for the output on the
+// commitment transaction paying to the remote party of said commitment
+// transaction. The money can only be spend after one confirmation.
+//
+// Possible Input Scripts:
+//     SWEEP: <sig>
+//
+// Output Script:
+//	<key> OP_CHECKSIGVERIFY
+//	1 OP_CHECKSEQUENCEVERIFY
+func CommitScriptToRemoteConfirmed(key *btcec.PublicKey) ([]byte, error) {
+	builder := txscript.NewScriptBuilder()
+
+	// Only the given key can spend the output.
+	builder.AddData(key.SerializeCompressed())
+	builder.AddOp(txscript.OP_CHECKSIGVERIFY)
+
+	// Check that the it has one confirmation.
+	builder.AddOp(txscript.OP_1)
+	builder.AddOp(txscript.OP_CHECKSEQUENCEVERIFY)
+
+	return builder.Script()
+}
+
+// CommitSpendToRemoteConfirmed constructs a valid witness allowing a node to
+// spend their settled output on the counterparty's commitment transaction when
+// it has one confirmetion. This is used for the anchor channel type. The
+// spending key will always be non-tweaked for this output type.
+func CommitSpendToRemoteConfirmed(signer Signer, signDesc *SignDescriptor,
+	sweepTx *wire.MsgTx) (wire.TxWitness, error) {
+
+	if signDesc.KeyDesc.PubKey == nil {
+		return nil, fmt.Errorf("cannot generate witness with nil " +
+			"KeyDesc pubkey")
+	}
+
+	// Similar to non delayed output, only a signature is needed.
+	sweepSig, err := signer.SignOutputRaw(sweepTx, signDesc)
+	if err != nil {
+		return nil, err
+	}
+
+	// Finally, we'll manually craft the witness. The witness here is the
+	// signature and the redeem script.
+	witnessStack := make([][]byte, 2)
+	witnessStack[0] = append(sweepSig, byte(signDesc.HashType))
+	witnessStack[1] = signDesc.WitnessScript
+
+	return witnessStack, nil
 }
 
 // SingleTweakBytes computes set of bytes we call the single tweak. The purpose
--- a/input/script_utils.go
+++ b/input/script_utils.go
@@ -1024,6 +1024,74 @@ func CommitSpendToRemoteConfirmed(signer Signer, signDesc *SignDescriptor,
 	return witnessStack, nil
 }
 
+// CommitScriptAnchor constructs the script for the anchor output spendable by
+// the given key immediately, or by anyone after 16 confirmations.
+//
+// Possible Input Scripts:
+//    By owner:				<sig>
+//    By anyone (after 16 conf):	<emptyvector>
+//
+// Output Script:
+//	<funding_pubkey> OP_CHECKSIG OP_IFDUP
+//	OP_NOTIF
+//	  OP_16 OP_CSV
+//	OP_ENDIF
+func CommitScriptAnchor(key *btcec.PublicKey) ([]byte, error) {
+	builder := txscript.NewScriptBuilder()
+
+	// Spend immediately with key.
+	builder.AddData(key.SerializeCompressed())
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	// Duplicate the value if true, since it will be consumed by the NOTIF.
+	builder.AddOp(txscript.OP_IFDUP)
+
+	// Otherwise spendable by anyone after 16 confirmations.
+	builder.AddOp(txscript.OP_NOTIF)
+	builder.AddOp(txscript.OP_16)
+	builder.AddOp(txscript.OP_CHECKSEQUENCEVERIFY)
+	builder.AddOp(txscript.OP_ENDIF)
+
+	return builder.Script()
+}
+
+// CommitSpendAnchor constructs a valid witness allowing a node to spend their
+// anchor output on the commitment transaction using their funding key. This is
+// used for the anchor channel type.
+func CommitSpendAnchor(signer Signer, signDesc *SignDescriptor,
+	sweepTx *wire.MsgTx) (wire.TxWitness, error) {
+
+	if signDesc.KeyDesc.PubKey == nil {
+		return nil, fmt.Errorf("cannot generate witness with nil " +
+			"KeyDesc pubkey")
+	}
+
+	// Create a signature.
+	sweepSig, err := signer.SignOutputRaw(sweepTx, signDesc)
+	if err != nil {
+		return nil, err
+	}
+
+	// The witness here is just a signature and the redeem script.
+	witnessStack := make([][]byte, 2)
+	witnessStack[0] = append(sweepSig, byte(signDesc.HashType))
+	witnessStack[1] = signDesc.WitnessScript
+
+	return witnessStack, nil
+}
+
+// CommitSpendAnchorAnyone constructs a witness allowing anyone to spend the
+// anchor output after it has gotten 16 confirmations. Since no signing is
+// required, only knowledge of the redeem script is necessary to spend it.
+func CommitSpendAnchorAnyone(script []byte) (wire.TxWitness, error) {
+	// The witness here is just the redeem script.
+	witnessStack := make([][]byte, 2)
+	witnessStack[0] = nil
+	witnessStack[1] = script
+
+	return witnessStack, nil
+}
+
 // SingleTweakBytes computes set of bytes we call the single tweak. The purpose
 // of the single tweak is to randomize all regular delay and payment base
 // points. To do this, we generate a hash that binds the commitment point to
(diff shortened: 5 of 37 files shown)


input/script_utils.go
```diff
@@ -249,6 +249,109 @@ func SenderHTLCScript(senderHtlcKey, receiverHtlcKey,
...
+//     OP_SWAP OP_SIZE 32 OP_EQUAL
+//     OP_NOTIF
+//         OP_DROP 2 OP_SWAP <sender htlc key> 2 OP_CHECKMULTISIG
+//     OP_ELSE
+//         OP_HASH160 <ripemd160(payment hash)> OP_EQUALVERIFY
+//         OP_CHECKSIG
+//     OP_ENDIF
+//     1 OP_CHECKSEQUENCEVERIFY
+// OP_ENDIF
+func SenderHTLCScriptAnchors(senderHtlcKey, receiverHtlcKey,
```
halseth: for these scripts we could pass a boolen/chantype in to determine whether to add the OP_CSV, but copied for now to make them totally independent.

lnwallet/commitment.go
```diff
@@ -0,0 +1,711 @@
...
+	}
+
+	// Generate the proper redeem scripts for the HTLC output modified by
+	// two-bits denoting if this is an incoming HTLC, and if the HTLC is
+	// being applied to their commitment transaction or ours.
+	switch {
+	// The HTLC is paying to us, and being applied to our commitment
+	// transaction. So we need to use the receiver's version of HTLC the
+	// script.
+	case isIncoming && ourCommit:
```
halseth: They keyring is already base on whether this is our commit our not... could we simplify this switch?

lnwallet/commitment.go
```diff
@@ -0,0 +1,711 @@
...
+		ourBalance -= commitFeeMSat
+
+	case !cb.chanState.IsInitiator && commitFee > theirBalance.ToSatoshis():
+		theirBalance = 0
+
+	case !cb.chanState.IsInitiator:
+		theirBalance -= commitFeeMSat
+	}
+
+	// If the commitment has anchors, we must also subtract the anchor
```
halseth: This is a bit awkward, but is a result of that the balances we store in the db has had their commit fee subtracted. should we do a db migration to change this?

lnwallet/commitment.go
```diff
@@ -12,6 +13,124 @@ import (
...
+	// redeem outputs from a revoked commitment transaction if it were to
+	// be published.
+	RevocationKey *btcec.PublicKey
+}
+
+// DeriveCommitmentKeys generates a new commitment key set using the base points
+// and commitment point. The keys are derived differently depending on the type
+// of channel, and whether the commitment transaction is ours or the remote
+// peer's.
+func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
```
joostjager: Separate move from change in two commits, for easier review.

lnwallet/commitment.go
```diff
@@ -151,9 +154,31 @@ type ScriptInfo struct {
 // CommitScriptToRemote creates the script that will pay to the non-owner of
 // the commitment transaction, adding a delay to the script based on the
 // channel type.
-func CommitScriptToRemote(_ channeldb.ChannelType, csvTimeout uint32,
+func CommitScriptToRemote(chanType channeldb.ChannelType, csvTimeout uint32,
 	key *btcec.PublicKey) (*ScriptInfo, error) {
 
+	// If this channel type has anchors, we derive the delayed to_remote
+	// script.
+	if chanType.HasAnchors() {
```
joostjager: Is this then the thing that prompts the creation of a `CommitmentBuilder` interface?

lnwallet/commitment.go
halseth: would also need to flip the `isIncoming` based on the local/remote perspecitve. Seems to be a bigger change that would, leaving for later.

input/script_utils.go
```diff
@@ -948,6 +1156,126 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+//    By owner:				<sig>
+//    By anyone (after 16 conf):	<emptyvector>
+//
+// Output Script:
+//	<funding_pubkey> OP_CHECKSIG OP_NOTIF
+//	16 OP_CSV
+func CommitScriptAnchor(key *btcec.PublicKey) ([]byte, error) {
+	builder := txscript.NewScriptBuilder()
+
+	builder.AddOp(txscript.OP_DEPTH)
```
cfromknecht: stray op

input/script_utils.go
```diff
@@ -948,6 +1156,126 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+
+	// Spend immediately with key.
+	builder.AddData(key.SerializeCompressed())
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	// Otherswise spendable by anyone after 16 confirmations.
```
cfromknecht: Otherswise

input/script_utils.go
```diff
@@ -948,6 +1156,126 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+	builder.AddOp(txscript.OP_NOTIF)
+	builder.AddInt64(16)
```
cfromknecht: OP_16?

input/script_utils.go
```diff
@@ -249,6 +249,109 @@ func SenderHTLCScript(senderHtlcKey, receiverHtlcKey,
...
+	builder.AddOp(txscript.OP_EQUALVERIFY)
+
+	// This checks the receiver's signature so that a third party with
+	// knowledge of the payment preimage still cannot steal the output.
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	// Close out the OP_IF statement above.
+	builder.AddOp(txscript.OP_ENDIF)
+
+	// 1 block CSV delay.
```
cfromknecht: can this be simplified by having a subroutine with a boolean arg as to whether or not to add these two ops? same for the receiver script.

input/script_utils.go
```diff
@@ -473,6 +576,122 @@ func ReceiverHTLCScript(cltvExpiry uint32, senderHtlcKey,
...
+	builder.AddInt64(int64(cltvExpiry))
+	builder.AddOp(txscript.OP_CHECKLOCKTIMEVERIFY)
+	builder.AddOp(txscript.OP_DROP)
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	// Close out the inner if statement.
+	builder.AddOp(txscript.OP_ENDIF)
+
+	// 1 block CSV delay.
+	builder.AddInt64(1)
```
cfromknecht: OP_1?

input/script_utils.go
```diff
@@ -948,6 +1156,126 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+	builder := txscript.NewScriptBuilder()
+
+	// Only the given key can spend the output.
+	builder.AddData(key.SerializeCompressed())
+	builder.AddOp(txscript.OP_CHECKSIGVERIFY)
+
+	// Check that the CSV delay has passed.
+	builder.AddInt64(int64(csvTimeout))
+	builder.AddOp(txscript.OP_CHECKSEQUENCEVERIFY)
+	return builder.Script()
```
cfromknecht: nit: add newline

input/size.go
```diff
@@ -168,6 +165,32 @@ const (
...
+	//	- CountTxOut: 1 byte
+	//	- TxOut: 4*43 + 43 * num-htlc-outputs bytes
+	//		OutputPayingToThem,
+	//		OutputPayingToUs,
+	//		AnchorPayingToThem,
+	//		AnchorPayingToUs,
+	//		....HTLCOutputs...
+	//	- LockTime: 4 bytes
+	BaseAnchorCommitmentTxSize = 4 + 1 + FundingInputSize + 1 +
+		2*CommitmentDelayOutput + 2*CommitmentAnchorOutput + 4
```
cfromknecht: does this mean we overpay when <4 outputs are present?

input/size.go
```diff
@@ -168,6 +165,32 @@ const (
 	// WitnessCommitmentTxWeight 224 weight
 	WitnessCommitmentTxWeight = WitnessHeaderSize + WitnessSize
 
+	// BaseAnchorCommitmentTxSize 223 + 43 * num-htlc-outputs bytes
+	//	- Version: 4 bytes
+	//	- WitnessHeader <---- part of the witness data
+	//	- CountTxIn: 1 byte
+	//	- TxIn: 41 bytes
+	//		FundingInput
+	//	- CountTxOut: 1 byte
```
cfromknecht: Can’t this be >1 byte?

lnwallet/reservation.go
```diff
@@ -151,7 +151,13 @@ func NewChannelReservation(capacity, localFundingAmt btcutil.Amount,
 	// TODO(halseth): make method take remote funding amount directly
 	// instead of inferring it from capacity and local amt.
 	capacityMSat := lnwire.NewMSatFromSatoshis(capacity)
+
+	// The total fee paid by the initiator will be the commitment fee in
+	// addition to the two anchor outputs.
 	feeMSat := lnwire.NewMSatFromSatoshis(commitFee)
+	if anchorOutputs {
+		feeMSat -= 2 * lnwire.NewMSatFromSatoshis(anchorSize)
```
cfromknecht: +=?

input/script_utils.go
halseth: Yeah as mentioned in the comment above this can be done, but didn't do it to keep script independent . But it's a lot of duplication so maybe better to just do it.

input/size.go
halseth: I don't see how this is different than the current commitment type, where it is 1 byte? Since there are just 2 extra outputs

input/size.go
halseth: Yeah, similar to the current commitment type, where we overpay (it goes to fees) when < 2 outputs are present.

lnwallet/commitment.go
```diff
@@ -184,9 +187,31 @@ type ScriptInfo struct {
...
+			return nil, err
+		}
+
+		return &ScriptInfo{
+			PkScript:      p2wsh,
+			WitnessScript: script,
+		}, nil
+	}
+
+	// Otherwise the te_remote will be a simple p2wkh.
```
joostjager: te_remote

lnwallet/commitment.go
```diff
@@ -200,6 +225,50 @@ func CommitScriptToRemote(_ channeldb.ChannelType, csvTimeout uint32,
...
+		localChanCfg.MultiSigKey.PubKey,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	localAnchorScriptHash, err := input.WitnessScriptHash(localAnchorScript)
+	if err != nil {
+		return nil, nil, err
+	}
```
joostjager: Extract into function that returns `ScriptInfo`?

lnwallet/commitment.go
```diff
@@ -465,18 +570,53 @@ func CreateCommitTx(chanType channeldb.ChannelType,
 	commitTx := wire.NewMsgTx(2)
 	commitTx.AddTxIn(&fundingOutput)
 
+	var (
+		localOutput  = false
+		remoteOutput = false
+	)
+
 	// Avoid creating dust outputs within the commitment transaction.
 	if amountToLocal >= localChanCfg.DustLimit {
```
joostjager: could assign this condition to `localOutput` and make it `if localOutput`

lnwallet/commitment.go
```diff
@@ -229,6 +229,18 @@ func CommitScriptToRemote(chanType channeldb.ChannelType, csvTimeout uint32,
 	}, nil
 }
 
+func InitiatorFee(chanType channeldb.ChannelType, commitWeight int64,
```
joostjager: Would there be a better term for `miner_fee + anchors` then "fee"? To prevent introducing a second definition of fee. Cost, overhead, ..?

lnwallet/channel.go
```diff
@@ -5639,6 +5639,100 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
...
+			}
+
+			return &wire.OutPoint{
+				Hash:  tx.TxHash(),
+				Index: uint32(i),
+			}
+		}
+
+		// Not found.
+		return nil
```
joostjager: Could make this stricter and return an error if the channel type says there must be anchors.

lnwallet/channel.go
```diff
@@ -1898,8 +1911,16 @@ func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
 		return nil, err
 	}
 
+	localAnchorScript, _, err := CommitScriptAnchors(
```
joostjager: Switching back from our/their to local/remote?

lnwallet/channel.go
```diff
@@ -5133,7 +5143,8 @@ func newOutgoingHtlcResolution(signer input.Signer,
 // they can just sweep the output immediately with knowledge of the pre-image.
 //
 // TODO(roasbeef) consolidate code with above func
-func newIncomingHtlcResolution(signer input.Signer, localChanCfg *channeldb.ChannelConfig,
+func newIncomingHtlcResolution(signer input.Signer,
```
joostjager: Are these funcs also candidates to move into `commitment.go`, now that they are channel type-dependent?

fundingmanager.go
```diff
@@ -1241,6 +1241,7 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 		Flags:            msg.ChannelFlags,
 		MinConfs:         1,
 		Tweakless:        tweaklessCommitment,
+		AnchorOutputs:    false,
```
halseth: reminder: must also be set in the other reservation msg

lnwallet/commitment.go
```diff
@@ -200,6 +230,80 @@ func CommitScriptToRemote(_ channeldb.ChannelType, csvTimeout uint32,
...
+		return nil, nil, err
+	}
+
+	return &ScriptInfo{
+			PkScript:      localAnchorScript,
+			WitnessScript: localAnchorScriptHash,
+		},
+		&ScriptInfo{
+			PkScript:      remoteAnchorScript,
+			WitnessScript: remoteAnchorScriptHash,
```
joostjager: hash and script swapped.

input/script_utils.go
```diff
@@ -948,6 +1155,126 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+// CommitScriptAnchor constructs the script for the anchor output spendable by
+// the given key immediately, or by anyone after 16 confirmations.
+//
+// Possible Input Scripts:
+//    By owner:				<sig>
+//    By anyone (after 16 conf):	<emptyvector>
+//
+// Output Script:
+//	<funding_pubkey> OP_CHECKSIG OP_NOTIF
+//	16 OP_CSV
```
halseth: missing op_endif

lnwallet/reservation.go
halseth: Yes 🙀

lnwallet/commitment.go
halseth: I like it :)

lnwallet/commitment.go
halseth: I like `cost` perhaps?

lnwallet/channel.go
```diff
@@ -4955,24 +4932,19 @@ func newOutgoingHtlcResolution(signer input.Signer,
 		Index: uint32(htlc.OutputIndex),
 	}
 
+	// First, we'll re-generate the script used to send the HTLC to
+	// the remote party within their commitment transaction.
+	htlcScriptHash, htlcScript, err := genHtlcScript(
```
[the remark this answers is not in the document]
halseth: `genHtlcScript` uses the `isIncoming` and `ourCommit` arguments to determine which scripts to create.  So this commit is really just a reuse of that logic instead of doing the same here.

lnwallet/channel.go
```diff
@@ -5146,17 +5108,10 @@ func newIncomingHtlcResolution(signer input.Signer, localChanCfg *channeldb.Chan
...
-		keyRing.RevocationKey, htlc.RHash[:],
-	)
-	if err != nil {
-		return nil, err
-	}
 	successSignDesc := input.SignDescriptor{
 		KeyDesc:       localChanCfg.HtlcBasePoint,
 		SingleTweak:   keyRing.LocalHtlcKeyTweak,
-		WitnessScript: htlcCreationScript,
+		WitnessScript: htlcScript,
```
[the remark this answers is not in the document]
halseth: The `genHtlcScript` should choose the correct script based on its parameters?

lnwallet/commitment.go
```diff
@@ -216,6 +283,11 @@ type CommitmentBuilder struct {
 
 // NewCommitmentBuilder creates a new CommitmentBuilder from chanState.
 func NewCommitmentBuilder(chanState *channeldb.OpenChannel) *CommitmentBuilder {
+	// The anchor chennel type MUST be tweakless.
+	if chanState.ChanType.HasAnchors() && !chanState.ChanType.IsTweakless() {
+		panic("invalid channel type combination")
```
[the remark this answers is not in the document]
halseth: We should never allow the combination to be set, so I think a panic is okay.

lnwallet/channel.go
halseth: Since we now just pass the `chanType` through, I think keeping them here is okay.

lnwallet/reservation.go
```diff
@@ -213,6 +219,16 @@ func NewChannelReservation(capacity, localFundingAmt btcutil.Amount,
 		)
 	}
 
+	// Similarly we ensure their balance is reasonable if we are not the
+	// initiator.
+	if !initiator && theirBalance.ToSatoshis() <= 2*DefaultDustLimit() {
```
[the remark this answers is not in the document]
halseth: It is not a new requirement, it was just something I noticed wasn't checked. Made it a separate commit.

halseth:
New version pushed. Highlight are command line flag for enabling it, disabling of watchtower and channel backups, handling of the delayed to_remote output in breach cases, and properly setting the sequence number in order to sweep the delayed HTLC outputs.

lnwire/features.go
```diff
@@ -101,6 +101,16 @@ const (
 	// HTLC.
 	MPPOptional FeatureBit = 17
 
+	// AnchorsRequired is a required feature bit that signals that the node
+	// requires channels to be made using commitments having anchor
+	// outputs.
+	AnchorsRequired FeatureBit = 1336
```
halseth: need to choose actual xzbit

halseth: same as for TLV types?

config.go
```diff
@@ -1125,6 +1125,15 @@ func loadConfig() (*config, error) {
 		return nil, fmt.Errorf("Unable to parse node color: %v", err)
 	}
 
+	// For now we don't allow watchtowers to be active while anchpr
+	// commitments are, as it'll require a change to how we communicate
+	// with the tower.
+	// TODO(halseth): make the change you want to see in the world.
+	if cfg.Watchtower.Active && cfg.ProtocolOptions.AnchorCommitments() {
```
halseth: remove this check

htlcswitch/link.go
```diff
@@ -415,7 +415,9 @@ func (l *channelLink) Start() error {
 
 	// If the config supplied watchtower client, ensure the channel is
 	// registered before trying to use it during operation.
-	if l.cfg.TowerClient != nil {
+	// TODO(halseth): support anchor types for watchtower.
+	state := l.channel.State()
+	if l.cfg.TowerClient != nil && !state.ChanType.HasAnchors() {
```
halseth: log?

fundingmanager.go
```diff
@@ -1231,6 +1231,20 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
 		lnwire.StaticRemoteKeyOptional,
 	)
 	tweaklessCommitment := localTweakless && remoteTweakless
+
+	// If both peers are signalling support for anchor commitments, this
+	// implicitly mean we'll create the channel of this type. Note that
+	// this also enables tweakless commitments, as anchor commitments are
+	// always tweakless.
+	localAnchors := fmsg.peer.LocalFeatures().HasFeature(
```
halseth: extract into helper

fundingmanager.go
```diff
@@ -1231,6 +1231,20 @@ func (f *fundingManager) handleFundingOpen(fmsg *fundingOpenMsg) {
...
+		lnwire.AnchorsOptional,
+	)
+	remoteAnchors := fmsg.peer.RemoteFeatures().HasFeature(
+		lnwire.AnchorsOptional,
+	)
+	anchorCommitment := localAnchors && remoteAnchors
+	tweaklessCommitment = tweaklessCommitment || anchorCommitment
```
halseth: not necessary? dependency might already be validated

input/size.go
```diff
@@ -247,11 +247,14 @@ const (
 	//                      - OP_DROP: 1 byte
 	//                      - OP_CHECKSIG: 1 byte
 	//              - OP_ENDIF: 1 byte
+	//              - OP_1: 1 byte
+	//              - OP_CSV: 1 byte
+	//              - OP_DROP: 1 byte
```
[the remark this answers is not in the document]
joostjager: Discussed offline: does the remote need to use the exact same weight to make sure the commitment txes reconstructed on both sides are identical?

input/size.go
```diff
@@ -260,7 +263,7 @@ const (
 	//      - witness_script: (accepted_htlc_script)
 	AcceptedHtlcTimeoutWitnessSize = 1 + 1 + 73 + 1 + 1 + AcceptedHtlcScriptSize
 
-	// AcceptedHtlcSuccessWitnessSize 322 bytes
+	// AcceptedHtlcSuccessWitnessSize 325 bytes
```
halseth: must also change here? https://github.com/lightningnetwork/lnd/blob/b75259a/input/size.go#L170

joostjager:
I tested aggregation of the 2nd level success txes that are generated in this PR. This is possible now that the remote party signs with sighash single+acp. Works on both `bitcoind` and `btcd`. Also attached a wallet utxo input and change output (bump the fee of the 2nd level aggregation).

It isn't easy to test this though. For the record: this is the (hacky) code I used to do it: https://github.com/lightningnetwork/lnd/compare/master...joostjager:test-htlc-sighash (last commit). When a channel is force closed with two pending incoming htlcs, a hex string is written to stdout that encodes the aggregate tx. This can be used to do a manual `sendrawtransaction`.

Roasbeef:
Why do we need to disable SCB for channels with anchor outputs, given that the to_remote output doesn't change?

halseth:
> Why do we need to disable SCB for channels with anchor outputs, given that the to_remote output doesn't change?

The `to_remote` output *does* change: it now has a 1 block CSV enforced.

input/script_utils.go
```diff
@@ -168,9 +171,11 @@ func Ripemd160H(d []byte) []byte {
 //         OP_HASH160 <ripemd160(payment hash)> OP_EQUALVERIFY
 //         OP_CHECKSIG
 //     OP_ENDIF
+//     [1 OP_CHECKSEQUENCEVERIFY OP_DROP] <- if allowing confirmed spend only.
 // OP_ENDIF
 func SenderHTLCScript(senderHtlcKey, receiverHtlcKey,
-	revocationKey *btcec.PublicKey, paymentHash []byte) ([]byte, error) {
+	revocationKey *btcec.PublicKey, paymentHash []byte,
+	confirmedSpend bool) ([]byte, error) {
```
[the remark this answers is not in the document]
halseth: Fixed

input/script_utils_test.go
```diff
@@ -285,6 +311,9 @@ func TestHTLCSenderSpendValidation(t *testing.T) {
 			// revoke w/ sig
 			// TODO(roasbeef): test invalid revoke
 			makeWitnessTestCase(t, func() (wire.TxWitness, error) {
+				genCommitTx(false)
```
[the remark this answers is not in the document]
halseth: Done

input/script_utils_test.go
```diff
@@ -367,6 +463,65 @@ func TestHTLCSenderSpendValidation(t *testing.T) {
...
+				genSweepTx(true)
+
+				signDesc := &SignDescriptor{
+					KeyDesc: keychain.KeyDescriptor{
+						PubKey: aliceKeyPub,
+					},
+					SingleTweak:   aliceCommitTweak,
+					WitnessScript: htlcWitnessScript,
+					Output:        htlcOutput,
+					HashType:      txscript.SigHashAll,
```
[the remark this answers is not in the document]
halseth: Since Alice is assembling the final tx, she can used `sighash_all`. Only Bob needs to use the other sighash type.

input/script_utils_test.go
```diff
@@ -258,23 +221,86 @@ func TestHTLCSenderSpendValidation(t *testing.T) {
...
+	var (
+		htlcWitnessScript, htlcPkScript []byte
+		htlcOutput                      *wire.TxOut
+		sweepTxSigHashes                *txscript.TxSigHashes
+		senderCommitTx, sweepTx         *wire.MsgTx
+		bobRecvrSig                     []byte
+	)
+
+	// genCommitTx generates a commitment tx where the htlc output requires
+	// confirmation to be sent according to 'confirmed'.
```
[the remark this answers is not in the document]
halseth: Fixed. Should be "spent"

input/size.go
```diff
@@ -213,6 +213,22 @@ const (
 	//      - witness_script (to_local_script)
 	ToLocalPenaltyWitnessSize = 1 + 1 + 73 + 1 + 1 + ToLocalScriptSize
 
+	// ToRemoteConfirmedScriptSize 40 bytes
+	//      - OP_DATA: 1 byte
+	//      - to_remote_key: 33 bytes
+	//      - OP_CHECKSIGVERIFY: 1 byte
+	//      - OP_1: 1 byte
+	//      - OP_CHECKSEQUENCEVERIFY: 1 byte
+	ToRemoteConfirmedScriptSize = 1 + 33 + 1 + 1 + 1
```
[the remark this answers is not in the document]
halseth: Indeed. Math is hard.

input/script_utils.go
```diff
@@ -1024,6 +1024,74 @@ func CommitSpendToRemoteConfirmed(signer Signer, signDesc *SignDescriptor,
 	return witnessStack, nil
 }
 
+// CommitScriptAnchor constructs the script for the anchor output spendable by
+// the given key immediately, or by anyone after 16 confirmations.
+//
+// Possible Input Scripts:
```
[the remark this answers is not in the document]
halseth: We use both, but input script is used throughout this file

lnwallet/commitment.go
```diff
@@ -430,7 +503,8 @@ func (cb *CommitmentBuilder) createUnsignedCommitmentTx(ourBalance,
 func CreateCommitTx(chanType channeldb.ChannelType,
 	fundingOutput wire.TxIn, keyRing *CommitmentKeyRing,
 	localChanCfg, remoteChanCfg *channeldb.ChannelConfig,
-	amountToLocal, amountToRemote btcutil.Amount) (*wire.MsgTx, error) {
+	amountToLocal, amountToRemote btcutil.Amount,
+	numHTLCs int64) (*wire.MsgTx, error) {
```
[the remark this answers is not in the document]
halseth: Would be a bigger change, but could indeed be done.

utxonursery_test.go
```diff
@@ -620,8 +619,10 @@ func createOutgoingRes(onLocalCommitment bool) *lnwallet.OutgoingHtlcResolution
 		}
 
 		outgoingRes.SignedTimeoutTx = timeoutTx
+		outgoingRes.CsvDelay = 2
```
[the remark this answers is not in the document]
halseth: It was preexisting for the local commitment case

lnwallet/reservation.go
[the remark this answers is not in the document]
halseth: From 10 lines above

lnwallet/reservation.go
```diff
@@ -151,7 +151,13 @@ func NewChannelReservation(capacity, localFundingAmt btcutil.Amount,
 	// TODO(halseth): make method take remote funding amount directly
 	// instead of inferring it from capacity and local amt.
 	capacityMSat := lnwire.NewMSatFromSatoshis(capacity)
+
+	// The total fee paid by the initiator will be the commitment fee in
+	// addition to the two anchor outputs.
 	feeMSat := lnwire.NewMSatFromSatoshis(commitFee)
+	if anchorOutputs {
+		feeMSat += 2 * lnwire.NewMSatFromSatoshis(anchorSize)
```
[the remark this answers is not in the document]
halseth: Fixed.

input/size.go
halseth: Done

fundingmanager.go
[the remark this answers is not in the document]
halseth: enum created

input/size.go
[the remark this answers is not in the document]
halseth: Done

lnwallet/reservation.go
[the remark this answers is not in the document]
joostjager: Isn't 10 lines above 2 times the anchor size?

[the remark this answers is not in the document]
halseth: https://github.com/lightningnetwork/lnd/blob/3aca9d24b849822b0c345985f2c4d27cad688b22/lnwallet/wallet.go#L69

Roasbeef:
As discussed, offline (not a blocker for this PR), we should add logic to calculate the _empirical_ weight of a commitment (for all types generically) and verify that our total fee rate isn't below the min relay amount. This safe guards us against weight _over_ estimation, which may cause us to attempt to sign+broadcast a commitment with insufficient fees. This logic can be inserted at either the state machine level, or within the contract court.

cfromknecht:
> This safe guards us against weight over estimation, which may cause us to attempt to sign+broadcast a commitment with insufficient fees. This logic can be inserted at either the state machine level, or within the contract court.

The intent is to prevent underestimation no? I would vote to add this in the state machine, as by the time the contractcourt can't prevent us from signing such a state and sending signatures to our peer.

lnwire/features.go
halseth: I don't think it is defined yet in the spec what the range for experimental feature bits is?

Also I don't think picking something `>= 2^16` is what we want in this case, as they are actual bit indices, which would lead to the serialized feature vector being `~ 2^16 / 8 = 8k` bytes long (correct me if I'm wrong). 

Maybe it makes sense to define some a range `>= 128` for experimental features? The would reserve 8 bytes (64 bits => 32 features?) for spec features. Maybe we should reserve even more.

lnwallet/channel.go
```diff
@@ -2185,31 +2180,13 @@ func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
...
-			if err != nil {
-				return nil, err
-			}
-		}
-
-		htlcPkScript, err := input.WitnessScriptHash(htlcWitnessScript)
+		// re-generate the sender HTLC script. Otherwise, is this was
+		// an outgoing HTLC that we sent, then from the PoV of the
+		// remote commitment state, they're the receiver of this HTLC.
+		htlcPkScript, htlcWitnessScript, err := genHtlcScript(
```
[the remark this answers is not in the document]
halseth: Probably introduced after the fact, and not used in all possible places.

halseth:
> As discussed, offline (not a blocker for this PR), we should add logic to calculate the _empirical_ weight of a commitment (for all types generically) and verify that our total fee rate isn't below the min relay amount. This safe guards us against weight _over_ estimation, which may cause us to attempt to sign+broadcast a commitment with insufficient fees. This logic can be inserted at either the state machine level, or within the contract court.

What is the right way to handle such a failed verification? Panic, silent error?

input/size.go
halseth: Good observation, it looks like the spec calculations are not accounting for that: https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#expected-weight-of-the-commitment-transaction

It's an existing issue, but it could possibly be worse in the cases where we are targeting a min-relay feerate, as we could be underestimating. We change this constant to target a worst-case nunmber of outputs, but that would lead to overpaying again in most cases. I think the best solution would be to rely less on this base constant, and dynamically calculate the commit size as we create it.

lnwallet/channel.go
```diff
@@ -5475,6 +5477,7 @@ func newIncomingHtlcResolution(signer input.Signer,
 				},
 				HashType: txscript.SigHashAll,
 			},
+			CsvDelay: HtlcSecondLevelInputSequence(chanType),
```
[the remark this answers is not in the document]
halseth: Github showing commits out of order... Should be fixed now.

lntest/itest/lnd_test.go
```diff
@@ -4247,7 +4247,7 @@ func testMultiHopPayments(net *lntest.NetworkHarness, t *harnessTest) {
 	//
 	// First, we'll create Dave and establish a channel to Alice. Dave will
 	// be running an older node that requires the legacy onion payload.
-	daveArgs := []string{"--legacyprotocol.onion"}
+	daveArgs := []string{"--protocol.onion"}
```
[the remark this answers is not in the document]
halseth: Agreed. Changed.

Roasbeef:
> What is the right way to handle such a failed verification? Panic, silent error?

Minimally, halt channel operations other than co-op or force close IMO.

input/size.go
```diff
@@ -213,7 +213,7 @@ const (
 	//      - witness_script (to_local_script)
 	ToLocalPenaltyWitnessSize = 1 + 1 + 73 + 1 + 1 + ToLocalScriptSize
 
-	// AcceptedHtlcScriptSize 139 bytes
+	// AcceptedHtlcScriptSize 142 bytes
```
Roasbeef: Don't we need an entirely new set of constants for anchors? Or turn all these constants into functions so they can always return the correct size estimate based on the commitment type. 

Or are we OK with just always over estimating the _weight_ of the commitment? If we always over estimate (across the board), but the other implementations don't (in favor of more accurate estimates), then seems we could de-synchronize.

input/script_utils.go
```diff
@@ -973,6 +961,69 @@ func CommitSpendNoDelay(signer Signer, signDesc *SignDescriptor,
...
+// commitment transaction paying to the remote party of said commitment
+// transaction. The money can only be spend after one confirmation.
+//
+// Possible Input Scripts:
+//     SWEEP: <sig>
+//
+// Output Script:
+//	<key> OP_CHECKSIGVERIFY
+//	1 OP_CHECKSEQUENCEVERIFY
+func CommitScriptToRemoteConfirmed(key *btcec.PublicKey) ([]byte, error) {
```
Roasbeef: Any particular reason you opted for a _new_ function here, instead of a new parameter as was done with the HTLC scripts? Non-blocking, just the inconsistency jumped out.

contractcourt/commit_sweep_resolver.go
```diff
@@ -172,24 +173,45 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
-	isLocalCommitTx := c.commitResolution.MaturityDelay != 0
-
-	// There're two types of commitments, those that have tweaks
-	// for the remote key (us in this case), and those that don't.
-	// We'll rely on the presence of the commitment tweak to to
-	// discern which type of commitment this is.
+	// The output is on our local commitment if the script starts with
+	// OP_IF for the revocation clause. On the remote commitment it will
+	// either be a regular P2WKH or a simple sig spend with a CSV delay.
+	isLocalCommitTx := c.commitResolution.SelfOutputSignDesc.WitnessScript[0] == txscript.OP_IF
```
Roasbeef: Seems a bit fragile, but I guess it works for now.

lnwallet/reservation.go
```diff
@@ -15,6 +15,32 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
+// CommitmentType is an enum indicating the commitment type we should use for
+// the channel we are opening.
+type CommitmentType int
+
+const (
+	// CommitmentTypeLegacy is the legacy commitment format with a tweaked
+	// to_remoe key.
```
Roasbeef: to_remoe -> to_remote

input/size.go
```diff
@@ -202,6 +202,18 @@ const (
 	// which will transition an incoming HTLC to the delay-and-claim state.
 	HtlcSuccessWeight = 703
 
+	// HtlcTimeoutWeightConfirmed is the weight of the HTLC timeout
+	// transaction which will transition an outgoing HTLC to the
+	// delay-and-claim state, for the confirmed HTLC outputs. It is 3 bytes
+	// larger because of the cdditional CSV check in the input script.
+	HtlcTimeoutWeightConfirmed = 666
```
Roasbeef: 😈

Roasbeef: Perhaps we should just have this (and also the above similarly), be: 
```go
ConfirmedScriptOverhead = 3
HtlcTimeoutWeightConfirmed = HtlcTimeoutWeight + ConfirmedScriptOverhead
```

Eliminates the `+3` magic number in the diff, and also some of these constants.

input/size.go
halseth: It is only used from the `SizeUpperBound` method, so the decision to slightly overestimate in some cases was made. Alternative would be, as you say, to make a new set of constants, or make it a method, but this would also require us to create a new `WitnessType` for these confirmed types, which feels a bit overkill.

Ultimately we should probably rewrite weight estimation to rely less on constants, and instead inspect the actual transaction being created.

> If we always over estimate (across the board), but the other implementations don't (in favor of more accurate estimates), then seems we could de-synchronize.

These constants are only being used for weight estimation, so we should be okay here.

input/script_utils.go
halseth: In the HTLC script cases it was adding a few new op codes to the existing scripts. In this case it is an entirely new script, going from p2wkh to p2wsh.

contractcourt/commit_sweep_resolver.go
halseth: Yeah, no straight forward way of adding this new information about output type, since this is all fetched from the DB. We could do a db migration, but inspecting the script (which is already persisted) seems easiest.

lnwire/features.go
halseth: This must still be resolved before we can merge this.

input/size.go
halseth: Yeah, that is better. Done

halseth:
> > What is the right way to handle such a failed verification? Panic, silent error?
> 
> Minimally, halt channel operations other than co-op or force close IMO.

Added: https://github.com/lightningnetwork/lnd/pull/3821/commits/6b1f37d3d7e6a12f9b433a1f7a4eee1180b3c95b

lnwire/features.go
Roasbeef: Rather it should be resolved before we cut a _release_ with this included. Feature bits are easy to change after the fact.

input/size.go
```diff
@@ -202,6 +202,23 @@ const (
 	// which will transition an incoming HTLC to the delay-and-claim state.
 	HtlcSuccessWeight = 703
 
+	// HtlcConfirmedScriptOverhead is the extra length of an HTLC script
+	// that requires confirmation before it can be spent. These extra bytes
+	// is a result of the extra CSV check.
+	HtlcConfirmedScriptOverhead = 3
```
Roasbeef: 👍

PR #3824, opened 2019-12-12
lnrpc: expose features on rpc invoices

cfromknecht:
In #3802 we exposed the features on a payreq decoded using the `decodepayreq` rpc. In this PR, we take this one step further by displaying it on all `lnrpc.Invoice` objects. This allows the invoice generator to see what features are being advertised on their invoices.


lnrpc/invoicesrpc/utils.go
```diff
@@ -125,6 +127,23 @@ func CreateRPCInvoice(invoice *channeldb.Invoice,
...
+func CreateRPCFeatures(fv *lnwire.FeatureVector) []*lnrpc.Feature {
+	features := fv.Features()
+	rpcFeatures := make([]*lnrpc.Feature, 0, len(features))
+	for bit := range features {
+		name := fv.Name(bit)
+		rpcFeatures = append(rpcFeatures, &lnrpc.Feature{
+			Bit:        uint32(bit),
+			Name:       name,
+			IsRequired: bit.IsRequired(),
+			IsKnown:    name != "unknown",
```
joostjager: This looks like we need an `IsKnown()` method on the features

PR #3828, opened 2019-12-12
channeldb: custom records sanity check

joostjager:
This PR adds a sanity check right before serializing the custom records hop data. It should never happen that this check fails because we check custom records at every entry point. Only because the consequences of a bad record slipping through could be bad, we think that a sanity check here is justified.

### DIFF
--- a/lnrpc/routerrpc/router_backend.go
+++ b/lnrpc/routerrpc/router_backend.go
@@ -231,8 +231,8 @@ func (r *RouterBackend) QueryRoutes(ctx context.Context,
 	// If we have any TLV records destined for the final hop, then we'll
 	// attempt to decode them now into a form that the router can more
 	// easily manipulate.
-	err = ValidateCustomRecords(in.DestCustomRecords)
-	if err != nil {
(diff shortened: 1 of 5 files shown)


lnrpc/routerrpc/router_backend.go
```diff
@@ -231,8 +231,8 @@ func (r *RouterBackend) QueryRoutes(ctx context.Context,
 	// If we have any TLV records destined for the final hop, then we'll
 	// attempt to decode them now into a form that the router can more
 	// easily manipulate.
-	err = ValidateCustomRecords(in.DestCustomRecords)
-	if err != nil {
+	customRecords := record.CustomSet(in.DestCustomRecords)
```
bhandras: I guess it is not a problem that copy is by reference here?

joostjager: No, it is no problem. Those rpc parameters are not used anywhere else.

joostjager:
Sanity check is catching some insane test records... Need to fix

joostjager:
Ok, it is sane again now

PR #3829, opened 2019-12-12
[anchor] Pluggable commitments

halseth:
This lays the groundwork for making it easy to introduce anchor commitments.

Replaces #3773


lnwallet/commitment.go
```diff
@@ -0,0 +1,450 @@
...
+	}, nil
+}
+
+// CreateCommitTx creates a commitment transaction, spending from specified
+// funding output. The commitment transaction contains two outputs: one local
+// output paying to the "owner" of the commitment transaction which can be
+// spent after a relative block delay or revocation event, and a remote output
+// paying the counterparty within the channel, which can be spent immediately
+// or after a delay depending on the commitment type..
+func CreateCommitTx(chanType channeldb.ChannelType,
```
joostjager: Shouldn't this also be a method on `CommitmentBuilder`?

lnwallet/commitment.go
```diff
@@ -0,0 +1,450 @@
...
+	// redeem outputs from a revoked commitment transaction if it were to
+	// be published.
+	RevocationKey *btcec.PublicKey
+}
+
+// DeriveCommitmentKeys generates a new commitment key set using the base points
+// and commitment point. The keys are derived differently depending on the type
+// of channel, and whether the commitment transaction is ours or the remote
+// peer's.
+func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
```
joostjager: Method on `CommitmentBuilder`?

lnwallet/commitment.go
halseth: Not sure what we gain by doing that, as there are less dependencies without (it only need the `ChannelType` instead of the whole `OpenChannel`)

lnwallet/commitment.go
halseth: same

lnwallet/commitment.go
joostjager: My idea is that that would prevent pass in wrong parameters here. Or is this called from places where we don't have a `CommitmentBuilder` yet?

lnwallet/channel.go
```diff
@@ -2430,23 +2430,27 @@ func (lc *LightningChannel) createCommitmentTx(c *commitment,
...
+		remoteBalance = theirBalance.ToSatoshis()
 	} else {
-		delay = uint32(lc.remoteChanCfg.CsvDelay)
-		delayBalance = theirBalance.ToSatoshis()
-		p2wkhBalance = ourBalance.ToSatoshis()
+		localCfg = lc.remoteChanCfg
+		remoteCfg = lc.localChanCfg
+		localBalance = theirBalance.ToSatoshis()
+		remoteBalance = ourBalance.ToSatoshis()
 	}
```
joostjager: Would it be better to have two different calls to `CreateCommitTx`, on in each `if` block? To prevent assigning to those intermediate vars of which the name may be confusing perhaps.

lnwallet/commitment.go
```diff
@@ -127,27 +127,48 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
...
+	// obfuscator is a 48-bit state hint that's used to obfuscate the
+	// current state number on the commitment transactions.
+	obfuscator [StateHintSize]byte
+}
+
+// NewCommitmentBuilder creates a new CommitmentBuilder from chanState.
+func NewCommitmentBuilder(chanState *channeldb.OpenChannel) *CommitmentBuilder {
+	return &CommitmentBuilder{
+		chanState:  chanState,
+		obfuscator: createStateHintObfuscator(chanState),
```
joostjager: Don't need to wait for restoring the commitment state, because this call doesn't use it right?

lnwallet/channel.go
```diff
@@ -2169,33 +2169,41 @@ func (lc *LightningChannel) fetchCommitmentView(remoteChain bool,
...
+		dustLimit = lc.remoteChanCfg.DustLimit
 	}
 
 	nextHeight := commitChain.tip().height + 1
 
 	// Run through all the HTLCs that will be covered by this transaction
 	// in order to update their commitment addition height, and to adjust
-	// the balances on the commitment transaction accordingly.
+	// the balances on the commitment transaction accordingly. Note that
+	// these nalances will be *before* taking a commitment fee from the
```
joostjager: typo (nalances)

lnwallet/commitment.go
```diff
@@ -122,6 +122,12 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
 	// we'll use that directly, and ignore the commitPoint tweak.
 	if tweaklessCommit {
 		keyRing.RemoteKey = remoteBasePoint
+
+		// If this is a tweakless remote commitment, then we can safely
+		// blank out the SingleTweak value as it isn't needed.
+		if !isOurCommit {
+			keyRing.LocalCommitKeyTweak = nil
+		}
```
joostjager: I get a bit confused with all the local and remote perspectives, but isn't it the case now that both local and remote keys are untweaked?

lnwallet/commitment.go
```diff
@@ -137,6 +137,36 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
...
+	// output with PkSript. This field will only be populated if a PkScript
+	// is p2wsh or p2sh.
+	WitnessScript []byte
+}
+
+// CommitScriptToRemote creates the script that will pay to the non-owner of
+// the commitment transaction, adding a delay to the script based on the
+// channel type.
+func CommitScriptToRemote(_ channeldb.ChannelType, csvTimeout uint32,
+	key *btcec.PublicKey) (*ScriptInfo, error) {
```
joostjager: Can this then be a method on `CommitmentBuilder`, or also not?

joostjager: Ok, also called from the naked `CreateCommitTx` function. Still wondering if it wouldn't be better eventually to also instantiate a commitment builder in the funding flow.

lnwallet/channel.go
```diff
@@ -1884,7 +1884,13 @@ func NewBreachRetribution(chanState *channeldb.OpenChannel, stateNum uint64,
...
 		return nil, err
 	}
-	localPkScript, err := input.CommitScriptUnencumbered(keyRing.RemoteKey)
+
+	// Since it is the remote breach we are reconstructing, the output going
+	// to us will be a to-remote script with our local params.
+	localDelay := uint32(chanState.LocalChanCfg.CsvDelay)
+	localScript, err := CommitScriptToRemote(
+		chanState.ChanType, localDelay, keyRing.RemoteKey,
+	)
```
joostjager: Would "our" be more clear in this context?

lnwallet/commitment.go
```diff
@@ -365,7 +395,8 @@ func (cb *CommitmentBuilder) createCommitmentTx(ourBalance,
 // spent after a relative block delay or revocation event, and a remote output
 // paying the counterparty within the channel, which can be spent immediately
 // or after a delay depending on the commitment type..
-func CreateCommitTx(fundingOutput wire.TxIn, keyRing *CommitmentKeyRing,
+func CreateCommitTx(chanType channeldb.ChannelType,
```
joostjager: Or can there be a lighter base class above `CommitmentBuilder` that contains just what we need in the funding flow. `CommitmentBuilder` extends that with what else is required for a live channel.

lnwallet/commitment.go
```diff
@@ -313,17 +336,18 @@ func (cb *CommitmentBuilder) createCommitmentTx(c *commitment,
...
+			height, cb.chanState.FundingOutpoint,
 			totalOut, cb.chanState.Capacity)
 	}
 
-	c.txn = commitTx
-	c.fee = commitFee
-	c.ourBalance = ourBalance
-	c.theirBalance = theirBalance
-	return nil
+	return &commitmentView{
```
cfromknecht: beautiful commit

lnwallet/channel.go
```diff
@@ -2169,33 +2169,41 @@ func (lc *LightningChannel) fetchCommitmentView(remoteChain bool,
 	keyRing *CommitmentKeyRing) (*commitment, error) {
 
 	commitChain := lc.localCommitChain
+	dustLimit := lc.localChanCfg.DustLimit
```
cfromknecht: "scine" in commit message

lnwallet/commitment.go
```diff
@@ -137,6 +137,36 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
...
+	PkScript []byte
+
+	// WitnessScript is the full script required to properly redeem the
+	// output with PkSript. This field will only be populated if a PkScript
+	// is p2wsh or p2sh.
+	WitnessScript []byte
+}
+
+// CommitScriptToRemote creates the script that will pay to the non-owner of
+// the commitment transaction, adding a delay to the script based on the
```
cfromknecht: does it add the delay in this commit?

lnwallet/channel.go
halseth: Yeah, I think that's nice. Done: https://github.com/lightningnetwork/lnd/pull/3829/commits/bfd026cf5e88b24710183d6322586767d8bae452

lnwallet/channel.go
```diff
@@ -2430,23 +2430,27 @@ func (lc *LightningChannel) createCommitmentTx(c *commitment,
 	}
 
 	var (
-		delay                      uint32
-		delayBalance, p2wkhBalance btcutil.Amount
+		localCfg, remoteCfg         *channeldb.ChannelConfig
+		localBalance, remoteBalance btcutil.Amount
```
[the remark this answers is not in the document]
halseth: Added a comment.

lnwallet/channel.go
```diff
@@ -6213,32 +6217,39 @@ func (lc *LightningChannel) generateRevocation(height uint64) (*lnwire.RevokeAnd
...
-func CreateCommitTx(fundingOutput wire.TxIn,
-	keyRing *CommitmentKeyRing, csvTimeout uint32,
-	amountToSelf, amountToThem, dustLimit btcutil.Amount) (*wire.MsgTx, error) {
+// funding output. The commitment transaction contains two outputs: one local
+// output paying to the "owner" of the commitment transaction which can be
+// spent after a relative block delay or revocation event, and a remote output
+// paying the counterparty within the channel, which can be spent immediately
+// or after a delay depending on the commitment type..
+func CreateCommitTx(fundingOutput wire.TxIn, keyRing *CommitmentKeyRing,
+	localChanCfg, remoteChanCfg *channeldb.ChannelConfig,
```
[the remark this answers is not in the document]
halseth: Yes, it is to be used for the delay when symmetric delays are introduced

lnwallet/channel.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/be5d74662b7a9f164e54e427a1ae2ef35bfa9b9f

lnwallet/commitment.go
halseth: I did it this way to not change current behavior, where we only nil it for our own. But you are right, also from the remote perspective it could/should be tweakless, it just wasn't needed since we never use this variable from that perspective.
Changed: https://github.com/lightningnetwork/lnd/pull/3829/commits/38c11a43297830c443d69b6f65c6274da9aae2aa

lnwallet/channel.go
halseth: Done: https://github.com/lightningnetwork/lnd/pull/3829/commits/80f4ece42aacb0988820c031cc9d4d0b8a3302a3

lnwallet/channel.go
```diff
@@ -1224,10 +1224,6 @@ type LightningChannel struct {
 	// Capacity is the total capacity of this channel.
```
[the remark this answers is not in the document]
halseth: Fixed.

lnwallet/commitment.go
halseth: what do you mean?

input/witnessgen.go
```diff
@@ -114,11 +114,6 @@ const (
 	// output that sends to a nested P2SH script that pays to a key solely
 	// under our control. The witness generated needs to include the
 	NestedWitnessKeyHash StandardWitnessType = 11
-
-	// CommitSpendNoDelayTweakless is similar to the CommitSpendNoDelay
-	// type, but it omits the tweak that randomizes the key we need to
-	// spend with a channel peer supplied set of randomness.
-	CommitSpendNoDelayTweakless StandardWitnessType = 12
```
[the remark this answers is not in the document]
halseth: Where are they persisted exactly?

What we can do is just define it there (or change the name here to indicate it should not be used anymore), and convert it to the common type when reading it from disk.

lnwallet/channel.go
halseth: fixed

lnwallet/commitment.go
halseth: No, that's why I wrote it as "adding a delay based on the channel type". Maybe that isn't very clear?

halseth:
> With regards to naming, it can get quite confusing here and there with the definition of local/remote. Maybe in some cases, our and their may be more clear.

I did a sweep to find a few places where this could be made more clear. Opted for `ours/theirs` in cases of static perspective (where we are always handling keys, outputs, commitments from our local node perspective), and `local/remote` for the dynamic perspective (where we are sharing logic from both perspectives).

Obviously this naming is used interchangeably throughout the codebase, so it was an effort way to large for this PR to change it everywhere, but would be nice to decide on a naming scheme and stick to it going forward.

I did the change in the parts already touched by this PR and for the `ChannelConfigs`:
https://github.com/lightningnetwork/lnd/pull/3829/commits/07859528852f697cf9d5ec89750e2fe69162bce1
https://github.com/lightningnetwork/lnd/pull/3829/commits/7f7449d937b1118812bc3f80d349eeb094471556
https://github.com/lightningnetwork/lnd/pull/3829/commits/03a6d377e6cc8cbad9cb4502abcdf9cd784a6e5f
https://github.com/lightningnetwork/lnd/pull/3829/commits/4f02cdea60f6ef261b47f044c9eee5629f35554c
https://github.com/lightningnetwork/lnd/pull/3829/commits/80f4ece42aacb0988820c031cc9d4d0b8a3302a3

lmk what you think!

halseth:
Ended up rever

> > With regards to naming, it can get quite confusing here and there with the definition of local/remote. Maybe in some cases, our and their may be more clear.
> 
> I did a sweep to find a few places where this could be made more clear. Opted for `ours/theirs` in cases of static perspective (where we are always handling keys, outputs, commitments from our local node perspective), and `local/remote` for the dynamic perspective (where we are sharing logic from both perspectives).
> 
> Obviously this naming is used interchangeably throughout the codebase, so it was an effort way to large for this PR to change it everywhere, but would be nice to decide on a naming scheme and stick to it going forward.
> 
> I did the change in the parts already touched by this PR and for the `ChannelConfigs`:
> [0785952](https://github.com/lightningnetwork/lnd/commit/07859528852f697cf9d5ec89750e2fe69162bce1)
> [7f7449d](https://github.com/lightningnetwork/lnd/commit/7f7449d937b1118812bc3f80d349eeb094471556)
> [03a6d37](https://github.com/lightningnetwork/lnd/commit/03a6d377e6cc8cbad9cb4502abcdf9cd784a6e5f)
> [4f02cde](https://github.com/lightningnetwork/lnd/commit/4f02cdea60f6ef261b47f044c9eee5629f35554c)
> [80f4ece](https://github.com/lightningnetwork/lnd/commit/80f4ece42aacb0988820c031cc9d4d0b8a3302a3)
> 
> lmk what you think!

Ended up reverting most of these changes, since it made a large PR even larger. I think we should agree on a proper way of naming things, and do a PR that does only that change if wanted.

lnwallet/channel.go
joostjager: I guess that is reverted now

joostjager: Ok, it is in the next commit

lnwallet/commitment.go
joostjager: Previously we were creating the state hint obfuscator after we restored the commitment state. But I answered it myself already, it doesn't matter.

lnwallet/commitment.go
joostjager: Leaving for later

watchtower/wtclient/backup_task_internal_test.go
```diff
@@ -193,10 +191,6 @@ func genTaskTest(
 		}
 
 		witnessType := input.CommitmentNoDelay
-		if tweakless {
```
cfromknecht: seems we don't test tweakless at all now?

input/witnessgen.go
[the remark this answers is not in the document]
cfromknecht: breacharbiter and nursery for sure, probably a few other places.

> What we can do is just define it there (or change the name here to indicate it should not be used anymore), and convert it to the common type when reading it from disk.

would need to see in how many places the serialization logic lives, but that's definitely one path. or we just leave the switch statements in place. since the tweakless-ness is now signaled through the sign descriptor, it might not be so bad

halseth:
Removed the last commit, we can move that discussion to #3864, as it is not essential for anchor progress.

lnwallet/channel.go
```diff
@@ -1449,8 +1443,8 @@ func NewLightningChannel(signer input.Signer,
 // createSignDesc derives the SignDescriptor for commitment transactions from
 // other fields on the LightningChannel.
 func (lc *LightningChannel) createSignDesc() error {
-	localKey := lc.localChanCfg.MultiSigKey.PubKey.SerializeCompressed()
-	remoteKey := lc.remoteChanCfg.MultiSigKey.PubKey.SerializeCompressed()
+	localKey := lc.channelState.LocalChanCfg.MultiSigKey.PubKey.SerializeCompressed()
+	remoteKey := lc.channelState.RemoteChanCfg.MultiSigKey.PubKey.SerializeCompressed()
```
cfromknecht: nit: long lines?

lnwallet/channel.go
```diff
@@ -5998,8 +5992,8 @@ func (lc *LightningChannel) CompleteCooperativeClose(localSig, remoteSig []byte,
 
 	// Finally, construct the witness stack minding the order of the
 	// pubkeys+sigs on the stack.
-	ourKey := lc.localChanCfg.MultiSigKey.PubKey.SerializeCompressed()
-	theirKey := lc.remoteChanCfg.MultiSigKey.PubKey.SerializeCompressed()
+	ourKey := lc.channelState.LocalChanCfg.MultiSigKey.PubKey.SerializeCompressed()
+	theirKey := lc.channelState.RemoteChanCfg.MultiSigKey.PubKey.SerializeCompressed()
```
cfromknecht: nit: also long line

watchtower/wtclient/client_test.go
```diff
@@ -291,8 +291,8 @@ func (c *mockChannel) createRemoteCommitTx(t *testing.T) {
 
 	commitKeyRing := &lnwallet.CommitmentKeyRing{
 		RevocationKey: c.revPK,
-		NoDelayKey:    c.toLocalPK,
-		DelayKey:      c.toRemotePK,
+		RemoteKey:     c.toLocalPK,
```
cfromknecht: nice 👍 it case it isn't clear (from the lack of comments 😅) it's important to flip local and reverse before giving this to the wtclient since the logic is from the opposite perspective.

input/witnessgen.go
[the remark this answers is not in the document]
Roasbeef: Had the same thought when I saw this, IIRC they're also persisted in some tower logic as well. Even independent of that, it's a useful witness type, as I may want to pass the very same sign descriptor for the two commitment types, an example being in the future if we opt to allow _existing_ channels to switch over to this new witness type.

Roasbeef:
I disagree with the move from local/remote to ours/theirs, we actually did the exact reverse in the past as local/remote is arguably more objective. It just adds a rather large unnecessary diff, and breaks existing models of the code for those already familiar with these areas.

lnwallet/commitment.go
```diff
@@ -211,16 +232,17 @@ func (lc *LightningChannel) createCommitmentTx(c *commitment,
 	// CreateCommitTx with parameters mathcing the perspective, to generate
 	// a new commitment transaction with all the latest unsettled/un-timed
 	// out HTLCs.
+	fundingTxIn := *wire.NewTxIn(&cb.chanState.FundingOutpoint, nil, nil)
```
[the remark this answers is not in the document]
halseth: Created https://github.com/lightningnetwork/lnd/pull/3829/commits/ac03e6ed4e8ec8fb6632a9e0fd2e6bc74e3d612f to not do `channeldb` changes.

lnwallet/channel.go
```diff
@@ -2163,33 +2163,41 @@ func (lc *LightningChannel) fetchCommitmentView(remoteChain bool,
...
 
-	// Determine how many current HTLCs are over the dust limit, and should
-	// be counted for the purpose of fee calculation.
-	var dustLimit btcutil.Amount
-	if remoteChain {
-		dustLimit = lc.channelState.RemoteChanCfg.DustLimit
-	} else {
-		dustLimit = lc.channelState.LocalChanCfg.DustLimit
+	// Actually generate unsigned commitment transaction for this view.
+	commitView, err := lc.commitBuilder.createCommitmentTx(
```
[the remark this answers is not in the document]
halseth: Yeah, true. Better name `commitTxInfo` perhaps?

lnwallet/commitment.go
halseth: Reverted back to the original change here, since it the `to_local` key stays tweaked even for tweakless commits.

lnwallet/channel.go
halseth: After recent changes our/their is removed again, but I'm open to using it only here.

contractcourt/chain_watcher.go
```diff
@@ -351,7 +351,7 @@ func isOurCommitment(localChanCfg, remoteChanCfg channeldb.ChannelConfig,
 	// With the keys derived, we'll construct the remote script that'll be
 	// present if they have a non-dust balance on the commitment.
 	remotePkScript, err := input.CommitScriptUnencumbered(
-		commitKeyRing.NoDelayKey,
+		commitKeyRing.RemoteKey,
```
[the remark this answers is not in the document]
halseth: Fixed.

lnwallet/commitment.go
```diff
@@ -122,6 +122,10 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
 	// we'll use that directly, and ignore the commitPoint tweak.
 	if tweaklessCommit {
 		keyRing.RemoteKey = remoteBasePoint
+
+		// If this is a tweakless commitment, then we can safely blank
+		// out the commitment tweak, as it isn't needed.
+		keyRing.LocalCommitKeyTweak = nil
```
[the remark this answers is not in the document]
halseth: Reverted back to original change, only nilling when not our commit.

halseth:
> I disagree with the move from local/remote to ours/theirs, we actually did the exact reverse in the past as local/remote is arguably more objective. It just adds a rather large unnecessary diff, and breaks existing models of the code for those already familiar with these areas.

Yeah, change turned our large, reverted.

To indicate the difference between the "static/dynamic" keys based on perspective, I renamed the `(No)DelayKey` to `ToLocalKey` and `ToRemoteKey`, and kept everything else intact.

lnwallet/commitment.go
```diff
@@ -36,11 +39,14 @@ type CommitmentKeyRing struct {
...
 	// transaction.
+	//
+	// NOTE: This will always refer to "our" local HTLC key, regardless of
+	// whether this is our commit or not.
 	LocalHtlcKeyTweak []byte
 
-	// LocalHtlcKey is the key that will be used in the "to self" clause of
-	// any HTLC scripts within the commitment transaction for this key ring
-	// set.
+	// LocalHtlcKey is the key that will be used in the any clause paying
```
joostjager: "any clause" ?

lnwallet/commitment.go
joostjager: We can safely blank it out, but why would we? Does this protect us against something?

lnwallet/commitment.go
```diff
@@ -83,13 +83,20 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
 
 	tweaklessCommit := chanType.IsTweakless()
 
+	// Depending on if this is our commit or not, we'll choose the correct
+	// base point.
+	localBasePoint := localChanCfg.PaymentBasePoint
+	if isOurCommit {
+		localBasePoint = localChanCfg.DelayBasePoint
+	}
+
```
joostjager: Merge this with `if isOurCommit {` block below where the other base points are calculated? (bring everything up)

lnwallet/commitment.go
```diff
@@ -83,13 +83,20 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
...
 	// First, we'll derive all the keys that don't depend on the context of
 	// whose commitment transaction this is.
 	keyRing := &CommitmentKeyRing{
 		CommitPoint: commitPoint,
 
 		LocalCommitKeyTweak: input.SingleTweakBytes(
-			commitPoint, localChanCfg.PaymentBasePoint.PubKey,
+			commitPoint, localBasePoint.PubKey,
```
joostjager: Further down in this func, `TweakPubKey` is used that calculated the tweak bytes again.

lnwallet/channel.go
joostjager: I think our/their would be more clear, because there is only a single perspective in this function. But up to you given our release cycle.

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: It seems that the problem is that the type `commitment` is actually a `commitmentView` and that `commitmentView` is not much more than a `commitment` tx. We should probably swap the names.

At least renaming the local var here would be useful.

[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/79c7f5681a71cf4e63ad406d2745b785af0dd3e6

lnwallet/commitment.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/557386cd36aaf631d31c91b7bd044ff99826047e

lnwallet/commitment.go
halseth: The comment was just copied, and yes a bit confusing. Updated: https://github.com/lightningnetwork/lnd/pull/3829/commits/0e46272440be020f82e107e7698ac5265595e443

lnwallet/channel.go
halseth: Using our/their: https://github.com/lightningnetwork/lnd/pull/3829/commits/3eecaa3bf5b64b3494301b3c19d2cf7cb53c33b5

Let me know what you think.

joostjager: Looks good

lnwallet/commitment.go
joostjager: Discussed offline. It is a different tweak. Naming remains a challenge.

lnwallet/commitment.go
joostjager: Discussed offline, excluded from thispr

lnwallet/commitment.go
```diff
@@ -149,6 +149,36 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
...
+func CommitScriptToRemote(_ channeldb.ChannelType, csvTimeout uint32,
+	key *btcec.PublicKey) (*ScriptInfo, error) {
+
+	p2wkh, err := input.CommitScriptUnencumbered(key)
+	if err != nil {
+		return nil, err
+	}
+
+	// Since this is a regular P2WKH, the WitnessScipt and PkScript should
+	// be both seth to the script hash.
```
cfromknecht: seth -> set

lnwallet/commitment.go
```diff
@@ -149,6 +149,36 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
...
+	return &ScriptInfo{
+		WitnessScript: p2wkh,
```
cfromknecht: the godoc on `WitnessScript` says this will only be populated for `p2wsh` or `p2sh`, does the comment need updating?

lnwallet/commitment.go
```diff
@@ -149,6 +149,36 @@ func DeriveCommitmentKeys(commitPoint *btcec.PublicKey,
 	return keyRing
 }
 
+// ScriptInfo holds a redeem script and hash.
+type ScriptInfo struct {
+	// PkScript is the outputs' PkScript.
```
cfromknecht: nit: output's

input/signdescriptor.go
```diff
@@ -56,8 +56,9 @@ type SignDescriptor struct {
 	DoubleTweak *btcec.PrivateKey
 
 	// WitnessScript is the full script required to properly redeem the
-	// output. This field will only be populated if a p2wsh or a p2sh
-	// output is being signed.
+	// output. This field should be set to the full script if a p2wsh
+	// output is being signed. For p2wkh it should be set to the hashed
```
cfromknecht: okay, i see the comment is updated here, squash?

lnwallet/channel.go
halseth: FIxed: https://github.com/lightningnetwork/lnd/pull/3829/commits/907d1f30f1ac065b4011707f4190aaf78be98a3d

lnwallet/channel.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/907d1f30f1ac065b4011707f4190aaf78be98a3d

lnwallet/commitment.go
```diff
@@ -166,27 +167,50 @@ func createStateHintObfuscator(state *channeldb.OpenChannel) [StateHintSize]byte
...
+	// transaction as one party must pay the commitment fee.
+	ourBalance   lnwire.MilliSatoshi
+	theirBalance lnwire.MilliSatoshi
+}
+
+// createCommitmentTx generates the unsigned commitment transaction for a
+// commitment view and returns it as part of the unsignedCommitmentTx. The
+// passed in balances should be balances *before* subtracting any commitment
+// fees.
+func (cb *CommitmentBuilder) createCommitmentTx(ourBalance,
```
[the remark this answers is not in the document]
halseth: Renamed: https://github.com/lightningnetwork/lnd/pull/3829/commits/4950f8e743ed1dd0a92c6c9c28f3901956de145d

lnwallet/commitment.go
```diff
@@ -36,11 +39,14 @@ type CommitmentKeyRing struct {
 	// the local HTLC base point. This value is needed in order to
 	// derive the final key used within the HTLC scripts in the commitment
 	// transaction.
+	//
+	// NOTE: This will always refer to "our" local HTLC key, regardless of
```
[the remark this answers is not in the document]
halseth: Yeah, this led to some confusion during review. 

I also tried switching the perspective always, but that turned out to be a large change since it will ripple through many parts of the codebase, so felt it was better to leave it as-is for now, with some clarifying commentary.

input/signdescriptor.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/95cac3f2e74a18bdeb664f2bb043a0b21e2bd266

lnwallet/commitment.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/95cac3f2e74a18bdeb664f2bb043a0b21e2bd266

lnwallet/commitment.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/95cac3f2e74a18bdeb664f2bb043a0b21e2bd266

lnwallet/commitment.go
halseth: Fixed: https://github.com/lightningnetwork/lnd/pull/3829/commits/5a2fe946e1a7af85895263c7ff6630a81d107af6

lnwallet/channel.go
```diff
@@ -2229,8 +2229,8 @@ func (lc *LightningChannel) fetchCommitmentView(remoteChain bool,
 	return c, nil
 }
 
-func (lc *LightningChannel) fundingTxIn() wire.TxIn {
-	return *wire.NewTxIn(&lc.channelState.FundingOutpoint, nil, nil)
+func fundingTxIn(chanState *channeldb.OpenChannel) wire.TxIn {
```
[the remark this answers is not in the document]
halseth: Figured it was better to leave it like this, restrict changes to this package.

lnwallet/channel.go
```diff
@@ -4995,7 +4995,7 @@ func genHtlcScript(isIncoming, ourCommit bool, timeout uint32, rHash [32]byte,
 // locate the added HTLC on the commitment transaction from the
 // PaymentDescriptor that generated it, the generated script is stored within
 // the descriptor itself.
-func (lc *LightningChannel) addHTLC(commitTx *wire.MsgTx, ourCommit bool,
```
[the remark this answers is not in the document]
halseth: Fixed

PR #3831, opened 2019-12-12
Add MPPTimeout ResolutionResult and expose to caller

carlaKC:
This change adds a `ResolutionResult` to the `hodlEvent` struct to provide calling functions with additional information about the outcome of an invoice update in the invoice registry. 

Purpose:
1. Choose wire message based on update result (FailInvalidDetails vs MPPTimeout)
2. More details for htlc notifier


lnwire/onion_error.go
```diff
@@ -1182,6 +1183,26 @@ func (f *InvalidOnionPayload) Encode(w io.Writer, pver uint32) error {
 	return WriteElements(w, f.Offset)
 }
 
+// FailMPPTimeout is returned if the complete amount for a multi part payment was
+// not receives within a reasonable time.
```
carlaKC: s: receives -> received

invoices/invoiceregistry.go
```diff
@@ -745,46 +756,35 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
-			AcceptHeight: acceptHeight,
-		}, nil
+		return hodlEvent, nil
 
 	case channeldb.HtlcStateSettled:
 		// Also settle any previously accepted htlcs. The invoice state
 		// is leading. If an htlc is marked as settled, we should follow
 		// now and settle the htlc with our peer.
-		for key, htlc := range invoice.Htlcs {
+		for _, htlc := range invoice.Htlcs {
```
joostjager: This change looks like it breaks something. There can be multiple htlcs on the invoice with distinct keys. With this change, all keys for the notifications will be the same.

invoices/invoiceregistry.go
```diff
@@ -745,46 +756,35 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+
+			i.notifyHodlSubscribers(*hodlEvent)
 		}
 
-		return &HodlEvent{
-			CircuitKey:   circuitKey,
-			Preimage:     &invoice.Terms.PaymentPreimage,
-			AcceptHeight: acceptHeight,
-		}, nil
+		return hodlEvent, nil
```
joostjager: Preimage is now only set inside the loop. I think it is okay, but the logic becomes more implicit.

I think it would be good to at least split this commit in two. One commit that introduces the 'default' hodlEvent and another one that adds the `ResolutionResult`.

We need to be careful in this area to not break anything unnecessarily. The 'override defaults' approach isn't as explicit as how it was before. Maybe it is safer to have some kind of helper function that still makes it very clear what the hodl event is that is returned.

htlcswitch/link.go
```diff
@@ -2817,21 +2820,7 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		invoiceHash, pd.Amount, pd.Timeout, int32(heightNow),
 		circuitKey, l.hodlQueue.ChanIn(), payload,
 	)
-
-	switch err {
-
-	// Cancel htlc if we don't have an invoice for it.
-	case channeldb.ErrInvoiceNotFound:
-		failure := lnwire.NewFailIncorrectDetails(pd.Amount, heightNow)
-		l.sendHTLCError(pd.HtlcIndex, failure, obfuscator, pd.SourceRef)
```
joostjager: Nice removal

invoices/invoiceregistry.go
```diff
@@ -740,9 +740,11 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
 			return updateDesc, nil
 		},
 	)
-	if err != nil {
+	if err == channeldb.ErrInvoiceNotFound {
```
joostjager: convert to switch?

invoices/invoiceregistry.go
```diff
@@ -857,9 +860,10 @@ func (i *InvoiceRegistry) SettleHodlInvoice(preimage lntypes.Preimage) error {
 		}
 
 		i.notifyHodlSubscribers(HodlEvent{
-			CircuitKey:   key,
-			Preimage:     &preimage,
-			AcceptHeight: int32(htlc.AcceptHeight),
+			CircuitKey:       key,
+			Preimage:         &preimage,
+			AcceptHeight:     int32(htlc.AcceptHeight),
+			ResolutionResult: ResultSettled,
```
joostjager: It seems this result isn't set when `NotifyExitHop` returns a settle directly.

I think that a constructor for `HodlEvent` that forces callers to populate all fields could be useful here, to prevent subtle omissions.

htlcswitch/link.go
```diff
@@ -1220,6 +1217,25 @@ func (l *channelLink) processHodlEvent(hodlEvent invoices.HodlEvent,
...
+// getHodlFailureMessage returns the wire message that a hodl event should be
+// failed with.
+func getHodlFailureMessage(event invoices.HodlEvent,
+	amount lnwire.MilliSatoshi) lnwire.FailureMessage {
+
+	// If the event has been resolved as part of a MPP timeout, we need to fail
+	// the htlc with lnwire.FailMppTimeout.
+	if event.ResolutionResult == invoices.ResultMPPTimeout {
+		return &lnwire.FailMPPTimeout{}
+	}
```
joostjager: Easy enough

invoices/invoiceregistry.go
carlaKC: I think that going with a `HodlEvent` constructor will be a better way to go, as you say default + override can introduce subtle changes.

invoices/invoiceregistry.go
carlaKC: > It seems this result isn't set when NotifyExitHop returns a settle directly.

Doesn't `updateLegacy` return `SettleEvent` when `NotifyExitHopHtlc` settled?

Or does this make it look like it isn't included? Not sure if I am following this comment.

joostjager: Yes, I thought there was a case where it wasn't included. But adding the constructors will surface that, if it was already the case and not my misunderstanding.

carlaKC: I could be missing it too. Going to do another check and see if I can add any more tests which check resResult then will re-request review.

carlaKC: Constructors added and `ResolutionResult` checks added to all tests that call `NotifyExitHopHtlc`.

invoices/invoiceregistry.go
```diff
@@ -50,6 +50,17 @@ type HodlEvent struct {
 	AcceptHeight int32
 }
 
+// NewHodlEvent returns a new hodl event with the values specified.
+func NewHodlEvent(preimage *lntypes.Preimage, key channeldb.CircuitKey,
+	acceptHeight int32) *HodlEvent {
```
joostjager: Just syntactic sugar, but a separate constructor for settles would be nice. It makes it more explicit what the resolution is and also allows us to set a fixed ResolutionResult Settled.

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -205,19 +205,21 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
 		return nil, err
 	}
 
+	defer h.Registry.HodlUnsubscribeAll(hodlChan)
+
+	// If the event is non-nil, and the invoice is known to the registry
+	// (indicating that the htlc is not a forward), try to resolve the htlc
+	// directly.
+	if event != nil &&
+		event.ResolutionResult != invoices.ResultInvoiceNotFound {
```
joostjager: I would move the InvoiceNotFound refactor to a separate commit after introducing the `ResolutionResult`

invoices/invoiceregistry.go
carlaKC: The settles constructor isn't quite as sugary because it still has to take a resolution result (since we have different settle resolution types for replays etc, and imo it's nice to have them for tests) but I think it's still nice enough to keep 🍭

joostjager:
Note: look at result interpretation of mpp timeout failure

lnwire/onion_error.go
```diff
@@ -1182,6 +1183,26 @@ func (f *InvalidOnionPayload) Encode(w io.Writer, pver uint32) error {
...
+	return CodeMPPTimeout
+}
+
+// Returns a human readable string describing the target FailureMessage.
+//
+// NOTE: Implements the error interface.
+func (f *FailMPPTimeout) Error() string {
+	return f.Code().String()
+}
+
```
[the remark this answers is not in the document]
carlaKC: What do we need encode/decode for? I thought they were only necessary for  `EncodeFailureMessage` and `DecodeFailureMessage` if there is additional information to be included (eg a channel update)? Since we optionally check `Serializable` in those methods? 

I've added it to `TestEncodeDecodeCode` (TIL this test exists) and it passes without adding encode/decode, just had to add the code to `makeEmptyOnionError`.

invoices/invoiceregistry.go
```diff
@@ -48,16 +48,21 @@ type HodlEvent struct {
 
 	// AcceptHeight is the original height at which the htlc was accepted.
 	AcceptHeight int32
+
+	// ResolutionResult indicates the outcome of the invoice registry update
+	// for the invoice.
+	ResolutionResult
```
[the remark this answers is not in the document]
carlaKC: While we're here, I'm wondering about the name `HodlEvent` generally, since we return a nil event for accepts it's not really an event in the typical sense because it doesn't cover the full set events that can happen for a hodlInvoice (settle/cancel/accept). This struct is more of a `HodlResolution` since it's only populated for settles/cancels.

Thoughts on a rename of `HodlEvent` -> `HodlResolution` and then this field can just be `Outcome`/ `Result`? `HodlResolution.Outcome`, reads nicely imo. cc @joostjager ?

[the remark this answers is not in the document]
joostjager: Nice thinking. Yes, it definitely is a resolution. But it isn't even specific to hodling. Maybe it should just be `HtlcResolution` and `HtlcResolution.Outcome`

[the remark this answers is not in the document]
carlaKC: Cool, I'll add another commit renaming :)
`processHodlEvent` -> `processHTLCResolution` will also be a nice rename.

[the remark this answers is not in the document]
joostjager: I am no fan of `HTLC`. For me it is dictionary word now and I write `Htlc` to make it not stand out so much. But that is a personal preference.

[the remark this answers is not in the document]
carlaKC: `Htlc` in the oxford dictionary 2020 🤔

carlaKC:
Ready for another look. I haven't added nop `Encode`/`Decode` because `FailMppTimeout` passes the unit test without them (will happily add, just not sure if it's needed when there is no metadata).

Ended up doing a fair amount of renaming in this one, so I have split the test renames into separate commits so that it's hopefully more digestible to review.

joostjager:
Looks like the commit order is off again.

contractcourt/htlc_incoming_resolver_test.go
```diff
@@ -96,10 +97,10 @@ func TestHtlcIncomingResolverExitSettle(t *testing.T) {
 	defer timeout(t)()
 
 	ctx := newIncomingResolverTestContext(t)
-	ctx.registry.notifyResolution = &invoices.HtlcResolution{
-		CircuitKey: testResCircuitKey,
-		Preimage:   &testResPreimage,
-	}
+	ctx.registry.notifyResolution = invoices.NewSettleResolution(
+		&testResPreimage, testResCircuitKey, testAcceptHeight,
```
joostjager: For settle, preimage is non-nil and can be passed by value.

invoices/invoiceregistry.go
```diff
@@ -50,6 +50,28 @@ type HtlcResolution struct {
 	AcceptHeight int32
 }
 
+// NewHtlcResolution returns a non-settle htlc resolution.
+func NewHtlcResolution(key channeldb.CircuitKey,
```
joostjager: Non-settle, that is fail right? `NewFailResolution`?

invoices/update.go
```diff
@@ -49,6 +49,10 @@ const (
 	// ResultSettled is returned when we settle an invoice.
 	ResultSettled
 
+	// ResultCancelled is returned when we cancel an invoice and its associated
+	// htlcs.
+	ResultCancelled
```
joostjager: We did a recent full repo rename to standardize on `Canceled`

invoices/update.go
```diff
@@ -59,6 +63,10 @@ const (
 	// ResultMppInProgress is returned when we are busy receiving a mpp payment.
 	ResultMppInProgress
 
+	// ResultMPPTimeout is returned when an invoice paid with multiple partial
+	// payments times out before it is fully paid.
+	ResultMPPTimeout
```
joostjager: `MPP` all upper not consistent with `ResultMppInProgress`

invoices/invoiceregistry.go
```diff
@@ -48,27 +48,32 @@ type HtlcResolution struct {
...
 
 // NewHtlcResolution returns a non-settle htlc resolution.
 func NewHtlcResolution(key channeldb.CircuitKey,
-	acceptHeight int32) *HtlcResolution {
+	acceptHeight int32, outcome ResolutionResult) *HtlcResolution {
 
 	return &HtlcResolution{
 		CircuitKey:   key,
 		AcceptHeight: acceptHeight,
+		Outcome:      outcome,
```
joostjager: If you really like constructors, you could add a check here to see if `outcome` is a settle-specific outcome. Same for other constructor.

To not incur runtime overhead, it may also work by splitting up the `ResolutionResult` type in two (internal) types `SettleResult` and `FailResult`. The compiler can then catch misuse and in the constructor it can be cast into the more generic `ResolutionResult`.

But this is totally optional.

lnwire/onion_error.go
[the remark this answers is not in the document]
joostjager: Maybe it is worth validating that `lnd` as a sender doesn't do anything strange when it receives this failure. The failure is for example serialized in mission control.

htlcswitch/link.go
```diff
@@ -1223,6 +1219,25 @@ func (l *channelLink) processHtlcResolution(resolution invoices.HtlcResolution,
...
 
+// getResolutionFailure returns the wire message that a htlc resolution should
+// be failed with.
+func getResolutionFailure(resolution invoices.HtlcResolution,
+	amount lnwire.MilliSatoshi) lnwire.FailureMessage {
+
+	// If the resolution has been resolved as part of a MPP timeout, we need
+	// to fail the htlc with lnwire.FailMppTimeout.
+	if resolution.Outcome == invoices.ResultMPPTimeout {
+		return &lnwire.FailMPPTimeout{}
```
joostjager: Below we use a constructor. Don't know if there are already constructors for every failure type.

carlaKC: We tend to not have them for failures with no fields, `FailFinalExpiryTooSoon` has a constructor but it's the exception among the no field failures. The majority is no fields -> no constructor so I'm going to leave as is.

invoices/invoiceregistry.go
carlaKC: I'm going to leave this one, I think the setting of results is sufficiently tested (ie, we're setting settles where we should) and the name/presence of preimage is a strong enough indicator for future code. 

I'd worry about setting up aliased types then casting in the constructor because we can't use iota across types so would have to declare mutually exclusive settle/fail result int values. Which would probably be fine, but would be a very nasty bug if it snuck in.

invoices/update.go
carlaKC: Agreed on matching case in this enum. We already have `MPP` and `Mpp` in the codebase, maybe we should pick one?

invoices/update.go
carlaKC: This spelling hurts me but ok :(

carlaKC:
> Looks like the commit order is off again.

Back in order 🕐 

> Only serious comment is to check whether we handle the new failure message as a receiver. We don't expect to receive it, but it can happen.

Updated the mission control test to check that we're ok on that side of things. Also ran a hideous hack locally returning `FailMppTimeout' for any failed invoice and that seemed to be ok, sender side logs: `RPCS: Unable to send payment: MPP Timeout@1`.

invoices/invoiceregistry.go
joostjager: Good point

invoices/update.go
joostjager: I don't know if `Mpp` is also in the Oxford 2020. I would prefer that, but it is your call.

htlcswitch/link.go
joostjager: Ok

invoices/invoiceregistry.go
```diff
@@ -1201,9 +1201,10 @@ func (i *InvoiceRegistry) SubscribeSingleInvoice(
 	return client, nil
 }
 
-// notifyHodlSubscribers sends out the hodl event to all current subscribers.
-func (i *InvoiceRegistry) notifyHodlSubscribers(hodlEvent HodlEvent) {
-	subscribers, ok := i.hodlSubscriptions[hodlEvent.CircuitKey]
+// notifyHodlSubscribers sends out the htlc resolution to all current
+// subscribers.
```
joostjager: 'htlc resolution' definitely reads better than 'hodl event' :+1:

invoices/invoiceregistry.go
```diff
@@ -789,13 +796,13 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
 
 			i.notifyHodlSubscribers(
 				*NewSettleResolution(
-					invoice.Terms.PaymentPreimage, key, acceptHeight,
+					invoice.Terms.PaymentPreimage, key, acceptHeight, result,
```
joostjager: Small question mark here whether all the previously added htlcs should be settled with this same result? Perhaps add a comment detailing this.

joostjager: Check line lengths in this file

lnrpc/routerrpc/router.proto
```diff
@@ -254,6 +254,7 @@ message Failure {
         PERMANENT_NODE_FAILURE = 20;
         PERMANENT_CHANNEL_FAILURE = 21;
         EXPIRY_TOO_FAR = 22;
+        MPP_TIMEOUT = 23;
```
joostjager: Ah nice, there is also this indeed.

channeldb/payments.go
```diff
@@ -109,6 +109,10 @@ const (
 	// balance to complete the payment.
 	FailureReasonInsufficientBalance FailureReason = 4
 
+	// FailureReasonMPPTimeout indicates that we attempted a mpp payment which
+	// timed out.
+	FailureReasonMPPTimeout FailureReason = 5
```
joostjager: So these are payment-level failures. I don't think that mpp timeout is a valid final payment-level failure, because we just need to try again (eventually). I would for now just map that onto `FailureReasonError`.

carlaKC: Seems like a slightly strange mapping to me because `FailureReasonError` in channelDB is for unexpected errors (according to its comment at least). But it's good enough since we don't expect MPP timeouts.

invoices/invoiceregistry.go
carlaKC: If this is the first time we are settling the htlc, the other htlcs will correctly be notified with `ResultSettled` (because they are also being settled for the first time). On replays of the htlc settle, the already settled htlcs will not notify, although there could be some race between the settle and replay?

I think the only way this can be inconsistent is if we settle a htlc and have a failure before we notify the other htlcs and then reply it. They will be notified with `ReplaySettle` even though they are being settled for the first time. 

I'll add a comment for this, it shouldn't be too much of an issue since I can't see us using settle result vs replaySettle for anything actionable.

carlaKC: They look fine to me? Anything sticking out?

joostjager: Ok, sounds good.

joostjager: Maybe your tab size is 4? It should be 8

channeldb/payments.go
joostjager: That is indeed slightly strange. We currently mix up unexpected (internal) errors and certain received failure messages (search for `reasonError` in `result_interpretation.go`). Needs to be fixed at some point.

joostjager:
Commits look out of order. You need that script :)

invoices/invoiceregistry.go
carlaKC: Should it? Contribution guidelines say 4 [here]( Editors should be set to treat a tab as 4 spaces.)?

joostjager: I am sure we had a discussion about that some time ago and that it was def 8

lnwire/onion_error.go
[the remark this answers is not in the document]
cfromknecht: @carlaKC you're right my mistake, i see that the other `struct{}` failures also don't have `Encode/Decode` since we don't serialize the messages directly to the wire

routing/result_interpretation.go
```diff
@@ -216,6 +217,12 @@ func (i *interpretedResult) processPaymentOutcomeFinal(
 		// deliberately. What to penalize?
 		i.finalFailureReason = &reasonIncorrectDetails
 
+	case *lnwire.FailMPPTimeout:
```
cfromknecht: i don't think we need to handle this quite yet because the router never sends MPP payments. we only need this if we send HTLCs smaller than `total_amt_msat` which isn't the case rn. that will allow us to defer final judgement on how to penalize as well

joostjager: There is the crowd source case? The idea of doing this now is that we at least don't blacklist our destination.

Btw, marking all the other hops as success can still happen for mpptimeout.

cfromknecht: yes good point, discussed more offline we still want it since sendtoroute can feed into mission control and bc other implementations will send this error. 

longer term I’m not sure if we need to penalize for mpp_timeout. the router or sendtorroute user should try again until they feel like giving up. so I think this is good as is?

joostjager: As it is now in this pr, we don't penalize for mpp_timeout. we just abort the payment. that should indeed change in the future.

I see that the scucess marking was indeed already in.

carlaKC: Also not sure about penalizing for mpp timeout, since we're failing a payment based on another payment not arriving. Difficult to attribute blame there 🤔 

As is, just updated to attribute success to all hops except the last one and just take no action on the last hop? I think that's nice because we still get all the successes from the intermediate hops, and then take no action on the one we're unsure about (which is just a single hop, so we aren't losing a large amount of insight).

joostjager: There is also a lingering bug in here: https://github.com/lightningnetwork/lnd/pull/3863

carlaKC: Ah I was wondering about that!

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -169,25 +169,25 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
 
 	// Define closure to process htlc resolutions either direct or triggered by
 	// later notification.
-	processHodlEvent := func(e invoices.HtlcResolution) (ContractResolver,
-		error) {
+	processHtlcResolution :=
```
halseth: style nit:
```go
processHtlcResolution := func(e invoices.HtlcResolution) (
       ContractResolver, error) {
```

Would that help with the extra identation also? 🤔

contractcourt/htlc_incoming_contest_resolver.go
```diff
@@ -201,23 +201,25 @@ func (h *htlcIncomingContestResolver) Resolve() (ContractResolver, error) {
...
+	if err != nil {
 		return nil, err
 	}
 
+	defer h.Registry.HodlUnsubscribeAll(hodlChan)
+
+	// If the event is non-nil, and the invoice is known to the registry
+	// (indicating that the htlc is not a forward), try to resolve the htlc
+	// directly.
+	if resolution != nil &&
```
halseth: can both the resolution and err be nil?

htlcswitch/link.go
```diff
@@ -2823,21 +2823,7 @@ func (l *channelLink) processExitHop(pd *lnwallet.PaymentDescriptor,
 		invoiceHash, pd.Amount, pd.Timeout, int32(heightNow),
 		circuitKey, l.hodlQueue.ChanIn(), payload,
 	)
-
-	switch err {
-
-	// Cancel htlc if we don't have an invoice for it.
```
halseth: not obvious to me why this can be removed? Add explanation to commit message perhaps?

contractcourt/htlc_incoming_contest_resolver.go
carlaKC: Yes, this comment should actually be re-phrased to say `if the resolution is non-nil`, will fix it now. 

We have a nil resolution and nil error when we are holding the htlc.

contractcourt/htlc_incoming_contest_resolver.go
carlaKC: yeah that's much better!

htlcswitch/link.go
joostjager: This is not removed, but moved to invoice registry. It will return a cancel resolution if the invoice is not found.

halseth: Ah, I see it now. Thanks!

PR #3834, opened 2019-12-13
feature: add transitive feature bit validation

cfromknecht:
Split off from #3679

This PR adds a new `feature.ValidateDeps` method to the `feature` package, which can be used to ensure that all transitive dependencies of a given feature are advertised in the feature vector. With base amp, for example, the dependency chain is `option_mpp` -> `option_payment_addr` -> `option_onion_tlv`. When validating, we want to fail if `option_mpp` is set but either of two dependencies are not. As more feature bits are added, these chains could become several layers deep. Things like AMP or extended gossip queries will continue to add to these constraints, and this gives a simple way of expressing and validating these relations.

We then add this transitive validation to remote init features and when parsing BOLT11 invoices, additionally verifying that the payment address is set if the corresponding feature bit is advertised.

### DIFF
--- a/feature/deps.go
+++ b/feature/deps.go
@@ -1,0 +1,124 @@
+package feature
+
+import (
+	"fmt"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+type (
+	// featureSet contains a set of feature bits.
+	featureSet map[lnwire.FeatureBit]struct{}
+
+	// supportedFeatures maps the feature bit from a feature vector to a
+	// boolean indicating if this features dependencies have already been
+	// verified. This allows us to short circuit verification if multiple
+	// features have common dependencies, or map traversal starts verifying
+	// from the bottom up.
+	supportedFeatures map[lnwire.FeatureBit]bool
+
+	// depDesc maps a features to its set of dependent features, which must
+	// also be present for the vector to be valid. This can be used to
+	// recursively check the dependency chain for features in a feature
+	// vector.
+	depDesc map[lnwire.FeatureBit]featureSet
+)
+
+// ErrMissingFeatureDep is an error signaling that a transitive dependency in a
+// feature vector is not set properly.
+type ErrMissingFeatureDep struct {
+	dep lnwire.FeatureBit
+}
+
+// Error returns a human-readable description of the missing dep error.
+func (e ErrMissingFeatureDep) Error() string {
+	return fmt.Sprintf("missing feature dependency: %v", e.dep)
+}
+
+// deps is the default set of dependencies for assigned feature bits. If a
+// feature is not present in the depDesc it is assumed to have no dependencies.
+//
+// NOTE: For proper functioning, only the optional variant of feature bits
+// should be used in the following descriptor. In the future it may be necessary
+// to distinguish the dependencies for optional and required bits, but for now
+// the validation code maps required bits to optional ones since it simplifies
+// the number of constraints.
+var deps = depDesc{
+	lnwire.PaymentAddrOptional: {
+		lnwire.TLVOnionPayloadOptional: {},
+	},
+	lnwire.MPPOptional: {
+		lnwire.PaymentAddrOptional: {},
+	},
+}
+
+// ValidateDeps asserts that a feature vector sets all features and their
+// transitive dependencies properly. It assumes that the dependencies between
+// optional and required features are identical, e.g. if a feature is required
+// but its dependency is optional, that is sufficient.
+func ValidateDeps(fv *lnwire.FeatureVector) error {
+	features := fv.Features()
+	supported := initSupported(features)
+
+	return validateDeps(features, supported)
+}
+
+// validateDeps is a subroutine that recursively checks that the passed features
+// have all of their associated dependencies in the supported map.
+func validateDeps(features featureSet, supported supportedFeatures) error {
+	for bit := range features {
+		// Convert any required bits to optional.
+		bit = mapToOptional(bit)
+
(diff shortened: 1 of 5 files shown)


feature/deps.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ErrMissingFeatureDep is an error signaling that a transitive dependency in a
+// feature vector is not set properly.
+type ErrMissingFeatureDep struct {
+	dep lnwire.FeatureBit
+}
+
+// Error returns a human-readable description of the missing dep error.
+func (e ErrMissingFeatureDep) Error() string {
+	return fmt.Sprintf("missing feature: %v", e.dep)
```
joostjager: Maybe update msg to "missing feature dependency"?

feature/deps.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// deps is the default set of dependencies for assigned feature bits. If a
+// feature is not present in the depDesc it is assumed to have no depdencies.
+var deps = depDesc{
```
joostjager: Add in comment that this must only contain the optional variants

feature/deps.go
```diff
@@ -0,0 +1,109 @@
...
+
+// initChecked sets all bits from the feature vector as known but not visited.
+// This signals that the validity of their dependencies has not been verified.
+// All required bits are mapped to optional to simplify the DAG.
+func initChecked(features featureSet) checkedFeatures {
+	checked := make(checkedFeatures)
+	for bit := range features {
+		if bit.IsRequired() {
+			bit ^= 0x01
+		}
```
joostjager: This could be a (local) helper func, to make it even more clear why we do this.

feature/deps.go
```diff
@@ -0,0 +1,109 @@
...
+
+		// If the checked features doesn't contain the dependency, this
+		// vector is invalid. Alternatively, if we know that this
+		// dependencies are valid, we can short circuit and continue
+		// verifying other bits.
+		visited, ok := checked[bit]
+		if !ok {
+			return ErrMissingFeatureDep{bit}
```
joostjager: Maybe the name `checked` isn't sufficiently covered what it does. Besides checked, it also is a quick look up into the features that we have. I didn't immediately get that because of the name.

feature/manager.go
```diff
@@ -63,14 +63,27 @@ func newManager(cfg Config, desc setDesc) (*Manager, error) {
...
+	for set, raw := range fsets {
 		if cfg.NoTLVOnion {
-			fv.Unset(lnwire.TLVOnionPayloadOptional)
-			fv.Unset(lnwire.TLVOnionPayloadRequired)
+			raw.Unset(lnwire.TLVOnionPayloadOptional)
+			raw.Unset(lnwire.TLVOnionPayloadRequired)
+			raw.Unset(lnwire.PaymentAddrOptional)
+			raw.Unset(lnwire.PaymentAddrRequired)
+			raw.Unset(lnwire.MPPOptional)
+			raw.Unset(lnwire.MPPRequired)
```
joostjager: This change looks like it belongs more to the first commit where the bits are added?

cfromknecht: we haven't begun to set `PaymentAddr` or `MPP` in any of our feature sets, that doesn't happen until the end. So this is the preemptive commit, that allows the one-line changes later

cfromknecht:
> My main question is whether we need to validate features also at the remaining entry points. Two that are missing are:
>  * paying an invoice
>  * node announcement gossip

Paying an invoice should be covered since we validate payreq features while decoding. Even if the invoice is paid manually, this is checked during `validate()` of #3679.

Node announcements are a little trickier, but we could start to validate them when we receive them. We will still have older announcements that will be grandfathered into our routing tables even if they are invalid. This can also happen if we update our dependencies that retroactively makes announcements invalid. For these reasons, I went with just validating them in the router as they're being used, but maybe there is a better long term strategy.

Do you think we should add the dependency validation on new node anns? Perhaps it makes sense, there's no reason to store them if we will end up failing to build a path/route with them.

joostjager:
Hmm, yes the stored announcements are a problem. Even if we start treating new ones differently, we still can't leave out the check for the old ones. Probably best then to leave it as you do it now, when we use them.

peer.go
```diff
@@ -2476,6 +2477,14 @@ func (p *peer) handleInitMsg(msg *lnwire.Init) error {
 		return err
 	}
 
+	// Ensure the remote party's feature vector contains all transistive
+	// dependencies. We know ours are are correct since they are validated
+	// during the feature manager's instantiation.
+	err = feature.ValidateDeps(p.remoteFeatures)
+	if err != nil {
+		return err
```
halseth: should we add some more context to the error message here?

lnwire/features.go
```diff
@@ -81,6 +81,26 @@ const (
...
+	// node requires payment addresses, which are used to mitigate probing
+	// attacks on the receiver of a payment.
+	PaymentAddrRequired FeatureBit = 14
+
+	// PaymentAddrOptional is an optional feature bit that signals that a
+	// node supports payment addresses, which are used to mitigate probing
+	// attacks on the receiver of a payment.
+	PaymentAddrOptional FeatureBit = 15
+
+	// MPPOptional is a required feature bit that signals that the receiver
```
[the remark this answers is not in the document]
cfromknecht: i think it's correct, since is that says more than one is supported, but doesn't say that only one is not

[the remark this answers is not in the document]
halseth: > requires settlement of an invoice with more than one HTLC.

PR #3836, opened 2019-12-14
discovery: use block ranges from ReplyChannelRange to determine end marker

wpaulino:
In this PR, we make a series of changes to properly adhere to the specification with the `QueryChannelRange` and `ReplyChannelRange` messages. Previously, we'd look at the `Complete` field of a `ReplyChannelRange` message to determine if we had received all replies to a query. This isn't the use of the field however, and instead we should be inspecting the block range of each `ReplyChannelRange` message to determine this.

One of the main takeaways that motivates most of the changes outlined in this PR is the following:

```
  - MUST respond with one or more `reply_channel_range` whose combined range
	cover the requested `first_blocknum` to `first_blocknum` plus
	`number_of_blocks` minus one.
```

Due to the large number of nodes out there that rely on the previous mechanism for graph sync, we still maintain it for backwards compatibility. In the future, once most nodes have upgraded, we can consider removing this.

I did a series of graph sync tests and I was able to successfully sync against other `lnd` nodes with and without this change, and `c-lightning` nodes. Syncs against `eclair` failed as they don't adhere to the point above, their responses don't cover the complete requested range.

This is a follow-up of https://github.com/lightningnetwork/lnd/pull/3785 and fixes https://github.com/lightningnetwork/lnd/issues/3728.

### DIFF
--- a/discovery/syncer.go
+++ b/discovery/syncer.go
@@ -825,13 +825,13 @@ func (g *GossipSyncer) replyChanRangeQuery(query *lnwire.QueryChannelRange) erro
 	// TODO(roasbeef): means can't send max uint above?
 	//  * or make internal 64
 
-	// In the base case (no actual response) the first block and the last
-	// block in the query will be the same. In the loop below, we'll update
-	// these two variables incrementally with each chunk to properly
-	// compute the starting block for each response and the number of
-	// blocks in a response.
-	firstBlockHeight := query.FirstBlockHeight
-	lastBlockHeight := query.FirstBlockHeight
+	// In the base case (no actual response) the first block and last block
+	// will match those of the query. In the loop below, we'll update these
+	// two variables incrementally with each chunk to properly compute the
+	// starting block for each response and the number of blocks in a
+	// response.
+	firstBlockHeight := startBlock
+	lastBlockHeight := endBlock
 
 	numChannels := int32(len(channelRange))
 	numChansSent := int32(0)
@@ -867,8 +867,25 @@ func (g *GossipSyncer) replyChanRangeQuery(query *lnwire.QueryChannelRange) erro
 		// update our pointers to the first and last blocks for each
 		// response.
 		if len(channelChunk) > 0 {
-			firstBlockHeight = channelChunk[0].BlockHeight
-			lastBlockHeight = channelChunk[len(channelChunk)-1].BlockHeight
+			// If this is the first response we'll send, we'll point
+			// the first block to the first block in the query.
+			// Otherwise, we'll continue from the block we left off
+			// at.
+			if numChansSent == 0 {
+				firstBlockHeight = startBlock
+			} else {
+				firstBlockHeight = lastBlockHeight
+			}
+
+			// If this is the last response we'll send, we'll point
+			// the last block to the last block of the query.
+			// Otherwise, we'll set it to the height of the last
+			// channel in the chunk.
+			if isFinalChunk {
+				lastBlockHeight = endBlock
+			} else {
+				lastBlockHeight = channelChunk[len(channelChunk)-1].BlockHeight
+			}
 		}
 
 		// The number of blocks contained in this response (the total
--- a/discovery/syncer.go
+++ b/discovery/syncer.go
@@ -298,6 +298,17 @@ type GossipSyncer struct {
 	// received over, these will be read by the replyHandler.
 	queryMsgs chan lnwire.Message
 
+	// curQueryRangeMsg keeps track of the latest QueryChannelRange message
+	// we've sent to a peer to ensure we've consumed all expected replies.
+	// This field is primarily used within the waitingQueryChanReply state.
+	curQueryRangeMsg *lnwire.QueryChannelRange
+
+	// prevReplyChannelRange keeps track of the previous ReplyChannelRange
+	// message we've received from a peer to ensure they've fully replied to
+	// our query by ensuring they covered our requested block range. This
+	// field is primarily used within the waitingQueryChanReply state.
+	prevReplyChannelRange *lnwire.ReplyChannelRange
+
 	// bufferedChanRangeReplies is used in the waitingQueryChanReply to
 	// buffer all the chunked response to our query.
 	bufferedChanRangeReplies []lnwire.ShortChannelID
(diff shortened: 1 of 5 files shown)


discovery/syncer.go
```diff
@@ -669,10 +680,70 @@ func (g *GossipSyncer) synchronizeChanIDs() (bool, error) {
...
+// communicating with a legacy node so we can properly sync with them.
+func isLegacyReplyChannelRange(query *lnwire.QueryChannelRange,
+	reply *lnwire.ReplyChannelRange) bool {
+
+	return reply.QueryChannelRange == *query
+}
+
+// lastHeightInRange is a helper function to return the last block height of a
+// QueryChannelRange message.
+func lastHeightInRange(msg *lnwire.QueryChannelRange) uint32 {
```
cfromknecht: could be method on `lnwire.QueryChannelRange`?

peer.go
```diff
@@ -1335,8 +1335,10 @@ func messageSummary(msg lnwire.Message) string {
 			msg.Complete)
 
 	case *lnwire.ReplyChannelRange:
-		return fmt.Sprintf("complete=%v, encoding=%v, num_chans=%v",
-			msg.Complete, msg.EncodingType, len(msg.ShortChanIDs))
+		return fmt.Sprintf("start_height=%v, end_height=%v, "+
+			"num_chans=%v, encoding=%v", msg.FirstBlockHeight,
+			msg.FirstBlockHeight+msg.NumBlocks-1,
```
cfromknecht: can use `LastHeightInRange()`?

discovery/syncer.go
wpaulino: Good idea, done.

peer.go
wpaulino: Done.

lnwire/query_channel_range.go
```diff
@@ -75,3 +75,9 @@ func (q *QueryChannelRange) MaxPayloadLength(uint32) uint32 {
 	// 32 + 4 + 4
 	return 40
 }
+
+// LastBlockHeight returns the last block height covered by the range of a
+// QueryChannelRange message.
```
halseth: nit: note that the range is inclusive

Roasbeef:
> Syncs against eclair failed as they don't adhere to the point above, their responses don't cover the complete requested range.

Shouldn't the logic to support legacy `lnd` nodes also cover the quirks in eclair's implementation?

discovery/syncer.go
```diff
@@ -814,8 +814,9 @@ func (g *GossipSyncer) replyChanRangeQuery(query *lnwire.QueryChannelRange) erro
 	// Next, we'll consult the time series to obtain the set of known
 	// channel ID's that match their query.
 	startBlock := query.FirstBlockHeight
+	endBlock := startBlock + query.NumBlocks - 1
```
Roasbeef: Nice catch!

discovery/syncer_test.go
```diff
@@ -783,52 +783,52 @@ func TestGossipSyncerReplyChanRangeQuery(t *testing.T) {
...
-			// other iteration, we want to move forward by the
-			// chunkSize (from the staring block height).
-			offset := 0
-			if i != 0 {
-				offset = 1
-			}
-			expectedFirstBlockHeight := (i+offset)*2 + startingBlockHeight
+			// We'll determine the correct values of each field in
+			// each response based on the order that they were sent.
+			var expectedFirstBlockHeight, expectedNumBlocks uint32
```
Roasbeef: Style nit: could be collapsed into a single `var` statement.

discovery/syncer_test.go
```diff
@@ -783,52 +783,52 @@ func TestGossipSyncerReplyChanRangeQuery(t *testing.T) {
...
-				queryResp[len(queryResp)-1].BlockHeight:
-
-				t.Fatalf("FirstBlockHeight incorrect: "+
-					"expected %v got %v",
-					rangeResp.FirstBlockHeight,
-					queryResp[len(queryResp)-1].BlockHeight)
+			// The first reply should range from our starting block
+			// height until it reaches its maximum capacity of
+			// channels.
+			case i == 0:
```
Roasbeef: Great simplification of the prior `switch` statement, I had an itch to massage it a bit when I touched this area last, but ended up deferring it.

discovery/syncer.go
```diff
@@ -750,17 +753,18 @@ func (g *GossipSyncer) genChanRangeQuery(
...
+		bestHeight-startHeight)
 
 	// Finally, we'll craft the channel range query, using our starting
 	// height, then asking for all known channels to the foreseeable end of
 	// the main chain.
 	return &lnwire.QueryChannelRange{
 		ChainHash:        g.cfg.chainHash,
 		FirstBlockHeight: startHeight,
-		NumBlocks:        math.MaxUint32 - startHeight,
+		NumBlocks:        bestHeight - startHeight,
```
Roasbeef: So from reading the commit message, this change actually isn't required at all? IIRC CL also sends a max uint32 num blocks in their requests. Alternatively, could've we just use the fact that the num blocks message extends beyond our current best known chain tip as a termination condition?

discovery/syncer.go
```diff
@@ -669,10 +680,64 @@ func (g *GossipSyncer) synchronizeChanIDs() (bool, error) {
...
 
+// isLegacyReplyChannelRange determines where a ReplyChannelRange message is
+// considered legacy. There was a point where lnd used to include the same query
+// over multiple replies, rather than including the portion of the query the
+// reply is handling. We'll use this as a way of detecting whether we are
+// communicating with a legacy node so we can properly sync with them.
+func isLegacyReplyChannelRange(query *lnwire.QueryChannelRange,
+	reply *lnwire.ReplyChannelRange) bool {
+
+	return reply.QueryChannelRange == *query
```
Roasbeef: For cases where the range requested fits into a single response, won't this trigger a false positive? I think this only applies to our new message pattern as we'll no longer use max uint32.

discovery/syncer.go
```diff
@@ -669,10 +680,64 @@ func (g *GossipSyncer) synchronizeChanIDs() (bool, error) {
...
+			return fmt.Errorf("reply includes channels for height "+
+				"%v after query %v", replyLastHeight,
+				queryLastHeight)
+		}
+
+		// If we've previously received a reply for this query, look at
+		// its last block to ensure the current reply properly follows
+		// it.
+		if g.prevReplyChannelRange != nil {
+			prevReplyLastHeight := g.prevReplyChannelRange.
```
Roasbeef: Weird line folding, might as well use two lines to unpack the variables.

discovery/syncer.go
```diff
@@ -669,10 +680,64 @@ func (g *GossipSyncer) synchronizeChanIDs() (bool, error) {
...
+				QueryChannelRange.LastBlockHeight()
+
+			// The current reply can either start from the previous'
```
Roasbeef: extra apostrophe at the end of the line

discovery/syncer.go
```diff
@@ -669,10 +680,64 @@ func (g *GossipSyncer) synchronizeChanIDs() (bool, error) {
...
+			// reply's last block, if there are still more channels
+			// for the same block, or the block after.
+			if msg.FirstBlockHeight != prevReplyLastHeight &&
```
Roasbeef: Why are both variatns allowed? Feels like the spec should be stricter in its requirements here (either overlapping ranges are allowed or they aren't). Or is this an artefact from resolving the issues with the implicit inclusivity bounds?

peer.go
```diff
@@ -1336,17 +1336,19 @@ func messageSummary(msg lnwire.Message) string {
...
 
 	case *lnwire.QueryShortChanIDs:
 		return fmt.Sprintf("chain_hash=%v, encoding=%v, num_chans=%v",
 			msg.ChainHash, msg.EncodingType, len(msg.ShortChanIDs))
 
 	case *lnwire.QueryChannelRange:
 		return fmt.Sprintf("chain_hash=%v, start_height=%v, "+
-			"num_blocks=%v", msg.ChainHash, msg.FirstBlockHeight,
-			msg.NumBlocks)
+			"end_height=%v", msg.ChainHash, msg.FirstBlockHeight,
```

Roasbeef: Actually, what does this and the above look like for older nodes that send `math.MaxUint32`?

lnwire/query_channel_range.go
Roasbeef: What about instances where `NumBlocks` is `math.MaxUint32`?

discovery/syncer.go
wpaulino: Indeed, it's not needed after all. Termination is now based on whether the last reply received satisfies the queried range end height.

discovery/syncer.go
wpaulino: Good point, but we don't know if our request will span multiple replies at this point.

discovery/syncer.go
wpaulino: This can happen if there are too many short channel IDs for a given range that exceed the maximum we're allowed to send in a single message.

peer.go
wpaulino: It overflows, should be fixed now.

wpaulino:
>> Syncs against eclair failed as they don't adhere to the point above, their responses don't cover the complete requested range.

> Shouldn't the logic to support legacy lnd nodes also cover the quirks in eclair's implementation?

The legacy way is based on the same query channel range message always being sent, whereas eclair implements it correctly by replying with a query channel range message that corresponds to the short channel IDs sent, but their replies don't cover the complete requested range.

discovery/syncer.go
Roasbeef: However, if the change from max uint32 was reverted, then we should properly fall into this legacy replay bucket.

PR #3837, opened 2019-12-14
lnrpc: display node and peer features via rpc

cfromknecht:
This PR exposes features in three new places on over rpc:
 * `GetNodeInfo`: node features
 * `DescribeGraph`: node features
 * `ListPeers`: remote init features

In addition, all rpc feature vectors are now displayed as maps instead of lists since this is more useful when interacting with the rpc responses and closer to how they are represented internally.


rpcserver.go
```diff
@@ -4137,13 +4138,16 @@ func (r *rpcServer) GetNodeInfo(ctx context.Context,
 		nodeAddrs = append(nodeAddrs, nodeAddr)
 	}
 
+	features := invoicesrpc.CreateRPCFeatures(node.Features)
```
carlaKC: Can inline this in `NodeInfo`?

lnrpc/rpc.proto
```diff
@@ -2069,6 +2069,7 @@ message LightningNode {
     string alias = 3 [ json_name = "alias" ];
     repeated NodeAddress addresses = 4 [ json_name = "addresses" ];
     string color = 5 [ json_name = "color" ];
+    repeated Feature features = 6 [ json_name = "features" ];
```
carlaKC: nit: we have `[json_name` (without a space) more than `[ json_name` (with space) in `rpc.proto`, not sure of which is correct, just noticed it.

cfromknecht: yeah lol both styles are used throughout the file, i just tried to match the surrounding definitions

rpcserver.go
cfromknecht: ends up being 1 char too long :/

PR #3694, opened 2019-11-08
invoices+channeldb: reject payments to expired invoices

bhandras:
invoices+channeldb: reject payments to expired invoices

---
This change is [Reviewable](https://reviewable.io/reviews/lightningnetwork/lnd/3694)

PR #3838, opened 2019-12-14
channeldb+invoices: fix what pending invoice means in ChannelDB

bhandras:
This pull request is a partial fix for issue lightningnetwork#1404 where part of the problem comes from
not being able to filter out expired invoices due to those not being canceled 
(fixed in PR lightningnetwork#3694) while also having a misleading definition of what pending invoice
state means.

The PR changes the meaning of the `pending_only` flag in `listinvoices`
RPC to list invoices that are either in `ContractOpen` or `ContractAccepted`
state. Previously when the `pending_only` flag was set the `listinvoices`
RPC returned all invoices that are not settled which equals to invoices in `ContractOpen`, `ContractAccepted` or `ContractCanceled` state.

### DIFF
--- a/channeldb/invoice_test.go
+++ b/channeldb/invoice_test.go
@@ -52,6 +52,29 @@ func randInvoice(value lnwire.MilliSatoshi) (*Invoice, error) {
 	return i, nil
 }
 
+// Tests that pending invoices are those which are either in ContractOpen or
+// in ContractAccepted state.
+func TestInvoiceIsPending(t *testing.T) {
+	contractStates := []ContractState{
+		ContractOpen, ContractSettled, ContractCanceled, ContractAccepted,
+	}
+
+	for _, state := range contractStates {
+		invoice := Invoice{
+			State: state,
+		}
+
+		// We expect that an invoice is pending if it's either in ContractOpen
+		// or ContractAccepted state.
+		pending := (state == ContractOpen || state == ContractAccepted)
+
+		if invoice.IsPending() != pending {
+			t.Fatalf("expected pending: %v, got: %v, invoice: %v",
+				pending, invoice.IsPending(), invoice)
+		}
+	}
+}
+
 func TestInvoiceWorkflow(t *testing.T) {
 	t.Parallel()
 
@@ -436,7 +459,7 @@ func TestFetchAllInvoicesWithPaymentHash(t *testing.T) {
 		invoice.State = states[i%len(states)]
 		paymentHash := invoice.Terms.PaymentPreimage.Hash()
 
-		if invoice.State != ContractSettled && invoice.State != ContractCanceled {
+		if invoice.IsPending() {
 			testPendingInvoices[paymentHash] = invoice
 		}
 
@@ -576,6 +599,69 @@ func TestDuplicateSettleInvoice(t *testing.T) {
 	if !reflect.DeepEqual(dbInvoice, invoice) {
 		t.Fatalf("wrong invoice after second settle, expected %v got %v",
 			spew.Sdump(invoice), spew.Sdump(dbInvoice))
+	}
+}
+
+// TestFetchAllInvoices tests that FetchAllInvoices works as expected.
+func TestFetchAllInvoices(t *testing.T) {
+	t.Parallel()
+
+	db, cleanUp, err := makeTestDB()
+	defer cleanUp()
+	if err != nil {
+		t.Fatalf("unable to make test db: %v", err)
+	}
+
+	contractStates := []ContractState{
+		ContractOpen, ContractSettled, ContractCanceled, ContractAccepted,
+	}
+
+	numInvoices := len(contractStates) * 2
+
+	var expectedPendingInvoices []Invoice
+	var expectedAllInvoices []Invoice
+
+	for i := 1; i <= numInvoices; i++ {
+		invoice, err := randInvoice(lnwire.MilliSatoshi(i))
+
+		if err != nil {
+			t.Fatalf("unable to create invoice: %v", err)
+		}
+
+		invoice.AddIndex = uint64(i)
+		// Set the contract state of the next invoice such that there's an equal
+		// number for all possbile states.
+		invoice.State = contractStates[i%len(contractStates)]
+
+		paymentHash := invoice.Terms.PaymentPreimage.Hash()
+		if invoice.IsPending() {
+			expectedPendingInvoices = append(expectedPendingInvoices, *invoice)
+		}
+		expectedAllInvoices = append(expectedAllInvoices, *invoice)
+
+		if _, err := db.AddInvoice(invoice, paymentHash); err != nil {
+			t.Fatalf("unable to add invoice: %v", err)
+		}
+	}
+
+	pendingInvoices, err := db.FetchAllInvoices(true)
+	if err != nil {
+		t.Fatalf("unable to fetch all pending invoices: %v", err)
+	}
+
+	allInvoices, err := db.FetchAllInvoices(false)
+	if err != nil {
+		t.Fatalf("unable to fetch all non pending invoices: %v", err)
+	}
+
+	if !reflect.DeepEqual(pendingInvoices, expectedPendingInvoices) {
+		t.Fatalf("pending invoices: %v\n != \n expected einvoices: %v",
+			spew.Sdump(pendingInvoices), spew.Sdump(expectedPendingInvoices))
+	}
+
+	if !reflect.DeepEqual(allInvoices, expectedAllInvoices) {
+		t.Fatalf("pending + non pending: %v\n != \n expected: %v",
+			spew.Sdump(allInvoices), spew.Sdump(expectedAllInvoices))
 	}
 }
 
--- a/channeldb/invoice_test.go
+++ b/channeldb/invoice_test.go
@@ -2,7 +2,6 @@ package channeldb
 
 import (
 	"crypto/rand"
-	mrand "math/rand"
 	"reflect"
 	"testing"
 	"time"
@@ -416,23 +415,25 @@ func TestFetchAllInvoicesWithPaymentHash(t *testing.T) {
 		t.Fatalf("expected empty list as a result, got: %v", empty)
 	}
 
-	// Now populate the DB and check if we can get all invoices with their
-	// payment hashes as expected.
-	const numInvoices = 20
-	testPendingInvoices := make(map[lntypes.Hash]*Invoice)
-	testAllInvoices := make(map[lntypes.Hash]*Invoice)
-
 	states := []ContractState{
 		ContractOpen, ContractSettled, ContractCanceled, ContractAccepted,
 	}
 
-	for i := lnwire.MilliSatoshi(1); i <= numInvoices; i++ {
-		invoice, err := randInvoice(i)
+	numInvoices := len(states) * 2
+	testPendingInvoices := make(map[lntypes.Hash]*Invoice)
+	testAllInvoices := make(map[lntypes.Hash]*Invoice)
+
+	// Now populate the DB and check if we can get all invoices with their
+	// payment hashes as expected.
+	for i := 1; i <= numInvoices; i++ {
+		invoice, err := randInvoice(lnwire.MilliSatoshi(i))
 		if err != nil {
 			t.Fatalf("unable to create invoice: %v", err)
 		}
 
-		invoice.State = states[mrand.Intn(len(states))]
+		// Set the contract state of the next invoice such that there's an equal
+		// number for all possbile states.
+		invoice.State = states[i%len(states)]
 		paymentHash := invoice.Terms.PaymentPreimage.Hash()
 
 		if invoice.State != ContractSettled && invoice.State != ContractCanceled {
--- a/channeldb/invoices.go
+++ b/channeldb/invoices.go
@@ -149,15 +149,13 @@ const (
 	// ContractOpen means the invoice has only been created.
 	ContractOpen ContractState = 0
 
-	// ContractSettled means the htlc is settled and the invoice has been
-	// paid.
+	// ContractSettled means the htlc is settled and the invoice has been paid.
 	ContractSettled ContractState = 1
 
 	// ContractCanceled means the invoice has been canceled.
 	ContractCanceled ContractState = 2
 
-	// ContractAccepted means the HTLC has been accepted but not settled
-	// yet.
+	// ContractAccepted means the HTLC has been accepted but not settled yet.
 	ContractAccepted ContractState = 3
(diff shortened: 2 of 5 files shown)


channeldb/invoice_test.go
```diff
@@ -578,6 +578,67 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+func TestFetchAllInvoices(t *testing.T) {
+	t.Parallel()
+
+	db, cleanUp, err := makeTestDB()
+	defer cleanUp()
+	if err != nil {
+		t.Fatalf("unable to make test db: %v", err)
+	}
+
+	const numInvoices = 100
```
joostjager: Why so many? One or two in every state seems enough to me

channeldb/invoice_test.go
```diff
@@ -578,6 +578,67 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+	for i := 1; i < numInvoices+1; i++ {
+		invoiceValue := lnwire.MilliSatoshi(i)
+		invoice, err := randInvoice(invoiceValue)
+		invoice.AddIndex = uint64(i)
+
+		if err != nil {
+			t.Fatalf("unable to create invoice: %v", err)
+		}
+
+		invoice.State = contractStates[mrand.Intn(len(contractStates))]
```
joostjager: Random in unit tests is risky. For a time it is unlikely to impact the execution flow, but for this the non-deterministically-ness gets more serious.

bhandras:
> Straight-forward change. But it is a breaking rpc change. How to deal with that?

To be fair #3694 already did this but this PR clears completes it.

channeldb/invoice_test.go
bhandras: Removed the random in favor of mod.

channeldb/invoice_test.go
bhandras: Yeah, with 4 states and mod, it's plenty more than needed. Changed.

joostjager:
>To be fair #3694 already did this but this PR completes it.

How did it do that? It auto-canceled invoices, but it didn't change which invoices where returned when listing?

channeldb/invoice_test.go
```diff
@@ -578,6 +577,67 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+
+	var expectedPendingInvoices []Invoice
+	var expectedAllInvoices []Invoice
+
+	for i := 1; i < numInvoices+1; i++ {
+		invoiceValue := lnwire.MilliSatoshi(i)
+		invoice, err := randInvoice(invoiceValue)
+		invoice.AddIndex = uint64(i)
+
+		if err != nil {
```
joostjager: nit: Check error right after `randInvoice` call

channeldb/invoice_test.go
```diff
@@ -578,6 +577,67 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+			t.Fatalf("unable to create invoice: %v", err)
+		}
+
+		invoice.State = contractStates[i%len(contractStates)]
```
joostjager: A line of comment explaining what is happening here wouldn't hurt

channeldb/invoice_test.go
```diff
@@ -578,6 +577,67 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+
+	const numInvoices = 12
+	contractStates := []ContractState{
+		ContractOpen, ContractSettled, ContractCanceled, ContractAccepted,
+	}
+
+	var expectedPendingInvoices []Invoice
+	var expectedAllInvoices []Invoice
+
+	for i := 1; i < numInvoices+1; i++ {
```
joostjager: `for i := 0; i < numInvoices; i++` ?

channeldb/invoice_test.go
```diff
@@ -578,6 +577,67 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+		invoice.AddIndex = uint64(i)
+
+		if err != nil {
+			t.Fatalf("unable to create invoice: %v", err)
+		}
+
+		invoice.State = contractStates[i%len(contractStates)]
+
+		paymentHash := invoice.Terms.PaymentPreimage.Hash()
+		if invoice.IsPending() {
```
joostjager: This test wouldn't catch a change to `IsPending`. Maybe you want to explicitly check the state value.

joostjager:
>We can keep compatibility by adding a new flag, but imo it's more confusing.

That is what we generally do. And mark the old flag as deprecated in at least one major release.

channeldb/invoice_test.go
```diff
@@ -53,6 +52,27 @@ func randInvoice(value lnwire.MilliSatoshi) (*Invoice, error) {
 	return i, nil
 }
 
+func TestInvoiceIsPending(t *testing.T) {
```
joostjager: Function comment missing.

channeldb/invoice_test.go
```diff
@@ -432,7 +452,7 @@ func TestFetchAllInvoicesWithPaymentHash(t *testing.T) {
 			t.Fatalf("unable to create invoice: %v", err)
 		}
 
-		invoice.State = states[mrand.Intn(len(states))]
+		invoice.State = states[int(i)%len(states)]
```
joostjager: This could have been a separate commit just to get rid of the random.

channeldb/invoice_test.go
```diff
@@ -578,6 +598,69 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+			t.Fatalf("unable to add invoice: %v", err)
+		}
+	}
+
+	pendingInvoices, err := db.FetchAllInvoices(true)
+	if err != nil {
+		t.Fatalf("unable to fetch all pending invoices: %v", err)
+	}
+
+	allInvoices, err := db.FetchAllInvoices(false)
```
joostjager: Wasn't it the idea that `FetchAllInvoices` and its test would move to `migration_11`?

channeldb/invoice_test.go
```diff
@@ -578,6 +598,69 @@ func TestDuplicateSettleInvoice(t *testing.T) {
...
+	}
+
+	numInvoices := len(contractStates) * 2
+
+	var expectedPendingInvoices []Invoice
+	var expectedAllInvoices []Invoice
+
+	for i := 1; i <= numInvoices; i++ {
+		invoice, err := randInvoice(lnwire.MilliSatoshi(i))
+
```
halseth: nit: remove newline

channeldb/invoices.go
```diff
@@ -385,6 +383,11 @@ func validateInvoice(i *Invoice) error {
 	return nil
 }
 
+// IsPending returns ture if the invoice is in ContractOpen state.
```
halseth: typo: ture

halseth: also comment is not exactly accurate

channeldb/invoice_test.go
bhandras: done

channeldb/invoice_test.go
bhandras: done

channeldb/invoice_test.go
bhandras: Removed FetchAllInvoices (in a separate commit). PTAL.

channeldb/invoice_test.go
bhandras: This whole test is now removed as (after discussing this with @joostjager) I decided to remove FetchAllInvoices completely.

channeldb/invoices.go
bhandras: Thansk :smile:, missed this. Fixed.

bhandras:
> Proto file also needs to be updated for new meaning of this boolean.

Thanks for pointing this out. Updated.

channeldb/invoice_test.go
```diff
@@ -330,29 +293,24 @@ func TestInvoiceAddTimeSeries(t *testing.T) {
...
 		)
 		if err != nil {
 			t.Fatalf("unable to settle invoice: %v", err)
 		}
-	}
 
-	invoices, err = db.FetchAllInvoices(false)
-	if err != nil {
-		t.Fatalf("unable to fetch invoices: %v", err)
+		settledInvoices = append(settledInvoices, *updatedInvoice)
```
joostjager: Doesn't this reduce the effect of this test, because we rely on the code under test to provides us with the expectations?

lnrpc/rpc.proto
```diff
@@ -2450,7 +2450,10 @@ message PaymentHash {
...
 
 message ListInvoiceRequest {
-    /// If set, only unsettled invoices will be returned in the response.
+    /**
+    If set, only invoices that are not settled and not canceled will be returned
+    in the response.
+    */
     bool pending_only = 1 [json_name = "pending_only"];
 
     /**
```
joostjager: I would squash this commit with the first one (optional)

channeldb/invoice_test.go
bhandras: I'd argue that it actually makes the test more strict as before we relied on `FetchAllInvoices` too. Ideally expected results should be constant and possbily included in the test itself rather then generated/selected etc.

channeldb/invoice_test.go
```diff
@@ -147,43 +147,6 @@ func TestInvoiceWorkflow(t *testing.T) {
...
-	// using big endian, the invoices should be retrieved in ascending
-	// order (and the primary key should be incremented with each
-	// insertion).
-	for i := 0; i < len(invoices)-1; i++ {
-		if !reflect.DeepEqual(*invoices[i], dbInvoices[i]) {
-			t.Fatalf("retrieved invoices don't match %v vs %v",
-				spew.Sdump(invoices[i]),
-				spew.Sdump(dbInvoices[i]))
-		}
-	}
```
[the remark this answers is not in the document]
bhandras: No because it tested that the `FetchAllInvoices` retrieves invoices added with `AddInvoice`. `FetchAllInvoices` is dead code and removed, and `AddInvoices` has coverage in other tests.

lnrpc/rpc.proto
bhandras: done

channeldb/invoices.go
```diff
@@ -1381,7 +1381,7 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
 		return &invoice, nil
 	}
 
-	now := d.Now()
+	now := d.Now().Truncate(0)
```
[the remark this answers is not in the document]
bhandras: Added a comment for this. Also fixed it as the canonical way of stripping it is not Truncate but Round (as described here: https://golang.org/pkg/time/)

channeldb/invoice_test.go
```diff
@@ -52,6 +52,29 @@ func randInvoice(value lnwire.MilliSatoshi) (*Invoice, error) {
 	return i, nil
 }
 
+// Tests that pendind state is correctly returned when calling the
```
joostjager: typo 'pendind'

channeldb/invoices.go
```diff
@@ -1381,7 +1381,10 @@ func (d *DB) updateInvoice(hash lntypes.Hash, invoices, settleIndex *bbolt.Bucke
 		return &invoice, nil
 	}
 
-	now := d.Now()
+	// Strip the monotonic clock reading. This is helpful when
+	// deep comparing time points in tests where updateInvoice
+	// has updated timestamps and expectations are pre generated.
+	now := d.Now().Round(0)
```
joostjager: Looking at this again, my opinion is that we shouldn't need this. For testing, we should to inject a mocked `Now()` to make the test deterministic and also not have the problem with the monotonic clock.

channeldb/invoice_test.go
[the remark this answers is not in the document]
joostjager: Good point

channeldb/invoice_test.go
bhandras: done

channeldb/invoices.go
bhandras: After some offline conversation, decided to use fixed Now() in the tests instead.

channeldb/invoices.go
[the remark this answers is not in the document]
bhandras: This is now removed.

channeldb/invoice_test.go
joostjager: The removal of `FetchAllInvoices` is good, but it is now replaced by something that uses the code under test to generate the expectations. This test is testing whether the filtering is correct. I'd say we don't need to do the deep equal on the exact invoice contents for that. It is tested elsewhere already. For me, building a list of payment hashes here and asserting on it in the various filter cases is enough.

bhandras: I'm not convinced that this is not a test relaxation (compared to what it was before). Maybe it's more correct this way though. PTAL.

joostjager: Isn't the `updatedInvoice` still used as an expectation now?

joostjager: Oh, it is changed in a later commit. Generally we don't want to touch the same code in one pr twice. Switching them around as you had before lowers the summed delta of all commits.

bhandras:
> Commit order seems to be not optimal in this revision?

Yeah, forgot to tamper with dates when rebasing interactively. It was in good order just GitHub orders by date. Fixed now.

channeldb/invoice_test.go
bhandras: Sorry, that was a rebase mistake on my end. Corrected.

channeldb/invoice_test.go
```diff
@@ -607,25 +579,23 @@ func TestQueryInvoices(t *testing.T) {
...
-		t.Fatalf("unable to retrieve pending invoices: %v", err)
+			// Save invoice to use it for expectation checks.
+			invoices = append(invoices, *updatedInvoice)
+		} else {
+			// Set AddIndex so that we can use DeepEqual later.
+			invoice.AddIndex = uint64(i)
+
+			// Save invoice to use it for expectation checks.
+			invoices = append(invoices, *invoice)
+			pendingInvoices = append(pendingInvoices, *invoice)
```
joostjager: Still using `updatedInvoice` for expectation?

bhandras:
After much discussion about what to include in the PR and how to change tests @joostjager and I decided to go back in time somewhat and just revert to the version where we keep `FetchAllInvoices` intact as fixing the tests introduces a lot of unrelated changes and should be done in subsequent cleanup PRs.

joostjager:
@bhandras it is ok to merge without a fresh approval from @halseth, because in the end the pr was reverted to the revision that he originally acked.

PR #3841, opened 2019-12-16
routing: check max routing info size

joostjager:
This PR modifies pathfinding to check the accumulated total payload size during search. This ensures that we will never find a route that exceeds the limit of 1300 bytes.

In order to execute this check, it is necessary to take into account whether nodes along the route support tlv (this influences the payload size). The size of custom records that are possibly present is added as well.


routing/pathfind.go
```diff
@@ -840,17 +889,8 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	// Overwrite final hop features.
 	pathEdges[len(pathEdges)-1].Node.Features = features
 
-	// The route is invalid if it spans more than 20 hops. The current
-	// Sphinx (onion routing) implementation can only encode up to 20 hops
-	// as the entire packet is fixed size. If this route is more than 20
-	// hops, then it's invalid.
-	numEdges := len(pathEdges)
-	if numEdges > HopLimit {
-		return nil, errMaxHopsExceeded
```
cfromknecht: remove this err?

channeldb/graph.go
```diff
@@ -2180,9 +2179,10 @@ func (l *LightningNode) isPublic(tx *bbolt.Tx, sourcePubKey []byte) (bool, error
 // FetchLightningNode attempts to look up a target node by its identity public
 // key. If the node isn't found in the database, then ErrGraphNodeNotFound is
 // returned.
-func (c *ChannelGraph) FetchLightningNode(pub *btcec.PublicKey) (*LightningNode, error) {
+func (c *ChannelGraph) FetchLightningNode(nodePub route.Vertex) (*LightningNode,
```
cfromknecht: 👍

routing/testdata/excessive_hops.json
```diff
@@ -1,410 +0,0 @@
-{
```
cfromknecht: 👏

routing/pathfind.go
```diff
@@ -677,6 +697,34 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+		// included. If we are coming from the source hop, the payload
+		// size is zero, because the original htlc isn't in the onion
+		// blob.
+		var payloadSize uint64
+		if fromVertex != source {
+			supportsTlv := fromFeatures.HasFeature(
+				lnwire.TLVOnionPayloadOptional,
+			)
+
+			hop := &route.Hop{
```
cfromknecht: remove `&`, here and above?

routing/pathfind.go
```diff
@@ -536,6 +537,23 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
 	}
 
+	// Build a preliminary destination hop structure to obtain the payload
+	// size.
+	var mpp *record.MPP
+	if r.PaymentAddr != nil {
+		mpp = record.NewMPP(amt, *r.PaymentAddr)
+	}
+
+	finalHop := &route.Hop{
```
cfromknecht: i see what you mean now about using the hops directly, very cool!

joostjager: Yes, so initially i planned to store these in the map. Maybe we'll do it in the future.

routing/pathfind.go
joostjager: done

routing/pathfind.go
```diff
@@ -792,7 +841,7 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 			//
 			// TODO(joostjager): Pass in fromFeatures to processEdge
 			// for payload size calculation.
-			processEdge(fromNode, policy, partialPath)
+			processEdge(fromNode, *fromFeatures, policy, partialPath)
```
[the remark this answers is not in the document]
joostjager: done

routing/pathfind.go
joostjager: removed

routing/pathfind.go
```diff
@@ -813,6 +826,19 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		}
 	}
 
+	// For the final hop, we'll set the node features to those determined
+	// above. These are either taken from the destination features, e.g.
+	// virual or invoice features, or loaded as a fallback from the graph.
```
bhandras: s/virual/virtual

routing/pathfind.go
```diff
@@ -766,6 +741,44 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+				// vector.
+				default:
+					fromFeatures = lnwire.EmptyFeatureVector()
+				}
+
+				// Update cache.
+				featureCache[fromNode] = fromFeatures
+			}
+			if fromFeatures == nil {
+				continue
```
bhandras: If this happens, should we set an empty feature vector?

routing/pathfind.go
```diff
@@ -587,11 +588,10 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 			timeLockDelta = edge.TimeLockDelta
 		}
 
-		incomingCltv := toNodeDist.incomingCltv +
-			uint32(timeLockDelta)
+		incomingCltv := toNodeDist.incomingCltv + int32(timeLockDelta)
 
 		// Check that we are within our CLTV limit.
-		if incomingCltv > r.CltvLimit {
+		if uint32(incomingCltv-finalHtlcExpiry) > r.CltvLimit {
```
bhandras: Wouldn't this be a bit cleaner in the form of `incomingCltv > finalHtlcExpiry + r.CltvLimit`?

routing/pathfind.go
joostjager: This happens if the node features in the graph are invalid. In that case we skip the node completely. If they want to route, they need to advertize a valid feature set.

Extracted this closure a little bit, too much nesting that makes it harder to understand.

routing/pathfind.go

routing/pathfind.go
joostjager: maybe safer indeed. but we can't cast `r.CltvLimit` because sometimes maxuint32 is used as a value.

joostjager: Ok, also not ok, because with the addition we wrap around. Left as it was, but added comment about safety of this comparison.

joostjager:
Itest failing, need to investigate

joostjager:
Itest fixed, db transaction problem for which I added a commit

routing/pathfind.go
```diff
@@ -824,17 +877,21 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
+	// NOTE: This may overwrite features loaded from the graph if
+	// destination features were provided. This is fine though, since our
+	// route construction does not care where the features are actually
+	// taken from. In the future we may wish to do route construction within
+	// findPath, and avoid using ChannelEdgePolicy altogether.
+	pathEdges[len(pathEdges)-1].Node.Features = features
 
 	log.Debugf("Found route: probability=%v, hops=%v, fee=%v\n",
-		distance[source].probability, numEdges,
+		distance[source].probability, len(pathEdges),
```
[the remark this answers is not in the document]
joostjager: removed

routing/pathfind.go
joostjager: is it needed? i think nothing can go wrong as it is?

joostjager: changed just in case

routing/pathfind.go
```diff
@@ -590,11 +591,13 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
 		}
 
-		incomingCltv := toNodeDist.incomingCltv +
-			uint32(timeLockDelta)
+		incomingCltv := toNodeDist.incomingCltv + int32(timeLockDelta)
+
+		// Check that we are within our CLTV limit. Take care with this
+		// comparison, because the cltv limit may be MaxUint32.
+		if uint64(incomingCltv) >
+			uint64(r.CltvLimit)+uint64(finalHtlcExpiry) {
```
cfromknecht: can this line be extracted sooner, and calculated once?

joostjager: extracted

PR #3843, opened 2019-12-17
[htlcnotifier 1/4]: Introduce Internal Errors

carlaKC:
This PR is change 1 of 4 proposed to add a htlc event notifier to lnd.
1: This PR
2: #3844
3: #3781
4: #3848 

--------------

This PR changes the structure of the errors that we use within htlcswitch. This change is necessary for introducing metadata (which should not be exposed on the wire) to  failures that occur within the switch.
- The `ClearTextError` interface is introduced to represent all of the non-opaque errors that the switch handles. 
- `ForwardingError` is updated to implement this interface, and covers the case where receive an error from a peer along the route of one of our own payments
- A `LinkError` implementation of the interface is added to cover the case where we fail a send/receive/forward on our incoming/outgoing link
- A `FailureDetail` enum that adds metadata to `LinkError` is added to enrich locally produced errors (to be used in follow up PRs); if we find that we need additional data, this can be replaced with an interface, or we can create a new implementation of `ClearTextError`

The added overhead of this change is that wire messages which do not need to be enriched with additional details need to be wrapped in this internal type. This is the cost of creating stricter barriers between internal errors and wire errors.

### DIFF
--- a/htlcswitch/failure.go
+++ b/htlcswitch/failure.go
@@ -25,6 +25,40 @@ type ClearTextError interface {
 	WireMessage() lnwire.FailureMessage
 }
 
+// LinkError is an implementation of the ClearTextError interface which
+// represents failures that occur on our incoming or outgoing link.
+type LinkError struct {
+	// msg returns the wire failure associated with the error.
+	// This value should *not* be nil, because we should always
+	// know the failure type for failures which occur at our own
+	// node.
+	msg lnwire.FailureMessage
+}
+
+// NewLinkError returns a LinkError with the failure message provided.
+// The failure message provided should *not* be nil, because we should
+// always know the failure type for failures which occur at our own node.
+func NewLinkError(msg lnwire.FailureMessage) *LinkError {
+	return &LinkError{msg: msg}
+}
+
+// WireMessage extracts a valid wire failure message from an internal
+// error which may contain additional metadata (which should not be
+// exposed to the network). This value should never be nil for LinkErrors,
+// because we are the ones failing the htlc.
+//
+// Note this is part of the ClearTextError interface.
+func (l *LinkError) WireMessage() lnwire.FailureMessage {
+	return l.msg
+}
+
+// Error returns the string representation of a link error.
+//
+// Note this is part of the ClearTextError interface.
+func (l *LinkError) Error() string {
+	return l.msg.Error()
+}
+
 // ForwardingError wraps an lnwire.FailureMessage in a struct that also
 // includes the source of the error.
 type ForwardingError struct {
--- a/htlcswitch/failure.go
+++ b/htlcswitch/failure.go
@@ -40,6 +40,27 @@ func (f *ForwardingError) Error() string {
 	)
 }
 
+// NewForwardingError creates a new payment error which wraps a wire error
+// with additional metadata.
+func NewForwardingError(failure lnwire.FailureMessage, index int,
+	extraMsg string) *ForwardingError {
+
+	return &ForwardingError{
+		FailureSourceIdx: index,
+		FailureMessage:   failure,
+		ExtraMsg:         extraMsg,
+	}
+}
+
+// NewUnknownForwardingError returns a forwarding error which has a nil failure
+// message. This constructor should only be used in the case where we cannot
+// decode the failure we have received from a peer.
+func NewUnknownForwardingError(index int) *ForwardingError {
+	return &ForwardingError{
+		FailureSourceIdx: index,
+	}
+}
+
 // ErrorDecrypter is an interface that is used to decrypt the onion encrypted
 // failure reason an extra out a well formed error.
 type ErrorDecrypter interface {
@@ -94,15 +115,10 @@ func (s *SphinxErrorDecrypter) DecryptError(reason lnwire.OpaqueReason) (
 	r := bytes.NewReader(failure.Message)
 	failureMsg, err := lnwire.DecodeFailure(r, 0)
 	if err != nil {
-		return &ForwardingError{
-			FailureSourceIdx: failure.SenderIdx,
-		}, nil
+		return NewUnknownForwardingError(failure.SenderIdx), nil
 	}
 
-	return &ForwardingError{
-		FailureSourceIdx: failure.SenderIdx,
-		FailureMessage:   failureMsg,
-	}, nil
+	return NewForwardingError(failureMsg, failure.SenderIdx, ""), nil
 }
 
 // A compile time check to ensure ErrorDecrypter implements the Deobfuscator
--- a/htlcswitch/failure.go
+++ b/htlcswitch/failure.go
@@ -33,6 +33,9 @@ type LinkError struct {
 	// know the failure type for failures which occur at our own
 	// node.
 	msg lnwire.FailureMessage
+
+	// FailureDetail enriches the wire error with additional information.
+	FailureDetail
 }
 
 // NewLinkError returns a LinkError with the failure message provided.
@@ -40,6 +43,17 @@ type LinkError struct {
 // always know the failure type for failures which occur at our own node.
 func NewLinkError(msg lnwire.FailureMessage) *LinkError {
 	return &LinkError{msg: msg}
+}
+
+// NewDetailedLinkError returns a link error that enriches a wire message with
+// a failure detail.
+func NewDetailedLinkError(msg lnwire.FailureMessage,
+	detail FailureDetail) *LinkError {
+
+	return &LinkError{
+		msg:           msg,
+		FailureDetail: detail,
+	}
 }
 
 // WireMessage extracts a valid wire failure message from an internal
@@ -56,7 +70,13 @@ func (l *LinkError) WireMessage() lnwire.FailureMessage {
 //
 // Note this is part of the ClearTextError interface.
 func (l *LinkError) Error() string {
-	return l.msg.Error()
+	// If the link error has no failure detail, return the wire message's
+	// error.
+	if l.FailureDetail == FailureDetailNone {
+		return l.msg.Error()
+	}
+
+	return fmt.Sprintf("%v: %v", l.msg.Error(), l.FailureDetail)
 }
 
 // ForwardingError wraps an lnwire.FailureMessage in a struct that also
--- a/htlcswitch/failure.go
+++ b/htlcswitch/failure.go
@@ -88,10 +88,6 @@ type ForwardingError struct {
 	// zero is the self node.
 	FailureSourceIdx int
 
-	// ExtraMsg is an additional error message that callers can provide in
-	// order to provide context specific error details.
-	ExtraMsg string
-
 	// msg is the wire message associated with the error. This value may
 	// be nil in the case where we fail to decode failure message sent by
 	// a peer.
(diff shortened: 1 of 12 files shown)


htlcswitch/failure.go
```diff
@@ -9,45 +9,121 @@ import (
...
 )
 
-// ForwardingError wraps an lnwire.FailureMessage in a struct that also
-// includes the source of the error.
-type ForwardingError struct {
+// LinkError is an interface which wraps lnwire failure messages for
+// internal use within the switch. This interface is intended to create a
+// barrier between the switch and wire.
+type LinkError interface {
+	GetWireMessage() lnwire.FailureMessage
```
halseth: needs godoc

htlcswitch/failure.go
```diff
@@ -36,8 +36,17 @@ func (s *WireError) Error() string {
 	return s.Msg.Error()
 }
 
-// PaymentError wraps an lnwire.FailureMessage in a struct that also
-// includes the source of the error.
+// NewWireError returns a switch error which wraps the wire message provided.
+func NewWireError(msg lnwire.FailureMessage) LinkError {
```
cfromknecht: return the concrete type `*WireError`?

htlcswitch/failure_detail.go
```diff
@@ -0,0 +1,60 @@
...
+// additional information.
+type FailureDetail interface {
+	// FailureString returns a string representation of a failure detail.
+	FailureString() string
+}
+
+// FailureDetailOnionDecode indicates that we could not decode an onion
+// error.
+type FailureDetailOnionDecode struct {
+	PaymentHash lntypes.Hash
```
cfromknecht: exported fields in this file need godocs

htlcswitch/failure_detail.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/lightningnetwork/lnd/lnwire"
+
+	"github.com/lightningnetwork/lnd/lntypes"
+)
+
+// FailureDetail is an interface which is used to enrich failures with
+// additional information.
+type FailureDetail interface {
+	// FailureString returns a string representation of a failure detail.
+	FailureString() string
```
cfromknecht: is this called anywhere in this pr?

a more concise name might be `Failure()` or `Detail()`, but use whatever you prefer :)

htlcswitch/failure.go
```diff
@@ -43,6 +43,39 @@ func NewWireError(msg lnwire.FailureMessage) LinkError {
...
+// Error returns an error string for a link failure.
+//
+// Note this is part of the LinkError interface.
+func (s *DetailError) Error() string {
+	return fmt.Sprintf("%v: %v", s.Msg.Error(), s.FailureDetail)
+}
+
+// newDetailError returns a link error which wraps the wire message provided
+// with an additional failure detail.
+func newDetailError(msg lnwire.FailureMessage, detail FailureDetail) LinkError {
```
cfromknecht: should we export this so that it can be unit tested like the others?

htlcswitch/failure_detail.go
carlaKC: Didn't want to have an empty interface so just picked something that seemed unique :)

This isn't used (could have been used in the `Error` funcs) but since we don't really need these strings and the metadata isn't used anywhere I'm switching these over to an enum. for now. If we end up needing metadata, it'll be easy to switch back.

carlaKC:
> Perhaps something like having `LocalPaymentError` and `ForwardingPaymentError` both implementing a common interface. 

Gone ahead with this approach, I think it cleans things up nicely :)

The switch had three kinds of errors it can experience when routing a htlc:
1. Link errors: failed on our incoming or outgoing link
2. Forwarding errors: fails down the line (and is our send)
3. Opaque errors: fails down the line (but we did not initiate the send)

We need a way to identify local errors, so I've added a `RouteHTLCError` interface which is implemented by link and forwarding errors. The router can identify routing related errors by checking for this interface. Decided on an interface because it lets us add different information to different errors - forwarding errors only have a wire message but also have a node index, link errors might have additional failure detail but don't have a source index (since we are the source).
`RouteHTLCError` is a dreadful name, alternate name suggestions are most welcome!

The switch doesn't really care about opaque errors because they are just passed downstream.
Follow up PRs will update the incoming link code to use `LinkError` as well, so that htlcnotifier can have additional info.

htlcswitch/failure.go
```diff
@@ -9,6 +9,16 @@ import (
...
 )
 
+// RouteHtlcError is an interface which is implemented by errors obtained when
+// trying to route a HTLC. A routing error can occur at our node, when we fail
+// a HTLC on the incoming/outgoing link or further down the line.
+type RouteHtlcError interface {
+	error
+
+	// GetWireMessage returns the wire failure associated with the failure.
+	GetWireMessage() lnwire.FailureMessage
```
joostjager: I think the go convention is to leave off `Get`

carlaKC:
> I have one question though: what is the future of `ExtraMsg`, the ugly string based message? This PR introduces `FailureDetail`. Can that be used instead of `ExtraMsg`?

There are two options we can go with here:
1. Failure detail enum (current impl)
[pro] brief declaration of errors, provides set strings which cover almost all `ExtraMsg` uses
[con] Cannot enrich these strings with error-specific metadata. 

2. Failure detail interface (explored in previous version of PR)
[pro] error specific metadata, potentially nice for unit testing
[con] has to be implemented for 20+ failure detail types, error declaration a lot more verbose

I've gone with the enum for now, because we only use `ExtraMsg` with metadata in 3 places and (link ineligible, onion decode failure and payment resolved on chain before first hop) and I think that logging lnd side is sufficient (users that have the knowledge to investigate prob have the knowledge to check logs), but I have the code for both so happy to go either way really.

joostjager:
I also think that extending the enum with the cases that we use `extramsg` for atm is the best option.

htlcswitch/failure.go
```diff
@@ -9,6 +9,16 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
+// RouteHtlcError is an interface which is implemented by errors obtained when
+// trying to route a HTLC. A routing error can occur at our node, when we fail
+// a HTLC on the incoming/outgoing link or further down the line.
+type RouteHtlcError interface {
```
carlaKC: Could also be called `KnownError` or `ClearTextError` since this covers all cases where we have a non-opaque error?

htlcswitch/switch.go
```diff
@@ -1163,12 +1169,14 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 // The ciphertext will be derived from the failure message proivded by context.
 // This method returns the failErr if all other steps complete successfully.
 func (s *Switch) failAddPacket(packet *htlcPacket,
-	failure lnwire.FailureMessage, failErr error) error {
+	failure *LinkError, failErr error) error {
```
halseth: can just take the `FailureMessage` as before?

htlcswitch/failure_detail.go
```diff
@@ -0,0 +1,65 @@
...
+		return "htlc exceeds maximum policy amount"
+
+	case FailureDetailInsufficientBalance:
+		return "insufficient bandwidth to route htlc"
+
+	case FailureDetailIncompleteForward:
+		return "incomplete forward"
+
+	default:
+		return ""
```
halseth: return "unknown failure detail"?

htlcswitch/failure.go
```diff
@@ -74,10 +74,6 @@ type ForwardingError struct {
 	// zero is the self node.
 	FailureSourceIdx int
 
-	// ExtraMsg is an additional error message that callers can provide in
-	// order to provide context specific error details.
-	ExtraMsg string
```
halseth: was this even used anywhere?

htlcswitch/failure.go
```diff
@@ -9,6 +9,62 @@ import (
...
+// represents failures that occur on our incoming or outgoing link.
+type LinkError struct {
+	// msg returns the wire failure associated with the error.
+	msg lnwire.FailureMessage
+
+	// FailureDetail enriches the wire error with additional information.
+	FailureDetail
+}
+
+// WireMessage unwraps a LinkError to obtain a wire message.
```
halseth: what is this wire message for? Maybe add an explanation to the godoc

htlcswitch/failure.go
```diff
@@ -9,6 +9,62 @@ import (
...
+
+	// WireMessage returns the wire failure associated with the failure.
+	WireMessage() lnwire.FailureMessage
+}
+
+// LinkError is an implementation of the RouteHtlcError interface which
+// represents failures that occur on our incoming or outgoing link.
+type LinkError struct {
+	// msg returns the wire failure associated with the error.
+	msg lnwire.FailureMessage
```
halseth: can be nil?

htlcswitch/failure.go
```diff
@@ -9,6 +9,62 @@ import (
...
+}
+
+// NewLinkError returns a LinkError with the failure message provided.
+func NewLinkError(msg lnwire.FailureMessage) *LinkError {
+	return &LinkError{msg: msg}
+}
+
+// NewDetailedError returns a link error that enriches a wire message with
+// a failure detail.
+func NewDetailedError(msg lnwire.FailureMessage,
```
halseth: `NewDetailedLinkError`

htlcswitch/switch.go
```diff
@@ -1016,18 +1014,20 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
 		var destination ChannelLink
 		for _, link := range interfaceLinks {
-			var failure lnwire.FailureMessage
+			var failure *LinkError
 
 			// We'll skip any links that aren't yet eligible for
 			// forwarding.
 			if !link.EligibleToForward() {
-				failure = &lnwire.FailUnknownNextPeer{}
+				failure = NewLinkError(
```
halseth: should have a detail?

htlcswitch/failure.go
carlaKC: It was used in [handleLocalDispatch](https://github.com/lightningnetwork/lnd/blob/3c6be62b18b9ea84369ac92d48ebcd2325807d3c/htlcswitch/switch.go#L775) to set a link not available error and in [parseFailedPayment](https://github.com/lightningnetwork/lnd/blob/3c6be62b18b9ea84369ac92d48ebcd2325807d3c/htlcswitch/switch.go#L935) if we could not decode the onion, or the payment went on chain before the first hop.

These are replaced with `FailureDetail`, so the error presented to the user no longer has the exact link that failed/ error that occurred, but the full original error is logged and I doubt anybody has been parsing onion decode payment errors for details.

htlcswitch/failure.go
carlaKC: Shouldn't be nil in this case because we're the source of the error. Thought about adding a panic in the constructor, but that felt a bit extreme. I think the comment here is enough?

htlcswitch/switch.go
carlaKC: Ah yes, getting ahead of myself for htlcNotifier things :D

routing/router.go
```diff
@@ -1910,15 +1910,28 @@ func (r *ChannelRouter) processSendError(paymentID uint64, rt *route.Route,
...
 
-	failureMessage := fErr.WireMessage()
-	failureSourceIdx := fErr.FailureSourceIdx
+	var (
+		failureMessage = rtErr.WireMessage()
+
+		// failureSourceIdx is the index of the node which the failure
+		// occurred at. A zero value indicates that the failure
+		// occurred locally at our node.
+		failureSourceIdx int
```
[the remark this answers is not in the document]
carlaKC: Nice, makes it much clearer :)

htlcswitch/failure.go
halseth: 👍 

Could be removed in the same commit that replaces it with `FailureDetail`, but that is no blocker :)

lnrpc/routerrpc/router_server.go
```diff
@@ -425,7 +425,16 @@ func marshallError(sendError error) (*Failure, error) {
...
 	}
 
-	response.FailureSourceIndex = uint32(fErr.FailureSourceIdx)
+	// If the ClearTextError received is a ForwardingError, the error
+	// originated from a node along the route, not locally on our outgoing
+	// link. We set failureSourceIdx to the index of the node where the
+	// failure occurred. If the error is not a ForwardingError, the failure
+	// occurred at our node, so we leave the index as 0 to indicate that
+	// we failed locally.
+	fErr, ok := rtErr.(*htlcswitch.ForwardingError)
```
cfromknecht: Do we always assume 0 for the other error types? If so the index could be part of the ClearTextError iface to avoid the type conversions. I see though we only do so in two places, so may not be worth the effort now

carlaKC: Yes, only other type is a `LinkError` which implies a 0 index. Going to leave as is for now, but will keep it in mind!

PR #3845, opened 2019-12-17
[trivial] lncfg+lncli: replace WithDialer->WitchContextDialer

halseth:
WithDialer is deprecated, use WitchContextDialer instead.


alexbosworth:
🧙‍♀️

PR #3846, opened 2019-12-17
itest: make harnesses usable in external projects

guggero:
To use the node and network harnesses from the lntest package in other projects, it is necessary to fix the macaroon path to have all macaroons in the same directory (without this fix, the admin, readonly and invoice macaroon are in the `data/` directory of lnd while all the sub-server macaroons are in the `data/chain/bitcoin/regtest/` folder).
Also it is useful if we can specify the location of the `lnd-itest` binary so we don't have to use the same folder structure/depth in other projects.
Finally, to get access to a node's data directory, we need to export the node's config struct.

### DIFF
--- a/lntest/node.go
+++ b/lntest/node.go
@@ -130,7 +130,7 @@ type BackendConfig interface {
 	Name() string
 }
 
-type nodeConfig struct {
+type NodeConfig struct {
 	Name       string
 	BackendCfg BackendConfig
 	NetParams  *chaincfg.Params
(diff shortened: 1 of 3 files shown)


lntest/node.go
```diff
@@ -232,7 +232,7 @@ func (cfg nodeConfig) genArgs() []string {
 // harness. Each HarnessNode instance also fully embeds an RPC client in
 // order to pragmatically drive the node.
 type HarnessNode struct {
-	cfg *nodeConfig
+	Cfg *nodeConfig
```
wpaulino: Should also export `nodeConfig`.

lntest/node.go
```diff
@@ -298,9 +298,13 @@ func newNode(cfg nodeConfig) (*HarnessNode, error) {
...
 	cfg.TLSCertPath = filepath.Join(cfg.DataDir, "tls.cert")
 	cfg.TLSKeyPath = filepath.Join(cfg.DataDir, "tls.key")
-	cfg.AdminMacPath = filepath.Join(cfg.DataDir, "admin.macaroon")
-	cfg.ReadMacPath = filepath.Join(cfg.DataDir, "readonly.macaroon")
-	cfg.InvoiceMacPath = filepath.Join(cfg.DataDir, "invoice.macaroon")
+
+	networkDir := filepath.Join(
+		cfg.DataDir, "chain", "bitcoin", cfg.NetParams.Name,
+	)
+	cfg.AdminMacPath = filepath.Join(networkDir, "admin.macaroon")
```
wpaulino: I'd imagine these don't need to be specified since we'd want them in their default path?

lntest/node.go
guggero: Done.

lntest/node.go
guggero: They are used in different places within the harness and need to be set.

lntest/node.go
```diff
@@ -298,9 +298,13 @@ func newNode(cfg NodeConfig) (*HarnessNode, error) {
 	cfg.LogDir = filepath.Join(cfg.BaseDir, "log")
 	cfg.TLSCertPath = filepath.Join(cfg.DataDir, "tls.cert")
 	cfg.TLSKeyPath = filepath.Join(cfg.DataDir, "tls.key")
-	cfg.AdminMacPath = filepath.Join(cfg.DataDir, "admin.macaroon")
-	cfg.ReadMacPath = filepath.Join(cfg.DataDir, "readonly.macaroon")
-	cfg.InvoiceMacPath = filepath.Join(cfg.DataDir, "invoice.macaroon")
+
+	networkDir := filepath.Join(
```
Roasbeef: The commit message should have some rationale w.r.t why this change was required.

guggero: Yes, I agree. I added some context to the commit message.

PR #3847, opened 2019-12-17
lnd test: fix csv mempool sweep time

halseth:
Since CSV locked outputs specifies the first block where they are
allowed to be included, they can actually be added one block earlier
into the mempool.

This led to a flake, where the sweep tx was already in the mempool at
the time we mined the last block, causing the next mempool check to
fail.

Fixes 
```
    --- FAIL: TestLightningNetworkDaemon/channel_force_closure (49.83s)
        lnd_test.go:100: Failed: (channel force closure): exited with error: 
            *errors.errorString failed to get sweep tx from mempool: wanted 1, found 0 txs in mempool: []
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:3266 (0xd993f9)
            	testChannelForceClosure: t.Fatalf("failed to get sweep tx from mempool: %v", err)
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:123 (0xd884da)
            	(*harnessTest).RunTestCase: testCase.test(h.lndHarness, h)
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:15298 (0xdf6d2c)
            	TestLightningNetworkDaemon.func4: ht.RunTestCase(testCase)
            /home/travis/.gimme/versions/go1.13.5.linux.amd64/src/testing/testing.go:909 (0x50af09)
            	tRunner: fn(t)
            /home/travis/.gimme/versions/go1.13.5.linux.amd64/src/runtime/asm_amd64.s:1357 (0x4610e1)
            	goexit: BYTE	$0x90	// NOP
FAIL
FAIL	github.com/lightningnetwork/lnd/lntest/itest	253.229s
FAIL
```

### DIFF
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -3186,10 +3186,10 @@ func testChannelForceClosure(net *lntest.NetworkHarness, t *harnessTest) {
 	}
 
 	// Currently within the codebase, the default CSV is 4 relative blocks.
-	// For the persistence test, we generate three blocks, then trigger
+	// For the persistence test, we generate two blocks, then trigger
 	// a restart and then generate the final block that should trigger
 	// the creation of the sweep transaction.
-	if _, err := net.Miner.Node.Generate(defaultCSV - 1); err != nil {
+	if _, err := net.Miner.Node.Generate(defaultCSV - 2); err != nil {
 		t.Fatalf("unable to mine blocks: %v", err)
 	}
 
@@ -3225,12 +3225,12 @@ func testChannelForceClosure(net *lntest.NetworkHarness, t *harnessTest) {
 		}
 
 		// At this point, the nursery should show that the commitment
-		// output has 1 block left before its CSV delay expires. In
+		// output has 2 block left before its CSV delay expires. In
 		// total, we have mined exactly defaultCSV blocks, so the htlc
 		// outputs should also reflect that this many blocks have
 		// passed.
 		err = checkCommitmentMaturity(
-			forceClose, commCsvMaturityHeight, 1,
+			forceClose, commCsvMaturityHeight, 2,
 		)
 		if err != nil {
 			return err
@@ -3258,9 +3258,9 @@ func testChannelForceClosure(net *lntest.NetworkHarness, t *harnessTest) {
 		t.Fatalf("unable to mine blocks: %v", err)
 	}
 
-	// At this point, the sweeping transaction should now be broadcast. So
-	// we fetch the node's mempool to ensure it has been properly
-	// broadcast.
+	// At this point, the CSV will expire in the next block, meaning that
+	// the sweeping transaction should now be broadcast. So we fetch the
+	// node's mempool to ensure it has been properly broadcast.
 	sweepingTXID, err := waitForTxInMempool(net.Miner.Node, minerMempoolTimeout)
 	if err != nil {
 		t.Fatalf("failed to get sweep tx from mempool: %v", err)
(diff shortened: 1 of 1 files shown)


joostjager:
I see in the commit sweep resolver:
`err := c.waitForHeight(unlockHeight - 1)`

Maybe we should pad with one more block to allow for propagation?

halseth:
Alternative waiting one extra block: https://github.com/lightningnetwork/lnd/pull/3852

cfromknecht:
fwiw I think #3852 was the correct fix. the padding was originally added to be more tolerant of block propagation for neutrino peers, since we don’t have a reliable connection to a single backend like we would with bitcoind or btcd. 

neutrino may hear about a new block, but also may lose that peer before publishing. if your other peers haven’t seen the last block they will fail to accept the transaction because it will appear one block too early.

In either of these cases, the only other possible spend is a revocation. If that hasn’t happened by this point then waiting one block has no real impact on our ability to play the timeout.

halseth:
> fwiw I think #3852 was the correct fix. the padding was originally added to be more tolerant of block propagation for neutrino peers, since we don’t have a reliable connection to a single backend like we would with bitcoind or btcd.
> 
> neutrino may hear about a new block, but also may lose that peer before publishing. if your other peers haven’t seen the last block they will fail to accept the transaction because it will appear one block too early. if that fails the txn won’t be rebroadcasted until the node restarts.
> 
> In either of these cases, the only other possible spend is a revocation. If that hasn’t happened by this point then waiting one block has no real impact on our ability to play the timeout.

That's an interesting point. However when connecting to a peer that is one block ahead of you, won't you sync up before requesting the transaction? Similarly if you are already connected, shouldn't the node request the block before any new transactions?

I also think the neutrino republisher will make sure we rebroadcast every block, so that might help as well.

cfromknecht:
> However when connecting to a peer that is one block ahead of you, won't you sync up before requesting the transaction? Similarly if you are already connected, shouldn't the node request the block before any new transactions?

In the case that you’re already connected, idt there’d be any syncing?

> I also think the neutrino republisher will make sure we rebroadcast every block, so that might help as well.

That’s a good point. I recall being able to reproduce this reliably in testing, and we ultimately decided the padding would fix this for the vast majority of cases. 

However, the rebroadcaster didn’t exist back when the padding was added, so it’s possible this alleviates the issue. Would probably need to do some testing to confirm.

joostjager:
I am wondering if this flake has the same origin:

```
--- FAIL: TestLightningNetworkDaemon/test_multi-hop_htlc_local_chain_claim (11.90s)
        lnd_test.go:100: Failed: (test multi-hop htlc local chain claim): exited with error: 
            *errors.errorString tx did not spend from commitment tx
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_multi-hop_htlc_local_chain_claim_test.go:194 (0xd8c632)
            	testMultiHopHtlcLocalChainClaim: t.Fatalf("tx did not spend from commitment tx")
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:123 (0xd92a5a)
            	(*harnessTest).RunTestCase: testCase.test(h.lndHarness, h)
            /home/travis/gopath/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:15177 (0xe0018c)
            	TestLightningNetworkDaemon.func4: ht.RunTestCase(testCase)
            /home/travis/.gimme/versions/go1.13.5.linux.amd64/src/testing/testing.go:909 (0x50ae99)
            	tRunner: fn(t)
            /home/travis/.gimme/versions/go1.13.5.linux.amd64/src/runtime/asm_amd64.s:1357 (0x4610e1)
            	goexit: BYTE	$0x90	// NOP
```

PR #3848, opened 2019-12-17
[htlcnotifier 4/4]: lnrpc: Add HTLC Event Subscription

### LINKED ISSUE #3420: htlcswitch+rpcserver: add new streaming ForwardingEvent API

carlaKC:
This PR is change 4 of 4 proposed to add a htlc event notifier to lnd.
1: ~#3843~
2: ~#3844~
3: ~#3781~
4: This PR

This PR adds a `htlcnotifier` subscription to `routerrpc`:
- We can either present wire message + detail separately (current approach), or combine them into one error code (I think this isn't great for adding more details, permutations go up). 

Fixes #3420


lnrpc/routerrpc/router.proto
```diff
@@ -490,6 +490,136 @@ message BuildRouteResponse {
...
+    uint32 incoming_timelock = 1 [json_name="incoming_timelock"];
+
+    // The timelock on the outgoing htlc.
+    uint32 outgoing_timelock = 2 [json_name="outgoing_timelock"];
+
+    // The amount of the incoming htlc.
+    uint64 incoming_amount_msat = 3 [json_name="incoming_amount_msat"];
+
+    // The amount of the outgoing htlc.
+    uint64 outgoing_amt_msat = 4 [json_name ="outgoing_amount_msat"];
```
joostjager: inconsistency: amt vs amount

lnrpc/routerrpc/router.proto
```diff
@@ -490,6 +490,136 @@ message BuildRouteResponse {
...
+    uint64 incoming_htlc_id = 3 [json_name="incoming_htlc_id"];
+
+    /**
+    Outgoing id is the index of the outgoing htlc in the outgoing channel.
+    This value is zero for receives.
+    */
+    uint64 outgoing_htlc_id = 4 [json_name="outgoing_htlc_id"];
+}
+
+message HtlcInfo{
```
joostjager: I know we have the split between `HtlcInfo` and `HtlcKey` because some available isn't accessible when the event is generated. I think it is ok internally, but do we also want to do that on the external rpc interface? It would promote something that really is 'not yet implemented' to a factor that shapes the response structure in a way that isn't easily undone.

lnrpc/routerrpc/router_server.go
```diff
@@ -80,6 +81,10 @@ var (
 			Entity: "offchain",
 			Action: "read",
 		}},
+		"/routerrpc.Router/SubscribeHtlcEvents": {{
+			Entity: "offchain",
```
joostjager: It isn't (or won't be) just offchain?

lnrpc/routerrpc/router_server.go
```diff
@@ -747,3 +752,244 @@ func (s *Server) BuildRoute(ctx context.Context,
...
+				return fmt.Errorf("unexpected event type: %T",
+					event)
+			}
+
+			rpcEvent, err := rpcHtlcEvent(htlcEvent)
+			if err != nil {
+				return err
+			}
+
+			switch e := event.(type) {
```
joostjager: There are now two type switches. Consider making a single marshall function that marshalls all of `htlcEvent`

lnrpc/routerrpc/router_server.go
```diff
@@ -747,3 +752,244 @@ func (s *Server) BuildRoute(ctx context.Context,
...
+
+		case <-s.quit:
+			return errors.New("server shutting down")
+		}
+	}
+}
+
+// rpcHtlcEvent returns a rpc htlc event. Note that the event field is not set
+// by this function, only the fields that are common to all events are set.
+func rpcHtlcEvent(htlcEvent htlcswitch.HtlcEvent) (*HtlcEvent, error) {
```
joostjager: Move some of this code to a new file, to not go down the way that `rpcserver.go` went? (very long file)

lnrpc/routerrpc/router_server.go
carlaKC: We can add another entity when on chain is added.

lnrpc/routerrpc/router.proto
```diff
@@ -490,6 +490,120 @@ message BuildRouteResponse {
...
 
+
+message SubscribeHtlcEventsRequest{}
+
+message HtlcEvent{
+    /**
+    IncomingChannel is the channel that the incoming htlc arrived at our node
+    on. This value is zero for sends.
+    */
+    uint64 incoming_channel = 1 [json_name="incoming_channel"];
```
cfromknecht: i might suggest making this `incoming_channel_id` or `incoming_chan_id` since we typically have the `_id` prefix elsewhere in the rpc interface, similar comment for `outgoing_channel`

lnrpc/routerrpc/router.proto
```diff
@@ -490,6 +490,120 @@ message BuildRouteResponse {
...
+    and may be 'no detail' if the wire message requires no additional metadata.
+    */
+    FailureDetail failure_detail = 2 [json_name = "failure_detail"];
+
+    // A string representation of the link failure.
+    string failure_string = 3 [json_name = "failure_string"];
+}
+
+enum FailureDetail {
+    UNKNOWN = 0;
```
cfromknecht: 👍

lnrpc/routerrpc/router.proto
```diff
@@ -490,6 +490,120 @@ message BuildRouteResponse {
...
+    IncomingChannel is the channel that the incoming htlc arrived at our node
+    on. This value is zero for sends.
+    */
+    uint64 incoming_channel = 1 [json_name="incoming_channel"];
+
+    /**
+    Outgoing channel is the channel that the outgoing htlc left our node on.
+    This value is zero for receives.
+    */
+    uint64 outgoing_channel = 2 [json_name="outgoing_channel"];
```
cfromknecht: @guggero has a PR open to remove all use of `json_name`, i think we can leave them out. iirc it doesn't do anything if the names match the field name but i could be mistaken

carlaKC: sgtm, will remove

lnrpc/routerrpc/subscribe_events.go
```diff
@@ -0,0 +1,214 @@
...
+
+	rpcEvent := &HtlcEvent{
+		IncomingChannelId: key.IncomingCircuit.ChanID.ToUint64(),
+		OutgoingChannelId: key.OutgoingCircuit.ChanID.ToUint64(),
+		IncomingHtlcId:    key.IncomingCircuit.HtlcID,
+		OutgoingHtlcId:    key.OutgoingCircuit.HtlcID,
+		Timestamp:         uint64(htlcEvent.Timestamp().Unix()),
+	}
+
+	switch e := htlcEvent.(type) {
```
joostjager: If you have this switch here anyway, do you need the `HtlcEvent` interface?

carlaKC: Just a lot more lines of duplicate code to have to extract the htlc key and timestamp for each event type when it can be done all in one.

I think the interface is also generally nice to have as a prescription for what a htlc event should have; eg if something like a htlc interceptor/forwarding predicate goes in, we may want to add another event, which would need to fit this interface.

joostjager: I don't think it is more duplicated code, the extraction just happens in the rpc server instead of in htlcnotifier.

My general thought is that if we need to flatten events just for the rpc (at the moment), it should happen as close to the rpc code as possible.

I quickly tried it out: https://github.com/lightningnetwork/lnd/commit/f7ff5eb0f238a764b5efcb37eb191e642c953209

It is not a blocker if you like the interface more.

carlaKC: Fair enough, actually less code relative to adding the interface. Fewer changes always better :) 

Cut the interface and flattened in rpc.

lnrpc/routerrpc/router.proto
carlaKC: Decided to keep this separate from for `LinkFail`/`Forwarding` events, because I'm not sure it's worth a change to the switch just to add these values for rpc purposes. If we decide to add them, htlc info fields can be deprecated and moved into the main `HtlcEvent` struct.

carlaKC:
Rebased to address some merge conflicts, ready for a look @cfromknecht @joostjager :)

lnrpc/routerrpc/router_server.go
```diff
@@ -167,6 +170,7 @@ func (s *Server) Start() error {
 //
 // NOTE: This is part of the lnrpc.SubServer interface.
 func (s *Server) Stop() error {
+	close(s.quit)
```
joostjager: Does this need a waitgroup too, to wait for completion?

lnrpc/routerrpc/router_server.go
```diff
@@ -596,3 +600,32 @@ func (s *Server) BuildRoute(ctx context.Context,
...
+			rpcEvent, err := rpcHtlcEvent(event)
+			if err != nil {
+				return err
+			}
+
+			if err := stream.Send(rpcEvent); err != nil {
+				return err
+			}
+
+		case <-s.quit:
```
joostjager: I see we in other places also have `case <-stream.Context().Done():`, in case the client disconnects

joostjager: Also noticed that `trackPayment` doesn't listen to a quit signal, maybe worth adding that one-liner.

lnrpc/routerrpc/router.proto
```diff
@@ -335,6 +335,136 @@ message BuildRouteResponse {
...
+    */
+    EventType event_type = 6;
+
+    /*
+    Event contains the htlc event that was processed. These are served on a
+    best-effort basis; events are not persisted, delivery is not guaranteed
+    (in the event of a crash in the switch, forward events may be lost) and
+    some events may be replayed upon restart. Events consumed from this package
+    should be de-duplicated by the htlc's unique combination of incoming and
+    outgoing channel id and htlc id. [EXPERIMENTAL]
```
joostjager: Maybe this could be moved one level up. It applies to all of the `HtlcEvent` proto message?

lnrpc/routerrpc/router.proto
joostjager: Ok. The message is marked experimental, so we have the opportunity to change it later.

lnrpc/routerrpc/router_server.go
```diff
@@ -596,3 +600,32 @@ func (s *Server) BuildRoute(ctx context.Context,
...
+
+	htlcClient, err := s.cfg.RouterBackend.SubscribeHtlcEvents()
+	if err != nil {
+		return err
+	}
+	defer htlcClient.Cancel()
+
+	for {
+		select {
+		case event := <-htlcClient.Updates():
```
joostjager: Is it possible that this channel closes?

carlaKC: This channel is not closed, but the subscription does also have a quit channel which will be closed in the case of an unexpected server shutdown which I can add 👍

lnrpc/routerrpc/router_server.go
carlaKC: Offline discussion about this, no reason for subscriptions not to be wait grouped (I thought there could be some grpc magic in the background). Added wait group for routerrpc subserver + `trackPayment`.

lnrpc/routerrpc/router_server.go
```diff
@@ -449,6 +454,9 @@ func (s *Server) trackPayment(paymentHash lntypes.Hash,
 		}
 	}
 
+	s.wg.Add(1)
```
joostjager: Still this feels a bit strange. Normally we add to the waitgroup right before we start the goroutine. In this case, grpc started the goroutine already, so are we too late here? I don't know the answer really, but maybe it doesn't make sense after all to do this?

carlaKC: On the router server level, adding this wait group means that we don't consider the server to be stopped before all the subscriptions have received the quit signal. So I think that adds something, although all the other subservers are shutting down fine without it, so also unsure.

lnrpc/routerrpc/router_server.go
```diff
@@ -609,3 +613,45 @@ func (s *Server) BuildRoute(ctx context.Context,
...
+	stream Router_SubscribeHtlcEventsServer) error {
+
+	htlcClient, err := s.cfg.RouterBackend.SubscribeHtlcEvents()
+	if err != nil {
+		return err
+	}
+	defer htlcClient.Cancel()
+
+	s.wg.Add(1)
+	defer s.wg.Done()
```
cfromknecht: hmm on second thought i suspect this is unnecessary since this doesn't spawn any _new_ goroutines. [`GracefulStop` will wait until all pending rpc calls are processed](https://github.com/grpc/grpc-go/blob/master/server.go#L1555)

lnrpc/routerrpc/router_server.go
cfromknecht: do we want to add `sync.Once` here? o/w calling `Stop()` multiple times will panic

lnrpc/routerrpc/router.proto
```diff
@@ -335,6 +335,136 @@ message BuildRouteResponse {
...
+
+    // The timelock on the outgoing htlc.
+    uint32 outgoing_timelock = 2;
+
+    // The amount of the incoming htlc.
+    uint64 incoming_amt_msat = 3;
+
+    // The amount of the outgoing htlc.
+    uint64 outgoing_amt_msat = 4;
+}
```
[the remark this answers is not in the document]
carlaKC: Yeah we discussed it in the previous notifier PRs, wanted to force people towards identifying htlcs by incoming/outgoing circuit keys rather than payment hash, since it's not a unique identifier. If it's there, chances are people with use it as a uuid. It's specifically a concern here since forwards need to be matched with settles/fails, and doing so with payment hash could go very wrong.

lnrpc/routerrpc/subscribe_events.go
```diff
@@ -0,0 +1,241 @@
...
+	default:
+		return nil, fmt.Errorf("unknown event type: %T", e)
+	}
+
+	rpcEvent := &HtlcEvent{
+		IncomingChannelId: key.IncomingCircuit.ChanID.ToUint64(),
+		OutgoingChannelId: key.OutgoingCircuit.ChanID.ToUint64(),
+		IncomingHtlcId:    key.IncomingCircuit.HtlcID,
+		OutgoingHtlcId:    key.OutgoingCircuit.HtlcID,
+		Timestamp:         uint64(timestamp.Unix()),
```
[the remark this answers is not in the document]
carlaKC: yeah sgtm!

carlaKC:
@joostjager re-requesting to make sure approval still stands. Removed wait group (see [comment](https://github.com/lightningnetwork/lnd/pull/3848#discussion_r391949728)) and converted timestamps to ns.

lnrpc/routerrpc/router_server.go
carlaKC: Yeah, felt weird to have it. So there are different `Stop` levels here? Stopping the internals of the subserver (achieved with quit) and stopping of the rpc server which happen separately.

joostjager:
Yes approval still stands.

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
cfromknecht: makes sense, i think that's a reasonable assumption 👍

carlaKC:
> it seems like this file is missing `make rpc-format`...

Last push build, so must have missed the travis build check merge in some werid order. Fixed in #4105.

PR #3849, opened 2019-12-18
zpay32: update spec test vectors, write features last

cfromknecht:
This commit updates the BOLT11 test vectors to use the updated versions
from the spec (with TLV bit set). This also pointed out that the
ordering was inconsistent with the spec, in that the payment secret
should be serialized before the feature vector.

This uses the updated test vectors stemming from [this comment](https://github.com/lightningnetwork/lightning-rfc/pull/719#discussion_r358910138)

### DIFF
--- a/zpay32/invoice.go
+++ b/zpay32/invoice.go
@@ -1094,6 +1094,14 @@ func writeTaggedFields(bufferBase32 *bytes.Buffer, invoice *Invoice) error {
 			return err
 		}
 	}
+	if invoice.PaymentAddr != nil {
+		err := writeBytes32(
+			bufferBase32, fieldTypeS, *invoice.PaymentAddr,
+		)
+		if err != nil {
+			return err
+		}
+	}
 	if invoice.Features.SerializeSize32() > 0 {
 		var b bytes.Buffer
 		err := invoice.Features.RawFeatureVector.EncodeBase32(&b)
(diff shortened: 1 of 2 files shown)


zpay32/invoice_test.go
```diff
@@ -499,19 +506,20 @@ func TestDecodeEncode(t *testing.T) {
 		},
 		{
 			// On mainnet, please send $30 coffee beans supporting
-			// features 1 and 9.
-			encodedInvoice: "lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9qzsze992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq73t7cl",
+			// features 15 and 99, using secret 0x11...
```
joostjager: Also include 9 in comment here and below?

zpay32/invoice.go
```diff
@@ -1094,6 +1094,14 @@ func writeTaggedFields(bufferBase32 *bytes.Buffer, invoice *Invoice) error {
...
 		}
 	}
+	if invoice.PaymentAddr != nil {
+		err := writeBytes32(
+			bufferBase32, fieldTypeS, *invoice.PaymentAddr,
+		)
+		if err != nil {
+			return err
+		}
+	}
```
joostjager: Where exactly in the spec is this ordering requirement? I only see this:

>MUST specify the most-preferred field first, followed by less-preferred fields, in order.

But what is preferred?

Maybe add a comment that the fields needs to be order.

joostjager: Ok, misread that.

PR #3852, opened 2019-12-18
contractcourt: wait one extra block before sweeping commit

halseth:
The integration tests are written with the assumption, since this was
the old bahavior before removal of the utxo nursery.

Alternative to #3847

### DIFF
--- a/contractcourt/commit_sweep_resolver.go
+++ b/contractcourt/commit_sweep_resolver.go
@@ -164,9 +164,11 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 		c.log.Debugf("waiting for csv lock to expire at height %v",
 			unlockHeight)
 
-		// We only need to wait for the block before the block that
-		// unlocks the spend path.
-		err := c.waitForHeight(unlockHeight - 1)
+		// We really only need to wait for the block before the block
+		// that unlocks the spend path, but we wait one extra block to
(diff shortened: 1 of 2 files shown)


contractcourt/commit_sweep_resolver.go
```diff
@@ -164,9 +164,11 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 		c.log.Debugf("waiting for csv lock to expire at height %v",
 			unlockHeight)
 
-		// We only need to wait for the block before the block that
-		// unlocks the spend path.
-		err := c.waitForHeight(unlockHeight - 1)
+		// We really only need to wait for the block before the block
+		// that unlocks the spend path, but we wait one extra block to
+		// ensure propagation and to keep the old behavior from when
```
joostjager: Returning your own question: is it needed to ensure propagation or do we just want to keep the old behavior?

Roasbeef:
Why should we wait an additional block? We want to being sweeping as soon as possible, if the integration tests have some unnecessary delay assumption, that's where we should point our attention instead.

halseth:
Fixed by https://github.com/lightningnetwork/lnd/pull/3847 instead

cfromknecht:
linking for original rationale behind the delay https://github.com/lightningnetwork/lnd/pull/3847#issuecomment-567810324

PR #3853, opened 2019-12-18
build: Prepend newline to first byte slice logged using LogWriter Write method

### LINKED ISSUE #3801: Cosmetics: Missing newline in "Shutdown complete" log message
After shutdown, lnd logs "Shutdown complete" without a trailing newline. Upon restart, the line continues:

`2019-12-05 22:39:18.593 [INF] LTND: Shutdown complete2019-12-05 22:39:39.821 [INF] LTND: Version: 0.8.2-beta commit=v0.8.0-beta-386-gf3398c0c0ecc469c31e1f7e18e0fa2d63b5729ed, build=production, logging=default`

ellemouton:
This PR fixes #3801. It results in a newline being written to the log for the first byte slice written.

The problem reported was that lnd logged the "Shutdown complete" message without a trailing newline. 
However, after further investigation it was found that the `err := logWriter.Close()` line in the 
lnd.go Main function was to blame for this as it trims any new lines in the log file.

The fix was to change the LogWriter Write function so that it adds a newline to the first byte
slice to be written to the log

### DIFF
--- a/build/log.go
+++ b/build/log.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
 	"github.com/btcsuite/btclog"
 )
@@ -45,6 +46,12 @@ type LogWriter struct {
 	// is written to by the Write method of the LogWriter type. This only
 	// needs to be set if neither the stdlog or nolog builds are set.
 	RotatorPipe *io.PipeWriter
+
+	// startNewline is used in the LogWriter’s Write function to prepend a
+	// newline to the first byte slice that is logged. This is required to start
+	// logging on a new line on restart, rather than appending to the last log
+	// written
+	startNewLine sync.Once
 }
 
 // NewSubLogger constructs a new subsystem log from the current LogWriter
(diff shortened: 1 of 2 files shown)


ellemouton:
Thanks for the feedback @carlaKC! I rebased and squashed the commits 👍 

Yeah I know the solution seems a bit hacky and that it may introduce problems in the future for anything using the LogWriter Write function. So then maybe best left as is?

carlaKC:
Another approach would be to make a change in `lnd/build/logrotater.go` to `(r *RotatingLogWriter) InitLogRotator` to make it check whether there are any existing lines in the file and append a newline somewhere. Also a bit of a hack, but at least it cuts out all weird use cases.

@guggero any thoughts on this?

guggero:
Thanks for the fix @ellemouton!
It does work and I think the idea is pretty clever. But to be honest, it feels a bit too hacky.

Couldn't we achieve the same effect if we just print a newline directly after creating `logWriter` in `NewRotatingLogWriter`? We know that we always use the `RotatingLogWriter` if a log file is involved. And because it is initialized statically, we can also be sure that it's only done once.

PR #3858, opened 2019-12-19
autopilot: fix memChannelGraph channel edge addition

bhandras:
This PR fixes an issue that happens when adding a new channel edge
between two nodes in a memChannelGraph. Originally a channel edge held a
node value which made the graph different when iterating from the two
endpoints of an edge. This is simply fixed by holding pointers instead.


bhandras:
This is just a minifix that came up during betweenness centrality testing that made results flaky and depend on order of iteration.

PR #3859, opened 2019-12-19
lnrpc: allow not specifying features

joostjager:
It isn't possible to distinguish between an empty list and no list in protobuf. But for dest_features, we need a way to signal that the graph features should be used. Leaving the list empty is not the solution,
because this will be interpreted as a destination node that supports nothing.

Making an exception for the empty list and interpreting it as no list rules out the possibility to specify a node that does not support any feature.


joostjager:
One thing that I noticed is an error message that the user may not expect:

If you send custom records (for example key send) to a node that isn't in the graph and leave the destination features unspecified, it says "destination hop doesn't understand new TLV payloads". This is because we check the required features before pathfinding.

Not sure if it is worth fixing though.

joostjager:
After discussion offline, closed in favor of https://github.com/lightningnetwork/lnd/pull/3862

PR #3863, opened 2019-12-20
routing: fix gap in result interpretation

joostjager:
Fixes an off by one error that skipped assigning a success pair. Added a test that previously failed because of this bug.

### DIFF
--- a/routing/result_interpretation.go
+++ b/routing/result_interpretation.go
@@ -191,10 +191,8 @@ func (i *interpretedResult) processPaymentOutcomeFinal(
 		i.failPair(route, n-1)
 
 		// The other hops relayed corectly, so assign those pairs a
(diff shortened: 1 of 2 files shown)


routing/result_interpretation.go
```diff
@@ -192,7 +192,7 @@ func (i *interpretedResult) processPaymentOutcomeFinal(
 
 		// The other hops relayed corectly, so assign those pairs a
 		// success result.
-		if n > 2 {
+		if n >= 2 {
```
carlaKC: Do we need this check at all? We won't have n=0, and we return on n=1 above so at this stage n>=2 is implied?

joostjager: Fixed

PR #3865, opened 2019-12-20
calculate betweenness centrality of nodes

bhandras:
This PR adds NodeMetric as a common interface for arbitrary graph metrics used in composite autopilot metrics and betweenness centrality metric based on Brandes algorithm. The PR also integrates optional betweenness centrality calculation to the DescribeGraph RPC call.


autopilot/interface.go
```diff
@@ -147,6 +147,22 @@ type AttachmentHeuristic interface {
 		map[NodeID]*NodeScore, error)
 }
 
+// NodeMetric is a common interface for all graph metrics that are not
+// directly used as autopilot node scores but may be used in compositional
+// heuristics or statistical information exposed to users.
+type NodeMetric interface {
+
```
cfromknecht: nit: extra new line

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	graph := &simpleGraph{
+		Nodes: make([]NodeID, len(nodes)),
+		Adj:   make([][]int, len(nodes)),
+	}
+
+	for key, u := range nodes {
```
cfromknecht: is `u` here the `idx` from above?

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,208 @@
...
+		graph.Nodes[u] = key
+		graph.Adj[u] = adj[u]
+	}
+
+	return graph, nil
+}
```
cfromknecht: nit: imo the code above could be in it's own file, and leave `betweeness_centrality.go` just for the impl of the algo

autopilot/interface.go
```diff
@@ -182,6 +206,11 @@ func init() {
 	for _, h := range availableHeuristics {
 		AvailableHeuristics[h.Name()] = h
 	}
+
+	// Fill available node metrics.
+	for _, h := range availableNodeMetrics {
+		AvailableNodeMetrics[h.Name()] = h
```
cfromknecht: should we check for overwrites, here and above?

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Name returns the name of the metric.
+func (bc *BetweennessCentrality) Name() string {
+	return "betweeness_centrality"
+}
+
+// betweennessCentrality is the core of Brandes' algorithm. For detailed
+// explanation please read:
+// https://www.cl.cam.ac.uk/teaching/1617/MLRD/handbook/brandes.html
+func betweennessCentrality(g *simpleGraph, node int, centrality []float64) {
```
cfromknecht: some light comments in this method would help comprehension without needing to read the paper ;)

bhandras: You're totally right, sorry. Changed variable names to make it more readable and added a few comments to help understand what's happening. The final part where we use Brande's MAGIC equation to calculate delta is where it's still hard to understand imho and where reading the "paper" is more important.

autopilot/interface.go
bhandras: Dunno, I feel like this should be more of a static assert if we'd have something like that, otherwise maybe review time assert :)

autopilot/betweenness_centrality.go
cfromknecht: looks good, thanks!

cmd/lncli/commands.go
```diff
@@ -2953,6 +2953,13 @@ var describeGraphCommand = cli.Command{
 				"graph. Unannounced channels are both private channels, and " +
 				"public channels that are not yet announced to the network.",
 		},
+		cli.BoolFlag{
+			Name: "calculate_node_centrality",
```
Roasbeef: How about just `--centrality`? As is, it's a bit long for a command line param IMO.

rpcserver.go
```diff
@@ -4521,14 +4533,20 @@ func (r *rpcServer) DescribeGraph(ctx context.Context,
...
 			PubKey:     hex.EncodeToString(node.PubKeyBytes[:]),
 			Addresses:  nodeAddrs,
 			Alias:      node.Alias,
 			Color:      routing.EncodeHexColor(node.Color),
 			Features:   invoicesrpc.CreateRPCFeatures(node.Features),
-		})
+		}
+
+		if req.CalculateNodeCentrality {
+			lnNode.BetweennessCentrality = centrality[node.PubKeyBytes]
```
Roasbeef: When running this, I find that some nodes have a value of zero. Is it the case that these nodes are just unconnected, or that they don't have at least _two_ channels?

Roasbeef: Also I'm getting rather large values as well, like `2738.2895441051587` as an example. I was under the impression that centrality would output a value between 0 and 1, as it's itself a ratio of the total number of paths relative to the total number that pass through a target node.

Roasbeef: How about we normalize things so we have a value between 0 and 1 which'll make the value easier to interpret (most connected node has the highest relative value, my node has a smaller value, but not my orders of magnitude). The wiki page for betweenness centrality has a suggested normalization algorithm.

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,177 @@
...
+func (bc *BetweennessCentrality) Name() string {
+	return "betweeness_centrality"
+}
+
+// betweennessCentrality is the core of Brandes' algorithm.
+// We first calculate the shortest paths from the start node s to all other
+// nodes with BFS, then update the betweenness centrality values by using
+// Brandes' dependency trick.
+// For detailed explanation please read:
+// https://www.cl.cam.ac.uk/teaching/1617/MLRD/handbook/brandes.html
```
Roasbeef: Thinking about this a bit more, this algorithm is fine for the case where we want to calculate the value for _every_ node. However, for the autopilot case, we only care about _our_ centrality metric, so perhaps we'll need to adopt a slightly modified algorithm which can be more efficient when we only care about a single node.

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,177 @@
...
+func betweennessCentrality(g *SimpleGraph, s int, centrality []float64) {
+	// Pred[w] is the list of nodes that immediately precede w on a
+	// shortest path from s to t for each node t.
+	Pred := make([][]int, len(g.Nodes))
```
Roasbeef: Variable names should be all lower case, applies below as well.

Roasbeef:
Oh and just for reference, the total computation took about 20 seconds on a spare VPS of mine (mainnet graph).

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,177 @@
+package autopilot
+
+// stack is a simple int stack to help with readability of Brandes'
+// betweenness centrality implementation below.
+type stack struct {
+	stack []int
+}
+
+func (s *stack) Push(v int) {
```
halseth: exported functions should have godoc?

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Refresh recaculates and stores centrality values.
+func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
+	cache, err := NewSimpleGraph(graph)
+	if err != nil {
+		return err
+	}
+
+	// TODO: parallelize updates to centrality.
```
halseth: should be pretty straight forward to launch this in `N` go routines?

autopilot/simple_graph.go
```diff
@@ -0,0 +1,67 @@
+package autopilot
+
+// SimpleGraph stores a simplifed adj graph of a channel graph to speed
+// up graph processing by eliminating all unnecessary hasing and map access.
```
halseth: typo: hasing

rpcserver.go
```diff
@@ -4508,6 +4508,18 @@ func (r *rpcServer) DescribeGraph(ctx context.Context,
 	// transactional model.
 	graph := r.server.chanDB.ChannelGraph()
 
+	// Calculate betweenness centrality if requested. Note that depending on the
+	// graph size, this may take up to a few minutes.
+	var centrality map[autopilot.NodeID]float64
+	if req.CalculateNodeCentrality {
+		channelGraph := autopilot.ChannelGraphFromDatabase(graph)
+		centralityMetric := autopilot.NewBetweennessCentralityMetric()
+		if err := centralityMetric.Refresh(channelGraph); err != nil {
```
halseth: do we ever use the `refresh` functionality, or could we just pass the channel graph into the contructor above? (or `GetMetric`)

autopilot/betweenness_centrality.go
bhandras: There are two reasons for calculating the betweenness centrality for all nodes:

1) to get the centrality value for a single node, we have to calculate all shortest paths for every s, t pairs where s != t, this is required as the centrality of a single vertex is the sum(sigma(s,t,v) / sigma(s,t)) [for every s != t != v], where sigma(s,t) is the number of shortest paths between s,t and sigma(s,t,v) is the number of the shortest paths between s,t that pass trough v. As calculating all paths is the bulk of the computation, the rest of the algorithm is much less time consuming, and so we wouldn't spare much.

2) betweenness centrality for a single node doesn't really have much meaning as it depends on how connected the graph is not an absolute measure.

rpcserver.go
bhandras: re: what does a centrality value of zero mean?

This means that no shortest paths pass trough that node. Note that as per the definition, (for reference: sum(sigma(s,t) / sigma(s,t,v)) [s != t != v]) leaf nodes don't have shortest paths passing trough, therefore their centrality is always zero.

bhandras: re: why aren't values between 0 and 1?

The definition is a bit tricky, but if you take a closer look at it in reality betweenness centrality isn't a ratio, but a sum of ratios for all distinct pairs of vertices. We may normalize the centrality values, eg. by using the factor 2/(n*(n-1)), or with a better normalizing factor if needed but it may get harder to interpret for humans.

autopilot/betweenness_centrality.go
bhandras: Yeah, I kept variable names like this to make it easier to follow the linked paper. Can rename if needed though.

cmd/lncli/commands.go
bhandras: done

autopilot/betweenness_centrality.go
bhandras: we really don't need to export them so instead of adding the trivial godoc, made them local to this file instead.

autopilot/betweenness_centrality.go
bhandras: added a new commit to do this, ptal

autopilot/simple_graph.go
bhandras: thanks, done

rpcserver.go
bhandras: we don't use it for this metric as this algo requires full recalculation, but in general i'd say it's better to be able to refresh rather then calculate once upon construction as there may be partial result that is cheaper to update. wdyt?

bhandras:
> Oh and just for reference, the total computation took about 20 seconds on a spare VPS of mine (mainnet graph).

Should be faster now with parallelization.

rpcserver.go
Roasbeef: Gotcha, IMO normalization is desirable, as otherwise it's hard to compare two values between distinct nodes. As an example, one of my personal nodes has a value of 26k, but without scanning all the other values for a max, I have no idea where I stand compared to all the other nodes.

Roasbeef:
Latest version doesn't build.

Roasbeef:
With 4 workers, my very primitive benchmark now computes centrality in ~10 seconds.

autopilot/betweenness_centrality.go
```diff
@@ -150,10 +162,49 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
...
+	work := make(chan int)
+	partials := make(chan []float64, bc.workers)
+
+	// Each worker will compute a partial result.
+	worker := func() {
+		defer wg.Done()
+		partial := make([]float64, len(cache.Nodes))
+		for {
+			node, ok := <-work
+
```
Roasbeef: Style nit: extra new line.

autopilot/interface.go
```diff
@@ -187,7 +188,7 @@ var (
 
 	// AvailableNodeMetrics holds all available node metrics.
 	availableNodeMetrics = []NodeMetric{
-		NewBetweennessCentralityMetric(),
+		NewBetweennessCentralityMetric(runtime.NumCPU()),
```
Roasbeef: Missing this in `rpcsever.go`.

autopilot/betweenness_centrality.go
bhandras: removed

autopilot/interface.go
bhandras: sorry, missed this one

rpcserver.go
bhandras: I agree, added min/max normalization. To be fair though, one still has to scan other nodes to figure out ranking.

rpcserver.go
halseth: I think there's a lot to that argument, but maybe do that change where we have a metric that actually requires this. The interface can easily be changed later.

autopilot/betweenness_centrality.go
```diff
@@ -150,10 +162,48 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
...
+	// Each worker will compute a partial result.
+	worker := func() {
+		defer wg.Done()
+		partial := make([]float64, len(cache.Nodes))
+		for {
+			node, ok := <-work
+			if !ok {
+				break
+			}
+			betweennessCentrality(cache, node, partial)
```
halseth: IMO it is a bit confusing that the result of the calculation is added by reference to the partial slice, instead of returned. 

Now that we are collecting and summing the final results below anyway, what do you think about letting each worker return the result for each calculation immediately, then collect them all below? So it would be something like

```golang
work := make(chan int, len(cache.Nodes))
for node := range cahce.Nodes {
    work <- node
}
close(work)

worker := func() {
     defer wg.Done()
      
     for node := range work {
             partials <- betweennessCentrality(cache, node)
      }
}

wg.Add(bc.workers)
for i := range bc.workers {
    go worker()
}

wg.Wait()
...
```

Not sure if this would have a performance impact though, sending on the channe after every node calculation, but it should be unblocking

autopilot/betweenness_centrality_test.go
```diff
@@ -87,48 +85,52 @@ func TestBetweennessCentralityWithNonEmptyGraph(t *testing.T) {
 		},
 	}
 
+	workers := []int{-1, 1, 3, 9, 100}
```
halseth: Just return an error in case of non-valid num workers?

autopilot/betweenness_centrality.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// betweennessCentrality is the core of Brandes' algorithm.
+// We first calculate the shortest paths from the start node s to all other
+// nodes with BFS, then update the betweenness centrality values by using
+// Brandes' dependency trick.
+// For detailed explanation please read:
+// https://www.cl.cam.ac.uk/teaching/1617/MLRD/handbook/brandes.html
+func betweennessCentrality(g *SimpleGraph, s int, centrality []float64) {
+	// Pred[w] is the list of nodes that immediately precede w on a
```
cfromknecht:
```suggestion
	// pred[w] is the list of nodes that immediately precede w on a
```

autopilot/betweenness_centrality.go
```diff
@@ -50,12 +54,20 @@ func (q *queue) empty() bool {
...
 	centrality map[NodeID]float64
 }
 
 // NewBetweennessCentralityMetric creates a new BetweennessCentrality instance.
-func NewBetweennessCentralityMetric() *BetweennessCentrality {
-	return &BetweennessCentrality{}
+// Users can specify the number of workers to use for calculating centrality.
+func NewBetweennessCentralityMetric(workers int) *BetweennessCentrality {
+	// There should be at least one worker.
+	if workers < 1 {
```
cfromknecht: return error when invalid value is passed? for configuration imo it's better to fail than silently succeed

autopilot/betweenness_centrality.go
```diff
@@ -150,10 +162,48 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
...
+		defer wg.Done()
+		partial := make([]float64, len(cache.Nodes))
+		for {
+			node, ok := <-work
+			if !ok {
+				break
+			}
+			betweennessCentrality(cache, node, partial)
+		}
+		partials <- partial
```
cfromknecht: something to note for the future—it may be desirable to have the ability to cancel the calculation, either via `context.Context` or a `quit` channel to avoid blocking on shutdown for example

autopilot/betweenness_centrality.go
bhandras: With the current approach we allocate NumCPU() * N buffers and create partials without sending the result on a channel, resulting in less contention/blocking on channel communication. Also in the end we do much less operations with the current approach as the one above would require summing N * N numbers vs N * NumCPU().

autopilot/betweenness_centrality.go
```diff
@@ -206,10 +206,30 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
...
+	}
+
+	// Normalization factor.
+	z := max - min
+
+	// Normalize and create result map.
 	bc.centrality = make(map[NodeID]float64)
 	for u, value := range centrality {
-		bc.centrality[cache.Nodes[u]] = value / 2.0
+		bc.centrality[cache.Nodes[u]] = (value - min) / z
```
cfromknecht: is it possible to divide by zero?

autopilot/betweenness_centrality_test.go
```diff
@@ -88,7 +89,16 @@ func TestBetweennessCentralityWithNonEmptyGraph(t *testing.T) {
 	workers := []int{-1, 1, 3, 9, 100}
 
 	expectedCentrality := []float64{
-		3.0, 0.0, 3.0, 15.0, 6.0, 6.0, 7.0, 0.0, 0.0,
+		0.2, 0.0, 0.2, 1.0, 0.4, 0.4, 0.466, 0.0, 0.0,
+	}
+
+	// almostEq returns true of a and b are close enough.
+	almostEq := func(a, b float64) bool {
```
cfromknecht: is this just needed to handle FPA?

autopilot/betweenness_centrality.go
halseth: discussed offline. The current approach is better since it allocates less memory for the partial results.

autopilot/betweenness_centrality.go
```diff
@@ -206,10 +206,30 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
 		}
 	}
 
-	// Divide by two as the graph is undirected.
+	// Get min/max to be able to normalize
```
cfromknecht: should this be squashed into the commit where the metric is added?

autopilot/betweenness_centrality_test.go
bhandras: done

autopilot/betweenness_centrality.go
bhandras: left it separate to make it easier to review, now squashed.

rpcserver.go
bhandras: I'd like to keep it this way for now. As we don't really have information right now whether or not the graph is dirty it makes sense to separate calculation from result retrieval. Once this is wired into autopilot, recalculation may happen less frequently then retrieval based on user preference.

autopilot/betweenness_centrality_test.go
bhandras: Yeah, it's a bit impractical to implement so removed for now and used a division instead.

autopilot/betweenness_centrality.go
bhandras: done

autopilot/betweenness_centrality.go
bhandras: Initially I thought it's not possible, but then realized that it is if there are no edges just lone nodes. Corrected, ptal.

lnrpc/rpc.proto
```diff
@@ -2505,6 +2512,12 @@ message ChannelGraphRequest {
     channels, and public channels that are not yet announced to the network.
     */
     bool include_unannounced = 1;
+
+    /**
+    Whether to calculate node centrality values. Currently this will only
+    calculate node betweenness centrality.
+    */
+    bool calculate_node_centrality = 2;
```
halseth: name suggestion: just `node_centrality`?

bhandras:
After some offline discussion, we decided to move this to a new RPC call. This is now implemented in: https://github.com/lightningnetwork/lnd/pull/3865/commits/61fbda58a091fbc8570e53754ab289f148848649 PTAL

autopilot/betweenness_centrality.go
```diff
@@ -150,10 +163,46 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
...
+	work := make(chan int)
+	partials := make(chan []float64, bc.workers)
+
+	// Each worker will compute a partial result. This
+	// partial result is a sum of centrality updates on
+	// roughly N / workers nodes.
+	worker := func() {
+		defer wg.Done()
+		partial := make([]float64, len(cache.Nodes))
+		// Consume the next node, update centrality
```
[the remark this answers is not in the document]
bhandras: done

autopilot/simple_graph.go
```diff
@@ -0,0 +1,67 @@
...
+		u := getNodeIndex(node)
+
+		return node.ForEachChannel(func(edge ChannelEdge) error {
+			v := getNodeIndex(edge.Peer)
+
+			adj[u] = append(adj[u], v)
+			return nil
+		})
+	})
+
```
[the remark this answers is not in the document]
bhandras: done

lnrpc/rpc.proto
bhandras: This option is now removed in favor of the new GetNodeMetrics RPC call which doesn't have this option. PTAL

bhandras:
As discussed offline, `GetNodeMetrics` now returns a map of `NodeMetrics`. Centrality is now returned normalized and non-normalized too.

cmd/lncli/commands.go
```diff
@@ -2974,6 +2974,29 @@ func describeGraph(ctx *cli.Context) error {
 	return nil
 }
 
+var getNodeMetricsCommand = cli.Command{
+	Name:        "getnodemetrics",
+	Category:    "Peers",
```
cfromknecht: this doesn't exactly pertain to peers, though i see `DescribeGraph` and `GetNodeInfo` already use this category. Perhaps we should group `DescribeGraph`, `GetNodeInfo`, `GetChanInfo`, and `GetNodeMetrics` under `"Graph"`?

lnrpc/rpc.proto
```diff
@@ -641,6 +641,17 @@ service Lightning {
         };
     }
 
+    /** lncli: `getnodemetrics`
+    GetNodeMetrics returns node metrics calculated from the graph. Currently
+    the only supported metric is betweenness centrality of individual nodes.
+    */
+    rpc GetNodeMetrics (NodeMetricsRequest) returns (NodeMetricsResponse) {
+        option (google.api.http) = {
+            get: "/v1/graph"
```
cfromknecht: doesn't this need to be distinct rom `DescribeGraph`'s?

rpcserver.go
```diff
@@ -4657,6 +4663,43 @@ func marshalDbEdge(edgeInfo *channeldb.ChannelEdgeInfo,
...
+	centralityMetric := autopilot.NewBetweennessCentralityMetric()
+	if err := centralityMetric.Refresh(channelGraph); err != nil {
+		return nil, err
+	}
+	resp.Metrics = make(map[string]*lnrpc.NodeMetrics)
+
+	// Fill normalized and non normalized centrality.
+	centrality := centralityMetric.GetMetric(true)
+	for nodeID, val := range centrality {
+		resp.Metrics[hex.EncodeToString(nodeID[:])] = &lnrpc.NodeMetrics{
```
cfromknecht: i believe we can cast these directly as `string(nodeID[:])`, they will still be printed properly in lncli due to hex-protobuf-display

rpcserver.go
```diff
@@ -4657,6 +4663,43 @@ func marshalDbEdge(edgeInfo *channeldb.ChannelEdgeInfo,
...
+	graph := r.server.chanDB.ChannelGraph()
+
+	// Calculate betweenness centrality if requested. Note that depending on the
+	// graph size, this may take up to a few minutes.
+	channelGraph := autopilot.ChannelGraphFromDatabase(graph)
+	centralityMetric := autopilot.NewBetweennessCentralityMetric()
+	if err := centralityMetric.Refresh(channelGraph); err != nil {
+		return nil, err
+	}
+	resp.Metrics = make(map[string]*lnrpc.NodeMetrics)
```
cfromknecht: initialize along with `resp`?

lnrpc/rpc.proto
```diff
@@ -2516,6 +2527,25 @@ message ChannelGraph {
     repeated ChannelEdge edges = 2;
 }
 
+message NodeMetricsRequest {
```
cfromknecht: imo the request should accept an (repeated) enum stating which metric(s) we want to collect. otherwise we will always need to compute betweenness centrality even if we add more metrics because we won't have a way to stop computing betweenness centrality in a backwards compatible way.

rpcserver.go
```diff
@@ -4657,6 +4663,43 @@ func marshalDbEdge(edgeInfo *channeldb.ChannelEdgeInfo,
...
+
+	// Obtain the pointer to the global singleton channel graph, this will
+	// provide a consistent view of the graph due to bolt db's
+	// transactional model.
+	graph := r.server.chanDB.ChannelGraph()
+
+	// Calculate betweenness centrality if requested. Note that depending on the
+	// graph size, this may take up to a few minutes.
+	channelGraph := autopilot.ChannelGraphFromDatabase(graph)
+	centralityMetric := autopilot.NewBetweennessCentralityMetric()
```
cfromknecht: along the lines of my comment on the request message, i think this should be computed conditionally depending on which metric the user requested.

autopilot/betweenness_centrality_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	tests := []struct {
+		name       string
+		normalize  bool
+		centrality []float64
+	}{
+		{
+			normalize: true,
+			centrality: []float64{
+				0.2, 0.0, 0.2, 1.0, 0.4, 0.4, 7.0 / 15.0, 0.0, 0.0,
```
cfromknecht: 7.0 / 15.0?

autopilot/interface.go
```diff
@@ -210,9 +206,11 @@ func init() {
 	}
 
 	// Fill available node metrics.
-	for _, h := range availableNodeMetrics {
-		AvailableNodeMetrics[h.Name()] = h
+	centrality, err := NewBetweennessCentralityMetric(runtime.NumCPU())
+	if err != nil {
+		panic(err)
 	}
+	AvailableNodeMetrics[centrality.Name()] = centrality
```
cfromknecht: is this ever used? should the rpcserver share this singleton?

rpcserver.go
bhandras: Unfortunately this is not the case, hex encoding is needed.

`Internal desc = grpc: error while marshaling: proto: field "lnrpc.NodeMetricsResponse.Metrics" contains invalid UTF-8`

cfromknecht: bummer, strings it is!

cmd/lncli/commands.go
bhandras: agree, added a new mini commit for DescribeGraph, GetNodeInfo and GetChanInfo.

lnrpc/rpc.proto
bhandras: sorry, copy pasta

rpcserver.go
bhandras: done

lnrpc/rpc.proto
bhandras: Agree PTAL

rpcserver.go
bhandras: done

autopilot/betweenness_centrality_test.go
bhandras: 0.46666666666666667 :)

autopilot/interface.go
bhandras: yeah, removed until needed

lnrpc/rpc.proto
```diff
@@ -2516,6 +2527,32 @@ message ChannelGraph {
     repeated ChannelEdge edges = 2;
 }
 
+enum NodeMetricType {
```
halseth: nice, i like this!

cmd/lncli/commands.go
```diff
@@ -2942,7 +2942,7 @@ func listInvoices(ctx *cli.Context) error {
 
 var describeGraphCommand = cli.Command{
 	Name:     "describegraph",
-	Category: "Peers",
+	Category: "Graph",
```
halseth: 👍

lnrpc/rpc.proto
```diff
@@ -2516,6 +2527,32 @@ message ChannelGraph {
...
+}
+
+message NodeMetricsRequest {
+  /// The requesteded node metrics.
+  repeated NodeMetricType types = 1;
+}
+
+message NodeMetricsResponse {
+    /// Node metrics by node pubkey.
+    map<string, NodeMetrics> metrics = 1;
```
halseth: how will `NodeMetrics` work for other metrics? Since it has centrality as explicit fields.

bhandras: Changed as we discussed offline. PTAL

halseth: I'm okay doing it this way.

lnrpc/rpc.proto
```diff
@@ -2516,6 +2527,34 @@ message ChannelGraph {
     repeated ChannelEdge edges = 2;
 }
 
+enum NodeMetricType {
+    BETWEENNESS_CENTRALITY = 0;
+}
+
+message NodeMetricsRequest {
+  /// The requesteded node metrics.
```
cfromknecht: requsteded

lnrpc/rpc.proto
```diff
@@ -2516,6 +2527,34 @@ message ChannelGraph {
...
+    Betweenness centrality is the sum of the ratio of shortest paths that pass
+    through the node for each pair of nodes in the graph (not counting paths
+    starting or ending at this node).
+    Map of node pubkey to betweenness centrality of the node. Normalized
+    values are in the [0,1] closed interval.
+    */
+    map<string, FloatValue> betweenness_centrality = 1;
+}
+
+message FloatValue {
```
cfromknecht: naming suggestion: `FloatMetric`?

autopilot/betweenness_centrality.go
```diff
@@ -158,22 +174,59 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
...
+		partials <- partial
+	}
+
+	// Now start the N workers.
+	wg.Add(bc.workers)
+	for i := 0; i < bc.workers; i++ {
+		go worker()
+	}
+
+	// Distribute work amongst workers Should be
```
cfromknecht: early line wrap?

autopilot/betweenness_centrality.go
```diff
@@ -158,22 +174,59 @@ func (bc *BetweennessCentrality) Refresh(graph ChannelGraph) error {
 		return err
 	}
 
-	// TODO: parallelize updates to centrality.
-	centrality := make([]float64, len(cache.Nodes))
+	var wg sync.WaitGroup
+	work := make(chan int)
+	partials := make(chan []float64, bc.workers)
+
+	// Each worker will compute a partial result. This
```
cfromknecht: early line wrap?

autopilot/betweenness_centrality_test.go
```diff
@@ -110,47 +134,54 @@ func TestBetweennessCentralityWithNonEmptyGraph(t *testing.T) {
...
-							node, test.centrality[node], calculatedCentrality)
+					for node, nodeCentrality := range expected.centrality {
+						nodeID := NewNodeID(graphNodes[node])
+						calculatedCentrality, ok := centrality[nodeID]
+						if !ok {
+							t1.Fatalf("no result for node: %x (%v)", nodeID, node)
+						}
+
+						if nodeCentrality != calculatedCentrality {
+							t1.Errorf("centrality for node: %v should be %v, got: %v",
```
cfromknecht: lines in this file look > 80 chars

Issue #1226, opened 2018-05-10
contractcourt: implement smarter "go to chain" heuristics in the channelArbitrator 


Roasbeef:
The `channelArbitrator`, amongst may other things is responsible for deciding when we should go to chain in order to sweep outputs on-chain from expired contracts. We use the [`shouldGoOnChain`](https://github.com/lightningnetwork/lnd/blob/master/contractcourt/channel_arbitrator.go#L747) function currently to decide exactly _when_ we should go on chain. Atm, the function is very basic, and will determine that we need to go on chain if the HTLC is _about_ to expire. However, this doesn't capture the cost of going on chain. For example, if we have a 1000 satoshi HTLC that's about to expire, and it would take 20k satoshis to actually fully sweep the output, then the cost simply isn't worth it. The `shouldGoOnChain` function should be modified to capture such opportunity cost heuristics. 

# Steps To Completion 

- [ ] Extend the `shouldGoOnChain` method to take into account (all or some of): the on chain fees to sweep an HTLC, the duration that an HTLC has been active, the fee revenue related to the channel we're examining (as if the channel is popular we shouldn't close it to sweep a single HTLC), the CSV value. 

- [ ] A series of tests should be written to ensure that each of the heuristics are properly implemented and are adhered to by the `channelArbitrator`.

34ro:
@Roasbeef I'd like to try this.

> the on chain fees to sweep an HTLC, the duration that an HTLC has been active

I think this is not so difficult if it is implemented like `estimatesmartfee` in bitcoind.
They are the first tasks.

>  the fee revenue related to the channel we're examining (as if the channel is popular we shouldn't close it to sweep a single HTLC), the CSV value

It is difficult to define what is CSV for LN.

carlaKC:
>     For example, if we have a 1000 satoshi HTLC that's about to expire, and it would take 20k satoshis to actually fully sweep the output, then the cost simply isn't worth it.

I think that we'd need some way to record that we deemed a htlc non-economical to sweep? Otherwise if they're swept by our peer we've got less funds than we'd expect and have to start following outputs on chain to figure out what happened.

GordianLN:
Being the author of #6749 which was closed as duplicate of this one, I'm surprised to notice that the issue has been open for quite a while, and without much activity in general.
I would think this issue would involve many operators, and remove the possibility of wasting funds which is always good.
Does it pose any particular difficulty to implement it? Like:
```
[...]

PR #3876, opened 2019-12-26
contractcourt: keep channels with timed-out initiated htlcs.

roeierez:
This PR main goal is to reduce force closed channels in a mobile node.
In case of an outgoing pending htlc, a forwarding node needs to make sure it is either cancelled or resolved on-chain after timeout.
While for an online node it is more likely that its counterpart peer will send `update_fail_htlc` before it will need to use the timeout transaction, in mobile node it is a bit different.
By its nature, mobile node is most of the time not online, which in case of pending htlc it is likely to miss the time slot for getting the `update_fail_htlc` and when comes online it will close the channel on-chain.
While in a forwarding payment the timeout transaction should be used immediately to eliminate the risk in loosing funds, it looks to me that in an initiated payment it is not risky to keep the channel at this situation. The advantage is that when the mobile node is online it will have a chance to receive the `udpate_fail_htlc`, preventing the channel from being closed.
This behavior also makes paying hold invoices in mobile a lot more practical. Assuming the hold invoice is waiting for some external info to be resolved, it can take time in which the mobile goes offline.
If the counterpart peer misbehaves and doesn't send the `udpate_fail_htlc`, then the user can always force close the channel by himself.

A flag was added to the configuration to be able to explicitly specify this behavior and in such case for every outgoing htlc that is closed to timeout and beyond the code doesn't take on-chain action if the htlc is associated with a direct payment of the user node.

### DIFF
--- a/config.go
+++ b/config.go
@@ -58,15 +58,16 @@ const (
 	// pending channels permitted per peer.
 	DefaultMaxPendingChannels = 1
 
-	defaultNoSeedBackup             = false
-	defaultTrickleDelay             = 90 * 1000
-	defaultChanStatusSampleInterval = time.Minute
-	defaultChanEnableTimeout        = 19 * time.Minute
-	defaultChanDisableTimeout       = 20 * time.Minute
-	defaultMaxLogFiles              = 3
-	defaultMaxLogFileSize           = 10
-	defaultMinBackoff               = time.Second
-	defaultMaxBackoff               = time.Hour
+	defaultNoSeedBackup                  = false
+	defaultPaymentsExpirationGracePeriod = time.Duration(0)
+	defaultTrickleDelay                  = 90 * 1000
+	defaultChanStatusSampleInterval      = time.Minute
+	defaultChanEnableTimeout             = 19 * time.Minute
+	defaultChanDisableTimeout            = 20 * time.Minute
+	defaultMaxLogFiles                   = 3
+	defaultMaxLogFileSize                = 10
+	defaultMinBackoff                    = time.Second
+	defaultMaxBackoff                    = time.Hour
 
 	defaultTorSOCKSPort            = 9050
 	defaultTorDNSHost              = "soa.nodes.lightning.directory"
@@ -298,10 +299,11 @@ type config struct {
 
 	NoSeedBackup bool `long:"noseedbackup" description:"If true, NO SEED WILL BE EXPOSED AND THE WALLET WILL BE ENCRYPTED USING THE DEFAULT PASSPHRASE -- EVER. THIS FLAG IS ONLY FOR TESTING AND IS BEING DEPRECATED."`
 
-	TrickleDelay             int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
-	ChanEnableTimeout        time.Duration `long:"chan-enable-timeout" description:"The duration that a peer connection must be stable before attempting to send a channel update to reenable or cancel a pending disables of the peer's channels on the network (default: 19m)."`
-	ChanDisableTimeout       time.Duration `long:"chan-disable-timeout" description:"The duration that must elapse after first detecting that an already active channel is actually inactive and sending channel update disabling it to the network. The pending disable can be canceled if the peer reconnects and becomes stable for chan-enable-timeout before the disable update is sent. (default: 20m)"`
-	ChanStatusSampleInterval time.Duration `long:"chan-status-sample-interval" description:"The polling interval between attempts to detect if an active channel has become inactive due to its peer going offline. (default: 1m)"`
+	PaymentsExpirationGracePeriod time.Duration `long:"payments-expiration-grace-period" description:"A period to wait before force closing channels with outgoing htlcs that have timed-out and are a result of this node initiated payments."`
+	TrickleDelay                  int           `long:"trickledelay" description:"Time in milliseconds between each release of announcements to the network"`
+	ChanEnableTimeout             time.Duration `long:"chan-enable-timeout" description:"The duration that a peer connection must be stable before attempting to send a channel update to reenable or cancel a pending disables of the peer's channels on the network (default: 19m)."`
+	ChanDisableTimeout            time.Duration `long:"chan-disable-timeout" description:"The duration that must elapse after first detecting that an already active channel is actually inactive and sending channel update disabling it to the network. The pending disable can be canceled if the peer reconnects and becomes stable for chan-enable-timeout before the disable update is sent. (default: 20m)"`
+	ChanStatusSampleInterval      time.Duration `long:"chan-status-sample-interval" description:"The polling interval between attempts to detect if an active channel has become inactive due to its peer going offline. (default: 1m)"`
 
 	Alias       string `long:"alias" description:"The node alias. Used as a moniker by peers and intelligence services"`
 	Color       string `long:"color" description:"The color of the node in hex format (i.e. '#3399FF'). Used to customize node appearance in intelligence services"`
@@ -416,15 +418,16 @@ func loadConfig() (*config, error) {
 				"preferential": 1.0,
 			},
 		},
-		TrickleDelay:             defaultTrickleDelay,
-		ChanStatusSampleInterval: defaultChanStatusSampleInterval,
-		ChanEnableTimeout:        defaultChanEnableTimeout,
-		ChanDisableTimeout:       defaultChanDisableTimeout,
-		Alias:                    defaultAlias,
-		Color:                    defaultColor,
-		MinChanSize:              int64(minChanFundingSize),
-		NumGraphSyncPeers:        defaultMinPeers,
-		HistoricalSyncInterval:   discovery.DefaultHistoricalSyncInterval,
+		PaymentsExpirationGracePeriod: defaultPaymentsExpirationGracePeriod,
+		TrickleDelay:                  defaultTrickleDelay,
+		ChanStatusSampleInterval:      defaultChanStatusSampleInterval,
+		ChanEnableTimeout:             defaultChanEnableTimeout,
+		ChanDisableTimeout:            defaultChanDisableTimeout,
+		Alias:                         defaultAlias,
+		Color:                         defaultColor,
+		MinChanSize:                   int64(minChanFundingSize),
+		NumGraphSyncPeers:             defaultMinPeers,
+		HistoricalSyncInterval:        discovery.DefaultHistoricalSyncInterval,
 		Tor: &torConfig{
 			SOCKS:   defaultTorSOCKS,
 			DNS:     defaultTorDNS,
--- a/contractcourt/chain_arbitrator.go
+++ b/contractcourt/chain_arbitrator.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/clock"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
@@ -153,6 +155,19 @@ type ChainArbitratorConfig struct {
 	// OnionProcessor is used to decode onion payloads for on-chain
 	// resolution.
 	OnionProcessor OnionProcessor
+
+	// PaymentsExpirationGracePeriod indicates is a time window we let the
+	// other node to cancel an outgoing htlc that our node has initiated and
+	// has timed out.
+	PaymentsExpirationGracePeriod time.Duration
+
+	// IsForwardedHTLC checks for a given htlc, identified by channel id and
+	// htlcIndex, if it is a forwarded one.
+	IsForwardedHTLC func(chanID lnwire.ShortChannelID, htlcIndex uint64) bool
+
+	// Clock is the clock implementation that ChannelArbitrator uses.
+	// It is useful for testing.
+	Clock clock.Clock
 }
 
 // ChainArbitrator is a sub-system that oversees the on-chain resolution of all
--- a/contractcourt/channel_arbitrator.go
+++ b/contractcourt/channel_arbitrator.go
@@ -1113,17 +1113,17 @@ func (c ChainActionMap) Merge(actions ChainActionMap) {
 // we should go on chain to claim.  We do this rather than waiting up until the
 // last minute as we want to ensure that when we *need* (HTLC is timed out) to
 // sweep, the commitment is already confirmed.
-func (c *ChannelArbitrator) shouldGoOnChain(htlcExpiry, broadcastDelta,
-	currentHeight uint32) bool {
+func (c *ChannelArbitrator) shouldGoOnChain(htlc channeldb.HTLC,
+	broadcastDelta, currentHeight uint32) bool {
 
 	// We'll calculate the broadcast cut off for this HTLC. This is the
 	// height that (based on our current fee estimation) we should
 	// broadcast in order to ensure the commitment transaction is confirmed
 	// before the HTLC fully expires.
-	broadcastCutOff := htlcExpiry - broadcastDelta
+	broadcastCutOff := htlc.RefundTimeout - broadcastDelta
 
 	log.Tracef("ChannelArbitrator(%v): examining outgoing contract: "+
-		"expiry=%v, cutoff=%v, height=%v", c.cfg.ChanPoint, htlcExpiry,
+		"expiry=%v, cutoff=%v, height=%v", c.cfg.ChanPoint, htlc.RefundTimeout,
 		broadcastCutOff, currentHeight)
 
 	// TODO(roasbeef): take into account default HTLC delta, don't need to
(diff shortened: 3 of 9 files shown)


Roasbeef:
If a node never goes on chain for timed out HTLCs (ever), then the funds will be suspended until the channel is closed. Even then, you cannot perform a cooperative channel close while you have outstanding outgoing HTLCs. As a result, the channel is locked into force closing at _some time_ in the future once it holds onto a timed out HTLC. It's also the case that as more HTLCs build up on the commitment transaction, it becomes more expensive to completely sweep all the funds. 

Rather than _never_ going on chain to time out HTLCs, a more flexible path would be following through with this issue: https://github.com/lightningnetwork/lnd/issues/1226. The goal here is to factor additional context other than the expiry of an HTLC when deciding to go to chain. In this case, rather than never going to chain, we could only go to chain after some % threshold of funds become unavailable.

roeierez:
Thanks @Roasbeef  for going over this and for all the details you added. 
I believe https://github.com/lightningnetwork/lnd/issues/1226 handles the case where the other peer didn't behave well by not cancelling the htlc on time or maybe it wasn't online to do so.

> In this case, rather than never going to chain, we could only go to chain after some % threshold of funds become unavailable.

The intention of this PR is to give a chance for the other peer to cancel (because the user node was offline) rather than never going on-chain. I think if the user node won't get the cancel then https://github.com/lightningnetwork/lnd/issues/1226 comes to play.

joostjager:
It could be that the outgoing payment serves to learn a preimage that is needed to settle an on-chain contract, like for example in the case of loop out.

If we wouldn't go to chain, the on-chain contract may expire and thereafter the off-chain payment can still be settled.

Isn't there a less impactful solution possible? A way to give peers some time to send their updates, but with a timeout after which we still go to chain.

roeierez:
> If we wouldn't go to chain, the on-chain contract may expire and thereafter the off-chain payment can still be settled.

A good point.

> Isn't there a less impactful solution possible? A way to give peers some time to send their updates, but with a timeout after which we still go to chain.

Agree with the approach.
One way to lower the impact is to record the start time when creating the `ChannelArbitrator` and then on `shouldGoOnChain`, that is called for every new block, check that a certain timeout has passed (1 minute as default should be enough) before going to chain.
During this time it is expected that both peers will be connected and the update htlc will be received while failing to do so will result in going to chain.
I will push a change at this direction looking for your feedback.

roeierez:
@joostjager added a grace period before force closing the channel.

joostjager:
Ah, indeed, loop in. For loop out, the payer knows the preimage already

contractcourt/channel_arbitrator.go
```diff
@@ -1114,8 +1114,28 @@ func (c ChainActionMap) Merge(actions ChainActionMap) {
...
-
+	currentHeight uint32, rHash [32]byte) (bool, error) {
+
+	// In the case where the user specified to not force close channels that
+	// have outgoing htlc which is a result of its own payment, then we execute
+	// this check and try to exit early.
+	// This shouldn't add any security risk as there is no incoming htlc to
+	// fulfill at this case and the expectation is that when the channel is
+	// is active the other node will send update_fail_htlc to remove the htlc
+	// without closing the channel. it is up to the user to force close the
```
cfromknecht: it -> It

contractcourt/channel_arbitrator.go
```diff
@@ -1114,8 +1114,28 @@ func (c ChainActionMap) Merge(actions ChainActionMap) {
 // last minute as we want to ensure that when we *need* (HTLC is timed out) to
 // sweep, the commitment is already confirmed.
 func (c *ChannelArbitrator) shouldGoOnChain(htlcExpiry, broadcastDelta,
-	currentHeight uint32) bool {
-
+	currentHeight uint32, rHash [32]byte) (bool, error) {
+
+	// In the case where the user specified to not force close channels that
```
cfromknecht: nice comment!

contractcourt/chain_arbitrator.go
```diff
@@ -154,10 +155,10 @@ type ChainArbitratorConfig struct {
 	// resolution.
 	OnionProcessor OnionProcessor
 
-	// KeepChannelsWithPendingPayments indicates if we will force close
-	// channels with outgoing htlcs that have timed-out and they are
-	// initiated by this node (e.g initiated payment).
-	KeepChannelsWithPendingPayments bool
+	// PaymentsExpirationGracePriod indicates is a time window we let the
+	// other node to cancel an ougoing htlc that our node has initiated and
```
cfromknecht: ougoing -> outgoing

contractcourt/channel_arbitrator.go
```diff
@@ -1116,22 +1121,25 @@ func (c ChainActionMap) Merge(actions ChainActionMap) {
...
-		}
-		if initiated {
+	initiated, err := c.cfg.IsInitiatedPayment(rHash)
+	if err != nil {
+		return false, err
+	}
+	if initiated {
+		runningDuration := time.Now().Sub(c.startTimestamp)
+		if runningDuration < c.cfg.PaymentsExpirationGracePriod ||
+			currentHeight < htlcExpiry {
```
cfromknecht: should this case cause the following log statement? perhaps it should be checked above with `initiated`? not totally clear where this was checked made before and why it's added here.

roeierez: This case is actually for the return statement following the log statement.
Anyway, looking again I realized there is a more intuitive way to implement this case that is also easier to understand as an increment to the old code.
Pushed the changes.

contractcourt/chain_arbitrator.go
roeierez: fixed.

contractcourt/channel_arbitrator.go
roeierez: fixed

contractcourt/channel_arbitrator.go
```diff
@@ -1194,10 +1228,13 @@ func (c *ChannelArbitrator) checkCommitChainActions(height uint32,
 			continue
 		}
 
-		toChain := c.shouldGoOnChain(
+		toChain, err := c.shouldGoOnChain(
 			htlc.RefundTimeout, c.cfg.IncomingBroadcastDelta,
-			height,
+			height, htlc.RHash,
```
joostjager: Could it be worth to split `shouldGoOnChain` in two versions, one for incoming and one for outgoing? Then we also don't need to pass in the hash and do the payment lookup.

server.go
```diff
@@ -915,11 +915,19 @@ func newServer(listenAddrs []net.Addr, chanDB *channeldb.DB,
...
-		NotifyClosedChannel: s.channelNotifier.NotifyClosedChannelEvent,
-		OnionProcessor:      s.sphinx,
+		DisableChannel:               s.chanStatusMgr.RequestDisable,
+		Sweeper:                      s.sweeper,
+		Registry:                     s.invoices,
+		NotifyClosedChannel:          s.channelNotifier.NotifyClosedChannelEvent,
+		OnionProcessor:               s.sphinx,
+		PaymentsExpirationGracePriod: cfg.PaymentsExpirationGracePriod,
+		IsInitiatedPayment: func(hash [32]byte) (bool, error) {
+			payment, err := paymentControl.FetchPayment(hash)
```
joostjager: This looks like a layering violation. Payments are a higher level concept and can get complicated with mpp (multiple outgoing htlcs with the same hash) and amp (multiple htlcs with different hashes). Maybe it is better to check the circuit map to see if this is a forward?

contractcourt/channel_arbitrator.go
roeierez: > Could it be worth to split `shouldGoOnChain` in two versions

I will give it a try and see if it is more clear this way.


> Then we also don't need to pass in the hash and do the payment lookup

We will need the lookup anyway to differentiate forward from initiated payment.

joostjager: Isn't an incoming htlc never an initiated payment?

roeierez: > Isn't an incoming htlc never an initiated payment?

Indeed, it wasn't clear to me that you intended to avoid the lookup only for the incoming. Understood now.

server.go
roeierez: I agree. I think instead we can query the Switch for whether this is a forward.
@joostjager  Is it correct to say that if the field `AddRef` of `PaymentCircuit` struct is **NOT** nil then we are dealing with a forward?

joostjager: I think we mostly check `incomingChanID == hop.Source`

contractcourt/channel_arbitrator.go
roeierez: I passed the htlc struct to the `shouldGoToChain` method so we can take a decision there separately for incoming/outgoing htlcs and avoid the query in case of incoming.

server.go
roeierez: @joostjager  The implementation of the check for forward/initiated htlc is now moved to the switch.

htlcswitch/switch.go
```diff
@@ -460,6 +460,18 @@ func (s *Switch) UpdateForwardingPolicies(
 	s.indexMtx.RUnlock()
 }
 
+// HasForwardedHTLCWithHash checks for a given htlc hash if it is related
+// to an existing circuit that represents a forwarded payment.
+func (s *Switch) HasForwardedHTLCWithHash(paymentHash [32]byte) bool {
+	paymentCircuites := s.circuits.LookupByPaymentHash(paymentHash)
```
joostjager: Circuites -> circuits

contractcourt/channel_arbitrator.go
```diff
@@ -1132,7 +1137,29 @@ func (c *ChannelArbitrator) shouldGoOnChain(htlcExpiry, broadcastDelta,
...
+	// both nodes to connect and give a chance to the other node to send its
+	// updates and cancel the htlc.
+	// This shouldn't add any security risk as there is no incoming htlc to
+	// fulfill at this case and the expectation is that when the channel is
+	// active the other node will send update_fail_htlc to remove the htlc
+	// without closing the channel. It is up to the user to force close the
+	// channel if the peer misbehaves and doesn't send the update_fail_htlc.
+	// It is useful when this node is most of the time not online and is
+	// likely to miss the time slot where the htlc may be cancelled.
+	hasForward := c.cfg.HasForwardedHTLCWithHash(htlc.RHash)
```
joostjager: We know the channel id here and the htlc index, so it could be a lookup by outgoing circuit instead of looking up all circuits with that hash.

roeierez: I see now you probably mean to use `LookupOpenCircuit` to fetch the exact circuit.
To execute this query I need to build the `CirtuitKey`. Quick question: Is the `HtlcIndex` of the `channeldb.HTLC` represents the `HtlcID` of the `CirtuitKey` ?

roeierez: I pushed changes with this feedback.

joostjager: Ok, looks good. Yes, `HtlcIndex` is `HtlcId`.

joostjager:
Implementation LGTM, but it would be good to have some unit test coverage on this. In order to do this, the call to `time.Now` can be replaced by injecting a `clock.Clock` dependency.

roeierez:
> it would be good to have some unit test coverage on this

I Wanted to see that the changes are accepted before moving on to this. I will definitely work on that now.

roeierez:
Test added.

contractcourt/channel_arbitrator_test.go
```diff
@@ -1811,3 +1818,87 @@ func TestChannelArbitratorDanglingCommitForceClose(t *testing.T) {
...
+		resolvers: make(map[ContractResolver]struct{}),
+	}
+	chanArbCtx, err := createTestChannelArbitrator(t, log)
+	if err != nil {
+		t.Fatalf("unable to create ChannelArbitrator: %v", err)
+	}
+	chanArb := chanArbCtx.chanArb
+
+	// We'll inject a test clock implementation so we can control the uptime.
+	startTime := time.Now()
```
joostjager: For a deterministic test, it is better to inject a constant time here.

contractcourt/channel_arbitrator_test.go
```diff
@@ -1811,3 +1818,87 @@ func TestChannelArbitratorDanglingCommitForceClose(t *testing.T) {
...
+
+	// We will advance the uptime to 10 seconds which should be still within
+	// the grace period and should not trigger going to chain.
+	testClock.SetTime(startTime.Add(time.Second * 10))
+	chanArbCtx.blockEpochs <- &chainntnfs.BlockEpoch{Height: 5}
+	chanArbCtx.AssertState(StateDefault)
+
+	// We will advance the uptime to 16 seconds which should trigger going
+	// to chain.
+	testClock.SetTime(startTime.Add(time.Second * 16))
```
joostjager: Nice unit test :+1:

htlcswitch/switch.go
```diff
@@ -460,6 +460,18 @@ func (s *Switch) UpdateForwardingPolicies(
...
+// to an opened circuit that represents a forwarded payment.
+func (s *Switch) IsForwardedHTLC(chanID lnwire.ShortChannelID,
+	htlcIndex uint64) bool {
+
+	circuit := s.circuits.LookupOpenCircuit(channeldb.CircuitKey{
+		ChanID: chanID,
+		HtlcID: htlcIndex,
+	})
+	return circuit != nil && circuit.Incoming.ChanID != hop.Source
+}
```
joostjager: This function isn't covered by the test above, but is important for proper behavior of the mechanism introduced in this pr. A low-effort option could be to add a few asserts in the existing switch tests.

contractcourt/channel_arbitrator.go
```diff
@@ -1113,17 +1118,17 @@ func (c ChainActionMap) Merge(actions ChainActionMap) {
 // we should go on chain to claim.  We do this rather than waiting up until the
 // last minute as we want to ensure that when we *need* (HTLC is timed out) to
 // sweep, the commitment is already confirmed.
-func (c *ChannelArbitrator) shouldGoOnChain(htlcExpiry, broadcastDelta,
+func (c *ChannelArbitrator) shouldGoOnChain(htlc channeldb.HTLC, broadcastDelta,
```
joostjager: For ease of reviewing, the non-functional refactorings (introducing `Clock` and the function signature change of `shouldGoOnChain`) could be isolated in a separate commit.

contractcourt/channel_arbitrator.go
```diff
@@ -1132,7 +1137,29 @@ func (c *ChannelArbitrator) shouldGoOnChain(htlcExpiry, broadcastDelta,
...
+	// both nodes to connect and give a chance to the other node to send its
+	// updates and cancel the htlc.
+	// This shouldn't add any security risk as there is no incoming htlc to
+	// fulfill at this case and the expectation is that when the channel is
+	// active the other node will send update_fail_htlc to remove the htlc
+	// without closing the channel. It is up to the user to force close the
+	// channel if the peer misbehaves and doesn't send the update_fail_htlc.
+	// It is useful when this node is most of the time not online and is
+	// likely to miss the time slot where the htlc may be cancelled.
+	isForwarded := c.cfg.IsForwardedHTLC(c.cfg.ShortChanID, htlc.HtlcIndex)
```
joostjager: Is there a possibility of any unexpected behavior because of concurrency / race conditions? That the switch already modified the circuit when we get here?

roeierez: If he htlc is settled/cancelled just before this line is executed and therefore no longer exist in the switch the `isForwarded` will be `false` which will cause going on-chain after the grace period.
Such modification by the switch may as well happen before this code change and in that case  will cause going to chain immediately. 
I don't find this difference as problematic.

contractcourt/channel_arbitrator.go
roeierez: @joostjager Let me verify I understood it correct:
On top of the original code of this file:
1. First commit for the signature change only.
2. Second commit for the shouldGoOnChain body and introducing Clock (As `Clock` is used in the new implementation).

Was that the suggestion?

joostjager: Ah yes, the `Clock` wasn't already there. In that case it is indeed only the signature change in a separate commit. But up to you whether you think it is worth it, it is non-blocking.

contractcourt/channel_arbitrator.go
joostjager: Make sense. What happens in link and what happens in the arbitrator is indeed already desynchronized.

htlcswitch/switch.go
roeierez: I took the low effort path which seems to me also the right one.

joostjager: Yes, very compact :+1:

config.go
```diff
@@ -416,15 +418,16 @@ func loadConfig() (*config, error) {
...
-		TrickleDelay:             defaultTrickleDelay,
-		ChanStatusSampleInterval: defaultChanStatusSampleInterval,
-		ChanEnableTimeout:        defaultChanEnableTimeout,
-		ChanDisableTimeout:       defaultChanDisableTimeout,
-		Alias:                    defaultAlias,
-		Color:                    defaultColor,
-		MinChanSize:              int64(minChanFundingSize),
-		NumGraphSyncPeers:        defaultMinPeers,
-		HistoricalSyncInterval:   discovery.DefaultHistoricalSyncInterval,
+		PaymentsExpirationGracePriod: defaultPaymentsExpirationGracePriod,
```
Roasbeef: PaymentsExpirationGracePriod -> PaymentsExpirationGrace _Period_ 

Propagates to several other areas in this PR.

roeierez: Thanks for noticing. Fixed.

contractcourt/chain_arbitrator.go
```diff
@@ -153,6 +155,19 @@ type ChainArbitratorConfig struct {
...
+	// PaymentsExpirationGracePeriod indicates is a time window we let the
+	// other node to cancel an outgoing htlc that our node has initiated and
+	// has timed out.
+	PaymentsExpirationGracePeriod time.Duration
+
+	// IsForwardedHTLC checks for a given htlc, identified by channel id and
+	// htlcIndex, if it is a forwarded one.
+	IsForwardedHTLC func(chanID lnwire.ShortChannelID, htlcIndex uint64) bool
+
+	// clock is the clock implementation that ChannelArbitrator uses.
```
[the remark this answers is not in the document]
roeierez: Thanks @cfromknecht, spelling fixed.

Issue #3875, opened 2019-12-26
Log performance impact of running without freelist sync


MrManPew:
### Background
It would be very efficient if there was some-kind of a progress-bar for lnd startup
See here
```
Dec 26 07:48:57 rock64 systemd[1]: Started LND Lightning Daemon.
Dec 26 07:48:57 rock64 lnd[13413]: Attempting automatic RPC configuration to bitcoind
Dec 26 07:48:57 rock64 lnd[13413]: Automatically obtained bitcoind's RPC credentials
Dec 26 07:48:57 rock64 lnd[13413]: 2019-12-26 07:48:57.762 [INF] LTND: Version: 0.8.0-beta commit=v0.8.0-beta, build=production, logging=default
Dec 26 07:48:57 rock64 lnd[13413]: 2019-12-26 07:48:57.764 [INF] LTND: Active chain: Bitcoin (network=mainnet)
Dec 26 07:54:16 rock64 lnd[13413]: 2019-12-26 07:54:16.998 [INF] CHDB: Checking for schema update: latest_version=11, db_version=11
```
It took  6 minutes for the start-up. This is long, and a progress-bar for this will be very helpful. Or at least to edit the log so it will state: "this will take a few minutes" (The bitcoin startup has some nice progressbars).

### Your environment

* version of `lnd`
`v0.8.0-beta,`
* which operating system (`uname -a` on *Nix)
`Linux rock64 4.4.132-1075-rockchip-ayufan-ga83beded8524 #1 SMP Thu Jul 26 08:22:22 UTC 2018 aarch64 GNU/Linux`
* version of `btcd`, `bitcoind`, or other backend
bitcoind

Roasbeef:
Start up is only slow if `lnd` needs to reconstruct the free list for its default database on start up. As the code to do this lives in another package, we can't easily add something like a progress bar to track the current progress. However, we should add some more logic in this area. 

If you want to have faster start up (like it was before), and don't mind the additional memory and hit to write performance, then you can run with this flag: `--sync-freelist`. 

> It took 6 minutes for the start-up. 

If you want to improve this time without running with the flag specified above, then you can _compact_ the database with these instructions: https://github.com/lightningnetwork/lnd/issues/2318#issuecomment-446407008. If you choose this route, make sure to take care to properly backup the database before compaction. Compaction will eliminate fragmentation on-disk, resulting in a smaller database, and quicker start up time as the free list will be much slimmer.

[...]

PR #3879, opened 2019-12-27
Update channeldb opening log

MrManPew:
It might take a few minutes and it should therefore state it...

See here: https://github.com/lightningnetwork/lnd/issues/3875


Roasbeef:
Committed as (19bdbe7f8d4c13d5eb9d21fd0d6fe8467f443da7), I also added an additional log line with timing information so users can easily track how long it takes to actually open the database.

PR #3882, opened 2019-12-29
config: Make bools in sample-lnd.conf consistent

### LINKED ISSUE #3810: Unclear/inconsistent handling of bools in lnd.conf
### Background

It is unclear which boolean values in lnd.conf can be specified as `0`/`1` or `true`/`false`. 

### Your environment

Irrelevant

### Steps to reproduce

1. Open the [sample conf file](https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf)
2. Find semantically boolean values (e.g. `nat` and `nobootstrap`)
3. See that there are inconsistent example values `true` and `1`
4. see that there's no explanation at the top of the sample

### Expected behaviour

One of these possible explanations written at the top of the config file:

* All boolean values might be specified with `true`/`false` or `1`/`0`
* Some boolean values must be specified as `true`/`false`, others as `1`/`0` - see the examples.

### Actual behaviour

It's unclear whether the values can be mixed.

Note: This is mainly important for systems that automate config generation.

philippgille:
Fixes #3810.

#### Pull Request Checklist

  logging level
  per default)
- ❌ Running `make check` does not fail any tests
  - `github.com/lightningnetwork/lnd/chainntnfs` tests fail, but I don't think this is because of the changes in the PR.
  already exist

### DIFF
--- a/sample-lnd.conf
+++ b/sample-lnd.conf
@@ -1,3 +1,7 @@
+; Example configuration for lnd.
+;
+; Boolean values can be specified as true/false or 1/0.
+
 [Application Options]
 
 ; The directory that lnd stores all wallet, chain, and channel related data
@@ -68,7 +72,7 @@
 
 ; Disable listening for incoming p2p connections.  This will override all
 ; listeners.
-; nolisten=1
+; nolisten=true
 
 ; Specify the interfaces to listen on for gRPC connections.  One listen
 ; address per line.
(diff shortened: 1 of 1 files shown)


philippgille:
P.S.: I decided to use true/false instead of 1/0 because it's more clear to beginners what this means. The distinction between a bool and a number is clear immediately. Also the Go programming language uses true/false instead of 1/0.

PR #3884, opened 2020-01-02
routing: clean up locking on topology change

bhandras:
routing: clean up locking on topology change

This is a trivial code clean up to avoid acquiring the reader lock twice. Fixes a race condition where the topologyClients could have changed while releasing the reader lock resulting in a wrongly logged count or worst case, return from the function.


bhandras:
> Would be useful to describe that this fixes a race condition with the log statement.

Updated the PR description to better describe the issue.

PR #3886, opened 2020-01-03
build: add version checks to release script

joostjager:
Enhances the release script to check that the specified tag is also the tag that is currently checked out. In addition to that, the script also ensures that the tag and the version specified in version.go match.

### DIFF
--- a/build/release/release.sh
+++ b/build/release/release.sh
@@ -16,6 +16,35 @@ if [[ $1x = x ]]; then
     TAG=$DATE-$VERSION
 else
     TAG=$1
+
+    # If a tag is specified, ensure that that tag is present and checked out.
+    if [[ $TAG != $(git tag -l --points-at HEAD) ]]; then
+        echo "tag $TAG not checked out"
+        exit 1
+    fi
+
+    # Build lnd to extract version.
+    go build github.com/lightningnetwork/lnd/cmd/lnd
+
+    # Extract version command output.
+    LND_VERSION_OUTPUT=`./lnd --version`
+
+    # Use a regex to isolate the version string.
+    LND_VERSION_REGEX="lnd version (.+) commit"
+    if [[ $LND_VERSION_OUTPUT =~ $LND_VERSION_REGEX ]]; then
+        # Prepend 'v' to match git tag naming scheme.
+        LND_VERSION="v${BASH_REMATCH[1]}"
+        echo "version: $LND_VERSION"
+
+        # Match git tag with lnd version.
+        if [[ $TAG != $LND_VERSION ]]; then
+            echo "lnd version $LND_VERSION does not match tag $TAG"
+            exit 1
+        fi
+    else
+        echo "malformed lnd version output"
+        exit 1
+    fi
 fi
 
 go mod vendor


build/release/release.sh
```diff
@@ -16,6 +16,34 @@ if [[ $1x = x ]]; then
...
     TAG=$1
+
+    # If a tag is specified, ensure that that tag is present and checked out.
+    if [[ $TAG != $(git tag -l --points-at HEAD) ]]; then
+        echo "tag $TAG not checked out"
+        exit 1
+    fi
+
+    # Build lnd to extract version.
+    go build github.com/lightningnetwork/lnd/cmd/lnd
```
guggero: In case lnd isn't checked out in that directory, wouldn't `go build ./cmd/lnd` be better?

build/release/release.sh
```diff
@@ -16,6 +16,34 @@ if [[ $1x = x ]]; then
...
+
+    # Extract version command output.
+    LND_VERSION_OUTPUT=`./lnd --version`
+
+    # Use a regex to isolate the version string.
+    if [[ $LND_VERSION_OUTPUT =~ lnd\ version\ (.+)\ commit ]]; then
```
guggero: To make this more readable and prevent bash version dependency, I'd suggest to change this to:

```
    LND_VERSION_REGEX="lnd version (.+) commit"
    if [[ $LND_VERSION_OUTPUT =~ $LND_VERSION_REGEX ]]; then
```
See https://stackoverflow.com/questions/218156/bash-regex-with-quotes/218217#218217

build/release/release.sh
joostjager: I've taken this from how the release binaries are built further down in this file. I think the whole script assumes that it is started from the repo root.

build/release/release.sh
joostjager: Interesting, didn't know that. Fixed

build/release/release.sh
guggero: Ah, I didn't see that. Makes sense to do it the same way here then.

build/release/release.sh
```diff
@@ -16,6 +16,35 @@ if [[ $1x = x ]]; then
     TAG=$DATE-$VERSION
 else
     TAG=$1
+
```
cfromknecht: while we are here, should we remove the date-version fallback above? there's no way for it to pass the new validation, since the output isn't in the `X.X.X` semver format produced by `lnd --version`

cfromknecht: instead we should just fail to build if it no `TAG` is specified

joostjager: I don't know what the original intention of the fallback is. But I can imagine for local testing of the script?

For the date-version fallback, we don't validate so the script should still work.

Roasbeef: The date fallback was just an old default, but with this new verification logic it can be used to just make sure things build without needing to checkout out a tag, etc.

build/release/release.sh
```diff
@@ -16,6 +16,35 @@ if [[ $1x = x ]]; then
...
+    go build github.com/lightningnetwork/lnd/cmd/lnd
+
+    # Extract version command output.
+    LND_VERSION_OUTPUT=`./lnd --version`
+
+    # Use a regex to isolate the version string.
+    LND_VERSION_REGEX="lnd version (.+) commit"
+    if [[ $LND_VERSION_OUTPUT =~ $LND_VERSION_REGEX ]]; then
+        # Prepend 'v' to match git tag naming scheme.
+        LND_VERSION="v${BASH_REMATCH[1]}"
```
Roasbeef: So the tags _do_ have that prefix, but the formatted version output for the `lnd/lncli` commands _don't_: 
```
lnd --version
lnd version 0.8.2-beta commit=queue/v1.0.2-165-g4273bc0ba22550536625c73d54539a7f6bc6030e
```

joostjager: What is the requested change? The comment says "prepend 'v' (to parsed version) to match git tag naming". I think that is correct.

PR #3887, opened 2020-01-03
channeldb: remove FetchAllInvoices which was test only and unused

bhandras:
This PR removes channeldb.FetchAllInvoices which was only used in tests to prepare expectation sets. The refactor also changes expectation set creation to be independent from the DB.

### DIFF
--- a/channeldb/invoice_test.go
+++ b/channeldb/invoice_test.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	emptyFeatures = lnwire.NewFeatureVector(nil, lnwire.Features)
+	testNow       = time.Unix(1, 0)
 )
 
 func randInvoice(value lnwire.MilliSatoshi) (*Invoice, error) {
@@ -23,9 +24,7 @@ func randInvoice(value lnwire.MilliSatoshi) (*Invoice, error) {
 	}
 
 	i := &Invoice{
-		// Use single second precision to avoid false positive test
-		// failures due to the monotonic time component.
-		CreationDate: time.Unix(time.Now().Unix(), 0),
+		CreationDate: testNow,
 		Terms: ContractTerm{
 			Expiry:          4000,
 			PaymentPreimage: pre,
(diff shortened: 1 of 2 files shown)


channeldb/invoice_test.go
```diff
@@ -351,28 +373,27 @@ func TestInvoiceAddTimeSeries(t *testing.T) {
...
-		t.Fatalf("unable to fetch invoices: %v", err)
-	}
+		// Create the settled invoice for the expectation set.
+		settleTestInvoice(invoice, settleIndex)
+		settleIndex++
 
-	// We'll slice off the first 10 invoices, as we only settled the last
-	// 10.
-	invoices = invoices[10:]
+		settledInvoices = append(settledInvoices, *copyInvoice(invoice))
```
joostjager: Is this copy necessary?

joostjager: Same question about other `copyInvoice` calls in test.

channeldb/invoices.go
```diff
@@ -1327,6 +1327,26 @@ func copySlice(src []byte) []byte {
 	return dest
 }
 
+// copyInvoiceHTLC makes a deep copy of the supplied invoice HTLC.
+func copyInvoiceHTLC(src *InvoiceHTLC) *InvoiceHTLC {
```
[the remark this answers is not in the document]
bhandras: Agree, commit message was a bit misleading. Reworded.

channeldb/invoices.go
```diff
@@ -1327,6 +1327,26 @@ func copySlice(src []byte) []byte {
...
+	result := &InvoiceHTLC{
+		Amt:           src.Amt,
+		MppTotalAmt:   src.MppTotalAmt,
+		AcceptHeight:  src.AcceptHeight,
+		AcceptTime:    src.AcceptTime,
+		ResolveTime:   src.ResolveTime,
+		Expiry:        src.Expiry,
+		State:         src.State,
+		CustomRecords: make(record.CustomSet),
+	}
```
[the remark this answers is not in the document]
bhandras: Done.

channeldb/invoice_test.go
bhandras: Removed as these are indeed not necessary. (Imo it's a bit cleaner to copy this way as we guarantee immutability if the test is refactored later though.)

channeldb/invoice_test.go
```diff
@@ -675,6 +673,7 @@ func TestQueryInvoices(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to make test db: %v", err)
 	}
+	db.Now = func() time.Time { return testNow }
```
guggero: nit: To make it more explicit that the same now function is now used everywhere we could extract this function declaration to the top as well and maybe call it `testNowFn`.

joostjager: Probably best fix is to add `Clock` to the database.

PR #3889, opened 2020-01-06
chanfitness: Bugfix save initial peer online event

carlaKC:
Uptime calculations rely on an initial online event to determine peer state which was removed in #3799. The change was appropriate for channel that are offline at start (since an event will be recorded if the peer comes online), but it introduces a race between peers coming online and existing channels being added to the store. If the peer comes online before the channel is added to the store, no online event is added (and the channel with an online peer appears to be offline). This change re-adds the channel online event in `addChannel` so that channels with already online peers have an initial online event.  

This edge case of needing a first online event seems unavoidable, and buggy, so I am going to follow up with a refactor of the unit tests to better cover this code. The issue here is that it's difficult to test a whole set of events, so I'm going to add an interface for `Subscribe.Client` and test a full event flow. 

For now, tested all possible scenarios with a local node:
On restart:
- Existing channels with online peers: lifetime equals time since restart, uptime potentially less than lifetime if the peer takes long to reconnect
- Existing channels with offline peers: lifetime equals time since restart, uptime is zero

New channels:
- New channel with online peer: uptime and lifetime equal time since open
- New channel with offline peer: zero uptime, lifetime set to time since open

Peer flaps:
- Peer goes offline: lifetime continues to increase, offline stops increasing
- Peer comes online: uptime and lifetime increase

### DIFF
--- a/chanfitness/chaneventstore_test.go
+++ b/chanfitness/chaneventstore_test.go
@@ -469,23 +469,61 @@ func TestGetUptime(t *testing.T) {
 
 // TestAddChannel tests that channels are added to the event store with
 // appropriate timestamps. This test addresses a bug where offline channels
-// did not have an opened time set.
+// did not have an opened time set, and checks that an online event is set for
+// peers that are online at the time that a channel is opened.
 func TestAddChannel(t *testing.T) {
 	_, vertex, chanPoint := getTestChannel(t)
 
-	store := NewChannelEventStore(&Config{})
+	tests := []struct {
+		name string
 
-	// Add channel to the store.
-	store.addChannel(chanPoint, vertex)
+		// peers maps peers to an online state.
+		peers map[route.Vertex]bool
 
-	// Check that the eventlog is successfully added.
-	eventlog, ok := store.channels[chanPoint]
-	if !ok {
-		t.Fatalf("channel should be in store")
+		expectedEvents []eventType
+	}{
+		{
+			name:           "peer offline",
+			peers:          make(map[route.Vertex]bool),
+			expectedEvents: []eventType{},
+		},
+		{
+			name: "peer online",
+			peers: map[route.Vertex]bool{
+				vertex: true,
+			},
+			expectedEvents: []eventType{peerOnlineEvent},
+		},
 	}
 
-	// Ensure that open time is always set.
-	if eventlog.openedAt.IsZero() {
-		t.Fatalf("channel should have opened at set")
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


carlaKC:
> Isn't this something that needs to be addressed in the startup sequence? Make sure the fitness subsystem is properly started before starting network connections.

Had a closer look at the start sequence and we start connections after the chanfitness subsystem, so channels that exist at startup will always have offline peers, and then add an online event if the peer re-connects on start. 

The issue is actually with new channels, which don't record the online event for already online peers (looked like a race to me because my startup script kept confirming channels when it mines blocks on restart). Adding an online event is how the original PR had [start online events](https://github.com/lightningnetwork/lnd/pull/3332/files#diff-cbc74c9a8edc6e13c50216d469018944R207), but my last bugfix moved it around and then deleted it (which made a worse bug). 

With this fix: channels that exist on startup will not have their peers tracked as online yet, because their peers have not connected, so `addChannel` will not record an event, and new channels will appropriately have an online event added it their peer is online.

joostjager:
In that case, isn't it better to not add the online event in `addChannel` (where it is no-op for existing channels), but do it when a `OpenChannelEvent` is received?

carlaKC:
> In that case, isn't it better to not add the online event in `addChannel` (where it is no-op for existing channels), but do it when a `OpenChannelEvent` is received?

Yes it could also be there. I think it's more visible in `addChannel`, which is better than having it hidden in the large main loop? Even if we handled existing channels in the main loop through some sort of historic subscription and only had to call `addChannel` once (ie did not need to get the open channels and add them), I think I'd still want to keep it there.

chanfitness/chaneventstore_test.go
```diff
@@ -469,23 +469,64 @@ func TestGetUptime(t *testing.T) {
...
 	}
 
-	// Ensure that open time is always set.
-	if eventlog.openedAt.IsZero() {
-		t.Fatalf("channel should have opened at set")
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+
```
joostjager: newline

chanfitness/chaneventstore_test.go
```diff
@@ -469,23 +469,64 @@ func TestGetUptime(t *testing.T) {
...
+					t.Fatalf("expected: %v, got: %v",
+						e, eventLog.events[i].eventType)
+				}
+			}
+
+			// Ensure that open time is always set.
+			if eventLog.openedAt.IsZero() {
+				t.Fatalf("channel should have opened at set")
+			}
+
```
joostjager: newline

PR #3891, opened 2020-01-07
routing: use updated sphinx.NewOnionPacket method

Roasbeef:
In this PR, we update the routing package to use the new
`sphinx.NewOnionPacket` method. The new version of this method allows us
to specify _how_ the packet should be filled before it's used to create
a mix-header. This isn't a fundamental change (totally backwards
compatible), instead it plugs a privacy leak that may have revealed to
the destination how long the true route was.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/kkdai/bstream v0.0.0-20181106074824-b3251f7901ec
 	github.com/lightninglabs/neutrino v0.11.0
 	github.com/lightninglabs/protobuf-hex-display v1.3.3-0.20191212020323-b444784ce75d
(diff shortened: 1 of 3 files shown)


Roasbeef:
There's some weirdness with modules: the new tag is pushed, the `go.sum+go.mod` were updated, but compilation doesn't work for me locally.... 

If I pin it to the latest commit hash instead, then things work. Perhaps their crawler just needs more time to pick up the latest tag?

joostjager:
It seems the tag isn't on the latest commit. It is on f34e9dc while this pr needs 25954be99ab05ad18ae745a00f81f271ae3590c4

Roasbeef:
Figured out the issue with the tag, should be updated properly and build now.

PR #3892, opened 2020-01-07
contractcourt: fix lingering contract after local breach

matheusd:
Fix an issue that started happening after commit 9acb2366656481c94f365e2e12d02f3129c4fb91.

After that commit, the commit sweeper for a channel arbitrator would no longer wait for a spend notification but rather for a specific number of blocks leaving open the possibility that a sweep by the remote party (such as when the channel was actually breached by the local party) would leave the channel forever in a pending state.

This fixes that by noticing that sweep errors where the sweep was not made by us should still cause the channel to be considered fully resolved.

The first commit surfaces the issue by verifying the state of pending channels after the testRevokedCloseRetribution integration test.

### DIFF
--- a/contractcourt/commit_sweep_resolver.go
+++ b/contractcourt/commit_sweep_resolver.go
@@ -239,24 +239,40 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 	// possible and publish the sweep tx. When the sweep tx
 	// confirms, it signals us through the result channel with the
 	// outcome. Wait for this to happen.
+	recovered := true
 	select {
 	case sweepResult := <-resultChan:
-		if sweepResult.Err != nil {
+		switch sweepResult.Err {
+		case sweep.ErrRemoteSpend:
+			// If the remote party was able to sweep this output
+			// it's likely what we sent was actually a revoked
+			// commitment. Report the error and continue to wrap up
+			// the contract.
+			c.log.Errorf("local commitment output was swept by "+
+				"remote party via %v", sweepResult.Tx.TxHash())
+			recovered = false
+		case nil:
+			// No errors, therefore continue processing.
+			c.log.Infof("local commitment output fully resolved by "+
+				"sweep tx: %v", sweepResult.Tx.TxHash())
+		default:
+			// Unknown errors.
 			c.log.Errorf("unable to sweep input: %v",
 				sweepResult.Err)
 
 			return nil, sweepResult.Err
 		}
-
-		c.log.Infof("commit tx fully resolved by sweep tx: %v",
-			sweepResult.Tx.TxHash())
 	case <-c.quit:
 		return nil, errResolverShuttingDown
 	}
 
 	// Funds have been swept and balance is no longer in limbo.
 	c.reportLock.Lock()
-	c.currentReport.RecoveredBalance = c.currentReport.LimboBalance
+	if recovered {
+		// We only record the balance as recovered if it actually came
+		// back to us.
+		c.currentReport.RecoveredBalance = c.currentReport.LimboBalance
+	}
 	c.currentReport.LimboBalance = 0
 	c.reportLock.Unlock()
 
--- a/contractcourt/commit_sweep_resolver.go
+++ b/contractcourt/commit_sweep_resolver.go
@@ -248,7 +248,7 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 			// it's likely what we sent was actually a revoked
 			// commitment. Report the error and continue to wrap up
 			// the contract.
-			c.log.Errorf("local commitment output was swept by "+
+			c.log.Warnf("local commitment output was swept by "+
 				"remote party via %v", sweepResult.Tx.TxHash())
 			recovered = false
 		case nil:
--- a/contractcourt/commit_sweep_resolver_test.go
+++ b/contractcourt/commit_sweep_resolver_test.go
@@ -96,6 +96,7 @@ func (i *commitSweepResolverTestContext) waitForResult() {
 type mockSweeper struct {
 	sweptInputs   chan input.Input
 	updatedInputs chan wire.OutPoint
+	sweepErr      error
 }
 
 func newMockSweeper() *mockSweeper {
@@ -112,7 +113,8 @@ func (s *mockSweeper) SweepInput(input input.Input, params sweep.Params) (
 
 	result := make(chan sweep.Result, 1)
 	result <- sweep.Result{
-		Tx: &wire.MsgTx{},
+		Tx:  &wire.MsgTx{},
+		Err: s.sweepErr,
 	}
 	return result, nil
 }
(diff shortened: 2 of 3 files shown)


lntest/itest/lnd_test.go
```diff
@@ -7441,6 +7441,10 @@ func testRevokedCloseRetribution(net *lntest.NetworkHarness, t *harnessTest) {
 	}
 
 	assertNodeNumChannels(t, carol, 0)
+
```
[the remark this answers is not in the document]
matheusd: Done. Added an extra commit with a unit test.

lntest/itest/lnd_test.go
```diff
@@ -7441,6 +7441,11 @@ func testRevokedCloseRetribution(net *lntest.NetworkHarness, t *harnessTest) {
...
+	// Mine enough blocks for Bob's channel arbitrator to wrap up the
+	// references to the breached channel.
+	mineBlocks(t, net, lntest.DefaultCSV-2, 0)
```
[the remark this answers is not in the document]
matheusd: Expanded the comment.

contractcourt/commit_sweep_resolver.go
```diff
@@ -219,15 +219,25 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
+			// it's likely what we sent was actually a revoked
+			// commitment. Report the error and continue to wrap up
+			// the contract.
+			c.log.Errorf("commit tx was swept by remote party via %v",
+				sweepResult.Tx.TxHash())
+		} else {
+			c.log.Infof("commit tx fully resolved by sweep tx: %v",
+				sweepResult.Tx.TxHash())
+
+		}
```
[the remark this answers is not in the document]
matheusd: Done.

matheusd:
> What is the current state of reporting of breaches by us or the remote party in the various overviews (pending channels etc) that exist?

Programatically? Almost completely inexistant. We mark closed channels as breached in the db and return that via RPC when the _remote_ party breached them, but this particular fix addresses a situation where the _local_ party breached the channel which isn't tracked at all anywhere. We also don't track any of the relevant breached balances as you note.

I added a unit test to contractcourt as well.

> And justice transactions can be applied to 2nd level htlc transactions too.

Should this be done in this PR?

contractcourt/commit_sweep_resolver.go
```diff
@@ -217,24 +217,40 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
 	select {
 	case sweepResult := <-resultChan:
-		if sweepResult.Err != nil {
+		switch sweepResult.Err {
+		case sweep.ErrRemoteSpend:
+			// If the remote party was able to sweep this output
+			// it's likely what we sent was actually a revoked
+			// commitment. Report the error and continue to wrap up
+			// the contract.
+			c.log.Errorf("commit tx was swept by remote party via %v",
```
joostjager, line 229: `commit tx`, it is pre-existing, but it is more accurate to say 'local commitment output` (also in log below).

contractcourt/commit_sweep_resolver.go
```diff
@@ -217,24 +217,40 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
 	// possible and publish the sweep tx. When the sweep tx
 	// confirms, it signals us through the result channel with the
 	// outcome. Wait for this to happen.
+	recovered := true
```
joostjager, line 220: usually we start the commit message with lower case after the colon

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -223,3 +225,87 @@ func TestCommitSweepResolverDelay(t *testing.T) {
...
+	}
+
+	ctx := newCommitSweepResolverTestContext(t, &res)
+
+	// Setup the sweeper so that it returns an error of the output being
+	// already swept.
+	ctx.sweeper.sweepErr = sweep.ErrRemoteSpend
+
+	report := ctx.resolver.report()
+	if !reflect.DeepEqual(report, &ContractReport{
```
joostjager, line 256: I think a normal `*report ==` works too, because there are no references?

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -223,3 +225,87 @@ func TestCommitSweepResolverDelay(t *testing.T) {
...
+	report = ctx.resolver.report()
+	if !reflect.DeepEqual(report, &ContractReport{
+		Outpoint:       outpoint,
+		Type:           ReportOutputUnencumbered,
+		Amount:         btcutil.Amount(amt),
+		MaturityHeight: testInitialBlockHeight + 2,
+
+		// RecoveredBalance is zero due to the output having been swept
+		// by the remote party.
+		RecoveredBalance: 0,
```
joostjager, line 307: Nice

contractcourt/commit_sweep_resolver_test.go
matheusd: Switched

matheusd:
Rebased to fix the conflict and fixed the review items. Once those are reviewed I'll squash them into their original commits.

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -273,14 +274,16 @@ func TestCommitSweepResolverLocalBreach(t *testing.T) {
 	// already swept.
 	ctx.sweeper.sweepErr = sweep.ErrRemoteSpend
 
-	report := ctx.resolver.report()
-	if !reflect.DeepEqual(report, &ContractReport{
+	gotReport := ctx.resolver.report()
+	wantReport := ContractReport{
```
joostjager, line 278: Mostly we would use `report` and `expectedReport`, if you want to stay in line with the usual naming (not a hard standard).

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -273,14 +274,16 @@ func TestCommitSweepResolverLocalBreach(t *testing.T) {
...
 		Outpoint:     outpoint,
 		Type:         ReportOutputUnencumbered,
 		Amount:       btcutil.Amount(amt),
 		LimboBalance: btcutil.Amount(amt),
-	}) {
-		t.Fatal("unexpected resolver report")
+	}
+	if *gotReport != wantReport {
+		t.Fatalf("unexpected resolver report. want=%+v got=%+v",
```
joostjager, line 285: Is the `+` needed?

joostjager:
PR title can be updated to match the `subsystem: message` format

contractcourt/commit_sweep_resolver_test.go
matheusd: I fel it looks better with the `+` but I'll remove it

joostjager: Oh, I now see that `+` prints field names. Yes, that is actually better indeed. I falsely assumed there was a `String()` method. Whatever you prefer.

matheusd:
Fixed the last items. Since you approved already, I went ahead and squashed the review fixes.

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -244,3 +246,92 @@ func TestCommitSweepResolverDelay(t *testing.T) {
 		t.Fatal("unexpected resolver report")
 	}
 }
+
+// TestCommitSweepResolverLocalBreach tests resolution when the local node
+// publishes a breached output (one that is swept by the remote party).
+func TestCommitSweepResolverLocalBreach(t *testing.T) {
```
bhandras, line 252: nb: you might want to consider making a table based sweep resolver test instead, to not duplicate most of `TestCommitSweepResolverDelay`.

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -244,3 +246,92 @@ func TestCommitSweepResolverDelay(t *testing.T) {
...
+	}
+
+	// Notify initial block height. The csv lock is still in effect, so we
+	// don't expect any sweep to happen yet.
+	ctx.notifyEpoch(testInitialBlockHeight)
+
+	select {
+	case <-ctx.sweeper.sweptInputs:
+		t.Fatal("no sweep expected")
+	case <-time.After(100 * time.Millisecond):
```
bhandras, line 311: maybe it's a good idea to also make these constant

contractcourt/commit_sweep_resolver_test.go
matheusd: Done on the new commit.

joostjager:
Our brand new itest error detector (see https://github.com/lightningnetwork/lnd/pull/4256) catches this new error:

`<time> [ERR] CNCT: *contractcourt.commitSweepResolver(<chan_point>): local commitment output was swept by remote party via <hex>`

As this is not an internal system error (although a little bit debatable), it may be better to downgrade to warning.

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -192,14 +192,20 @@ func TestCommitSweepResolverDelay(t *testing.T) {
...
-	if !reflect.DeepEqual(report, &ContractReport{
+	expectedReport := ContractReport{
 		Outpoint:     outpoint,
 		Type:         ReportOutputUnencumbered,
 		Amount:       btcutil.Amount(amt),
 		LimboBalance: btcutil.Amount(amt),
-	}) {
-		t.Fatal("unexpected resolver report")
+	}
+	if *report != expectedReport {
```
bhandras, line 206: I think it's safer to use `DeepEqual` here (even this is currently fine).

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -235,103 +241,52 @@ func TestCommitSweepResolverDelay(t *testing.T) {
...
-
-	report := ctx.resolver.report()
-	expectedReport := ContractReport{
-		Outpoint:     outpoint,
-		Type:         ReportOutputUnencumbered,
-		Amount:       btcutil.Amount(amt),
-		LimboBalance: btcutil.Amount(amt),
+		RecoveredBalance: expectedRecoveredBalance,
 	}
 	if *report != expectedReport {
```
bhandras, line 260: we should use `DeepEqual` here too.

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -1,7 +1,6 @@
 package contractcourt
```
bhandras, line 1: Nice, thanks for doing this!

contractcourt/commit_sweep_resolver_test.go
joostjager: I actually suggested to do the direct comparison. I think both work.

bhandras: I agree, in this case both work just fine. Perhaps we can keep direct comparison as it's much faster.

contractcourt/commit_sweep_resolver.go
```diff
@@ -239,24 +239,40 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
-		c.log.Infof("commit tx fully resolved by sweep tx: %v",
-			sweepResult.Tx.TxHash())
 	case <-c.quit:
 		return nil, errResolverShuttingDown
 	}
 
 	// Funds have been swept and balance is no longer in limbo.
 	c.reportLock.Lock()
-	c.currentReport.RecoveredBalance = c.currentReport.LimboBalance
+	if recovered {
```
Roasbeef, line 271: 👍

contractcourt/commit_sweep_resolver.go
```diff
@@ -239,24 +239,40 @@ func (c *commitSweepResolver) Resolve() (ContractResolver, error) {
...
 	select {
 	case sweepResult := <-resultChan:
-		if sweepResult.Err != nil {
+		switch sweepResult.Err {
+		case sweep.ErrRemoteSpend:
+			// If the remote party was able to sweep this output
+			// it's likely what we sent was actually a revoked
+			// commitment. Report the error and continue to wrap up
+			// the contract.
+			c.log.Errorf("local commitment output was swept by "+
```
Roasbeef, line 251: Should be demoted to `Warnf`. We're trying to make a change in the codebase that we only use `Errorf` for things that are actually "critical".

contractcourt/commit_sweep_resolver_test.go
Roasbeef: Just the deref is ok here since the struct only contains integers or fixed sized byte slices and has no pointer references.

contractcourt/commit_sweep_resolver_test.go
```diff
@@ -235,103 +241,52 @@ func TestCommitSweepResolverDelay(t *testing.T) {
...
-	// is 3, so a spend will be valid in the first block after height +1.
-	ctx.notifyEpoch(testInitialBlockHeight + 1)
-
-	<-ctx.sweeper.sweptInputs
+}
 
-	ctx.waitForResult()
+// TestCommitSweepResolverDelay tests resolution of a direct commitment output
+// that is encumbered by a time lock.
+func TestCommitSweepResolverDelay(t *testing.T) {
```
Roasbeef, line 269: 👍

contractcourt/commit_sweep_resolver.go
matheusd: Fixed in edcef88

cfromknecht:
@matheusd this looks ready for merge, but needs a rebase first. the build system has undergone some changes in the past week that should be run under these commits.

PR #3893, opened 2020-01-08
routing: ignore unknown required features in pathfinding

cfromknecht:
This PR brings us inline with recent modifications to the [spec](https://github.com/lightningnetwork/lightning-rfc/pull/723), that
say we shouldn't pay nodes whose feature vectors signal unknown required
features, and also that we shouldn't route through nodes signaling
unknown required features.

Currently we assert that invoices don't have such features during
decoding, but now that users can specify feature vectors via the rpc
interface, it makes sense to perform this check deeper in call stack.
This will also allow us to remove the check from decoding entirely,
making decodepayreq more useful for debugging.

We also fix one of the test vectors which had been incorrectly copied
from https://github.com/lightningnetwork/lightning-rfc/pull/719.
Feature bit 9 was missing from the current encoding, which is caught
now that we consider the decoding valid and actually perform the
comparison with the expected struct.

### DIFF
--- a/routing/pathfind.go
+++ b/routing/pathfind.go
@@ -456,8 +456,14 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 		}
 	}
 
-	// With the destination's feature vector selected, ensure that all
-	// transitive depdencies are set.
+	// Ensure that the destination's features don't include unknown
+	// required features.
+	err = feature.ValidateRequired(features)
+	if err != nil {
+		return nil, err
+	}
+
+	// Ensure that all transitive dependencies are set.
 	err = feature.ValidateDeps(features)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


routing/pathfind.go
```diff
@@ -752,11 +759,24 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
-				if err == nil {
-					fromFeatures = targetNode.Features
+				nodeFeatures := targetNode.Features
+
+				// Don't route through nodes that contain
+				// unknown required features.
+				req := nodeFeatures.UnknownRequiredFeatures()
+				if len(req) > 0 {
+					break
 				}
```
joostjager: Non blocking suggestion: create `feature.ValidateRequired` that returns an error describing the missing features. This can then be used in three locations where we currently check the length of the returned list.

routing/pathfind_test.go
```diff
@@ -1645,6 +1649,94 @@ func TestMissingFeatureDep(t *testing.T) {
...
+		target route.Vertex) ([]*channeldb.ChannelEdgePolicy, error) {
+
+		return findPath(
+			&graphParams{
+				graph: ctx.graphParams.graph,
+			},
+			r, testPathFindingConfig,
+			sourceNode.PubKeyBytes, target, 100, 0,
+		)
+	}
```
joostjager: I think this test can be written shorter by using `newPathFindingTestContext`

routing/pathfind.go
cfromknecht: i also thought this, and since you've brought it up i decided to implement. fixed

routing/pathfind_test.go
cfromknecht: indeed, much nicer

routing/pathfind_test.go
```diff
@@ -1645,6 +1649,73 @@ func TestMissingFeatureDep(t *testing.T) {
...
+				MinHTLC: 1,
+				MaxHTLC: 100000000,
+			}, 0,
+		),
+	}
+
+	ctx := newPathFindingTestContext(t, testChannels, "roasbeef")
+	defer ctx.cleanup()
+
+	conner := ctx.testGraphInstance.aliasMap["conner"]
```
joostjager: `ctx.keyFromAlias(...)`

PR #3894, opened 2020-01-08
build: bump version to 0.9.0-beta

joostjager:


joostjager:
Looking back in the history of `version.go`, I noticed that we never updated the version with the rc number. I did it now, because it seems helpful for preventing confusion about which release binaries are being used by users.

joostjager:
After discussion offline, reverted to not including the rc in the version reported by `lnd`

PR #3899, opened 2020-01-10
lncli: improve last_hop help message

cfromknecht:
Clarifies that last_hop is not the destination, but the penultimate node
in the path, i.e. the last intermediary.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -2087,8 +2087,9 @@ var (
 	}
 
 	lastHopFlag = cli.StringFlag{
-		Name:  "last_hop",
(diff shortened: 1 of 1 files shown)


cfromknecht:
> Should we go ahead and rename the flag and rpc field to incoming_node? Or won't it be clear that this refers to the incoming node of the destination? Incoming is more consistent with the outgoingChannel restriction.

`outgoing_chan_id` make sense because it always originates from the user's node. imo the only case `incoming_node` makes more sense than `last_hop` is when rebalancing, since the source and destination are the same. when constraining regular payments where `source != target`,  `incoming_node` is much more confusing because it's unclear what the reference point is.

PR #3900, opened 2020-01-10
lnrpc+lncli: add send support for custom data records

joostjager:
With the introduction of custom record sending and receiving, it became possible to attach arbitrary data to a payment. One obvious use case is attaching a human-readable message to a payment. Especially in the case of spontaneous key send payments, this can give the receiver some context on the payment.

For example: tipping. Usually people sending a tip would want to include some information on who they are or what the tip is for. For Lightning this may be even more desired than for other payment methods, because payments are anonymous by default. 

Attaching and retrieving the message was already possible before, but this PR makes it easier to do so by introducing an additional `lncli` flag `--data` to attach one or more custom records:

`lncli sendpayment -d 0274e7fb33eafd74fe1acb6db7680bb4aa78e9c839a6e954e38abfad680f645ef7 -a 100 --key_send --data 323442=00,3234556=ffff080812`

To specify a string value, the standard command line tool `xxd` can be used (the example record id here is the 3-byte ascii string 'tip' converted to an integer):

`--data 7629168=$(echo -n "Thank you!" | xxd -pu -c 10000)`

(The `-c` parameter is to prevent `xxd` from inserting line breaks)

Note: The available onion blob space of 1300 bytes is used for routing info and custom records. The bigger the size of the custom records, the fewer bytes remain for routing info and the shorter the maximum route length will be.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -2091,6 +2091,14 @@ var (
 		Usage: "pubkey of the last hop (penultimate node in the path) " +
 			"to route through for this payment",
 	}
+
+	dataFlag = cli.StringFlag{
+		Name: "data",
+		Usage: "attach custom data to the payment. The required " +
+			"format is: <record_id>=<hex_value>,<record_id>=" +
+			"<hex_value>,.. For example: --data 3438382=0a21ff. " +
+			"Custom record ids start from 65536.",
+	}
 )
 
 // paymentFlags returns common flags for sendpayment and payinvoice.
@@ -2127,6 +2135,7 @@ func paymentFlags() []cli.Flag {
 			Name:  "allow_self_payment",
 			Usage: "allow sending a circular payment to self",
 		},
+		dataFlag,
 	}
 }
 
@@ -2270,8 +2279,9 @@ func sendPayment(ctx *cli.Context) error {
 	}
 
 	req := &lnrpc.SendRequest{
-		Dest: destNode,
-		Amt:  amount,
+		Dest:              destNode,
+		Amt:               amount,
+		DestCustomRecords: make(map[uint64][]byte),
 	}
 
 	var rHash []byte
@@ -2286,9 +2296,10 @@ func sendPayment(ctx *cli.Context) error {
 			return err
 		}
 
-		req.DestCustomRecords = map[uint64][]byte{
-			record.KeySendType: preimage[:],
-		}
+		// Set the preimage. If the user supplied a preimage with the
+		// data flag, the preimage that is set here will be overwritten
+		// later.
+		req.DestCustomRecords[record.KeySendType] = preimage[:]
 
 		hash := preimage.Hash()
 		rHash = hash[:]
@@ -2354,6 +2365,33 @@ func sendPaymentRequest(ctx *cli.Context, req *lnrpc.SendRequest) error {
 	req.CltvLimit = uint32(ctx.Int(cltvLimitFlag.Name))
 
 	req.AllowSelfPayment = ctx.Bool("allow_self_payment")
+
+	// Parse custom data records.
+	data := ctx.String(dataFlag.Name)
+	if data != "" {
+		records := strings.Split(data, ",")
+		for _, r := range records {
+			kv := strings.Split(r, "=")
+			if len(kv) != 2 {
+				return errors.New("invalid data format: " +
+					"multiple equal signs in record")
+			}
+
+			recordID, err := strconv.ParseUint(kv[0], 10, 64)
+			if err != nil {
+				return fmt.Errorf("invalid data format: %v",
+					err)
+			}
+
+			hexValue, err := hex.DecodeString(kv[1])
+			if err != nil {
+				return fmt.Errorf("invalid data format: %v",
+					err)
+			}
+
+			req.DestCustomRecords[recordID] = hexValue
+		}
+	}
 
 	amt := req.Amt
 
@@ -2434,8 +2472,9 @@ func payInvoice(ctx *cli.Context) error {
 	}
 
 	req := &lnrpc.SendRequest{
-		PaymentRequest: payReq,
-		Amt:            ctx.Int64("amt"),
+		PaymentRequest:    payReq,
+		Amt:               ctx.Int64("amt"),
+		DestCustomRecords: make(map[uint64][]byte),
 	}
 
 	return sendPaymentRequest(ctx, req)


joostjager:
Message record id updated to a different random number.

My thought was that if we don't pick a field and provide a few hooks to make it easy to use (string flag on lncli and string type invoice htlc field), payment messages are less likely to be used. The payment message functionality can serve as a demonstrator of everything else that is possible using custom records. it seemed nice to me to get people thinking about that.

But as @Roasbeef suggests, we can also generalize this to just a new `lncli` flag. But the question there is what a good format is. Any thoughts on that?

`lncli sendpayment --data=34378273:"hello world!",124343:0feab32827cba23`

It gets complicated with the spaces and special chars. Also not sure what different shells do there. In my bash, at least the data argument comes through as one block, but the quotes have disappeared.

joostjager:
Removed hard-coded record and replaced by generic `--data` flag.

cmd/lncli/commands.go
```diff
@@ -2286,9 +2295,7 @@ func sendPayment(ctx *cli.Context) error {
 			return err
 		}
 
-		req.DestCustomRecords = map[uint64][]byte{
-			record.KeySendType: preimage[:],
-		}
+		req.DestCustomRecords[record.KeySendType] = preimage[:]
```
cfromknecht: What should we do if this overwrites a user supplied preimage? It may be confusing if users supplies one preimage and it isn’t used.

We could change to only generating one if this field isn’t set or simply fail.

The bigger question in my mind is if we want to allow user supplied preimages altogether. The former may resolve some complications in tracking keysend payments, perhaps it is okay to permit it given we already allow arbitrary preimages for hodl invoices. Thoughts?

cmd/lncli/commands.go
```diff
@@ -2355,6 +2362,30 @@ func sendPaymentRequest(ctx *cli.Context, req *lnrpc.SendRequest) error {
...
+		records := strings.Split(data, ",")
+		for _, r := range records {
+			kv := strings.Split(r, "=")
+			if len(kv) != 2 {
+				return errors.New("invalid data format")
+			}
+
+			recordID, err := strconv.ParseUint(kv[0], 10, 64)
+			if err != nil {
+				return errors.New("invalid data format")
```
cfromknecht: error messages could be more descriptive, and describe why it’s invalid

cmd/lncli/commands.go
```diff
@@ -2091,6 +2091,13 @@ var (
 		Usage: "pubkey of the last hop (penultimate node in the path) " +
 			"to route through for this payment",
 	}
+
+	dataFlag = cli.StringFlag{
+		Name: "data",
+		Usage: "attach custom data to the payment. The required " +
+			"format is: <record_id>=<hex_value>,<record_id>=" +
+			"<hex_value>,.. For example: --data 3438382=0a21ff",
```
cfromknecht: also mention start of custom range?

cmd/lncli/commands.go
joostjager: I think it is okay for this experimental release of key send. Not only was it already possible to choose a preimage for hodl invoices, but one can even do this for normal invoices via `lncli addinvoice --preimage`. 

When we unify key send with moon amp, it seems like a good time to me to rethink where the preimage is generated.

joostjager: The current implementation lets a user specified preimage override the generated one. Added comment to clarify that.

cmd/lncli/commands.go
joostjager: fixed

cmd/lncli/commands.go
joostjager: added

curly60e:
> With the introduction of custom record sending and receiving, it became possible to attach arbitrary data to a payment. One obvious use case is attaching a human-readable message to a payment. Especially in the case of spontaneous key send payments, this can give the receiver some context on the payment.
> 
> For example: tipping. Usually people sending a tip would want to include some information on who they are or what the tip is for. For Lightning this may be even more desired than for other payment methods, because payments are anonymous by default.
> 
> Attaching and retrieving the message was already possible before, but this PR makes it easier to do so by introducing an additional `lncli` flag `--data` to attach one or more custom records:
> 
> `lncli sendpayment -d 0274e7fb33eafd74fe1acb6db7680bb4aa78e9c839a6e954e38abfad680f645ef7 -a 100 --key_send --data 323442=00,3234556=ffff080812`
> 
> To specify a string value, the standard command line tool `xxd` can be used (the example record id here is the 3-byte ascii string 'tip' converted to an integer):
> 
> `--data 7629168=$(echo -n "Thank you!" | xxd -pu -c 10000)`
> 
> (The `-c` parameter is to prevent `xxd` from inserting line breaks)
> 
> Note: The available onion blob space of 1300 bytes is used for routing info and custom records. The bigger the size of the custom records, the fewer bytes remain for routing info and the shorter the maximum route length will...

This is the only method to send a message over lightning?

There is a short command method like a standard?

cfromknecht:
Currently yes, there is no standardized way of sending messages apart from these custom data records. Note that this require sender and receiver to be aware of the record type and format. There isn't any BOLT standard on how to do so, though some folks have implemented their own protocols on top of this primitive. Personally I wouldn't expect to see a messaging standard integrated into LND any time soon given now nascent these higher-level specifications are.

drmartinberger:
Dear lnd developers @Roasbeef @cfromknecht,

I am investigating custom data records attached to a Lightning Payment as should be available in lnd with this PR #3900. Thank you for that!

@joostjager was mentioning:
> Note: The available onion blob space of 1300 bytes is used for routing info and custom records. The bigger the size of the custom records, the fewer bytes remain for routing info and the shorter the maximum route length will be.

Do you have an idea or experience, how many bytes can be consumed by the custom_records field and still getting a routeable onion payment? Do you have an estimate `x` if one wants to route a payment with custom data record via `y` hops, there usually remain `1300/y*x` bytes for the data?

Can you point me to any sources to learn more on that? Any other way to transport encrypted custom data with a lightning payment, maybe compatible with other lightning implementations?

Thank you!

Cheers, Martin

bitromortac:
Perhaps you could constrain your route length to about the diameter of the network, `lncli getnetworkinfo` shows me a value of 12, or a bit less, which should be plenty. Here [some comment](https://github.com/lightningnetwork/lightning-onion/blob/b62f49fcadfc013d81c639e196a7f09b428a4dc1/path.go#L307) about the maximal number of hops in an onion.

drmartinberger:
> Perhaps you could constrain your route length to about the diameter of the network, `lncli getnetworkinfo` shows me a value of 12, or a bit less, which should be plenty. Here [some comment](https://github.com/lightningnetwork/lightning-onion/blob/b62f49fcadfc013d81c639e196a7f09b428a4dc1/path.go#L307) about the maximal number of hops in an onion.

Thanks @bitromortac  for your hint. If I understand [your mentioned comment](https://github.com/lightningnetwork/lightning-onion/blob/b62f49fcadfc013d81c639e196a7f09b428a4dc1/path.go#L307) correctly, in a route with `x` non-exist hops there should remain at most `1300-37-x*65` bytes for a custom data record (assuming larger legacy hop payloads), right?

bitromortac:
Right, but you can assume a base data size of 47 bytes as legacy onions will be phased out soon. If no other data is transported in there this should be the limit.

ZoltanAB:
If I make a payment using lncli payinvoice ... --data 65536=$myuniquehex, is there any way to (later) easily find all payments that matches that hex?
Listing all payments (lncli listpayments) and manually comparing the hex is not the ideal solution when you might have hundreds of thousands of payments but only 5-10 has $myuniquehex. Any advise?

PR #3902, opened 2020-01-10
Fix zlib empty short channel ids

t-bast:
I noticed a weird issue with lnd's handling of `query_short_channel_ids` without any `short_channel_id`.
It looks like `var buf bytes.Buffer` initializes a buffer of size 11.
So if we don't have anything to add to it, we'll call zlib on that buffer containing 11 bytes `0` which isn't the output we expect.

It's a bit unclear in the spec if we should omit the `encoded_short_ids` field entirely in that case: that may be cleaner than a single byte representing the encoding followed by nothing.
I need to check what c-lightning does in that case (Eclair does what I implemented here, but it feels a bit weird).

### DIFF
--- a/lnwire/query_short_chan_ids.go
+++ b/lnwire/query_short_chan_ids.go
@@ -147,6 +147,9 @@ func decodeShortChanIDs(r io.Reader) (ShortChanIDEncoding, []ShortChannelID, err
 	// Before continuing, we'll snip off the first byte of the query body
 	// as that was just the encoding type.
 	queryBody = queryBody[1:]
+	if len(queryBody) == 0 {
+		return encodingType, nil, nil
+	}
 
 	// Otherwise, depending on the encoding type, we'll decode the encode
 	// short channel ID's in a different manner.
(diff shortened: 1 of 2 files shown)


t-bast:
Thinking about it some more, we shouldn't even be sending a `query_short_channel_ids` if we don't have any `short_channel_id` to query for :confused:

t-bast:
Actually this isn't where I want to be fixing that; there's no reason to send a `query_short_channel_ids` without any `short_channel_id`.
The bug is in the code that handles an empty list in `reply_channel_range`.

Roasbeef:
> The bug is in the code that handles an empty list in reply_channel_range.

If there's a bug it would be when we're _decoding_ the message rather than handling it, which is located here: https://github.com/lightningnetwork/lnd/blob/master/lnwire/query_short_chan_ids.go#L121. If the response is empty, then it should only be a single byte, which may cause this line to exit with an error as there's nothing more to read: https://github.com/lightningnetwork/lnd/blob/master/lnwire/query_short_chan_ids.go#L138.

PR #3903, opened 2020-01-10
invoices: streamline key send cltv delta picking and validation

joostjager:
This PR:
* Does a pre-check for key send htlc expiry values to prevent empty invoices filling up the database.
* Increases the default final cltv delta that payment RPCs use so that a receiver running `lnd` won't reject it. This allows key send payments via `lncli` without specifying the `final_cltv_delta` flag.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3535,7 +3535,11 @@ func (r *rpcServer) extractPaymentIntent(rpcPayReq *rpcPaymentRequest) (rpcPayme
 	if rpcPayReq.FinalCltvDelta != 0 {
 		payIntent.cltvDelta = uint16(rpcPayReq.FinalCltvDelta)
 	} else {
-		payIntent.cltvDelta = zpay32.DefaultFinalCLTVDelta
+		// If no final cltv delta is given, assume the default that we
+		// use when creating an invoice. We do not assume the default of
+		// 9 blocks that is defined in BOLT-11, because this is never
+		// enough for other lnd nodes.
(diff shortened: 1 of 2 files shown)


rpcserver.go
```diff
@@ -3535,7 +3535,11 @@ func (r *rpcServer) extractPaymentIntent(rpcPayReq *rpcPaymentRequest) (rpcPayme
 	if rpcPayReq.FinalCltvDelta != 0 {
 		payIntent.cltvDelta = uint16(rpcPayReq.FinalCltvDelta)
 	} else {
-		payIntent.cltvDelta = zpay32.DefaultFinalCLTVDelta
+		// If no final cltv delta is given, assume the default that we
+		// use when creating an invoice. We do not assume the default of
+		// 9 blocks that is defined in BOLT-11, because this is never
+		// enough for other lnd nodes.
+		payIntent.cltvDelta = uint16(cfg.Bitcoin.TimeLockDelta)
```
Roasbeef: Hmm, I see the motivation here, but I'm not sure about this, as then distinct version of `lnd`, or different implementations may end up displaying different values for the final CLTV delta for a given invoice.

joostjager: This doesn't affect invoice creation. It is just the final cltv delta used for a payment if no value is specified.

Roasbeef: Sure, it's a display thing (as in us displaying a decoded invoice vs CL displaying the same one). In the end, it's just cosmetic though.

joostjager: I still don't understand your comment. This line of code is only hit when there is no payment request to decode. Which exact display of the final cltv delta do you mean?

PR #3906, opened 2020-01-13
Fix channel range queries with empty results

t-bast:
The spec is a bit unclear on how to encode empty `encoded_short_ids`.
It can either be done with `len` set to `0000` or `len` set to `0001` followed by only the encoding `00` or `01` (this is the current behavior in lnd, eclair and c-lightning).

I'm fixing the following issues in this PR:

* Decoding an empty list with the zlib encoding doesn't work ("unexpected EOF")
* Encoding an empty list with zlib produces unexpected bytes because an empty `bytes.Buffer` is in fact initialized with 11 0-value bytes
* Decoding a `len` set to `0000` should work: no need to force the sender to provide an encoding when nothing is actually encoded :thinking:

### DIFF
--- a/lnwire/query_short_chan_ids.go
+++ b/lnwire/query_short_chan_ids.go
@@ -132,7 +132,7 @@ func decodeShortChanIDs(r io.Reader) (ShortChanIDEncoding, []ShortChannelID, err
 	}
 
 	if numBytesResp == 0 {
-		return 0, nil, fmt.Errorf("No encoding type specified")
+		return 0, nil, nil
 	}
 
 	queryBody := make([]byte, numBytesResp)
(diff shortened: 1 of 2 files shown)


t-bast:
It looks like c-lightning doesn't support setting `len` to `0000` (and not providing any encoding).
I'll try to ping them to have their view on whether that should be supported or not.
Meanwhile I can revert that part of the PR and only keep the part that fixes the `zlib` error?

Roasbeef:
> Encoding an empty list with zlib produces unexpected bytes because an empty bytes.Buffer is in fact initialized with 11 0-value bytes

This isn't correct, an empty `bytes.Buffer` has no backing buffer (`var b bytes.Buffer`). You may be seeing something else here related to how the `zlib` library encodes an empty data set.

Roasbeef:
I ended up going with a slightly different approach, and committed it as: 17200afc57cec82313a55b066483027f8feaeaaa

Thanks for getting the ball rolling on this fix!

t-bast:
LGTM, thanks for the quick turnaround!

PR #3907, opened 2020-01-13
[mobile] Set walletunlocker TLS credentials

### LINKED ISSUE #3905: [mobile] grpc: no transport security set
### Background

I have launched LND on iOS via `gomobile` and all your instructions (using falafel). I am able to start LND node and see all logs. But I cannot make any gRPC requests because of the following error:

`Error Domain=go Code=1 "grpc: no transport security set (use grpc.WithInsecure() explicitly or set credentials)" UserInfo={NSLocalizedDescription=grpc: no transport security set (use grpc.WithInsecure() explicitly or set credentials)}`

### Your environment

Also please note that I am using new `nio` branch of `grpc-swift` as they recommend.

* version of `lnd`:  lnd v0.9.0-beta-rc1
* which operating system (`uname -a` on *Nix): OSX
* version of `btcd`, `bitcoind`, or other backend: Neutrino

### Steps to reproduce

`LndmobileStart` succeeds and I see all the logs. But any gRPC request (`LndmobileInitWallet` or `LndmobileGenSeed`) does fail with the error above.

halseth:
This PR fixes a regression from #3783, where TLS support was added on mobile, but we forgot to add the TLS credentials also to the wallet unlocker client.

Fixes #3905

### DIFF
--- a/lnd.go
+++ b/lnd.go
@@ -60,6 +60,25 @@ var (
 	networkDir string
 )
 
+// WalletUnlockerAuthOptions returns a list of DialOptions that can be used to
+// authenticate with the wallet unlocker service.
+//
+// NOTE: This should only be called after the WalletUnlocker listener has
+// signaled it is ready.
+func WalletUnlockerAuthOptions() ([]grpc.DialOption, error) {
+	creds, err := credentials.NewClientTLSFromFile(cfg.TLSCertPath, "")
+	if err != nil {
+		return nil, fmt.Errorf("unable to read TLS cert: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


lnd.go
```diff
@@ -60,6 +60,25 @@ var (
 	networkDir string
 )
 
+// WalletUnlockerAuthOptions returns a list of DialOptions that can be used to
+// authenticate with the wallet unlocker service.
+//
+// NOTE: This should only be called after the WalletUnlocker listerner has
```
guggero: nit: s/listerner/listener/

halseth: Fixed

PR #3910, opened 2020-01-14
lnwallet: limit received htlc's to MaxAcceptedHTLCs

Crypt-iQ:
This PR clamps down the number of received HTLCs to `MaxAcceptedHTLCs`.

### DIFF
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -3084,12 +3084,16 @@ func (lc *LightningChannel) validateCommitmentSanity(theirLogCounter,
 	view := lc.fetchHTLCView(theirLogCounter, ourLogCounter)
 
 	// If we are checking if we can add a new HTLC, we add this to the
-	// update log, in order to validate the sanity of the commitment
-	// resulting from _actually adding_ this HTLC to the state.
+	// appropriate update log, in order to validate the sanity of the
+	// commitment resulting from _actually adding_ this HTLC to the state.
 	if predictAdded != nil {
-		// If we are adding an HTLC, this will be an Add to the local
-		// update log.
-		view.ourUpdates = append(view.ourUpdates, predictAdded)
+		// If the remoteChain bool is true, add to ourUpdates.
+		if remoteChain {
+			view.ourUpdates = append(view.ourUpdates, predictAdded)
+		} else {
+			// Else add to theirUpdates.
+			view.theirUpdates = append(view.theirUpdates, predictAdded)
+		}
 	}
 
 	commitChain := lc.localCommitChain
@@ -4638,6 +4642,17 @@ func (lc *LightningChannel) ReceiveHTLC(htlc *lnwire.UpdateAddHTLC) (uint64, err
 		LogIndex:  lc.remoteUpdateLog.logIndex,
 		HtlcIndex: lc.remoteUpdateLog.htlcCounter,
 		OnionBlob: htlc.OnionBlob[:],
+	}
+
+	localACKedIndex := lc.remoteCommitChain.tail().ourMessageIndex
+
+	// Clamp down on the number of HTLC's we can receive by checking the
+	// commitment sanity.
+	err := lc.validateCommitmentSanity(
+		lc.remoteUpdateLog.logIndex, localACKedIndex, false, pd,
+	)
+	if err != nil {
+		return 0, err
 	}
 
 	lc.remoteUpdateLog.appendHtlc(pd)
--- a/lnwallet/channel_test.go
+++ b/lnwallet/channel_test.go
@@ -5462,22 +5462,25 @@ func TestMaxAcceptedHTLCs(t *testing.T) {
 	defer cleanUp()
 
 	// One over the maximum number of HTLCs that either can accept.
-	const numHTLCs = 20
-	const numHTLCsReceived = 12
+	const numHTLCs = 12
 
-	// Set the remote's required MaxAcceptedHtlcs. This means that alice
+	// Set the remote's required MaxAcceptedHtlcs. This means that Alice
 	// can only offer the remote up to numHTLCs HTLCs.
 	aliceChannel.channelState.LocalChanCfg.MaxAcceptedHtlcs = numHTLCs
 	bobChannel.channelState.RemoteChanCfg.MaxAcceptedHtlcs = numHTLCs
 
 	// Similarly, set the remote config's MaxAcceptedHtlcs. This means
-	// that the remote will be aware that Alice will only accept up to
-	// numHTLCsRecevied at a time.
-	aliceChannel.channelState.RemoteChanCfg.MaxAcceptedHtlcs = numHTLCsReceived
-	bobChannel.channelState.LocalChanCfg.MaxAcceptedHtlcs = numHTLCsReceived
+	// that the remote will be aware that Bob will only accept up to
+	// numHTLCs at a time.
+	aliceChannel.channelState.RemoteChanCfg.MaxAcceptedHtlcs = numHTLCs
+	bobChannel.channelState.LocalChanCfg.MaxAcceptedHtlcs = numHTLCs
 
 	// Each HTLC amount is 0.1 BTC.
 	htlcAmt := lnwire.NewMSatFromSatoshis(0.1 * btcutil.SatoshiPerBitcoin)
+
+	// htlcID is used to keep track of the HTLC that Bob will fail back to
+	// Alice.
+	var htlcID uint64
 
 	// Send the maximum allowed number of HTLCs.
 	for i := 0; i < numHTLCs; i++ {
@@ -5488,6 +5491,13 @@ func TestMaxAcceptedHTLCs(t *testing.T) {
 		if _, err := bobChannel.ReceiveHTLC(htlc); err != nil {
 			t.Fatalf("unable to recv htlc: %v", err)
 		}
+
+		// Just assign htlcID to the last received HTLC.
+		htlcID = htlc.ID
+	}
+
+	if err := ForceStateTransition(aliceChannel, bobChannel); err != nil {
+		t.Fatalf("unable to transition state: %v", err)
 	}
 
 	// The next HTLC should fail with ErrMaxHTLCNumber.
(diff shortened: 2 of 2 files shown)


lnwallet/channel.go
```diff
@@ -4331,6 +4331,14 @@ func (lc *LightningChannel) ReceiveHTLC(htlc *lnwire.UpdateAddHTLC) (uint64, err
 			"ID %d", htlc.ID, lc.remoteUpdateLog.htlcCounter)
 	}
 
+	// Clamp down on the number of HTLC's we can receive. It should never
+	// exceed our configurable MaxAcceptedHTLCs.
+	maxHTLCs := lc.channelState.LocalChanCfg.MaxAcceptedHtlcs
+	remoteACKedIndex := lc.localCommitChain.tail().theirHtlcIndex
```
halseth: This will be the index of the last acked htlc,  not the first one active, which results in checking the number of unACKed HTLCs, which I don't think is what we want?

What we want to check is the number of active HTLCs, which will include both ACKed and unACKed. I think we must check the update logs or the commitments to see which HTLCs are actually active.

Crypt-iQ: Ahhh you're right

Crypt-iQ:
PTAL @halseth

halseth:
Travis failing?

Crypt-iQ:
> Travis failing?

timeout flake with TestMppPayment I think

Crypt-iQ:
@halseth Updated the calculation to start from the tip rather than tail since any new adds will be on top of that. Also abstracted the calculation to a function + updated the unit test.

halseth:
> @halseth Updated the calculation to start from the tip rather than tail since any new adds will be on top of that. Also moved the calculation to a function + updated the unit test.

Yes, makes sense! Looks good 👍

halseth:
There might be some unexpected implication for the itests, looks like they are all failing the same test now.

lntest/itest/lnd_test.go
```diff
@@ -10145,7 +10145,7 @@ func testBidirectionalAsyncPayments(net *lntest.NetworkHarness, t *harnessTest)
 	}
 
 	// Calculate the number of invoices.
-	numInvoices := int(info.LocalBalance / paymentAmt)
+	numInvoices := input.MaxHTLCNumber / 2
```
halseth: Do we know why this didn't fail before?

halseth: Since the new check is on the receive side that could mean we are not properly checking sender side we are not overflowing the commitment.

Crypt-iQ: You're right - the last commit is not the correct fix.  The sender is overflowing the commitment, but there is a check there. I suspect an off-by-one error somewhere...

lnwallet/channel.go
```diff
@@ -6028,3 +6038,107 @@ func (lc *LightningChannel) RemoteCommitHeight() uint64 {
...
+		if remoteTail.height > commitHeight {
+			// Check if the HTLC exists on the remote's commitment tx.
+			if _, ok := indexMap[pd.HtlcIndex]; !ok {
+				// If it doesn't exist, decrement numCommitted.
+				numCommitted--
+			}
+		}
+	}
+
+	// First grab the htlc of the tip to determine how many HTLC's have been
```
cfromknecht: nit: "First" looks like it's left over from prior iterations

lnwallet/channel.go
```diff
@@ -6028,3 +6038,107 @@ func (lc *LightningChannel) RemoteCommitHeight() uint64 {
...
+	// sides are sending signatures at the same time. In step 3, the signature
+	// does not cover the recent settle/fail that Bob sent in step 1. Because
+	// of this, the add Alice sends to Bob in step 6 could overflow Bob's
+	// commitment transaction depending on our calculation. If we just count
+	// the number of committed HTLC's incoming to Bob and the number of in-mem
+	// HTLC's, then the add in step 6 will overflow and Bob will fail the
+	// channel. If, however, we adjust the committed HTLC's count to ignore
+	// HTLC's which will be removed in the next signature that Alice sends,
+	// then the add in step 6 won't overflow. This is because the signature
+	// received in step 7 will cover the settle/fail in step 1 and makes space
```
cfromknecht: excellent comments, very helpful!

lnwallet/channel.go
```diff
@@ -4331,6 +4331,16 @@ func (lc *LightningChannel) ReceiveHTLC(htlc *lnwire.UpdateAddHTLC) (uint64, err
 			"ID %d", htlc.ID, lc.remoteUpdateLog.htlcCounter)
 	}
 
+	// Clamp down on the number of HTLC's we can receive. It should never
+	// exceed our configurable MaxAcceptedHTLCs. The total number of HTLC's is
+	// the combination of uncommitted and committed HTLC's.
+	maxHTLCs := lc.channelState.LocalChanCfg.MaxAcceptedHtlcs
+	numHTLCs := lc.calcIncomingHtlcs(htlc.ID)
```
joostjager: Is this smart calculation now also applied on the side of the sender of the HTLC?

lnwallet/channel.go
```diff
@@ -6028,3 +6038,107 @@ func (lc *LightningChannel) RemoteCommitHeight() uint64 {
...
+
+// calcIncomingHtlcs calculates the number of committed and uncommitted HTLC's
+// from the PoV of the tip of the commitment chain. This function is used to
+// determine if one of the counterparty's adds will overflow the commitment
+// transaction before it happens in ReceiveNewCommitment. Since these Adds will
+// be appended to the tip of the chain, the tail is not used. It takes an
+// htlcID parameter that is used to determine the number of uncommitted HTLC's
+// in-memory.
+func (lc *LightningChannel) calcIncomingHtlcs(htlcID uint64) uint64 {
+	// There are two cases to consider when calculating the true number of
```
halseth: This is starting to become quite complex. Could we instead call `validateCommitmentSanity`, which I believe already performs these checks?

Crypt-iQ: Yup, cuts down the whole diff pretty much

lnwallet/channel.go
Crypt-iQ: It was, but turns out I was just mimicking `validateCommitmentSanity`

lnwallet/channel.go
Crypt-iQ: I updated the code to just use `validateCommitmentSanity` since it was doing the same thing. I didn't feel the need to add a unit test for the specific case outlined above since it's not a new function like `calcIncomingHtlcs`. Curious what you think.

lnwallet/channel.go
```diff
@@ -2788,7 +2788,7 @@ func (lc *LightningChannel) createCommitDiff(
 // in the state when adding a new HTLC, or nil otherwise.
 func (lc *LightningChannel) validateCommitmentSanity(theirLogCounter,
 	ourLogCounter uint64, remoteChain bool,
-	predictAdded *PaymentDescriptor) error {
+	predictAdded, predictTheirs *PaymentDescriptor) error {
```
cfromknecht: the godoc should be updated to describe this new argument. alternatively, seems like we can just pass it in via `predictAdded` and then add it to the appropriate log based on `remoteChain`

lnwallet/channel.go
```diff
@@ -4351,6 +4347,17 @@ func (lc *LightningChannel) ReceiveHTLC(htlc *lnwire.UpdateAddHTLC) (uint64, err
 		OnionBlob: htlc.OnionBlob[:],
 	}
 
+	localACKedIndex := lc.remoteCommitChain.tail().ourMessageIndex
+
+	// Clamp down on the number of HTLC's we can receive by checking the
+	// commitment sanity.
+	err := lc.validateCommitmentSanity(
```
halseth: yeah, nice I think this is it. If I'm not mistaken this will also handle the `CASE 2` you outlined earlier, since it will apply the settles/fails accordingly?

lnwallet/channel.go
halseth: Possibly we would need this extra argument, to allow the case where we are predicting adding `our` add to the `remote` commitment chain, but would need to look at the callsites for the method.

lnwallet/channel.go
Crypt-iQ: Yup it applies the same logic that the sender applies so it handles case 2

lnwallet/channel.go
Crypt-iQ: Yes can do this, if `remoteChain` is true, add to our updates. Otherwise, add to theirs.

lnwallet/channel_test.go
```diff
@@ -5443,6 +5443,158 @@ func TestMaxAcceptedHTLCs(t *testing.T) {
...
 
+// TestMaxAsynchronousHtlcs tests that Bob correctly receives (and does not
+// fail) an HTLC from Alice when exchanging asynchronous payments. We want to
+// mimic the following case where Bob's commitment transaction is full before
+// starting:
+// 	Alice                    Bob
+// 1.         <---settle/fail---
+// 2.         <-------sig-------
+// 3.	       --------sig------> (covers an add sent before step 1)
+// 4. 	       <-------rev-------
```
cfromknecht: nit: this line and line above are indented improperly

lnwallet/channel.go
```diff
@@ -4341,6 +4345,17 @@ func (lc *LightningChannel) ReceiveHTLC(htlc *lnwire.UpdateAddHTLC) (uint64, err
...
```
joostjager: Doesn't this early validation violate the protocol? For example if we'd receive and settle an htlc in the same batch with no more commitment slot left? If validating only for the whole batch, that would still be possible perhaps?

Crypt-iQ: I'm not entirely sure I follow - the prior behavior was to do no validation and wait until a commitment was received to do the validation. This just makes the validation earlier so there shouldn't be a problem. It also mimics lnd's sender-side logic (to validate upon calling `AddHTLC`) so if this is a protocol violation, then that logic would be too.

joostjager: Yes, the question is indeed if it is a protocol violation. If the commitment is full, should we allow another htlc to be added? Because before the signature comes, a settle may still be added.

Crypt-iQ: So there are two places in the spec I can see where this is elaborated on:

```
if result would be offering more than the remote's max_accepted_htlcs HTLCs, in the remote commitment transaction:

    MUST NOT add an HTLC.
```

and 

```
if a sending node adds more than receiver max_accepted_htlcs HTLCs to its local commitment transaction [...]:

    SHOULD fail the channel.
```

So they only seem concerned about the effect on the commitment transaction. So I think this won't violate the spec since `ReceiveHTLC` just goes into the update logs. Plus it's only applied to the commit tx because there is a settle/fail that ensures only the max can be applies to the commit tx.

lnwallet/channel_test.go
Crypt-iQ: squashed

lnwallet/channel.go
joostjager: It still feels a little inconclusive. But as the second quote is from the description of the add message, I indeed think that validation should take place already then and not be postponed until the signature arrives.

Crypt-iQ: Is this worth bringing up / making an issue in the spec repo? I don't see the receive logic being the problem, but rather the sending logic. If we think a send in this async case is ok but our peer thinks it's a violation, then I'd see a problem as we may be disconnected. All in all though, the receive should match the send logic.

halseth: Worth bringing up as a spec issue I think, although I think it is pretty obvious that each received message would have to lead to a valid commitment transaction. Otherwise how many updates would you allow while waiting fo a settel to open up the commitment. But agree that this could be more clearly spelled out in the spec.

halseth:
Noticed travis was failing, but unrelated to this change: https://github.com/lightningnetwork/lnd/pull/4010

lnwallet/channel_test.go
```diff
@@ -5562,6 +5562,158 @@ func TestMaxAcceptedHTLCs(t *testing.T) {
...
+// 5.         --------rev------>
+// 6.         --------add------>
+// 7.         - - - - sig - - ->
+// This represents an asynchronous commitment dance in which both sides are
+// sending signatures at the same time. In step 3, the signature does not
+// cover the recent settle/fail that Bob sent in step 1. However, the add that
+// Alice sends to Bob in step 6 does not overflow Bob's commitment transaction.
+// This is because validateCommitmentSanity counts the HTLC's by ignoring
+// HTLC's which will be removed in the next signature that Alice sends. Thus,
+// the add won't overflow. This is because the signature received in step 7
```
halseth: Won't this contradict with the conclusion in https://github.com/lightningnetwork/lnd/pull/3910#pullrequestreview-359213860 ?

Roasbeef: In the end, we can only really execute these heuristics on a best effort basis. There're other classes of the concurrent send case that aren't covered in this test/scenario. At the end of the day, the real defense point is at the _sender_, in that they should be careful to not send items that would result in a force close.

PR #3911, opened 2020-01-14
routing+lnrpc: add missing query routes parameters

### LINKED ISSUE #3150: queryroutes uses unrouteable final_cltv_delta value by default

joostjager:
Over time, several new parameters have been added to the payment RPCs. But `QueryRoutes` hasn't kept up with those. This PR adds what is missing:

* Route hints
* Last hop restriction
* Outgoing channel restriction
* Destination node features

Furthermore an inconsistency with the default final cltv delta is fixed.

Fixes #3150


C-Otto:
@alexbosworth https://github.com/lightningnetwork/lnd/pull/3926

routing/router.go
```diff
@@ -1402,14 +1402,8 @@ type routingMsg struct {
 func (r *ChannelRouter) FindRoute(source, target route.Vertex,
 	amt lnwire.MilliSatoshi, restrictions *RestrictParams,
```
Tryiimaggie, line 1403: Auburn at the end of the 4th

Tryiimaggie, line 1403: 1+4+0+3=8🩵™️

Tryiimaggie (approved):

PR #3914, opened 2020-01-15
docs: fix code_contribution_guidelines.md wrong git hashes for gRPC tools

hsjoberg:
In Section 4.7 of [docs/code_contribution_guidelines.md](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#Protobuf), the git hashes for the gRPC tools does not match what is said in [lnrpc/README.md](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/README.md#generate-protobuf-definitions), thus might produce erroneous files.

This PR fixes this, however, it might make sense to just remove these rows and only link to the lnrpc README.

### DIFF
--- a/docs/code_contribution_guidelines.md
+++ b/docs/code_contribution_guidelines.md
@@ -454,8 +454,8 @@ of `protoc`.
 
 The following commit hashes of related projects are also required in order to
 generate identical compiled protos and related files:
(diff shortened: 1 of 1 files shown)


guggero:
Thanks for the PR! These commit hashes are indeed incorrect in the guidelines.
I would prefer the suggested link to `lnrpc/README.md` though. Then we don't forget it again the next time we update the dependencies.

guggero:
Superseded by #4250.

Issue #3771, opened 2019-11-27
htlcswitch: reject self-loop HTLC forwarding attempts 


Roasbeef:
Circular routes in the protocol have many use cases as is today, such as rebalancing ones channels. However, they can also be used for nuisance activities such as saturating a channel with HTLCs. In the future, there're plans to extend the protocol to make these nuisance activities no longer costless. However, until these changes are rolled out, there're a number of low-hanging mitigations we can put in place. One such mitigation is: rejecting self-loop HTLC forwarding attempts. I refer to self-loops as instances wherein a nodes receives a an HTLC with payment hash `W` over a link, then forwards out the HTLC `W` over the outgoing link, only to receive that same HTLC `W` over another link. This link may even be the very _same_ link that was the target outgoing/incoming link. See [this paper for a formal treatment of this precise attack](https://eprint.iacr.org/2019/1149.pdf). 

# Steps To Completion 

- [ ] The [`CheckHtlcForward`](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L2119) method should be updated to reject _all_ HTLCs which are already present on the commitment transaction of the target link. This prevents us from cooperating with an attempt self-loop.

carlaKC:
I've been working in this part of the code base recently so I'd like to pick this up. 
Just wondering what failure message we'd send back to the peer on the wire for this kind of check? `TemporaryChannelFailure`?

From my understanding, there are two mitigations that we could implement, given nodes A and B:
1. Prevent loops on the same link (as described in steps to completion, can be done in `CheckHTLCForward`, seems straightforward)
```
+----+-----------------+----+
|  A |   HTLC(W) --->  |  B | 
|    |  <---- HTLC(w)  |    |
+----+-----------------+----+
```
2. Prevent loops that are incoming on one channel link and outgoing on another, probably a bit more complex, and would need to be handled on switch level.
```
+----+-----------------+----+
|  A |   HTLC(W) --->  |  B | 
|    |-----------------|    |
|    |  <---- HTLC(w)  |    |
+----+-----------------+----+
```

Thinking of going ahead with (1) for now and then we can consider (2)?

I can't think of many uses for looping HTLCs back to yourself in the same channel in the wild, but I think that the solution for (1) should be configurable just in case somebody has a use case out there? Enabled by default, with the option to disable in case issues arise?

alexbosworth:
I use this for sloshing

carlaKC:
[...]

PR #3915, opened 2020-01-15
htlcswitch: Disallow circular routes on same channel

carlaKC:
This PR adds some protection against circular liquidity attacks, as described in #3771. 

Two types of protection against circular routes are identified:
1. Same channel circles: disallow htlcs with the same incoming and outgoing channel ID

Update: after discussion, decided to only implement (1), which simplifies the PR.
~2. Pair channel circles: disallow forwarding of htlcs over a pair of channels when a htlc has already forwarded a htlc with that hash in the opposite direction~

Context for not implementing (2):
1. Effect on base MPP: base MPP uses duplicate htlcs, which are indistinguishable from circular routes (in almost all cases, except for payments which reverse over the same channel pair)
2. Pair channel effectiveness: pair channel protection is probably quite easy to get around, so not worth implementing

### DIFF
--- a/config.go
+++ b/config.go
@@ -341,6 +341,8 @@ type config struct {
 	Watchtower *lncfg.Watchtower `group:"watchtower" namespace:"watchtower"`
 
 	LegacyProtocol *lncfg.LegacyProtocol `group:"legacyprotocol" namespace:"legacyprotocol"`
+
+	AllowCircularRoute bool `long:"allow-circular-route" description:"If true, our node will allow htlc forwards that arrive and depart on the same channel."`
 }
 
 // loadConfig initializes and parses the config using a config file and command
--- a/htlcswitch/link_test.go
+++ b/htlcswitch/link_test.go
@@ -39,6 +39,7 @@ import (
 
 const (
 	testStartingHeight = 100
+	testDefaultDelta   = 6
 )
 
 // concurrentTester is a thread-safe wrapper around the Fatalf method of a
--- a/htlcswitch/switch.go
+++ b/htlcswitch/switch.go
@@ -167,6 +167,10 @@ type Config struct {
 	// fails in forwarding packages.
 	AckEventTicker ticker.Ticker
 
+	// AllowCircularRoute is true if the user has configured their node to
+	// allow forwards that arrive and depart our node over the same channel.
+	AllowCircularRoute bool
+
 	// RejectHTLC is a flag that instructs the htlcswitch to reject any
 	// HTLCs that are not from the source hop.
 	RejectHTLC bool
@@ -986,6 +990,22 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 			return s.handleLocalDispatch(packet)
 		}
 
+		// Before we attempt to find a non-strict forwarding path for
+		// this htlc, check whether the htlc is being routed over the
+		// same incoming and outgoing channel. If our node does not
+		// allow forwards of this nature, we fail the htlc early. This
+		// check is in place to disallow inefficiently routed htlcs from
+		// locking up our balance.
+		linkErr := checkCircularForward(
+			packet.incomingChanID, packet.outgoingChanID,
+			s.cfg.AllowCircularRoute, htlc.PaymentHash,
+		)
+		if linkErr != nil {
+			return s.failAddPacket(
+				packet, linkErr.WireMessage(), linkErr,
+			)
+		}
+
 		s.indexMtx.RLock()
 		targetLink, err := s.getLinkByShortID(packet.outgoingChanID)
 		if err != nil {
@@ -1168,6 +1188,37 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 	default:
 		return errors.New("wrong update type")
 	}
+}
+
+// checkCircularForward checks whether a forward is circular (arrives and
+// departs on the same link) and returns a link error if the switch is
+// configured to disallow this behaviour.
+func checkCircularForward(incoming, outgoing lnwire.ShortChannelID,
+	allowCircular bool, paymentHash lntypes.Hash) *LinkError {
+
+	// If the route is not circular we do not need to perform any further
+	// checks.
+	if incoming != outgoing {
+		return nil
+	}
+
+	// If the incoming and outgoing link are equal, the htlc is part of a
+	// circular route which may be used to lock up our liquidity. If the
+	// switch is configured to allow circular routes, log that we are
+	// allowing the route then return nil.
+	if allowCircular {
+		log.Debugf("allowing circular route over link: %v "+
+			"(payment hash: %x)", incoming, paymentHash)
+		return nil
+	}
+
+	// If our node disallows circular routes, return a temporary channel
+	// failure. There is nothing wrong with the policy used by the remote
+	// node, so we do not include a channel update.
+	return NewDetailedLinkError(
+		lnwire.NewTemporaryChannelFailure(nil),
+		FailureDetailCircularRoute,
+	)
 }
 
 // failAddPacket encrypts a fail packet back to an add packet's source.
--- a/htlcswitch/switch_test.go
+++ b/htlcswitch/switch_test.go
@@ -32,7 +32,9 @@ func genPreimage() ([32]byte, error) {
 func TestSwitchAddDuplicateLink(t *testing.T) {
 	t.Parallel()
 
-	alicePeer, err := newMockServer(t, "alice", testStartingHeight, nil, 6)
+	alicePeer, err := newMockServer(
+		t, "alice", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create alice server: %v", err)
 	}
@@ -90,7 +92,9 @@ func TestSwitchAddDuplicateLink(t *testing.T) {
 func TestSwitchHasActiveLink(t *testing.T) {
 	t.Parallel()
 
-	alicePeer, err := newMockServer(t, "alice", testStartingHeight, nil, 6)
+	alicePeer, err := newMockServer(
+		t, "alice", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create alice server: %v", err)
 	}
@@ -158,7 +162,9 @@ func TestSwitchHasActiveLink(t *testing.T) {
 func TestSwitchSendPending(t *testing.T) {
 	t.Parallel()
 
-	alicePeer, err := newMockServer(t, "alice", testStartingHeight, nil, 6)
+	alicePeer, err := newMockServer(
+		t, "alice", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create alice server: %v", err)
 	}
@@ -253,11 +259,15 @@ func TestSwitchSendPending(t *testing.T) {
 func TestSwitchForward(t *testing.T) {
 	t.Parallel()
 
-	alicePeer, err := newMockServer(t, "alice", testStartingHeight, nil, 6)
+	alicePeer, err := newMockServer(
+		t, "alice", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create alice server: %v", err)
 	}
-	bobPeer, err := newMockServer(t, "bob", testStartingHeight, nil, 6)
+	bobPeer, err := newMockServer(
+		t, "bob", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create bob server: %v", err)
 	}
@@ -358,11 +368,15 @@ func TestSwitchForwardFailAfterFullAdd(t *testing.T) {
 
 	chanID1, chanID2, aliceChanID, bobChanID := genIDs()
 
-	alicePeer, err := newMockServer(t, "alice", testStartingHeight, nil, 6)
+	alicePeer, err := newMockServer(
+		t, "alice", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create alice server: %v", err)
 	}
-	bobPeer, err := newMockServer(t, "bob", testStartingHeight, nil, 6)
+	bobPeer, err := newMockServer(
+		t, "bob", testStartingHeight, nil, testDefaultDelta,
+	)
 	if err != nil {
 		t.Fatalf("unable to create bob server: %v", err)
 	}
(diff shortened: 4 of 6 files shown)


carlaKC:
Quite simple change if we only disallow circles in the same channel. 
Started working on an itest for this as well, but unsure if it's worth adding (in terms of adding more for travis to do).

config.go
```diff
@@ -341,6 +341,8 @@ type config struct {
 	Watchtower *lncfg.Watchtower `group:"watchtower" namespace:"watchtower"`
 
 	LegacyProtocol *lncfg.LegacyProtocol `group:"legacyprotocol" namespace:"legacyprotocol"`
+
+	AllowCircularRoute bool `long:"allow-circular" description:"If true, our node will allow htlc forwards that arrive and depart on the same channel."`
```
joostjager: Match var name and flag name? Is there a use case for allowing this btw?

htlcswitch/link.go
```diff
@@ -2144,8 +2144,8 @@ func (l *channelLink) UpdateForwardingPolicy(newPolicy ForwardingPolicy) {
 // NOTE: Part of the ChannelLink interface.
 func (l *channelLink) CheckHtlcForward(payHash [32]byte,
 	incomingHtlcAmt, amtToForward lnwire.MilliSatoshi,
-	incomingTimeout, outgoingTimeout uint32,
-	heightNow uint32) *LinkError {
+	incomingTimeout, outgoingTimeout, heightNow uint32,
+	incomingLink lnwire.ShortChannelID, allowCircular bool) *LinkError {
```
joostjager: Maybe it is better to get this static config bool from the link configuration, rather than passing it every time.

joostjager: But I actually think that this check is something that should be done in the switch, when the forwarding decision is made.

carlaKC: I thought about putting it in the switch (in `handlepacketforward`) but then it's less easy to unit test. It's quite a simple change though so can move it in there as well, which saves having to get that bool down to the link?

config.go
carlaKC: I think that keysend will replace the need for this eventually, but I'm worried about this change breaking no-invoice apps which tip by dropping off fees (could be the case that they drop off over the same channel for single channel nodes) in the meantime.

carlaKC:
Moved check into the switch and added an itest. 

The downside of moving this check into the switch is that it exposes the switch to link level information in the form of channelupdate/temporary failure. Instead of adding a `ChannelUpdate` function to `ChannelLink` interface, I could add `TemporaryFailure` which returns the correct error?

htlcswitch/switch.go
```diff
@@ -1075,6 +1079,23 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
 			return s.failAddPacket(packet, linkErr.WireMessage(), addErr)
 		}
 
+		// We have found an outgoing link with a forwarding policy that
+		// the htlc satisfies. Before we proceed with forwarding, check
+		// whether the htlc will produce a circular forward through the
+		// same channel, and fail it if the switch does not allow
+		// forwards of this nature.
+		if linkErr := s.CheckCircularForward(
```
joostjager: A separate statement to call `CheckCircularForward` rather than embedding it in the if condition is probably more readable.

htlcswitch/switch.go
```diff
@@ -1170,6 +1191,47 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+	// If our node disallows circular routes, get a channel update for
+	// the outgoing link and return a temporary channel failure. If we
+	// fail getting the channel update, we return a temporary node failure.
+	update, err := outgoing.ChannelUpdate()
+	if err != nil {
+		return NewLinkError(&lnwire.FailTemporaryNodeFailure{})
+	}
+
+	return NewDetailedLinkError(
+		lnwire.NewTemporaryChannelFailure(update),
```
joostjager: I don't think we need to return an update. There isn't a problem with the policy.

lntest/itest/lnd_test.go
```diff
@@ -5060,6 +5063,104 @@ func testMultiHopSendToRoute(net *lntest.NetworkHarness, t *harnessTest) {
 	assertAmountPaid(t, "Alice(local) => Bob(remote)", net.Alice,
 		aliceFundPoint, amountPaid+(baseFee*numPayments), int64(0))
 
+	// Finally, we want to test allowing and disallowing of circular
+	// payments. Bob's node disallows circular payments and carol's node
+	// allows them. We will send payments over circular routes through
+	// each of these nodes.
```
joostjager: Wouldn't it be better to test this in a unit test like `TestSkipIneligibleLinksMultiHopForward`? To me it looks like there is nothing that an itest can exercise that can't be exercised in a lower level test.

htlcswitch/switch.go
joostjager: I am wondering how this check works together with the previous non-strict forwarding decision. Could it be that the same in and out chan was requested, but that the out channel for some reason (concurrent payment perhaps?) doesn't have enough balance and another channel to the same node is selected?

htlcswitch/switch.go
carlaKC: Nil update is fine because the spec has a length field, right? It doesn't specifically say whether we have to include an update. 

Also double checking, `TemporaryChannelUpdate` is the correct error to send?
Seems like the most fitting one in the bolts: "The channel from the processing node was unable to handle this HTLC, but may be able to handle it, or others, later."

htlcswitch/switch.go
carlaKC: I think that it's fair to treat these payments as 'suspicious' because they are taking an inefficient route. While there may be some custom use cases out there, node operators who are expecting to use them should configure their node to allow circular routes. 

If somebody tries to route a payment over the same channel (and we do not allow that), we could just catch that early and not apply non-strict forwarding? Which would just mean moving the check up in `handlePacketForward` and failing early.

joostjager: Yes moving it up sounds good. Or maybe basing the check on a pubkey comparison, maybe that is even better.

carlaKC: Disallowing based on pubkey would mean that people can't do 2 hop circular rebalances, which I think would cause some issues for existing nodes if their lnd peers suddenly stopped letting them rebalance.

joostjager: Two hop circular rebalances are problematic anyway, because of non-strict forwarding. You may just as well get back your money through the same channel.

Comparing on the exact chan ids is also a risk for peers that have multiple channels between them. The ping pong route can still be executed. I think it is more important to prevent that than to allow the two hop rebalance which isn't reliable anyway.

htlcswitch/switch.go
```diff
@@ -1170,6 +1194,40 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+// CheckCircularForward checks whether a forward is circular (arrives and
+// departs on the same link) and returns a link error if the switch is
+// configured to disallow this behaviour.
+func CheckCircularForward(incoming, outgoing lnwire.ShortChannelID,
+	allowCircular bool, paymentHash lntypes.Hash) *LinkError {
+
+	// Check that the incoming link is different to the outgoing link.
+	// If they are equal, the htlc is part of a circular route which may
+	// be used to lock up our liquidity.
+	isCircularRoute := incoming.ToUint64() == outgoing.ToUint64()
```
bhandras: I think the `ToUint64()` is not required here as these structs are comparable and comparison should actually be faster too.

htlcswitch/switch.go
```diff
@@ -1170,6 +1194,40 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+
+	// If the route is not circular we do not need to perform any further
+	// checks.
+	if !isCircularRoute {
```
bhandras: init: Maybe further simplify this to `if incoming != outgoing`.

htlcswitch/switch.go
```diff
@@ -1170,6 +1194,40 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+		return nil
+	}
+
+	// If the switch is configured to allow circular routes, log that we
+	// are allowing the route then return nil.
+	if allowCircular {
+		log.Info("allowing circular route over link: %v "+
```
bhandras: Do we want to info log each time this happens? Maybe a debug log is enough.

htlcswitch/switch.go
```diff
@@ -1170,6 +1194,40 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+			"(payment hash: %x)", paymentHash, incoming)
```
bhandras: `paymentHash` and `incoming` are switched in the format string.

htlcswitch/switch_test.go
```diff
@@ -1294,6 +1296,178 @@ type multiHopFwdTest struct {
...
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			alicePeer, err := newMockServer(
+				t, "alice", testStartingHeight, nil, 6,
```
bhandras: I know this is a pattern in this file, but it really bothers me that 6 is not a constant :)

htlcswitch/switch_test.go
```diff
@@ -1294,6 +1296,178 @@ type multiHopFwdTest struct {
...
+				s, chanID1, aliceChanID, alicePeer, true,
+			)
+
+			if err := s.AddLink(aliceChannelLink); err != nil {
+				t.Fatalf("unable to add alice link: %v", err)
+			}
+
+			// Create a new packet that loops through alice's link
+			// in a circle.
+			preimage := [sha256.Size]byte{1}
```
bhandras: As `preimage` and `rhash` are constants for all sub tests, you may consider moving them out from the loop.

htlcswitch/switch_test.go
```diff
@@ -1294,6 +1296,178 @@ type multiHopFwdTest struct {
...
+			failure := obfuscator.(*mockObfuscator).failure
+			if !reflect.DeepEqual(failure, test.expectedFailure) {
+				t.Fatalf("expected: %v, got: %v",
+					test.expectedFailure, failure)
+			}
+
+			// If the error is non-nil, check that the error
+			// contains the error string for a circular forward.
+			if err != nil {
+				if !strings.Contains(err.Error(),
```
bhandras: nit: Could be just one if.

htlcswitch/switch_test.go
```diff
@@ -1294,6 +1296,178 @@ type multiHopFwdTest struct {
...
+					test.expectedErrorString) {
+					t.Fatalf("expected: %v, got:"+
+						" %v", test.expectedErrorString,
+						err.Error())
+				}
+			}
+
+			// Ensure that no circuits were opened.
+			if s.circuits.NumOpen() != 0 {
```
bhandras: nitnitnit :nerd_face: `if s.curcuits.NumOpen() > 0`

htlcswitch/switch.go
carlaKC: It is unreliable but I think a lot of people rely on it at the moment. Would say that we'd probably disable by default if we disallow to the same peer, so that we don't break a lot of people's setups. 

I think that single channel is still a valuable defense to have. Without it, an attacker can lock up funds in every channel on your node by zig zagging back and fourth through each channel. The takes `capacity/(max htlc size*19)` payments to lock up your balance (assuming one hop where they lock up their own balance to route the first hop to your node. I think that to one peer would need wider discussion, @alexbosworth has an operational requirement for that ability, for example.

htlcswitch/switch_test.go
carlaKC: nit and thy shall receive! (a separate commit declaring a constant for these tests 😄 )

Thanks for pointing it out, it's always nice to do a little clean up while you're around. I think it would be cool to switch over to functional options for some or these test structs with a sensible default so that exactly what you're spinning up is more clear/controlled. But that's a separate discussion/PR.

htlcswitch/switch.go
carlaKC: Wasn't expecting it to happen too often, since it's the case where somebody has changed their config and there's a strange route going through their node, but downgraded to debug. Would imagine that operators who are using these special case routes are running debug logs anyway.

htlcswitch/switch.go
joostjager: Yes, single channel is better than nothing. I have no data myself on how many people rely on two hop rebalancing. Detecting peer-level ping-pong can be implemented later if there are doubts.

One way to lock up the channel is to use up all its capacity, but another way is to use up all commitment tx slots. There is a max number of htlcs that the tx can hold. That way it costs only negligible money to do the attack.

htlcswitch/link_test.go
```diff
@@ -39,6 +39,7 @@ import (
 
 const (
 	testStartingHeight = 100
+	testDefaultDelta   = 6
```
joostjager: For commits that only change test code, we often use `htlcswitch/test` as the commit message prefix

htlcswitch/switch.go
```diff
@@ -66,6 +66,12 @@ var (
 	// ErrUnreadableFailureMessage is returned when the failure message
 	// cannot be decrypted.
 	ErrUnreadableFailureMessage = errors.New("unreadable failure message")
+
+	// circularRouteErrTemplate is the string used to return errors to the
+	// router when htlcs fail due to circular routes. It is declared here
+	// to allow for exact error matching in tests.
+	circularRouteErrTemplate = "incoming HTLC(%x) disallowed due to " +
+		"circular forward over: %v "
```
joostjager: Downside of formatting is that it is no longer possible to check for this error via `err == circularRouteErr`

htlcswitch/switch.go
```diff
@@ -986,6 +996,25 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+		// locking up our balance.
+		linkErr := checkCircularForward(
+			packet.incomingChanID, packet.outgoingChanID,
+			s.cfg.AllowCircularRoute, htlc.PaymentHash,
+		)
+		if linkErr != nil {
+			addErr := fmt.Errorf(circularRouteErrTemplate,
+				htlc.PaymentHash[:], packet.outgoingChanID)
+
+			return s.failAddPacket(
```
joostjager: I am wondering, what is the purpose of returning addErr here (via `failAddPacket`) and not just `linkErr`? `FailureDetail` already describes what is the problem.

carlaKC: The `addrErr` provides the specific channel ID that the circle was attempted on. I've actually just removed this in #3844 though, I think we can get by without it. If we find ourselves wanting specific link errors, we can just make new `FailureDetail` types that have additional metadata.

htlcswitch/switch.go
carlaKC: I guess this is a more general question of how specific we want the errors we surface to the user to be. If we want to include metadata then I think this is an improvement on unit tests where we just copy the error string that is returned inline. If we're happy with more generic user errors, then the failure detail's error string can just be returned.

htlcswitch/switch.go
carlaKC: Going to kill `addrErr` here and just pass the `linkErr` in twice, since the change to remove `addErr` from `failAddPacket` is already in 3844.

htlcswitch/switch.go
joostjager: Yes, I think that is ok. It doesn't matter much which error we return, because the sender must know what they are doing.

htlcswitch/switch.go
```diff
@@ -1170,6 +1190,37 @@ func (s *Switch) handlePacketForward(packet *htlcPacket) error {
...
+	if incoming != outgoing {
+		return nil
+	}
+
+	// If the incoming and outgoing link are equal, the htlc is part of a
+	// circular route which may be used to lock up our liquidity. If the
+	// switch is configured to allow circular routes, log that we are
+	// allowing the route then return nil.
+	if allowCircular {
+		log.Debugf("allowing circular route over link: %v "+
```
cfromknecht: nit: capitalize log message

PR #3918, opened 2020-01-15
routing: remove path finding shortcut

joostjager:
This shortcut does not work when the destination is a private node. We also don't have this shortcut for regular payments. This commit aligns the behavior between SendPayment and QueryRoutes.

### DIFF
--- a/routing/errors.go
+++ b/routing/errors.go
@@ -7,14 +7,9 @@ import "github.com/go-errors/errors"
 type errorCode uint8
 
 const (
-	// ErrTargetNotInNetwork is returned when the target of a path-finding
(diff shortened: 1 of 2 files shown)


joostjager:
I thought about that, but the node itself is not in the route hints. It would then just come down to skipping the check if there are route hints. I thought that removing it for the sake of consistency is a better solution. This is probably also not the right place to execute this check. I would be better in `findPath`. But then it influences regular payments and maybe we introduce a new bug this late in the cycle.

joostjager:
If we would re-implement this, we can also consider to check whether at least one route hint has a starting point that is in the graph. But that is also a little fuzzy.

PR #3922, opened 2020-01-17
multi: rename key_send, key-send and key send to keysend

cfromknecht:
Converts multiple API level references of `key_send` or `key-send`, or `key send` to `keysend`.


cfromknecht:
> There are some variables remaining though that are named keySend or KeySend instead of keysend and Keysend.

@joostjager I had another commit that modifies the internal naming, but left it out bc it has no observable effect on 0.9

PR #3924, opened 2020-01-17
itest: use fast scrypt options for aezeed, macaroons DB and wallet DB

guggero:
This PR modifies the `aezeed`, `macaroons` and `lnwallet/btcwallet` packages to use fast scrypt options when compiled with the itest build tag to speed up the node creation/startup in the integration tests.

The faster wallet unlock also brought to light some flakes in the itests that have been addressed.

### DIFF
--- a/aezeed/cipherseed_rpctest.go
+++ b/aezeed/cipherseed_rpctest.go
@@ -1,0 +1,13 @@
+// +build rpctest
+
+package aezeed
+
+import "github.com/btcsuite/btcwallet/waddrmgr"
+
+func init() {
+	// For the purposes of our itest, we'll crank down the scrypt params a
+	// bit.
+	scryptN = waddrmgr.FastScryptOptions.N
+	scryptR = waddrmgr.FastScryptOptions.R
+	scryptP = waddrmgr.FastScryptOptions.P
+}
--- a/htlcswitch/link.go
+++ b/htlcswitch/link.go
@@ -627,7 +627,7 @@ func (l *channelLink) syncChanStates() error {
 
 	if err := l.cfg.Peer.SendMessage(true, localChanSyncMsg); err != nil {
 		return fmt.Errorf("Unable to send chan sync message for "+
-			"ChannelPoint(%v)", l.channel.ChannelPoint())
+			"ChannelPoint(%v): %v", l.channel.ChannelPoint(), err)
 	}
 
 	var msgsToReSend []lnwire.Message
@@ -3121,7 +3121,7 @@ func (l *channelLink) fail(linkErr LinkFailureError,
 		return
 	}
 
-	l.log.Errorf("failing link: %s", reason)
+	l.log.Errorf("failing link: %s with error: %v", reason, linkErr)
 
 	// Set failed, such that we won't process any more updates, and notify
 	// the peer about the failure.
--- a/lntest/itest/lnd_test.go
+++ b/lntest/itest/lnd_test.go
@@ -7328,10 +7328,12 @@ func testRevokedCloseRetribution(net *lntest.NetworkHarness, t *harnessTest) {
 
 	// Carol will be the breached party. We set --nolisten to ensure Bob
 	// won't be able to connect to her and trigger the channel data
-	// protection logic automatically.
+	// protection logic automatically. We also can't have Carol
+	// automatically re-connect too early, otherwise DLP would be initiated
+	// instead of the breach we want to provoke.
 	carol, err := net.NewNode(
 		"Carol",
-		[]string{"--hodl.exit-settle", "--nolisten"},
+		[]string{"--hodl.exit-settle", "--nolisten", "--minbackoff=1h"},
 	)
 	if err != nil {
 		t.Fatalf("unable to create new carol node: %v", err)
@@ -7591,10 +7593,12 @@ func testRevokedCloseRetributionZeroValueRemoteOutput(net *lntest.NetworkHarness
 
 	// Dave will be the breached party. We set --nolisten to ensure Carol
 	// won't be able to connect to him and trigger the channel data
-	// protection logic automatically.
+	// protection logic automatically. We also can't have Dave automatically
+	// re-connect too early, otherwise DLP would be initiated instead of the
+	// breach we want to provoke.
 	dave, err := net.NewNode(
 		"Dave",
-		[]string{"--hodl.exit-settle", "--nolisten"},
+		[]string{"--hodl.exit-settle", "--nolisten", "--minbackoff=1h"},
 	)
 	if err != nil {
 		t.Fatalf("unable to create new node: %v", err)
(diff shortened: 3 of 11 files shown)


guggero:
For some reason the itest `TestLightningNetworkDaemon/revoked_uncooperative_close_retribution` always fails with these changes but I couldn't figure out why.
Other than copying the channel DB (which is not even encrypted) this test doesn't do anything special as far as I can see.

chainregistry.go
```diff
@@ -261,6 +261,7 @@ func newChainControlFromConfig(cfg *config, chanDB *channeldb.DB,
 		walletConfig.ChainSource = chain.NewNeutrinoClient(
 			activeNetParams.Params, neutrinoCS,
 		)
+		time.Sleep(1 * time.Second)
```
Roasbeef: IIRC we deduced the precise reason why we need this?

wpaulino: Based on the logs @guggero sent to me, I was able to find https://github.com/lightninglabs/neutrino/pull/200. This solved a flake for him, but I believe there's another one we're still investigating.

guggero: The neutrino fix definitely removed the need for the `time.Sleep`.
There's still a flake in the SCB restore:
```
        --- FAIL: TestLightningNetworkDaemon/channel_backup_restore/restore_during_unlock (41.49s)
            lnd_test.go:106: Error outside of test: *errors.errorString unable to find Carol's force close tx in mempool: wanted 1, found 0 txs in mempool: []
                /home/guggero/projects/go/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:8555 (0xdc9db3)
                        assertDLPExecuted: t.Fatalf("unable to find Carol's force close tx in mempool: %v",
                /home/guggero/projects/go/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:14116 (0xdec0aa)
                        testChanRestoreScenario: assertDLPExecuted(
                /home/guggero/projects/go/src/github.com/lightningnetwork/lnd/lntest/itest/lnd_test.go:14437 (0xe08532)
                        testChannelBackupRestore.func7: testChanRestoreScenario(h, net, &testCase, password)
                /usr/lib/go-1.13.1/src/testing/testing.go:909 (0x50ac19)
                        tRunner: fn(t)
                /usr/lib/go-1.13.1/src/runtime/asm_amd64.s:1357 (0x460e81)
                        goexit: BYTE    $0x90   // NOP

```
It does also happen with `btcd` although much less often than with neutrino.
Should we continue with this PR without a fix for it and move that to a follow-up PR?

halseth:
what's the estimated speedup for a itest suite run?

guggero:
I added the wait after the Neutrino wallet creation in 8b370cbd to fix the remaining flake, hopefully. I'll run the Neutrino itest a few times to make sure.

The alternative to 8b370cbd would be btcsuite/btcwallet#68 but that would lead to other people having problems with old/stale simnet chains according to @wpaulino.
I'm fine with either fixes, not sure how to prioritize between them.

joostjager:
We can definitely use the speed up as we expand the test matrix with the anchor commitment format

joostjager:
This may even be a hard requirement for the anchor itests, because the suite now exceeds that travis max of 50 minutes

aezeed/cipherseed_rpctest.go
```diff
@@ -0,0 +1,13 @@
+// +build rpctest
```
wpaulino: Any reason to gate the unit tests with this flag as well? It'd be nice to have them when just running `make unit` as well.

lntest/itest/lnd_test.go
```diff
@@ -8598,9 +8602,18 @@ func assertDLPExecuted(net *lntest.NetworkHarness, t *harnessTest,
 	carol *lntest.HarnessNode, carolStartingBalance int64,
 	dave *lntest.HarnessNode, daveStartingBalance int64) {
 
+	// We disabled auto-reconnect for some tests to avoid timing issues.
+	// To make sure the nodes are initiating DLP now, we have to manually
+	// re-connect them.
+	ctxb := context.Background()
+	err := net.ConnectNodes(ctxb, carol, dave)
+	if err != nil && !strings.Contains(err.Error(), "already connected") {
```
wpaulino: If they're supposed to be disconnected before this point, wouldn't it be an error if they already were?

htlcswitch/link.go
```diff
@@ -627,7 +627,7 @@ func (l *channelLink) syncChanStates() error {
 
 	if err := l.cfg.Peer.SendMessage(true, localChanSyncMsg); err != nil {
 		return fmt.Errorf("Unable to send chan sync message for "+
-			"ChannelPoint(%v)", l.channel.ChannelPoint())
+			"ChannelPoint(%v): %v", l.channel.ChannelPoint(), err)
```
wpaulino: :+1:

lntest/timeouts.go
```diff
@@ -23,5 +23,5 @@ const (
 
 	// AsyncBenchmarkTimeout is the timeout used when running the async
 	// payments benchmark.
-	AsyncBenchmarkTimeout = time.Minute
+	AsyncBenchmarkTimeout = 2 * time.Minute
```
wpaulino: Was this increase needed for travis or when running this test locally?

chainregistry.go
```diff
@@ -262,6 +263,22 @@ func newChainControlFromConfig(cfg *config, chanDB *channeldb.DB,
...
+		// with the recovery. This can lead to the rescan missing some
+		// of the addresses. By just waiting half a second we can fix
+		// this flake in our integration tests. This cannot happen on
+		// testnet or mainnet because there we always wait for the
+		// wallet to be fully synced before we even start the recovery.
+		network := activeNetParams.Params.Net
+		if network == chaincfg.RegressionNetParams.Net ||
+			network == chaincfg.SimNetParams.Net {
+
+			time.Sleep(500 * time.Millisecond)
```
wpaulino: I'd prefer not having this sleep in production code, I wonder if we could have this be somewhere in the integration tests instead? Might not be possible though given that there isn't a hook between initializing the neutrino node and the wallet.

aezeed/cipherseed_rpctest.go
guggero: Yes, the unit test already has the fast options. The thing is, for the itest, the `lnd-itest` binary is built with `go build` and not `go test`. So this code needs to be in a non-test file but guarded with the build flag to only be picked up in itests.
But I added the options to the unit tests in the macaroons package, that shaves another 300ms off of the tests there.

lntest/itest/lnd_test.go
guggero: The tests relied on the auto-reconnect feature which was wrong IMO as it depends on the timing of the node startup and the `--minbackoff` setting. Now with the min backoff set very high (which is equal to disabling reconnect), we have to do an explicit reconnect which is what we want.

lntest/timeouts.go
guggero: This was for me locally, not sure if it also happens on Travis. I think this was just too short in any case, we should be able to run the tests on any system, especially locally.

chainregistry.go
guggero: Yeah, I don't really like that either. I tried adding it only to the itest but as you say, there is no hook for it. That's why I created https://github.com/btcsuite/btcwallet/pull/681 in the first place, to fix the problem instead of a workaround. Now that it got merged, I think it's better to use the fix instead of the wait. Switched commits.

lntest/itest/lnd_test.go
wpaulino: Gotcha, so now they should never reconnect unless done explicitly, but wouldn't ignoring the already connected error imply that they somehow did though?

guggero: A bunch of tests use this function so I didn't want to break them in case they implemented a manual reconnect for whatever reason.

PR #3927, opened 2020-01-17
keychain+walletunlocker: use fast scrypt options for unit tests

guggero:
With this PR we crank down the scrypt options used by the `btcwallet` to speed up unit tests, especially when running with the race detector.

Results on my machine **before** the change (run with `go clean -testcache && go test ./keychain -race && go test ./walletunlocker -race`):
```
ok      github.com/lightningnetwork/lnd/keychain        150.092s
ok      github.com/lightningnetwork/lnd/walletunlocker  123.689s
```
**After** the change:
```
ok      github.com/lightningnetwork/lnd/keychain        3.834s
ok      github.com/lightningnetwork/lnd/walletunlocker  6.728s
```

### DIFF
--- a/keychain/interface_test.go
+++ b/keychain/interface_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcwallet/snacl"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/btcsuite/btcwallet/walletdb"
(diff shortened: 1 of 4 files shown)


keychain/interface_test.go
```diff
@@ -38,6 +39,13 @@ var (
 		0x4f, 0x2f, 0x6f, 0x25, 0x98, 0xa3, 0xef, 0xb9,
 		0x69, 0x49, 0x18, 0x83, 0x31, 0x98, 0x47, 0x53,
 	}
+
+	// Below are the scrypt parameters that are used when creating the
+	// encryption key for the wallet database with snacl.NewSecretKey.
+	// We use very low values for the tests to speed things up.
+	fastScryptN = 8
```
cfromknecht: since we use these constants in a number of places, it'd be great if we just exposed them from the `waddrmgr`. definitely not blocking, but would be nice to have eventually

walletunlocker/service_test.go
```diff
@@ -461,3 +470,16 @@ func TestChangeWalletPassword(t *testing.T) {
 		t.Fatalf("password not received")
 	}
 }
+
+func init() {
```
halseth: hm, maybe try avoiding using init? Could we move it to `createTestWallet` instead?

guggero: Yes, good point. Copied this over from the itest PR (#3924) but it can be simpler for unit tests.

keychain/interface_test.go
guggero: Yeah, that would be nice. I'll create a PR there.

guggero: https://github.com/btcsuite/btcwallet/pull/675

cfromknecht: that pr has been merged, should be ready to import the newer version and use in this pr :)

guggero: Added the new version: d501e9a2
This also updates `btcwallet/walletdb` to version `v1.2.0`, I hope that doesn't have any side effects.

PR #3932, opened 2020-01-20
Re-create TLS files if IPs or domains changed

halseth:
This commit makes lnd recreate its TLS certificate if the config's
tlsextradomains or tlsextraips changed. This is useful, since earlier
user would have to manually delete the files to trigger lnd to recreate
them.

The new bahavior is gated behind a new flag `--tlsautorefresh `


cert/selfsigned.go
```diff
@@ -100,6 +100,52 @@ func dnsNames(tlsExtraDomains []string) (string, []string) {
 	return host, dnsNames
 }
 
+// Outdated returns whether the given certificate is outdated wrt the IPs and
```
Roasbeef: Naming suggestion: `IsOutdated`?

Roasbeef: wrt -> w.r.t

Roasbeef: `godoc` comment could also use further elaboration on what "outdated" means in this context (IPs or domains aren't a sub-set of what's currently in the cert), as a few of our related packages use this `cert` package to handle routine cert operations.

cert/selfsigned.go
```diff
@@ -100,6 +100,52 @@ func dnsNames(tlsExtraDomains []string) (string, []string) {
...
+
+	// Parse the slice of IP strings.
+	ips, err := ipAddresses(tlsExtraIPs)
+	if err != nil {
+		return false, err
+	}
+
+	// If the certificate has a different number of IP addresses, it is
+	// definitely out of date.
+	if len(ips) != len(cert.IPAddresses) {
```
Roasbeef: Alternatively, should we relax this to allow _more_ IPs in the cert than are specified in our config? A sample case being a cert that a user has generated themselves, and synced with the existing configuration options.

lnd.go
```diff
@@ -744,6 +744,12 @@ func getTLSConfig(tlsCertPath string, tlsKeyPath string, tlsExtraIPs,
 			return nil, nil, "", err
 		}
 		rpcsLog.Infof("Done renewing TLS certificates")
+
+		// Reload the certificate data.
+		certData, _, err = cert.LoadCert(tlsCertPath, tlsKeyPath)
```
Roasbeef: Depending on where we go with this PR, we may want to consider splitting off this commit (either into a new PR or to be cherry-picked onto the next release branch).

halseth:
Gate behind cmd line flag

cert/selfsigned.go
halseth: Done

cert/selfsigned.go
halseth: The problem then is that even though the user removes an IP from the config, the cert will still allow it.

cert/selfsigned_test.go
```diff
@@ -0,0 +1,65 @@
...
+			outdated, err := cert.IsOutdated(
+				parsedCert, extraIPs[:numIPs],
+				extraDomains[:numDomains],
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// We expect it to be considered outdated if the IPs or
+			// domains don't match exaactly what we created.
```
bhandras: nit: exactly

cert/selfsigned_test.go
```diff
@@ -0,0 +1,65 @@
...
+		extraIPs[:2], extraDomains[:2],
+		cert.DefaultAutogenValidity,
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// We'll attempt to check up-to-date status for all variants of 1-3
+	// number of IPs and domains.
+	for numIPs := 1; numIPs <= 3; numIPs++ {
```
bhandras: you could use `len(extraIPs)` and `len(extraDomains)` instead

cert/selfsigned_test.go
```diff
@@ -0,0 +1,65 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// We expect it to be considered outdated if the IPs or
+			// domains don't match exaactly what we created.
+			expected := numIPs != 2 || numDomains != 2
+			if outdated != expected {
+				t.Fatalf("expected certificate to be "+
+					"outdated=%v", expected)
```
bhandras: nit: i'd add "got=%v" to be more informative.

cert/selfsigned_test.go
halseth: Fixed

cert/selfsigned_test.go
halseth: That is much better indeed!

cert/selfsigned_test.go
halseth: done

cert/selfsigned.go
cfromknecht: this will return a false positive if the user accidentally includes a duplicate IP, when in fact we don't need to change the cert. imo it'd be better to convert both to sets, then compare length.

one possible downside of this is that certs generated before this may not have checked for duplicates, so it would cause those to be refreshed. perhaps we should start checking for duplicate entries before creating new certs as well?

cert/selfsigned.go
```diff
@@ -100,6 +100,53 @@ func dnsNames(tlsExtraDomains []string) (string, []string) {
...
+		return true, nil
+	}
+
+	// Go through each IP address, and check that they are equal. We expect
+	// the order to match exactly, as we assume the same tlsExtraIPs were
+	// used when the certificate was first created.
+	for i := range ips {
+		if !ips[i].Equal(cert.IPAddresses[i]) {
```
cfromknecht: related to above, imo order shouldn't matter here. plus we can reuse the sets created above to check. i think it'd be less obvious to users that reordering ip/dns/domains would cause their certs to be renewed

cert/selfsigned.go
```diff
@@ -100,6 +100,53 @@ func dnsNames(tlsExtraDomains []string) (string, []string) {
...
+			return true, nil
+		}
+	}
+
+	// Get the full list of DNS names to use.
+	_, dnsNames := dnsNames(tlsExtraDomains)
+
+	// If the number of domains are different, it is out of date.
+	if len(cert.DNSNames) != len(dnsNames) {
```
cfromknecht: similar comment here wrt to duplicates/set matching

lnd.go
cfromknecht: do we know why our unit tests didn't catch this?

config.go
```diff
@@ -232,21 +232,24 @@ type torConfig struct {
...
 	TLSExtraDomains []string `long:"tlsextradomain" description:"Adds an extra domain to the generated certificate"`
-	NoMacaroons     bool     `long:"no-macaroons" description:"Disable macaroon authentication"`
-	AdminMacPath    string   `long:"adminmacaroonpath" description:"Path to write the admin macaroon for lnd's RPC and REST services if it doesn't exist"`
-	ReadMacPath     string   `long:"readonlymacaroonpath" description:"Path to write the read-only macaroon for lnd's RPC and REST services if it doesn't exist"`
-	InvoiceMacPath  string   `long:"invoicemacaroonpath" description:"Path to the invoice-only macaroon for lnd's RPC and REST services if it doesn't exist"`
-	LogDir          string   `long:"logdir" description:"Directory to log output."`
-	MaxLogFiles     int      `long:"maxlogfiles" description:"Maximum logfiles to keep (0 for no rotation)"`
-	MaxLogFileSize  int      `long:"maxlogfilesize" description:"Maximum logfile size in MB"`
+	TLSAutoRefresh  bool     `long:"tlsautorefresh" description:"Re-generate TLS certificate and key if the IPs or domains are changed"`
```
cfromknecht: nit: recently i think we've been preferring the hyphenated config options, but i see that it would deviate from our other `--tls*` options :/

lnd.go
halseth: I don't think we have and `lnd.go` level unit tests?

config.go
halseth: yeah, would be nice to move all tls options into its own category, but maybe not for a minor release like this

cert/selfsigned.go
halseth: Yep, fixed: https://github.com/lightningnetwork/lnd/pull/3932/commits/7aba914bbe87456f039f245afc54b771baa398c8

config.go
cfromknecht: agreed!

cert/selfsigned.go
```diff
@@ -100,6 +100,83 @@ func dnsNames(tlsExtraDomains []string) (string, []string) {
...
+	for s, ip1 := range ips1 {
+		// Assert the IP string is found in both sets.
+		ip2, ok := ips2[s]
+		if !ok {
+			return true, nil
+		}
+
+		// And that the IPs are considered equal.
+		if !ip1.Equal(ip2) {
+			return true, nil
```
cfromknecht: is this reachable? is it possible for two IPs to have the same string representation, but be unequal?

halseth: No idea 😂

PR #3945, opened 2020-01-21
routing: penalize node-level failures harder

joostjager:
Previously we only penalized the outgoing connections of a failing node. This turned out not to be sufficient, because the next route sometimes went into the same failing node again to try a different outgoing connection that wasn't yet known to mission control and therefore not penalized before.

### DIFF
--- a/routing/result_interpretation.go
+++ b/routing/result_interpretation.go
@@ -424,11 +424,13 @@ func (i *interpretedResult) failNode(rt *route.Route, idx int) {
 	// Mark the incoming connection as failed for the node. We intent to
 	// penalize as much as we can for a node level failure, including future
 	// outgoing traffic for this connection. The pair as it is returned by
-	// getPair is directed towards the failed node. Therefore we first
-	// reverse the pair. We don't want to affect the score of the node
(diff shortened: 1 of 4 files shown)


routing/result_interpretation.go
```diff
@@ -424,11 +424,13 @@ func (i *interpretedResult) failNode(rt *route.Route, idx int) {
 	// Mark the incoming connection as failed for the node. We intent to
 	// penalize as much as we can for a node level failure, including future
 	// outgoing traffic for this connection. The pair as it is returned by
-	// getPair is directed towards the failed node. Therefore we first
-	// reverse the pair. We don't want to affect the score of the node
-	// sending towards the failing node.
+	// getPair is penalized in the original and the reversed direction. Note
+	// that this will also affect the score of the failing node's peers.
```
halseth: So node A can impact the score of node B by just not forwarding payments to it?

joostjager: Yes, that's true. You become less popular if you have bad friends. But this mainly affects A's link to B specifically.

PR #3950, opened 2020-01-22
invoices: fix slow startup with many expired invoices

bhandras:
This PR intends to fix slow first startup time when there are many
invoices that need to be canceled. The slowdown is caused by a combination
of adding invoices to the expiry watcher one-by-one and slow
cancellation. Due to slow cancellation and the unbuffered channel which
we use to pass invoices to the expiry watcher blocks the registry.
With this fix we'll instead batch add invoices to the expiry watcher and
thereby won't block the registry startup.

### DIFF
--- a/invoices/invoice_expiry_watcher.go
+++ b/invoices/invoice_expiry_watcher.go
@@ -49,7 +49,7 @@ type InvoiceExpiryWatcher struct {
 
 	// newInvoices channel is used to wake up the main loop when a new invoices
 	// is added.
-	newInvoices chan *invoiceExpiry
+	newInvoices chan []*invoiceExpiry
 
 	wg sync.WaitGroup
 
@@ -61,7 +61,7 @@ type InvoiceExpiryWatcher struct {
 func NewInvoiceExpiryWatcher(clock clock.Clock) *InvoiceExpiryWatcher {
 	return &InvoiceExpiryWatcher{
 		clock:       clock,
-		newInvoices: make(chan *invoiceExpiry),
+		newInvoices: make(chan []*invoiceExpiry),
 		quit:        make(chan struct{}),
 	}
 }
@@ -102,15 +102,14 @@ func (ew *InvoiceExpiryWatcher) Stop() {
 	}
 }
 
-// AddInvoice adds a new invoice to the InvoiceExpiryWatcher. This won't check
-// if the invoice is already added and will only add invoices with ContractOpen
-// state.
-func (ew *InvoiceExpiryWatcher) AddInvoice(
-	paymentHash lntypes.Hash, invoice *channeldb.Invoice) {
+// prepareInvoice checks if the passed invoice may be canceled and calculates
+// the expiry time.
+func (ew *InvoiceExpiryWatcher) prepareInvoice(
+	paymentHash lntypes.Hash, invoice *channeldb.Invoice) *invoiceExpiry {
 
 	if invoice.State != channeldb.ContractOpen {
 		log.Debugf("Invoice not added to expiry watcher: %v", invoice)
-		return
+		return nil
 	}
 
 	realExpiry := invoice.Terms.Expiry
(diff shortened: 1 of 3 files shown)


invoices/invoice_expiry_watcher.go
```diff
@@ -119,20 +118,54 @@ func (ew *InvoiceExpiryWatcher) AddInvoice(
...
 		PaymentHash: paymentHash,
 		Expiry:      expiry,
 	}
+}
+
+// AddInvoices adds multiple invoices to the InvoiceExpiryWatcher.
+func (ew *InvoiceExpiryWatcher) AddInvoices(
+	invoices []channeldb.InvoiceWithPaymentHash) {
+
+	var invoicesWithExpiry []*invoiceExpiry
```
Roasbeef: We can pre-size this slice as we know ahead of time how many invoices we'll need to be sending over.

invoices/invoice_expiry_watcher.go
```diff
@@ -119,20 +118,54 @@ func (ew *InvoiceExpiryWatcher) AddInvoice(
...
+		}
+	}
+
+	if len(invoicesWithExpiry) > 0 {
+		log.Debugf("Added %v invoices to the expiry watcher: %v",
+			len(invoicesWithExpiry))
+		select {
+		case ew.newInvoices <- invoicesWithExpiry:
+		case <-ew.quit:
+			// Select on quit too so that callers won't get blocked in case
```
Roasbeef: Typically we prefer to place comments like this right above the `case` statement.

invoices/invoice_expiry_watcher.go
```diff
@@ -119,20 +118,54 @@ func (ew *InvoiceExpiryWatcher) AddInvoice(
...
 		Expiry:      expiry,
 	}
+}
+
+// AddInvoices adds multiple invoices to the InvoiceExpiryWatcher.
+func (ew *InvoiceExpiryWatcher) AddInvoices(
+	invoices []channeldb.InvoiceWithPaymentHash) {
+
+	var invoicesWithExpiry []*invoiceExpiry
+	for idx := range invoices {
```
Roasbeef: Any particular reason you opted for indices here rather than the range loop (to pick up the item in a temp loop var). Mostly a style thing.

invoices/invoice_expiry_watcher.go
```diff
@@ -174,15 +207,17 @@ func (ew *InvoiceExpiryWatcher) mainLoop() {
 
 	for {
 		// Cancel any invoices that may have expired.
-		ew.cancelExpiredInvoices()
+		ew.cancelNextExpiredInvoice()
```
Roasbeef: Hmm, unsure if this is better than the prior diff. On one hand we'll always check for new invoices before consuming the existing set (partially addressing our starvation issue), but on the other-hand we'll now execute the main loop multiples time to chip away at the set of expired invoices.

invoices/invoiceregistry.go
```diff
@@ -198,12 +198,7 @@ func (i *InvoiceRegistry) populateExpiryWatcher() error {
 		return err
 	}
 
-	for idx := range pendingInvoices {
-		i.expiryWatcher.AddInvoice(
-			pendingInvoices[idx].PaymentHash, &pendingInvoices[idx].Invoice,
-		)
-	}
-
+	i.expiryWatcher.AddInvoices(pendingInvoices)
```
Roasbeef: May be useful to have log statements on either side of this line so we can track any long lived progress in the future.

invoices/invoice_expiry_watcher.go
Roasbeef: One option to counter act this is to buffer the `newInvoices` channel to a degree.

invoices/invoice_expiry_watcher.go
bhandras: Invoices not in ContractOpen state won't be added but at least we know the capacity. Fixed.

invoices/invoice_expiry_watcher.go
bhandras: Sorry, was copy-paste. Done.

invoices/invoice_expiry_watcher.go
bhandras: Done.

invoices/invoice_expiry_watcher.go
bhandras: Imo it's better as whoever called `InvoiceRegistry.AddInvoice` may still be blocked until we cancel everyting as we won't read from the `newInvoices` channel.

invoices/invoiceregistry.go
bhandras: Done.

invoices/invoice_expiry_watcher.go
```diff
@@ -119,20 +118,54 @@ func (ew *InvoiceExpiryWatcher) AddInvoice(
...
+}
+
+// AddInvoices adds multiple invoices to the InvoiceExpiryWatcher.
+func (ew *InvoiceExpiryWatcher) AddInvoices(
+	invoices []channeldb.InvoiceWithPaymentHash) {
+
+	invoicesWithExpiry := make([]*invoiceExpiry, 0, len(invoices))
+	for _, invoiceWithPaymentHash := range invoices {
+		newInvoiceExpiry := ew.prepareInvoice(
+			invoiceWithPaymentHash.PaymentHash, &invoiceWithPaymentHash.Invoice)
```
wpaulino: Nit: close paren should go on its own line.

invoices/invoiceregistry.go
```diff
@@ -198,12 +198,11 @@ func (i *InvoiceRegistry) populateExpiryWatcher() error {
...
-	for idx := range pendingInvoices {
-		i.expiryWatcher.AddInvoice(
-			pendingInvoices[idx].PaymentHash, &pendingInvoices[idx].Invoice,
-		)
-	}
-
+	log.Debugf("Adding %v pending invoices to the expiry watcher")
+	t0 := time.Now()
+	i.expiryWatcher.AddInvoices(pendingInvoices)
+	log.Debugf("Added %v pending invoices to the expiry watcher, took: %v",
```
wpaulino: Nit: I don't think this log is necessary, the one above gives enough context if things get stuck here for a while.

invoices/invoice_expiry_watcher.go
wpaulino: > Imo it's better as whoever called InvoiceRegistry.AddInvoice may still be blocked until we cancel everyting as we won't read from the newInvoices channel.

Good point, but in the current form we still can't guarantee that reading from `newInvoices` takes precedence over `nextExpiry`, so I think we'd want to add a non-blocking select statement before the current one:

```go
select {
case <-ew.newInvoices:
default:
}
```

Doing this will likely cause even higher CPU usage however.

bhandras:
> Good point, but in the current form we still can't guarantee that reading from newInvoices takes precedence over nextExpiry, so I think we'd want to add a non-blocking select statement before the current one.

That's a very good point. PTAL.

invoices/invoiceregistry.go
bhandras: Indeed, done.

invoices/invoice_expiry_watcher.go
bhandras: Done.

invoices/invoice_expiry_watcher.go
```diff
@@ -174,18 +208,32 @@ func (ew *InvoiceExpiryWatcher) mainLoop() {
...
-			// Wait until the next invoice expires, then cancel expired invoices.
+
+		case invoicesWithExpiry := <-ew.newInvoices:
+			// Take newly forwarded invoices with higher priority
+			// in order to not block the newInvoices channel.
+			for _, invoiceWithExpiry := range invoicesWithExpiry {
+				ew.expiryQueue.Push(invoiceWithExpiry)
+			}
 			continue
+		default:
```
wpaulino: Nit: add newline above.

bhandras: done

PR #3951, opened 2020-01-23
sweep: add exclusive groups

joostjager:
Allows certain sweep inputs to be kept in separate transactions at all times. This is a preparation for anchor outputs. Before the commitment tx confirms, there are three potential anchors that can be cpfp'ed. We want to cpfp them all, but if done in the same transaction, the transaction would guaranteed to be invalid.

Exponential backoff would eventually get the txes published too, but having exclusive groups makes the process faster.

### DIFF
--- a/sweep/bucket_list.go
+++ b/sweep/bucket_list.go
@@ -1,0 +1,45 @@
+package sweep
+
+// bucket contains a set of inputs that are not mutually exclusive.
+type bucket pendingInputs
+
+// tryAdd tries to add a new input to this bucket.
+func (b bucket) tryAdd(input *pendingInput) bool {
+	exclusiveGroup := input.params.ExclusiveGroup
+	if exclusiveGroup != nil {
+		for _, input := range b {
+			existingGroup := input.params.ExclusiveGroup
+			if existingGroup != nil &&
+				*existingGroup == *exclusiveGroup {
+
+				return false
+			}
+		}
+	}
+
+	b[*input.OutPoint()] = input
+
+	return true
+}
+
+// bucketList is a list of buckets that contain non-mutually exclusive inputs.
+type bucketList struct {
+	buckets []bucket
+}
+
+// add adds a new input. If the input is not accepted by any of the existing
+// buckets, a new bucket will be created.
+func (b *bucketList) add(input *pendingInput) {
+	for _, existingBucket := range b.buckets {
+		if existingBucket.tryAdd(input) {
+			return
+		}
+	}
+
+	// Create a new bucket and add the input. It is not necessary to check
+	// the return value of tryAdd because it will always succeed on an empty
+	// bucket.
+	newBucket := make(bucket)
+	newBucket.tryAdd(input)
+	b.buckets = append(b.buckets, newBucket)
+}
--- a/sweep/sweeper.go
+++ b/sweep/sweeper.go
@@ -402,10 +402,9 @@ func (s *UtxoSweeper) SweepInput(input input.Input,
 	}
 
 	log.Infof("Sweep request received: out_point=%v, witness_type=%v, "+
-		"time_lock=%v, amount=%v, fee_preference=%v, force=%v",
+		"time_lock=%v, amount=%v, params=(%v)",
 		input.OutPoint(), input.WitnessType(), input.BlocksToMaturity(),
-		btcutil.Amount(input.SignDesc().Output.Value),
-		params.Fee, params.Force)
+		btcutil.Amount(input.SignDesc().Output.Value), params)
 
 	sweeperInput := &sweepInputMessage{
 		input:      input,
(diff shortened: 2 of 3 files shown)


sweep/sweeper.go
```diff
@@ -402,10 +402,10 @@ func (s *UtxoSweeper) SweepInput(input input.Input,
 	}
 
 	log.Infof("Sweep request received: out_point=%v, witness_type=%v, "+
-		"time_lock=%v, amount=%v, fee_preference=%v, force=%v",
+		"time_lock=%v, amount=%v, params=(%v)",
 		input.OutPoint(), input.WitnessType(), input.BlocksToMaturity(),
 		btcutil.Amount(input.SignDesc().Output.Value),
-		params.Fee, params.Force)
+		params)
```
guggero: nit: looks like this could fit on the previous line now.

joostjager:
Not sure about repellent. Isn't that used for bug spray? :bug: 

>How would the group identifier be used for a commitment transaction? Would it be set to the channel ID that the outputs of the same commitment would never be in the same group but anchor outputs of other channels could?

Indeed, the short chan id is put in there. Anchor outputs of different commitment txes of the same channel would never end up in the same sweep tx and invalidate the whole tx.

sweep/sweeper.go

guggero:
I was thinking more of the same pole of magnets repelling each other. But yes, it's also what you call a bug spray :joy:

joostjager:
Need to add something to cancel all other inputs in the same exclusive group if one confirms. Otherwise they stay around in the sweeper causing failing txes for no reason.

sweep/bucket_list.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	b[*input.OutPoint()] = input
+
+	return true
+}
+
+// bucketList is a list of buckets that contain non-mutually exclusive inputs.
+type bucketList struct {
+	buckets []*bucket
```
cfromknecht: `bucket` is already a pointer since it's a map, `[]bucket` would be sufficient

sweep/bucket_list.go
```diff
@@ -0,0 +1,42 @@
...
+
+// bucket contains a set of inputs that are not mutually exclusive.
+type bucket pendingInputs
+
+// tryAdd tries to add a new input to this bucket.
+func (b bucket) tryAdd(input *pendingInput) bool {
+	exclusiveGroup := input.params.ExclusiveGroup
+	if exclusiveGroup != nil {
+		for _, input := range b {
+			if input.params.ExclusiveGroup == exclusiveGroup {
```
cfromknecht: shouldn't this test value equality rather than pointer equality? fwiw the unit test uses the same pointer for all inputs, which explains why this isn't caught

sweep/sweeper.go
```diff
@@ -679,7 +722,7 @@ func (s *UtxoSweeper) bucketForFeeRate(
 // sweep fee rate, which is determined by calculating the average fee rate of
 // all inputs within that cluster.
 func (s *UtxoSweeper) clusterBySweepFeeRate() []inputCluster {
-	bucketInputs := make(map[int]pendingInputs)
+	bucketInputs := make(map[int]*bucketList)
```
cfromknecht: similar to above, pointer type is redundant for slices

sweep/bucket_list.go
joostjager: good catch. unit test updated too.

sweep/bucket_list.go
joostjager: fixed

sweep/sweeper.go
joostjager: In this case, the slice is wrapped in a struct to allow it to be modified in-place using the `add` method. I tried to make `bucketList` just a type alias for the slice, but then you need to deal with the return value of `add` and also update the parent map that holds the (old) slice. I think it is easier to understand with the wrapping struct.

cfromknecht: gotcha that makes sense, i had forgotten about the member methods. thanks for the explanation

sweep/sweeper.go
```diff
@@ -640,6 +657,31 @@ func (s *UtxoSweeper) collector(blockEpochs <-chan *chainntnfs.BlockEpoch) {
 	}
 }
 
+// removeExclusiveGroup removes all inputs in the given exclusive group. This
+// function is called when one of the exclusive group inputs has been spent. The
+// other inputs won't ever be spendable and can be removed. This also prevents
+// them from being part of future sweep transactions that would fail.
+func (s *UtxoSweeper) removeExclusiveGroup(group uint64) {
+	for outpoint, input := range s.pendingInputs {
+		outpoint := outpoint
```
cfromknecht: line number irony?

joostjager: hehe, devilish design flag of go

PR #3952, opened 2020-01-23
channeldb: keep remote commit diff in memory

joostjager:
This change is preparation for anchor output resolution. By not querying the remote commit diff on demand from the database, it will be easier write unit tests for the channel arbitrator.

Additionally, this change will improve performance by eliminating a database lookup for every commitment signed.


joostjager:
I am probably going with a different approach for which this change isn't needed.

Issue #3939, opened 2020-01-21
routing+invoices: implement AMP secret sharing and combination 


Roasbeef:
This issue is related to https://github.com/lightningnetwork/lightning-rfc/pull/658. Completion of this issue entails creating a new `secretshare` package (naming suggestion) within `lnd` that's able to:
  * Statically create a set of N shares for apriori splitting use cases. 
  * Dynamically create a set of N shares (incrementally or adaptively) fo more adhoc splitting use case. 
  * Combine the relevant shares and perform key derivation to produce preimages and payment hashes related to an ultimate payment 

Once the above is done (the libraries at the edges), we'll also need to then integrate this new _combination_ logic into the invoice registry. Along the way we'll likely opt to create some sort of new HTLC combination+settling (deciding _when_ it's possible to combine and also how to derive the pre-images) within the invoice registry. 

I've left out most of the details w.r.t the additive secret sharing scheme as those are elaborated upon in detail within the linked issue. One other thing I've left out of this is the implication on the RPC interface, for calls such as `ListPayments` and `ListInvoices`.

cfromknecht:
Closed by #4162

PR #3957, opened 2020-01-25
record+routing/route: add AMP record

cfromknecht:
This PR adds the currently proposed AMP record from https://github.com/lightningnetwork/lightning-rfc/pull/658, and
implements the sanity checks when encoding a particular `route.Hop`. For now, this could will
remain dead until a number of other components are in place, but this serves as a starting point to
begin implementing the remainder of the proposal that depends on these primitive definitions
before ultimately enabling it externally.

Related to #3939.

### DIFF
--- a/record/amp.go
+++ b/record/amp.go
@@ -1,0 +1,107 @@
+package record
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// AMPOnionType is the type used in the onion to reference the AMP fields:
+// root_share, set_id, and child_index.
+const AMPOnionType tlv.Type = 10
+
+// AMP is a record that encodes the fields necessary for atomic multi-path
+// payments.
+type AMP struct {
+	rootShare  [32]byte
+	setID      [32]byte
+	childIndex uint16
+}
+
+// NewAMP generate a new AMP record with the given root_share, set_id, and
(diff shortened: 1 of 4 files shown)


record/amp.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// PayloadSize returns the size this record takes up in encoded form.
+func (a *AMP) PayloadSize() uint64 {
+	return 32 + 32 + tlv.SizeTUint16(a.childIndex)
+}
+
+// String returns a human-readble description of the amp payload fields.
+func (a *AMP) String() string {
+	return fmt.Sprintf("root_share=%x set_id=%x child_index=%d",
```
Roasbeef: 👍

routing/route/route.go
```diff
@@ -168,6 +176,18 @@ func (h *Hop) PackHopPayload(w io.Writer, nextChanID uint64) error {
 		}
 	}
 
+	// If an AMP record is destined for this hop, ensure that we only ever
```
Roasbeef: 👍

cfromknecht:
fixed the linter errors, build now passes

joostjager:
Does merging this now really make things easier? It doesn't look like this code has conflict potential. In general I think it is better to merge living code, unless we have for example two prs that build onto this. But not sure if that is already the case.

Issue #3753, opened 2019-11-21
ListPayments should offer pagination


alexbosworth:
When fetching invoices and other records like forwards via the RPC, those record sets are separated into pages to allow for scalable retrieval of data.

With `ListPayments` all of the payments are returned in a single response. If a node is doing regular payments or the incomplete payments flag is used, the number of results from this request can be very large.

In PR https://github.com/lightningnetwork/lnd/pull/1962 it was suggested that the fix should be to add pagination.

bitromortac:
I will try to do this :+1:

Roasbeef:
Closed by #3960. Note that there'll be a future follow up to make the implementation more efficient in the future as well.

Issue #3942, opened 2020-01-21
channeldb+invoices: 


Roasbeef:
This is a master tracking issue which follows a larger project that has a goal of plugging the remaining accounting holes within `lnd`, as well as providing end users with a more coherent interface that they can use to consume complete accounting data delated to `lnd`. 

See this post on the `lnd` mailing list for a outline of our current design w.r.t the end-user interface of this new sub-system:  https://groups.google.com/a/lightning.engineering/forum/#!topic/lnd/bEnRBZvsmB4

Related issues: 
 * #2579 
 * #1563
 * #3709

bitromortac:
Hey, I think #2936 and #3708 are also related. I could try to look a bit deeper into them.

Roasbeef:
Removing from milestone for now as we're in the process of gathering better requirements from the ecosystem to shape this issue properly.

saubyk:
Closing this issue as all of the linked issues are closed. Can be reopened if there is any unmet requirement still.

PR #1719, opened 2018-08-10
Control Tower: Sender-side checks for duplicate payments #2

cfromknecht:
This PR is an extension of #973, mostly just touching up documentation. The only thing that is new in the PR is the reordering of logic w/in `handleLocalDispatch`, which should result in better clean up of circuits and fwd references. This was going to be done in a separate PR, but the changes were overlapping enough that I made them here.

PR #3960, opened 2020-01-27
channeldb+lnrpc+lncli: listpayments pagination support

bitromortac:
This PR tries to address issue #3753, pagination support for the ListPayments api.

The goal of making the ListPayments command paginable is to turn the retrival of conducted
payments more scalable and seekable (e.g. for mobile wallets).

The implementation closely follows the behavior of the ListInvoices api,
modifying the grpc ListPaymentsRequest to also include an index offset, the option
to sepcify the pagination direction, and to give a maximal number of returned payments constraint.
The call returns a ListPaymentsResponse where in addition to the payments, the first and
last index of the payments are included, which can be used to further seek forwards or
backwards. By default, the api call returns the same payments as before.

The old behavior of ListPayments is to return all payments via the FetchPayments method.
In the implementation of queried ListPayments in this PR, still all payments are loaded
into memory first by using the FetchPayments method and then discarding the unwanted entries.

This means, that memory usage is the same as before, while adding more functionality for the caller.
The reason why I decided to do that instead of reproducing the implementaion of ListInvoices
(where the database is seeked with a cursor) is that the the payments database structure 
is complicated, with subbuckets for duplicate payments of the same payment hash on the one side and that the payments bucket is indexed by the payment hash instead of the sequence index (which could be used to walk through them in order). This could be addressed in a future PR, making the call more memory effective, by droppping duplicate payments and creating a bucket with a mapping from the squence index to the payment hash.

#### Pull Request Checklist

  per default)
  already exist


carlaKC:
Assigning myself because this falls under #3942.
Thanks for the PR @bitromortac 🎉

channeldb/payments_test.go
```diff
@@ -173,3 +174,178 @@ func TestRouteSerialization(t *testing.T) {
...
+		IncludeIncomplete: true,
+	})
+	for i, p := range allPayments {
+		q := querySlice.Payments[i]
+		if p.sequenceNum != q.sequenceNum {
+			t.Fatalf("sequence numbers do not match, "+
+				"got %v, want %v", p.sequenceNum, q.sequenceNum)
+		}
+	}
+
```
bjarnemagnussen: May need more test cases handling e.g. different starting `IndexOffset` values.

A naive test to currently trigger an index-out-of-range, see comment above:

```golang
// Should return an empty slice but currently triggers an index-out-of-range.
querySlice, _ = db.QueryPayments(PaymentsQuery{
	IndexOffset:       7,
	NumMaxPayments:    7,
	Reversed:          false,
	IncludeIncomplete: true,
})
if len(querySlice.Payments) != 0 {
	t.Fatalf("index-out-of-range is triggered before test is completed.")
}
```

channeldb/payments.go
```diff
@@ -443,6 +443,154 @@ func fetchPayment(bucket *bbolt.Bucket) (*MPPayment, error) {
...
+			return paymentsLength, true
+		case indexOffset > 1:
+			return indexOffset - 1, true
+		default:
+			return 0, false
+		}
+	}
+	// Normal ordering
+	switch {
+	case indexOffset <= paymentsLength:
```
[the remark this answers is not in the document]
bitromortac: That's true, nice catch! Removed the equal condition.

channeldb/payments_test.go
bitromortac: Perfect, test triggered the out-of-range error. Added the test, which is now passing.

bitromortac: Hey @bjarnemagnussen, thank you very much for your review!

bitromortac:
Hey @bjarnemagnussen, thank you very much for your review!

channeldb/payments.go
```diff
@@ -443,6 +443,154 @@ func fetchPayment(bucket *bbolt.Bucket) (*MPPayment, error) {
...
+}
+
+// QueryPayments is a query to the payments database which is restricted
+// to a subset of payments by the payments query, containing an offset
+// index and a maximum number of returned payments.
+func (db *DB) QueryPayments(query PaymentsQuery) (PaymentsQuerySlice, error) {
+	resp := PaymentsQuerySlice{
+		PaymentsQuery: query,
+	}
+	allPayments, err := db.FetchPayments()
```
joostjager: I checked out `FetchPayments` quickly and it looks like the list that is returned here, is sorted by payment hash. That means that if a new payment is initiated, the ordering will be different and everything shifts?

bitromortac: No, it's sorted by the sequence number, the equivalent to the add index in the invoices db, so no reordering should occur.

joostjager: Ah right, there is a sort at the end of `FetchPayments`

lnrpc/rpc.proto
```diff
@@ -2794,12 +2794,40 @@ message ListPaymentsRequest {
     that pending payments, as well as failed payments will show up if this
```
carlaKC: This commit should be squashed with the previous one so that each commit builds :)

cmd/lncli/commands.go
```diff
@@ -2913,11 +2913,40 @@ func describeGraph(ctx *cli.Context) error {
 var listPaymentsCommand = cli.Command{
 	Name:     "listpayments",
 	Category: "Payments",
-	Usage:    "List all outgoing payments.",
+	Usage:    "Query the payment history.",
```
carlaKC: List payments includes payments that are still in flight so 'history' doesn't quite parse.

cmd/lncli/commands.go
```diff
@@ -2913,11 +2913,40 @@ func describeGraph(ctx *cli.Context) error {
...
+	Description: "This command enables the retrieval of payments stored in the " +
+		"database. Pagination is supported by the usage of " +
+		"index_offset in combination with the reversed flag. " +
+		"Pagination can be resumed by using the returned " +
+		"last_index_offset (for normal order), or first_index_offset " +
```
carlaKC: nit: I'm not a fan of using the term normal to describe behaviour, it is somehow too descriptive and too general at the same time. Rather stick to forwards and reverse? Or a better word for forwards that I can't think of 🤔

channeldb/payments.go
```diff
@@ -460,6 +460,12 @@ type PaymentsQuery struct {
 	// true, the query will fetch payments with indices lower than the
 	// IndexOffset.
 	Reversed bool
+
+	// If IncludeIncomplete is true, then return payments that have
+	// not yet fully completed. This means that pending payments,
+	// as well as failed payments will show up if this
+	// field is set to True.
+	IncludeIncomplete bool
```
carlaKC: I think that this commit belongs before we introduce paginated queries? That way you don't need to touch this code twice, and the commit where you add pagination works as intended from the get go.

channeldb/payments.go
```diff
@@ -443,6 +443,141 @@ func fetchPayment(bucket *bbolt.Bucket) (*MPPayment, error) {
...
+		// of the payments range, we start from the last known payment.
+		case indexOffset == 0 || indexOffset > paymentsLength:
+			return paymentsLength, true
+		case indexOffset > 1:
+			return indexOffset - 1, true
+		default:
+			return 0, false
+		}
+	}
+	// Normal ordering
```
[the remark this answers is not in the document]
bitromortac: Corrected.

channeldb/payments.go
```diff
@@ -443,6 +443,141 @@ func fetchPayment(bucket *bbolt.Bucket) (*MPPayment, error) {
...
+// to a subset of payments by the payments query, containing an offset
+// index and a maximum number of returned payments.
+func (db *DB) QueryPayments(query PaymentsQuery) (PaymentsQuerySlice, error) {
+	resp := PaymentsQuerySlice{
+		PaymentsQuery: query,
+	}
+
+	allPayments, err := db.FetchPayments()
+	if err != nil {
+		return resp, err
```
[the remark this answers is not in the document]
bitromortac: Here the empty response struct (only query is populated) is returned, nil doesn't work here, but will keep it in mind.

channeldb/payments_test.go
```diff
@@ -173,3 +174,178 @@ func TestRouteSerialization(t *testing.T) {
...
+	if len(allPayments) != 7 {
+		t.Fatalf("Number of payments received does not "+
+			"match expected one. Got %v, want %v.",
+			len(allPayments), 7)
+	}
+
+	// Define test data for QueryPayments
+	data := []struct {
```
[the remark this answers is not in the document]
bitromortac: That's exactly what I wanted to do, thank you for the link!

rpcserver.go
```diff
@@ -4766,19 +4766,36 @@ func marshallTopologyChange(topChange *routing.TopologyChange) *lnrpc.GraphTopol
 	}
 }
 
-// ListPayments returns a list of all outgoing payments.
+// ListPayments returns a list of outgoing payments determined by a constrained
```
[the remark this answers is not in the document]
bitromortac: Fixed.

lnrpc/rpc.proto
bitromortac: Have squashed those commits.

cmd/lncli/commands.go
bitromortac: Reverted this change.

cmd/lncli/commands.go
bitromortac: Agree, have used forwards instead of normal.

channeldb/payments.go
bitromortac: I kept this commit separated, as I was not sure whether the incomplete payments should be discarded at the level of QueryPayments or at the rpcserver level and wanted to discuss this. I think it's best as it is now and also squashed the commit with the payments query commit as you suggest.

bitromortac:
Since #4003 is merged, I made the flag for pagination consistent with the listinvoices command.

channeldb/payments.go
```diff
@@ -421,6 +421,154 @@ func fetchPayment(bucket *bbolt.Bucket) (*MPPayment, error) {
...
+	NumMaxPayments uint64
+
+	// Reversed gives a meaning to the IndexOffset. If reversed is set to
+	// true, the query will fetch payments with indices lower than the
+	// IndexOffset.
+	Reversed bool
+
+	// If IncludeIncomplete is true, then return payments that have
+	// not yet fully completed. This means that pending payments,
+	// as well as failed payments will show up if this
```
carlaKC: nit, line wrap looks odd here + no capital for `True` in comment.

channeldb/payments.go
```diff
@@ -421,6 +421,154 @@ func fetchPayment(bucket *bbolt.Bucket) (*MPPayment, error) {
...
+		if query.Reversed {
+			resp.FirstIndexOffset = i
+			if resp.LastIndexOffset == 0 {
+				resp.LastIndexOffset = i
+			}
+		} else {
+			if resp.FirstIndexOffset == 0 {
+				resp.FirstIndexOffset = i
+			}
+			resp.LastIndexOffset = i
```
carlaKC: add some comments here explaining the reasoning behind this? It's quite hard to parse on first look.

lnrpc/rpc.proto
```diff
@@ -2817,12 +2817,40 @@ message ListPaymentsRequest {
...
+    to seek backwards, pagination style.
+    */
+    uint64 first_index_offset = 2 [json_name = "first_index_offset"];
+
+    /**
+    The index of the last item in the set of returned payments. This can be used
+    to seek further, pagination style.
+    */
+    uint64 last_index_offset = 3 [json_name = "last_index_offset"];
+
```
carlaKC: nit: unnecessary newline

bitromortac:
Thanks for the feedback! I agree with the total resource usage argument, it's not ideal. In that sense it is good that default behavior in this PR (returning all payments) is not changed and therefore the resource consumption stays the same for that call. I would rather see this PR as a preparatory one, which burdens a little bit more the server side (for subsequent pagination), but adds already benefit for callers not having to transfer as much data as before, improving latency and client side data handling. The code is already prepared for a cursor-style traversal of the data, which would need another bucket which maps from the `sequenceIndex` to the payment hash.

Concerning the propagation down the call stack, that could be done going down until `fetchPayment` to get parsing and memory benefits. Initially, I determined the first and last index for the range, but then encountered the problem that you don't know beforehand how many payments will be excluded with the `IncludeIncomplete` flag (in conjunction with the `NumMaxPayments`). I found the code also easier to understand in the way it's implemented now and was interested to work towards the ideal solution.

The ideal way of implementing this would be using a database cursor like it is done in `listinvoices`. We could postpone this PR until the payments database is cleaned up and then use the cursor. I would be interested in taking care of the cleanup.

The big question is how to deal with the duplicate payments to the same payment hashes. I don't know what's the policy about deleting user data, but I'd suggest to do a migration, which discards duplicate payments (they were forbidden already some time ago, see https://github.com/lightningnetwork/lnd/pull/1719) and log the discarded duplicate payments upon migration. That would get rid of the legacy duplicate payments database structure and would simplify things.

channeldb/payments_test.go
```diff
@@ -173,3 +174,195 @@ func TestRouteSerialization(t *testing.T) {
...
+				IncludeIncomplete: true,
+			},
+			firstIndex: 1,
+			lastIndex:  3,
+		},
+	}
+
+	// Test queried payments.
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
```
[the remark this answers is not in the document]
bitromortac: Fixed it, good to know!

channeldb/payments.go
bitromortac: Improved it.

channeldb/payments.go
bitromortac: Yes, have commented the code.

lnrpc/rpc.proto
bitromortac: Yes, done.

channeldb/payments_test.go
```diff
@@ -173,3 +174,195 @@ func TestRouteSerialization(t *testing.T) {
 			spew.Sdump(testRoute), spew.Sdump(route2))
 	}
 }
+
+// TestQueryPayments tests retrieval of payments with forwards and reversed
+// queries.
+func TestQueryPayments(t *testing.T) {
+	db, err := initDB()
```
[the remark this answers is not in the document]
bitromortac: I would leave it that way, if you're ok with it, otherwise I would have to repeat the setup code for the last part of the test and all the tests are read-only on the database anyhow.

carlaKC:
Some high sending nodes are already running into issues here, so going to go ahead with this for 0.10 😄 As you say, performance not affected for the default use case. 

Is this ready for another round of review @bitromortac?

bitromortac:
> Some high sending nodes are already running into issues here, so going to go ahead with this for 0.10 smile As you say, performance not affected for the default use case.
> 
> Is this ready for another round of review @bitromortac?

Oh, didn't know that it's already urgent. Yes, I tried to incorporate all requested changes.

Second thought on @joostjager's suggestion: we could drop the `IncludeIncomplete` flag and include all payments by default. In this way we can determine firstIndex and lastIndex deterministically and filter down in the stack, getting a performance improvement.

carlaKC:
> we could drop the `IncludeIncomplete` flag and include all payments by default.

From `rpcserver.go#ListPayments`: `To keep compatibility with the old API, we only return non-succeeded payments if requested.` Would say that we still want to maintain backwards compatibility here. People are pretty reliant on `ListPayments` and could see some unexpected behaviour if they don't notice this in the changelog.

carlaKC:
> Yes, I tried to incorporate all requested changes.

Great! Don't forget to re-request review when something is ready, otherwise we might miss it because the request doesn't show up on Github :) Also needs a rebase ⚾️

lnrpc/rpc.proto
```diff
@@ -2909,12 +2909,39 @@ message ListPaymentsRequest {
...
+    bool include_incomplete = 1 [json_name = "include_incomplete"];
+
+    /**
+    The index of a payment that will be used as either the start or end of a
+    query to determine which payments should be returned in the response.
+    */
+    uint64 index_offset = 2 [json_name = "index_offset"];
+
+    /// The max number of payments to return in the response to this query.
+    uint64 num_max_payments = 3 [json_name = "num_max_payments"];
```
joostjager: `num_max_payments`, what is that the abbreviation of? I read this as 'number of maximum payments'.

lnrpc/rpc.proto
```diff
@@ -2909,12 +2909,39 @@ message ListPaymentsRequest {
     that pending payments, as well as failed payments will show up if this
     field is set to True.
     */
-    bool include_incomplete = 1;
+    bool include_incomplete = 1 [json_name = "include_incomplete"];
+
+    /**
+    The index of a payment that will be used as either the start or end of a
+    query to determine which payments should be returned in the response.
+    */
```
joostjager: Document that it is zero-based?

joostjager: Probably also mention how the index works together with `include_incomplete`. Does this filter change the indexing?

lnrpc/rpc.proto
```diff
@@ -2909,12 +2909,39 @@ message ListPaymentsRequest {
...
 message ListPaymentsResponse {
     /// The list of payments
-    repeated Payment payments = 1;
+    repeated Payment payments = 1 [json_name = "payments"];
+
+    /**
+    The index of the first item in the set of returned payments. This can be
+    used to seek backwards, pagination style.
+    */
+    uint64 first_index_offset = 2 [json_name = "first_index_offset"];
```
joostjager: using both the words `index` and `offset`, isn't that redundant? Maybe `first_index` and `last_index` is sufficient

lnrpc/rpc.proto
```diff
@@ -2909,12 +2909,39 @@ message ListPaymentsRequest {
...
 }
 
 message ListPaymentsResponse {
     /// The list of payments
-    repeated Payment payments = 1;
+    repeated Payment payments = 1 [json_name = "payments"];
+
+    /**
+    The index of the first item in the set of returned payments. This can be
+    used to seek backwards, pagination style.
```
joostjager: To seek backwards, do I need to call `ListPayments` again with `index_offset` set to `first_index_offset - 1`?

channeldb/payments.go
```diff
@@ -383,6 +383,158 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
 	return deserializeHTLCFailInfo(r)
 }
 
+// PaymentsQuery represents a query to the payments database starting or ending
+// at a certain offset index. The number of retrieved records can be limited.
+type PaymentsQuery struct {
+	// IndexOffset determines the starting point of the payments query and
+	// is always exclusive. In normal order, the query starts one index
+	// higher than the IndexOffset, in reversed order, the query ends one
+	// index lower than the IndexOffset.
```
joostjager: I am wondering, isn't it more transparent to leave the minus and plus one to the rpc caller to apply?

channeldb/payments.go
```diff
@@ -383,6 +383,158 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+	if paymentsLength == 0 {
+		return resp, nil
+	}
+
+	i, ok := startIndex(query.IndexOffset, paymentsLength, query.Reversed)
+
+	var count uint64
+	for ; count < query.NumMaxPayments && ok; i, ok = nextIndex(
+		i, paymentsLength, query.Reversed) {
+		// To keep compatibility with the old API, we only return
```
joostjager: We generally insert a newline here

channeldb/payments.go
```diff
@@ -383,6 +383,158 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+			}
+			resp.LastIndexOffset = i
+		}
+
+		resp.Payments = append(resp.Payments, *allPayments[i-1])
+		count++
+	}
+
+	// Need to swap the payments slice order if reversed order.
+	if query.Reversed {
```
joostjager: Why swap actually? So the idea is that the pages are reversed, but per page the payments are in ascending order? Also something to document in the proto.

bitromortac:
Gosh, sorry messed up the commits. Trying to fix it.

bitromortac:
> Main comment is that the +1/-1 corrections seem to make the code more complicated than necessary. I don't think it is a problem to let rpc users specify the exact range, but I may be missing something.

Thank you for the review! I agree, it's a bit ambiguous, but the main motivation for this is to have the rpc work the exact same way as the listinvoices pagination. As a caller you just take the first_index or last_index of the reply and can go on paginating with this value as the offset_index. Also as it's called an offset, I am ok with making the offset exclusive as in listinvoices.

> The other question is why we add another index when we already have a sequence number. Isn't it better to just use that?

Are you referring to the suggested database change? Would just add another bucket which is keyed by the sequence index mapping to the payment hash. In that way one would be able to have an efficient cursor on the sequence index to be able to traverse the associated payment hashes. When looking at the db with boltbrowser I see entries ordered lexicographically, so I assume having an ordered index is more efficient to go through the payments, do you have a suggestion about that?

lnrpc/rpc.proto
bitromortac: Changed it to `max_payments` and `NumMaxPayments`->`MaxPayments`. In listinvoices we have `max_invoices`.

lnrpc/rpc.proto
bitromortac: Actually it's one-based, only zero has a special meaning. The filter doesn't change the index meaning, documented that in the rpc as well as in lncli.

lnrpc/rpc.proto
bitromortac: It's again due compatibility with listinvoices, have also thought about that. It's an offset for an index, which is ok, I think.

lnrpc/rpc.proto
bitromortac: You just need to take the `first_index_offset` to continue pagination.

channeldb/payments.go
bitromortac: Answered in main reply.

channeldb/payments.go
bitromortac: New line added.

channeldb/payments.go
bitromortac: It's again a replication of `listinvoices` behavior. Have documented in proto under the `reversed` flag documentation.

channeldb/payments.go
```diff
@@ -383,6 +383,159 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+	// index lower than the IndexOffset.
+	IndexOffset uint64
+
+	// MaxPayments is the maximal number of payments returned in the
+	// payments query.
+	MaxPayments uint64
+
+	// Reversed gives a meaning to the IndexOffset. If reversed is set to
+	// true, the query will fetch payments with indices lower than the
+	// IndexOffset.
```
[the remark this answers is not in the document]
bitromortac: Right, have added this.

channeldb/payments.go
```diff
@@ -383,6 +383,159 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+
+	// Payments is the set of payments returned from the database for the
+	// PaymentsQuery.
+	Payments []MPPayment
+
+	// FirstIndexOffset is the index of the first element in the set of
+	// returned MPPayments. Callers can use this to resume their query
+	// in the event that the slice has too many events to fit into a single
+	// response.
+	FirstIndexOffset uint64
```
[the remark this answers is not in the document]
bitromortac: Have added the typical usage to the comments of `PaymentQuerySlice` as well as to the description of the RPC fields.

channeldb/payments_test.go
```diff
@@ -175,3 +176,197 @@ func TestRouteSerialization(t *testing.T) {
...
+
+	// Define table driven test for QueryPayments.
+	tests := []struct {
+		desc       string
+		query      PaymentsQuery
+		firstIndex uint64
+		lastIndex  uint64
+	}{
+		{
+			desc: "Test IndexOffset at the end of the payments range.",
```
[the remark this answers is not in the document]
bitromortac: Made all test descriptions consistent, respecting the 80 chars.

channeldb/payments_test.go
```diff
@@ -175,3 +176,197 @@ func TestRouteSerialization(t *testing.T) {
...
+
+	if len(allPayments) != 7 {
+		t.Fatalf("Number of payments received does not "+
+			"match expected one. Got %v, want %v.",
+			len(allPayments), 7)
+	}
+
+	// Define table driven test for QueryPayments.
+	tests := []struct {
+		desc       string
```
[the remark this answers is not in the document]
bitromortac: Use now `name` instead of `desc`.

bitromortac:
Ok, now I get you :), thank you very much. The idea was to be close to a future cursor implementation, where one would essentially replace the `startIndex` functionality with initialization of the cursor and then perform the iteration with the `nextIndex`. On the other hand I thought better not rely on the sequence index being complete, because one could expect, e.g., deletion of failed payments in the future. I only assumed that the sequence index gives a time ordering of the payments. I like your approach too and will have a closer look and try to complete it for having an alternative.

carlaKC:
> On the other hand I thought better not rely on the sequence index being complete, because one could expect, e.g., deletion of failed payments in the future

I think that by sequence number would have less of an issue in a future where we delete payments? Picturing payments with sequence nums: [1, 2, 3]
- We query for `num_payments` = 2, `index_offset`=0: we get [1,2] and `last_offset`=2
- Now we delete with payment with sequence number 2. 
- If we then make a second call, with `index_offset`=2, we will miss payment 3 because the index has shifted with the deletion.
This wouldn't happen if we were indexing by sequence number? 

> I only assumed that the sequence index gives a time ordering of the payments

You can assume sequence number is unique and increasing with time order. There will be gaps in sequence nr (from older versions), but we can at least have a stable index over time?

bitromortac:
Ok, so it's the question about whether we want to keep the indices constant or make them incremental by one, if there is a gap in the `sequece_index`. `listinvoices` keeps the index constant and also exports the equivalent `add_index` in the RPC call. On the other hand it might be confusing for a consumer of the API if the `sequence_index` is gapped, should we also export the `sequence_index` in the RPC to inform the consumer?

I like your approach more (great work!), it is more in line what exists in `listinvoices` (also deletion of payments is hypothetical). I updated the commits to use your approach and also the tests (is there a better git way to acknowledge your work?).

channeldb/payments.go
```diff
@@ -383,6 +383,139 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+	allPayments, err := db.FetchPayments()
+	if err != nil {
+		return resp, err
+	}
+
+	if len(allPayments) == 0 {
+		return resp, nil
+	}
+
+	indexLimit := query.IndexOffset
```
joostjager: `indexExclusiveLimit`? The exclusive part helps me to understand the code better, but it may be a bit long

channeldb/payments.go
```diff
@@ -383,6 +383,139 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+			continue
+		}
+
+		// If we are paginating forwards, skip over all payments that
+		// have sequence numbers less than or equal to the index offset.
+		// We skip payments with equal indexes because the index offset
+		// is exclusive.
+		if !query.Reversed && payment.sequenceNum <= indexLimit {
+			continue
+		}
```
joostjager: Maybe it is clearer to group all code for forward and all code for reverse together in one if-then-else. But up to your preference.

lnrpc/rpc.proto
```diff
@@ -485,7 +485,9 @@ service Lightning {
     */
     rpc CloseChannel (CloseChannelRequest) returns (stream CloseStatusUpdate) {
         option (google.api.http) = {
-            delete: "/v1/channels/{channel_point.funding_txid_str}/{channel_point.output_index}"
+            delete:
```
joostjager: commit message: we use subsystems as prefixes, but lnd is not a subsystem

lnrpc/rpc.proto
```diff
@@ -498,7 +500,9 @@ service Lightning {
     rpc AbandonChannel (AbandonChannelRequest)
         returns (AbandonChannelResponse) {
         option (google.api.http) = {
-            delete: "/v1/channels/abandon/{channel_point.funding_txid_str}/{channel_point.output_index}"
+            delete:
+                "/v1/channels/abandon/{channel_point.funding_txid_str}/"
+                "{channel_point.output_index}"
```
joostjager: These changes can be reverted, probably caused by a different version of clang

lnrpc/rpc.proto
```diff
@@ -2863,14 +2868,43 @@ message ListPaymentsRequest {
...
+    The index of the first item in the set of returned payments. This can be
+    used as the index_offset to continue seeking backwards in the next request.
+    */
+    uint64 first_index_offset = 2 [json_name = "first_index_offset"];
+
+    /**
+    The index of the last item in the set of returned payments. This can be used
+    as the index_offset to continue seeking forwards in the next request.
+    */
+    uint64 last_index_offset = 3 [json_name = "last_index_offset"];
```
joostjager: Instead of these two fields, how about exposing the sequence number on the payment? The caller can then figure out min and max themselves and it also creates a stronger link between the query and the response

halseth:
> Ok, so it's the question about whether we want to keep the indices constant or make them incremental by one, if there is a gap in the `sequece_index`. `listinvoices` keeps the index constant and also exports the equivalent `add_index` in the RPC call. On the other hand it might be confusing for a consumer of the API if the `sequence_index` is gapped, should we also export the `sequence_index` in the RPC to inform the consumer?
> 
> I like your approach more (great work!), it is more in line what exists in `listinvoices` (also deletion of payments is hypothetical). I updated the commits to use your approach and also the tests (is there a better git way to acknowledge your work?). Haven't worked on the payment exclusion part yet.

I like the idea of just using the `sequence_num` as a payment identifier! Since some old database actually can have duplicate payments to the same payment hash, the `sequence_num` is the only unique payment identifier we have. There might be gaps in the list, but that's okay, since we also want to be able to delete payments without changing the identifiers.

What we could do is exposing this sequence number on the RPC (maybe call it something more descriptive, like `payment_id`) and then use that to paginate (making it clear that they are increasing in order of payment attempt time). We wouldn't even need the `first_index_offset` and `last_index_offset` values, as the next offset for pagination could be retrieved from the last list of payments retrieved.

channeldb/payments_test.go
```diff
@@ -175,3 +176,210 @@ func TestRouteSerialization(t *testing.T) {
 			spew.Sdump(testRoute), spew.Sdump(route2))
 	}
 }
+
+// TestQueryPayments tests retrieval of payments with forwards and reversed
+// queries.
+func TestQueryPayments(t *testing.T) {
+	// Define table driven test for QueryPayments.
```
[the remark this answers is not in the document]
bitromortac: I introduced a gap into the payment sequence numbers and added two new tests: setting the `IncludeIncomplete` flag to false and querying directly before the gap.

channeldb/payments.go
```diff
@@ -383,6 +383,139 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
+	if query.Reversed {
+		for l, r := 0, len(resp.Payments)-1; l < r; l, r = l+1, r-1 {
+			resp.Payments[l], resp.Payments[r] =
+				resp.Payments[r], resp.Payments[l]
+		}
+	}
+
+	// Set the first and last index of the returned payments so that the
+	// caller can resume from this point later on.
+	if len(resp.Payments) > 0 {
```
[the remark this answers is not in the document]
bitromortac: Thank you :)

channeldb/payments.go
```diff
@@ -383,6 +383,139 @@ func fetchHtlcFailInfo(bucket *bbolt.Bucket) (*HTLCFailInfo, error) {
...
 }
 
+// PaymentsQuery represents a query to the payments database starting or ending
+// at a certain offset index. The number of retrieved records can be limited.
+type PaymentsQuery struct {
+	// IndexOffset determines the starting point of the payments query and
+	// is always exclusive. In normal order, the query starts one index
+	// higher than the IndexOffset, in reversed order, the query ends one
+	// index lower than the IndexOffset.
+	IndexOffset uint64
```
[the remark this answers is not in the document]
bitromortac: Right, have corrected both comments. Concerning the maxint, it also works when supplied, but left zero to be the documented special value.

channeldb/payments.go
bitromortac: Sure, have renamed it to `indexExclusiveLimit`.

channeldb/payments.go
bitromortac: Yes, that's clearer. Regrouped the logic into reverse and forward directions.

lnrpc/rpc.proto
bitromortac: Good to know, have removed lnd from commit message.

lnrpc/rpc.proto
bitromortac: Have reverted the changes, which version of clang are you using?

bitromortac:
The latest state of the PR includes improved unit testing, where the case of a gap in the indices is covered. Also, the `SequenceIndex` is now exported in the rpc as the field `payment_index` (instead of Johan's suggestion, I took `index` instead of `id`, because we have `index` everywhere :slightly_smiling_face:).
The last point would be if we should leave away the first and last index in the response. Pro keeping it would be that it's working the same way as in `listinvoices` and it adds additional explicit documentation to the rpc, but on the other hand it's redundant :thinking:.

channeldb/payments.go
```diff
@@ -382,6 +382,143 @@ func fetchHtlcFailInfo(bucket kvdb.ReadBucket) (*HTLCFailInfo, error) {
...
+		return resp, err
+	}
+
+	if len(allPayments) == 0 {
+		return resp, nil
+	}
+
+	indexExclusiveLimit := query.IndexOffset
+	// In backward pagination, if the index limit is the default 0 value,
+	// we set our limit to our highest sequence number.
```
carlaKC: nit: We set it to highest sequence nr +1 so that we will still include the highest sequence nr (since it's exclusive)

channeldb/payments_test.go
```diff
@@ -175,3 +177,261 @@ func TestRouteSerialization(t *testing.T) {
 			spew.Sdump(testRoute), spew.Sdump(route2))
 	}
 }
+
+// deletePayment removes a payment with paymentHash from the payments database.
+func deletePayment(db *DB, paymentHash lntypes.Hash) error {
```
carlaKC: nit: pass `testing.T` into this and just `Fatal` directly if we have an error. Nice way of gatekeeping the function for testing only as well.

lnrpc/rpc.proto
```diff
@@ -2990,6 +2990,13 @@ message Payment {
 
     /// The HTLCs made in attempt to settle the payment [EXPERIMENTAL].
     repeated HTLCAttempt htlcs = 14;
+
+    /**
+    The creation index of this payment. Each payment can be uniquely identified
+    by this index, which is monotonically increasing by creation time but can
+    have gaps due to deleted payments.
```
carlaKC: Technically not due to deleted payments right now; we used to only save completed payments to the db, not failed attempts (😱 ).  This could be a bit misleading to a user, because they haven't deleted any payments, but may see gaps. I would go with, not strictly incremented by 1 for payments made in older versions of lnd, or something like that. If/when we add payment deletion, we can update the comment.

lnrpc/rpc.proto
```diff
@@ -3022,14 +3029,46 @@ message ListPaymentsRequest {
     /**
     If true, then return payments that have not yet fully completed. This means
     that pending payments, as well as failed payments will show up if this
-    field is set to True.
+    field is set to true. This flag doesn't change the meaning of the indices,
+    which are tied to individual payments.
+    */
+    bool include_incomplete = 1 [json_name = "include_incomplete"];
```
carlaKC: Since this PR was opened, we've remove json from `rpc.proto`. They're now just set to the field's name, so you can remove `json_name`...

channeldb/payments.go
bitromortac: Edited comment.

channeldb/payments_test.go
bitromortac: Gatekeeping was exactly what I wanted to achieve, thank you for the hint. Also added `t.Helper()`.

lnrpc/rpc.proto
bitromortac: Right, updated the comment.

lnrpc/rpc.proto
bitromortac: Missed that, good to know!

bitromortac:
Thank you already for the patience @carlaKC and @joostjager :slightly_smiling_face:. I would also leave the first and last index fields for now, if you're not opposed to this. I'll keep on rebasing.

channeldb/payments.go
```diff
@@ -382,6 +382,144 @@ func fetchHtlcFailInfo(bucket kvdb.ReadBucket) (*HTLCFailInfo, error) {
...
+	if len(allPayments) == 0 {
+		return resp, nil
+	}
+
+	indexExclusiveLimit := query.IndexOffset
+	// In backward pagination, if the index limit is the default 0 value,
+	// we set our limit to our highest sequence number + 1, to also include
+	// the highest sequence number since the limit is exclusive.
+	if query.Reversed && indexExclusiveLimit == 0 {
+		indexExclusiveLimit = allPayments[len(allPayments)-1].SequenceNum + 1
```
joostjager, line 460: I think it can also be set to maxint?

channeldb/payments.go
```diff
@@ -382,6 +382,144 @@ func fetchHtlcFailInfo(bucket kvdb.ReadBucket) (*HTLCFailInfo, error) {
...
+	// response. The offset can be used to continue forward pagination.
+	LastIndexOffset uint64
+}
+
+// QueryPayments is a query to the payments database which is restricted
+// to a subset of payments by the payments query, containing an offset
+// index and a maximum number of returned payments.
+func (db *DB) QueryPayments(query PaymentsQuery) (PaymentsQuerySlice, error) {
+	resp := PaymentsQuerySlice{
+		PaymentsQuery: query,
```
joostjager, line 444: Why return the input parameter?

channeldb/payments.go
```diff
@@ -382,6 +382,144 @@ func fetchHtlcFailInfo(bucket kvdb.ReadBucket) (*HTLCFailInfo, error) {
...
+
+	indexExclusiveLimit := query.IndexOffset
+	// In backward pagination, if the index limit is the default 0 value,
+	// we set our limit to our highest sequence number + 1, to also include
+	// the highest sequence number since the limit is exclusive.
+	if query.Reversed && indexExclusiveLimit == 0 {
+		indexExclusiveLimit = allPayments[len(allPayments)-1].SequenceNum + 1
+	}
+
+	var payment *MPPayment
```
joostjager, line 463: Can be declared inside the for loop?

channeldb/payments_test.go
```diff
@@ -175,3 +177,264 @@ func TestRouteSerialization(t *testing.T) {
...
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			db, err := initDB()
+			if err != nil {
+				t.Fatalf("unable to init db: %v", err)
+			}
```
joostjager, line 373: You could argue whether it is necessary to use a real db here instead of mocking `FetchPayments`, but it works.

cmd/lncli/commands.go
```diff
@@ -2171,10 +2171,42 @@ var listPaymentsCommand = cli.Command{
...
+				"excluded. If index_offset is set to zero in " +
+				"reversed mode, the query will end with the " +
+				"last payment made.",
+		},
+		cli.UintFlag{
+			Name: "max_payments",
+			Usage: "the max number of payments to return, by " +
+				"default, all completed payments are returned",
+		},
+		cli.BoolFlag{
```
joostjager, line 2205: Is the default to paginate backwards? I think previously it would always be forwards.

channeldb/payments.go
bitromortac: Yes, that's simpler, changed it.

channeldb/payments.go
bitromortac: Tried to replicate the invoices query (https://github.com/lightningnetwork/lnd/blob/77df8e3a4386326f3b9bb72db336b1b8641886a0/channeldb/invoices.go#L673), but don't want to use this always as a discussion stopper. Indeed, it's used nowhere.

channeldb/payments.go
bitromortac: Done

channeldb/payments_test.go
bitromortac: True, it's a bit heavy.

cmd/lncli/commands.go
bitromortac: Yes, backward pagination is the new default, but it doesn't make a difference compared to the previous output of the command, as before and after the PR all the (completed) payments are returned by default and the order is always oldest to newest payments. In the future when the command is more effective, we could limit the number of returned payments. The motivation as in `listinvoices` and `fwdinghistory` is to be able to quickly look up newest events.

channeldb/payments.go
joostjager: I would remove it. On the rpc interface, it may be worth keeping interfaces the same, but this is just internally used code.

bitromortac: Yes, should have done this already before.

bitromortac:
> Only comment standing from me is to remove the unused InvoiceQuery return parameter. If we need it (seems unlikely to me), we can add it back in.

Removed the initial query parameter form the PaymentsResponse. Thank you again for reviewing!

channeldb/payments.go
```diff
@@ -423,6 +424,140 @@ func fetchHtlcFailInfo(bucket kvdb.ReadBucket) (*HTLCFailInfo, error) {
...
+	if len(allPayments) == 0 {
+		return resp, nil
+	}
+
+	indexExclusiveLimit := query.IndexOffset
+	// In backward pagination, if the index limit is the default 0 value,
+	// we set our limit to maxint to include all payments from the highest
+	// sequence number on.
+	if query.Reversed && indexExclusiveLimit == 0 {
+		indexExclusiveLimit = math.MaxInt64
```
joostjager, line 498: Final comment about this is that this non-functional special casing could also be moved into `rpcserver.go`. For `MaxPayments`, the special case zero is handled there as well.

PR #3961, opened 2020-01-27
channeldb: Indicate cooperative close initator

### LINKED ISSUE #3709: The closed channels list, should return who initiated the channel-close operation close_initiator
### Background

The closed channels list, should return who initiated the channel-close operation close_initiator

### Expected behaviour

show "close_initiator": true/false, like "initiator": true for channels

### Actual behaviour

attribute not present

carlaKC:
Fixes #3709. 

Update channeldb to save channel statuses indicating whether the local or remote party initiated a cooperative close. These values are written to db at the time of cooperative close broadcast, and reloaded in chain watcher when we need to dispatch a cooperative close.


channeldb/channel.go
```diff
@@ -651,13 +651,9 @@ func (c *OpenChannel) hasChanStatus(status ChannelStatus) bool {
 	return c.chanStatus&status == status
 }
 
-// RefreshShortChanID updates the in-memory channel state using the latest
+// RefreshChanInfo updates the in-memory channel state using the latest
 // value observed on disk.
-//
-// TODO: the name of this function should be changed to reflect the fact that
```
wpaulino: I wonder why this wasn't done on the change to refresh all channel state... 🤔

channeldb/channel.go
```diff
@@ -978,14 +992,27 @@ func (c *OpenChannel) MarkCommitmentBroadcasted(closeTx *wire.MsgTx) error {
...
+	localInit bool) error {
+
+	// Choose the channel status based on whether we initiated the
+	// channel close or not.
+	status := ChanStatusRemoteInitCoopBroadcast
+	if localInit {
+		status = ChanStatusLocalInitCoopBroadcast
+	}
+
+	log.Debugf("marking channel: %v cooperatively broadcast "+
```
wpaulino: Nit: capitalize log statement.

contractcourt/chain_watcher.go
```diff
@@ -691,10 +691,36 @@ func (c *chainWatcher) dispatchCooperativeClose(commitSpend *chainntnfs.SpendDet
 	log.Infof("Cooperative closure for ChannelPoint(%v): %v",
 		c.cfg.chanState.FundingOutpoint, spew.Sdump(broadcastTx))
 
+	// Reload the channel state in memory to obtain the latest channel
+	// status. This is required to determine the close initiator of the
+	// channel.
+	if err := c.cfg.chanState.RefreshChanInfo(); err != nil {
```
wpaulino: Perhaps a better place for this would be within `closeObserver` right after we receive the spend notification so that it applies to all cases? I'm not too familiar with all the areas that can modify our channel state, but we should be careful of this change in general.

bhandras:
Overall solid PR, my only wish is that it'd be great to have test for the added functionality, as in the current form tests are only patched.

channeldb/channel.go
carlaKC: ¯\_(ツ)_/¯

contractcourt/chain_watcher.go
carlaKC: Yeah, we seem to follow this pattern in a bunch of places in chainwatcher; if we need a fresh value we re-query db then mutate those values in memory. 

Had a look at this one specifically, and the only fields that would change on this call are `c.IsPending`, `c.chanStatus`, `c.TotalMsatSend` and `c.TotalMsatReceived` because we are only updating chan info. Of these fields, only `chanStatus` is used in `chainwatcher` so I think this should be ok (especially since we don't clear channel statuses, just write more bits). But I will move it up so that it's used across close types 👍

carlaKC:
Don't review right now, want to check out the travis failure.

channeldb/channel.go
```diff
@@ -403,18 +403,30 @@ var (
 
 	// ChanStatusCoopBroadcasted indicates that a cooperative close for this
 	// channel has been broadcasted.
+	// Deprecated: use ChanStatusLocalInitCoopBroadcast or
```
halseth: just a drive by comment, so ignore if this has already been discussed: can't  we add this info (who initiated) under a new key in the db? So we wouldn't have to change the meaning and handling of this flag, and only check this key for the value, if not fall back to a default.

carlaKC: if adding a new key to the db is preferable to using up 2 more bits in channel status then this would definitely be the way to go. Would you say only set this new field for coop? Since we already know the initiator from the close tx in other scenarios.

channeldb/channel.go
```diff
@@ -976,28 +985,33 @@ func (c *OpenChannel) isBorked(chanBucket *bbolt.Bucket) (bool, error) {
...
 // markBroadcasted is a helper function which modifies the channel status of the
 // receiving channel and inserts a close transaction under the requested key,
-// which should specify either a coop or force close.
+// which should specify either a coop or force close. If a non-nil locally coop
+// initiated bool is provided, a closure to save the channel close initiator
+// is created. This is only used for cooperative closes, because the close
+// initiator is known based on the closeTx for other close types.
 func (c *OpenChannel) markBroadcasted(status ChannelStatus, key []byte,
-	closeTx *wire.MsgTx) error {
+	closeTx *wire.MsgTx, localCoopInitiated *bool) error {
```
carlaKC: Pointer to bool is not ideal, could also pass the whole closure in here?

carlaKC:
Updated to add a new field in the db as @halseth suggested. I think it simplifies the PR nicely, and we don't use us 2 of our 3 remaining status bits.

channeldb/channel.go
```diff
@@ -1016,7 +1030,22 @@ func (c *OpenChannel) markBroadcasted(status ChannelStatus, key []byte,
 		}
 	}
 
-	return c.putChanStatus(status, putClosingTx)
+	// If the localCoopInitiated bool is non-nil, generate a closure which
```
bhandras: Imho, it'd be much nicer if you could refactor these tree functions as it's already a bit messy. In this (`markBroadcasted`) function what we really do is create closures which could be helper functions themselves and rather move the locking and the call to `putChanStatus` to `MarkCoopBroadcasted` and `MarkCommitmentBroadcasted`. WDYT?

contractcourt/chain_watcher_test.go
```diff
@@ -575,3 +578,173 @@ func TestChainWatcherLocalForceCloseDetect(t *testing.T) {
...
+			closeType: channeldb.LocalCoopClose,
+		},
+		{
+			name:      "remote init cooperative close",
+			closeType: channeldb.RemoteCoopClose,
+		},
+	}
+
+	for _, test := range tests {
+		test := test
```
bhandras: nit: remove

carlaKC: this is required for the linter (goscope) to address concurrency issues with [parallel tests](https://github.com/golang/go/issues/16586#issuecomment-237355153)

channeldb/channel.go
carlaKC: Yeah I think this is a nicer way of going about it, have updated to remove `markBroadcast` entirely in favor of each function creating their own closures & locking.

lnrpc/rpc.proto
```diff
@@ -1426,6 +1426,8 @@ message ChannelCloseSummary {
         BREACH_CLOSE = 3;
         FUNDING_CANCELED = 4;
         ABANDONED = 5;
+        LOCAL_COOPERATIVE_CLOSE = 6;
```
halseth: same here, make `COOPERATIVE_CLOSE` mean local close? (need to consider whether renaming it could break some users)

lntest/itest/lnd_test.go
```diff
@@ -6401,6 +6401,9 @@ func verifyCloseUpdate(chanUpdate *lnrpc.ChannelEventUpdate,
 				chanUpdate.Type)
 		}
 
+		// If the channel was force closed, check that we have the
```
halseth: 👍

lntest/itest/lnd_test.go
```diff
@@ -6381,8 +6381,8 @@ func subscribeChannelNotifications(ctxb context.Context, t *harnessTest,
 
 // verifyCloseUpdate is used to verify that a closed channel update is of the
 // expected type.
-func verifyCloseUpdate(chanUpdate *lnrpc.ChannelEventUpdate,
-	force bool, forceType lnrpc.ChannelCloseSummary_ClosureType) error {
+func verifyCloseUpdate(chanUpdate *lnrpc.ChannelEventUpdate, force bool,
+	coopType, forceType lnrpc.ChannelCloseSummary_ClosureType) error {
```
halseth: could these three vars be condensed down to just one single `closeType lnrpc.ChannelCloseSummary_ClosureType`?

lnrpc/rpc.proto
carlaKC: Wouldn't that be incorrect for channels closed before this change? 
A channel marked `COOPERATIVE_CLOSE`, which we have repurposed to mean locally cooperatively closed, could actually have been closed by the remote party if was closed before this change. It makes it a bit ambiguous.

halseth: IMO that is okay, since we don't have more information about who closed it anyway. But I could see it being confusing, as suddenly it looks like all your old closes were locally initiated. Do you think it would be a deal breaker?

Just sounds like a maintenance headache to keep three possible coop close types: `local`, `remote` and `unknown`.

bhandras: I'd vote for keeping cooperative/unknown, local, remote to not confuse users.

lntest/itest/lnd_test.go
bhandras: +1

lnrpc/rpc.proto
carlaKC: I think it's a bit of a deal breaker tbh, want to reduce ambiguity not add a newer, more subtle form of it :(

halseth: Could we go about this another way: instead of overloading the close types, what about adding a new field `close_initiator` or similar, that would augment the coop close information. That way we wouldn't have change the meaning of the existing enum.

My main worry about the current approach is that we are deprecating a DB field, and without a migration we will have to keep maintaining  all codepaths where it is used.

carlaKC: So something that is exposed to the user along the lines of: 
`close_initiator` : `LOCAL`/`REMOTE` or `UNKNOWN` for older channels?

Sounds like a nice compromise to me, but we would need to migrate existing closes summaries.

halseth: Yeah, something like that. Why would you need to migrate? We could make it a field that by default is `Unknown` and store the initiator under a new key for new channels.

carlaKC: Discussed offline, close summaries store chanID keys rather than buckets, and have a bunch of optional fields at the end which makes quick addition of a new field tricky. We could add another bucket to `closed-chan-bucket` which has `chanID` buckets where new information can be stored. 

Since this is just a rpc change, opted to just get channel status from the historical channel bucket. If we want to fix up this bucket, it would be worth doing a full migration and getting rid of the optional fields as well, which is out of scope.

channeldb/db_test.go
```diff
@@ -693,3 +693,49 @@ func TestFetchChannels(t *testing.T) {
...
+	// minimum required for this call to run without panicking.
+	if err := channel.CloseChannel(&ChannelCloseSummary{
+		ChanPoint:      channel.FundingOutpoint,
+		RemotePub:      channel.IdentityPub,
+		SettledBalance: btcutil.Amount(500),
+	}); err != nil {
+		t.Fatalf("unexpected error closing channel: %v", err)
+	}
+
+	_, err = cdb.FetchHistoricalChannel(&channel.FundingOutpoint)
```
halseth: sanity check that the returned channel is the one you expect?

rpcserver.go
```diff
@@ -2922,13 +2922,55 @@ func (r *rpcServer) ClosedChannels(ctx context.Context,
...
 	}
 
 	return resp, nil
 }
 
+// getCooperativeCloseInitiator returns booleans which indicate whether the
+// cooperative close initiator status is set on a channel. For non-cooperatively
+// closed channels, false values will be returned (because these statues are
+// only set for cooperatively closed channels.
+func (r *rpcServer) getCooperativeCloseInitiator(
```
halseth: style nit: move to below `createRPCClosedChannel`, where it is used.

channeldb/channel.go
```diff
@@ -401,20 +401,36 @@ var (
...
+	// channels will only have this status set. Newer ones will also have
+	// close initiator information stored using the local/remote initiator
+	// status. This status is set in conjunction with the initiator status
+	// so that we do not need to check multiple channel statues for
+	// cooperative closes.
 	ChanStatusCoopBroadcasted ChannelStatus = 1 << 4
+
+	// ChanStatusLocalCoopInitiated is set in addition to
+	// ChanStatusCoopBroadcasted when we initiate a cooperative close.
+	ChanStatusLocalCoopInitiated ChannelStatus = 1 << 5
```
halseth: cool, I like this :) Looking at the chan statuses, we don't actually store whether we initiate a force close. This could be useful information also, like if we initiate, but the remote force close ends up in the chain. So we could make this a bit more general, and store it also in the cases where we/them force close?

carlaKC: Updated to include force closes. Not quite the same flow for local vs remote force closes because we have some warning for local force where we can mark broadcast and set the initiator status. We only detect remote forces once they've been broadcast, and we don't mark broadcast (because we use that field for rebroadcasting on startup).

carlaKC:
Ready for another look @bhandras!
 
A TL;DR on the changes since you last had a look:
1. Went back to using channel statuses since we are increasing the size of the field
2. The `CloseSummary` bucket is nasty to add fields to, channels are keys not buckets, and have a bunch of optional fields tacked on the end, decided not to use this. If we want to change it, a separate PR with full migration is in order IMO. 
3. Since this is just an aesthetic change for rpc, I decided to use the values in the historical channel bucket, since it is now written for all channels

rpcserver.go
```diff
@@ -3170,7 +3190,77 @@ func createRPCClosedChannel(
...
+	histChan, err := r.server.chanDB.FetchHistoricalChannel(chanPoint)
+	switch err {
+	// The node has upgraded from a version where we did not store
+	// historical channels, and has not closed a channel since. Do
+	// not return an error, initiator values are unknown.
+	case channeldb.ErrNoHistoricalBucket:
+
+	// The channel was closed before we started storing historical
+	// channels. Do  not return an error, initiator values are unknown.
+	case channeldb.ErrChannelNotFound:
```
halseth: return immediately to avoid indentation below

contractcourt/channel_arbitrator_test.go
```diff
@@ -1910,3 +1941,144 @@ func TestChannelArbitratorPendingExpiredHTLC(t *testing.T) {
...
+				newStates: make(chan ArbitratorState,
+					len(test.expectedStates)),
+				resolutions: &ContractResolutions{
+					CommitHash:       chainhash.Hash{},
+					CommitResolution: nil,
+				},
+			}
+
+			chanArbCtx, err := createTestChannelArbitrator(
+				t, log, withMarkClosed(alice.CloseChannel),
```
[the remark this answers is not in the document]
carlaKC: nice, cuts down the test length too 👍

PR #3963, opened 2020-01-27
doc: add Operational Safety Guidelines document

guggero:
This PR adds a document that describes the security/safety mechanisms that are implemented in
`lnd` and a list of established best practices concerning the safe operation of an `lnd` node.


gkrizek:
Big 👍 - very helpful doc

PR #3964, opened 2020-01-27
routing/test: integrated routing test

joostjager:
Adds an integrated routing test of probability extrapolation for untried channels. 

The difference between this test and the existing pathfinding tests, is that this test focuses on the feedback loop from result interpretation via mission control updates and probability estimation back to pathfinding. Improvements like probability extrapolation were previously only validated by reasoning, while this setup makes it possible to assert the improvement in a test and guard it for the future.

### DIFF
--- a/routing/graph.go
+++ b/routing/graph.go
@@ -1,4 +1,104 @@
 package routing
 
-// TODO(roasbeef): abstract out graph to interface
-//  * add in-memory version of graph for tests
+import (
+	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// routingGraph is an abstract interface that provides information about nodes
+// and edges to pathfinding.
+type routingGraph interface {
+	// forEachNodeChannel calls the callback for every channel of the given node.
+	forEachNodeChannel(nodePub route.Vertex,
+		cb func(*channeldb.ChannelEdgeInfo, *channeldb.ChannelEdgePolicy,
+			*channeldb.ChannelEdgePolicy) error) error
+
+	// sourceNode returns the source node of the graph.
+	sourceNode() route.Vertex
+
+	// fetchNodeFeatures returns the features of the given node.
+	fetchNodeFeatures(nodePub route.Vertex) (*lnwire.FeatureVector, error)
+}
+
+// dbRoutingTx is a routingGraph implementation that retrieves from the
+// database.
+type dbRoutingTx struct {
+	graph  *channeldb.ChannelGraph
+	tx     *bbolt.Tx
+	source route.Vertex
+}
+
+// newDbRoutingTx instantiates a new db-connected routing graph. It implictly
+// instantiates a new read transaction.
+func newDbRoutingTx(graph *channeldb.ChannelGraph) (*dbRoutingTx, error) {
+	sourceNode, err := graph.SourceNode()
+	if err != nil {
+		return nil, err
+	}
+
+	tx, err := graph.Database().Begin(false)
+	if err != nil {
+		return nil, err
+	}
+
+	return &dbRoutingTx{
+		graph:  graph,
+		tx:     tx,
+		source: sourceNode.PubKeyBytes,
+	}, nil
+}
+
+// close closes the underlying db transaction.
+func (g *dbRoutingTx) close() error {
+	return g.tx.Rollback()
+}
+
+// forEachNodeChannel calls the callback for every channel of the given node.
+//
+// NOTE: Part of the routingGraph interface.
+func (g *dbRoutingTx) forEachNodeChannel(nodePub route.Vertex,
+	cb func(*channeldb.ChannelEdgeInfo, *channeldb.ChannelEdgePolicy,
+		*channeldb.ChannelEdgePolicy) error) error {
+
+	txCb := func(_ *bbolt.Tx, info *channeldb.ChannelEdgeInfo,
+		p1, p2 *channeldb.ChannelEdgePolicy) error {
+
+		return cb(info, p1, p2)
+	}
+
+	return g.graph.ForEachNodeChannel(g.tx, nodePub[:], txCb)
+}
+
+// sourceNode returns the source node of the graph.
+//
+// NOTE: Part of the routingGraph interface.
+func (g *dbRoutingTx) sourceNode() route.Vertex {
+	return g.source
+}
+
+// fetchNodeFeatures returns the features of the given node. If the node is
+// unknown, assume no additional features are supported.
+//
+// NOTE: Part of the routingGraph interface.
+func (g *dbRoutingTx) fetchNodeFeatures(nodePub route.Vertex) (
+	*lnwire.FeatureVector, error) {
+
+	targetNode, err := g.graph.FetchLightningNode(g.tx, nodePub)
+	switch err {
+
+	// If the node exists and has features, return them directly.
+	case nil:
+		return targetNode.Features, nil
+
+	// If we couldn't find a node announcement, populate a blank feature
+	// vector.
+	case channeldb.ErrGraphNodeNotFound:
+		return lnwire.EmptyFeatureVector(), nil
+
+	// Otherwise bubble the error up.
+	default:
+		return nil, err
+	}
+}
--- a/routing/integrated_routing_context_test.go
+++ b/routing/integrated_routing_context_test.go
@@ -1,0 +1,194 @@
+package routing
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/coreos/bbolt"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// integratedRoutingContext defines the context in which integrated routing
+// tests run.
+type integratedRoutingContext struct {
+	graph *mockGraph
+	t     *testing.T
+
+	source *mockNode
+	target *mockNode
+
+	amt         lnwire.MilliSatoshi
+	finalExpiry int32
+
+	mcCfg          MissionControlConfig
+	pathFindingCfg PathFindingConfig
+}
+
+// newIntegratedRoutingContext instantiates a new integrated routing test
+// context with a source and a target node.
+func newIntegratedRoutingContext(t *testing.T) *integratedRoutingContext {
+	// Instantiate a mock graph.
+	source := newMockNode()
+	target := newMockNode()
+
+	graph := newMockGraph(t)
+	graph.addNode(source)
+	graph.addNode(target)
+	graph.source = source
+
+	// Initiate the test context with a set of default configuration values.
+	// We don't use the lnd defaults here, because otherwise changing the
+	// defaults would break the unit tests. The actual values picked aren't
+	// critical to excite certain behavior, but do need to be aligned with
+	// the test case assertions.
+	ctx := integratedRoutingContext{
+		t:           t,
+		graph:       graph,
+		amt:         100000,
+		finalExpiry: 40,
+
+		mcCfg: MissionControlConfig{
+			PenaltyHalfLife:       30 * time.Minute,
+			AprioriHopProbability: 0.6,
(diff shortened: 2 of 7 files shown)


routing/pathfind.go
```diff
@@ -425,14 +421,12 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	}
 	self := selfNode.PubKeyBytes
 
-	tx := g.tx
-	if tx == nil {
```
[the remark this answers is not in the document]
joostjager: I checked it out and found https://github.com/lightningnetwork/lnd/commit/646f79f566ece3f647acda565a6841375960e84c.

It seems that before that commit, new txes were opened for every query. Reducing that to only a single tx per pathfinding operation would have been sufficient too I'd say.

routing/graph.go
```diff
@@ -1,4 +1,101 @@
...
+
+// fetchNodeFeatures returns the features of the given node. If the node is
+// unknown, assume no additional features are supported.
+//
+// NOTE: Part of the routingGraph interface.
+func (g *dbRoutingGraph) fetchNodeFeatures(nodePub route.Vertex) (
+	*lnwire.FeatureVector, error) {
+
+	targetNode, err := g.graph.FetchLightningNode(g.tx, nodePub)
+	switch {
```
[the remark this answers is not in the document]
joostjager: fixed

routing/pathfind.go
```diff
@@ -380,28 +380,49 @@ func getMaxOutgoingAmt(node route.Vertex, outgoingChan *uint64,
...
+// available bandwidth.
 func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
 	source, target route.Vertex, amt lnwire.MilliSatoshi, finalHtlcExpiry int32) (
 	[]*channeldb.ChannelEdgePolicy, error) {
 
+	dbGraph, err := newDbRoutingGraph(g.graph)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = dbGraph.closeTx() }()
```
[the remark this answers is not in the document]

routing/integrated_routing_context_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// Initiate the test context with default configuration values.
+	ctx := integratedRoutingContext{
+		t:           t,
+		graph:       graph,
+		amt:         100000,
+		finalExpiry: 40,
+
+		mcCfg: MissionControlConfig{
+			PenaltyHalfLife:       30 * time.Minute,
```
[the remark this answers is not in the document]
joostjager: Comment added about this.

routing/integrated_routing_context_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	dbPath := file.Name()
+
+	db, err := bbolt.Open(dbPath, 0600, nil)
+	if err != nil {
+		c.t.Fatal(err)
+	}
+	defer db.Close()
+	defer os.Remove(dbPath)
```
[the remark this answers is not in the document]
joostjager: done

routing/integrated_routing_context_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+			if finalResult != nil {
+				fmt.Printf("final result: %v\n", finalResult)
+				break
+			}
+		}
+
+	}
+
+	fmt.Printf("Payment attempts: %v\n", nextPid)
```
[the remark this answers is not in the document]
joostjager: Good point, removed

routing/integrated_routing_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+		// The channels from intermediate1 all have insufficient balance.
+		g.nodes[intermediate1.pubkey].channels[imNode.pubkey].balance = 0
+	}
+
+	// It is expected that pathfinding will try to explore the routes via
+	// intermediate1 first, because those are free. But as failures happen,
+	// the node probability of intermediate1 will go down in favor of the
+	// paid route via expensiveNode.
+	ctx.testPayment(5)
```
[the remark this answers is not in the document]
joostjager: Comment added

routing/mock_graph_test.go
```diff
@@ -0,0 +1,261 @@
...
+	pubkey := createPubkey()
+	return &mockNode{
+		channels: make(map[route.Vertex]*mockChannel),
+		pubkey:   pubkey,
+	}
+}
+
+// fwd simulates an htlc forward through this node. If the from parameter is
+// nil, this node is considered to be the sender of the payment. The route
+// parameter describes the remaining route from this node onwards. If nil, this
```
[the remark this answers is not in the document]
joostjager: good point, changed

routing/mock_graph_test.go
```diff
@@ -0,0 +1,261 @@
...
+		// Update the incoming balance.
+		inChan.balance += route.amtToFwd
+
+		return htlcResult{}, nil
+	}
+
+	// Check if the outgoing channel has enough balance.
+	outChan, ok := m.channels[next.node.pubkey]
+	if !ok {
+		return htlcResult{}, fmt.Errorf("%v: unknown next %v", m.pubkey, next.node.pubkey)
```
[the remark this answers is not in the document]
joostjager: fixed

joostjager:
>One offhand thought while looking at this was that it would be nice to not need so much logic in the mock, but I do not think it's avoidable.

Yes, I already tried to minimize it by adding the graph interface. I could have returned canned responses from the mock graph and not instantiate those mock nodes, but didn't chose to do so because:
* It is trickier to set up the responses so that they really mimic what would happen on the network in reality. Every attempt needs a response that takes into account the previous fails and successes.
* If we decide to extend the test later, writing down the canned responses may become even more difficult. I am thinking for example of adding timing to it.

routing/integrated_routing_context_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+			// If the payment is successful, the control loop can be
+			// broken out of.
+			break
+		} else {
+			// Failure, update mission control and retry.
+			fmt.Printf("fail: %v @ %v\n", htlcResult.failure, htlcResult.failureSource)
+
+			var failureSourceIndex int
+			for i, h := range route.Hops {
```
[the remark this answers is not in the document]
joostjager: Hmm, then I'd need another struct for the return value. I opted to isolate this into a method. I think that also makes it better.

joostjager:
>Just nits from me, and not going to review the last commit if it'll be removed for the final PR.

Yes, it doesn't need to be reviewed. The intention was just to provide some background on where this is going.

routing/integrated_routing_test.go
```diff
@@ -0,0 +1,57 @@
+package routing
+
+import (
+	"testing"
+)
+
+// TestProbabilityExtrapolation tests that probabilities for tried channels are
+// extrapolated to untried channels. This is a way to improve pathfinding
+// success by steering away from bad nodes.
+func TestProbabilityExtrapolation(t *testing.T) {
```
halseth: this is very nice :)

Would it be possible to have it in its own sub package? Would possibly need to refactor the internal findPath call to export it, but feels like `routing` is big enough as is.

routing/mock_graph_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+// addNode adds the given mock node to the network.
+func (m *mockGraph) addNode(node *mockNode) {
+	m.nodes[node.pubkey] = node
+}
+
+// addChannel adds a new channel between two existing nodes on the network. It
+// sets the channel balance to 50/50%.
+//
+// nolint:unparam
```
halseth: what's the lint failure?

routing/graph.go
```diff
@@ -1,4 +1,101 @@
...
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// routingGraph is an abstract interface that provides information about nodes
+// and edges to pathfinding.
+type routingGraph interface {
+	// forEachNodeChannel calls the callback for every channel of the given node.
+	forEachNodeChannel(nodePub route.Vertex,
```
halseth: Suggestion to further decouple this from channeldb: figure out what information from `ChannelEdgeInfo` and `ChannelEdgePolicy` we need during path finding, and add that to a new type that can be part of this interface.

routing/integrated_routing_test.go
joostjager: It is possible of course, but I don't see much gain from that. We'd make the file listing of `routing` two files shorter, but have to pay for that by exporting symbols that we only need for testing. It is too bad that golang doesn't allow access to internals from test packages.

routing/graph.go
joostjager: I tried it out. It is possible, but there are multiple ways to shape the data structures. I think it is better to leave it as is for now, also because it enlarges the scope of the pr without a direct benefit, and revisit when in-memory pathfinding is implemented. It will then probably be more clear what the best structure is.

routing/graph.go
```diff
@@ -1,4 +1,101 @@
...
+// database.
+type dbRoutingGraph struct {
+	graph *channeldb.ChannelGraph
+	tx    *bbolt.Tx
+}
+
+// newDbRoutingGraph instantiates a new db-connected routing graph. It implictly
+// instantiates a new read transaction.
+func newDbRoutingGraph(graph *channeldb.ChannelGraph) (*dbRoutingGraph, error) {
+	tx, err := graph.Database().Begin(false)
```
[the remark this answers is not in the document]
joostjager: I changed the naming. I think that addresses your concern without changing any code. Let me know what you think of it.

routing/mock_graph_test.go
joostjager: It complained previously because I used this function with always the same channel capacity. But that isn't the case anymore. Removed the linter exception.

routing/pathfind.go
```diff
@@ -380,28 +380,54 @@ func getMaxOutgoingAmt(node route.Vertex, outgoingChan *uint64,
...
+
+	return findPathInternal(
+		g.additionalEdges, g.bandwidthHints, routingTx, r, cfg, source,
+		target, amt, finalHtlcExpiry,
+	)
+}
+
+// findPathInternal is the internal implementation of findPath.
+func findPathInternal(
+	additionalEdges map[route.Vertex][]*channeldb.ChannelEdgePolicy,
```
halseth: why remove `graphParams`?

joostjager:
Async payments itest is consistently failing. Need to see if there is a problem with the routing tx.

routing/pathfind.go
joostjager: `graphParams` contains `channeldb.Graph` and not our abstract interface.

joostjager:
Problem found: `SourceNode()` was now called inside the tx, while previously it was outside. `SourceNode()` creates its own tx and that doesn't work when we are already in a tx.

joostjager:
Dropped final commit that demonstrated testing mpp send. That commit will make its return in the mpp send pr.

routing/pathfind.go
```diff
@@ -415,44 +441,22 @@ func findPath(g *graphParams, r *RestrictParams, cfg *PathFindingConfig,
...
-	}
-	self := selfNode.PubKeyBytes
-
-	// Get a db transaction to execute the graph queries in.
-	tx, err := g.graph.Database().Begin(false)
-	if err != nil {
-		return nil, err
-	}
-	defer tx.Rollback()
+	self := graph.sourceNode()
```
halseth: above comment is no longer accurate and should be moved to where we actually fetch the source node

routing/router.go
```diff
@@ -2311,6 +2311,17 @@ func (r *ChannelRouter) BuildRoute(amt *lnwire.MilliSatoshi,
 		return nil, err
 	}
 
+	routingTx, err := newDbRoutingTx(r.cfg.Graph)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err := routingTx.close()
```
halseth: won't this tx be open much longer than necessary now? It actually looks like we'll do an RPC call inside it below...

routing/pathfind.go
joostjager: comment removed and call moved down to where it is first used

routing/router.go
joostjager: Moved rpc call outside the db tx

PR #3966, opened 2020-01-28
lnwire: create new ExtraOpaqueData type for parsing TLV extensions

Roasbeef:
In this Pr, we create a new `ExtraOpaqueData` based on the field with the same name that's present in all the announcement related messages. In later commits, we'll embed this new type in each message, so we'll have a generic way to add/parse TLV extensions from messages.

We then convert all the current known messages to include this new field, and parse it accordignly. The new `ExtraOpaqueData` also comes with some propsective helper methods for TLV encoding/decoding. I envision these will be used either _within_ the top-level `Encode/Decode` methods, or by callers to populate new (optional) pointer fields that get added in the wire messages similar to the structs in `htlcswitch/hop`.

### DIFF
--- a/channeldb/migration_01_to_11/codec.go
+++ b/channeldb/migration_01_to_11/codec.go
@@ -172,7 +172,8 @@ func WriteElement(w io.Writer, element interface{}) error {
 		}
 
 	case lnwire.Message:
-		if _, err := lnwire.WriteMessage(w, e, 0); err != nil {
+		_, err := lnwire.WriteMessage(w, e, lnwire.ProtocolVersionLegacy)
+		if err != nil {
 			return err
 		}
 
@@ -383,7 +384,7 @@ func ReadElement(r io.Reader, element interface{}) error {
 		*e = bytes
 
 	case *lnwire.Message:
-		msg, err := lnwire.ReadMessage(r, 0)
+		msg, err := lnwire.ReadMessage(r, lnwire.ProtocolVersionLegacy)
 		if err != nil {
 			return err
 		}
--- a/channeldb/migration_01_to_11/migrations.go
+++ b/channeldb/migration_01_to_11/migrations.go
@@ -724,7 +724,8 @@ func MigrateGossipMessageStoreKeys(tx kvdb.RwTx) error {
 
 		// Serialize the message with its wire encoding.
 		var b bytes.Buffer
-		if _, err := lnwire.WriteMessage(&b, msg, 0); err != nil {
+		_, err := lnwire.WriteMessage(&b, msg, lnwire.ProtocolVersionTLV)
+		if err != nil {
 			return err
 		}
 
--- a/channeldb/waitingproof.go
+++ b/channeldb/waitingproof.go
@@ -227,7 +227,8 @@ func (p *WaitingProof) Encode(w io.Writer) error {
 		return err
 	}
 
-	if err := p.AnnounceSignatures.Encode(w, 0); err != nil {
+	err := p.AnnounceSignatures.Encode(w, lnwire.ProtocolVersionTLV)
+	if err != nil {
 		return err
 	}
 
@@ -242,7 +243,7 @@ func (p *WaitingProof) Decode(r io.Reader) error {
 	}
 
 	msg := &lnwire.AnnounceSignatures{}
-	if err := msg.Decode(r, 0); err != nil {
+	if err := msg.Decode(r, lnwire.ProtocolVersionTLV); err != nil {
 		return err
 	}
 
--- a/lnwire/accept_channel.go
+++ b/lnwire/accept_channel.go
@@ -110,6 +110,7 @@ var _ Message = (*AcceptChannel)(nil)
 // This is part of the lnwire.Message interface.
 func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 	return WriteElements(w,
+		pver,
 		a.PendingChannelID[:],
 		a.DustLimit,
 		a.MaxValueInFlight,
@@ -137,6 +138,7 @@ func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 func (a *AcceptChannel) Decode(r io.Reader, pver uint32) error {
 	// Read all the mandatory fields in the accept message.
 	return ReadElements(r,
+		pver,
 		a.PendingChannelID[:],
 		&a.DustLimit,
 		&a.MaxValueInFlight,
--- a/lnwire/accept_channel.go
+++ b/lnwire/accept_channel.go
@@ -92,6 +92,11 @@ type AcceptChannel struct {
 	// and has a length prefix, so a zero will be written if it is not set
 	// and its length followed by the script will be written if it is set.
 	UpfrontShutdownScript DeliveryAddress
+
+	// ExtraData is the set of data that was appended to this message to
+	// fill out the full maximum transport message size. These fields can
+	// be used to specify optional data such as custom TLV fields.
+	ExtraData ExtraOpaqueData
 }
 
 // A compile time check to ensure AcceptChannel implements the lnwire.Message
@@ -104,7 +109,7 @@ var _ Message = (*AcceptChannel)(nil)
 //
 // This is part of the lnwire.Message interface.
 func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
-	return WriteElements(w,
+	err := WriteElements(w,
 		a.PendingChannelID[:],
 		a.DustLimit,
 		a.MaxValueInFlight,
@@ -121,6 +126,11 @@ func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 		a.FirstCommitmentPoint,
 		a.UpfrontShutdownScript,
 	)
+	if err != nil {
+		return err
+	}
+
+	return a.ExtraData.Encode(w)
 }
 
 // Decode deserializes the serialized AcceptChannel stored in the passed
@@ -156,7 +166,8 @@ func (a *AcceptChannel) Decode(r io.Reader, pver uint32) error {
 	if err != nil && err != io.EOF {
 		return err
 	}
-	return nil
+
+	return a.ExtraData.Decode(r)
 }
 
 // MsgType returns the MessageType code which uniquely identifies this message
@@ -172,11 +183,5 @@ func (a *AcceptChannel) MsgType() MessageType {
 //
 // This is part of the lnwire.Message interface.
 func (a *AcceptChannel) MaxPayloadLength(uint32) uint32 {
-	// 32 + (8 * 4) + (4 * 1) + (2 * 2) + (33 * 6)
-	var length uint32 = 270 // base length
-
-	// Upfront shutdown script max length.
-	length += 2 + deliveryAddressMaxSize
-
-	return length
+	return MaxMsgBody
 }
--- a/lnwire/accept_channel.go
+++ b/lnwire/accept_channel.go
@@ -91,7 +91,7 @@ type AcceptChannel struct {
 	// be paid when mutually closing the channel. This field is optional, and
 	// and has a length prefix, so a zero will be written if it is not set
 	// and its length followed by the script will be written if it is set.
-	UpfrontShutdownScript DeliveryAddress
+	UpfrontShutdownScript TypedDeliveryAddress
 
 	// ExtraData is the set of data that was appended to this message to
 	// fill out the full maximum transport message size. These fields can
@@ -109,7 +109,7 @@ var _ Message = (*AcceptChannel)(nil)
 //
 // This is part of the lnwire.Message interface.
 func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
-	err := WriteElements(w,
+	return WriteElements(w,
 		a.PendingChannelID[:],
 		a.DustLimit,
 		a.MaxValueInFlight,
@@ -125,12 +125,8 @@ func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 		a.HtlcPoint,
 		a.FirstCommitmentPoint,
 		a.UpfrontShutdownScript,
+		a.ExtraData,
 	)
-	if err != nil {
-		return err
-	}
-
-	return a.ExtraData.Encode(w)
 }
 
 // Decode deserializes the serialized AcceptChannel stored in the passed
@@ -140,7 +136,7 @@ func (a *AcceptChannel) Encode(w io.Writer, pver uint32) error {
 // This is part of the lnwire.Message interface.
 func (a *AcceptChannel) Decode(r io.Reader, pver uint32) error {
 	// Read all the mandatory fields in the accept message.
-	err := ReadElements(r,
+	return ReadElements(r,
 		a.PendingChannelID[:],
 		&a.DustLimit,
 		&a.MaxValueInFlight,
@@ -155,19 +151,9 @@ func (a *AcceptChannel) Decode(r io.Reader, pver uint32) error {
 		&a.DelayedPaymentPoint,
 		&a.HtlcPoint,
 		&a.FirstCommitmentPoint,
+		&a.UpfrontShutdownScript,
+		&a.ExtraData,
 	)
-	if err != nil {
-		return err
-	}
-
-	// Check for the optional upfront shutdown script field. If it is not there,
-	// silence the EOF error.
-	err = ReadElement(r, &a.UpfrontShutdownScript)
-	if err != nil && err != io.EOF {
-		return err
-	}
-
-	return a.ExtraData.Decode(r)
 }
 
 // MsgType returns the MessageType code which uniquely identifies this message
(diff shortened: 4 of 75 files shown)


lnwire/update_fulfill_htlc.go
```diff
@@ -75,8 +90,7 @@ func (c *UpdateFulfillHTLC) MsgType() MessageType {
 //
 // This is part of the lnwire.Message interface.
 func (c *UpdateFulfillHTLC) MaxPayloadLength(uint32) uint32 {
-	// 32 + 8 + 32
-	return 72
+	return 65533
```
cfromknecht: use max msg size constant? or maybe we just remove it entirely since they all now return the same constant

lnwire/update_add_htlc.go
```diff
@@ -68,29 +73,39 @@ var _ Message = (*UpdateAddHTLC)(nil)
...
 		&c.Amount,
 		c.PaymentHash[:],
 		&c.Expiry,
 		c.OnionBlob[:],
 	)
+	if err != nil {
+		return err
+	}
+
+	return c.ExtraData.Decode(r)
```
cfromknecht: can we add `ExtraOpaqueData` to the codec? then the diff is only two lines for most cases

lnwire/reply_channel_range.go
```diff
@@ -62,15 +81,21 @@ func (c *ReplyChannelRange) Decode(r io.Reader, pver uint32) error {
...
+		c.FirstBlockHeight,
+		c.NumBlocks,
+		c.Complete,
+	)
+	if err != nil {
 		return err
 	}
 
-	if err := WriteElements(w, c.Complete); err != nil {
+	err = encodeShortChanIDs(w, c.EncodingType, c.ShortChanIDs, c.noSort)
```
cfromknecht: does this compile? can return directly

lnwire/reply_channel_range.go
```diff
@@ -96,6 +101,7 @@ func (c *ReplyChannelRange) Encode(w io.Writer, pver uint32) error {
 		return err
 	}
 
+	return c.ExtraData.Encode(w)
```
cfromknecht: ah i see it is added back here

lnwire/extra_bytes.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// If we _do_ have some bytes, then we'll swap out our backing pointer.
+	// This ensures that any struct that embeds this type will properly
+	// store the bytes once this method exits.
+	if len(rawBytes) > 0 {
+		*e = ExtraOpaqueData(rawBytes)
+	}
+
+	return err
```
halseth: style nit: `return nil`

lnwire/extra_bytes.go
```diff
@@ -0,0 +1,77 @@
...
+	*e = ExtraOpaqueData(extraBytesWriter.Bytes())
+
+	return nil
+}
+
+// ExtractRecords attempts to decode any types in the internal raw bytes as if
+// it were a tlv stream. The set of raw parsed types is returned, and any
+// passed records (if found in the stream) will be parsed into the proper
+// tlv.Record.
+func (e *ExtraOpaqueData) ExtractRecords(records ...tlv.Record,
```
halseth: awkward line break

lnwire/extra_bytes_test.go
```diff
@@ -0,0 +1,147 @@
...
+			t.Fatalf("expected %x, got %x", extraData,
+				newBytes)
+			return false
+		}
+
+		return true
+	}
+
+	// We'll make a function to generate random test data. Half of the
+	// time, we'll actually feed in blank byes.
```
halseth: nit: typ "byes"

lnwire/extra_bytes_test.go
```diff
@@ -0,0 +1,147 @@
...
+	quickCfg := &quick.Config{
+		Values: func(v []reflect.Value, r *rand.Rand) {
+
+			var newTestCase testCase
+			if r.Int31()%2 == 0 {
```
halseth: seems like the empty bytes case could be tested once without property based testing

lnwire/accept_channel.go
```diff
@@ -156,7 +166,8 @@ func (a *AcceptChannel) Decode(r io.Reader, pver uint32) error {
 	if err != nil && err != io.EOF {
 		return err
 	}
-	return nil
+
```
halseth: reminder that how to handle the above upfront shutdown must be decided on in the spec

lnwire/update_fulfill_htlc.go
halseth: Would be in favor of removing it!

Roasbeef: Made it into a constant, IMO still useful if you're reeding out messages from a reader other than the underlying brontide transport.

lnwire/extra_bytes_test.go
Roasbeef: Don't think we lose anything by just leaving it as is.

Roasbeef:
I ran into a snag with this PR: in certain places we store a message, but then pack in additional data _after_ the wire message, like the commitment diffs we store for the state machine. The current logic here will attempt to read _all_ data after the base message as extra TLV data, so it ends up consuming all the extra bytes, leading to an EOF error when we attempt to read information serialized _after_ the wire message. 

A few solutions come to mind:
  1. Do a db migration to prepend a length in front of all wire messages we store in the database. This resolves things as we'd read the entire blob, _then_ pass it into the relevant `Decode` method. 
  2. Add another serialization that includes a length prefix in front of the extra data. This also requires a migration to re-serialize all the messages on disk. 
  3. Update the _spec_ to dictate that all TLV extensions include a length prefix. This resolves the issue as we'd stop reading (zero bytes length) at the end of the messages. 

Thoughts?

halseth:
1. sounds like the easiest one

Roasbeef:
#1 has been done in: #4099.

halseth:
> #1

Trip down memory lane

Roasbeef:
Here's a combined branch with the migration PR to show everything passes with both applied: https://github.com/Roasbeef/lnd/pull/new/combined-wire-tlv

The tests won't pass for this PR in isolation, as since the new wire parsing tries to read until EOF, it can read any of the HTLC in the commitment diff as an example.

Roasbeef:
Pushed up a new rebased version of https://github.com/Roasbeef/lnd/pull/new/combined-wire-tlv

Roasbeef:
> Alternatively, I can instead modify #3966 to actually use the protocol version (pver) argument of lnwire.[Read/Write]Message. 

Updated this PR to add this explicit version. It'll need to be updated further once #4099 lands since the _new_ migration needs to be forced to use the legacy protocol version when decoding wire messages before migrating the data.

Roasbeef:
The combined branch has been updated: https://github.com/Roasbeef/lnd/pull/new/combined-wire-tlv. Only the last commit we new (which'll eventually be in this PR after the dep PR is merged): https://github.com/lightningnetwork/lnd/commit/633cf78a9d2fd7d5034c314d88d2adcaea58fb1a.

lnwire/extra_bytes_test.go
halseth: You lose covereage? Since you spend half of your test cases testing the same thing.

lnwire/node_announcement.go
```diff
@@ -159,6 +161,9 @@ func (a *NodeAnnouncement) DataToSign() ([]byte, error) {
 	// We should not include the signatures itself.
 	var w bytes.Buffer
 	err := WriteElements(&w,
+		// We always use the modern protocol version as we need to
+		// include all data for forawrds compatability.
```
halseth, line 165: typo: forawrds

lnwire/reply_channel_range.go
```diff
@@ -96,12 +98,14 @@ func (c *ReplyChannelRange) Encode(w io.Writer, pver uint32) error {
...
-	err = encodeShortChanIDs(w, c.EncodingType, c.ShortChanIDs, c.noSort)
+	err = encodeShortChanIDs(
+		w, c.EncodingType, c.ShortChanIDs, c.noSort, pver,
+	)
 	if err != nil {
 		return err
 	}
 
-	return c.ExtraData.Encode(w)
+	return c.ExtraData.Encode(w, pver)
```
halseth, line 108: can use `WriteElements`

lnwire/node_announcement.go
```diff
@@ -159,6 +161,9 @@ func (a *NodeAnnouncement) DataToSign() ([]byte, error) {
...
+		ProtocolVersionTLV,
```
halseth, line 166: same, makes more sense to take pver as method argument?

channeldb/migration_01_to_11/migrations.go
```diff
@@ -724,7 +724,8 @@ func MigrateGossipMessageStoreKeys(tx kvdb.RwTx) error {
 
 		// Serialize the message with its wire encoding.
 		var b bytes.Buffer
-		if _, err := lnwire.WriteMessage(&b, msg, 0); err != nil {
+		_, err := lnwire.WriteMessage(&b, msg, lnwire.ProtocolVersionTLV)
```
halseth, line 727: why TLV here?

channeldb/waitingproof.go
```diff
@@ -242,7 +243,7 @@ func (p *WaitingProof) Decode(r io.Reader) error {
 	}
 
 	msg := &lnwire.AnnounceSignatures{}
-	if err := msg.Decode(r, 0); err != nil {
+	if err := msg.Decode(r, lnwire.ProtocolVersionTLV); err != nil {
```
halseth, line 246: does this work since this message never had a TLV blob?

lnwire/lnwire.go
```diff
@@ -854,10 +859,10 @@ func ReadElement(r io.Reader, element interface{}) error {
 
 // ReadElements deserializes a variable number of elements into the passed
 // io.Reader, with each element being deserialized according to the ReadElement
-// function.
-func ReadElements(r io.Reader, elements ...interface{}) error {
+// function. The passed protocol version may affect how the items are encoded.
+func ReadElements(r io.Reader, pver uint32, elements ...interface{}) error {
```
joostjager, line 863: If I understand it correctly, the propagation of `pver` and creation of a new version is only required for the migrations and a new node will never use the legacy version. I don't understand why we don't continue the same reasoning as before and copy over `lnwire`. In the past it was said that the copy is cheap and safe and not an issue. With the currently proposed change, we will always need to be careful not to break a migration still.

channeldb/migration_01_to_11/codec.go
```diff
@@ -172,7 +172,8 @@ func WriteElement(w io.Writer, element interface{}) error {
 		}
 
 	case lnwire.Message:
-		if _, err := lnwire.WriteMessage(w, e, 0); err != nil {
+		_, err := lnwire.WriteMessage(w, e, lnwire.ProtocolVersionLegacy)
```
joostjager, line 175: In https://github.com/lightningnetwork/lnd/commit/633cf78a9d2fd7d5034c314d88d2adcaea58fb1a, doesn't this change in `channeldb/migration18/enclosed_codec.go` need to be `ProtocolVersionTLV` to migrate into the new format?

lnwire/revoke_and_ack.go
```diff
@@ -30,11 +30,18 @@ type RevokeAndAck struct {
...
+	// fill out the full maximum transport message size. These fields can
+	// be used to specify optional data such as custom TLV fields.
+	ExtraData ExtraOpaqueData
 }
 
 // NewRevokeAndAck creates a new RevokeAndAck message.
 func NewRevokeAndAck() *RevokeAndAck {
-	return &RevokeAndAck{}
+	return &RevokeAndAck{
+		ExtraData: make([]byte, 0),
```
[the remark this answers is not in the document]
Roasbeef: Deep equal weirdness.

lnwire/error.go
```diff
@@ -123,8 +123,7 @@ func (c *Error) MsgType() MessageType {
 //
 // This is part of the lnwire.Message interface.
 func (c *Error) MaxPayloadLength(uint32) uint32 {
-	// 32 + 2 + 65501
-	return 65535
+	return MaxMsgBody
```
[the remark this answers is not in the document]
Roasbeef: It can't "fit" given that the error message is of dynamic length.

lnwire/accept_channel.go
```diff
@@ -155,19 +151,9 @@ func (a *AcceptChannel) Decode(r io.Reader, pver uint32) error {
 		&a.DelayedPaymentPoint,
 		&a.HtlcPoint,
 		&a.FirstCommitmentPoint,
+		&a.UpfrontShutdownScript,
```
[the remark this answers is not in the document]
Roasbeef: Not sure what you mean, it's done like this to keep the "hacked" extensions from the "pure" set of TLVs. Or you mean combine this with the TLV addr? Need to double check , but I think it's intended to be a standalone TLV stream.

lnwire/channel_announcement.go
```diff
@@ -126,6 +128,10 @@ func (a *ChannelAnnouncement) DataToSign() ([]byte, error) {
 	// We should not include the signatures itself.
 	var w bytes.Buffer
 	err := WriteElements(&w,
+		// We always use the modern protocol version here as we always
+		// need to include any optional data in the signature digest
+		// for forwards compatibility.
+		ProtocolVersionTLV,
```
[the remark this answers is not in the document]
Roasbeef: FWIW there're no other changes to the digest in sight, in that this is the most correct future looking version, not sure we need to make this a method and make this diff even larger.

lnwire/channel_update.go
```diff
@@ -142,12 +143,12 @@ func (a *ChannelUpdate) Decode(r io.Reader, pver uint32) error {
...
 	// Now check whether the max HTLC field is present and read it if so.
 	if a.MessageFlags.HasMaxHtlc() {
-		if err := ReadElements(r, &a.HtlcMaximumMsat); err != nil {
+		if err := ReadElements(r, pver, &a.HtlcMaximumMsat); err != nil {
 			return err
 		}
 	}
 
-	return a.ExtraOpaqueData.Decode(r)
+	return a.ExtraOpaqueData.Decode(r, pver)
```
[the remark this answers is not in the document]
Roasbeef: This is more explicit IMO, it also has to go _after_ the optional max HTLC read above.

lnwire/channel_reestablish.go
```diff
@@ -145,11 +148,11 @@ func (a *ChannelReestablish) Decode(r io.Reader, pver uint32) error {
 	// We'll conclude by parsing out the commitment point. We don't check
 	// the error in this case, as it they included the commit secret, then
 	// they MUST also include the commit point.
-	if err = ReadElement(r, &a.LocalUnrevokedCommitPoint); err != nil {
+	if err = ReadElement(r, pver, &a.LocalUnrevokedCommitPoint); err != nil {
 		return err
 	}
 
-	return a.ExtraData.Decode(r)
+	return a.ExtraData.Decode(r, pver)
```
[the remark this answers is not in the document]
Roasbeef: Same as above.

channeldb/migration_01_to_11/migrations.go
Roasbeef: Hmm may have been a mistake during the rebase. Looking at it with fresh eyes though, it needs to ensure we write any extra bytes that may have been stored with the announcement.

channeldb/waitingproof.go
Roasbeef: IIRC, `ioutil.ReadAll` will return `nil` if nothing was read.

lnwire/extra_bytes.go
```diff
@@ -0,0 +1,84 @@
...
+// of which we may not actually know how to iterate or parse. By holding onto
+// this data, we ensure that we're able to properly validate the set of
+// signatures that cover these new fields, and ensure we're able to make
+// upgrades to the network in a forwards compatible manner.
+type ExtraOpaqueData []byte
+
+// Encode attempts to encode the raw extra bytes into the passed io.Writer.
+func (e *ExtraOpaqueData) Encode(w io.Writer) error {
+	eBytes := []byte((*e)[:])
+	if err := WriteElements(w, eBytes); err != nil {
```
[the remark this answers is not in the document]
Roasbeef: It's a low level API, so it just accepts raw bytes. Many of these areas don't have any TLV semantics defined yet. When they do utilize the field, they can modify the API accordingly. This is a generic prep PR.

lnwire/lnwire.go
Roasbeef: You're suggesting copying over the entirety of `lnwire` for this minor change to conditionally continue to read bytes at the end of a message (the way it should've been all along really)? I don't think the bloat is worth it, the migration tests should also be such that if this section does break, then a migration test breaks as it runs into an EOF error.

channeldb/migration_01_to_11/codec.go
Roasbeef: You mean some portion of a commit we partially undone? They may have been the case, the end state is correct though. In any case this PR is going away in favor of a mega combined version so it can be tested all at once properly.

Roasbeef:
This PR is going away now (in favor of the combined version), but I pushed up a rebased version for posterity.

PR #3972, opened 2020-01-29
routing: extend trackpayment with htlc events

joostjager:
Currently there is inconsistency between the payment status as reported via `ListPayment` and the status notified through `routerrpc.SendPayment/TrackPayment`. With this PR, the same full payment struct is returned by both rpc calls.

In addition to that, `routerrpc` notifications are now sent out for every change to the payment. This will allow users to track progress of their (multi-parts) payments, either via `lncli` (use the newly added `--show_inflight` flag) or via graphical user interfaces/apps if support is added there. Previously only the final state of the payment was reported.

Note: The `HTLCAttempt_IN_FLIGHT` update will be sent out before the htlc is actually in flight on the link. Launching a `QueryRoutes` at that point will not see the balance change yet. This is relevant for applications that use the output of `TrackPayment` to implement their own payment loop.


joostjager:
Ready for concept ack

Todo: add `ConcurrentQueue` for notifications, to prevent rpc users from blocking the payment process

joostjager:
`ConcurrentQueue` added. Also used multimutex to prevent a global lock around db tx and notification. Initially I had the global lock, but that also caused the async payments benchmark to fail. Probably too slow.

routing/control_tower.go
```diff
@@ -205,82 +217,76 @@ func (p *controlTower) FetchInFlightPayments() ([]*channeldb.InFlightPayment, er
...
-		return false, nil, err
+		return nil, err
 	}
 
-	var event PaymentResult
+	subscriber := newControlTowerSubscriber()
 
-	switch payment.Status {
+	// Always write current payment state to the channel.
+	subscriber.queue.ChanIn() <- payment
```
Roasbeef, line 235: Ah ok scratch by prior comment.

routing/control_tower.go
```diff
@@ -205,82 +217,76 @@ func (p *controlTower) FetchInFlightPayments() ([]*channeldb.InFlightPayment, er
...
-	delete(p.subscribers, paymentHash)
 	p.subscribersMtx.Unlock()
 
-	// Notify all subscribers of the event. The subscriber channel is
-	// buffered, so it cannot block here.
+	// If the payment reached a terminal state, the subscriber list can be
+	// cleared. There won't be any more updates.
+	terminal := event.Status != channeldb.StatusInFlight
+	if terminal {
+		delete(p.subscribers, paymentHash)
```
Roasbeef, line 272: Access to the map without holding the mutex above. The mutex should continue to be held till the loop below.

cmd/lncli/cmd_pay.go
```diff
@@ -405,6 +413,10 @@ func sendPaymentRequest(ctx *cli.Context,
 
 			return nil
 		}
+
+		if showInflight {
```
Roasbeef, line 417: Nice, this was something I was missing when I was testing things out late last week. I ended up resorting to piping the final output of `payinvoice` into a file and doing `tail -f` on `lnd`'s logs.

routing/control_tower.go
```diff
@@ -203,82 +217,76 @@ func (p *controlTower) FetchInFlightPayments() ([]*channeldb.InFlightPayment, er
...
+	}
+
+	// Notify all subscribers of the event.
+	log.Infof("Notifying %v to %v listeners", event, len(list))
```
Roasbeef, line 276: This log is spammy as it attempts to format a raw struct.

routing/payment_lifecycle.go
```diff
@@ -678,7 +678,7 @@ func (p *shardHandler) handleSendError(attempt *channeldb.HTLCAttemptInfo,
 		return nil
 	}
 
-	log.Debugf("Payment %x failed: final_outcome=%v, raw_err=%v",
+	log.Debugf("Payment %v failed: final_outcome=%v, raw_err=%v",
```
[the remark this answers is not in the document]
Roasbeef: There's another instance of this above in `sendPaymentAttempt`.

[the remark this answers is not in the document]
joostjager: I see that there are quite a few more. Changed them all. Probably caused by change of the hash type to `lntypes.Hash` in the previous mpp pr

lnrpc/routerrpc/router.proto
```diff
@@ -465,15 +409,15 @@ service Router {
     /**
     SendPayment attempts to route a payment described by the passed
     PaymentRequest to the final destination. The call returns a stream of
-    payment status updates.
+    payment updates.
     */
-    rpc SendPayment (SendPaymentRequest) returns (stream PaymentStatus);
+    rpc SendPayment (SendPaymentRequest) returns (stream lnrpc.Payment);
```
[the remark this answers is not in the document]
joostjager: Yes. Previously there wasn't really an incremental update other than the payment status code. No in-flight htlcs were ever streamed back, only the final set.

I thought about sending back a delta containing the changed fields only, but it may get a little complicated to use. The caller needs to correctly apply those deltas to get the full payment state. In the end, it is an optimization for which I think it isn't the time yet.

lnrpc/routerrpc/router_server.go
```diff
@@ -450,95 +450,37 @@ func (s *Server) trackPayment(paymentHash lntypes.Hash,
 	case err != nil:
 		return err
 	}
+	defer subscription.Close()
```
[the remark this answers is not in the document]
joostjager: We still send it, but at the `ControlTower` side. See code comment `// Always write current payment state to the channel.`

routing/control_tower.go
joostjager: Oops, fixed

routing/control_tower.go
joostjager: Removed, debugging left-over

cmd/lncli/cmd_pay.go
joostjager: Yes, there is probably still a long tail of mpp diagnostics to be expected.

lnrpc/routerrpc/router_backend.go
```diff
@@ -1158,7 +1158,21 @@ func (r *RouterBackend) MarshallPayment(payment *channeldb.MPPayment) (
 		PaymentRequest:  string(payment.Info.PaymentRequest),
 		Status:          status,
 		Htlcs:           htlcs,
-	}, nil
+		FailureReason:   lnrpc.Payment_NONE,
+	}
+
+	// Marshall failure reason if any.
+	if payment.FailureReason != nil {
```
halseth, line 1165: style nit: do this above, before creating the final `lnrpc.Payment`

lnrpc/rpc.proto
```diff
@@ -3102,6 +3102,42 @@ message Payment {
 
     /// The HTLCs made in attempt to settle the payment [EXPERIMENTAL].
     repeated HTLCAttempt htlcs = 14;
+
+    enum PaymentFailureReason {
+        /**
+        Payment isn't failed (yet)
```
halseth, line 3108: nit: missing period

queue/queue.go
```diff
@@ -88,6 +95,22 @@ func (cq *ConcurrentQueue) start() {
...
+			case cq.chanOut <- nextElement.Value:
+				cq.overflow.Remove(nextElement)
+			case <-cq.quit:
+				return
+			}
+			nextElement = cq.overflow.Front()
+		}
+
+		// Close outgoing channel.
+		close(cq.chanOut)
```
halseth, line 113: maybe this should be done on quit also?

queue/queue.go
```diff
@@ -58,14 +58,18 @@ func (cq *ConcurrentQueue) start() {
...
 			nextElement := cq.overflow.Front()
 			if nextElement == nil {
 				// Overflow queue is empty so incoming items can be pushed
 				// directly to the output channel. If output channel is full
 				// though, push to overflow.
 				select {
-				case item := <-cq.chanIn:
+				case item, ok := <-cq.chanIn:
+					if !ok {
+						break readLoop
```
halseth, line 71: 👍

multimutex/hash_mutex.go
```diff
@@ -0,0 +1,90 @@
...
+
+// HashMutex is a struct that keeps track of a set of mutexes with a given hash.
+// It can be used for making sure only one goroutine gets given the mutex per
+// hash.
+type HashMutex struct {
+	// mutexes is a map of hashes to a cntMutex. The cntMutex for
+	// a given hash will hold the mutex to be used by all
+	// callers requesting access for the hash, in addition to
+	// the count of callers.
+	mutexes map[lntypes.Hash]*cntMutex
```
halseth, line 18: can't we just take 4 bytes of the hash and use the existing `uint64` mutex? Alternatively use `interface{}` as the key type.

routing/control_tower.go
```diff
@@ -52,43 +53,63 @@ type ControlTower interface {
 	// SubscribePayment subscribes to updates for the payment with the given
 	// hash. It returns a boolean indicating whether the payment is still in
 	// flight and a channel that provides the final outcome of the payment.
-	SubscribePayment(paymentHash lntypes.Hash) (bool, chan PaymentResult,
+	SubscribePayment(paymentHash lntypes.Hash) (*ControlTowerSubscriber,
```
halseth, line 56: godoc needs update

routing/control_tower.go
```diff
@@ -52,43 +53,63 @@ type ControlTower interface {
...
 }
 
-// PaymentResult is the struct describing the events received by payment
-// subscribers.
-type PaymentResult struct {
-	// Success indicates whether the payment was successful.
-	Success bool
+// ControlTowerSubscriber contains the state for a payment update subscriber.
+type ControlTowerSubscriber struct {
+	// Updates it the channel over which *channeldb.MPPayment updates can be
```
halseth, line 62: typo

routing/control_tower.go
```diff
@@ -107,8 +128,18 @@ func (p *controlTower) InitPayment(paymentHash lntypes.Hash,
 func (p *controlTower) RegisterAttempt(paymentHash lntypes.Hash,
 	attempt *channeldb.HTLCAttemptInfo) error {
 
-	_, err := p.db.RegisterAttempt(paymentHash, attempt)
-	return err
+	p.paymentsMtx.Lock(paymentHash)
```
halseth, line 131: Nice solution 👍

routing/control_tower.go
```diff
@@ -205,82 +217,75 @@ func (p *controlTower) FetchInFlightPayments() ([]*channeldb.InFlightPayment, er
...
-	return false, c, nil
+	return subscriber, nil
 }
 
-// notifyFinalEvent sends a final payment event to all subscribers of this
+// notifySubscribers sends a final payment event to all subscribers of this
 // payment. The channel will be closed after this.
-func (p *controlTower) notifyFinalEvent(paymentHash lntypes.Hash,
-	event *PaymentResult) {
+func (p *controlTower) notifySubscribers(paymentHash lntypes.Hash,
```
halseth, line 256: add do doc that it should be called with the multimutex held?

multimutex/hash_mutex.go
joostjager: Interesting suggestion. That will work, unless people use non-random hashes (for invoices that are never settled)? There is some cost in terms of understandability. Not sure if it weighs up.

halseth: Worst case it will just have to wait a bit longer before accessing the DB in those cases. I think it is nice to avoid the code duplication, as long as we don't have generics.

Do avoid the scenario you mention, could also a add a simple method
```go
func (h Hash) Uint64Key() uint64 {
      return uint64(sha256(h[:])[0:4])
```
but maybe that is taking it too far

joostjager: I will leave this thread open for others to comment on.

lnrpc/routerrpc/router_backend.go
joostjager: fixed

lnrpc/rpc.proto
joostjager: fixed

queue/queue.go
joostjager: I don't think so. It would signal a clean end to the queue, which isn't the case for quit

routing/control_tower.go
joostjager: updated

routing/control_tower.go
joostjager: fixed

routing/control_tower.go
joostjager: I knew you'd like it, as the original inventor of multi-mutexing :)

routing/control_tower.go
joostjager: added

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
alexbosworth: Can totally changing response messages be avoided?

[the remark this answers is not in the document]
joostjager: It can be avoided, but because we are starting the move to mainlining `routerrpc` now that would be a missed chance to make the api consistent.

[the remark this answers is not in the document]
alexbosworth: Sure, I wish that there were a better upgrade path but it seems extremely disruptive to me and potentially there are less disruptive ways to go about it that still allow for a cleaner future API direction

multimutex/hash_mutex.go
cfromknecht: no strong preference fro me, tho i think a hash map would be a useful addition to `multimutex` as a library

lnrpc/routerrpc/router_backend.go
```diff
@@ -1143,7 +1143,19 @@ func (r *RouterBackend) MarshallPayment(payment *channeldb.MPPayment) (
 	paymentHash := payment.Info.PaymentHash
 	creationTimeNS := MarshalTimeNano(payment.Info.CreationTime)
 
-	return &lnrpc.Payment{
+	// Marshall failure reason if any.
+	failureReason := lnrpc.Payment_NONE
```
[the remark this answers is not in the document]
cfromknecht: perhaps `lnrpc.PaymentStatus_NONE`?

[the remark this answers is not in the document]
joostjager: Failure reason is an embedded enum. I don't think I can get that to be generated as `lnrpc.PaymentStatus_NONE`. Or `lnrpc.FailureReason_NON`. Do you want me to move it to the global scope that get the naming right?

channeldb/payment_control.go
```diff
@@ -186,46 +186,47 @@ func (p *PaymentControl) InitPayment(paymentHash lntypes.Hash,
...
+		return nil, err
 	}
 	htlcInfoBytes := a.Bytes()
 
 	htlcIDBytes := make([]byte, 8)
 	binary.BigEndian.PutUint64(htlcIDBytes, attempt.AttemptID)
 
-	return kvdb.Batch(p.db.Backend, func(tx kvdb.RwTx) error {
+	var payment *MPPayment
+	err = kvdb.Batch(p.db.Backend, func(tx kvdb.RwTx) error {
```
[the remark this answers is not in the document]
joostjager: I don't know. My stand point is to remove all `Batch` calls because it is too early to optimize for high payments/sec rates.

routing/control_tower.go
```diff
@@ -50,45 +51,65 @@ type ControlTower interface {
...
+}
 
-	// FailureReason is a failure reason code indicating the reason the
-	// payment failed. It is only set for failed payments.
-	FailureReason channeldb.FailureReason
+// Close signals that the subscriber is no longer interested in updates.
+func (s *ControlTowerSubscriber) Close() {
+	// Close quit channel so that any pending writes to the queue are
+	// cancelled.
+	close(s.quit)
```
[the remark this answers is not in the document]
joostjager: But wait for what? There are no go-routines in control tower.

routing/control_tower.go
```diff
@@ -201,86 +213,81 @@ func (p *controlTower) FetchInFlightPayments() ([]*channeldb.InFlightPayment, er
...
 	list, ok := p.subscribers[paymentHash]
 	if !ok {
 		p.subscribersMtx.Unlock()
 		return
 	}
-	delete(p.subscribers, paymentHash)
+
+	// If the payment reached a terminal state, the subscriber list can be
+	// cleared. There won't be any more updates.
+	terminal := event.Status != channeldb.StatusInFlight
```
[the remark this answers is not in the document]
joostjager: If you subscribe to a non-existing payment (status unknown), the call will fail.

channeldb/payment_control.go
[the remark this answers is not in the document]
cfromknecht: i agree

lnrpc/routerrpc/router_backend.go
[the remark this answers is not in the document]
cfromknecht: yes was i think moving to global, i'm not much a fan of embedded enums anyway

routing/control_tower.go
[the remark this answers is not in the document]
cfromknecht: okay got it, just rare that we don't lol

routing/control_tower.go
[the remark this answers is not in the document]
cfromknecht: yes you're right, okay 👍

lnrpc/routerrpc/router_backend.go
```diff
@@ -1143,7 +1143,19 @@ func (r *RouterBackend) MarshallPayment(payment *channeldb.MPPayment) (
...
+	if payment.FailureReason != nil {
+		var err error
+		failureReason, err = marshallPaymentFailureReason(
```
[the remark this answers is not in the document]
joostjager: done

lnrpc/routerrpc/router_backend.go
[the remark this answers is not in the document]
joostjager: moved to global scope. too bad that those long prefixes are unavoidable

multimutex/hash_mutex.go
joostjager: leaving hash map in then

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
halseth: I feel it could be useful to wrap the return value in one of our standard return types (`SendPaymentResponse`). This would give use more flexibility going forward adding more data to it, independent of `TrackPayment`. 

`TrackPayment` would similarly have a `TrackPaymentResponse` that for now just wraps the `lnrpc.Payment`. Is there any reason at all that the two calls must have the same response type?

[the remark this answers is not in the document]
joostjager: `TrackPayment` is basically a streaming version of `ListPayments`. I can even see a future where both are integrated. In `ListPayments` we also return a list of plain `Payment` messages. `SendPayment` is a convenience method that combines initiating the payment and following up with `TrackPayment`.

I am wary of this kind of future proofing, but not strongly opposed to it. If another reviewer agrees that it is an improvement, I can add it. @cfromknecht @Roasbeef

joostjager:
This PR replaces the messages in the stream of `SendPayment` and `TrackPayment` with a completely different message. Running an application that isn't updated against this new rpc yields:

`grpc: failed to unmarshal the received message unexpected EOF`

alexbosworth:
> This PR replaces the messages in the stream of `SendPayment` and `TrackPayment` with a completely different message. Running an application that isn't updated against this new rpc yields:
> 
> `grpc: failed to unmarshal the received message unexpected EOF`

That's pretty rough, I it is because it is not a new RPC, it's an existing RPC that was overwritten in a way that gRPC was not designed to support

Roasbeef:
> That's pretty rough

That's perfectively acceptable as this is an experimental RPC server, which was only envisioned to only near a final form after MPP/AMP was implemented, which is now. Compare this to the external sendpayment and llist/subscribe invoice calls which haven't changed significantly since they were created.

alexbosworth:
By rough I mean that it's more difficult to deal with as a consumer. So if you were building on this, you will now have a harder job adapting to the change than otherwise might be possible

Issue #3688, opened 2019-11-07
utxonursery: delete


joostjager:
The on-chain sweeping sub-system 'utxo nursery' is now considered legacy code that needs to be phased out over time. Part of the functionality has already been ported to other sub-systems.

A sweeper sub-system was introduced to publish the chain transactions in #1960. This marked the beginning of the end of the nursery.

PR #2407 removed the nursery dependency for sweeps of direct commitment outputs for txes published by the remote party.

PR #3644 furthers this work by also bypassing nursery for locally published commitment txes.

This issue serves as a tracking issue for the remaining changes required:

* Claiming an htlc with the preimage on a locally published commit tx (`htlcSuccessResolver`).
* Timing out an htlc and reclaiming the funds (`htlcTimeoutResolver`)

When done, all code from `UtxoNursery` and `NurseryStore` can be removed from the code base.

joostjager:
One potential complication of converting the last two resolvers to not use nursery anymore is that on upgrade there may be pending resolvers. Those resolver may already have handed off sweeps to the nursery and those may then interfere with the new resolver code. Or there may be transactions in the mempool or in a block that originate from the nursery. These may not be expected by the new resolvers.

One way to address this is by combining all the following steps:
* Convert the remaining resolvers (`htlcTimeoutResolver` and `htlcSuccessResolver`) at the same time. That way, the nursery can be deleted completely and there is no chance of active interference of the nursery with the new style resolvers.
* Implement the new resolvers in a stateless way. Discard any persisted state and only rely on the chain to infer the current state. 
* To prevent prior nursery sweep txes from not being recognized as our own, carry out a light 'migration' that inserts tx ids of nursery transactions into the sweeper's "our txes" set.

I think that with these steps, the switch over will be safe and also not too complicated to reason about.

saubyk:
cc: @yyforyongyu is this issue relevant still?

yyforyongyu:
yeah we still need to remove it, can be tracked as the last task in #8680

saubyk:
Closing this issue as the ask is covered with the list of requirements in #8680

PR #3977, opened 2020-02-03
cnct: remove nursery dependency from success resolver

joostjager:


contractcourt/htlc_success_resolver.go
```diff
@@ -99,159 +93,108 @@ func (h *htlcSuccessResolver) Resolve() (ContractResolver, error) {
...
-		)
-		if err != nil {
+	// Wait for the sweep result.
+	select {
+	case result, ok := <-resultChan:
+		if !ok {
+			return nil, errResolverShuttingDown
+		}
+		if result.Err != nil {
 			return nil, err
```
chokoboko: Shouldn't this return result.Err?

joostjager: Nice catch, thank you. Fixed.

joostjager:
Todo: figure out migration path when nursery is already trying to sweep

joostjager:
For migration, it is probably enough to just run https://github.com/lightningnetwork/lnd/blob/2cd26d75564345c0c37fa83785ebe8e4f1fb8f28/sweep/store.go#L100 again so that sweeper recognizes nursery transactions as its own.

Other than that, the (new) resolvers are stateless and will pick up any state the resolution process was left in.

joostjager:
Code got outdated because we didn't prioritize this. Closing PR now, because priority is still low. If that changes in the future, it can still be found as a reference on the parent issue #3688.

PR #3980, opened 2020-02-04
invoices: fix htlc timer deadlock

joostjager:
Fixes a possible deadlock caused by sending on a channel inside a lock.

### DIFF
--- a/invoices/invoiceregistry.go
+++ b/invoices/invoiceregistry.go
@@ -693,8 +693,7 @@ func (i *InvoiceRegistry) cancelSingleHtlc(hash lntypes.Hash,
 
 // processKeySend just-in-time inserts an invoice if this htlc is a keysend
 // htlc.
-func (i *InvoiceRegistry) processKeySend(ctx invoiceUpdateCtx,
-	hash lntypes.Hash) error {
+func (i *InvoiceRegistry) processKeySend(ctx invoiceUpdateCtx) error {
 
 	// Retrieve keysend record if present.
 	preimageSlice, ok := ctx.customRecords[record.KeySendType]
@@ -704,7 +703,7 @@ func (i *InvoiceRegistry) processKeySend(ctx invoiceUpdateCtx,
 
 	// Cancel htlc is preimage is invalid.
 	preimage, err := lntypes.MakePreimage(preimageSlice)
-	if err != nil || preimage.Hash() != hash {
+	if err != nil || preimage.Hash() != ctx.hash {
 		return errors.New("invalid keysend preimage")
 	}
 
@@ -751,7 +750,7 @@ func (i *InvoiceRegistry) processKeySend(ctx invoiceUpdateCtx,
 
 	// Insert invoice into database. Ignore duplicates, because this
 	// may be a replay.
-	_, err = i.AddInvoice(invoice, hash)
+	_, err = i.AddInvoice(invoice, ctx.hash)
 	if err != nil && err != channeldb.ErrDuplicateInvoice {
 		return err
 	}
@@ -789,6 +788,7 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
 	// Create the update context containing the relevant details of the
 	// incoming htlc.
 	updateCtx := invoiceUpdateCtx{
+		hash:                 rHash,
 		circuitKey:           circuitKey,
 		amtPaid:              amtPaid,
 		expiry:               expiry,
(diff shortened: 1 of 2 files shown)


invoices/invoiceregistry.go
```diff
@@ -779,6 +779,56 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+	// autoRelease signals that the htlc should be automatically released
+	// after a timeout.
+	autoRelease bool
+
+	// acceptTime is the time at which this htlc was accepted.
+	acceptTime time.Time
+}
+
+// notifyExitHopHtlcLocked is the internal implementation of NotifyExitHopHtlc
+// that should be executed inside the registry lock.
```
bhandras: > that should be exectuted inside the registry lock

This is not true, locks itself.

invoices/invoiceregistry.go
```diff
@@ -779,6 +779,56 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+func (i *InvoiceRegistry) notifyExitHopHtlcLocked(rHash lntypes.Hash,
```
bhandras: Maybe I'd name this function something like `notifyExitHopHtlcImpl` to reflect better what it does.

invoices/invoiceregistry.go
```diff
@@ -779,6 +779,56 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+	amtPaid lnwire.MilliSatoshi, expiry uint32, currentHeight int32,
+	circuitKey channeldb.CircuitKey, hodlChan chan<- interface{},
+	payload Payload) (interface{}, error) {
+
```
bhandras: Can't comment below, but can't we restrict the scope of locking even further? All notifications could happen outside the lock right? That way this function extraction may not even be needed and the timer could stay also here.

invoices/invoiceregistry.go
joostjager: Yes good point. Now it doesn't lock itself anymore. It is also clearer now that the keysend logic (which also cannot operate under the lock) is moved out too.

invoices/invoiceregistry.go
joostjager: I am not sure what `Impl` reflects, it is very generic. But in the latest revision of the PR, the implementation of the method matches its name again.

invoices/invoiceregistry.go
joostjager: The notifications need to happen inside the lock to guarantee that they are delivered in order.

invoices/invoiceregistry.go
```diff
@@ -807,8 +807,53 @@ func (i *InvoiceRegistry) NotifyExitHopHtlc(rHash lntypes.Hash,
...
+	autoRelease bool
+
+	// acceptTime is the time at which this htlc was accepted.
+	acceptTime time.Time
+}
+
+// notifyExitHopHtlcLocked is the internal implementation of NotifyExitHopHtlc
+// that should be executed inside the registry lock.
+func (i *InvoiceRegistry) notifyExitHopHtlcLocked(
+	ctx *invoiceUpdateCtx, hodlChan chan<- interface{}) (
```
bhandras: I kind of like how `ctx` was called `updateCtx` before. Gives more meaning for readers.

invoices/invoiceregistry.go
bhandras: Maybe I'm misunderstanding this but isn't delivery independent from locking? And therefore you cannot enforce delivery order with a lock?

joostjager: No, suppose we make update A and B in the database and because of no locking, the notifications arrive in the order B, A. Then it seems as if the invoice is in state A while it really is in B

invoices/invoiceregistry.go
joostjager: The lines became long, so I used a shorter var name

invoices/invoiceregistry.go
bhandras: I understand that. To clarify, what I meant is that as notifications are sent by writing to channels, delivery order really is defined at the channel reader side. To answer my own question, if processing on the reader side is sequential, then the lock will indeed enforce order.

juscamarena:
I've tested this pull locally and it fixes the deadlock issue I had.

I can now quickly create invoices and shutdown cleanly.

Thanks for the quick PR!

invoices/invoiceregistry.go
joostjager: Yes, reading is sequential indeed.

PR #3981, opened 2020-02-05
channeldb/channel: migrate byte to varint chanstatus

cfromknecht:
This PR converts the `channeldb.ChannelStatus` from a `uint8` to a `uint64` **w/o needing a migration**.

If in the future we need to allocate more than 8 `ChannelStatus` flags, this provides an easy way of expanding the width of the bitfield. The main premise is to replace the existing single-byte encoding with a TLV-varint, aka BigSize.

BigSize encodes values less than 253 (`0xfe`) as their direct byte encoding; since we have yet to allocate the highest bit in the chan status field, it's safe to assume that all current values are less than `0xfe`. This changes allows us to reinterpret the future encodings of the field, w/o needing to migrate any existing data. With this new interpretation, the on-disk representation will automatically grow to the required width as chan status flags are written back to the db.

There are a number of other places in the codebase where a `bool` became a `uint8`, applying this technique gives us YA opportunity grow them into `uint64` fields w/o needing migrations 🎉

### DIFF
--- a/channeldb/channel.go
+++ b/channeldb/channel.go
@@ -372,7 +372,7 @@ type ChannelCommitment struct {
 
 // ChannelStatus is a bit vector used to indicate whether an OpenChannel is in
 // the default usable state, or a state where it shouldn't be used.
-type ChannelStatus uint8
+type ChannelStatus uint64
 
 var (
 	// ChanStatusDefault is the normal state of an open channel.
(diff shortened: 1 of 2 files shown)


Roasbeef:
Implemented in https://github.com/lightningnetwork/lnd/pull/5955

PR #3984, opened 2020-02-06
channeldb: Replace fetchChannels pending/waiting bools with optional filters

carlaKC:
The first commit is just a change to tests, replacing boiler plate channel setup with a default channel which can be overridden with functional options. 

In this PR, the booleans in `fetchChannels` are replaces with optional filters. This change allows us to decrease the number of calls we need to fetch certain sets of channels (`FetchAllOpenChannels` and `FetchWaitingCloseChannels`) and generally allows for more flexibility in channel queries. 

This is also a first step in investigating the addition of a `RequiresExport` flag that would allow channels to remain in the open channels bucket after close so that we can export accounting data. If we do go with this approach, we would need to filter these channels out somewhere (so they are not included with open channels calls), and this PR paves the path to that. If we don't go with this approach, it's still a nice cleanup.

### DIFF
--- a/channeldb/channel_test.go
+++ b/channeldb/channel_test.go
@@ -187,6 +187,13 @@ func fundingPointOption(chanPoint wire.OutPoint) testChannelOption {
 	}
 }
 
+// channelIDOption is an option which sets the short channel ID of the channel.
+var channelIDOption = func(chanID lnwire.ShortChannelID) testChannelOption {
+	return func(params *testChannelParams) {
+		params.channel.ShortChannelID = chanID
+	}
+}
+
 // createTestChannel writes a test channel to the database. It takes a set of
 // functional options which can be used to overwrite the default of creating
 // a pending channel that was broadcast at height 100.
--- a/channeldb/channel_test.go
+++ b/channeldb/channel_test.go
@@ -71,6 +71,17 @@ var (
 	wireSig, _ = lnwire.NewSigFromSignature(testSig)
 
 	testClock = clock.NewTestClock(testNow)
+
+	// defaultPendingHeight is the default height at which we set
+	// channels to pending.
+	defaultPendingHeight = 100
+
+	// defaultAddr is the default address that we mark test channels pending
+	// with.
+	defaultAddr = &net.TCPAddr{
+		IP:   net.ParseIP("127.0.0.1"),
+		Port: 18555,
+	}
 )
 
 // makeTestDB creates a new instance of the ChannelDB for testing purposes. A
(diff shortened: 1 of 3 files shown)


channeldb/channel_test.go
```diff
@@ -98,21 +109,141 @@ func makeTestDB() (*DB, func(), error) {
...
+	openChannel bool
+}
+
+// testChannelOption is a functional option which can be used to alter the
+// default channel that is creates for testing.
+type testChannelOption func(params *testChannelParams)
+
+// pendingHeightOption is an option which can be used to set the height the
+// channel is marked as pending at.
+var pendingHeightOption = func(height uint32) testChannelOption {
```
bhandras: Is there any advantage of using global variables vs declaring actual functions?

channeldb/db.go
```diff
@@ -556,70 +556,102 @@ func (d *DB) FetchChannel(chanPoint wire.OutPoint) (*OpenChannel, error) {
...
+
+	// description is a string which explains which channels are being
+	// included.
+	description string
+}
+
+// pendingChannelFilter returns a filter based on whether channels are pending
+// (ie, their funding transaction still needs to confirm). If pending is false,
+// channels with confirmed funding transactions are returned.
+var pendingChannelFilter = func(pending bool) fetchChannelsFilter {
```
bhandras: Same question, but is there any reason to have vars instead of functions? Somehow reminds me of javascript :)

channeldb/db.go
```diff
@@ -556,70 +556,102 @@ func (d *DB) FetchChannel(chanPoint wire.OutPoint) (*OpenChannel, error) {
...
+// database. It takes a set of filters which are applied to each channel to
+// obtain a set of channels with the desired set of properties. Only channels
+// which have a true value returned for *all* of the filters will be returned.
+// If no filters are provided, every channel in the open channels bucket will
+// be returned.
+func fetchChannels(d *DB, filters ...fetchChannelsFilter) ([]*OpenChannel, error) {
 	var channels []*OpenChannel
 
+	// Log the number of filters and their descriptions.
+	log.Infof("fetchChannels applying: %v filters", len(filters))
```
bhandras: Would it be better to have this as a debug log? Do we frequently use this information in logs?

channeldb/db.go
```diff
@@ -556,70 +556,102 @@ func (d *DB) FetchChannel(chanPoint wire.OutPoint) (*OpenChannel, error) {
...
+	for _, f := range filters {
+		log.Info(f.description)
```
bhandras: Same: info vs debug log?

carlaKC: I'll bump down to debug and make shorter descriptions + concat them with the length log line above so it's also fewer lines?

channeldb/channel_test.go
carlaKC: Just a design pattern I've picked up somewhere :) functions definitely better than global vars.

channeldb/db.go
bhandras: It should be better if you just keep it `log.Debugf` as that won't assemble the format string when the log level is below debug (ref: https://github.com/btcsuite/btclog/blob/84c8d2346e9fc8c7b947e243b9c24e6df9fd206a/log.go#L362)

channeldb/channel_test.go
bhandras: :+1: :slightly_smiling_face:

channeldb/db.go
wpaulino: Could be more of a trace log, or something not even worth logging possibly.

channeldb/db.go
```diff
@@ -556,70 +556,102 @@ func (d *DB) FetchChannel(chanPoint wire.OutPoint) (*OpenChannel, error) {
...
 func (d *DB) FetchWaitingCloseChannels() ([]*OpenChannel, error) {
-	waitingClose, err := fetchChannels(d, false, true)
-	if err != nil {
-		return nil, err
-	}
-	pendingWaitingClose, err := fetchChannels(d, true, true)
-	if err != nil {
-		return nil, err
+	return fetchChannels(
+		d, waitingCloseFilter(true),
```
wpaulino: This seems to be missing a pending channel filter?

channeldb/channel_test.go
wpaulino: Maybe convert them to functions to maintain consistency with the rest of the project? Not a blocker though.

channeldb/db.go
carlaKC: In the previous logic:
waitingClose = channels that are not pending and waiting close
pendingWaitingClose = channels that are pending and waiting close

So since we're getting pending and not pending channels (which are the only 2 states), we just need to filter by waiting close, since that's what we're after.

channeldb/db.go
carlaKC: Just going to kill the logging, not much value add and can be quite spammy.

PR #3988, opened 2020-02-07
docs: add terminal symbol to docker readme example commands

yancyribbens:
This PR fixes a small typo in the docker docs.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -64,8 +64,8 @@ bitcoin into.
 $ export NETWORK="simnet" 
 
 # Create persistent volumes for alice and bob.
-docker volume create simnet_lnd_alice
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -64,8 +64,8 @@ bitcoin into.
 $ export NETWORK="simnet" 
 
 # Create persistent volumes for alice and bob.
```
carlaKC: nit: we generally prefix commits with the affected package, as per [go guidelines](https://golang.org/doc/contribute.html#first_line). In this case there isn't strictly a package but `docker:` is probably more fitting :)

yancyribbens: np.  prefixed the commit with `docker`

PR #3989, opened 2020-02-07
MPP channeldb bucket structure

halseth:


channeldb/payment_control.go
```diff
@@ -296,6 +299,99 @@ func (p *PaymentControl) Success(paymentHash lntypes.Hash,
...
 }
 
+// AttemptFailureReason is the reason a payment attemot failed.
+type AttemptFailureReason byte
+
+const (
+	// AttemptFailureReasonUnknown is recorded for attempts that failed
+	// with an unknown reason.
+	AttemptFailureReasonUnknown AttemptFailureReason = 0
+)
```
joostjager: Perhaps this should then be the htlc-level wire failure?

channeldb/payments.go
```diff
@@ -28,20 +27,35 @@ var (
 	// root-bucket
 	//      |
 	//      |-- <paymenthash>
-	//      |        |--sequence-key: <sequence number>
-	//      |        |--creation-info-key: <creation info>
-	//      |        |--attempt-info-key: <attempt info>
-	//      |        |--settle-info-key: <settle info>
-	//      |        |--fail-info-key: <fail info>
+	//      |        |--mpp-sequence-key: <sequence number>
```
joostjager: why the prefix `mpp` actually? it is the only type we support in the db? or is it a prep. for amp?

channeldb/payment_control.go
```diff
@@ -217,10 +263,25 @@ func (p *PaymentControl) Success(paymentHash lntypes.Hash,
 			return nil
 		}
 
-		// Record the successful payment info atomically to the
-		// payments record.
-		err = bucket.Put(paymentSettleInfoKey, preimage[:])
-		if err != nil {
+		// We'll also mark all attempts successful, so get the
+		// active attempts bucket.
+		// TODO: this will change.
```
joostjager: Maybe just squash. I personally find it not as easy as one may expect to grasp these intermediate refactor states. I'd rather look at the whole diff even if it is longer (and the total diff is probably slightly smaller)

channeldb/payment_control.go
```diff
@@ -235,6 +299,99 @@ func (p *PaymentControl) Success(paymentHash lntypes.Hash,
...
+	if err := serializeHTLCFailInfo(&b, failInfo); err != nil {
+		return err
+	}
+	failBytes := b.Bytes()
+
+	htlcIDBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(htlcIDBytes, attempt.AttemptID)
+
+	var updateErr error
+	err := p.db.Batch(func(tx *bbolt.Tx) error {
```
joostjager: I keep saying it: batch is premature :)

channeldb/payment_control.go
```diff
@@ -235,6 +299,99 @@ func (p *PaymentControl) Success(paymentHash lntypes.Hash,
...
+			return err
+		}
+
+		// We can only mark in-flight payments as failed.
+		if err := ensureInFlight(bucket); err != nil {
+			updateErr = err
+			return nil
+		}
+
+		// TODO: ensure attempt is in flight and not other state?
```
joostjager: Isn't that checked above?

channeldb/payment_control.go
```diff
@@ -359,15 +517,29 @@ func nextPaymentSequence(tx *bbolt.Tx) ([]byte, error) {
...
+	htlcsBucket := bucket.Bucket(mppHtlcsBucket)
+	if htlcsBucket != nil {
+		htlcs, err := fetchHTLCAttempts(htlcsBucket)
+		if err != nil {
+			return StatusUnknown
+		}
+
+		// Go through all HTLCs, and return StatusSucceeded if any of
+		// them did succeed.
+		// TODO(halseth): Is this iteration a perf bottleneck?
```
joostjager: I wouldn't be worried about that at this stage. Once this is the bottleneck, adoption has happened :)

channeldb/payments.go
```diff
@@ -345,61 +266,152 @@ func (db *DB) FetchPayments() ([]*MPPayment, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return htlcs, nil
+}
+
+// fetchHTLCHTLCAttemptInfo fetches the payment attempt for this HTLC from
+// the bucket.
+func fetchHTLCHTLCAttemptInfo(bucket *bbolt.Bucket) (
```
joostjager: HTLCHTLCAttempt :thinking:

channeldb/mp_payment.go
```diff
@@ -25,20 +25,26 @@ type MPPaymentCreationInfo struct {
...
-type HTLCAttempt struct {
-	// PaymentID is the unique ID used for this attempt.
-	PaymentID uint64
-
-	// SessionKey is the ephemeral key used for this payment attempt.
+// HTLCAttemptInfo contains static information about a specific HTLC attempt
+// for a payment. This information is used by the router to handle any errors
+// coming back after an attempt is made, and to query the switch about the
+// status of the attempt.
+type HTLCAttemptInfo struct {
```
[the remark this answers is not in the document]
halseth: I'm all for using the shard term. `Shard` and `ShardInfo` for the static info?
cc @cfromknecht

channeldb/payment_control.go
halseth: Yeah, I think it is useful to keep track of the actual reason the HTLC failed. Maybe reuse the error codes added by @carlaKC ? :)

channeldb/payment_control.go
halseth: It is `O(N^2)` where N is the number of HTLCs sent for a payment I believe. So you don't need that may failed attempt before this becomes a problem.

channeldb/payments.go
halseth: I added it to distinguish it from the old key, but we can rename it when we know what the final structure should be.

channeldb/payments.go
```diff
@@ -28,20 +27,35 @@ var (
...
 	//      |
 	//      |-- <paymenthash>
-	//      |        |--sequence-key: <sequence number>
-	//      |        |--creation-info-key: <creation info>
-	//      |        |--attempt-info-key: <attempt info>
-	//      |        |--settle-info-key: <settle info>
-	//      |        |--fail-info-key: <fail info>
+	//      |        |--mpp-sequence-key: <sequence number>
+	//      |        |--mpp-creation-info-key: <creation info>
+	//      |        |--mpp-fail-info-key: <(optional) fail info>
```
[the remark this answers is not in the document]
halseth: it might be nil

channeldb/payments.go
```diff
@@ -28,20 +27,35 @@ var (
...
+	//      |        |
+	//      |        |--mpp-htlcs-bucket (shard-bucket)
+	//      |        |        |
+	//      |        |        |-- <htlc attempt ID>
```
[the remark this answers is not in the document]
halseth: You mean track payment?

channeldb/payments.go
```diff
@@ -54,33 +68,48 @@ var (
...
+	// mppHtlcsBucket is a bucket where we'll store the information
+	// about the HTLCs that were attempted for a payment.
+	mppHtlcsBucket = []byte("mpp-htlcs-bucket")
 
-	// paymentCreationInfoKey is a key used in the payment's sub-bucket to
-	// store the creation info of the payment.
-	paymentCreationInfoKey = []byte("payment-creation-info")
+	// htlcAttemptInfoKey is a key used in a HTLC's sub-bucket to store the
+	// info about the attempt that was done for the HTLC in question.
+	htlcAttemptInfoKey = []byte("htlc-attempt-info")
```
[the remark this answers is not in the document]
halseth: It is not a rename, they are actually new keys. But I think this will become more clear when we add the migration, as the old keys will stay around.

channeldb/payments.go
halseth: the only thing better than HTLC :)

channeldb/payment_control.go
joostjager: Well, that is another thing again. I am just thinking about how we are going to return the raw failure in `SendToRoute`. If we decide to use `TrackPayment` to (reliably) track the outcome of a `SendToRoute` call (conceptually being more like`SendShard`), then we need that raw failure message in the db.

channeldb/payment_control.go
joostjager: You mean that we are redetermining the payment status for every htlc result that comes in? I still think that this isn't a problem in practice. Do you have a number in mind when this really becomes problematic? Probably the fetch call is a lot hotter than this in memory loop.

channeldb/payments.go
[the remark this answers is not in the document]
joostjager: Ah ok. They will stay around, but only in our graveyard implementation of `channeldb` right? :) But I get why we don't want to rename, could become confusing.

channeldb/payments.go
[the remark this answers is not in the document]
joostjager: Yes

channeldb/payments.go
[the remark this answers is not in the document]
joostjager: Ok, this is the payment level failure which is nil if the payment is still ongoing or succeeded?

halseth:
Replaced by #4000

Issue #3944, opened 2020-01-21
lnd: further abstract the top-level lnd package to support external initialization of core interfaces


Roasbeef:
With the newer versions of `lnd`, we'd started to progressively refactor `lnd` as itself so it's more generically useable. We've recently reached the point where it's possible to initialize an `lnd` instance within another go process, and also generate a set of bindings to support usage of `lnd` within mobile deployments. 

The aim of this issue is to further abstract the main `lnd` package with the goal of being able to: 
1. Perform configuration parsing _externally_ at will. 
2. Related to #3833, be able to provide initialized versions of core interfaces when "building" up an `lnd` node incrementally. 

Once the two steps above have been completed, a user will be able to _embed_ `lnd` within another Go process or package, and also provide _new_ implementations of core interfaces such as the `Signer` or newly created `kvdb.Backend` interface. 

### Potential External API 

I figure the best way to provide guidance on precisely _how_ to achieve the two goals above is to provide an example using an hypothetical interface, then proceed backwards (a la top-down design) to refactor `lnd` to support our target use case. 

Here's an example of a user creating a new `lnd` instance in an new package and providing a custom implementation (really just re-directs or maybe proxies to a remote server) of the `kvdb.Backend` interface. 
```go

type IO struct {
    Stdout io.Writer 
    Stdin io.Reader
    Stderr io.writer
}

// ConfigBuilder is a type that represents a way to provide custom 
// implementations of base lnd interfaces as parmaters to lnd 
// initialization. The logic in `chainregistry.go` (and also server.go) as it 
// exists now will conditionally create things like the database based 
// on the default-ness of values like this. 
type ConfigBuilder func(*ChainControl)

// CustomDatabaseBuilder is an example of a builder that adds in a 
// custom DB implementation. 
func CustomDatabaseBuilder(db kvdb.Backend) func(*ChainControl) {
    return func(cc *ChainControl) {
        cc.db = db
    }
}

// This parses the main config as we normally do, but doesn't 
// yet carry out any of the steps which are found 
// today in `chainregistry.go` .
baseCfg, err := lnd.ParseMainConfig(configPath)
if err != nil {
      return err
}

// When then create a new instance that provides our own 
// database implementation and uses the standard set 
// of interfaces for stdout/stderr, etc. 
lndInstance, err := lnd.New(
    baseCfg, IO{}, CustomDatabaseBuilder(myNewDb),
)
if err != nil {
    return err
}
```

[...]

PR #3991, opened 2020-02-09
Add SigHashes parameters to SignOutputRaw() and ComputeInputScript()

yaslama:
With this addition, the rpc functions can be used to implement input.Signer
and use any library using input.Signer to sign transactions.


Roasbeef:
Why is this needed given that the transaction is accepted as parameter, which can then be used to generate the sighashes internally?

yaslama:
For instance, with support for sighashes at the rpc level, sweep.CraftSweepAllTx() can be called with an implementation of signer.input using rpc calls. One application is to have the transaction without publishing it. In fact any library using input.Signer interface can be used externally from lnd by using a rpc implementation of input.Signer

yaslama:
I created a small gist demoing the use of this PR: https://gist.github.com/yaslama/ade8eaa6ebe03eb587c1d701c07e5722

guggero:
The sighashes is just a cache. It can be recomputed at any time, it only exists to save computation time. Whether or not the TX should be published has nothing to do with this parameter.
I'm pretty sure your gist still works if you run it against the master branch and just don't pass the sighashes cache.

yaslama:
@guggero I agree that the TX can be obtained correctly without using sighashes and this PR, but with this PR, the signature computation will take O(n) of the previous output instead of O(n^2) without it.
So it's good to use sighashes using rpc for the same reason it's used internally in lnd: performance optimization. Don't you think it's worth this small PR?

guggero:
Are you running into performance problems without this fix? Or is this a premature optimization?

I think it's generally a bad idea to exchange cache state over any API. It makes the interface harder to use/understand, even if the parameter is optional. I'm also not sure of the security implications. Possibly you could trick the signer into signing something it's not supposed to by sending manipulated sighashes.

IMO we shouldn't do this...

yaslama:
> Are you running into performance problems without this fix? Or is this a premature optimization?

If it's a premature optimization, why lnd does that internally? For instance, in the gist we compute the transaction 3 times (with different target confirmations) in order let user choose the fees.
So we need more optimizations than the current code in lnd, not less. And it can be important especially in mobiles.
> 
> I think it's generally a bad idea to exchange cache state over any API. It makes the interface harder to use/understand, even if the parameter is optional. I'm also not sure of the security implications. Possibly you could trick the signer into signing something it's not supposed to by sending manipulated sighashes.

I don't follow. The client calling these rpc has the macaroon needed to access the api. So she can ask to sign everything. Do you think about a particular scenario?

cfromknecht:
> with this PR, the signature computation will take O(n) of the previous output instead of O(n^2) without it. So it's good to use sighashes using rpc for the same reason it's used internally in lnd: performance optimization. Don't you think it's worth this small PR?

Do you have concrete benchmarks showing that this is actually _the_ bottleneck?

cfromknecht:
Signing the inputs would still O(n^2), and I would wager that it dwarfs computing 3 hash values. Even if you make the hashing complexity linear, the operation as a whole remains quadratic.

yaslama:
> Signing the inputs would still O(n^2), and I would wager that it dwarfs computing 3 hash values. Even if you make the hashing complexity linear, the operation as a whole remains quadratic.

I completely agree. My argument is more that there is no reason to discriminate the rpc functions vs the api used in the same process.

cfromknecht:
> I completely agree. My argument is more that there is no reason to discriminate the rpc functions vs the api used in the same process.

There absolutely is a reason to discriminate. Everything exposed via rpc has a maintenance cost and a commitment to backwards compatibility, while changing things internally is cheap. Adding things with marginal utility only bloats the interface and makes usage confusing or error prone.

In general I think if this level of optimization is required, and a user is signing enough transactions such that asymptotics are at play, then there's probably a more fundamental issue with the API or usage of the rpc interface. 

In your example, you ask for `n` signed transactions in order to let the user decide which fee rate to use, but it would be more performant altogether just to request unsigned transactions and only sign the one that they decide on.

At this point in time I doubt anyone is trying to sign millions of transactions using this RPC. I'd suspect the performance increase is marginal, and until we have evidence that this fixes a practical performance bottleneck this is probably a premature optimization. I'd consider changing my opinion if the benchmarks show otherwise :)

yaslama:
@cfromknecht Thanks for your detailed answer. Let me detail a little more my argument.
lnd can be broken conceptually in two parts: 1) live services: onchain wallet, LN wallet, gossip, etc.. and 2) libraries/packages. Part of the libraries don't need any live service (for instance to decode a payment request) but the majority of libraries need to "call" one ore more live services.
Lot of "interesting" LN application can use libraries from lnd but need also to "call" the live services. Right now the only way to do that is to make changes to lnd and recompile it. A better way to work would be to have low level rpc functions which could be used as building blocks from a process outside lnd and use the "regular" lnd downloaded from github.
Using the logic, there is no reason to treat differently a public Go function than a grpc "low level" function.
In the case of the signrpc service, I completely agree that the optimization is premature for *this specific use case* but I wrote the gist to demonstrate the architecture: any package using the public Go interface to sign transaction can be used from *outside* lnd using low level rpc.
One can imagine to move all the onchain transactions creation from lnd to the "client" process if the right low level rpc exists. 
Some grpc functions can be marked as unstable (with no commitment that they will not change) the same way as lnd public interfaces are unstable.
In fact, it's a way to overcome the limitations of Go to use dynamic libraries to write extensions.

Roasbeef:
Seems related to #3944.

PR #3998, opened 2020-02-12
Fix TestNewBreachRetributionSkipsDustHtlcs incorrect state transtion

halseth:
The unit test TestNewBreachRetributionSkipsDustHtlcs triggered a state
transition from Bob, even though it was Alice that had added the HTLCs.
This is wrong since it will lead to Bob still owing Alice a commitment,
which is not accounted for in the unit tests.

We add a sanity check that the add heights has been set for all entries
found in the logs, and return an error otherwise. This won't happen
during normal operation, but it does reveal the mistake in the unit
test, which is fixed by making Alice trigger the transition.

In addition we resolve a long standing TODO by removing a (purposeful)
panic in the channel state machine. Old version of lnd had a bug that
could lead to the parent entries being lost during channel restore. A
panic was added to get to the bottom of if.

This is now fixed, so new nodes shouldn't encounter it. However, to be
on the safe side, instead of panicking we return an error back to
gracefully exit the channel state machine.

### DIFF
--- a/lnwallet/channel.go
+++ b/lnwallet/channel.go
@@ -2460,6 +2460,46 @@ func (lc *LightningChannel) evaluateHTLCView(view *htlcView, ourBalance,
 	skipUs := make(map[uint64]struct{})
 	skipThem := make(map[uint64]struct{})
 
+	// fetchParentEntry is a helper method that will fetch the parent of
+	// entry from the corresponding update log.
+	fetchParentEntry := func(entry *PaymentDescriptor,
+		remoteLog bool) *PaymentDescriptor {
+
+		var (
+			updateLog *updateLog
+			logName   string
+		)
+
+		if remoteLog {
+			updateLog = lc.remoteUpdateLog
+			logName = "remote"
+		} else {
+			updateLog = lc.localUpdateLog
+			logName = "local"
+		}
+
+		addEntry := updateLog.lookupHtlc(entry.ParentIndex)
+
+		switch {
+		// We check if the parent entry is not found at this point. We
+		// have seen this happening a few times and panic with some
+		// addtitional info to figure out why.
+		// TODO(halseth): remove when bug is fixed.
+		case addEntry == nil:
+			panic(fmt.Sprintf("unable to find parent entry %d "+
+				"in %v update log: %v\nUpdatelog: %v",
+				entry.ParentIndex, logName,
+				newLogClosure(func() string {
+					return spew.Sdump(entry)
+				}), newLogClosure(func() string {
+					return spew.Sdump(updateLog)
+				}),
+			))
+		}
+
+		return addEntry
+	}
+
 	// First we run through non-add entries in both logs, populating the
 	// skip sets and mutating the current chain state (crediting balances,
 	// etc) to reflect the settle/timeout entry encountered.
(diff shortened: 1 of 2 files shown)


halseth:
Small mistake in one of the lnwallet test I encountered when working on https://github.com/lightningnetwork/lnd/pull/3691. Could be worth considering for 0.9.1?

lnwallet/channel.go
```diff
@@ -2460,6 +2460,46 @@ func (lc *LightningChannel) evaluateHTLCView(view *htlcView, ourBalance,
...
+	// entry from the corresponding update log.
+	fetchParentEntry := func(entry *PaymentDescriptor,
+		remoteLog bool) *PaymentDescriptor {
+
+		var (
+			updateLog *updateLog
+			logName   string
+		)
+
+		if remoteLog {
```
joostjager: Can be moved inside the case statement

lnwallet/channel.go
```diff
@@ -2481,23 +2484,34 @@ func (lc *LightningChannel) evaluateHTLCView(view *htlcView, ourBalance,
...
+			)
+
+		// The parent add height should never be zero at this point. If
+		// that's the case we probably forgot to send a new commitment.
+		case remoteChain && addEntry.addCommitHeightRemote == 0:
+			return nil, fmt.Errorf("parent entry %d for update %d "+
+				"had zero remote add height", entry.ParentIndex,
+				entry.LogIndex)
+		case !remoteChain && addEntry.addCommitHeightLocal == 0:
+			return nil, fmt.Errorf("parent entry %d for update %d "+
```
joostjager: mention local or remote in log line?

lnwallet/channel.go
```diff
@@ -2481,23 +2484,34 @@ func (lc *LightningChannel) evaluateHTLCView(view *htlcView, ourBalance,
...
 				newLogClosure(func() string {
 					return spew.Sdump(entry)
 				}), newLogClosure(func() string {
 					return spew.Sdump(updateLog)
 				}),
-			))
+			)
+
+		// The parent add height should never be zero at this point. If
+		// that's the case we probably forgot to send a new commitment.
```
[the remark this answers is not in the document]
halseth: Yes. If it is zero it means it wasn't added to a commitment (yet), so a `CommitSig` that covers it must be sent. This is what was missing in the test case, and hence we wasn't actually testing settling the HTLCs, as they weren't added at that point.

lnwallet/channel.go
halseth: "zero **remote** add height" ?

lnwallet/channel.go
halseth: What can be moved exactly?

halseth:
> How does this play with settles pipelining? When we find out that the update log isn't sane, we already handed off the settle to the switch.

It shouldn't happen, since this is not really a situation that happens in practice. And if that happens I guess it is better to pipeline it to the switch so you can claim your money there then not to :)

lnwallet/channel.go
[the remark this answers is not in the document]
joostjager: I mean just using the magic value of zero to signal that it wasn't added, while commit height zero is also a valid commitment. But I guess it is ok

lnwallet/channel.go
joostjager: `updateLog` and `logName` are only used inside the switch case

lnwallet/channel.go
joostjager: haha. ok. :)

joostjager:
>It shouldn't happen, since this is not really a situation that happens in practice. And if that happens I guess it is better to pipeline it to the switch so you can claim your money there then not to :)

And there are no other side effects of the pipelining that we need to be aware of? Closing circuits, updating fwd log? I don't have a particular concern, but just checking

cfromknecht:
@joostjager pipelining of settles only [after processing the settle successfully](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L1660) so we should be fine in terms of circuits, fwding log, etc

lnwallet/channel.go
halseth: Update log is used on line 2481, felt right to do the log name in the same if

joostjager: ah, ok

PR #4003, opened 2020-02-14
lncli: change listinvoices reversed flag to paginate-forwards

### LINKED ISSUE #3986: listinvoices: reversed flag has no effect
Currently the `lncli listinvoices --reversed` flag has no effect, as it
is of type `cli.BoolTFlag` https://github.com/lightningnetwork/lnd/blob/ad0a89b844ec9f55b3ca4ddbb4d368dd8e3c0ea9/cmd/lncli/commands.go#L2848
which is true by default.

Correct behavior can be triggered by `lncli listinvoices --reversed=True`
 or `lncli listinvoices --reversed=False`.

Is this behavior intended? I get that it is more comfortable for lncli users to
see the latest invoices by default. An easy fix would be to change 
`cli.BoolTFlag` to `cli.BoolFlag`, however, an lncli user would have to really use
then `lncli listinvoices --reversed` for the latest invoices. Could take over and fix it.

bitromortac:
The listinvoices `--reversed` flag is changed to `--paginate-forwards` to fix #3986.

Currently, if one likes to start to paginate from the beginning of the invoices history
one has to set `--reversed=False`, which is not very intuitive. Otherwise the 
`--reversed` flag has no effect at all, as it is set by default. The downside of the change
is that it diverts from the grpc API, where the reversed flag still is present.

If this change is wanted I would apply it also to the payments listing in PR #3960.

### DIFF
--- a/cmd/lncli/commands.go
+++ b/cmd/lncli/commands.go
@@ -2819,10 +2819,10 @@ var listInvoicesCommand = cli.Command{
 	allowing users to query for specific invoices through their add_index.
 	This can be done by using either the first_index_offset or
 	last_index_offset fields included in the response as the index_offset of
-	the next request. The reversed flag is set by default in order to
-	paginate backwards. If you wish to paginate forwards, you must
-	explicitly set the flag to false. If none of the parameters are
-	specified, then the last 100 invoices will be returned.
+	the next request. Backward pagination is enabled by default to receive
+	current invoices first. If you wish to paginate forwards, set the 
+	paginate-forwards flag.	If none of the parameters are specified, then 
+	the last 100 invoices will be returned.
 
 	For example: if you have 200 invoices, "lncli listinvoices" will return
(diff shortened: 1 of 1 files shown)


cmd/lncli/commands.go
```diff
@@ -2845,11 +2845,10 @@ var listInvoicesCommand = cli.Command{
...
 		},
-		cli.BoolTFlag{
-			Name: "reversed",
-			Usage: "if set, the invoices returned precede the " +
-				"given index_offset, allowing backwards " +
-				"pagination",
+		cli.BoolFlag{
+			Name: "paginate-forwards",
+			Usage: "if set, invoices succeeding the " +
+				"index_offset will be displayed",
```
guggero: nit: s/be displayed/be returned/ to stay consistent with the wording in the other descriptions.

bitromortac: Agreed, have changed it :+1:

PR #4010, opened 2020-02-17
peer: avoid chansync resend loop

halseth:
If a peer receives a channel reestablish message shortly after the
channel has been closed, it will resend its own channel reestablish
message. In the meantime the other peer could also have seen the channel
being closed and will also resend its own message. This leads to a
resend loop that never terminates.

To avoid two peers getting into this situation, we now allow only one
such resent message per conenction.

Noticed during integration tests: https://travis-ci.org/lightningnetwork/lnd/jobs/650839757?utm_medium=notification&utm_source=github_status

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -210,6 +210,12 @@ type peer struct {
 	// the connection handshake.
 	remoteFeatures *lnwire.FeatureVector
 
+	// resentChanSyncMsg is a set that keeps track of which channels we
+	// have re-sent channel reestablishment messages for. This is done to
+	// avoid getting into loop where both peers will respond to the other
+	// peer's chansync message with its own over and over again.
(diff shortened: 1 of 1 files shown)


peer.go
```diff
@@ -2527,6 +2534,11 @@ func (p *peer) sendInitMsg() error {
 // resendChanSyncMsg will attempt to find a channel sync message for the closed
 // channel and resend it to our peer.
 func (p *peer) resendChanSyncMsg(cid lnwire.ChannelID) error {
+	// If we already re-sent the mssage for this channel, we won't do it again.
+	if _, ok := p.resentChanSyncMsg[cid]; ok {
```
Roasbeef: We don't appear to actually populate the map in this diff as is.

halseth: Wow, that's embarrassing ... 🤦‍♂ fixed

PR #4015, opened 2020-02-18
kvdb wrapper for etcd

bhandras:
This PR builds on top of [channeldb/kvdb: introduce new KV-store database abstraction](https://github.com/lightningnetwork/lnd/pull/3833) and adds a kvdb wrapper for etcd.

The PR is still somewhat WIP, but mostly ready for initial reviews.

Missing:

- [x] Batch
- [x] Inline comments from most places
- [x] Test coverage is ~90%
- [x] Configurable integration to LND/tests.
- [x] Handle errors in manual transactions
- [ ] Migration tool to mirror bbolt dump to etcd.
- [ ] A lot of additional testing.


channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	}
+
+	// Load prefetch to save a round trip.
+	s.rset.add(txnresp)
+	s.prefetch = s.rset
+
+	return CommitError{}
+}
+
+func (s *stm) Rollback() {
```
Roasbeef: Prefetch is kept in place as an optimization?

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	s.rset = make(map[string]stmGet)
+	s.wset = make(map[string]stmPut)
+	s.getOpts = nil
+	s.revision = math.MaxInt64 - 1
```
Roasbeef: Not clear on the purpose of this STM level revision.

go.mod
```diff
@@ -40,20 +52,26 @@ require (
...
 	github.com/rogpeppe/fastuuid v1.2.0 // indirect
+	github.com/soheilhy/cmux v0.1.4 // indirect
+	github.com/stretchr/testify v1.4.0
+	github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5 // indirect
 	github.com/tv42/zbase32 v0.0.0-20160707012821-501572607d02
 	github.com/urfave/cli v1.18.0
-	go.etcd.io/bbolt v1.3.3
-	golang.org/x/crypto v0.0.0-20190211182817-74369b46fc67
-	golang.org/x/net v0.0.0-20190206173232-65e2d4e15006
+	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
```
Roasbeef: How much of this goes away with a `go mod tidy`?

go.mod
```diff
@@ -64,4 +82,8 @@ replace github.com/lightningnetwork/lnd/cert => ./cert
 
 replace git.schwanenlied.me/yawning/bsaes.git => github.com/Yawning/bsaes v0.0.0-20180720073208-c0276d75487e
 
+replace github.com/coreos/go-systemd => github.com/coreos/go-systemd/v22 v22.0.0
+
+replace github.com/btcsuite/btcwallet/walletdb => github.com/bhandras/btcwallet/walletdb v0.0.0-20200218161033-f2ed9c1c77a2
```
Roasbeef: Can point to `btcwalelt` master as your dep PR has been merged in.

channeldb/kvdb/etcd/db.go
```diff
@@ -32,6 +34,28 @@ func newEtcdBackend(host, user, pass string) (*db, error) {
 	return backend, nil
 }
 
+// Copy writes a copy of the database to the provided writer.  This call will
```
Roasbeef: 👍

channeldb/kvdb/etcd/bucket.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"bytes"
+)
+
+func isValidKey(key []byte) bool {
+	return !bytes.ContainsAny(key, "/~")
+}
+
+func makeBucketPrefix(path string) string {
+	return "b/" + path + "~"
```
Roasbeef: Any reason the tilde is chosen? For the purpose of quickly terminating certain classes of range queries?

channeldb/kvdb/etcd/db.go
```diff
@@ -59,3 +105,6 @@ func (db *db) Copy(w io.Writer) error {
 func (db *db) Close() error {
 	return db.cli.Close()
 }
+
+// TODO
+//func (db *db) Batch(apply func(tx kvdb.RwTx) error) error { }
```
Roasbeef: Batch can just call `Update`, since `etcd` as is, is inherently multi-writer.

channeldb/kvdb/etcd/readwrite_bucket.go
```diff
@@ -0,0 +1,246 @@
...
+// the bucket.  This includes nested buckets, in which case the value
+// is nil, but it does not include the key/value pairs within those
+// nested buckets.
+func (b *readWriteBucket) ForEach(cb func(k, v []byte) error) error {
+	prefix := makeValuePrefix(b.path)
+	prefixLen := len(prefix)
+
+	kv := b.tx.stm.First(prefix)
+	for kv != nil {
+		if err := cb([]byte(kv.key[prefixLen:]), []byte(kv.val)); err != nil {
```
Roasbeef: Bounds check key before accessing `prefixLen`, both here and below?

channeldb/kvdb/etcd/readwrite_bucket.go
```diff
@@ -0,0 +1,246 @@
...
+			return err
+		}
+		kv = b.tx.stm.Next(prefix, kv.key)
+	}
+
+	prefix = makeBucketPrefix(b.path)
+	prefixLen = len(prefix)
+
+	kv = b.tx.stm.First(prefix)
```
Roasbeef: Is the repeat here on purpose? (the loop)

channeldb/kvdb/etcd/readwrite_cursor.go
```diff
@@ -0,0 +1,88 @@
...
+		prefix: makeValuePrefix(bucket.path),
+	}
+}
+
+// First positions the cursor at the first key/value pair and returns
+// the pair.
+func (c *readWriteCursor) First() (key, value []byte) {
+	if kv := c.bucket.tx.stm.First(c.prefix); kv != nil {
+		c.currKey = kv.key
+		return []byte(kv.key[len(c.prefix):]), []byte(kv.val)
```
Roasbeef: Similar comment here re bounds checking prefix key access before hand.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	// Put adds a value for a key to the txn's write set.
+	Put(key, val string, opts ...v3.OpOption)
+
+	// Del adds a delete operation for the key to the txn's write set.
+	Del(key string)
+
+	// First returns the first k/v that begins with prefix or nil if there's
+	// no such k/v pair. If the key is found it is inserted to the txn's
+	// read set.
+	First(prefix string) *KV
```
cfromknecht: any reason the interface operates on strings and not byte slices?

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+
+	// Commit attempts to apply the txn's changes to the server.
+	Commit() error
+
+	// Rollback emties the read and write sets such that a subsequent commit
+	// won't alter the database.
+	Rollback()
+}
+
+type CommitError struct{}
```
cfromknecht: would a sentinel `CommitError = errors.New("...")` suffice?

channeldb/kvdb/etcd/db.go
```diff
@@ -32,6 +34,28 @@ func newEtcdBackend(host, user, pass string) (*db, error) {
...
+	ctx, cancel := context.WithTimeout(ctx, etcdTimeout)
+	defer cancel()
+
+	readCloser, err := db.cli.Snapshot(ctx)
+	if err != nil {
+		return err
+	}
+
+	if _, err = io.Copy(w, readCloser); err != nil {
+		return err
```
cfromknecht: can return this error directly

channeldb/kvdb/etcd/db.go
```diff
@@ -34,6 +38,48 @@ func newEtcdBackend(host, user, pass string) (*db, error) {
...
+	}
+	return nil
+}
+
+func (db *db) BeginReadWriteTx() (kvdb.RwTx, error) {
+	return newReadWriteTx(NewSTM(db.cli)), nil
+}
+
+func (db *db) BeginReadTx() (kvdb.ReadTx, error) {
+	return newReadWriteTx(NewSTM(db.cli)), nil
```
cfromknecht: are we able to assert that the db isn't modified?

channeldb/kvdb/etcd/db.go
```diff
@@ -34,6 +38,48 @@ func newEtcdBackend(host, user, pass string) (*db, error) {
...
+// View opens a database read transaction and executes the function f with the
+// transaction passed as a parameter.  After f exits, the transaction is rolled
+// back.  If f errors, its error is returned, not a rollback error (if any
+// occur).
+func (db *db) View(f func(tx kvdb.ReadTx) error) error {
+	apply := func(stm STM) error {
+		return f(newReadWriteTx(stm))
+	}
+
+	if err := RunSTM(db.cli, apply); err != nil {
```
cfromknecht: style nit: can return error directly, also in `Update`

channeldb/kvdb/etcd/bucket.go
```diff
@@ -0,0 +1,29 @@
+package etcd
+
+import (
+	"bytes"
+)
+
+func isValidKey(key []byte) bool {
+	return !bytes.ContainsAny(key, "/~")
```
cfromknecht: does this mean we need to encode binary keys or values so they don't mess with the control characters? seems like we need to pick an encoding scheme that doesn't collide with `{'b': 98, 'v': 118, '$': 36, '/': 47, '~': 126}`

channeldb/kvdb/etcd/readwrite_bucket.go
```diff
@@ -0,0 +1,246 @@
...
+
+	return nil
+}
+
+func (b *readWriteBucket) Put(key, value []byte) error {
+	if len(key) == 0 {
+		return walletdb.ErrKeyRequired
+	}
+
+	if !isValidKey(key) {
```
cfromknecht: since we regularly insert random data, i think we need to have something that converts binary data into a character set that doesn't collide with the prefixing strategy.

channeldb/kvdb/etcd/readwrite_bucket.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+
+		return nil, err
+	}
+
+	return bucket, nil
+}
+
+func (b *readWriteBucket) DeleteNestedBucket(key []byte) error {
+
```
cfromknecht: nit: extra new line

channeldb/kvdb/etcd/readwrite_bucket.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil
+	}
+
+	valueKey := makeValuePrefix(b.path) + string(key)
+	val := b.tx.stm.Get(valueKey)
+	if val == "" {
+		return nil
+	}
+
+	return []byte(val)
```
cfromknecht: similarly we may have to decode hex/base64 back into binary if we needed to convert it when storing.

channeldb/kvdb/etcd/bucket.go
bhandras: We might do that, or simply hash part of the keys eg. v/path/key => value_prefix + hash(path) + key. This will change key locality though (not sure if this is a problem)

bhandras: Other possibly viable idea is to use a lenght/value encoding for path parts and the key.

channeldb/kvdb/etcd/bucket.go
bhandras: This (tilde may be anything really) is just to simplify the case where we iterate trough buckets in the current bucket. Instead of making etcd fetch all bucket keys with the path prefix, we just select the ones in the next level.

channeldb/kvdb/etcd/db.go
bhandras: So do I understand correctly that `testBatchInterface` in walletdb is incorrect as it calls `CreateTopLevelBucket` in each concurrent batch? As that'll make the first batch committed to succeed the rest to fail. It's simple to correct that test if what I assume is right and in that case `Batch` is indeed a simple `Update` call.

channeldb/kvdb/etcd/readwrite_bucket.go
bhandras: If I understand what you refer here: then yes. As `ForEach` iterates trough all keys and also all sub buckets (though returning empty value). As keys are currently prefixed with `v/path~` and buckets with `b/path~`.

go.mod
bhandras: There's a new PR in `btcwallet` that needs to be merged beforehand to support `Batch` tests.

channeldb/kvdb/etcd/db.go
bhandras: depends on https://github.com/btcsuite/btcwallet/pull/679

channeldb/kvdb/etcd/bucket.go
bhandras: @Roasbeef just pointed out there's a LevelDB implementation for walletdb, where the approach is pretty nice and flat: https://github.com/btcsuite/btcd/blob/06e5c43499a2f30a1f7d7c3b49de0073dfbbdd1f/database/ffldb/db.go#L565

I'll adopt something like this instead.

Roasbeef:
Can be rebased now that the `kvdb` PR has been merged.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+func (e DatabaseError) Error() string {
+	return fmt.Sprintf("etcd error: %v - %v", e.msg, e.err)
+}
+
+type stmGet struct {
+	val string
+	rev int64
+}
+
+type readSet map[string]stmGet
```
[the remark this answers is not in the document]
bhandras: yeah, comments are incoming

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	return puts
+}
+
+func (s *stm) fetch(key string, opts ...v3.OpOption) []KV {
+	resp, err := s.client.Get(
+		s.options.ctx, key, append(opts, s.getOpts...)...,
+	)
+
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
bhandras: When we do transactions inside a `View` or `Update` then panics are handled in the STM runloop. I think we may need to modify the walletdb/kvdb interface to support errors which happen in "free-standing" transactions (started with `BeginReadTx` or `BeginReadWriteTx`), as `ReadBucket.Get` doesn't support errors.

channeldb/kvdb/etcd/db.go
bhandras: Not really, what we can do is call `Commit` then retry manually. In general `BeginReadTx` and `BeginReadWriteTx` isn't really nice as we need to handle errors which the current interface doesn't support in all cases, whereas we can do this in `Update` and `View`

channeldb/kvdb/etcd/bucket.go
bhandras: So binary keys are supported for some time now.

channeldb/kvdb/etcd/readwrite_bucket.go
bhandras: Not sure I understand what you meant here. If you referred to binary keys, that's done.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+
+func (s *stm) Next(prefix string, key string) *KV {
+	return s.next(prefix, key, false)
+}
+
+func (s *stm) Seek(prefix, key string) *KV {
+	return s.next(prefix, key, true)
+}
+
+func (s *stm) next(prefix, startKey string, leftClosed bool) *KV {
```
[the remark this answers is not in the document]
bhandras: Renamed this to includeStartKey and added some more comments to explain what is happening.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	matchFound := false
+
+	for {
+		opts := []v3.OpOption{
+			v3.WithFromKey(),
+			v3.WithSort(v3.SortByKey, v3.SortAscend),
+			v3.WithLimit(1),
+		}
+
+		if !leftClosed || !firstFetch {
```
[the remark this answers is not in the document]
bhandras: done

channeldb/kvdb/etcd/stm.go
bhandras: Yes, unless ranges are requested (in which case we always fetch), prefetch holds values that'd otherwise be re-fetched.

channeldb/kvdb/etcd/stm.go
bhandras: It is the transactions base revision.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	Commit() error
+
+	// Rollback emties the read and write sets such that a subsequent commit
+	// won't alter the database.
+	Rollback()
+}
+
+type CommitError struct{}
+
+func (e CommitError) Error() string {
```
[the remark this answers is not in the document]
bhandras: Added `Unwrap` for `DatabaseError`, not sure how it would be useful for `CommitError` as it doesn't wrap a real error (just a type).

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+	return "commit failed"
+}
+
+type DatabaseError struct {
+	msg string
+	err error
```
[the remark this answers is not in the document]
bhandras: done.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+
+		s.Rollback()
+	}
+
+	return err
+}
+
+func (rs readSet) add(txnResp *v3.TxnResponse) {
+	for _, resp := range txnResp.Responses {
+		getResp := (*v3.GetResponse)(resp.GetResponseRange())
```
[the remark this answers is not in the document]
bhandras: I think this is not possible as `RangeResponse` is not an interface. (etcd uses this cast too)

channeldb/kvdb/etcd/db.go
bhandras: yeah, originally I was thinking about whether or not we should do some kind of error translation.

channeldb/kvdb/etcd/db.go
bhandras: done

channeldb/kvdb/etcd/readwrite_bucket.go
bhandras: Are you sure we need prefix check here? `stm.First` will either return `nil` or a match in which case it contains prefix.

channeldb/kvdb/etcd/readwrite_cursor.go
bhandras: Similarly to my above comment, `stm.First` will either return `nil` or a match which contains prefix, so I believe bounds check is unnecessary here.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,566 @@
...
+}
+
+type stmResponse struct {
+	resp *v3.TxnResponse
+	err  error
+}
+
+func runSTM(s *stm, apply func(STM) error) error {
+	var err error
+	for {
```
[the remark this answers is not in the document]
Roasbeef: Hmm, good questions, it seems like if we hit our soft-limit, then we may either need a background goroutine that tries again, or we just abandon the transaction all together (panic).

channeldb/kvdb/etcd/embed.go
```diff
@@ -0,0 +1,73 @@
...
+
+// NewEmbeddedEtcdInstance creates an embedded etcd instance for testing,
+// listening on random open ports. Returns the connection config and
+// a cleanup func that will stop the etcd instance.
+func NewEmbeddedEtcdInstance(path string) (*ConnConfig, func(), error) {
+	cfg := embed.NewConfig()
+	cfg.Dir = path
+
+	// To ensure that we can submit large transactions.
+	cfg.MaxTxnOps = 1000
```
Roasbeef, line 43: Is this tuned based on the current testnet testing? In the end though, I think this is only meant to be used in tests, correct?

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,736 @@
...
+
+			// Rollback before trying to re-apply.
+			s.Rollback()
+		}
+
+		// Return the error to the caller.
+		out <- err
+	}()
+
+	return <-out
```
Roasbeef, line 281: FWIW, this will likely block a daemon shutdown until the transaction commits or it runs into a fatal error, which I exactly what we want I think.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,736 @@
...
+func (ws writeSet) cmps(rev int64) []v3.Cmp {
+	cmps := make([]v3.Cmp, 0, len(ws))
+	for key := range ws {
+		cmps = append(cmps, v3.Compare(v3.ModRevision(key), "<", rev))
+	}
+
+	return cmps
+}
+
+// puts is the list of ops for all pending writes
```
Roasbeef, line 331: Style nit: missing a period here and the above two methods.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,736 @@
...
+		}
+
+		// Do not panic when executing a manual transaction.
+		if s.manual {
+			return nil, dbErr
+		}
+
+		// Panic when executing inside the STM runloop.
+		panic(dbErr)
+
```
Roasbeef, line 362: Style nit: extra new line.

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,736 @@
...
+		kv    KV
+		found bool
+	)
+
+	if len(resp) > 0 {
+		kv = resp[0]
+		found = true
+	}
+
+	// Now make sure there's noting in the write set
```
Roasbeef, line 467: nothing -> nothing

channeldb/kvdb/etcd/stm.go
```diff
@@ -0,0 +1,736 @@
...
+	cmps := append(s.rset.cmps(), s.wset.cmps(s.revision+1)...)
+	// Create a transaction with the optional abort context.
+	txn := s.client.Txn(s.options.ctx)
+
+	// If the compare set holds, try executing the puts.
+	txn = txn.If(cmps...)
+	txn = txn.Then(s.wset.puts()...)
+
+	// Prefetch keys in case of conflict to save
+	// a round trip to etcd.
```
Roasbeef, line 727: 👍

channeldb/kvdb/etcd/bucket.go
```diff
@@ -0,0 +1,86 @@
...
+	prefixBuf := make([]byte, len(prefix)+len(parent))
+	copy(prefixBuf, prefix)
+	copy(prefixBuf[len(prefix):], parent)
+
+	return prefixBuf
+}
+
+// makeBucketKey returns a bucket key from the passed parent
+// bucket id and the key.
+func makeBucketKey(parent []byte, key []byte) []byte {
```
Roasbeef, line 58: Much easier to follow this new bucket mapping 👍

channeldb/kvdb/etcd/bucket.go
```diff
@@ -0,0 +1,86 @@
...
+// rootBucketId returns a zero filled 32 byte array
+func rootBucketID() []byte {
+	var rootID [bucketIDLength]byte
+	return rootID[:]
+}
+
+// makeBucketID returns a deterministic key for the passed byte slice.
+// Currently it returns the sha256 hash of the slice.
+func makeBucketID(key []byte) [bucketIDLength]byte {
+	return sha256.Sum256(key)
```
Roasbeef, line 26: An alternative to creating the bucket ID based on the actual bucket _name_ would be to use a sequence number here. So the first bucket has an ID of zero, the next one and so one. I don't envision in the short term that we'll run into performance issues with usage of `sha2` here, but if we want to drop tat dependnacy, then we can go with the sequence number route instead.

channeldb/kvdb/etcd/db.go
```diff
@@ -50,7 +59,73 @@ func newEtcdBackend(config ConnConfig) (*db, error) {
...
+// returned.
+func (db *db) Update(f func(tx walletdb.ReadWriteTx) error) error {
+	apply := func(stm STM) error {
+		return f(newReadWriteTx(stm))
+	}
+
+	return RunSTM(db.cli, apply)
+}
+
+func (db *db) BeginReadWriteTx() (walletdb.ReadWriteTx, error) {
```
Roasbeef, line 158: Missing a `godoc` comment here and below.

channeldb/channel_test.go
```diff
@@ -98,13 +98,19 @@ func makeTestDB() (*DB, func(), error) {
 	}
 
 	// Next, create channeldb for the first time.
-	cdb, err := Open(tempDirName, OptionClock(testClock))
+	backend, backendCleanup, err := kvdb.GetTestBackend(tempDirName, "cdb")
+	if err != nil {
+		return nil, nil, err
+	}
+	cdb, err := OpenWithBackend(backend, OptionClock(testClock))
+
```
Roasbeef, line 106: Style nit: extra new line.

channeldb/kvdb/backend.go
```diff
@@ -0,0 +1,91 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/lightningnetwork/lnd/channeldb/kvdb/etcd"
+)
+
+// getEnvBackendConfig reads LND_TEST_DB_BACKEND env variable and
+// returns its value if it equals to "etcd" or "bdb" otherwise.
+func getEnvBackendConfig() string {
+	backend, _ := os.LookupEnv("LND_TEST_DB_BACKEND")
```
Roasbeef, line 14: We should switch this over to using a build tag before this lands. 

One way to do this would be something like the following: 
```go
// +build !etcd-db
const TestBackend = "bdb"
```
Then we have a file with the opposite that's gated on specifying that build flag: 
```go
// +build etcd-db
const TestBackend = "etcd"
```

lncfg/db.go
```diff
@@ -0,0 +1,60 @@
+package lncfg
+
+import (
+	"fmt"
+
+	"github.com/lightningnetwork/lnd/channeldb/kvdb"
+)
+
+// DB holds database configuration for LND.
+type DB struct {
```
Roasbeef, line 10: What do you think about unpacking things a bit so we can spcify thing son the command line like: `--db.bolt.syncfreelist`? 

I think the the following structure would allow us to do so poretty easily: 
```go
type BoltDb struct {
    NoSyncFreeList bool `long:"nosyncfreelist`
}

type EtcdDb struct {
    <the rest of the fields as contained now>
}

type DB struct {
    Etcd *BoltDb `group:"bolt" namespace:"bolt"`

    Bolt *EtcdDb `group:"bolt" namespace:"bolt"`
}
```

Roasbeef: With the above, a user would specify `etcd` on the command line like `--db.etcd.host=x`, etc. When parsing these values, we'd ensure that only one of those internal pointers has been specified.

channeldb/kvdb/etcd/embed.go
bhandras: This is only used in test and required to make channeldb tests succeed. You can test it yourself with: `LND_TEST_DB_BACKEND=etcd go test -v` in the channeldb package.

channeldb/kvdb/etcd/bucket.go
bhandras: This would not work if there are multiple db instances. Also we need to make sure bucket ids are deterministic when running multiple nodes on the same db (in the far-far future :))

bhandras: Also the sequence number thing would require deterministic order of bucket creation, otherwise failover would be tricky.

channeldb/kvdb/etcd/stm.go
bhandras: done

channeldb/kvdb/etcd/stm.go
bhandras: done

channeldb/kvdb/etcd/stm.go
bhandras: done

channeldb/kvdb/etcd/stm.go
bhandras: :tada:

channeldb/kvdb/etcd/bucket.go
bhandras: yeah, I realized it was over-complicated after taking a fresh look

channeldb/kvdb/etcd/db.go
bhandras: done

channeldb/channel_test.go
bhandras: done

channeldb/kvdb/backend.go
bhandras: ptal

bhandras:
Just marking this as ready for review, as it's already being reviewed anyway.

I think it'd be better to move the config part to a different PR, also stats collection could be removed entirely as it's not that important, was mostly just for myself to see how it works out with actual LND.

Roasbeef:
@bhandras so then you want to remove the config commit? I think it's better to leave it in, as otherwise if this lands there's no way to actually even run w/ `etcd` mode on. My suggested changes shouldn't be too involved (re the sub-configs, which is also related to #3944).

channeldb/kvdb/etcd/bucket.go
Roasbeef: Ah makes sense, great decision.

channeldb/db.go
```diff
@@ -147,32 +146,35 @@ type DB struct {
...
+	if err == nil {
+		db.dbPath = dbPath
+	}
+	return db, err
+}
+
+// OpenWithBackend opens or creates channeldb using the passed kvdb.Backend.
+// Any necessary schemas migrations due to updates will take place as necessary.
+func OpenWithBackend(backend kvdb.Backend, modifiers ...OptionModifier) (*DB, error) {
+	if err := initChannelDB(backend); err != nil {
```
Roasbeef, line 167: As is this, more or less assumes the backend is already "open", as it just does initialization. What about just passing in the args and name, then using `kvdb.Open(name, args...)`? We'd get a`kvdb.Backend` instance out of it, then can proceed with set up as normal.

channeldb/kvdb/kvdb_test_bdb.go
```diff
@@ -0,0 +1,5 @@
+// +build !kvdb_etcd
```
Roasbeef, line 1: 👍 

Once we want them to run on travis (both versions), we'll also need to update the script we use to execute the buid.

lnd.go
```diff
@@ -226,11 +228,19 @@ func Main(lisCfg ListenerCfg) error {
 	ltndLog.Infof("Opening the main database, this might take a few " +
 		"minutes...")
 
+	cfg.DB.Path = graphDir // TODO(bhandras)
+	cfg.DB.Name = "channel.db"
+	chanDbBackend, err := cfg.DB.GetBackend()
```
Roasbeef, line 233: Here's what I mean that `OpenWithBackend` assumes the backend is already obtained. The method is more like create/init with backend.

lnd.go
```diff
@@ -226,11 +228,19 @@ func Main(lisCfg ListenerCfg) error {
 	ltndLog.Infof("Opening the main database, this might take a few " +
 		"minutes...")
 
+	cfg.DB.Path = graphDir // TODO(bhandras)
```
Roasbeef, line 231: This should likely go before this is merged. The path from the config should be used.

lnd.go
```diff
@@ -729,6 +739,9 @@ func Main(lisCfg ListenerCfg) error {
 	// Wait for shutdown signal from either a graceful server stop or from
 	// the interrupt handler.
 	<-signal.ShutdownChannel()
+	if v, ok := chanDbBackend.(kvdb.BackendWithStats); ok {
+		ltndLog.Infof(v.PrintStats())
```
Roasbeef, line 743: Should be removed before this is merged.

channeldb/db.go
bhandras: The main reason this is done as this way is to support testing with multiple backends without much code rewrite. So when we test we first instantiate the `kvdb.Backend` then "open" the channeldb.
Another strong reason is that the `kvdb.Backend` becomes a dependency this way and it's much simpler to stub/mock it if needed.

lnd.go
bhandras: Yeah, I agree fully. Renamed `OpenWithBackend` to `CreateWithBackend`.

lncfg/db.go
bhandras: I believe go-flags creates subgroups by default so pointer checking is not possible (although maybe possible to hack this together). So I instead created a top level `--db.backend` flag which by default is set to `bolt` and the sub configs proposed in your comment. PTAL.

lnd.go
bhandras: Removed. Also I reworked the commit stats collector code a little bit to be less coupled with the database.

lnd.go
bhandras: I think the config must be almost parsed to fill the path, so rather I created a function in `config.go` to return the local database directory. This is then passed as a param to the `GetBackend` function. We could of course expose database path in the config, but I'm not sure if we should? Maybe this is something our users want though.

bhandras:
Delta since last review:
- reworked stats collector (https://github.com/lightningnetwork/lnd/pull/4015/commits/d94be59e3816957b51a0d5bd2157d4ebb167b899)
- config changes (https://github.com/lightningnetwork/lnd/pull/4015/commits/d9e3afcc2f06702e9268695dd8f8625631335f23)
- OpenWithBackend => CreateWithBackend

lncfg/db.go
```diff
@@ -0,0 +1,85 @@
...
+
+// EtcdDB hold etcd configuration.
+type EtcdDB struct {
+	Host string `long:"host" description:"Etcd database host."`
+
+	User string `long:"user" description:"Etcd database user."`
+
+	Pass string `long:"pass" description:"Password for the database user."`
+
+	CollectStats bool `long:"collect_stats" description:"Wheter to collect etcd commit stats."`
```
Roasbeef, line 29: This also needs TLS config info for certain configurations.

Roasbeef:
Looks like a clean shutdown attempt if a db transaction is open just hangs.

go.mod
```diff
@@ -68,4 +88,6 @@ replace git.schwanenlied.me/yawning/bsaes.git => github.com/Yawning/bsaes v0.0.0
 // btcsuite/btcutil package requests a newer version.
 replace golang.org/x/crypto => golang.org/x/crypto v0.0.0-20190211182817-74369b46fc67
 
+replace github.com/coreos/go-systemd => github.com/coreos/go-systemd/v22 v22.0.0
```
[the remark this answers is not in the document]
Roasbeef: IIRC, there's an issue with the `go.mod` of this package that forces us to do this replace manually.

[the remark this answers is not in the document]
Roasbeef: https://github.com/coreos/go-systemd/issues/321

channeldb/kvdb/etcd/bucket.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"crypto/sha256"
+)
+
+const (
+	bucketIDLength = 32
+)
+
+var (
+	bucketPrefix   = []byte("b")
```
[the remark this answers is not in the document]
bhandras: :+1:

channeldb/kvdb/etcd/bucket.go
```diff
@@ -0,0 +1,86 @@
...
+	return sha256.Sum256(key)
+}
+
+// isValidBucketID checks if the passed slice is the required length
+// to be a valid bucket id.
+func isValidBucketID(s []byte) bool {
+	return len(s) == bucketIDLength
+}
+
+// makeKey concatenates prefx, parent and key into one byte slice.
```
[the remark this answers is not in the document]
bhandras: done

channeldb/kvdb/etcd/bucket_test.go
```diff
@@ -0,0 +1,38 @@
...
+		bucketKey = makeBucketKey(parent, []byte(bucketName))
+		id := makeBucketID(bucketKey)
+		parent = id[:]
+	}
+
+	return string(bucketKey)
+}
+
+// bval is a helper function used in tests to create a bucket value
+// (the value part of a bucket key) from the passed bucket list.
```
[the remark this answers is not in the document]
bhandras: corrected, ptal

channeldb/kvdb/interface.go
```diff
@@ -36,6 +36,13 @@ var Create = walletdb.Create
 // through read or read+write transactions.
 type Backend = walletdb.DB
 
+// BackendWithStats is and interface to debug/uncover database access patterns.
+type BackendWithStats interface {
+	Backend
+
+	PrintStats() string
```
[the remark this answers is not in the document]
bhandras: done (In a latter commit, as this is now renamed. Will squash after review)

lncfg/db.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Validate validates the DB config.
+func (db *DB) Validate() error {
+	switch db.Backend {
+	case boltBackend:
+
+	case etcdBackend:
+		if db.Etcd.Host == "" {
+			return fmt.Errorf("host must be set")
```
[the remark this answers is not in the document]
bhandras: done

lncfg/db.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+	default:
+		return fmt.Errorf("unknwon backend, must be either \"%v\" or \"%v\"",
```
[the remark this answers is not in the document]
bhandras: thx, done

lncfg/db.go
bhandras: done in separate commit (can squash after review).

channeldb/db.go
```diff
@@ -141,11 +142,43 @@ var (
...
+	if v, ok := db.Backend.(kvdb.ExtendedBackend); ok {
+		return v.Update(f)
+	}
+	return walletdb.Update(db, f)
+}
+
+// View is a wrapper around walletdb.View which calls into the extended
+// backend when available. This call is needed to be able to cast DB to
+// ExtendedBackend.
+func (db *DB) View(f func(tx walletdb.ReadTx) error) error {
```
Roasbeef, line 164: Is this layer also needed? Since all callers end up using `kvdb.Update/View`? 

Guess it gives callers a way to just interact directly w/ the DB itself.

Roasbeef: Also in certain areas, there's a subtlety w.r.t what actually gets passed into `kvdb.Update`. In that behavior can differ if it's this struct (that embeds) a backend vs passing in the raw backend value.

channeldb/kvdb/etcd/db.go
```diff
@@ -130,18 +131,40 @@ type BackendConfig struct {
...
 // newEtcdBackend returns a db object initialized with the passed backend
 // config. If etcd connection cannot be estabished, then returns error.
 func newEtcdBackend(config BackendConfig) (*db, error) {
+	tlsInfo := transport.TLSInfo{
+		CertFile:           config.CertFile,
+		KeyFile:            config.KeyFile,
+		InsecureSkipVerify: config.InsecureSkipVerify,
+	}
+
+	tlsConfig, err := tlsInfo.ClientConfig()
```
Roasbeef, line 157: 👍

channeldb/kvdb/etcd/db.go
```diff
@@ -130,18 +131,40 @@ type BackendConfig struct {
 	// Pass is the password for the etcd peer.
 	Pass string
 
+	// CertFile holds the path to the TLS certificate for etcd RPC.
+	CertFile string
+
+	// KeyFile holds the path to the TLS private key for etcd RPC.
+	KeyFile string
```
Roasbeef, line 138: Odd that we would need the _private_ key here right? In my local branch, seemed it wouldn't proceed at runtime unless this was set. Perhaps we need to specify some custom validation for the cert. Note that this is only needed for self signed certs, if etcd host as a domain, and it has a "real" cert, then it'll validate as normal.

bhandras: PTAL

channeldb/db.go
bhandras: This is needed to keep the current global `kvdb.Update/View` functions while also allowing extended transaction logic when a backend implements that. With the `kvdb.ExtendedBackend` we keep proper separation, so no need to bake in reference to etcd here.

bhandras: > Also in certain areas, there's a subtlety w.r.t what actually gets passed into `kvdb.Update`. In that behavior can differ if it's this struct (that embeds) a backend vs passing in the raw backend value.

In my tests when `channeldb.DB` itself did not implement the `ExtendedBackend` interface, it was impossible to to the type assertions as with the embedding the extended type information is lost and `db.Backend` cannot be "upcasted" anymore.

channeldb/kvdb/etcd/db.go
bhandras: In the end I reverted back to the way `etcd` does the TLS config as that actually seems to be correct to me. Could do TLS with auto generated certs and insecure skip verify. I'm not sure why the trusted cert requires the private key or if it requires at all, but please also try with `etcdctl --debug` to see if that works and what error it gives.

lncfg/db.go
```diff
@@ -50,12 +51,14 @@ func (db *DB) Validate() error {
 }
 
 // GetBackend returns a kvdb.Backend as set in the DB config.
-func (db *DB) GetBackend(path string) (kvdb.Backend, error) {
+func (db *DB) GetBackend(dbPath string) (kvdb.Backend, error) {
 	if db.Backend == etcdBackend {
-		return kvdb.GetEtcdBackend(db.Etcd)
+		// Prefix will separate key/values in the db.
+		prefix := path.Join(dbPath, dbName)
```
Roasbeef, line 57: We can shorten this a bit, by just extracting which _network_ the user is running on and using that as a prefix. Otherwise, we'll end up writing part of the user's file system path as the main bucket key for everything, which can possibly change depending on where they're running `lnd` and what they have set as the `datadir`, etc. 

The other item we should tack on here is the pubkey of the node, which is known before we update the database. So the final prefix would be: `/network/<nodePubKey/`.

channeldb/kvdb/etcd/readwrite_tx.go
```diff
@@ -25,7 +35,51 @@ func newReadWriteTx(stm STM) *readWriteTx {
...
 func rootBucket(tx *readWriteTx) *readWriteBucket {
-	return newReadWriteBucket(tx, rootBucketID())
+	return newReadWriteBucket(tx, rootBucketID(), rootBucketID())
+}
+
+// lock adds a key value to the lock set.
+func (tx *readWriteTx) lock(key, val string) {
+	tx.stm.Lock(key)
+	if !tx.dirty {
+		tx.lset = append(tx.lset, KV{key, val})
```
Roasbeef, line 52: Should this also be a map? IIRC there was an issue with using a slice for the main read/write set re duplicates.

channeldb/kvdb/etcd/readwrite_bucket.go
```diff
@@ -20,7 +21,12 @@ type readWriteBucket struct {
 
 // newReadWriteBucket creates a new rw bucket with the passed transaction
 // and bucket id.
-func newReadWriteBucket(tx *readWriteTx, id []byte) *readWriteBucket {
+func newReadWriteBucket(tx *readWriteTx, key, id []byte) *readWriteBucket {
+	if !bytes.Equal(id, rootBucketID()) {
+		// Add the bucket key/value to the lock set.
+		tx.lock(string(key), string(id))
```
Roasbeef, line 29: This is cool! It's basically like tree locking.

lncfg/db.go
bhandras: This is not an actual prefix, perhaps the naming is not the best as it confuses readers :). It is the same as if there were an extra bucket, which means that in the end the whole path (except the key) is flattened to 32 bytes.

channeldb/kvdb/etcd/readwrite_tx.go
bhandras: yes, indeed! fixed

channeldb/kvdb/etcd/readwrite_bucket.go
bhandras: yeah, this locks the keys along the bucket path, (although possibly creates more conflicts than simply locking the on the keys in the bucket but is also more correct)

lncfg/db.go
bhandras: So the "flattening" (key generation?) algorithm is a bit easier to grasp if you take a look at `bucket_test.go` and how the `bkey` function is used in the tests. The way it is currently done, we do create the key when we create a `ReadWriteBucket` and just use the cached key after that. Each bucket key is prefixed with a byte containing `b` and each value key is prefixed with a byte containing `v` (for details please look at: `bucket.go`.

cfromknecht: i like the idea of not prefixing all keys with redundant data, having a root bucket as H(network | nodepub) make sense 👍

it does seem like that path in full should be removed, otherwise one could have multiple states for the same (node, network) pair which is dangerous

channeldb/kvdb/etcd/readwrite_tx.go
```diff
@@ -12,6 +12,16 @@ type readWriteTx struct {
 	// active is true if the transaction hasn't been
 	// committed yet.
 	active bool
+
+	// dirty is true if we intent to update a value
```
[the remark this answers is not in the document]
bhandras: done

channeldb/kvdb/interface.go
```diff
@@ -36,11 +47,27 @@ var Create = walletdb.Create
 // through read or read+write transactions.
 type Backend = walletdb.DB
 
-// BackendWithStats is and interface to debug/uncover database access patterns.
-type BackendWithStats interface {
+// ExtendedBackend is and interface that supports View and Update and als
```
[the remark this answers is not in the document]
bhandras: done

channeldb/db.go
```diff
@@ -407,7 +407,7 @@ func (d *DB) FetchOpenChannels(nodeID *btcec.PublicKey) ([]*OpenChannel, error)
 // stored currently active/open channels associated with the target nodeID. In
 // the case that no active channels are known to have been created with this
 // node, then a zero-length slice is returned.
-func (d *DB) fetchOpenChannels(tx kvdb.ReadTx,
+func (db *DB) fetchOpenChannels(tx kvdb.ReadTx,
```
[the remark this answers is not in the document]
bhandras: your inner voice is right, maybe I should reword the commit too :)

channeldb/channel_test.go
```diff
@@ -98,13 +98,19 @@ func makeTestDB() (*DB, func(), error) {
 	}
 
 	// Next, create channeldb for the first time.
-	cdb, err := Open(tempDirName, OptionClock(testClock))
+	backend, backendCleanup, err := kvdb.GetTestBackend(tempDirName, "cdb")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cdb, err := CreateWithBackend(backend, OptionClock(testClock))
```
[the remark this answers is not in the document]
bhandras: Good catch! Not the most elegant to fix, but there it is. PTAL

channeldb/db_test.go
```diff
@@ -33,7 +34,13 @@ func TestOpenWithCreate(t *testing.T) {
...
 	// Next, open thereby creating channeldb for the first time.
 	dbPath := filepath.Join(tempDirName, "cdb")
-	cdb, err := Open(dbPath)
+	backend, cleanup, err := kvdb.GetTestBackend(dbPath, "cdb")
+	if err != nil {
+		t.Fatalf("unable to get test db backend: %v", err)
+	}
+	defer cleanup()
+	cdb, err := CreateWithBackend(backend)
+
```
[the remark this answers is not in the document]
bhandras: done

channeldb/db_test.go
```diff
@@ -73,7 +80,13 @@ func TestWipe(t *testing.T) {
...
```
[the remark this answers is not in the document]
bhandras: done

bhandras:
> note: in testing i was unable to shut down the daemon without kill -9 due to the large read requests. we can follow up with passing cancel contexts into the queries so we can break out if needed

Yes this is indeed a thing to solve. It is probably not caused by large tx's as much as frequent conflicts, keeping one tx retrying over and over (holding lnd from stopping). There's probably multiple layers to this: one is to limit the retries to a certain number (could be context dependent?) and the other is the cancellation context.

 I'll address this along with the itests in an upcoming PR.

bhandras:
> 

i like the idea of not prefixing all keys with redundant data, having a root bucket as H(network | nodepub) make sense +1

it does seem like that path in full should be removed, otherwise one could have multiple states for the same (node, network) pair which is dangerous

Is that because people migrate their DB?

bhandras:
Added a commit to separate etcd databases by network name instead of full path. PTAL

channeldb/kvdb/etcd/stm.go
cfromknecht: nope, actually i prefer the typing now that you mention it 👍

cfromknecht:
> Added a commit to separate etcd databases by network name instead of full path. PTAL

Looks correct to me! Are we planning to add node pubkey to the root bucket? Or did I miss where that's done?

bhandras:
> > Added a commit to separate etcd databases by network name instead of full path. PTAL
> 
> Looks correct to me! Are we planning to add node pubkey to the root bucket? Or did I miss where that's done?

This is a great idea and we should do it, but imho this is perhaps a bit out of scope for this PR as to achieve this we need to refactor the init order to be able to derive the private key before the channeldb is constructed.

githorray:
Is the db time_to_open reported by LND under-reported after this merge?

```
[INF] LTND: Opening the main database, this might take a few minutes...
[INF] CHDB: Checking for schema update: latest_version=13, db_version=13
[INF] LTND: Database now open (time_to_open=19.680575ms)!
```


I'm used to a few seconds, but not milliseconds, especially after a reboot.

Before merge:

```
[INF] LTND: Database now open (time_to_open=14.952646033s)!
[INF] LTND: Database now open (time_to_open=19.449286406s)!
[INF] LTND: Database now open (time_to_open=14.656393946s)!
[INF] LTND: Database now open (time_to_open=15.319765911s)!
[INF] LTND: Database now open (time_to_open=14.79603768s)!
[INF] LTND: Database now open (time_to_open=14.659297753s)!
```

After merge:

```
[INF] LTND: Database now open (time_to_open=6.325834ms)!
[INF] LTND: Database now open (time_to_open=19.680575ms)!
```

Roasbeef:
@githorray nice catch, should be fixed in this PR: https://github.com/lightningnetwork/lnd/pull/4309

PR #4017, opened 2020-02-19
build: udpate to latest btcwallet version with rescan addr fix

Roasbeef:
In this commit, we update to the latest `btcwallet` version that
includes a fix for how we perform rescans. Before this commit, the
wallet would load ALL the created keys into the wallet to perform a
rescan. This is unnecessary, as many of the keys we create are actually
used in contracts, so the wallet can't spend them directly anyway.

For neutrino nodes, this would've caused them to attempt o match more
items in the filter than necessary, possibly resulting in an increased
number of false positive block fetches.


Roasbeef:
> My question however is: Doesn't this affect the to_remote output of tweakless commitments?

We handle manually anyway as is with SCB. It's also the case that only _after_ a restart for SCB would that work. Also remember that with the new anchor output stuff, this'll be a P2WSH script anyway (CSV of 1).

Issue #3953, opened 2020-01-24
htlc callback acceptance 


nicolasburtey:
There is a PR about better reporting for HTLC forwarding: https://github.com/lightningnetwork/lnd/issues/3420#issuecomment-576853370

It would relevant to go one step further and allow callback for acceptance.

I see at least one use case where this would be relevant:
If a routing node is connected to knowingly mobile phone nodes that are offline most of the time but could be notified when a payment is pending. One of the requirement would be to make payment pending when receiving an HTLC for some channels, instead of automatically fail if the node is offline, as is currently happening with lnd.

carlaKC:
>  One of the requirement would be to make payment pending when receiving an HTLC for some channels, instead of automatically fail if the node is offline, as is currently happening with lnd.

Would the idea be to get the incoming htlc, see that it's for a known mobile client and then notify them to come online so that the htlc can be locked into their commitment?

champo:
@carlaKC at Muun we would love to have that feature! Ideally we would be able to decide whether to accept an htlc meant for a non existant channel with the mobile node, so actual channels can be created on demand.

I'd be more than happy to contribute to make this possible.

carlaKC:
> Ideally we would be able to decide whether to accept an htlc meant for a non existant channel with the mobile node, so actual channels can be created on demand.

If the channel doesn't exist yet, nodes wouldn't be able to find a route to the mobile device. If you were sure your routing node would be the second last hop, I think you could do some moon magic with hop hints but would have to think about it more. 

This would be a non-trivial change to lnd, because the switch (where we forward htlcs) is one of the more complex subsystems. Some things you'd have to think about are htlcs timing out while channel acceptor is waiting for the mobile node to come online, and handling this kind of blocking in the switch (which I think would be most challenging).

> I'd be more than happy to contribute to make this possible.

That would be great! A good place to start would be `handlePacketForward` in `htlcSwitch/switch.go` and having a look at the `ChannelAcceptor` in lnd which is a similar principle for channels being opened to our node. Feel free to ping me on the lnd slack/request for review if you decide to take this on :D

nicolasburtey:
@carlaKC my request was for existing channels (to get started :) )

But as @champo suggested, having this for non-existing channels (the "Phoenix" wallet way) would be a great addition.

Roasbeef:
Implemented by #4018

PR #4018, opened 2020-02-19
Intercept forward htlc

roeierez:
This PR introduces the ability to intercept htlc forward events that are handled at the htlcswitch package and to override the switch default behavior. Basically, similar to HODL invoices, this PR enables routing nodes to HODL HTLCs and to run custom validations or processes before forwarding (or cancelling) the HTLC.
More specifically for every HTLC that is intended to be forwarded, the implementor can choose to hold the forward and signal the switch to wait for a later resolution. Such resolution can be one of the following:
1. _Resume_ - Resume execution to the default behavior (probably just forward).
2. _Fail_ - Fail the htlc backward, the implementor is not interested in forwarding it.
3. _Settle_ - For a given preimage that was revealed in a way not known to the switch the implementor can settle the htlc.
Having more control over the forward behavior enables new important use cases to be implemented.
We are specifically looking to support the following use cases:

**Pay to Open**
This is about being able to pay someone with no channel or with a low capacity channel by opening a channel during the payment.
Let's say Alice (that has a channel) wants to pay Bob (that doesn't have a channel).
1. Bob is connected to a node and issues an invoice with a routing hint that points to a fake channel between Bob and the node. 
2. When Alice pays Bob's invoice, the node intercepts the HTLC and holds it.
3. At this stage one of the two options can follow:
    - The node opens a channel to Bob and upon confirmation (or not) resumes the forward.
    - The node opens a channel to Bob, pays Bob the original Alices amount minus service fee (in a different invoice), gets the preimage from Bob (in exchange to the payment) and settles the hold forward.

**Notify to Pay**
Let's say Alice wants to pay Bob, a mobile user, which is connected to a node via private channel.
1. Alice pays, the node intercept the forward, checks Bob's availability and notices Bob is offline (inactive).
2. The node notifies Bob using an agreed-upon notification mechanism.
2. The node waits for Bob to be online and once the channel is active, it resumes the forward.

**Implementation flow & Usage**
1. Link forward a packet, packet intercepted and sent to the RPC layer.
2. For each packet the RPC layer do the following:

- Hold the packet immediately and signal hold backwards to the switch.
- Send the packet for the client

3. In parallel RPC listens to client resolutions responses on the same stream and resolve packets accordingly.
4. On disconnection all remaining held packets are resolved using the default behavior (currently `Resume`)

I am attaching a sample file to show a sample usage that filter packets according to amount
[cmd_intercept_forward.txt](https://github.com/lightningnetwork/lnd/files/4686651/cmd_intercept_forward.txt)


halseth:
Can be integrated with the HTLC notifier? #3781 
cc @carlaKC

carlaKC:
> Can be integrated with the HTLC notifier? #3781

I think they're different sides of the same coin? `HTLCNotifier` notifies forwarding decisions post-fact, once we have made the forwarding/failing decision. 

My understanding of this is that it's preemptive: receive forward, callout to external forwarding predicate and then return, as specified in #3953.

roeierez:
> Can be integrated with the HTLC notifier? #3781

@halseth thanks for bringing this to the discussion.
Some things to note when thinking about integration:
1.  The notifications are dispatched after the forward has been processed while the interception should be hooked before that, this means the current htlcNotifier interface needs to be extended.
2. Interception is bidirectional and need to handle responses as well, this affects both the middleware and the RPC layer.
3. Controlling the forwards will probably need different RPC permissions than getting htlc notifications.
4. One example I see in the code that seems similar to this case is the channel events vs the channel acceptor. Two different mechanisms that have similar path in the code.

My take is that it will be an attempt to mix different things together but I am interested in @carlaKC opinion as well.
Hope that helps.

Roasbeef:
Interception should be done at the switch, not the link. Also _extreme care_ must be taken to not block the entire forwarding fabric, as the switch manages its state synchronously atm.

champo:
@roeierez I really like the look of the API here. I've also been working on something similar. It allows you to register _virtual_ channels and control HTLCs there through gRPC. It adds a new `Link` implementation and the intercept happens there. As far as the switch is concerned, nothing changes from a regular channel, which I think goes along the lines of what @Roasbeef suggests.

The highly-WIP code is at https://github.com/lightningnetwork/lnd/commit/b3b2f7f4433463691919f85081b0f74d6c9749ed and an usage example at https://gist.github.com/champo/8e90ce1ef38f2dd546814ff7898988c3 A knonw issue is that it doesn't close circuits so it kinda stops working after a success settle. It's easy to fix but I haven't gotten around to it yet.

roeierez:
> Interception should be done at the switch, not the link. Also _extreme care_ must be taken to not block the entire forwarding fabric, as the switch manages its state synchronously atm.

I have moved the interception to the switch, added unit tests and the Router RPC implementation.

lnrpc/routerrpc/router.proto
```diff
@@ -490,6 +490,67 @@ message BuildRouteResponse {
...
 
+message CircuitKey {    
+    /// The id of the channel that the htlc arrived at our node on.
+    uint64 incoming_chan_id = 1;
+
+    /// The index of the incoming htlc in the incoming channel.
+    uint64 htlc_id = 2;
+}
+
+message ForwardHtlcInterceptRequest {
```
champo: It'd be great to include the sphinx packet in this message. It would allow a receiver to check amount and expiry as it would if the HTLC had been forwarded through a link. This is useful for reverse swaps.

roeierez: @champo thanks for the feedback.
Currently the fields `amount_sat` `expiry` are what you are looking for as they are the actual values that would have been used in the forwarded packet in the default switch behavior.

champo: A third-party receiver (say a mobile app) shouldn't trust the fields. They would want to check the information against the sphinx. It would also enable usage of new fields like payment secret which is only known to the receiver.

roeierez: @champo I agree. At first I exposed only the fields that are required for the use cases I mentioned in the PR description. I do think it makes sense to expose all the forwarding info (such as outgoing channel, tlv fields and raw sphinx packet) as other use cases may need this info in order to take a decision at interception time.

htlcinterceptor/middleware.go
```diff
@@ -0,0 +1,146 @@
...
+		fwdHandlers:  make(map[uint64]ForwardHtlcHandler),
+		holdForwards: make(map[channeldb.CircuitKey]ForwardResolver),
+	}
+}
+
+// AddForwardHtlcHandler adds a ForwardHtlcHandler.
+func (c *Middleware) AddForwardHtlcHandler(
+	handler ForwardHtlcHandler) uint64 {
+
+	id := atomic.AddUint64(&c.fwdHandlerID, 1)
```
champo: do you need an atomic here when `fwdHandlers` is always accessed with a lock?

htlcinterceptor/middleware.go
```diff
@@ -0,0 +1,146 @@
...
+	resolver, ok := c.holdForwards[circuitKey]
+	c.Unlock()
+	if !ok {
+		return ErrFwdNotExists
+	}
+	if err := resolveFunc(resolver); err != nil {
+		return err
+	}
+
+	delete(c.holdForwards, circuitKey)
```
champo: shoudn't this be protected by the lock too? concurrent writes to maps are not supported

htlcswitch/switch.go
```diff
@@ -63,6 +64,50 @@ var (
 	ErrUnreadableFailureMessage = errors.New("unreadable failure message")
 )
 
+// forwardResolver implements the htlcinterceptor.Interceptor interface.
```
champo:
```suggestion
// forwardResolver implements the htlcinterceptor.ForwardResolver interface.
```

htlcswitch/switch.go
```diff
@@ -63,6 +64,50 @@ var (
...
+}
+
+// Resume resumes the default behavior as if the packet was not intercepted.
+func (h *forwardResolver) Resume() error {
+	return h.htlcSwitch.handlePacketForward(h.packet)
+}
+
+// Fail forward a failed packet to the switch.
+func (h *forwardResolver) Fail() error {
+	return h.resolve(&lnwire.UpdateFailHTLC{})
```
champo: is it possible to set a `Reason`? what will it default to?

htlcswitch/switch.go
```diff
@@ -63,6 +64,50 @@ var (
...
+	circuitInKey := h.packet.inKey()
+	pkt := &htlcPacket{
+		outgoingChanID: h.packet.outgoingChanID,
+		outgoingHTLCID: h.packet.outgoingHTLCID,
+		isResolution:   true,
+		circuit:        h.packet.circuit,
+		circuitInKey:   &circuitInKey,
+		htlc:           message,
+	}
+	return h.htlcSwitch.handlePacketForward(pkt)
```
champo: I _think_ this should either be calling `ForwardPackets`, `routeAsync` or `route`. As far as I can tell, calls to `handlePacketForward` always happen in the context of `hltcForwarder` go routine to keep the switch "sync". Calling it here directly might be messing up some assumptions made down the line. The 3 methods suggested before make sure the message ends up in `hltcForwarder`.

htlcswitch/switch.go
```diff
@@ -1364,7 +1442,19 @@ func (s *Switch) closeCircuit(pkt *htlcPacket) (*PaymentCircuit, error) {
...
 	// Otherwise, this is packet was received from the remote party.  Use
 	// circuit map to find the incoming link to receive the settle/fail.
-	circuit, err := s.circuits.CloseCircuit(pkt.outKey())
+	var circuit *PaymentCircuit
+	var err error
+	if pkt.circuitInKey != nil {
+		// In case we got the circuitInKey populated it means that we should
+		// use it to find the circuit as there is no open circuit probably due
+		// to interception of this packet.
+		circuit = s.circuits.LookupCircuit(*pkt.circuitInKey)
```
champo: is this circuit ever deleted?

htlcinterceptor/middleware.go
roeierez: No, moved to within the lock scope.

htlcinterceptor/middleware.go
roeierez: We should! Done.

htlcswitch/switch.go
roeierez: Good one! Used route.

htlcswitch/switch.go
roeierez: Yes, this test verifies that for all resolve options: https://github.com/lightningnetwork/lnd/pull/4018/files#diff-9ca6c00aa4cbaaa08aaa8bd697703398R3115

htlcswitch/switch.go
```diff
@@ -63,6 +64,52 @@ var (
...
+type forwardResolver struct {
+	packet     *htlcPacket
+	htlcSwitch *Switch
+}
+
+// Resume resumes the default behavior as if the packet was not intercepted.
+func (h *forwardResolver) Resume() error {
+	pkt := *h.packet
+	pkt.intercepted = true
+	return h.htlcSwitch.route(&pkt)
```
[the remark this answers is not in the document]
roeierez: Yes, this is the main event loop. The only change is the `intercepted` flag that is sent to true to prevent from the switch to intercept the packet again.

htlcswitch/switch.go
```diff
@@ -1682,6 +1774,12 @@ out:
 		// packet concretely, then either forward it along, or
 		// interpret a return packet to a locally initialized one.
 		case cmd := <-s.htlcPlex:
+			// We allow first to external handlers to intercept this packet.
+			// In case it was intercepted we skip the default paket handling.
+			if !cmd.pkt.intercepted && s.interceptForward(cmd.pkt) {
+				cmd.err <- nil
```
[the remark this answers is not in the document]
roeierez: In my understanding the errors in this case are handled here: https://github.com/lightningnetwork/lnd/blob/50ddb53e28f45d5f8c50031e9dd59a7ae2eeb0ab/htlcswitch/switch.go#L2004
And the current implementation only logs them.
It seems ok to to return them to the caller of `resume` in our case. Maybe add log as well to preserver the original behavior?

lnrpc/routerrpc/router_server.go
```diff
@@ -656,3 +669,192 @@ func (s *Server) SubscribeHtlcEvents(req *SubscribeHtlcEventsRequest,
...
+				CircuitKey: &CircuitKey{
+					IncomingChanId: inKey.ChanID.ToUint64(),
+					HtlcId:         inKey.HtlcID,
+				},
+				HtlcPaymentHash: htlc.PaymentHash[:],
+				AmountSat:       uint64(htlc.Amount.ToSatoshis()),
+				Expiry:          htlc.Expiry,
+			}
+
+			if err := stream.Send(interceptionRequest); err != nil {
```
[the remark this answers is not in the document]
roeierez: I understand the concern here.
Interceptions always delegate some responsibility to the user and if we want to mitigate bad use of this API it will come with the price of complicating the switch implementation which I am not sure worth it.
I tend to think that the way to handle this is to use small timeout by default and document it to express this concern.
The API should be used with care and the caller must have all the information ready for the decision on whether to intercept and avoid depending on external resources that may block.
What do you think?

htlcswitch/switch.go
```diff
@@ -1370,7 +1450,19 @@ func (s *Switch) closeCircuit(pkt *htlcPacket) (*PaymentCircuit, error) {
...
 	// Otherwise, this is packet was received from the remote party.  Use
 	// circuit map to find the incoming link to receive the settle/fail.
-	circuit, err := s.circuits.CloseCircuit(pkt.outKey())
+	var circuit *PaymentCircuit
+	var err error
+	if pkt.intercepted {
+		// In case we got the circuitInKey populated it means that we should
+		// use it to find the circuit as there is no open circuit probably due
+		// to interception of this packet.
+		circuit = s.circuits.LookupCircuit(pkt.circuit.InKey())
```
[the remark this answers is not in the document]
roeierez: After restart the link re-sends all forwarded htlc to the switch (https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L2463)
So they should re-offered to the interceptor.

lnrpc/routerrpc/router_server.go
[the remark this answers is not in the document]
joostjager: I was thinking about using a buffered `Queue`. Could that work? So that the main loop can keep pushing htlcs into the queue and move on.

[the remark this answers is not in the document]
roeierez: So the concern is not only blocking htlc handling system-wide but also blocking other channels that are polled in the main loop right?
Buffered queue can help here unless the system load is such that the queue ends up full and this also means that these buffered packets need to be re-send to the main loop later.
If that is the case why not introduce a separate channel for fresh forwarded htlcs that will be polled in a separate go routine? 
This go routine will only handle the interception without blocking the main loop.
After the interception handling the packets will be routed to the main loop to be processed regularly as today.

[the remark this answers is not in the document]
roeierez: Alternatively we can add the interception code in `route` & `routeAsync` which are the only ones that send forwards to the main loop and are executed in the caller go routine.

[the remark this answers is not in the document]
joostjager: I think it is the same thing, blocking channel receives in the main loop is blocked htlc handling system-wide. In the PR as is, it seems that the main loop can be blocked on a network send to the client. Hand-off to a queue can disconnect the main loop from the rpc code. With queue I don't mean a regular queue, but `queue.ConcurrentQueue`. That already contains its own goroutine, which I think is also what you are suggesting with separate channel.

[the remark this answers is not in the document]
roeierez: @joostjager I pushed an alternative implementation that introduces a very minimal change to the switch and uses a kind of Proxy structure that wraps the switch and intercepts requests.

In fact the only change to the switch is because settled/failed packets that are forwarded from the interceptor don't have an open circuit which the code expects.
I think conceptually it is not a problem but as part of reviewing this approach we need to find the right solution for that as I am not sure if the current change doesn't comes with the tradeoff of not filtering duplicate settles/failures. 

The blocking issue is now solved without the need for buffering since the ForwardPackets interface method is supposed to be called (and called) in its own go routine.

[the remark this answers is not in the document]
joostjager: >In fact the only change to the switch is because settled/failed packets that are forwarded from the interceptor don't have an open circuit which the code expects.

Isn't it possible to also not let the return path flow through the switch? So return the failure directly from the `ForwardInterceptor` to the link?

> The blocking issue is now solved without the need for buffering since the ForwardPackets interface method is supposed to be called (and called) in its own go routine.

The only thing remaining is that one packet in the forwarding batch can block the others. But that is much better than a system-wide block that was introduced in the prior version.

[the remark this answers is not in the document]
roeierez: @joostjager I just pushed another version where the fails/settles from the interceptor are delivered directly to the link using the `mailOrchestrator`
This bypass the attempt to close the circuit that doesn't exist.

A few more notes:
1. For fail packets the switch attempt to encode the error which we don't populate at the moment so we don't loos anything. If in a later time we want to add a failure reason then we have the obfuscator so we can do that in the interceptor.

2. For settled packets the switch log the forward events so if this doesn't count as forward event then all is good, otherwise it is fairly easy to add this logging.

3. It looks like the `mailOrchestrator` is built for concurrent use so it seems OK to use it from the interceptor.

4. Since we isolated the blocking issue to only affect the intercepted packets and not system wide, I tend to think that we don't need to add more concurrency and it is the user responsibility to make sure the packet he intercepts do not block the others. The caller can always handle these packets in an asynchronous way.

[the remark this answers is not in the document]
joostjager: Looks good, cleaner layering.

1. But we need to do that already in this PR, right? Otherwise what will the sender of the payment see as a failure message?

2. Sounds correct. No forward happened and injecting an external preimage is a special case

3. It may be better to reference the orchestrator as a separate dependency (not via switch), but that is mostly theoretical as there is already a reference to the switch. I am still thinking a little about how this compares to handing the packet off to the link directly (assuming the link is modified to also accept direct packets, like in #4183)

4. Yes it is much safer now. But maybe still need to look at some of the failure modes like network connection lost. Does a timeout still apply?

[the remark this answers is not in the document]
roeierez: > 1. But we need to do that already in this PR, right? Otherwise what will the sender of the payment see as a failure message?

What do you think is the message type that needed to be sent? Is there anything that supports custom error message?


> 3\. I am still thinking a little about how this compares to handing the packet off to the link directly (

To be honest I think that from the same reason we didn't want the link to talk directly to the interceptor it makes sense for the interceptor to talk to the transport layer rather than find the link itself.

> 4\. But maybe still need to look at some of the failure modes like network connection lost. Does a timeout still apply?

You mean timeout in settle/fail packets? 
As far as I understand the link will re-send packets on startup and the interceptor can/should retry upon error from the orchestrator. Any specific use case you had in mind?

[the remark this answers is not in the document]
joostjager: >What do you think is the message type that needed to be sent? Is there anything that supports custom error message?

I would think `FailTemporaryChannelFailure`?

>To be honest I think that from the same reason we didn't want the link to talk directly to the interceptor it makes sense for the interceptor to talk to the transport layer rather than find the link itself.

I have no strong opinion about it. Just wondering how the "reliable" mailbox (as long as no restarts) vs direct hand off would compare.

>You mean timeout in settle/fail packets? As far as I understand the link will re-send packets on startup and the interceptor can/should retry upon error from the orchestrator. Any specific use case you had in mind?

I mean what happens if the connection to an application is lost and the packet that is considered for forwarding doesn't get a response over the rpc. Also no connection lost, just nothing happens. And then if the application comes back online, what state does it need to signal forwarding for the 'hanging' forward request?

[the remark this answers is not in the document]
roeierez: > Just wondering how the "reliable" mailbox (as long as no restarts) vs direct hand off would compare

Using the mailbox let me avoid handling the case where the link is down. The mailbox is reliable in a sense that it will attempt to deliver the packet when the link comes live again.

> I would think `FailTemporaryChannelFailure`?

Added encryption.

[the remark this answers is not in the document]
joostjager: And if it intercepts, but nothing is ever heard back? I guess that is the application's responsibility then. Similar to how hodl invoices can currently cause channels to close if no resolution is given

htlcinterceptor/interface.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/lightningnetwork/lnd/channeldb"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// ForwardHtlcHandler is a method that is invoked from the switch for every
+// incoming htlc that is intended to be forwarded.
+// The return value indicates if this handler will take control of this forward
+// and resolve it later or let the switch execute its default behavior.
+type ForwardHtlcHandler func(inKey channeldb.CircuitKey,
```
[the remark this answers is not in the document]
roeierez: Isn't it the current behavior? In case false is returned the switch executes its default forward. In case true the interceptor takes control.

htlcinterceptor/interface.go
```diff
@@ -0,0 +1,52 @@
...
+	req lnwire.UpdateAddHTLC) bool
+
+// ForwardResolver is the resolver interface that is intended to be called from
+// outside the switch to resolve an existing forward.
+type ForwardResolver interface {
+	// Resume notifies the intention to resume an existing hold forward. This
+	// basically means the caller wants to resume with the default behavior for
+	// this htlc which usually means forward it.
+	Resume() error
```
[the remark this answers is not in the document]
roeierez: The `ForwardResolver` is already associated internally with a specific htlc (and circuit key).

htlcinterceptor/interface.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Settle notifies the intention to settle an existing hold
+	// forward with a given preimage.
+	Settle(lntypes.Preimage) error
```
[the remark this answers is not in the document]
roeierez: See my comment above, a specific circuit key is associated with this `ForwardResolver` so the resolution only affect that one.

htlcinterceptor/middleware.go
```diff
@@ -0,0 +1,147 @@
...
+// * ResumeHoldForward
+// * SettleHoldForward
+// * FailHoldForward.
+func (c *Middleware) InterceptForwardHtlc(
+	circuitKey channeldb.CircuitKey,
+	htlc lnwire.UpdateAddHTLC, resolver ForwardResolver) bool {
+
+	c.Lock()
+	defer c.Unlock()
+	for _, actionResolver := range c.fwdHandlers {
```
joostjager, line 78: Are multiple handlers even needed at all for the described use cases? Without it, the `Middleware` may not be needed at all, simplifying the first version of this

roeierez: @joostjager I think this is a valid point. This can be simplified to support only one interceptor. In case an RPC interceptor is already registered, a new RPC one would probably need to disconnect the previous stream or reject the new registration.
@Roasbeef what do you think? I prefer we all agree on this before going for it.

roeierez: I will simplify it now. We can only support more interceptors later.

lnrpc/routerrpc/router_server.go
```diff
@@ -656,3 +669,192 @@ func (s *Server) SubscribeHtlcEvents(req *SubscribeHtlcEventsRequest,
...
+			return false
+		case <-s.quit:
+			return false
+		}
+	}
+
+	// Add our forwardHandler to the Middleware.
+	interceptMiddleware := s.cfg.RouterBackend.HtlcInterceptMiddleware
+	handlerID := interceptMiddleware.AddForwardHtlcHandler(forwardHandler)
+	defer interceptMiddleware.RemoveForwardHtlcHandler(handlerID)
```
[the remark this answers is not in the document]
roeierez: > I think we still need to cancel back any outstanding HTLC resolution intents.

I understand the need to do some action for outstanding HTLC , I just wondering if the default behavior shouldn't be `Resume` rather than `Cancel`.
This is mainly because I don't see a different between timeout, disconnection or no interceptor at all.

htlcinterceptor/middleware.go
roeierez: I have pushed a new version that support only one interceptor at a time.
This simplifies the implementation as I removed the middleware and replaced it by a thin implementation at the RPC layer. 
One more change was to use the same bidirectional stream for the client to resolve packets.

So in general the flows goes like this:
1. Link forward a packet, packet intercepted and sent to the RPC layer.
2. For each packet the RPC layer do the following:

- Hold the packet immediately and signal hold backwards to the switch.
- Send the packet for the client

3. In parallel RPC listens to client resolutions responses on the same stream and resolve packets accordingly.
4. On disconnection all remaining held packets are resolved using the default behavior (currently `Resume`)

htlcswitch/InterceptableSwitch.go
```diff
@@ -0,0 +1,168 @@
...
+	s.Unlock()
+
+	//optimize for the case we don't have an interceptor.
+	if interceptor == nil {
+		return s.htlcSwitch.ForwardPackets(linkQuit, packets...)
+	}
+
+	var notIntercepted []*htlcPacket
+	for _, p := range packets {
+		if !s.interceptForward(p, interceptor, linkQuit) {
```
joostjager, line 72: Shouldn't somehow the return value of the delayed `ForwardPackets` call in the interceptor be tunneled into the error channel that is returned in this function?

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// resolveFromClient handles a resolution arrived from the client.
+func (r *rpcInterceptor) resolveFromClient(in *ForwardHtlcInterceptResponse) error {
+	circuitKey := channeldb.CircuitKey{
+		ChanID: lnwire.NewShortChanIDFromInt(in.CircuitKey.IncomingChanId),
+		HtlcID: in.CircuitKey.HtlcId,
+	}
+	var interceptedForward htlcswitch.InterceptedForward
+	r.Lock()
```
joostjager, line 219: Is there a way to avoid this lock? There is already a main message loop to synchronize access

htlcswitch/InterceptableSwitch.go
```diff
@@ -0,0 +1,168 @@
...
+
+		intercepted := &interceptedForward{
+			linkQuit:   linkQuit,
+			htlc:       htlc,
+			packet:     packet,
+			htlcSwitch: s.htlcSwitch,
+		}
+
+		// If this htlc was intercepted, don't handle the forward.
+		if interceptor(intercepted) {
```
joostjager, line 101: `return interceptor(...)`

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	rpcInterceptor := interceptor
+	s.currentRPCInterceptor = rpcInterceptor
+	return nil
+}
+
+// onInterceptorDisconnected removes all previousely held forwards from
+// the store. Before they are removed it ensure to resume as the default
+// behavior.
+func (s *Server) onInterceptorDisconnected(interceptor *rpcInterceptor) {
```
joostjager, line 141: interceptor is available as a field on `s`

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,259 @@
...
+			return false
+		}
+	}
+
+	// Register our interceptor and make sure we resolve the held packets
+	// on disconnect.
+	interceptableForwarder := s.cfg.RouterBackend.InterceptableForwarder
+	interceptableForwarder.SetInterceptor(fromSwitchInterceptor)
+
+	for {
```
joostjager, line 101: Perhaps the main loop would also work well as a method on `rpcInterceptor`

htlcswitch/InterceptableSwitch.go
roeierez: I changed the interface of `ForwardPackets` to return error. The behavior didn't change as the previous error channel was only used to log the errors and I added that logging.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Yes, good one. I moved the previous `hold` call to the main loop and now the `rpcInterceptor` is only used within the main loop. I also added a comment that it should only be called from one go routine.

htlcswitch/InterceptableSwitch.go
roeierez: Fixed.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Yes, used that.

lnrpc/routerrpc/forward_interceptor.go
roeierez: I understand the motivation and actually tried that initially and TBH it didn't look better. The function selects from 4 channels that the underline `rpcInterceptor` would need access to...

htlcswitch/switch.go
```diff
@@ -621,7 +604,7 @@ func (s *Switch) ForwardPackets(linkQuit chan struct{},
 		default:
 			err := s.routeAsync(packet, fwdChan, linkQuit)
 			if err != nil {
-				return errChan
+				return fmt.Errorf("failed to forward packet %v", err)
```
joostjager, line 607: Hmm this is an error that previously disappeared?

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+func (s *Server) setRPCInterceptor(interceptor *rpcInterceptor) error {
+	s.rpcInterceptorMtx.Lock()
+	defer s.rpcInterceptorMtx.Unlock()
+	if s.currentRPCInterceptor != nil {
+		return ErrInterceptorAlreadyExists
+	}
+	rpcInterceptor := interceptor
```
joostjager, line 132: Why this assignment?

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+	interceptableForwarder.SetInterceptor(fromSwitchInterceptor)
+
+	for {
+		select {
+		case intercepted := <-interceptedPacketsChan:
+			rpcInterceptor.hold(intercepted)
+			log.Tracef("sending intercepted packet to client %v", intercepted)
+			// in case we couldn't forward we exit the loop and drain the
+			// current interceptor as this indicates on a connection problem.
+			if err := rpcInterceptor.forwardToClient(intercepted); err != nil {
```
joostjager, line 106: I think `hold` and `forwardToClient` can be merged into a single function? They are only called here in sequence

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+// drain removes all previousely held forwards and allows to pass a callback
+// that will be called for each one before it is removed.
+func (r *rpcInterceptor) drain(
+	resolver func(forward htlcswitch.InterceptedForward)) {
+
+	for key, forward := range r.holdForwards {
+		resolver(forward)
```
joostjager, line 250: Why can't this just be `forward.Resume()`?

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+		CircuitKey: &CircuitKey{
+			IncomingChanId: inKey.ChanID.ToUint64(),
+			HtlcId:         inKey.HtlcID,
+		},
+		HtlcPaymentHash: htlc.PaymentHash[:],
+		AmountSat:       uint64(htlc.Amount.ToSatoshis()),
+		Expiry:          htlc.Expiry,
+	}
+
+	return r.stream.Send(interceptionRequest)
```
joostjager, line 208: It feels that the separation of responsibilties in this file `forward_interceptor.go` isn't optimal yet. We are doing sending here, but receive happens outside of the `rpcInterceptor` object.

htlcswitch/switch.go
roeierez: Yes but it only happened if the switch or the link has been shut down.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Removed.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Indeed, merged.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Left over from a previous iteration. Simplified now.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Thanks for the feedback @joostjager. 
In my eyes the main loop is a thin layer that is responsible to consuming input of both sides (switch and the client stream) and delegates the handling of the packets to the `rpcInterceptor` for the actual work. In my mind the "Send" to the stream is equivalent to the "Settle/Fail" on the other side and both are done in the `rpcInterceptor`.
Of course I am opened to suggestions on how to optimize it.

lnrpc/routerrpc/router.proto
```diff
@@ -579,3 +588,55 @@ message PaymentStatus {
...
+    /** 
+    The key of this forwarded htlc. It defines the incoming channel id and
+    the index in this channel.
+    */
+    CircuitKey circuit_key = 1;
+
+    /// The htlc payment hash.
```
cfromknecht, line 606: add note that this value may not be unique?

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,247 @@
...
+
+	interceptor.drain()
+}
+
+// rpcInterceptor is a helper struct that handles the lifecycle of an rpc
+// interceptor streaming session.
+// It is created when the stream opens and drains when the stream closes.
+// It should be used from the same go routine and not built for concurrent
+// access.
+type rpcInterceptor struct {
```
joostjager, line 162: I would also move this into a separate file. Just easier to navigate

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,247 @@
...
+func newRPCInterceptor(stream Router_HtlcInterceptorServer) *rpcInterceptor {
+	return &rpcInterceptor{
+		stream: stream,
+		holdForwards: make(
+			map[channeldb.CircuitKey]htlcswitch.InterceptedForward),
+	}
+}
+
+// hold just store the intercepted forward.
+func (r *rpcInterceptor) hold(forward htlcswitch.InterceptedForward) {
```
joostjager, line 182: method unused now?

lnrpc/routerrpc/forward_interceptor.go
joostjager: I looked at I again and maybe you should just move almost everything from `Server.HtlcInterceptor` into `rpcInterceptor`.

htlcswitch/InterceptableSwitch.go
```diff
@@ -0,0 +1,168 @@
...
+
+// ForwardPackets attempts to forward the batch of htlcs through the
+// switch, any failed packets will be returned to the provided
+// ChannelLink. The link's quit signal should be provided to allow
+// cancellation of forwarding during link shutdown.
+func (s *InterceptableSwitch) ForwardPackets(linkQuit chan struct{},
+	packets ...*htlcPacket) chan error {
+
+	var interceptor ForwardInterceptor
+	s.Lock()
```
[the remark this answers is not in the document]
roeierez: Removed use of Mutex.

lnrpc/routerrpc/router.proto
```diff
@@ -579,3 +588,55 @@ message PaymentStatus {
     repeated lnrpc.HTLCAttempt htlcs = 4;
 }
 
+message CircuitKey {    
+    /// The id of the channel that the htlc arrived at our node on.
+    uint64 incoming_chan_id = 1;
```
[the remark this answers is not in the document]
roeierez: Agree, fixed also `ForwardHtlcInterceptRequest` & `ForwardHtlcInterceptResponse`

lnrpc/routerrpc/router.proto
```diff
@@ -579,3 +588,55 @@ message PaymentStatus {
...
+    The key of this forwarded htlc. It defines the incoming channel id and
+    the index in this channel.
+    */
+    CircuitKey circuit_key = 1;
+
+    /// The htlc payment hash.
+    bytes htlc_payment_hash = 2;
+
+    /// The htlc amount.
+    uint64 amount_sat = 3;
```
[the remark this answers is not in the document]
roeierez: Changed the amount to msat.

lnrpc/routerrpc/router.proto
```diff
@@ -579,3 +588,55 @@ message PaymentStatus {
...
+    The key of this forwarded htlc. It defines the incoming channel id and
+    the index in this channel.
+    */
+    CircuitKey circuit_key = 1;
+
+    // The resolve action for this intercepted htlc.
+    ResolveHoldForwardAction action = 2;
+
+    // The preimage in case the resolve action is Settle.
+    bytes r_preimage = 3;
```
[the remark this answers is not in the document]
roeierez: Removed

lnrpc/routerrpc/router.proto
roeierez: Added.

lnrpc/routerrpc/forward_interceptor.go
roeierez: Yes, removed.

lnrpc/routerrpc/forward_interceptor.go
roeierez: I changed the RPC structure, `Server.HtlcInterceptor` only responsible for starting the session, making sure there is only one active `forwardInterceptor` and the interceptor is responsible for the rest.

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,224 @@
...
+// to read from the client stream.
+// To coordinate all this and make sure it is safe for concurrent access all
+// packets are sent to the main where they are handled.
+func (r *forwardInterceptor) run() error {
+	// make sure we disconnect and resolves all remaining packets if any.
+	defer r.onDisconnect()
+
+	// Register our interceptor so we receive all forwarded packets.
+	interceptableForwarder := r.server.cfg.RouterBackend.InterceptableForwarder
+	interceptableForwarder.SetInterceptor(r.onIntercept)
```
joostjager, line 74: It is nice to have the set and clear on the same level. Maybe take the clear out of disconnect and do an additional defer call?

lnrpc/routerrpc/forward_interceptor.go
joostjager: Much improved again

lnrpc/routerrpc/router.proto
```diff
@@ -97,6 +97,15 @@ service Router {
...
     }
+    
+    /**
+    HtlcInterceptor dispatches a bi-directional streaming RPC in which
+    Forwarded HTLC requests are sent to the client and the client responds with
+    a boolean that tells LND if this htlc should be intercepted.
+    In case of interception, the htlc can be either settled, cancelled or
+    resumed later by using the ResolveHoldForward endpoint.
+    */
+    rpc HtlcInterceptor (stream ForwardHtlcInterceptResponse) returns (stream ForwardHtlcInterceptRequest);
```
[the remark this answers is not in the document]
roeierez: Maybe there is a place for another subserver and I agree this PR alone doesn't yet justify it.
BTW currently the `routerrpc` is not totally mapped to the sender side as it contains the 
`SubscribeHtlcEvents` RPC endpoint that notifies about forward events.

Roasbeef:
Let me know when I should do another pass over this. Great work so far @roeierez!

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
cfromknecht: good point, yes then definitely we can reconsider moving these at a later point

htlcswitch/InterceptableSwitch.go
```diff
@@ -0,0 +1,168 @@
+package htlcswitch
```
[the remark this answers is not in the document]
cfromknecht: `I` in interceptable still capitalized :)

roeierez:
> Let me know when I should do another pass over this. Great work so far @roeierez!

Thanks. At this point I am only aware of missing  itest which I am working on so I think it is ready for another pass.

htlcswitch/InterceptableSwitch.go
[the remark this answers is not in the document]
roeierez: :open_mouth: Fixed again.

htlcswitch/link.go
```diff
@@ -135,10 +135,10 @@ type ChannelLinkConfig struct {
 	Switch *Switch
 
 	// ForwardPackets attempts to forward the batch of htlcs through the
-	// switch, any failed packets will be returned to the provided
-	// ChannelLink. The link's quit signal should be provided to allow
+	// switch. The function returns and error in case it fails to send one or
+	// more packets. The link's quit signal should be provided to allow
 	// cancellation of forwarding during link shutdown.
-	ForwardPackets func(chan struct{}, ...*htlcPacket) chan error
+	ForwardPackets func(chan struct{}, ...*htlcPacket) error
```
Roasbeef, line 141: 👍  for removing error channels

htlcswitch/interceptable_switch.go
```diff
@@ -0,0 +1,164 @@
...
+	fwdInterceptor ForwardInterceptor
+}
+
+// NewInterceptableSwitch returns an instance of InterceptableSwitch.
+func NewInterceptableSwitch(s *Switch) *InterceptableSwitch {
+	return &InterceptableSwitch{htlcSwitch: s}
+}
+
+// SetInterceptor sets the ForwardInterceptor to be used.
+func (s *InterceptableSwitch) SetInterceptor(
```
Roasbeef, line 45: As we intend to (atm) only have a global interceptor, perhaps this should return an error if one is already set. Alternatively, this can be handled at the RPC level.

htlcswitch/interceptable_switch.go
```diff
@@ -0,0 +1,164 @@
...
+// cancellation of forwarding during link shutdown.
+func (s *InterceptableSwitch) ForwardPackets(linkQuit chan struct{},
+	packets ...*htlcPacket) error {
+
+	var interceptor ForwardInterceptor
+	s.Lock()
+	interceptor = s.fwdInterceptor
+	s.Unlock()
+
+	//optimize for the case we don't have an interceptor.
```
Roasbeef, line 65: Style nit: optimize should be capitalized, and should be a space in front of it.

htlcswitch/interceptable_switch.go
```diff
@@ -0,0 +1,164 @@
...
+		return interceptor(intercepted)
+	default:
+		return false
+	}
+}
+
+// interceptedForward implements the InterceptedForward interface.
+// It is passed from the switch to external interceptors that are interested
+// in holding forwards and resolve them manually.
+type interceptedForward struct {
```
Roasbeef, line 110: Very nice, this logic is now much easier to follow IMO with the approach of passing around "interception intents".

htlcswitch/interceptable_switch.go
```diff
@@ -0,0 +1,164 @@
...
+	htlc       *lnwire.UpdateAddHTLC
+	packet     *htlcPacket
+	htlcSwitch *Switch
+}
+
+func (f *interceptedForward) Packet() lnwire.UpdateAddHTLC {
+	return *f.htlc
+}
+
+func (f *interceptedForward) CircuitKey() channeldb.CircuitKey {
```
Roasbeef, line 121: Missing `godoc` comments here and above.

htlcswitch/interceptable_switch.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// Settle forwards a settled packet to the switch.
+func (f *interceptedForward) Settle(preimage lntypes.Preimage) error {
+	return f.resolve(&lnwire.UpdateFulfillHTLC{
+		PaymentPreimage: preimage,
+	})
+}
+
+// resolve is used for both Settle and Fail and forwards the message to the
```
Roasbeef, line 151: Incomplete sentence here.

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,224 @@
...
+	// ErrMissingPreimage is an error returned when the caller tries to settle
+	// a forward and doesn't provide a preimage.
+	ErrMissingPreimage = errors.New("missing preimage")
+)
+
+// forwardInterceptor is a helper struct that handles the lifecycle of an rpc
+// interceptor streaming session.
+// It is created when the stream opens and disconnects when the stream closes.
+type forwardInterceptor struct {
+
```
Roasbeef, line 28: Style nit: extra space.

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,224 @@
...
+	// First de-register the current interceptor so we'll stop getting
+	// more interceptions.
+	interceptableForwarder := r.server.cfg.RouterBackend.InterceptableForwarder
+	interceptableForwarder.SetInterceptor(nil)
+
+	// Then close the channel so all go routine will exit.
+	close(r.quit)
+
+	log.Infof("RPC interceptor disconnected, resolving held packets")
+	for key, forward := range r.holdForwards {
```
Roasbeef, line 217: 👍

lnrpc/routerrpc/router.proto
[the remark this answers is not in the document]
Roasbeef: Yeah I think this is just the current "best" place we have atm. Could say a similar thing about the HTLC even notification end point.

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,224 @@
...
+	switch in.Action {
+	case ResolveHoldForwardAction_RESUME:
+		return interceptedForward.Resume()
+	case ResolveHoldForwardAction_FAIL:
+		return interceptedForward.Fail()
+	case ResolveHoldForwardAction_SETTLE:
+		if in.Preimage == nil {
+			return ErrMissingPreimage
+		}
+		preimage, err := lntypes.MakePreimage(in.Preimage)
```
Roasbeef, line 193: We should catch an _invalid_ pre-image for the payment hash here early (so verify the pay hash and preimage match up). Otherwise, a user could mess up and cause a channel force close since that's considered a dire offense atm.

lnrpc/routerrpc/router_server.go
```diff
@@ -586,3 +596,22 @@ func (s *Server) SubscribeHtlcEvents(req *SubscribeHtlcEventsRequest,
...
+// requests to the caller.
+// Upon connection it does the following:
+// 1. Check if there is already a live stream, if yes it rejects the request.
+// 2. Regsitered a ForwardInterceptor
+// 3. Delivers to the caller every √√ and detect his answer.
+// It uses a local implementation of holdForwardsStore to keep all the hold
+// forwards and find them when manual resolution is later needed.
+func (s *Server) HtlcInterceptor(stream Router_HtlcInterceptorServer) error {
+	// We ensure there is only one interceptor at a time.
+	if !atomic.CompareAndSwapInt32(&s.forwardInterceptorActive, 0, 1) {
```
Roasbeef, line 610: Nice, can ignore my other comment about enforcing a singleton interceptor.

lntest/itest/log_error_whitelist.txt
```diff
@@ -188,3 +188,4 @@
 <time> [ERR] FNDG: Unable to advance state(<chan_point>): failed adding to router graph: error sending channel announcement: gossiper is shutting down
 <time> [ERR] PEER: unable to close channel, ChannelID(<hex>) is unknown
 <time> [ERR] HSWC: ChannelLink(<chan>): unable to update signals
+<time> [ERR] RPCS: [/routerrpc.Router/HtlcInterceptor]: rpc error: code = Canceled desc = context canceled
```
joostjager, line 191: If this is newly introduced, it may be worth to log context canceled specifically as warn. A tiny step towards getting rid of all errors that aren't really errors.

lntest/itest/lnd_forward_interceptor_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+	// initialize the test context with 3 connected nodes.
+	testContext := newInterceptorTestContext(t, net)
+	defer testContext.shutdownNodes()
+
+	const (
+		chanAmt = btcutil.Amount(300000)
+	)
+
+	// open and wait for channels.
```
joostjager, line 48: (Implicit) standard is to start sentences with a capital.

lntest/itest/lnd_forward_interceptor_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+	// run the interceptor loop in its own go routine.
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			request, err := interceptor.Recv()
+			if err != nil {
+				return
```
joostjager, line 80: Any other action needed here? Log or fail test?

lntest/itest/lnd_forward_interceptor_test.go
```diff
@@ -0,0 +1,389 @@
...
+					IncomingCircuitKey: request.IncomingCircuitKey,
+					Action:             routerrpc.ResolveHoldForwardAction_RESUME,
+				})
+			case routerrpc.ResolveHoldForwardAction_SETTLE:
+				interceptor.Send(&routerrpc.ForwardHtlcInterceptResponse{
+					IncomingCircuitKey: request.IncomingCircuitKey,
+					Action:             routerrpc.ResolveHoldForwardAction_SETTLE,
+					Preimage:           testCase.invoice.RPreimage,
+				})
+			}
```
joostjager, line 150: Is this switch needed at all? Maybe the test case parameters can be copied straight into the `ForwardHtlcInterceptResponse`?

lnrpc/routerrpc/router.proto
champo: @Roasbeef in our case, the end node and the interceptor communicate over HTTP to negotiate the settling of the HTLC. The end node can then decode the sphinx packet to reduce trust.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: In case this error is a result of context cancellation, nothing is needed as this is the test behavior. In other error cases it is better to fail the test. I added the code to support this behavior.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: Yes, much cleaner. Done.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: Changed all over the file.

lntest/itest/log_error_whitelist.txt
roeierez: This is not something added explicitly. I think there is somewhere a code that logs every RPC call that returns with error. In this case the error is the context cancellation when the stream is closed.

lnrpc/routerrpc/forward_interceptor.go
joostjager: Still interested to hear your thought on this. It is a non-blocking suggestion though

lnrpc/routerrpc/forward_interceptor.go
joostjager: Looks good now

lnrpc/routerrpc/router_server.go
```diff
@@ -656,3 +669,192 @@ func (s *Server) SubscribeHtlcEvents(req *SubscribeHtlcEventsRequest,
...
+// htlcInterceptInfo is used in the HtlcInterceptor bidirectional stream and
+// encapsulates the request information sent from the HtlcForwardHandler to the
+// RPCServer.
+type htlcInterceptInfo struct {
+	inKey        channeldb.CircuitKey
+	htlc         lnwire.UpdateAddHTLC
+	responseChan chan bool
+}
+
+func (s *Server) HtlcInterceptor(stream Router_HtlcInterceptorServer) error {
```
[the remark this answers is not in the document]
joostjager: Resolved, the method isn't long anymore

lntest/itest/lnd_forward_interceptor_test.go
joostjager: No capital here

lntest/itest/lnd_forward_interceptor_test.go
```diff
@@ -0,0 +1,386 @@
...
+// Alice sends 4 different payments to Carol while the interceptor handles
+// differently the htlcs.
+// The test ensures that:
+// 1. Intercepted failed htlcs result in no payment (invoice is not settled).
+// 2. Intercepted resumed htlcs result in a payment (invoice is settled).
+// 3. Intercepted held htlcs result in no payment (invoice is not settled).
+// 4. When Interceptor disconnects it resumes all held htlcs, which result in
+//    valid payment (invoice is settled).
+func testForwardInterceptor(net *lntest.NetworkHarness, t *harnessTest) {
+
```
joostjager, line 41: Remove \n

lntest/itest/lnd_forward_interceptor_test.go
joostjager: To problem is that calling `t.Fatalf` isn't allowed in a goroutine

lntest/itest/log_error_whitelist.txt
joostjager: Ok, out of scope for this pr then

lntest/itest/lnd_forward_interceptor_test.go
```diff
@@ -0,0 +1,386 @@
...
+			switch tc.interceptorAction {
+			// For 'fail' interceptor action we make sure the payment failed.
+			case routerrpc.ResolveHoldForwardAction_FAIL:
+				if attempt.Status != lnrpc.HTLCAttempt_FAILED {
+					t.Fatalf("expected payment to fail, instead got %v", attempt.Status)
+				}
+			// For settle and resume we make sure the payment is successfull.
+			case routerrpc.ResolveHoldForwardAction_SETTLE:
+				fallthrough
+			case routerrpc.ResolveHoldForwardAction_RESUME:
```
joostjager, line 122: Newlines between cases

joostjager:
Proof of concept for rate limiting in Faraday: https://github.com/lightninglabs/faraday/pull/47

htlcswitch/link.go
```diff
@@ -2900,8 +2900,10 @@ func (l *channelLink) forwardBatch(packets ...*htlcPacket) {
 		filteredPkts = append(filteredPkts, pkt)
 	}
 
-	errChan := l.cfg.ForwardPackets(l.quit, filteredPkts...)
-	go handleBatchFwdErrs(errChan, l.log)
+	if err := l.cfg.ForwardPackets(l.quit, filteredPkts...); err != nil {
+		log.Errorf("Unhandled error while reforwarding htlc "+
```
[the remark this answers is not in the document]
roeierez: Missed this question... I believe the answer is no but since this commit is a pure refactoring I prefer to maintain the previous behavior.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: Fixed.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: Fixed.

htlcswitch/switch_test.go
```diff
@@ -1634,18 +1676,30 @@ func testSkipIneligibleLinksMultiHopForward(t *testing.T,
...
+		t.Fatal(err)
+	}
+	var linkError *LinkError
+	select {
+	case p := <-aliceChannelLink.packets:
+		linkError = p.linkFailure
+	case p := <-bobChannelLink1.packets:
+		linkError = p.linkFailure
+	case p := <-bobChannelLink2.packets:
+		linkError = p.linkFailure
```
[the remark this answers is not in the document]
roeierez: Added comment.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: Fixed.

lntest/itest/lnd_forward_interceptor_test.go
roeierez: Changed to `Errorf`

lnrpc/routerrpc/forward_interceptor.go
roeierez: > Still interested to hear your thought on this. It is a non-blocking suggestion though

I agree it is better so I changed it.

htlcswitch/interceptable_switch.go
roeierez: Fixed

htlcswitch/interceptable_switch.go
roeierez: Added

lnrpc/routerrpc/forward_interceptor.go
roeierez: Done.

lnrpc/rest-annotations.yaml
```diff
@@ -211,6 +211,8 @@ http:
       # deprecated, no REST endpoint
     - selector: routerrpc.Router.TrackPayment
       # deprecated, no REST endpoint
+    - selector: routerrpc.HtlcInterceptor
```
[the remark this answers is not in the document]
roeierez: I added it because it is required by the `make rpc-check`

roeierez:
> Don't think we need the REST annotation since this type of RPC isn't supported atm

The build requires it atm: `make rpc-check`


> ready to land after a rebase!

Rebased.

lntest/itest/log_error_whitelist.txt
```diff
@@ -89,13 +89,13 @@
...
-<time> [ERR] HSWC: ChannelLink(<chan>): Unhandled error while reforwarding htlc settle/fail over htlcswitch: circuit has already been closed
-<time> [ERR] HSWC: ChannelLink(<chan>): Unhandled error while reforwarding htlc settle/fail over htlcswitch: FeeInsufficient(htlc_amt==<amt>, update=(lnwire.ChannelUpdate) {
-<time> [ERR] HSWC: ChannelLink(<chan>): Unhandled error while reforwarding htlc settle/fail over htlcswitch: insufficient bandwidth to route htlc
-<time> [ERR] HSWC: ChannelLink(<chan>): Unhandled error while reforwarding htlc settle/fail over htlcswitch: node configured to disallow forwards
-<time> [ERR] HSWC: ChannelLink(<chan>): Unhandled error while reforwarding htlc settle/fail over htlcswitch: UnknownNextPeer
 <time> [ERR] HSWC: ChannelLink(<chan>): failing link: unable to synchronize channel states: unable to send chan sync message for ChannelPoint(<chan_point>): set tcp <ip>: use of closed network connection with error: unable to resume channel, recovery required
+<time> [ERR] HSWC: Unhandled error while reforwarding htlc settle/fail over htlcswitch: AmountBelowMinimum(amt=<amt>, update=(lnwire.ChannelUpdate) {
```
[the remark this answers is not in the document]
joostjager: Ah, the error changed and it is sorted

htlcswitch/link.go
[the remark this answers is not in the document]
joostjager: Yes valid point

lnrpc/routerrpc/forward_interceptor.go
```diff
@@ -0,0 +1,218 @@
...
+		if err != nil {
+			return err
+		}
+		err = interceptedForward.Settle(preimage)
```
joostjager, line 196: `return interceptedForward.Settle(preimage)`

lnrpc/routerrpc/router_server.go
joostjager: I think it is still useful to make sure that any user of interceptable switch uses it correctly. But non-blocking

lnrpc/routerrpc/forward_interceptor.go
roeierez: Fixed.

lnrpc/routerrpc/router.proto
```diff
@@ -98,6 +98,15 @@ service Router {
     rpc TrackPayment (TrackPaymentRequest) returns (stream PaymentStatus) {
         option deprecated = true;
     }
+    
+    /**
+    HtlcInterceptor dispatches a bi-directional streaming RPC in which
+    Forwarded HTLC requests are sent to the client and the client responds with
+    a boolean that tells LND if this htlc should be intercepted.
+    In case of interception, the htlc can be either settled, cancelled or
+    resumed later by using the ResolveHoldForward endpoint.
```
hsjoberg, line 107: Hi @roeierez. This description text looks wrong. I don't find any `ResolveHoldForward` endpoint.
AFAICT how to resume the incoming payment is by sending with `action` set to `RESUME` in this gRPC stream.  
Perhaps I'm missing something?

Also by reading the protobuf, it doesn't look lke the client is supposed to respond with a boolean if the HTLC should be intercepted or not (rather it is decided by the `incoming_circuit_key`?)

roeierez: @hsjoberg You are right. Unfortunately this description was left from a previous iteration and is indeed wrong.
The interface was changes and the client should return a ForwardHtlcInterceptResponse with action that maps to one of the follows:

**SETTLE**:  to settle the htlc (a preimage is expected as well).
**FAIL**:  to fail the htlc
**RESUME**: to resume with default behavior.

hsjoberg: @roeierez Thanks, that makes sense.

Cheers!

hsjoberg:
@roeierez Correct me if I'm wrong about this, but AFAICT as it stands right now, this API is dangerous if an amountless invoice is used (and action is `SETTLE`).  
Only the receiving node knows about `amt_to_forward` and `payment_secret` in their TLV payload, so isn't there a vulnerability here where the LSP node could give up the preimage for too little satoshi? (Basically the amountless invoice problem again.)

roeierez:
> @roeierez Correct me if I'm wrong about this, but AFAICT as it stands right now, this API is dangerous if an amountless invoice is used (and action is `SETTLE`).
> Only the receiving node knows about `amt_to_forward` and `payment_secret` in their TLV payload, so isn't there a vulnerability here where the LSP node could give up the preimage for too little satoshi? (Basically the amountless invoice problem again.)

@hsjoberg  Not following. Settling an htlc backwards requires a preimage which is passed to the interceptor by the next node in the path either using regular lightning mechanism or some custom swap logic. In any case interceptor can make sure it has the required funds before settling. Don't see any risk.

hsjoberg:
@roeierez In the case of a non-determined amount ("amountless invoice"), the interceptor node must somehow know how much it should reveal the preimage for. AFAICT basically what [feature bits 14/15](https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md) is for.

roeierez:
> @roeierez In the case of an undetermined amount ("amountless invoice"), the interceptor node must somehow know how much it should reveal the preimage for to settle the payment. AFAICT basically what [feature bits 14/15](https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md) is for.

@hsjoberg  The interceptor doesn't know the total amount of a payment regardless of amount-less invoice or not. Invoice is between the payer and the payee. Also the way the interceptor gets the pre-image to settle backwards is totally implementation dependent so in that sense it can be risky or not. 
The implementation in Breez won't change the way lightning works, means the interceptor will get the pre-image from the payee (once his invoice is settled) so I don't see how this adds risk to the interceptor node.

hsjoberg:
>  The interceptor doesn't know the total amount of a payment regardless of amount-less invoice or not. Invoice is between the payer and the payee

But that could just be trivially communicated with the interceptor somehow in the case of a normal invoice with amount.

---

@roeierez Sorry, I will try to explain my concern better. I'll take it from the beginning:
There exists a vulnerability with amountless invoices where a malicious party Mallory could attempt to make the receiver reveal the preimage by forwarding another payment with same hash but with less amount.

Let's say A wants to pay C 1 BTC through M:
A -> M -> C

Instead of forwarding the real payment, M could make a new payment with 1 sat using the same hash in an attempt lure C to reveal the preimage.

This issue has been demonstrated in the past and fixed with the 14/15 payment secret and/or enforcing TLV/MPP/keysend. My high-level understanding is that a nonce is introduced that the payer is sending to the payee, which can be used to find out tampering.
The payment secret is encrypted in a TLV for the _receiving node_.

Now let's say we're using Pay to Open, and interceptor settling the payment and opening channel to C (either paying through push amount or another invoice, shouldn't matter AFAICT).
A wants to pay C 1 BTC through M and interceptor node I. C has no inbound channel:
A->M->I->C

The malicious party M could attempt to lure I and thus also C by forwarding 1 sat instead of the real payment.
Now, this is the gut of the whole thing, and this is more of a question rather than a conclusion, but even if the new payment scheme using TLV is enforced (or 15 payment secret), lnd and the code as it stands right now wouldn't check for tampering because it won't decrypt and check the actual receiving nodes (C) packet (as I is just settling it), thus it looks vulnerable to the old amountless invoice problem.

roeierez:
> But that could just be trivially communicated with the interceptor somehow in the case of a normal invoice with amount.

Then what prevents the payee to communicate the amount he expects in the zero invoice case as well?

> as I is just settling it and not letting C see his packet

If this is the case how the interceptor has the pre-image? It seems that you described a flow where the interceptor has the pre-image beforehand and this is not how we (at Breez) implementing **Pay to Open**. We do forward the packet to the destination and only once the destination reveals the pre-image the interceptor settles backwards using regular lightning mechanism.
The interceptor, in our implementation, also agrees with the end node on the amount to forward and the LSP fee (for opening a.channel) beforehand.
So in fact the interceptor has all the means to hold the forward until he receives enough htlcs amount to take his fees, before resuming the forwards.

hsjoberg:
> Then what prevents the payee to communicate the amount he expects in the zero invoice case as well?

The payee doesn't know what amount to expect, hence amountless invoice.
The final TLV payload has to be read by either the interceptor or payee somehow, as there lies the `payment_secret` and `amt_to_forward` that the payer sent to the payee.

> If this is the case how the interceptor has the pre-image? It seems that you described a flow where the interceptor has the pre-image beforehand

The wallet simply hands the preimage to the LSP and the LSP uses the `SETTLE` action via the `HtlcIntercept` API.
The flow I am thinking about is _not_ where the interceptor continues the forwarding (`RESUME`), rather it settles it themselves (`SETTLE`), and then afterwards opens a channel with the wallet.
In short:

1. Wallet creates an invoice with route hint which goes to the LSP node with fake short chan id.
2. Wallet hands preimage to LSP
3. Payer pays invoice and uses the fake short chan id provided in route hint
4. LSP intercepts the incoming HTLC and `SETTLE`'s it
5. LSP opens a new channel to Wallet (and pays using push amount, another invoice or something else)

I explained briefly in the [OP](https://github.com/lightningnetwork/lnd/pull/4018#issuecomment-692115987) that the concern is about the `SETTLE` action.

> this is not how we (at Breez) implementing Pay to Open

I'm not talking about Breez's implementation per se but I think it should apply to the 3.2 scenario in Pay to Open described in your OP:
> The node opens a channel to Bob, pays Bob the original Alices amount minus service fee (in a different invoice), gets the preimage from Bob (in exchange to the payment) and settles the hold forward.

The interceptor requests a new invoice from the wallet to pay through and then settles the hold forward in this scenario(?). Thus the original sender's TLV payload would not reach the receiver.

> So in fact the interceptor has all the means to hold the forward until he receives enough htlcs amount to take his fees, before resuming the forwards.

Yes, if the interceptor just resumed the forwarding and uses a newly created channel to the payee, the problem I've described does not apply as 14 payment secret (or any other method) can be enforced.

Issue #619, opened 2018-01-18
[Request] sendcoins all / openchannel all


junderw:
Reason: Currently, if I want to spend 100% of my funds onchain OR spend 100% of my funds to a channel, I need to dance back and forth by overshooting it and seeing the error message telling me how much I need, subtract, then send the exact amount.

Flow
- User specifies in the args that "I want to send as much as possible accounting for fees" ( `ALL` in place of the local_amt etc.)
- lnd calculates the fees for the opening channel transaction normally, and subtracts that from available witness outputs (for openchannel) or all inputs (for sendcoins) and performs the send operation without interaction from the user.

t4sk:
I will give it crack!

githubsands:
@t4sk are you actively working on this?

t4sk:
@githubsands I am planning to work on this weekend but I am also happy to hand it over to you :)

Roasbeef:
AFAIK, this issue is still up for grabs!

simonhorlick:
This would actually be super useful. I might pick it up if nobody else has raised a PR.

simonhorlick:
For the openchannel case I feel like the main issue here is that the user has to guess the fee. For example the following flow is frustrating:
```shell
$ lncli walletbalance
{
    "total_balance": "16768865",
    "confirmed_balance": "16768865",
    "unconfirmed_balance": "0"
}
$ lncli openchannel 03dfb7ad0bda82ecb7b516a57c914bbcb71a58cf4f70d22d313731b33a9a7de3d0 16768865
[lncli] rpc error: code = Unknown desc = not enough witness outputs to create funding transaction, need 0.16776515 BTC only have 0.16768865 BTC  available
```

We could take the fee from local_amt by default, but the change of behaviour might surprise users. We could also add a new option such as --include-fee that makes it explicit that the local_amt includes the fee.

Finally, do we want to commit 100% to channel balances, or should we reserve a certain amount in case fees increase later on? From BOLT2:

> Given the variance in fees, and the fact that the transaction may be spent in the future, it's a good idea for the fee payer to keep a good margin (say 5x the expected fee requirement); but, due to differing methods of fee estimation, an exact value is not specified.

dabura667:
> a new option such as --include-fee

This should be sufficient.

> do we want to commit 100% to channel balances

This could also be another arg.

So in summary:

1. Add `--include-fee` arg to openchannel and sencoins.
[...]

PR #4024, opened 2020-02-20
Adds a new flag `commitmax` to the command `openchannel` of `lncli`

bjarnemagnussen:
### Description

This PR introduces a new flag `commitmax` for `openchannel` that for the channel capacity will use the whole available wallet balance up to the allowed maximum funding limit (currently ~0.16 btc) when opening a new channel. 

### Motivation

The command `sendcoins` has a flag `sweepall` that will spend all the coins from the wallet. However no such flag exists when opening new channels using `openchannel`.

Wanting to use the total available wallet balance in the local amount when opening a channel requires calculating the needed fee and subtracting it from the balance manually.

Further, it seems that this possibility has at least come up as most of the functionality to make this possible already exists, e.g.:
https://github.com/lightningnetwork/lnd/blob/80af2957ac37544f13b7945e5c8af1c834bce33b/lnwallet/chanfunding/assembler.go#L67-L72

### Edge Case

The `commitmax` flag makes use of the `CoinSelectSubtractFees` function to select funds. This implies that if the wallet balance is greater than the sum of the maximum allowed funding limit (~0.16 btc) and the fee, then for the new channel the capacity will be the maximum allowed (~0.16 btc) subtracted by the fee, instead of exactly the maximum allowed (~0.16 btc). This is a slightly wrong behaviour, as in this case it would have been possible to exactly max out the channel capacity and additionally have enough available funds to pay the fee.

I think a solution to this would require adding a field e.g. `CommitMax` to the `chanfunding.Request` and adding logic to the `ProvisionChannel` function inside `lnd/lnwallet/chanfunding/wallet_assembler.go` to distinguish between opening channels with `CommitMax` and without. But I think this is too much to specifically change for such an edge case.

### Disclaimer

I added this feature because I felt that such a flag was missing and could be useful. But I also implemented it just to get a better understanding of the `lnd` codebase. Let me know if this feature is useful or not for others.

### Testing

I am unsure if this change requires to add tests. But since this change does only use functionality that already existed, I don't see where tests should be added.

#### Pull Request Checklist

- [ ] For new code: Code is accompanied by tests which exercise both
  per default)
  already exist


Roasbeef:
A better way to do this is to use functionality added along side [`CreateSweepAllTx`](https://github.com/lightningnetwork/lnd/blob/master/sweep/walletsweep.go#L151) call to do it from a _coin selection_ perspective. In this case the "delivery" address is just the funding output. 

As implemented in this PR, it isn't concurrent safe, as the balance may shift when it's obtained on the CLI compared to when we actually go to fund the transaction. If the above function is used, then this isn't an issue as it locks all coin selection (and funding). 

Also a better name for this feature imo is something along the lines of `--fundall`. See this issue for some prior context as well: https://github.com/lightningnetwork/lnd/issues/619

bjarnemagnussen:
Thanks for your input @roasbeef ! Those are some great insights you gave me. I will take another look at it and try to update this PR.

bjarnemagnussen:
Will reopen PR as this has undergone major changes.

rpcserver.go
```diff
@@ -1564,6 +1564,22 @@ func (r *rpcServer) OpenChannel(in *lnrpc.OpenChannelRequest,
...
+	// If the FundAll flag is set, ensure that the
+	// acceptable minimum local amount adheres to the
+	// amount to be pushed to the remote, and to current
+	// rules.
+	if in.FundAll {
+		// Use as minmimum local amount the larger of
+		// the amount to be pushed to the remote or
+		// the allowed minimum funding size.
+		if remoteInitialBalance >= minChanFundingSize {
+			localFundingAmt = remoteInitialBalance + 1
```
ziggie1984, line 1577: would be cool to add the information in the comment which was discussed below why the +1 is needed here

PR #4025, opened 2020-02-20
makefile: upgrade linter

carlaKC:
Upgrading linter to latest to see whether it fixes the build.


carlaKC:
Nope, also doesn't build.

PR #4026, opened 2020-02-20
travis: remove git clone depth limitation to fix linter

guggero:
By default, Travis-CI does `git clone --depth=50` which limits the
repo to the last 50 commits. Because we do linting against an old
commit, it cannot be found any more. We remove the limit completely
to not run into this problem again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ cache:
     - $GOPATH/src/github.com/golang
     - $GOPATH/src/gopkg.in/alecthomas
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,6 +8,9 @@ cache:
     - $GOPATH/src/github.com/golang
     - $GOPATH/src/gopkg.in/alecthomas
 
+git:
+  depth: false
```
halseth: prob add a comment here

PR #4027, opened 2020-02-21
sweep: additionally detect our sweep transactions with the wallet

### LINKED ISSUE #3321: Don't allow manual close of restored channels
### Background

I was running LND 0.6.1 on bitcoin 0.18.0 on my old phone (Termux). Copying channel updates by inotify. Then my micro sd containing .lnd died. 

After I received my new card, I tried restoring both offchain and onchain funds using 
`lncli create -multi_file=channels.backup` (on 0.6.1 again), but LND crashed. Sorry, I didn't get any logs.

Instead, I tried the same on the just released 0.7.0. This time, it didn't crash, and the onchain balance got restored, but the channels didn't get closed. I used to have two channels open with funds on my side. One didn't show up at all, the other remained in waiting_close_channels. When I tried to force-close this one, LND crashed again, saying something along the lines of "chain-link missing".

So I tried it a third time, this time first doing just `lncli create` (with the seed), then later `lncli restorechanbackup`. This again restored the onchain funds. One channel is again in waiting_close_channels, and this time, the other channel is also here, but it's in pending_force_close_channels (and the commitment-tx has long been confirmed). 

### Your environment

* version of `lnd` 

0.7.0 (from 0.6.1)

Linux localhost 3.4.0-perf-gbccb33a #1 SMP PREEMPT Fri Jul 10 09:23:01 2015 armv7l Android

bitcoind 0.18.0

### Steps to reproduce

probably hard to reproduce. Kill your .lnd with open channels, then try to restore.

### Expected behaviour

All open channels should be closed and the funds be returned to my wallet.

### Actual behaviour

Onchain funds got returned, but it appears that one force-closed channel isn't recognized as returned to us while another is kept open, and makes lnd crash on force-close.

wpaulino:
This is only useful for detecting the case where users have attempted to restore their channels multiple times and one of their earlier attempts was actually successful in sweeping their funds. Without this, outputs that have already been swept by us on-chain would never have their corresponding contract resolvers resolved, causing channels to remain as pending closed, even though its been fully resolved.

The transaction lookup in the wallet is done by checking if any of the outputs are under the wallet's control. Alternatively, we could just look up the transaction itself through its hash, but the current wallet interface does not expose this functionality.

Testing this is a bit involved, but here's how you can confirm the fix:

1. Open a channel from Alice to Bob.
2. Delete all of Alice's data except for their channel backup.
3. Restore Alice's channel backup running master and wait until the channel is fully resolved.
4. Attempt another restore attempt running master and note how the channel never fully resolves due to a remote spend error.
5. Attempt another restore with this PR and wait until the channel fully resolves.

Note: this fix may not be enough if the input is swept to a custom output (not currently supported) that is not under the wallet's control.

Fixes #3321.

### DIFF
--- a/sweep/backend_mock_test.go
+++ b/sweep/backend_mock_test.go
@@ -96,6 +96,10 @@ func (b *mockBackend) WithCoinSelectLock(f func() error) error {
 	return f()
 }
 
+func (b *mockBackend) FetchInputInfo(*wire.OutPoint) (*lnwallet.Utxo, error) {
+	return nil, lnwallet.ErrNotMine
+}
+
 func (b *mockBackend) deleteUnconfirmed(txHash chainhash.Hash) {
 	b.lock.Lock()
 	defer b.lock.Unlock()
--- a/sweep/interface.go
+++ b/sweep/interface.go
@@ -24,4 +24,10 @@ type Wallet interface {
 	// ability to execute a function closure under an exclusive coin
 	// selection lock.
 	WithCoinSelectLock(f func() error) error
+
+	// FetchInputInfo queries for the wallet's knowledge of the passed
+	// outpoint. If the wallet determines this output is under its control,
+	// then the corresponding output should be returned. Otherwise,
+	// lnwallet.ErrNotMine should be returned instead.
+	FetchInputInfo(prevOut *wire.OutPoint) (*lnwallet.Utxo, error)
 }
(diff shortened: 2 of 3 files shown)


joostjager:
Doesn't this generate false positives if someone else publishes a tx that also contains an output paying to us?

PR #4029, opened 2020-02-25
adds `fundmax` flag to `openchannel`

bjarnemagnussen:
Related to PR #4024, but reopened as a new PR as this feature has undergone major changes. 

### Description

This PR introduces a new flag `fundall` for `openchannel`, which when opening a new channel uses the whole available wallet balance for the channel capacity up to the allowed maximum funding limit (currently ~0.16 btc).

### Motivation

See also #619 

The command `sendcoins` has a flag `sweepall` that will spend all the coins from the wallet. However no such flag exists when opening new channels using `openchannel`.

Wanting to use the total available wallet balance in the local amount when opening a channel requires calculating the needed fee and subtracting it from the balance manually.

### Implementation

A new coin select function `CoinSelectUpToAmount` has been introduced, which selects coins up to  the requested amount, or below if there are not sufficient funds available. Fees and dust amounts are added or subtracted from this amount or the change output depending on the available funds, see also the new test cases.

### Considerations

This implementation should be concurrent safe, becuase it locks the coins prior to selecting them, as currently done by the wallet assembler for the coin select subroutine.

#### Pull Request Checklist

  per default)
  already exist


rpcserver.go
```diff
@@ -1564,6 +1564,22 @@ func (r *rpcServer) OpenChannel(in *lnrpc.OpenChannelRequest,
...
+	// If the FundAll flag is set, ensure that the
+	// acceptable minimum local amount adheres to the
+	// amount to be pushed to the remote, and to current
+	// rules.
+	if in.FundAll {
+		// Use as minmimum local amount the larger of
+		// the amount to be pushed to the remote or
+		// the allowed minimum funding size.
+		if remoteInitialBalance >= minChanFundingSize {
+			localFundingAmt = remoteInitialBalance + 1
```
bjarnemagnussen: One requirement for local amount is to be greater than the remote initial balance. Is it therefore enough for the local amount to be 1 Satoshi greater?

lnwallet/chanfunding/coin_select.go
```diff
@@ -214,3 +214,136 @@ func CoinSelectSubtractFees(feeRate chainfee.SatPerKWeight, amt,
 
 	return selectedUtxos, outputAmt, changeAmt, nil
 }
+
+// CoinSelectUpToAmount attempts to select coins such that we'll spend up
```
Roasbeef: Why can't `CraftSweepAllTx` be used for this purpose?

Roasbeef: This also doesn't lock outpoints at all, which can lead to race conditions.

halseth:
The wallet/fundingmanager already has a `subtractFees` flag that can be used for this purpose: https://github.com/lightningnetwork/lnd/blob/3aca9d24b849822b0c345985f2c4d27cad688b22/lnwallet/wallet.go#L69

It should significantly reduce the complexity of this PR I think.

lnwallet/chanfunding/coin_select.go
bjarnemagnussen: As far as I can see, the `CraftSweepAllTx` will spend _all_ inputs. But for opening channels there is a maximum constant `lnd.MaxFundingAmount` that a channel funding is not allowed to exceed. 
It would probably be possible to change `CraftSweepAllTx` to allow to specify a maximum amount, but would logically not make sense in a sweep package?

In addition, opening channels requires sending many messages. E.g. before attempting to open a channel LND will check if the remote node is responding, using dust limits depending on the chain used, etc. From the workflow of opening a channel it wasn't obvious to me how to get the funding output to use as "delivery" address, without specifying the funding amount in the first place. And I also didn't want to change this workflow when `fundall` is used.

Regarding locking of outpoints, as far as I see this should already happen in the `wallet_assembler.go` like for the other coin select subroutines:
https://github.com/lightningnetwork/lnd/blob/38b521d87d3fd9cff628e5dc09b764aeabaf011a/lnwallet/chanfunding/wallet_assembler.go#L299-L306

bjarnemagnussen:
> The wallet/fundingmanager already has a `subtractFees` flag that can be used for this purpose:
> 
> https://github.com/lightningnetwork/lnd/blob/3aca9d24b849822b0c345985f2c4d27cad688b22/lnwallet/wallet.go#L69
> 
> It should significantly reduce the complexity of this PR I think.

I was trying to use this flag, but since the maximum funding amount is limited to the constant `lnd.MaxFundingAmount`, we don't always want to use it. It is only if the balance is less than `lnd.MaxFundingAmount`, and in special cases where the balance is just slightly greater than `lnd.MaxFundingAmount` but not enough to cover the fees. Then we actually will subtract the fee from the total balance which will give a funding amount a little less than this constant.

But I did use the `CoinSelectSubtractFee` function for inspiration a lot, and it may be that I should try to combine the new `CoinSelectUpToAmount` with it?

I saw there may also be an open TODO in that regard for the excess fee. Furthermore, `CoinSelect` does not make correct fee calculation if change amount is dust, which is solved by `CoinSelectSubtractFee` and `CoinSelectUpToAmount`.

halseth:
@bjarnemagnussen I'm not sure I understand the limitation of the current API. If you do a regular channel opening with the `MaxFundingAmt` value, it will open that channel if the balance is available for that channel size + the fee. If the balance is not available, one can use the `subtractFees` flag, wit a value set to the remaining wallet balance. Wouldn't this work?

bjarnemagnussen:
@halseth This works well, and I have added a commit to do as you suggest! :) Hopefully this is something along the line of what you were thinking? 

### Mismatch between coin select functions

There is this little mismatch between `CoinSelect` and `CoinSelectSubtractFees` that probably should be considered.

The `CoinSelect` function always uses a fixed fee calculation for one input and one change output. Let's assume the balance is one satoshi greater than the sum of `MaxFundingAmt` and a fee for one input and **no** change output. Then the `CoinSelect` function will return an error for insufficient funds, and the `CoinSelectSubtractFees` function takes over.

Inside `CoinSelectSubtractFees` we use the whole balance but the smaller fee calculation for one input and no change output. The final output amount returned from `CoinSelectSubtractFees` will therefore be the smaller fee subtracted from the total balance, which will result in an amount one satoshi greater than the `MaxFundingAmt`.

An easy fix is to use a sanity check that simply makes sure that we do not return an amount greater than what we asked for. Any difference should be small (fee difference between one and no change output) and can directly go towards the fee.

### Test cases

I have for now added a test case "WIP: mismatch of fee calculation between functions" that will pass due to the added sanity check explained above, but otherwise would give rise to an output amount greater than the specified maximum.

There is also an oddity regarding testing for dust change, as it depends on which branch inside `CoinSelectUpToAmount` is used. Since `CoinSelectUpToAmount` now makes use of `CoinSelect` and `CoinSelectSubtractFees` it depends on their different ways of handling dust change. I have removed testing for dust change as those should be handled by the other coin select functions. I have therefore also removed the test case "no amount to pay fee for change".

cmd/lncli/commands.go
```diff
@@ -613,6 +613,12 @@ var openChannelCommand = cli.Command{
 			Name:  "connect",
 			Usage: "(optional) the host:port of the target node",
 		},
+		cli.BoolFlag{
+			Name: "fundall",
+			Usage: "(optional) if set, then the local_amt field will " +
+				"be ignored, and the wallet will attempt to commit the " +
```
halseth: should state that should not be set together with `localAmt`

rpcserver.go
halseth: Where does this requirement come from?

fundingmanager.go
```diff
@@ -2888,6 +2888,14 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		}
 	}
 
+	// If the caller specified to spend all funds, then we
+	// define the maximum allowed funding amount to be the
+	// current limit of the maximum funding amount.
+	var fundUpToMaxAmt btcutil.Amount
+	if msg.fundAll {
+		fundUpToMaxAmt = MaxFundingAmount
```
halseth: should we return an error if the user attempts `fundAll` but have more funds than `MaxFundingAmount`?

bjarnemagnussen:
I am still not sure how to handle both the `MaxFundingAmount` and the `minChanFundingSize` better with the `fundall` flag. Any other idea on how to streamline this?

guggero:
Could you rebase the PR please? I'm going to take over the review for @halseth.

> I am still not sure how to handle both the MaxFundingAmount and the minChanFundingSize better with the fundall flag. Any other idea on how to streamline this?

Didn't read through all comments yet, but couldn't we just check the `walletbalance` from the command line and if it's higher than `MaxFundingAmount` we just say "wallet balance exceeds minimum channel size, --fundall flag cannot be used, please specify exact amount" or something?

bjarnemagnussen:
I have rebased to the current v0.10.0-beta.rc5 master commit.

Just to make sure we expect the same behaviour I am listing some cases how the `fundall` flag should open channels:

* In general if `balance < minChanFundingSize`: Fail just as is current behaviour for `openchannel`.
* If `balance < MaxFundingAmount + open-tx-fee`: Open a channel by maxing out the capacity resulting in no change, equivalent to opening a channel using the whole balance with `SubtractFees` set. 
* If `balance >= MaxFundingAmount + open-tx-fee`: Open a channel using `MaxFundingAmount` with a change amount to the wallet.

**Remark**: In the current implementation an edge case can raise a bug resulting in a channel capacity _greater_ than `MaxFundingAmount`. This is caught with a sanity check, see https://github.com/lightningnetwork/lnd/pull/4029#issuecomment-592955663 above.

rpcserver.go
guggero: In practice, the remote amount cannot go over a certain percent because of the channel reserve. The same probably also applies to the push amount though I haven't tried this.
For this PR I think it either makes sense to disallow setting a push amount (`remoteInitialBalance`) in the command line when using `--fundall/--fundmax` or try to actually calculate the channel reserve in percent and then scale the local amount accordingly.

lnwallet/wallet.go
```diff
@@ -99,6 +99,10 @@ type InitFundingReserveMsg struct {
 	// to this channel.
 	RemoteFundingAmt btcutil.Amount
 
+	// FundUpToMaxAmt defines if channel funding should try to add as many
+	// funds to LocalAmt as possible until FundMaxAmt is reached.
+	FundUpToMaxAmt btcutil.Amount
```
guggero, line 104: Because of the check in the `wallet_assembler`, the `LocalAmt` becomes the minimum amount of coins that need to be available if this is set. Maybe this should be mentioned here.

bjarnemagnussen: I have tried to clarify this with the new commit.

cmd/lncli/cmd_open_channel.go
```diff
@@ -81,6 +81,12 @@ var openChannelCommand = cli.Command{
 			Name:  "connect",
 			Usage: "(optional) the host:port of the target node",
 		},
+		cli.BoolFlag{
+			Name: "fundall",
```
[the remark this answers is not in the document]
bjarnemagnussen: Changed to `--fundmax` with the newest commit! :)

rpcserver.go
bjarnemagnussen: I have tried to figure this out but am not completely sure. 

The way I see it, here in the code (https://github.com/lightningnetwork/lnd/blob/master/rpcserver.go#L1633-L1645) it "lazily" checks only if the push amount is greater or equal the local amount.

The logic to actually make sure the funds are sufficient happens later in the workflow here (https://github.com/lightningnetwork/lnd/blob/0dfb946afd36175dfc0e37acf45150784c87e375/lnwallet/reservation.go#L218-L260).

Therefore I was thinking to simply set the "minimum amount" (local_amt) to the greater of the existing channel opening minimum (`minChanFundingSize`) and the push amount, and let the existing logic later on make sure we actually have the funds to pay for fees, etc?

bjarnemagnussen:
I commented above _before_ committing, so sorry for that! 

I have renamed `fundall` to `fundmax` and added comments in the code to explain a bit more what is going on.

guggero:
No worries. I'll have a look at it today and test it more thoroughly.

bjarnemagnussen:
Let me just one more time mention https://github.com/lightningnetwork/lnd/pull/4029#issuecomment-592955663 for when you find the time to test this more thoroughly.

I am not particularly happy about this edge case and how it is currently resolved. I think this should be solved by aligning the fee calculation of the two coin-selection functions (`CoinSelect` and `CoinSelectSubtractFees`). But I am not sure if there are concrete plans to do this in the near future, since TODOs already exist in the code of the `CoinSelect` function.

If needed, I could take a look at the TODO inside `CoinSelect` to _make coinSelect not estimate change output for dust change_, which is already correctly handled by ´CoinSelectSubtractFee`.

guggero:
I tested this a bit on regtest today. Ran the following test cases:
1. Wallet amount is dust (2000sat): `output amount(-0.00004087 BTC) after subtracting fees(0.00006087 BTC) below dust limit(0.00000573 BTC)` :heavy_check_mark: 
2. Wallet amount < min chan size (18000sat): `available funds(0.00017739 BTC) below the minimum amount(0.0002 BTC)` :heavy_check_mark: 
3. Wallet amount > min chan size (20400sat): new channel over 20k sats :heavy_check_mark: 
4. Wallet amount > max chan size (20000000sat): new max size channel :heavy_check_mark: 
5. Wallet amount > max chan size, push amount == amount: `Funder balance too small (-9049000) with fee=9050 sat, minimum=1146 sat required` :heavy_check_mark: 
6. Wallet amount > max chan size, push amount 16766000: good (local balance 2165, remote balance 16766000, fee 9050, reserve: 167772) :heavy_check_mark: 
7. Test the above without --fundmax flag: good, same result :heavy_check_mark: 

So far so good. I'll have a look at the fee calculation difference between `CoinSelectSubtractFee` and `CoinSelect` you mentioned.

rpcserver.go
guggero: Yes, you are right, there is a check later on. Found that out while testing :)
So I think that `remoteInitialBalance+1` is ok here.

lnwallet/chanfunding/coin_select_test.go
```diff
@@ -381,3 +381,260 @@ func TestCoinSelectSubtractFees(t *testing.T) {
...
+			coins: []Coin{
+				{
+					TxOut: wire.TxOut{
+						PkScript: p2wkhScript,
+						Value:    int64(fundingFee(feeRate, 1, false) + dust),
+					},
+				},
+			},
+			spendValue: fundingFee(feeRate, 1, false) + dust,
+
```
guggero, line 511: nit: extra newline

bjarnemagnussen:
I am looking into your suggested changes but have a bit of noob question regarding where to add the tests. 

Most of them can indirectly be tested with the coin select function, such as the coin select function not picking up dust amounts; behaviour of selecting coins close to some max amount; etc.

But I cannot find a dedicated place where "channel opening"/`wallet_assembler.go` is tested. Since [the logic](https://github.com/lightningnetwork/lnd/pull/4029/files#diff-96e38b3dbb364a2303ef75daa1183cbfR268-R271) to make sure the selected amount is not below `LocalAmount` is inside the `wallet_assembler.go` file, I am not sure where to add the test for it without an existing `wallet_assembler_test.go` file. 

This yields another question: Would it make sense to add a parameter `minAmt` to `CoinSelectUpToAmount` and instead have the logic to not return selected coins below this amount inside this function and test it there?

guggero:
I think the unit test you added is sufficient. It does test the added coin selection code.

What I mean is to add a new itest (see `lntest/itest`). There you can start a new node and send the amount you want to test with to it, then open channels.
It's a bit of work to set everything up. But there isn't an easier way to test the interplay between the funding manager and the wallet currently.

bjarnemagnussen:
I have taken a look at the suggested changes and hopefully addressed them with the new commits. 

Regarding the integration tests: I am not sure if they are fine like that. I think it could have been made more compact with a single testing function, but I wasn't sure how to simply "reset" the nodes "carol" and "dave", so that they have no UTXOs in the wallet. Therefore I "killed" those nodes and made new ones with each new testing function.

//edit: Rebased to current master since the rpc-proto seems to have been adjusted a bit.

lntest/itest/lnd_test.go
```diff
@@ -2895,6 +2921,465 @@ func testChannelBalance(net *lntest.NetworkHarness, t *harnessTest) {
 	closeChannelAndAssert(ctxt, t, net, net.Alice, chanPoint, false)
 }
 
+// testChannelFundMaxLessThanDust tries to create a new channel using fundmax
+// flag with a wallet balance less than dust. Should fail.
+func testChannelFundMaxLessThanDust(net *lntest.NetworkHarness, t *harnessTest) {
```
guggero, line 2926: Yay, nice to have these integration tests!

However, there are a few things to optimize here:
- It would be nice if we had one test for this that runs several sub test cases in a table driven manner. An example for this is `testChannelBackupRestore`.
- The two nodes `carol` and `dave` could be created once and reused for each (sub) test. To make sure they start with a zero balance before each test, you can use the `SendCoins` RPC with the `SendAll` flag to send all remaining balance to some address of the miner for example (see `testSweepAllCoins`). That should speed up things considerably as the nodes can keep running between the tests (the previously opened channels can remain open I think).
- We started moving new tests to their own files to make keep `lnd_test.go` from growing even further. Can you move the test to `lnd_channel-funding_fund_max.go` or something like that?
- Move everything related to the itest into a separate commit.

bjarnemagnussen:
Thanks for your inputs @guggero! It took me a few weekends but I feel that the integration tests are now much better structured and clear. 

### Two "special" test cases

I left two "special" test cases (`wallet amount > max chan size, push amount == max-chan-size` and `wallet amount > max chan size, push amount == max-chan-size - 1`), which also contain
comments marked with `//!` that will be deleted after discussing them. I think they are important tests but are actually not directly related to the `fundmax` flag. They also trigger an error with implementation details in mind, and therefore I am not sure if they should be kept or not.

### Regarding the failed CI pipeline

* I am not sure why the linter reports failure. I can see that inside the `fundChannel` function of `fundmanager_test.go` the parameter `numConfs` is unsued, but I didn't touch that and only added one more parameter `fundMax`, which is used. 
* The neutrino integration fails with an error that I think should be unrelated to my added tests?

Let me know if you can see how to improve it, or if I missed something!

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		ht.Fatalf("unable to connect peers: %v", err)
+	}
+
+	// Helper function that returns the fee estimate used for a tx
+	// with the given number of inputs and the optional change output.
+	// This matches the estimate done by the wallet.
+	fundingFee := func(numInput int, change bool) btcutil.Amount {
+		var (
+			feeRate        = chainfee.SatPerKWeight(12500)
```
guggero, line 70: `feeRate` can be defined as a package level variable, it's never changed here. Then the var statement can become one line.

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -0,0 +1,278 @@
...
+			pushAmt:      16766000,
+			carolBalance: lnd.MaxBtcFundingAmount - 16766000,
+		},
+	}
+
+	for _, testCase := range testCases {
+		success := ht.t.Run(testCase.name, func(t *testing.T) {
+
+			sendCoinsAndAssert(testCase.amount, carol)
+			defer func() {
```
guggero, line 209: could make this a helper function as well that accepts the node and test case as parameter.

lnwallet/chanfunding/coin_select_test.go
```diff
@@ -381,3 +381,216 @@ func TestCoinSelectSubtractFees(t *testing.T) {
...
+	testCases := []testCase{
+		{
+			// We have 1.0 BTC available, spend them all. This
+			// should lead to a funding TX with one output, the
+			// rest goes to fees.
+			name: "spend exactly all",
+			coins: []Coin{{
+				TxOut: wire.TxOut{
+					PkScript: p2wkhScript,
+					Value:    1 * btcutil.SatoshiPerBitcoin,
```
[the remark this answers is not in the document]
bjarnemagnussen: Great! I have done that with the new commits. However, should this also be done for the other tests in the same file? I think that would be better and could therefore add that if you agree?

fundingmanager_test.go
```diff
@@ -3019,6 +3020,91 @@ func TestFundingManagerFundAll(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		alice, bob := setupFundingManagers(t)
+		defer tearDownFundingManagers(t, alice, bob)
+
+		// We set up our mock wallet to control a list of UTXOs that sum to
+		// more than the max channel size.
+		alice.fundingMgr.cfg.Wallet.WalletController.(*mockWalletController).utxos = test.coins
```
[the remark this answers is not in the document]
bjarnemagnussen: The new commits does that now. The only difference is that inside `setupFundingManagers()` both `alice` and `bob` will then get funds, whereas this would only apply to `alice`. In the existing `TestFundingManagerFundAll` inside this file the same technique is used. If you agree I would also change it there?

bjarnemagnussen:
Thanks @guggero! I have looked into the issues you raised and hope that they are fixed now. See also my two comments regarding your suggestions to the rest of the code.

I think that the code now follows the style guidelines better, but if you should find anything more just let me know.

lnwallet/chanfunding/coin_select_test.go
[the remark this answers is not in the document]
guggero: Sure, feel free to change those too. It looks like the existing tests also don't fully adhere to the style guide and replacing those long variable names might fix that as well.

fundingmanager_test.go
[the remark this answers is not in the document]
guggero: Ah, I see. Didn't look at the other tests in this file at all :sweat_smile: It should be fine to give both nodes some UTXOs, as long as it isn't a dual-funding test only Alice should use them anyway.
Feel free to use the following form (shouldn't use additional memory, the compiler optimizes things like this):
```
		addFunds := func(fundingCfg *fundingConfig) {
			wc := fundingCfg.Wallet.WalletController
			wc.(*mockWalletController).utxos = test.coins
		}
```

fundingmanager.go
```diff
@@ -3140,6 +3140,14 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
 		return
 	}
 
+	// If the caller specified to spend all funds, then we
+	// define the maximum allowed funding amount to be the
+	// current limit of the maximum funding amount.
+	var fundUpToMaxAmt btcutil.Amount
+	if msg.fundMax {
+		fundUpToMaxAmt = MaxFundingAmount
```
bjarnemagnussen, line 3148: I just saw that PR #4429 directly affects the maximum funding amount used here. With "wumbo" channels being limitless, one way to handle it could be to just set `fundUpToMaxAmt` to the maximum integer value (`math.MaxInt64`) when "wumbo" feature is enabled. 

To check if "wumbo" is enabled, at this point I think it is necessary to check if "wumbo" is active in both the local and remote feature set (using `msg.peer.LocalFeatures()` and `msg.peer.RemoteFeatures()`).

Roasbeef: `MaxFundingAmount` is unchanged in that PR.

bjarnemagnussen: But if both the local and remote nodes support wumbo channels, the flag `fundmax` should fund a channel using _all_ funds, not only up to the existing `MaxFundingAmount`?

Roasbeef: Yeah, so you'd need to modify `fundUpToMaxAmt` based on if wumbo was set or not as you mentioned.

halseth:
Needs rebase!

fundingmanager.go
```diff
@@ -3148,6 +3149,20 @@ func (f *fundingManager) handleInitFundingMsg(msg *initFundingMsg) {
...
+	// If the caller specified to spend all funds, then we define the
+	// maximum allowed funding amount depending on if both local and remote
+	// peer supports wumbo channels.
+	var fundUpToMaxAmt btcutil.Amount
+	if msg.fundMax {
+		fundUpToMaxAmt = MaxFundingAmount
+		remoteWumbo := msg.peer.RemoteFeatures().HasFeature(
+			lnwire.WumboChannelsOptional,
+		)
+		if !f.cfg.NoWumboChans && remoteWumbo {
```
bjarnemagnussen, line 3161: As far as I understand it the fundingmanager has the field `NoWumboChannels` inactive if Wumbo is enabled with `--protocol.wumbo-channels`:
https://github.com/lightningnetwork/lnd/blob/77549f1fb8f046fc565d42151a4ae3900391170e/server.go#L970

So there is no need to check what the local node is advertising as enabled features and it suffices to check if Wumbo is enabled in the fundingmanager and advertised as features with the remote node?

fundingmanager_test.go
```diff
@@ -3090,6 +3091,117 @@ func TestFundingManagerFundAll(t *testing.T) {
...
+
+	for _, test := range tests {
+		// We set up our mock wallet to control a list of UTXOs that
+		// sum to more than the max channel size.
+		test := test
+		addFunds := func(fundingCfg *fundingConfig) {
+			wc := fundingCfg.Wallet.WalletController
+			wc.(*mockWalletController).utxos = test.coins
+		}
+		alice, bob := setupFundingManagers(t, addFunds)
```
bjarnemagnussen, line 3156: By default TestFundingManager is created with `NoWumboChannels` field inactive, which has the effect that Wumbo is enabled but not advertised as an active feature in the TestNode. Test cases that don't have `wumbo` specified are hence indirectly being tested with Wumbo enabled on the local node but not _advertised_ by the remote node.

Alternatively, `NoWumboChannels` could be actived for test cases that don't use Wumbo. But then it would probably be better to add yet another test case that explicitly ensures that the edge case mentioned above is tested for.

bjarnemagnussen:
Rebased and added a commit to adjust the maximum funding amount based on if Wumbo is enabled both locally and advertised by the remote node.

lnwallet/chanfunding/coin_select.go
```diff
@@ -214,3 +214,57 @@ func CoinSelectSubtractFees(feeRate chainfee.SatPerKWeight, amt,
...
+	switch d := outputAmt - maxAmt; {
+	case d > dustLimit:
+		return nil, 0, 0, fmt.Errorf(
+			"non-negative difference in requested maximum funding "+
+				"amount %v and obtained amount %v", maxAmt,
+			outputAmt,
+		)
+
+	case d > 0:
+		outputAmt = maxAmt
```
halseth, line 266: shouldn't this be an error?

halseth: Since in the case `outputAmt != maxAmt` we already determined above that we don't have funds to fund `maxAmt`.

lnwallet/chanfunding/coin_select.go
```diff
@@ -214,3 +214,57 @@ func CoinSelectSubtractFees(feeRate chainfee.SatPerKWeight, amt,
 
 	return selectedUtxos, outputAmt, changeAmt, nil
 }
+
+// CoinSelectUpToAmount attempts to select coins such that we'll spend up
+// to amt in total before fees, or all of the coins if not sufficient amount
+// available. The specified fee rate should be expressed in sat/kw for coin
+// selection to function properly.
+func CoinSelectUpToAmount(feeRate chainfee.SatPerKWeight, maxAmt, minAmt,
```
halseth, line 222: minAmt not used

lnrpc/rpc.proto
```diff
@@ -24,868 +24,863 @@ option go_package = "github.com/lightningnetwork/lnd/lnrpc";
...
-    rpc SubscribeChannelBackups (ChannelBackupSubscription)
-        returns (stream ChanBackupSnapshot);
-
-    /* lncli: `bakemacaroon`
-    BakeMacaroon allows the creation of a new macaroon with custom read and
-    write permissions. No first-party caveats are added since this can be done
-    offline.
-    */
-    rpc BakeMacaroon (BakeMacaroonRequest) returns (BakeMacaroonResponse);
+  /* lncli: `walletbalance`
```
halseth, line 27: something went wrong on rebase here? Diff is huge

rpcserver.go
```diff
@@ -1709,6 +1709,22 @@ func (r *rpcServer) parseOpenChannelReq(in *lnrpc.OpenChannelRequest,
 
 	localFundingAmt := btcutil.Amount(in.LocalFundingAmount)
 	remoteInitialBalance := btcutil.Amount(in.PushSat)
+
+	// If the FundMax flag is set, ensure that the
+	// acceptable minimum local amount adheres to the
+	// amount to be pushed to the remote, and to current
+	// rules.
+	if in.FundMax {
+		// Use as minmimum local amount the larger of
```
halseth, line 1718: typo. minminum

cmd/lncli/cmd_open_channel.go
```diff
@@ -197,6 +203,7 @@ func openChannel(ctx *cli.Context) error {
 		SpendUnconfirmed:           minConfs == 0,
 		CloseAddress:               ctx.String("close_address"),
 		RemoteMaxValueInFlightMsat: ctx.Uint64("remote_max_value_in_flight_msat"),
+		FundMax:          ctx.Bool("fundmax"),
```
halseth, line 206: File not formatted

cmd/lncli/cmd_open_channel.go
```diff
@@ -250,10 +257,15 @@ func openChannel(ctx *cli.Context) error {
 			return fmt.Errorf("unable to decode local amt: %v", err)
 		}
 		args = args.Tail()
-	default:
+	case !ctx.Bool("fundmax"):
```
halseth, line 260: I don't think we need this check here

fundingmanager.go
halseth: Could also leave this max limit cap for the fundall flag for now, to avoid user by accident making channels using all their wumbo funds.

fundingmanager_test.go
halseth: As mentioned above, I would be in favor of not allowing wumbo channels to be created using the `fundmax` option (at least for the time being).

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -0,0 +1,307 @@
...
+			err = checkChannelBalance(dave, testCase.pushAmt)
+			if err != nil {
+				t.Errorf("dave's balance is wrong: %v", err)
+			}
+		}
+	}
+
+	var testCases = []*chanFundMaxTestCase{
+		{
+			name:               "wallet amount is dust",
```
halseth, line 242: tbh this looks more like checks that should be done on the unit test level...

lnwallet/chanfunding/coin_select_test.go
```diff
@@ -73,18 +74,18 @@ func TestCoinSelect(t *testing.T) {
 				{
 					TxOut: wire.TxOut{
 						PkScript: p2wkhScript,
-						Value:    1 * btcutil.SatoshiPerBitcoin,
+						Value:    1 * coin,
```
halseth, line 77: changes like this should be squashed with the original commit 😄

lnwallet/chanfunding/coin_select.go
bjarnemagnussen: In case `outputAmt != maxAmt` we would expect the routine to select coins below `maxAmt`. But due to the edge case (see https://github.com/lightningnetwork/lnd/pull/4029#issuecomment-592955663 and the corresponding test case https://github.com/lightningnetwork/lnd/pull/4029/files#diff-a19edac697b6152ce1c2fec4f6ac1d80R511), it is possible that coins are selected slightly above `maxAmt`. In case the difference is below the dust limit we set `outputAmt` to `maxAmt` and implicitly let the small difference go towards the fee, otherwise `outputAmt` is left unchanged.

lnwallet/chanfunding/coin_select.go
bjarnemagnussen: Fixed

lnrpc/rpc.proto
bjarnemagnussen: Oh yes! My editor auto-formatted this file and changed the indentation. I have restored it.

rpcserver.go
bjarnemagnussen: Fixed

cmd/lncli/cmd_open_channel.go
bjarnemagnussen: Fixed

cmd/lncli/cmd_open_channel.go
bjarnemagnussen: Ah yes, fixed

lntest/itest/lnd_channel-funding_fund_max_test.go
bjarnemagnussen: You are thinking something like a test `TestFundingManagerFailDust` inside `fundingmanager_test.go`?

lnwallet/chanfunding/coin_select_test.go
bjarnemagnussen: I have made the cosmetic change to squash this with its corresponding commit ;)

fundingmanager_test.go
bjarnemagnussen: Mh, personally I would prefer this flag to fund up the maximum possible (similar to how the `sendall` flag works), however up to the soft-limit if it is _active_. I feel that letting this flag suddenly respect a soft-limit that is proactively deactivated by turning on Wumbo is a bit strange. However if the general feeling is that for security reasons this makes sense I will revert the commit to always respect this limit.

bjarnemagnussen:
I have pushed the changes! See comments for a few points that are still unclear.

lntest/itest/lnd_channel-funding_fund_max_test.go
halseth: I'm saying that all these tests don't really belong on the integration test level, since they test very specific behaviour of a particular unit (the coin selection).

lnwallet/chanfunding/coin_select.go
halseth: Aha, I see! Thanks for the explanation :)

Your fix does indeed seem to handle this edge case, but it feels a bit hacky. Could we instead try to fix the change output calculation in `CoinSelect`? (I see I already added a TODO there for this case).

halseth: (preferably as its own independent PR with corresponding test cases)

bjarnemagnussen: Yes, I think that's a good idea. I will look into that!

lntest/itest/lnd_channel-funding_fund_max_test.go
bjarnemagnussen: The test cases that are more specific for the `fundmax` flag are indeed already tested for inside `coin_select_test.go` (dust values, min/max channel sizes...).

However, as far as I can tell there does not exist integration tests for the `openchannel` command. Do you think it would make sense to later add a PR with only the integration tests for the `openchannel` command?

lnwallet/chanfunding/coin_select.go
bjarnemagnussen: (I am still looking into this. I just didn't find the time prior to now.)

guggero (commented):
Refreshed my memory on this by doing another round of review since this is scheduled as a potential candidate for 0.13.
There are still two remaining open discussion items:
 - Fix change handling in `CoinSelect` instead of adding a new function
 - How to handle a situation where there are more than `MaxFundingAmount` of sats in the wallet

Also, needs a rebase.
Do you have time to pick this up in the next few weeks so we can get it in, @bjarnemagnussen?

bjarnemagnussen:
@guggero I will take a look in the next weeks!

bjarnemagnussen:
I will wait for PR #4588 to be merged, which allows some minor simplification of this PR as then the fee calculation is streamlined. We then don't need to make the sanity check for edge cases.

guggero:
@bjarnemagnussen #4588 has been merged, this can now be rebased.

lnwallet/chanfunding/coin_select.go
```diff
@@ -214,3 +214,57 @@ func CoinSelectSubtractFees(feeRate chainfee.SatPerKWeight, amt,
...
+	// Get balance from coins.
+	var balance btcutil.Amount
+	for _, coin := range coins {
+		balance += btcutil.Amount(coin.Value)
+	}
+
+	// First we try to select coins up to the specified maxAmt.
+	outputAmt := maxAmt
+	selectedCoins, changeAmt, err := CoinSelect(
+		feeRate, maxAmt, coins,
```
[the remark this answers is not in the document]
bjarnemagnussen: Since `CoinSelect` now takes four arguments by checking for the dust limit, it does not fit into one line anymore.

lnwallet/chanfunding/coin_select.go
```diff
@@ -214,3 +214,57 @@ func CoinSelectSubtractFees(feeRate chainfee.SatPerKWeight, amt,
 
 	return selectedUtxos, outputAmt, changeAmt, nil
 }
+
+// CoinSelectUpToAmount attempts to select coins such that we'll spend up
+// to amt in total before fees, or all of the coins if not sufficient amount
+// available. The specified fee rate should be expressed in sat/kw for coin
```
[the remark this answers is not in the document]
bjarnemagnussen: Removed!

bjarnemagnussen:
This is still a WIP as I will add new logic in the next days.

I have rebased the code and it was a bit more involved as a lot of refactoring has happened, including deleted files, etc. 

I have removed logic for wumbo for now, since I need to clarify a few things. It seems that a new flag `maxchansize` was introduced, which overrides the default values. I was looking through the code, but it seems there is no way to communicate between the nodes their own maxChanSize parameter, which they will accept for incoming channels?

The problem then is (with and without Wumbo), that currently `fundmax` uses the default max-value (0.16... BTC), but should respect the funding managers maxChanSize, which is easy to fix. But even then it is not known what value the remote node would accept.

A solution may be something along the lines of allowing an optional value with the `fundmax` flag, which sets the maximum value to try to open a channel node with and can be set to the remote's maxChanSize if it is known, and otherwise if the value is set to default 0 will use the local node's maxChanSize.

I will incorporate this with additional commits in the next days.

bjarnemagnussen:
I have rebased and added a new commit 810d331 with the following changes. 

When using the `fundmax` flag it will now use the options set with `maxchansize` and `minchansize`. This way if wumbo is enabled but the maximum channel size has been set lower, the `fundmax` flag will respect this limit even though this setting is meant for incoming channels.

I think this is a good balance of staying safe in regard to not unexpectedly spend all wumbo funds with a channel open, while keeping this flag useful. However, I would also consider e.g. letting fundmax only spend up to the "legacy" maximum channel balance (or should it be the configured maxchansize if it is lower?) and adding a fundmaxwumbo to spend up to the new wumbo maximum channel size.

The rest of the commits are as before, however with changes due to the rebase.

guggero:
@bjarnemagnussen sorry this didn't make it into 0.14.0. Are you willing to give this another attempt for 0.15? I think we're really close and just need to find another reviewer (and currently a rebase plus an addition to the not-yet-existing 0.15 release notes).

bjarnemagnussen:
Hey @guggero, this PR feels like my little "lifework" in LND for so long that it has been running under review already :) So yes, I am definitely willing to give this another attempt!

As a first step I will look into getting it working again on current master.

guggero:
@bjarnemagnussen cool, glad to hear you're willing to see this one through :sweat_smile: can you please re-request review once you were able to rebase? Thanks!

guggero:
!lightninglabs-deploy mute 2022-Feb-01

bjarnemagnussen:
I have rebased this PR and added two commits to respect the reserve wallet balance in case of anchor channels.

One thing to consider is regarding the different commit types. Should the integration test also check the "lease" commit type, and should there be an integration test that fails whenever a new type is introduced that is not tested for to be sure the behaviour is always correct?

saubyk:
@guggero waiting eagerly for this feature.

guggero:
Okay cool. Feel free to test and leave a review to make sure this works as expected for your use case(s). Will try to pick it up this week too.

lnwallet/chanfunding/coin_select.go
guggero: I stumbled over this again. Somehow this still makes me nervous. But reading this comment again and seeing the test case I see why this is needed. Perhaps just call the variable `overshoot` instead of `d` to make it a bit more clear?

funding/manager_test.go
```diff
@@ -3280,6 +3281,94 @@ func TestFundingManagerFundAll(t *testing.T) {
...
+				len(fundingTx.TxOut))
+		}
+
+		if !test.change && len(fundingTx.TxOut) != 1 {
+			t.Fatalf("expected 1 output, had %v",
+				len(fundingTx.TxOut))
+		}
+
+		// Inputs should be all funds in the wallet.
+		if len(fundingTx.TxIn) != len(test.coins) {
```
guggero, line 3376: Should we add a test case with multiple inputs so this check actually does make sense?

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -0,0 +1,277 @@
...
+		mineBlocks(ht, net, 1, 1)
+
+		// Make sure all funds were swept.
+		checkChannelBalance(node, 0)
+	}
+
+	runTestCase := func(carol, dave *lntest.HarnessNode,
+		testCase *chanFundMaxTestCase) func(t *testing.T) {
+
+		return func(t *testing.T) {
```
guggero, line 167: To avoid this nesting and the code within reaching over the 80 character limit, we shouldn't return a function here, but instead just define a function that takes `t, carol, dave, testCase` and then _call_ it in the test case loop below.

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -0,0 +1,277 @@
...
+			feeRate:            1,
+			chanOpenShouldFail: true,
+		},
+
+		{
+			name:   "wallet amount > min chan size (37000sat)",
+			amount: 37000,
+			// The transaction fee to open the channel must be subtracted from
+			// Carol's balance (since wallet balance < max-chan-size).
+			carolBalance: btcutil.Amount(37000) - fundingFee(1, false),
```
guggero, line 271: We should add a test with a low balance and the anchor reserve, just to make sure the interplay here is correct.

cmd/lncli/cmd_open_channel.go
```diff
@@ -103,8 +103,9 @@ var openChannelCommand = cli.Command{
 		cli.BoolFlag{
 			Name: "fundmax",
 			Usage: "(optional) if set, the wallet will attempt to " +
-				"commit the maximum possible local amount to the channel. " +
-				"Should not be set together with local_amt",
+				"commit the maximum possible local amount to " +
```
guggero, line 106: This whole commit should be squashed with the respective commits where the initial changes were introduced.

rpcserver.go
```diff
@@ -1878,19 +1878,22 @@ func (r *rpcServer) parseOpenChannelReq(in *lnrpc.OpenChannelRequest,
...
+	// channel size.
+	var fundUpToMaxAmt btcutil.Amount
 	if in.FundMax {
-		// Use as minimum local amount the larger of
-		// the amount to be pushed to the remote or
-		// the allowed minimum funding size.
-		if remoteInitialBalance >= funding.MinChanFundingSize {
+		// As default value for the minimum local amount use the larger
+		// of the amount to be pushed to the remote or the allowed
+		// minimum channel size.
```
guggero, line 1889: To be honest, these comments don't really help me to understand _why_ something is done. They just describe the code below in words. We should add the context from the discussions in this PR instead. Otherwise someone looking at this in a year or so won't know what was discussed.

lnwallet/chanfunding/coin_select.go
```diff
@@ -267,14 +277,30 @@ func CoinSelectUpToAmount(feeRate chainfee.SatPerKWeight, maxAmt,
 	selectedCoins, changeAmt, err := CoinSelect(
 		feeRate, maxAmt, dustLimit, coins,
 	)
+	if err == nil {
```
guggero, line 280: This whole method is very hard to grasp without staring at it for a long time. Since this selects coins, it makes me very nervous.

So here two suggestions to streamline the code in this area.

1. Use one `switch` statement:

```
switch {
case err == nil:

case err.(type) == *ErrInsufficientFunds:

default:

}
```

2. Instead of setting the err to `errReservedValueInvalidated`, do the actual call to `CoinSelectSubtractFees` directly.

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -204,6 +237,13 @@ func testChannelFundMax(net *lntest.NetworkHarness, ht *harnessTest) {
 			if err != nil {
 				t.Errorf("dave's balance is wrong: %v", err)
 			}
+
+			if commitTypeHasAnchors(testCase.commitmentType) {
+				err = checkWalletBalance(carol, lnwallet.ReservedValue(1))
```
guggero, line 242: :+1:

lnwallet/wallet.go
```diff
@@ -739,6 +739,7 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
 
 	localFundingAmt := req.LocalFundingAmt
 	remoteFundingAmt := req.RemoteFundingAmt
+	hasAnchors := req.CommitType.HasAnchors()
```
guggero, line 742: These changes to the wallet probably deserve their own commit.

guggero (commented):
We're getting pretty close, I think. The added complexity with the reserve balance does make everything a bit harder. I hope you understand why this PR undergoes so much scrutiny since it touches the core funding logic, where (as you've probably noticed) a lot of things can have an influence.

The integration tests look really good now, though! I don't think we need to also add a test with the new commitment type, since fee wise they look just like anchor channels.

cmd/lncli/cmd_open_channel.go
```diff
@@ -302,10 +310,19 @@ func openChannel(ctx *cli.Context) error {
 			return fmt.Errorf("unable to decode local amt: %v", err)
 		}
 		args = args.Tail()
-	default:
+	case !ctx.Bool("fundmax"):
 		return fmt.Errorf("local amt argument missing")
 	}
 
+	// Note:
+	// The fundmax flag is NOT allowed to be combiend with local_amt above.
```
sputn1ck, line 318: typo (combiend -> combined)

ziggie1984:
I was wondering why the author commit dates are not in a chronological order ?
```
Bjarne Magnussen: 6d3de89b3: 5 months ago
Bjarne Magnussen: 1a9f2807b: 5 months ago
Bjarne Magnussen: 3ac041fcc: 1 year, 1 month ago
Bjarne Magnussen: 827685603: 5 months ago
Bjarne Magnussen: 2a2d6ca9c: 1 year, 1 month ago
Bjarne Magnussen: e9a0c1478: 1 year, 1 month ago
Bjarne Magnussen: 87b957a8b: 1 year, 1 month ago
Bjarne Magnussen: bb4013440: 1 year, 1 month ago
```

ziggie1984:
Could we also allow the --fundmax option to consider unconfirmed outputs allowing effectively also to CPFP for the unconfirmed transaction into a new channel?

funding/manager_test.go
```diff
@@ -3280,6 +3281,113 @@ func TestFundingManagerFundAll(t *testing.T) {
...
+		}
+		alice, bob := setupFundingManagers(t, addFunds)
+		defer tearDownFundingManagers(t, alice, bob)
+
+		// We will consume the channel updates as we go, so no
+		// buffering is needed.
+		updateChan := make(chan *lnrpc.OpenStatusUpdate)
+
+		// Initiate a fund channel, and inspect the funding tx.
+		pushAmt := btcutil.Amount(0)
```
ziggie1984, line 3358: should we also test a pushamt greater than the MinChanFundingSize?

ziggie1984:
will this PR also allow for --fundmax with wumbo channels, meaning that the UpperLimit is adjusted when wumbo channels are active ? (As far as I understood the code, its capped at the 2^24-1 limit)

ziggie1984:
> will this PR also allow for --fundmax with wumbo channels, meaning that the UpperLimit is adjusted when wumbo channels are active ? (As far as I understood the code, its capped at the 2^24-1 limit)

Justed tested it with wumbo channels works nice 👍

lnwallet/chanfunding/coin_select.go
hsjoberg: Agree on variable name change. It's a bit difficult to understand what's going on, so naming it `overshoot` would make it more clear.

lntest/itest/lnd_channel-funding_fund_max_test.go
```diff
@@ -0,0 +1,324 @@
+// +build rpctest
+
+package itest
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/btcsuite/btcutil"
```
hsjoberg, line 10: Erroneous import:

```suggestion
	"github.com/btcsuite/btcd/btcutil"
```

Also I'm no go expert but other test files contain a `+build xxx` in the top:

```diff
+//go:build !rpctest
// +build !rpctest
```

hsjoberg (changes requested):

lnwallet/wallet.go
```diff
@@ -752,15 +759,34 @@ func (l *LightningWallet) handleFundingReserveRequest(req *InitFundingReserveMsg
...
 		fundingReq := &chanfunding.Request{
-			RemoteAmt:    req.RemoteFundingAmt,
-			LocalAmt:     req.LocalFundingAmt,
-			MinConfs:     req.MinConfs,
-			SubtractFees: req.SubtractFees,
-			FeeRate:      req.FundingFeePerKw,
+			RemoteAmt:      req.RemoteFundingAmt,
+			LocalAmt:       req.LocalFundingAmt,
+			FundUpToMaxAmt: req.FundUpToMaxAmt,
+			ReservedAmt:    ReservedValue(numAnchorChans),
```
hsjoberg, line 786: This will lead to inconsistent behavior.
For private channels (i.e mobile wallets), we do not right now reserve any sats for anchor outputs.
If you specify the amount when opening a private channel, you'll be able to use all funds (no sats will be reserved).
Related commit: https://github.com/lightningnetwork/lnd/commit/8e1087d1cdcbb21b495cffa74ac0afc6c2ffa533

Also related topic: https://github.com/lightningnetwork/lnd/issues/6574

Roasbeef:
@bjarnemagnussen hey this is on the milestone for our next release, do you have time to resume work on it so it can finally land?

hieblmi:
@saubyk @Roasbeef I see this is up for grabs and it looks like there's a lot to learn about the lnd code base so I'd like to take a look at this issue but I assume it'll take me some time to delve into it.

saubyk:
@hieblmi keen on making progress on this PR, so it would be great if you can pick it up and run with it. We are hoping this issue can make it in release 16. Happy to work with you to make progress on this. You may as well want to reach out to @bjarnemagnussen if he can provide any assistance on this.

lntest/itest/lnd_channel-funding_fund_max_test.go
hieblmi: Fixed in the continuation of this PR, here: https://github.com/lightningnetwork/lnd/pull/6903

lnwallet/chanfunding/coin_select.go
```diff
@@ -37,6 +37,16 @@ func (e *errUnsupportedInput) Error() string {
 	return fmt.Sprintf("unsupported address type: %x", e.PkScript)
 }
 
+// errReservedValueInvalidated is an error used internally for the coin select
+// algorithm to determine that
```
hieblmi, line 41: incomplete comment.

cmd/lncli/cmd_open_channel.go
hieblmi: Done here https://github.com/lightningnetwork/lnd/pull/6903/commits/b8e351e04a948152dbfd8cd965848c47619d2d0c and here
https://github.com/lightningnetwork/lnd/pull/6903/commits/7915d0dd52683ebecc291843ccc802079d333a48

cmd/lncli/cmd_open_channel.go
hieblmi: done.

funding/manager_test.go
hieblmi: done.

funding/manager_test.go
hieblmi: done.

lntest/itest/lnd_channel-funding_fund_max_test.go
hieblmi: Has been covered here: https://github.com/lightningnetwork/lnd/pull/6903

lntest/itest/lnd_channel-funding_fund_max_test.go
hieblmi: Was added by @bjarnemagnussen, to be reviewed here: https://github.com/lightningnetwork/lnd/pull/6903/commits/f85113904bed65eb650f5dd3c5de8445b1c59586#diff-583407cd98e4e423efd8a6afe4acb5b67395da6ae79927b11f9e3b989f822313R142

rpcserver.go
hieblmi: Tried to make it clearer here: https://github.com/lightningnetwork/lnd/pull/6903/commits/58bc9bdc6e9a0d335c92585665e02d8fad8c653b#diff-674c79ba02b6e9c45a994388392689814f3538800351b2fd43fb3ade21e1effbR1916

lnwallet/chanfunding/coin_select.go
hieblmi: Covered here: https://github.com/lightningnetwork/lnd/pull/6903/commits/60ddaa719a0e36f104304ffa9adb901764c9e314#diff-40ac2ca08026898bb3f9c383bb3ddf7c232b5d2a46c31c8673a80c17c87cf037R277

lnwallet/wallet.go
hieblmi: Addressed and described here with comments from the related commit: https://github.com/lightningnetwork/lnd/pull/6903/commits/60ddaa719a0e36f104304ffa9adb901764c9e314#diff-3ebebe355b386e69fdd9742a8d6ed60cb3b2377a9c10a3d3b831650e47c7d258R779

hieblmi:
> I have rebased this PR and added two commits to respect the reserve wallet balance in case of anchor channels.
> 
> One thing to consider is regarding the different commit types. Should the integration test also check the "lease" commit type, and should there be an integration test that fails whenever a new type is introduced that is not tested for to be sure the behaviour is always correct?


I've added a test in this commit: https://github.com/lightningnetwork/lnd/pull/6903/commits/1301bdef9616977bee1dfcd5727b3ea0549333ae

guggero:
It looks like all comments were addressed in the replacement PR #6903, so I'm going to close this one.
And I haven't heard any protest from @bjarnemagnussen against it being taken over, so I assume that's okay.
Thanks a lot for getting this started!

PR #4034, opened 2020-02-26
input/size: correct NP2WKH and NP2SH input count

cfromknecht:
This PR corrects a bug in TxWeightEstimator that could result in
underestimations for transactions involving NestedP2WPKH and NestedP2WSH
inputs. The scriptSig data push is now accounted for in a proper size
constant, and the input count is now incremented in both. This would
only be detectable in the event that the number of non-nested inputs and
the total number of inputs straddle the discontinuities in the
CompactSize encoding, e.g. 253, 2^16-1, or 2^32-1.

Fortunately this doesn’t impact any critical LN fee calculations (that could result
in invalid commitment signatures and a force close) since we don’t use these input
types.

### DIFF
--- a/input/size.go
+++ b/input/size.go
@@ -25,11 +25,21 @@ const (
 	//	- PublicKeyHASH160: 20 bytes
 	P2WPKHSize = 1 + 1 + 20
 
+	// NestedP2WPKHSize 23 bytes
+	//      - OP_DATA: 1 byte (P2WPKHSize)
+	//      - P2WPKHWitnessProgram: 22 bytes
+	NestedP2WPKHSize = 1 + P2WPKHSize
+
 	// P2WSHSize 34 bytes
 	//	- OP_0: 1 byte
 	//	- OP_DATA: 1 byte (WitnessScriptSHA256 length)
 	//	- WitnessScriptSHA256: 32 bytes
 	P2WSHSize = 1 + 1 + 32
+
+	// NestedP2WSHSize 35 bytes
+	//      - OP_DATA: 1 byte (P2WSHSize)
+	//      - P2WSHWitnessProgram: 35 bytes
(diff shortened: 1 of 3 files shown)


halseth:
Unit tests actually fail now, which means they are wrong also?

Roasbeef:
Yeah looks like they were coded around this bug.

lnwallet/size_test.go
```diff
@@ -68,41 +68,54 @@ func TestTxWeightEstimator(t *testing.T) {
...
 		numP2SHOutputs       int
 	}{
+		{},
 		{
-			numP2PKHInputs:  1,
-			numP2PKHOutputs: 2,
+			numP2PKHInputs: 1,
 		},
 		{
-			numP2PKHInputs:  1,
```
halseth: Looks like combinations of inputs/outputs are not tested anymore? I think that is valuable to keep to some degree, as it ensures the weight estimator handles it.

cfromknecht: readded tests and also expanded the test suite a good bit

input/size.go
```diff
@@ -25,12 +25,22 @@ const (
...
+	//      - P2WPKHWitnessProgram: 22 bytes
+	NestedP2WPKHSize = 1 + P2WPKHSize
+
 	// P2WSHSize 34 bytes
 	//	- OP_0: 1 byte
 	//	- OP_DATA: 1 byte (WitnessScriptSHA256 length)
 	//	- WitnessScriptSHA256: 32 bytes
 	P2WSHSize = 1 + 1 + 32
 
+	// NestedP2WSHSize 35 bytes
```
Roasbeef: 36 bytes*

cfromknecht: ahh good catch, fixed in https://github.com/lightningnetwork/lnd/pull/4062

PR #4040, opened 2020-03-02
lnrpc+rpcserver: expose short chan id on feereport

cfromknecht:
This commit adds each channel's short chan id to the `feereport` rpc.
Without this, it can be tedious to lookup more info about a particular
channel since most rpcs only accept short chan ids and not channel
points. For instance, now one can take a channel id from `feereport` and
look it up directly via `getchaninfo` to examine the policy in more detail.


lnrpc/rpc.proto
```diff
@@ -2895,6 +2895,9 @@ message Feature {
 
 message FeeReportRequest {}
 message ChannelFeeReport {
+    /// The short channel id that this fee report belongs to.
```
carlaKC: Is this at the top of the message for the sake of lncli? If so, rather leave the proto in order and add an ad-hoc struct which prints the order we want in lncli?

joostjager:
Two reviewers already, unassigning myself

lnrpc/rpc.proto
cfromknecht: It’s not only for lncli, but also rest. In 0.9 we went through and removed a bunch of structs that were purely for the purpose of formatting so I’m hesitant to start down that road again

cfromknecht: fwiw the rationale for doing so is that `jsonpb` now automatically converts byte slices to hex, but it only operates on `proto.Message` objects. if we had custom formatting they'd either have to implement this interface or add back this custom serialization, but all of this is avoided by just reordering the protos.

carlaKC: Ah I see, good to know!

PR #3112, opened 2019-05-23
[WIP]: test 2

cfromknecht:

PR #4042, opened 2020-03-02
travis: format, compile and verify protos

### LINKED ISSUE #3094: [feature request] Verify Compiled Protos on Travis
As pointed out by @xsb in #2312, different authors may have different proto versions installed when doing development. To ensure all PRs use the correct of protoc and plugins, we should move to have Travis verify that the output is correct, and fail the build otherwise.

## Steps to Completion
1. Have travis build and install the protobuf dependencies specified [here](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/README.md#generate-protobuf-definitions).
2. Cache the protobuf and plugin source directories in the travis build cache, so that it doesn't need to be recompiled from scratch on every build.
3. Add a new make target to `Makefile`, e.g. `rpc-check`, that first runs the `rpc` target, and then asserts that the source tree is clean. This can be done by using `$(shell git describe --dirty)` and asserting that the output doesn't contain the string `"dirty"`
4. Add the `rpc-check` target as a dependency to `travis-race`, `travis-itest`, and `travis-cover`, preferably after `lint` so that the more expensive operations are done only if the protos are sound.

guggero:
Replaces/continues #3112/#3110 by @cfromknecht.

This PR installs/compiles the `protoc`, `protobuf`, `genproto` and `grpc-gateway` libraries/binaries in the exact version as specified in `lnrpc/README.md` on Travis and then compiles all `.proto` files.
If the git worktree becomes dirty in the process, that means the protos weren't compiled correctly or with a wrong version and the CI build is failed.

The build process also makes sure the protos are properly formatted, using `clang-format` with custom rules (`lnrpc/.clang-format`) to do so.

Fixes #3094

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,14 @@ rpc:
 	@$(call print, "Compiling protos.")
 	cd ./lnrpc; ./gen_protos.sh
 
+rpc-format:
+	@$(call print, "Formatting protos.")
+	cd ./lnrpc; find . -name "*.proto" | xargs clang-format --style=file -i
+
+rpc-check: rpc-format rpc
+	@$(call print, "Verifying protos.")
+	if test -n "$$(git describe --dirty | grep dirty)"; then echo "Protos not properly formatted or not compiled with v3.4.0"; git status; git diff; exit 1; fi
+
 mobile-rpc:
 	@$(call print, "Creating mobile RPC from protos.")
 	cd ./mobile; ./gen_bindings.sh
@@ -223,6 +231,8 @@ clean:
 	lint \
 	list \
 	rpc \
+	rpc-format \
+	rpc-check \
 	mobile-rpc \
 	vendor \
 	ios \
--- a/lnrpc/.clang-format
+++ b/lnrpc/.clang-format
@@ -1,0 +1,7 @@
+---
+Language: Proto
+BasedOnStyle: Google
+IndentWidth: 4
+AllowShortFunctionsOnASingleLine: None
+SpaceBeforeParens: Always
+CompactNamespaces: false
--- a/lnrpc/README.md
+++ b/lnrpc/README.md
@@ -146,6 +146,17 @@ $ go get -u github.com/lightningnetwork/lnd/lnrpc
 
 ## Generate protobuf definitions
 
+### Linux
+
+For linux there is an easy install script that is also used for the Travis CI
+build. Just run the following command (requires `sudo` permissions and the tools
+`make`, `go`, `wget` and `unzip` to be installed) from the repository's root
+folder:
+
+`./scripts/install_travis_proto.sh`
+
+### MacOS / Unix like systems
+
 1. Download [v.3.4.0](https://github.com/google/protobuf/releases/tag/v3.4.0) of
 `protoc` for your operating system and add it to your `PATH`.
 For example, if using macOS:
@@ -178,4 +189,24 @@ $ git reset --hard v1.8.6
 $ go install ./protoc-gen-grpc-gateway ./protoc-gen-swagger
 ```
 
-5. Run [`gen_protos.sh`](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/gen_protos.sh) to generate new protobuf definitions.
+5. Run [`gen_protos.sh`](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/gen_protos.sh)
+or `make rpc` to generate new protobuf definitions.
+
+## Format .proto files
+
+We use `clang-format` to make sure the `.proto` files are formatted correctly.
+You can install the formatter on Ubuntu by running `apt install clang-format`.
+
+Consult [this page](http://releases.llvm.org/download.html) to find binaries
+for other operating systems or distributions.
+
+## Makefile commands
+
+The following commands are available with `make`:
+
+* `rpc`: Compile `.proto` files (calls `lnrpc/gen_protos.sh`).
+* `rpc-format`: Formats all `.proto` files according to our formatting rules.
+  Requires `clang-format`, see previous chapter.
+* `rpc-check`: Runs both previous commands and makes sure the git work tree is
+  not dirty. This can be used to check that the `.proto` files are formatted
+  and compiled properly.
--- a/lnrpc/routerrpc/router.proto
+++ b/lnrpc/routerrpc/router.proto
@@ -19,7 +19,7 @@ message SendPaymentRequest {
 
     /**
     Number of millisatoshis to send.
-    
+
     The fields amt and amt_msat are mutually exclusive.
     */
     int64 amt_msat = 12;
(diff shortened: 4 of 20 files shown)


lnrpc/routerrpc/router.proto
```diff
@@ -60,15 +60,15 @@ message SendPaymentRequest {
...
-     /**
-    The maximum number of millisatoshis that will be paid as a fee of the
-    payment. If this field is left to the default value of 0, only zero-fee
-    routes will be considered. This usually means single hop routes connecting
-    directly to the destination. To send the payment without a fee limit, use
-    max int here.
-    
-    The fields fee_limit_sat and fee_limit_msat are mutually exclusive.
-    */
+    /**
```
bhandras: nit: looks like the formatter failed to catch the extra space here.

lnrpc/watchtowerrpc/watchtower.proto
```diff
@@ -5,24 +5,24 @@ package watchtowerrpc;
...
 service Watchtower {
-        /** lncli: tower info
-        GetInfo returns general information concerning the companion watchtower
-        including it's public key and URIs where the server is currently
-        listening for clients.
-        */
-        rpc GetInfo(GetInfoRequest) returns (GetInfoResponse);
+    /** lncli: tower info
+    GetInfo returns general information concerning the companion watchtower
+    including it's public key and URIs where the server is currently
```
bhandras: nit: ~~it's~~ its

lnrpc/rpc.proto
```diff
@@ -819,7 +819,7 @@ service Lightning {
 
 message Utxo {
     /// The type of address
-    AddressType type = 1;
+    AddressType address_type = 1;
```
bhandras: note: usually it is forbidden to change name for existing fields. We may choose to do this if our users use submitted generated code and json, otherwise I'd reconsider.

Makefile
```diff
@@ -174,6 +174,14 @@ rpc:
 	@$(call print, "Compiling protos.")
 	cd ./lnrpc; ./gen_protos.sh
 
+rpc-format:
+	@$(call print, "Formatting protos.")
+	cd ./lnrpc; clang-format -i *.proto **/*.proto
+
+rpc-check: rpc-format rpc
+	@$(call print, "Verifying protos.")
+	if test -n "$$(git describe --dirty | grep dirty)"; then echo "Protos not properly formatted or not compiled with v3.4.0"; exit 1; fi
```
bhandras: Nice way to ensure formatting :slightly_smiling_face:

lnrpc/.clang-format
```diff
@@ -0,0 +1,7 @@
+---
+Language: Proto
+BasedOnStyle: Google
+IndentWidth: 4
+AllowShortFunctionsOnASingleLine: None
```
bhandras: personal opinion: I think single line short functions are nice and readable. I'm fine with your choice though.

scripts/install_travis_proto.sh
```diff
@@ -0,0 +1,70 @@
...
+    echo "Using cached version of protoc"
+  else
+    wget -O /tmp/protoc.zip $PROTOC_URL
+    mkdir -p "${PROTOC_DL_CACHE_DIR}"
+    unzip -o /tmp/protoc.zip -d "${PROTOC_DL_CACHE_DIR}"
+    chmod +x "${PROTOC_DL_CACHE_DIR}/bin/protoc"
+  fi
+  sudo cp "${PROTOC_DL_CACHE_DIR}/bin/protoc" /usr/local/bin
+  sudo cp -r "${PROTOC_DL_CACHE_DIR}/include" /usr/local
+  sudo chmod +r /usr/local/include -R
```
bhandras: Wouldn't a simple `cp -a` be enough`?

scripts/install_travis_proto.sh
```diff
@@ -0,0 +1,70 @@
...
+  fi
+  pushd $GOPATH/src/google.golang.org/genproto
+  git reset --hard $GENPROTO_VERSION
+  popd
+}
+
+# install_grpc_gateway downloads and installs the gRPC gateway that converts
+# .proto files into REST gateway code.
+install_grpc_gateway() {
+  if [ ! -d $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway ]; then
```
bhandras: nit: for this and the above, you could move the paths to local variables.

lnrpc/README.md
```diff
@@ -146,6 +146,16 @@ $ go get -u github.com/lightningnetwork/lnd/lnrpc
 
 ## Generate protobuf definitions
 
+### Linux
+
+For linux there is an easy install script that is also used for the Travis CI
```
bhandras: Note that the script only works for Debian/Ubuntu and does not check if there's a preinstalled toolset. For those users (and in general) a hermetic, reproducible build system like Bazel may be a better choice.

lnrpc/routerrpc/router.proto
guggero: That commit is from the dependent PR where the formatter was not yet introduced. I noticed this and fixed it in the previous PR because the formatter found it.

bhandras: Sorry missed the parent PR.

lnrpc/rpc.proto
guggero: Yes, we are aware of that. See the discussion in #3996.

Makefile
guggero: Yes, I like it too! It's @cfromknecht's idea, btw.

lnrpc/.clang-format
guggero: This mainly affects how empty messages are displayed (closing bracket on new line or same line). I wanted to imitate my IDE's formatting as close as possible to not have a huge diff to #3996. But I think it's indeed nicer to have it on one line.

scripts/install_travis_proto.sh
guggero: No, I think this was required because the copy was done with `sudo` and in the original zip the "anyone can read" permission was missing. So only root could have read it.
But I fixed it in a slightly different way now.

scripts/install_travis_proto.sh
guggero: Good idea, thanks.

lnrpc/README.md
guggero: Yes, that's true. But quite a few users have struggled with this doing it manually. So having a script already helps a lot in my opinion. I added a description what tools need to be installed to make it more clear that this is not an all-in-one solution.

guggero:
Thanks for the review! I think this is ready to look at, I just forgot to remove the "[WIP]" prefix from the title. Please be aware that only the last 4 commits belong to this PR.

lnrpc/README.md
bhandras: Sorry, didn't articulate really what I meant. My main point is that as this script applies to Debian/Ubuntu only, I'd specifically say that instead of referring to all "linux" users.

scripts/install_travis_proto.sh
bhandras: :+1:

lnrpc/README.md
guggero: Which part is Ubuntu specific? The `apt-get` stuff is only in `.travis.yml`. The bash script should work on most systems.

bhandras: Yeah, sorry. Somehow I remembered that part to be in the `installl_travis_proto.sh` but just checked and it's indeed in `travis.yml`.

Makefile
```diff
@@ -109,7 +109,7 @@ scratch: build
 # TESTING
 # =======
 
-check: unit itest
+check: rpc-check unit itest
```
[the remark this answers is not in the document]
guggero: Haha, no, it did in fact not run on Travis yet. I assumed we ran `check`. But with a few additional tweaks it now seems to work :tada:

guggero:
Rebased.

cfromknecht:
@guggero build is failing

guggero:
Haha, nice. Real-life test passed :joy: 
Fixed the build.

PR #4045, opened 2020-03-03
cmd/lncli: add --with flag to list channels

cfromknecht:
This commit adds a flag to `listchannels` that filters by remote pubkey.


cmd/lncli/commands.go
```diff
@@ -1990,6 +1990,12 @@ var listChannelsCommand = cli.Command{
 			Name:  "private_only",
 			Usage: "only list channels which are currently private",
 		},
+		cli.StringFlag{
+			Name: "with",
```
bhandras: `with_peer` may be a better name, wdyt?

cfromknecht: yeah i considered a more descriptive name but then thought it sounded nice when read like a sentence. i don't have a strong opinion either way, i'll let others chime in with their preference and then go with herd 👍

bhandras: sgtm

cfromknecht:
we don't typically add tests for lncli since it's predominately a debugging tool. I have tested against my node though:
 - `list channels --with`

cmd/lncli/commands.go
joostjager: I would do `--peer`. The "with-" prefix isn't what we usually do.

cmd/lncli/commands.go
```diff
@@ -1999,6 +2005,20 @@ func listChannels(ctx *cli.Context) error {
...
+	// Assert the with filter is empty or 66 bytes in length.
+	withPeer := ctx.String("with")
+	if len(withPeer) > 0 && len(withPeer) != 66 {
+		return fmt.Errorf("--with pubkey length of %d is invalid, "+
+			"must be 0 or 66 bytes", len(withPeer))
+	}
+
+	// Assert the with filter is hex-encoded, the empty string trivially
+	// passes.
+	_, err := hex.DecodeString(withPeer)
```
joostjager: `route.NewVertexFromStr` ?

cfromknecht:
> Why the decision to do this client side, while other filtering is taking place server side?

@joostjager didn't seem performance critical, plus rpc users can do this fairly easily. regardless, changed it to filter internally, ptal.

cmd/lncli/commands.go
cfromknecht: done

lnrpc/rpc.proto
```diff
@@ -1385,6 +1385,12 @@ message ListChannelsRequest {
     bool inactive_only = 2;
     bool public_only = 3;
     bool private_only = 4;
+
+    /**
+    Filters the response for channels with a target peer's pubkey. If with_peer
```
joostjager: s/with_peer/peer

rpcserver.go
```diff
@@ -2948,6 +2948,11 @@ func (r *rpcServer) ListChannels(ctx context.Context,
 			"`private_only` can be set, but not both")
 	}
 
+	if len(in.Peer) > 0 && len(in.Peer) != 33 {
```
joostjager: Could use `route.NewVertexFromBytes`

PR #4051, opened 2020-03-05
peers: Track errors across connections

### LINKED ISSUE #3000: peer: cache and display the last error(s) sent by a peer in listpeers
Whenever a peer encounters an error, the [`Error` message](https://github.com/lightningnetwork/lnd/blob/master/lnwire/error.go#L64) is sent across. Today we lack any sort of error type, but in the future once we gain an error type, it may be a valuable source of information to diagnose cross implementation issues. We should start to show the last few errors sent/recv'd via the `ListPeers` RPC call so we can collect this data preemptively. 

# Steps To Completion 

- [ ] Store the last N errors sent/received to/from the peer within [the main peer struct](https://github.com/lightningnetwork/lnd/blob/master/peer.go#L106). 

- [ ] Extend the [`ListPeers` RPC call to also](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/rpc.proto#L1283) expose this new data. 

- [ ] Update the RPC server implementation to locate and populate this new error information.

carlaKC:
This PR adds tracking of peer errors across peer connections. 
- The latest 10 errors are cached in a circular array in the peer struct (10 is a magic number, seemed reasonable). 
- On disconnection, the peer hands off this cache to the server
- On reconnection, the server creates the peer struct with their historic error cache
- The latest error for each peer is displayed in `ListPeers`; showing the whole list felt like overkill (?) but could do so (since we'd need to add an extra endpoint to expose them otherwise)

Fixes #3000


lnrpc/rpc.proto
```diff
@@ -1515,6 +1515,21 @@ message Peer {
 
     /// Features advertised by the remote peer in their init message.
     map<uint32, Feature> features = 11 [json_name = "features"];
+
+    /*
+    Latest error is the last error we received from the peer. This error is
+    tracked across peer connections, but is not currently persisted across lnd
+    restarts.
+    */
+    TimestampError latest_error = 12;
```
bhandras: I'd probably call this `TimestampedError`. Somehow `TimestampError` to me reads like there as an error with the timestamp :)

lnrpc/rpc.proto
```diff
@@ -1515,6 +1515,21 @@ message Peer {
...
+}
+
+message TimestampError{
+    // The unix timestamp in seconds when the error occurred.
```
bhandras: nit: strictly speaking, a unix timestamp is always in seconds.

lnrpc/rpc.swagger.json
```diff
@@ -3646,6 +3646,10 @@
             "$ref": "#/definitions/lnrpcFeature"
           },
           "description": "/ Features advertised by the remote peer in their init message."
+        },
+        "latest_error": {
+          "$ref": "#/definitions/lnrpcTimestampError",
+          "description": "Latest error is the last error we received from the peer. This error is\ntracked across peer connections, but is not currently persisted across lnd\nrestarts."
```
bhandras: nit: Instead of "latest error is the last error..", just "The latest error with timestamp that we received from the peer".

lnrpc/rpc.swagger.json
```diff
@@ -4097,6 +4101,20 @@
     "lnrpcStopResponse": {
       "type": "object"
     },
+    "lnrpcTimestampError": {
+      "type": "object",
+      "properties": {
+        "timestamp": {
+          "type": "string",
+          "format": "uint64",
+          "description": "The unix timestamp in seconds when the error occurred."
```
bhandras: nit: unix timestamp is in seconds by def

lnrpc/rpc.swagger.json
```diff
@@ -4097,6 +4101,20 @@
...
+        },
+        "error": {
+          "type": "string",
+          "description": "The error string of the error our peer sent us."
```
bhandras: Maybe something like this: "String representation/value of the error the peer sent us"

cache/cache.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return &Circular{
+		index: noIndex,
+		total: 0,
+		size:  size,
+
+		// Create a slice with length and capacity equal to the size of
+		// the cache so that we do not need to resize the underlying
+		// array when we add items.
+		items: make([]interface{}, size),
```
bhandras: Instead of taking up a fixed size, this could grow as well. Not a big thing as for now we only use this with a handful of peers.

roeierez:
> * The latest error for each peer is displayed in `ListPeers`; showing the whole list felt like overkill (?) but could do so (since we'd need to add an extra endpoint to expose them otherwise)

Suggestion: A boolean parameter can be added to the RPC `show-all-errors` that will add the errors history only if true.
It will eliminate the need to add another RPC.

lnrpc/rpc.proto
guggero: I agree with the naming suggestion. I also think it would make sense to make this field `repeated` and send the whole list of errors if we go through all the trouble of caching them.

lnrpc/rpc.proto
guggero: We do sometimes serialize timestamps as `UnixNano()` in `lnd` (I think) that's why usually the comment mentions seconds explicitly.

cache/cache.go
```diff
@@ -0,0 +1,130 @@
...
+	// If no items have been cached yet, we can just return a nil list.
+	case c.index == noIndex:
+		return nil
+
+	// If we have added fewer items than the cache size, we can simply
+	// return the total number of items from the beginning of the list
+	// to the total. This special case is added because the oldest item
+	// is at the beginning of the underlying array, not after the index,
+	// when we have not filled the array yet.
+	case c.total < c.size:
```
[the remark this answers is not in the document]
carlaKC: Not sure I'm following this comment? 
The logic after this doesn't work for an array that isn't full yet, it'll return things in the wrong order. Theis covers the special case where the oldest item is at index 0, rather than being after the index.

peer.go
```diff
@@ -91,6 +95,13 @@ type channelCloseUpdate struct {
 	Success     bool
 }
 
+// timestampError is a timestamped error that is used to cache the most recent
+// errors sent from our peers.
+type timestampError struct {
```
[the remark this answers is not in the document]
carlaKC: I think I'll pass for this case since this isn't really an error that we're propagating up through a call stack. Would just need to add an `Error` and `Unwrap` function to this struct, and I think it server it purpose as is?

cache/cache.go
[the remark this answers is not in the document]
bhandras: Sorry, misunderstood something and made an incorrect comment. All is fine here.

cache/cache.go
```diff
@@ -0,0 +1,130 @@
...
+// is full.
+func (c *Circular) Add(item interface{}) {
+	c.Lock()
+	defer c.Unlock()
+
+	// Increment the index and wrap by size if necessary.
+	c.index = (c.index + 1) % c.size
+
+	// Set the item in the next free index in the items array.
+	c.items[c.index] = item
```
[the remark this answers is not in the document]
carlaKC: nope, can cut it :)

cache/cache.go
```diff
@@ -0,0 +1,130 @@
...
+// noIndex is the value of the cache's index when no values have been added.
+const noIndex = -1
+
+// errInvalidSize is returned when an invalid size for a cache is provided.
+var errInvalidSize = errors.New("cache size must be > 0")
+
+// Circular is a cache which retains a set of values in memory, and
+// overwrites the oldest item in the cache when a new item needs to be
+// written.
+type Circular struct {
```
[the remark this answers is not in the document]
carlaKC: Renamed to buffer :) 

> Maybe a better following this reasoning is the queue package.

Do you mean move this into the queue package?

cache/cache.go
carlaKC: > make([]interface{}, 0, size)

Thought of going this route but then you end up having custom logic for the first set of elements you add - need to append for the first `size` x elements (so that slice len is increased) and then store at index after that. Since we don't really need `len` for much, I thought it would be better to have more simple add logic?

server.go
```diff
@@ -2791,6 +2812,10 @@ func (s *server) peerConnected(conn net.Conn, connReq *connmgr.ConnReq,
 
 	s.addPeer(p)
 
+	// Once we have successfully added the peer to the server, we can
+	// delete the previous error cache from the server.
+	delete(s.peerErrors, pkStr)
```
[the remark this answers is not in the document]
carlaKC: Would it? Here, once the peer has connected, we copy the error buffer across to the peer struct and delete it from the server (no need to maintain 2 copies). If the peer disconnects, we copy it back to the server in `removePeer`.

lnrpc/rpc.proto
carlaKC: Got a comment on another PR to explicitly indicate that it's in seconds, so now I just add it by default :)

carlaKC:
> A boolean parameter can be added to the RPC include-all-errors that will add the errors history only if true.

@roeierez nice idea! Added this :) 

> Before adding those tho I think we need to decide whether this is meant to be for general debugging, or whether it should be reserved for in-protocol errors. imo i think the debugging tool is more useful so we can help node operators surface actionable errors.

I think debugging is also the primary use case here. Rather have more errors tracked and then potentially decrease what we track/ make cache size configurable etc. 

For init, I'd say track invalid features and DLP required, pretty straightforward.

Disconnect is an interesting one, because I almost think you'd want to have the last disconnection reason around permanently, not have it in a buffer which may get over written after some number of errors. Maybe overkill, but it does seem to be a big debugging issue that people bring up so could be worth adding a dedicated last disconnection error field?

peer.go
```diff
@@ -1174,13 +1215,39 @@ func (p *peer) isActiveChannel(chanID lnwire.ChannelID) bool {
 	return ok
 }
 
+// storeError stores an error in our peer's buffer of recent errors with the
+// current timestamp. Errors are only stored if we have at least one active
+// channel with the peer to mitigate dos attack vectors where a peer costlessly
+// connects to us and spams us with errors.
+func (p *peer) storeError(err error) {
+	p.activeChanMtx.RLock()
+	channelCount := len(p.activeChannels)
```
Crypt-iQ: The current `peer.go` is a little wonky in that pending channels are also in `activeChannels` on startup due to `loadActiveChannels`. I have a slight refactor of this so that pending channels have `nil` in the `activeChannels` map and open channels actually have the channel data.  This could wait until that lands, or you could loop through all channels and check if one isn't pending.

carlaKC: Link to PR? If that change is far off I'll update this, otherwise will wait for it to land.

rpcserver.go
```diff
@@ -2459,6 +2459,33 @@ func (r *rpcServer) ListPeers(ctx context.Context,
...
+			if latestErr != nil {
+				peerErrors = []interface{}{latestErr}
+			}
+		} else {
+			peerErrors = serverPeer.errorBuffer.List()
+		}
+
+		// Add the relevant peer errors to our response.
+		for _, error := range peerErrors {
+			tsError := error.(*timestampedError)
```
guggero: To avoid a panic _if_ there ever is something other in there, we could just skip the item if the cast is not ok.

queue/circular_buf.go
```diff
@@ -0,0 +1,116 @@
...
+	// Copy the second set of times into the response slice offset by the
+	// length of the first set of items so that we return a response which
+	// is ordered from oldest to newest entry.
+	copy(resp[len(firstHalf):], secondHalf)
+
+	return resp
+}
+
+// Total returns the total number of items that have been added to the buffer.
+func (c *CircularBuffer) Total() int {
```
bhandras: Wouldn't it be more useful if this method would return the number of items currently in the buffer instead of total items added?

queue/circular_buf.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+)
+
+// errInvalidSize is returned when an invalid size for a buffer is provided.
+var errInvalidSize = errors.New("buffer size must be > 0")
+
+// CircularBuffer is a buffer which retains a set of values in memory, and
+// overwrites the oldest item in the buffer when a new item needs to be
+// written.
+type CircularBuffer struct {
```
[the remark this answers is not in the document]
bhandras: Agreed, much cleaner now! :)

peer.go
Crypt-iQ: It will get into 0.10 but I don't see a reason to block this PR, I can just make the change on my end

queue/circular_buf.go
carlaKC: Kinda useless after you hit the total size, since it'll just always return the size. If we end up needing that, I'll add a `Size()` func so you can figure out how full it is.

rpcserver.go
carlaKC: If we do make a change which adds something else, I'm pretty confident that the itests will panic. Should be picked up developer end when testing changes anyway, so I'd rather fail loudly?

peer.go
carlaKC: Cool, thanks! Tag me in the PR when it's up so I can specifically test this?

carlaKC:
Rebased and updated some nits, going to wait on the change from @Crypt-iQ before merge, just so I can do one last set of tests.

peer.go
Crypt-iQ: Will do

Roasbeef:
Needs rebase. @Crypt-iQ's PR referenced hasn't yet entered the review cycle yet, so I think this can land, with that PR being updating the referenced area as needed.

rpcserver.go
guggero: Yes, failing loudly also sounds good to me. It _should_ never happen but a nice error is always nicer than just a panic.

PR #4055, opened 2020-03-06
peer: remove unused failedChannels channels map

cfromknecht:
This commit removes the `failedChannels` map from the peer.
Currently it is only read on peer startup before anything is entered
into the map.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -210,14 +210,6 @@ type peer struct {
 	// the connection handshake.
 	remoteFeatures *lnwire.FeatureVector
 
-	// failedChannels is a set that tracks channels we consider `failed`.
-	// This is a temporary measure until we have implemented real failure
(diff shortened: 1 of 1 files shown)


peer.go
```diff
@@ -1205,10 +1188,6 @@ func (p *peer) handleError(msg *lnwire.Error) bool {
 	case msg.ChanID == lnwire.ConnectionWideID:
 		for chanID, chanStream := range p.activeMsgStreams {
```
bhandras: Seems like this `chanID` won't be used anymore.

cfromknecht: good catch, fixed!

cfromknecht:
> This is also done in #2018 btw

I was looking for that! Didn't realize it was that old lol

PR #4056, opened 2020-03-07
tor+server: add OnionStore to AddOnionConfig with file-based implementation

wpaulino:
In this PR, we modify the `AddOnionConfig` struct to include an abstract `OnionStore`, which will be responsible for storing all relevant information of an onion service. We also add a file-based implementation of the interface to maintain the same behavior of storing an onion service's private key in a file.


tor/add_onion.go
```diff
@@ -53,24 +116,31 @@ func (c *Controller) AddOnion(cfg AddOnionConfig) (*OnionAddr, error) {
...
-		privateKey, err := ioutil.ReadFile(cfg.PrivateKeyPath)
-		if err != nil {
+	switch cfg.Type {
+	case V2:
+		keyParam = "NEW:RSA1024"
+	case V3:
+		keyParam = "NEW:ED25519-V3"
+	}
+
+	if cfg.Store != nil {
```
Roasbeef: Is this check necessary? What about failing during construction if a store isn't provided. Line 202 seems to assume that a store will always be present if we were able to successfully send the command to the tor daemon.

wpaulino: The intention is to make it optional for users of the library that aren't interested in recovering their onion services after shutdown. Line 202 also checks that a store is provided before writing it.

PR #4062, opened 2020-03-10
input/size: correct NestedP2WSHSize comment

cfromknecht:


input/size.go
```diff
@@ -36,7 +36,7 @@ const (
 	//	- WitnessScriptSHA256: 32 bytes
 	P2WSHSize = 1 + 1 + 32
 
-	// NestedP2WSHSize 35 bytes
+	// NestedP2WSHSize 36 bytes
```
joostjager: But now it doesn't add up? `P2WSHSize` is 34, plus one is 35?

cfromknecht: you're right, fixed

Issue #899, opened 2018-03-20
Wallet password should be optional


petertodd:
If the attacker has a copy of your wallet, in most sane threat models they probably also backdoor lnd to capture the password later anyway, rendering it pointless.

I set mine to 12345678, but really, an empty password should be accepted.

Roasbeef:
If you don't want to use the wallet password, there's the `--noencryptwallet` setting. It was put in for usage within integration tests (and will likely soon be gated behind a build flag). It just uses a canned password in this setting. 

As you note, it depends on your threat model though. If the attacker doesn't have the power to backdoor their running `lnd` instance, and they obtained a copy of the database from like a compromised Dropbox account or something, then the password is a level of defense.

robtex:
if you use `--noencryptwallet` you will not get any cipher seed though?

Roasbeef:
That's correct.

petertodd:
@Roasbeef But why would I have the database on a Dropbox account? Lightning doesn't really do backups at this point. Equally, even in a normal wallet, I'd just use GPG for that encryption.

chpio:
here's a workaround: https://github.com/Stadicus/RaspiBolt/blob/master/raspibolt_6A_auto-unlock.md

gaillard:
How to automate the current unlock with the rest api without relying on timing? The unlockwallet endpoint seems to return many things in success cases such as `{"error":"context canceled","code":1}` a 408 to `{"error":"transport is closing","code":14}` a 503. I have yet to see a 2xx. But usually any response besides connection refused or a 404 does unlock the wallet. The final resting case for that endpoint after unlocking seems to be 404, but the problem is you can get that before it issues the prompt to unlock too. So putting an infinite loop in that hits the endpoint until 404 isn't reliable. Using expect as in #1611 is probably better but I think all this illustrates that a simple no password option would be best.

gaillard:
btw this expect script and systemd unit file I believe are at least a reliable way to automate it even without an optional password. For example on a raspberry pi of mine:

/home/pi/lnd/lnd_expect.sh
```
#!/usr/bin/expect
set lndDist /home/pi/lnd
spawn $lndDist/lnd
set lndID $spawn_id
expect {
	-timeout 300
	"Waiting for wallet encryption password" {}
	default { send_user "default hit 1\n"; exit 1 }	
}
spawn $lndDist/lncli unlock
set cliID $spawn_id
expect {
	-i $cliID
	-timeout 30
[...]

PR #4066, opened 2020-03-10
lncli: allow unlock with password from stdin

guggero:
As discussed with @lispmeister, this is needed to unlock lnd in a containerized environment like Kubernetes. Without this fix, additional unlocking software is needed that performs the unlock call through gRPC.

This could perhaps also be an acceptable compromise to what was suggested in #899.


githorray:
Trouble building on Windows. go version go1.14 windows/amd64

```
cmd\lncli\commands.go:1810:34: cannot use syscall.Stdin (type syscall.Handle) as
 type int in argument to terminal.ReadPassword
Makefile:91: recipe for target 'build' failed
make: *** [build] Error 2
```

githorray:
```		pw, err = terminal.ReadPassword(int(syscall.Stdin))```

seems to have fixed it.

guggero:
Thanks for reporting this, @githorray! Fix is up: #4069.

githorray:
working great! thanks.

